PR #421, opened 2021-08-05
zcash_primitives: Benchmark trial decryption of compact outputs

str4d:
These are effectively identical to full-output trial decryption (as the
primary cost is the scalar multiplication), but it's good to check.

This PR also now enables flamegraph profiling of the `zcash_primitives` benchmarks.

### DIFF
--- a/zcash_primitives/benches/note_decryption.rs
+++ b/zcash_primitives/benches/note_decryption.rs
@@ -95,6 +95,6 @@ criterion_group! {
     config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
     targets = bench_note_decryption
 }
-#[cfg(windows)]
+#[cfg(not(unix))]
 criterion_group!(benches, bench_note_decryption);
 criterion_main!(benches);
--- a/zcash_primitives/benches/note_decryption.rs
+++ b/zcash_primitives/benches/note_decryption.rs
@@ -18,6 +18,9 @@ use zcash_primitives::{
     },
 };
 
+#[cfg(unix)]
+use pprof::criterion::{Output, PProfProfiler};
+
 fn bench_note_decryption(c: &mut Criterion) {
     let mut rng = OsRng;
     let height = TEST_NETWORK.activation_height(Canopy).unwrap();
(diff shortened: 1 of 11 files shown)


daira (approved):
utACK

zcash_primitives/benches/note_decryption.rs
```diff
@@ -66,7 +72,29 @@ fn bench_note_decryption(c: &mut Criterion) {
...
+    });
 }
 
+#[cfg(unix)]
+criterion_group! {
+    name = benches;
+    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    targets = bench_note_decryption
+}
+#[cfg(windows)]
```
daira, line 98:
```suggestion
#[cfg(not(unix))]
```

zcash_primitives/benches/pedersen_hash.rs
```diff
@@ -2,17 +2,27 @@ use criterion::{criterion_group, criterion_main, Criterion};
...
     });
 }
 
+#[cfg(unix)]
+criterion_group! {
+    name = benches;
+    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    targets = bench_pedersen_hash
+}
+#[cfg(windows)]
```
daira, line 26:
```suggestion
#[cfg(not(unix))]
```

str4d:
Looking at the flamegraphs, I noticed that the `compact-invalid` case wasn't doing note commitment computations. It turns out that this is because for an invalid `ivk` we decrypt to random noise, and in 255/256 cases we get the wrong version byte. This reality is reflected across multiple runs of the benchmark, but not within the benchmark itself (which uses a single fixed output and invalid `ivk`). We should alter these compact benchmarks to have a large number of `ivk`s we trial against, to get the average case reflected in their benchmarks; I'll do this in a follow-up PR.

str4d:
Update on the above: we don't actually see the above behavior for Sapling in reality, because during trial decryption we reject notes with values greater than `MAX_MONEY`, which for random nose has probability 2^13. We do see this behaviour for Orchard, however.

PR #423, opened 2021-08-06
Store OutputDescription `ephemeral_key` as bytes

str4d:
This removes an unnecessary `to_bytes` during trial decryption of notes,
and more closely matches the protocol spec. We retain the consensus rule
canonicity check on epk due to `SaplingVerificationContext::check_output`
taking a `jubjub::ExtendedPoint`, forcing `zcashd` to parse the bytes.

### DIFF
--- a/zcash_client_backend/src/wallet.rs
+++ b/zcash_client_backend/src/wallet.rs
@@ -3,6 +3,7 @@
 
 use subtle::{Choice, ConditionallySelectable};
 
+use zcash_note_encryption::EphemeralKeyBytes;
 use zcash_primitives::{
     merkle_tree::IncrementalWitness,
     sapling::{
@@ -54,7 +55,7 @@ pub struct WalletShieldedSpend {
 pub struct WalletShieldedOutput<N> {
     pub index: usize,
     pub cmu: bls12_381::Scalar,
-    pub ephemeral_key: [u8; 32],
+    pub ephemeral_key: EphemeralKeyBytes,
     pub account: AccountId,
     pub note: Note,
     pub to: PaymentAddress,
--- a/zcash_client_backend/src/wallet.rs
+++ b/zcash_client_backend/src/wallet.rs
@@ -54,7 +54,7 @@ pub struct WalletShieldedSpend {
 pub struct WalletShieldedOutput<N> {
     pub index: usize,
     pub cmu: bls12_381::Scalar,
-    pub epk: jubjub::ExtendedPoint,
+    pub ephemeral_key: [u8; 32],
     pub account: AccountId,
     pub note: Note,
     pub to: PaymentAddress,
--- a/zcash_primitives/src/sapling/note_encryption.rs
+++ b/zcash_primitives/src/sapling/note_encryption.rs
@@ -436,8 +436,8 @@ mod tests {
     use std::convert::TryInto;
 
     use zcash_note_encryption::{
-        NoteEncryption, OutgoingCipherKey, ENC_CIPHERTEXT_SIZE, NOTE_PLAINTEXT_SIZE,
-        OUT_CIPHERTEXT_SIZE, OUT_PLAINTEXT_SIZE,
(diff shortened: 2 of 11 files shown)


str4d:
The effect of this change is on the order of 5%, around the noise threshold for the current single-fixed-output benchmark. We'd see more of an effect from this if we removed the much larger time cost from the `from_bytes` square root for deriving `epk` from `ephemeral_key` (which _is_ needed in the trial decryption protocol).

str4d:
Before:
image

After:
image

(Reminder that the colours don't actually mean anything, they are just for distinguishing things.)

zcash_client_backend/src/wallet.rs
```diff
@@ -54,7 +54,7 @@ pub struct WalletShieldedSpend {
 pub struct WalletShieldedOutput<N> {
     pub index: usize,
     pub cmu: bls12_381::Scalar,
-    pub epk: jubjub::ExtendedPoint,
+    pub ephemeral_key: [u8; 32],
```
nuttycom, line 57: Would it make sense for this to be typed as `EphemeralKeyBytes`, which we're using elsewhere?

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -551,12 +551,17 @@ mod tests {
         ovk: &OutgoingViewingKey,
         cv: &jubjub::ExtendedPoint,
         cmu: &bls12_381::Scalar,
-        epk: &jubjub::ExtendedPoint,
+        ephemeral_key: &[u8; 32],
```
nuttycom, line 554: Same question here.

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -339,15 +345,15 @@ impl OutputDescriptionV5 {
 }
 
 pub struct CompactOutputDescription {
-    pub epk: jubjub::ExtendedPoint,
+    pub ephemeral_key: [u8; 32],
```
nuttycom, line 348: And here.

nuttycom (approved):
utACK with a minor suggestion.

daira:
What we should have done is made [KA<sup>Orchard</sup>](https://zips.z.cash/protocol/protocol.pdf#concreteorchardkeyagreement) return only the x-coordinate of sharedSecret (with a corresponding change to KDF<sup>Orchard</sup>). That way, it would be possible to compute the shared secret with an x-only ladder, and so no square root would be required in the fast path of trial decryption.

str4d:
The `zcash_note_encryption` crate API is compatible with this. `Domain::SharedSecret` is produced only by `Domain::{ka_agree_enc, ka_agree_dec}` interface, and consumed only by `Domain::kdf` (and soon `Domain::batch_kdf`), so it's entirely internal and can be whatever type a particularl protocol needs.

nuttycom (approved):
re-utACK 0facec094f3ea71cb3a94cdc687d1ac11f7c3e4c

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -255,9 +255,10 @@ impl OutputDescription<GrothProofBytes> {
         let cmu = read_base(&mut reader, "cmu")?;
 
         // Consensus rules (§4.5):
-        // - Canonical encoding is enforced here.
+        // - Canonical encoding is enforced in librustzcash_sapling_check_output by zcashd
```
daira, line 258: Are there any clients other than zcashd for which this change would be unsafe?

daira (approved):
utACK with a question.

zcash_primitives/src/transaction/components/sapling.rs
str4d: - `zcashd` enforces this as documented, so this will continue to be enforced for all mined notes.
- `zebrad` (I believe) does its own thing, so shouldn't be affected by this change.
- `ephemeral_key` is always parsed (and thus validated) as part of trial-decryption (for light clients in particular), so this will continue to be enforced for individual users' notes.

PR #425, opened 2021-08-10
Batch note decryption

### LINKED ISSUE #422: `zcash_note_encryption`: Batch inversions for trial decryption
Trial decryption involves a minimum of one scalar multiplication with a unique base per output, and a KDF invocation. We can't batch the scalar multiplications, as we need to produce the individual shared secrets. However, we then need to encode the shared secrets as bytes for input to the KDF, which requires a field inversion. If we have a large batch of outputs that we want to decrypt (and/or a large number of `ivk`s we want to decrypt with), we could compute the KDFs together and perform a batch inversion.

For the `orchard` crate, the inversion inside the KDF is around 9% of the base-line cost (for trial-decrypting outputs that are not ours). So this optimisation is small but significant over the lifetime of a wallet.

str4d:
Closes #422.

### DIFF
--- a/components/zcash_note_encryption/src/batch.rs
+++ b/components/zcash_note_encryption/src/batch.rs
@@ -35,18 +35,15 @@ fn batch_note_decryption<D: Domain, Output: ShieldedOutput<D>, F, FR>(
 where
     F: Fn(&D, &D::IncomingViewingKey, &EphemeralKeyBytes, &Output, D::SymmetricKey) -> Option<FR>,
 {
-    // Fetch the ephemeral keys for each output.
-    let ephemeral_keys: Vec<_> = outputs
-        .iter()
-        .map(|(_, output)| output.ephemeral_key())
-        .collect();
+    // Fetch the ephemeral keys for each output and batch-parse them.
+    let ephemeral_keys = D::batch_epk(outputs.iter().map(|(_, output)| output.ephemeral_key()));
 
     // Derive the shared secrets for all combinations of (ivk, output).
-    // None of this work can benefit from batching.
+    // The scalar multiplications cannot benefit from batching.
     let items = ivks.iter().flat_map(|ivk| {
-        ephemeral_keys.iter().map(move |ephemeral_key| {
+        ephemeral_keys.iter().map(move |(epk, ephemeral_key)| {
             (
-                D::epk(ephemeral_key).map(|epk| D::ka_agree_dec(ivk, &epk)),
+                epk.as_ref().map(|epk| D::ka_agree_dec(ivk, epk)),
                 ephemeral_key,
             )
         })
@@ -64,7 +61,7 @@ where
                 .zip(outputs.iter())
         })
         .zip(keys)
-        .map(|(((ivk, ephemeral_key), (domain, output)), key)| {
+        .map(|(((ivk, (_, ephemeral_key)), (domain, output)), key)| {
             // The `and_then` propagates any potential rejection from `D::epk`.
             key.and_then(|key| decrypt_inner(domain, ivk, ephemeral_key, output, key))
         })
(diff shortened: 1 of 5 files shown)


str4d:
Here are some rough benchmarks of base-line (invalid) batched trial decryption throughput:
- i7-8700K, overclocked to 4.8 GHz
- Throughput is in Kelem/s (thousand trial-decryptions per second)

| Code | Type | Single | Batch of 10 | Batch of 1000 |
| --- | --- | --- | --- | --- |
| Main branch | Full | 8692 | - | - |
| Main branch | Compact | 8814 | - | - |
| This PR | Full | 8691 | 9286 | 9342 |
| This PR | Compact | 8824 | 9437 | 9484 |
| This PR + https://github.com/zkcrypto/jubjub/pull/51 | Full | 8660 | 9891 | 10147 |
| This PR + https://github.com/zkcrypto/jubjub/pull/51 | Compact | 8768 | 9935 | 10256 |

So we see a throughput improvement for a size-10 batch of around 7% from this PR, rising to around 14% with https://github.com/zkcrypto/jubjub/pull/51 (which causes two inversions per trial-decryption to be batched instead of one, since we can do that for Edwards curves). Larger batch sizes do have higher throughput but with clearly diminishing returns.

components/zcash_note_encryption/src/batch.rs
```diff
@@ -0,0 +1,69 @@
...
+    decrypt_inner: F,
+) -> Vec<Option<FR>>
+where
+    F: Fn(&D, &D::IncomingViewingKey, &EphemeralKeyBytes, &Output, D::SymmetricKey) -> Option<FR>,
+{
+    // Fetch the ephemeral keys for each output and batch-parse them.
+    let ephemeral_keys = D::batch_epk(outputs.iter().map(|(_, output)| output.ephemeral_key()));
+
+    // Derive the shared secrets for all combinations of (ivk, output).
+    // The scalar multiplications cannot benefit from batching.
```
daira, line 42: They could benefit from parallelisation, maybe. But rayon or crossbeam is not currently a dependency.

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -178,6 +185,37 @@ impl<P: consensus::Parameters> Domain for SaplingDomain<P> {
...
+            .iter()
+            .filter_map(|s| s.map(ExtendedPoint::from))
+            .collect();
+        let mut secrets_affine = vec![AffinePoint::identity(); shared_secrets.len()];
+        group::Curve::batch_normalize(&secrets, &mut secrets_affine);
+
+        let mut secrets_affine = secrets_affine.into_iter();
+        shared_secrets
+            .into_iter()
+            .map(|s| s.and_then(|_| secrets_affine.next()))
```
daira, line 203: This is kinda funky, but I can't think of a better way. Maybe add a comment explaining it?

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -1359,4 +1428,29 @@ mod tests {
...
+                    random_enc_ciphertext_with(height, &valid_ivk, &mut rng).2,
+                )
+            })
+            .collect();
+
+        let res = batch::try_note_decryption(&[invalid_ivk, valid_ivk], &outputs);
+        assert_eq!(res.len(), 20);
+        assert_eq!(&res[..10], &vec![None; 10][..]);
+        for result in &res[10..] {
+            assert!(result.is_some());
```
daira, line 1453: Check that this gives the same result as unbatched trial decryption.

daira (approved):
utACK with comments.

components/zcash_note_encryption/src/batch.rs
str4d: Yes, I expect any of that parallelism to be managed by the caller outside these APIs.

zcash_primitives/src/sapling/note_encryption.rs
str4d: I think we maybe need a rectangular return value (`HashMap` maybe?) instead of a linear one; I haven't actually tried using this in a client yet, just benchmarks. We can always improve on this later.

PR #341, opened 2021-02-06
A first pass at autoshielding via the data access API

nuttycom:
This is mostly @pacu's work, now integrated with the data access API.

PR #429, opened 2021-08-23
Bump dependencies

str4d:
This bumps everything that isn't a patched dependency, except:
- `rusqlite` (we should merge #341 first).
- `quickcheck` (the `1.0` is painful, we should switch to `proptest` instead like everything else).


nuttycom (approved):
utACK :tada:

daira (approved):
utACK

PR #432, opened 2021-08-30
zcash_note_encryption: Switch to chacha20poly1305 crate

str4d:


components/zcash_note_encryption/src/lib.rs
```diff
@@ -481,7 +492,9 @@ fn try_compact_note_decryption_inner<D: Domain, Output: ShieldedOutput<D>>(
     // Start from block 1 to skip over Poly1305 keying output
     let mut plaintext = [0; COMPACT_NOTE_SIZE];
     plaintext.copy_from_slice(output.enc_ciphertext());
-    ChaCha20Ietf::xor(key.as_ref(), &[0u8; 12], 1, &mut plaintext);
+    let mut keystream = ChaCha20::new(key.as_ref().into(), [0u8; 12][..].into());
+    keystream.seek(64);
```
therealyingtong, line 496: How do we get 64 here?

therealyingtong (approved):
utACK

components/zcash_note_encryption/src/lib.rs
str4d: ChaCha20 uses 64-byte blocks, so this is equivalent to incrementing the counter by 1.

PR #433, opened 2021-08-30
Safe compact size

nuttycom:
Builds upon #428

PR #428, opened 2021-08-18
Move the CompactSize, Vector, Array and Option encodings to a new zcash_encoding crate.

nuttycom:

PR #434, opened 2021-08-31
Update unified addresses to use CompactSize for typecode & length fields.

nuttycom:
Builds upon #428, #433


components/zcash_address/Cargo.toml
```diff
@@ -15,9 +15,11 @@ edition = "2018"
 bech32 = "0.8"
 blake2b_simd = "0.5"
 bs58 = { version = "0.4", features = ["check"] }
+zcash_encoding = { version = "0.0", path = "../zcash_encoding" }
 
 [dev-dependencies]
 proptest = "1"
+assert_matches = "1.3.0"
```
str4d, line 22:
```suggestion
assert_matches = "1.3.0"
proptest = "1"
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -34,7 +35,7 @@ pub enum Typecode {
     P2sh,
     Sapling,
     Orchard,
-    Unknown(u8),
+    Unknown(u64),
```
str4d, line 38: We should double-check that ZIP 316 is clear on whether it allows the full CompactSize space, or whether it is retaining the Zcash consensus rule variant with a limit below a `u32` (in which case this could be a `u32`). Open an issue for this.

components/zcash_address/src/kind/unified.rs
```diff
@@ -208,37 +211,55 @@ impl TryFrom<(&str, &[u8])> for Address {
...
+                    e
+                ))
+            })?;
+            let length: usize = CompactSize::read_t(&mut cursor).map_err(|e| {
+                ParseError::InvalidEncoding(format!(
+                    "Failed to deserialize CompactSize-encoded length {}",
+                    e
+                ))
+            })?;
+            println!("Got length for address {}", length);
```
str4d:
```suggestion
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -208,37 +211,55 @@ impl TryFrom<(&str, &[u8])> for Address {
     type Error = ParseError;
 
     fn try_from((hrp, buf): (&str, &[u8])) -> Result<Self, Self::Error> {
-        let encoded = f4jumble::f4jumble_inv(buf).ok_or(ParseError::InvalidEncoding)?;
+        fn read_receiver<R: Read>(mut cursor: R) -> Result<Receiver, ParseError> {
+            let typecode: u64 = CompactSize::read(&mut cursor).map_err(|e| {
```
str4d, line 215:
```suggestion
            let typecode = CompactSize::read(&mut cursor).map_err(|e| {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -211,28 +211,24 @@ impl TryFrom<(&str, &[u8])> for Address {
...
             })?;
-            let mut addr_buf = vec![0; length];
-            cursor.read_exact(&mut addr_buf).map_err(|e| {
-                ParseError::InvalidEncoding(format!(
-                    "Error reading {} bytes of address data: {}",
-                    length, e
-                ))
-            })?;
+            let addr_end = cursor.position() + length;
+            let addr = &cursor.get_ref()[cursor.position() as usize..addr_end as usize];
```
str4d, line 228: We need to check the length here, so we can return an error instead of panicking (as this implies a maliciously-constructed UA that is attempting to perform a buffer overflow read).

nuttycom: Fixed.

components/zcash_address/src/kind/unified.rs
nuttycom: https://github.com/zcash/zips/pull/551#pullrequestreview-745416777

components/zcash_address/src/kind/unified.rs
```diff
@@ -208,37 +211,64 @@ impl TryFrom<(&str, &[u8])> for Address {
...
+                ))
+            })?;
+            let addr_end = cursor.position().checked_add(length).ok_or_else(|| {
+                ParseError::InvalidEncoding(format!(
+                    "Length value {} caused an overflow error",
+                    length
+                ))
+            })?;
+            let addr = cursor.get_ref();
+            if (addr.len() as u64) < length {
```
str4d, line 234: We need to check we can read `length` bytes from `cursor.position()`, not from the beginning (`Cursor::get_ref` returns a reference to the entire underlying data, not the remaining slice):
```suggestion
            if (addr.len() as u64) < addr_end {
```

str4d (changes requested):

components/zcash_address/src/kind/unified.rs
nuttycom: Ahh, that shouldn't be called `addr` then. Fixed.

components/zcash_address/src/kind/unified.rs
```diff
@@ -208,37 +211,64 @@ impl TryFrom<(&str, &[u8])> for Address {
...
-                _ => Some(Err(ParseError::InvalidEncoding)),
-            })
-            .collect::<Result<_, _>>()
-            .and_then(|receivers: Vec<Receiver>| receivers.try_into())
+        let mut cursor = std::io::Cursor::new(encoded);
+        let mut result = vec![];
+        while cursor.position() < encoded.len().try_into().unwrap() {
+            result.push(read_receiver(&mut cursor)?);
+        }
+        result.try_into()
```
str4d, line 271:
```suggestion
        assert_eq!(cursor.position(), encoded.len().try_into().unwrap());
        result.try_into()
```

str4d (approved):
utACK

components/zcash_address/src/kind/unified/f4jumble.rs
```diff
@@ -113,15 +128,17 @@ mod tests {
...
     #[test]
     fn h_pers() {
-        assert_eq!(&H_PERS!(7), b"UA_F4Jumble_H_\x07\x00");
+        assert_eq!(&H_PERS!(7), b"UA_F4Jumble_H\x07\x00\x00");
     }
 
     #[test]
     fn g_pers() {
-        assert_eq!(&G_PERS!(7, 13), b"UA_F4Jumble_G_\x07\x0d");
+        assert_eq!(&G_PERS!(7, 13), b"UA_F4Jumble_G\x07\x0d\x00");
```
daira, line 136:
```suggestion
        assert_eq!(&G_PERS!(7, 13), b"UA_F4Jumble_G\x07\x0d\x00");
        assert_eq!(&G_PERS!(7, 65535), b"UA_F4Jumble_G\x07\xff\xff");
```

components/zcash_address/src/kind/unified/f4jumble.rs
```diff
@@ -113,15 +128,17 @@ mod tests {
...
     }
```
daira, line 138: Relating to `f4jumble_check_vectors` below:
In https://github.com/daira/zcash-test-vectors/commit/bdf4c9ff9b4ed298bc8fa91b6e0998898d5f1ef8 I added test vectors for very large inputs, where the input is easy to generate and the output is BLAKE2b-hashed. Can you update this to check those vectors as well?

nuttycom: Yup, will do, thanks for adding those.

components/zcash_address/src/kind/unified/f4jumble.rs
```diff
@@ -5,20 +5,35 @@ use std::ops::RangeInclusive;
...
+            74,
+            117,
+            109,
+            98,
+            108,
+            101,
+            95,
+            71,
+            $i,
+            $j as u8,
```
daira, line 35:
```suggestion
            ($j & 0xFF) as u8,
```

components/zcash_address/src/kind/unified/f4jumble/test_vectors_long.rs
```diff
@@ -0,0 +1,21 @@
+pub(crate) struct TestVector {
+    pub(crate) length: usize,
+    pub(crate) jumbled_hash: &'static[u8],
+}
+
+// From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/f4jumble.py
```
daira, line 6:
```suggestion
// From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/f4jumble_long.py
```

daira (approved):
utACK

components/zcash_address/src/kind/unified.rs
```diff
@@ -87,7 +88,7 @@ impl From<u8> for Typecode {
     }
 }
 
-impl From<Typecode> for u8 {
+impl From<Typecode> for u64 {
```
str4d, line 91:
```suggestion
impl From<Typecode> for u32 {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -125,9 +126,9 @@ impl fmt::Display for ParseError {
         match self {
             ParseError::BothP2phkAndP2sh => write!(f, "UA contains both P2PKH and P2SH receivers"),
             ParseError::DuplicateTypecode(typecode) => {
-                write!(f, "Duplicate typecode {}", u8::from(*typecode))
+                write!(f, "Duplicate typecode {}", u64::from(*typecode))
```
str4d, line 129:
```suggestion
                write!(f, "Duplicate typecode {}", u32::from(*typecode))
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -142,7 +143,7 @@ pub enum Receiver {
     Sapling(kind::sapling::Data),
     P2pkh(kind::p2pkh::Data),
     P2sh(kind::p2sh::Data),
-    Unknown { typecode: u8, data: Vec<u8> },
+    Unknown { typecode: u64, data: Vec<u8> },
```
str4d, line 146:
```suggestion
    Unknown { typecode: u32, data: Vec<u8> },
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -160,10 +161,10 @@ impl cmp::PartialOrd for Receiver {
     }
 }
 
-impl TryFrom<(u8, &[u8])> for Receiver {
+impl TryFrom<(u64, &[u8])> for Receiver {
```
str4d, line 164:
```suggestion
impl TryFrom<(u32, &[u8])> for Receiver {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -160,10 +161,10 @@ impl cmp::PartialOrd for Receiver {
...
     type Error = ParseError;
 
-    fn try_from((typecode, addr): (u8, &[u8])) -> Result<Self, Self::Error> {
+    fn try_from((typecode, addr): (u64, &[u8])) -> Result<Self, Self::Error> {
         match typecode.into() {
```
str4d, line 168:
```suggestion
    fn try_from((typecode, addr): (u32, &[u8])) -> Result<Self, Self::Error> {
        match typecode.try_into()? {
```

str4d (changes requested):

daira (approved):
utACK 4b67977ec0c890ced9754d8a7cd4688cec09696f. I have not checked that the test vectors match those generated by the Python code in zcash-test-vectors.

components/zcash_address/src/kind/unified.rs
```diff
@@ -75,19 +76,25 @@ impl PartialOrd for Typecode {
...
-            0x00 => Typecode::P2pkh,
-            0x01 => Typecode::P2sh,
-            0x02 => Typecode::Sapling,
-            0x03 => Typecode::Orchard,
-            _ => Typecode::Unknown(typecode),
+impl TryFrom<u32> for Typecode {
+    type Error = ParseError;
+
+    fn try_from(typecode: u32) -> Result<Self, Self::Error> {
+        if typecode < 0x02000000 {
```
str4d, line 83: This is the wrong bound:
https://github.com/zcash/librustzcash/blob/a3df9dd47d6c3fa89b08483cb8276515464de321/components/zcash_encoding/src/lib.rs#L57-L61

components/zcash_address/src/kind/unified.rs
```diff
@@ -75,19 +76,25 @@ impl PartialOrd for Typecode {
...
+            Ok(match typecode {
+                0x00 => Typecode::P2pkh,
+                0x01 => Typecode::P2sh,
+                0x02 => Typecode::Sapling,
+                0x03 => Typecode::Orchard,
+                _ => Typecode::Unknown(typecode),
+            })
+        } else {
+            Err(ParseError::InvalidTypecodeValue(typecode as u64))
         }
```
str4d, line 93: I would prefer this syntax, which is simpler (and the compiler can reason about) while keeping the invalid range explicit:
```suggestion
        match typecode {
            0x00 => Ok(Typecode::P2pkh),
            0x01 => Ok(Typecode::P2sh),
            0x02 => Ok(Typecode::Sapling),
            0x03 => Ok(Typecode::Orchard),
            0x02000001..=u32::MAX => Err(ParseError::InvalidTypecodeValue(typecode as u64)),
            _ => Ok(Typecode::Unknown(typecode)),
        }
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -208,37 +218,67 @@ impl TryFrom<(&str, &[u8])> for Address {
...
-        let encoded = f4jumble::f4jumble_inv(buf).ok_or(ParseError::InvalidEncoding)?;
+        fn read_receiver(mut cursor: &mut std::io::Cursor<&[u8]>) -> Result<Receiver, ParseError> {
+            let typecode = CompactSize::read(&mut cursor)
+                .map_err(|e| {
+                    ParseError::InvalidEncoding(format!(
+                        "Failed to deserialize CompactSize-encoded typecode {}",
+                        e
+                    ))
+                })
+                .and_then(|v| u32::try_from(v).map_err(|_| ParseError::InvalidTypecodeValue(v)))?;
```
str4d, line 229: This error-map is unnecessary, as `CompactSize::read` enforces the max size limit, so `v` is guaranteed to fit in a `u32`:
```suggestion
                .and_then(|v| u32::try_from(v).expect("CompactSize::read enforces MAX_SIZE limit"))?;
```

str4d (changes requested):
One bug, two suggestions.

components/zcash_address/src/kind/unified.rs
str4d: Or to be even more explicit (and have the compiler check that all values are covered):
```suggestion
        match typecode {
            0x00 => Ok(Typecode::P2pkh),
            0x01 => Ok(Typecode::P2sh),
            0x02 => Ok(Typecode::Sapling),
            0x03 => Ok(Typecode::Orchard),
            0x04..=0x02000000 => Ok(Typecode::Unknown(typecode)),
            0x02000001..=u32::MAX => Err(ParseError::InvalidTypecodeValue(typecode as u64)),
        }
```

PR #439, opened 2021-09-09
ff 0.11, group 0.11, etc.

str4d:


zcash_client_backend/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 ### Changed
 - MSRV is now 1.51.0.
+- Bumped dependencies to `ff 0.11`, `group 0.11`, `bls12_381 0.6`, `jubjub 0.8`.
```
daira, line 11:
```suggestion
- Bumped dependencies to `ff 0.11`, `group 0.11`, `bls12_381 0.6`, `jubjub 0.8`,
  and the most recent unreleased versions of `halo2` and `orchard`.
```

zcash_primitives/src/sapling.rs
```diff
@@ -216,7 +217,7 @@ impl ViewingKey {
         // Drop the most significant five bits, so it can be interpreted as a scalar.
         h[31] &= 0b0000_0111;
 
-        SaplingIvk(jubjub::Fr::from_repr(h).expect("should be a valid scalar"))
+        SaplingIvk(jubjub::Fr::from_repr(h).unwrap())
```
ebfull, line 220: We should eventually add an `expect` method to `CtOption`.

ebfull (approved):

zcash_primitives/CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.51.0.
+- Bumped dependencies to `ff 0.11`, `group 0.11`, `bls12_381 0.6`, `jubjub 0.8`.
```
daira, line 57:
```suggestion
- Bumped dependencies to `ff 0.11`, `group 0.11`, `bls12_381 0.6`,
  `jubjub 0.8`, `pasta_curves 0.2`
```

str4d: It's fine as-is; it's within 80 characters, which is what this time is wrapped at IIRC.

daira: Right, but `pasta_curves` was bumped.

str4d: We haven't cut a release containing Orchard logic in a public API yet, so bumping the `pasta_curves` crate doesn't change anything.

PR #441, opened 2021-09-14
`impl memuse::DynamicUsage for Amount`

str4d:
Required to use `DynamicUsage` with `orchard::Bundle<_, Amount>`.


nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK

PR #442, opened 2021-09-16
Bump Orchard patch to include Merkle path cleanups

str4d:


daira (approved):
utACK

nuttycom (approved):
utACK, verified the `orchard` dependency hash.

PR #445, opened 2021-09-22
Update dependencies

str4d:
Includes `halo2` and `orchard` crate versions with the BOSL exception for our crates.


nuttycom (approved):
utACK, verified commit hashes & consistency of pasta_curves versions.

PR #446, opened 2021-09-27
Add `#![no_std]` support for f4jumble.

jarys:
This PR adds no-std support to `f4jumble`.

I have to xor G's output piece by piece to the message buffer. Separation of these two operations (computing G and xor) would require another auxiliary buffer.


components/f4jumble/src/lib.rs
```diff
@@ -40,92 +55,110 @@ macro_rules! G_PERS {
...
     }
 }
 
-fn xor(a: &[u8], b: &[u8]) -> Vec<u8> {
-    a.iter().zip(b.iter()).map(|(a0, b0)| a0 ^ b0).collect()
+// xor bytes of the `source` to bytes of the `target`
+fn xor(target: &mut [u8], source: &[u8]) {
+    for i in 0..min(source.len(), target.len()) {
+        target[i] ^= source[i];
+    }
```
str4d, line 107:
```suggestion
    for (source, target) in source.iter().zip(target.iter_mut()) {
        *target ^= source;
    }
```

components/f4jumble/src/lib.rs
```diff
@@ -1,13 +1,28 @@
...
 mod test_vectors_long;
 
-const VALID_LENGTH: RangeInclusive<usize> = 48..=4194368;
+pub const VALID_LENGTH: RangeInclusive<usize> = 48..=4194368;
+const INVALID_LENGTH_ERROR_MESSAGE: &str = formatcp!(
+    "Message length must be in interval ({}..={})",
+    *VALID_LENGTH.start(),
+    *VALID_LENGTH.end()
+);
+//more andvanced formatcp! only in nightly Rust
```
str4d, line 25: Rather than using a string (which is hard to keep track of in callers), we should use an `Error` enum with a single case. Then if `std` is enabled, we can implement `std::error::Error` for it.

We should also return this `Error` enum from the allocating methods.

str4d (commented):
Generally LGTM.

components/f4jumble/src/lib.rs
str4d: Something like:
```suggestion
use core::fmt;

#[derive(Debug)]
pub enum Error {
    InvalidLength,
}

impl fmt::Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            Self::InvalidLength => [...],
        }
    }
}

#[cfg(feature = "std")]
impl std::error::Error for Error {}
```

components/f4jumble/src/lib.rs
```diff
@@ -40,92 +71,110 @@ macro_rules! G_PERS {
...
+    let mut result = message.to_vec();
+    let res = f4jumble_inv_mut(&mut result);
+    if res.is_ok() {
+        Some(Vec::from(result))
     } else {
         None
     }
 }
 
+#[cfg(feature = "std")]
```
daira, line 170: This means that the tests will not be run with no_std. So if there were a platform-dependent bug on an embedded platform, for example, we wouldn't catch it. Please add tests that run with no_std.

components/f4jumble/src/test_vectors.rs
```diff
@@ -1,9 +1,18 @@
...
+extern crate std;
+#[cfg(feature = "std")]
+use std::vec;
+#[cfg(feature = "std")]
+use std::vec::Vec;
+
+#[cfg(feature = "std")]
 pub(crate) struct TestVector {
     pub(crate) normal: Vec<u8>,
     pub(crate) jumbled: Vec<u8>,
```
daira, line 11: We could maybe change these to slices with static lifetime, to allow the test vectors to be used in no_std.

daira (changes requested):
Please add tests for no_std. Otherwise this looks good.

components/f4jumble/src/test_vectors.rs
```diff
@@ -1,4872 +1,4879 @@
...
-                0x17, 0x8a, 0xc9, 0x51, 0x31, 0x05, 0xfc, 0xc7, 0xe3, 0x0b, 0xa8, 0xe0, 0x16, 0xaa,
-                0x36, 0xa6, 0xb5, 0xdf, 0x5e, 0x5a, 0x19, 0x09, 0xf6, 0x3a, 0xba, 0x09, 0x5d, 0x98,
-                0x77, 0xa8, 0xf2, 0xdc, 0x53, 0xf4, 0x6f, 0x6c, 0x9b, 0x07, 0xad, 0xdf, 0x14, 0x6f,
-                0x4f, 0xfa, 0x50, 0x1f, 0x9d, 0xd3, 0xcf, 0xf9, 0x24, 0xe3, 0x01, 0x0f, 0xaf, 0x50,
-                0x4e, 0x2b, 0x8a, 0xca, 0x73, 0x57, 0xac, 0xbf, 0xfe, 0xc7, 0x3a, 0xc3, 0x4c, 0x1a,
-                0x73, 0x16, 0x0f, 0x2c, 0xea, 0x1e, 0x05, 0x10, 0xf8, 0x4d, 0x2f, 0xe2, 0xf7, 0x3b,
-                0x6e, 0x92, 0x19, 0x07, 0xa1, 0xb7, 0xb3, 0x75, 0x12, 0x13, 0x24, 0x1b, 0x2c, 0xfa,
-                0xa5, 0x5a, 0x5e, 0xa4, 0xdd, 0x51, 0x7e, 0x7b, 0x49, 0xd2, 0xde, 0x8c, 0x09, 0x08,
-                0x43, 0x73, 0x0d, 0x24, 0x08, 0xa2, 0xa3, 0x04, 0xaa, 0x1e, 0x2e, 0x13, 0x70, 0xa6,
-                0xbf, 0x6c, 0x2b, 0xc7, 0x3f, 0xf0, 0x0d, 0x89,
```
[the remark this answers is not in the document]
jarys: I was thinking of adding `long-test-vectors` feature. Then I changed my mind and I forgot to remove it properly.

str4d (approved):

PR #447, opened 2021-09-28
NU5 testnet update

str4d:


defuse:
utACK I double-checked hashes and constants

zcash_primitives/src/consensus.rs
```diff
@@ -240,7 +240,7 @@ impl Parameters for TestNetwork {
             NetworkUpgrade::Blossom => Some(BlockHeight(584_000)),
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_800)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_028_500)),
-            NetworkUpgrade::Nu5 => None,
+            NetworkUpgrade::Nu5 => Some(BlockHeight(1_599_200)),
```
str4d, line 243: I calculate this to be 8:28pm UTC, October 7th.

zcash_primitives/src/consensus.rs
```diff
@@ -458,7 +458,7 @@ impl TryFrom<u32> for BranchId {
             0x2bb4_0e60 => Ok(BranchId::Blossom),
             0xf5b9_230b => Ok(BranchId::Heartwood),
             0xe9ff_75a6 => Ok(BranchId::Canopy),
-            0xf919_a198 => Ok(BranchId::Nu5),
+            0x3751_9621 => Ok(BranchId::Nu5),
```
str4d, line 461: Thanks `/dev/urandom` for your contribution 😄

zcash_primitives/src/consensus.rs
defuse: Did we miss this change in the previous release?

str4d:
ZIP 244 test vectors are updated using https://github.com/zcash-hackworks/zcash-test-vectors/pull/54.

defuse:
utACK again :)

zcash_primitives/src/consensus.rs
therealyingtong: I've double-checked this calculation.

Cargo.toml
```diff
@@ -20,8 +20,7 @@ codegen-units = 1
 
 [patch.crates-io]
 # In development.
-halo2 = { git = "https://github.com/zcash/halo2.git", rev = "a7cd600eb60b1528159b92af5e426adcc615de1a" }
-orchard = { git = "https://github.com/zcash/orchard.git", rev = "8779ce8f1a638ebbc9b229d4eff3a29ef4de7ac0" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "2c8241f25b943aa05203eacf9905db117c69bd29" }
```
therealyingtong, line 23: I've checked that this commit merges the circuit bugfixes in https://github.com/zcash/orchard/pull/209.

nuttycom (approved):
utACK

therealyingtong (approved):
utACK

zcash_primitives/src/consensus.rs
daira: @defuse yes we did.

daira (commented):
Post-hoc utACK

zcash_primitives/src/consensus.rs
str4d: It wasn't technically necessary, as we don't use these parameters in zcashd. But it's good practice to update them in sync.

Issue #443, opened 2021-09-17
Ensure that GitHub's "View license" link points to complete license info for all our repos


daira:
GitHub's "View license" link can only point to one file. For example, if there is `LICENSE-APACHE` and `LICENSE-MIT` but no `COPYRIGHT` or `COPYING`, it will point to `LICENSE-APACHE`. This is misleading because it disregards the MIT licensing option.

Here are examples of dual MIT/Apache repos where the "View license" link points to a file with complete license info:
* https://github.com/rust-lang/rust (points to `COPYRIGHT`)
* https://github.com/zcash/zcash (points to `COPYING`)

Dual-licensed repos that this affects:
* [x] https://github.com/zcash/librustzcash/pull/449
* [x] https://github.com/zcash/incrementalmerkletree/pull/8
* [x] https://github.com/zcash/pasta_curves/pull/23
* [x] https://github.com/zcash-hackworks/zcash-test-vectors/pull/53
  * [x] updated in https://github.com/zcash-hackworks/zcash-test-vectors/pull/60
* [x] https://github.com/zcash-hackworks/bn/pull/23

Other repos with no or unrecognized license information:
* [ ] https://github.com/zcash-hackworks/zcash-testnet-in-a-box [should be MIT/Apache]
* [ ] https://github.com/zcash/mpc [should be MIT/Apache]
* [ ] https://github.com/zcash/zcash-gitian [should be MIT/Apache]
* [ ] https://github.com/zcash/gitian.sigs [should be MIT/Apache]
* [ ] https://github.com/zcash/sapling-security-analysis [should be MIT/Apache]
* [ ] https://github.com/zcash/developers [should be MIT/Apache]
* [ ] https://github.com/zcash/zcash-seeder (forked from https://github.com/sipa/bitcoin-seeder which also has unclear licensing)
* [ ] https://github.com/zcash/zeclink [deleted, but shouldn't have been?]
* [ ] https://github.com/zcash/homebrew-formulae
* [ ] https://github.com/zcash/zcash-blog
* [ ] https://github.com/zcash/zcash-gitcoin (basically empty)
* [ ] https://github.com/zcash/zcash-wordpress-theme

daira:
Review needed for https://github.com/zcash-hackworks/zcash-test-vectors/pull/53; if that is okay then I'll file similar PRs for the other four MIT/Apache repos.

PR #449, opened 2021-10-17
Ensure that GitHub's "View license" link points to complete license information

daira:
refs https://github.com/zcash/librustzcash/issues/443

Signed-off-by: Daira Hopwood <daira@jacaranda.org>


README.md
```diff
@@ -9,16 +9,4 @@ These libraries are currently under development and have not been fully-reviewed
...
-
-at your option.
-
-### Contribution
-
-Unless you explicitly state otherwise, any contribution intentionally
-submitted for inclusion in the work by you, as defined in the Apache-2.0
-license, shall be dual licensed as above, without any additional terms or
-conditions.
+This repository is [MIT/Apache 2.0 dual-licensed](COPYING.md).
```
str4d, line 12: The existing preamble above is an ecosystem-wide standard for Rust crates, because what crates.io shows is the `README.md`, and the `Cargo.toml`'s `license` field, not any `COPYING.md` file. I would greatly prefer that the above preamble is left exactly as-is, and we instead duplicate it in `COPYING.md` for GitHub's benefit. If you particularly want a link from the `README.md` to `COPYING.md`, linkify "Licensed under" in the first sentence.

str4d (changes requested):

daira (commented):
[commented on the wrong PR]

nuttycom (approved):
ACK

PR #450, opened 2021-10-20
Add `#![no_std]` support for zcash_note_encryption

jarys:
This PR adds no_std compatibility for zcash_note_encryption.

All functionalities and API are preserved. Modifications are light and straightforward.

Motivation: implementation into Trezor and other hardware wallets.

### DIFF
--- a/components/zcash_note_encryption/Cargo.toml
+++ b/components/zcash_note_encryption/Cargo.toml
@@ -12,18 +12,22 @@ license = "MIT OR Apache-2.0"
 edition = "2018"
 
 [dependencies]
-blake2b_simd = "0.5"
-byteorder = "1"
-chacha20 = "0.8"
-chacha20poly1305 = "0.9"
-ff = "0.11"
-group = "0.11"
-rand_core = "0.6"
-subtle = "2.2.3"
+blake2b_simd = { version = "0.5", default-features = false }
+byteorder = { version = "1", default-features = false }
+chacha20 = { version = "0.8", default-features = false }
+chacha20poly1305 = { version = "0.9", default-features = false }
+ff = { version = "0.11", default-features = false }
+group = { version = "0.11", default-features = false }
+rand_core = { version = "0.6", default-features = false }
+subtle = { version = "2.2.3", default-features = false }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


components/zcash_note_encryption/src/lib.rs
```diff
@@ -120,6 +127,7 @@ pub trait Domain {
     ///
     /// For each item in the batch, if the shared secret is `None`, this returns `None` at
     /// that position.
+    #[cfg(feature = "alloc")]
```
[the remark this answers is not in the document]
jarys: I see.

str4d:
This can be rebased now that #452 has merged.

components/zcash_note_encryption/Cargo.toml
```diff
@@ -12,18 +12,22 @@ license = "MIT OR Apache-2.0"
...
 [dependencies]
-blake2b_simd = "0.5"
-byteorder = "1"
-chacha20 = "0.8"
-chacha20poly1305 = "0.9"
-ff = "0.11"
-group = "0.11"
-rand_core = "0.6"
-subtle = "2.2.3"
+blake2b_simd = { version = "0.5", default-features = false }
```
str4d, line 15: Hmm, this crate uses `std` to perform CPU detection for performance. We should keep that enabled by default, so I think we should add a `std` feature flag that isn't used by the crate itself, but just enables all the underlying dependency `std` flags.

components/zcash_note_encryption/Cargo.toml
```diff
@@ -12,18 +12,22 @@ license = "MIT OR Apache-2.0"
...
+rand_core = { version = "0.6", default-features = false }
+subtle = { version = "2.2.3", default-features = false }
 
 [dev-dependencies]
 zcash_primitives = { version = "0.5", path = "../../zcash_primitives" }
 jubjub = "0.8"
 
+[features]
+default = ["alloc"]
+alloc = []
```
str4d, line 30:
```suggestion
default = ["std"]
alloc = []
std = ["alloc", "blake2b_simd/std"]
```

str4d (approved):
utACK with comment.

PR #452, opened 2021-10-28
Introduce `zcash_note_encryption::BatchDomain` trait

str4d:
This change makes it possible to make the crate no-std compatible.


therealyingtong (approved):
utACK

PR #459, opened 2021-11-22
Download sprout parameters in zcash_proofs

teor2345:
## Motivation

Zebra wants to use `zcash_proofs` to download both Sprout and Sapling parameters.

## Changes

Add sprout:
- Add a `download_sprout_parameters` function.
- Stream sprout and sapling downloads to disk, rather than loading the whole file into RAM.

Functionality changes:
- If the parameter files already exist, check them instead of starting a download.
- Allow the caller to specify a download timeout.
- Check file sizes as well as file hashes (useful for debugging).
- Remove downloaded files on error (but leave existing files alone) .

Interface changes:
- Add `download_sapling_parameters` and deprecate `download_parameters`. This avoids confusion between sprout and sapling downloads, while maintaining backward compatibility.
- Make the Sprout and Sapling parameter file names public. This makes it easier for callers to supply the correct paths to `load_parameters`.

Bug fixes:
- Add a missing feature dependency: `download_params` -> `directories`. This makes sure `download_parameters` will work if it's the only feature specified.

Tweaks:
- Refactor file loads to use the same verifying function as downloads.

### DIFF
--- a/zcash_proofs/examples/download-sprout-and-sapling-params.rs
+++ b/zcash_proofs/examples/download-sprout-and-sapling-params.rs
@@ -1,0 +1,17 @@
+fn main() -> Result<(), minreq::Error> {
+    const DOWNLOAD_TIMEOUT_SECONDS: u64 = 3600;
+
+    // Always do a download to /tmp, if compiled with `RUSTFLAGS="--cfg always_download"`
+    #[cfg(always_download)]
+    {
+        std::env::set_var("HOME", "/tmp");
+        let _ = std::fs::remove_dir(
+            zcash_proofs::default_params_folder().expect("unexpected missing HOME env var"),
+        );
+    }
+
+    zcash_proofs::download_sapling_parameters(Some(DOWNLOAD_TIMEOUT_SECONDS))?;
+    zcash_proofs::download_sprout_parameters(Some(DOWNLOAD_TIMEOUT_SECONDS))?;
+
+    Ok(())
+}
--- a/zcash_proofs/examples/download-sprout-and-sapling-params.rs
+++ b/zcash_proofs/examples/download-sprout-and-sapling-params.rs
@@ -1,16 +1,25 @@
 fn main() -> Result<(), minreq::Error> {
     const DOWNLOAD_TIMEOUT_SECONDS: u64 = 3600;
 
+    #[allow(unused_mut, unused_assignments)]
+    let mut params_folder =
+        zcash_proofs::default_params_folder().expect("unexpected missing HOME env var");
+
     // Always do a download to /tmp, if compiled with `RUSTFLAGS="--cfg always_download"`
     #[cfg(always_download)]
     {
         std::env::set_var("HOME", "/tmp");
-        let _ = std::fs::remove_dir(
-            zcash_proofs::default_params_folder().expect("unexpected missing HOME env var"),
-        );
+        params_folder =
+            zcash_proofs::default_params_folder().expect("unexpected missing HOME env var");
+
+        println!("removing temporary parameters folder: {:?}", params_folder);
+        let _ = std::fs::remove_dir_all(&params_folder);
     }
 
+    println!("downloading sapling parameters to: {:?}", params_folder);
     zcash_proofs::download_sapling_parameters(Some(DOWNLOAD_TIMEOUT_SECONDS))?;
+
+    println!("downloading sprout parameters to: {:?}", params_folder);
     zcash_proofs::download_sprout_parameters(Some(DOWNLOAD_TIMEOUT_SECONDS))?;
 
     Ok(())
--- a/zcash_proofs/src/downloadreader.rs
+++ b/zcash_proofs/src/downloadreader.rs
@@ -1,0 +1,79 @@
+//! [`io::Read`] implementations for [`minreq`].
+
+use std::io;
+
+/// A wrapper that implements [`io::Read`] on a [`minreq::ResponseLazy`].
+pub enum ResponseLazyReader {
+    Request(minreq::Request),
+    Response(minreq::ResponseLazy),
+    Complete(Result<(), String>),
+}
+
+impl From<minreq::Request> for ResponseLazyReader {
+    fn from(request: minreq::Request) -> ResponseLazyReader {
+        ResponseLazyReader::Request(request)
+    }
+}
+
+impl From<minreq::ResponseLazy> for ResponseLazyReader {
+    fn from(response: minreq::ResponseLazy) -> ResponseLazyReader {
+        ResponseLazyReader::Response(response)
+    }
+}
+
+impl io::Read for ResponseLazyReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        use ResponseLazyReader::*;
+
+        // Zero-sized buffer. This should never happen.
+        if buf.is_empty() {
+            return Ok(0);
+        }
+
+        loop {
+            match self {
+                // Launch a lazy response for this request
+                Request(request) => match request.clone().send_lazy() {
+                    Ok(response) => *self = Response(response),
+                    Err(error) => {
+                        let error = Err(format!("download request failed: {:?}", error));
+
+                        *self = Complete(error);
+                    }
+                },
+
+                // Read from the response
+                Response(response) => {
+                    // minreq has a very limited lazy reading interface.
+                    match &mut response.next() {
+                        // Read one byte into the buffer.
+                        // We ignore the expected length, because we have no way of telling the BufReader.
+                        Some(Ok((byte, _length))) => {
+                            buf[0] = *byte;
+                            return Ok(1);
+                        }
+
+                        // Reading failed.
+                        Some(Err(error)) => {
+                            let error = Err(format!("download response failed: {:?}", error));
+
+                            *self = Complete(error);
+                        }
+
+                        // Finished reading.
+                        None => *self = Complete(Ok(())),
+                    }
+                }
+
+                Complete(result) => {
+                    return match result {
+                        // Return a zero-byte read for download success and EOF.
+                        Ok(()) => Ok(0),
+                        // Keep returning the download error,
+                        Err(error) => Err(io::Error::new(io::ErrorKind::Other, error.clone())),
+                    };
+                }
+            }
+        }
+    }
+}
--- a/zcash_proofs/src/downloadreader.rs
+++ b/zcash_proofs/src/downloadreader.rs
@@ -23,7 +23,7 @@ impl From<minreq::ResponseLazy> for ResponseLazyReader {
 
 impl io::Read for ResponseLazyReader {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        use ResponseLazyReader::{Request, Response};
+        use ResponseLazyReader::{Complete, Request, Response};
 
         // Zero-sized buffer. This should never happen.
         if buf.is_empty() {
--- a/zcash_proofs/src/downloadreader.rs
+++ b/zcash_proofs/src/downloadreader.rs
@@ -23,7 +23,7 @@ impl From<minreq::ResponseLazy> for ResponseLazyReader {
 
 impl io::Read for ResponseLazyReader {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        use ResponseLazyReader::*;
+        use ResponseLazyReader::{Request, Response};
 
         // Zero-sized buffer. This should never happen.
         if buf.is_empty() {
--- a/zcash_proofs/src/lib.rs
+++ b/zcash_proofs/src/lib.rs
@@ -68,13 +68,29 @@ pub fn default_params_folder() -> Option<PathBuf> {
     })
 }
 
+/// Download the Zcash Sapling parameters, check their hash, and store them in the default location.
+///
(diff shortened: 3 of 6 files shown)


zcash_proofs/src/lib.rs
```diff
@@ -60,56 +78,158 @@ pub fn default_params_folder() -> Option<PathBuf> {
     })
 }
 
-/// Download the Zcash Sapling parameters, storing them in the default location.
+/// Download the Zcash Sapling parameters if needed, and store them in the default location.
+/// Always checks the hashes of the files, even if they didn't need to be downloaded.
```
teor2345, line 82: Missing docs:
```suggestion
/// Always checks the hashes of the files, even if they didn't need to be downloaded.
///
/// A timeout can be set using the `MINREQ_TIMEOUT` environmental variable.
```

zcash_proofs/src/lib.rs
```diff
@@ -60,56 +78,160 @@ pub fn default_params_folder() -> Option<PathBuf> {
...
         io::Error::new(io::ErrorKind::Other, "Could not load default params folder")
     })?;
     std::fs::create_dir_all(&params_dir)?;
 
-    let fetch_params = |name: &str, expected_hash: &str| -> Result<(), minreq::Error> {
-        use std::io::Write;
-
-        // Download the parts directly (Sapling parameters are small enough for this).
-        let part_1 = minreq::get(format!("{}/{}.part.1", DOWNLOAD_URL, name)).send()?;
-        let part_2 = minreq::get(format!("{}/{}.part.2", DOWNLOAD_URL, name)).send()?;
```
defuse, line 80: It's necessary for us to host these files in two parts because of [CloudFlare's maximum cached file size limit of 512 MB](https://developers.cloudflare.com/workers/platform/limits#response-limits). 

(It's very important that the files fit in the cache to prevent "denial of wallet" attacks. We are actually on an Enterprise plan at the moment so our *hard* limit is currently 5GB which is why it's safe to keep hosting those single-file deprecated URLs, but after people stopped attacking us it's looking like we might not actually need to pay for Enterprise plan so I don't want to guarantee that the unsplit files will remain.)

teor2345: Thanks, fixed in commit 76f1c15.

defuse: Thanks!

teor2345:
Chaining the Sprout requests is failing, probably because the second connection times out.

I'll try to fix it tomorrow.

conradoplg (commented):
@teor2345 asked me to take a look. It seems good.

I think it would improve usability to delete downloaded files on errors, so that the user won't need to manually delete them. But it's not a big deal.

teor2345:
> I think it would improve usability to delete downloaded files on errors, so that the user won't need to manually delete them. But it's not a big deal.

Sounds like a good idea, I'll see if I can make it happen today.

I was concerned we could accidentally delete files we haven't created. But now that we're handling existing and downloaded files separately, we can just do the deletes for downloaded files.

teor2345:
> Chaining the Sprout requests is failing, probably because the second connection times out.

This is fixed now.

Tested using:
```sh
RUSTFLAGS="--cfg always_download" cargo run --release --features=download-params --example download-sprout-and-sapling-params
```

teor2345:
I also checked the code with clippy using:
```sh
cargo clippy --all-features --no-deps --lib --examples
```

zcash_proofs/examples/download-sprout-and-sapling-params.rs
```diff
@@ -0,0 +1,26 @@
...
+        std::env::set_var("HOME", "/tmp");
+        params_folder =
+            zcash_proofs::default_params_folder().expect("unexpected missing HOME env var");
+
+        println!("removing temporary parameters folder: {:?}", params_folder);
+        let _ = std::fs::remove_dir_all(&params_folder);
+    }
+
+    println!("downloading sapling parameters to: {:?}", params_folder);
+    zcash_proofs::download_sapling_parameters(Some(DOWNLOAD_TIMEOUT_SECONDS))?;
```
daira, line 20: Instead, add `download_sapling_parameters_to_directory` and `download_sprout_parameters_to_directory` that allow specifying the target directory.

zcash_proofs/examples/download-sprout-and-sapling-params.rs
```diff
@@ -0,0 +1,26 @@
+fn main() -> Result<(), minreq::Error> {
+    const DOWNLOAD_TIMEOUT_SECONDS: u64 = 3600;
+
+    #[allow(unused_mut, unused_assignments)]
+    let mut params_folder =
+        zcash_proofs::default_params_folder().expect("unexpected missing HOME env var");
+
+    // Always do a download to /tmp, if compiled with `RUSTFLAGS="--cfg always_download"`
```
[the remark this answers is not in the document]
teor2345: When I was changing the download code multiple times per day, I ran this example with this configuration to test a fresh download.

(Running `download_sapling_parameters` repeatedly will download the files on the first attempt, then check the existing files on subsequent attempts.)

zcash_proofs/examples/download-sprout-and-sapling-params.rs
teor2345: Zebra doesn't need to be able to change the download directory in the public API, and setting `$HOME` is sufficient for testing purposes.

zcash_proofs/examples/download-sprout-and-sapling-params.rs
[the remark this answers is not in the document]
teor2345: I'm happy to add a comment here if that would help.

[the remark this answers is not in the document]
teor2345: In particular, I used it to resolve a bug where I was opening the connection for the second file too early, so it was timing out:
https://github.com/zcash/librustzcash/pull/459#issuecomment-978676910

teor2345:
@nuttycom @str4d I rebased on `master`, fixed a clippy warning, and updated this PR.

zcash_proofs/src/lib.rs
```diff
@@ -60,58 +88,196 @@ pub fn default_params_folder() -> Option<PathBuf> {
...
+        // Either:
+        // - return Ok if the paths exist, or
+        // - always load and return the parameters, for newly downloaded and existing files.
+
+        let file_path_string = params_path.to_string_lossy();
+
+        // Check the file size is correct before hashing large amounts of data.
+        verify_file_size(&params_path, expected_bytes, name, &file_path_string).expect(
+            "parameter file size is not correct, \
+             please clean your Zcash parameters directory and re-run `fetch-params`.",
```
defuse, line 199: Should this error reference `fetch-params` or the `download-params` tool in this repo?

zcash_proofs/src/lib.rs
```diff
@@ -206,3 +431,81 @@ pub fn parse_parameters<R: io::Read>(
...
+                name, expected_bytes, file_size, params_source,
+            ),
+        ));
+    }
+
+    Ok(())
+}
+
+/// Check if the Blake2b hash from `hash_reader` matches `expected_hash`,
+/// while streaming from `data` into `sink`.
```
defuse, line 463: What `data` is this comment referring to?

defuse (approved):
utACK with comments/questions

zcash_proofs/src/lib.rs
teor2345: I wrote something similar to the previous `panic!()` text, but I'm happy to change it if you'd like:
> panic!("Sapling spend parameter file is not correct, please clean your `~/.zcash-params/` and re-run `fetch-params`.");

https://github.com/zcash/librustzcash/pull/459/files#diff-67a539a91d2370483ef0aef0e6d89f1d1b480eaf9767975709ff767ec3a282b8L182

zcash_proofs/src/lib.rs
teor2345: I think the comment is outdated, the `data` is inside `hash_reader`.

I'll rebase this PR and fix up the comment.

zcash_proofs/src/downloadreader.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl From<minreq::ResponseLazy> for ResponseLazyReader {
+    fn from(response: minreq::ResponseLazy) -> ResponseLazyReader {
+        ResponseLazyReader::Response(response)
+    }
+}
+
+impl io::Read for ResponseLazyReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        use ResponseLazyReader::*;
```
nuttycom, line 26: nit: we prefer to use individual imports rather than wildcard imports

zcash_proofs/src/lib.rs
```diff
@@ -33,20 +33,48 @@ pub mod sprout;
...
+#[cfg(any(feature = "local-prover", feature = "download-params"))]
+pub const SPROUT_NAME: &str = "sprout-groth16.params";
 
 // Circuit hashes
 const SAPLING_SPEND_HASH: &str = "8270785a1a0d0bc77196f000ee6d221c9c9894f55307bd9357c3f0105d31ca63991ab91324160d8f53e2bbd3c2633a6eb8bdf5205d822e7f3f73edac51b2b70c";
 const SAPLING_OUTPUT_HASH: &str = "657e3d38dbb5cb5e7dd2970e8b03d69b4787dd907285b5a7f0790dcc8072f60bf593b32cc2d1c030e00ff5ae64bf84c5c3beb84ddc841d48264b4a171744d028";
 const SPROUT_HASH: &str = "e9b238411bd6c0ec4791e9d04245ec350c9c5744f5610dfcce4365d5ca49dfefd5054e371842b3f88fa1b9d7e8e075249b3ebabd167fa8b0f3161292d36c180a";
 
+// Circuit parameter file sizes
+const SAPLING_SPEND_BYTES: u64 = 47958396;
```
nuttycom, line 60: I have verified these sizes against my local copies of these files.

zcash_proofs/src/lib.rs
nuttycom:
```suggestion
/// while streaming from `hash_reader` into `sink`.
```

nuttycom (approved):
utACK

zcash_proofs/src/lib.rs
```diff
@@ -206,3 +431,81 @@ pub fn parse_parameters<R: io::Read>(
...
+        ));
+    }
+
+    Ok(())
+}
+
+/// Check if the Blake2b hash from `hash_reader` matches `expected_hash`,
+/// while streaming from `data` into `sink`.
+///
+/// `hash_reader` can be used to partially read `data`,
```
teor2345, line 465:
```suggestion
/// `hash_reader` can be used to partially read its inner reader's data,
```

zcash_proofs/src/downloadreader.rs
teor2345: Fixed in 9c331797

teor2345:
Thanks for the review, everything should be rebased and fixed now.

I ran `cargo test` locally, and everything passed.

teor2345:
I forgot to test with and without features, I pushed some minor fixes.

zcash_proofs/src/lib.rs
```diff
@@ -33,20 +33,48 @@ pub mod sprout;
 )]
 pub mod prover;
```
daira, line 35: Move the imports of `std::fs::File` and `std::io::BufReader` on lines 14-15 into the functions that use them, so that they're not unused imports when "download-params" and "local-prover" features are not enabled.

zcash_proofs/src/lib.rs
```diff
@@ -169,28 +375,48 @@ pub fn parse_parameters<R: io::Read>(
...
-        panic!("Sapling spend parameter file is not correct, please clean your `~/.zcash-params/` and re-run `fetch-params`.");
-    }
+    // TODO: use the correct paths for Windows and macOS
+    //       use the actual file paths supplied by the caller
+    verify_hash(
+        spend_fs,
+        &mut sink,
+        SAPLING_SPEND_HASH,
+        SAPLING_SPEND_BYTES,
+        SAPLING_SPEND_NAME,
```
daira, line 386: This function is compiled unconditionally, but `SAPLING_SPEND_NAME`, `SAPLING_OUTPUT_NAME`, and `SPROUT_NAME` are conditional on `any(feature = "local-prover", feature = "download-params")`.

daira (changes requested):
The use of constants that are not always compiled-in is blocking.

zcash_proofs/src/lib.rs
teor2345: Some functions were being compiled conditionally, but they should have always been available.

Fixed in 40d9115cc51bdfe4f55c986c30216ce65e8410ea

zcash_proofs/src/lib.rs
teor2345: They need to be unconditional now, because they are used in error messages.

Fixed in 40d9115cc51bdfe4f55c986c30216ce65e8410ea

teor2345: I was confused because compiling `librustzcash` with `--no-default-features` gives different results to `zcash_proofs` by itself, due to inter-crate dependencies. So I submitted a fix that didn't actually work.

I have tested this fix with all the features (and no features) on `zcash_proofs` by itself.

daira (approved):
utACK

PR #461, opened 2021-12-01
Factors out sapling address generation from SaplingExtendedFullViewingKey

nuttycom:
Sapling address generation depends only on an FVK and
a diversifier key. For the purpose of unified full viewing
keys, we need to break the dependence upon Sapling extended
full viewing keys, so this makes the address generation
methods of SaplingExtendedFullViewingKey usable without
coupling them to the ability to derive child full viewing keys.

### DIFF
--- a/zcash_primitives/src/zip32.rs
+++ b/zcash_primitives/src/zip32.rs
@@ -181,6 +181,42 @@ impl DiversifierKey {
     }
 }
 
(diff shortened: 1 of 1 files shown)


daira (approved):
utACK.

Codecov is saying this code is not covered by tests; it really should be (although that's not the responsibility of this PR).

PR #420, opened 2021-08-04
`zcash_address`: Parse unified viewing keys

therealyingtong:
Closes #372.

Based on https://github.com/zcash/librustzcash/pull/460.

PR #466, opened 2021-12-04
Unified receivers were inadvertently kept private

nuttycom:


nuttycom:
This was just a fragment of #420 that was necessary to keep zcashd/zcashd compiling.

PR #467, opened 2021-12-13
Add decryption of Sapling diversifiers.

nuttycom:
Given a diversifier key, decrypts a diversifier to obtain
the diversifier index from which the diversifier was originally
produced.

### DIFF
--- a/zcash_primitives/src/zip32.rs
+++ b/zcash_primitives/src/zip32.rs
@@ -149,6 +149,7 @@ impl DiversifierKey {
         d_j.copy_from_slice(&enc.to_bytes_le());
         let diversifier = Diversifier(d_j);
 
+        // validate that the generated diversifier maps to a jubjub subgroup point.
         diversifier.g_d().map(|_| diversifier)
     }
 
@@ -157,6 +158,18 @@ impl DiversifierKey {
     pub fn diversifier(&self, j: DiversifierIndex) -> Option<Diversifier> {
         let ff = FF1::<Aes256>::new(&self.0, 2).unwrap();
         Self::try_diversifier_internal(&ff, j)
+    }
+
+    /// Decrypts a diversifier using this diversifier key to obtain the
+    /// diversifier index from which it was originally created.
+    pub fn diversifier_index(&self, d: &Diversifier) -> DiversifierIndex {
+        let ff = FF1::<Aes256>::new(&self.0, 2).unwrap();
+        let dec = ff
+            .decrypt(&[], &BinaryNumeralString::from_bytes_le(&d.0[..]))
+            .unwrap();
+        let mut j = DiversifierIndex::new();
+        j.0.copy_from_slice(&dec.to_bytes_le());
+        j
     }
 
     /// Returns the first index starting from j that generates a valid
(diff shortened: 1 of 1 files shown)


zcash_primitives/src/zip32.rs
```diff
@@ -159,6 +160,18 @@ impl DiversifierKey {
         Self::try_diversifier_internal(&ff, j)
     }
 
+    /// Decrypts a diversifier using this diversifier key to obtain the
+    /// diversifier index from which it was originally created.
+    pub fn diversifier_index(&self, d: &Diversifier) -> Option<DiversifierIndex> {
+        let ff = FF1::<Aes256>::new(&self.0, 2).unwrap();
+        let dec = ff
+            .decrypt(&[], &BinaryNumeralString::from_bytes_le(&d.0[..]))
+            .ok()?;
```
str4d, line 169: We can unwrap here; `FF1::decrypt` only returns an error if the numeral string is in the incorrect radix, but we hard-code the matching radixes here (`2` and `BinaryNumeralString`):
```suggestion
            .unwrap();
```

zcash_primitives/src/zip32.rs
```diff
@@ -159,6 +160,18 @@ impl DiversifierKey {
         Self::try_diversifier_internal(&ff, j)
     }
 
+    /// Decrypts a diversifier using this diversifier key to obtain the
+    /// diversifier index from which it was originally created.
+    pub fn diversifier_index(&self, d: &Diversifier) -> Option<DiversifierIndex> {
```
str4d, line 165: And since we no longer use `?`, we can always return the diversifier index (unlike when deriving the diversifier, we know here that `Diversifier` is valid).
```suggestion
    pub fn diversifier_index(&self, d: &Diversifier) -> DiversifierIndex {
```

zcash_primitives/src/zip32.rs
```diff
@@ -159,6 +160,18 @@ impl DiversifierKey {
...
+        let ff = FF1::<Aes256>::new(&self.0, 2).unwrap();
+        let dec = ff
+            .decrypt(&[], &BinaryNumeralString::from_bytes_le(&d.0[..]))
+            .ok()?;
+        let mut j = DiversifierIndex::new();
+        j.0.copy_from_slice(&dec.to_bytes_le());
+        Some(j)
```
str4d, line 172:
```suggestion
        j
```

zcash_primitives/src/zip32.rs
```diff
@@ -578,6 +591,7 @@ mod tests {
         // j = 0
         let d_j = dk.diversifier(j_0).unwrap();
         assert_eq!(d_j.0, d_0);
+        assert_eq!(dk.diversifier_index(&Diversifier(d_0)), Some(j_0));
```
str4d, line 594:
```suggestion
        assert_eq!(dk.diversifier_index(&Diversifier(d_0)), j_0);
```

zcash_primitives/src/zip32.rs
```diff
@@ -588,6 +602,7 @@ mod tests {
         // j = 3
         let d_j = dk.diversifier(j_3).unwrap();
         assert_eq!(d_j.0, d_3);
+        assert_eq!(dk.diversifier_index(&Diversifier(d_3)), Some(j_3));
```
str4d, line 605:
```suggestion
        assert_eq!(dk.diversifier_index(&Diversifier(d_3)), j_3);
```

str4d (changes requested):
Generally LGTM.

zcash_primitives/src/zip32.rs
str4d: Opened https://github.com/str4d/fpe/issues/24 to clean up the API now that const generics are a thing.

PR #469, opened 2021-12-17
`zcash_note_encryption` API cleanups

str4d:


daira (approved):
utACK.
I think that `NoteValidity` could probably be removed in favour of booleans or `Choice`, but it's no longer public so that's not needed in this PR.

nuttycom (commented):
post-hoc utACK

PR #471, opened 2021-12-17
Update dependencies

str4d:


nuttycom (approved):
utACK. Confirmed that checkpoint serialization looks correct.

PR #474, opened 2022-01-04
Unified container parsing enforces typecode order.

nuttycom:


nuttycom:
UA test vectors need to be updated for the tests here to pass.

components/zcash_address/src/kind/unified.rs
```diff
@@ -286,15 +286,24 @@ pub(crate) mod private {
         /// items in their given order.
```
daira, line 286: This can be done more simply (IMHO) and efficiently without the `typecodes` HashSet:
```rust
        /// A private function that constructs a unified container with the
        /// specified items, which must be in ascending typecode order.
        fn try_from_items_internal(items: Vec<Self::Item>) -> Result<Self, ParseError> {
            assert!(u32::from(Typecode::P2sh) == u32::from(Typecode::P2pkh) + 1);

            let mut only_transparent = true;
            let mut prev_code = None; // less than any Some
            for item in &items {
                let t = item.typecode();
                let t_code = Some(u32::from(t));
                if t_code < prev_code {
                    return Err(ParseError::InvalidEncoding(
                        "Receivers out of order.".to_owned(),
                    ));
                } else if t_code == prev_code {
                    return Err(ParseError::DuplicateTypecode(t));
                } else if t == Typecode::P2sh && prev_code == Some(u32::from(Typecode::P2pkh)) {
                    // P2pkh and P2sh can only be in that order and next to each other,
                    // otherwise we would detect an out-of-order or duplicate typecode.
                    return Err(ParseError::BothP2phkAndP2sh);
                } else {
                    prev_code = t_code;
                    only_transparent = only_transparent && t.is_transparent();
                }
            }

            if only_transparent {
                Err(ParseError::OnlyTransparent)
            } else {
                // All checks pass!
                Ok(Self::from_inner(items))
            }
        }
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -327,21 +336,7 @@ pub trait Encoding: private::SealedContainer {
     /// * the item list may not contain two items having the same typecode
     /// * the item list may not contain only a single transparent item
```
daira, line 337:
```suggestion
    /// * the item list may not contain only transparent items
    /// * the item list may not contain both P2PKH and P2SH items.
```

components/zcash_address/src/kind/unified/address.rs
```diff
@@ -164,8 +164,9 @@ mod tests {
     }
 
     fn arb_unified_address_for_typecodes(
-        typecodes: Vec<Typecode>,
+        mut typecodes: Vec<Typecode>,
     ) -> impl Strategy<Value = Vec<Receiver>> {
+        typecodes.sort_unstable_by_key(|tc| <u32>::from(*tc));
```
daira, line 169: Should this be cloning `typecodes` before mutating it? It doesn't seem common for `arb_*` functions to mutate their arguments, and I'm not sure whether there are any special requirements to make shrinking work.

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -200,7 +208,9 @@ mod tests {
             shielded in arb_shielded_fvk(),
             transparent in prop::option::of(arb_transparent_fvk()),
         ) -> Ufvk {
-            Ufvk(shielded.into_iter().chain(transparent).collect())
+            let mut items = shielded.into_iter().chain(transparent).collect::<Vec<_>>();
```
daira, line 211: Style nit:
```suggestion
            let mut items: Vec<_> = shielded.into_iter().chain(transparent).collect();
```
(The turbofish is fine when needed, but I prefer to put the type up-front when I can.)

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -193,7 +201,9 @@ mod tests {
             shielded in arb_shielded_ivk(),
             transparent in prop::option::of(arb_transparent_ivk()),
         ) -> Uivk {
-            Uivk(shielded.into_iter().chain(transparent).collect())
+            let mut items = shielded.into_iter().chain(transparent).collect::<Vec<_>>();
```
daira, line 204: Same style nit here.

daira (approved):
utACK with suggestions.

components/zcash_address/src/kind/unified.rs
```diff
@@ -286,15 +286,24 @@ pub(crate) mod private {
...
             for item in &items {
                 let t = item.typecode();
                 if typecodes.contains(&t) {
                     return Err(ParseError::DuplicateTypecode(t));
+                } else if prev_typecode
+                    .iter()
+                    .any(|tc| <u32>::from(*tc) > <u32>::from(t))
+                {
+                    return Err(ParseError::InvalidEncoding(
+                        "Receivers out of order.".to_owned(),
```
daira, line 299:
```suggestion
                        "Items out of order.".to_owned(),
```
(This is used for UVKs as well.)

components/zcash_address/src/kind/unified/address.rs
nuttycom: Since this takes ownership, there's no need to clone it, right?

components/zcash_address/src/kind/unified/ivk.rs
nuttycom:
```suggestion
            let mut items: Vec<_> = shielded.into_iter().chain(transparent).collect();
```

nuttycom:
One style issue that I note this PR raises: we now have two canonical orderings for typecodes (preference order and encoding order) and as such it feels like the `Ord` instance that prioritizes preference order should instead just be a function that can be used with `sort_by`.

str4d (approved):
utACK. I verified that the updated test vectors here match the ones generated by https://github.com/zcash-hackworks/zcash-test-vectors/commit/61894e7793d29fc236697ee187e486d84a31d158.

PR #475, opened 2022-01-06
Derive Sapling internal keys.

therealyingtong:
As specified in https://github.com/zcash/zips/pull/576.

### DIFF
--- a/zcash_primitives/src/zip32.rs
+++ b/zcash_primitives/src/zip32.rs
@@ -544,6 +544,35 @@ impl ExtendedFullViewingKey {
     pub fn default_address(&self) -> (DiversifierIndex, PaymentAddress) {
         sapling_default_address(&self.fvk, &self.dk)
     }
+
+    pub fn derive_internal(&self) -> Self {
+        let i = Blake2bParams::new()
+            .hash_length(64)
+            .personal(crate::zip32::ZIP32_SAPLING_INT_PERSONALIZATION)
+            .hash(&self.fvk.to_bytes());
+        let i_nsk = jubjub::Fr::from_bytes_wide(prf_expand(i.as_bytes(), &[0x17]).as_array());
+        let r = prf_expand(i.as_bytes(), &[0x18]);
+        let r = r.as_bytes();
+        // PROOF_GENERATION_KEY_GENERATOR = \mathcal{H}^Sapling
+        let nk_internal = PROOF_GENERATION_KEY_GENERATOR * i_nsk + self.fvk.vk.nk;
+        let dk_internal = DiversifierKey(r[..32].try_into().unwrap());
+        let ovk_internal = OutgoingViewingKey(r[32..].try_into().unwrap());
+
+        ExtendedFullViewingKey {
+            depth: self.depth,
+            parent_fvk_tag: self.parent_fvk_tag,
+            child_index: self.child_index,
+            chain_code: self.chain_code,
+            fvk: FullViewingKey {
+                vk: ViewingKey {
+                    ak: self.fvk.vk.ak,
+                    nk: nk_internal,
+                },
+                ovk: ovk_internal,
+            },
+            dk: dk_internal,
+        }
+    }
 }
 
 #[cfg(test)]
--- a/zcash_primitives/src/zip32.rs
+++ b/zcash_primitives/src/zip32.rs
@@ -6,6 +6,7 @@ use aes::Aes256;
 use blake2b_simd::Params as Blake2bParams;
 use byteorder::{ByteOrder, LittleEndian, ReadBytesExt, WriteBytesExt};
 use fpe::ff1::{BinaryNumeralString, FF1};
+use std::convert::TryInto;
 use std::ops::AddAssign;
 
 use crate::{
@@ -20,6 +21,7 @@ use crate::sapling::keys::{
 
 pub const ZIP32_SAPLING_MASTER_PERSONALIZATION: &[u8; 16] = b"ZcashIP32Sapling";
 pub const ZIP32_SAPLING_FVFP_PERSONALIZATION: &[u8; 16] = b"ZcashSaplingFVFP";
+pub const ZIP32_SAPLING_INT_PERSONALIZATION: &[u8; 16] = b"Zcash_SaplingInt";
 
 // Common helper functions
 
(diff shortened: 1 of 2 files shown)


zcash_primitives/src/zip32.rs
```diff
@@ -409,6 +411,35 @@ impl ExtendedSpendingKey {
...
         ExtendedFullViewingKey::from(self).default_address()
     }
+
+    pub fn derive_internal(&self) -> Self {
+        let i = {
+            let fvk = FullViewingKey::from_expanded_spending_key(&self.expsk);
+            Blake2bParams::new()
+                .hash_length(64)
+                .personal(crate::zip32::ZIP32_SAPLING_INT_PERSONALIZATION)
+                .hash(&fvk.to_bytes())
```
nuttycom, line 421: This isn't correct; `fvk.to_bytes()` doesn't include the encoding of `dk`. This should be `EncodeExtFVKParts(ak, nk, ovk, dk)`

nuttycom (changes requested):
The computation of `I` needs to be fixed to include `dk` for both IXSK and IFVK derivation.

zcash_primitives/src/zip32.rs
```diff
@@ -513,6 +544,35 @@ impl ExtendedFullViewingKey {
...
     }
+
+    pub fn derive_internal(&self) -> Self {
+        let i = Blake2bParams::new()
+            .hash_length(64)
+            .personal(crate::zip32::ZIP32_SAPLING_INT_PERSONALIZATION)
+            .hash(&self.fvk.to_bytes());
+        let i_nsk = jubjub::Fr::from_bytes_wide(prf_expand(i.as_bytes(), &[0x17]).as_array());
+        let r = prf_expand(i.as_bytes(), &[0x18]);
+        let r = r.as_bytes();
```
nuttycom, line 555:
```suggestion
        let r = prf_expand(i.as_bytes(), &[0x18]).as_bytes();
```

defuse (approved):
utACK from @daira and me.

str4d:
**DO NOT MERGE THIS**

str4d (changes requested):
We messed up merging #472 directly into the `master` branch. It altered the NU5 consensus rules, but did _not_ alter the consensus branch ID (which needs to happen whenever this changes). But whatever we change the consensus branch ID to will change _again_ when the `orchard` dependency is updated.

Basically, now that we've messed up, anything that `zcashd` needs outside of the NU5 consensus changes _must_ be based on a commit prior to #472 merging. I'm going to set up an "anti-feature" branch, treating `master` as the NU5 consensus rules feature branch for the time being. Once this "anti-feature" branch is up, this PR needs to be rebased on it to exclude the #472 changes.

str4d:
Okay, I've changed the base of this PR to target the new `non-consensus-changes-on-branchid-37519621` branch. This PR now needs to be rebased to remove the commits from #472, so it only includes the three commits it is intending to add.

zcash_primitives/src/zip32.rs
```diff
@@ -233,6 +235,41 @@ pub fn sapling_default_address(
...
 
+/// Returns the internal full viewing key and diversifier key
+/// for the provided external fvk and dk
+pub fn sapling_derive_internal_fvk(
+    fvk: &FullViewingKey,
+    dk: &DiversifierKey,
+) -> (FullViewingKey, DiversifierKey) {
+    let i = {
+        let mut h = Blake2bParams::new()
+            .hash_length(64)
```
str4d, line 246: ZIP 32 specifies `BLAKE2b-256` here:
```suggestion
            .hash_length(32)
```
This bug should be caught by test vectors.

zcash_primitives/src/zip32.rs
```diff
@@ -233,6 +235,41 @@ pub fn sapling_default_address(
     sapling_find_address(fvk, dk, DiversifierIndex::new()).unwrap()
 }
 
+/// Returns the internal full viewing key and diversifier key
+/// for the provided external fvk and dk
+pub fn sapling_derive_internal_fvk(
```
str4d, line 240: Document what this API is for (specifically implementing UFVKs, since otherwise this would be duplicative of `ExtendedFullViewingKey::derive_internal`).

zcash_primitives/src/zip32.rs
```diff
@@ -409,6 +446,38 @@ impl ExtendedSpendingKey {
     pub fn default_address(&self) -> (DiversifierIndex, PaymentAddress) {
         ExtendedFullViewingKey::from(self).default_address()
     }
+
+    pub fn derive_internal(&self) -> Self {
+        let i = {
+            let fvk = FullViewingKey::from_expanded_spending_key(&self.expsk);
+            let mut h = Blake2bParams::new()
+                .hash_length(64)
```
str4d, line 454: ZIP 32 specifies `BLAKE2b-256` here:
```suggestion
                .hash_length(32)
```
This bug should be caught by test vectors.

zcash_primitives/src/zip32.rs
```diff
@@ -409,6 +446,38 @@ impl ExtendedSpendingKey {
     pub fn default_address(&self) -> (DiversifierIndex, PaymentAddress) {
         ExtendedFullViewingKey::from(self).default_address()
     }
+
+    pub fn derive_internal(&self) -> Self {
```
str4d, line 450: Document this (yes the existing documentation is sparse; we should improve it as we go).

zcash_primitives/src/zip32.rs
```diff
@@ -513,6 +582,19 @@ impl ExtendedFullViewingKey {
     pub fn default_address(&self) -> (DiversifierIndex, PaymentAddress) {
         sapling_default_address(&self.fvk, &self.dk)
     }
+
+    pub fn derive_internal(&self) -> Self {
```
str4d, line 586: Document this.

str4d (changes requested):
Two blocking implementation bugs.

zcash_primitives/src/zip32.rs
nuttycom: We should add the `DiversifiableFullViewingKey` type here as we have in zcashd, and then this should be a member function of that type. If we had that, then the `ExtendedFullViewingKey` impl could be removed.

nuttycom: Also, it's not critical for the moment but now that zip32 has non-Sapling stuff in it, I'd really like this to be moved into the Sapling module.

nuttycom (approved):
re-utACK https://github.com/zcash/librustzcash/pull/475/commits/eb80138cf9cb1fab1bdb1006cdd5013068d580f8

str4d (approved):
utACK

daira (approved):
utACK.

Note: All of the constructions of the form "I<sub>foo</sub> = ToScalar<sup>Sapling</sup>(PRF<sup>expand</sup>(I, [fixed_domain_separator]))", where I is 256 bits, have a slight entropy bottleneck. This is because ToScalar<sup>Sapling</sup> ∘ PRF<sup>expand</sup> behaves as a random function from 256 bits to the Jubjub scalar field, but 256 bits is only a few bits larger than the scalar field, so this function is expected to have some collisions. That is, the resulting scalar distribution will not be *statistically* indistinguishable from a uniform distribution.

However, I don't see how this could be exploitable with work less than the 2<sup>125</sup> target security level, or remotely close to it, so I think it's fine. (We couldn't change this for the existing ZIP 32 key derivations which also have this slight entropy bottleneck, anyway.)

Orchard doesn't have this issue at all, because we're deriving (ask, nk, rivk) from (sk, c) which has 512 bits of entropy, not 256, and there is no entropy bottleneck in this derivation (sk is in the PRF<sup>expand</sup> key and c in the input).

Edit: I computed the Shannon entropy of the ToScalar<sup>Sapling</sup> ∘ PRF<sup>expand</sup> output; it is 251.815847 bits, compared to 251.857089 bits for a uniformly random scalar. So the entropy loss is not significant. Here's the code to compute that: https://github.com/daira/zcash-security/blob/main/sapling_entropy.py

PR #480, opened 2022-01-11
[ZIP 316] Transparent internal and external ovk

therealyingtong:
Implements ZIP 316 changes as specified in https://github.com/zcash/zips/pull/576/.

This is based on the autoshielding branch https://github.com/nuttycom/librustzcash/tree/autoshield-poc-daa, i.e. on top of #341.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -2,7 +2,7 @@ use std::fmt::Debug;
 use zcash_primitives::{
     consensus::{self, NetworkUpgrade},
     memo::MemoBytes,
-    sapling::prover::TxProver,
+    sapling::{keys::OutgoingViewingKey, prover::TxProver},
     transaction::{
         builder::Builder,
         components::{amount::DEFAULT_FEE, Amount},
@@ -398,13 +398,14 @@ where
         .and_then(|x| x.ok_or_else(|| Error::ScanRequired.into()))?;
 
     // derive the t-address for the extpubkey at child index 0
-    let taddr = sk.to_external_pubkey().to_address();
+    let t_ext_pubkey = sk.to_external_pubkey();
+    let taddr = t_ext_pubkey.to_address();
+    let ovk = OutgoingViewingKey(t_ext_pubkey.internal_ovk().as_bytes());
 
     // derive own shielded address from the provided extended spending key
     // TODO: this should become the internal change address derived from
     // the wallet's UFVK
     let z_address = extfvk.default_address().1;
-    let ovk = extfvk.fvk.ovk;
 
     // get UTXOs from DB
     let utxos = wallet_db.get_unspent_transparent_outputs(&taddr, latest_anchor)?;
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -8,6 +8,7 @@ use zcash_primitives::{
         components::{amount::DEFAULT_FEE, Amount},
         Transaction,
     },
+    transparent,
     zip32::{ExtendedFullViewingKey, ExtendedSpendingKey},
 };
 
@@ -20,9 +21,6 @@ use crate::{
     wallet::{AccountId, OvkPolicy},
     zip321::{Payment, TransactionRequest},
 };
-
-#[cfg(feature = "transparent-inputs")]
-use crate::keys::transparent;
 
 /// Scans a [`Transaction`] for any information that can be decrypted by the accounts in
 /// the wallet, and saves it to the wallet.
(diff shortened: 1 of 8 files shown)


nuttycom (changes requested):
The source of the chain code in `shield_transparent_funds` needs to be fixed, otherwise this is looking good.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -407,12 +409,17 @@ where
         .get_target_and_anchor_heights(min_confirmations)
         .and_then(|x| x.ok_or_else(|| Error::ScanRequired.into()))?;
 
-    // derive the corresponding t-address
+    // derive the corresponding transparent internal ovk
+    let t_fvk: transparent::FullViewingKey = {
+        let secp = secp256k1::Secp256k1::new();
+        let pk = secp256k1::PublicKey::from_secret_key(&secp, sk);
+        (extfvk.chain_code(), pk).into()
```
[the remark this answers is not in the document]
nuttycom: Actually, as it turns out I'm making changes to #341 that will make it so that the chain code is available, so I will resolve this issue once that's done.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -385,6 +387,7 @@ pub fn shield_transparent_funds<E, N, P, D, R>(
     wallet_db: &mut D,
     params: &P,
     prover: impl TxProver,
+    chain_code: ChainCode,
     sk: &secp256k1::SecretKey,
```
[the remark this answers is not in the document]
nuttycom: Oh. This is just the `ExtendedPrivKey` type from `hdwallet`.

zcash_primitives/src/zip316/transparent.rs
```diff
@@ -0,0 +1,99 @@
...
+        SaplingOvk(t_int_ovk.0)
+    }
+}
+
+/// Finds the internal or external ovk for the transparent component
+/// of a given Ufvk, if it exists.
+pub fn find_ovk(key: &Ufvk, external: bool) -> Option<[u8; 32]> {
+    for r in &key.items() {
+        if let Fvk::P2pkh(data) = r {
+            let fvk: FullViewingKey = (*data).try_into().ok()?;
```
daira, line 89: This is similar but different to https://github.com/zcash/zcash/pull/5470/files#diff-98a0abefb7ae486bd76f5e8d28d7fd65cead55c7344b31cc90caa56583917ca3R155-R159 . Why the difference?

zcash_client_backend/src/data_api/wallet.rs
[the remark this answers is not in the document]
daira: Add a TODO and/or use a chain code that is obviously wrong, like all-zeros, for now.

zcash_primitives/src/zip316/transparent.rs
daira: Oh it's just because the `TryInto` implementation above fills in the gap.

daira (approved):
utACK modulo comments.

zcash_primitives/src/lib.rs
```diff
@@ -17,6 +17,7 @@ pub mod memo;
 pub mod merkle_tree;
 pub mod sapling;
 pub mod transaction;
+pub mod transparent;
```
nuttycom, line 20: This module should still be behind the `transparent-inputs` feature flag, for those wallet vendors that do not choose to include transparent support, we don't want to require it.

zcash_primitives/src/transparent.rs
```diff
@@ -42,4 +43,54 @@ impl ExternalPubKey {
...
+    /// Returns the chain code component of this public key.
+    pub fn chain_code(&self) -> &[u8] {
+        &self.0.chain_code
+    }
+
+    /// Derives the internal ovk and external ovk corresponding to this
+    /// transparent fvk. As specified in [ZIP 316][transparent-ovk].
+    ///
+    /// [transparent-ovk]: https://zips.z.cash/zip-0316#deriving-internal-keys
+    fn ovk_for_shielding(&self) -> (InternalOvk, ExternalOvk) {
```
nuttycom, line 65: This method name confused me, as the OVK for shielding is just the internal part. I might just call this `derive_ovks` instead.

nuttycom (approved):
ACK

zcash_primitives/src/transparent.rs
```diff
@@ -0,0 +1,105 @@
...
```
str4d, line 65: This is a bug in the PR: `ExternalPubKey` is derived at the address level, but ZIP 316 specifies that the internal and external OVKs must be derived from the transparent FVK, which is at the account level.

PR #495, opened 2022-02-01
Restrict cargo to `protobuf >=2.20,<2.26`

str4d:
`protobuf 2.26.0` bumped its MSRV to 1.52.1, which is incompatible with
our current MSRV.


nuttycom (approved):
utACK

PR #501, opened 2022-02-01
Fix Clippy beta linting complaints.

nuttycom:


zcash_primitives/src/transaction/components/amount.rs
```diff
@@ -200,6 +200,12 @@ impl Sum<Amount> for Option<Amount> {
     }
 }
 
+impl<'a> Sum<&'a Amount> for Option<Amount> {
```
str4d, line 203: Where was this needed in this PR? Also add this impl to the changelog under `Added`:
```
- `impl Sum<&Amount> for Option<Amount>`
```

nuttycom: This was needed due to this change: https://github.com/zcash/librustzcash/pull/501/files#diff-536864bb9b96c14c23f82e72d025e944216c4e1305836eae30dc1e7c38fac7a7L270-R269

Clippy identified `array::IntoIter::new` as being deprecated.

str4d: Ah yeah, that's because we were forced to bump MSRV to 1.54 recently, and that was deprecated in 1.53. But we can't fully migrate to `array_var.into_iter()` until we bump to MSRV 1.56 and enable 2021 edition. So yeah, this isn't technically needed once we get to 1.56 because we can just use the by-value sum, but this is fine.

str4d (approved):
utACK a304c3953b5a58f0d512cb23de8997b1863e0405

PR #504, opened 2022-02-02
Merge all non-consensus changes staged for NU5 to `master`.

nuttycom:


str4d (approved):
utACK 60d9124453558acb658e85d829a07f6dae94c6ef. I confirmed that this is merging the correct non-consensus commit (4f4a25252f0fc6b84c44316d810107bc7ea7f32a).

daira (commented):
Post-hoc utACK cdd899dda181cbda1b1ffa20142ad5c86f519baf..60d9124453558acb658e85d829a07f6dae94c6ef, which according to @nuttycom is the correct range of commits for the merge resolution. (I didn't check it is the correct range.)

PR #507, opened 2022-02-03
Define transaction::TransactionData::map_authorization

nuttycom:


zcash_primitives/src/transaction/mod.rs
```diff
@@ -370,6 +370,68 @@ impl<A: Authorization> TransactionData<A> {
...
+            transparent_bundle: f_transparent(self.transparent_bundle),
+            sprout_bundle: self.sprout_bundle,
+            sapling_bundle: f_sapling(self.sapling_bundle),
+            orchard_bundle: f_orchard(self.orchard_bundle),
+            #[cfg(feature = "zfuture")]
+            tze_bundle: f_tze(self.tze_bundle),
+        }
+    }
+
+    pub fn map_authorization<B: Authorization>(
```
str4d, line 407: I think this one is technically not needed anymore now that `TransactionData::map_bundles` exists? But I don't mind it being here.

str4d (approved):
ACK (aside from the `TransactionData::map_bundles` method that I added).

nuttycom (commented):
ACK https://github.com/zcash/librustzcash/pull/507/commits/54e3dded92ccd42ccfb669f056523a2e157e8e7a (ACK as comment because I was the original PR author)

PR #510, opened 2022-02-11
Update unified address test vectors.

nuttycom:
This updates the UA test vectors after https://github.com/zcash-hackworks/zcash-test-vectors/pull/81


daira (commented):
I would be surprised if this passes: some of the test vector entries now contain unknown items, but the current test will drop those items, so won't construct the same serialized UA.

daira:
Oh! This is a bug in zcash-hackworks/zcash-test-vectors#81! I added the optional unknown item to the test vector fields, but forgot to include it in the list of items to be serialized in the Python code. (This bug is only in `unified_address.py`, not in `unified_full_viewing_keys.py` or `unified_incoming_viewing_keys.py`.)

I will put up another PR on zcash-test-vectors to fix it, and then this test will also have to be changed.

daira:
> I will put up another PR on zcash-test-vectors to fix it, and then this test will also have to be changed.

This is zcash-hackworks/zcash-test-vectors#82.

daira (approved):
utACK. I have not checked that this matches the content of the test vectors generated by https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/zcash_test_vectors/unified_address.py (but it does match the format as updated by https://github.com/zcash-hackworks/zcash-test-vectors/pull/82).

PR #512, opened 2022-02-15
CI: Run tests with cargo-nextest

str4d:


str4d:
Hmm, this actually ends up taking longer, because for some reason the sequence of compiles it triggers takes longer overall, plus we have several minutes compiling and installing `cargo-nextest` itself. So I'm putting this on the backburner until we can figure out build efficiencies.

sunshowers:
BTW nextest is now available as pre-built binaries, if you're interested in using them: https://nexte.st/book/pre-built-binaries

.github/workflows/ci.yml
```diff
@@ -8,14 +8,24 @@ jobs:
     runs-on: ${{ matrix.os }}
```
sunshowers, line 8: You'll want to change this to `matrix.os.runs-on` btw

str4d:
(Below I ignore the 2m 50s for building the param-fetching binary, which is unchanged)

[Without this PR](https://github.com/zcash/librustzcash/runs/5325128468?check_suite_focus=true):
- Run tests and doc tests (7m 34s)
  - 6m 06s building, 1m 28s testing.
- Run slow tests (1m 41s)
  - 0.15s building, so basically all testing.
- Total build time: 6m 06s
- Total test time: 3m 09s

[With this PR](https://github.com/zcash/librustzcash/runs/5342327856?check_suite_focus=true):
- Run tests and slow tests (8m 56s)
  - 6m 06s building, 2m 50s testing.
- Run doc tests (3m 52s)
  - 3m 39s building, 13s testing.
- Total build time: 9m 45s
- Total test time: 3m 03s

Switching to `nextest` saves us just over 3% during testing (6 seconds), but increases build times by almost 60% due to the need to recompile with doctests enabled. I'm a little surprised by this, as I saw a 25% reduction locally, but I suspect this is due to overcontention on the more resource-constrained CI builders (since our proving tests are internally running multithreaded code). In any case, it's not worth us merging this PR unless `nextest` has a way to build the tests with doc tests enabled (even if it can't run them due to [limitations in stable Rust](https://github.com/nextest-rs/nextest/issues/16)).

nathan-at-least:
Speculation: is there any way we can do sophisticated / fine-grained build caching in CI, since that's the bottleneck?

I've experimented with a few cargo build caching github actions without much luck. I think I just don't understand gh CI well enough to understand how to do it well.

sunshowers:
Maintainer of nextest here -- it's a bit surprising that doctests cause a rebuild because I don't see that on my own projects, it's worth digging into why that is happening. Wondering if there's something related to proc macros going on (not sure if you have any).

> In any case, it's not worth us merging this PR unless nextest has a way to build the tests with doc tests enabled

Sadly the two are related -- `cargo test --doc` is just a really special snowflake.

Issue #456, opened 2021-11-18
Update orchard crate and NU5 branch ID, and set activation heights


str4d:
We will be rolling back testnet alongside the mainnet activation definition, because we are making final cleanup changes to the Orchard Action circuit.

daira:
This requires updates to take account of zcash/orchard#305.

steven-ecc:
Done

PR #514, opened 2022-02-15
Migrate to `orchard` revision with circuit change

str4d:
This also bumps MSRV for `zcash_primitives` and below to 1.54.0, and fixes up the `protobuf` dependency (now that we are above its latest MSRV).

Part of #456.


nuttycom (approved):
utACK

Cargo.toml
```diff
@@ -19,6 +19,7 @@ panic = 'abort'
 codegen-units = 1
 
 [patch.crates-io]
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "f1795f8068c8c83af89af8a061131759f973d656" }
```
nuttycom, line 22: I have verified that this matches the `HEAD` of Orchard.

ebfull (approved):

daira (commented):
Post-hoc ACK

PR #515, opened 2022-02-17
zcash_client_backend: Bring in latest `compact_formats.proto`

str4d:
Source: https://github.com/zcash/lightwalletd/blob/83bb19918fb36a5f6d5013fba7938daf3d73db1d/walletrpc/compact_formats.proto

### DIFF
--- a/zcash_client_backend/proto/compact_formats.proto
+++ b/zcash_client_backend/proto/compact_formats.proto
@@ -1,6 +1,11 @@
+// Copyright (c) 2019-2021 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
 syntax = "proto3";
 package cash.z.wallet.sdk.rpc;
 option go_package = "walletrpc";
+option swift_prefix = "";
 
 // Remember that proto3 fields are all optional. A field that is not present will be set to its zero value.
 // bytes fields of hashes are in canonical little-endian format.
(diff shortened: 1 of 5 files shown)


zcash_client_backend/proto/compact_formats.proto
```diff
@@ -33,16 +41,30 @@ message CompactTx {
...
-    bytes cmu = 1;
-    bytes epk = 2;
-    bytes ciphertext = 3;
+// https://github.com/zcash/zips/blob/main/zip-0225.rst#orchard-action-description-orchardaction
+// (but not all fields are needed)
+message CompactOrchardAction {
+    bytes nullifier = 1;        // [32] The nullifier of the input note
+    bytes cmx = 2;              // [32] The x-coordinate of the note commitment for the output note
+    bytes ephemeralKey = 3;     // [32] An encoding of an ephemeral Pallas public key
+    bytes encCiphertext = 4;    // [580] The encrypted contents of the note plaintext
```
str4d, line 69: @LarryRuane why was this field documented in https://github.com/zcash/lightwalletd/pull/371 as being the full 580 bytes? We only plan to send the first 52 bytes of it, i.e. as documented in https://zips.z.cash/zip-0244#t-4a-orchard-actions-compact-digest

```suggestion
    bytes ciphertext = 4;       // [52] The note plaintext component of the encCiphertext field
```

LarryRuane: Ah yes, I vaguely remember that at one time the lightwalletd code thought that this field of the compact format comprised all 580 bytes, then I discovered that's incorrect, but I forgot to update this comment. I just went ahead and merged this PR to fix the comment: https://github.com/zcash/lightwalletd/pull/377 so please pull a copy of `compact_formats.proto` from master again.

zcash_client_backend/proto/compact_formats.proto
```diff
@@ -33,16 +41,30 @@ message CompactTx {
     //    valueBalance + (sum(vPubNew) - sum(vPubOld) - sum(tOut))
```
daira, line 41:
```
    // If there are no transparent inputs, the fee will be calculable as:
    //    valueBalanceSapling + valueBalanceOrchard + sum(vPubNew) - sum(vPubOld) - sum(tOut)

File: zcash_client_backend/proto/compact_formats.proto (hunk 1 above)
Comment (daira) [reply zcash_client_backend/proto/compact_formats.proto:69]: The suggested description is not quite accurate, but "// first 52 bytes of ciphertext" as used in https://github.com/zcash/lightwalletd/pull/377 is.

("Note plaintext" includes the memo; also, it's not accurate to refer to components of `encCiphertext`.)

File: zcash_client_backend/proto/compact_formats.proto [hunk 3]
```diff
@@ -33,16 +41,30 @@ message CompactTx {
...
+// CompactSaplingSpend is a Sapling Spend Description as described in 7.3 of the Zcash
+// protocol specification.
+message CompactSaplingSpend {
+    bytes nf = 1;   // nullifier (see the Zcash protocol specification)
 }
 
-message CompactSpend {
-    bytes nf = 1;
+// output is a Sapling Output Description as described in section 7.4 of the
+// Zcash protocol spec. Total size is 948.
```
daira, line 56: The size of an OutputDescriptionV4 as encoded in a v4 transaction is 948 bytes. The size of an OutputDescriptionV5 as encoded in a v5 transaction is 756 bytes. The size of a CompactSaplingOutput as encoded here is 116 bytes. The latter does not include `cv`, `outCiphertext`, or `zkproof` fields, as well as truncating the `encCiphertext` field.

Also, there is an important semantic distinction between "epk" and `ephemeralKey` in the spec: `ephemeralKey` is a sequence of 32 bytes (which prior to activation of ZIP 214 with Canopy could be a non-canonical encoding), whereas epk refers to the abstract public key. Please consider renaming this field (I'm not including that in the suggestion because I know it requires other Go code changes).

```suggestion
// output encodes the `cmu` field, `ephemeralKey` field (here inaccurately called
// `epk`), and a 52-byte prefix of the `encCiphertext` field of a Sapling Output
// Description. These fields are described in section 7.4 of the Zcash protocol spec.
// Total size is 116 bytes.
```

daira (approved):
utACK with suggestions.

zcash_client_backend/proto/compact_formats.proto
str4d: > Please consider renaming this field (I'm not including that in the suggestion because I know it requires other Go code changes).

We can easily do this, because fields are identified by their index, not their name.

str4d:
Oops, I'd meant to merge this into the non-consensus branch (which I based it on). Oh well, the difference between the two shouldn't persist for too much longer, I hope.

PR #526, opened 2022-03-23
Update NU5 consensus branch ID and remove testnet activation height

str4d:
Part of zcash/librustzcash#456.


str4d:
Test vector updates are from https://github.com/zcash-hackworks/zcash-test-vectors/pull/88.

zcash_primitives/src/consensus.rs
```diff
@@ -240,7 +240,7 @@ impl Parameters for TestNetwork {
             NetworkUpgrade::Blossom => Some(BlockHeight(584_000)),
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_800)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_028_500)),
-            NetworkUpgrade::Nu5 => Some(BlockHeight(1_599_200)),
+            NetworkUpgrade::Nu5 => None,
```
str4d, line 243: This is unset for now because this height doesn't apply to the new consensus branch ID. We will set it to the actual height just before 4.7.0.

therealyingtong (approved):
utACK

daira (approved):
utACK. Note that this doesn't include an intended rewind (see 8dcb533226557041c0b745566057131a1178959b for example) back to height 1599199.

str4d:
The rewind logic is handled in `zcash/zcash`.

PR #534, opened 2022-04-08
Add conversion from incrementalmerkletree::bridgetree::Frontier -> CommitmentTree

### LINKED ISSUE #533: Add `CommitmentTree::from_frontier` method
This is the inverse of `CommitmentTree::to_frontier`, and will enable converting from the dense frontier format used by `incrementalmerkletree` to the sparse format used by this crate (and by the current `z_gettreestate` RPC output).

nuttycom:
Closes #533.

### DIFF
--- a/zcash_primitives/src/merkle_tree.rs
+++ b/zcash_primitives/src/merkle_tree.rs
@@ -1,13 +1,17 @@
 //! Implementation of a Merkle tree of commitments used to prove the existence of notes.
 
 use byteorder::{LittleEndian, ReadBytesExt};
-use incrementalmerkletree::{self, bridgetree, Altitude};
+use incrementalmerkletree::{
+    self,
+    bridgetree::{self, Leaf},
+    Altitude,
+};
 use std::collections::VecDeque;
 use std::convert::TryFrom;
 use std::io::{self, Read, Write};
 use zcash_encoding::{Optional, Vector};
 
-use crate::sapling::{SAPLING_COMMITMENT_TREE_DEPTH, SAPLING_COMMITMENT_TREE_DEPTH_U8};
+use crate::sapling::SAPLING_COMMITMENT_TREE_DEPTH;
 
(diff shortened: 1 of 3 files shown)


daira (approved):
utACK

ebfull (approved):

zcash_primitives/src/merkle_tree.rs
```diff
@@ -117,7 +121,35 @@ impl<Node> CommitmentTree<Node> {
...
+            let (left, right) = match f.leaf() {
+                Leaf::Left(v) => (Some(v.clone()), None),
+                Leaf::Right(l, r) => (Some(l.clone()), Some(r.clone())),
+            };
+            let mut ommers_iter = f.ommers().iter().cloned();
+            let upos: usize = f.position().into();
+            Self {
+                left,
+                right,
+                parents: (1..DEPTH)
```
str4d, line 138: I was concerned there might be an off-by-one here, but I now believe this to be correct. `parents.len()` is at most `depth - 1`, because the final layer in the depth is stored in `left / right`. Starting from `1` is correct here because the first bit of `upos` corresponds to `left / right`.

zcash_primitives/src/merkle_tree.rs
```diff
@@ -117,7 +121,35 @@ impl<Node> CommitmentTree<Node> {
...
+                    .into_iter()
+                    .map(|i| {
+                        if upos & (1 << i) == 0 {
+                            None
+                        } else {
+                            ommers_iter.next()
+                        }
```
str4d, line 145: I confirmed that this lines up with how `CommitmentTree::size` is implemented:

https://github.com/zcash/librustzcash/blob/43c18d000fcbe45363b2d53585d5102841eff99e/zcash_primitives/src/merkle_tree.rs#L156-L160

str4d (approved):
utACK

PR #538, opened 2022-04-11
Add `ZcashAddress::encode` method.

### LINKED ISSUE #463: `zcash_address`: Add explicit string encoding method to `ZcashAddress`
Currently `ZcashAddress` follows the standard Rust approach for string encodings: it provides implementations of the stdlib traits `FromStr` (which enables `string.parse()` to work) and `Display` (which provides `ZcashAddress::to_string()`).

@nuttycom pointed out that `Display` doesn't always get used for an encoding format; it can sometimes mean "pretty-print this please", and there's no good way to distinguish the two. What `Display` does for a given struct is always fixed by the struct author (as foreign traits can't be impled for foreign types), but it's a fair point that it would be useful to have another way to discover this, for usability.

We should add a method on `ZcashAddress` that can place the serialization documentation front-and-centre while also pointing to `Display`. This is equivalent to when we added `ZcashAddress::try_from_encoded`: https://github.com/zcash/librustzcash/commit/c7fcee27a229ea55ce20738c606c719628d791bf

nuttycom:
This provides the encoding corresponding to
`ZcashAddress::try_from_encoded` and documents the fact that the
`Display` instance can also provide this encoding.

Fixes #463

### DIFF
--- a/components/zcash_address/src/lib.rs
+++ b/components/zcash_address/src/lib.rs
@@ -41,6 +41,19 @@ enum AddressKind {
 }
 
 impl ZcashAddress {
+    /// Encodes this Zcash address in its canonical string representation.
+    ///
+    /// This provides the encoded string representation of the address as defined by the
+    /// [Zcash protocol specification](https://zips.z.cash/protocol.pdf) and/or
+    /// [ZIP 316](https://zips.z.cash/zip-0316). The [`Display` implementation] can also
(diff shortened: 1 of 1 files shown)


daira (approved):
utACK

str4d (commented):
Post-hoc utACK.

PR #547, opened 2022-05-04
Allow CommitmentTree completeness & auth paths to be computed at arbitrary depths.

### LINKED ISSUE #546: Fix the trailing parents in `CommitmentTree`
The function [from_frontier()](https://github.com/zcash/librustzcash/blob/master/zcash_primitives/src/merkle_tree.rs#L124) always produces `CommitmentTree`s with `const DEPTH - 1` parents.

---

I noticed this when looking at serialized `CommitmentTree`s produced by [write()](https://github.com/zcash/librustzcash/blob/master/zcash_primitives/src/merkle_tree.rs#L219). For example, the serialized tree at height 419201 looks like this:
```
                                       0xf1 is 31, which is the length of the vector with the parents.
                                                                    vv                  
019eb30778ddeea84c72e69e07a1689f3c8def3dc0a1939f0edcbe47279069d931001f000150715810d52caf35471d10feb487213fbd95ff209122225b7b65d27a7fb1a44d0000000000000000000000000000000000000000000000000000000000
```

On the other hand, `zcashd` returns:
```
                                                   The length of the parents vector is 2.
                                                                    vv
019eb30778ddeea84c72e69e07a1689f3c8def3dc0a1939f0edcbe47279069d9310002000150715810d52caf35471d10feb487213fbd95ff209122225b7b65d27a7fb1a44d
```

---

The trees are the same except for the trailing zeroes, which is caused by the number of parents. This is likely to cause issues at least on the following places:
https://github.com/zcash/librustzcash/blob/master/zcash_primitives/src/merkle_tree.rs#L199,
https://github.com/zcash/librustzcash/blob/master/zcash_primitives/src/merkle_tree.rs#L296.

nuttycom:
Fixes #546

### DIFF
--- a/zcash_primitives/src/lib.rs
+++ b/zcash_primitives/src/lib.rs
@@ -5,7 +5,7 @@
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
 // Catch documentation errors caused by code changes.
-#![deny(broken_intra_doc_links)]
+#![deny(rustdoc::broken_intra_doc_links)]
 // Temporary until we have addressed all Result<T, ()> cases.
 #![allow(clippy::result_unit_err)]
 
--- a/zcash_primitives/src/merkle_tree.rs
+++ b/zcash_primitives/src/merkle_tree.rs
@@ -194,10 +194,14 @@ impl<Node> CommitmentTree<Node> {
     }
 
     fn is_complete(&self, depth: usize) -> bool {
-        self.left.is_some()
-            && self.right.is_some()
-            && self.parents.len() == depth - 1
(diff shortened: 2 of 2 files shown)


upbqdn:
Wouldn't it be beneficial to trim the empty parents as well? The reason I noticed these bugs was that I'm using the method `from_frontier` in Zebra, and I noticed the trailing zeroes in serialized trees, so I was wondering if it would cause any issues.

daira (approved):
utACK

ebfull (approved):

daira (commented):
Post-hoc utACK

PR #554, opened 2022-05-11
Update NU5 mainnet activation height.

nuttycom:


steven-ecc (approved):
utACK

zcash_primitives/src/consensus.rs
```diff
@@ -659,9 +659,13 @@ mod tests {
...
             BranchId::Canopy,
         );
+        assert_eq!(
+            BranchId::for_height(&MAIN_NETWORK, BlockHeight(1_687_104)),
+            BranchId::Nu5,
+        );
         assert_eq!(
             BranchId::for_height(&MAIN_NETWORK, BlockHeight(5_000_000)),
-            BranchId::Canopy,
+            BranchId::Nu5,
```
steven-ecc, line 668: Should this be "NU5"?

steven-ecc: Nm, looks like the constant is mixed case.

nuttycom: Yes; Rust convention is to use CamelCase even for acronyms.

daira (approved):
utACK

PR #555, opened 2022-05-19
Backport NU5 branch ID and activation heights for mobile SDKs

str4d:
This enables mobile wallets using versions of the mobile SDKs based on zcash/librustzcash@8e3e7de50c3c35bd7c18988bb629dd1baeb1a3d1 to send transactions after NU5 activation. These wallets will however still not support receiving v5 transactions.

PR #559, opened 2022-05-24
Legacy SDK helper: Store the original transaction data to the wallet database without parsing roundtrip.

nuttycom:
Builds on #555 and #558.


daira:
This PR appears to have no changes? Is it targeted to the right branch?

nuttycom:
> This PR appears to have no changes? Is it targeted to the right branch?

It discards the changes; it exists solely to ensure that the backport commit exists in the history of `master` so that it can be referred to by a cargo `ref=` and not get lost due to a branch being deleted; however, the changes will not actually become part of `master`.

str4d (approved):
ACK. I confirmed that the top merge is a merge of current `master` and the commit we used in the most recent mobile SDK changes.

Issue #563, opened 2022-06-06
zcash_address: Enable parser errors to be exposed


str4d:
The `zcash_address` crate focuses on parsing address encodings, down to the bytes of the represented raw addresses. Those bytes will likely be further validated by code that uses `zcash_address` (e.g. checking for valid curve points), but currently the `FromAddress` trait methods only have the error type `UnsupportedAddress`. We should generalise it to an enum with two cases:
- `UnsupportedAddress`
- `InvalidAddress` containing downstream-specific errors.

This might work nicely by adding an associated type on `FromAddress` to store the parsing error.

This is a breaking change, and would require a `zcash_address 0.2.0` release.

str4d:
The further I get into #387, the more I'm convinced it blocks on this (or us reimplementing the `zcash_address` logic which is pointless).

PR #565, opened 2022-06-07
zcash_address: Replace `FromAddress` with `TryFromAddress`

str4d:
This enables the user-defined conversions to be fallible, which they will almost always want to be (as address data needs to be validated before it can be used).

Closes zcash/librustzcash#563.


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/547ad12117f4a6c0b156ad64ac24ecc33cf63da7..16938d1c4f087fa61c3be35f911f7d25293df41a) to fix `rustdoc` lint.

Issue #564, opened 2022-06-06
zcash_address: Enable a parsing workflow that checks against an expected network


str4d:
Currently, the `FromAddress` trait methods provide the parsed network as part of the constructor. This makes it more noisy to construct network-agnostic types while checking that the network matches an expected value, as `FromAddress` needs to be implemented on an intermediate type.

We should instead add a `ZcashAddress::convert_if_network(self, Network)` method that handles the network check internally (and also allows testnet Sprout and transparent addresses to be "converted" to regtest in the same way we currently handle the reverse in `ToAddress`).

PR #567, opened 2022-06-10
zcash_address: Add `ZcashAddress::convert_if_network`

str4d:
This, along with the corresponding `TryFromRawAddress` trait, enables converting `ZcashAddress` into a network-agnostic type.

Closes zcash/librustzcash#564.


daira (approved):
utACK

Issue #387, opened 2021-05-24
Implement Unified Address key derivation


str4d:
We need APIs that can take a BIP 39 seed phrase (or some root seed derived from it), and an account number, and return a Unified Spending Key. This can then be used by downstream wallet developers like the mobile SDKs.

PR #568, opened 2022-06-10
Add `RecipientAddress::Unified`

str4d:
Part of zcash/librustzcash#387.

### DIFF
--- a/zcash_client_backend/src/address.rs
+++ b/zcash_client_backend/src/address.rs
@@ -1,11 +1,17 @@
 //! Structs for handling supported address types.
 
-use zcash_primitives::{consensus, legacy::TransparentAddress, sapling::PaymentAddress};
+use zcash_address::{ConversionError, Network, ToAddress, TryFromRawAddress, ZcashAddress};
+use zcash_primitives::{consensus, constants, legacy::TransparentAddress, sapling::PaymentAddress};
 
-use crate::encoding::{
-    decode_payment_address, decode_transparent_address, encode_payment_address_p,
-    encode_transparent_address_p,
-};
+fn params_to_network<P: consensus::Parameters>(params: &P) -> Network {
+    // Use the Sapling HRP as an indicator of network.
+    match params.hrp_sapling_payment_address() {
+        constants::mainnet::HRP_SAPLING_PAYMENT_ADDRESS => Network::Main,
+        constants::testnet::HRP_SAPLING_PAYMENT_ADDRESS => Network::Test,
+        constants::regtest::HRP_SAPLING_PAYMENT_ADDRESS => Network::Regtest,
+        _ => panic!("Unsupported network kind"),
+    }
+}
 
 /// An address that funds can be sent to.
 // TODO: rename to ParsedAddress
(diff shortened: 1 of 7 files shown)


zcash_client_backend/src/address.rs
```diff
@@ -1,18 +1,151 @@
...
+impl UnifiedAddress {
+    /// Constructs a Unified Address from a given set of receivers.
+    ///
+    /// Returns `None` if the receivers would produce an invalid Unified Address (namely,
+    /// if no shielded receiver is provided).
+    pub fn from_receivers(
+        orchard: Option<orchard::Address>,
+        sapling: Option<PaymentAddress>,
+        transparent: Option<TransparentAddress>,
+    ) -> Option<Self> {
```
pacu, line 92: This could be conceptually wrong because I'm a Rust newbie. 

Should this be a wrapper a function that actually takes the unknown receiver as well?

````Rust
 pub fn from_receivers(
        orchard: Option<orchard::Address>,
        sapling: Option<PaymentAddress>,
        transparent: Option<TransparentAddress>,
        unknown: Vec<(u32, Vec<u8>)>,
    ) -> Option<Self> {
````

File: zcash_client_backend/src/address.rs [hunk 2]
```diff
@@ -1,18 +1,151 @@
...
+    }
+}
 
 /// An address that funds can be sent to.
 // TODO: rename to ParsedAddress
 #[derive(Debug, PartialEq, Clone)]
 pub enum RecipientAddress {
     Shielded(PaymentAddress),
     Transparent(TransparentAddress),
+    Unified(UnifiedAddress),
```
pacu, line 148: 😍

zcash_client_backend/src/address.rs
str4d: "No-ish"

This is currently sort of a hybrid of two structs: a receiver-aware struct that can handle any address, and a receiver-aware struct that is only for wallet-generated addresses. I currently don't want to expose the ability to construct a new `UnifiedAddress` with arbitrary receivers (I only want known receivers), hence `unknown` is not exposed. However `unknown` is necessary in order to _parse_ arbitrary UAs, since we need to store them to ensure round-trip serialization.

This API will likely end up changing once I've dug us out from under this particular tech debt hill, but for now I want to keep things as minimal and straightforward as possible while we go through the multiple PRs to fix the structure of `UnifiedFullViewingKey`.

zcash_client_backend/src/keys.rs
```diff
@@ -164,6 +183,79 @@ impl UnifiedFullViewingKey {
...
+    ///
+    /// Returns `None` if the specified index does not produce a valid diversifier.
+    // TODO: Allow filtering down by receiver types?
+    pub fn address(&self, j: DiversifierIndex) -> Option<UnifiedAddress> {
+        let sapling = if let Some(extfvk) = self.sapling.as_ref() {
+            match extfvk.address(j) {
+                Some(pa) => Some(pa),
+                // Diversifier doesn't generate a valid Sapling address.
+                None => return None,
+            }
```
daira, line 197:
```suggestion
            Some(extfvk.address(j)?)
```

zcash_client_backend/src/zip321.rs
```diff
@@ -28,7 +28,7 @@ pub enum Zip321Error {
     InvalidBase64(base64::DecodeError),
     MemoBytesError(memo::Error),
     TooManyPayments(usize),
-    DuplicateParameter(parse::Param, usize),
+    DuplicateParameter(Box<parse::Param>, usize),
```
daira, line 31: clippy is now warning in a different place.

daira (approved):
utACK modulo the new clippy warning.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/b0b69209cdd01d1863f40d8f139838c78e017b04..97aee898b2fb5e8e039497608b2e3dc486e3368e) to address @daira's comments.

Issue #489, opened 2022-01-31
Implement UA-specific migrations for `zcash_client_sqlite`


str4d:
The `zcash_client_sqlite` data DB needs to be updated to account for UAs. Specifically:

- We should have an `addresses` table for storing the derived diversified UAs within the known account(s).
- The `accounts` table should have its address-related columns removed (migrating the Sapling and transparent content into the `addresses` table), and a `transparent_fvk` column added.
  - The UFVK can then be derived from the component columns, and e.g. if a wallet doesn't want transparent support, they'd represent this by leaving the `transparent_fvk` column set to `NULL`.
  - The alternative would be to just have a `ufvk` column that stores the encoded UFVK, and parse it from there, but storing the individual components is probably more introspectable.
  - This migration requires the caller to provide access to the UFVKs.

pacu:
I've seen other wallets like Yoroi (cardano) that keep track of the addresses given out and if they have received funds or not. 
This way they don't roll you a new address automatically if the latest one hasn't received funds.

str4d:
For Bitcoin, that's usually not a choice: wallets following BIP 44 semantics require users to receive funds in (at least some of) generated addresses before subsequent new ones are generated, otherwise they would exceed the gap limit and become unable to detect funds. We will have this issue for our transparent components, but less so: we can always detect any shielded funds received at any index within an account, so we can use that to skip forward over any unused transparent receivers.

pacu:
Great, we ought to weed out what information of all this is useful to be available for end users so that we can have te proper APIs in place.

str4d:
Wrt storing the UFVK as one column per receiver vs just storing the UFVK directly in the table, both of these will require migrations:
- One column per receiver: we need to migrate the `extfvk` TEXT column to a `dfvk` BLOB column (can be done without the seed), add BLOB columns for the transparent FVK and Orchard FVK, and then migrate the existing accounts to UFVKs (via the seed).
  - We can't serialize a `dfvk` as an `extfvk` as the former omits data that the latter contained.
  - We don't have a string encoding for Sapling diversifiable FVKs. We could just store them encoded as a Sapling-only UFVK, but that's rather wasteful.
- Storing UFVK directly: we need to replace the `extfvk` TEXT column with a `ufvk` TEXT column, and then migrate the existing Sapling extended FVKs to UFVKs (via the seed).
  - Upside: UFVKs are directly inspectable in the SQL table.
[...]

Issue #403, opened 2021-06-15
Add Orchard support to `zcash_client_backend`


str4d:
The main task here is to add support for Orchard viewing keys in trial-decryption. Decrypted wallet transactions will gain the ability to contain received Orchard notes and spent Orchard nullifiers.

nuttycom:
Completed by #1105

PR #570, opened 2022-06-14
`zcash_client_*`: Migrate from `ExtendedFullViewingKey` to `UnifiedFullViewingKey`

str4d:
Closes zcash/librustzcash#387.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -115,6 +115,7 @@ pub trait WalletRead {
     ///
     /// This will return `Ok(None)` if the account identifier does not correspond
     /// to a known account.
+    // TODO: This does not appear to be the case.
     fn get_address(&self, account: AccountId) -> Result<Option<PaymentAddress>, Self::Error>;
 
     /// Returns all extended full viewing keys known about by this wallet.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -17,6 +17,7 @@ use zcash_primitives::{
 use crate::{
     address::RecipientAddress,
     decrypt::DecryptedOutput,
+    keys::UnifiedFullViewingKey,
     proto::compact_formats::CompactBlock,
     wallet::{SpendableNote, WalletTx},
 };
@@ -118,10 +119,10 @@ pub trait WalletRead {
     // TODO: This does not appear to be the case.
     fn get_address(&self, account: AccountId) -> Result<Option<PaymentAddress>, Self::Error>;
 
-    /// Returns all extended full viewing keys known about by this wallet.
-    fn get_extended_full_viewing_keys(
+    /// Returns all unified full viewing keys known to this wallet.
+    fn get_unified_full_viewing_keys(
         &self,
-    ) -> Result<HashMap<AccountId, ExtendedFullViewingKey>, Self::Error>;
+    ) -> Result<HashMap<AccountId, UnifiedFullViewingKey>, Self::Error>;
 
     /// Checks whether the specified extended full viewing key is
     /// associated with the account.
@@ -330,6 +331,7 @@ pub mod testing {
     };
 
     use crate::{
+        keys::UnifiedFullViewingKey,
         proto::compact_formats::CompactBlock,
         wallet::{SpendableNote, WalletTransparentOutput},
     };
@@ -386,9 +388,9 @@ pub mod testing {
             Ok(None)
         }
 
-        fn get_extended_full_viewing_keys(
+        fn get_unified_full_viewing_keys(
             &self,
-        ) -> Result<HashMap<AccountId, ExtendedFullViewingKey>, Self::Error> {
+        ) -> Result<HashMap<AccountId, UnifiedFullViewingKey>, Self::Error> {
             Ok(HashMap::new())
         }
 
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -215,7 +215,7 @@ where
     // https://github.com/zcash/librustzcash/issues/403
     let dfvks: Vec<_> = ufvks
         .iter()
-        .map(|(account, ufvk)| (account, ufvk.sapling().expect("TODO Add Orchard support")))
+        .filter_map(|(account, ufvk)| ufvk.sapling().map(move |k| (account, k)))
         .collect();
 
     // Get the most recent CommitmentTree
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -84,7 +84,6 @@ use zcash_primitives::{
     consensus::{self, BlockHeight, NetworkUpgrade},
     merkle_tree::CommitmentTree,
     sapling::Nullifier,
-    zip32::{AccountId, ExtendedFullViewingKey},
 };
 
 use crate::{
@@ -214,7 +213,7 @@ where
     let ufvks = data.get_unified_full_viewing_keys()?;
     // TODO: Change `scan_block` to also scan Orchard.
     // https://github.com/zcash/librustzcash/issues/403
-    let extfvks: Vec<(&AccountId, &ExtendedFullViewingKey)> = ufvks
+    let dfvks: Vec<_> = ufvks
         .iter()
         .map(|(account, ufvk)| (account, ufvk.sapling().expect("TODO Add Orchard support")))
         .collect();
@@ -249,7 +248,7 @@ where
             scan_block(
                 params,
                 block,
-                &extfvks,
+                &dfvks,
                 &nullifiers,
                 &mut tree,
                 &mut witness_refs[..],
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -210,9 +210,14 @@ where
             .unwrap_or(sapling_activation_height - 1)
     })?;
 
-    // Fetch the ExtendedFullViewingKeys we are tracking
-    let extfvks = data.get_extended_full_viewing_keys()?;
-    let extfvks: Vec<(&AccountId, &ExtendedFullViewingKey)> = extfvks.iter().collect();
+    // Fetch the UnifiedFullViewingKeys we are tracking
+    let ufvks = data.get_unified_full_viewing_keys()?;
+    // TODO: Change `scan_block` to also scan Orchard.
+    // https://github.com/zcash/librustzcash/issues/403
+    let extfvks: Vec<(&AccountId, &ExtendedFullViewingKey)> = ufvks
+        .iter()
+        .map(|(account, ufvk)| (account, ufvk.sapling().expect("TODO Add Orchard support")))
+        .collect();
 
     // Get the most recent CommitmentTree
     let mut tree = data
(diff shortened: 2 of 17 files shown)


zcash_primitives/src/sapling/keys.rs
```diff
@@ -157,6 +161,151 @@ impl FullViewingKey {
...
+        let mut bytes = [0; 128];
+        self.fvk
+            .write(&mut bytes[..96])
+            .expect("slice should be the correct length");
+        bytes[96..].copy_from_slice(&self.dk.0);
+        bytes
+    }
+
+    /// Derives the internal `DiversifiableFullViewingKey` corresponding to `self` (which
+    /// is assumed here to be an external DFVK).
```
daira, line 229: It's a little bit error-prone that the type doesn't record whether this is an internal or external FVK, and so you can call `derive_internal` incorrectly on a `DiversifiableFullViewingKey` that is already internal. (OTOH, in most places internal and external keys are handled uniformly.)

zcash_primitives/src/sapling/keys.rs
```diff
@@ -210,4 +359,24 @@ mod tests {
...
+            DiversifiableFullViewingKey::from(extfvk)
+        };
+
+        // Check value -> bytes -> parsed round trip.
+        let dfvk_bytes = dfvk.to_bytes();
+        let dfvk_parsed = DiversifiableFullViewingKey::from_bytes(&dfvk_bytes).unwrap();
+        assert_eq!(dfvk_parsed.fvk.vk.ak, dfvk.fvk.vk.ak);
+        assert_eq!(dfvk_parsed.fvk.vk.nk, dfvk.fvk.vk.nk);
+        assert_eq!(dfvk_parsed.fvk.ovk, dfvk.fvk.ovk);
+        assert_eq!(dfvk_parsed.dk, dfvk.dk);
```
daira, line 384: Why not derive `Eq`?

zcash_client_backend/src/data_api.rs
```diff
@@ -115,6 +115,7 @@ pub trait WalletRead {
     ///
     /// This will return `Ok(None)` if the account identifier does not correspond
     /// to a known account.
+    // TODO: This does not appear to be the case.
```
daira, line 119: Document the actual behaviour.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -210,9 +210,13 @@ where
             .unwrap_or(sapling_activation_height - 1)
     })?;
 
-    // Fetch the ExtendedFullViewingKeys we are tracking
-    let extfvks = data.get_extended_full_viewing_keys()?;
-    let extfvks: Vec<(&AccountId, &ExtendedFullViewingKey)> = extfvks.iter().collect();
+    // Fetch the UnifiedFullViewingKeys we are tracking
+    let ufvks = data.get_unified_full_viewing_keys()?;
+    // TODO: Change `scan_block` to also scan Orchard.
```
daira, line 214: Say which issue this is.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -44,10 +42,12 @@ use {
...
 pub fn init_wallet_db<P>(wdb: &WalletDb<P>) -> Result<(), rusqlite::Error> {
+    // TODO: Add migrations (https://github.com/zcash/librustzcash/issues/489)
+    // - extfvk column -> ufvk column
     wdb.conn.execute(
         "CREATE TABLE IF NOT EXISTS accounts (
             account INTEGER PRIMARY KEY,
-            extfvk TEXT,
+            ufvk TEXT,
             address TEXT,
             transparent_address TEXT
```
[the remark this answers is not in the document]
str4d: `teansparent_address` is going to be removed, just not in this PR.

zcash_client_backend/src/keys.rs
```diff
@@ -139,6 +143,8 @@ pub struct UnifiedFullViewingKey {
     #[cfg(feature = "transparent-inputs")]
     transparent: Option<legacy::AccountPubKey>,
     sapling: Option<sapling_keys::DiversifiableFullViewingKey>,
+    orchard: Option<orchard::keys::FullViewingKey>,
+    unknown: Vec<(u32, Vec<u8>)>,
```
daira, line 147: It's a public struct in a library crate, so I don't understand why clippy thinks this field can't be read. Consider suppressing the warning.

zcash_client_backend/src/keys.rs
```diff
@@ -155,70 +162,107 @@ impl UnifiedFullViewingKey {
...
+        let mut transparent = None;
+
+        // We can use as-parsed order here for efficiency, because we're breaking out the
+        // receivers we support from the unknown receivers.
+        let unknown = ufvk
+            .items_as_parsed()
+            .iter()
+            .filter_map(|receiver| match receiver {
+                unified::Fvk::Orchard(data) => {
+                    Option::from(orchard::keys::FullViewingKey::from_bytes(data))
```
daira, line 198: Suppress the clippy warning.

daira (approved):
utACK with comments.

zcash_client_sqlite/src/wallet/init.rs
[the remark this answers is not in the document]
str4d: The specific issue for this is #489.

zcash_primitives/src/sapling/keys.rs
str4d: Note that this API does not expose a way to produce an internal `DiversifiableFullViewingKey`. This assumption is referring to the fact that we can't know whether `DiversifiableFullViewingKey::from_bytes` is parsing something that is external or internal (or not HD-derived at all), since that information is not stored in the encoding.

zcash_primitives/src/sapling/keys.rs
str4d: I tend to avoid deriving `PartialEq, Eq` on key types that could be considered secret, or for which a comparison doesn't really make sense in the public API. In those situations, reaching into the internals is fine. But if it is useful in the public API, someone could add a `PartialEq` implementation backed by `ConstantTimeEq`.

zcash_client_backend/src/data_api/chain.rs
str4d: This is #403.

zcash_client_backend/src/keys.rs
str4d: Nope, this is actually a bug! We are parsing `unknown` from the encoding, but not including it when serializing. I will fix.

zcash_client_backend/src/keys.rs
str4d: Fixed instead.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/a33082477df8ec78c48dcbd4df188e7d2ac1db1f..b52e949bd6ee1dbbc0324437440bed8b3c423f79) to fix doctest error and address @daira's comments.

zcash_client_backend/src/data_api.rs
str4d: #571

daira (approved):
utACK

zcash_client_backend/src/welding_rig.rs
```diff
@@ -127,6 +128,26 @@ pub trait ScanningKey {
     fn nf(&self, note: &Note, witness: &IncrementalWitness<Node>) -> Self::Nf;
 }
 
+impl ScanningKey for DiversifiableFullViewingKey {
+    type Nf = Nullifier;
```
nuttycom, line 132:
```suggestion
    type Nf = sapling::Nullifier;
```

zcash_client_backend/src/welding_rig.rs
```diff
@@ -9,6 +9,7 @@ use zcash_primitives::{
     consensus::{self, BlockHeight},
     merkle_tree::{CommitmentTree, IncrementalWitness},
     sapling::{
+        keys::DiversifiableFullViewingKey,
```
nuttycom, line 12:
```suggestion
        self,
        keys::DiversifiableFullViewingKey,
```

nuttycom (commented):
Added some questions/suggestions about the migration approach. This PR does things in a  more hard-cut-over fashion than I think might be necessary. Overall it looks like it'll work, but ideally I'd like to avoid the "fake UFVK" encoding.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -199,7 +199,7 @@ pub fn init_accounts_table<P: consensus::Parameters>(
 
     // Insert accounts atomically
     wdb.conn.execute("BEGIN IMMEDIATE", NO_PARAMS)?;
-    for key in keys.iter() {
+    for (account, key) in (0u32..).zip(keys) {
```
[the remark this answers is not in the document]
str4d: I'm fine with making this refactor, but it should be done in a separate PR. This PR keeps us within the documented requirements of the method.

zcash_client_backend/src/decrypt.rs
```diff
@@ -33,17 +35,18 @@ pub struct DecryptedOutput {
...
     tx: &Transaction,
-    extfvks: &HashMap<AccountId, ExtendedFullViewingKey>,
+    ufvks: &HashMap<AccountId, UnifiedFullViewingKey>,
 ) -> Vec<DecryptedOutput> {
     let mut decrypted = vec![];
 
     if let Some(bundle) = tx.sapling_bundle() {
-        for (account, extfvk) in extfvks.iter() {
+        for (account, ufvk) in ufvks.iter() {
+            let extfvk = ufvk.sapling().expect("TODO: Add Orchard support");
```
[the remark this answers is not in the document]
nuttycom: Fixed by #574

zcash_client_sqlite/src/wallet/init.rs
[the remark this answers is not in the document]
nuttycom: Fixed by #574

zcash_client_backend/src/keys.rs
```diff
@@ -137,38 +140,135 @@ impl UnifiedSpendingKey {
...
 pub struct UnifiedFullViewingKey {
-    account: AccountId,
     #[cfg(feature = "transparent-inputs")]
     transparent: Option<legacy::AccountPubKey>,
-    // TODO: This type is invalid for a UFVK; create a `sapling::DiversifiableFullViewingKey`
-    // to replace it.
-    sapling: Option<sapling::ExtendedFullViewingKey>,
+    sapling: Option<sapling_keys::DiversifiableFullViewingKey>,
+    orchard: Option<orchard::keys::FullViewingKey>,
+    unknown: Vec<(u32, Vec<u8>)>,
```
nuttycom, line 147: I still don't think we should include the unknown components here, nor should we seek roundtrip serialization capability at this level.

The data flow I envision for UFVK import (the only case where this might matter) is that the original encoding that was imported is persisted, and then it is decoded to a type which does not contain the unknown parts. That decoded value should be in-memory only, and should never need to be reencoded; if the encoded form is required, the original encoded value should be used.

nuttycom (commented):
I'm pretty strongly -0 on making UnifiedFullViewingKey handle roundtrip serialization to strings; I think that conversions between this and the zcash_address types should be explicit, and that we should generally not ever need to serialize UnifiedFullViewingKey except at initial creation from the spending key; in that circumstance, the explicit conversion to Ufvk is not a big deal.

I think that if we include the unknown elements, it'll ultimately come back to bite us in the form of added unnecessary complexity.

str4d (commented):
utACK 7236204b14c13ed76bdf4ad2d70a3f8f3663d870

PR #573, opened 2022-06-24
`zcash_primitives` and `zcash_proofs` 0.7.0

str4d:


daira (approved):
utACK. I didn't not check whether all of the changelogs and `Cargo.toml`s were consistent with each other in each crate.

nuttycom (approved):
utACK

PR #574, opened 2022-06-27
Fix account key init

nuttycom:
This builds atop #570

### DIFF
--- a/zcash_client_backend/src/keys.rs
+++ b/zcash_client_backend/src/keys.rs
@@ -364,7 +364,10 @@ mod tests {
     #[cfg(feature = "transparent-inputs")]
     use {
         crate::encoding::AddressCodec,
-        zcash_primitives::{legacy, legacy::keys::IncomingViewingKey},
+        zcash_primitives::legacy::{
+            self,
+            keys::{AccountPrivKey, IncomingViewingKey},
+        },
     };
 
     #[cfg(feature = "transparent-inputs")]
@@ -409,7 +412,11 @@ mod tests {
         };
 
         #[cfg(feature = "transparent-inputs")]
-        let transparent = { None };
+        let transparent = {
+            let privkey =
+                AccountPrivKey::from_seed(&MAIN_NETWORK, &[0; 32], AccountId::from(0)).unwrap();
+            Some(privkey.to_account_pubkey())
+        };
 
         let ufvk = UnifiedFullViewingKey::new(
             #[cfg(feature = "transparent-inputs")]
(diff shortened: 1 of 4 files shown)


zcash_client_backend/src/keys.rs
```diff
@@ -107,10 +109,11 @@ impl UnifiedSpendingKey {
 
     pub fn to_unified_full_viewing_key(&self) -> UnifiedFullViewingKey {
         UnifiedFullViewingKey {
-            account: self.account,
             #[cfg(feature = "transparent-inputs")]
             transparent: Some(self.transparent.to_account_pubkey()),
```
daira, line 113: If the `transparent-inputs` feature is not enabled, will any transparent component be dropped (which seems incorrect), or will it be stored in `unknown` (to preserve round-trip serialisation)?

zcash_client_backend/src/keys.rs
```diff
@@ -137,38 +140,135 @@ impl UnifiedSpendingKey {
 #[derive(Clone, Debug)]
 #[doc(hidden)]
 pub struct UnifiedFullViewingKey {
-    account: AccountId,
     #[cfg(feature = "transparent-inputs")]
     transparent: Option<legacy::AccountPubKey>,
```
daira, line 144: Same question for when `transparent-inputs` is not enabled.

zcash_client_backend/src/keys.rs
```diff
@@ -137,38 +140,135 @@ impl UnifiedSpendingKey {
...
+                unified::Fvk::P2pkh(data) => legacy::AccountPubKey::deserialize(data)
+                    .map_err(|_| "Invalid transparent FVK in Unified FVK")
+                    .map(|tfvk| {
+                        transparent = Some(tfvk);
+                        None
+                    })
+                    .transpose(),
+                #[cfg(not(feature = "transparent-inputs"))]
+                unified::Fvk::P2pkh(data) => {
+                    Some(Ok((unified::Typecode::P2pkh.into(), data.to_vec())))
```
daira, line 223: Ok it looks as though when `transparent-inputs` is disabled this will treat the P2PKH component as though it were unknown. If that's true then feel free to resolve my questions above.

zcash_client_backend/src/keys.rs
```diff
@@ -291,4 +393,46 @@ mod tests {
...
+            ufvk.transparent.map(|t| t.serialize()),
+        );
+        assert_eq!(
+            decoded.sapling.map(|s| s.to_bytes()),
+            ufvk.sapling.map(|s| s.to_bytes()),
+        );
+        assert_eq!(
+            decoded.orchard.map(|o| o.to_bytes()),
+            ufvk.orchard.map(|o| o.to_bytes()),
+        );
```
daira, line 436: How about also having a round trip test from a UFVK and UA that each contain an unknown component, to the internal repr and back again?

daira (approved):
utACK (modulo the clippy failure).

daira (approved):
utACK

PR #575, opened 2022-06-27
Return the network type for encoding addresses as part of the network parameters.

nuttycom:
This is a better solution for https://github.com/zcash/librustzcash/pull/570/files#diff-02d0f986b89bf4015dc96d572a01613b93abcfcced6b678c7e9332e3ac542017R11


zcash_primitives/Cargo.toml
```diff
@@ -45,6 +45,7 @@ ripemd = { version = "0.1", optional = true }
 secp256k1 = { version = "0.21", optional = true }
 sha2 = "0.9"
 subtle = "2.2.3"
+zcash_address = { version = "0.1", path = "../components/zcash_address" }
```
str4d, line 48: The previous approach avoided the need for this dependency interaction, which otherwise goes unused. However, at some point we will refactor our fully-parsed address types out of `zcash_client_backend` (once we're happy with them), and at that point we'd likely need this dependency, so it's fine to add now.

str4d (approved):
utACK

str4d (approved):
re-utACK

PR #580, opened 2022-07-04
Sapling batch validation

str4d:
As this only makes additions to the `zcash_proofs` public API, we can publish this as `zcash_proofs 0.7.1`.


ebfull (approved):

therealyingtong (approved):
utACK

str4d:
Note that we will _only_ publish `zcash_proofs` after this merges. We cannot publish a `zcash_primitives 0.7.1` because of the PR @nuttycom made that altered the `consensus::Parameters` trait. Fortunately we do not need to do that in order to publish `zcash_proofs 0.7.1`.

zcash_proofs/src/sapling/verifier.rs
```diff
@@ -1,37 +1,37 @@
...
             cv_sum: jubjub::ExtendedPoint::identity(),
-            zip216_enabled,
         }
     }
 
     /// Perform consensus checks on a Sapling SpendDescription, while
     /// accumulating its value commitment inside the context for later use.
     #[allow(clippy::too_many_arguments)]
-    pub fn check_spend(
+    fn check_spend<C>(
```
nuttycom, line 34: Do these visibility changes need to be noted in the changelog?

nuttycom (approved):
testedACK - this in conjunction with zcash/zcash#6048 successfully validates mainnet blocks to the current chain tip.

zcash_proofs/src/sapling/verifier.rs
str4d: No. Note that I changed the name of the struct these functions are in, and the old struct name (which is the "new" struct in `single.rs`) has the old function signatures with the same public visibility.

zcash_proofs/src/sapling/verifier/batch.rs
```diff
@@ -0,0 +1,175 @@
...
+            signatures: redjubjub::batch::Verifier::new(),
+        }
+    }
+
+    /// Checks the bundle against Sapling-specific consensus rules, and adds its proof and
+    /// signatures to the validator.
+    ///
+    /// Returns `false` if the bundle doesn't satisfy all of the consensus rules. This
+    /// `BatchValidator` can continue to be used, but it may include some or all of the
+    /// bundle's proofs and signatures.
```
daira, line 43:
```suggestion
    /// `BatchValidator` can continue to be used regardless, but some or all of the proofs
    /// and signatures from this bundle may have already been added to the batch even if
    /// it fails other consensus rules.
```

zcash_proofs/src/sapling/verifier/batch.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub fn validate<R: RngCore + CryptoRng>(
+        self,
+        spend_vk: &groth16::VerifyingKey<Bls12>,
+        output_vk: &groth16::VerifyingKey<Bls12>,
+        mut rng: R,
+    ) -> bool {
+        if !self.bundles_added {
+            // An empty batch is always valid, but is not free to run; skip it. Note that
+            // a bundle has at least a binding signature, so if there are no binding
+            // signatures, there are also no proofs or spend authorization signatures.
```
daira, line 154:
```suggestion
            // An empty batch is always valid, but is not free to run; skip it.
```
We're not relying on a bundle having at least a binding signature here.

daira (approved):
utACK

therealyingtong (approved):
utACK

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/a64f0020920c239673699157ac99be767b8e4989..e545002dfa70429eaa7028730b7f17cb7a7c4d67) to use `bellman 0.13.1`.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/2b20131de4dd75f788645f9e9ed8793d5ea96d3c..407eeca1e2d29e92a6cc546f8f935d7e947aa243) to fix the threadpool commit to use the `multicore` feature flag, so it still works on `wasm32-wasi`.

ebfull (approved):

nuttycom (approved):
re-utACK  6f0e26a

daira (commented):
Post-hoc reACK

Issue #406, opened 2021-06-15
Add Orchard support to the transaction builder


str4d:
Now that the refactors are all done, we can integrate the modular Orchard bundle builder into our Rust transaction builder.

- [ ] The standard ZIP 317 fee algorithm applies to fully-constructed transactions. However, in the transaction builders, the Orchard (and potentially the Sapling) bundle may be padded to 2 actions. We need to ensure that this padding is taken into account by the ZIP 317 fee estimation code.

nuttycom:
Resolved by #1065

PR #366, opened 2021-03-27
Lint fixes

str4d:
This gets us clippy-clean (aside from #367 which I've deferred as it is non-trivial work).

PR #583, opened 2022-07-05
Add orchard bundle creation to `zcash_primitives::transaction::builder::Builder`

AloeareV:
This adds a `with_orchard_anchor` constructor to the transaction builder, that allows for adding orchard components to transactions.

### DIFF
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -472,6 +472,11 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng, O: MaybeOrchard> Buil
                 self.progress_notifier.as_ref(),
             )
             .map_err(Error::SaplingBuild)?;
+        let orchard_bundle: Option<orchard::Bundle<_, Amount>> = self
+            .orchard_builder
+            .build(&mut rng)
+            .transpose()
+            .map_err(Error::OrchardBuild)?;
 
         #[cfg(feature = "zfuture")]
         let (tze_bundle, tze_signers) = self.tze_builder.build();
@@ -488,7 +493,7 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng, O: MaybeOrchard> Buil
             // for both a `Bundle<InProgress<Unproven, Unauthorized>` and a `Bundle<Authorized, V>`
             // As such, it's created and authorized at the same time, right before being added to the
             // authorized bundle
-            orchard_bundle: None,
+            orchard_bundle,
             #[cfg(feature = "zfuture")]
             tze_bundle,
         };
@@ -535,11 +540,8 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng, O: MaybeOrchard> Buil
 
         let orchard_saks = self.orchard_saks.clone();
 
-        let orchard_bundle: Option<orchard::Bundle<_, Amount>> = self
-            .orchard_builder
-            .build(&mut rng)
-            .transpose()
-            .map_err(Error::OrchardBuild)?
+        let orchard_bundle = unauthed_tx
+            .orchard_bundle
             .map(|b| {
                 b.create_proof(&orchard::circuit::ProvingKey::build(), &mut rng)
                     .and_then(|b| {
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -66,11 +66,11 @@ pub enum Error<FeeError> {
     /// An error occurred in constructing the Sapling parts of a transaction.
     SaplingBuild(sapling_builder::Error),
     /// An error occurred in constructing the Orchard parts of a transaction.
-    OrchardBuild(orchard::builder::Error),
+    OrchardBuild(orchard::builder::BuildError),
     /// An error occurred in adding an Orchard Spend a transaction.
-    OrchardSpend(&'static str),
+    OrchardSpend(orchard::builder::SpendError),
     /// An error occurred in adding an Orchard Output a transaction.
-    OrchardRecipient(&'static str),
+    OrchardRecipient(orchard::builder::OutputError),
     /// Orchard is not yet active on the network, and the user attempted to add
     /// Orchard parts to a transaction.
     NU5Inactive,
@@ -179,6 +179,10 @@ pub struct Builder<'a, P, R, O: MaybeOrchard = WithoutOrchard> {
     transparent_builder: TransparentBuilder,
     sapling_builder: SaplingBuilder<P>,
     orchard_builder: O,
+    // TODO: These should be calculated by the orchard builder and not cached here
+    // This requires publicizing methods or fields of the orchard builder
+    num_orchard_spends: usize,
+    num_orchard_outputs: usize,
     orchard_saks: Vec<orchard::keys::SpendAuthorizingKey>,
     #[cfg(feature = "zfuture")]
     tze_builder: TzeBuilder<'a, TransactionData<Unauthorized>>,
@@ -286,6 +290,8 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng, O: MaybeOrchard> Buil
             #[cfg(not(feature = "zfuture"))]
             tze_builder: std::marker::PhantomData,
             progress_notifier: None,
+            num_orchard_spends: 0,
+            num_orchard_outputs: 0,
         }
     }
 
@@ -387,6 +393,10 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng, O: MaybeOrchard> Buil
                 self.transparent_builder.outputs(),
                 self.sapling_builder.inputs().len(),
                 self.sapling_builder.bundle_output_count(),
+                match std::cmp::max(self.num_orchard_spends, self.num_orchard_outputs) {
+                    1 => 2,
+                    n => n,
+                },
             )
             .map_err(Error::Fee)?;
         self.build_internal(prover, fee)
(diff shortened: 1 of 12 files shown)


str4d:
I had started working on this change (which is #406) a month and a half ago, but had to put it aside while doing other NU5 work, and forgot to open a draft PR (which I've now done: #584). The complexity was indeed in figuring out the areas you have run into 😄

> Add a `with_orchard` function to transaction builder, which will enable the builder to create orchard bundles. This is likely the most feasible solution, but I'm not sold on it yet, as it's a potentially confusing interface.

This is the direction I was taking, specifically to add a `with_orchard_anchor` API.

AloeareV:
Great! If I were to do some work atop that, would you prefer I open a PR into your feature, or push it to here?

AloeareV:
I've marked this as ready for review. It needs tests added, but I'm finding myself at a bit of a loss on how to test this in a meaningful way, as adding an orchard spend needs a note, and currently (I've opened [this PR](https://github.com/zcash/orchard/pull/344) which might also help here) there seems to be no way to get a note aside from decrypting one, and that requires creating one. All the tests are currently using mock tx provers, which can't actually bind signatures. 

So do I need to either access sapling parameters, or feature-gate behind transparent inputs, in order to create a transparent-to-orchard or sapling-to-orchard transaction, in order to have a spendable orchard note?

AloeareV:
[Zingo](https://github.com/zingolabs/zingolib) is currently relying on a cargo patch in order to use this builder (it's also patching librustzcash to use https://github.com/zcash/orchard/pull/344). This means all downstream libraries need to copy our patch into their `Cargo.toml`s in order to use the library. 
Is there something more I can/should be doing to get this merged in? Is this blocked on orchard/#344 (in order to make some mock notes for testing)?

AloeareV:
Re-marked as draft, as this is blocked on zcash/orchard#352

AloeareV:
Some of my code is squashed into str4d's initial commit, as doing so made rebasing a lot easier.

AloeareV:
Alright! Librustzcash has moved far enough since I opened this that a rebase proved really difficult, so I reimplemented instead. 

Ideally, we should get the number of orchard actions from the orchard builder, instead of the builder keeping its own count of inputs and outputs. This (and the later work of adding orchard to change handling in zcash_client_backend) will require adding access to `spends` and `outputs` of the orchard builder (and for change, `note` of `SpendInfo` and `value` of `RecipientInfo`).

AloeareV:
There is also a change to `test_only_new_with_rng` so that it returns an existential type `Builder<'a, P, impl RngCore + CryptoRng>` instead of a `Builder<'a, P, R>`. This allows `new_internal` to require a CryptoRng bound, which is faked by the test fn. This is technically a breaking change to a public interface. It makes some things easier but isn't strictly necessary for the rest of the changes, so it could be reverted if it needs to be, but it makes calls to `new_internal` safer.

AloeareV:
Testing this locally with zingolib (branch https://github.com/zingolabs/zingolib/pull/185), I'm getting `error_code: -26, error_message: \"16: bad-sapling-bundle-authorization\"`. I haven't touched the sapling authorization code. Have I done something wrong, or is this a bug that exists on main? I can't trivially test that, as it would involve digging up a version of zingolib that can work with a transaction builder that doesn't support orchard. 

I plan to test that, but I won't get a change until monday.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -68,6 +69,15 @@ pub enum Error<FeeError> {
     TransparentBuild(transparent::builder::Error),
     /// An error occurred in constructing the Sapling parts of a transaction.
     SaplingBuild(sapling::builder::Error),
+    /// An error occurred in constructing the Orchard parts of a transaction.
+    OrchardBuild(orchard::builder::Error),
+    /// An error occurred in adding an Orchard Spend a transaction.
+    OrchardSpend(&'static str),
+    /// An error occurred in adding an Orchard Output a transaction.
+    OrchardRecipient(&'static str),
```
nuttycom, line 77: Using strings to communicate these errors does not permit programmatic analysis without parsing the strings; it would be better to add these cases to `orchard::builder::Error` or to add a separate enum that carries appropriate well-typed data describing the errors.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -134,21 +175,47 @@ impl Progress {
...
     target_height: BlockHeight,
     expiry_height: BlockHeight,
     transparent_builder: TransparentBuilder,
     sapling_builder: SaplingBuilder<P>,
+    orchard_builder: O,
+    // TODO: These should be calculated by the orchard builder and not cached here
+    // This requires publicizing methods or fields of the orchard builder
+    num_orchard_spends: usize,
+    num_orchard_outputs: usize,
+    orchard_saks: Vec<orchard::keys::SpendAuthorizingKey>,
```
nuttycom, line 190: Can you please open a PR against the `orchard` repository that proposes the required changes? I want to make sure that this is not misunderstanding the Orchard builder API; for example, the Orchard builder is used within the `zcashd` transaction builder unmodified.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -405,6 +483,10 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
             transparent_bundle,
             sprout_bundle: None,
             sapling_bundle,
+            // This is not of the right type, and it's not obvious how to implement this generically
+            // for both a `Bundle<InProgress<Unproven, Unauthorized>` and a `Bundle<Authorized, V>`
+            // As such, it's created and authorized at the same time, right before being added to the
+            // authorized bundle
```
nuttycom, line 489: It might be useful to look at https://github.com/zcash/zcash/blob/master/src/rust/src/builder_ffi.rs to see how the builder is being used inside of zcashd. It is likely that the `build_internal` method should be split into two different operations and the appropriate changes be made to the Sapling and transparent builders to support that separation.

AloeareV: Thanks, I'll look there next!

nuttycom (commented):
@str4d, @daira, @sellout and I had a look at this in PR gardening and determined that the Orchard integration process would be more straightforward if we first changed the Sapling builder to match the builder pattern that we introduced for the `orchard` crate, and then updated the existing transaction builder to make use of this *without* introducing the Orchard builder, at which point it should be easier for you to handle the Orchard builder integration.

zcash_primitives/src/transaction/builder.rs
AloeareV: Orchard PR here: https://github.com/zcash/orchard/pull/365. There are more changes needed, but I may as well get simple changes out-of-the-way and not risk making a too-big PR.

zcash_primitives/src/transaction/builder.rs
AloeareV: https://github.com/zcash/orchard/pull/366 adds concrete error types here. This should be the last of the simple changes to orchard.

AloeareV:
> @str4d, @daira, @sellout and I had a look at this in PR gardening and determined that the Orchard integration process would be more straightforward if we first changed the Sapling builder to match the builder pattern that we introduced for the `orchard` crate, and then updated the existing transaction builder to make use of this _without_ introducing the Orchard builder, at which point it should be easier for you to handle the Orchard builder integration.

Sounds good! I'll see if I can contrib towards this in the next few days.

zcash_primitives/src/transaction/builder.rs
zancas: Hi @nuttycom I added a review to [orchard #366](https://github.com/zcash/orchard/pull/366).  Let me know if there's something else I can do to move that work along!  Thanks.

AloeareV:
I pushed to this last week, there are still a couple changes suggested that I need to get to before this should be merged (plus I should probably get it so it passes CI), but as far as I can tell it's once again functional. It turns out the `unauthed_tx` needed the orchard bundle.

AloeareV:
@nuttycom Alright, I've gotten this rebased and working, and addressed the review comments!

zcash_primitives/src/transaction/builder.rs
```diff
@@ -51,8 +52,17 @@ use crate::{
 
 const DEFAULT_TX_EXPIRY_DELTA: u32 = 20;
 
-/// Errors that can occur during transaction construction.
+/// Network Upgrade that could be inactive, causing a transaction creation error.
+/// Note that currently NU5 (Orchard) is the only upgrade we check for the presence
+/// of during transaction construction.
 #[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum NetworkUpgrade {
```
nuttycom, line 60: I would prefer that we use the crate-wide type https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/consensus.rs#L365 here instead of a new bespoke type.

AloeareV: Ah! Didn't see that one, that's much better than a new type

AloeareV: Fixed

nuttycom:
@AloeareV just to give you a heads-up, we likely won't get to a full review of this for a few days because the team is trying to get zcashd 5.5.0 out the door.

zancas:
WOOOT!  On the 5.5.0 release!!!

zancas:
Hi!   I notice we're approaching our 1 year anniversary on this PR.   `zinglib` has to patch because this is not landed.   We have had to rebase our feature several times over the months.

AloeareV:
@nuttycom @str4d

zancas:
Do we need to do more work to update this?   I am concerned that it may have been superseded, again.

str4d:
It was pushed back by the necessary 5.6.0 timeline. Now that's out, this is near the top of my review list.

daira:
@AloeareV The first comment in a PR goes into the merge commit, so it should reflect the current state of the PR. Can you update it please?

nuttycom:
Closing in favor of #863

PR #586, opened 2022-07-12
Allow reading address kind and network

adityapk00:
Allow reading the type of address and the network that it can be used on. Useful for tests.

### DIFF
--- a/components/zcash_address/src/lib.rs
+++ b/components/zcash_address/src/lib.rs
@@ -34,7 +34,7 @@ pub enum Network {
 
 /// Known kinds of Zcash addresses.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-enum AddressKind {
+pub enum AddressKind {
     Sprout(kind::sprout::Data),
     Sapling(kind::sapling::Data),
     Unified(unified::Address),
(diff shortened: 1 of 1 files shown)


str4d:
> 1. Be able to parse a user string from an "address" field, and identify what type of address it is. Like if the user types in "ztestsapling....." into an address field, I want to show "You typed a sapling address, but it is for the Test network. Please type in a mainnet address"

For this specific example, it sounds like what you want is `ZcashAddress::convert_if_network`, which returns `ConversionError::IncorrectNetwork` if the networks don't match. This error does not expose the _kind_ of address though (both because as mentioned above `AddressKind` is an implementation detail, and because I'm not currently convinced that the kind matters to a user in this context).

More generally, the `ZcashAddress` type is not intended for this kind of introspection, as it is only partially parsed. Full introspection is what the user-defined type is for; we intend to provide our own type for this in `zcash_primitives` or somewhere similar that people _could_ use for this kind of custom input validation, but haven't yet due to time constraints (and because it's somewhat impacted by other ongoing refactors).

str4d:
Thinking about this again. I think I'd prefer one of the following approaches:
- We get address types into `zcash_primitives` (we currently have been developing types in `zcash_client_backend` but will likely refactor them soon). This in combination with `ZcashAddress::convert_if_network` provides all the requested functionality.
- We add a dedicated "address checker" struct that acts as a conversion target which drops the address data. That is equivalent to exposing `AddressKind` but does not commit us to supporting a particular internal structure for it, and it avoids the issue of dealing with `pk_d` parsing etc. Technically the checker would not detect all possible invalid address encodings, but it would (in combination with `ZcashAddress::convert_if_network`) be sufficient to cover the requested functionality.

nuttycom:
We have added introspection methods to the `ZcashAddress` type to reveal the properties that this PR was intended to expose.

PR #588, opened 2022-07-26
Perform batch decryption in wallet scanning (legacy poc branch)

nuttycom:
At present, we batch on a per-block basis.


components/zcash_note_encryption/src/batch.rs
```diff
@@ -14,7 +12,7 @@ use crate::{
 pub fn try_note_decryption<D: BatchDomain, Output: ShieldedOutput<D>>(
     ivks: &[D::IncomingViewingKey],
     outputs: &[(D, Output)],
-) -> Vec<Option<(D::Note, D::Recipient, D::Memo)>> {
+) -> Vec<Option<((D::Note, D::Recipient, D::Memo), usize)>> {
```
daira, line 15: Explain this in the doc comment.

components/zcash_note_encryption/src/batch.rs
```diff
@@ -24,25 +22,29 @@ pub fn try_note_decryption<D: BatchDomain, Output: ShieldedOutput<D>>(
 pub fn try_compact_note_decryption<D: BatchDomain, Output: ShieldedOutput<D>>(
     ivks: &[D::IncomingViewingKey],
     outputs: &[(D, Output)],
-) -> Vec<Option<(D::Note, D::Recipient)>> {
+) -> Vec<Option<((D::Note, D::Recipient), usize)>> {
```
daira, line 25: Same here.

pacu:
I tested this on an internal Build of ECC wallet. and seems ok. Improvement is not noticeable to the naked eye but any enhancement is very welcome 🙂

nuttycom:
Closing as it seems likely we won't make another wallet release based on the legacy POC branch.

PR #589, opened 2022-07-27
Use batch decryption for wallet scanning

nuttycom:


components/zcash_note_encryption/src/batch.rs
```diff
@@ -55,17 +65,18 @@ where
...
+        .map(|(key_chunk, ((_, ephemeral_key), (domain, output)))| {
+            key_chunk
+                .iter()
+                .zip(ivks.iter().enumerate())
+                .filter_map(|(key, (i, ivk))| {
+                    key.as_ref()
+                        .and_then(|key| decrypt_inner(domain, ivk, ephemeral_key, output, key))
+                        .map(|out| (out, i))
+                })
+                .next()
```
str4d, line 83: The commit message claims that `m * n` decryptions is only the worst case, and this change to return early results in a 50% performance improvement to multi-key scanning on average. I disagree on both counts. `m * n` decryptions is indeed the worst case, but it is also the average case: the chain will contain many more outputs that are _not_ for the wallet than it does that _are_ for the wallet. The 50% improvement only applies to the cost of decrypting notes that are for us; it is not a 50% improvement to the average cost of scanning. The commit text needs to be fixed to clarify this.

Separately, a downside to making this return early is that trial decryption cannot be constant time with respect to the IVKs (by definition). This should be noted in the documentation. I think it is currently not constant time with respect to _outputs_ because the post-decryption validation steps in `zcash_note_encryption` have early returns, but that's a separate issue. cc @defuse, who was the primary motivator for closer-to-constant-time detection when the early mobile SDK code was built, resulting in code like this:

https://github.com/zcash/librustzcash/blob/37fc28634e811fdf8db9274a0080cff17c6b6a09/zcash_client_backend/src/welding_rig.rs#L238-L265).

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -1450,22 +1452,16 @@ mod tests {
...
-        assert_eq!(res.len(), 20);
-        // The batched trial decryptions with invalid_ivk failed.
-        assert_eq!(&res[..10], &vec![None; 10][..]);
-        for (result, (_, output)) in res[10..].iter().zip(outputs.iter()) {
-            // Confirm that the outputs should indeed have failed with invalid_ivk
-            assert_eq!(
-                try_sapling_note_decryption(&TEST_NETWORK, height, &invalid_ivk, output),
-                None
-            );
-
```
str4d, line 1463: This check should have been left in.

components/zcash_note_encryption/src/batch.rs
```diff
@@ -11,39 +10,50 @@ use crate::{
...
+/// Returns a vector containing the decrypted result for each output,
+/// with the same length and in the same order as the outputs were
+/// provided, along with the index in the `ivks` slice associated with
+/// the IVK that successfully decrypted the output.
 #[allow(clippy::type_complexity)]
 pub fn try_note_decryption<D: BatchDomain, Output: ShieldedOutput<D, ENC_CIPHERTEXT_SIZE>>(
     ivks: &[D::IncomingViewingKey],
     outputs: &[(D, Output)],
-) -> Vec<Option<(D::Note, D::Recipient, D::Memo)>> {
+) -> Vec<Option<((D::Note, D::Recipient, D::Memo), usize)>> {
```
str4d, line 22: Why make this a nested tuple, instead of just extending the existing tuple? Call sites will need to be updated either way.
```suggestion
) -> Vec<Option<(D::Note, D::Recipient, D::Memo, usize)>> {
```

components/zcash_note_encryption/src/batch.rs
```diff
@@ -11,39 +10,50 @@ use crate::{
...
+/// Returns a vector containing the decrypted result for each output,
+/// with the same length and in the same order as the outputs were
+/// provided, along with the index in the `ivks` slice associated with
+/// the IVK that successfully decrypted the output.
+#[allow(clippy::type_complexity)]
 pub fn try_compact_note_decryption<D: BatchDomain, Output: ShieldedOutput<D, COMPACT_NOTE_SIZE>>(
     ivks: &[D::IncomingViewingKey],
     outputs: &[(D, Output)],
-) -> Vec<Option<(D::Note, D::Recipient)>> {
+) -> Vec<Option<((D::Note, D::Recipient), usize)>> {
```
str4d, line 38: Ditto.

str4d (changes requested):
Partial review up to the end of the first commit.

components/zcash_note_encryption/src/batch.rs
nuttycom: The reason this doesn't work is that the arity of the first element of this tuple varies according to the `decrypt_inner` argument to `batch_note_decryption`, and so we can't destructure it internally and reconstruct; it would take a whole extra allocation of the result vector and a needless pass over the result.

zcash_client_backend/src/welding_rig.rs
```diff
@@ -87,47 +35,45 @@ fn scan_output<P: consensus::Parameters, K: ScanningKey>(
...
-        P: consensus::Parameters,
-        Output: ShieldedOutput<SaplingDomain<P>, COMPACT_NOTE_SIZE>,
-    >(
-        &self,
-        params: &P,
-        height: BlockHeight,
-        output: &Output,
-    ) -> Option<(Note, PaymentAddress)>;
+    /// Obtain the underlying Sapling incoming viewing key(s) for this scanning key.
+    fn to_sapling_keys(&self) -> Vec<(SaplingIvk, Self::SaplingNk)>;
```
str4d, line 46: Does this need to be a `Vec`, or can it be an iterator (via an associated type)?

zcash_client_backend/src/welding_rig.rs
```diff
@@ -265,17 +201,43 @@ pub fn scan_block<P: consensus::Parameters, K: ScanningKey>(
...
+            let decrypted = batch::try_compact_note_decryption(&ivks, decoded);
+
+            for (index, ((_, output), dec_output)) in decoded.iter().zip(decrypted).enumerate() {
                 // Grab mutable references to new witnesses from previous outputs
                 // in this transaction so that we can update them. Scoped so we
                 // don't hold mutable references to shielded_outputs for too long.
                 let new_witnesses: Vec<_> = shielded_outputs
                     .iter_mut()
-                    .map(|output| &mut output.witness)
+                    .map(|out| &mut out.witness)
```
str4d, line 238: Nit: this line didn't need to be changed.

str4d (changes requested):

components/zcash_note_encryption/src/batch.rs
```diff
@@ -47,6 +47,10 @@ fn batch_note_decryption<D: BatchDomain, Output: ShieldedOutput<D, CS>, F, FR, c
 where
     F: Fn(&D, &D::IncomingViewingKey, &EphemeralKeyBytes, &Output, &D::SymmetricKey) -> Option<FR>,
 {
+    if ivks.is_empty() {
+        return (0..outputs.len()).map(|_| None).collect();
+    };
```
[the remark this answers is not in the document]
nuttycom: This whole operation can't be constant time anyway, so I think this is moot.

zcash_client_backend/src/welding_rig.rs
```diff
@@ -283,26 +265,38 @@ pub fn scan_block<P: consensus::Parameters, K: ScanningKey>(
...
+                let new_witnesses: Vec<_> = shielded_outputs
                     .iter_mut()
                     .map(|output| &mut output.witness)
                     .collect();
 
+                let output = CompactOutputDescription::try_from(output).ok().unwrap();
+
+                // Increment tree and witnesses
+                let node = Node::new(output.cmu.to_repr());
+                for witness in &mut *existing_witnesses {
```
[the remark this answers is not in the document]
nuttycom: These two errors: ```error[E0382]: use of moved value: `existing_witnesses`
   --> zcash_client_backend/src/welding_rig.rs:243:32
    |
145 |     existing_witnesses: &mut [&mut IncrementalWitness<Node>],
    |     ------------------ move occurs because `existing_witnesses` has type `&mut [&mut IncrementalWitness<Node>]`, which does not implement the `Copy` trait
...
243 |                 for witness in existing_witnesses {
    |                                ^^^^^^^^^^^^^^^^^^ `existing_witnesses` moved due to this implicit call to `.into_iter()`, in previous iteration of loop
    |
note: this function takes ownership of the receiver `self`, which moves `existing_witnesses`
   --> /home/nuttycom/.rustup/toolchains/1.56.1-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/collect.rs:234:18
    |
234 |     fn into_iter(self) -> Self::IntoIter;
    |                  ^^^^
help: consider creating a fresh reborrow of `existing_witnesses` here
    |
243 |                 for witness in &mut *existing_witnesses {
    |                                ++++++

For more information about this error, try `rustc --explain E0382`.
error: could not compile `zcash_client_backend` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
nuttycom@sword:~/work/librustzcash/canon (wallet_scan_batch_decrypt *)λ cargo check
   Compiling zcash_client_backend v0.5.0 (/home/nuttycom/work/librustzcash/canon/zcash_client_backend)
error[E0277]: `[&mut IncrementalWitness<Node>]` is not an iterator
   --> zcash_client_backend/src/welding_rig.rs:243:32
    |
243 |                 for witness in &mut existing_witnesses {
    |                                ^^^^^^^^^^^^^^^^^^^^^^^ `[&mut IncrementalWitness<Node>]` is not an iterator
    |
    = help: the trait `Iterator` is not implemented for `[&mut IncrementalWitness<Node>]`
    = note: required because of the requirements on the impl of `Iterator` for `&mut [&mut IncrementalWitness<Node>]`
    = note: 1 redundant requirements hidden
    = note: required because of the requirements on the impl of `Iterator` for `&mut &mut [&mut IncrementalWitness<Node>]`
    = note: required because of the requirements on the impl of `IntoIterator` for `&mut &mut [&mut IncrementalWitness<Node>]`
note: required by `into_iter`
   --> /home/nuttycom/.rustup/toolchains/1.56.1-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/collect.rs:234:5
    |
234 |     fn into_iter(self) -> Self::IntoIter;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

File: zcash_client_backend/src/welding_rig.rs (hunk 5 above)
Comment (nuttycom) [reply zcash_client_backend/src/welding_rig.rs]: I think that the associated type should be `IntoIterator` instead. Fixed.

File: zcash_client_backend/src/welding_rig.rs [hunk 9]
```diff
@@ -105,47 +35,48 @@ fn scan_output<P: consensus::Parameters, K: ScanningKey>(
 /// [`CompactSaplingOutput`]: crate::proto::compact_formats::CompactSaplingOutput
 /// [`scan_block`]: crate::welding_rig::scan_block
 pub trait ScanningKey {
```
nuttycom, line 37: Document this in the changelog.

zcash_client_backend/src/welding_rig.rs
nuttycom: I didn't like the shadowing, it confused me.

zcash_primitives/src/sapling/note_encryption.rs
nuttycom: I guess that the check can now be that it fully fails if you only pass in an invalid Ivk.

zcash_client_backend/src/welding_rig.rs
```diff
@@ -35,29 +35,11 @@ fn scan_output<P: consensus::Parameters, K: ScanningKey>(
...
     height: BlockHeight,
     index: usize,
-    output: CompactSaplingOutput,
+    output: CompactOutputDescription,
     vks: &[(&AccountId, &K)],
     spent_from_accounts: &HashSet<AccountId>,
     tree: &mut CommitmentTree<Node>,
-    existing_witnesses: &mut [&mut IncrementalWitness<Node>],
-    block_witnesses: &mut [&mut IncrementalWitness<Node>],
-    new_witnesses: &mut [&mut IncrementalWitness<Node>],
```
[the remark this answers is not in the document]
nuttycom: I don't think it's worthwhile to document this, since the method is subsequently removed entirely; this was just split out as a separate commit to make the history easier to follow.

zcash_client_backend/src/welding_rig.rs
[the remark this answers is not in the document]
str4d: I think that `existing_witnesses.iter_mut()` might have worked here instead, but it doesn't matter.

zcash_primitives/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this library adheres to Rust's notion of
 ### Added
 - `zcash_primitives::sapling::keys::DiversifiableFullViewingKey`
 - `zcash_primitives::sapling::keys::Scope`
+- `zcash_primitives::sapling::NullifierDerivingKey`
+- `zcash_primitives::sapling::DiversifiableFullViewingKey::to_nk`
```
str4d, line 13: If this is the same `DiversifiableFullViewingKey` as was added above then we can simplify this by not mentioning `to_nk` as it's part of the same new type (I hadn't seen the changelog contents when making my prior comment, whereas now that there are other things mentioned here it's visible in the diff). Non-blocking, better to have more listed here than less, and we can clean this up before the next release.

zcash_primitives/src/sapling.rs
```diff
@@ -185,15 +185,36 @@ impl ProofGenerationKey {
...
         ViewingKey {
             ak: self.ak,
-            nk: constants::PROOF_GENERATION_KEY_GENERATOR * self.nsk,
+            nk: NullifierDerivingKey(constants::PROOF_GENERATION_KEY_GENERATOR * self.nsk),
         }
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct NullifierDerivingKey(jubjub::SubgroupPoint);
```
str4d, line 194: This is still missing documentation.

zcash_primitives/src/sapling.rs
```diff
@@ -185,15 +185,36 @@ impl ProofGenerationKey {
...
+
+impl From<jubjub::SubgroupPoint> for NullifierDerivingKey {
+    fn from(p: jubjub::SubgroupPoint) -> Self {
+        NullifierDerivingKey(p)
+    }
+}
```
str4d, line 200: Hmm.. _do_ we want to expose this capability (remembering that `From` impls will _always_ be public and can't be scoped down)? Where would it be used (outside of this crate's code where we ensure it is constructed correctly from the FVK)?

The only rationale I can see for having either `From` impl is to preserve the current API of `ViewingKey` (where it can be constructed from arbitrary `jubjub::SubgroupPoint`s, but that's not something I want to allow long-term).

zcash_primitives/src/sapling.rs
```diff
@@ -185,15 +185,36 @@ impl ProofGenerationKey {
...
+
+impl From<NullifierDerivingKey> for jubjub::SubgroupPoint {
+    fn from(nk: NullifierDerivingKey) -> Self {
+        nk.0
+    }
+}
```
str4d, line 206: I don't see this used anywhere. I presume it was added for parity with the other `From` impl, so if that one is removed then this likely also should be.

zcash_primitives/src/sapling.rs
```diff
@@ -185,15 +185,36 @@ impl ProofGenerationKey {
...
+
+impl NullifierDerivingKey {
+    pub fn to_bytes(&self) -> [u8; 32] {
```
str4d, line 209: Documentation. Also should this be public? Are we likely to want to serialize this independently of doing so inside other viewing key serializations? (Presumably no given that we can't do anything with the encoding, as no `from_bytes` method is present.)

str4d (approved):
utACK with documentation nits and API concerns.

zcash_primitives/src/sapling.rs
nuttycom: It's used here: https://github.com/zcash/librustzcash/pull/589/commits/44b808d9cfe032b1ab3b5b2d4677a7f373b6b132#diff-65eea7ee46cbc6fdbcd35372d1907c9fe1c39f4e49e0cc760bd0662f25e0941bL284-R594

zcash_primitives/src/sapling.rs
nuttycom: I'll drop this back out. I initially added it with the idea that it helped a bit with encapsulation, but you're right that it's only really relevant inside of ZIP 32.

zcash_primitives/src/sapling.rs
nuttycom: I can just make the internal value crate-public, I guess. So nothing outside the crate can construct one of these.

zcash_client_backend/CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this library adheres to Rust's notion of
...
+
+### Changed
+- The api of `welding_rig::ScanningKey` has changed to accommodate batch
+  decryption and to correctly handle scanning with the internal (change) keys
+  derived from ZIP 316 UFVKs and UIVKs.
+- `welding_rig::scan_block` now uses batching for trial-decryption of 
+  transaction outputs.
+
+## [0.6.0] - 2022-05-11
+
```
str4d, line 18: This also seems to have slipped in; we have not cut a `zcash_client_backend 0.6.0`.

```suggestion
```

The `Changed` section should be merged into the one below.

str4d (approved):
utACK 67bc0d3c2a4696e3425daa80dff83edffe13c316 modulo one small issue I found in the `zcash_client_backend` changelog.

str4d (approved):
utACK 8a6e979cb9a9f16cc975dbf7bbaa79032d4bd91e

Issue #591, opened 2022-07-28
Change the cache DB format for `zcash_client_sqlite`


str4d:
Currently we store compact blocks in rows of a SQLite table (specifically `(height, blockbytes)`), in a database specifically for caching. @ccjernigan noted that this is quite inefficient when the compact blocks become large, because SQLite is not good at data reclamation. We can "easily" reclaim that data by just deleting the database file, but this becomes more complex on e.g. Android (which uses Room for managing its SQLite databases).

In a meeting today with @ccjernigan, @pacu, @nuttycom and myself, we decided to define a new cache DB system. Compact blocks will be written to a known on-disk path with a known filename format `HEIGHT-BLOCKHASHHEX`, and then metadata about the block (which is much smaller) is what gets stored in the SQLite database. We specifically want the following metadata:
- Height
- Block hash
- Time
- number of Sapling outputs
- number of Orchard outputs
  - These two are for figuring out how blocks line up against subtree boundaries in the note commitment trees.

The SQLite cache DB will be managed on the Rust side instead of the SDK side. The data flow will be:
- When the cache is being filled:
  - SDK requests via FFI the current height at which cached data is stored.
    - If no cached data, SDK can request the wallet's synced-to height instead.
  - SDK rounds the height down to some 100- / 1000-block boundary (for privacy).
  - SDK calls the `getblock` gRPC method and gets a stream of compact blocks.
  - SDK writes each compact block to the defined filesystem folder as a separate file.
    - Overwrites are fine because we assume block hashes are both immutable and do not collide.
  - Once "enough" blocks have been written to disk successfully, SDK provides the list of corresponding metadata to Rust via the FFI.
  - Rust updates the cache DB with this metadata atomically.
- When Rust is later told to scan the cache for wallet data:
  - It reads the cache DB to obtain `(height, hash)`.
  - It reads the corresponding file `height-hash` from the defined filesystem folder and parses the compact block.
  - It proceeds with scanning the compact block as usual.
- When part of the cache is no longer needed:
  - SDK determines some height `H` that is the earliest block data it needs to preserve.
    - This might be determined based on where the wallet is fully-synced to, or other heuristics.
  - SDK searches the defined filesystem folder for all files beginning in `HEIGHT-*` where `HEIGHT < H`, and deletes those files.

[...]

PR #596, opened 2022-07-29
Add a BlockSource implementation that reads blocks from files in the filesystem.

nuttycom:
Also, this adds functionality to create and insert records into the
block metadata db that is used as the source for which blocks to read.


zcash_client_sqlite/CHANGELOG.md
```diff
@@ -16,6 +15,13 @@ and this library adheres to Rust's notion of
     transparent address decoding.
   - `SqliteClientError::RequestedRewindInvalid`, to report when requested
     rewinds exceed supported bounds.
+- A filesystem-backed `BlockSource` implementation
+  `zcash_clinet_sqlite::FsBlockDb`.  This block source expects blocks to be
+  stored on disk in individual files named following the pattern
+  `<blocks_dir>/<blockheight>-<blockhash>`. A SQLite database stores metadata
```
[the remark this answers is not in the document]
nuttycom: Should the `compactblock`/`nullifiers` part of the name be a prefix or a suffix? I don't know how much sorting matters, but should we sort files of the same type together, or files for the same block together?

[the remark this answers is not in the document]
nuttycom: I've changed the filename to `compactblock-<blockheight>-<blockhash>`

zcash_client_sqlite/src/chain.rs
```diff
@@ -64,6 +67,121 @@ where
...
+    mut with_block: F,
+) -> Result<(), SqliteClientError>
+where
+    F: FnMut(CompactBlock) -> Result<(), SqliteClientError>,
+{
+    // Fetch the CompactBlocks we need to scan
+    let mut stmt_blocks = cache.conn.prepare(
+        "SELECT height, blockhash, time, sapling_outputs_count, orchard_actions_count
+         FROM compactblocks_meta
+         WHERE height > ?
```
daira, line 143: The doc comment says "starting at `from_height`", which implies `from_height` should be included.
```suggestion
         WHERE height >= ?
```

nuttycom: Hm. This is currently written to be consistent with the existing behavior: https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/chain.rs#L36

I agree that the name of the parameter is misleading, but we should be consistent, and I think it's hazardous to change this now.

nuttycom: Renamed to `last_scanned_height` and updated the documentation accordingly.

str4d: To be clear, there is no error here. The SDK ingress points correctly ensure that this ends up being set to the height of the last validated block, _from which_ we want to scan. English is hard.

str4d (changes requested):

zcash_client_sqlite/src/lib.rs
```diff
@@ -717,7 +719,42 @@ impl BlockSource for BlockDb {
...
+#[derive(Debug)]
+pub enum FsBlockDbError<P: AsRef<Path>> {
+    FsError(io::Error),
+    DbError(rusqlite::Error),
+    InvalidBlocksDir(P),
+    InvalidBlockPath(P),
+}
+
+impl<P: AsRef<Path>> FsBlockDb<P> {
+    pub fn for_paths(db_path: P, blocks_dir: P) -> Result<Self, FsBlockDbError<P>> {
```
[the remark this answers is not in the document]
nuttycom: I opted to instead now take only a single path, and hard-code the child paths.

zcash_client_sqlite/src/chain/migrations/blockmeta.rs
```diff
@@ -0,0 +1,54 @@
...
+    use schemer::{self, migration};
+    use schemer_rusqlite::RusqliteMigration;
+    use uuid::Uuid;
+
+    pub struct Migration;
+
+    /// The migration that removed the address columns from the `accounts` table, and created
+    /// the `accounts` table.
+    ///
+    /// d956978c-9c87-4d6e-815d-fb8f088d094c
```
str4d, line 18:
```suggestion
    /// The migration that added the `compactblocks_meta` table.
    ///
    /// 68525b40-36e5-46aa-a765-720f8389b99d
```

str4d (changes requested):
Reviewed 0ba1ea5b9916deb5f1241c04fb1fb34cc340d67e. Other than my comment on the filename format and my unanswered question, all of my comments are about comments.

zcash_client_sqlite/src/chain.rs
```diff
@@ -64,6 +76,148 @@ where
...
+    pub block_time: u32,
+    pub sapling_outputs_count: u32,
+    pub orchard_actions_count: u32,
+}
+
+#[cfg(feature = "unstable")]
+impl BlockMeta {
+    pub fn block_file_path<P: AsRef<Path>>(&self, blocks_dir: &P) -> PathBuf {
+        blocks_dir.as_ref().join(Path::new(&format!(
+            "compactblock-{}-{}",
```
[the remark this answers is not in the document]
nuttycom: Ideally humans aren't going to be looking at this. Even so, I think that grouping by type is more valuable than grouping by block, particularly since when we start doing out-of-order operations, it will be commonplace to have some parts for some blocks and not others, and in that case one will likely be grepping the output of `ls` for a particular block height or block hash anyway.

zcash_client_sqlite/src/lib.rs
```diff
@@ -739,7 +746,141 @@ impl BlockSource for BlockDb {
...
+///
+/// where `<block_height>` is the decimal value of the height at which the block was mined, and
+/// `<block_hash>` is the hexidecimal representaion of the block hash, as produced by the
+/// [`Display`] implementation for [`zcash_primitives::block::BlockHash`].
+///
+/// This block source is intended to be used with the following data flow:
+/// * When the cache is being filled:
+///   * The caller requests the current maximum height height at which cached data is available
+///     using [`FsBlockDb::get_max_cached_height`]. If no cached data is available, the caller
+///     can should use the wallet's synced-to height for the following operations instead.
```
nuttycom, line 770:
```suggestion
///     can use the wallet's synced-to height for the following operations instead.
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -739,7 +746,141 @@ impl BlockSource for BlockDb {
...
+    /// Returns the maximum height of blocks known to the block metadata database.
+    pub fn get_max_cached_height(&self) -> Result<Option<BlockHeight>, FsBlockDbError> {
+        Ok(chain::blockmetadb_get_max_cached_height(&self.conn)?)
+    }
+
+    /// Adds a set of block metadata entries to the metadata database.
+    ///
+    /// This will return an error if any block file corresponding to one of these metadata records
+    /// is absent from the blocks directory.
+    pub fn write_block_metadata(&self, block_meta: &[BlockMeta]) -> Result<(), FsBlockDbError> {
```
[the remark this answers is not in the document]
nuttycom: This is the behavior that @ccjernigan had originally requested. I don't have a strong opinion as to whether we should re-parse the block data to recover the block time and the Sapling and Orchard output counts, but it seems superfluous to me when the caller will have this information anyway. IIRC we don't actually use those extra fields for anything at the moment, so the most minimal implementation would elide everything but the block height and the block hash.

str4d (approved):
utACK 4868a377b903479b47d02d90b3d9402eb34ce7fd

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -25,13 +25,21 @@ and this library adheres to Rust's notion of
...
   be replaced by accessors provided by the data access API at some point
   in the future:
   - `v_transactions`
   - `v_tx_received`
   - `v_tx_sent`
 - `zcash_client_sqlite::wallet::init::WalletMigrationError`
+- A filesystem-backed `BlockSource` implementation
+  `zcash_client_sqlite::FsBlockDb`. This block source expects blocks to be
+  stored on disk in individual files named following the pattern
+  `<blockmeta_root>/blocks/compactblock-<blockheight>-<blockhash>`. A SQLite
```
ccjernigan, line 38: There are some differences here compared to our prior discussion in the [GitHub issue](https://github.com/zcash/librustzcash/issues/591).  Were there additional considerations that came up as you were implementing this?

The implementation should work as-is.  However the primary key is the the blockheight-blockhash, and to me it seems more logical to put those in the front.  The compactblock suffix is more like a filetype.

Thoughts?

ccjernigan, line 38: This is very minor and can be ignored.  However I had an Opinion to share.

I've gone back and forth over the years about singular versus plural for group names (either directories or java packages).  I think the singular is probably better because not all words pluralize, not all things should be pluralized, and pluralization goes against the underlying type which has a singular name.  It becomes more obvious when looking at a whole lot of java packages than when adding just a single directory.

E.g. consider Java packages with:
 - co.electriccoin.zcash.models - For typesafe model objects
 - co.electriccoin.zcash.db - For a single database, but which has multiple helper classes and thus justifies a singular package.

From that example, I think making models singular enhances the naming consistency and avoids introducing an additional complex naming rule (we only pluralize when it contains multiple items of the same type) by sticking with a simple naming rule (we always use singular).

Again, this is very minor and can be ignored as it has no impact on correctness.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/4868a377b903479b47d02d90b3d9402eb34ce7fd..f8fd27653cd108e54a8901ac3edf0f3f2cb39b32) to change the order of filename components.

PR #600, opened 2022-08-02
Add database migrations to zcash_client_sqlite.

nuttycom:
This change adds database migration infrastructure to zcash_client_sqlite,
and replaces wallet database initialization with running database migrations.

The migrations introduced by this change should be able to safely upgrade 
wallet databases based on `librustzcash` versions 0.3.0 and also any databases
that have already applied the changes to wallet initialization expected by the
09f4d90a32434462fcd37e9f843cb015cf313fa7 lineage of commits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,5 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
+schemer = { git = "https://github.com/aschampion/schemer.git", rev = "2370c96f43eda5b2cb267bbc9355f0fefff850bb" }
+schemer-rusqlite = { git = "https://github.com/aschampion/schemer.git", rev = "2370c96f43eda5b2cb267bbc9355f0fefff850bb" }
--- a/zcash_client_backend/src/encoding.rs
+++ b/zcash_client_backend/src/encoding.rs
@@ -26,15 +26,55 @@ where
     bech32::encode(hrp, data.to_base32(), Variant::Bech32).expect("hrp is invalid")
 }
 
-fn bech32_decode<T, F>(hrp: &str, s: &str, read: F) -> Result<Option<T>, Error>
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Bech32DecodeError {
+    Bech32Error(Error),
+    IncorrectVariant(Variant),
+    ReadError,
+    HrpMismatch { expected: String, actual: String },
+}
+
+impl From<Error> for Bech32DecodeError {
+    fn from(err: Error) -> Self {
+        Bech32DecodeError::Bech32Error(err)
+    }
+}
+
+impl fmt::Display for Bech32DecodeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self {
+            Bech32DecodeError::Bech32Error(e) => write!(f, "{}", e),
+            Bech32DecodeError::IncorrectVariant(variant) => write!(
+                f,
+                "Incorrect bech32 encoding (wrong variant: {:?})",
+                variant
+            ),
+            Bech32DecodeError::ReadError => {
+                write!(f, "Failed to decode key from its binary representation.")
+            }
+            Bech32DecodeError::HrpMismatch { expected, actual } => write!(
+                f,
+                "Key was encoded for a different network: expected {}, got {}.",
+                expected, actual
+            ),
+        }
+    }
+}
+
+fn bech32_decode<T, F>(hrp: &str, s: &str, read: F) -> Result<T, Bech32DecodeError>
 where
     F: Fn(Vec<u8>) -> Option<T>,
 {
-    match bech32::decode(s)? {
-        (decoded_hrp, data, Variant::Bech32) if decoded_hrp == hrp => {
-            Vec::<u8>::from_base32(&data).map(read)
-        }
-        _ => Ok(None),
+    let (decoded_hrp, data, variant) = bech32::decode(s)?;
+    if variant != Variant::Bech32 {
+        Err(Bech32DecodeError::IncorrectVariant(variant))
+    } else if decoded_hrp != hrp {
+        Err(Bech32DecodeError::HrpMismatch {
+            expected: hrp.to_string(),
+            actual: decoded_hrp,
+        })
+    } else {
+        read(Vec::<u8>::from_base32(&data)?).ok_or(Bech32DecodeError::ReadError)
     }
 }
 
@@ -121,7 +161,7 @@ pub fn encode_extended_spending_key(hrp: &str, extsk: &ExtendedSpendingKey) -> S
 pub fn decode_extended_spending_key(
     hrp: &str,
     s: &str,
-) -> Result<Option<ExtendedSpendingKey>, Error> {
+) -> Result<ExtendedSpendingKey, Bech32DecodeError> {
     bech32_decode(hrp, s, |data| ExtendedSpendingKey::read(&data[..]).ok())
 }
 
@@ -155,7 +195,7 @@ pub fn encode_extended_full_viewing_key(hrp: &str, extfvk: &ExtendedFullViewingK
 pub fn decode_extended_full_viewing_key(
     hrp: &str,
     s: &str,
-) -> Result<Option<ExtendedFullViewingKey>, Error> {
+) -> Result<ExtendedFullViewingKey, Bech32DecodeError> {
     bech32_decode(hrp, s, |data| ExtendedFullViewingKey::read(&data[..]).ok())
 }
 
@@ -240,11 +280,11 @@ pub fn encode_payment_address_p<P: consensus::Parameters>(
 ///         HRP_SAPLING_PAYMENT_ADDRESS,
 ///         "ztestsapling1qqqqqqqqqqqqqqqqqqcguyvaw2vjk4sdyeg0lc970u659lvhqq7t0np6hlup5lusxle75ss7jnk",
 ///     ),
-///     Ok(Some(pa)),
+///     Ok(pa),
 /// );
 /// ```
 /// [`PaymentAddress`]: zcash_primitives::sapling::PaymentAddress
-pub fn decode_payment_address(hrp: &str, s: &str) -> Result<Option<PaymentAddress>, Error> {
+pub fn decode_payment_address(hrp: &str, s: &str) -> Result<PaymentAddress, Bech32DecodeError> {
     bech32_decode(hrp, s, |data| {
         if data.len() != 43 {
             return None;
@@ -392,6 +432,7 @@ mod tests {
     use super::{
         decode_extended_full_viewing_key, decode_extended_spending_key, decode_payment_address,
         encode_extended_full_viewing_key, encode_extended_spending_key, encode_payment_address,
+        Bech32DecodeError,
     };
 
     #[test]
@@ -414,7 +455,7 @@ mod tests {
                 encoded_main
             )
             .unwrap(),
-            Some(extsk.clone())
+            extsk
         );
 
         assert_eq!(
@@ -430,7 +471,7 @@ mod tests {
                 encoded_test
             )
             .unwrap(),
-            Some(extsk)
+            extsk
         );
     }
 
@@ -454,7 +495,7 @@ mod tests {
                 encoded_main
             )
             .unwrap(),
-            Some(extfvk.clone())
+            extfvk
         );
 
         assert_eq!(
@@ -470,7 +511,7 @@ mod tests {
                 encoded_test
             )
             .unwrap(),
-            Some(extfvk)
+            extfvk
         );
     }
 
@@ -500,7 +541,7 @@ mod tests {
                 encoded_main
             )
             .unwrap(),
-            Some(addr.clone())
+            addr
         );
 
         assert_eq!(
@@ -513,7 +554,7 @@ mod tests {
                 encoded_test
             )
             .unwrap(),
-            Some(addr)
+            addr
         );
     }
 
@@ -535,9 +576,8 @@ mod tests {
             decode_payment_address(
                 constants::mainnet::HRP_SAPLING_PAYMENT_ADDRESS,
                 &encoded_main
-            )
-            .unwrap(),
-            None
+            ),
+            Err(Bech32DecodeError::ReadError)
         );
     }
 }
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -19,7 +19,10 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - Various **BREAKING CHANGES** have been made to the database tables. These will
-  require migrations, which may need to be performed in multiple steps.
+  require migrations, which may need to be performed in multiple steps. Migrations
+  will now be automatically performed for any user using
+  `zcash_client_sqlite::wallet::init_wallet_db`and it is recommended to use this
+  method to maintain the state of the database going forward.
   - The `extfvk` column in the `accounts` table has been replaced by a `ufvk`
     column. Values for this column should be derived from the wallet's seed and
     the account number; the Sapling component of the resulting Unified Full
@@ -48,6 +51,11 @@ and this library adheres to Rust's notion of
   method to be used in contexts where a transaction has just been
   constructed, rather than only in the case that a transaction has
   been decrypted after being retrieved from the network.
+- `zcash_client_sqlite::wallet::init_wallet_db` has been modified to
+  take the wallet seed as an argument so that it can correctly perform
+  migrations that require re-deriving key material. In particular for
+  this upgrade, the seed is used to derive UFVKs to replace the currently
+  stored Sapling extfvks as part of the migration process.
 
 ### Removed
 - `zcash_client_sqlite::wallet`:
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -21,7 +21,7 @@ and this library adheres to Rust's notion of
 - Various **BREAKING CHANGES** have been made to the database tables. These will
   require migrations, which may need to be performed in multiple steps. Migrations
   will now be automatically performed for any user using
-  `zcash_client_sqlite::wallet::init_wallet_db`and it is recommended to use this
+  `zcash_client_sqlite::wallet::init_wallet_db` and it is recommended to use this
   method to maintain the state of the database going forward.
   - The `extfvk` column in the `accounts` table has been replaced by a `ufvk`
     column. Values for this column should be derived from the wallet's seed and
@@ -55,7 +55,8 @@ and this library adheres to Rust's notion of
   take the wallet seed as an argument so that it can correctly perform
   migrations that require re-deriving key material. In particular for
   this upgrade, the seed is used to derive UFVKs to replace the currently
-  stored Sapling extfvks as part of the migration process.
+  stored Sapling extfvks (without loss of information) as part of the
+  migration process.
 
 ### Removed
 - `zcash_client_sqlite::wallet`:
--- a/zcash_client_sqlite/src/error.rs
+++ b/zcash_client_sqlite/src/error.rs
@@ -3,7 +3,10 @@
 use std::error;
 use std::fmt;
 
-use zcash_client_backend::{data_api, encoding::TransparentCodecError};
+use zcash_client_backend::{
+    data_api,
+    encoding::{Bech32DecodeError, TransparentCodecError},
+};
 use zcash_primitives::consensus::BlockHeight;
 
 use crate::{NoteId, PRUNING_HEIGHT};
@@ -27,8 +30,8 @@ pub enum SqliteClientError {
     /// Illegal attempt to reinitialize an already-initialized wallet database.
     TableNotEmpty,
 
-    /// Bech32 decoding error
-    Bech32(bech32::Error),
+    /// A Bech32-encoded key or address decoding error
+    Bech32DecodeError(Bech32DecodeError),
 
     /// Base58 decoding error
     Base58(bs58::decode::Error),
@@ -58,7 +61,7 @@ impl error::Error for SqliteClientError {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match &self {
             SqliteClientError::InvalidMemo(e) => Some(e),
-            SqliteClientError::Bech32(e) => Some(e),
+            SqliteClientError::Bech32DecodeError(Bech32DecodeError::Bech32Error(e)) => Some(e),
             SqliteClientError::DbError(e) => Some(e),
             SqliteClientError::Io(e) => Some(e),
             _ => None,
@@ -78,7 +81,7 @@ impl fmt::Display for SqliteClientError {
                 write!(f, "The note ID associated with an inserted witness must correspond to a received note."),
             SqliteClientError::RequestedRewindInvalid(h, r) =>
                 write!(f, "A rewind must be either of less than {} blocks, or at least back to block {} for your wallet; the requested height was {}.", PRUNING_HEIGHT, h, r),
-            SqliteClientError::Bech32(e) => write!(f, "{}", e),
+            SqliteClientError::Bech32DecodeError(e) => write!(f, "{}", e),
             SqliteClientError::Base58(e) => write!(f, "{}", e),
             SqliteClientError::TransparentAddress(e) => write!(f, "{}", e),
             SqliteClientError::TableNotEmpty => write!(f, "Table is not empty"),
@@ -102,9 +105,9 @@ impl From<std::io::Error> for SqliteClientError {
     }
 }
 
-impl From<bech32::Error> for SqliteClientError {
-    fn from(e: bech32::Error) -> Self {
-        SqliteClientError::Bech32(e)
+impl From<Bech32DecodeError> for SqliteClientError {
+    fn from(e: Bech32DecodeError) -> Self {
+        SqliteClientError::Bech32DecodeError(e)
     }
 }
 
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -197,13 +197,7 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
         for row in rows {
             let (account, address) = row?;
             let decoded_address =
-                decode_payment_address(self.params.hrp_sapling_payment_address(), &address)?
-                    .ok_or_else(|| {
-                        SqliteClientError::CorruptedData(format!(
-                            "Not a valid Sapling payment address: {}",
-                            address
-                        ))
-                    })?;
+                decode_payment_address(self.params.hrp_sapling_payment_address(), &address)?;
 
             let usk =
                 UnifiedSpendingKey::from_seed(&self.params, self.seed.expose_secret(), account)
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -229,7 +229,8 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
         //
 
         transaction.execute_batch(
-            "CREATE TABLE sent_notes_new (
+            "PRAGMA foreign_keys = ON;
+            CREATE TABLE sent_notes_new (
                 id_note INTEGER PRIMARY KEY,
                 tx INTEGER NOT NULL,
                 output_pool INTEGER NOT NULL ,
@@ -273,7 +274,8 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
         }
 
         transaction.execute_batch(
-            "DROP TABLE sent_notes;
+            "PRAGMA foreign_keys = OFF;
+            DROP TABLE sent_notes;
             ALTER TABLE sent_notes_new RENAME TO sent_notes;
             PRAGMA foreign_keys = ON;",
         )?;
@@ -540,7 +542,7 @@ mod tests {
             account: AccountId,
         ) -> Result<(), rusqlite::Error> {
             wdb.conn.execute(
-                "CREATE TABLE IF NOT EXISTS accounts (
+                "CREATE TABLE accounts (
                     account INTEGER PRIMARY KEY,
                     extfvk TEXT NOT NULL,
                     address TEXT NOT NULL
@@ -548,7 +550,7 @@ mod tests {
                 NO_PARAMS,
             )?;
             wdb.conn.execute(
-                "CREATE TABLE IF NOT EXISTS blocks (
+                "CREATE TABLE blocks (
                     height INTEGER PRIMARY KEY,
                     hash BLOB NOT NULL,
                     time INTEGER NOT NULL,
@@ -557,7 +559,7 @@ mod tests {
                 NO_PARAMS,
             )?;
             wdb.conn.execute(
-                "CREATE TABLE IF NOT EXISTS transactions (
+                "CREATE TABLE transactions (
                     id_tx INTEGER PRIMARY KEY,
                     txid BLOB NOT NULL UNIQUE,
                     created TEXT,
@@ -570,7 +572,7 @@ mod tests {
                 NO_PARAMS,
             )?;
             wdb.conn.execute(
-                "CREATE TABLE IF NOT EXISTS received_notes (
+                "CREATE TABLE received_notes (
                     id_note INTEGER PRIMARY KEY,
                     tx INTEGER NOT NULL,
                     output_index INTEGER NOT NULL,
@@ -590,7 +592,7 @@ mod tests {
                 NO_PARAMS,
             )?;
             wdb.conn.execute(
-                "CREATE TABLE IF NOT EXISTS sapling_witnesses (
+                "CREATE TABLE sapling_witnesses (
                     id_witness INTEGER PRIMARY KEY,
                     note INTEGER NOT NULL,
                     block INTEGER NOT NULL,
@@ -602,7 +604,7 @@ mod tests {
                 NO_PARAMS,
             )?;
             wdb.conn.execute(
-                "CREATE TABLE IF NOT EXISTS sent_notes (
+                "CREATE TABLE sent_notes (
                     id_note INTEGER PRIMARY KEY,
                     tx INTEGER NOT NULL,
                     output_index INTEGER NOT NULL,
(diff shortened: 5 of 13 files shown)


pacu (approved):
I understand that this PR is to introduce migrations into the crate without considering the migrations that were conducted by clients like zcash-android-wallet-sdk or ZcashLightClientKit and that those will be cover, if needed on other PRs.


that being said. utACK !

Cargo.toml
```diff
@@ -21,3 +21,5 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
+schemer = { git = "https://github.com/nuttycom/schemer.git", rev = "b3f7c8a3abb69b67d68258ae5284c5743bdb380b" }
```
ccjernigan, line 24: Should we we relying on a repo fork that is coming from your personal github account?

A denial of service could occur through a failed build if your GitHub user account is deleted or you delete the repo.

Do we have a place we fork things that are kept under the zcash org?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,6 +25,240 @@ use {
...
+
+    fn up(&self, transaction: &Transaction) -> Result<(), SqliteClientError> {
+        transaction.execute_batch(
+            "CREATE TABLE IF NOT EXISTS accounts (
+                account INTEGER PRIMARY KEY,
+                extfvk TEXT NOT NULL,
+                address TEXT NOT NULL
+            );
+            CREATE TABLE IF NOT EXISTS blocks (
+                height INTEGER PRIMARY KEY,
```
ccjernigan, line 48: Since we have stronger types elsewhere in our codebase, do we ever consider constraints on some columns like height which could be confined to [0, UInt32.max]?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,6 +25,240 @@ use {
...
+    WalletMigration0,
+    "bc4f5e57-d600-4b6c-990f-b3538f0bfce1",
+    [],
+    "Initialize the wallet database."
+);
+
+impl RusqliteMigration for WalletMigration0 {
+    type Error = SqliteClientError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), SqliteClientError> {
```
ccjernigan, line 40: Since I don't know exactly how we use this data, this is an open-ended question.  Are there any indexes that would help with querying for data?  INTEGER PRIMARY KEY and UNIQUE columns are already indexed by default but I'm curious if there are any other columns that we would benefit by indexing?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,6 +25,240 @@ use {
...
+impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
+    type Error = SqliteClientError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), SqliteClientError> {
+        //
+        // Update the accounts table to store ufvks rather than extfvks
+        //
+
+        transaction.execute_batch(
+            "PRAGMA foreign_keys = OFF;
```
ccjernigan, line 179: Would it make sense to move the pragmas into their own transaction?

I'm just thinking of making it visually easier for humans to balance the foreign keys being turned on/off here and at the very end.  (The one at the end is easy to miss since its tucked inside a bunch of other SQL).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,6 +25,240 @@ use {
...
+                FOREIGN KEY (from_account) REFERENCES accounts(account),
+                CONSTRAINT tx_output UNIQUE (tx, output_pool, output_index)
+            );
+            INSERT INTO sent_notes_new 
+                (id_note, tx, output_pool, output_index, from_account, address, value, memo)
+                SELECT id_note, tx, 2, output_index, from_account, address, value, memo 
+                FROM sent_notes;
+            DROP TABLE sent_notes;
+            ALTER TABLE sent_notes_new RENAME TO sent_notes;
+            PRAGMA foreign_keys = ON;",
```
ccjernigan, line 250: If we don't make it all the way here, does the database potentially end up in a weird state where foreign keys are turned off?

Is there a kind of try-finally we can use to ensure that foreign keys are turned back on regardless of whether the migration succeeds?

I imagine a potential recovery mode if migration fails is to drop the tables, recreate the tables, then rescan the wallet since all data should be recoverable given time+seed.  If we ever do some kind of recovery, it would be nice to ensure that foreign keys aren't turned of by accident here.

ccjernigan (commented):
Since I lack s some context around these changes I have some questions to consider although I don't think any are truly blocking.

Cargo.toml
pacu: We can probably use  "zcash-hackworks" if it's something that's in the works

nuttycom: We can fork it either to `zcash` or `zcash-hackworks`. Depending upon whether the original maintainer wants to bother with accepting patches, we might also publish it as our own crate (it's MIT/Apache2 so we just need to attribute properly.)

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Everything in a migration runs in a single transaction, with multiple statements. I'd be fine with splitting the text here, but it's purely a textual change, so I didn't see much of a reason for it.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Everything is in the scope of a single SQLite transaction; I'll need to check to see whether SQLite DDL operations are transactional.

zcash_client_sqlite/src/wallet/init.rs
ccjernigan: Yes this is correct from the machine perspective.

It is a suggestion for humans reading the code, rather than that machine.  

If we ever have to modify this code, I want the human reading it to be able to balance the `"PRAGMA foreign_keys = OFF;` at the beginning with another one at the end.  Right now, these statements hide and don't stand out so a human reader could miss it.

ccjernigan: This is non-blocking, but a small change here improves readability.

Cargo.toml
str4d: The original maintainer [seems interested in these changes](https://github.com/aschampion/schemer/pull/13#issuecomment-1203774567), so I think it's fine for now to rely on a fork. If we don't want to have that fork be under @nuttycom's repo (though we _did_ do that for temporary `incrementalmerkletree` changes that were merged, due to them referring to an open PR on a fork of our repo), then I'd make it under `zcash-hackworks`, but given that this is pinning a revision rather than a branch, I'm not too concerned (the moment anyone builds this branch locally, they will have a cached copy of the git repo in their `~/.cargo` subdir, so we can recreate it in the event of repo loss).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -40,106 +280,34 @@ use {
...
-            height INTEGER NOT NULL,
-            spent_in_tx INTEGER,
-            FOREIGN KEY (spent_in_tx) REFERENCES transactions(id_tx),
-            CONSTRAINT tx_outpoint UNIQUE (prevout_txid, prevout_idx)
-        )",
-        NO_PARAMS,
-    )?;
+pub fn init_wallet_db<P: consensus::Parameters + 'static>(
+    wdb: &mut WalletDb<P>,
+    seed: Vec<u8>,
```
str4d, line 288:
```suggestion
    seed: SecretVec<u8>,
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,6 +25,240 @@ use {
...
+    fn up(&self, transaction: &Transaction) -> Result<(), SqliteClientError> {
+        //
+        // Update the accounts table to store ufvks rather than extfvks
+        //
+
+        transaction.execute_batch(
+            "PRAGMA foreign_keys = OFF;
+            CREATE TABLE IF NOT EXISTS accounts_new (
+                account INTEGER PRIMARY KEY,
+                ufvk TEXT NOT NULL,
```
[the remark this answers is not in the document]
nuttycom: I never attempt to construct the `autoshielding` state; I just ensure that we can safely migrate *from* that state.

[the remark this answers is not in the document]
str4d: That's not what I mean. What I'm saying is we should confirm that, from the state that the mobile SDKs are currently in, we should be able to reach a successful end state.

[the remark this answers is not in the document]
str4d: I'm also saying that the end result of this migration step is `ufvk TEXT NOT NULL`, which is _not_ equivalent to what current `main` has (`ufvk TEXT`). If that is unintentional (i.e. it is not intended that these migrations be applied to current `main` schema) then we should remove the `NOT NULL` so we don't have two subtly different schemas deployed in wallets after this merges. But if the intent is to apply the migrations to _any_ of the three current schemas, so that they all end up in a single fourth schema, that's fine.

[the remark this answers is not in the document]
nuttycom: Okay, yes. I think that the right thing to do here is add a test that just manually constructs each current state, and performs the migration on each such state.

The addition of the `NOT NULL` constraint is intentional; there's no way to have a valid row in this table if we can't obtain the UFVK.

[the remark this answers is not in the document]
nuttycom: Also, the migration *must* be applied to any database derived from the current `main` schema, in order to correctly initialize the migration state.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/ce50831407ccc97a4ca6518f5a4378b4609b0f0c..b15c4747f41cce184634dc3dceaf08ea3d3929a4) to address review comments, add `secrets::SecretVec` wrappers for seed values, and add tests.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/2280fc1c457e4bf05d42f659e1f2228a84041ff2..0f433fc38959f0be39b89220e7acaf9e7567f4d2) to address test failures and fix clippy lints.

Cargo.toml
```diff
@@ -21,3 +21,5 @@ codegen-units = 1
...
+schemer-rusqlite = { git = "https://github.com/nuttycom/schemer.git", rev = "b3f7c8a3abb69b67d68258ae5284c5743bdb380b" }
```
daira, line 25: Why do these need to be forked? I cannot see any differences between these forks, and [schemer](https://crates.io/crates/schemer) / [schemer-rusqlite](https://crates.io/crates/schemer-rusqlite) 0.1.1, apart from https://github.com/aschampion/schemer/commit/cd224eb325f5bcfd9150520546a48f24ddf6e66b (which was made after the schemer 0.1.2 release but does not change behaviour at all).

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -19,7 +19,10 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - Various **BREAKING CHANGES** have been made to the database tables. These will
-  require migrations, which may need to be performed in multiple steps.
+  require migrations, which may need to be performed in multiple steps. Migrations
+  will now be automatically performed for any user using
+  `zcash_client_sqlite::wallet::init_wallet_db`and it is recommended to use this
```
daira, line 24:
```suggestion
  `zcash_client_sqlite::wallet::init_wallet_db` and it is recommended to use this
```

zcash_client_sqlite/src/error.rs
```diff
@@ -58,7 +61,7 @@ impl error::Error for SqliteClientError {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match &self {
             SqliteClientError::InvalidMemo(e) => Some(e),
-            SqliteClientError::Bech32(e) => Some(e),
+            SqliteClientError::KeyError(KeyError::Bech32Error(e)) => Some(e),
```
daira, line 64: Why only the `KeyError::Bech32Error` case?

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -48,6 +51,11 @@ and this library adheres to Rust's notion of
   method to be used in contexts where a transaction has just been
   constructed, rather than only in the case that a transaction has
   been decrypted after being retrieved from the network.
+- `zcash_client_sqlite::wallet::init_wallet_db` has been modified to
+  take the wallet seed as an argument so that it can correctly perform
+  migrations that require re-deriving key material. In particular for
+  this upgrade, the seed is used to derive UFVKs to replace the currently
+  stored Sapling extfvks as part of the migration process.
```
daira, line 58:
```suggestion
  stored Sapling extfvks (without loss of information) as part of the
  migration process.
```

zcash_client_sqlite/src/wallet/init.rs
daira: `blocks.hash` should probably be indexed I think.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,19 +26,286 @@ use {
...
+        Ok(())
+    }
+}
+
+struct WalletMigration2<P: consensus::Parameters> {
+    params: P,
+    seed: SecretVec<u8>,
+}
+
+impl<P: consensus::Parameters> Migration for WalletMigration2<P> {
```
daira, line 154: Why isn't this expressible using the `migration!` macro?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,19 +26,286 @@ use {
...
+impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
+    type Error = SqliteClientError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), SqliteClientError> {
+        //
+        // Update the accounts table to store ufvks rather than extfvks
+        //
+
+        transaction.execute_batch(
+            "PRAGMA foreign_keys = OFF;
```
daira, line 180: https://www.sqlite.org/pragma.html#pragma_foreign_keys says:
> This pragma is a no-op within a transaction; foreign key constraint enforcement may only be enabled or disabled when there is no pending [BEGIN](https://www.sqlite.org/lang_transaction.html) or [SAVEPOINT](https://www.sqlite.org/lang_savepoint.html).

Cargo.toml
nuttycom: The changes are those represented by these PRs: https://github.com/aschampion/schemer/pull/13, https://github.com/aschampion/schemer/pull/14

daira: 👍

zcash_client_sqlite/src/wallet/init.rs
nuttycom: I'm not sure what to think about this; I know that I had to add this pragma in order to avoid getting a foreign key constraint violation when moving the new accounts table into place.

zcash_client_sqlite/src/error.rs
nuttycom: That's the only one of the existing error types that implements `error::Error`; we don't implement that for any of the other variants of SqliteClientError that we control, so I also did not implement it here.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: At present we don't ever look up blocks by their hash or join on that column, which is I think why we don't have an index on that column.

nuttycom: A unique constraint does make sense though.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: If we want to do this, we should do so as a separate migration, for the same reason as is described in https://github.com/zcash/librustzcash/pull/600/files/8575d74dbb4af54cb45c1f1585f39c23754c3c8a#r946220645

zcash_client_sqlite/src/wallet/init.rs
nuttycom: It's because the macro doesn't deal properly with the `<P>` type parameter.

zcash_client_sqlite/src/wallet/init.rs
daira: Huh. Maybe the sqlite documentation is wrong? It also says:
> As of SQLite [version 3.6.19](https://www.sqlite.org/releaselog/3_6_19.html), the default setting for foreign key enforcement is OFF. However, that might change in a future release of SQLite.

(`rusqlite` is [version 0.24 used with the "bundled" feature](https://github.com/zcash/librustzcash/blob/276bcc34f58dd2b97a9685e3fbe20f583c24a341/zcash_client_sqlite/Cargo.toml#L23) which I believe [bundles](https://github.com/rusqlite/rusqlite/tree/v0.24.2/libsqlite3-sys/sqlite3) sqlite 3.33.0.)

The default being OFF isn't consistent with needing to explicitly set it to OFF to avoid a constraint violation, so I'm confused. I guess there's no harm in including this pragma, though.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -19,19 +26,286 @@ use {
...
+            Ok((AccountId::from(account), address))
+        })?;
+
+        for row in rows {
+            let (account, address) = row?;
+            let decoded_address =
+                decode_payment_address(self.params.hrp_sapling_payment_address(), &address)?;
+
+            let usk =
+                UnifiedSpendingKey::from_seed(&self.params, self.seed.expose_secret(), account)
```
daira, line 204: This is potentially doing a lot of recomputation, but I think it's fine.

zcash_client_sqlite/src/wallet/init.rs
daira: Note that scoping the pragma more narrowly:
```sql
PRAGMA foreign_keys = ON;

CREATE TABLE sent_notes_new [...];
INSERT INTO sent_notes_new [...];

PRAGMA foreign_keys = OFF;
DROP TABLE sent_notes;
ALTER TABLE sent_notes_new RENAME TO sent_notes;
PRAGMA foreign_keys = ON;
```
has a different effect because it enforces the foreign key constraints while `sent_notes_new` is being created. (Setting the pragma to ON does not check existing rows added when it was OFF, I think.) I believe those constraints should hold for `sent_notes_new`, so this may be preferable.

If this is changed then I would suggest doing the same for `accounts_new`, although that has no foreign key constraints.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -40,106 +314,34 @@ use {
...
-            CONSTRAINT tx_outpoint UNIQUE (prevout_txid, prevout_idx)
-        )",
-        NO_PARAMS,
-    )?;
+// TODO: It would be possible to make the transition from providing transparent support to no
+// longer providing transparent support safe, by including a migration that verifies that no
+// unspent transparent outputs exist in the wallet at the time of upgrading to a version of
+// the library that does not support transparent use. It might be a good idea to add an explicit
+// check for unspent transparent outputs whenever running initialization with a version of the
+// library *not* compiled with the `transparent-inputs` feature flag, and fail if any are present.
```
daira, line 325: Please add an issue for this.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -204,28 +407,43 @@ pub fn init_accounts_table<P: consensus::Parameters>(
...
+            .map(|k| k.default_address().0.encode(network))
+    });
+    #[cfg(not(feature = "transparent-inputs"))]
+    let taddress_str: Option<String> = None;
+
+    conn.execute(
+        &format!(
+            "INSERT INTO {} (account, ufvk, address, transparent_address)
+            VALUES (?, ?, ?, ?)",
+            accounts_table
```
daira, line 439: There's no injection vuln here because `add_account_internal` is private and only called with the strings `accounts` and `accounts_new`.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -310,11 +533,365 @@ mod tests {
...
+        }
+
+        let seed = [0xab; 32];
+        let account = AccountId::from(0);
+        let secret_key = sapling::spending_key(&seed, tests::network().coin_type(), account);
+        let extfvk = ExtendedFullViewingKey::from(&secret_key);
+        let data_file = NamedTempFile::new().unwrap();
+        let mut db_data = WalletDb::for_path(data_file.path(), tests::network()).unwrap();
+        init_0_3_0(&mut db_data, &extfvk, account).unwrap();
+        init_wallet_db(&mut db_data, Secret::new(seed.to_vec())).unwrap();
```
daira, line 649: Do we want to do any checks on the resulting db here (as opposed to checking just that `init_wallet_db` succeeds)?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -310,11 +533,365 @@ mod tests {
...
 
+    #[test]
+    fn init_migrate_from_0_3_0() {
+        fn init_0_3_0<P>(
+            wdb: &mut WalletDb<P>,
+            extfvk: &ExtendedFullViewingKey,
+            account: AccountId,
+        ) -> Result<(), rusqlite::Error> {
+            wdb.conn.execute(
+                "CREATE TABLE IF NOT EXISTS accounts (
```
daira, line 544:
```suggestion
                "CREATE TABLE accounts (
```
These don't need to use `IF NOT EXISTS`, and shouldn't because we want to assert that we're creating these tables from scratch.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -310,11 +533,365 @@ mod tests {
 
     use super::{init_accounts_table, init_blocks_table, init_wallet_db};
 
+    #[test]
+    fn init_migrate_from_0_3_0() {
```
daira, line 537: I have not checked that these tests initialize their respective db states correctly.

daira (approved):
utACK with comments.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: It runs only once, so yeah I think this should be okay.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: If we need additional checks, they should actually be implemented inside the migrations themselves, so that they will abort the migration if our invariants are violated. I've done this with WalletMigration2, and WalletMigration0 and WalletMigration1 don't have any invariants that need to hold.

zcash_client_sqlite/src/wallet/init.rs
ccjernigan: Is this database shared between the Rust and Kotlin code on Android?

For some of the shared databases, our database initialization in the Android SDK is turning on foreign key constraints.  So daira is correct that SQLite has constraints turned off by default, but our SDK is turning them on for the derived data db.

daira (approved):
utACK 1088164b5d3570729b6d7b8e4a6b22e4547a9363

zcash_client_sqlite/src/wallet/init.rs
daira: Do you mean the foreign key constraints? It doesn't look as if those will be enforced for the migration, because of the shipping of `PRAGMA foreign_keys = OFF;` as I pointed out at https://github.com/zcash/librustzcash/pull/600/files#r946241489 .

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Ah, okay, I understand now. Fixed.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Got it. I fixed that scoping of the pragmas now.

zcash_client_sqlite/src/wallet/init.rs
daira: #605

daira (approved):
utACK 982c3f8676554915a5bcf5f1878171f7705ed5d7

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -537,6 +539,360 @@ mod tests {
...
+                )",
+                NO_PARAMS,
+            )?;
+            Ok(())
+        }
+
+        let data_file = NamedTempFile::new().unwrap();
+        let mut db_data = WalletDb::for_path(data_file.path(), tests::network()).unwrap();
+        init_main(&db_data).unwrap();
+        init_wallet_db(&mut db_data, Secret::new(vec![])).unwrap();
```
str4d, line 889: If you make the change to handle UAs in the `address` column, then we can also add a test data case here like for the other two.

zcash_client_backend/src/encoding.rs
```diff
@@ -26,15 +26,55 @@ where
     bech32::encode(hrp, data.to_base32(), Variant::Bech32).expect("hrp is invalid")
 }
 
-fn bech32_decode<T, F>(hrp: &str, s: &str, read: F) -> Result<Option<T>, Error>
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum KeyError {
```
str4d, line 30: This should probably actually be `Bech32DecodeError`, as it is not specific to keys (it can be returned from parsing Sapling payment addresses). Similarly below the messages shouldn't refer to "key", but either "data" or "value".

zcash_client_sqlite/src/error.rs
```diff
@@ -27,8 +30,8 @@ pub enum SqliteClientError {
     /// Illegal attempt to reinitialize an already-initialized wallet database.
     TableNotEmpty,
 
-    /// Bech32 decoding error
-    Bech32(bech32::Error),
+    /// A Bech32-encoded key or address decoding error
+    KeyError(KeyError),
```
str4d, line 34:
```suggestion
    Bech32DecodeError(Bech32DecodeError),
```

zcash_client_backend/src/encoding.rs
```diff
@@ -26,15 +26,55 @@ where
...
+    Bech32Error(Error),
+    IncorrectVariant(Variant),
+    ReadError,
```
str4d, line 33: @nuttycom notes that this also was hard to name. The fact that our `read` functions return `Option<T>` is also eliding parsing error information (somewhat intentionally), but that also suggests that this error is not _quite_ `Bech32DecodeError`, in that the errors are not just from decoding Bech32. It is an error for decoding things that are expected to be encoded as Bech32. For now though, `Bech32DecodeError` is fine as a name for this, and we can revisit this error stack when we refactor the parsing logic.

zcash_client_sqlite/src/error.rs
```diff
@@ -78,7 +81,7 @@ impl fmt::Display for SqliteClientError {
                 write!(f, "The note ID associated with an inserted witness must correspond to a received note."),
             SqliteClientError::RequestedRewindInvalid(h, r) =>
                 write!(f, "A rewind must be either of less than {} blocks, or at least back to block {} for your wallet; the requested height was {}.", PRUNING_HEIGHT, h, r),
-            SqliteClientError::Bech32(e) => write!(f, "{}", e),
+            SqliteClientError::KeyError(e) => write!(f, "{}", e),
```
str4d, line 84: We should ensure we still tell the user what kind of data failed to be parsed:
```suggestion
            SqliteClientError::KeyError(kind, e) => write!(f, "Not a valid {}: {}", kind, e),
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -198,13 +198,7 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
             let (account, address) = row?;
             let decoded_address =
-                decode_payment_address(self.params.hrp_sapling_payment_address(), &address)?
-                    .ok_or_else(|| {
-                        SqliteClientError::CorruptedData(format!(
-                            "Not a valid Sapling payment address: {}",
-                            address
-                        ))
-                    })?;
+                decode_payment_address(self.params.hrp_sapling_payment_address(), &address)?;
```
str4d, line 201: This would then need to use `.map_err` to provide the context `"Sapling payment address", address`.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -230,7 +230,8 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
         //
 
         transaction.execute_batch(
-            "CREATE TABLE sent_notes_new (
+            "PRAGMA foreign_keys = ON;
```
str4d, line 233: Move this into the `DROP TABLE accounts` part of the batch (and also move the `OFF` command into there as well, if we can safely respect foreign keys when inserting into `accounts_new`).

str4d (changes requested):

nuttycom:
[Force-pushed](https://github.com/zcash/librustzcash/compare/982c3f8676554915a5bcf5f1878171f7705ed5d7..ec557ad187b573ce52743cf046eb6f58ddcc9bd7) to address comments from review.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -1,37 +1,368 @@
...
+                let decoded_address =
+                    RecipientAddress::decode(&self.params, &address).ok_or_else(|| {
+                        SqliteClientError::CorruptedData(format!(
+                            "Could not decode {} as a valid Zcash address.",
+                            address
+                        ))
+                    })?;
+                let output_pool = match decoded_address {
+                    RecipientAddress::Shielded(_) => Ok(PoolType::Sapling.typecode()),
+                    RecipientAddress::Transparent(_) => Ok(PoolType::Sapling.typecode()),
```
str4d:
```suggestion
                    RecipientAddress::Transparent(_) => Ok(PoolType::Transparent.typecode()),
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -310,11 +597,422 @@ mod tests {
...
+                    NO_PARAMS,
+                )?;
+                wdb.conn.execute(
+                    "INSERT INTO transactions (block, id_tx, txid) VALUES (0, 0, '')",
+                    NO_PARAMS,
+                )?;
+                wdb.conn.execute(
+                    "INSERT INTO sent_notes (tx, output_pool, output_index, from_account, address, value)
+                    VALUES (0, 0, 0, ?, ?, 0)",
+                    &[u32::from(account).to_sql()?, taddr.to_sql()?])?;
```
str4d, line 996: I have a mild preference to use the `PoolType::Transparent` enum here.

str4d (approved):
utACK f67830ad1661ad4412939440f4b8ef7521b29efa modulo replacing the current `schemer` patches with `schemer 0.2.0` (which it sounds like should get published this week).

Cargo.toml
```diff
@@ -21,3 +21,5 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
+schemer = { git = "https://github.com/aschampion/schemer.git", rev = "2370c96f43eda5b2cb267bbc9355f0fefff850bb" }
+schemer-rusqlite = { git = "https://github.com/aschampion/schemer.git", rev = "2370c96f43eda5b2cb267bbc9355f0fefff850bb" }
```
daira, line 25: I verified that these are the correct hash.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -1,37 +1,425 @@
...
+                transparent_address TEXT
+            );",
+        )?;
+
+        let mut stmt_fetch_accounts =
+            transaction.prepare("SELECT account, address FROM accounts")?;
+
+        let mut rows = stmt_fetch_accounts.query(NO_PARAMS)?;
+        while let Some(row) = rows.next()? {
+            if let Some(seed) = &self.seed {
```
str4d, line 237: We shouldn't run any of the migration if we don't have the seed:
```suggestion
        let seed = self.seed.as_ref().ok_or(WalletMigrationError::SeedRequired)?;

        transaction.execute_batch(
            "CREATE TABLE accounts_new (
                account INTEGER PRIMARY KEY,
                ufvk TEXT NOT NULL,
                address TEXT,
                transparent_address TEXT
            );",
        )?;

        let mut stmt_fetch_accounts =
            transaction.prepare("SELECT account, address FROM accounts")?;

        let mut rows = stmt_fetch_accounts.query(NO_PARAMS)?;
        while let Some(row) = rows.next()? {
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -1,37 +1,425 @@
...
+                add_account_internal::<P, WalletMigrationError>(
+                    &self.params,
+                    transaction,
+                    "accounts_new",
+                    account,
+                    &ufvk,
+                )?;
+            } else {
+                return Err(WalletMigrationError::SeedRequired);
+            }
```
str4d, line 292:
```suggestion
```
(and also undo the indentation).

str4d (changes requested):

zcash_client_sqlite/src/wallet/init.rs
nuttycom: That's not necessarily true; for example, we don't need the seed if we're initializing a brand-new database.

str4d, line 237: Document why the seed is checked inside the row iterator.

Cargo.toml
str4d: I also verified this.

str4d (approved):
utACK 913d5720876794fe26440976024596fd172aadf6. We can merge this before upstream cuts a release, because we are pinning a revision in the upstream repo.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -40,106 +431,40 @@ use {
...
+        params: wdb.params.clone(),
+        seed,
+    });
+
+    migrator
+        .register_multiple(vec![migration0, migration1, migration2])
+        .expect("Wallet migration registration should have been successful.");
+    migrator.up(None)?;
+    wdb.conn
+        .execute("PRAGMA foreign_keys = ON", NO_PARAMS)
```
daira, line 466: This again doesn't enforce foreign_key constraints when creating the `sent_notes_new` table. I don't object, but why was this changed from a previous commitment that did?

nuttycom: The pragmas only work per-connection, not within a transaction (as you identified in the docs) so we now drop constraints prior to all the migrations and reenable them all afterward.

daira: That makes sense as far as it goes, but how was the PR passing tests when the PRAGMA was inside the transaction?

nuttycom: At the time, there was no data in the tables, so the foreign key constraints (at the row level) weren't being checked. I'm not sure why the DDL was succeeding though.

PR #608, opened 2022-08-18
Implement parallel, batched trial decryption in wallet scanning.

nuttycom:
This ports/reimplements the batched decryption approach used in zcashd to implement light wallet scanning.

### DIFF
--- a/zcash_client_backend/src/scan.rs
+++ b/zcash_client_backend/src/scan.rs
@@ -1,0 +1,526 @@
+use core::fmt;
+use std::collections::HashMap;
+use std::io;
+use std::mem;
+
+use crossbeam_channel as channel;
+use group::GroupEncoding;
+use zcash_note_encryption::{batch, BatchDomain, Domain, ShieldedOutput, ENC_CIPHERTEXT_SIZE};
+use zcash_primitives::{
+    block::BlockHash,
+    consensus, constants,
+    sapling::{self, note_encryption::SaplingDomain},
+    transaction::{
+        components::{sapling::GrothProofBytes, OutputDescription},
+        Transaction, TxId,
+    },
+};
+
+#[cxx::bridge]
+mod ffi {
+    #[namespace = "wallet"]
+    struct SaplingDecryptionResult {
+        txid: [u8; 32],
+        output: u32,
+        ivk: [u8; 32],
+        diversifier: [u8; 11],
+        pk_d: [u8; 32],
+    }
+
+    #[namespace = "wallet"]
+    extern "Rust" {
+        type Network;
+        type BatchScanner;
+        type BatchResult;
+
+        fn network(
+            network: &str,
+            overwinter: i32,
+            sapling: i32,
+            blossom: i32,
+            heartwood: i32,
+            canopy: i32,
+            nu5: i32,
+        ) -> Result<Box<Network>>;
+
+        fn init_batch_scanner(
+            network: &Network,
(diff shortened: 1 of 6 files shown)


zcash_client_backend/src/scan.rs
```diff
@@ -0,0 +1,221 @@
+use crossbeam_channel as channel;
```
daira, line 1: This isn't a standard library import; shouldn't it go in a different section?

zcash_client_backend/src/welding_rig.rs
```diff
@@ -141,16 +144,65 @@ impl ScanningKey for SaplingIvk {
...
+        tree,
+        existing_witnesses,
+        None,
+    )
+}
+
+pub(crate) fn add_block_to_runner<P: consensus::Parameters + Send + 'static>(
+    params: &P,
+    block: CompactBlock,
+    batch_runner: &mut BatchRunner<SaplingDomain<P>, CompactOutputDescription>,
```
daira, line 169: How hard will it be to adapt this to cover Orchard batch scanning? I'm concerned that we could end up duplicating quite a lot of code and I want to minimise that as far as possible.

str4d: I've already started the process of adding Orchard to the zcashd version of this in https://github.com/zcash/zcash/pull/6105. Once that is done I'll apply the same changes to this code, assuming it still exists (and hasn't been replaced with async logic).

zcash_client_backend/src/scan.rs
nuttycom: We generally group imports from all "external" libraries together.

str4d: I do try and group `std / core / alloc` imports separately from external crate dependencies, but it's easy for an editor doing automatic imports to not pay attention to that sometimes.

zcash_client_backend/src/scan.rs
```diff
@@ -1,200 +1,19 @@
-use core::fmt;
+use crossbeam_channel as channel;
 use std::collections::HashMap;
-use std::io;
+use std::fmt;
 use std::mem;
 
-use crossbeam_channel as channel;
```
str4d: To @daira's point, this change in the second commit would not be present if the split between `core/std` imports and external crate imports had been preserved.

zcash_client_backend/src/scan.rs
```diff
@@ -1,200 +1,19 @@
...
     /// The incoming viewing key used to decrypt the note.
-    ivk: D::IncomingViewingKey,
+    pub(crate) ivk: D::IncomingViewingKey,
     /// The recipient of the note.
-    recipient: D::Recipient,
+    pub(crate) recipient: D::Recipient,
     /// The note!
-    note: D::Note,
-    /// The memo sent with the note.
-    memo: D::Memo,
```
str4d: Non-blocking: we should figure out how to generalise this logic over compact and non-compact outputs, so we can potentially share it with `zcashd` (assuming the logic doesn't change significantly when we move to async).

str4d (approved):
utACK 397c76ca8d3972855392b7a3a4ec975a6958f674

zcash_client_backend/src/scan.rs
nuttycom: Agreed; I made a quick attempt at it before we started pairing on this but it added enough complexity that think that should be a separate refactoring.

PR #613, opened 2022-08-24
zcash_client_backend: Generate Orchard component for USKs

str4d:
This ensures that the resulting UFVKs and UAs also contain Orchard components.

### DIFF
--- a/zcash_client_backend/src/keys.rs
+++ b/zcash_client_backend/src/keys.rs
@@ -69,6 +69,7 @@ fn to_transparent_child_index(j: DiversifierIndex) -> Option<u32> {
 #[derive(Debug)]
 #[doc(hidden)]
 pub enum DerivationError {
+    Orchard(orchard::zip32::Error),
     #[cfg(feature = "transparent-inputs")]
     Transparent(hdwallet::error::Error),
 }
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK

zcash_client_backend/src/keys.rs
```diff
@@ -283,6 +294,11 @@ impl UnifiedFullViewingKey {
         self.sapling.as_ref()
     }
 
+    /// Returns the Orchard full viewing key component of this unified key.
```
daira, line 297:
```suggestion
    /// Returns the Orchard full viewing key component of this unified key.
    /// (Orchard full viewing keys are always diversifiable.)
```

nuttycom: I don't understand what it means for a full viewing key to be diversifiable?

str4d: I think @daira was contrasting this to the docstring for the Sapling side, which says it returns a "diversifiable full viewing key". I think a better resolution would be to remove "diversifiable" from the Sapling text, as it's more of an implementation detail that users shouldn't need to care about.

Issue #614, opened 2022-08-26
Remove `extfvk` argument from `shield_transparent_funds`


str4d:
`shield_transparent_funds` shields funds from the transparent part of a given `account` to its shielded part. It currently takes an `extfvk` explicitly as an argument, and internally confirms it exists in the `accounts` table. Instead of that, we should just _use_ the viewing key from the `accounts` table, and derive an appropriate shielded recipient to send funds to.

PR #615, opened 2022-08-26
Shield funds to the internal Sapling key for a specified account.

nuttycom:
This updates `shield_transparent_funds` to look up the internal
(change) address for the account specified, and use that as the
destination for shielding transparent funds.

### DIFF
--- a/zcash_client_backend/src/data_api/error.rs
+++ b/zcash_client_backend/src/data_api/error.rs
@@ -24,6 +24,9 @@ pub enum ChainInvalid {
 
 #[derive(Debug)]
 pub enum Error<NoteId> {
+    /// No account with the given identifier was found in the wallet.
+    AccountNotFound(AccountId),
+
     /// The amount specified exceeds the allowed range.
     InvalidAmount,
 
@@ -34,12 +37,6 @@ pub enum Error<NoteId> {
 
     /// Chain validation detected an error in the block at the specified block height.
     InvalidChain(BlockHeight, ChainInvalid),
-
-    /// A provided extsk is not associated with the specified account.
-    AccountNotFound(AccountId),
-
-    /// No key of the given type was associated with the specified account.
-    KeyNotFound(AccountId, Typecode),
 
     /// A provided extsk is not associated with the specified account.
     InvalidExtSk(AccountId),
@@ -54,6 +51,9 @@ pub enum Error<NoteId> {
     /// The root of an output's witness tree in a previously stored transaction
     /// does not correspond to root of the current commitment tree.
     InvalidWitnessAnchor(NoteId, BlockHeight),
+
+    /// No key of the given type was associated with the specified account.
+    KeyNotFound(AccountId, Typecode),
 
     /// The wallet must first perform a scan of the blockchain before other
     /// operations can be performed.
@@ -86,6 +86,9 @@ impl ChainInvalid {
 impl<N: fmt::Display> fmt::Display for Error<N> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match &self {
+            Error::AccountNotFound(account) => {
+                write!(f, "Wallet does not contain account {}", u32::from(*account))
+            }
             Error::InvalidAmount => write!(
                 f,
                 "The value lies outside the valid range of Zcash amounts."
@@ -97,12 +100,6 @@ impl<N: fmt::Display> fmt::Display for Error<N> {
             ),
             Error::InvalidChain(upper_bound, cause) => {
                 write!(f, "Invalid chain (upper bound: {}): {:?}", u32::from(*upper_bound), cause)
-            }
-            Error::AccountNotFound(account) => {
-                write!(f, "Unrecognized account {}", u32::from(*account))
-            }
-            Error::KeyNotFound(account, typecode) => {
-                write!(f, "No {:?} key was available for account {}", typecode, u32::from(*account))
             }
             Error::InvalidExtSk(account) => {
                 write!(f, "Incorrect ExtendedSpendingKey for account {}", u32::from(*account))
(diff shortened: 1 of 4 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -448,11 +449,20 @@ where
...
+    // Obtain the UFVK for the specified account & use its internal change address
+    // as the destination for shielded funds.
+    let shielding_address = wallet_db
+        .get_unified_full_viewing_keys()
+        .and_then(|ufvks| {
+            ufvks
+                .get(&account)
+                .ok_or_else(|| E::from(Error::AccountNotFound(account)))
+                .and_then(|ufvk| {
+                    ufvk.sapling()
```
str4d, line 461: Add a TODO here (with an issue that is blocked by #614) to have this formally select the most preferred shielded receiver. Currently that is Sapling because we can't receive or spend Orchard funds, but that will change soon and we want to remember to update this.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -34,6 +35,12 @@ pub enum Error<NoteId> {
     /// Chain validation detected an error in the block at the specified block height.
     InvalidChain(BlockHeight, ChainInvalid),
 
+    /// A provided extsk is not associated with the specified account.
+    AccountNotFound(AccountId),
+
+    /// No key of the given type was associated with the specified account.
+    KeyNotFound(AccountId, Typecode),
+
```
str4d, line 43: These are maximally non-sorted 😝 I'd prefer they be sorted into the (currently mostly sorted) list of enum cases.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -34,6 +35,12 @@ pub enum Error<NoteId> {
     /// Chain validation detected an error in the block at the specified block height.
     InvalidChain(BlockHeight, ChainInvalid),
 
+    /// A provided extsk is not associated with the specified account.
```
str4d, line 38: This comment doesn't make sense (copy-paste error?) - the error kind is AFAICT for when a given account ID is not present in the wallet.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -91,6 +98,12 @@ impl<N: fmt::Display> fmt::Display for Error<N> {
             Error::InvalidChain(upper_bound, cause) => {
                 write!(f, "Invalid chain (upper bound: {}): {:?}", u32::from(*upper_bound), cause)
             }
+            Error::AccountNotFound(account) => {
+                write!(f, "Unrecognized account {}", u32::from(*account))
```
str4d, line 102:
```suggestion
                write!(f, "Wallet does not contain account {}", u32::from(*account))
```

zcash_primitives/src/sapling/keys.rs
```diff
@@ -292,6 +292,13 @@ impl DiversifiableFullViewingKey {
         zip32::sapling_default_address(&self.fvk, &self.dk)
     }
 
+    /// Returns the internal address corresponding to the smallest valid diversifier index,
+    /// along with that index.
```
str4d, line 296:
```suggestion
    /// Returns the internal address corresponding to the smallest valid diversifier index,
    /// along with that index.
    ///
    /// This address **MUST NOT** be encoded and exposed to end users. User interfaces
    /// should instead mark these notes as "change notes" or "internal wallet operations".
```

zcash_primitives/CHANGELOG.md
```diff
@@ -108,6 +108,9 @@ and this library adheres to Rust's notion of
   refactored to become a member function of a new `DiversifiableFullViewingKey`
   type, which represents the ability to derive IVKs, OVKs, and addresses, but
   not child viewing keys.
+- `zcash_primitives::sapling::keys::DiversifiableFullViewingKey::change_address`
+  has been added as a convenience method for obtaining the change address
+  at the default diversifier.
```
str4d, line 113:
```suggestion
  at the default diversifier. This address **MUST NOT** be encoded and exposed
  to users. User interfaces should instead mark these notes as "change notes" or
  "internal wallet operations".
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -418,11 +421,10 @@ where
 /// * `prover`: The TxProver to use in constructing the shielded transaction.
 /// * `sk`: The secp256k1 secret key that will be used to detect and spend transparent
 ///   UTXOs.
-/// * `extfvk`: The extended full viewing key that will be used to produce the
-///   Sapling address to which funds will be sent.
-/// * `account`: The ZIP32 account identifier associated with the the extended
-///   full viewing key. This procedure will return an error if this does not correctly
-///   correspond to `extfvk`.
+/// * `account`: The ZIP32 account identifier for the account to which funds will
+///   be shielded. Funds will be shielded to the internal (change) address associated with the
```
str4d, line 425: Hmm; is it intended that `shield_transparent_funds` can shield funds from an _arbitrary_ account (specifically whichever account `sk` corresponds to) to `account`? Or should it only be shielding funds within `account` itself?

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -418,11 +421,10 @@ where
...
+///   Sapling key corresponding to this account, or if a Sapling key is not available for this
+///   account this function will return an error.
```
str4d, line 427:
```suggestion
///   most preferred shielded receiver corresponding to this account, or if no shielded
///   receiver can be used for this account, this function will return an error.
```

str4d (changes requested):

zcash_client_backend/src/data_api/wallet.rs
nuttycom: If we want to enforce that constraint, we will just have to check that correspondence and return an error if it doesn't hold.  Opened https://github.com/zcash/librustzcash/issues/616 to track this.

zcash_client_backend/src/data_api/error.rs
nuttycom, line 38:
```suggestion
    /// No account with the given identifier was found in the wallet.
```

zcash_client_backend/src/data_api/wallet.rs
nuttycom, line 461:
```suggestion
                    // TODO: select the most preferred shielded receiver once we have the ability to
                    // spend Orchard funds.
                    ufvk.sapling()
```

nuttycom: Added #617

str4d (approved):
utACK ba1bb65a5f52477d21bd24d2962667812a2a7e94

PR #619, opened 2022-08-29
`zcash_client_sqlite`: Improve type safety for prepared statements

str4d:
This PR is overall a no-op, but collects all usages of `DataConnStmtCache` prepared statements into a private submodule `zcash_client_sqlite::prepared`. This forces the rest of the crate to only interact with the statements through type-safe APIs, and keeps the definition and execution of prepared statements close together for reviewability.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -8,7 +8,7 @@
 //! [`WalletWrite`]: zcash_client_backend::data_api::WalletWrite
 
 use ff::PrimeField;
-use rusqlite::{params, OptionalExtension, ToSql, NO_PARAMS};
+use rusqlite::{OptionalExtension, ToSql, NO_PARAMS};
 use std::collections::HashMap;
 use std::convert::TryFrom;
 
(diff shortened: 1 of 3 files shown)


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/4688985af27ae0ee8f939a30d30c7136cef193d9..04b1d505b283727b458f295fbbaaa931ed3ba65c) to fix clippy lints.

nuttycom (approved):
utACK

zcash_client_sqlite/src/wallet.rs
```diff
@@ -978,14 +919,7 @@ pub fn delete_utxos_above<'a, P: consensus::Parameters>(
...
 ) -> Result<usize, SqliteClientError> {
-    let sql_args: &[(&str, &dyn ToSql)] = &[
-        (":address", &taddr.encode(&stmts.wallet_db.params)),
-        (":above_height", &u32::from(height)),
-    ];
-
-    let rows = stmts.stmt_delete_utxos.execute_named(sql_args)?;
-
-    Ok(rows)
+    stmts.stmt_delete_utxos(taddr, height)
```
nuttycom, line 922: some of these are now so trivial as to almost not need to exist, but I guess that it's fine

str4d: Yep, these amount to a move in their cases. But they are all in the public API (deprecated), so I left them alone for now (we can decide what to do with them separately).

PR #620, opened 2022-08-29
Add views that join transactions & notes for Android SDK support.

nuttycom:
Android does not provide a good way to perform outer joins in
its type-safe query builder API, so we expose views build using
those joins instead.


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,102 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+                   blocks.time                  AS block_time,
+                   IFNULL(received_notes.is_change, false) as is_change
+            FROM   transactions
+                   LEFT JOIN received_notes
+                          ON transactions.id_tx = received_notes.tx
+                   LEFT JOIN sent_notes
+                          ON transactions.id_tx = sent_notes.tx
+                   LEFT JOIN blocks
+                          ON transactions.block = blocks.height
+            ORDER  BY ( minedheight IS NOT NULL ),
```
ccjernigan, line 491: Again I think we can eliminate ORDER BY.

ccjernigan (dismissed):

str4d (changes requested):

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,87 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+    "Add views over transaction & note data."
+);
+
+impl RusqliteMigration for WalletMigrationAddTxViews {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "CREATE VIEW IF NOT EXISTS v_tx_sent_notes AS
+            SELECT transactions.id_tx         AS id,
```
str4d, line 429: We should be consistent in how these are referenced:
```suggestion
            SELECT transactions.id_tx         AS id_tx,
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,87 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+                   transactions.block         AS mined_height,
+                   transactions.tx_index      AS transaction_index,
+                   transactions.txid          AS raw_transaction_id,
+                   transactions.expiry_height AS expiry_height,
+                   transactions.raw           AS raw,
+                   sent_notes.address         AS to_address,
+                   sent_notes.value           AS value,
+                   sent_notes.memo            AS memo,
+                   sent_notes.id_note         AS note_id,
```
str4d, line 438:
```suggestion
                   sent_notes.id_note         AS id_note,
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,87 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+impl RusqliteMigration for WalletMigrationAddTxViews {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "CREATE VIEW IF NOT EXISTS v_tx_sent_notes AS
+            SELECT transactions.id_tx         AS id,
+                   transactions.block         AS mined_height,
+                   transactions.tx_index      AS transaction_index,
+                   transactions.txid          AS raw_transaction_id,
```
str4d, line 432:
```suggestion
                   transactions.txid          AS txid,
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,87 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+            CREATE VIEW IF NOT EXISTS v_tx_received_notes AS
+            SELECT transactions.id_tx     AS id,
+                   transactions.block     AS mined_height,
+                   transactions.tx_index  AS transaction_index,
+                   transactions.txid      AS raw_transaction_id,
+                   received_notes.value   AS value,
+                   received_notes.memo    AS memo,
+                   received_notes.id_note AS note_id,
+                   blocks.time            AS block_time,
+                   IFNULL(received_notes.is_change, false) as is_change
```
str4d, line 454: Semantically, "is this note change" cannot be answered for a non-received note. So this really _should_ be null. It would be non-null if the view used an inner join instead (excluding transactions with no received notes).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,87 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+                   transactions.tx_index        AS transaction_index,
+                   transactions.txid            AS raw_transaction_id,
+                   transactions.expiry_height   AS expiry_height,
+                   transactions.raw             AS raw,
+                   sent_notes.address           AS to_address,
+                   IFNULL(sent_notes.value, received_notes.value) AS value,
+                   IFNULL(sent_notes.memo, received_notes.memo)   AS memo,
+                   IFNULL(sent_notes.id_note, received_notes.id_note) AS note_id,
+                   blocks.time                  AS block_time,
+                   IFNULL(received_notes.is_change, false) as is_change
```
str4d, line 472: AFAICT this is going to cause received notes to never show up in this view, when a sent note is also present.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,103 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+                   transactions.txid          AS txid,
+                   transactions.expiry_height AS expiry_height,
+                   transactions.raw           AS raw,
+                   SUM(sent_notes.value)      AS sent_total,
+                   COUNT(sent_notes.id_note)  AS sent_note_count,
+                   SUM(CASE WHEN sent_notes.memo IS NULL OR sent_notes.memo = '' THEN 0 ELSE 1 END) AS memo_count,
+                   blocks.time                AS block_time
+            FROM   transactions
+                   JOIN sent_notes
+                          ON transactions.id_tx = sent_notes.tx
```
ccjernigan, line 441: I see `transactions.id_tx` has a unique constraint, which means the column is implicitly indexed.  But there's no unique constraint on `sent_notes.tx` (just a NOT NULL).  Should we add an index to `sent_notes.tx` to improve the performance of this join?

ccjernigan (commented):
I think this all looks reasonable.

Have you done any performance checks against querying `v_transactions`?  Since it has 3 select statements, I was curious how it performs.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Are columns that have FOREIGN KEY constraints on them not indexed in sqlite? If that's the case, then yes we should probably add an index; most of the time it shouldn't matter too much since for ordinary wallets this won't be huge numbers of notes.

str4d: Would adding this index require modifying the view, or can the index be added independently? If the latter then we can leave it until we actually know what the performance of the view is like, but if we'd have to drop and recreate the view to use the index then I'd be more inclined to fold it into this migration.

ccjernigan: Foreign key wouldn't automatically cause sent_notes.tx to be indexed.  But my understanding is that a foreign key constraint in the sent_notes table would require transactions.id_tx to be indexed (e.g. with PRIMARY KEY, UNIQUE).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -1178,6 +1349,69 @@ mod tests {
...
+            .conn
+            .prepare("SELECT net_value, has_change FROM v_transactions")
+            .unwrap();
+        let mut rows = q.query(NO_PARAMS).unwrap();
+        let mut row_count = 0;
+        while let Some(row) = rows.next().unwrap() {
+            row_count += 1;
+            let net_value: i64 = row.get(0).unwrap();
+            let has_change: bool = row.get(1).unwrap();
+            assert_eq!(net_value, 7);
```
str4d, line 1429: Yep, this test shows the bug I raised above. This is ostensibly computing `5received + 7change - 2sent - 3sent`, but:
- if we treat all notes as separate, then we should have `0` instead of `7` because we want to ignore change (as it is wallet-internal, and the sum doesn't count the values of the wallet's own notes that were spent in the sent transactions).
- if the inconsistency is fixed, then we should also have `0` instead of `7` because we could not have any change note (as by definition change notes don't appear in `sent_notes`.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -434,6 +434,7 @@ impl RusqliteMigration for WalletMigrationAddTxViews {
                    transactions.raw           AS raw,
                    SUM(sent_notes.value)      AS sent_total,
                    COUNT(sent_notes.id_note)  AS sent_note_count,
+                   SUM(CASE WHEN sent_notes.memo IS NULL OR sent_notes.memo = '' THEN 0 ELSE 1 END) AS memo_count,
```
str4d, line 437: How do we store empty memos? Should `memo_count` also exclude the encoding of the empty memo?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -448,6 +449,7 @@ impl RusqliteMigration for WalletMigrationAddTxViews {
                    transactions.txid             AS txid,
                    SUM(received_notes.value)     AS received_total,
                    COUNT(received_notes.id_note) AS received_note_count,
+                   SUM(CASE WHEN received_notes.memo IS NULL OR received_notes.memo = '' THEN 0 ELSE 1 END) AS memo_count,
```
str4d, line 452: Ditto.

str4d (changes requested):

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -411,6 +411,98 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+                   transactions.expiry_height AS expiry_height,
+                   transactions.raw           AS raw,
+                   SUM(sent_notes.value)      AS sent_total,
+                   COUNT(sent_notes.id_note)  AS sent_note_count,
+                   blocks.time                AS block_time
+            FROM   transactions
+                   JOIN sent_notes
+                          ON transactions.id_tx = sent_notes.tx
+                   LEFT JOIN blocks
+                          ON transactions.block = blocks.height
```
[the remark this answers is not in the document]
nuttycom: My understanding is that left joins of this sort are associative; i.e. `(a left join b) left join c == a left join (b left join c)` so long as the `on` clauses do not create a dependency between rows in tables `b` and `c`.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -1178,6 +1349,69 @@ mod tests {
...
+        init_wallet_db(&mut db_data, None).unwrap();
+
+        db_data.conn.execute_batch(
+            "INSERT INTO accounts (account, ufvk) VALUES (0, '');
+            INSERT INTO blocks (height, hash, time, sapling_tree) VALUES (0, 0, 0, '');
+            INSERT INTO transactions (block, id_tx, txid) VALUES (0, 0, '');
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, address, value) VALUES (0, 2, 0, 0, '', 2);
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, address, value) VALUES (0, 2, 1, 0, '', 3);
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change) VALUES (0, 0, 0, '', 5, '', 'a', false);
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change) VALUES (0, 1, 0, '', 7, '', 'b', true);",
```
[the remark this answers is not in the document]
nuttycom: How does the `output_pool` field factor in here? if they're in different pools, we could have overlapping output indices, right? However, the `received_notes` table doesn't currently track output pool information at all, so we actually can't tell that these are the same notes?

zcash_client_sqlite/src/wallet/init.rs
nuttycom: This reminds me of a change that I'm making in the zcashd wallet, which I also want to make in the light wallet, which is that change be explicit, not implicit. The fact that change notes currently never show up as sent, only as received, means that we aren't really doing proper double-entry accounting in the wallet; we're eliding the change spent and then therefore have to skip counting the change received. Also, we're probably not accounting correctly for fees at all.

zcash_client_sqlite/src/wallet/init.rs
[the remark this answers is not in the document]
str4d: Aha, currently the `received_notes` table's `output_pool` is inherently Sapling, because we don't have Orchard scanning support yet, and that table isn't used for transparent. And you've hard-coded typecode 2 into the `sent_notes` data, which is for Sapling. Hence the `output_index` values do overlap in this test data.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -278,7 +278,7 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
     pub fn build(
         mut self,
         prover: &impl TxProver,
-    ) -> Result<(Transaction, SaplingMetadata), Error> {
+    ) -> Result<(Transaction, SaplingMetadata, Amount), Error> {
```
str4d, line 281: This is a breaking API change, and it is also unnecessary because we know that the builder hard-codes `DEFAULT_FEE` which is part of the public crate API.

Revert this change; we should design the API changes for potentially-variable fee setting as a cohesive feature, separately from this PR.

str4d (changes requested):
Still reviewing, but NACK until the changes to the transaction builder API are removed from the PR.

zcash_primitives/src/transaction/builder.rs
nuttycom: The "we know the builder hard-codes DEFAULT_FEE" is another one of those "implicit knowledge, you have to read the code or the docs to figure it out" things that I don't like having in our codebase. I'd like to keep this change.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/3a3827b559178ed8e9664b06c5f8425be0186d4c..2497826a32e45118b6cf894f43e14fa4828d5e2f) to address review comments

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/3a3827b559178ed8e9664b06c5f8425be0186d4c..6e228d35ecb33b605a3aece439d473c3c0424dd6) to address review comments and rebase to `main`.

nuttycom:
The most recent [force-push](https://github.com/zcash/librustzcash/pull/620) is a zero-change history cleanup.

zcash_primitives/src/transaction/builder.rs
nuttycom: Fixed by https://github.com/zcash/librustzcash/compare/3a3827b559178ed8e9664b06c5f8425be0186d4c..6e228d35ecb33b605a3aece439d473c3c0424dd6

str4d (changes requested):
Flushing my review comments.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -154,7 +158,16 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
...
     /// The fee will be set to the default fee (0.0001 ZEC).
     pub fn new_with_rng(params: P, target_height: BlockHeight, rng: R) -> Builder<'a, P, R> {
-        Self::new_internal(params, target_height, rng)
+        Self::new_internal(params, rng, target_height, DEFAULT_FEE)
+    }
+
+    pub fn new_with_rng_and_fee(
+        params: P,
+        rng: R,
+        target_height: BlockHeight,
```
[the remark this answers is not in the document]
nuttycom: I think we made a mistake with the previous order of constructor arguments: I generally think of function arguments (including constructor arguments) from a perspective of how those arguments work under partial application. In imagining how this constructor could be used under partial application, the parameters and RNG are the arguments that are most likely to be applied to yield a function that takes the remainder of the arguments. I don't feel strongly about this, but I think it makes sense to group arguments that affect the contents of the constructed transaction and arguments that relate to *how to construct* the transaction separately.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -163,13 +176,18 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
     ///
     /// WARNING: THIS MUST REMAIN PRIVATE AS IT ALLOWS CONSTRUCTION
     /// OF BUILDERS WITH NON-CryptoRng RNGs
-    fn new_internal(params: P, target_height: BlockHeight, rng: R) -> Builder<'a, P, R> {
+    fn new_internal(
+        params: P,
+        rng: R,
+        target_height: BlockHeight,
```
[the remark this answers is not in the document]
nuttycom: Same as here: https://github.com/zcash/librustzcash/pull/620/files#r965089751 - I think that we made a mistake in how we previously grouped constructor arguments.

zcash_primitives/src/transaction/builder.rs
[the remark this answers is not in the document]
str4d: This is another reason why I feel this change should be made in a separate PR, so we can discuss these kinds of changes in the broader context of the transaction builder.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/11dc9fd9ee35d0eb6be6e0279f11d637d469e420..2d2ba112b241d25dafb140600534a5ad5c395316) to rebase to `main` and address comments from code review.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -36,6 +39,9 @@ pub enum WalletMigrationError {
 
     /// Wrapper for rusqlite errors.
     DbError(rusqlite::Error),
+
+    /// Wrapper for amount balance violations
+    BalanceError(BalanceError),
```
str4d, line 44: When was `WalletMigrationError` added? Should this extra case be added to the changelog?

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -0,0 +1,24 @@
...
+) -> Vec<Box<dyn RusqliteMigration<Error = WalletMigrationError>>> {
+    vec![
+        Box::new(WalletMigration0 {}),
+        Box::new(WalletMigration1 {}),
+        Box::new(WalletMigration2 {
+            params: params.clone(),
+            seed,
+        }),
+        Box::new(add_transaction_views::Migration {
+            params: params.clone(),
```
str4d, line 21: This feels like it will trigger a clippy lint for "unnecessary clone". I think this should also work:
```suggestion
            params,
```

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+                            WalletMigrationError::CorruptedData(format!(
+                                "Parsing failed for transaction {:?}: {:?}",
+                                id_tx, e
+                            ))
+                        })?;
+
+                let fee_paid = tx.fee_paid(|op| {
+                    stmt_find_utxo_value
+                        .query_row(&[op.hash().to_sql()?, op.n().to_sql()?], |row| {
+                            row.get(0).map(|i| Amount::from_i64(i).unwrap())
```
str4d, line 78: This unwrap _should_ be fine (it won't trigger for any data that's been inserted by `zcash_client_sqlite`), but there's technically a crash chance here. Open an issue to fix this if you don't want to do so in this PR.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+        Ok(())
+    }
+
+    fn down(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "DROP VIEW v_tx_sent_notes;
+            DROP VIEW v_tx_received_notes;
+            DROP VIEW v_tx_notes;",
+        )?;
+        Ok(())
```
str4d, line 192: This needs to also remove the `fee` column in order to be a correct down-migration. That is quite involved however. There are several possible alternatives:
- Remove the column via "create table without `fee` column, copy all data across, delete old table, rename new table".
- Alter the up-migration to correctly handle the case where the `fee` column exists:
  - If the column doesn't exist, add it and fill it.
  - If the column does exist, check its contents matches what we would have filled in.
- Make the entire down-migration panic (the "simple" solution).

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change, memo)
+            VALUES (0, 0, 0, '', 5, '', 'a', false, X'62');
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change, memo)
+            VALUES (0, 1, 0, '', 7, '', 'b', true, X'63');",
```
str4d, line 239: This test data is still inconsistent:
- `sent_notes` contains notes with `(tx, output_pool, output_index)` of `(0, 2, 0)`, `(0, 2, 1)`, `(0, 2, 2)`.
- `received_notes ` contains notes with `(tx, output_pool, output_index)` of `(0, 2, 0)`, `(0, 2, 1)` (because the entire `received_notes` table is implicitly Sapling).
- The two notes that appear in both `sent_notes` and `received_notes` have different values and memos.

https://github.com/zcash/librustzcash/pull/620#discussion_r961133991 doesn't apply, because if it _did_, we would have no way to ever use or migrate the `received_notes` table. We _have_ to treat all notes in the (current) `received_notes` table as Sapling notes, because that's all that has ever been added to that table, and that is how we interpret the notes in that table when spending notes. When we add Orchard support, we'll need to migrate the table to have an `output_pool` column like `sent_notes` does (or do some other more involved migration to deduplicate notes if we decide that's a more useful data model, which in any case will require treating the existing notes as Sapling notes).

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+                   transactions.tx_index      AS tx_index,
+                   transactions.txid          AS txid,
+                   transactions.expiry_height AS expiry_height,
+                   transactions.raw           AS raw,
+                   SUM(sent_notes.value)      AS sent_total,
+                   COUNT(sent_notes.id_note)  AS sent_note_count,
+                   SUM(
+                       CASE
+                           WHEN sent_notes.memo IS NULL THEN 0
+                           WHEN SUBSTR(sent_notes.memo, 0, 2) = X'F6' THEN 0
```
str4d, line 100: This is not technically correct. Per [ZIP 302](https://zips.z.cash/zip-0302#specification):
> - the byte `0xF6` followed by 511 `0x00` bytes, indicating "no memo";
>
> ...
>
> If the memo matches any of these patterns, then this memo is from the future, because these ranges are reserved for future updates to this specification:
> - The first byte has a value of `0xF6`, and the remaining 511 bytes are not all `0x00`.

The entire memo needs to be checked to verify it is an empty memo. (Just being able to check the first byte was [my rationale for keeping the original definition of empty memos from the initial draft](https://github.com/zcash/zips/pull/105#discussion_r597257764), but it was decided to go with the protocol spec definition instead for various reasons.)

That being said, a memo of `null` has previously meant "we don't know what the memo is yet", so a rationale we could use here is that we are just not counting memos that we don't know how to represent to the user. That rationale would technically imply excluding all "future" memos as well, and that feels weird because it's likely that future memo types are meaningful. But if what the SDK is going to use this count for is "how many memos are we going to show the user" then there's no need to count future memo types, in which case we can leave this check as-is (and we should add additional exclusions). Adding exclusions into the view does mean that adding a new future memo type will require migrating the view.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+            SELECT transactions.id_tx            AS id_tx,
+                   transactions.block            AS mined_height,
+                   transactions.tx_index         AS tx_index,
+                   transactions.txid             AS txid,
+                   SUM(received_notes.value)     AS received_total,
+                   COUNT(received_notes.id_note) AS received_note_count,
+                   SUM(
+                       CASE
+                           WHEN received_notes.memo IS NULL THEN 0
+                           WHEN SUBSTR(received_notes.memo, 0, 2) = X'F6' THEN 0
```
str4d, line 121: Ditto.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+                       transactions.raw              AS raw,
+                       0                             AS fee,
+                       CASE
+                            WHEN received_notes.is_change THEN 0
+                            ELSE value
+                       END AS value,
+                       received_notes.is_change      AS is_change,
+                       CASE
+                           WHEN received_notes.memo IS NULL THEN 0
+                           WHEN SUBSTR(received_notes.memo, 0, 2) = X'F6' THEN 0
```
str4d, line 157: Ditto.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+                       transactions.tx_index         AS tx_index,
+                       transactions.txid             AS txid,
+                       transactions.expiry_height    AS expiry_height,
+                       transactions.raw              AS raw,
+                       transactions.fee              AS fee,
+                       -sent_notes.value             AS value,
+                       false                         AS is_change,
+                       CASE
+                           WHEN sent_notes.memo IS NULL THEN 0
+                           WHEN SUBSTR(sent_notes.memo, 0, 2) = X'F6' THEN 0
```
str4d, line 174: Ditto.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -0,0 +1,356 @@
...
+                          ON transactions.block = blocks.height
+            GROUP BY received_notes.tx;
+            CREATE VIEW v_transactions AS
+            SELECT id_tx,
+                   mined_height,
+                   tx_index,
+                   txid,
+                   expiry_height,
+                   raw,
+                   SUM(value) + MAX(fee) AS net_value,
```
str4d, line 139: Is the `MAX(fee)` here necessary because this is collating multiple rows with distinct `value` column entries? And in the previous fields, because all collated rows have the same entries you can just select the column and get a single copy of the row?

str4d (changes requested):

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
str4d: And now that I look at this again, how does this work? `memo` is a `BLOB`, and `SUBSTR` is [defined as](https://www.sqlite.org/lang_corefunc.html#substr) (emphasis mine):

> The substr(X,Y,Z) function returns a substring of input string X that begins with the Y-th character and which is Z characters long. If Z is omitted then substr(X,Y) returns all characters through the end of the string X beginning with the Y-th. **The left-most character of X is number 1.** If Y is negative then the first character of the substring is found by counting from the right rather than the left. If Z is negative then the abs(Z) characters preceding the Y-th character are returned. If X is a string then characters indices refer to actual UTF-8 characters. If X is a BLOB then the indices refer to bytes.

Doesn't this mean that `SUBSTR(sent_notes.memo, 0, 2)` is going to attempt to return 2 bytes starting from index 0, which is before the start of the `BLOB`? Should this not be `SUBSTR(sent_notes.memo, 1, 1)`?

And additionally, can we not check against the entire memo with `SUBSTR(sent_notes.memo, 1, 512) = X'F6000000....'`?

str4d: https://www.sqlite.org/lang_corefunc.html#quote suggests that what we maybe need here is `QUOTE(sent_notes.memo) = X'F60000....'`

nuttycom: I tried this with `QUOTE` and it still will not match the empty memo literal.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
nuttycom: Since we're grouping by the transaction's row ID, we'll get only a single row per transaction (a single row per grouped identifier) so the query cannot even compile without using an aggregate function for each column retrieved from a joined table. `MAX` is correct here because we explicitly set the fee to 0 on the received "side" of the union; when we have two rows with the same transaction ID, we will therefore get the fee amount from the sent-note associated row.

zcash_client_sqlite/src/prepared.rs
```diff
@@ -478,7 +482,8 @@ impl<'a, P> DataConnStmtCache<'a, P> {
             u32::from(account),
             to_str,
             ivalue,
-            memo.map(|m| m.as_slice()),
+            memo.filter(|m| *m != &MemoBytes::empty())
+                .map(|m| m.as_slice()),
```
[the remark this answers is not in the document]
nuttycom: I think it would be a really bad thing if downstream consumers used that as a sentinel for download state, which leads me to this resolution: https://github.com/zcash/librustzcash/pull/620/commits/3ca50520d440f6111277a2f6d8b0ea2088afbf18 - uniformly avoid storing the empty memo; this also should avoid a potentially-costly memcmp in the new views. This might be a premature optimization, but I would much prefer that callers not be able to use the nullity of the memo field for determining download state; if we want that information to be available, it should be via an explicit status flag.

zcash_client_sqlite/src/wallet/init/migrations.rs
nuttycom: Since we get `params` by reference, we need to clone here. It might be that we should actually make `params` a `Copy` type, but that's not in scope for this PR.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Added.

nuttycom:
[changes in response to @str4d's comments](https://github.com/zcash/librustzcash/compare/11dc9fd9ee35d0eb6be6e0279f11d637d469e420..7d1e976d0ad998a8bbc5a80ac82d6b077b81a66e) (plus rebase)

zcash_client_sqlite/src/prepared.rs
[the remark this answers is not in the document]
str4d: My concern was that downstream consumers might _already_ be using the nullity of the memo field for this purpose. But I believe it's more likely that they are using nullity of the `raw` field of the corresponding entry in the `transactions` table for this purpose, so this change _should be fine.

str4d (approved):
utACK 7d1e976d0ad998a8bbc5a80ac82d6b077b81a66e

PR #621, opened 2022-08-29
zcash_client_backend: Add `WalletWrite::remove_unmined_tx` method

str4d:
This is to replace the database mutations in the Android SDK. It is placed behind an `unstable` feature flag until we are satisfied that it is suitable as a general-purpose API (or replace it).

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -558,6 +558,59 @@ impl<'a, P: consensus::Parameters> WalletWrite for DataConnStmtCache<'a, P> {
         })
     }
 
+    #[cfg(feature = "unstable")]
+    fn remove_tx(&mut self, txid: &TxId) -> Result<(), Self::Error> {
+        self.transactionally(|up| {
+            // Find the transaction to be deleted.
+            let tx_ref = up.stmt_select_tx_ref(txid)?;
+
+            // Delete witnesses for received notes that will be deleted.
+            up.wallet_db.conn.execute_named(
+                "DELETE FROM sapling_witnesses WHERE note IN (
+                    SELECT rn.id_note
+                    FROM received_notes rn
+                    WHERE rn.tx = :tx
+                )",
+                &[(":tx", &tx_ref)],
+            )?;
+
+            // Delete received notes that were outputs of the given transaction.
+            up.wallet_db.conn.execute_named(
+                "DELETE FROM received_notes WHERE tx = :tx",
+                &[(":tx", &tx_ref)],
+            )?;
+
+            // Unmine received notes that were inputs to the given transaction.
+            up.wallet_db.conn.execute_named(
+                "UPDATE received_notes SET spent = null WHERE spent = :tx",
+                &[(":tx", &tx_ref)],
+            )?;
+
+            // We don't normally want to delete sent notes, as this can contain data that
+            // is not recoverable from the chain. However, given that the caller wants to
+            // delete the transaction in which these notes were sent, we cannot keep these
+            // database entries around.
+            up.wallet_db
+                .conn
+                .execute_named("DELETE FROM sent_notes WHERE tx = :tx", &[(":tx", &tx_ref)])?;
+
+            // Delete the UTXOs because these are effectively cached and we don't have a
+            // good way of knowing whether they're spent.
+            up.wallet_db.conn.execute_named(
+                "DELETE FROM utxos WHERE spent_in_tx = :tx",
+                &[(":tx", &tx_ref)],
+            )?;
+
+            // Now that it isn't depended on, delete the transaction.
+            up.wallet_db.conn.execute_named(
+                "DELETE FROM transactions WHERE id_tx = :tx",
+                &[(":tx", &tx_ref)],
+            )?;
+
+            Ok(())
+        })
+    }
+
     fn rewind_to_height(&mut self, block_height: BlockHeight) -> Result<(), Self::Error> {
         wallet::rewind_to_height(self.wallet_db, block_height)
     }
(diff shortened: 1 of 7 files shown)


str4d:
`WalletWrite::remove_tx()` is specifically for replacing [`DerivedDataDb.removeInvalidOutboundTransaction()`](https://github.com/zcash/zcash-android-wallet-sdk/blob/345af7aaccecf313aaa64b050dc0764545eaf84a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/DerivedDataDb.kt#L460-L486).

zcash_client_sqlite/src/lib.rs
```diff
@@ -558,6 +558,59 @@ impl<'a, P: consensus::Parameters> WalletWrite for DataConnStmtCache<'a, P> {
...
+                .execute_named("DELETE FROM sent_notes WHERE tx = :tx", &[(":tx", &tx_ref)])?;
+
+            // Delete the UTXOs because these are effectively cached and we don't have a
+            // good way of knowing whether they're spent.
+            up.wallet_db.conn.execute_named(
+                "DELETE FROM utxos WHERE spent_in_tx = :tx",
+                &[(":tx", &tx_ref)],
+            )?;
+
+            // Now that it isn't depended on, delete the transaction.
```
str4d, line 604: It occurs to me that this is not necessarily true: it may be that the caller is deleting a transaction with notes that have already been spent in some downstream transaction. I checked locally and the test works fine if I delete the first transaction (that is, the balance is detected only as the change), because there is only a chain of two transactions (so no spent note becomes falsely unspent). I expect that the actual bug could be reproduced (though it would take some work as the currentl test harness can't create three-tx chains).

This was a bug in the Android SDK logic, but not one that I think is reachable, because the deletion logic is only triggered for cancelled transactions (that have only just been created) and expired transactions, neither of which by definition can have outputs that get spent. But we should either fix this before publishing a `zcash_client_sqlite` release, or remove the API.

zcash_client_sqlite/src/lib.rs
```diff
@@ -840,4 +894,92 @@ mod tests {
...
+        // Account balance should only reflect the first received note.
+        assert_eq!(get_balance(&db_data, AccountId::from(0)).unwrap(), value);
+
+        // Scanning the cache again won't change the balance, as the wallet believes it
+        // has already scanned all the blocks in the cache.
+        scan_cached_blocks(&network(), &db_cache, &mut db_write, None).unwrap();
+        assert_eq!(get_balance(&db_data, AccountId::from(0)).unwrap(), value);
+
+        // Rewind the wallet by one block, and scan the cache again.
+        rewind_to_height(&db_data, sapling_activation_height()).unwrap();
```
nuttycom, line 976: This inconsistency in database state bothers me a little bit. It really seems like we shouldn't be exposing the ability to remove a transaction directly, but should instead be manipulating the chain state to get rid of the transaction so that the two don't get out of sync.

zcash_client_sqlite/src/lib.rs
str4d: Opened #622.

zcash_client_sqlite/src/lib.rs
str4d: The Android SDK code that wants this, is wanting to manipulate _non-chain_ state: either just-created transactions that haven't been mined, or transactions that were created and then expired before getting mined. So there's no chain state to keep in sync from that perspective. However, see my other comment where I identify a bug in the upstream logic that this PR preserves around mined transactions.

str4d: For clarity, I'm leveraging in this test the fact that it is equally fine to remove a transaction that is unmined, and has _just_ been mined, in order to make defining the test easy. The bug is that the Android code (and this code by extension) allows long-mined transactions to also be removed.

zcash_client_sqlite/src/lib.rs
```diff
@@ -558,6 +558,59 @@ impl<'a, P: consensus::Parameters> WalletWrite for DataConnStmtCache<'a, P> {
...
+
+            // Delete witnesses for received notes that will be deleted.
+            up.wallet_db.conn.execute_named(
+                "DELETE FROM sapling_witnesses WHERE note IN (
+                    SELECT rn.id_note
+                    FROM received_notes rn
+                    WHERE rn.tx = :tx
+                )",
+                &[(":tx", &tx_ref)],
+            )?;
```
str4d, line 575: This code was not present in the upstream Android SDK logic, but was necessary to enable deleting just-mined transactions (so I could use the test harness). Note however that omitting this logic would not have prevented the bug from being triggered, as old witnesses get periodically pruned, so old-enough transactions would not run into any foreign key issue.

nuttycom: Another idea: would it make sense to join to the `blocks` table and only delete transactions where they aren't associated with any block?

str4d: If we want this to specifically be `remove_unmined_tx` then yes that should work. That's not how the Android SDK logic was implemented, but based on how it was _used_ I think that should be sufficient. @ccjernigan?

str4d: I've now implemented this change.

zcash_client_sqlite/src/lib.rs
str4d: We now only allow unmined transactions to be removed.

zcash_client_sqlite/src/lib.rs
str4d: I've now resolved this issue; #622 is now about whether we want to keep the API itself.

nuttycom (approved):
utACK

PR #623, opened 2022-08-31
`zcash_client_*`: Various changes to how addresses are handled

str4d:


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -289,12 +289,34 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
+                let ufvk_str: String = ufvk.encode(&self.params);
+                let address_str: String = ufvk.default_address().0.encode(&self.params);
+
+                // This migration, and the wallet behaviour before it, stored the default
+                // transparent address in the `accounts` table. This does not necessarily
+                // match the transparent receiver in the default Unified Address. Starting
+                // from `AddressesTableMigration` below, we no longer store transparent
+                // addresses directly, but instead extract them from the Unified Address
+                // (or from the UFVK if the UA was derived without a transparent receiver,
+                // which is not the case for UAs generated by this crate).
```
nuttycom, line 301: I don't understand this comment fully; it corresponds to the code below but I don't think this is doing the right thing. We still need to retain the transparent address that was previously generated, so we shouldn't be overwriting it with the taddr that's the default for the UFVK.

nuttycom (changes requested):
This does not seem to preserve the t-address at "diversifier index 0", which we will continue to need to use for detecting received funds in wallets that had auto-shielding enabled and who had distributed their taddrs to counterparties who might send funds to them in the future. So, if we're not going to store those taddrs directly, we need to programmatically derive that special taddr in code that retrieves transparent tx associated with this wallet.

zcash_client_sqlite/src/wallet/init.rs
str4d: Per our earlier discussion, what I will do is ensure that the `transparent_address` column is indeed storing the address we expect it to be, and then continue to delete the column. We will instead have a new API that provides a list of all the transparent receivers that should be scanned for funds; it will be composed of all transparent receivers for all UAs in the `addresses` table, plus the default transparent address under each account (because the autoshielding code was generating that transparent address, and we cannot predict whether it was generated for accounts other than account 0 that the SDKs were using, so we have to scan it in perpetuity).

zcash_client_sqlite/src/lib.rs
```diff
@@ -150,9 +150,33 @@ impl<P: consensus::Parameters> WalletRead for WalletDb<P> {
         wallet::get_unified_full_viewing_keys(self)
     }
 
-    fn get_address(&self, account: AccountId) -> Result<Option<PaymentAddress>, Self::Error> {
-        #[allow(deprecated)]
-        wallet::get_address(self, account)
+    fn get_address(&self, account: AccountId) -> Result<Option<UnifiedAddress>, Self::Error> {
```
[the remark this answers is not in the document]
str4d: That method is not crate-private, that's the problem. Its implementation now diverges from `WalletRead::get_address`.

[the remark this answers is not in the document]
str4d: Given that the vast majority of every function in `zcash_client_sqlite::wallet` is both public and deprecated, it seemed to me like the split was purely historic. If you want this impl to also be there, then it will be in a separate crate-private method.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/0d3d1523796045af113cd0d2131deaaa2bf1c58f..faef150d839af3aa275052ed2d17c355d6950178) to move the new migration into a sub-module, to make merging concurrent migrations simpler.

nuttycom:
> [Force-pushed](https://github.com/zcash/librustzcash/compare/0d3d1523796045af113cd0d2131deaaa2bf1c58f..faef150d839af3aa275052ed2d17c355d6950178) to move the new migration into a sub-module, to make merging concurrent migrations simpler.

We should figure out a way to name the sub-modules such that the next time we need a migration on the address tables, we don't get confusion.

zcash_client_sqlite/src/lib.rs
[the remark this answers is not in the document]
nuttycom: So, the deprecation was just to make it possible to eventually remove it from the public API. But the fact that the methods were originally public is, in my mind, a *feature* because I think that one should also expose the internals used to implement higher-level functionality, in general.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/faef150d839af3aa275052ed2d17c355d6950178..9799703b5354232187b93b4741f44edefc8c12d8) to address most of @nuttycom's comments. The remaining comment is the new API to get the transparent addresses, which I suspect should take an account as a parameter like most other APIs; I'll add that in a separate commit.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -289,12 +291,34 @@ impl<P: consensus::Parameters> RusqliteMigration for WalletMigration2<P> {
...
```
daira, line 303:
```suggestion
                // addresses directly, but instead extract them from the UFVK.
```
As far as I can see the code does always extract them from the UFVK, regardless of whether they exist in the UA.

daira (approved):
utACK. There are no tests; is that okay?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -173,6 +174,36 @@ pub fn get_address<P: consensus::Parameters>(
...
 
+pub(crate) fn get_address_ua<P: consensus::Parameters>(
+    wdb: &WalletDb<P>,
+    account: AccountId,
+) -> Result<Option<UnifiedAddress>, SqliteClientError> {
+    // This returns the first diversified address, which will be the default one.
+    let addr: Option<String> = wdb
+        .conn
+        .query_row(
+            "SELECT address FROM addresses WHERE account = :account",
```
nuttycom, line 185:
```suggestion
            "SELECT address FROM addresses WHERE account = :account ORDER BY diversifier_index LIMIT 1",
```
Please make the "first diversified address" condition explicit.

zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
```diff
@@ -0,0 +1,176 @@
...
+use super::super::{add_account_internal, WalletMigrationError};
+
+#[cfg(feature = "transparent-inputs")]
+use zcash_primitives::legacy::keys::IncomingViewingKey;
+
+/// The migration that removed the address columns from the `accounts` table, and created
+/// the `accounts` table.
+///
+/// d956978c-9c87-4d6e-815d-fb8f088d094c
+pub(super) const ADDRESSES_TABLE_MIGRATION: Uuid = Uuid::from_fields(
```
nuttycom, line 19: This is a nice approach, I'll have to handle the ID for the views migration the same way.

zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
```diff
@@ -0,0 +1,176 @@
...
+                            transparent_address,
+                        )));
+                    }
+                }
+
+                // If the transparent_address column is not empty, and we can't check its
+                // value, return an error.
+                #[cfg(not(feature = "transparent-inputs"))]
+                {
+                    let _ = decoded_transparent_address;
```
nuttycom, line 147: This line is just here to avoid a warning that `decoded_transparent_address` is unused?

nuttycom (changes requested):
One minor change requested; otherwise still awaiting the transparent addresses accessor.

zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
str4d: Correct.

zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
str4d: Yep; I figured this way we could use code imports for DAG edges.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/9799703b5354232187b93b4741f44edefc8c12d8..e37693c2a8dcc60df0618c991ccabb45f221ebd0).

zcash_client_sqlite/src/wallet.rs
str4d: In a pairing we established that this didn't work, because diversifier indices are stored as `BLOB` types, and ordering there would not order numerically. As it is, this method will always return a single address because we don't expose a way to derive more, and when we do this method will need to take a diversifier index.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/e37693c2a8dcc60df0618c991ccabb45f221ebd0..4eda5f4351853608a9c5e4664cbcaaac6d84338b) with a small fix, and to add `WalletReadTransparent::get_transparent_receivers`.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/4eda5f4351853608a9c5e4664cbcaaac6d84338b..7112ed9da5d8168e670fe34b7fa11568e6f6717e) to fix clippy lints.

zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
```diff
@@ -0,0 +1,176 @@
...
+                // so we can confidently delete the column knowing we can regenerate the
+                // values from the stored UFVKs.
+
+                // We can only check if it is the expected transparent address if the
+                // transparent-inputs feature flag is enabled.
+                #[cfg(feature = "transparent-inputs")]
+                {
+                    let expected_address = ufvk
+                        .transparent()
+                        .and_then(|k| k.derive_external_ivk().ok().map(|k| k.default_address().0));
```
nuttycom, line 134: This is only a valid check because we're assuming that no wallets exist that were initialized after the previous migration (in which case the taddr would be the taddr from the default UA, rather than the one derived at index 0.) I'm not sure that this corruption check is actually useful, but I guess it's okay.

zcash_client_sqlite/src/lib.rs
```diff
@@ -900,6 +925,36 @@ mod tests {
...
+
+        let data_file = NamedTempFile::new().unwrap();
+        let mut db_data = WalletDb::for_path(data_file.path(), network()).unwrap();
+        init_wallet_db(&mut db_data, Some(Secret::new(vec![]))).unwrap();
+
+        // Add an account to the wallet.
+        let (ufvk, taddr) = init_test_accounts_table_ufvk(&db_data);
+        let taddr = taddr.unwrap();
+
+        let receivers = db_data.get_transparent_receivers(0.into()).unwrap();
```
nuttycom, line 949: It would be nice to run this test with two different seeds; one that triggers the case where `receivers` contains two different addresses, and one that triggers the case where there is only one.

nuttycom (approved):
utACK with nonblocking comments.

PR #624, opened 2022-08-31
Add conversions from u32 and u64 to DiversifierIndex

nuttycom:
This implementation is copied from https://github.com/zcash/zcash/blob/master/src/zcash/address/zip32.h#L84-L93

### DIFF
--- a/zcash_primitives/src/zip32.rs
+++ b/zcash_primitives/src/zip32.rs
@@ -134,6 +134,20 @@ impl Default for DiversifierIndex {
     }
 }
 
+impl From<u32> for DiversifierIndex {
+    fn from(i: u32) -> Self {
+        u64::from(i).into()
+    }
+}
+
+impl From<u64> for DiversifierIndex {
+    fn from(i: u64) -> Self {
+        let mut result = DiversifierIndex([0; 11]);
+        (&mut result.0[..8]).copy_from_slice(&i.to_le_bytes());
+        result
+    }
+}
+
 impl DiversifierIndex {
     pub fn new() -> Self {
         DiversifierIndex([0; 11])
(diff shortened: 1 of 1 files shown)


zcash_primitives/src/zip32.rs
```diff
@@ -134,6 +134,27 @@ impl Default for DiversifierIndex {
...
 }
 
+impl From<u32> for DiversifierIndex {
+    fn from(i: u32) -> Self {
+        u64::from(i).into()
+    }
+}
+
+impl From<u64> for DiversifierIndex {
+    fn from(i: u64) -> Self {
```
pacu, line 144: shouldn't these be `pub fn` ?

pacu (changes requested):

zcash_primitives/src/zip32.rs
nuttycom: No; trait implementation methods are public by default.

pacu (approved):

zcash_primitives/src/zip32.rs
```diff
@@ -134,6 +134,20 @@ impl Default for DiversifierIndex {
...
+        let mut result = DiversifierIndex([0; 11]);
+        (&mut result.0).copy_from_slice(&i.to_le_bytes());
```
str4d, line 146: This will panic, because [`&[T]::copy_from_slice` requires that `src` and `self` are the same length](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.copy_from_slice), but here `result.0` has length 11, and `i.to_le_bytes()` has length 8. This should be:
```suggestion
        (&mut result.0[..8]).copy_from_slice(&i.to_le_bytes());
```

str4d (changes requested):
Add a test that covers these conversions (which would have caught the bug).

str4d (approved):
utACK

daira (approved):
utACK

PR #625, opened 2022-09-01
Add a binary encoding format for unified spending keys.

nuttycom:


components/zcash_address/src/kind/unified.rs
```diff
@@ -88,6 +88,18 @@ impl From<Typecode> for u32 {
     }
 }
 
+impl From<Typecode> for usize {
+    fn from(t: Typecode) -> Self {
+        match t {
```
daira, line 93: This could be implemented as `u32::from(t).try_into().unwrap()`. Non-blocking.

zcash_client_backend/src/keys.rs
```diff
@@ -74,12 +86,56 @@ pub enum DerivationError {
     Transparent(hdwallet::error::Error),
 }
 
+/// A version identifier for the encoding of unified spending keys.
+///
+/// Each epoch corresponds to a range of block heights during which the unified spending key parsed
+/// from an encoded form tagged with that epoch's identifier is expected to provide sufficient
+/// spending authority to spend any non-Sprout shielded note created in a transaction within the
```
daira, line 93: This might not be a correct description in future if, e.g. we close a pool (after Sprout) for spending. I don't know how important it is to document that now. (Non-blocking.)

zcash_client_backend/src/keys.rs
```diff
@@ -74,12 +86,56 @@ pub enum DerivationError {
...
+    fn id(&self) -> u32 {
+        // We use the consensus branch id of the network upgrade that introduced a
+        // new USK format as the identifier for the epoch.
+        match self {
+            Epoch::Orchard => u32::from(BranchId::Nu5),
+        }
+    }
+
+    #[cfg(feature = "unstable")]
+    fn from_id(id: u32) -> Option<Self> {
```
daira, line 126:
```suggestion
    fn try_from_id(id: u32) -> Option<Self> {
```

zcash_client_backend/src/keys.rs
```diff
@@ -74,12 +86,56 @@ pub enum DerivationError {
...
+        })
+    }
+}
+
 /// A set of viewing keys that are all associated with a single
 /// ZIP-0032 account identifier.
 #[derive(Clone, Debug)]
 #[doc(hidden)]
 pub struct UnifiedSpendingKey {
-    account: AccountId,
```
daira, line 82: Should we put the account id in the encoding rather than omitting it from this type? How was it used?

zcash_client_backend/src/keys.rs
```diff
@@ -144,16 +195,151 @@ impl UnifiedSpendingKey {
...
+                    let mut key = [0u8; 169];
+                    source
+                        .read_exact(&mut key)
+                        .map_err(|_| DecodingError::InsufficientData(Typecode::Sapling))?;
+                    sapling = Some(
+                        sapling::ExtendedSpendingKey::from_bytes(&key)
+                            .map_err(|_| DecodingError::KeyDataInvalid(Typecode::Sapling))?,
+                    );
+                }
+                #[cfg(feature = "transparent-inputs")]
```
daira, line 299: If the transparent feature is not enabled, should a USK that includes a P2PKH key still be accepted? Or do we require that all components communicating with each other are consistently compiled either to use or not use transparent (even across languages)?

daira (changes requested):
utACK with one blocking question.

zcash_client_backend/src/keys.rs
nuttycom: The inclusion of the account ID was a misfeature, since we would need at least an account ID and a seed fingerprint to correctly identify the relationship of a USK to a particular seed. The association with the account will be checked when the USK is used for spending, using information from the wallet database.

zcash_client_backend/src/keys.rs
nuttycom: That's a good point; this is a situation where the use of a feature flag to control transparent functionality is a bit awkward. I guess that for parsing purposes, if transparent functionality is not enabled then we should raise an error if we are asked to parse a USK containing a transparent component.

pacu: Given how we intend the USK to be used an error like this should probably be treated as a developer error, shouldn't it? I can't think of a users getting into this problem by just using the wallet because the USK is encoded in binary for internal interchange between the SDK/Wallet and the Rust crate. 

If I'm not mistaken is not possible to call `shield_funds` if the `transparent-inputs` is not enabled. There could be the case that a developer mistakenly saves the bytes somewhere and then if they move out of a build that includes such feature, they end up passing a USK with transparent keys. This is probably an error in migrating their software that was already using these keys as they are not intended to. :-/

I think an error must be raised to prevent this situation go unnoticed.

nuttycom: Actually, if `feature = "transparent-inputs"` is not enabled, at present this *will* return a `TypecodeInvalid` error.

daira: Right. My question was, should the transparent part be ignored instead?

nuttycom: Ah, okay. No, I think that this should be an error, because if the key was originally created with a transparent part, it may control transparent funds that will be otherwise unrecoverable. So I think an error is correct here.

daira (approved):
utACK 8ce53692d6d6db6380a852542a6245a33a114161

daira (approved):
re-utACK f1e8c04b6ce22430e533fd97a4f21bbb76ea531f

str4d (commented):
Flushing my comments.

zcash_client_backend/src/keys.rs
```diff
@@ -74,12 +86,56 @@ pub enum DerivationError {
...
 }
 
+/// A version identifier for the encoding of unified spending keys.
+///
+/// Each epoch corresponds to a range of block heights. During an epoch, the unified spending key
+/// parsed from an encoded form tagged with that epoch's identifier is expected to provide
+/// sufficient spending authority to spend any non-Sprout shielded note created in a transaction
+/// within the epoch's block range.
+#[derive(Debug, PartialEq, Eq)]
+pub enum Epoch {
```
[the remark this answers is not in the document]
str4d: In PR Gardening today I suggested renaming this to `Era`.

components/zcash_address/src/kind/unified.rs
```diff
@@ -88,6 +88,12 @@ impl From<Typecode> for u32 {
     }
 }
 
+impl From<Typecode> for usize {
+    fn from(t: Typecode) -> Self {
+        u32::from(t).try_into().unwrap()
```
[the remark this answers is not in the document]
daira: In `pasta_curves` we statically assert that `usize` is at least 32 bits: https://github.com/zcash/pasta_curves/blob/682a0e65b87a835df9cd1e1e55468d969269d992/src/arithmetic/fields.rs#L18 . We should just do the same here.

zcash_client_backend/src/keys.rs
```diff
@@ -74,12 +86,56 @@ pub enum DerivationError {
...
+    /// The Orchard epoch begins at Orchard activation, and will end if a new pool that requires a
+    /// change to unified spending keys is introduced.
+    Orchard,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum DecodingError {
+    ReadError(&'static str),
+    EpochInvalid,
+    EpochMismatch(Epoch),
```
daira, line 106: Don't forget to change these (e.g. to `EraInvalid` and `EraMismatch(Era)`) as well.

daira (approved):
re-utACK modulo @str4d's comments.

zcash_client_backend/src/keys.rs
```diff
@@ -144,16 +195,151 @@ impl UnifiedSpendingKey {
...
+    /// within wallets when required for storage and/or crossing FFI boundaries;
+    /// unified spending keys should not be exposed to users, and consequently
+    /// no string-based encoding is defined. This encoding does not include any
+    /// internal validation metadata (such as checksums) as keys decoded from
+    /// this form will necessarily be validated when the attempt is made to
+    /// spend a note that they have authority for.
+    #[cfg(feature = "unstable")]
+    pub fn to_bytes(&self, epoch: Epoch) -> Vec<u8> {
+        let mut result = vec![];
+        result.write_u32::<LittleEndian>(epoch.id()).unwrap();
```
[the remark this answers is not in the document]
nuttycom: Why are either of these better than what is written?

zcash_client_backend/src/keys.rs
```diff
@@ -144,16 +195,151 @@ impl UnifiedSpendingKey {
...
+    pub fn from_bytes(epoch: Epoch, encoded: &[u8]) -> Result<Self, DecodingError> {
+        let mut source = std::io::Cursor::new(encoded);
+        let decoded_epoch = source
+            .read_u32::<LittleEndian>()
+            .map_err(|_| DecodingError::ReadError("epoch"))
+            .and_then(|id| Epoch::try_from_id(id).ok_or(DecodingError::EpochInvalid))?;
+
+        if decoded_epoch != epoch {
+            return Err(DecodingError::EpochMismatch(decoded_epoch));
+        }
```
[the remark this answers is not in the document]
nuttycom: Given that we use `byteorder` extensively in `zcash_primitives`, so we have it as a transitive dependency anyway, what's the issue with using it here? I'll put it behind the unstable feature flag, but I think there's value in retaining the consistency of how we write little-endian data.

zcash_client_backend/src/keys.rs
```diff
@@ -144,16 +195,151 @@ impl UnifiedSpendingKey {
...
+
+        let mut orchard = None;
+        let mut sapling = None;
+        #[cfg(feature = "transparent-inputs")]
+        let mut transparent = None;
+        loop {
```
str4d, line 260: This loop allows the components to be read in any order. I guess that's fine, since we don't really care that this encoding is visibly canonical, since it is not visible to users.

zcash_client_backend/src/keys.rs
```diff
@@ -1,22 +1,34 @@
 //! Helper functions for managing light client key material.
-use zcash_address::unified::{self, Container, Encoding};
+use orchard;
+use zcash_address::unified::{self, Container, Encoding, Typecode};
 use zcash_primitives::{
-    consensus,
-    sapling::keys as sapling_keys,
+    consensus::{self},
```
str4d, line 5: ?
```suggestion
    consensus,
```

str4d (changes requested):
Changelog updates, and moving everything consistently behind the `unstable` feature flag, are blocking.

zcash_client_backend/src/keys.rs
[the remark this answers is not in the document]
str4d: It moves to stdlib APIs, removing the dependency on `byteorder`.

zcash_client_backend/src/keys.rs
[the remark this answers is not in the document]
str4d: I want to remove `byteorder` usage in `zcash_primitives`; it was only necessary while the stdlib did not have that support.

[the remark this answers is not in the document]
str4d: So my issue with using it here is just that we're adding more dependency on it. But it's a weak objection; we can always remove it later, as its usage isn't part of the public API.

zcash_client_backend/src/keys.rs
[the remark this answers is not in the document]
nuttycom: Since we want to change to `no_std` for these encodings anyway, I think we should wait until then to make that change.

zcash_client_backend/src/keys.rs
nuttycom: Yes, this is intentional.

zcash_client_backend/src/keys.rs
nuttycom: I wonder why cargo fmt didn't fix this; it was left over from adding (and then removing) other types from the consensus crate.

components/zcash_address/src/kind/unified.rs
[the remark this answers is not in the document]
nuttycom: Since this is just for typecodes and is in `zcash_address` (which doesn't provide any operations that require the curves associated with the underlying keys) I think that providing `TryFrom` and allowing the caller to unwrap is fine. In particular, where we're using this it's always safe to unwrap because we're using specific `TypeCode` constants.

zcash_primitives/src/legacy/keys.rs
```diff
@@ -64,6 +67,16 @@ impl AccountPrivKey {
             .derive_private_key(KeyIndex::Normal(child_index))
             .map(|k| k.private_key)
     }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.0.serialize()
```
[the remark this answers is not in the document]
nuttycom: Yes.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/f1e8c04..ebe5b626a8121a49e063de39125b0e35f5e28677) to address review comments,

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/f1e8c04..e8fce059d8b494e1711927187f6931e14beefa9b) to address review comments and rebase to current `main`

str4d (approved):
utACK 7b7288c750d4a2e12357f1fde045808dbbfa64ea

zcash_primitives/src/zip32.rs
```diff
@@ -423,15 +484,23 @@ impl ExtendedSpendingKey {
...
+    /// Encodes the extended spending key to the its seralized representation as defined in
+    /// [ZIP 32](https://zips.z.cash/zip-0032)
+    pub fn to_bytes(&self) -> [u8; 169] {
+        let mut result = [0u8; 169];
+        result[0] = self.depth;
+        (&mut result[1..5]).copy_from_slice(&self.parent_fvk_tag.as_bytes()[..]);
+        (&mut result[5..9]).copy_from_slice(&self.child_index.value().to_le_bytes()[..]);
+        (&mut result[9..41]).copy_from_slice(&self.chain_code.as_bytes()[..]);
+        (&mut result[41..137]).copy_from_slice(&self.expsk.to_bytes()[..]);
+        (&mut result[137..169]).copy_from_slice(&self.dk.as_bytes()[..]);
```
daira, line 496: I verified that this encoding is consistent with ZIP 32.

daira (commented):
Post-hoc ACK

PR #626, opened 2022-09-02
zcash_primitives: Make `NullifierDerivingKey` internals public

str4d:
The `zcashd` Rust code relies on being able to construct the Sapling
types transparently. This part of the "public API" of the crate was
broken when the `NullifierDerivingKey` newtype was introduced. We do
want to migrate to all of these types having stronger type safety
guarantees (by only constructing them via constructors), but that should
be done consistently across the types. For now we maintain the existing
API by changing `NullifierDerivingKey` to be a transparent newtype.


nuttycom (approved):
utACK

PR #629, opened 2022-09-06
[DO NOT MERGE] Experiment to track down memory leak

daira:
Signed-off-by: Daira Hopwood <daira@jacaranda.org>


daira:
This is no longer needed.

PR #632, opened 2022-09-12
Replace `get_address` with `get_current_address` and `get_next_available_address`

nuttycom:
This updates the data access API to provide diversified address functionality. In order to support this change, the addresses table is updated to store diversifier index information in big-endian order to allow sorting by diversifier index, and account initialization is updated to store the diversifier index accordingly. The currently unreleased `addresses_table` migration is updated to reflect this change.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -48,7 +48,7 @@ use zcash_primitives::{
     merkle_tree::{CommitmentTree, IncrementalWitness},
     sapling::{Node, Nullifier},
     transaction::{components::Amount, Transaction, TxId},
-    zip32::{AccountId, ExtendedFullViewingKey},
+    zip32::{AccountId, DiversifierIndex, ExtendedFullViewingKey},
 };
 
 use zcash_client_backend::{
@@ -153,8 +153,11 @@ impl<P: consensus::Parameters> WalletRead for WalletDb<P> {
         wallet::get_unified_full_viewing_keys(self)
     }
 
-    fn get_address(&self, account: AccountId) -> Result<Option<UnifiedAddress>, Self::Error> {
-        wallet::get_address_ua(self, account)
+    fn get_current_address(
+        &self,
+        account: AccountId,
+    ) -> Result<Option<UnifiedAddress>, Self::Error> {
+        wallet::get_current_address(self, account).map(|res| res.map(|(addr, _)| addr))
     }
 
     fn is_valid_account_extfvk(
@@ -276,8 +279,11 @@ impl<'a, P: consensus::Parameters> WalletRead for DataConnStmtCache<'a, P> {
         self.wallet_db.get_unified_full_viewing_keys()
     }
 
-    fn get_address(&self, account: AccountId) -> Result<Option<UnifiedAddress>, Self::Error> {
-        self.wallet_db.get_address(account)
+    fn get_current_address(
+        &self,
+        account: AccountId,
+    ) -> Result<Option<UnifiedAddress>, Self::Error> {
+        self.wallet_db.get_current_address(account)
     }
 
     fn is_valid_account_extfvk(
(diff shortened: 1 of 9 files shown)


daira (approved):
utACK

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e25e6fc..763945ca8efc68923fcde3c929acc328faeb7d4d) to fix test failures & add a test.

zcash_client_sqlite/src/lib.rs
```diff
@@ -396,6 +402,30 @@ impl<'a, P: consensus::Parameters> DataConnStmtCache<'a, P> {
...
+    ) -> Result<Option<UnifiedAddress>, Self::Error> {
+        match self.get_unified_full_viewing_keys()?.get(&account) {
+            Some(ufvk) => {
+                let mut search_from = wallet::get_current_address(self.wallet_db, account)?
+                    .map_or_else(DiversifierIndex::default, |(_, last_diversifier_index)| {
+                        last_diversifier_index
+                    });
+                search_from
+                    .increment()
+                    .map_err(|_| SqliteClientError::DiversifierIndexOutOfRange)?;
```
str4d, line 417: This is subtly incorrect, in a way that I believe doesn't actually break:

- If `wallet::get_current_address()` returns `Ok(Some(_)`, that means we've generated addresses for this account before, and we start searching from `last_diversifier_index + 1`, as expected.
- If `wallet::get_current_address()` returns `Ok(None)`, that means there is no entry in the `addresses` table for the given account. That implies no address has ever been generated for the account. In this scenario, we should start searching from `DiversifierIndex::zero()`, but we actually start from `DiversifierIndex::zero() + 1`, which skips an address.

In practice this doesn't break because when adding an account, we also always add the default UA, so we _never_ get `Ok(None)`. However, we should for correctness fix this:
```suggestion
                let search_from = wallet::get_current_address(self.wallet_db, account)?
                    .map(|(_, last_diversifier_index)| {
                        last_diversifier_index
                            .increment()
                            .map_err(|_| SqliteClientError::DiversifierIndexOutOfRange)
                    })
                    .transpose()?
                    .unwrap_or_else(DiversifierIndex::default);
```

str4d (changes requested):

zcash_client_sqlite/src/wallet.rs
```diff
@@ -177,21 +179,31 @@ pub fn get_address<P: consensus::Parameters>(
         })
 }
 
-pub(crate) fn get_address_ua<P: consensus::Parameters>(
+pub(crate) fn get_current_address<P: consensus::Parameters>(
     wdb: &WalletDb<P>,
     account: AccountId,
-) -> Result<Option<UnifiedAddress>, SqliteClientError> {
+) -> Result<Option<(UnifiedAddress, DiversifierIndex)>, SqliteClientError> {
     // This returns the first diversified address, which will be the default one.
```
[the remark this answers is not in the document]
nuttycom: I intentionally did not return the diversifier index here, because I'm not exposing any API that the user can take advantage of with it

zcash_client_sqlite/src/lib.rs
nuttycom: This doesn't work, because `increment` does mutation in place, so the `map` here results in `Option<()>`. But I agree that the increment is in the wrong place.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/71bd1c977eb66a00e473274ac30da13829166bb9..03b0ff7e0a6aeb27b40f0b1268d906fb4b6e986b) to fix the increment semantics & make small documentation fixes.

str4d (approved):
utACK 03b0ff7e0a6aeb27b40f0b1268d906fb4b6e986b. This alters an existing migration, but we have not cut a release containing that migration so it's not stable yet.

PR #636, opened 2022-09-13
zcash_client_backend: Add tags to IVKs in the batch scanner

str4d:
This removes the dependency on `SaplingIvk::to_repr()`, and enables us to alter the type of `D::IncomingViewingKey` to improve the performance of batch scanning.

For the welding rig, we already annotate the viewing keys with `AccountId`, so we use `(AccountId, Scope)` as the tag.

### DIFF
--- a/zcash_client_backend/src/scan.rs
+++ b/zcash_client_backend/src/scan.rs
@@ -7,17 +7,18 @@ use zcash_note_encryption::{batch, BatchDomain, Domain, ShieldedOutput, COMPACT_
 use zcash_primitives::{block::BlockHash, transaction::TxId};
 
 /// A decrypted note.
-pub(crate) struct DecryptedNote<D: Domain> {
-    /// The incoming viewing key used to decrypt the note.
-    pub(crate) ivk: D::IncomingViewingKey,
+pub(crate) struct DecryptedNote<A, D: Domain> {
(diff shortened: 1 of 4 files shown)


zcash_client_backend/src/scan.rs
```diff
@@ -123,30 +128,35 @@ impl<D: BatchDomain, Output: ShieldedOutput<D, COMPACT_NOTE_SIZE> + Clone> Batch
...
+            acc: Batch::new(tags, ivks),
             pending_results: HashMap::default(),
         }
     }
 }
 
-impl<D, Output> BatchRunner<D, Output>
+impl<A, D, Output> BatchRunner<A, D, Output>
 where
+    A: Clone + Send + 'static,
```
daira, line 159: Why `'static`?

str4d: This is necessary to use it with `rayon`:
image

str4d: Specifically, it's because we spawn a new task, which may outlive the current thread. Thus we need to ensure that `A` contains no lifetimes that could outlive the current thread.

daira (approved):
utACK

zcash_client_backend/src/scan.rs
daira: 👍

nuttycom (approved):
utACK

PR #637, opened 2022-09-13
Add WalletWrite::create_account function

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -1,5 +1,6 @@
 //! Interfaces for wallet data persistence & low-level wallet utilities.
 
+use secrecy::SecretVec;
```
str4d, line 3: This is in the wrong import block for our convention. Move it to the same block as `zcash_primitives` (or a new block immediately above it, but from a dependency perspective `secrecy` and `zcash_primitives` are both crates.io dependencies).

zcash_client_backend/src/data_api.rs
```diff
@@ -269,6 +270,14 @@ pub struct SentTransactionOutput<'a> {
 /// This trait encapsulates the write capabilities required to update stored
 /// wallet data.
 pub trait WalletWrite: WalletRead {
+    /// Creates a new account-level spending authority by derivation from the provided
+    /// seed using the next available unused account identifier, and returns this identifier
+    /// along with the generated unified spending key.
```
str4d, line 275:
```suggestion
    /// Tells the wallet to track the next available account-level spend authority, given
    /// the current set of [ZIP 316] account identifiers known to the wallet database.
    ///
    /// Returns the account identifier for the newly-created wallet database entry, along
    /// with the associated [`UnifiedSpendingKey`].
    ///
    /// If `seed` was imported from a backup and this method is being used to restore a
    /// previous wallet state, you should use this method to add all of the desired
    /// accounts before scanning the chain from the seed's birthday height.
    ///
    /// By convention, wallets should only allow a new account to be generated after funds
    /// have been received by the currently-available account (in order to enable
    /// automated account recovery).
    ///
    /// [ZIP 316]: https://zips.z.cash/zip-0316
```

zcash_client_backend/src/data_api.rs
```diff
@@ -269,6 +270,14 @@ pub struct SentTransactionOutput<'a> {
...
+    fn create_account(
+        &mut self,
+        seed: SecretVec<u8>,
```
str4d, line 278: Make this a reference to enable the caller to reuse the same seed for multiple calls to this method:
```suggestion
        seed: &SecretVec<u8>,
```

zcash_client_backend/src/data_api.rs
```diff
@@ -521,6 +533,16 @@ pub mod testing {
     }
 
     impl WalletWrite for MockWalletDb {
+        fn create_account(
+            &mut self,
+            seed: SecretVec<u8>,
```
str4d, line 538:
```suggestion
            seed: &SecretVec<u8>,
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -402,6 +403,29 @@ impl<'a, P: consensus::Parameters> DataConnStmtCache<'a, P> {
 
 #[allow(deprecated)]
 impl<'a, P: consensus::Parameters> WalletWrite for DataConnStmtCache<'a, P> {
+    fn create_account(
+        &mut self,
+        seed: SecretVec<u8>,
```
str4d, line 408:
```suggestion
        seed: &SecretVec<u8>,
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -402,6 +403,29 @@ impl<'a, P: consensus::Parameters> DataConnStmtCache<'a, P> {
...
+    ) -> Result<(AccountId, UnifiedSpendingKey), Self::Error> {
+        self.transactionally(|stmts| {
+            let account = wallet::get_max_account_id(stmts.wallet_db)?
+                .map(|a| AccountId::from(u32::from(a) + 1))
```
str4d, line 412: In a pairing, @nuttycom and I confirmed that if `a` here is (the theoretical) `u31::MAX`, then this increment will be detected inside `UnifiedSpendingKey::from_seed` (by the `orchard` crate) and an error will be returned.

We separately identified that the Sapling ZIP 32 impl could really use some love.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -179,6 +179,60 @@ pub fn get_address<P: consensus::Parameters>(
         })
 }
 
+pub(crate) fn get_max_account_id<P>(
+    wdb: &WalletDb<P>,
+) -> Result<Option<AccountId>, SqliteClientError> {
+    // This returns the most recently generated address.
+    Ok(wdb
+        .conn
+        .query_row("SELECT MAX(account) FROM accounts", NO_PARAMS, |row| {
```
str4d, line 188: We should maybe exclude `u31::MAX` from this selection, just in case the legacy account from `zcashd` somehow ends up in the wallet database (it shouldn't as there is by definition no ZIP 316 account at that position, but databases are fun). This would mean the code continues to correctly add accounts sequentially from whatever is the earliest non-legacy account ID. We would need to separately handle the case where this returns `u31::MAX - 1`, in which case the caller would currently increment that and then either incorrectly add the legacy account ID, or clobber whatever is there.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -516,12 +520,17 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
     Ok(())
 }
 
-/// Initialises the data database with the given [`UnifiedFullViewingKey`]s.
+/// Initialises the data database with the given set of account [`UnifiedFullViewingKey`]s.
+///
+/// **WARNING** This method should be used with care, and should ordinarily be unnecessary.
+/// Prefer to use [`zcash_client_backend::data_api::WalletWrite::create_account`] instead.
```
str4d, line 526:
```suggestion
/// Prefer to use [`WalletWrite::create_account`] instead.
///
/// [`WalletWrite::create_account`]: zcash_client_backend::data_api::WalletWrite::create_account
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -516,12 +520,17 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
...
 ///
 /// The [`UnifiedFullViewingKey`]s are stored internally and used by other APIs such as
-/// [`get_address`], [`scan_cached_blocks`], and [`create_spend_to_address`]. `extfvks` **MUST**
-/// be arranged in account-order; that is, the [`UnifiedFullViewingKey`] for ZIP 32
-/// account `i` **MUST** be at `extfvks[i]`.
+/// [`get_address`], [`scan_cached_blocks`], and [`create_spend_to_address`]. Account identifiers
+/// in `keys` **MUST** form a consecutive sequence beginning at account 0, and the
```
str4d, line 530: We should add a check for this (that the provided list's account IDs forms the appropriate set).

str4d (changes requested):
Reviewed d0062a87d4be6ad435974644828796d004a0a1f1. utACK after the various changes.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/637/commits/51a14464a2697deabcf5a10718043e8476d2c306

zcash_client_sqlite/src/wallet.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/637/commits/55e3afdbcf3a499d1dd59d5c33104c0d4e0e9d2e

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/637/commits/55e3afdbcf3a499d1dd59d5c33104c0d4e0e9d2e

str4d (approved):
utACK d086c57f2f105f2260670f36452956fa38249edc

PR #643, opened 2022-09-17
Add `encrypt-to-recipient` feature to `zcash_note_encryption`

nuttycom:
This adds a feature that provides the ability to encrypt an arbitrary payload such that it may be decrypted by both the recipient of a particular Zcash note and the holder of an OVK that can decrypt that note, if any such OVK exists.

### DIFF
--- a/components/zcash_note_encryption/CHANGELOG.md
+++ b/components/zcash_note_encryption/CHANGELOG.md
@@ -33,7 +33,7 @@ and this library adheres to Rust's notion of
   - `Domain::prepare_epk` method, which produces the above type.
 - Feature `encrypt-to-recipient`, which allows the encryption of an arbitrary
   payload decryptable by the recipient of a shielded note. This adds the
-  following trats and functions:
+  following traits and functions:
   - `PayloadEncryptionDomain`
   - `decrypt_associated_ciphertext_ivk`
   - `decrypt_associated_ciphertext_ovk`
--- a/components/zcash_note_encryption/CHANGELOG.md
+++ b/components/zcash_note_encryption/CHANGELOG.md
@@ -31,6 +31,16 @@ and this library adheres to Rust's notion of
 - `zcash_note_encryption::Domain`:
   - `Domain::PreparedEphemeralPublicKey` associated type.
   - `Domain::prepare_epk` method, which produces the above type.
+- Feature `encrypt-to-recipient`, which allows the encryption of an arbitrary
+  payload decryptable by the recipient of a shielded note. This adds the
+  following trats and functions:
+  - `PayloadEncryptionDomain`
+  - `decrypt_associated_ciphertext_ivk`
+  - `decrypt_associated_ciphertext_ovk`
+- A new `KeyedOutput` supertrait for `zcash_note_encryption::ShieldedOutput`
+- A new `RecoverableOutput` subtrait for `zcash_note_encryption::ShieldedOutput`,
+  which is used to simplify the type signatures of `try_output_recovery_with_ock`
+  and `try_output_recovery_with_ovk`.
 
 ### Changed
 - MSRV is now 1.56.1.
@@ -45,6 +55,17 @@ and this library adheres to Rust's notion of
     decryption results of the same length and in the same order as the `outputs`
     argument to the function. Each successful result includes the index of the
     entry in `ivks` used to decrypt the value.
+- Changes to the `zcash_note_encryption::ShieldedOutput` trait:
+  - The `ephemeral_key` and `cmstar_bytes` method have been moved into
+    a `KeyedOutput` supertrait of `ShieldedOutput`, for use in methods that
+    do not require the `enc_ciphertext` capability, such as
+    `decrypt_associated_ciphertext_ivk`.
+- The signatures of `try_output_recovery_with_ovk` and
+  `try_output_recovery_with_ock` have been changed to no longer take `cv` and
+  `out_ciphertext` parameters; instead, the output passed to these functions
+  must implement the `RecoverableOutput` trait. This ensures that the value
+  commitment and `out_ciphertext` values must correspond to the output being
+  decrypted, and simplifies the API.
 
 ## [0.1.0] - 2021-12-17
 Initial release.
--- a/components/zcash_note_encryption/src/lib.rs
+++ b/components/zcash_note_encryption/src/lib.rs
@@ -331,7 +331,7 @@ pub trait PayloadEncryptionDomain: Domain {
     /// Derives a `SymmetricKey` used to encrypt the an arbitrary payload distinct from the note
     /// plaintext.
     ///
-    /// This method provides the ability to use arbitrary personalization for the derivation of the
+    /// This method provides the ability to use personalization for the derivation of the
     /// secret key, so as to provide domain separation of the symmetric key when encrypting a
     /// payload to be readable by the recipient of a shielded note.
     ///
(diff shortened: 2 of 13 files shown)


daira:
This needs to be specified: https://github.com/zcash/zips/issues/633

components/zcash_note_encryption/CHANGELOG.md
```diff
@@ -10,6 +10,16 @@ and this library adheres to Rust's notion of
...
   - `Domain::prepare_epk` method, which produces the above type.
+- Feature `encrypt-to-recipient`, which allows the encryption of an arbitrary
+  payload decryptable by the recipient of a shielded note. This adds the
+  following trats and functions:
+  - `PayloadEncryptionDomain`
+  - `decrypt_associated_ciphertext_ivk`
+  - `decrypt_associated_ciphertext_ovk`
+- A new `KeyedOutput` supertrait for `zcash_note_encryption::ShieldedOutput`
+- A new `RecoverableOutput` subtrait for `zcash_note_encryption::ShieldedOutput`,
+  which is used to simplify the type signatures of `try_output_recovery_with_okc`
```
daira, line 21:
```suggestion
  which is used to simplify the type signatures of `try_output_recovery_with_ock`
```

components/zcash_note_encryption/src/lib.rs
```diff
@@ -321,22 +331,69 @@ pub trait BatchDomain: Domain {
     }
 }
 
-/// Trait that provides access to the components of an encrypted transaction output.
-///
-/// Implementations of this trait are required to define the length of their ciphertext
-/// field. In order to use the trial decryption APIs in this crate, the length must be
-/// either [`ENC_CIPHERTEXT_SIZE`] or [`COMPACT_NOTE_SIZE`].
-pub trait ShieldedOutput<D: Domain, const CIPHERTEXT_SIZE: usize> {
+#[cfg(feature = "encrypt-to-recipient")]
```
daira, line 334: Document this trait.

components/zcash_note_encryption/src/lib.rs
```diff
@@ -462,6 +519,31 @@ impl<D: Domain> NoteEncryption<D> {
...
+    /// Encrypts a payload that may decrypted with the same shared secret and epk as can
+    /// be used to decrypt the note plaintext for a given note. Returns a newly generated
+    /// nonce that must be used in decryption along with the ciphertext.
+    ///
+    /// This can be used to make a separate payload passed out-of-band decryptable to the
+    /// recipient of a Zcash payment.
+    pub fn encrypt_to_note_recipient<R: RngCore + CryptoRng>(
+        &self,
+        rng: &mut R,
+        kdf_personalization: &D::KdfPersonalization,
```
daira, line 533: The KDF personalization used for this feature should not be arbitrary, because then it can coincide with "Zcash_OrchardKDF" or some other already-used BLAKE2b-256 personalization, complicating security analysis. Also, it should be different depending on whether the domain is for Sapling or Orchard or some future shielded protocol. One option is to require it to start with, say, "ZcashSAK" or "ZcashOAK" (Zcash Sapling/Orchard Associated Key), with the remaining 8 bytes left arbitrary.

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -5,7 +5,8 @@ use group::GroupEncoding;
 use std::io::{self, Read, Write};
 
 use zcash_note_encryption::{
-    EphemeralKeyBytes, ShieldedOutput, COMPACT_NOTE_SIZE, ENC_CIPHERTEXT_SIZE,
+    EphemeralKeyBytes, KeyedOutput, RecoverableOutput, ShieldedOutput, COMPACT_NOTE_SIZE,
+    ENC_CIPHERTEXT_SIZE, OUT_CIPHERTEXT_SIZE,
 };
```
daira, line 10: Fix the clippy warning below by deriving `Eq`.

daira:
The nonce for IETF_CHACHA20_POLY1305 is 96 bits. That means that if you encrypt a large number of plaintexts with this feature using the same epk, you might get a nonce collision with larger probability than we have targeted for the security of the rest of the protocol. (We'd want a 192-bit nonce at least, I think.)

We can either say something like "Don't reuse the same epk for more than 2^20 encryptions.", or we can use XChaCha20-Poly1305 for this feature. I realise that the latter complicates the implementation significantly.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/5c70114b91b59caef0d10fdf488f7d9abc75fd8e..571cb460a8dd9dddcd452eaa1f2a011d19af0062) to use XChaCha20Poly1305 for associated payload encryption.

zcash_primitives/src/transaction/components/sapling.rs
nuttycom: Fixed in #645

components/zcash_note_encryption/src/lib.rs
nuttycom: I think that the name should remain `KdfPersonalization`, so that in the event that we move away from Blake2b in a future protocol and use a hash that requires a different approach to personalization, that the trait remains generic. But I will change the type at the implementation site and document it in those places as 8-byte suffixes.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/571cb460a8dd9dddcd452eaa1f2a011d19af0062..baf4f926a8fbedd4ccf0c3da7a97f03a511595de) to only allow key personalization suffixes instead of arbitrary personalization.

components/zcash_note_encryption/CHANGELOG.md
```diff
@@ -10,6 +10,16 @@ and this library adheres to Rust's notion of
 - `zcash_note_encryption::Domain`:
   - `Domain::PreparedEphemeralPublicKey` associated type.
   - `Domain::prepare_epk` method, which produces the above type.
+- Feature `encrypt-to-recipient`, which allows the encryption of an arbitrary
+  payload decryptable by the recipient of a shielded note. This adds the
+  following trats and functions:
```
daira, line 15:
```suggestion
  following traits and functions:
```

components/zcash_note_encryption/src/lib.rs
```diff
@@ -321,22 +331,79 @@ pub trait BatchDomain: Domain {
...
+/// shielded protocol, to ensure that keys cannot be made to collide with those produced by the
+/// KDF used for note encryption.
+#[cfg(feature = "encrypt-to-recipient")]
+pub trait PayloadEncryptionDomain: Domain {
+    type KdfPersonalization;
+
+    /// Derives a `SymmetricKey` used to encrypt the an arbitrary payload distinct from the note
+    /// plaintext.
+    ///
+    /// This method provides the ability to use arbitrary personalization for the derivation of the
```
daira, line 351:
```suggestion
    /// This method provides the ability to use personalization for the derivation of the
```

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -257,26 +258,40 @@ pub struct OutputDescription<Proof> {
     pub cmu: bls12_381::Scalar,
     pub ephemeral_key: EphemeralKeyBytes,
     pub enc_ciphertext: [u8; 580],
```
daira, line 260:
```suggestion
    pub enc_ciphertext: [u8; ENC_CIPHERTEXT_SIZE],
```

daira (approved):
utACK with minor suggestions.

nuttycom:
Rebased to fix conflicts; the diff is not readily reviewable so this will require a fresh set of reviews.

nuttycom:
With the extraction of the `zcash_note_encryption` crate, these proposed changes are now available [here](https://github.com/nuttycom/zcash_note_encryption/tree/payload_encryption)

daira:
I suggest opening a PR on the new repo, even if it's just a draft PR. I almost never look at branches that aren't a PR branch.

PR #644, opened 2022-09-17
zcash_client_sqlite: Fix `FsBlockDb::get_max_cached_height`

str4d:
The `MAX` SQLite function returns `null` when the table is empty. The code was expecting zero rows to be returned in this case, and was trying to parse the `null` as an integer.


nuttycom (approved):
utACK

PR #653, opened 2022-09-26
zcash_client_sqlite: fix wallet::get_max_account_id

nuttycom:
This had the same bug as was fixed in 5b8ec380a006ccc0ccb38c11cc6832e1176015c0


pacu (approved):

pacu:
utACK

PR #657, opened 2022-10-01
Use upsert functionality for transparent UTXOs, rather than delete/repopulate.

nuttycom:
The ios FFI was still relying upon a deprecated approach to handling transparent UTXOs, deleting
all of them above a certain height and repopulating them each time they were retrieved. Now, instead,
the `put_received_transparent_utxo` data access API method actually performs a proper upsert, so
there should be no need for the clear function.

This builds on #656


pacu (approved):
utACK

daira (approved):
utACK

str4d (approved):
utACK 6df6fec860c9cb475332e69c494d2583b2c4a88b

PR #385, opened 2021-05-19
ZIP-225/244 #1: Minor refactoring and preparatory updates.

nuttycom:
This is a set of minor improvements to type safety factored out from #375.

PR #658, opened 2022-10-02
Update the transaction builder to make change outputs explicit

nuttycom:
This introduces a FeeStrategy abstraction (and `FutureFeeStrategy` for `zfuture`) that can be used to compute fees and required change during the process of transaction build. This also modifies `TransactionBuilder` to require the caller to construct a change output directly, instead of doing so implicitly.

### DIFF
--- a/zcash_client_backend/src/data_api/error.rs
+++ b/zcash_client_backend/src/data_api/error.rs
@@ -6,7 +6,11 @@ use zcash_address::unified::Typecode;
 use zcash_primitives::{
     consensus::BlockHeight,
     sapling::Node,
-    transaction::{builder, components::amount::Amount, TxId},
+    transaction::{
+        builder,
+        components::amount::{Amount, BalanceError},
+        TxId,
+    },
     zip32::AccountId,
 };
 
@@ -33,8 +37,8 @@ pub enum Error<NoteId> {
     /// No account with the given identifier was found in the wallet.
     AccountNotFound(AccountId),
 
-    /// The amount specified exceeds the allowed range.
-    InvalidAmount,
+    /// Zcash amount computation encountered an overflow or underflow.
+    BalanceError(BalanceError),
 
     /// Unable to create a new spend because the wallet balance is not sufficient.
     /// The first argument is the amount available, the second is the amount needed
@@ -113,9 +117,9 @@ impl<N: fmt::Display> fmt::Display for Error<N> {
             Error::AccountNotFound(account) => {
                 write!(f, "Wallet does not contain account {}", u32::from(*account))
             }
-            Error::InvalidAmount => write!(
+            Error::BalanceError(e) => write!(
                 f,
-                "The value lies outside the valid range of Zcash amounts."
+                "The value lies outside the valid range of Zcash amounts: {:?}.", e
             ),
             Error::InsufficientBalance(have, need) => write!(
                 f,
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -5,9 +5,10 @@ use zcash_primitives::{
     sapling::prover::TxProver,
     transaction::{
         builder::Builder,
-        components::{amount::DEFAULT_FEE, Amount},
+        components::amount::{Amount, BalanceError, DEFAULT_FEE},
         Transaction,
     },
+    zip32::Scope,
 };
 
 use crate::{
@@ -17,6 +18,7 @@ use crate::{
         SentTransactionOutput, WalletWrite,
     },
     decrypt_transaction,
+    fees::{BasicFixedFeeChangeStrategy, ChangeError, ChangeStrategy, ChangeValue},
     keys::UnifiedSpendingKey,
     wallet::OvkPolicy,
     zip321::{Payment, TransactionRequest},
@@ -92,19 +94,18 @@ where
 /// * `wallet_db`: A read/write reference to the wallet database
 /// * `params`: Consensus parameters
 /// * `prover`: The TxProver to use in constructing the shielded transaction.
-/// * `account`: The ZIP32 account identifier associated with the extended spending
-///   key that controls the funds to be used in creating this transaction.  This
-///   procedure will return an error if this does not correctly correspond to `extsk`.
-/// * `extsk`: The extended spending key that controls the funds that will be spent
-///   in the resulting transaction.
-/// * `amount`: The amount to send.
+/// * `usk`: The unified spending key that controls the funds that will be spent
+///   in the resulting transaction. This procedure will return an error if the
+///   USK does not correspond to an account known to the wallet.
 /// * `to`: The address to which `amount` will be paid.
+/// * `amount`: The amount to send.
 /// * `memo`: A memo to be included in the output to the recipient.
 /// * `ovk_policy`: The policy to use for constructing outgoing viewing keys that
 ///   can allow the sender to view the resulting notes on the blockchain.
 /// * `min_confirmations`: The minimum number of confirmations that a previously
 ///   received note must have in the blockchain in order to be considered for being
 ///   spent. A value of 10 confirmations is recommended.
+///
 /// # Examples
 ///
 /// ```
@@ -236,10 +237,8 @@ where
 /// * `params`: Consensus parameters
 /// * `prover`: The TxProver to use in constructing the shielded transaction.
 /// * `usk`: The unified spending key that controls the funds that will be spent
-///   in the resulting transaction.
-/// * `account`: The ZIP32 account identifier associated with the extended spending
-///   key that controls the funds to be used in creating this transaction.  This
-///   procedure will return an error if this does not correctly correspond to `extsk`.
+///   in the resulting transaction. This procedure will return an error if the
+///   USK does not correspond to an account known to the wallet.
 /// * `request`: The ZIP-321 payment request specifying the recipients and amounts
 ///   for the transaction.
 /// * `ovk_policy`: The policy to use for constructing outgoing viewing keys that
@@ -267,17 +266,17 @@ where
         .get_account_for_ufvk(&usk.to_unified_full_viewing_key())?
         .ok_or(Error::KeyNotRecognized)?;
 
-    let extfvk = usk.sapling().to_extended_full_viewing_key();
+    let dfvk = usk.sapling().to_diversifiable_full_viewing_key();
 
     // Apply the outgoing viewing key policy.
     let ovk = match ovk_policy {
-        OvkPolicy::Sender => Some(extfvk.fvk.ovk),
+        OvkPolicy::Sender => Some(dfvk.fvk().ovk),
         OvkPolicy::Custom(ovk) => Some(ovk),
         OvkPolicy::Discard => None,
     };
 
     // Target the next block, assuming we are up-to-date.
-    let (height, anchor_height) = wallet_db
+    let (target_height, anchor_height) = wallet_db
         .get_target_and_anchor_heights(min_confirmations)
         .and_then(|x| x.ok_or_else(|| Error::ScanRequired.into()))?;
 
@@ -286,8 +285,9 @@ where
         .iter()
         .map(|p| p.amount)
         .sum::<Option<Amount>>()
-        .ok_or_else(|| E::from(Error::InvalidAmount))?;
-    let target_value = (value + DEFAULT_FEE).ok_or_else(|| E::from(Error::InvalidAmount))?;
+        .ok_or_else(|| E::from(Error::BalanceError(BalanceError::Overflow)))?;
+    let target_value = (value + DEFAULT_FEE)
+        .ok_or_else(|| E::from(Error::BalanceError(BalanceError::Overflow)))?;
     let spendable_notes =
         wallet_db.select_spendable_sapling_notes(account, target_value, anchor_height)?;
 
@@ -296,7 +296,7 @@ where
         .iter()
         .map(|n| n.note_value)
         .sum::<Option<_>>()
-        .ok_or_else(|| E::from(Error::InvalidAmount))?;
+        .ok_or_else(|| E::from(Error::BalanceError(BalanceError::Overflow)))?;
     if selected_value < target_value {
         return Err(E::from(Error::InsufficientBalance(
             selected_value,
@@ -305,10 +305,10 @@ where
     }
 
     // Create the transaction
-    let mut builder = Builder::new_with_fee(params.clone(), height, DEFAULT_FEE);
+    let mut builder = Builder::new(params.clone(), target_height);
     for selected in spendable_notes {
-        let from = extfvk
-            .fvk
+        let from = dfvk
+            .fvk()
             .vk
             .to_payment_address(selected.diversifier)
             .unwrap(); //DiversifyHash would have to unexpectedly return the zero point for this to be None
@@ -361,7 +361,42 @@ where
         }?
     }
 
-    let (tx, tx_metadata) = builder.build(&prover).map_err(Error::Builder)?;
+    let fee_strategy = BasicFixedFeeChangeStrategy::new(DEFAULT_FEE);
+    let balance = fee_strategy
+        .compute_balance(
+            params,
+            target_height,
+            builder.transparent_inputs(),
+            builder.transparent_outputs(),
+            builder.sapling_inputs(),
+            builder.sapling_outputs(),
+        )
+        .map_err(|e| match e {
+            ChangeError::InsufficientFunds {
+                available,
+                required,
+            } => Error::InsufficientBalance(available, required),
+            ChangeError::StrategyError(e) => Error::BalanceError(e),
+        })?;
+
+    for change_value in balance.proposed_change() {
+        match change_value {
+            ChangeValue::Sapling(amount) => {
+                builder
+                    .add_sapling_output(
+                        Some(dfvk.to_ovk(Scope::Internal)),
+                        dfvk.change_address().1,
+                        *amount,
+                        MemoBytes::empty(),
+                    )
+                    .map_err(Error::Builder)?;
+            }
+        }
+    }
+
+    let (tx, tx_metadata) = builder
+        .build(&prover, &fee_strategy.fee_rule())
+        .map_err(Error::Builder)?;
 
     let sent_outputs = request.payments().iter().enumerate().map(|(i, payment)| {
         let (output_index, recipient) = match &payment.recipient_address {
@@ -405,7 +440,7 @@ where
         created: time::OffsetDateTime::now_utc(),
         account,
         outputs: sent_outputs,
-        fee_amount: DEFAULT_FEE,
+        fee_amount: balance.fee_required(),
         #[cfg(feature = "transparent-inputs")]
         utxos_spent: vec![],
     })
@@ -423,12 +458,12 @@ where
 /// * `wallet_db`: A read/write reference to the wallet database
 /// * `params`: Consensus parameters
 /// * `prover`: The TxProver to use in constructing the shielded transaction.
-/// * `sk`: The secp256k1 secret key that will be used to detect and spend transparent
-///   UTXOs.
-/// * `account`: The ZIP32 account identifier for the account to which funds will
-///   be shielded. Funds will be shielded to the internal (change) address associated with the
-///   most preferred shielded receiver corresponding to this account, or if no shielded
-///   receiver can be used for this account, this function will return an error.
+/// * `usk`: The unified spending key that will be used to detect and spend transparent UTXOs,
+///   and that will provide the shielded address to which funds will be sent. Funds will be
+///   shielded to the internal (change) address associated with the most preferred shielded
+///   receiver corresponding to this account, or if no shielded receiver can be used for this
+///   account, this function will return an error. This procedure will return an error if the
+///   USK does not correspond to an account known to the wallet.
 /// * `memo`: A memo to be included in the output to the (internal) recipient.
 ///   This can be used to take notes about auto-shielding operations internal
 ///   to the wallet that the wallet can use to improve how it represents those
@@ -462,7 +497,7 @@ where
         .to_diversifiable_full_viewing_key()
         .change_address()
         .1;
-    let (latest_scanned_height, latest_anchor) = wallet_db
+    let (target_height, latest_anchor) = wallet_db
         .get_target_and_anchor_heights(min_confirmations)
         .and_then(|x| x.ok_or_else(|| Error::ScanRequired.into()))?;
 
@@ -476,21 +511,15 @@ where
         utxos.append(&mut outputs);
     }
 
-    let total_amount = utxos
+    let _total_amount = utxos
         .iter()
         .map(|utxo| utxo.txout().value)
         .sum::<Option<Amount>>()
-        .ok_or_else(|| E::from(Error::InvalidAmount))?;
-
-    let fee = DEFAULT_FEE;
-    if fee >= total_amount {
-        return Err(E::from(Error::InsufficientBalance(total_amount, fee)));
-    }
-
-    let amount_to_shield = (total_amount - fee).ok_or_else(|| E::from(Error::InvalidAmount))?;
+        .ok_or_else(|| E::from(Error::BalanceError(BalanceError::Overflow)))?;
 
     let addr_metadata = wallet_db.get_transparent_receivers(account)?;
-    let mut builder = Builder::new_with_fee(params.clone(), latest_scanned_height, fee);
+    let mut builder = Builder::new(params.clone(), target_height);
+
     for utxo in &utxos {
         let diversifier_index = addr_metadata
             .get(utxo.recipient_address())
@@ -510,15 +539,44 @@ where
             .map_err(Error::Builder)?;
     }
 
-    // there are no sapling notes so we set the change manually
-    builder.send_change_to(ovk, shielding_address.clone());
+    // Compute the balance of the transaction. We have only added inputs, so the total change
+    // amount required will be the total of the UTXOs minus fees.
+    let fee_strategy = BasicFixedFeeChangeStrategy::new(DEFAULT_FEE);
+    let balance = fee_strategy
+        .compute_balance(
+            params,
+            target_height,
+            builder.transparent_inputs(),
+            builder.transparent_outputs(),
+            builder.sapling_inputs(),
+            builder.sapling_outputs(),
+        )
+        .map_err(|e| match e {
+            ChangeError::InsufficientFunds {
+                available,
+                required,
+            } => Error::InsufficientBalance(available, required),
+            ChangeError::StrategyError(e) => Error::BalanceError(e),
+        })?;
 
-    // add the sapling output to shield the funds
-    builder
-        .add_sapling_output(Some(ovk), shielding_address, amount_to_shield, memo.clone())
+    let fee = balance.fee_required();
+    let mut total_out = Amount::zero();
+    for change_value in balance.proposed_change() {
+        total_out = (total_out + change_value.value())
+            .ok_or(Error::BalanceError(BalanceError::Overflow))?;
+        match change_value {
+            ChangeValue::Sapling(amount) => {
+                builder
+                    .add_sapling_output(Some(ovk), shielding_address.clone(), *amount, memo.clone())
+                    .map_err(Error::Builder)?;
+            }
+        }
+    }
+
+    // The transaction build process will check that the inputs and outputs balance
+    let (tx, tx_metadata) = builder
+        .build(&prover, &fee_strategy.fee_rule())
         .map_err(Error::Builder)?;
-
-    let (tx, tx_metadata) = builder.build(&prover).map_err(Error::Builder)?;
     let output_index = tx_metadata.output_index(0).expect(
         "No sapling note was created in autoshielding transaction. This is a programming error.",
     );
@@ -529,8 +587,8 @@ where
         account,
         outputs: vec![SentTransactionOutput {
             output_index,
-            value: amount_to_shield,
             recipient: Recipient::InternalAccount(account, PoolType::Sapling),
+            value: total_out,
             memo: Some(memo.clone()),
         }],
         fee_amount: fee,
--- a/zcash_client_backend/src/fees.rs
+++ b/zcash_client_backend/src/fees.rs
@@ -1,0 +1,163 @@
+use zcash_primitives::{
+    consensus::{self, BlockHeight},
+    transaction::{
+        components::{
+            amount::{Amount, BalanceError},
+            sapling::builder::{SaplingInput, SaplingOutput},
+            transparent::{builder::TransparentInput, TxOut},
+        },
+        fees::{FeeRule, FixedFeeRule},
+    },
+};
+
+/// A proposed change amount and output pool.
+pub enum ChangeValue {
+    Sapling(Amount),
+}
+
+impl ChangeValue {
+    pub fn value(&self) -> Amount {
+        match self {
+            ChangeValue::Sapling(value) => *value,
+        }
+    }
+}
+
+/// The amount of change and fees required to make a transaction's inputs and
+/// outputs balance under a specific fee rule, as computed by a particular
+/// [`ChangeStrategy`] that is aware of that rule.
+pub struct TransactionBalance {
+    proposed_change: Vec<ChangeValue>,
+    fee_required: Amount,
+}
+
+impl TransactionBalance {
+    /// Constructs a new balance from its constituent parts.
+    pub fn new(proposed_change: Vec<ChangeValue>, fee_required: Amount) -> Self {
+        TransactionBalance {
+            proposed_change,
+            fee_required,
+        }
+    }
+
+    /// The change values proposed by the [`ChangeStrategy`] that computed this balance.  
+    pub fn proposed_change(&self) -> &[ChangeValue] {
+        &self.proposed_change
+    }
+
+    /// Returns the fee computed for the transaction, assuming that the suggested
+    /// change outputs are added to the transaction.
+    pub fn fee_required(&self) -> Amount {
+        self.fee_required
+    }
+}
+
+/// Errors that can occur in balance
+pub enum ChangeError<E> {
+    InsufficientFunds { available: Amount, required: Amount },
+    StrategyError(E),
+}
+
+/// A trait that represents the ability to compute the suggested change and fees that must be paid
+/// by a transaction having a specified set of inputs and outputs.
+pub trait ChangeStrategy {
+    type FeeRule: FeeRule;
+    type Error;
+
+    /// Returns the fee rule that this change strategy will respect when performing
+    /// balance computations.
+    fn fee_rule(&self) -> Self::FeeRule;
+
+    /// Computes the totals of inputs, suggested change amounts, and fees given the
+    /// provided inputs and outputs being used to construct a transaction.
+    ///
+    /// The fee computed as part of this operation should take into account the prospective
+    /// change outputs recommended by this operation. If insufficient funds are available to
+    /// supply the requested outputs and required fees, implementations should return
+    /// [`ChangeError::InsufficientFunds`].
+    fn compute_balance<P: consensus::Parameters>(
+        &self,
+        params: &P,
+        target_height: BlockHeight,
+        transparent_inputs: &[impl TransparentInput],
+        transparent_outputs: &[TxOut],
+        sapling_inputs: &[impl SaplingInput],
+        sapling_outputs: &[SaplingOutput],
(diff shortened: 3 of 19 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -365,7 +364,25 @@ where
         }?
     }
 
-    let (tx, tx_metadata) = builder.build(&prover).map_err(Error::Builder)?;
+    let fee_strategy = FixedFeeStrategy::new(DEFAULT_FEE);
```
daira, line 367: Line 429 below always stores `fee_amount: DEFAULT_FEE` in the wallet db regardless of the strategy.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -63,11 +60,19 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Error::ChangeIsNegative(amount) => {
-                write!(f, "Change is negative ({:?} zatoshis)", amount)
+            Error::BalanceFailure(amount) => {
```
daira, line 63: Please document the meaning of this value's sign, both in the doc for `Error::BalanceFailure`, and the error message below.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -107,8 +112,164 @@ impl Progress {
...
-enum ChangeAddress {
-    SaplingChangeAddress(OutgoingViewingKey, PaymentAddress),
+pub struct TransactionBalances {
+    total_in: Amount,
+    change_required: Amount,
+    fee_required: Amount,
+}
+
+impl TransactionBalances {
+    pub fn new(
```
daira, line 122: Needs documentation.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -323,32 +496,16 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
-                }
-                None => {
-                    let (ovk, addr) = self
-                        .sapling_builder
-                        .get_candidate_change_address()
-                        .ok_or(Error::NoChangeAddress)?;
-                    self.add_sapling_output(Some(ovk), addr, change, MemoBytes::empty())?;
-                }
-            }
+        // After fees are accounted for, the value balance of the transaction must be zero.
```
daira, line 504: Add a comment about the apparent redundancy between this check and the one for change above.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -546,7 +728,10 @@ mod tests {
         );
     }
 
+    // This test only works with the transparent_inputs feature because we have to
+    // be able to create a tx with a valid balance, without using Sapling inputs.
```
daira, line 732: File a ticket to use Orchard instead when that's supported.

zcash_primitives/src/transaction/components/transparent/builder.rs
```diff
@@ -40,6 +40,26 @@ impl fmt::Display for Error {
...
 
+pub trait TransparentInput {
+    fn outpoint(&self) -> &OutPoint;
+    fn coin(&self) -> &TxOut;
+}
+
+/// A never-instantiated zero-sized struct to allow the type of [`Builder::inputs`]
+/// to resolve when the transparent-inputs feature is not turned on.
+#[cfg(not(feature = "transparent-inputs"))]
+struct InvalidTransparentInput;
```
daira, line 51: Make this an uninhabited type (enum with no variants).

daira (commented):
Looks good so far.

daira (approved):
utACK modulo the comments from my previous review. In particular there are still some bare uses of `DEFAULT_FEE` that assume fees are fixed without going through `FixedFeeStrategy`.

nuttycom:
@daira the remaining uses of `DEFAULT_FEE` will be addressed in https://github.com/zcash/librustzcash/pull/689, which builds atop this PR.

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -225,6 +243,14 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
         }
     }
 
+    pub fn inputs(&self) -> &[impl SaplingInput] {
```
str4d, line 246: Does this need to be public? If so:
- Needs documentation.
- Document this addition in the changelog.

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -225,6 +243,14 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+        &self.spends
+    }
+
+    pub fn outputs(&self) -> &[SaplingOutput] {
```
str4d, line 250: Does this need to be public? If so:
- Needs documentation.
- Document this addition in the changelog.

zcash_primitives/src/transaction/components/transparent/builder.rs
```diff
@@ -40,6 +40,26 @@ impl fmt::Display for Error {
     }
 }
 
+pub trait TransparentInput {
```
str4d, line 43: - Needs documentation (the trait and its methods).
- Document this addition in the changelog.

zcash_primitives/src/transaction/components/transparent/builder.rs
```diff
@@ -40,6 +40,26 @@ impl fmt::Display for Error {
...
+    fn outpoint(&self) -> &OutPoint;
+    fn coin(&self) -> &TxOut;
+}
+
+/// An uninhabited type that allows the type of [`Builder::inputs`]
```
str4d, line 48:
```suggestion
/// An uninhabited type that allows the type of [`TransparentBuilder::inputs`]
```

zcash_primitives/src/transaction/components/transparent/builder.rs
```diff
@@ -40,6 +40,26 @@ impl fmt::Display for Error {
...
+/// to resolve when the transparent-inputs feature is not turned on.
+#[cfg(not(feature = "transparent-inputs"))]
+pub enum InvalidTransparentInput {}
```
str4d, line 51: Does this need to be public? AFAICT it is only used as a type inside this module, and otherwise is exposed as `impl TransparentInput` which does not require this type to be public (similarly to `TransparentInputInfo`).
```suggestion
enum InvalidTransparentInput {}
```

zcash_primitives/src/transaction/components/transparent/builder.rs
```diff
@@ -80,6 +111,21 @@ impl TransparentBuilder {
         }
     }
 
+    pub fn inputs(&self) -> &[impl TransparentInput] {
```
str4d, line 114: - Needs documentation.
- Document this addition in the changelog.

zcash_primitives/src/transaction/components/transparent/builder.rs
```diff
@@ -80,6 +111,21 @@ impl TransparentBuilder {
...
+        return &self.inputs;
+
+        #[cfg(not(feature = "transparent-inputs"))]
+        {
+            let invalid: &[InvalidTransparentInput] = &[];
+            return invalid;
+        }
+    }
+
+    pub fn outputs(&self) -> &[TxOut] {
```
str4d, line 125: - Needs documentation.
- Document this addition in the changelog.

zcash_primitives/src/transaction/components/tze.rs
```diff
@@ -212,6 +212,11 @@ impl TzeOut {
     }
 }
 
+pub trait TzeInput {
```
str4d, line 215: - Needs documentation (the trait and its methods).
- Document this addition in the changelog.

zcash_primitives/src/transaction/components/tze/builder.rs
```diff
@@ -32,13 +32,27 @@ impl fmt::Display for Error {
 
 #[allow(clippy::type_complexity)]
 pub struct TzeSigner<'a, BuildCtx> {
-    prevout: TzeOut,
     builder: Box<dyn FnOnce(&BuildCtx) -> Result<(u32, Vec<u8>), Error> + 'a>,
 }
 
+#[derive(Clone)]
+pub struct TzeBuildInput {
```
str4d, line 39: Does this need to be public? AFAICT it is only used as a type inside this module, and otherwise is exposed as `impl TzeBuildInput` which does not require this type to be public.
```suggestion
struct TzeBuildInput {
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -107,8 +120,179 @@ impl Progress {
...
+        sapling_outputs: &[SaplingOutput],
+    ) -> Result<TransactionBalances, Self::Error>;
+}
+
+/// A trait that represents the ability to compute the fees that must be paid
+/// by a transaction having a specified set of inputs and outputs, for use
+/// when experimenting with the TZE feature.
+#[cfg(feature = "zfuture")]
+pub trait FutureFeeStrategy: FeeStrategy {
+    fn compute_balances_zfuture(
```
str4d, line 189: Needs documentation.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -18,56 +19,68 @@ use crate::{
...
-
 const DEFAULT_TX_EXPIRY_DELTA: u32 = 20;
 
+/// Errors that can occur during transaction construction.
 #[derive(Debug, PartialEq, Eq)]
 pub enum Error {
-    ChangeIsNegative(Amount),
+    /// Insufficient funds were provided to the transaction builder; the given
+    /// additional amount is required in order to construct the transaction.
+    InsufficientFunds(Amount),
```
str4d, line 55: Document removals and additions in the changelog.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -107,8 +120,179 @@ impl Progress {
     }
 }
 
-enum ChangeAddress {
-    SaplingChangeAddress(OutgoingViewingKey, PaymentAddress),
+/// The balance of inputs, outputs, and fees of a transaction under construction.
+pub struct TransactionBalances {
```
str4d, line 124: Document this addition in the changelog.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -107,8 +120,179 @@ impl Progress {
...
+    total_in: Amount,
+    change_required: Amount,
+    fee_required: Amount,
+}
+
+impl TransactionBalances {
+    /// Constructs a new balance from its constituent parts. Returns an error
+    /// on overflow or underflow of amount arithmetic.
+    pub fn new(
```
str4d, line 133: This appears to be public so that people can implement `FeeStrategy` externally. But currently `FeeStrategy` is sealed because the types in its method can't be named externally. If it is intended to be sealed, then this constructor doesn't need to be public:
```suggestion
    fn new(
```

str4d (changes requested):
Partial review of 6280ddebc556ef06064e603c2f04b4e43d24bd73.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -136,23 +318,9 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, OsRng> {
     ///
     /// The expiry height will be set to the given height plus the default transaction
     /// expiry delta (20 blocks).
-    ///
-    /// The fee will be set to the default fee (0.0001 ZEC).
```
str4d, line 140: Document this change in the changelog.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -136,23 +318,9 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, OsRng> {
...
-
-    /// Creates a new `Builder` targeted for inclusion in the block with the given height, using
-    /// the specified fee, and otherwise default values for general transaction fields and the
-    /// default OS random.
-    ///
-    /// # Default values
-    ///
-    /// The expiry height will be set to the given height plus the default transaction
-    /// expiry delta (20 blocks).
-    pub fn new_with_fee(params: P, target_height: BlockHeight, fee: Amount) -> Self {
```
str4d, line 153: Document this removal in the changelog.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -163,27 +331,8 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
     ///
     /// The expiry height will be set to the given height plus the default transaction
     /// expiry delta (20 blocks).
-    ///
-    /// The fee will be set to the default fee (0.0001 ZEC).
```
str4d, line 167: Document this change in the changelog.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -163,27 +331,8 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
...
-
-    /// Creates a new `Builder` targeted for inclusion in the block with the given height, and
-    /// randomness source, using the specified fee, and otherwise default values for general
-    /// transaction fields and the default OS random.
-    ///
-    /// # Default values
-    ///
-    /// The expiry height will be set to the given height plus the default transaction
-    /// expiry delta (20 blocks).
-    pub fn new_with_rng_and_fee(
```
str4d, line 180: Document this removal in the changelog.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -269,14 +411,6 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
             .map_err(Error::TransparentBuild)
     }
 
-    /// Sets the Sapling address to which any change will be sent.
-    ///
-    /// By default, change is sent to the Sapling address corresponding to the first note
-    /// being spent (i.e. the first call to [`Builder::add_sapling_spend`]).
-    pub fn send_change_to(&mut self, ovk: OutgoingViewingKey, to: PaymentAddress) {
```
str4d, line 276: Document this removal in the changelog.

zcash_primitives/src/transaction/builder.rs
nuttycom: FeeStrategy isn't intended to be sealed, so I'll have to look at what you mean by the types in its method not being able to be named externally.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
             .ok_or(Error::InvalidAmount)
     }
 
+    pub fn compute_balances<BalanceError>(
```
str4d, line 443: - Needs documentation.
- Document this addition in the changelog.
- This generic name is confusing because we _also_ import a `BalanceError` type. AFAICT we don't need it:
  ```suggestion
      pub fn compute_balances<S: FeeStrategy>(
  ```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+        &self,
+        fee_strategy: &impl FeeStrategy<Error = BalanceError>,
+    ) -> Result<TransactionBalances, BalanceError>
+    where
+        Error: From<BalanceError>,
+    {
```
str4d, line 449:
```suggestion
        fee_strategy: &S,
    ) -> Result<TransactionBalances, S::Error> {
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+    }
+
+    #[cfg(feature = "zfuture")]
+    pub fn compute_balances_zfuture<BalanceError>(
+        &self,
+        fee_strategy: &impl FutureFeeStrategy<Error = BalanceError>,
+    ) -> Result<TransactionBalances, BalanceError>
+    where
+        Error: From<BalanceError>,
+    {
```
str4d, line 465: - Needs documentation.
- Document this addition in the changelog.
- Ditto:

```suggestion
    pub fn compute_balances_zfuture<S: FutureFeeStrategy>(
        &self,
        fee_strategy: &S,
    ) -> Result<TransactionBalances, S::Error> {
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+        )
+    }
+
     /// Builds a transaction from the configured spends and outputs.
     ///
     /// Upon success, returns a tuple containing the final transaction, and the
     /// [`SaplingMetadata`] generated during the build process.
-    pub fn build(
-        mut self,
+    pub fn build<BalanceError>(
```
str4d, line 480: - Document this change in the changelog.
- Similarly to above, this should not take the error as the generic:

```suggestion
    pub fn build<S: FeeStrategy>(
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+        self,
         prover: &impl TxProver,
-    ) -> Result<(Transaction, SaplingMetadata), Error> {
+        balance_calculator: &impl FeeStrategy<Error = BalanceError>,
+    ) -> Result<(Transaction, SaplingMetadata), Error>
+    where
+        Error: From<BalanceError>,
```
str4d, line 486:
```suggestion
        balance_calculator: &S,
    ) -> Result<(Transaction, SaplingMetadata), Error>
    where
        Error: From<S::Error>,
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+    }
+
+    #[cfg(feature = "zfuture")]
+    pub fn build_zfuture<BalanceError>(
+        self,
+        prover: &impl TxProver,
+        balance_calculator: &impl FutureFeeStrategy<Error = BalanceError>,
+    ) -> Result<(Transaction, SaplingMetadata), Error>
+    where
+        Error: From<BalanceError>,
```
str4d, line 499: - Needs documentation.
- Document this addition in the changelog.
- Ditto:

```suggestion
    pub fn build_zfuture<S: FutureFeeStrategy>(
        self,
        prover: &impl TxProver,
        balance_calculator: &S,
    ) -> Result<(Transaction, SaplingMetadata), Error>
    where
        Error: From<S::Error>,
```

Aside: I have not checked, but maybe the following is also possible?

```suggestion
    pub fn build_zfuture(
        self,
        prover: &impl TxProver,
        balance_calculator: &impl FutureFeeStrategy<Error: Into<Error>>,
    ) -> Result<(Transaction, SaplingMetadata), Error>
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -306,14 +440,76 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+    }
+
+    fn build_internal<BalanceError>(
+        self,
+        prover: &impl TxProver,
+        balances: &TransactionBalances,
+    ) -> Result<(Transaction, SaplingMetadata), Error>
+    where
+        Error: From<BalanceError>,
+    {
```
str4d, line 512: I don't understand why this generic is present at all; it isn't used anywhere AFAICT.
```suggestion
    fn build_internal(
        self,
        prover: &impl TxProver,
        balances: &TransactionBalances,
    ) -> Result<(Transaction, SaplingMetadata), Error> {
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -650,8 +891,8 @@ mod tests {
         {
             let builder = Builder::new(TEST_NETWORK, tx_height);
             assert_eq!(
-                builder.build(&MockTxProver),
-                Err(Error::ChangeIsNegative(
+                builder.mock_build(),
+                Err(Error::InsufficientFunds(
                     (Amount::zero() - DEFAULT_FEE).unwrap()
                 ))
```
str4d, line 897: Bug: `Error::InsufficientFunds` has the inverse semantics to `Error::ChangeIsNegative`:
```suggestion
                Err(Error::InsufficientFunds(DEFAULT_FEE))
```
This and the subsequent equivalent changes should fix the CI test failure.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -674,8 +915,8 @@ mod tests {
                 )
                 .unwrap();
             assert_eq!(
-                builder.build(&MockTxProver),
-                Err(Error::ChangeIsNegative(
+                builder.mock_build(),
+                Err(Error::InsufficientFunds(
                     (Amount::from_i64(-50000).unwrap() - DEFAULT_FEE).unwrap()
```
str4d, line 920:
```suggestion
                    (Amount::from_u64(50000).unwrap() + DEFAULT_FEE).unwrap()
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -692,8 +933,8 @@ mod tests {
...
```
str4d, line 938:
```suggestion
                    (Amount::from_u64(50000).unwrap() + DEFAULT_FEE).unwrap()
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -734,8 +975,8 @@ mod tests {
                 )
                 .unwrap();
             assert_eq!(
-                builder.build(&MockTxProver),
-                Err(Error::ChangeIsNegative(Amount::from_i64(-1).unwrap()))
+                builder.mock_build(),
+                Err(Error::InsufficientFunds(Amount::from_i64(-1).unwrap()))
```
str4d, line 979:
```suggestion
                Err(Error::InsufficientFunds(Amount::from_u64(1).unwrap()))
```

str4d (changes requested):
Reviewed 6280ddebc556ef06064e603c2f04b4e43d24bd73.

zcash_primitives/src/transaction/builder.rs
nuttycom: Ah, yeah, I just changed this and didn't try the tests locally.

zcash_primitives/src/transaction/builder.rs
str4d: What I mean is that you have a public type in a private module, and it isn't publicly exported elsewhere in a public module. So outside of the crate, no one can `use zcash_primitives::path::to::Type` to import the types needed to implement the trait.

However, I've looked at the crate documentation again, and it seems that the `builder` submodules of the transaction components _are_ exposed publicly (I didn't think they were based on my initial look at the PR), so these types are nameable and `FeeStrategy` isn't sealed AFAICT.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -34,7 +38,7 @@ pub enum Error<NoteId> {
     AccountNotFound(AccountId),
 
     /// The amount specified exceeds the allowed range.
-    InvalidAmount,
+    BalanceError(BalanceError),
```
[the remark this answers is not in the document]
nuttycom: The `InvalidAmount` constructor was added since the last release.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -107,8 +120,179 @@ impl Progress {
...
+    /// Constructs a new balance from its constituent parts. Returns an error
+    /// on overflow or underflow of amount arithmetic.
+    pub fn new(
+        total_in: Amount,
+        change_required: Amount,
+        fee_required: Amount,
+    ) -> Result<Self, BalanceError> {
+        // check for overflow and underflow at construction.
+        (total_in - (change_required + fee_required).ok_or(BalanceError::Overflow)?)
+            .ok_or(BalanceError::Underflow)?;
```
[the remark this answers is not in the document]
nuttycom: My thinking here is that it may be necessary to compute balances for partially-constructed transactions; for example, one might supply outputs before supplying inputs, in which case the output balance could be negative. I don't think we want to constrain transaction balance at this point, since where we actually need to do that is in the builder itself.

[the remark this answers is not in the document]
nuttycom: I narrowed the scope of TransactionBalance to just represent the balance.

nuttycom:
[force-pushed(https://github.com/zcash/librustzcash/compare/6280dde..9496fc6118812fc8cc89d8b155d8b5b501348e43) to address review comments.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -34,7 +38,7 @@ pub enum Error<NoteId> {
     AccountNotFound(AccountId),
 
     /// The amount specified exceeds the allowed range.
```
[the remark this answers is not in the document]
str4d: Mmk, that wasn't clear from the type.

zcash_client_backend/src/data_api/error.rs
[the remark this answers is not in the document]
str4d: I hadn't seen it in the changelog which is why I assumed it had always been there. But it was added in https://github.com/zcash/librustzcash/pull/385, and I wasn't as stringent in my review of that PR about changelog updates.

zcash_client_backend/src/fees.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+/// The amount of change and fees required to make a transaction's inputs and
+/// outputs balance under a specific fee rule, as computed by a particular
+/// [`ChangeStrategy`] that is aware of that rule.
+pub struct TransactionBalance {
+    proposed_change: Vec<ChangeValue>,
+    fee_required: Amount,
+}
```
str4d, line 32: I like this approach; it's going to enable note splitting easily. I'm now wondering about note merging; `ChangeStrategy` assumes that all of the provided inputs will be part of the transaction (which makes sense given that it is focused on change strategy), so does this mean to have a variable number of inputs beyond what is necessary, we will need a third layer around `ChangeStrategy` and `FeeRule`? Or would it make more sense to do both note merging and note splitting in the same place? I guess the outer layer could equivalently implement that itself...

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -361,7 +361,42 @@ where
...
+            builder.sapling_inputs(),
+            builder.sapling_outputs(),
+        )
+        .map_err(|e| match e {
+            ChangeError::InsufficientFunds {
+                available,
+                required,
+            } => Error::InsufficientBalance(available, required),
+            ChangeError::StrategyError(e) => Error::BalanceError(e),
+        })?;
```
str4d, line 380: We do this match in two places, so we should add an `impl From<ChangeError> for Error`.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -361,7 +361,42 @@ where
...
+
+    for change_value in balance.proposed_change() {
+        match change_value {
+            ChangeValue::Sapling(amount) => {
+                builder
+                    .add_sapling_output(
+                        Some(dfvk.to_ovk(Scope::Internal)),
+                        dfvk.change_address().1,
+                        *amount,
```
str4d, line 389: Why does this need to be dereferenced?

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -510,15 +539,44 @@ where
...
+
+    let fee = balance.fee_required();
+    let mut total_out = Amount::zero();
+    for change_value in balance.proposed_change() {
+        total_out = (total_out + change_value.value())
+            .ok_or(Error::BalanceError(BalanceError::Overflow))?;
+        match change_value {
+            ChangeValue::Sapling(amount) => {
+                builder
+                    .add_sapling_output(Some(ovk), shielding_address.clone(), *amount, memo.clone())
```
str4d, line 570: Ditto.

str4d (changes requested):
Reviewed 9496fc6118812fc8cc89d8b155d8b5b501348e43.

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,55 @@ and this library adheres to Rust's notion of
...
+    - `Builder::sapling_outputs()`
+- `zcash_primitives::transaction::fees` a new module containing abstractions
+  and types related to fee calculations.
+  - `TransactionBalance` A type representing the unspent balance of transaction inputs,
+    allocated separately to change and fee portions.
+  - `FeeRule` a function that computes the fee required for a transaction given
+    inputs and outputs to the transaction.
+  - `Infalliable` an uninhabited error type for conditions where no error can occur.
+  - `FixedFeeRule` A `FeeRule` implementation that always computes the same fixed
+    fee irrespective of the inputs and outputs to the transaction being constructed.
```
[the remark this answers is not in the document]
nuttycom: I mean, okay, it's a trait. But it's really just a specially-named function type. That's all *any* trait with a single method is.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: `proposed_change` returns a slice, and so it's borrowed here.

zcash_client_backend/src/fees.rs
nuttycom: Your inference is correct: we *do* need a third layer! See #689

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Error handling changes *substantially* in #689 so let's wait to alter this.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: See above; `proposed_change` returns a slice.

zcash_client_backend/src/fees.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// supply the requested outputs and required fees, implementations should return
+    /// [`ChangeError::InsufficientFunds`].
+    fn compute_balance<P: consensus::Parameters>(
+        &self,
+        params: &P,
+        target_height: BlockHeight,
+        transparent_inputs: &[impl TransparentInput],
+        transparent_outputs: &[TxOut],
+        sapling_inputs: &[impl SaplingInput],
+        sapling_outputs: &[SaplingOutput],
```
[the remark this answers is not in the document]
nuttycom: The benefit of adding an `impl SaplingOutput` trait would be additional information hiding from the caller; I'm not convinced that such information hiding is particularly useful; the `TransparentInput` and `SaplingInput` traits are necessary because the structs that they expose information from are private to their modules.

zcash_primitives/CHANGELOG.md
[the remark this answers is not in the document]
nuttycom: `FeeRule` is mentioned elsewhere in the changelog, so I thought it was valuable to mention here; changelogs are generally intended to let users know what interesting things they should specifically be aware of so I think we should keep the documentation here. I'll change the wording relating to the trait though.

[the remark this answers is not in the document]
nuttycom: Oh, and `TransactionBalance` was actually moved to zcash_client_backend

zcash_primitives/src/transaction/fees.rs
```diff
@@ -0,0 +1,115 @@
...
+        sapling_inputs: &[impl SaplingInput],
+        sapling_outputs: &[SaplingOutput],
+        tze_inputs: &[impl TzeInput],
+        tze_outputs: &[TzeOut],
+    ) -> Result<Amount, Self::Error>;
+}
+
+/// An uninhabited error type used to indicate when an operation
+/// that returns a `Result` cannot fail.
+pub enum Infallible {}
```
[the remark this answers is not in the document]
nuttycom: Oh, nice. The bottom type has always been something that seemed like it was missing to me. :)

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -69,8 +76,18 @@ pub struct SpendDescriptionInfo {
...
+        // amount value.
+        Amount::from_u64(self.note.value).unwrap()
+    }
+}
+
+/// A struct containing the information required in order to construct a
+/// Sapling output to a transaction.
 #[derive(Clone)]
-struct SaplingOutput {
+pub struct SaplingOutput {
```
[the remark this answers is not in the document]
nuttycom: We'll need the amounts for change computation, and I don't think it's worth having two different traits, so let's just make a `SaplingOutput` trait and rename the struct to `SaplingOutputInfo`

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -213,7 +236,7 @@ impl Authorization for Unauthorized {
     type AuthSig = SpendDescriptionInfo;
 }
 
-impl<P: consensus::Parameters> SaplingBuilder<P> {
+impl<P> SaplingBuilder<P> {
```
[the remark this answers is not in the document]
nuttycom: This is an additional block of methods that don't depend upon the constraint on `P`.

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -60,6 +60,13 @@ impl fmt::Display for Error {
     }
 }
 
+/// A trait that provides a minimized view of a Sapling input suitable for use in
+/// fee calculation.
+pub trait SaplingInput {
```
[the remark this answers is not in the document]
nuttycom: Same comment here as above: since this is *only* relevant to tx under construction, and not anything in the protocol, I think it's better to keep this here and move `TzeInput` the other way.

nuttycom:
[updated](https://github.com/zcash/librustzcash/compare/9496fc6..c92d81b15d7adab6a460aa99eee4591e91c0aad1) to address requested changes.

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,51 @@ and this library adheres to Rust's notion of
...
+    - `Builder::sapling_outputs()`
+- `zcash_primitives::transaction::fees` a new module containing abstractions
+  and types related to fee calculations.
+  - `FeeRule` a trait that describes how to compute the fee required for a
+    transaction given inputs and outputs to the transaction.
+- Added to `zcash_primitives::transaction::components::sapling::builder`
+  - `SaplingInput` a trait that provides a minimized view of a Sapling input suitable
+    for use in change and fee computation.
+  - `SaplingOutput` a trait that provides a minimized view of a Sapling output suitable
+    for use in change and fee computation.
```
str4d, line 31: These have been renamed and moved to a new module; changelog needs updating.

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,51 @@ and this library adheres to Rust's notion of
...
+  - `SaplingBuilder::inputs` and `SaplingBuilder::outputs`: accessors for Sapling
+    builder state.
+- Added to `zcash_primitives::transaction::components::transparent::builder`
+  - `TransparentInput` a trait that provides a minimized view of a transparent input suitable
+    for use in fee computation.
```
str4d, line 36: This was renamed and moved to a new module; changelog needs updating. Also needs `OutputView` if both will be listed directly (but could also just say that the new modules were added for X purpose).

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,51 @@ and this library adheres to Rust's notion of
...
+  argument which is used to compute the fee for the transaction as part of the
+  build process.
+- `zcash_primitives::transaction::builder::Builder::{new, new_with_rng}` no
+  longer fixes the fee for transactions to 0.00001 ZEC; the builder instead
+  computes the fee using a `FeeRule` implementation at build time.
+
+### Removed
+- Removed from `zcash_primitives::transaction::builder::Builder`
+  - `Builder::{new_with_fee, new_with_rng_and_fee`} (use `Builder::{new, new_with_rng}` 
+    instead along with a `FeeRule` implementation passed to `Builder::build`.
```
str4d, line 50:
```suggestion
    instead along with a `FeeRule` implementation passed to `Builder::build`).
```

str4d (approved):
utACK 9c894ebf3eba4dc394e8a7ab1e555fb47c372937 modulo changelog fixes.

str4d (approved):
Re-utACK 28db1e324f620c811124ef6326d30bbd826bcfb1

PR #659, opened 2022-10-02
Track inputs sent to wallet-internal recipients

nuttycom:
This adds trial decryption with the internal IVK and correctly tracks payments sent to the wallet-internal address in the wallet database.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -11,15 +11,16 @@ use secrecy::SecretVec;
 use zcash_primitives::{
     block::BlockHash,
     consensus::BlockHeight,
+    legacy::TransparentAddress,
     memo::{Memo, MemoBytes},
     merkle_tree::{CommitmentTree, IncrementalWitness},
-    sapling::{Node, Nullifier},
+    sapling::{Node, Nullifier, PaymentAddress},
     transaction::{components::Amount, Transaction, TxId},
     zip32::{AccountId, ExtendedFullViewingKey},
 };
 
 use crate::{
-    address::{RecipientAddress, UnifiedAddress},
+    address::UnifiedAddress,
     decrypt::DecryptedOutput,
     keys::{UnifiedFullViewingKey, UnifiedSpendingKey},
     proto::compact_formats::CompactBlock,
@@ -27,10 +28,7 @@ use crate::{
 };
 
 #[cfg(feature = "transparent-inputs")]
-use {
-    crate::wallet::WalletTransparentOutput,
-    zcash_primitives::{legacy::TransparentAddress, transaction::components::OutPoint},
-};
+use {crate::wallet::WalletTransparentOutput, zcash_primitives::transaction::components::OutPoint};
 
 pub mod chain;
 pub mod error;
@@ -254,13 +252,6 @@ pub struct SentTransaction<'a> {
     pub utxos_spent: Vec<OutPoint>,
 }
 
-#[derive(Debug, Clone)]
-#[allow(clippy::large_enum_variant)]
-pub enum Recipient {
-    Address(RecipientAddress),
-    InternalAccount(AccountId),
-}
-
 /// A value pool to which the wallet supports sending transaction outputs.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum PoolType {
@@ -270,9 +261,19 @@ pub enum PoolType {
     Sapling,
 }
 
+/// A type that represents the recipient of a transaction output; a recipient address (and, for
+/// unified addresses, the pool to which the payment is sent) in the case of outgoing output, or an
+/// internal account ID and the pool to which funds were sent in the case of a wallet-internal
+/// output.
+#[derive(Debug, Clone)]
+pub enum Recipient {
+    Transparent(TransparentAddress),
+    Sapling(PaymentAddress),
+    Unified(UnifiedAddress, PoolType),
+    InternalAccount(AccountId, PoolType),
+}
+
 pub struct SentTransactionOutput {
-    /// The value in which the output has been created
-    pub output_pool: PoolType,
     /// The index within the transaction that contains the recipient output.
     ///
     /// - If `recipient_address` is a Sapling address, this is an index into the Sapling
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -248,13 +248,19 @@ pub struct SentTransaction<'a> {
     pub tx: &'a Transaction,
     pub created: time::OffsetDateTime,
     pub account: AccountId,
-    pub outputs: Vec<SentTransactionOutput<'a>>,
+    pub outputs: Vec<SentTransactionOutput>,
     pub fee_amount: Amount,
     #[cfg(feature = "transparent-inputs")]
     pub utxos_spent: Vec<OutPoint>,
 }
 
-pub struct SentTransactionOutput<'a> {
+#[derive(Debug, Clone)]
+pub enum Recipient {
+    Address(RecipientAddress),
+    InternalAccount(AccountId),
+}
+
+pub struct SentTransactionOutput {
     /// The index within the transaction that contains the recipient output.
     ///
     /// - If `recipient_address` is a Sapling address, this is an index into the Sapling
@@ -262,7 +268,7 @@ pub struct SentTransactionOutput<'a> {
     /// - If `recipient_address` is a transparent address, this is an index into the
     ///   transparent outputs of the transaction.
     pub output_index: usize,
-    pub recipient_address: &'a RecipientAddress,
+    pub recipient: Recipient,
     pub value: Amount,
     pub memo: Option<MemoBytes>,
 }
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -255,12 +255,24 @@ pub struct SentTransaction<'a> {
 }
 
 #[derive(Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
 pub enum Recipient {
     Address(RecipientAddress),
     InternalAccount(AccountId),
 }
 
+/// A value pool to which the wallet supports sending transaction outputs.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum PoolType {
+    /// The transparent value pool
+    Transparent,
+    /// The Sapling value pool
+    Sapling,
+}
+
 pub struct SentTransactionOutput {
+    /// The value in which the output has been created
+    pub output_pool: PoolType,
     /// The index within the transaction that contains the recipient output.
     ///
     /// - If `recipient_address` is a Sapling address, this is an index into the Sapling
@@ -268,8 +280,12 @@ pub struct SentTransactionOutput {
     /// - If `recipient_address` is a transparent address, this is an index into the
     ///   transparent outputs of the transaction.
     pub output_index: usize,
+    /// The recipient address of the transaction, or the account
+    /// id for wallet-internal transactions.
     pub recipient: Recipient,
+    /// The value of the newly created output
     pub value: Amount,
+    /// The memo that was attached to the output, if any
     pub memo: Option<MemoBytes>,
 }
 
--- a/zcash_client_backend/src/decrypt.rs
+++ b/zcash_client_backend/src/decrypt.rs
@@ -10,10 +10,21 @@ use zcash_primitives::{
         Note, PaymentAddress,
     },
     transaction::Transaction,
-    zip32::AccountId,
+    zip32::{AccountId, Scope},
 };
 
 use crate::keys::UnifiedFullViewingKey;
+
+/// An enumeration of the possible relationships a TXO can have to the wallet.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum TransferType {
+    /// The transfer was received on one of the wallet's external addresses.
+    Incoming,
+    /// The transfer was received on one of the wallet's internal-only addresses.
+    WalletInternal,
+    /// The transfer was decrypted using one of the wallet's outgoing viewing keys.
+    Outgoing,
+}
 
 /// A decrypted shielded output.
 pub struct DecryptedOutput {
@@ -33,7 +44,7 @@ pub struct DecryptedOutput {
     /// this is a logical output of the transaction.
     ///
     /// [`OutgoingViewingKey`]: zcash_primitives::keys::OutgoingViewingKey
-    pub outgoing: bool,
+    pub transfer_type: TransferType,
 }
 
 /// Scans a [`Transaction`] for any information that can be decrypted by the set of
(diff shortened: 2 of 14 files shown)


zcash_client_backend/src/decrypt.rs
```diff
@@ -15,6 +16,17 @@ use zcash_primitives::{
 
 use crate::keys::UnifiedFullViewingKey;
 
+/// An enumeration of the possible relationships a TXO can have to the wallet.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum TransferType {
```
str4d, line 21: Needs to be mentioned in the changelog as an addition.

zcash_client_backend/src/decrypt.rs
```diff
@@ -15,6 +16,17 @@ use zcash_primitives::{
...
+    /// The transfer was received on one of the wallet's external addresses.
+    Incoming,
+    /// The transfer was received on one of the wallet's internal-only addresses.
+    WalletInternal,
+    /// The transfer was decrypted using one of the wallet's outgoing viewing keys.
+    Outgoing,
```
str4d, line 27: Either these cases need more documentation, or the enum as a whole needs more documentation. In particular, it is _almost never_ true that a wallet will detect an outgoing transaction by decrypting with an OVK; instead a change output will be decrypted with an IVK. There are potential edge cases here as well when a transaction is e.g. sending funds between two accounts in a wallet, or involves an explicit send to a wallet's external address as well as a wallet-external address, which cannot be represented as a single instance of this enum.

zcash_client_backend/src/decrypt.rs
```diff
@@ -33,7 +45,7 @@ pub struct DecryptedOutput {
     /// this is a logical output of the transaction.
     ///
     /// [`OutgoingViewingKey`]: zcash_primitives::keys::OutgoingViewingKey
-    pub outgoing: bool,
+    pub transfer_type: TransferType,
```
str4d, line 48: - This is a change to a public type. Document it in the changelog.
- Per my above comment, a single instance of `TransferType` is insufficient to fully describe the possible transaction types. This needs to either be a set or a bitflag, or `TransferType` needs additional cases.

zcash_client_backend/src/decrypt.rs
```diff
@@ -49,27 +61,37 @@ pub fn decrypt_transaction<P: consensus::Parameters>(
...
+                        try_sapling_note_decryption(params, height, &ivk_external, output)
+                            .map(|ret| (ret, TransferType::Incoming))
+                            .or_else(|| {
+                                try_sapling_note_decryption(params, height, &ivk_internal, output)
+                                    .map(|ret| (ret, TransferType::WalletInternal))
+                            })
+                            .or_else(|| {
+                                try_sapling_output_recovery(params, height, &ovk, output)
+                                    .map(|ret| (ret, TransferType::Outgoing))
+                            });
```
str4d: This looks vaguely like some similar code in `zcashd`. We should document the logic here, and try to keep it in line with what we do in `zcashd`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1232,26 +1231,21 @@ pub fn put_sent_utxo<'a, P: consensus::Parameters>(
...
-)]
-pub fn insert_sent_note<'a, P: consensus::Parameters>(
+pub(crate) fn insert_sent_note<'a, P: consensus::Parameters>(
     stmts: &mut DataConnStmtCache<'a, P>,
     tx_ref: i64,
     output_index: usize,
-    account: AccountId,
-    to: &PaymentAddress,
+    from_account: AccountId,
+    to: &Recipient,
```
str4d, line 1239: The type here is invalid, as it allows a transparent address to be passed in.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1170,14 +1169,14 @@ pub fn put_sent_note<'a, P: consensus::Parameters>(
     tx_ref: i64,
     output_index: usize,
     account: AccountId,
-    to: &PaymentAddress,
+    to: &Recipient,
```
str4d, line 1172: - The type here is invalid, as this allows a transparent address to be passed in.
- This is a change to a public deprecated API, making the deprecation meaningless. We should not change the public deprecated API, and instead either map to the new API internally (as was done below for `put_sent_utxo`) or remove the deprecated API.

zcash_client_sqlite/src/prepared.rs
```diff
@@ -500,78 +579,6 @@ impl<'a, P> DataConnStmtCache<'a, P> {
...
-    /// Inserts a sent note into the wallet database.
-    ///
-    /// `output_index` is the index within the transaction that contains the recipient output:
-    ///
-    /// - If `to` is a Sapling address, this is an index into the Sapling outputs of the
-    ///   transaction.
-    /// - If `to` is a transparent address, this is an index into the transparent outputs of
-    ///   the transaction.
-    #[allow(clippy::too_many_arguments)]
-    pub(crate) fn stmt_insert_sent_note(
```
str4d, line 512: Why were these moved? It makes review significantly harder (for me on my 13" laptop screen that can't show two copies of this PR side-by-side to check what is moved and what is altered without line wrapping).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -122,8 +122,13 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
     seed: Option<SecretVec<u8>>,
     target_migration: Option<Uuid>,
 ) -> Result<(), MigratorError<WalletMigrationError>> {
+    // Turn off foreign keys, and ensure that table replacement/modification
+    // does not break views
     wdb.conn
-        .execute("PRAGMA foreign_keys = OFF", [])
+        .execute_batch(
+            "PRAGMA foreign_keys = OFF;
+                        PRAGMA legacy_alter_table = TRUE;",
```
str4d, line 130:
```suggestion
            "PRAGMA foreign_keys = OFF;
            PRAGMA legacy_alter_table = TRUE;",
```

I also get suspicious when flags named `legacy_*` are being enabled; I'd prefer some more documentation explaining why this is necessary, and why it won't break with some future version of SQLite.

zcash_client_sqlite/src/wallet/init/migrations/sent_notes_to_internal.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch("ALTER TABLE sent_notes ADD COLUMN to_account INTEGER;")?;
+
+        // `to_account` should be null for all migrated rows, since internal addresses
+        // have not been used for change or shielding prior to this migration.
```
str4d, line 48: This comment is confusing. `to_account` is guaranteed to be null because we just added it, and we'd have hit a database error if the column already existed.

I also do not understand why we are altering the table to add a `to_account` column in the first place, given that we then immediately replace that table with a new one. We should instead check that the column does not exist in `sent_notes` (to replace the error that we'd get above) and then just use the new table creation.

zcash_client_sqlite/src/lib.rs
```diff
@@ -611,35 +622,26 @@ impl<'a, P: consensus::Parameters> WalletWrite for DataConnStmtCache<'a, P> {
...
+                        wallet::insert_sent_utxo(
+                            up,
+                            tx_ref,
+                            output.output_index,
+                            sent_tx.account,
+                            addr,
+                            output.value,
+                        )?
+                    }
+                    shielded_recipient => wallet::insert_sent_note(
```
str4d, line 636: This incorrectly resolves the above TODO and introduces a bug. `wallet::insert_sent_note` internally hard-codes Sapling as the pool, which made sense because it explicitly took in a `sapling::PaymentAddress`. This change coerces all shielded output types to Sapling, which will cause corruption and (in the case of mixed transactions) data loss.

The correct approach is provide what the TODO says: more information about the output index. We need to pass through the output pool type all thr way through from where `sent_tx.outputs` is populated. Or alternatively, map the UA to its Sapling case, preserving the approach we were taking before (deferring Orchard support to a later PR).

zcash_client_sqlite/src/prepared.rs
```diff
@@ -331,6 +336,80 @@ impl<'a, P> DataConnStmtCache<'a, P> {
...
+    /// Inserts a sent note into the wallet database.
+    ///
+    /// `output_index` is the index within the transaction that contains the recipient output:
+    ///
+    /// - If `to` is a Sapling address, this is an index into the Sapling outputs of the
+    ///   transaction.
+    /// - If `to` is a transparent address, this is an index into the transparent outputs of
+    ///   the transaction.
+    /// - If `to` is an internal account, this is an index into the Sapling outputs of the
+    ///   transaction.
```
str4d, line 348: This does not mention what happens when `to` is a UA, which is indicative of the bug I identified above.

str4d (changes requested):
Reviewed 52e20ca195614ac7dcb15bf71b2cc53be20cbd66.

zcash_client_sqlite/src/lib.rs
nuttycom: I don't think this is correct. We need to be tracking the entire unified address to which we were requested to send the transaction; otherwise we're throwing away that critical information and can't retrieve it in the wallet history. This change was intentional. Right now, of course, we can only send to the Sapling pool, so that's why I didn't propagate the pool information through, but we need to retain the full UA somewhere.

zcash_client_sqlite/src/wallet/init/migrations/sent_notes_to_internal.rs
```diff
@@ -0,0 +1,88 @@
...
+        "Adds an identifier for the account that received an internal note to the sent_notes table"
+    }
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch("ALTER TABLE sent_notes ADD COLUMN to_account INTEGER;")?;
+
```
nuttycom, line 46:
```suggestion
```

zcash_client_sqlite/src/prepared.rs
nuttycom: These were moved because they now require the `consensus::Parameters` constraint on `P` where previously they did not.

zcash_client_sqlite/src/wallet.rs
nuttycom: I guess that this was a copy-paste error in https://github.com/zcash/librustzcash/commit/cb6a99384059b0bacff56a0e1e4de498b0a610bd

nuttycom: Oh, wait. No. It's that the `sent_notes` table is actually also being used to track UTXOs, and the intermediate APIs are just obscuring this fact: https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet.rs#L1230-L1288

So the correct thing, as of right now, is to get rid of basically all of this and instead maybe rename the `sent_notes` table to something that more accurately reflects its role, and I guess re-create the views so that they're showing the right information.

nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/659/commits/179e22b76d8a34eff7cfc995683dcf34692816ac

zcash_client_sqlite/src/wallet.rs
nuttycom: It's actually not; see https://github.com/zcash/librustzcash/pull/659/commits/179e22b76d8a34eff7cfc995683dcf34692816ac

zcash_client_sqlite/src/lib.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/659/commits/179e22b76d8a34eff7cfc995683dcf34692816ac by propagating the pool information in from where it's definitively known.

zcash_client_sqlite/src/prepared.rs
nuttycom: Not a bug.

zcash_client_backend/src/data_api.rs
```diff
@@ -248,21 +248,27 @@ pub struct SentTransaction<'a> {
     pub tx: &'a Transaction,
     pub created: time::OffsetDateTime,
     pub account: AccountId,
-    pub outputs: Vec<SentTransactionOutput<'a>>,
+    pub outputs: Vec<SentTransactionOutput>,
```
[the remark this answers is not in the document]
nuttycom: The `SentTransactionOutput` is new as of this release, but it had been only incidentally mentioned in the `Changed` section of the changelog. Adding it to `Added`.

zcash_client_backend/src/data_api.rs
```diff
@@ -248,21 +248,27 @@ pub struct SentTransaction<'a> {
...
 }
 
-pub struct SentTransactionOutput<'a> {
+#[derive(Debug, Clone)]
+pub enum Recipient {
+    Address(RecipientAddress),
+    InternalAccount(AccountId),
+}
+
+pub struct SentTransactionOutput {
```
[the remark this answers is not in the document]
nuttycom: This type is new.

zcash_client_sqlite/src/prepared.rs
str4d: It _was_ a bug at the time, but it looks like the most recent commit fixes it.

zcash_client_sqlite/src/wallet.rs
str4d: It _was_ invalid at the time (it's not possible to have a transparent recipient in the Sapling pool), but it looks like the most recent commit fixes it.

zcash_client_sqlite/src/prepared.rs
```diff
@@ -346,6 +354,103 @@ impl<'a, P> DataConnStmtCache<'a, P> {
 }
 
 impl<'a, P: consensus::Parameters> DataConnStmtCache<'a, P> {
+    /// Inserts a sent note into the wallet database.
+    ///
+    /// `output_index` is the index within the transaction that contains the recipient output:
+    ///
+    /// - If `to` is a Sapling address, this is an index into the Sapling outputs of the
+    ///   transaction.
```
str4d, line 364: This does not mention UAs.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -131,7 +134,6 @@ and this library adheres to Rust's notion of
     - `insert_witness`
     - `prune_witnesses`
     - `update_expired_notes`
-    - `put_sent_note`
     - `put_sent_utxo`
     - `insert_sent_note`
     - `insert_sent_utxo`
```
```suggestion
```

str4d (changes requested):
Reviewed a8df4c2587682166575aaa7a167032ab9b815db1

str4d (approved):
utACK e666e692301f7fde2e873d04c84f95e3c24a24a0.

PR #660, opened 2022-10-02
Upgrade rusqlite to version 0.25

nuttycom:


str4d (approved):
utACK f7aa7b2c84ea277ad299ff56e2dfcb03c6e5b89f. Merging this will block a release of the `zcash_client_sqlite` crate, which we had hoped to do for the upcoming SDK releases, so we should take that into consideration.

nuttycom:
> utACK [f7aa7b2](https://github.com/zcash/librustzcash/commit/f7aa7b2c84ea277ad299ff56e2dfcb03c6e5b89f). Merging this will block a release of the `zcash_client_sqlite` crate, which we had hoped to do for the upcoming SDK releases, so we should take that into consideration.

Why will that necessarily block the release?

PR #665, opened 2022-10-05
Add more information to v_transactions, v_sent_tx, and v_received_tx

nuttycom:
This adds sent and received note count information, transaction fees, account information, and makes the information returned about sent notes and received notes consistent with one another.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -114,13 +114,13 @@ pub fn init_wallet_db<P: consensus::Parameters + 'static>(
     wdb: &mut WalletDb<P>,
     seed: Option<SecretVec<u8>>,
 ) -> Result<(), MigratorError<WalletMigrationError>> {
-    init_wallet_db_internal(wdb, seed, None)
+    init_wallet_db_internal(wdb, seed, &[])
 }
 
 fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
     wdb: &mut WalletDb<P>,
     seed: Option<SecretVec<u8>>,
-    target_migration: Option<Uuid>,
+    target_migrations: &[Uuid],
 ) -> Result<(), MigratorError<WalletMigrationError>> {
     // Turn off foreign keys, and ensure that table replacement/modification
     // does not break views
@@ -137,7 +137,13 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
     migrator
         .register_multiple(migrations::all_migrations(&wdb.params, seed))
         .expect("Wallet migration registration should have been successful.");
-    migrator.up(target_migration)?;
+    if target_migrations.is_empty() {
+        migrator.up(None)?;
+    } else {
+        for target_migration in target_migrations {
+            migrator.up(Some(*target_migration))?;
+        }
+    }
     wdb.conn
         .execute("PRAGMA foreign_keys = ON", [])
         .map_err(|e| MigratorError::Adapter(WalletMigrationError::from(e)))?;
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -454,7 +454,7 @@ mod tests {
                    notes.raw,
                    SUM(notes.value) + MAX(notes.fee) AS net_value,
                    MAX(notes.fee)                    AS fee_paid,
-                   SUM(notes.is_wallet_internal) > 0 AS is_wallet_internal,
+                   SUM(notes.sent_count) == 0        AS is_wallet_internal,
                    SUM(notes.is_change) > 0          AS has_change,
                    SUM(notes.sent_count)             AS sent_note_count,
                    SUM(notes.received_count)         AS received_note_count,
@@ -472,7 +472,7 @@ mod tests {
                             WHEN received_notes.is_change THEN 0
                             ELSE value
                        END AS value,
-                       0                             AS is_wallet_internal,
+                       0                             AS sent_count,
                        CASE
                             WHEN received_notes.is_change THEN 1
                             ELSE 0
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -186,17 +196,25 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+            "CREATE VIEW v_transactions AS
+            SELECT notes.id_tx,
+                   notes.mined_height,
+                   notes.tx_index,
+                   notes.txid,
+                   notes.expiry_height,
+                   notes.raw,
+                   SUM(notes.value) + MAX(notes.fee) AS net_value,
+                   MAX(notes.fee)                    AS fee_paid,
+                   SUM(notes.is_wallet_internal) > 0 AS is_wallet_internal,
```
str4d, line 212: `is_wallet_internal` is maybe misleading at a transaction level; this will be set to `true` for a regular user-created transaction that has a change output. If that's the desired effect then this should be renamed to `has_wallet_internal`. However, I think the desired check here is that _every_ output is wallet-internal, so this should probably be: 
```suggestion
                   SUM(notes.is_not_wallet_internal) == 0 AS is_wallet_internal,
```

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -225,15 +252,22 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                        transactions.raw              AS raw,
                        transactions.fee              AS fee,
                        -sent_notes.value             AS value,
-                       false                         AS is_change,
+                       CASE
+                           WHEN sent_notes.from_account = sent_notes.to_account THEN 1
+                           ELSE 0
+                       END AS is_wallet_internal,
```
str4d, line 258:
```suggestion
                       CASE
                           WHEN sent_notes.from_account = sent_notes.to_account THEN 0
                           ELSE 1
                       END AS is_not_wallet_internal,
```

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -225,15 +252,22 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+                       0                             AS is_change,
+                       0                             AS received_count,
+                       1                             AS sent_count,
```
str4d, line 261: Now that the `sent_notes` table is also storing change / wallet-internal outputs, this is going to count these outputs as part of `sent_note_count`. But I presumed that `sent_note_count` is intended to be a count of the number of logical (e.g. from a payment request) recipients in the transaction? If that's the case, then this could be replaced by `is_not_wallet_internal` above (or separately count the number of change / wallet-internal notes).

str4d (commented):
Reviewed e6f039d0f9c35c69e1dc123b3e17e226c6d8cb17.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
nuttycom, line 212:
```suggestion
                   SUM(notes.is_not_wallet_internal) == 0 AS is_wallet_internal,
```

nuttycom: Good suggestion; I did this as `is_outgoing` instead.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
nuttycom: Actually, we can just change the definition of `sent_count` and fix both these issues at once.

str4d (approved):
utACK 5a2f6595947c5f3b82ef2176667819e3dc5a5cdc

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -504,29 +534,31 @@ mod tests {
                           ON transactions.id_tx = received_notes.tx
                    LEFT JOIN blocks
                           ON transactions.block = blocks.height
-            GROUP BY received_notes.tx",
+            GROUP BY received_notes.tx, received_notes.account",
+            // v_tx_received
```
ccjernigan, line 538: Is this comment a typo?

ccjernigan (commented):
Can a single transaction have multiple accounts?  And should be account number be included in the v_transactions table?  I'm asking because we have account ID in the received and sent views, but not the v_transactions view.  I don't know whether that might have future usage or implications for our UX.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -16,6 +16,15 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
 ) -> Vec<Box<dyn RusqliteMigration<Error = WalletMigrationError>>> {
+    //      initial_setup
+    //      /           \
+    // utxos_table     ufvk_support ----------
+    //      \                \                \
+    //       \         addresses_table   sent_notes_to_internal
+    //        \              /                /
+    //        add_utxo_account               /
+    //                       \              /
+    //                    add_transaction_views
```
daira, line 27: This is so helpful, thankyou!

nuttycom:
> Can a single transaction have multiple accounts? And should be account number be included in the v_transactions table? I'm asking because we have account ID in the received and sent views, but not the v_transactions view. I don't know whether that might have future usage or implications for our UX.

Yes, of course; two different received notes in the same transaction could be received  by addresses associated with different accounts, and in the future we may want to support support account consolidation, i.e. send all your funds out of the wallet irrespective of what account they were received on. `v_transactions` shouldn't have account information associated with it, unless we want to have multiple records associated with each transaction, grouped by account. That would be fine if we wanted to do that; I would actually prefer a more account-centric approach at the API level (which we've been moving toward anyway) and since these views are effectively another public API, we should consider that. But it does make things a bit less straightforward from a UX perspective, because you have to make sure that e.g. transactions that go from one account in the wallet to another are properly represented.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Oh, yup.

PR #666, opened 2022-10-05
zcash_proofs: Temporarily fix params download

upbqdn:
The implementation of `io:Read` for `ResponseLazyReader` used to return
only one byte regardless of the size of the provided `buf`, which
degraded the performance of loading the response.

This change makes use of the provided `buf`. On my machine with Rust
1.64, the downloading speed went up from ~100 KiB/s to ~2 MiB/s.
However, as before this change, the process still uses 100% of a single
CPU thread when downloading the response. The most likely reason is that
we still manually read each byte of the response. I can download the
same data with ~ 50 MiB/s on my machine using a different HTTP client.

This fix seems to be sufficient for now.

### DIFF
--- a/zcash_proofs/src/downloadreader.rs
+++ b/zcash_proofs/src/downloadreader.rs
@@ -44,25 +44,30 @@ impl io::Read for ResponseLazyReader {
 
                 // Read from the response
                 Response(response) => {
-                    // minreq has a very limited lazy reading interface.
-                    match &mut response.next() {
-                        // Read one byte into the buffer.
-                        // We ignore the expected length, because we have no way of telling the BufReader.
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK

nuttycom:
Clippy lints need fixing.

upbqdn:
> Clippy lints need fixing.

It looks like `codecov` is reporting untested changes. I didn't figure out how I could fix those, unfortunately.

str4d:
@upbqdn The clippy lint CI check is also failing:

image

upbqdn:
Thanks, I overlooked that check.

daira (approved):
utACK

PR #667, opened 2022-10-06
Allow spent UTXOs to be absent in migrations when adding tx fee.

nuttycom:
The previous approach to UTXO handling involved UTXO data being deleted from the wallet after the relevant UTXOs had been spent. However, this means that we can no longer accurately compute transaction fees for the transactions spending those UTXOs. However, since the transaction fee has been a constant 1000 ZAT for the life of any wallet where this issue might occur, we can just ignore those cases and assume the 1000 ZAT fee.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
@@ -8,7 +8,10 @@ use uuid::Uuid;
 
 use zcash_primitives::{
     consensus::{self, BlockHeight, BranchId},
-    transaction::{components::amount::Amount, Transaction},
+    transaction::{
+        components::amount::{Amount, BalanceError},
+        Transaction,
+    },
 };
 
 use super::{ufvk_support, utxos_table};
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -81,27 +103,40 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
-                stmt_set_fee.execute([i64::from(fee_paid), id_tx])?;
+                });
+
+                match fee_paid {
+                    Ok(fee_paid) => {
+                        stmt_set_fee.execute([i64::from(fee_paid), id_tx])?;
+                    }
+                    Err(FeeError::UtxoNotFound) => {
+                        // In the cases where we threw away UTXO data and so can no longer
+                        // determine the fee, use the legacy 1000 ZAT default fee.
```
nuttycom, line 127:
```suggestion
                        // determine the fee, use the legacy 1000 ZAT default fee.
                        // The 1000 ZAT fee was introduced in commit 53ebc378c, on Nov 16 19:59:28 2020.
                        // The autoshielding-poc branch was first introduced in commit cff457ff15af, on 
                        // Dec 22 11:10:13 2020 so there is no case where a wallet could have made an
                        // autoshielding transaction using the old 10000 ZAT fee.
```

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -81,27 +103,40 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+                        stmt_set_fee.execute([1000, id_tx])?;
```
nuttycom, line 128:
```suggestion
                        // The fee and net value will end up being null in the transactions view.
```

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -45,6 +48,25 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
     type Error = WalletMigrationError;
 
     fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        enum FeeError {
+            DbError(rusqlite::Error),
+            UtxoNotFound,
+            BalanceError(BalanceError),
```
str4d, line 54:
```suggestion
            Db(rusqlite::Error),
            UtxoNotFound,
            Balance(BalanceError),
```

str4d (approved):
utACK modulo Rust naming conventions.

str4d (approved):
utACK

PR #668, opened 2022-10-06
Implement ZIP 302 structured memos

### LINKED ISSUE #2234: Implement ZIP 302 structured memos
Once https://github.com/zcash/zips/issues/1208 is finalised, we should include it in the Rust crates.

str4d:
Implements https://github.com/zcash/zips/pull/638.

Closes #2234.
Closes COR-1121.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5966,6 +5966,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "wasabi_leb128"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1ef95b2fc26dab4b706f4763b297aa7277a1660e83450fa39e6832e64aeeb637"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "wasi"
 version = "0.11.0+wasi-snapshot-preview1"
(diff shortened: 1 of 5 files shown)


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/4edf9fef61b6560e32a3f5c81e49aa4c1bccd712..e7a12ea96a8c05e38e3bbfb61eb374ba9df3309b) to fix a bug and CI lints.

zcash_primitives/src/memo/structured.rs
```diff
@@ -0,0 +1,325 @@
...
+use wasabi_leb128::{ReadLeb128, WriteLeb128};
+
+use super::{Error, TextMemo};
+use crate::sapling::PaymentAddress;
+
+/// A payload within a [`StructuredMemo`].
+#[derive(Clone, Debug, PartialEq)]
+pub enum Payload {
+    /// A Sapling return address.
+    ReturnAddress(PaymentAddress),
```
str4d, line 12: IIRC I copied this type code from some existing implementation of structured memos. It was Sapling-specific then, but we can easily expand its definition to be any valid recipient address encoding.

Note that this uses a byte encoding rather than a string encoding for more efficient packing, so we may need to consider how this interacts with Unified Addresses (which now exist).

PR #672, opened 2022-10-13
Remove `received_by_account` field from WalletTransparentOutput

nuttycom:
Due to how the wallets retrieved unspent transparent outputs from the light wallet server, the account associated with a particular UTXO may not be known by the light wallet. Instead of requiring the caller to perform a separate lookup and match the address of the received UTXO with a known account, it's simpler to perform this lookup internally at the time of insertion or update.

In order to make this operation more efficient, the `addresses_table` migration is modified to add a column containing the transparent receiver so that they may be used in the joins in the UTXO insert and update operations.

### DIFF
--- a/zcash_client_sqlite/src/prepared.rs
+++ b/zcash_client_sqlite/src/prepared.rs
@@ -67,6 +67,7 @@ impl<'a> InsertAddress<'a> {
         mut diversifier_index: DiversifierIndex,
         address: &UnifiedAddress,
     ) -> Result<(), rusqlite::Error> {
+        // the diversifier index is stored in big-endian order to allow sorting
         diversifier_index.0.reverse();
         self.stmt.execute(named_params![
             ":account": &u32::from(account),
--- a/zcash_client_sqlite/src/prepared.rs
+++ b/zcash_client_sqlite/src/prepared.rs
@@ -33,6 +33,52 @@ use {
     zcash_primitives::transaction::components::transparent::OutPoint,
 };
 
+pub(crate) struct InsertAddress<'a> {
+    stmt: Statement<'a>,
+}
+
+impl<'a> InsertAddress<'a> {
+    pub(crate) fn new(conn: &'a rusqlite::Connection) -> Result<Self, rusqlite::Error> {
+        Ok(InsertAddress {
+            stmt: conn.prepare(
+                "INSERT INTO addresses (
+                    account,
+                    diversifier_index_be,
+                    address,
+                    cached_transparent_receiver_address
+                )
+                VALUES (
+                    :account,
+                    :diversifier_index_be,
+                    :address,
+                    :cached_transparent_receiver_address
+                )",
+            )?,
+        })
+    }
+
+    /// Adds the given address and diversifier index to the addresses table.
+    ///
+    /// Returns the database row for the newly-inserted address.
+    pub(crate) fn execute<P: consensus::Parameters>(
+        &mut self,
+        params: &P,
+        account: AccountId,
+        mut diversifier_index: DiversifierIndex,
+        address: &UnifiedAddress,
+    ) -> Result<(), rusqlite::Error> {
+        diversifier_index.0.reverse();
+        self.stmt.execute(named_params![
+            ":account": &u32::from(account),
+            ":diversifier_index_be": &&diversifier_index.0[..],
+            ":address": &address.encode(params),
+            ":cached_transparent_receiver_address": &address.transparent().map(|r| r.encode(params)),
+        ])?;
+
+        Ok(())
+    }
+}
+
 /// The primary type used to implement [`WalletWrite`] for the SQLite database.
 ///
 /// A data structure that stores the SQLite prepared statements that are
@@ -70,7 +116,7 @@ pub struct DataConnStmtCache<'a, P> {
     stmt_prune_witnesses: Statement<'a>,
     stmt_update_expired: Statement<'a>,
 
-    stmt_insert_address: Statement<'a>,
+    stmt_insert_address: InsertAddress<'a>,
 }
 
 impl<'a, P> DataConnStmtCache<'a, P> {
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
```diff
@@ -52,6 +52,8 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                 account INTEGER NOT NULL,
                 diversifier_index_be BLOB NOT NULL,
                 address TEXT NOT NULL,
+                sapling_receiver TEXT,
```
str4d, line 55: We can't store Orchard receivers as `TEXT`, so we shouldn't store Sapling this way either. But as this column is unused, we should just remove it:
```suggestion
```

zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
```diff
@@ -52,6 +52,8 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+                transparent_receiver TEXT,
```
str4d, line 56: This is only for caching purposes, and should not be relied on externally. Name the column to indicate this:
```suggestion
                cached_transparent_receiver TEXT,
```

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/b7475ddfca95563df6bc2239d382e62ed08bd076..f1f9465) to make clear that the transparent receiver column is only a cache.

zcash_client_sqlite/src/prepared.rs
```diff
@@ -119,22 +165,26 @@ impl<'a, P> DataConnStmtCache<'a, P> {
...
                         value_zat, height)
-                    VALUES (
-                        :received_by_account, :address,
+                    SELECT
+                        addresses.account, :address,
                         :prevout_txid, :prevout_idx, :script,
-                        :value_zat, :height)
+                        :value_zat, :height
+                    FROM addresses
+                    WHERE addresses.cached_transparent_receiver_address = :address
```
str4d, line 173: If `:address` does not exist in `addresses`, what happens? `utxos.received_by_account` is `NOT NULL`, so that would cause an error, but does it reach that point if the `SELECT` returns nothing due to the `WHERE` clause filtering everything out? It would be bad if this silently failed to insert.

zcash_client_sqlite/src/prepared.rs
```diff
@@ -119,22 +165,26 @@ impl<'a, P> DataConnStmtCache<'a, P> {
...
-                    SET received_by_account = :received_by_account,
+                    SET received_by_account = addresses.account,
                         height = :height,
                         address = :address,
                         script = :script,
                         value_zat = :value_zat
+                    FROM addresses
                     WHERE prevout_txid = :prevout_txid
                       AND prevout_idx = :prevout_idx
+                      AND addresses.cached_transparent_receiver_address = :address
```
str4d, line 187: I believe this `UPDATE` will silently fail if `:address` is not present in `addresses`?

zcash_client_sqlite/src/prepared.rs
```diff
@@ -516,19 +561,17 @@ impl<'a, P: consensus::Parameters> DataConnStmtCache<'a, P> {
...
+    ) -> Result<(), SqliteClientError> {
+        self.stmt_insert_address.execute(
+            &self.wallet_db.params,
+            account,
+            diversifier_index,
+            address,
+        )?;
 
-        Ok(self.wallet_db.conn.last_insert_rowid())
+        Ok(())
```
str4d, line 574: I confirmed that the return value of this function wasn't being used for anything.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -299,8 +299,6 @@ mod tests {
     use std::collections::HashMap;
     use tempfile::NamedTempFile;
 
-    use zcash_address::test_vectors;
```
str4d, line 302: I don't know why all these import changes have been made.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -929,8 +921,7 @@ pub(crate) fn get_unspent_transparent_outputs<P: consensus::Parameters>(
     max_height: BlockHeight,
 ) -> Result<Vec<WalletTransparentOutput>, SqliteClientError> {
     let mut stmt_blocks = wdb.conn.prepare(
-        "SELECT u.received_by_account,
```
str4d, line 932: To clarify, we were using the same `WalletTransparentOutput` struct to both load UTXOs into the database, and return them to the caller of this method? Does the caller of this method not require knowledge of the account?

str4d (approved):
utACK f1f9465f375a0aa77e86b7f586f990d5e2d47a57 modulo question about inserts and updates.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: They were unused when building without `--all-features`

zcash_client_sqlite/src/wallet.rs
nuttycom: The only thing we're doing with the results of this method is computing the transparent balance, so all we use is actually just the `value` field. When we go to implement the ZIP 315 recommendations, we'll probably want to add methods to the data access api that give this information more directly; right now, all the lookups are done by address rather than account and so it's a bit of a mess in any case.

zcash_client_sqlite/src/prepared.rs
daira: https://www.sqlite.org/lang_update.html
> It is not an error if the WHERE clause does not evaluate to true for any row in the table - this just means that the UPDATE statement affects zero rows.

But yes, a test would be good.

zcash_client_sqlite/src/prepared.rs
```diff
@@ -516,19 +561,17 @@ impl<'a, P: consensus::Parameters> DataConnStmtCache<'a, P> {
...
-            (":diversifier_index_be", &&diversifier_index.0[..]),
-            (":address", &address.encode(&self.wallet_db.params)),
-        ];
-
-        self.stmt_insert_address.execute(sql_args)?;
+    ) -> Result<(), SqliteClientError> {
+        self.stmt_insert_address.execute(
+            &self.wallet_db.params,
+            account,
+            diversifier_index,
```
daira, line 570: We are no longer reversing the bytes of `diversifier_index`, but the field is still called `diversifier_index_be`. Is that correct?

daira (approved):
The question about `diversifier_index_be` is blocking. Otherwise utACK modulo the suggested test.

daira (changes requested):
Changing to "request changes" just to clarify that the question about `diversifer_index_be` is blocking.

zcash_client_sqlite/src/prepared.rs
nuttycom: We perform the reverse here: https://github.com/zcash/librustzcash/pull/672/files/f1f9465f375a0aa77e86b7f586f990d5e2d47a57#diff-1e7d5d95a00896e1916695c2509d888591b4eb647ec38b7e750855e0b72c609dR70

zcash_client_sqlite/src/prepared.rs
nuttycom: Test: https://github.com/zcash/librustzcash/pull/672/commits/306d37e706ae56759ba754b1e35d2d00cb46a04a

zcash_client_sqlite/src/prepared.rs
```diff
@@ -33,6 +33,52 @@ use {
...
+    ///
+    /// Returns the database row for the newly-inserted address.
+    pub(crate) fn execute<P: consensus::Parameters>(
+        &mut self,
+        params: &P,
+        account: AccountId,
+        mut diversifier_index: DiversifierIndex,
+        address: &UnifiedAddress,
+    ) -> Result<(), rusqlite::Error> {
+        diversifier_index.0.reverse();
```
daira, line 70:
```suggestion
        // the diversifier index is stored in big-endian order to allow sorting
        diversifier_index.0.reverse();
```

daira (approved):
utACK

PR #674, opened 2022-10-13
Remove unused dependencies

str4d:
These were mostly detected with `cargo-udeps`, for which I've also added exclusions for the dependencies it can't detect are used in doc-tests.


ebfull (approved):

nuttycom (approved):
utACK

PR #676, opened 2022-10-14
Remove the `WalletReadTransparent` and `WalletWriteTransparent` extension traits.

nuttycom:
These traits introduce a problem, in that constraints on a method cannot be conditionally required based upon the presence or absence of a feature flag. Instead, we make the methods previously introduced by the removed traits present in all cases on the `WalletRead` and `WalletWrite` traits, but ensure that their implementations return an error if the caller attempts to use them in a wallet that has not been configured with support for transparent inputs functionality.

### DIFF
--- a/zcash_client_backend/src/data_api/error.rs
+++ b/zcash_client_backend/src/data_api/error.rs
@@ -78,6 +78,11 @@ pub enum Error<NoteId> {
     /// An error occurred deriving a spending key from a seed and an account
     /// identifier.
     KeyDerivationError(AccountId),
+
+    /// An error indicating that a call was attempted to a method providing
+    /// support
+    #[cfg(not(feature = "transparent-inputs"))]
+    TransparentInputsNotSupported,
 }
 
 impl ChainInvalid {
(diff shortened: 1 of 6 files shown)


zcash_client_backend/src/data_api/error.rs
```diff
@@ -78,6 +78,11 @@ pub enum Error<NoteId> {
     /// An error occurred deriving a spending key from a seed and an account
     /// identifier.
     KeyDerivationError(AccountId),
+
+    /// An error indicating that a call was attempted to a method providing
+    /// support
```
str4d, line 83: Unfinished comment.

zcash_client_sqlite/src/lib.rs
```diff
@@ -253,23 +244,32 @@ impl<P: consensus::Parameters> WalletRead for WalletDb<P> {
...
+        _account: AccountId,
     ) -> Result<HashSet<TransparentAddress>, Self::Error> {
-        wallet::get_transparent_receivers(&self.params, &self.conn, account)
+        #[cfg(feature = "transparent-inputs")]
+        return wallet::get_transparent_receivers(&self.params, &self.conn, _account);
+
+        #[cfg(not(feature = "transparent-inputs"))]
+        return Err(SqliteClientError::BackendError(
+            Error::TransparentInputsNotSupported,
+        ));
```
str4d, line 258: You can avoid `return`s by wrapping the first half in a scope:
```suggestion
        #[cfg(feature = "transparent-inputs")]
        {
            wallet::get_transparent_receivers(&self.params, &self.conn, _account)
        }

        #[cfg(not(feature = "transparent-inputs"))]
        Err(SqliteClientError::BackendError(
            Error::TransparentInputsNotSupported,
        ))
```

str4d (approved):
utACK 47a0d0d2b789c816525aa1d07c0ebbfd67758cf1

zcash_client_sqlite/src/lib.rs
nuttycom: I actually think that the `return` is helpful here, because it make the "the following code may not be executed" situation a bit more obvious. In general I don't like early `return`, but this is an exception.

PR #677, opened 2022-10-14
Mark our utxos spent when we detect them as inputs to a transaction.

nuttycom:
This modifies `decrypt_and_store_transaction` to check for inputs
to a transaction being decrypted that correspond to addresses in
our wallet. For each such UTXO found, it is marked spent in our
`utxos` table if present.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -10,9 +10,6 @@ use zcash_primitives::{
     },
 };
 
-#[cfg(feature = "transparent-inputs")]
-use zcash_primitives::{legacy::keys::IncomingViewingKey, sapling::keys::OutgoingViewingKey};
-
 use crate::{
     address::RecipientAddress,
     data_api::{
@@ -24,6 +21,9 @@ use crate::{
     wallet::OvkPolicy,
     zip321::{Payment, TransactionRequest},
 };
+
+#[cfg(feature = "transparent-inputs")]
+use zcash_primitives::{legacy::keys::IncomingViewingKey, sapling::keys::OutgoingViewingKey};
 
 /// Scans a [`Transaction`] for any information that can be decrypted by the accounts in
 /// the wallet, and saves it to the wallet.
@@ -50,14 +50,10 @@ where
         .or_else(|| params.activation_height(NetworkUpgrade::Sapling))
         .ok_or(Error::SaplingNotActive)?;
 
-    let sapling_outputs = decrypt_transaction(params, height, tx, &ufvks);
-
-    if !(sapling_outputs.is_empty() && tx.transparent_bundle().iter().all(|b| b.vout.is_empty())) {
-        data.store_decrypted_tx(&DecryptedTransaction {
-            tx,
-            sapling_outputs: &sapling_outputs,
-        })?;
-    }
+    data.store_decrypted_tx(&DecryptedTransaction {
+        tx,
+        sapling_outputs: &decrypt_transaction(params, height, tx, &ufvks),
+    })?;
 
     Ok(())
 }
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -475,7 +475,7 @@ where
     let utxos = wallet_db.get_unspent_transparent_outputs(&taddr, latest_anchor)?;
     let total_amount = utxos
         .iter()
-        .map(|utxo| utxo.txout.value)
+        .map(|utxo| utxo.txout().value)
         .sum::<Option<Amount>>()
         .ok_or_else(|| E::from(Error::InvalidAmount))?;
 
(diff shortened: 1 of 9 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -236,6 +236,8 @@ pub struct PrunedBlock<'a> {
 pub struct DecryptedTransaction<'a> {
     pub tx: &'a Transaction,
     pub sapling_outputs: &'a Vec<DecryptedOutput>,
+    #[cfg(feature = "transparent-inputs")]
+    pub utxos_spent: &'a Vec<OutPoint>,
```
str4d, line 240: Needs to be added to the changelog. At this point, the entry for `DecryptedTransaction` under "changed" should be split into an "added" and "removed (use instead)".

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -50,12 +56,53 @@ where
...
+        }
+
+        tx.transparent_bundle()
+            .iter()
+            .flat_map(|b| b.vin.iter())
+            .filter_map(|txin| {
+                // If the txin belongs to one of our addresses
+                if txin
+                    .script_sig
+                    .address()
```
str4d, line 75: This doesn't work, sorry 😅 `Script::address` is written to match on the contents of a `scriptPubKey` (which has a well-defined format for P2PKH and P2SH which directly contains the address), not a `scriptSig` (which only makes sense in the context of a `scriptPubKey`). We could in theory extract an address from a `scriptSig` (hash the pubkey in the case of a P2PKH redeem script, and hash the included script in the case of a P2SH redeem script), but we technically do not know whether a `scriptSig` is for a P2PKH or P2SH output without actually looking up that output (it could in fact be for some other output script that is not P2PKH or P2SH but happens to have a similar-looking `scriptSig`).

There are two things we should do:
- For this PR, we should extract the full list of outpoints from the transaction, and then call the data access API to determine whether any of the UTXOs appear in our data store (from which we can set `spends_our_utxos` and then either filter the list there, or just pass the full list over for the data access API to update only the ones that exist).
- We should document `Script::address` to indicate that it is for use specifically on `scriptPubKey`. We could maybe also just move it onto `TxOut` instead (making the existing method crate-private and exposing it publicly via a wrapper method on the output type).

str4d (changes requested):

zcash_client_backend/src/data_api.rs
nuttycom: Actually, there's no need to pass the outpoints separately at all; they're already part of the transaction that is passed through, so the only change needed is to always pass the `DecryptedTransaction` to the back end, not just when we successfully decrypted Sapling outputs.

zcash_client_backend/src/wallet.rs
```diff
@@ -30,15 +30,44 @@ pub struct WalletTx<N> {
...
+    outpoint: OutPoint,
+    txout: TxOut,
+    height: BlockHeight,
+    recipient_address: TransparentAddress,
 }
 
 impl WalletTransparentOutput {
-    pub fn address(&self) -> TransparentAddress {
-        self.txout.script_pubkey.address().unwrap()
+    pub fn from_parts(
```
str4d, line 42: Should really be documented, but in the interest of not blocking the SDKs I won't block on it this time.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -50,14 +50,10 @@ where
         .or_else(|| params.activation_height(NetworkUpgrade::Sapling))
         .ok_or(Error::SaplingNotActive)?;
 
-    let sapling_outputs = decrypt_transaction(params, height, tx, &ufvks);
-
-    if !(sapling_outputs.is_empty() && tx.transparent_bundle().iter().all(|b| b.vout.is_empty())) {
```
str4d, line 55: This change means that the transaction passed to `decrypt_and_store_transaction` is always saved to the wallet, instead of only when it is of interest to us. I think this is within the documented behaviour of this method, so it's fine.

zcash_client_sqlite/src/lib.rs
```diff
@@ -579,8 +578,15 @@ impl<'a, P: consensus::Parameters> WalletWrite for DataConnStmtCache<'a, P> {
                 }
             }
 
+            // If any of the utxos spent in the transaction are ours, mark them as spent.
+            #[cfg(feature = "transparent-inputs")]
+            for txin in d_tx.tx.transparent_bundle().iter().flat_map(|b| b.vin.iter()) {
+                wallet::mark_transparent_utxo_spent(up, tx_ref, &txin.prevout)?;
```
str4d, line 584: I confirmed that this uses `UPDATE` filtering on `prevout`, and therefore ignores spent TXOs that we aren't tracking.

str4d (approved):
utACK c773ea3b3e76afcdd6e6b9b6ddadb05cad3c45d0

PR #679, opened 2022-10-14
zcash_client_backend: Fix missing bound

str4d:
In 47a0d0d2b789c816525aa1d07c0ebbfd67758cf1 `WalletWriteTransparent` was merged into `WalletWrite`, including its associated type `UtxoRef`. However, `shield_transparent_funds` placed a bound on this associated type that was not moved to `WalletWrite`. This left the generic parameter `U` unconstrained, which didn't cause a local failure because the method has no local tests, but was immediately apparent when trying to use the method in the mobile SDKs.


str4d:
Confirmed that this compiles with `zcash-android-wallet-sdk`.

nuttycom (approved):
utACK

PR #680, opened 2022-10-17
Add accessor for `ZcashAddress::net`

nuttycom:


ebfull (approved):

sellout (approved):
I think the message from Slack would make a good justification in the commit or PR description.

nuttycom:
This information turns out to be available via the `convert` catamorphism and the `TryFromAddress` algebra. It's just not evident that this is how you'd have to get at it - by defining your own target type for the conversion.

PR #681, opened 2022-10-19
Update zcash_primitives to orchard 0.3; release zcash_primitives and zcash_proofs 0.8.

ebfull:
* This updates the `zcash_primitives` crate to use `orchard 0.3` which is a breaking change; we do a release of `zcash_primitives 0.8`.
* We update `zcash_proofs` to use `zcash_primitives 0.8` which is a breaking change; we do a release of `zcash_proofs 0.8`.
* We update `zcash_extensions`, `zcash_client_backend`, `zcash_client_sqlite` to use `zcash_primitives 0.8` which is a breaking change; we are not releasing these crates in this PR.
* We also update `zcash_client_backend` to use `orchard 0.3`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,5 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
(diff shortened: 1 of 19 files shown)


nuttycom (approved):
utACK of the first commit; I authored the second (albeit mechanically, using clippy +beta --fix) so it should be independently checked.

sellout (approved):

daira (commented):
Post-hoc ACK

PR #682, opened 2022-10-19
Fix changelog for `zcash_encoding`

nuttycom:


ebfull (approved):
ACK, PR only changes changelogs, can be merged as is.

Issue #490, opened 2022-01-31
Enable `#![deny(missing_docs)]` on `zcash_client_backend`


str4d:
Once we've ensured that the full API is documented, we can enable this to ensure we don't merge undocumented APIs in future.

nuttycom:
Also fix https://github.com/zcash/librustzcash/pull/685#discussion_r1004803808

PR #685, opened 2022-10-24
Allow insertion of UTXOs associated with the legacy taddr.

nuttycom:
The legacy transparent address is never added to the `addresses` table, but we still need to be able to receive UTXOs sent to that address. So, we add a special case for when a UTXO matches that legacy address, and set the account ID to 0 manually.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -23,7 +23,10 @@ use crate::{
 };
 
 #[cfg(feature = "transparent-inputs")]
-use zcash_primitives::{legacy::TransparentAddress, sapling::keys::OutgoingViewingKey};
+use {
+    crate::wallet::WalletTransparentOutput,
+    zcash_primitives::{legacy::TransparentAddress, sapling::keys::OutgoingViewingKey},
+};
 
 /// Scans a [`Transaction`] for any information that can be decrypted by the accounts in
 /// the wallet, and saves it to the wallet.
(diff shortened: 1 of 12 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -440,6 +440,7 @@ pub fn shield_transparent_funds<E, N, P, D, R, U>(
     params: &P,
     prover: impl TxProver,
     usk: &UnifiedSpendingKey,
```
daira, line 442: Presumably when Orchard is supported, this is going to shield to the most preferred pool in the USK, which will prefer Orchard over Sapling. Should that be specified now in the doc comment, rather than saying it shields to the default Sapling address?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -262,15 +263,24 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
...
+    let mut ua_query =
+        conn.prepare("SELECT address, diversifier_index_be FROM addresses WHERE account = :account")?;
     let mut rows = ua_query.query(named_params![":account": &u32::from(account)])?;
 
     while let Some(row) = rows.next()? {
         let ua_str: String = row.get(0)?;
+        let di_vec: Vec<u8> = row.get(1)?;
+        let mut di_be: [u8; 11] = di_vec.try_into().map_err(|_| {
+            SqliteClientError::CorruptedData(
+                "Diverisifier index is not an 11-byte value".to_owned(),
```
daira, line 279:
```suggestion
                "Diversifier index is not an 11-byte value".to_owned(),
```

daira (approved):
utACK with minor comments.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/274af779c101adbb19e1d26d02551ebad92c0041..1dbb04d1ca0ee4bc17801f926ddd822427d5a2bb) to resolve review comments.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Noted in https://github.com/zcash/librustzcash/issues/490

daira (commented):
Post-hoc ACK

PR #689, opened 2022-10-31
Add dynamic note selection that is responsive to changes in required fees.

nuttycom:
This PR is best reviewed hiding whitespace changes.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -71,6 +71,8 @@ and this library adheres to Rust's notion of
   - `TransactionBalance`
   - `BasicFixedFeeChangeStrategy` - a `ChangeStrategy` implementation that
     reproduces current wallet change behavior
+  - `fixed` a new module containing of change selection strategies for the
+    existing fixed fee rule.
 - New experimental APIs that should be considered unstable, and are
   likely to be modified and/or moved to a different module in a future
   release:
@@ -107,9 +109,11 @@ and this library adheres to Rust's notion of
     the anchor height being returned; this had previously been hardcoded to
     `data_api::wallet::ANCHOR_OFFSET`.
   - `WalletRead::get_spendable_notes` has been renamed to
-    `get_spendable_sapling_notes`
+    `get_spendable_sapling_notes` and now takes as an argument a vector of
+    note IDs to be excluded from consideration.
   - `WalletRead::select_spendable_notes` has been renamed to
-    `select_spendable_sapling_notes`
+    `select_spendable_sapling_notes` and now takes as an argument a vector of
+    note IDs to be excluded from consideration.
   - The `WalletRead::NoteRef` and `WalletRead::TxRef` associated types are now
     required to implement `Eq` and `Ord`
 - The `zcash_client_backend::data_api::SentTransaction` type has been
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,6 +28,7 @@ and this library adheres to Rust's notion of
   - `AddressMetadata`
 - `zcash_client_backend::data_api`:
   - `PoolType`
+  - `ShieldedPool`
   - `Recipient`
   - `SentTransactionOutput`
   - `WalletRead::get_unified_full_viewing_keys`
@@ -42,6 +43,11 @@ and this library adheres to Rust's notion of
   - `WalletWrite::get_next_available_address`
   - `WalletWrite::put_received_transparent_utxo`
   - `impl From<prost::DecodeError> for error::Error`
+  - `chain::error`: a module containing error types type that that can occur only
+    in chain validation and sync have been separated out from errors related to
+    other wallet operations.
+  - `input_selection`: a module containing types related to the process
+    of selecting inputs to be spent, given a transaction request.
 - `zcash_client_backend::decrypt`:
   - `TransferType`
 - `zcash_client_backend::proto`:
@@ -50,6 +56,7 @@ and this library adheres to Rust's notion of
   - gRPC bindings for the `lightwalletd` server, behind a `lightwalletd-tonic`
     feature flag.
 - `zcash_client_backend::zip321::TransactionRequest` methods:
+  - `TransactionRequest::empty` for constructing a new empty request.
   - `TransactionRequest::new` for constructing a request from `Vec<Payment>`.
   - `TransactionRequest::payments` for accessing the `Payments` that make up a
     request.
@@ -72,6 +79,7 @@ and this library adheres to Rust's notion of
   - `zcash_client_backend::encoding::AddressCodec`
   - `zcash_client_backend::encoding::encode_payment_address`
   - `zcash_client_backend::encoding::encode_transparent_address`
+- `impl Eq for zcash_client_backend::address::UnifiedAddress`
 
 ### Changed
 - MSRV is now 1.56.1.
@@ -90,13 +98,8 @@ and this library adheres to Rust's notion of
     been replaced by `ephemeral_key`.
   - `zcash_client_backend::proto::compact_formats::CompactSaplingOutput`: the
     `epk` method has been replaced by `ephemeral_key`.
-- `data_api::wallet::spend_to_address` now takes a `min_confirmations`
-  parameter, which the caller can provide to specify the minimum number of
-  confirmations required for notes being selected. A default value of 10
-  confirmations is recommended.
 - Renamed the following in `zcash_client_backend::data_api` to use lower-case
   abbreviations (matching Rust naming conventions):
-  - `error::Error::InvalidExtSK` to `Error::InvalidExtSk`
   - `testing::MockWalletDB` to `testing::MockWalletDb`
 - Changes to the `data_api::WalletRead` trait:
   - `WalletRead::get_target_and_anchor_heights` now takes
@@ -107,6 +110,8 @@ and this library adheres to Rust's notion of
     `get_spendable_sapling_notes`
   - `WalletRead::select_spendable_notes` has been renamed to
     `select_spendable_sapling_notes`
+  - The `WalletRead::NoteRef` and `WalletRead::TxRef` associated types are now
+    required to implement `Eq` and `Ord`
 - The `zcash_client_backend::data_api::SentTransaction` type has been
   substantially modified to accommodate handling of transparent inputs.
   Per-output data has been split out into a new struct `SentTransactionOutput`
@@ -116,24 +121,28 @@ and this library adheres to Rust's notion of
   `store_decrypted_tx`.
 - `data_api::ReceivedTransaction` has been renamed to `DecryptedTransaction`,
   and its `outputs` field has been renamed to `sapling_outputs`.
-- `data_api::error::Error::Protobuf` now wraps `prost::DecodeError` instead of
-  `protobuf::ProtobufError`.
-- `data_api::error::Error` has the following additional cases:
-  - `Error::BalanceError` in the case of amount addition overflow
-    or subtraction underflow.
-  - `Error::MemoForbidden` to report the condition where a memo was
-    specified to be sent to a transparent recipient.
-  - `Error::TransparentInputsNotSupported` to represent the condition
-    where a transparent spend has been requested of a wallet compiled without
-    the `transparent-inputs` feature.
-  - `Error::AddressNotRecognized` to indicate that a transparent address from
-    which funds are being requested to be spent does not appear to be associated
-    with this wallet.
-  - `Error::ChildIndexOutOfRange` to indicate that a diversifier index for an
-    address is out of range for valid transparent child indices.
-  - `Error::NoteMismatch` to indicate that a note being spent is not associated
-    with either the internal or external full viewing keys corresponding to the
-    provided spending key.
+- `data_api::error::Error` has been substantially modified. It now wraps database,
+  note selection, builder, and other errors
+  - `Error::DataSource` has been added.
+  - `Error::NoteSelection` has been added.
+  - `Error::BalanceError` has been added.
+  - `Error::MemoForbidden` has been added.
+  - `Error::AddressNotRecognized` has been added.
+  - `Error::ChildIndexOutOfRange` has been added.
+  - `Error::NoteMismatch` has been added.
+  - `Error::InsufficientBalance` has been renamed `InsufficientFunds` and
+    restructured to have named fields.
+  - `Error::Protobuf` has been removed; these decoding errors are now
+    produced as data source and/or block-source implementation-specific
+    errors.
+  - `Error::InvalidChain` has been removed; its former purpose is now served
+    by `zcash_client_backend::data_api::chain::ChainError`.
+  - `Error::InvalidNewWitnessAnchor` and `Error::InvalidWitnessAnchor` have
+    been moved to `zcash_client_backend::data_api::chain::error::ContinuityError`
+  - `Error::InvalidExtSk` (now unused) has been removed.
+  - `Error::KeyNotFound` (now unused) has been removed.
+  - `Error::KeyDerivationError` (now unused) has been removed.
+  - `Error::SaplingNotActive` (now unused) has been removed.
 - `zcash_client_backend::decrypt`:
   - `decrypt_transaction` now takes a `HashMap<_, UnifiedFullViewingKey>`
     instead of `HashMap<_, ExtendedFullViewingKey>`.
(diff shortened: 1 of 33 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -18,56 +19,75 @@ use crate::{
...
+/// Errors that can occur during transaction construction.
 #[derive(Debug, PartialEq, Eq)]
-pub enum Error {
-    ChangeIsNegative(Amount),
-    InvalidAmount,
-    NoChangeAddress,
+pub enum Error<FeeError> {
+    /// Insufficient funds were provided to the transaction builder; the given
+    /// additional amount is required in order to construct the transaction.
+    InsufficientFunds(Amount),
```
pacu, line 59: This will surely be a UX problem.  Is there a way of doing a `dry-run` to know the estimated fee for a given amount to show on wallets' send flow? Waiting for the user to complete the whole transaction to deny it because of variable fees not calculated well beforehand seems pretty inconvenient

nuttycom: Usually insufficient funds will be detected at note selection time, so the build-time check is really just a fall-back to ensure that the transaction doesn't violate balance rules. Also, the next generation of transaction construction will provide an intermediate result that has the privacy implications of the transaction listed out for the user's approval.

pacu: > Also, the next generation of transaction construction will provide an intermediate result that has the privacy implications of the transaction listed out for the user's approval.

I'm not following you here. In terms of wallet UX, can we have a function that given an amount, it will return an approximate fee so that wallet developers can validate their "send funds" forms UI?

str4d (commented):
Partial review.

zcash_client_backend/src/data_api.rs
```diff
@@ -268,6 +271,23 @@ pub enum PoolType {
     Transparent,
     /// The Sapling value pool
     Sapling,
```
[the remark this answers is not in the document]
nuttycom: This type ended up being unnecessary.

zcash_primitives/src/transaction/builder.rs
nuttycom: There's a subtlety here which is that we would need such a function to essentially prepare the transaction and "reserve" the notes/UTXOs that would be spent, so that we can be sure that the fee doesn't change between the time that the UI displays the fee and the time that the user goes to actually confirm the transaction; for example, if a new note comes in that would be preferentially selected by the note selection algorithm, the fees and the notes being spent could change in that interim. So, what we need to do is what we've discussed before: prepare the transaction and lock the notes to be spent to it, and then return that prepared transaction (along with its privacy implications) to be confirmed by the user. The fee should be displayed on the confirmation page.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e400f8ac..f0cff4391a66e57ef45a622841003bfd2ec92fe8) to address review comments.

zcash_primitives/src/transaction/builder.rs
pacu: The problem with this approach is that it doesn't solve the problem I'm referring to, which is: "don't let the user proceed to a place that inevitably is an error state if it can be known beforehand". 

The SDK will update every 20~ seconds and being the user in the send flow, it will refresh the available information and that information in the local state will be used to create the transaction. Any changes in balance should trigger a UI refresh in send flow so that it can be validated against the latest current state.

pacu: We do know that there are certain combinations of amount and notes used that will make the fees an impediment for the transaction to be sent. The "dry-run" function I'm proposing is aimed to catch and report that particular situation so that wallet developers can validate their send form before proceeding. There will be situations like you mention where timing and chance will plot against it. *

Also, it can be the case that a wallet developer does not want a confirmation screen. So this additional step to safety check will not be present. Which makes this validation function more helpful to avoid failed transactions for consensus reasons which are a pain to explain to users.




* I hadn't thought of this. Can users be dusted into oblivion thanks to these new fee scheme?

nuttycom: No; it is the responsibility of the ZIP 321 note selection implementation to ignore notes that are below the value of the fee required to spend them (though low-valued notes can have their value recovered by using them as a free input when spending a note that is by itself large enough to cover the full transaction amount.)

nuttycom: In any case, this error needs to exist because we must *always* validate the transaction at construction time anyway. The transaction preparation step you're talking about is effectively the input selection step itself; it's been renamed 'propose_transaction' to make this a bit clearer: https://github.com/zcash/librustzcash/pull/689/files#diff-fedecff319cb43f7a4f8b913cdee2332ba452a6310e2470cf6cc7631ca83291cR126

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+    pub fn sapling_inputs(&self) -> &[SpendableNote<NoteRef>] {
+        &self.sapling_inputs
+    }
+    pub fn balance(&self) -> &TransactionBalance {
+        &self.balance
+    }
+    pub fn fee_rule(&self) -> &FeeRuleT {
+        &self.fee_rule
+    }
```
str4d, line 90: Needs documentation.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+/// A strategy for selecting transaction inputs and proposing transaction outputs.
+///
+/// Inputs returned should include only economically useful inputs with economic usefulness
+/// determined according to the fee estimator; that is, do not return inputs that cause fees to
+/// increase by an amount greater than the value of the input.
```
str4d, line 97:
```suggestion
/// Proposals should only include economically useful inputs, as determined by
/// `Self::FeeRule`; that is, do not return inputs that cause fees to increase by an
/// amount greater than the value of the input.
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+    /// selectors that may use specialized knowledge of the internals of a
+    /// particular backing data store, if the generic API of `WalletRead`
+    /// does not provide sufficiently fine-grained operations for a particular
+    /// backing store to optimally perform input selection.
+    type DataSource: WalletRead;
+    /// The type of the fee rule that this input selector uses when computing fees.
+    type FeeRule: FeeRule;
+
+    /// Returns a list of spendable inputs sufficient to cover the amount required to satisfy the
+    /// given [`TransactionRequest`], with fees as estimated by the provided [`FeeEstimator`]
```
str4d, line 112: Needs updating.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+        params: &ParamsT,
+        wallet_db: &Self::DataSource,
+        account: AccountId,
+        anchor_height: BlockHeight,
+        target_height: BlockHeight,
+        transaction_request: TransactionRequest,
+    ) -> Result<
+        Proposal<
+            Self::FeeRule,
+            std::convert::Infallible,
```
str4d, line 137: Is this the right type to use here, or should this use `()`? Is it valid to construct an empty vector of uninhabitable types (clearly Rust lets you)? What happens if the caller attempts to iterate over them?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+            <<Self as InputSelector>::DataSource as WalletRead>::NoteRef,
+        >,
+        InputSelectorError<<<Self as InputSelector>::DataSource as WalletRead>::Error, Self::Error>,
+    >
+    where
+        ParamsT: consensus::Parameters;
+
+    /// Returns a list of transparent inputs to spend in a wallet-internal shielding transaction.
```
str4d, line 145: Needs updating.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+
+impl sapling::OutputView for SaplingPayment {
+    fn value(&self) -> Amount {
+        self.0
+    }
+}
+
+/// An InputSelector implementation that uses only methods available via the [`WalletRead`]
+/// interface to perform input selection, and that uses a greedy strategy to select between
+/// available notes.
```
str4d, line 219:
```suggestion
/// An [`InputSelector`] implementation that uses a greedy strategy to select between
/// available notes.
///
/// This implementation performs input selection using methods available via the
/// [`WalletRead`] interface.
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,398 @@
...
+pub struct GreedyInputSelector<DbT, ChangeT> {
+    change_strategy: ChangeT,
+    _ds_type: PhantomData<DbT>,
+}
+
+impl<DbT, ChangeT: ChangeStrategy> GreedyInputSelector<DbT, ChangeT> {
+    pub fn new(change_strategy: ChangeT) -> Self {
```
str4d, line 226: Needs documentation.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -583,23 +559,69 @@ where
...
+            // determine the Sapling output indices; `enumerate` will no longer suffice
+            outputs: proposal
+                .balance()
+                .proposed_change()
+                .iter()
+                .enumerate()
+                .map(|(idx, change_value)| match change_value {
+                    ChangeValue::Sapling(value) => {
+                        let output_index = tx_metadata.output_index(idx).expect(
+                            "Missing autoshielding transaction. This is a programming error.",
```
str4d, line 580:
```suggestion
                            "Missing Sapling output of autoshielding transaction. This is a programming error.",
```

str4d (changes requested):
Reviewed a9e955e862986abcf1c88a418b1234def97126ce. Overall this is looking good!

zcash_client_sqlite/src/chain.rs
```diff
@@ -385,16 +394,15 @@ mod tests {
...
-            Err(SqliteClientError::BackendError(Error::InvalidChain(lower_bound, _))) => {
-                assert_eq!(lower_bound, sapling_activation_height() + 2)
-            }
-            _ => panic!(),
-        }
+        );
+
+        assert!(
+            matches!(val_result, Err(Error::Chain(e)) if e.at_height() == sapling_activation_height() + 2)
+        );
```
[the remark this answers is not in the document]
nuttycom: The generics cause issues for compilation for a number of these macro calls.

zcash_client_sqlite/src/chain.rs
```diff
@@ -21,53 +21,51 @@ use {
...
+pub(crate) fn blockdb_with_blocks<F, DbErrT, NoteRef>(
+    block_source: &BlockDb,
     last_scanned_height: BlockHeight,
     limit: Option<u32>,
     mut with_row: F,
-) -> Result<(), SqliteClientError>
+) -> Result<(), Error<DbErrT, SqliteClientError, NoteRef>>
 where
-    F: FnMut(CompactBlock) -> Result<(), SqliteClientError>,
+    F: FnMut(CompactBlock) -> Result<(), Error<DbErrT, SqliteClientError, NoteRef>>,
```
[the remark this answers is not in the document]
nuttycom: This is crate-private.

zcash_client_sqlite/src/chain.rs
```diff
@@ -160,53 +158,65 @@ pub(crate) fn blockmetadb_get_max_cached_height(
...
+pub(crate) fn fsblockdb_with_blocks<F, DbErrT, NoteRef>(
     cache: &FsBlockDb,
     last_scanned_height: BlockHeight,
     limit: Option<u32>,
     mut with_block: F,
-) -> Result<(), SqliteClientError>
+) -> Result<(), Error<DbErrT, FsBlockDbError, NoteRef>>
 where
-    F: FnMut(CompactBlock) -> Result<(), SqliteClientError>,
+    F: FnMut(CompactBlock) -> Result<(), Error<DbErrT, FsBlockDbError, NoteRef>>,
```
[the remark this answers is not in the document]
nuttycom: This is crate-private.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -233,26 +242,26 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
     ///
     /// Returns an error if the given Merkle path does not have the same anchor as the
     /// paths for previous Sapling notes.
-    pub fn add_sapling_spend(
+    pub fn add_sapling_spend<FE>(
```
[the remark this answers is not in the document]
nuttycom: I think that we should replace `FE` with `!` once that stabilizes.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -233,26 +242,26 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
-    ) -> Result<(), Error> {
+    ) -> Result<(), Error<FE>> {
         self.sapling_builder
             .add_spend(&mut self.rng, extsk, diversifier, note, merkle_path)
             .map_err(Error::SaplingBuild)
     }
 
     /// Adds a Sapling address to send funds to.
-    pub fn add_sapling_output(
+    pub fn add_sapling_output<FE>(
```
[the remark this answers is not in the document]
nuttycom: Same here; `!` is the correct type and then we can remove the parameter.

zcash_primitives/src/transaction/builder.rs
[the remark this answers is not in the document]
str4d: Hmm, that seems _worse_ from a UX perspective, because then the errors returned by this `Builder` API are incompatible with the errors returned by `Builder::build` and can't be handled together in any way. Returning `sapling::builder::Error` here would at least permit the caller to merge them if they want.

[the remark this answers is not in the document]
nuttycom: I thought that `!` was going to be the bottom type? The bottom type should be treated by the compiler as every type, including the one you want here.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/a9e955e862986abcf1c88a418b1234def97126ce..a5cef6c) to address review comments and add handling of dust inputs.

zcash_primitives/src/transaction/builder.rs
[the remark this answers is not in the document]
str4d: The above example can be [fixed](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=4a18ed948da2de31e1525e20491fb048) by adding an `impl From<Error<!>> for Error<u32>`, but that can't be done generically as `impl<T> From<Error<!>> for Error<T>` because it collides with `impl<T> From<T> for T` (precisely because `!` can coerce to anything).

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/a9e955e862986abcf1c88a418b1234def97126ce..bd16035) to address review comments and add handling of dust inputs.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/a9e955e862986abcf1c88a418b1234def97126ce..bd16035) to address review comments and add handling of dust inputs.

str4d (approved):
utACK bd16035c77163606ff24658569d29a0bfbcbe892 modulo comments.

nuttycom:
@str4d I realized that I hadn't properly accounted for handling dust *outputs* in the change APIs, so I have force-pushed the second commit here to address the problem: https://github.com/zcash/librustzcash/compare/bd16035c77163606ff24658569d29a0bfbcbe892..847ba49761fdfec1213a6d6417db6165c549241e

zcash_client_backend/src/fees.rs
```diff
@@ -76,10 +81,76 @@ pub enum ChangeError<E> {
...
+    /// Explicitly allow the creation of dust change amounts greater than the specified value.
+    AllowDustChange,
+    /// Allow dust amounts to be added to the transaction fee
+    AddDustToFee,
+}
+
+/// A policy describing how a [`ChangeStrategy`] should treat potentially dust-valued change
+/// outputs (outputs that are likely to be without economic value due to fee rules.)
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct DustOutputPolicy {
```
daira, line 118: `DustFilter`

zcash_client_backend/CHANGELOG.md
```diff
@@ -64,6 +71,8 @@ and this library adheres to Rust's notion of
   - `TransactionBalance`
   - `BasicFixedFeeChangeStrategy` - a `ChangeStrategy` implementation that
     reproduces current wallet change behavior
+  - `fixed` a new module containing of change selection strategies for the
```
str4d, line 74:
```suggestion
  - `fixed`, a new module containing of change selection strategies for the
```
(or `:` as was used above)

zcash_primitives/CHANGELOG.md
```diff
@@ -24,6 +27,8 @@ and this library adheres to Rust's notion of
   and types related to fee calculations.
   - `FeeRule` a trait that describes how to compute the fee required for a
     transaction given inputs and outputs to the transaction.
+  - `zcash_primitives::transaction::fees::fixed` a new module containing an implementation
```
str4d, line 30:
```suggestion
  - `fixed`, a new module containing an implementation
```

str4d (approved):
utACK 847ba49761fdfec1213a6d6417db6165c549241e mod comments.

nuttycom:
I will address the changelog comments in #694, which will follow immediately.

zcash_client_backend/CHANGELOG.md
```diff
@@ -42,6 +43,11 @@ and this library adheres to Rust's notion of
   - `WalletWrite::get_next_available_address`
   - `WalletWrite::put_received_transparent_utxo`
   - `impl From<prost::DecodeError> for error::Error`
+  - `chain::error`: a module containing error types type that that can occur only
```
daira, line 46:
```suggestion
  - `chain::error`: a module containing error types that that can occur only
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -42,6 +43,11 @@ and this library adheres to Rust's notion of
...
+    in chain validation and sync have been separated out from errors related to
+    other wallet operations.
```
daira, line 48:
```suggestion
  - `chain::error`: a module containing error types that can occur only in chain
    validation and sync, which have been separated out from errors related to
    other wallet operations.
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -160,8 +173,38 @@ and this library adheres to Rust's notion of
...
+  `zcash_client_backend::data_api::chain::error`.
+- The error type of `zcash_client_backend::data_api::wallet::decrypt_and_store_transaction`
+  has been changed; the error type produced by the provided `WalletWrite` instance is
+  returned directly.
+
+### Deprecated
+- `zcash_client_backend::data_api::wallet::create_spend_to_address` has been
+  deprecated. Use `zcash_client_backend::data_api::wallet::spend` instead. If
+  you wish to continue using `create_spend_to_address`, note that the arguments
+  to the function has been modified to take a unified spending key instead of a
```
daira, line 203:
```suggestion
  to the function have been modified to take a unified spending key instead of a
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -160,8 +173,38 @@ and this library adheres to Rust's notion of
...
+  Sapling extended spending key, and now also requires a `min_confirmations`
```
daira, line 204:
```suggestion
  Sapling extended spending key, and it now also requires a `min_confirmations`
```

daira, line 204:
```suggestion
  to the function have been modified to take a unified spending key instead of a
  Sapling extended spending key, and it now also requires a `min_confirmations`
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -0,0 +1,452 @@
...
+        >,
+        InputSelectorError<<<Self as InputSelector>::DataSource as WalletRead>::Error, Self::Error>,
+    >
+    where
+        ParamsT: consensus::Parameters;
+
+    /// Performs input selection and returns a proposal for the construction of a shielding
+    /// transaction.
+    ///
+    /// Implementations should return the maximum possible number of economically useful inputs
```
daira, line 154: Is "maximum possible" correct? That would be all the notes held by the wallet.

I don't think "minimum possible" is correct either; while it's not useful to return additional unneeded notes, there's no guarantee or requirement that the returned set of notes be the minimum possible number.

PR #690, opened 2022-11-01
Dependency refactor

str4d:


zcash_client_backend/Cargo.toml
```diff
@@ -17,29 +17,55 @@ rust-version = "1.56.1"
...
-zcash_note_encryption = { version = "0.2", path = "../components/zcash_note_encryption" }
-zcash_primitives = { version = "0.8", path = "../zcash_primitives" }
+
+# - Shielded protocols
+bls12_381 = "0.7"
+group = "0.12"
+orchard = "0.3"
+
+# - Test dependencies
+proptest = { version = "1.0.0", optional = true }
```
daira, line 55: Should this be a dev dependency?

daira (approved):
utACK. (Thanks for separating the changes from the reorganisation, that helped a lot with review.)

zcash_client_backend/Cargo.toml
nuttycom: This is included as a normal dependency because we expose it in the `testing` modules under the `test-dependency` feature flags, so that if someone needs proptest generators for their own tests, they have them.

zcash_client_backend/Cargo.toml
```diff
@@ -17,49 +17,72 @@ rust-version = "1.56.1"
 development = ["zcash_proofs"]
 
 [dependencies]
+zcash_address = { version = "0.2", path = "../components/zcash_address" }
```
nuttycom, line 20: A problem that I encountered in the zcash-light-client-ffi crate is that these path-based dependencies can cause problems if you neglect to override one of them with a patch version, even if you're not directly depending upon the crate. Is it possible to remove the path dependency, and instead have that scoped to the patch directives in the workspace file?

nuttycom (approved):
utACK

PR #691, opened 2022-11-02
zcash_client_backend: Switch from `protobuf 2` to `prost 0.11`

str4d:
The latter is maintained by the Tokio developers, and has easy integration with the `tonic` gRPC library which is actively maintained.


nuttycom (approved):
utACK modulo Clippy complaints.

PR #693, opened 2022-11-02
Fix a bug in migrations when transaction block height is unknown.

nuttycom:
Un-mined transactions cannot currently be migrated to retroactively store fee information, because the encoding used to persist them is unknown.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
@@ -7,7 +7,7 @@ use schemer_rusqlite::RusqliteMigration;
 use uuid::Uuid;
 
 use zcash_primitives::{
-    consensus::{self, BlockHeight, BranchId},
+    consensus::BranchId,
     transaction::{
         components::amount::{Amount, BalanceError},
         Transaction,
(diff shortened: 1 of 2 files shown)


daira (approved):
utACK

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -85,20 +85,22 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
-                        .map_err(|e| {
-                            WalletMigrationError::CorruptedData(format!(
-                                "Parsing failed for transaction {:?}: {:?}",
-                                id_tx, e
-                            ))
-                        })?;
+            if let Some((tx_bytes, block_height)) = tx_bytes.zip(block_height) {
+                let tx = Transaction::read(
+                    &tx_bytes[..],
+                    BranchId::for_height(&self.params, block_height),
```
str4d, line 96: The problem here is that we don't necessarily know the block height in order to figure out the consensus branch ID. But doing so only matters for v4 transactions; v5 transactions ignore the `BranchId` passed in and use the one stored internally. So at a minimum we can reliably parse v5 transactions by passing a dummy value, and then if the resulting transaction is v4 and `block_height.is_none()` then we could filter those out as not reliable.

But my second question is, does anything that `tx` is used for below need the consensus branch ID? If it doesn't affect the `Transaction::fee_paid` result, then we could just not read `block_height` from the database at all and use a dummy value all the time (like we do in a few places in `zcashd`).

nuttycom: Oh, that's a good point; we're not doing anything with consensus_branch_id, so we can just always use Nu5, since all we actually want is the prevouts and value balances.

str4d (approved):
utACK 77efc18dd9154786a626b5694f8081359299b198

PR #694, opened 2022-11-04
Add ZIP 317 fee calculation strategy.

nuttycom:
This builds atop #689

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -71,6 +71,8 @@ and this library adheres to Rust's notion of
   - `TransactionBalance`
   - `BasicFixedFeeChangeStrategy` - a `ChangeStrategy` implementation that
     reproduces current wallet change behavior
+  - `fixed` a new module containing of change selection strategies for the
+    existing fixed fee rule.
 - New experimental APIs that should be considered unstable, and are
   likely to be modified and/or moved to a different module in a future
   release:
@@ -107,9 +109,11 @@ and this library adheres to Rust's notion of
     the anchor height being returned; this had previously been hardcoded to
     `data_api::wallet::ANCHOR_OFFSET`.
   - `WalletRead::get_spendable_notes` has been renamed to
-    `get_spendable_sapling_notes`
+    `get_spendable_sapling_notes` and now takes as an argument a vector of
+    note IDs to be excluded from consideration.
   - `WalletRead::select_spendable_notes` has been renamed to
-    `select_spendable_sapling_notes`
+    `select_spendable_sapling_notes` and now takes as an argument a vector of
+    note IDs to be excluded from consideration.
   - The `WalletRead::NoteRef` and `WalletRead::TxRef` associated types are now
     required to implement `Eq` and `Ord`
 - The `zcash_client_backend::data_api::SentTransaction` type has been
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,6 +28,7 @@ and this library adheres to Rust's notion of
   - `AddressMetadata`
 - `zcash_client_backend::data_api`:
   - `PoolType`
+  - `ShieldedPool`
   - `Recipient`
   - `SentTransactionOutput`
   - `WalletRead::get_unified_full_viewing_keys`
@@ -42,6 +43,11 @@ and this library adheres to Rust's notion of
   - `WalletWrite::get_next_available_address`
   - `WalletWrite::put_received_transparent_utxo`
   - `impl From<prost::DecodeError> for error::Error`
+  - `chain::error`: a module containing error types type that that can occur only
+    in chain validation and sync have been separated out from errors related to
+    other wallet operations.
+  - `input_selection`: a module containing types related to the process
+    of selecting inputs to be spent, given a transaction request.
 - `zcash_client_backend::decrypt`:
   - `TransferType`
 - `zcash_client_backend::proto`:
@@ -50,6 +56,7 @@ and this library adheres to Rust's notion of
   - gRPC bindings for the `lightwalletd` server, behind a `lightwalletd-tonic`
     feature flag.
 - `zcash_client_backend::zip321::TransactionRequest` methods:
+  - `TransactionRequest::empty` for constructing a new empty request.
   - `TransactionRequest::new` for constructing a request from `Vec<Payment>`.
   - `TransactionRequest::payments` for accessing the `Payments` that make up a
     request.
@@ -72,6 +79,7 @@ and this library adheres to Rust's notion of
   - `zcash_client_backend::encoding::AddressCodec`
   - `zcash_client_backend::encoding::encode_payment_address`
   - `zcash_client_backend::encoding::encode_transparent_address`
+- `impl Eq for zcash_client_backend::address::UnifiedAddress`
 
 ### Changed
 - MSRV is now 1.56.1.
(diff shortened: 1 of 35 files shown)


zcash_client_backend/CHANGELOG.md
```diff
@@ -71,6 +71,10 @@ and this library adheres to Rust's notion of
   - `TransactionBalance`
   - `BasicFixedFeeChangeStrategy` - a `ChangeStrategy` implementation that
     reproduces current wallet change behavior
+  - `fixed` a new module containing of change selection strategies for the
+    existing fixed fee rule.
+  - `zip317` a new module containing change selection strategies for the ZIP
+    317 fee rule.
```
str4d, line 77:
```suggestion
  - `fixed`, a new module containing of change selection strategies for the
    existing fixed fee rule.
  - `zip317`, a new module containing change selection strategies for the ZIP
    317 fee rule.
```

zcash_primitives/CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ and this library adheres to Rust's notion of
   and types related to fee calculations.
   - `FeeRule` a trait that describes how to compute the fee required for a
     transaction given inputs and outputs to the transaction.
+- `zcash_primitives::transaction::fees::zip317` a new module containing an implementation
+  of the existing fixed fee rule.
+- `zcash_primitives::transaction::fees::zip317` a new module containing an implementation
+  of the ZIP 317 fee rules.
```
str4d, line 33:
```suggestion
  - `fixed`, a new module containing an implementation of the existing fixed fee rule.
  - `zip317`, a new module containing an implementation of the ZIP 317 fee rules.
```

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -0,0 +1,101 @@
...
+    grace_actions: usize,
+    p2pkh_standard_input_size: usize,
+    p2pkh_standard_output_size: usize,
+}
+
+impl FeeRule {
+    /// Construct a new FeeRule instance with the specified parameter values.
+    /// Returns `None` if either `p2pkh_standard_input_size` or `p2pkh_standard_output_size`
+    /// are zero.
+    pub fn new(
```
str4d, line 31: Is there a benefit to enabling non-standard parameters to be used? If so, `FeeRule::new` is too innocuous for such a constructor. We should rename and document it appropriately:
```suggestion
    pub fn non_standard(
```

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -0,0 +1,101 @@
...
+        &self,
+        _params: &P,
+        _target_height: BlockHeight,
+        transparent_inputs: &[impl transparent::InputView],
+        transparent_outputs: &[impl transparent::OutputView],
+        sapling_input_count: usize,
+        sapling_output_count: usize,
+    ) -> Result<Amount, Self::Error> {
+        let t_in_total_size = transparent_inputs.len() * 150;
+        let t_out_total_size = transparent_outputs.len() * 34;
```
str4d, line 90: Bug: this assumes the kind of each transparent input and output, such that `p2pkh_standard_input_size` and `p2pkh_standard_output_size` cancel out and cause the transparent contribution to only use the number of inputs and outputs. Instead it needs to use the as-serialized-in-transaction size of `transparent_inputs` and `transparent_outputs` (which can be calculated from the data accessible via `transparent::{inputView, OutputView}`).

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        let ceildiv = |num: usize, den: usize| (num / den) + if num % den == 0 { 0 } else { 1 };
```
str4d, line 92: Why not:
```suggestion
        let ceildiv = |num: usize, den: usize| ((num + den - 1) / den);
```

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        let logical_actions = max(
+            ceildiv(t_in_total_size, self.p2pkh_standard_input_size),
+            ceildiv(t_out_total_size, self.p2pkh_standard_output_size),
+        ) + max(sapling_input_count, sapling_output_count);
+
+        (self.marginal_fee * max(logical_actions, self.grace_actions)).ok_or(BalanceError::Overflow)
```
str4d, line 99: Nit: to more closely match ZIP 317:
```suggestion
        (self.marginal_fee * max(self.grace_actions, logical_actions)).ok_or(BalanceError::Overflow)
```

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -0,0 +1,173 @@
+//! Change strategies designed to implement the ZIP 317 fee rules.
+//!
+//! Change selection in ZIP 317 requires careful handling of low-valued inputs
+//! to ensure that inputs added to a transaction do
```
str4d, line 4: Unfinished sentence.

str4d (changes requested):
Partial review of bac6024f9e8b7fc93e2edd42a0b593829e53a299 (finished everything except `zcash_client_backend::fees::zip317`).

zcash_primitives/src/transaction/fees/zip317.rs
nuttycom: I mean, this isn't a bug, this is a known not-fully-implemented.

nuttycom: Resolved by making passing anything but a p2pkh input an error.

zcash_client_sqlite/src/wallet/transact.rs
```diff
@@ -832,4 +838,113 @@ mod tests {
...
+                test_prover(),
+                &input_selector,
+                &usk,
+                req,
+                OvkPolicy::Sender,
+                1,
+            ),
+            Err(Error::InsufficientFunds { available, required })
+                if available == Amount::from_u64(51000).unwrap()
+                && required == Amount::from_u64(60000).unwrap()
```
nuttycom, line 921: A note here: we do not yet do anything to indicate that notes have been rejected as dust when performing wallet balance calculation. We might want to do that.

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -0,0 +1,336 @@
...
+                .grace_actions()
+                .saturating_sub(t_non_dust)
+                .saturating_sub(s_non_dust);
+
+            let mut t_disallowed_dust = transparent_dust.len().saturating_sub(t_allowed_dust);
+            let mut s_disallowed_dust = sapling_dust.len().saturating_sub(s_allowed_dust);
+
+            if available_grace_inputs > 0 {
+                // if we have available grace inputs, allocate them first to transparent dust
+                // and then to sapling dust
```
str4d, line 107: To clarify, the assumption here is that if any transparent inputs have been provided to this `ChangeStrategy`, the caller is okay with any of those inputs being spent, so we don't need to consider the privacy implications here. It might be useful to call these out in a comment.

str4d (approved):
utACK 73ab884073183cf7a87e05ddf5e303b88c761cf0

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -0,0 +1,336 @@
...
+
+                let s_grace_dust = available_grace_inputs
+                    .saturating_sub(t_grace_dust)
+                    .saturating_sub(s_disallowed_dust);
+                s_disallowed_dust = s_disallowed_dust.saturating_sub(s_grace_dust);
+            }
+
+            // truncate the lists of inputs to be disregarded in input selection to just the
+            // disallowed lengths this has the effect of prioritizing inputs for inclusion by the
+            // order of th original input slices, with the most preferred inputs first
```
str4d, line 119:
```suggestion
            // Truncate the lists of inputs to be disregarded in input selection to just the
            // disallowed lengths. This has the effect of prioritizing inputs for inclusion by the
            // order of the original input slices, with the most preferred inputs first.
```

zcash_client_backend/src/fees/zip317.rs
str4d, line 107:
```suggestion
                // If we have available grace inputs, allocate them first to transparent dust
                // and then to Sapling dust. The caller has provided inputs that it is willing
                // to spend, so we don't need to consider privacy effects at this layer.
```

daira (commented):
Post-hoc ACK

PR #695, opened 2022-11-04
Fix clippy complaints about generated protobuf code.

nuttycom:
Clippy currently fails on all PRs based on `main` due to linter errors in the generated protobuf code.


str4d (approved):
utACK fb8681adbb601f66580db7d549ac2244d26ac6b3

PR #700, opened 2022-11-11
Release `zcash_client_sqlite 0.4.0` and dependencies

### LINKED ISSUE #487: Publish `zcash_client_backend 0.6.0`
We likely won't publish this until we are ready to publish `zcash_client_sqlite 0.4.0`, in case the latter requires changes to this crate.

### LINKED ISSUE #488: Publish `zcash_client_sqlite 0.4.0`

### LINKED ISSUE #497: Ensure that `decrypt_and_store_transaction` only persists data for transactions that involve the wallet.
In #341, `zcash_client_backend::data_api::wallet::decrypt_and_store_transaction` was modified to persist transactions if any transparent inputs are present. While this is presently safe because we expect this method to only be called for transactions that we have already detected are related to the wallet, it is not safe for general use and should be modified to perform this check internally. See https://github.com/zcash/librustzcash/pull/341#discussion_r794960501

str4d:
Closes #487.
Closes #488.
Closes #497.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -85,7 +85,7 @@ and this library adheres to Rust's notion of
 ### Changed
 - MSRV is now 1.56.1.
 - Bumped dependencies to `ff 0.12`, `group 0.12`, `bls12_381 0.7`
-  `zcash_primitives 0.8`, `orchard 0.3`.
+  `zcash_primitives 0.9`, `orchard 0.3`.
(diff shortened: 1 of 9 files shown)


sellout (approved):
utACK

It looks like `zcash_extensions` doesn’t have a CHANGELOG update because it hasn’t been released yet. Just wanted to call that out here.

daira (commented):
Post-hoc ACK

PR #701, opened 2022-11-12
Update generated protobufs

str4d:
`prost 0.11.2` added code formatting support, so the generated files now better match what rustfmt produces.


nuttycom (approved):
utACK

Issue #705, opened 2022-11-17
add `limit` to `validate_chain`


pacu:
Current:
````Rust
pub fn validate_chain<ParamsT, BlockSourceT>(
    parameters: &ParamsT,
    block_source: &BlockSourceT,
    validate_from: Option<(BlockHeight, BlockHash)>,
) -> Result<(), Error<Infallible, BlockSourceT::Error, Infallible>>
````

suggestion:

````Rust
pub fn validate_chain<ParamsT, BlockSourceT>(
    parameters: &ParamsT,
    block_source: &BlockSourceT,
    validate_from: Option<(BlockHeight, BlockHash)>,
    limit: Option<u32>,
) -> Result<(), Error<Infallible, BlockSourceT::Error, Infallible>>
````

### call is unable to be stopped in a timely fashion neither limited

wallets being out of date (most wallets currently are thanks to the increase load problem) take a lot of time to validate the combined chain if the cache is worth a couple hundred thousand blocks.  `validate_chain` takes several seconds and can't be stopped.

### validation can't be broken into smaller batches for better parallelization

Download is faster than scanning. Cache will be filled and scan will follow slowly. It is useful to be able to limit the range to validate since the scan batch is maybe a 100 blocks and `validate_chain` would validate the whole cache, when reorgs will appear only at chaintip.

PR #706, opened 2022-11-17
Adds `limit` parameter to `validate_chain()`

pacu:
Closes zcash/librustzcash#705

This change adds a `limit` parameter to `validate_chain()` function and also makes the `validate_from` parameters non-optional to avoid falling back to `sapling_activation_height` which is an implicit default parameter that is visible to the caller and I consider it hurts the "locality" of the function itself. 

I have empirical reasons to suspect that these assumptions on DB state don’t hold up and it can trigger a DB rewind on clients using validate_chain without specifying a concrete `validate_from` value and that might make the clients delete the whole database.

If you pass None to the validate_from  parameter to a BlockSourceT that has blocks on it and does not start in sapling_activation  this will fail with a weird state that would signal a gap of sapling_activation_height - 1 and first_existing_block_in_block_source regardless of it’s valid state.

I might have seen this bug in action in debug environments.

I also cleaned up the side-effects of the Optional `validate_from` parameter which results in the removal of the `ParamsT` argument since it's no longer needed because there is no more fallback to `sapling_activation_height`  when no value is provided. The `if () else` structure resulting of this change as been simplified as well. 

I only have one doubt I'd like to run by @str4d and @nuttycom:  Is it possible that there's no `prev_hash` to be provided?

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -25,26 +25,21 @@ and this library adheres to Rust's notion of
   - `validate_chain` now requires a non-optional `validate_from` parameter that
     indicates the starting point of the `BlockSourceT` validation. An Optional
     `limit` can be specified as well.  This allows callers to validate smaller
-    intervals of the given `BlockSourceT` shortening processing times of the
-    function call at the expense of obtaining a partial result on a given
-    section of interest of the block source. 
+    intervals of blocks already present on the provided `BlockSource` shortening
+    processing times of the function call at the expense of obtaining a partial
+    result. 
 
   - `params: &ParamsT` has been removed from the arguments since they were only
     needed to fall back to `sapling_activation_height` when `None` as passed as
-    the `validate_from` argument. Passing `None` as validation start point on a
-    pre-populated `block_source` would result in an error
-    `ChainError::block_height_discontinuity(sapling_activation_height - 1, current_height)` 
+    the `validate_from` argument. Implementors of `BlockSource` are resposible
+    of definining how they will fall back to a suitable value when `validate_from`
+    is `None`. 
 
-  - With this new API callers must specify a concrete `validate_from` argument
-    and assume that `validate_chain` will not take any default fallbacks to
-    chain `ParamsT`.
-
-  - The addition of a `limit` to the chain validation function changes the
-    meaning of its successful output, being now a `BlockHeight, BlockHash)`
-    tuple indicating the block height and block has up to which the chain as
-    been validated on its continuity of heights and hashes. Callers providing a
-    `limit` aregumente are responsible of subsequent calls to
-    `validate_chain()` to complete validating the totality of the block_source. 
+  - When passing a `limit` to the chain validation function, a successful output,
+    indicates that the chain has been validated on its continuity of heights and hashes
+    within the limits of the range `[validate_from, validate_from + limit)`. 
+    Callers providing a `limit` argument are responsible for making subsequent calls to
+    `validate_chain()` in order to complete validating the totality of the block_source. 
 
 ### Removed 
 - `zcash_client_backend::data_api`:
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -9,20 +9,47 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.60.0.
+
 - `zcash_client_backend::data_api::wallet::shield_transparent_funds` now
   takes a `shielding_threshold` argument that can be used to specify the 
   minimum value allowed as input to a shielding transaction. Previously 
   the shielding threshold was fixed at 100000 zatoshis.
+
 - Note commitments now use
   `zcash_primitives::sapling::note::ExtractedNoteCommitment` instead of
   `bls12_381::Scalar` in the following places:
   - The `cmu` field of `zcash_client_backend::wallet::WalletShieldedOutput`.
   - `zcash_client_backend::proto::compact_formats::CompactSaplingOutput::cmu`.
 
+- **breaking changes** to `zcash_client_backend::data_api::chain::validate_chain`
+  - `validate_chain` now requires a non-optional `validate_from` parameter that
+    indicates the starting point of the `BlockSourceT` validation. An Optional
+    `limit` can be specified as well.  This allows callers to validate smaller
+    intervals of the given `BlockSourceT` shortening processing times of the
+    function call at the expense of obtaining a partial result on a given
+    section of interest of the block source. 
+
+  - `params: &ParamsT` has been removed from the arguments since they were only
+    needed to fall back to `sapling_activation_height` when `None` as passed as
+    the `validate_from` argument. Passing `None` as validation start point on a
+    pre-populated `block_source` would result in an error
+    `ChainError::block_height_discontinuity(sapling_activation_height - 1, current_height)` 
+
+  - With this new API callers must specify a concrete `validate_from` argument
+    and assume that `validate_chain` will not take any default fallbacks to
+    chain `ParamsT`.
+
+  - The addition of a `limit` to the chain validation function changes the
+    meaning of its successful output, being now a `BlockHeight, BlockHash)`
+    tuple indicating the block height and block has up to which the chain as
+    been validated on its continuity of heights and hashes. Callers providing a
+    `limit` aregumente are responsible of subsequent calls to
+    `validate_chain()` to complete validating the totality of the block_source. 
 
 ### Removed 
 - `zcash_client_backend::data_api`:
   - `WalletWrite::remove_unmined_tx` (was behind the `unstable` feature flag).
+
 ## [0.6.1] - 2022-12-06
 ### Added
 - `zcash_client_backend::data_api::chain::scan_cached_blocks` now generates
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -88,7 +88,7 @@ use std::convert::Infallible;
 
 use zcash_primitives::{
     block::BlockHash,
-    consensus::{self, BlockHeight, NetworkUpgrade},
+    consensus::{self, BlockHeight},
     merkle_tree::CommitmentTree,
     sapling::{note_encryption::PreparedIncomingViewingKey, Nullifier},
     zip32::Scope,
@@ -111,7 +111,8 @@ pub trait BlockSource {
     type Error;
 
     /// Scan the specified `limit` number of blocks from the blockchain, starting at
-    /// `from_height`, applying the provided callback to each block.
+    /// `from_height`, applying the provided callback to each block. If `from_height`
+    /// is `None` then scanning will begin at the first available block.
     ///
     /// * `WalletErrT`: the types of errors produced by the wallet operations performed
     ///   as part of processing each row.
@@ -119,7 +120,7 @@ pub trait BlockSource {
     ///   reporting errors related to specific notes.
     fn with_blocks<F, WalletErrT, NoteRefT>(
         &self,
-        from_height: BlockHeight,
+        from_height: Option<BlockHeight>,
         limit: Option<u32>,
         with_row: F,
     ) -> Result<(), error::Error<WalletErrT, Self::Error, NoteRefT>>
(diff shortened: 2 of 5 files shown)


zcash_client_backend/src/proto/compact_formats.rs
```diff
@@ -1,32 +1,29 @@
...
-
 /// CompactBlock is a packaging of ONLY the data from a block that's needed to:
 ///    1. Detect a payment to your shielded Sapling address
 ///    2. Detect a spend of your shielded Sapling notes
 ///    3. Update your witnesses to generate new Sapling spend proofs.
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct CompactBlock {
     /// the version of this wire format, for storage
-    #[prost(uint32, tag="1")]
+    #[prost(uint32, tag = "1")]
```
nuttycom, line 8: These changes shouldn't appear in this PR; can you rebase?

nuttycom (changes requested):
Looks like this PR is incomplete? Where is the change to `BlockSource::with_blocks`?

pacu:
I think something's wrong with the copy of my repo. I'm drafting this. please hold...

pacu:
Fixed. It's ready for review now

str4d:
CI is failing because a doctest needs updating.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -44,7 +44,7 @@
 //! // Given that we assume the server always gives us correct-at-the-time blocks, any
 //! // errors are in the blocks we have previously cached or scanned.
 //! let max_height_hash = db_data.get_max_height_hash().map_err(Error::Wallet)?;
-//! if let Err(e) = validate_chain(&network, &block_source, max_height_hash) {
+//! if let Err(e) = validate_chain(&network, &block_source, max_height_hash, limit) {
```
str4d, line 47: `limit` needs to be defined somewhere in the doctest source, or a constant provided like
```suggestion
//! if let Err(e) = validate_chain(&network, &block_source, max_height_hash, None) {
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -140,6 +140,8 @@ pub trait BlockSource {
 /// - `block_source` Source of compact blocks
 /// - `from_tip` Height & hash of last validated block; if no validation has previously
 ///    been performed, this will begin scanning from `sapling_activation_height - 1`
+/// - `limit` amount of blocks that will be fetched from `block_source` and attempt to 
+///    be validated
 ///
 /// Returns:
 /// - `Ok(())` if the combined chain is valid.
```
str4d, line 147: This case needs more explanation given the presence of `limit` (as the chain is not valid any more, but only "valid up to the given limit height").

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -140,6 +140,8 @@ pub trait BlockSource {
 /// - `block_source` Source of compact blocks
 /// - `from_tip` Height & hash of last validated block; if no validation has previously
 ///    been performed, this will begin scanning from `sapling_activation_height - 1`
+/// - `limit` amount of blocks that will be fetched from `block_source` and attempt to 
+///    be validated
```
str4d, line 144:
```suggestion
/// - `limit`: Maximum number of blocks that should be validated in this call.
```

The documentation also needs to explain that if `limit` is provided, then the caller is responsible for repeatedly calling this function until no more validation occurs. It's not immediately clear to me whether the caller can do this however, or if the return value of this function needs to be altered to return the `(height, hash)` of the last validated block (so that the caller can use this on the next call, and terminate once the returned tuple equals `from`).

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -151,6 +153,7 @@ pub fn validate_chain<ParamsT, BlockSourceT>(
     parameters: &ParamsT,
     block_source: &BlockSourceT,
     validate_from: Option<(BlockHeight, BlockHash)>,
+    limit: Option<u32>,
```
str4d, line 156: This is a change to a public API; it needs to be documented as a change in the changelog.

str4d (dismissed):
Reviewed 0c8d2092159b6a1ba4b0370dc57b3d37af1a2e61.

zcash_client_backend/src/data_api/chain.rs
pacu: I'm hesitating on how to solve this tuple as return value. Mainly because when limit provided is `None`, there's a fallback on sapling activation height, which is just a height and its has is an Optional.

```
 let from_height = validate_from
        .map(|(height, _)| height)
        .unwrap_or(sapling_activation_height - 1);
```

nuttycom:
`prev_hash` can always have a value, making it non-optional is fine, because we always know the hash of the last block prior to Sapling activation.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
```
nuttycom, line 10: Oops, this changelog entry should be in `zcash_client_backend`.

nuttycom (changes requested):
utACK modulo moving the CHANGELOG entry to `zcash_client_backend`

nuttycom (approved):
utACK

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -134,59 +134,52 @@ pub trait BlockSource {
...
-        .activation_height(NetworkUpgrade::Sapling)
-        .expect("Sapling activation height must be known.");
-
     // The block source will contain blocks above the `validate_from` height.  Validate from that
     // maximum height up to the chain tip, returning the hash of the block found in the block
     // source at the `validate_from` height, which can then be used to verify chain integrity by
     // comparing against the `validate_from` hash.
-    let from_height = validate_from
-        .map(|(height, _)| height)
-        .unwrap_or(sapling_activation_height - 1);
```
str4d, line 169: The reason that `sapling_activation_height - 1` existed in the original chain validation code that I wrote, was as a way to ensure we checked all cached CompactBlocks. We could have equivalently used 0, or any other height that was guaranteed to never be seen by the mobile SDKs (since we designed them for Sapling, and thus would never send them pre-Sapling blocks). https://github.com/zcash/librustzcash/blob/7ced7f3c5669360e9d3cfbe5622552b40406cddc/zcash_client_sqlite/src/chain.rs#L102-L130

The current code before this PR is instead interpreting it as the actual height we should start at, due in part to how `BlockSource::with_blocks` is constructed and documented. Removing the default here just moves the problem to the caller (per my above comment); we need to solve this at `BlockSource`.

zcash_client_backend/CHANGELOG.md
```diff
@@ -7,6 +7,32 @@ and this library adheres to Rust's notion of
...
+  - With this new API callers must specify a concrete `validate_from` argument
+    and assume that `validate_chain` will not take any default fallbacks to
+    chain `ParamsT`.
+
+  - The addition of a `limit` to the chain validation function changes the
+    meaning of its successful output, being now a `BlockHeight, BlockHash)`
+    tuple indicating the block height and block has up to which the chain as
+    been validated on its continuity of heights and hashes. Callers providing a
+    `limit` aregumente are responsible of subsequent calls to
+    `validate_chain()` to complete validating the totality of the block_source.
```
str4d, line 34: This needs rewriting, but given the other API changes needed, I'll hold off suggestions until they are done.

str4d (changes requested):

zcash_client_backend/src/data_api/chain.rs
nuttycom: The safest thing to do here is to query the wallet for the latest scanned height each time. For that reason, I've changed this back to returning `Result<(), ...>`.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -134,59 +134,52 @@ pub trait BlockSource {
...
+/// of heights and hashes. Callers providing a `limit` argument are responsible of
+/// making subsequent calls to `validate_chain()` to complete validating the remaining
+/// blocks stored on the `block_source`.
 ///
 /// Arguments:
-/// - `parameters` Network parameters
 /// - `block_source` Source of compact blocks
-/// - `from_tip` Height & hash of last validated block; if no validation has previously
-///    been performed, this will begin scanning from `sapling_activation_height - 1`
+/// - `validate_from` Height & hash of last validated block;
```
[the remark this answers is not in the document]
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/706/commits/4fa67fe00f889a5c2a0e4cf7c6cae4d59ea9a5d7

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -134,59 +134,52 @@ pub trait BlockSource {
 /// block source is more likely to be accurate than the previously-scanned information.
 /// This follows from the design (and trust) assumption that the `lightwalletd` server
 /// provides accurate block information as of the time it was requested.
+/// The addition of a `limit` to the chain validation function changes the meaning of
+/// its successful output, being now a `BlockHeight, BlockHash)` tuple indicating the
+/// block height and block hash up to which the chain as been validated on its continuity
+/// of heights and hashes. Callers providing a `limit` argument are responsible of
+/// making subsequent calls to `validate_chain()` to complete validating the remaining
+/// blocks stored on the `block_source`.
```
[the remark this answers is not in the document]
pacu: Changed this in 533c75fc8193374118c05944da71d30bc460b9de

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -134,59 +134,52 @@ pub trait BlockSource {
...
+/// - `limit` Maximum number of blocks that should be validated in this call.
 ///
 /// Returns:
-/// - `Ok(())` if the combined chain is valid.
+/// - `Ok((BlockHeight, BlockHash))` if the combined chain is valid up to the given height
+/// and block hash.
 /// - `Err(Error::Chain(cause))` if the combined chain is invalid.
 /// - `Err(e)` if there was an error during validation unrelated to chain validity.
 ///
 /// This function does not mutate either of the databases.
```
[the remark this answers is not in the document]
pacu: Fixed in b031812238496224e4197946da20026bad8faa55

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -134,59 +134,52 @@ pub trait BlockSource {
...
 /// - `block_source` Source of compact blocks
-/// - `from_tip` Height & hash of last validated block; if no validation has previously
-///    been performed, this will begin scanning from `sapling_activation_height - 1`
+/// - `validate_from` Height & hash of last validated block;
+/// - `limit` Maximum number of blocks that should be validated in this call.
 ///
 /// Returns:
-/// - `Ok(())` if the combined chain is valid.
+/// - `Ok((BlockHeight, BlockHash))` if the combined chain is valid up to the given height
+/// and block hash.
```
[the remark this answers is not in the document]
pacu: Current code no longer returns tuple

zcash_client_backend/src/data_api/chain.rs
pacu: I think this is now being addressed by `blockdb_with_blocks` implementation

````Rust
let mut stmt_blocks = block_source
        .0
        .prepare(
            "SELECT height, data FROM compactblocks 
            WHERE height > ? 
            ORDER BY height ASC LIMIT ?",
        )
        .map_err(to_chain_error)?;

    let mut rows = stmt_blocks
        .query(params![
            last_scanned_height.map_or(0u32, u32::from),
            limit.unwrap_or(u32::max_value()),
        ])
        .map_err(to_chain_error)?;
````

Where the lower bound of the block source falls back to zero if `None` is provided

pacu: Equivalent implementation can be found in `pub(crate) fn fsblockdb_with_blocks<F, DbErrT, NoteRef>` for the FsBlock block source implementation

zcash_client_backend/CHANGELOG.md
pacu: Rewrote on 66e4df3eb182d416c287e75d1dfc2f56e3e266c0  and 58251c41a6ba1ef75ae45141e5ebc566cce3c6fa

nuttycom (approved):
utACK

nuttycom (approved):
re-utACK after rebasing to resolve conflicts.

daira (approved):
utACK

str4d (approved):
utACK

PR #708, opened 2022-11-18
Parse ZIP 321 URIs without query parameters

### LINKED ISSUE #707: Fails to parse ZIP 321 URIs without a query string
Coinbase URIs don’t include query parameters (or the `?` before them), which is
valid. But zcash_client_backend errors when it encounters one.

sellout:
Fixes #707.


nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK

PR #725, opened 2022-12-08
Fix transparent balance error when the wallet is empty.

### LINKED ISSUE #724: `get_transparent_balances` fails with SQLite error when there are rows on UTXOS table
`get_transparent_balances` fails with SQLite error when there are rows on UTXOS table

````
 RustWeldingError
 ▿ genericError : 1 element
  - message : "Error while fetching transparent balances for AccountId(0): Invalid column type Null at index: 0, name: address"
````

if you initialize a wallet and right out of the box query the get_transparent_balances from walletDb, it will throw that error.

The error does not happen anymore if you insert a UTXO

nuttycom:
Fixes #724


str4d (approved):
utACK

PR #732, opened 2022-12-09
Fix clippy complaints.

nuttycom:


zcash_client_backend/src/welding_rig.rs
```diff
@@ -191,6 +191,7 @@ pub(crate) fn add_block_to_runner<P, S, T>(
 }
 
 #[tracing::instrument(skip_all, fields(height = block.height))]
+#[allow(clippy::suspicious_else_formatting)]
```
str4d, line 194: Clippy is still failing because this doesn't work. The reason is that `tracing` is _already_ handling this issue with an `allow` attribute, but there is a second instance of the issue nested inside the first one, and our annotation is insulated from the second instance by their fix of the first instance. I've opened https://github.com/tokio-rs/tracing/issues/2410 to track this.

```suggestion
```

str4d (approved):
utACK after fixing the commit to remove the non-functional attribute.

zcash_client_backend/src/welding_rig.rs
nuttycom: Aw. So we don't actually get green clippy lints until that's fixed. :(

PR #733, opened 2022-12-11
Update generated protobuf bindings

str4d:


str4d (commented):
Self-ACK (only adds clippy lint allows to generated bindings).

nuttycom (commented):
post-hoc utACK

PR #734, opened 2022-12-13
Improve Sapling type safety

### LINKED ISSUE #739: Introduce type-safe value types for Sapling
This is the Sapling equivalent of the `orchard::value` module.

### LINKED ISSUE #740: Make Sapling transaction component types opaque
Their fields should not be `pub`.

str4d:
Closes #739.
Closes #740.

### DIFF
--- a/zcash_primitives/src/sapling/value.rs
+++ b/zcash_primitives/src/sapling/value.rs
@@ -1,0 +1,243 @@
+//! Monetary values within the Sapling shielded pool.
+//!
+//! Values are represented in three places within the Sapling protocol:
+//! - [`NoteValue`], the value of an individual note. It is an unsigned 64-bit integer
+//!   (with maximum value [`MAX_NOTE_VALUE`]), and is serialized in a note plaintext.
+//! - [`ValueSum`], the sum of note values within a Sapling [`Bundle`]. It is represented
+//!   as an `i128` and places an upper bound on the maximum number of notes within a
+//!   single [`Bundle`].
+//! - `valueBalanceSapling`, which is a signed 63-bit integer. This is represented
+//!   by a user-defined type parameter on [`Bundle`], returned by
+//!   [`Bundle::value_balance`] and [`SaplingBuilder::value_balance`].
+//!
+//! If your specific instantiation of the Sapling protocol requires a smaller bound on
+//! valid note values (for example, Zcash's `MAX_MONEY` fits into a 51-bit integer), you
+//! should enforce this in two ways:
+//!
+//! - Define your `valueBalanceSapling` type to enforce your valid value range. This can
+//!   be checked in its `TryFrom<i64>` implementation.
+//! - Define your own "amount" type for note values, and convert it to `NoteValue` prior
+//!   to calling [`SaplingBuilder::add_output`].
+//!
+//! Inside the circuit, note values are constrained to be unsigned 64-bit integers.
+//!
+//! # Caution!
+//!
+//! An `i64` is _not_ a signed 64-bit integer! The [Rust documentation] calls `i64` the
+//! 64-bit signed integer type, which is true in the sense that its encoding in memory
+//! takes up 64 bits. Numerically, however, `i64` is a signed 63-bit integer.
+//!
+//! Fortunately, users of this crate should never need to construct [`ValueSum`] directly;
+//! you should only need to interact with [`NoteValue`] (which can be safely constructed
+//! from a `u64`) and `valueBalanceSapling` (which can be represented as an `i64`).
+//!
+//! [`Bundle`]: crate::transaction::components::sapling::Bundle
+//! [`Bundle::value_balance`]: crate::transaction::components::sapling::Bundle::value_balance
+//! [`SaplingBuilder::value_balance`]: crate::transaction::components::sapling::builder::SaplingBuilder::value_balance
+//! [`SaplingBuilder::add_output`]: crate::transaction::components::sapling::builder::SaplingBuilder::add_output
+//! [Rust documentation]: https://doc.rust-lang.org/stable/std/primitive.i64.html
+
+use ff::Field;
+use group::GroupEncoding;
+use rand::RngCore;
+use subtle::CtOption;
+
+use crate::constants::{VALUE_COMMITMENT_RANDOMNESS_GENERATOR, VALUE_COMMITMENT_VALUE_GENERATOR};
+
+mod sums;
+pub use sums::{CommitmentSum, OverflowError, TrapdoorSum, ValueSum};
+
+/// Maximum note value.
+pub const MAX_NOTE_VALUE: u64 = u64::MAX;
+
+/// The non-negative value of an individual Sapling note.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct NoteValue(u64);
+
+impl NoteValue {
+    /// Returns the raw underlying value.
+    pub fn inner(&self) -> u64 {
+        self.0
+    }
+
+    /// Creates a note value from its raw numeric value.
+    ///
+    /// This only enforces that the value is an unsigned 64-bit integer. Callers should
+    /// enforce any additional constraints on the value's valid range themselves.
+    pub fn from_raw(value: u64) -> Self {
+        NoteValue(value)
+    }
+}
+
+/// The blinding factor for a [`ValueCommitment`].
+#[derive(Clone, Debug)]
+pub struct ValueCommitTrapdoor(jubjub::Scalar);
+
+impl ValueCommitTrapdoor {
+    /// Generates a new value commitment trapdoor.
+    ///
+    /// This is public for access by `zcash_proofs`.
+    pub fn random(rng: impl RngCore) -> Self {
+        ValueCommitTrapdoor(jubjub::Scalar::random(rng))
+    }
+
+    /// Returns the inner Jubjub scalar representing this trapdoor.
+    ///
+    /// This is public for access by `zcash_proofs`.
+    pub fn inner(&self) -> jubjub::Scalar {
+        self.0
+    }
+}
+
+/// A commitment to a [`ValueSum`].
+///
+/// # Consensus rules
+///
+/// The Zcash Protocol Spec requires Sapling Spend Descriptions and Output Descriptions to
+/// not contain a small order `ValueCommitment`. However, the `ValueCommitment` type as
+/// specified (and implemented here) may contain a small order point. In practice, it will
+/// not occur:
+/// - [`ValueCommitment::derive`] will only produce a small order point if both the given
+///   [`NoteValue`] and [`ValueCommitTrapdoor`] are zero. However, the only constructor
+///   available for `ValueCommitTrapdoor` is [`ValueCommitTrapdoor::random`], which will
+///   produce zero with negligible probability (assuming a non-broken PRNG).
+/// - [`ValueCommitment::from_bytes_not_small_order`] enforces this by definition, and is
+///   the only constructor that can be used with data received over the network.
+#[derive(Clone, Debug)]
+pub struct ValueCommitment(jubjub::ExtendedPoint);
+
+impl ValueCommitment {
+    /// Derives a `ValueCommitment` by $\mathsf{ValueCommit^{Sapling}}$.
+    ///
+    /// Defined in [Zcash Protocol Spec § 5.4.8.3: Homomorphic Pedersen commitments (Sapling and Orchard)][concretehomomorphiccommit].
+    ///
+    /// [concretehomomorphiccommit]: https://zips.z.cash/protocol/protocol.pdf#concretehomomorphiccommit
+    pub fn derive(value: NoteValue, rcv: ValueCommitTrapdoor) -> Self {
+        let cv = (VALUE_COMMITMENT_VALUE_GENERATOR * jubjub::Scalar::from(value.0))
+            + (VALUE_COMMITMENT_RANDOMNESS_GENERATOR * rcv.0);
+
+        ValueCommitment(cv.into())
+    }
+
+    /// Returns the inner Jubjub point representing this value commitment.
+    ///
+    /// This is public for access by `zcash_proofs`.
+    pub fn as_inner(&self) -> &jubjub::ExtendedPoint {
+        &self.0
+    }
+
+    /// Deserializes a value commitment from its byte representation.
+    ///
+    /// Returns `None` if `bytes` is an invalid representation of a Jubjub point, or the
+    /// resulting point is of small order.
+    ///
+    /// This method can be used to enforce the "not small order" consensus rules defined
+    /// in [Zcash Protocol Spec § 4.4: Spend Descriptions][spenddesc] and
+    /// [§ 4.5: Output Descriptions][outputdesc].
+    ///
+    /// [spenddesc]: https://zips.z.cash/protocol/protocol.pdf#spenddesc
+    /// [outputdesc]: https://zips.z.cash/protocol/protocol.pdf#outputdesc
+    pub fn from_bytes_not_small_order(bytes: &[u8; 32]) -> CtOption<ValueCommitment> {
+        jubjub::ExtendedPoint::from_bytes(bytes)
+            .and_then(|cv| CtOption::new(ValueCommitment(cv), !cv.is_small_order()))
+    }
+
+    /// Serializes this value commitment to its canonical byte representation.
+    pub fn to_bytes(&self) -> [u8; 32] {
+        self.0.to_bytes()
+    }
+}
+
+/// Generators for property testing.
+#[cfg(any(test, feature = "test-dependencies"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "test-dependencies")))]
+pub mod testing {
+    use proptest::prelude::*;
+
+    use super::{NoteValue, ValueCommitTrapdoor, MAX_NOTE_VALUE};
+
+    prop_compose! {
+        /// Generate an arbitrary value in the range of valid nonnegative amounts.
+        pub fn arb_note_value()(value in 0u64..MAX_NOTE_VALUE) -> NoteValue {
+            NoteValue(value)
+        }
+    }
+
+    prop_compose! {
+        /// Generate an arbitrary value in the range of valid positive amounts less than a
+        /// specified value.
+        pub fn arb_note_value_bounded(max: u64)(value in 0u64..max) -> NoteValue {
+            NoteValue(value)
+        }
+    }
+
+    prop_compose! {
+        /// Generate an arbitrary value in the range of valid positive amounts less than a
+        /// specified value.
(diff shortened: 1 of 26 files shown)


zcash_proofs/src/sapling/verifier.rs
```diff
@@ -33,18 +34,19 @@ impl SaplingVerificationContextInner {
...
+        spend_auth_sig: &Signature,
         zkproof: Proof<Bls12>,
         verifier_ctx: &mut C,
-        spend_auth_sig_verifier: impl FnOnce(&mut C, PublicKey, [u8; 64], Signature) -> bool,
+        spend_auth_sig_verifier: impl FnOnce(&mut C, &PublicKey, [u8; 64], &Signature) -> bool,
         proof_verifier: impl FnOnce(&mut C, Proof<Bls12>, [bls12_381::Scalar; 7]) -> bool,
     ) -> bool {
-        if (cv.is_small_order() | rk.0.is_small_order()).into() {
+        // `cv` is not small order by construction.
+        if rk.0.is_small_order().into() {
```
str4d, line 49: This is consensus-critical code, so the PR requires 3 reviews. My intention in this refactor was to encapsulate this consensus check inside the `ValueCommitment` type (which is partly why I added a separate `CommitmentSum` type, to allow the latter to hold the value "zero", which would fail this check if we did `cv` summing with the same type as in `orchard`).

zcash_proofs/src/sapling/verifier.rs
```diff
@@ -100,13 +102,14 @@ impl SaplingVerificationContextInner {
...
-        cv: jubjub::ExtendedPoint,
+        cv: &ValueCommitment,
         cmu: bls12_381::Scalar,
         epk: jubjub::ExtendedPoint,
         zkproof: Proof<Bls12>,
         proof_verifier: impl FnOnce(Proof<Bls12>, [bls12_381::Scalar; 5]) -> bool,
     ) -> bool {
-        if (cv.is_small_order() | epk.is_small_order()).into() {
+        // `cv` is not small order by construction.
+        if epk.is_small_order().into() {
```
str4d, line 112: Ditto.

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -83,13 +83,52 @@ impl MapAuth<Authorized, Authorized> for () {
...
+        }
+    }
+
+    /// Returns the list of spends in this bundle.
+    pub fn shielded_spends(&self) -> &[SpendDescription<A>] {
+        &self.shielded_spends
+    }
+
+    /// Returns the list of outputs in this bundle.
+    pub fn shielded_outputs(&self) -> &[OutputDescription<A::Proof>] {
```
AloeareV, line 114: Zingolib's test framework constructs mock bundles, and adds outputs to mock bundles, could we get a getter fn for a `mut &` to the vec of outputs (probably behind a `#[cfg(feature = "test-dependences")]`), and public access to `from_parts`? Probably something like the `OutputDescription` test impl?

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -282,12 +353,79 @@ impl SpendDescriptionV5 {
...
+    pub(crate) fn cmu_mut(&mut self) -> &mut bls12_381::Scalar {
+        &mut self.cmu
+    }
+    pub(crate) fn ephemeral_key_mut(&mut self) -> &mut EphemeralKeyBytes {
+        &mut self.ephemeral_key
+    }
+    pub(crate) fn enc_ciphertext_mut(&mut self) -> &mut [u8; 580] {
+        &mut self.enc_ciphertext
+    }
+    pub(crate) fn out_ciphertext_mut(&mut self) -> &mut [u8; 80] {
```
AloeareV, line 426: Zingolib's test framework also uses these. Could we
```suggestion
#[cfg(any(test, feature = "test-dependencies"))]
impl<Proof> OutputDescription<Proof> {
    pub fn from_parts(
        cv: ValueCommitment,
        cmu: bls12_381::Scalar,
        ephemeral_key: EphemeralKeyBytes,
        enc_ciphertext: [u8; 580],
        out_ciphertext: [u8; 80],
        zkproof: Proof,
    ) -> Self {
        OutputDescription {
            cv,
            cmu,
            ephemeral_key,
            enc_ciphertext,
            out_ciphertext,
            zkproof,
        }
    }
    pub fn cv_mut(&mut self) -> &mut ValueCommitment {
        &mut self.cv
    }
    pub fn cmu_mut(&mut self) -> &mut bls12_381::Scalar {
        &mut self.cmu
    }
    pub fn ephemeral_key_mut(&mut self) -> &mut EphemeralKeyBytes {
        &mut self.ephemeral_key
    }
    pub fn enc_ciphertext_mut(&mut self) -> &mut [u8; 580] {
        &mut self.enc_ciphertext
    }
    pub fn out_ciphertext_mut(&mut self) -> &mut [u8; 80] {
```

AloeareV (changes requested):
The rest of the public-facing changes that affect zingolabs should be straightforward to adapt to.

zcash_primitives/src/transaction/components/sapling.rs
str4d: Perhaps instead we should have an API for constructing these mock bundles. We do already have `MockTxProver` which means you can use `SaplingBuilder` to create fake Sapling bundles with invalid proofs while everything else is valid. But that doesn't let you construct arbitrarily invalid bundles (as this would necessarily violate type safety).

Can you link me to how you currently do this?

AloeareV: We create one here: https://github.com/zingolabs/zingolib/blob/8bcad1020357047b1fbe4915e102868a45aa0280/lib/src/lightclient/testmocks.rs#L9-L14 and https://github.com/zingolabs/zingolib/blob/6ad127e73a41f7bfd612597677b292437812f7d1/lib/src/blaze/test_utils.rs#L119-L124

And we add outputs here: https://github.com/zingolabs/zingolib/blob/6ad127e73a41f7bfd612597677b292437812f7d1/lib/src/blaze/test_utils.rs#L217

zcash_primitives/src/sapling/value/sums.rs
```diff
@@ -0,0 +1,220 @@
...
+impl std::error::Error for OverflowError {}
+
+/// A sum of Sapling note values.
+///
+/// [Zcash Protocol Spec § 4.13: Balance and Binding Signature (Sapling)][saplingbalance]
+/// constrains the range of this type to between `[-(r_J - 1)/2..(r_J - 1)/2]` in the
+/// abstract protocol, and `[−38913406623490299131842..104805176454780817500623]` in the
+/// concrete Zcash protocol. We represent it as an `i128`, which has a range large enough
+/// to handle Zcash transactions while small enough to ensure the abstract protocol bounds
+/// are not breached.
```
daira, line 28: :+1:

zcash_primitives/src/sapling/value/sums.rs
```diff
@@ -0,0 +1,220 @@
...
+        CommitmentSum(jubjub::ExtendedPoint::identity())
+    }
+
+    /// Transform this value commitment sum into the corresponding RedJubjub public key.
+    ///
+    /// This is public for access by `zcash_proofs`.
+    pub fn into_bvk<V: Into<i64>>(self, value_balance: V) -> Option<redjubjub::PublicKey> {
+        let value: i64 = value_balance.into();
+
+        // Compute the absolute value (failing if -i64::MAX is the value).
```
daira, line 155:
```suggestion
        // Compute the absolute value (failing if i64::MIN is the value).
```
`-i64::MAX` is wrong. Independently of that: i64::MIN cannot occur in Zcash but could occur in the abstract protocol omitting the block size limit, and it might be able to occur in some future version of Zcash with ZSAs. It is possible to just compute bvk correctly in this case; should we do that?

zcash_primitives/src/sapling/value/sums.rs
```diff
@@ -0,0 +1,220 @@
...
+        let abs_value = value
+            .checked_abs()
+            .map(|v| u64::try_from(v).expect("v is non-negative"))?;
```
daira, line 158:
```suggestion
            .map(|v| u64::try_from(v).expect("v is non-negative"))
            .ok_or(1u64 << 63);
```

zcash_primitives/src/sapling/value/sums.rs
```diff
@@ -0,0 +1,220 @@
...
+impl CommitmentSum {
+    /// Initializes a sum of `ValueCommitment`s to zero.
+    pub fn zero() -> Self {
+        CommitmentSum(jubjub::ExtendedPoint::identity())
+    }
+
+    /// Transform this value commitment sum into the corresponding RedJubjub public key.
+    ///
+    /// This is public for access by `zcash_proofs`.
+    pub fn into_bvk<V: Into<i64>>(self, value_balance: V) -> Option<redjubjub::PublicKey> {
```
daira, line 152:
```suggestion
    pub fn into_bvk<V: Into<i64>>(self, value_balance: V) -> redjubjub::PublicKey {
```
Consider this change as well if the change below is taken.

zcash_primitives/src/sapling/value/sums.rs
```diff
@@ -0,0 +1,220 @@
...
+        let value_balance = if value.is_negative() {
+            -jubjub::Scalar::from(abs_value)
+        } else {
+            jubjub::Scalar::from(abs_value)
+        };
+
+        // Subtract `value_balance` from the sum to get the final bvk.
+        let bvk = self.0 - VALUE_COMMITMENT_VALUE_GENERATOR * value_balance;
+
+        Some(redjubjub::PublicKey(bvk))
```
daira, line 170:
```suggestion
        redjubjub::PublicKey(bvk)
```

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -142,22 +181,54 @@ impl<A: Authorization> std::fmt::Debug for SpendDescription<A> {
...
+        &self.spend_auth_sig
+    }
+}
+
 /// Consensus rules (§4.4) & (§4.5):
 /// - Canonical encoding is enforced here.
 /// - "Not small order" is enforced in SaplingVerificationContext::(check_spend()/check_output())
 ///   (located in zcash_proofs::sapling::verifier).
-pub fn read_point<R: Read>(mut reader: R, field: &str) -> io::Result<jubjub::ExtendedPoint> {
+fn read_value_commitment<R: Read>(mut reader: R, field: &str) -> io::Result<ValueCommitment> {
```
daira, line 220: The `field` argument is redundant, since it is always `cv` (as implied now by the function name).

zcash_proofs/src/circuit/sapling.rs
```diff
@@ -28,10 +28,26 @@ use group::ff::PrimeFieldBits;
 
 pub const TREE_DEPTH: usize = SAPLING_COMMITMENT_TREE_DEPTH;
 
+/// The preimage of a Sapling value commitment.
+#[derive(Clone)]
+pub struct ValueCommitmentPreimage {
```
daira, line 33: Consider renaming this to `ValueCommitmentOpening`. (I was initially confused what this was when calling it a preimage. In particular, "opening" unambiguously includes the randomness.)

zcash_proofs/src/circuit/sapling.rs
```diff
@@ -28,10 +28,26 @@ use group::ff::PrimeFieldBits;
 
 pub const TREE_DEPTH: usize = SAPLING_COMMITMENT_TREE_DEPTH;
 
+/// The preimage of a Sapling value commitment.
```
daira, line 31:
```suggestion
/// The opening (value and randomness) of a Sapling value commitment.
```

zcash_proofs/src/circuit/sapling.rs
```diff
@@ -57,7 +73,7 @@ pub struct Spend {
 /// This is an output circuit instance.
 pub struct Output {
     /// Pedersen commitment to the value being spent
```
daira, line 75:
```suggestion
    /// The opening of a Pedersen commitment to the value being spent
```

zcash_primitives/src/sapling/value/sums.rs
daira: I also mildly prefer using a match as in https://github.com/zcash/librustzcash/pull/734/files#diff-09b75bb59427e9d8d8efcdb2c8033ec30a82fcd6e44e661da06d5929c77d895eL17 , i.e.
```rust
        let abs_value = match value.checked_abs() {
            Some(v) => u64::try_from(v).expect("v is non-negative"),
            None => 1u64 << 63,
        };
```
(because that makes it clearer that the error case can only come from the `checked_abs()` call).

zcash_proofs/src/sapling/prover.rs
```diff
@@ -56,27 +55,22 @@ impl SaplingProvingContext {
...
-            tmp.add_assign(&self.bsk);
-
-            // Update the context
-            self.bsk = tmp;
-        }
+        self.bsk += &rcv;
 
         // Construct the value commitment
-        let value_commitment = ValueCommitment {
+        let value_commitment_inputs = crate::circuit::sapling::ValueCommitmentPreimage {
```
daira, line 69: Consider renaming to `value_commitment_opening`.

daira: Also I suggest importing `ValueCommitmentPreimage` (or `ValueCommitmentOpening` if it's renamed). It is used more tab once in the file.

zcash_proofs/src/sapling/verifier.rs
daira: As far as I can see, `ValueCommitment::derive` can return a `ValueCommitment` for the zero point, so this is not equivalent. (`ValueCommitment::from_bytes` indeed does the small order check, but we can't tell by local code review whether `cv` was constructed by `derive` or `from_bytes`.)

zcash_primitives/src/sapling/value.rs
```diff
@@ -0,0 +1,228 @@
...
+        self.0
+    }
+}
+
+/// A commitment to a [`ValueSum`].
+#[derive(Clone, Debug)]
+pub struct ValueCommitment(jubjub::ExtendedPoint);
+
+impl ValueCommitment {
+    /// Derives a `ValueCommitment` by $\mathsf{ValueCommit^{Sapling}}$.
```
daira, line 97: If this is intended to exclude small-order points, please document that and enforce it in `derive`/`derive_inner`.

zcash_proofs/src/sapling/verifier.rs
daira: Aside: I wish we hadn't excluded the zero point here. It's ugly and unnecessary.

daira: Note that $\mathsf{ValueCommitment^{Sapling}.Output}$ is $\mathcal{J}$, i.e. it doesn't exclude the zero point. The exclusion of small-order points is in the descriptions of the $\mathsf{cv}$ fields of Spend descriptions in section 4.4, and Output descriptions in section 4.5.

zcash_proofs/src/sapling/verifier.rs
daira: Same problem as above.

zcash_proofs/src/sapling/verifier.rs
```diff
@@ -143,18 +146,12 @@ impl SaplingVerificationContextInner {
...
-        // Obtain current cv_sum from the context
-        let mut bvk = PublicKey(self.cv_sum);
-
-        // Compute value balance
-        let value_balance = match compute_value_balance(value_balance) {
+        // Compute the final bvk.
+        let bvk = match self.cv_sum.into_bvk(value_balance) {
             Some(a) => a,
             None => return false,
         };
```
daira, line 153: My suggestion to make `into_bvk` infallible would simplify this.

zcash_proofs/src/sapling/verifier/batch.rs
```diff
@@ -47,22 +47,22 @@ impl BatchValidator {
 
         let mut ctx = SaplingVerificationContextInner::new();
 
-        for spend in bundle.shielded_spends {
+        for spend in bundle.shielded_spends() {
```
daira, line 50: This code not being covered by tests is concerning. (I know that is a preexisting problem.)

zcash_proofs/src/sapling/verifier/single.rs
```diff
@@ -42,18 +45,13 @@ impl SaplingVerificationContext {
             cv,
             anchor,
             nullifier,
-            rk,
+            &rk,
```
daira, line 48: Similarly, about this not being covered by tests.

daira (changes requested):
NACK, potential consensus change because `ValueCommitment::derive` can return a `ValueCommitment` for the zero point which is no longer explicitly excluded.

zcash_proofs/src/sapling/prover.rs
str4d: The reason for the long path import is that this was originally named `ValueCommitment` and there were two of them. We can fix that now.

zcash_primitives/src/transaction/components/sapling.rs
str4d: Mmk, all of these examples could definitely be replaced by usage of `SaplingBuilder` (see the changes to [zcash_primitives/benches/note_decryption.rs](https://github.com/zcash/librustzcash/pull/734/files#diff-09d5a6ce4867b5daf6d0c6b4fb7815ab80432d14c20025660ef9a625747b7253R40-R54) in this PR). However, once we pull this code out into `sapling-crypto`, we will need `Bundle::from_parts` to be public (the same way it is for `orchard`) in order for transaction parsing logic to remain in `zcash_primitives`. So I'd be okay with exposing that.

I am not keen on exposing `&mut` references to the internals of a `Bundle`; the only reason those methods exist on `OutputDescription` is for the purpose of testing the internals of note decryption, which is implemented as part of the crate.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/108b2f8767f029261c8a89415bd7440f4b999f79..23922ca290e300db3252e484dfdcd18c17ed75ee) to address @daira's comments.

zcash_primitives/src/transaction/components/sapling.rs
nuttycom: We could add them behind the `test-dependencies` feature instead, but I think that they're better off only for internal testing. Even better would be if we could exclude the need for any mutation from our tests.

nuttycom (approved):
utACK with minor comment about test API

ebfull (approved):
ACK

daira (approved):
utACK. It's a little difficult to review on my phone, but my comments seem to have been addressed.

zcash_proofs/src/sapling/verifier.rs
daira: 👍

PR #417, opened 2021-07-12
Add manual serialization for bridgetree::Frontier

nuttycom:
This also implements parsing of bridgetree::Frontier from the CommitmentTree serialized form.

PR #753, opened 2023-01-06
Sapling refactor

str4d:
This refactors the internals of the `zcash_primitives::sapling` module. It does not alter the existing public API (which is being done in #755).


zcash_primitives/src/sapling/tree.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::io::{self, Read, Write};
+
+use super::{
+    note::Note,
+    pedersen_hash::{pedersen_hash, Personalization},
+};
+use crate::merkle_tree::{HashSer, Hashable};
+
+pub const SAPLING_COMMITMENT_TREE_DEPTH: usize = 32;
+pub const SAPLING_COMMITMENT_TREE_DEPTH_U8: u8 = 32;
```
daira, line 14: Is this really needed? `SAPLING_COMMITMENT_TREE_DEPTH as u8` [is a const expression](https://doc.rust-lang.org/reference/const_eval.html#constant-expressions).

zcash_primitives/src/sapling/address.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+impl Eq for PaymentAddress {}
+
+impl PaymentAddress {
+    /// Constructs a PaymentAddress from a diversifier and a Jubjub point.
+    ///
+    /// Returns None if `pk_d` is the identity.
+    pub fn from_parts(diversifier: Diversifier, pk_d: jubjub::SubgroupPoint) -> Option<Self> {
```
daira, line 32: Pre-existing issue: this does not check that `diversifier` is valid, i.e. it does not check that `diversifier.g_d()` succeeds (which it may not since `Diversifier` is only a `pub` wrapper for `[u8; 11]`).

zcash_primitives/src/sapling/tree.rs
str4d: This constant was added by @nuttycom in #417. Nothing in this repo uses it, so I presume it was added for a downstream user, `zcash/zcash` perhaps?

zcash_primitives/src/sapling/address.rs
str4d: Indeed. Fixing this is a change to the public API (both here, and in places where we no longer need to fallibly derive `g_d`), so I'm not doing it in this PR but plan to do so.

zcash_primitives/src/sapling/address.rs
```diff
@@ -0,0 +1,118 @@
...
+        if pk_d.is_identity().into() {
+            None
+        } else {
```
daira, line 35:
```suggestion
        } else {
            // Check that the diversifier is valid
            diversifier.g_d()?;
```
and add a notice of the incompatible change to the release notes.

(The protocol spec omits to specify in [section 5.6.3.1](https://zips.z.cash/protocol/protocol.pdf#saplingpaymentaddrencoding) that `\mathsf{DiversifyHash^{Sapling}}(d)` MUST NOT return `\bot`. It was intended to specify that, and there should be no incompatibility from also enforcing it here, since it was already enforced by `PaymentAddress::from_bytes`. I have filed https://github.com/zcash/zips/issues/659 for the spec omission.)

str4d: Per https://github.com/zcash/librustzcash/pull/753#discussion_r1064859474, this PR is intentionally not changing the existing public API. I will be doing this in a subsequent PR.

zcash_primitives/src/sapling/address.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// Parses a PaymentAddress from bytes.
+    pub fn from_bytes(bytes: &[u8; 43]) -> Option<Self> {
+        let diversifier = {
+            let mut tmp = [0; 11];
+            tmp.copy_from_slice(&bytes[0..11]);
+            Diversifier(tmp)
+        };
+        // Check that the diversifier is valid
+        diversifier.g_d()?;
```
daira, line 59:
```suggestion
        // PaymentAddress::from_parts will check that the diversifier is valid.
```
(This depends on the change above, and is just an optimization / avoids redundancy. It does not matter what order the components are checked in.)

zcash_primitives/src/sapling/address.rs
```diff
@@ -0,0 +1,118 @@
...
+        if pk_d.is_identity().into() {
+            None
+        } else {
+            Some(PaymentAddress { pk_d, diversifier })
+        }
+    }
+
+    /// Constructs a PaymentAddress from a diversifier and a Jubjub point.
+    ///
+    /// Only for test code, as this explicitly bypasses the invariant.
```
daira, line 42:
```suggestion
    /// Only for test code, as this explicitly bypasses the invariant.
    /// The `g_d()` method will panic on a `PaymentAddress` that has been
    /// constructed by passing an invalid diversifier to `from_parts_unchecked`.
```

zcash_primitives/src/sapling/note.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Note {
+    /// The value of the note
+    pub value: u64,
+    /// The diversified base of the address, GH(d)
```
daira, line 44:
```suggestion
    /// The diversified base of the address, DiversifyHash(d)
```

zcash_primitives/src/sapling/note.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub g_d: jubjub::SubgroupPoint,
+    /// The public key of the address, g_d^ivk
```
daira, line 46:
```suggestion
    /// The public key of the address, [ivk] g_d
```

zcash_primitives/src/sapling/note.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// rseed
+    pub rseed: Rseed,
+}
+
+impl PartialEq for Note {
+    fn eq(&self, other: &Self) -> bool {
+        self.value == other.value
+            && self.g_d == other.g_d
+            && self.pk_d == other.pk_d
+            && self.rcm() == other.rcm()
```
daira, line 57: Pre-existing issue: if, for the sake of argument, collisions could be found for $\mathsf{rseed} \mapsto \mathsf{PRF^{expand}_{rseed}}([4])$, then there could be two distinct `Note` instances that are equal. This is because `Note` actually represents a Sapling *note plaintext* minus the memo, not a Sapling note $(\mathsf{v}, \mathsf{d}, \mathsf{pk_d}, \mathsf{rcm})$ as defined in the spec.

(The protocol spec does not specify collision-resistance on the key as a security property of $\mathsf{PRF^{expand}_{rseed}}$. We can infer collision-resistance of its instantiation from that of BLAKE2b-512, so this is not a problem in practice.)

zcash_primitives/src/sapling/address.rs
daira: 👍

zcash_primitives/src/sapling/keys.rs
```diff
@@ -254,6 +257,72 @@ impl Diversifier {
...
+            .into_iter()
+            .map(move |s| s.and_then(|_| secrets_affine.next()))
+    }
+
+    /// Defined in [Zcash Protocol Spec § 5.4.5.4: Sapling Key Agreement][concretesaplingkdf].
+    ///
+    /// [concretesaplingkdf]: https://zips.z.cash/protocol/protocol.pdf#concretesaplingkdf
+    pub(crate) fn kdf_sapling(self, ephemeral_key: &EphemeralKeyBytes) -> Blake2bHash {
+        Self::kdf_sapling_inner(
+            jubjub::ExtendedPoint::from(self.0).to_affine(),
```
nuttycom, line 388: Is this a type-safety/spec-accuracy improvement relative to https://github.com/zcash/librustzcash/pull/753/commits/7fb80d55d64c3d25ff73542cc671c1f224372584#diff-9674216338af68a74b0dc7f5ce9b3f070551ca9971e2a5334a45cb5a6beb9127L89? I can't tell whether this is the same.

zcash_primitives/src/sapling/keys.rs
```diff
@@ -254,6 +257,72 @@ impl Diversifier {
...
+            .filter_map(|s| s.as_ref().map(|s| jubjub::ExtendedPoint::from(s.0)))
+            .collect();
+
+        // Batch-normalize the shared secrets.
+        let mut secrets_affine = vec![jubjub::AffinePoint::identity(); secrets.len()];
+        group::Curve::batch_normalize(&secrets, &mut secrets_affine);
+
+        // Re-insert the invalid ephemeral_key positions.
+        let mut secrets_affine = secrets_affine.into_iter();
+        shared_secrets
```
nuttycom, line 378: nit: I don't like the use of the side-effecting `next` operation inside of `map`; I know that `map` doesn't require the function passed to it to be referentially transparent, but dang it it should. It took me a solid 10 minutes to convince myself that this was correct.

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -649,8 +650,8 @@ mod tests {
             MemoBytes::empty(),
             &mut rng,
         );
-        let epk = *ne.epk();
-        let ock = prf_ock(&ovk, &cv, &cmu.to_repr(), &epk_bytes(&epk));
+        let epk = EphemeralPublicKey::from_inner(*ne.epk());
+        let ock = prf_ock(&ovk, &cv, &cmu.to_repr(), &epk.to_bytes());
```
nuttycom, line 654: Clippy complaining on line 660 below after these changes.

nuttycom (approved):
utACK with one question where I wasn't fully able to sort through the change. Otherwise I have confirmed that this is move-only followed by functionally equivalent refactorings.

Also, one clippy lint to fix.

zcash_primitives/src/sapling/keys.rs
daira: I considered using `std::iter::from_fn` but it didn't turn out any more clearly correct. This is fine.

zcash_primitives/src/sapling/keys.rs
daira: [jubjub::SubgroupPoint::to_bytes](https://github.com/zkcrypto/jubjub/blob/6af5abf31c38e2ee006314743941c2912e39cfc4/src/lib.rs#L1435-L1437), which delegates to [jubjub::ExtendedPoint::to_bytes](https://github.com/zkcrypto/jubjub/blob/6af5abf31c38e2ee006314743941c2912e39cfc4/src/lib.rs#L1419-L1421), has to do an extended-to-affine conversion, and that [involves expensive field inversions](https://github.com/zkcrypto/jubjub/blob/6af5abf31c38e2ee006314743941c2912e39cfc4/src/lib.rs#L227-L243). In contrast [jubjub::AffinePoint::to_bytes](https://github.com/zkcrypto/jubjub/blob/main/src/lib.rs#L455-L464) does not involve any field operations. So for efficiency we want to have `kdf_sapling_inner`, defined below, for the case where we already have an `AffinePoint` (e.g. obtained by *batch* extended-to-affine conversion).

We want to avoid duplicating the body of `kdf_sapling_inner`, but we cannot use `AffinePoint::from(SubgroupPoint)` directly because there is no `impl From<SubgroupPoint> for AffinePoint`, and we cannot access `SubgroupPoint`'s `.0` member directly. Hence this indirection. I've checked it is correct and is doing the same thing as the existing code.

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -215,33 +198,30 @@ impl<P: consensus::Parameters> Domain for SaplingDomain<P> {
...
         esk: &Self::EphemeralSecretKey,
     ) -> Self::EphemeralPublicKey {
-        // epk is an element of jubjub's prime-order subgroup,
-        // but Self::EphemeralPublicKey is a full group element
-        // for efficiency of encryption. The conversion here is fine
-        // because the output of this function is only used for
-        // encoding and the byte encoding is unaffected by the conversion.
-        (note.g_d * esk).into()
+        EphemeralSecretKey(*esk)
+            .derive_public(note.g_d.into())
```
daira, line 202: This ends up calling [`ka_sapling_derive_public`](https://github.com/zcash/librustzcash/blob/19d8e2125e03df88783b23ae6dcad0ae69436e97/zcash_primitives/src/sapling/spec.rs#L83) which uses `PreparedBase` and `PreparedScalar`. Remind me, is it efficient to do that even for a single scalar mul with fresh base and scalar?

daira (approved):

zcash_primitives/src/sapling/note_encryption.rs
str4d: Yes: #332.

zcash_primitives/src/sapling/note_encryption.rs
str4d: I'll fix clippy issues in a follow-up PR once the dust settles.

PR #756, opened 2023-01-11
[#751] add support for rewind_to_height to FsBlockDb

### LINKED ISSUE #751: add support for rewind_to_height to `FsBlockDb`
a BlockSource might be rewound in different scenarios like chain re-orgs or user triggered re-scans.

Whatever the case, wallets need BlockSource implementations to be able to go back a few blocks to resume scanning. 

Reorgs would mean a max rewind distance of 100 blocks, whereas quick re-scans will depend on the wallet's state. On current linear ascending scan, quick re-scans go back to the minedHeight of the oldest unspent note.

the rust sqlite_crate assumes that caller will delete the files rewound and will roll back the database only.

pacu:
Closes #751


zcash_client_sqlite/src/chain.rs
```diff
@@ -139,6 +139,17 @@ pub(crate) fn blockmetadb_insert(
     }
 }
 
+#[cfg(feature = "unstable")]
+pub(crate) fn blockmetadb_rewind_to_height(
+    conn: &Connection,
+    block_height: BlockHeight,
+) -> Result<(), rusqlite::Error> {
+    conn.prepare("DELETE FROM compactblocks_meta WHERE height > ?")
+        .unwrap()
```
nuttycom, line 148:
```suggestion
    conn.prepare("DELETE FROM compactblocks_meta WHERE height > ?")?
```

nuttycom (changes requested):
This looks fine apart from the `unwrap`.

zcash_client_sqlite/src/chain.rs
```diff
@@ -139,6 +139,16 @@ pub(crate) fn blockmetadb_insert(
     }
 }
 
+#[cfg(feature = "unstable")]
+pub(crate) fn blockmetadb_rewind_to_height(
+    conn: &Connection,
+    block_height: BlockHeight,
+) -> Result<(), rusqlite::Error> {
+    conn.prepare("DELETE FROM compactblocks_meta WHERE height > ?")
```
nuttycom, line 147:
```suggestion
    conn.prepare("DELETE FROM compactblocks_meta WHERE height > ?")?
```

nuttycom: Missed the question mark (to return the result as an error) rather than `unwrap`.

nuttycom (changes requested):
Not compiling.

str4d:
Compilation failure is due to `rustls 0.20.8` bumping its MSRV to 1.57.0 (as a result of its own dependencies bumping MSRV). We only use `rustls` when the `download-params` feature of `zcash_proofs` is enabled; here it is being enabled during testing.

I'll open a separate issue for handling or bumping MSRV.

nuttycom (approved):
utACK

Issue #759, opened 2023-01-19
Raise MSRV for `zcash_primitives` and its dependents


str4d:
Our current MSRV for `zcash_primitives` is 1.56.1 (which was a security fix release that some of our dependencies raised their MSRV to).

`rustls 0.20.8` bumped its MSRV to 1.57.0 (as a result of its own dependencies bumping MSRV). We only use `rustls` when the `download-params` feature of `zcash_proofs` is enabled, but that is part of our public API, so we now need to consider bumping our MSRV, and if we're going to bump for `zcash_proofs` then we should just bump for `zcash_primitives` and its dependents. (Component crates like `zcash_encoding` should be considered separately, given that they are going to be factored out.)

str4d:
I performed some analysis to help guide our MSRV choice.

# Features

I went through the release blog posts and picked out reasons we might want to set our MSRV to a specific version:

- [1.57](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html) (`n - 9`):
  - `panic!` in const contexts, enabling static assertions.
  - Custom compiler profiles.
  - Fallible allocation APIs.
- [1.59](https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html) (`n - 7`):
  - Inline assembly.
  - Const generics defaults and interleaving.
  - `std::thread::available_parallelism` (replacing `num_cpus` crate).
- [1.60](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html) (`n - 6`):
  - Weak dependency features and namespaced features.
  - `Instant` monotonicity guarantees (IDK if we rely on `Instant` anywhere though).
- [1.61](https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html) (`n - 5`):
  - More capabilities for `const fn`.
- [1.62](https://blog.rust-lang.org/2022/06/30/Rust-1.62.0.html) (`n - 4`):
  - `#[default]` enum variants.
  - `bool::then_some`
- [1.63](https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html) (`n - 3`):
  - Scoped threads.
  - `const Mutex, RwLock, Condvar` (enables avoiding `lazy_static` usage for creating global statics using these).
- [1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html) (`n - 2`):
  - C-compatible FFI types in `core` and `alloc` (for `no-std` usage).
  - Workspace inheritance.
[...]

PR #760, opened 2023-01-20
Bump MSRV for `zcash_primitives` and dependents to 1.60

str4d:
The MSRVs of the component crates are left as-is, partly because our dependencies don't require us to bump them, and partly because those crates have no pending changes and are relatively stable. We also plan to split the component crates out into a separate repository, where it will be easier to have a separate MSRV.

Closes zcash/librustzcash#759.


str4d:
I bumped `criterion` because it reduced duplicate dependencies. I didn't bump any other dependency because doing so increased duplicate dependencies.

nuttycom (approved):
utACK

PR #763, opened 2023-01-24
zcash_client_sqlite: Add `FsBlockDb::find_block`

str4d:


daira (approved):
utACK (but I'm not familiar with the use case for this, so needs another review from someone who is).

zcash_client_sqlite/src/lib.rs
```diff
@@ -1285,6 +1301,32 @@ mod tests {
...
+        let block_path = meta.block_file_path(&blocks_dir);
+
+        File::create(block_path)
+            .unwrap()
+            .write_all(&cb.encode_to_vec())
+            .unwrap();
+
+        meta
+    }
+
```
nuttycom, line 1329: Should we actually make a method like this part of the `BlockStore` interface?

nuttycom (approved):
utACK

zcash_client_sqlite/src/lib.rs
str4d: Maybe in future when we need it from Rust; for now this is logic that the mobile SDKs will be implementing themselves.

PR #770, opened 2023-01-31
Remove unstable function `WalletWrite::remove_unmined_tx`

### LINKED ISSUE #622: `zcash_client_sqlite`: Decide whether to keep or remove `WalletWrite::remove_unmined_tx`
#621 adds a `WalletWrite::remove_tx` API to replace wallet database mutation logic in the Android SDK (that should not be there, as it violates the wallet database contract). The logic preserves a bug from the Android SDK where transactions that have spent outputs are allowed to be deleted, and their full downstream effects are not fully unwound.

We should do one of three things before the next `zcash_client_sqlite` release:
- Change `WalletWrite::remove_tx` to cascade and delete downstream transactions of the given transaction.
- Change `WalletWrite::remove_tx` to return an error if deleting this transaction would cause other transactions to be deleted.
- Remove `WalletWrite::remove_tx`.

pacu:
Closes #622 (if the decision is to actually remove this)

Used PR https://github.com/zcash/librustzcash/pull/621 as reference to know what had been added and remove it.


nuttycom (approved):
utACK

str4d (approved):
utACK

PR #773, opened 2023-02-01
CI: Replace most unmaintained `actions-rs/*` actions

str4d:
The `actions-rs/*` actions are unmaintained. See actions-rs/toolchain#216 for more information.

- `actions-rs/toolchain` is replaced by `rust-toolchain.toml`.
- `actions-rs/cargo` is replaced by direct cargo usage.
- `actions-rs/tarpaulin` is replaced by the `cargo-tarpaulin` container.

We currently retain a few usages due to a current lack of alternatives:

- `actions-rs/toolchain` to enable beta (for Clippy warnings) and nightly (for building the book).
- `actions-rs/clippy-check` so Clippy results are published as annotations on PRs.

We also now require via `rust-toolchain.toml` that people building from the repo have the `clippy` and `rustfmt` components installed. This is fine because they will need to ensure these pass when making PRs.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -16,10 +16,7 @@ jobs:
           override: true
 
       - name: Build latest rustdocs
-        uses: actions-rs/cargo@v1
-        with:
-          command: doc
-          args: --no-deps --workspace --all-features
+        run: cargo doc --no-deps --workspace --all-features
         env:
           RUSTDOCFLAGS: -Z unstable-options --enable-index-page --cfg docsrs
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: 1.60.0
-          override: true
-
       - name: Fetch path to Zcash parameters
         working-directory: ./zcash_proofs
         shell: bash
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -76,35 +57,22 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: 1.60.0
-          components: clippy
-          override: true
       - name: Run clippy
         uses: actions-rs/clippy-check@v1
```
str4d, line 73: I did some searching and found a few possible alternatives for this, but I'm not currently sold on any of them:
- https://crates.io/crates/clippy-sarif
  - Integrates with GitHub Advanced Security, so the lints appear as security warnings.
- https://github.com/giraffate/clippy-action
  - Requires integration with `reviewdog` AFAICT.
- https://github.com/qernal/github-actions-rust-clippy
  - Looks equivalent, but also appears to only be used by its author.

.github/workflows/book.yml
```diff
@@ -16,10 +16,7 @@ jobs:
           override: true
```
str4d, line 16: I considered switching to `dtolnay/rust-toolchain`, but it [will not support `rust-toolchain`](https://github.com/dtolnay/rust-toolchain/issues/12), and [because it uses `default` instead of `override` it doesn't work with repos that have a `rust-toolchain.toml`](https://github.com/dtolnay/rust-toolchain/issues/29).

sellout (approved):
utACK.

This seems like a clear win to me.

nuttycom (approved):
utACK

.github/workflows/book.yml
str4d: A few hours after I posted this, some changes were made to `dtolnay/rust-toolchain` that now make this workflow possible without manual hackery. I'll open another PR to implement it.

PR #775, opened 2023-02-02
CI: Replace `actions-rs/toolchain` with `dtolnay/rust-toolchain`

str4d:
Now that dtolnay/rust-toolchain#53 has been closed, we can migrate away from the unmaintained actions-rs/toolchain.


nuttycom (approved):
utACK

PR #784, opened 2023-03-08
Temporarily re-expose ability to construct invalid Sapling bundles

str4d:
Until zcash/zcash#6397 is closed, this ability is needed by `zcashd` for crossing the FFI.


daira (commented):
Post-hoc ACK

PR #793, opened 2023-03-15
[#792] zcash_primitives: implement Seed Signature section of ZIP-32

### LINKED ISSUE #792: `zcash_primitives`: implement `Seed Signature` section of ZIP-32
Provide a function to produce a fingerprint from seed bytes according to ZIP-32 specification

Context:  We'd like to safely support multi-seed wallet developers by using seed fingerprints as general purpose unique identifiers for the application domain. 

https://github.com/zcash/zcash-android-wallet-sdk/issues/848

Reference: 
https://zips.z.cash/zip-0032#seed-fingerprints

pacu:
Implements a type to enclose a Seed Fingerprint byte sequence called `SeedFingerprint`

implements a public function to create a `SeedFingerprint` from seed bytes

Closes #792


zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub fn from_seed(seed_bytes: &[u8]) -> SeedFingerprint {
+        let mut h = Blake2bParams::new()
+            .hash_length(32)
+            .personal(ZIP32_SEED_FP_PERSONALIZATION)
+            .to_state();
+        let seed_len_bytes = seed_bytes.len().to_le_bytes();
+
+        let concatenated_input = [&seed_len_bytes, seed_bytes].concat();
+
+        h.update(&concatenated_input);
```
nuttycom, line 23:
```suggestion
        h.update(seed_bytes.len().to_le_bytes());
        h.update(&seed_bytes);
```

pacu: Oh I didn't know update would mean to concatenate.

I have a question about the implementation: is the `to_le_bytes()` correct? The ZIP says all bytes are `le` but IDK if seed bytes are already `le`.

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+//! [Section Seed Fingerprints]: https://zips.z.cash/zip-0032#seed-fingerprints
+
+use blake2b_simd::Params as Blake2bParams;
+
+pub const ZIP32_SEED_FP_PERSONALIZATION: &[u8; 16] = b"Zcash_HD_Seed_FP";
+
+pub struct SeedFingerprint([u8; 32]);
+
+impl SeedFingerprint {
+    pub fn from_seed(seed_bytes: &[u8]) -> SeedFingerprint {
```
daira, line 14:
```suggestion
    /// Return the seed fingerprint of the wallet as defined in
    /// <https://zips.z.cash/zip-0032#seed-fingerprints>.
    pub fn from_seed(seed_bytes: &[u8]) -> SeedFingerprint {
```

zcash_primitives/src/zip32/fingerprint.rs
daira: The [specification](https://zips.z.cash/zip-0032#seed-fingerprints) says to use a single-byte length, so it is not necessary to define a byte order.

In any case, `len()` returns a `usize`, and [usize::to_le_bytes](https://doc.rust-lang.org/std/primitive.usize.html#method.to_le_bytes) is not correct because, as the documentation says, it returns an array of length 2, 4 or 8 bytes depending on the target pointer size.

daira:
```suggestion
        let seed_len = seed_bytes.len();
        assert!(32 <= seed_len && seed_len <= 252);
        h.update(&[u8::try_from(seed_len).unwrap()]);
        h.update(seed_bytes);
```

daira, line 23:
```suggestion
        let seed_len = seed_bytes.len();
        assert!(32 <= seed_len && seed_len <= 252);

        let mut h = Blake2bParams::new()
            .hash_length(32)
            .personal(ZIP32_SEED_FP_PERSONALIZATION)
            .to_state();
        h.update(&[u8::try_from(seed_len).unwrap()]);
        h.update(seed_bytes);
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+        let seed_len_bytes = seed_bytes.len().to_le_bytes();
+
+        let concatenated_input = [&seed_len_bytes, seed_bytes].concat();
+
+        h.update(&concatenated_input);
+
+        let mut seed_fp = [0u8; 32];
+        seed_fp.copy_from_slice(h.finalize().as_bytes());
+
+        SeedFingerprint(seed_fp)
```
daira, line 28:
```suggestion
        let seed_len = seed_bytes.len();
        assert!(32 <= seed_len && seed_len <= 252);

        SeedFingerprint(Blake2bParams::new()
            .hash_length(32)
            .personal(ZIP32_SEED_FP_PERSONALIZATION)
            .to_state()
            .update(&[u8::try_from(seed_len).unwrap()])
            .update(seed_bytes)
            .finalize()
            .as_bytes())
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+#[test]
+fn test_seed_fingerprint() {
+    struct TestVector {
+        root_seed: Vec<u8>,
```
daira, line 35:
```suggestion
        master_seed: Vec<u8>,
        fingerprint: Vec<u8>,
```

daira, line 35:
```suggestion
        root_seed: Vec<u8>,
        fingerprint: Vec<u8>,
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+        ],
+    }];
+
+    let fp = SeedFingerprint::from_seed(&test_vectors[0].root_seed);
+
+    assert_eq!(fp.0.len(), 32);
+}
+
+#[cfg(any(test, feature = "test-dependencies"))]
+pub mod testing {}
```
daira, line 53:
```suggestion
```

zcash_primitives/src/zip32/fingerprint.rs
nuttycom: Given the restriction on `seed_bytes.len()`, I think that this should return an `Option<SeedFingerprint>` such that it returns `None` if the length of the seed exceeds what can be represented in a lead byte.

zcash_primitives/src/zip32/fingerprint.rs
nuttycom: Instead of `assert`...`unwrap` here, use `u8::try_from(seed_len).ok().map( ... )`

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+        root_seed: Vec<u8>,
+    }
+
+    // From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/unified_address.py
+    let test_vectors = vec![TestVector {
+        root_seed: vec![
+            0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d,
+            0x0e, 0x0f, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x1a, 0x1b,
+            0x1c, 0x1d, 0x1e, 0x1f,
+        ],
```
daira, line 44:
```suggestion
        ],
        fingerprint: vec![
            0xde, 0xff, 0x60, 0x4c, 0x24, 0x67, 0x10, 0xf7, 0x17, 0x6d, 0xea, 0xd0, 0x2a, 0xa7,
            0x46, 0xf2, 0xfd, 0x8d, 0x53, 0x89, 0xf7, 0x7, 0x25, 0x56, 0xdc, 0xb5, 0x55, 0xfd,
            0xbe, 0x5e, 0x3a, 0xe3,
        ],
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+    }];
+
+    let fp = SeedFingerprint::from_seed(&test_vectors[0].root_seed);
+
+    assert_eq!(fp.0.len(), 32);
```
daira, line 49:
```suggestion
    for tv in test_vectors {
        let fp = SeedFingerprint::from_seed(&tv.root_seed);
        assert_eq!(fp.0, &tv.fingerprint[..]);
    }
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+        let concatenated_input = [&seed_len_bytes, seed_bytes].concat();
+
+        h.update(&concatenated_input);
+
+        let mut seed_fp = [0u8; 32];
+        seed_fp.copy_from_slice(h.finalize().as_bytes());
+
+        SeedFingerprint(seed_fp)
+    }
+}
```
daira, line 30:
```suggestion
    pub fn inner(&self) -> [u8; 32] {
        self.0
    }
}
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[test]
+fn test_seed_fingerprint() {
+    struct TestVector {
+        root_seed: Vec<u8>,
+    }
+
+    // From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/unified_address.py
```
daira, line 38: This URL is invalid.

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,81 @@
...
+            .to_state()
+            .update(&[seed_len_byte])
+            .update(seed_bytes)
+            .finalize()
+            .as_bytes()
+            .try_into()
+        {
+            Ok(blake2b_hash) => Some(SeedFingerprint(blake2b_hash)),
+            Err(_) => None,
+        }
```
daira, line 41: In general:
* Only use error/`None` returns for cases that can happen as a result of invalid input.
* Use panics for cases that cannot occur or would indicate a programming error. These panics don't have to be documented.
* It can also be okay to use a panic for invalid input (but in this case I agree returning `None` is better). Such cases must be documented in a `Panics` section.
* When unwrapping an `Option` where a panic is wanted in the error case: it's okay to use `.unwrap()` when the reason why the error can't happen is clear from immediate context, otherwise use `.expect(...)`.

```suggestion
        if (32..=252).contains(&seed_len) {
            let seed_len: u8 = seed_len.try_into().unwrap();
            Some(SeedFingerprint(
                Blake2bParams::new()
                    .hash_length(32)
                    .personal(ZIP32_SEED_FP_PERSONALIZATION)
                    .to_state()
                    .update(&[seed_len])
                    .update(seed_bytes)
                    .finalize()
                    .as_bytes()
                    .try_into()
                    .expect("hash length should be 32 bytes"),
            ))
        } else {
            None
        }
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[test]
+fn test_seed_fingerprint() {
+    struct TestVector {
+        root_seed: Vec<u8>,
+        fingerprint: Vec<u8>,
+    }
+
+    // From https://github.com/zcash-hackworks/zcash-test-vectors/
```
daira, line 57: The test vector below is not from `zcash-hackworks/zcash-test-vectors`, so we shouldn't claim it is.
```suggestion
```

zcash_primitives/src/zip32/fingerprint.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    for tv in test_vectors {
+        match SeedFingerprint::from_seed(&tv.root_seed) {
+            Some(fp) => {
+                assert_eq!(&fp.to_bytes(), &tv.fingerprint[..]);
+            }
+            None => {
+                panic!("Coudn't create a fingerprint for the given seed")
+            }
+        }
```
daira, line 79:
```suggestion
        let fp = SeedFingerprint::from_seed(&tv.root_seed).expect("root_seed has valid length");
        assert_eq!(&fp.to_bytes(), &tv.fingerprint[..]);
```

daira (approved):
utACK with suggestions.

zcash_primitives/src/zip32/fingerprint.rs
pacu: Thank you for the feedback! I added it to the PR and also included a test for when `None` is returned.

nuttycom (approved):
utACK with @str4d and @sellout; for completeness we should at some point implement https://github.com/zcash-hackworks/zcash-test-vectors/issues/93 that reproduce the test vectors used in these tests, and then reference those test vectors from here.

PR #795, opened 2023-03-18
Update dependencies to reflect changes in ff 0.13

nuttycom:


zcash_proofs/Cargo.toml
```diff
@@ -17,7 +17,7 @@ categories = ["cryptography::cryptocurrencies"]
 all-features = true
 
 [dependencies]
-zcash_primitives = { version = "0.10", path = "../zcash_primitives" }
+zcash_primitives = { version = "0.10", path = "../zcash_primitives", default-features = false }
```
str4d, line 20: Add `zcash_primitives/multicore` to the `multicore` feature flag below. They are logically tied together, and if we ever move either Orchard proving into `zcash_proofs` or Sapling proving out of it, we'd be doing the same for `bellman` and `halo2_proofs` directly.

zcash_client_backend/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- Bumped dependencies to `bls12_381 0.8`, `group 0.13`, 
+- Bumped dev dependency to `jubjub 0.10`
```
str4d, line 12: We don't need to document changes to dev dependencies; they aren't observable outside the crate.
```suggestion
```

Cargo.toml
```diff
@@ -21,3 +21,7 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
+bellman = { git = "https://github.com/zkcrypto/bellman.git", rev = "29c82c7e432c1b32a50ef58e42790c3963ef2b4b" }
```
str4d, line 24: Remove this and use `bellman = "0.14"` instead.
```suggestion
```

Cargo.toml
```diff
@@ -21,3 +21,7 @@ codegen-units = 1
...
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "61a67f018f456da980de7f93a5ea2a1dc5a61e63" }
```
str4d: I verified that this commit includes the `ff 0.13` changes and is just before the `halo2_proofs` bump with the WASM fix.

Cargo.toml
```diff
@@ -21,3 +21,7 @@ codegen-units = 1
...
+halo2_gadgets = { git = "https://github.com/zcash/halo2.git", rev = "35eb268e202d76721abde3d6e50e719ae9140605" }
+halo2_proofs = { git = "https://github.com/zcash/halo2.git", rev = "35eb268e202d76721abde3d6e50e719ae9140605" }
```
str4d: I verified that this is the same commit that the `orchard` revision above depended on.

Cargo.toml
```diff
@@ -22,6 +22,6 @@ codegen-units = 1
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
 bellman = { git = "https://github.com/zkcrypto/bellman.git", rev = "29c82c7e432c1b32a50ef58e42790c3963ef2b4b" }
-orchard = { git = "https://github.com/zcash/orchard.git", rev = "61a67f018f456da980de7f93a5ea2a1dc5a61e63" }
-halo2_gadgets = { git = "https://github.com/zcash/halo2.git", rev = "35eb268e202d76721abde3d6e50e719ae9140605" }
-halo2_proofs = { git = "https://github.com/zcash/halo2.git", rev = "35eb268e202d76721abde3d6e50e719ae9140605" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "6cbde279e90974201bedbd9b5ddf155e8f8b1e8e" }
```
str4d, line 25: I verified that this is the current revision of the `main` branch for `zcash/orchard`.

Cargo.toml
```diff
@@ -22,6 +22,6 @@ codegen-units = 1
...
+halo2_gadgets = { git = "https://github.com/zcash/halo2.git", rev = "642924d614305d882cc122739c59144109f4bd3f" }
+halo2_proofs = { git = "https://github.com/zcash/halo2.git", rev = "642924d614305d882cc122739c59144109f4bd3f" }
```
str4d, line 27: I verified that this is the current revision of the `main` branch for `zcash/halo2`.

str4d (changes requested):

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/d75db2cb5a7312fce14b04d1362b2d732d00f803..7c1a4c898d38548216bb40b285d63175461a6d23) to address comments from review.

zcash_client_backend/Cargo.toml
```diff
@@ -23,7 +23,7 @@ development = ["zcash_proofs"]
 zcash_address = { version = "0.2", path = "../components/zcash_address" }
 zcash_encoding = { version = "0.2", path = "../components/zcash_encoding" }
 zcash_note_encryption = { version = "0.2", path = "../components/zcash_note_encryption" }
-zcash_primitives = { version = "0.10", path = "../zcash_primitives" }
+zcash_primitives = { version = "0.10", path = "../zcash_primitives", default-features = false }
```
str4d, line 26: This change should be documented in the changelog. Almost all downstream users likely don't have to do anything as they will already have a dependency on `zcash_primitives` that doesn't disable default features (because `default-features = false` previously didn't do anything for it), but anyone who added it by accident or by force of habit, or who doesn't have a direct dependency on `zcash_primitives` (unlikely as we don't re-export types) will see a performance regression.

zcash_client_sqlite/Cargo.toml
```diff
@@ -15,7 +15,7 @@ rust-version = "1.60"
 
 [dependencies]
 zcash_client_backend = { version = "0.7", path = "../zcash_client_backend" }
-zcash_primitives = { version = "0.10", path = "../zcash_primitives" }
+zcash_primitives = { version = "0.10", path = "../zcash_primitives", default-features = false }
```
str4d, line 18: Ditto.

zcash_extensions/Cargo.toml
```diff
@@ -11,7 +11,7 @@ rust-version = "1.60"
 
 [dependencies]
 blake2b_simd = "1"
-zcash_primitives = { version = "0.10", path = "../zcash_primitives", features = ["zfuture" ] }
+zcash_primitives = { version = "0.10", path = "../zcash_primitives", default-features = false, features = ["zfuture" ] }
```
str4d, line 14: Here we don't need to bother, because this crate has never been published (and even if it had, a performance regression in Orchard wouldn't matter to this crate).

zcash_primitives/Cargo.toml
```diff
@@ -90,13 +90,15 @@ criterion = "0.4"
...
-orchard = { version = "0.3", features = ["test-dependencies"] }
+orchard = { version = "0.3", default-features = false, features = ["test-dependencies"] }
 
 [target.'cfg(unix)'.dev-dependencies]
 pprof = { version = "0.11", features = ["criterion", "flamegraph"] } # MSRV 1.56
 inferno = ">=0.11, <0.11.5" # MSRV 1.59
 
 [features]
+default = ["multicore"]
+multicore = ["orchard/multicore"]
```
str4d, line 101: Document this added feature flag in the changelog (as exposing a way for users to selectively disable multicore support for Orchard proof creation).

zcash_proofs/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- Bumped dependencies to `bls12_381 0.8`, `group 0.13`, `jubjub 0.10`
```
str4d, line 11: Missing `bellman 0.14`.

str4d (approved):
utACK 7c1a4c898d38548216bb40b285d63175461a6d23 after changelog fixes.

PR #801, opened 2023-03-30
Upgrade to redjubjub 0.7.0

nuttycom:
This removes our reliance on the following out-of-date transitive dependencies that are duplicative of more recent versions of crates we depend on:
- bls12_381 v0.7
- ff v0.12
- group v0.12
- jubjub v0.9
- pasta_curves v0.4
- reddsa v0.3


daira (approved):
utACK

str4d (approved):
utACK

Issue #786, opened 2023-03-09
`zcash_proofs`: Theoretical possibility of overflow leading to panic


mpguerra:
As part of the zebra audit the following potential issue has been highlighted:

> In [src/hashreader.rs](https://github.com/zcash/librustzcash/blob/zcash_proofs-0.8.0/zcash_proofs/src/hashreader.rs#L14), a streamed hashing structure is defined, which hashes bytes (with the BLAKE2b hash function) as they flow, but also keeps track of how many bytes have been processed so far; the `byte_count` field maintains that information, and has type `usize` :

https://github.com/zcash/librustzcash/blob/ccb0444b6d799313604163738091e9c0ca94cf2a/zcash_proofs/src/hashreader.rs#L10-L15

> On 32-bit architectures, `usize` has size 32 bits, and thus any inputs larger than about 4.29 gigabytes will overflow that counter. In particular, the addition on [line 51](https://github.com/zcash/librustzcash/blob/ccb0444b6d799313604163738091e9c0ca94cf2a/zcash_proofs/src/hashreader.rs#L51) may then trigger a panic (if the code was compiled in debug mode), or silently truncate the count to its low 32 bits (if compilation used release mode). This issue cannot be triggered with the implementation in its current state, since hashing is performed only on files whose size has been explicitly verified to match the expected size for parameter files, with a maximum of about 0.73 gigabytes (for Sprout Groth16 parameters). Defining `byte_count` to have type `u64` would make the implementation more robust with regard to future development and protocol versions.

str4d:
I looked into this again during the 0.11 release process, and the only effect in release mode of this would be that error messages report incorrect byte counts.

PR #805, opened 2023-04-04
Fix a theoretical possibility of overflow leading to panic

upbqdn:
Closes https://github.com/zcash/librustzcash/issues/786 by changing the type of `byte_count` from `usize` to `u64`.


str4d (approved):
utACK. It's unlikely this ever becomes necessary, as it's only used for proving system parameters and none of ours are over 1GiB (and from Orchard they can be generated instead of distributed), but it's not an incorrect change (as the amount of data read in a single go will never exceed `u64`).

PR #806, opened 2023-04-06
Ensure that Sapling padding outputs are taken account in ZIP 317 change computation.

### LINKED ISSUE #799: Ensure that ZIP 317 fee calculation in wallet code takes padding into account.
The standard ZIP 317 fee algorithm applies to fully-constructed transactions. However, in the transaction builders, the Orchard (and potentially the Sapling) bundle may be padded to 2 actions. We need to ensure that this padding is taken into account by the ZIP 317 fee estimation code.

nuttycom:
Fixes #799


str4d (approved):
utACK

PR #807, opened 2023-04-07
Fix double-entry accounting errors in transaction views.

nuttycom:


pacu (approved):
utACK.

I pushed this branch on the iOS FFI repo and pointed it to this branch. It has a small compiler error, could you check it out? I never understood the zcash_address crate and how it worked so I need some help over there to test this out

https://github.com/zcash-hackworks/zcash-light-client-ffi/pull/85

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,318 @@
...
+    fn description(&self) -> &'static str {
+        "Fix transaction views to correctly handle double-entry accounting for change."
+    }
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
```
str4d, line 40:
```suggestion
        // As of this migration, the `received_notes` table only stores Sapling notes, so
        // we can hard-code the `output_pool` value.
        transaction.execute_batch(
```

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,318 @@
...
+            "INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_account, value)
+             SELECT tx, 2, output_index, account, account, value
```
str4d, line 42: Can we avoid actually hard-coding `2` here, and instead use an existing constant?

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_account, value)
+             SELECT tx, 2, output_index, account, account, value 
+             FROM received_notes
+             WHERE received_notes.is_change
+             EXCEPT
+             SELECT tx, 2, output_index, from_account, from_account, value
+             FROM sent_notes")?;
```
str4d, line 47: What is this `EXCEPT` necessary for? Do we have a test that exercises it?

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        db_data
+            .conn
+            .execute(
+                "INSERT INTO accounts (account, ufvk) VALUES (0, ?)",
+                params![ufvk.encode(&tests::network())],
+            )
+            .unwrap();
+
+        db_data.conn.execute_batch(
```
str4d, line 223:
```suggestion
        // - Tx 0 creates two notes of 2 and 5 zatoshis for account 0.
        // - Tx 1 creates two notes of 2 and 3 zatoshis for an external address, and a
        //   change note of 2 zatoshis.
        db_data.conn.execute_batch(
```
Writing this comment helped me clarify that this database is in a state that would never be produced by the existing wallet. Tx 1 is presumably intended to spend the two notes received in tx 0, but that requires setting the `spent` column for those two notes. Alternatively tx 1 represents a shielding transaction, but then there should be corresponding entries in the `utxos` table.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,318 @@
...
+            VALUES (0, 3, 0, '', 5, '', 'b', false);
+
+            INSERT INTO blocks (height, hash, time, sapling_tree) VALUES (1, 1, 1, '');
+            INSERT INTO transactions (block, id_tx, txid) VALUES (1, 1, 'b');
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_address, value)
+            VALUES (1, 2, 0, 0, '', 2);
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_address, value, memo)
+            VALUES (1, 2, 1, 0, '', 3, X'61');
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change)
+            VALUES (1, 3, 0, '', 2, '', 'c', true);",
```
str4d, line 239: Add a check that after the new migration is run, the only change to the data in the database is that `sent_notes` contains a row corresponding to this change note in `received_notes`.

str4d (changes requested):

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
nuttycom: This EXCEPT is the critical part that ensures we don't create redundant `sent_notes` entries.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
nuttycom: `tx 0` simply represents a transaction received by the wallet from some external source; such inbound transactions can of course have multiple outputs to the same recipient.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -541,8 +538,13 @@ mod tests {
...
                    LEFT JOIN blocks
                           ON transactions.block = blocks.height
-            GROUP BY received_notes.tx, received_notes.account",
-            // v_tx_received
+            GROUP BY received_notes.tx, received_notes.account
+            HAVING id_tx NOT IN (
+                SELECT tx FROM sent_notes
+                UNION
+                SELECT spent_in_tx FROM utxos
+            )",
```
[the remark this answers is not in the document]
nuttycom: The cross-account transaction case gets really complicated, which is part of the reason that I think the whole notion of "received transaction" is a leaky abstraction. For example, if you send funds from one account to another in a way that generates change, the sending account should not consider this to be a "received transaction" but the receiving account *should* do so. I hadn't realized it before now, but it does look like sqlite supports tuple expressions, which are the only way I can think of to make this achievable, by having `(id_tx, account) NOT IN ...`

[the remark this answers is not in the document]
nuttycom: Please see the latest commit.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
str4d: Sure, that part is fine; the issue was that tx 1's input state was not represented in the database. The fact that 7 zatoshis were received in tx 0 and then 7 zatoshis were spent in tx 1 led me to believe that it was tx 0 state that was missing from the test database state.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,451 @@
...
+            VALUES (1, 2, 1, 0, '', 3, X'61');
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change)
+            VALUES (1, 3, 0, '', 2, '', 'nf_c', true);
+
+            INSERT INTO blocks (height, hash, time, sapling_tree) VALUES (2, 2, 2, '');
+            INSERT INTO transactions (block, id_tx, txid) VALUES (2, 2, 'tx2');
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_address, value)
+            VALUES (2, 2, 0, 0, '', 2);
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change)
+            VALUES (2, 1, 1, '', 2, '', 'nf_d', false);",
```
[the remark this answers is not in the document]
str4d: I'm confused; the suggestion above just splits the received note into two smaller notes, but they both go to account 1. It also violates `CONSTRAINT tx_output UNIQUE (tx, output_index)`.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
str4d: Ah, I see. The `from_account, from_account` part was what confused me; it seems overly restrictive since we know that the relevant row will have `(tx, 2, output_index)` be unique (so that is what we really want to match on). If we need to select values to have this side be compatible with the `EXCEPT`, we should select whatever is in the `to_account` column rather than selecting `from_account` twice.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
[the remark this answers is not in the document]
nuttycom: The issue here is that `zcash_client_sqlite` is a publicly available library; we want to ensure that the semantics are consistent even for users who have wallets that *do* support multiple accounts.

There was definitely a bug here prior to these changes, but my contention remains that the notion of `sent transaction` and `received transaction` are leaky abstractions.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -565,6 +567,10 @@ mod tests {
                           ON transactions.id_tx = sent_notes.tx
                    LEFT JOIN blocks
                           ON transactions.block = blocks.height
+                   LEFT JOIN received_notes
+                          ON transactions.id_tx = received_notes.tx
```
nuttycom, line 571:
```suggestion
                          ON transactions.id_tx = received_notes.tx
                          AND sent_notes.output_pool = 2
```

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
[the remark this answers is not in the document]
pacu: I don't know what you mean by "leaky abstraction". But it's a notion that's on every wallet, centralized or decentralized, Fiat or Crypto, so we need to show what you received and what you sent. Can crypto be more complicated? yes for sure. But **we** as wallet SDK developers need to adapt to existing requirements regardless of what's possible protocol-wise to make sure that we have the same concepts as the other wallets of the world.

If Zcash can do more, that's great! but we want to focus on the 99,99% of users that just want to know what went out  and what came in.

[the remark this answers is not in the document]
pacu: Actually wallets are specific interpretation of the blockchain data. So it would be expected that a tool made for building wallets has these specific domain specific abstractions to serve a certain scope. 

It can be the case that these views ended up being scoped too broadly and we need others that are tailored to mobile wallets. That's ok. Let's build them if so. 

If I shift my perspective and put my hat of a 9-to-5 wallet developer that has to deal with a lot of coins and other problems, I'd like the SDK to be straight forward and that does not require me to be giving it too much thought, meaning that I should be able to paint the screen with that info pretty much 1-to-1 without having to think about notes, etc.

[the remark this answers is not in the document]
nuttycom: @pacu I think you're misunderstanding. The notion of "sent value" and "received value" are perfectly fine. The notions of "sent transaction" and "received transaction" are muddled, because a single transaction can be both.

Here's an example: you use your phone to scan to send a payment to a ZIP 321 QR code. One of the recipients is an address for some account in your wallet, either the one that you're sending the funds from or a different one.
I could, for example, in about 30 minutes build a trivially simple note management application that accepts an address, a balance, and a number of notes `n` and renders a QR code that splits up that balance in proportion to the first `n` numbers of the Fibonacci sequence. Enter the balance of the notes in your wallet, whatever `n` you want, scan the QR code and you have a transaction that spends all your notes and also where you receive the entire balance.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,488 @@
...
+                         CASE
+                             WHEN sent_notes.memo IS NULL THEN 0
+                             ELSE 1
+                         END
+                       ) AS memo_count
+                FROM sent_notes
+                WHERE (sent_notes.tx, sent_notes.output_index) NOT IN (
+                    SELECT received_notes.tx, received_notes.output_index FROM received_notes
+                    WHERE received_notes.is_change = 1
+                )
```
str4d, line 136: This needs to sub-filter on Sapling.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,498 @@
...
+                SELECT MAX(blocks.height) as max_height FROM blocks
+            )
+            SELECT notes.account_id                  AS account_id,
+                   transactions.id_tx                AS id_tx,
+                   transactions.block                AS mined_height,
+                   transactions.tx_index             AS tx_index,
+                   transactions.txid                 AS txid,
+                   transactions.expiry_height        AS expiry_height,
+                   transactions.raw                  AS raw,
+                   SUM(notes.value)                  AS net_transfer,
```
pacu, line 152: @nuttycom I've tested it and I have a doubt. It could be either an error in the query or a misunderstanding.

When I send 20000 Zats I see that the mined transaction is read back from the chain with and the value present in `net_transfer` is -21000 and `fee_paid` of 1000.

I would expect that `net_transfer` is the value of the transaction minus all deductions (in this case the fee), and when I inspect the value I find the "gross" value, which is the whole value transferred, which I consider it to be the value indicated by user intent to be transferred to the recipient, plus the value of the fee (which in this case is 1000 but if could be something else less noticeable).

I see a couple of paths here:
a) this is intentional. The value property of a transaction in the SDK will show the sum total value transferred to recipients and the naming and to show the value transferred to recipients without the fee the wallet must add the fee to the `net_transfer` column

b) there's a mistake. The value property of a transaction in the SDK will show the value transferred to recipients and to show the whole value of the transaction, wallets must add the fee to the value present  in the `net_transfer` column. 

Also, fee value for received transactions is zero

pacu (commented):
left a comment, not specifically approving or requesting changes.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
nuttycom: @pacu your answer is (a), this is intentional. The reason for this is, consider this scenario:

Your wallet recovers a transaction that:
* Spends a 500 ZAT note belonging to account `0`
* Spends a 20500 ZAT note belonging to account `1`
* Has an output of 20000 ZAT that goes to an external address.

This transaction will result in two entries in `v_transactions`, one associated with each account. The problem is, where to associate the fee? 

Consider further: the ECC wallet only attempts recovery for account `0`. Now, you show as one of "your" wallet's transactions... what? The answer we have here is a transfer of `-500`, a fee of 1000, and 1 sent output.

pacu: there is no emoji to reaction to express my "mindblown-ness" 😅

Should wallets display the `net_transfer` straightforward in any case then? or shall they add or substract  the fee?

nuttycom: I think that wallets should show `net_transfer` directly.

str4d: > The value property of a transaction in the SDK will show the sum total value transferred to recipients and the naming and to show the value transferred to recipients without the fee the wallet must add the fee to the `net_transfer` column

Yes. Framing this another way, the `net_transfer` column is the total amount that the balance of the account increases or decreases by.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -492,80 +470,95 @@ mod tests {
...
+            FROM transactions
+            JOIN notes ON notes.id_tx = transactions.id_tx
+            JOIN blocks_max_height
+            LEFT JOIN blocks ON blocks.height = transactions.block
+            LEFT JOIN sent_note_counts
+                      ON sent_note_counts.account_id = notes.account_id
+                      AND sent_note_counts.id_tx = notes.id_tx
+            GROUP BY notes.account_id, transactions.id_tx",
+            // v_tx_events
+            "CREATE VIEW v_tx_events AS
```
pacu, line 532: Is this the "Wallet Events" details that we discussed on our last sync?

I assume that the `v_transactions` is the `ZcashTransaction.Overview` and then there would be a drill down to details where the wallet requests all events for a given `id_tx` (the internal id in the database present on the Overview entity) and it would be able to "enrich" the Overview information with the wallet events. (we could make a `ZcashTransaction.Detail` entity to reflect this)

nuttycom: This is effectively (and maybe should be named) `v_tx_outputs`. It's the list of outputs of the transaction that are associated with this wallet: all outputs received (with change marked) and outputs sent.

nuttycom: To determine whether an output is "sent" or "received" from the perspective of a given account, a wallet should consider the value sof the `from_account`, `to_account`, and `to_address` fields. Either `from_account`, `to_account`, or both will always be populated; there should never be an occasion where both are null. The valid combinations are:

(from_account, to_account, null)
(from_account, null, null)
(from_account, null, to_address)
(null, to_account, null)

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
nuttycom: Or maybe `account_net_value`?

pacu: I think the main naming conflict here is that the "net value" of something is the total value minus all possible deductions which is not quite what's going on here.

pacu (approved):
I think this fixes the bug unstoppable reported plus other ones and paves the way ahead for future changes. I'll point them to the branch of the SDK that implements these changes so they can validate as well.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
```diff
@@ -0,0 +1,586 @@
...
+        db_data.conn.execute_batch(
+            "INSERT INTO blocks (height, hash, time, sapling_tree) VALUES (1, 1, 1, '');
+            INSERT INTO transactions (block, id_tx, txid) VALUES (1, 1, 'tx1');
+            UPDATE received_notes SET spent = 1 WHERE tx = 0;
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_account, to_address, value)
+            VALUES (1, 2, 0, 0, NULL, 'addra', 2);
+            INSERT INTO sent_notes (tx, output_pool, output_index, from_account, to_account, to_address, value, memo)
+            VALUES (1, 2, 1, 0, NULL, 'addrb', 3, X'61');
+            INSERT INTO received_notes (tx, output_index, account, diversifier, value, rcm, nf, is_change)
+            VALUES (1,    2, 0, '', 2, '', 'nf_c', true);").unwrap();
```
str4d, line 276: Non-blocking:
```suggestion
            VALUES (1, 2, 0, '', 2, '', 'nf_c', true);").unwrap();
```

str4d (approved):
utACK. I'm sufficiently satisfied that we can cut a release with this. If the edge cases of the new views turn out to cause issues, we can refine them further in future releases (or replace them with a Rust API-based approach).

Missing changelog updates, but we can do that in the release PR.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_net.rs
nuttycom: This space was actually intended, so that the output indices would line up.

PR #808, opened 2023-04-11
Update dependencies

str4d:


Cargo.toml
```diff
@@ -21,4 +21,3 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
```
nuttycom, line 23: Can we get rid of the path-based dependencies here? IIRC right now we're depending upon the released versions of these crates in `zcash_primitives` and the other crates.

nuttycom (approved):
utACK with nonblocking suggestion

components/zcash_address/Cargo.toml
```diff
@@ -15,7 +15,7 @@ categories = ["cryptography::cryptocurrencies", "encoding"]
 keywords = ["zcash", "address", "sapling", "unified"]
 
 [dependencies]
-bech32 = "0.8"
+bech32 = "0.9"
```
str4d, line 18: AFAICT this dependency is not exposed, so we can cut a point release of `zcash_address` at the same time we cut `zcash_primitives 0.11` et al.

zcash_client_backend/src/proto/service.rs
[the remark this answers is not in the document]
str4d: Yes; it is generated if `protoc` is present in the path and the protobuf files are present. We don't include the protobuf files in the published releases (so that downstream users don't have a hard dependency on `protoc`), instead checking in the generated output.

nuttycom (approved):
re-utACK f82866dbb965849d4c4355c09af8d8ec006538dd

daira (approved):
utACK

ebfull (approved):

PR #811, opened 2023-04-12
zcash_primitives: Re-export language types in `zip339` module

str4d:
`bip0039 0.11` switched from language function arguments to a generic parameter.


str4d:
Closing because we removed this re-export in #1391.

Issue #757, opened 2023-01-14
Eliminate duplicate versions of dependent crates


daira:
Duplicate versions make the executable larger and potentially could make debugging confusing, as well as hiding bugs in dependencies that have since been fixed. The following is based on the output of `cargo duplicates`, with redundant information suppressed and some manual research on how to improve things.

rand 0.7.3 (could be rand 0.8.5):
- Because of librustzcash 0.2.0 => clearscreen 2.0.0 => terminfo 0.7.3 => phf_codegen 0.8.0 => phf_generator 0.8.0 => rand 0.7.3
- => Tell terminfo to depend on phf 0.11 (because phf_generator 0.11.1 [depends on rand 0.8](https://github.com/rust-phf/rust-phf/blob/f8e9d279c528cb6985badc3ca3a60117ef92d51b/phf_generator/Cargo.toml#L14)), [instead of phf 0.8](https://github.com/meh/rust-terminfo/blob/0ce04e89362677c3c854e663ccbf4a0bc2b9a981/Cargo.toml#L19).

rand_core 0.5.1 (could be rand_core 0.6.4):
- Because of librustzcash 0.2.0 => ed25519-zebra 3.1.0 => curve25519-dalek 3.2.0 => rand_core 0.5.1
- => Tell ZF to make ed25519-zebra depend on curve25519-dalek 4 (which [depends on rand_core 0.6](https://github.com/dalek-cryptography/curve25519-dalek/blob/fedb1450dec58d86851ba67f2363ccde4e45ae16/Cargo.toml#L39)), [instead of curve255-dalek 3](https://github.com/ZcashFoundation/ed25519-zebra/blob/612e51af2e7eaa228f9a08ad38b0b0660e510d9e/Cargo.toml#L20). Note that this requires an MSRV bump of ed25519-zebra and librustzcash to [at least 1.60.0](https://github.com/dalek-cryptography/curve25519-dalek#breaking-changes-in-400).

wasi 0.10.2+wasi-snapshot-preview1 (could be wasi 0.11.0+wasi-snapshot-preview1):
- Because of librustzcash 0.2.0 => metrics-exporter-prometheus 0.11.0 => quanta 0.10.1 => wasi 0.10.2+wasi-snapshot-preview1
- Because of librustzcash 0.2.0 => metrics-util 0.14.0 => quanta 0.10.1 => wasi 0.10.2+wasi-snapshot-preview1
[...]

PR #813, opened 2023-04-13
Use proptest instead of quickcheck for zcash_history tests.

nuttycom:
Part of #757; resolves duplicate `rand-core` dependencies.

Best viewed ignoring whitespace changes.

### DIFF
--- a/zcash_history/src/tree.rs
+++ b/zcash_history/src/tree.rs
@@ -325,12 +325,11 @@ fn combine_nodes<'a, V: Version>(left: IndexedNode<'a, V>, right: IndexedNode<'a
 
 #[cfg(test)]
 mod tests {
-
     use super::{Entry, EntryKind, EntryLink, Tree};
     use crate::{node_data, NodeData, Version, V2};
 
     use assert_matches::assert_matches;
-    use quickcheck::{quickcheck, TestResult};
+    use proptest::prelude::*;
 
     fn leaf(height: u32) -> node_data::V2 {
         node_data::V2 {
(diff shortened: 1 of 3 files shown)


zcash_history/src/tree.rs
```diff
@@ -640,100 +639,88 @@ mod tests {
...
         }
 
-        fn parity_with_truncate(add: u32, delete: u32) -> TestResult {
+        #[test]
+        fn prop_parity_with_truncate(
+            add_and_delete in (0u32..=2048).prop_flat_map(
+                |add| (Just(add), 0..=add)
+            )
+        ) {
+            let (add, delete) = add_and_delete;
```
str4d, line 686: I believe this works:
```suggestion
            (add, delete) in (0u32..=2048).prop_flat_map(
                |add| (Just(add), 0..=add)
            )
        ) {
```

zcash_history/src/tree.rs
```diff
@@ -640,100 +639,88 @@ mod tests {
...
-                for _ in 0..delete {
-                    tree.truncate_leaf().expect("Failed to truncate");
-                }
+        #[test]
+        fn prop_stored_length(
+            add_and_delete in (0u32..=2048).prop_flat_map(
+                |add| (Just(add), 0..=add)
+            )
+        ) {
+            let (add, delete) = add_and_delete;
```
str4d, line 712:
```suggestion
            (add, delete) in (0u32..=2048).prop_flat_map(
                |add| (Just(add), 0..=add)
            )
        ) {
```

str4d (approved):
utACK

zcash_history/src/tree.rs
nuttycom: I tried that first; the macro didn't like it.

str4d: Hmm, maybe because of the `prop_flat_map`.

Issue #816, opened 2023-04-14
cargo audit shows 3 vulnerabilities and 2 warnings.


nathan-at-least:
# Problem

The `cargo audit` tool shows 3 vulnerabilities and 2 warnings:

## Vulnerabilities

- https://rustsec.org/advisories/RUSTSEC-2022-0090
- https://rustsec.org/advisories/RUSTSEC-2023-0018
- https://rustsec.org/advisories/RUSTSEC-2023-0001

## Warnings

- https://rustsec.org/advisories/RUSTSEC-2020-0056
- `crossbeam-channel` is yanked from crates.io (signaling it is disrecommended)

# Proposed Solution

1. Immediate solution: resolve all vulnerabilities; all 3 are known-resolved by dependency upgrades.
2. Preventative solution: run `cargo audit` in CI. Because vulnerabilities may be reported at any time, this should be a cronjob-style CI.

# Reproduction

```
$ cargo install cargo-audit
[…]

$ git rev-parse HEAD                      
23a9a26fe38b1caac3e4ee9e76df6fccfcad1b15     
                                                
$ cargo audit                                  
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 540 security advisories (from /home/user/.cargo/advisory-db)                        
    Updating crates.io index                                                                     
    Scanning Cargo.lock for vulnerabilities (350 crate dependencies)                             
Crate:     libsqlite3-sys                                                                        
Version:   0.22.2                                                                                
Title:     `libsqlite3-sys` via C SQLite CVE-2022-35737                                          
Date:      2022-08-03                                                                            
ID:        RUSTSEC-2022-0090                  
URL:       https://rustsec.org/advisories/RUSTSEC-2022-0090
Solution:  Upgrade to >=0.25.1         
Dependency tree:                   
libsqlite3-sys 0.22.2                  
└── rusqlite 0.25.4                       
    ├── zcash_client_sqlite 0.5.0      
    └── schemer-rusqlite 0.2.1             
        └── zcash_client_sqlite 0.5.0                                                            
                                                
Crate:     remove_dir_all                       
Version:   0.5.3           
Title:     Race Condition Enabling Link Following and Time-of-check Time-of-use (TOCTOU)
Date:      2023-02-24                                                                            
ID:        RUSTSEC-2023-0018         
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0018
Solution:  Upgrade to >=0.8.0
[...]

PR #817, opened 2023-04-14
WIP: Add CI to do daily `cargo audit` checks.

nathan-at-least:
This is an untested attempt at the "preventative solution" to #816.

It attempts to run `cargo audit` every 24 hours. I'm calling it WIP because I don't have confidence it will work well as-is:

- How are failing "cron-like" CI workflows represented? Do they mark `main` with a red flag or block other merges?
- Are these run steps correct?
- Is there a `rustsec` action we can use to cut down on run time by skipping the `cargo-audit` install?
- How can we test that this workflow functions correctly prior to merging (if at all)?

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,14 @@
+name: cargo audit
+
+on:
+  schedule:
+    - cron: '30 5 * * *'
+
+jobs:
+  cargo-audit:
+    name: cargo-audit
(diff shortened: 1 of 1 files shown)


nuttycom:
Marked as draft until this leaves WIP status.

str4d:
Closing as not planned. Dependabot effectively provides this functionality for us.

PR #818, opened 2023-04-14
Bump `tempfile` to 3.5.0 to mitigate RUSTSEC-2023-0018…

nathan-at-least:
See https://rustsec.org/advisories/RUSTSEC-2023-0018

This is resolves one of multiple audit failures reported in #816


nuttycom (approved):
utACK

PR #829, opened 2023-04-28
Add receiver type selection to unified address derivation.

nuttycom:


zcash_client_backend/src/keys.rs
```diff
@@ -136,7 +135,7 @@ impl Era {
     }
 }
 
-/// A set of spending keys that are all associated with a single
+/// A set of viewing keys that are all associated with a single
```
daira, line 138: This says viewing keys, but the struct is still named `UnifiedSpendingKey` and refers to other spending key types, which seems wrong.

zcash_client_backend/src/keys.rs
```diff
@@ -537,7 +548,13 @@ impl UnifiedFullViewingKey {
...
         let transparent = None;
 
-        UnifiedAddress::from_receivers(None, sapling, transparent)
+        let orchard = self
+            .orchard
+            .as_ref()
+            .filter(|_| receiver_types.contains(&Typecode::Orchard))
+            .map(|ofvk| ofvk.address_at(j, Scope::External));
+
+        UnifiedAddress::from_receivers(orchard, sapling, transparent)
```
daira, line 557: I would suggest putting the Orchard part first so that the code is in receiver priority order (like the arguments to `from_receivers`).

daira (commented):
The comment on `UnifiedSpendingKey` is blocking.

zcash_client_backend/src/keys.rs
```diff
@@ -507,19 +503,38 @@ impl UnifiedFullViewingKey {
...
+            .filter(|_| receiver_types.contains(&Typecode::Sapling))
+        {
             Some(extfvk.address(j)?)
         } else {
             None
         };
 
         #[cfg(feature = "transparent-inputs")]
-        let transparent = if let Some(tfvk) = self.transparent.as_ref() {
+        let transparent = if let Some(tfvk) = self
```
daira, line 533: Similarly this can use `.flatten()`.

zcash_client_backend/src/keys.rs
```diff
@@ -507,19 +503,38 @@ impl UnifiedFullViewingKey {
...
+            .transparent
+            .as_ref()
+            .filter(|_| receiver_types.contains(&Typecode::P2pkh))
+        {
             match to_transparent_child_index(j) {
                 Some(transparent_j) => match tfvk
                     .derive_external_ivk()
```
daira, line 540: Re: the `return None`, are we sure we want to return from the whole method if we can't derive the transparent address, rather than just omitting it from the UA? (If we want the latter then just use `Result::ok()` rather than the inner match.)

daira (approved):
utACK modulo comments.

zcash_client_backend/src/keys.rs
```diff
@@ -507,19 +503,38 @@ impl UnifiedFullViewingKey {
...
+
+        let sapling = if let Some(extfvk) = self
+            .sapling
+            .as_ref()
+            .filter(|_| receiver_types.contains(&Typecode::Sapling))
+        {
             Some(extfvk.address(j)?)
         } else {
             None
         };
```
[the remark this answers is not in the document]
nuttycom: This doesn't work, because `exfvk.address(j)` returns a `Result` which must be able to be returned from the top level.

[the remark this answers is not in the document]
nuttycom: Oh, wait, no: `extfvk.address` returns an Option, but your `flatten` would subtly break the semantics here: we want to require a Sapling receiver to exist if a Sapling receiver type was requested. If the Sapling receiver type is not requested, we don't want to eagerly return `None` here.

daira (dismissed):
utACK

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/1031b036f13d165f8e6de2e19a83ff0799344c3e..8298a8040aab307abd096a6d59b4d623b1f07b8a) to fix a Clippy lint.

zcash_client_backend/src/keys.rs
```diff
@@ -536,19 +537,46 @@ impl UnifiedFullViewingKey {
...
+    /// receiver types.
     ///
     /// Returns `None` if the specified index does not produce a valid diversifier.
-    // TODO: Allow filtering down by receiver types?
-    pub fn address(&self, j: DiversifierIndex) -> Option<UnifiedAddress> {
-        let sapling = if let Some(extfvk) = self.sapling.as_ref() {
+    pub fn address(
+        &self,
+        j: DiversifierIndex,
+        receiver_types: &[Typecode],
```
str4d, line 547: This change should maybe be documented, although we did introduce this type as experimental so maybe we just change them unannounced.

zcash_client_backend/src/keys.rs
```diff
@@ -536,19 +537,46 @@ impl UnifiedFullViewingKey {
...
+    ) -> Option<UnifiedAddress> {
+        #[cfg(feature = "orchard")]
+        let orchard = {
+            let orchard_j = orchard::keys::DiversifierIndex::from(*j.as_bytes());
```
str4d, line 551: This reminded me that we still need to migrate `orchard` to use `zip32` (which would remove this line). Opened https://github.com/zcash/orchard/issues/410.

zcash_client_backend/src/keys.rs
```diff
@@ -536,19 +537,46 @@ impl UnifiedFullViewingKey {
...
     ///
     /// Returns `None` if the specified index does not produce a valid diversifier.
-    // TODO: Allow filtering down by receiver types?
-    pub fn address(&self, j: DiversifierIndex) -> Option<UnifiedAddress> {
-        let sapling = if let Some(extfvk) = self.sapling.as_ref() {
+    pub fn address(
+        &self,
+        j: DiversifierIndex,
+        receiver_types: &[Typecode],
+    ) -> Option<UnifiedAddress> {
```
str4d, line 548: We need to confirm that `receiver_types` only contains `Typecode`s for which we can generate receivers. If an unsupported `Typecode` is requested, we should fail:
```suggestion
    ) -> Option<UnifiedAddress> {
        // If the user requests an unsupported receiver type, we can't produce an address.
        if receiver_types.iter()
            .any(|typecode| match typecode {
                #[cfg(feature = "orchard")]
                Typecode::Orchard => false,
                Typecode::Sapling => false,
                #[cfg(feature = "transparent-inputs")]
                Typecode::P2pkh => false,
                _ => true,
            })
        {
            return None;
        }

```
Previously this wasn't a problem because we encoded a default preference, but now that the caller has to be explicit, we need to respect that explicitness.

We might also want to make this `Option<&[Typecode]>` so if the caller specifies `None` we do the old default behaviour.

zcash_client_backend/src/keys.rs
```diff
@@ -593,7 +627,7 @@ impl UnifiedFullViewingKey {
 
         // Find a working diversifier and construct the associated address.
         loop {
-            let res = self.address(j);
+            let res = self.address(j, receiver_types);
```
str4d, line 630: `self.address()` currently returns `None` only if the diversifier index is invalid for the receiver types (previously implied by the `UnifiedFullViewingKey` components, and now explicitly stated), which means that incrementing `j` below is the correct behaviour here.

If we make my change above to validate `receiver_types` then we will also have to either pre-validate `receiver_types` in `find_address` before this call, or change the return type of `self.address()` to distinguish these two cases. Otherwise we'll get stuck in a seemingly-infinite loop here (because an unsupported receiver type will be invalid for all `j`).

zcash_client_backend/src/keys.rs
```diff
@@ -606,8 +640,11 @@ impl UnifiedFullViewingKey {
     /// Returns the Unified Address corresponding to the smallest valid diversifier index,
     /// along with that index.
     pub fn default_address(&self) -> (UnifiedAddress, DiversifierIndex) {
-        self.find_address(DiversifierIndex::new())
-            .expect("UFVK should have at least one valid diversifier")
+        self.find_address(
+            DiversifierIndex::new(),
+            &[Typecode::P2pkh, Typecode::Sapling],
```
str4d, line 645: TODO (maybe not in this PR): conditionally add Orchard here.

zcash_client_backend/src/keys.rs
```diff
@@ -792,7 +828,7 @@ mod tests {
                 continue;
             }
 
-            let ua = ufvk.address(d_idx).unwrap_or_else(|| panic!("diversifier index {} should have produced a valid unified address for account {}",
+            let ua = ufvk.address(d_idx, &[Typecode::P2pkh, Typecode::Sapling]).unwrap_or_else(|| panic!("diversifier index {} should have produced a valid unified address for account {}",
```
str4d, line 831: TODO: update the test vectors to include Orchard.

str4d (changes requested):
Reviewed 72d9ad2f9b0b4b90e4bb074731c882aa59ccc727.

zcash_client_backend/src/keys.rs
nuttycom: This makes sense. I think though that instead of returning `Option<UnifiedAddress>` we should at this point probably then switch to a `Result` with a bespoke error enum, since generation can fail in multiple ways.

zcash_client_backend/src/keys.rs
nuttycom: See also https://github.com/zcash/librustzcash/pull/829#discussion_r1442391323

zcash_client_backend/src/keys.rs
nuttycom: Since we now have the `orchard` feature in `main`, I think it makes sense in this PR.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/f693eba49e6af6feac9c78440b227de70c661a30..dfc79ed0eaebad940b1524600123681383982441) to address comments from code review and align address requests with enabled feature flags.

zcash_client_backend/src/keys.rs
```diff
@@ -373,6 +374,65 @@ impl UnifiedSpendingKey {
...
+    has_orchard: bool,
+    has_sapling: bool,
+    #[cfg(feature = "transparent-inputs")]
+    has_p2pkh: bool,
+}
+
+impl UnifiedAddressRequest {
+    pub const DEFAULT: UnifiedAddressRequest = Self {
+        #[cfg(feature = "orchard")]
+        has_orchard: false, // FIXME: Always request Orchard receivers once we can receive Orchard funds
```
nuttycom, line 406: Added https://github.com/zcash/librustzcash/issues/1099 to track this FIXME.

zcash_client_backend/src/data_api.rs
```diff
@@ -995,6 +995,7 @@ pub trait WalletWrite: WalletRead {
     fn get_next_available_address(
         &mut self,
         account: AccountId,
+        request: UnifiedAddressRequest,
```
str4d, line 998: Document this trait change in the changelog.

zcash_client_backend/CHANGELOG.md
```diff
@@ -191,10 +194,19 @@ and this library adheres to Rust's notion of
...
+    `Option<UnifiedAddress>` so that we may better report to the user how
+    address generation has failed.
   - `UnifiedSpendingKey::orchard` is now only available when the `orchard`
     feature is enabled.
+  - `UnifiedSpendingKey::transparent` is now only available when the
+    `transparent-inputs` feature is enabled.
   - `UnifiedFullViewingKey::new` no longer takes an Orchard full viewing key
     argument unless the `orchard` feature is enabled.
+  - `UnifiedFullViewingKey` and `UnifiedSpendingKey` now only contain transparent
+    key components in the case that the `transparent-inputs` feature is enabled.
```
str4d, line 209: This is not new; it's been the case since `UnifiedSpendingKey` was introduced in https://github.com/zcash/librustzcash/commit/dec395a5b0c3bd5ee1fb2661ded06005ce00b81f. Did you mean something else here?

str4d (approved):
utACK 9e141f252270721fca7d0aa7675e38f016e950c3 after changelog fixes.

zcash_client_backend/CHANGELOG.md
daira: It's also ambiguously written. I think it's intended to mean "When the `transparent-inputs` feature is disabled, `UnifiedFullViewingKey` and `UnifiedSpendingKey` omit transparent key components.", but can also be read as "When the `transparent-inputs` feature is enabled, `UnifiedFullViewingKey` and `UnifiedSpendingKey` now contain only transparent key components (and no shielded components)."

str4d (approved):
utACK 6cbdd494cf92ed9223164efa33d14acfa777fc59

PR #830, opened 2023-04-28
Update DEFAULT_FEE to 10_000 from 1000

AloeareV:
This makes the default fee equal to the minimum possible zip317 fee.

### DIFF
--- a/zcash_client_backend/src/fees/fixed.rs
+++ b/zcash_client_backend/src/fees/fixed.rs
@@ -185,7 +185,7 @@ mod tests {
             &Vec::<TxOut>::new(),
             &[TestSaplingInput {
                 note_id: 0,
-                value: Amount::from_u64(45000).unwrap(),
+                value: Amount::from_u64(60000).unwrap(),
             }],
             &[SaplingPayment::new(Amount::from_u64(40000).unwrap())],
             &DustOutputPolicy::default(),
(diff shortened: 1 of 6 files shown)


nuttycom:
Tests are failing

nuttycom:
This can be rebased now.

AloeareV:
> This can be rebased now.

Great, rebased!

AloeareV:
D'oh! Local test run was non-comprehensive. Fixed!

nuttycom (approved):
utACK

I think that this is an acceptable change. Overall what we really need to do is get rid of `DEFAULT_FEE` entirely, but for the moment this change puts us in a state that's arguably better than the existing `DEFAULT_FEE` value.

AloeareV:
> utACK
> 
> I think that this is an acceptable change. Overall what we really need to do is get rid of `DEFAULT_FEE` entirely, but for the moment this change puts us in a state that's arguably better than the existing `DEFAULT_FEE` value.

Yeah, this is intended as a bandaid. We're in a similar situation in zingo where the next release will just hardcode a fee of 10_000 instead of 1_000, and that'll be functional enough that full support for zip317 fees can wait for us to be careful about implementing it right

AloeareV:
@nuttycom Is this ready to merge, or is it awaiting another reviewer?

daira:
I will file a PR to revert this and then do the rename.

PR #833, opened 2023-05-03
zcash_history: Migrate to `primitive-types 0.12`

str4d:
We also disable the default `std` feature, to remove transitive dependencies we don't use.


str4d:
Removing the default `std` feature was not technically necessary to remove the dependencies from our dependency tree (because `primitive-types 0.12` uses Cargo weak dependencies), but it was necessary to remove it from the `Cargo.lock` (everything still compiled fine anyway because of Cargo weak dependencies, but since we weren't using anything from the `std` flag, it makes sense to just remove it).

daira (approved):
utACK

nuttycom (commented):
post-hoc utACK

PR #838, opened 2023-05-09
Revert #830 and instead modify `zcash_primitives::transaction::fees::fixed::FeeRule::standard()`

daira:
#830 updated `zcash_primitives::components::amount::DEFAULT_FEE` to be 10000 zatoshis. This is not consistent with the value of `LEGACY_DEFAULT_FEE` in zcashd, and might have unintended side effects in code that uses it. Also, the effect on the behaviour of `zcash_primitives::transaction::fees::fixed::FeeRule::standard()` was not documented in the release notes.

This PR reverts #830 (as the first commit) and instead updates `zcash_primitives::transaction::fees::fixed::FeeRule::standard()` to use a new constant, `zcash_primitives::transaction::fees::zip317::MINIMUM_FEE`. It also deprecates `zcash_primitives::components::amount::DEFAULT_FEE` and `zcash_primitives::transaction::fees::fixed::FeeRule::standard()`. (You can still use `fixed::FeeRule` via its `non_standard` constructor, which is correct because any fixed fee is by definition non-standard.)

### DIFF
--- a/zcash_primitives/src/transaction/components/amount.rs
+++ b/zcash_primitives/src/transaction/components/amount.rs
@@ -32,6 +32,14 @@ impl Amount {
         Amount(0)
     }
 
+    /// Creates a constant Amount from an i64.
+    ///
+    /// Panics: if the amount is outside the range `{-MAX_MONEY..MAX_MONEY}`.
+    pub const fn const_from_i64(amount: i64) -> Self {
+        assert!(-MAX_MONEY <= amount && amount <= MAX_MONEY); // contains is not const
+        Amount(amount)
+    }
+
     /// Creates an Amount from an i64.
     ///
     /// Returns an error if the amount is outside the range `{-MAX_MONEY..MAX_MONEY}`.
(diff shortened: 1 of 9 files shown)


nuttycom (approved):
utACK

zcash_primitives/src/transaction/components/amount.rs
```diff
@@ -32,6 +38,14 @@ impl Amount {
         Amount(0)
     }
 
+    /// Creates a constant Amount from an i64.
+    ///
+    /// Panics: if the amount is outside the range `{-MAX_MONEY..MAX_MONEY}`.
+    pub const fn const_from_i64(amount: i64) -> Self {
+        assert!(-MAX_MONEY <= amount && amount <= MAX_MONEY); // contains is not const
+        Amount(amount)
+    }
```
daira, line 47: Despite the codecov warning, this is in fact covered by tests. (A test uses `MINIMUM_FEE` which uses `const_from_i64`.)

PR #840, opened 2023-05-15
fix: Avoid a redundant check containing curve point multiplication

### LINKED ISSUE #802: Redundant Computation in Sapling and Orchard Note Validation
The following issue (https://github.com/ZcashFoundation/zebra/issues/6392 on zebra github repo) was raised by the auditors working on the Zebra audit:

## Impact

During Sapling and Orchard output validation, one redundant elliptic curve scalar multiplication is performed. It may be considered for removal for efficiency purposes.

## Description

In order to verify that Sapling and Orchard outputs are decryptable and consistent with ZIP 212 rules, Zebra uses the `zcash_primitives` crate’s `try_sapling_output_recovery` [function](https://github.com/zcash/librustzcash/blob/277d07c7/zcash_primitives/src/sapling/note_encryption.rs#L553). This function is called, for example, when coinbase transaction outputs are [validated](https://github.com/ZcashFoundation/zebra/blob/5a88fe78/zebra-consensus/src/transaction/check.rs#L320) to adhere to the rules specified in ZIP 212.

Once decrypted, an output’s note is [parsed](https://github.com/zcash/librustzcash/blob/277d07c7/components/zcash_note_encryption/src/lib.rs#L681). An ephemeral key validation function is passed as a lambda:

https://github.com/zcash/librustzcash/blob/fb49cd412f02b65def53a096f536c2d4aeda68f4/zcash_primitives/src/sapling/note_encryption.rs#L254-L268

The check implemented by the highlighted lambda function is mandated by ZIP 212. A few lines below, the `check_note_validity` function is [called](https://github.com/zcash/librustzcash/blob/277d07c7/components/zcash_note_encryption/src/lib.rs#L693):

https://github.com/zcash/librustzcash/blob/fb49cd412f02b65def53a096f536c2d4aeda68f4/components/zcash_note_encryption/src/lib.rs#L514-L537

The validation highlighted in the first code snippet happens regardless of whether ZIP 212 is activated or not. This is also what the original Zcash client does; see [`zcash/Note.cpp`](https://github.com/zcash/zcash/blob/219b9c68/src/zcash/Note.cpp#L400).
The validation highlighted in the second code snippet aims to only validate the public key post-ZIP 212 and is likely meant to be a blanket end-of-function validation helper. This helper includes an ECC point multiplication and as such is not inexpensive.

## Recommendation

It appears that the highlighted check inside the `check_note_validity` function overlaps with the previous validation, and, as such, may be removed.

upbqdn:
## Motivation

The function [try_output_recovery_with_ock](https://github.com/zcash/librustzcash/blob/bc558932679247b56d69c508c4172abda8b8d1f6/components/zcash_note_encryption/src/lib.rs#L614) calls [parse_note_plaintext_without_memo_ovk](https://github.com/zcash/librustzcash/blob/bc558932679247b56d69c508c4172abda8b8d1f6/zcash_primitives/src/sapling/note_encryption.rs#L254) from [here](https://github.com/zcash/librustzcash/blob/bc558932679247b56d69c508c4172abda8b8d1f6/components/zcash_note_encryption/src/lib.rs#L659). The second function validates the ephemeral keys. The first function then calls [check_note_validity](https://github.com/zcash/librustzcash/blob/bc558932679247b56d69c508c4172abda8b8d1f6/components/zcash_note_encryption/src/lib.rs#L514), which also contains the same check of the same ephemeral keys. These checks both contain a curve point multiplication.

Closes #802.

## Solution

- Don't call `check_note_validity`, but still check the commitment bytes.

As mentioned in https://github.com/ZcashFoundation/zebra/issues/6392#issuecomment-1520405426, I initially thought of refactoring `check_note_validity` directly, which wouldn't be worth it since it's called in other places. Only then I thought of not calling `check_note_validity` at all, which is a much simpler solution that avoids the redundant check.

### DIFF
--- a/components/zcash_note_encryption/src/lib.rs
+++ b/components/zcash_note_encryption/src/lib.rs
@@ -667,9 +667,7 @@ pub fn try_output_recovery_with_ock<D: Domain, Output: ShieldedOutput<D, ENC_CIP
         }
     }
 
-    if let NoteValidity::Valid =
-        check_note_validity::<D>(&note, &ephemeral_key, &output.cmstar_bytes())
-    {
+    if D::ExtractedCommitmentBytes::from(&D::cmstar(&note)) == output.cmstar_bytes() {
         Some((note, to, memo))
     } else {
         None


daira:
This will require three ACKs since `try_output_recovery_with_ock` is used in consensus code.

str4d:
In a gardening PR today, @daira, @nuttycom and I looked at this. We determined that while this change is correct, it leaves the burden of enforcing this necessary check on the domain implementations. I've opened #848 with an alternative approach that instead reduces the requirements on the domain implementations, and relies on the central check instead.

nuttycom:
Subsumed by #848

PR #842, opened 2023-05-16
Bump MSRV for `zcash_primitives` and dependents to 1.65

str4d:
This corresponds to an MSRV of `n - 4`, in terms of previous analysis from #759.


zcash_primitives/Cargo.toml
```diff
@@ -96,7 +96,6 @@ orchard = { version = "0.4", default-features = false, features = ["test-depende
 
 [target.'cfg(unix)'.dev-dependencies]
 pprof = { version = "0.11", features = ["criterion", "flamegraph"] } # MSRV 1.56
```
daira, line 98:
```suggestion
pprof = { version = "0.11", features = ["criterion", "flamegraph"] }
```

str4d: I'd prefer to leave this in as it's useful for documenting the MSRV of a dependency that has caused us problems before.

nuttycom (approved):
utACK

PR #843, opened 2023-05-16
Update memo retrieval API to reflect that memos may not be present.

### LINKED ISSUE #384: Document potential memory use of functions that deserialize CompactSize-containing data structures
(Originally reported by @teor2345 through security channels.)

The maximum value of a CompactSize field is [32MiB](https://github.com/zcash/librustzcash/blob/05bd98b79691b1d89b5cba304b6228ac192c5d17/zcash_primitives/src/serialize.rs#L4). Parsers that make use of the CompactSize value to pre-allocate memory for parsing the remainder of the data structure are at risk of creating a memory denial of service attack, especially when those parsers are parallelized and run in direct response to network traffic (such as might be the case in zebrad).

Wherever this library does preallocation like this, the maximum potential memory use should be documented so that developers who call the function are aware of the potential for memory DoS.

nuttycom:
Memos may be absent for both sent and received notes in cases where only compact block information has been used to populate the wallet database. This fixes a potential crash in the case that we attempt to decode a SQLite `NULL` as a byte array. It does, however, introduce a slight semantic confusion that will need to be considered in the case of future updates where a note may not have an associated memo; at present, the only reason we might not have the memo is that we might not have retrieved the full transaction information from the chain, but in the future there might be other possible reasons for this absence.

Fixes #384

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -155,8 +155,10 @@ pub trait WalletRead {
     /// Returns the memo for a note.
     ///
     /// Implementations of this method must return an error if the note identifier
-    /// does not appear in the backing data store.
-    fn get_memo(&self, id_note: Self::NoteRef) -> Result<Memo, Self::Error>;
+    /// does not appear in the backing data store. Returns `Ok(None)` if the note
+    /// is known to the wallet but memo data has not yet been populated for that
+    /// note.
+    fn get_memo(&self, id_note: Self::NoteRef) -> Result<Option<Memo>, Self::Error>;
 
     /// Returns a transaction.
     fn get_transaction(&self, id_tx: Self::TxRef) -> Result<Transaction, Self::Error>;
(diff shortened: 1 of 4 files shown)


daira (approved):
utACK

zcash_client_backend/src/data_api.rs
```diff
@@ -155,8 +155,10 @@ pub trait WalletRead {
     /// Returns the memo for a note.
     ///
     /// Implementations of this method must return an error if the note identifier
-    /// does not appear in the backing data store.
-    fn get_memo(&self, id_note: Self::NoteRef) -> Result<Memo, Self::Error>;
+    /// does not appear in the backing data store. Returns `Ok(None)` if the note
+    /// is known to the wallet but memo data has not yet been populated for that
+    /// note.
+    fn get_memo(&self, id_note: Self::NoteRef) -> Result<Option<Memo>, Self::Error>;
```
str4d, line 161: Change needs to be documented in the changelog.

str4d (approved):
utACK

zcash_client_backend/src/data_api.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/843/commits/ea84243e0f4f61b997ca79c152c55824a7ded0d8

PR #845, opened 2023-05-18
DRAFT: radicle contrib init

patitieso:
Introduction and basic instructions for Radicle peer-to-peer network for code collaboration, and seednode.link


str4d:
There is nothing specific to Radicle that needs to exist in this repository for its users to be able to replicate this repository there.

PR #846, opened 2023-05-18
Return the empty memo when memo is not present.

### LINKED ISSUE #384: Document potential memory use of functions that deserialize CompactSize-containing data structures

nuttycom:
Memos may be absent for both sent and received notes in cases where only compact block information has been used to populate the wallet database. This fixes a potential crash in the case that we attempt to decode a SQLite `NULL` as a byte array.

Fixes #384

(cherry picked from commit d99b4d4d6ec1126447159ffe649a3edac2f322d5)

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -7,6 +7,14 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+## [0.7.1] - 2023-05-17
+
+### Fixed
+- Fixes a potential crash that could occur when attempting to read a memo from 
+  sqlite when the memo value is `NULL`. At present, we return the empty memo
+  in this case; in the future, the `get_memo` API will be updated to reflect
+  the potential absence of memo data.
+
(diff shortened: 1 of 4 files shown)


str4d (approved):
utACK

str4d:
CI failures are expected because this is based off the `zcash_client_sqlite-0.7.0` tag. I verified that the changes here exactly backport changes from d99b4d4d6ec1126447159ffe649a3edac2f322d5, and the only difference is a trivially-correct map.

PR #849, opened 2023-05-25
Upgrade `incrementalmerkletree` & `orchard` patch versions.

nuttycom:


Cargo.toml
```diff
@@ -21,6 +21,5 @@ codegen-units = 1
 [patch.crates-io]
 zcash_encoding = { path = "components/zcash_encoding" }
 zcash_note_encryption = { path = "components/zcash_note_encryption" }
-bridgetree = { git = "https://github.com/zcash/incrementalmerkletree.git", rev = "ea1686e8f8f6c1e41aa97251a7eb4fadfd33df47" }
```
nuttycom, line 24: The removal of `bridgetree` as an `orchard` optional dependency means we no longer need this here.

ebfull (approved):

PR #852, opened 2023-06-02
Consolidate sapling functionality in zcash_client_sqlite in a dedicated module.

nuttycom:


zcash_client_sqlite/src/lib.rs
```diff
@@ -43,7 +43,7 @@ use zcash_primitives::{
     consensus::{self, BlockHeight},
     legacy::TransparentAddress,
     memo::{Memo, MemoBytes},
-    sapling::{self, Nullifier},
+    sapling::{self},
```
daira, line 46: Is this `rustfmt`ted?

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -173,6 +228,182 @@ pub(crate) fn select_spendable_sapling_notes<P>(
...
+            rcm,
+            nf,
+            memo,
+            is_change,
+        )
+    } else {
+        // It was there, so grab its row number.
+        stmts.stmt_select_received_note(tx_ref, output.index())
+    }
+}
```
daira, line 405: I have not checked that the code here has only been moved.

daira (approved):
utACK, modulo not checking the changes that should be move-only.

zcash_client_sqlite/src/lib.rs
nuttycom: For some reason rustfmt doesn't collapse this to just `sapling`. This code is likely to change further, so I'm going to leave it like this for now.

PR #853, opened 2023-06-06
doc-only: Document ZIP 321 error variants.

nuttycom:


zcash_client_backend/src/zip321.rs
```diff
@@ -26,12 +26,22 @@ use crate::address::RecipientAddress;
 /// Errors that may be produced in decoding of payment requests.
 #[derive(Debug)]
 pub enum Zip321Error {
+    /// A memo field in the ZIP 321 URI was not properly base-64 encoded
     InvalidBase64(base64::DecodeError),
+    /// A memo value exceeded 512 bytes in length or could not be interpreted as a UTF-8 string
+    /// when using a valid UTF-8 lead byte.
     MemoBytesError(memo::Error),
+    /// The ZIP 321 request included more payments than can be created within a single Zcash
+    /// transaction.
```
daira, line 35:
```suggestion
    /// transaction. The wrapped value is the number of payments in the request.
```

daira (approved):
ACK with non-blocking suggestion. Doc-only.

PR #857, opened 2023-06-09
Use `rusqlite::Connection::prepare_cached` instead of manual statement caching.

nuttycom:
`rusqlite` includes a mechanism for creating prepared statements that
automatically caches them and reuses the caches when possible. This
means that it's unnecessary for us to do our own caching, and also
offers a minor performance improvement in that we don't need to eagerly
prepare statements that we may not execute in the lifetime of a given
`WalletDb` object. It also improves code locality, because the prepared
statements are now adjacent in the code to the parameter assignment
blocks that correspond to those statements.

This also updates a number of `put_x` methods to use sqlite upsert
functionality via the `ON CONFLICT` clause, instead of having to perform
separate inserts and updates.


zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -318,11 +318,11 @@ pub(crate) fn get_sapling_nullifiers<P>(
...
 
 /// Returns the nullifiers for the notes that this wallet is tracking.
-pub(crate) fn get_all_sapling_nullifiers<P>(
-    wdb: &WalletDb<P>,
+pub(crate) fn get_all_sapling_nullifiers(
+    conn: &Connection,
 ) -> Result<Vec<(AccountId, Nullifier)>, SqliteClientError> {
     // Get the nullifiers for the notes we are tracking
-    let mut stmt_fetch_nullifiers = wdb.conn.prepare(
+    let mut stmt_fetch_nullifiers = conn.prepare(
```
daira, line 334: Why `prepare` rather than `prepare_cached` here?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -891,48 +891,107 @@ pub(crate) fn update_expired_notes<P>(
...
+// and `put_sent_output`
+//
+// - If `to` is a Unified address, this is an index into the outputs of the transaction
+//   within the bundle associated with the recipient's output pool.
+// - If `to` is a Sapling address, this is an index into the Sapling outputs of the
+//   transaction.
+// - If `to` is a transparent address, this is an index into the transparent outputs of
+//   the transaction.
+// - If `to` is an internal account, this is an index into the Sapling outputs of the
+//   transaction.
```
daira, line 991: I don't understand what "this" refers to. The returned tuple? In any case the description doesn't seem to match what the function computes.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -891,48 +891,107 @@ pub(crate) fn update_expired_notes<P>(
...
+            to_address, to_account, value, memo)
+        VALUES (
+            :tx, :output_pool, :output_index, :from_account,
+            :to_address, :to_account, :value, :memo)
+        ON CONFLICT (tx, output_pool, output_index) DO UPDATE
+        SET from_account = :from_account,
+            to_address = :to_address,
+            to_account = :to_account,
+            value = :value,
+            memo = IFNULL(:memo, memo)",
```
daira, line 1066: Check whether this has the same issue as https://github.com/zcash/librustzcash/pull/857/commits/c7f759929cd3157039842cd0a4a04c8dc648dc21#r1224660086

zcash_client_sqlite/src/lib.rs
```diff
@@ -107,12 +102,21 @@ impl fmt::Display for NoteId {
...
-pub struct WalletDb<P> {
-    conn: Connection,
+pub struct WalletDb<C, P> {
+    conn: C,
     params: P,
 }
 
-impl<P: consensus::Parameters> WalletDb<P> {
+/// A wrapper for a SQLite transaction affecting the wallet database.
+pub struct WalletTransaction<'conn>(pub(crate) rusqlite::Transaction<'conn>);
```
daira, line 111: This name is ambiguous; maybe `WalletSqlTransaction`?

(`WalletDbTransaction` would still be ambiguous since it could be read as `(WalletDb)Transaction`, not the intended `Wallet(DbTransaction)`.)

zcash_client_sqlite/src/lib.rs
```diff
@@ -32,11 +32,9 @@
 // Catch documentation errors caused by code changes.
 #![deny(rustdoc::broken_intra_doc_links)]
```
daira, line 34: Codecov seems to be complaining that this file has poor test coverage.

daira:
I'm unconvinced by the changes to take `conn: &rusqlite::Connection` (and sometimes `params: &P`) instead of `wdb: &WalletDb<P>`. Aren't these logically operations on the `WalletDb`, so that it's proper for the callee rather than the caller to be extracting the connection and parameters from the `WalletDb`?

daira (commented):
Partial review.

nuttycom:
> I'm unconvinced by the changes to take `conn: &rusqlite::Connection` (and sometimes `params: &P`) instead of `wdb: &WalletDb<P>`. Aren't these logically operations on the `WalletDb`, so that it's proper for the callee rather than the caller to be extracting the connection and parameters from the `WalletDb`?

These are all crate-private methods; taking `&rusqlite::Connection` is useful in circumstances where we want to access these methods directly and don't have a `WalletDb` in scope, such as in a database migration (where all that we have is a `&rusqlite::Transaction` and we can't construct a `WalletDb<rusqlite::Connection, P>` because we don't have an owned `rusqlite::Connection` value at all.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: The only cases where I changed to use `prepare_cached` were those where we were previously caching prepared statements. It might be reasonable to cache here as well; I haven't looked closely to determine whether it would make sense to do so.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -345,63 +345,70 @@ pub(crate) fn get_all_sapling_nullifiers<P>(
...
+        (tx, output_index, account, diversifier, value, rcm, memo, nf, is_change)
+        VALUES
+        (:tx, :output_index, :account, :diversifier, :value, :rcm, :memo, :nf, :is_change)
+        ON CONFLICT (tx, output_index) DO UPDATE
+        SET account = :account,
+            diversifier = :diversifier,
+            value = :value,
+            rcm = :rcm,
+            nf = IFNULL(:nf, nf),
+            memo = IFNULL(:memo, memo),
```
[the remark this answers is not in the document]
nuttycom: This is a direct transcription of https://github.com/zcash/librustzcash/pull/857/files#diff-1e7d5d95a00896e1916695c2509d888591b4eb647ec38b7e750855e0b72c609dL223-L233, as it is used in https://github.com/zcash/librustzcash/pull/857/files#diff-1e7d5d95a00896e1916695c2509d888591b4eb647ec38b7e750855e0b72c609dL743-L746 via https://github.com/zcash/librustzcash/pull/857/files#diff-9cea07a2b90189d927a2737a964494dfea6ab1bc5a51342e481210b93586716eL378-L404

This sort of thing is part of the reason for this consolidation; having this logic about how parameters are transformed scattered across (literally, when you consider inserts) 5 different pieces of code is a disaster for understandability.

zcash_client_sqlite/src/lib.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/857/commits/8ebda21655755d8dcfe49e4abe09adf7f9659d34

zcash_client_sqlite/src/wallet.rs
```diff
@@ -592,13 +618,13 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
-        wdb.conn.execute(
+        conn.execute(
             "DELETE FROM sapling_witnesses WHERE block > ?",
             [u32::from(block_height)],
         )?;
 
         // Rewind received notes
-        wdb.conn.execute(
+        conn.execute(
             "DELETE FROM sapling_received_notes
```
daira, line 628: In some SQL implementations you can [directly delete from an outer join](https://stackoverflow.com/a/5808865/393146) (specifying which table is to be deleted from). I don't know whether this is supported by SQLite.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -592,13 +618,13 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
                 WHERE id_note IN (
```
daira, line 629: In some SQL implementations you can [directly delete from an outer join](https://stackoverflow.com/a/5808865/393146) (specifying which table is to be deleted from). I don't know whether this is supported by SQLite.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -200,10 +197,47 @@ pub(crate) fn get_current_address<P: consensus::Parameters>(
...
+    Ok(())
+}
+
 #[cfg(feature = "transparent-inputs")]
 pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
+    conn: &rusqlite::Connection,
     params: &P,
-    conn: &Connection,
     account: AccountId,
 ) -> Result<HashMap<TransparentAddress, AddressMetadata>, SqliteClientError> {
```
daira, line 242: Nit for preexisting code: handling of invalid addresses is duplicated between here and `get_current_address`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -732,166 +760,322 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+    conn: &rusqlite::Connection,
     height: BlockHeight,
 ) -> Result<(), SqliteClientError> {
-    stmts.stmt_update_expired(height)
+    let mut stmt_update_expired = conn.prepare_cached(
+                    "UPDATE sapling_received_notes SET spent = NULL WHERE EXISTS (
+                        SELECT id_tx FROM transactions
+                        WHERE id_tx = sapling_received_notes.spent AND block IS NULL AND expiry_height < ?
+                    )",
+                )?;
```
daira, line 976: It's odd that this is `rustfmt`ted with such a large indent.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -732,166 +760,322 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+    let (to_address, to_account, pool_type) = recipient_params(params, recipient);
+    let sql_args = named_params![
+        ":tx": &tx_ref,
+        ":output_pool": &pool_code(pool_type),
+        ":output_index": &i64::try_from(output_index).unwrap(),
+        ":from_account": &u32::from(from_account),
+        ":to_address": &to_address,
+        ":to_account": &to_account,
+        ":value": &i64::from(value),
+        ":memo": &memo.filter(|m| *m != &MemoBytes::empty()).map(|m| m.as_slice()),
```
daira, line 1075: `memo.filter(|m| *m != &MemoBytes::empty()).map(|m| m.as_slice())` is used three times I think; consider extracting it to a function.

zcash_client_sqlite/src/wallet/sapling.rs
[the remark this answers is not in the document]
daira: You're technically correct (the best kind of correct), but the point is that if the caller passes `output` with `output.memo()` as `MemoBytes::empty()`, then leaving the existing memo unchanged is wrong. Please file an issue if it isn't to be fixed in this PR.

zcash_client_sqlite/src/wallet/sapling.rs
daira: There appears to be no down-side. This isn't a method that is only used at startup or anything like that.

daira (approved):
utACK modulo comments.

zcash_client_sqlite/src/wallet/sapling.rs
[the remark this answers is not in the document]
nuttycom: The problem here is that we use SQL NULL both to represent the absence of a memo *and* to represent the empty memo. If we want an explicit operation to remove the memo from an output record, we should add that, but I think  we shouldn't make a special case that sending the empty memo removes the existing value of the memo field, overwriting whatever was previously stored there. Under normal circumstances the memo value for an output should never change between calls.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -732,166 +760,322 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+    let (to_address, to_account, pool_type) = recipient_params(params, output.recipient());
+    let sql_args = named_params![
+        ":tx": &tx_ref,
+        ":output_pool": &pool_code(pool_type),
+        ":output_index": &i64::try_from(output.output_index()).unwrap(),
+        ":from_account": &u32::from(from_account),
+        ":to_address": &to_address,
+        ":to_account": &to_account,
+        ":value": &i64::from(output.value()),
+        ":memo": output.memo().filter(|m| *m != &MemoBytes::empty()).map(|m| m.as_slice()),
```
str4d, line 1021:
```suggestion
        ":memo": memo_repr(output.memo()),
```

str4d (commented):
Post-hoc utACK with notes on missing documentation.

zcash_client_sqlite/src/wallet/sapling.rs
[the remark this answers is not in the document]
daira: (Sorry, I missed your reply earlier.)

My suggestion would *remove* a special case. Currently, `memo_repr` converts an empty memo to a `None` option. `memo_repr` is called [here](https://github.com/zcash/librustzcash/pull/857/files#diff-a1724f7547ecabc613a07520eaf7917d9084d4277219d186698b0c8e352495a4R1075) and below. Therefore, in the statements that use `IFNULL(:memo, memo)`, `:memo` will be NULL when the memo passed into the function is `MemoBytes::empty()`. So when `Some(new_memo)` is passed into the function, `new_memo` will replace the existing one *except* when `new_memo == MemoBytes::empty()`.

This is propagating the implementation detail that `MemoBytes::empty()` is represented as an SQL NULL to the visible behaviour of the Rust functions in a way that isn't necessitated by that representation. No-one using this API would expect that if you have stored the memo "foo" and overwrite it with the empty memo, you still get "foo". You should get the same thing as if you'd written the empty memo in the first place (which happens to be not distinguishable from a missing memo, but that's a separate issue).

I take the point that the memo value for an output should not change between calls. It's still irregular and error-prone.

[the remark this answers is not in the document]
nuttycom: Hm. Perhaps a better solution all around would be to store the empty memo as a single `0xf6` byte, and special-case that for both serialization and parsing.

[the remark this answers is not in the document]
daira: Filed as #866.

Issue #768, opened 2023-01-30
Extract "component" crates into a separate repository


str4d:
Due to the `orchard` crate living in a separate repository, we currently have an annoying cyclic dependency between the two repositories, due to `zcash_primitives` depending on `orchard`, which depends on `zcash_note_encryption`. We also intend to move the Sapling protocol logic out into its own repository again (#738).

To improve development interactions between the crates, we should extract the "component crates" into a separate repository and Cargo workspace. By "component crates" I mean "crates that don't depend on the protocols", but this could also be the narrower "crates that the protocols depend on".

str4d:
Here is the current dependency graph between the Rust crates:
```mermaid
graph TD
	n150[equihash]
	n151[f4jumble]
	n230[orchard]
	n232[zcash_note_encryption]
	n336[zcash_address]
	n337[zcash_encoding]
	n338[zcash_client_backend]
	n339[zcash_primitives]
	n340[zcash_proofs]
	n341[zcash_client_sqlite]

	n341 --> n338
	n341 --> n339
	n338 --> n230
	n338 --> n336
	n338 --> n337
	n338 --> n232
	n338 --> n339
	n338 --> n340
	n340 --> n339
	n339 --> n150
	n339 --> n230
	n339 --> n336
	n339 --> n337
	n339 --> n232
	n230 --> n232
	n336 --> n151
	n336 --> n337
```

If we define "component crates" as "crates that don't depend on the protocols" then we get this split:
```mermaid
graph TD
	subgraph components
	n150[equihash]
	n151[f4jumble]
	n232[zcash_note_encryption]
	n337[zcash_encoding]
	end

	subgraph protocols
	n230[orchard]
	end

	subgraph main
	n336[zcash_address]
	n338[zcash_client_backend]
	n339[zcash_primitives]
	n340[zcash_proofs]
	n341[zcash_client_sqlite]
	end

	n341 --> n338
	n341 --> n339
	n338 --> n230
	n338 --> n336
	n338 --> n337
	n338 --> n232
	n338 --> n339
	n338 --> n340
	n340 --> n339
	n339 --> n150
	n339 --> n230
	n339 --> n336
	n339 --> n337
	n339 --> n232
	n230 --> n232
	n336 --> n151
	n336 --> n337
```

If we define "component crates" as "crates that the protocols depend on" then we get this split:
```mermaid
graph TD
	subgraph components
	n232[zcash_note_encryption]
	end

	subgraph protocols
[...]

PR #858, opened 2023-06-17
Document `zcash_note_encryption` non-local dependency rationale & doc work-around.

nathan-at-least:
I had forgotten about the issue in #768 and just ran `cargo doc` and was confused by the warning. It took me a bit to track down via `cargo tree -i zcash_note_encryption`, then I rediscovered #768.

So I figured many devs/consumers who are poking around here may also get confused and I added in-band docs about the issue.

### DIFF
--- a/components/zcash_note_encryption/README.md
+++ b/components/zcash_note_encryption/README.md
@@ -12,6 +12,15 @@ users with their own existing types can similarly implement the trait themselves
 [`zcash_primitives`]: https://crates.io/crates/zcash_primitives
 [`orchard`]: https://crates.io/crates/orchard
 
+## Cross-Workspace Dependency Cycle
+
+There is a cross-workspace cyclic dependency: `zcash_client_backend` (local)
+depends on [`orchard`](https://github.com/zcash/orchard) which depends on this crate
+`zcash_note_encryption`. For this reason `zcash_client_backend` does not depend on the local
+`zcash_note_encryption` crate, but rather the published crate release.
+
+This issue is tracked in [#768](https://github.com/zcash/librustzcash/issues/768).
+
 ## License
 
 Licensed under either of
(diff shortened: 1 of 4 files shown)


nuttycom:
I think that rather than change this documentation, and then immediately have to remove it once we remove `zcash_note_encryption` from the `librustzcash` repository, we should instead just pull `zcash_note_encryption` out of the workspace into its own repository and fix the problem once and for all.

components/zcash_note_encryption/README.md
```diff
@@ -12,6 +12,15 @@ users with their own existing types can similarly implement the trait themselves
 [`zcash_primitives`]: https://crates.io/crates/zcash_primitives
 [`orchard`]: https://crates.io/crates/orchard
 
+## Cross-Workspace Dependency Cycle
+
+There is a cross-workspace cyclic dependency: `zcash_client_backend` (local)
+depends on [`orchard`](https://github.com/zcash/orchard) which depends on this crate
+`zcash_note_encryption`. For this reason `zcash_client_backend` does not depend on the local
+`zcash_note_encryption` crate, but rather the published crate release.
```
daira, line 20: That isn't cyclic, right? The issue is that `zcash_client_backend` might depend on a different version of `zcash_note_encryption` to `orchard`; not that there is any cycle in the dependency graph (which Rust and `cargo` do not support).
```suggestion
There is a cross-workspace dependency that requires care with versioning:
`zcash_client_backend` (local) depends on [`orchard`](https://github.com/zcash/orchard)
which depends on this crate `zcash_note_encryption`. For this reason `zcash_client_backend`
does not depend on the local `zcash_note_encryption` crate, but rather the published crate
release.
```

daira (approved):
utACK modulo comment.

nuttycom:
@nathan-at-least this is now conflicted and you have not responded to review, can you please revisit this?

str4d:
This is obsoleted by #1041.

PR #859, opened 2023-06-18
[Impacts API versioning] Update dependencies to resolve RUSTSEC-2022-0090

nathan-at-least:
This resolves [RUSTSEC-2022-0090](https://rustsec.org/advisories/RUSTSEC-2022-0090), and with both this patch and after I remembered to also run `cargo update`, there are now no `cargo audit` reports.

Note that this updates dependencies that are re-exported in the public interface, so I'm not super clear on if that's invasive for dependent crates or not. If it were invasive, it may imply something about our release versioning and semver, which I haven't thought about too carefully.

### DIFF
--- a/zcash_client_backend/Cargo.toml
+++ b/zcash_client_backend/Cargo.toml
@@ -29,7 +29,7 @@ zcash_primitives = { version = "0.12", path = "../zcash_primitives", default-fea
 # Dependencies exposed in a public API:
 # (Breaking upgrades to these require a breaking upgrade to this crate.)
 # - Data Access API
-time = "0.2"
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK. I also reviewed the `schemer-rusqlite` delta https://github.com/aschampion/schemer/compare/schemer-v0.2.1...schemer-rusqlite-v0.2.2

nuttycom:
@nathan-at-least this is the perfect time to do these updates, because we have not yet made a release of `zcash-client-backend` or `zcash-client-sqlite` that makes use of `zcash_primitive-0.12`

PR #861, opened 2023-06-20
Add interfaces & support for subtree root insertion.

nuttycom:
Builds atop #831

### DIFF
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -66,6 +66,35 @@ use crate::{
 pub mod error;
 use error::Error;
 
+/// A struct containing metadata about a subtree root of the note commitment tree.
+///
+/// This stores the block height at which the leaf that completed the subtree was
+/// added, and the root hash of the complete subtree.
+pub struct CommitmentTreeRoot<H> {
+    subtree_end_height: BlockHeight,
+    root_hash: H,
+}
+
+impl<H> CommitmentTreeRoot<H> {
+    /// Construct a new `CommitmentTreeRoot` from its constituent parts.
+    pub fn from_parts(subtree_end_height: BlockHeight, root_hash: H) -> Self {
+        Self {
+            subtree_end_height,
+            root_hash,
+        }
+    }
+
+    /// Returns the block height at which the leaf that completed the subtree was added.
+    pub fn subtree_end_height(&self) -> BlockHeight {
+        self.subtree_end_height
+    }
+
+    /// Returns the root of the complete subtree.
+    pub fn root_hash(&self) -> &H {
+        &self.root_hash
+    }
+}
+
 /// This trait provides sequential access to raw blockchain data via a callback-oriented
 /// API.
 pub trait BlockSource {
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -36,6 +36,7 @@ use either::Either;
 use rusqlite::{self, Connection};
 use secrecy::{ExposeSecret, SecretVec};
 use std::{borrow::Borrow, collections::HashMap, convert::AsRef, fmt, io, ops::Range, path::Path};
+use wallet::commitment_tree::put_shard_roots;
 
 use incrementalmerkletree::Position;
 use shardtree::{ShardTree, ShardTreeError};
@@ -55,9 +56,10 @@ use zcash_primitives::{
 use zcash_client_backend::{
     address::{AddressMetadata, UnifiedAddress},
     data_api::{
-        self, chain::BlockSource, BlockMetadata, DecryptedTransaction, NullifierQuery, PoolType,
-        Recipient, ScannedBlock, SentTransaction, WalletCommitmentTrees, WalletRead, WalletWrite,
-        SAPLING_SHARD_HEIGHT,
+        self,
+        chain::{BlockSource, CommitmentTreeRoot},
+        BlockMetadata, DecryptedTransaction, NullifierQuery, PoolType, Recipient, ScannedBlock,
+        SentTransaction, WalletCommitmentTrees, WalletRead, WalletWrite, SAPLING_SHARD_HEIGHT,
     },
     keys::{UnifiedFullViewingKey, UnifiedSpendingKey},
     proto::compact_formats::CompactBlock,
@@ -643,9 +645,30 @@ impl<P: consensus::Parameters> WalletCommitmentTrees for WalletDb<rusqlite::Conn
             let mut shardtree = ShardTree::new(shard_store, PRUNING_DEPTH.try_into().unwrap());
             callback(&mut shardtree)?
         };
+
         tx.commit()
             .map_err(|e| ShardTreeError::Storage(Either::Right(e)))?;
         Ok(result)
+    }
+
+    fn put_sapling_subtree_roots(
+        &mut self,
+        start_index: u64,
+        roots: &[CommitmentTreeRoot<sapling::Node>],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        let tx = self
+            .conn
+            .transaction()
+            .map_err(|e| ShardTreeError::Storage(Either::Right(e)))?;
+        put_shard_roots::<_, { sapling::NOTE_COMMITMENT_TREE_DEPTH }, SAPLING_SHARD_HEIGHT>(
+            &tx,
+            SAPLING_TABLES_PREFIX,
+            start_index,
+            roots,
+        )?;
+        tx.commit()
+            .map_err(|e| ShardTreeError::Storage(Either::Right(e)))?;
+        Ok(())
     }
 }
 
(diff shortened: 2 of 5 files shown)


zcash_client_backend/src/data_api/chain.rs
```diff
@@ -66,6 +66,27 @@ use crate::{
 pub mod error;
 use error::Error;
 
+pub struct CommitmentTreeRoot<H> {
```
str4d, line 69: Needs documentation and a changelog addition entry.

zcash_client_sqlite/src/lib.rs
```diff
@@ -643,10 +645,31 @@ impl<P: consensus::Parameters> WalletCommitmentTrees for WalletDb<rusqlite::Conn
...
+        start_index: u64,
+        roots: &[CommitmentTreeRoot<sapling::Node>],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        let tx = self
+            .conn
+            .transaction()
+            .map_err(|e| ShardTreeError::Storage(Either::Right(e)))?;
+        put_shard_roots::<_, { sapling::NOTE_COMMITMENT_TREE_DEPTH }, SAPLING_SHARD_HEIGHT>(
+            &tx,
+            "sapling",
```
str4d, line 665: Use the constant.

zcash_client_sqlite/src/lib.rs
```diff
@@ -674,6 +697,19 @@ impl<'conn, P: consensus::Parameters> WalletCommitmentTrees for WalletDb<SqlTran
...
     }
+
+    fn put_sapling_subtree_roots(
+        &mut self,
+        start_index: u64,
+        roots: &[CommitmentTreeRoot<sapling::Node>],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        put_shard_roots::<_, { sapling::NOTE_COMMITMENT_TREE_DEPTH }, SAPLING_SHARD_HEIGHT>(
+            self.conn.0,
+            "sapling",
```
str4d, line 708: Use the constant.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -257,23 +262,29 @@ type Error = Either<io::Error, rusqlite::Error>;
...
     .optional()
     .map_err(Either::Right)?
-    .map(|shard_data| {
+    .map(|(shard_data, root_hash)| {
         let shard_tree = read_shard(&mut Cursor::new(shard_data)).map_err(Either::Left)?;
-        Ok(LocatedPrunableTree::from_parts(shard_root, shard_tree))
+        let located_tree = LocatedPrunableTree::from_parts(shard_root_addr, shard_tree);
+        if let Some(root_hash_data) = root_hash {
+            let root_hash = H::read(Cursor::new(root_hash_data)).map_err(Either::Left)?;
+            Ok(located_tree.reannotate_root(Some(Rc::new(root_hash))))
```
str4d, line 284: What does this do? #831 gave me the impression that the `root_hash` column was only being stored for caching purposes, and that it was re-derivable from `shard_data`.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -746,15 +757,98 @@ pub(crate) fn truncate_checkpoints(
...
+                        id: (),
+                        is_marked: false,
+                    },
+                )
+            }),
+        )
+        .map_err(ShardTreeError::Insert)?
+        .expect("slice of inserted roots was verified to be nonempty");
+
+    put_cap(conn, table_prefix, cap_result.subtree.take_root()).map_err(ShardTreeError::Storage)?;
```
str4d, line 799: `subtree.take_root()` is a confusing method name, as it suggests this takes the root node out of the cap tree leaving `subtree` usable, when in fact it consumes `self` and returns the entire cap tree structure (so that we can later reconstruct it using `LocatedTree::from_parts` above. I expect it would be more accurately named `into_tree` or `into_inner` or `into_unlocated`.

We should do a full pass over the `subtree` crate to fix this and similar non-standard usages of the `take_` prefix.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -746,15 +757,98 @@ pub(crate) fn truncate_checkpoints(
...
+    for (root, i) in roots.iter().zip(0u64..) {
+        // To avoid deserializing the subtree just to annotate its root, we simply cache the downloaded
+        // root alongside of the persisted subtree and reannotate the root, handling conflicts, at the
+        // time that we deserialize the tree.
+        let mut stmt = conn
+            .prepare_cached(&format!(
+            "INSERT INTO {}_tree_shards (shard_index, subtree_end_height, root_hash, shard_data)
+            VALUES (:shard_index, :subtree_end_height, :root_hash, :shard_data)
+            ON CONFLICT (shard_index) DO UPDATE
+            SET subtree_end_height = :subtree_end_height, root_hash = :root_hash",
```
str4d, line 810: When there is a conflict, why isn't `shard_data` updated? The comment after this says that `shard_data` is only used on the insert side of the upsert, but none of the comments appear to explain why.

It's also possible that I am getting confused because "root" is being used in some places to refer to an entire subtree and in others to refer to just the root node (and its hash) of that subtree.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -746,15 +757,98 @@ pub(crate) fn truncate_checkpoints(
...
+            table_prefix
+        ))
+            .map_err(|e| ShardTreeError::Storage(Either::Right(e)))?;
+
+        // The `shard_data` value will only be used in the case that no tree already exists.
+        let mut shard_data: Vec<u8> = vec![];
+        let tree = PrunableTree::empty().reannotate_root(Some(Rc::new(root.root_hash().clone())));
+        write_shard(&mut shard_data, &tree)
+            .map_err(|e| ShardTreeError::Storage(Either::Left(e)))?;
```
str4d, line 819: It is not at all clear to me why this is useful. AFAICT `reannotate_root` replaces the annotation on the root node of the tree, but only if the root node is a `Node::Parent`; this is an empty tree with a placeholder root node `Node::Nil`, so the `reannotate_root` does nothing.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -804,4 +898,65 @@ mod tests {
...
+
+    #[test]
+    fn put_shard_roots() {
+        let data_file = NamedTempFile::new().unwrap();
+        let mut db_data = WalletDb::for_path(data_file.path(), tests::network()).unwrap();
+        data_file.keep().unwrap();
+
+        init_wallet_db(&mut db_data, None).unwrap();
+        let tx = db_data.conn.transaction().unwrap();
+        let store = SqliteShardStore::<_, String, 3>::from_connection(&tx, "sapling").unwrap();
```
str4d, line 910: Use the constant.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -804,4 +898,65 @@ mod tests {
...
+                    BlockHeight::from((idx + 1) * 3),
+                    if idx == 3 {
+                        "abcdefgh".to_string()
+                    } else {
+                        idx.to_string()
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
+        super::put_shard_roots::<_, 6, 3>(store.conn, "sapling", 0, &roots).unwrap();
```
str4d, line 926: Use the constant.

str4d (commented):
Reviewed c2c52f8873757d4881035f3cf01cf67583ca97ab.

zcash_client_sqlite/src/wallet/commitment_tree.rs
nuttycom: Ah, good point, this should be a `PrunableTree::leaf()` call.

zcash_client_sqlite/src/wallet/commitment_tree.rs
nuttycom: The root can only be re-derived from `shard_data` if the subtree is complete; if there are any `nil` children then we can't re-derive the root, so we have to store it as an annotation.

zcash_client_sqlite/src/wallet/commitment_tree.rs
nuttycom: Yeah, I keep forgetting about the `into` convention.

zcash_client_sqlite/src/wallet/commitment_tree.rs
nuttycom: If no entry exists at this index, we need to write a valid serialized tree; however, if a tree already exists we don't want to spend the cycles to deserialize and reserialize it just to change the annotation on the root; instead, we just cache the root and then can reannotate the root with that hash when we actually need to use or update the deserialized subtree at some later point.

str4d: I think I've figured out the source of my confusion. `put_shard_roots` only changes the value of the root node of each subtree in the range, it does not touch the rest of the subtree's data. But the `subtree` crate uses `root` to refer to the entire tree rooted at a given node, so my brain was expecting the subtree's data to be altered by this method.

str4d: Oh wait, is it better to think of it as we _should_ be updating `subtree_data` as well, but the only part of it that would get updated is the annotation because none of the rest of the tree structure stores the root's information? That would clarify the comment above `stmt`.

nuttycom: That's correct; if it didn't require an expensive round-trip through the serialization logic, we'd reannotate the root node with its hash immediately here.

str4d (approved):
utACK 1e5b23aeba4e1da4f814fb75191c6f99cb724c3e

PR #862, opened 2023-06-20
Add derive Ord to Node

AloeareV:
For any of the new features in incrementalmerkletree, to make a `BridgeTree ` or `ShardTree` of  `Node`s, we need an `Ord` impl

### DIFF
--- a/zcash_primitives/src/sapling/tree.rs
+++ b/zcash_primitives/src/sapling/tree.rs
@@ -64,7 +64,7 @@ pub fn merkle_hash(depth: usize, lhs: &[u8; 32], rhs: &[u8; 32]) -> [u8; 32] {
 }
 
 /// A node within the Sapling commitment tree.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


nuttycom:
> For any of the new features in incrementalmerkletree, to make a `BridgeTree ` or `ShardTree` of `Node`s, we need an `Ord` impl

This is not correct; checkpoint identifiers must be ordered but there is no need for hashes to be. Seehttps://github.com/zcash/librustzcash/pull/831/files#diff-57d91750cf8573e5cfebdee8bd31b2d635cb02adf7f5e5fdd20ff2253e43782fR616-R674 and https://github.com/zcash/librustzcash/pull/831/files#diff-802691a03e5797ebc59e5aa3bd80d3d5c0067754ac0a74f7a7a5666420780d7d for an example of a `ShardStore` implementation using `sapling::Node`

AloeareV:
It's needed for BridgeTree....but it's not needed for ShardTree, which is the better solution anyway. Closing!

PR #863, opened 2023-06-23
Add Orchard to transaction builder

nuttycom:
Subsumes #583

### DIFF
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -7,6 +7,10 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Removed
+- Removed `impl {PartialEq, Eq} for transaction::builder::Error` in favor
+  of using `assert_matches!` where error comparisons are required.
+
 ## [0.12.0] - 2023-06-06
 ### Added
 - `zcash_primitives::transaction`:
@@ -27,7 +31,7 @@ and this library adheres to Rust's notion of
   `incrementalmerkletree::Hashable` and `merkle_tree::HashSer`.
 - The `Hashable` bound on the `Node` parameter to the `IncrementalWitness`
   type has been removed.
-- `sapling::SAPLING_COMMITMENT_TREE_DEPTH_U8` and `sapling::SAPLING_COMMITMENT_TREE_DEPTH` 
+- `sapling::SAPLING_COMMITMENT_TREE_DEPTH_U8` and `sapling::SAPLING_COMMITMENT_TREE_DEPTH`
   have been removed; use `sapling::NOTE_COMMITMENT_TREE_DEPTH` instead.
 - `merkle_tree::{CommitmentTree, IncrementalWitness, MerklePath}` have been removed in
   favor of versions of these types that are now provided by the
@@ -89,7 +93,7 @@ and this library adheres to Rust's notion of
 - The bounds on the `H` parameter to the following methods have changed:
   - `merkle_tree::incremental::read_frontier_v0`
   - `merkle_tree::incremental::read_auth_fragment_v1`
-- The depth of the `merkle_tree::{CommitmentTree, IncrementalWitness, and MerklePath}` 
+- The depth of the `merkle_tree::{CommitmentTree, IncrementalWitness, and MerklePath}`
   data types are now statically constrained using const generic type parameters.
 - `transaction::fees::fixed::FeeRule::standard()` now uses the ZIP 317 minimum fee
   (10000 zatoshis rather than 1000 zatoshis) as the fixed fee. To be compliant with
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -15,11 +15,12 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - `zcash_primitives::transaction`:
-  - `builder::Builder::{new, new_with_rng}` now takes an optional `orchard_anchor` 
+  - `builder::Builder::{new, new_with_rng}` now take an optional `orchard_anchor` 
     argument which must be provided in order to enable Orchard spends and recipients.
-  - `builder::Builder::test_only_new_with_rng`
-    now returns an existential type: `Builder<'a, P, impl RngCore + CryptoRng>` 
-    instead of `Builder<'a, P, R>`
+  - All `builder::Builder` methods now require the bound `R: CryptoRng` on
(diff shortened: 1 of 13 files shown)


nuttycom:
Subsumes #583

nuttycom (commented):
utACK 4fbdd64c89703053eeaeee51750ea6f5b93a3ec9

AloeareV (commented):
Ah, no more `transaction::components::orchard::builder` mod, with all that `MaybeOrchard` stuff... It means that a builder now has the orchard-related methods even if it's not provided with an orchard anchor, which is now a runtime error when you try to call them with no anchor...but on the other hand, it's makes the implementation so much cleaner. 

It's a tradeoff, but one that I'm fine with! The other change I've noticed is assert_matches, which just makes more sense than `Eq`able errors

nuttycom:
> Ah, no more `transaction::components::orchard::builder` mod, with all that `MaybeOrchard` stuff... It means that a builder now has the orchard-related methods even if it's not provided with an orchard anchor, which is now a runtime error when you try to call them with no anchor...but on the other hand, it's makes the implementation so much cleaner.

Not only that, there was *already* a runtime error that had to be raised if NU5 wasn't active, and so statically making those methods unavailable strictly _increased_ the space of types that implementations had to deal with without reducing the number of possible runtime errors at all. So overall, the `MaybeOrchard` stuff wasn't paying its way.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -66,6 +66,15 @@ pub enum Error<FeeError> {
...
     /// An error occurred in constructing the Sapling parts of a transaction.
     SaplingBuild(sapling_builder::Error),
+    /// An error occurred in constructing the Orchard parts of a transaction.
+    OrchardBuild(orchard::builder::BuildError),
+    /// An error occurred in adding an Orchard Spend a transaction.
+    OrchardSpend(orchard::builder::SpendError),
+    /// An error occurred in adding an Orchard Output a transaction.
+    OrchardRecipient(orchard::builder::OutputError),
+    /// Orchard is not yet active on the network, and the user attempted to add
+    /// Orchard parts to a transaction.
```
str4d, line 76:
```suggestion
    /// The builder was constructed without an Orchard anchor, but an Orchard spend or
    /// recipient was added. This might be because Orchard is not yet active on the
    /// network.
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -229,6 +275,48 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+    /// Adds an Orchard note to be spent in this bundle.
+    ///
+    /// Returns an error if the given Merkle path does not have the required anchor for
+    /// the given note.
+    pub fn add_orchard_spend<FR: FeeRule>(
+        &mut self,
+        sk: orchard::keys::SpendingKey,
+        note: orchard::Note,
+        merkle_path: orchard::tree::MerklePath,
+    ) -> Result<(), Error<FR::Error>> {
```
str4d, line 287: Hmm, the other adders (`add_sapling_spend` etc) return the underlying protocol-specific errors, instead of mapping to `builder::Error`. We should probably do the same here for consistency, which would enable removing the `FR` generic parameter.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -327,6 +422,17 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
                 self.transparent_builder.outputs(),
                 self.sapling_builder.inputs().len(),
                 self.sapling_builder.bundle_output_count(),
+                match std::cmp::max(
+                    self.orchard_builder
+                        .as_ref()
+                        .map_or_else(|| 0, |builder| builder.outputs().len()),
```
str4d, line 428:
```suggestion
                        .map_or(0, |builder| builder.outputs().len()),
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -327,6 +422,17 @@ impl<'a, P: consensus::Parameters, R: RngCore> Builder<'a, P, R> {
...
+                    self.orchard_builder
+                        .as_ref()
+                        .map_or_else(|| 0, |builder| builder.spends().len()),
```
str4d, line 431:
```suggestion
                        .map_or(0, |builder| builder.spends().len()),
```

zcash_primitives/CHANGELOG.md
```diff
@@ -7,6 +7,32 @@ and this library adheres to Rust's notion of
...
 
+### Added
+- `zcash_primitives::transaction::builder`:
+  - `Builder::add_orchard_spend`
+  - `Builder::add_orchard_output`
+- `zcash_primitives::transaction::components::orchard::builder` module
+
+### Changed
+- `zcash_primitives::transaction`:
+  - `builder::Builder::{new, new_with_rng}` now takes an optional `orchard_anchor`
```
str4d, line 18:
```suggestion
  - `builder::Builder::{new, new_with_rng}` now take an optional `orchard_anchor` 
```

zcash_primitives/CHANGELOG.md
```diff
@@ -7,6 +7,32 @@ and this library adheres to Rust's notion of
...
+  - `Builder::add_orchard_output`
+- `zcash_primitives::transaction::components::orchard::builder` module
+
+### Changed
+- `zcash_primitives::transaction`:
+  - `builder::Builder::{new, new_with_rng}` now takes an optional `orchard_anchor` 
+    argument which must be provided in order to enable Orchard spends and recipients.
+  - `builder::Builder::test_only_new_with_rng`
+    now returns an existential type: `Builder<'a, P, impl RngCore + CryptoRng>` 
+    instead of `Builder<'a, P, R>`
```
str4d, line 22:
```suggestion
  - All `builder::Builder` methods now require the bound `R: CryptoRng` on
    `Builder<'a, P, R>`. A non-`CryptoRng` randomness source is still accepted
    by `builder::Builder::test_only_new_with_rng`, which **MUST NOT** be used in
    production.
```

zcash_primitives/src/transaction/builder.rs
nuttycom: We can't, because we're returning `Error::OrchardAnchorNotAvailable`.

nuttycom: However, this should be universally quantified without a bound.

str4d (approved):
utACK

PR #864, opened 2023-06-28
Replace `directories` dependency with `home, known-folders, xdg`

str4d:
`directories 5.0.1` added a dependency on `option-ext`, which is licensed as MPL (a copyleft license), and which upstream intends to keep (dirs-dev/dirs-sys-rs#21). The replacement dependencies are all licensed as `MIT OR Apache-2.0`.

### DIFF
--- a/zcash_proofs/CHANGELOG.md
+++ b/zcash_proofs/CHANGELOG.md
@@ -6,6 +6,9 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Replaced internal `directories` dependency which now transitively depends on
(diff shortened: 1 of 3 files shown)


nuttycom (dismissed):
utACK. Amazing that it takes 3 separate dependencies to handle this, all for the sake of a method on `Option` that ought to be in the stdlib anyway.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/730d8e80350198428ffbc40821283d1193b77da3..6f4d68f93580966b5cba575f1f450fffbe3e8b97) to add a missing `Option::map` for Windows.

nuttycom (approved):
re-utACK 6f4d68f93580966b5cba575f1f450fffbe3e8b97

PR #867, opened 2023-07-03
zcash_client_backend: Make scan range bounds required in `data_api::chain::scan_cached_blocks`

nuttycom:
Builds on #861

### DIFF
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -51,6 +51,7 @@
 //! ```
 
 use zcash_primitives::{
+    block::BlockHash,
     consensus::{self, BlockHeight},
     sapling::{self, note_encryption::PreparedIncomingViewingKey},
     zip32::Scope,
@@ -196,6 +197,19 @@ where
             // check block continuity
             if let Some(scan_error) = check_continuity(&block, continuity_check_metadata.as_ref()) {
                 return Err(Error::Scan(scan_error));
+            }
+
+            if from_height == BlockHeight::from(0) {
+                // We can always derive a valid `continuity_check_metadata` for the
+                // genesis block, even if the block source doesn't have
+                // `sapling_commitment_tree_size`. So briefly set it to a dummy value that
+                // ensures the `map` below produces the correct genesis block value.
+                assert!(continuity_check_metadata.is_none());
+                continuity_check_metadata = Some(BlockMetadata::from_parts(
+                    BlockHeight::from(0),
+                    BlockHash([0; 32]),
+                    0,
+                ));
             }
             continuity_check_metadata = continuity_check_metadata.as_ref().map(|m| {
                 BlockMetadata::from_parts(
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -45,7 +45,7 @@
 //! // At this point, the cache and scanned data are locally consistent (though not
 //! // necessarily consistent with the latest chain tip - this would be discovered the
 //! // next time this codepath is executed after new blocks are received).
-//! scan_cached_blocks(&network, &block_source, &mut db_data, None, None)
+//! scan_cached_blocks(&network, &block_source, &mut db_data, BlockHeight::from(0), 10)
 //! # }
 //! # }
 //! ```
@@ -60,8 +60,10 @@ use crate::{
     data_api::{NullifierQuery, WalletWrite},
     proto::compact_formats::CompactBlock,
     scan::BatchRunner,
-    scanning::{add_block_to_runner, scan_block_with_runner},
+    scanning::{add_block_to_runner, check_continuity, scan_block_with_runner},
 };
+
+use super::BlockMetadata;
 
 pub mod error;
 use error::Error;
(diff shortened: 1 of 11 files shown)


zcash_client_backend/src/scanning.rs
```diff
@@ -124,9 +124,21 @@ pub enum ScanError {
...
+        computed: u32,
+    },
+
+    /// The size of the note commitment tree for the given protocol was not provided as part of a
+    /// [`CompactBlock`] being scanned, making it impossible to construct the nullifier for a
+    /// detected note.
+    TreeSizeUnknown {
+        protocol: ShieldedProtocol,
+        at_height: BlockHeight,
+    },
```
str4d, line 141: Changes should be noted in the changelog. IIRC `TreeSizeUnknown` case was previously added in this set of PRs, so its entry just needs to be adjusted, but `TreeSizeMismatch` is new.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -133,8 +135,8 @@ pub fn scan_cached_blocks<ParamsT, DbT, BlockSourceT>(
     params: &ParamsT,
     block_source: &BlockSourceT,
     data_db: &mut DbT,
-    from_height: Option<BlockHeight>,
-    limit: Option<u32>,
+    from_height: BlockHeight,
+    limit: u32,
```
str4d, line 139: Mention in changelog

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -170,61 +172,73 @@ where
...
+            if let Some(scan_error) = check_continuity(&block, continuity_check_metadata.as_ref()) {
+                return Err(Error::Scan(scan_error));
+            }
+            continuity_check_metadata = continuity_check_metadata.as_ref().map(|m| {
+                BlockMetadata::from_parts(
+                    block.height(),
+                    block.hash(),
+                    m.sapling_tree_size()
+                        + u32::try_from(block.vtx.iter().map(|tx| tx.outputs.len()).sum::<usize>())
+                            .unwrap(),
```
str4d, line 198: We iterated over `block.vtx` inside `check_continuity` in order to validate its chain metadata if present, so we should use it if present to avoid duplicate iteration:
```suggestion
                    block.chain_metadata
                        .as_ref()
                        .map(|m| m.sapling_commitment_tree_size)
                        .unwrap_or_else(|| {
                            m.sapling_tree_size()
                                + u32::try_from(block.vtx.iter().map(|tx| tx.outputs.len())
                                    .sum::<usize>())
                                    .unwrap()
                        }),
```

str4d (commented):
utACK 8300c4ab89795318c8564e0099b360a6ead3125b with suggestions. Blocked on #861 merging.

zcash_client_backend/src/scanning.rs
nuttycom: Since the `Chain` error type was removed and restructured, I just note `ScanError` as being an entirely new addition; a user migrating from the `Chain` and `ChainError` types will need to look at `ScanError` in its entirety anyway.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -149,61 +180,81 @@ where
...
+    let mut continuity_check_metadata = prior_block_metadata;
+    block_source.with_blocks::<_, DbT::Error>(
+        Some(from_height),
+        Some(limit),
+        |block: CompactBlock| {
+            // check block continuity
+            if let Some(scan_error) = check_continuity(&block, continuity_check_metadata.as_ref()) {
+                return Err(Error::Scan(scan_error));
+            }
+            continuity_check_metadata = continuity_check_metadata.as_ref().map(|m| {
```
str4d, line 200: There's a small issue here: if the user calls this method with `from_height = 0`, `prior_block_metadata` was `None`, which means that `continuity_check_metadata` is initialized as `None`, and this `map` propagates that for all time, even if every block in `block_source` has `sapling_commitment_tree_size` set correctly.

The fix here is to recognise that if `from_height = 0`, then the previous block's continuity metadata implicitly has `sapling_commitment_tree_size = 0`, so we can always derive a correct `continuity_check_metadata` for height 0 itself. So we should special-case this height for correctness (of downstream usage on chains that activated Sapling from genesis):
```suggestion
            if from_height == BlockHeight::from(0) {
                // We can always derive a valid `continuity_check_metadata` for the
                // genesis block, even if the block source doesn't have
                // `sapling_commitment_tree_size`. So briefly set it to a dummy value that
                // ensures the `map` below produces the correct genesis block value.
                assert!(continuity_check_metadata.is_none());
                continuity_check_metadata =
                    Some(BlockMetadata::from_parts(0, BlockHash([0; 32]), 0));
            }
            continuity_check_metadata = continuity_check_metadata.as_ref().map(|m| {
```

zcash_client_backend/src/scanning.rs
```diff
@@ -140,8 +152,11 @@ impl fmt::Display for ScanError {
             ScanError::BlockHeightDiscontinuity { previous_tip, new_height } => {
                 write!(f, "Block height discontinuity at height {}; next height is : {}", previous_tip, new_height)
             }
-            ScanError::SaplingTreeSizeUnknown { at_height } => {
-                write!(f, "Unable to determine Sapling note commitment tree size at height {}", at_height)
+            ScanError::TreeSizeMismatch { protocol, at_height, given, computed } => {
+                write!(f, "The the {:?} note commitment tree size provided by a compact block did not match the expected size at height {}; given {}, expected {}", protocol, at_height, given, computed)
```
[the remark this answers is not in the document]
str4d: This was not resolved.

str4d (changes requested):
utACK 98147e0d7dda54ce451a28e8f8cf87d2571e19ba after one fix.

str4d (approved):
utACK 81a32f284e4d411b04d98a1b12a844ed4be85bc0

PR #868, opened 2023-07-06
Restore `zcash_client_backend::data_api::chain::validate_chain`

nuttycom:
Builds atop #867

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -63,6 +63,13 @@ pub trait WalletRead {
     /// directly.
     type TxRef: Copy + Debug + Eq + Ord;
 
+    /// Returns the wallet's view of the chain tip to the given depth.
+    ///
+    /// This may return fewer than `depth` blocks worth of data if insufficient block data is
+    /// available in the wallet database to provide metadata contiguous blocks to the requested
+    /// depth. Metadata values for returned blocks are sequential and in height order.
+    fn chain_tip(&self, depth: usize) -> Result<Vec<BlockMetadata>, Self::Error>;
+
     /// Returns the minimum and maximum block heights for stored blocks.
     ///
     /// This will return `Ok(None)` if no block data is present in the database.
@@ -616,6 +623,10 @@ pub mod testing {
         type Error = ();
         type NoteRef = u32;
         type TxRef = TxId;
+
+        fn chain_tip(&self, _depth: usize) -> Result<Vec<BlockMetadata>, Self::Error> {
+            Ok(vec![])
+        }
 
         fn block_height_extrema(&self) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error> {
             Ok(None)
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -63,7 +63,8 @@ pub trait WalletRead {
     /// directly.
     type TxRef: Copy + Debug + Eq + Ord;
 
-    /// Returns the wallet's view of the chain tip to the given depth.
+    /// Returns the wallet's view of the chain tip to the given depth, in ascending
+    /// block height order.
     ///
     /// This may return fewer than `depth` blocks worth of data if insufficient block data is
     /// available in the wallet database to provide metadata contiguous blocks to the requested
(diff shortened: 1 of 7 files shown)


zcash_client_sqlite/src/chain.rs
```diff
@@ -26,10 +26,14 @@ pub mod migrations;
 /// Starting at `from_height`, the `with_row` callback is invoked with each block retrieved from
 /// the backing store. If the `limit` value provided is `None`, all blocks are traversed up to the
 /// maximum height.
+///
+/// # Panics
+///
+/// Panics if the provided `limit` value exceeds the range of a u32
```
str4d, line 32: Rather than panicking, we could just saturate; it doesn't really matter either way as `u32::MAX` is the maximum block height.

zcash_client_sqlite/src/chain.rs
```diff
@@ -194,11 +198,15 @@ pub(crate) fn blockmetadb_find_block(
 /// Starting at `from_height`, the `with_row` callback is invoked with each block retrieved from
 /// the backing store. If the `limit` value provided is `None`, all blocks are traversed up to the
 /// maximum height for which metadata is available.
+///
+/// # Panics
+///
+/// Panics if the provided `limit` value exceeds the range of a u32
```
str4d, line 204: Ditto.

zcash_client_sqlite/src/chain.rs
```diff
@@ -52,7 +56,7 @@ where
     let mut rows = stmt_blocks
         .query(params![
             from_height.map_or(0u32, u32::from),
-            limit.unwrap_or(u32::max_value()),
+            limit.map_or(u32::max_value(), |l| u32::try_from(l).unwrap())
```
daira, line 59:
```suggestion
            limit.and_then(|l| u32::try_from(l).ok()).unwrap_or(u32::MAX)
```

zcash_client_backend/src/data_api.rs
```diff
@@ -63,6 +63,13 @@ pub trait WalletRead {
     /// directly.
     type TxRef: Copy + Debug + Eq + Ord;
 
+    /// Returns the wallet's view of the chain tip to the given depth.
```
daira, line 66: Document the ordering.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -114,13 +115,71 @@ pub trait BlockSource {
...
+/// This function is built on the core assumption that the information provided in the
+/// block source is more likely to be accurate than the previously-scanned information.
+/// This follows from the design (and trust) assumption that the `lightwalletd` server
+/// provides accurate block information as of the time it was requested.
+///
+/// Arguments:
+/// - `block_source` Source of compact blocks
+/// - `data_db` Source of wallet chain tip data
+/// - `validation_depth` the number of blocks prior to the chain tip to validate against
+///   blocks available from the block source.
```
daira, line 138: Document what happens for `validation_depth == 0`.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -114,13 +115,71 @@ pub trait BlockSource {
...
+/// - `limit` the maximum number of blocks that will be valididated. If `None`
```
daira, line 139:
```suggestion
/// - `limit` the maximum number of blocks that will be validated. If `None`
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -114,13 +115,71 @@ pub trait BlockSource {
...
+/// provides accurate block information as of the time it was requested.
+///
+/// Arguments:
+/// - `block_source` Source of compact blocks
+/// - `data_db` Source of wallet chain tip data
+/// - `validation_depth` the number of blocks prior to the chain tip to validate against
+///   blocks available from the block source.
+/// - `limit` the maximum number of blocks that will be validated. If `None`
+///   is provided, there will be no limit set to the validation and upper bound of the
+///   validation range will be the latest height present in the `block_source`.
```
daira, line 141: It is unclear how this would actually be used to validate in multiple calls. Consider removing this parameter, because `validation_depth` already bounds the maximum amount of work done.

Also consider making this a new API with a better name rather than calling it `validate_chain`.

PR #869, opened 2023-07-06
Replace `put_block` with `put_blocks`

nuttycom:


zcash_client_sqlite/src/error.rs
```diff
@@ -57,6 +57,9 @@ pub enum SqliteClientError {
     /// different hash. This indicates that a required rewind was not performed.
     BlockConflict(BlockHeight),
 
+    /// A range of blocks provided to the database as a unit was non-sequential
+    NonSequentialBlocks,
```
str4d, line 61: Non-blocking: Note this in the changelog.

str4d (approved):
ACK d55fa094644f989371275dd5480372141e2f9fed.

PR #870, opened 2023-07-06
zcash_client_sqlite: Fix `WalletDb::get_transaction` for unmined txs

str4d:


nuttycom (dismissed):
utACK

I'll note that we do also store an `expiry_height` value in the `transactions` table, which I believe should be non-null for all transactions created by the wallet. Should we just use that to obtain the branch ID?

str4d:
> I'll note that we do also store an `expiry_height` value in the `transactions` table, which I believe should be non-null for all transactions created by the wallet. Should we just use that to obtain the branch ID?

Hmm, that raises an important point: if a transaction has an expiry height of zero, then it doesn't expire, so we can't use that as a way to derive the consensus branch ID. It happens that for transactions created by our wallet, we always set it to non-zero, so indeed we should just query that column instead of the read-and-clone approach. But we should still defer to the block height if we have it (and then we can just keep returning an error if it's both unmined and has no expiry height).

nuttycom (commented):
utACK modulo Clippy errros

nuttycom (dismissed):
utACK d3e4514

daira (approved):
utACK

PR #875, opened 2023-07-19
Use an upsert for `FsBlockDb::write_block_metadata`

str4d:


daira (approved):
utACK

nuttycom (commented):
post-hoc utACK with request for documentation:

This PR doesn't document what motivated this change; I presume it expects a usage pattern where the wallet goes offline, and when it comes back on it re-downloads a series of blocks that have changed due to a reorg. This should be fine given that integrity checks are performed via scan_cached_blocks, but it would be good for the PR description to clarify the motivation.

PR #878, opened 2023-07-21
zcash_client_sqlite: Maintain a nullifier map from out-of-order scanning

str4d:
Also fixes the `WalletDb::block_fully_scanned` implementation, which this PR depends on.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -300,6 +300,7 @@ pub struct ScannedBlock<Nf> {
     metadata: BlockMetadata,
     block_time: u32,
     transactions: Vec<WalletTx<Nf>>,
+    sapling_nullifier_map: Vec<(TxId, u16, Vec<sapling::Nullifier>)>,
     sapling_commitments: Vec<(sapling::Node, Retention<BlockHeight>)>,
 }
 
@@ -308,12 +309,14 @@ impl<Nf> ScannedBlock<Nf> {
         metadata: BlockMetadata,
         block_time: u32,
         transactions: Vec<WalletTx<Nf>>,
+        sapling_nullifier_map: Vec<(TxId, u16, Vec<sapling::Nullifier>)>,
         sapling_commitments: Vec<(sapling::Node, Retention<BlockHeight>)>,
     ) -> Self {
         Self {
             metadata,
             block_time,
             transactions,
+            sapling_nullifier_map,
             sapling_commitments,
         }
     }
@@ -336,6 +339,10 @@ impl<Nf> ScannedBlock<Nf> {
 
     pub fn transactions(&self) -> &[WalletTx<Nf>] {
         &self.transactions
+    }
+
+    pub fn sapling_nullifier_map(&self) -> &[(TxId, u16, Vec<sapling::Nullifier>)] {
+        &self.sapling_nullifier_map
     }
 
     pub fn sapling_commitments(&self) -> &[(sapling::Node, Retention<BlockHeight>)] {
@@ -498,7 +505,7 @@ pub trait WalletWrite: WalletRead {
     /// `blocks` must be sequential, in order of increasing block height
     fn put_blocks(
         &mut self,
-        block: Vec<ScannedBlock<sapling::Nullifier>>,
+        blocks: Vec<ScannedBlock<sapling::Nullifier>>,
     ) -> Result<Vec<Self::NoteRef>, Self::Error>;
 
     /// Updates the wallet's view of the blockchain.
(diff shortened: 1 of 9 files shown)


ebfull (approved):
Pair reviewed this with @str4d. ACK

zcash_client_backend/src/data_api.rs
```diff
@@ -300,6 +300,7 @@ pub struct ScannedBlock<Nf> {
     metadata: BlockMetadata,
     block_time: u32,
     transactions: Vec<WalletTx<Nf>>,
+    sapling_nullifier_map: Vec<(TxId, u16, Vec<sapling::Nullifier>)>,
```
daira, line 303: Please add documentation for what the `u16` is.

zcash_client_backend/src/data_api.rs
```diff
@@ -338,6 +341,10 @@ impl<Nf> ScannedBlock<Nf> {
         &self.transactions
     }
 
+    pub fn sapling_nullifier_map(&self) -> &[(TxId, u16, Vec<sapling::Nullifier>)] {
```
daira, line 344: Again please add documentation (it can be in one place and referred to from the other place).

zcash_client_backend/src/scanning.rs
```diff
@@ -355,36 +355,44 @@ pub(crate) fn scan_block_with_runner<
 
     let compact_block_tx_count = block.vtx.len();
     let mut wtxs: Vec<WalletTx<K::Nf>> = vec![];
+    let mut sapling_nullifier_map = Vec::with_capacity(block.vtx.len());
     let mut sapling_note_commitments: Vec<(sapling::Node, Retention<BlockHeight>)> = vec![];
     for (tx_idx, tx) in block.vtx.into_iter().enumerate() {
         let txid = tx.txid();
+        let tx_index =
+            u16::try_from(tx.index).expect("Cannot fit more than 2^16 transactions in a block");
```
daira, line 363: Does the lightwalletd security model explicitly permit lightwalletd to cause a connected wallet to panic? (I know there are other cases where it can.)

zcash_client_backend/src/scanning.rs
```diff
@@ -505,7 +513,7 @@ pub(crate) fn scan_block_with_runner<
         if !(shielded_spends.is_empty() && shielded_outputs.is_empty()) {
             wtxs.push(WalletTx {
                 txid,
-                index: tx.index as usize,
+                index: tx_index as usize,
```
daira, line 516: There's an infallible [`From<u16> for usize`](https://doc.rust-lang.org/std/convert/trait.From.html#impl-From%3Cu16%3E-for-usize), so this need not use `as`.
```suggestion
                index: tx_index.into(),
```

daira (commented):
Post-hoc ACK with a few minor comments.

zcash_client_backend/src/data_api.rs
str4d: The entire struct needed documenting, so we can do this before the next release.

nuttycom: Created https://github.com/zcash/librustzcash/issues/885

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1294,6 +1324,165 @@ pub(crate) fn put_sent_output<P: consensus::Parameters>(
...
+    let (height, index, txid) = match locator {
+        Some(res) => res,
+        None => return Ok(None),
+    };
+
+    // Find or create a corresponding row in the `transactions` table. Usually a row will
+    // have been created during the same scan that the locator was added to the nullifier
+    // map, but it would not happen if the transaction in question spent the note with no
+    // change or explicit in-wallet recipient.
+    put_tx_meta(
```
nuttycom, line 1458: Given that this procedure can perform an insert, I don't think that `query_nullifier_map` is a good name; it's also not indicative of its purpose. Maybe something like `spending_tx_ref_for_nullifier`?

zcash_client_sqlite/src/wallet/init/migrations/nullifier_map.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        // We don't enforce any foreign key constraint to the blocks table, to allow
+        // loading the nullifier map separately from block scanning.
+        debug!("Creating tables for nullifier map");
+        transaction.execute_batch(
+            "CREATE TABLE tx_locator_map (
```
nuttycom, line 42: nit: I don't understand why these tables have the `_map` suffix; we don't use that convention elsewhere.

nuttycom (commented):
post-hoc utACK with a minor documentation suggestion.

Issue #877, opened 2023-07-20
`zcash_client_sqlite`: Improve performance of `shardtree` database migration


str4d:
Various PRs ending with #872 added a database migration that enables `shardtree` support. As part of the migration, existing note commitment and witness data is migrated into the `ShardTree` to pre-populate it and ensure that existing notes remain spendable via it.

In testing, we found that loading the existing witness data was fast enough to not be noticeable: the wallet only maintains witnesses for the last 100 blocks, for the notes in our wallet, so the amount of data is linear in the wallet size rather than the chain size.

However, loading in the Sapling commitment tree frontier data is _very_ slow: on a Ryzen 9 5950X it took around 10 minutes to load in 1.5 million frontiers, and in the iOS simulator on a MacBook M1 it took several minutes to load in 200k frontiers.

The code loading in the frontiers is here:

https://github.com/zcash/librustzcash/blob/8e582109cad9af14b461f76429f9f0437e860115/zcash_client_sqlite/src/wallet/init/migrations/shardtree_support.rs#L119-L172

The core problem is that the data migration is occurring inside a database migration, and database migrations are all-or-nothing: if any failure happens, the database is left in its unmigrated state. This means that if e.g. a mobile wallet gets killed 10 minutes into a 15-minute migration, the work done is lost and the migration restarts from the beginning on next app start.

At the same time, we _want_ the data to be migrated inside the database migration, to keep our previously-spendable notes spendable at the same height when the wallet updates to be `shardtree`-compatible.

There are two possible fixes we can consider here:
- Load the existing witnesses into the `ShardTree` first, and then use the positions of the marked notes to figure out which Sapling frontiers should be loaded into the `ShardTree` to fill out its data.
  - If this is possible to do, it would concretely reduce the number of frontiers we load in (minimizing the number of `insert_frontier_nodes` calls), to something that doesn't depend on the full chain size.
  - This might require the database to know the heights at which subtree boundaries occur; we might still need to do that pass inside the database migration (similar to the code in `zcashd`) before we can then identify which block heights have frontiers we need.
- Move the frontier-loading code out of the database migration and expose it via a new post-DB-migration API (or perhaps have it run as a last step inside the existing `init_data_db` API that currently drives the migration), and write to SQLite periodically during the frontier loading process.
  - This would mean that if a wallet app is killed during a call to `init_data_db`, then the database migration would be preserved, as would any partial loading of the frontiers.

PR #879, opened 2023-07-21
zcash_client_sqlite: Only import recent frontiers into `ShardTree`

str4d:
We only need to load frontiers into the ShardTree that are close enough to the wallet's known chain tip to fill `PRUNING_DEPTH` checkpoints, so that ShardTree's witness generation will be able to correctly handle anchor depths. Loading frontiers further back than this doesn't add any useful nodes to the ShardTree (as we don't support rollbacks beyond `PRUNING_DEPTH`, and we won't be finding notes in earlier blocks), and hurts performance (as frontier importing has a significant Merkle tree hashing cost).

Closes zcash/librustzcash#877.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/shardtree_support.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/shardtree_support.rs
@@ -103,6 +103,8 @@ impl RusqliteMigration for Migration {
             );",
         )?;
 
+        let block_height_extrema = block_height_extrema(transaction)?;
+
         let shard_store =
             SqliteShardStore::<_, sapling::Node, SAPLING_SHARD_HEIGHT>::from_connection(
                 transaction,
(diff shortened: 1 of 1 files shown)


ebfull (approved):

zcash_client_sqlite/src/wallet/init/migrations/shardtree_support.rs
```diff
@@ -149,25 +151,40 @@ impl RusqliteMigration for Migration {
...
+                // hashing cost).
+                if let Some((nonempty_frontier, (_, latest_height))) = block_end_tree
+                    .to_frontier()
+                    .value()
+                    .zip(block_height_extrema)
+                {
+                    let block_height = BlockHeight::from(block_height);
+                    if block_height + PRUNING_DEPTH >= latest_height {
+                        trace!(
+                            height = u32::from(block_height),
```
daira, line 170: We already had `block_height` as a `u32` above before it was shadowed. So we can avoid this reconversion by choosing a different name (e.g. `block_height_u32` for the `u32` version.

str4d: Eh, I'd prefer that we use `BlockHeight` in more places. I added the shadowing here because I wanted it for the addition and comparison, but we need to unwrap for `tracing` macros. The unwrapping cost is literally zero, the compiler will eliminate it entirely as `BlockHeight` is a newtype wrapper around `u32`.

daira (commented):
Post-hoc utACK

nuttycom (commented):
post-hoc utACK.

I had originally thought that these insertions might be necessary in order to ensure that wallets that connect to not-yet-upgraded lightwalletd servers had all the necessary roots to perform spends, but on reflection importing the incremental witnesses is sufficient for this.

PR #883, opened 2023-07-25
zcash_client_sqlite: Build subtrees from new commitments in a threadpool

str4d:
The new `multicore` feature flag can be used to disable this behaviour.

Depends on https://github.com/zcash/incrementalmerkletree/pull/88.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -33,6 +33,10 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 
 use either::Either;
+use maybe_rayon::{
+    prelude::{IndexedParallelIterator, ParallelIterator},
+    slice::ParallelSliceMut,
+};
 use rusqlite::{self, Connection};
 use secrecy::{ExposeSecret, SecretVec};
 use std::{borrow::Borrow, collections::HashMap, convert::AsRef, fmt, io, ops::Range, path::Path};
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/lib.rs
```diff
@@ -474,7 +478,7 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                 }));
 
                 last_scanned_height = Some(block.height());
-                sapling_commitments.extend(block.into_sapling_commitments().into_iter());
+                sapling_commitments.extend(block.into_sapling_commitments().into_iter().map(Some));
```
nuttycom, line 481: Why do we map this into `Some` and then use `n.take().expect(...)`? Is this just a common idiom that makes it possible to take ownership of parts of an iterator (because `.take()` replaces the iterator's owned chunk with `None`?

nuttycom (commented):
utACK; I'm surprised by how little code it takes to build the subtrees in parallel.

zcash_client_sqlite/src/lib.rs
str4d: Yep. `rayon` only provides mutable references to sub-slices, not owned references (which makes sense as you'd otherwise need allocations to split the slice). I don't know if this pattern is common per se for this kind of code, but it's used frequently to implement state machines (e.g. manual `Future` impls), and this is kinda a single-step state machine.

PR #886, opened 2023-08-01
Fix a number of bugs related to the creation of change memos and storage of empty memos.

nuttycom:


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -545,30 +545,29 @@ where
...
                     *ua.sapling().expect("TODO: Add Orchard support to builder"),
                     payment.amount,
-                    payment.memo.clone().unwrap_or_else(MemoBytes::empty),
+                    memo.clone(),
                 )?;
                 sapling_output_meta.push((
                     Recipient::Unified(ua.clone(), PoolType::Shielded(ShieldedProtocol::Sapling)),
                     payment.amount,
-                    payment.memo.clone(),
+                    Some(memo),
```
str4d, line 561: Bug: we are now mapping "memo is unset" to "empty memo", and then setting that on a metadata field that has a "memo is unset" state. We should either not do this (preserving the existing behaviour where we internally map the unset state to empty for the builder, but expose the unset state externally), or we should make the metadata field non-optional (as it no longer indicates whether the user actually set a memo or not).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -545,30 +545,29 @@ where
...
-                    Recipient::Sapling(*addr),
-                    payment.amount,
-                    payment.memo.clone(),
-                ));
+                let memo = payment
+                    .memo
+                    .as_ref()
+                    .map_or_else(MemoBytes::empty, |m| m.clone());
+                builder.add_sapling_output(external_ovk, *addr, payment.amount, memo.clone())?;
+                sapling_output_meta.push((Recipient::Sapling(*addr), payment.amount, Some(memo)));
```
str4d, line 570: Bug: ditto.

str4d (changes requested):

zcash_client_backend/src/data_api/wallet.rs
nuttycom: This is not a bug; it's an intended change in behavior. If the payment request does not specify a memo, the transaction that is created will have the empty memo, and that is what the recipient will receive. At the payment request level, we can't make the memo required because a `Payment` may be to a transparent address. The goal here is for the notes table(s) to store the same information as actually appears in the transaction.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -440,6 +445,164 @@ pub(crate) mod tests {
...
+        let sent_note_ids = stmt_sent_notes
+            .query(rusqlite::params![sent_tx_id])
+            .unwrap()
+            .mapped(|row| row.get::<_, i64>(0).map(NoteId::SentNoteId))
+            .collect::<Result<Vec<_>, _>>()
+            .unwrap();
+
+        assert_eq!(sent_note_ids.len(), 2);
+
+        // The sent memo should be the empty memo for both the sent output and change
```
nuttycom, line 580:
```suggestion
        // The sent memo should be the empty memo for the sent output, and the
        // change output's memo should be as specified.
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -129,8 +129,14 @@ pub(crate) fn pool_code(pool_type: PoolType) -> i64 {
...
-    memo.filter(|m| *m != &MemoBytes::empty())
-        .map(|m| m.as_slice())
+    memo.map(|m| {
+        if m == &MemoBytes::empty() {
+            // we store the empty memo as a single 0xf6 byte
+            &[0xf6]
+        } else {
+            m.as_slice()
+        }
+    })
```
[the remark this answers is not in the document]
nuttycom: The memo may be `None` in the `sent_notes` table as a consequence of wallet recovery; the note may have been detected but the rest of the transaction data will not have been populated. The intent of this change overall is to make `null` reflect explicitly and only this state; in every case where we've observed the memo (because we created the transaction, or decrypted it) it should be identical to the memo data that's encoded into the transaction itself.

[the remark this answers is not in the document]
nuttycom: (Remember that the `sent_notes` table is badly named at this point; it also contains metadata for transparent outputs.)

zcash_client_backend/src/data_api/wallet.rs
str4d: I didn't say this was a bug in the table, but a bug in the metadata. The `Some(memo)` that I left this comment on, ends up being passed to this constructor further down:
https://github.com/zcash/librustzcash/blob/110a9b31d3aebc832bb82537be05047059ee0529/zcash_client_backend/src/data_api/wallet.rs#L635
and is then exposed via this method:
https://github.com/zcash/librustzcash/blob/eade2acab9dcef282ca3298bb14fdcef787bb6e0/zcash_client_backend/src/data_api.rs#L454-L457

Note "if any". That is where the bug lies: the change in this PR is explicitly mapping away the `None` case that would constitute "not any", but the metadata API is still representing to the user that it can and should be expected to occur. The previous code preserved the "if any" case from the part of the code being changed by this PR, to allow the caller to distinguish "implicit empty memo" from "explicit empty memo"; this PR prevents them from doing so. If that is an intentional change, then per my original comment, we should make `SentTransactionOutput::memo` return `&MemoBytes`.

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
str4d: If that is the desired change, then this should be done with a database migration. The current `null` value is aliasing "implicit empty memo" (which previously was the only state this column stored) with "unknown memo" (present only since your change to store received change notes in the `sent_notes` table). To properly fix this, we need a database migration that checks whether the `raw` field is non-null (indicating we do have the transaction data populated), and decrypts the outputs to re-fill the memo column.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: We can't make `SentTransactionOutput::memo` return `&MemoBytes` because `SentTransactionOutput` is also used to represent transparent outputs, which cannot have memos.

str4d: Aha, okay. Then this is a documentation bug, and we should fix the documentation of that method to say that `None` now only means "a transparent output".

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
pacu: Getting rid of these assumptions would be much appreciated by our future selves. 

Maybe it is too much of a lift, or out of scope for this PR?

It feels much safer to actually keep the things an they are with the current conventions ( extra docs will help for sure ) and keep the PR more concise.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -588,7 +587,7 @@ where
                     internal_ovk(),
                     dfvk.change_address().1,
                     *amount,
-                    MemoBytes::empty(),
+                    change_memo.clone().unwrap_or_else(MemoBytes::empty),
```
str4d, line 590: The change to the definition of `SentTransactionOutput::memo` means that the `sapling_output_meta` tuple below this still doesn't match this fixed line. It needs to also be set to `Some(this_value)` so that it isn't using `None` to mean "the user didn't set a change memo so there's an implicit one", instead of the updated definition of "not applicable to this output pool".

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
str4d: `memo_repr` is undocumented, so I went digging to determine its in-practice definition. Prior to this PR, the only two places we call `memo_repr` are:
- `put_received_note` using `ReceivedSaplingOutput::memo`, which is undocumented but in practice for our impls `None` means "we don't have the memo data".
- `put_sent_output` using a provided `Option<&MemoBytes>`. This is only called inside `store_decrypted_tx`: once with `Some(&memo)` for decrypted Sapling outputs, and once with `None` for transparent outputs.

So just from this we already have two definitions of `None` as the output of `memo_repr`, but they happen to not be aliased as they are used in different tables. The `memo` column in the `sapling_received_notes` table uses `null` in this context to mean "unknown memo", while the `memo` column in the `sent_notes` table uses `null` in this context to mean "not applicable to this output pool".

This PR alters `insert_sent_output` to use `memo_repr` as well, replacing code that is equivalent to the previous implementation of `memo_repr`. The field there is `SentTransactionOutput::memo`, which per the other comment thread uses `None` to mean either "transparent output" or "no change memo specified" as of this PR, but previously also used it to mean "unknown memo". So AFAICT if the "no change memo specified" case is fixed per my new comment below, then the `sent_notes` table's `memo` column has a single definition of `null` after this PR.

str4d (approved):
utACK once the change memo bug above is fixed. The database migration and the documentation fixes are release-blocking but don't need to block this PR; please open issues for them if not addressed here.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Ah, yep, I missed that one.

nuttycom:
> utACK once the change memo bug above is fixed. The database migration and the documentation fixes are release-blocking but don't need to block this PR; please open issues for them if not addressed here.

Changes addressed in https://github.com/zcash/librustzcash/pull/886/commits/a5a8562ddde1a5d7b7a2dbe9497fb7adfabe65c4

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/886/commits/90412ebdc230cbc113945a42ec86124116d09dca

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
```diff
@@ -0,0 +1,218 @@
...
+                "UPDATE sent_notes 
+                SET memo = :memo 
+                WHERE tx = :id_tx
+                AND output_index = :output_index",
+            )?;
+
+            for d_out in decrypted_outputs {
+                if sapling_outputs
+                    .iter()
+                    .any(|(idx, _, _)| &(u32::try_from(d_out.index).unwrap()) == idx)
```
str4d, line 105: When could this be false? We looked up this transaction via the `sent_notes` table, so if any change had been made to the `transaction` row that caused it to conflict with the `sent_notes` table, that seems like it would be a bug that we should raise a corruption error for instead of ignoring.

str4d (changes requested):
Reviewed as of cb0d16cfa97c2532d97d58dfc890213f96cddf45

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
```diff
@@ -0,0 +1,218 @@
...
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        let mut stmt_raw_tx = transaction.prepare(
+            "SELECT DISTINCT sent_notes.tx, sent_notes.output_index, accounts.account, accounts.ufvk
+             FROM sent_notes 
+             JOIN accounts ON sent_notes.from_account = accounts.account
+             JOIN transactions ON transactions.id_tx = sent_notes.tx
+             WHERE transactions.raw IS NOT NULL"
```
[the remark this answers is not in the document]
nuttycom: Hm, yeah, that would be fine, I think. I figured doing it this way was more comprehensive (and could potentially fix any memo inconsistencies that were introduced by bugs that might have existed in the past) but if migration time is a concern then we can narrow it down.

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
nuttycom: I... don't remember what I was thinking here. The update statement also skips anything where the decrypted note doesn't have an entry in the `sent_notes` table, so this is entirely unnecessary.

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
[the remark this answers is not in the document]
str4d: Let's leave it as-is in that case; we can always reduce the scope of this migration if we see performance problems in testing.

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
```diff
@@ -0,0 +1,215 @@
...
+                WalletMigrationError::CorruptedData(format!(
+                    "Could not decode unified full viewing key for account {}: {:?}",
+                    account, e
+                ))
+            })?;
+
+            tx_sent_notes
+                .entry(id_tx)
+                .and_modify(|v| v.push((output_index, AccountId::from(account), ufvk.clone())))
+                .or_insert_with(|| vec![(output_index, AccountId::from(account), ufvk)]);
```
str4d, line 75: Huh, I thought I left a comment here in my previous review, but I must have forgotten to click "Add". The comment was that we can avoid the `ufvk.clone()` here; merging that with the `output_index` removal we get:
```suggestion
            tx_sent_notes
                .entry(id_tx)
                .or_default()
                .insert(AccountId::from(account), ufvk);
```

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        for (id_tx, sapling_outputs) in tx_sent_notes {
```
str4d, line 78:
```suggestion
        for (id_tx, ufvks) in tx_sent_notes {
```

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
```diff
@@ -0,0 +1,215 @@
...
+                        "An error was encountered decoding transaction data: {:?}",
+                        other
+                    )),
+                })?;
+
+            let ufvks: HashMap<AccountId, UnifiedFullViewingKey> = sapling_outputs
+                .iter()
+                .map(|(_, account, ufvk)| (*account, ufvk.clone()))
+                .collect();
+
```
str4d, line 95:
```suggestion
```

str4d (approved):
utACK 522f1e4da0d1e9a63777881ec534f56834ea6f4f

PR #887, opened 2023-08-01
Pin `time` dependency to avoid forced MSRV bump.

nuttycom:


daira (approved):
utACK

str4d (changes requested):
NACK. Pinning non-dev depending is the wrong way to solve this, as it prevents downstreams that don't pin to that MSRV from updating (e.g. our own moving SDKs).

Can we get CI working by pinning like this as a dev-dependency?

str4d:
This is going to be a somewhat-ongoing issue, because [`time` has adopted a "somewhat lazy `N - 2`" MSRV policy](https://github.com/time-rs/time/discussions/535): they reserve the right to bump at any time to `N - 2` if there's a feature they need, and `N - six_months` otherwise. Our own policy is around `N - 4`.

str4d:
I've verified locally that adding this line to `zcash_client_backend/Cargo.toml` is sufficient to enable us to compile:

```rust
[dev-dependencies]
...
time = ">=0.3.22, <0.3.24" # time 0.3.24 has MSRV 1.67
```

str4d (approved):
ACK

PR #888, opened 2023-08-02
Restrict use of backend-specific note identifiers.

nuttycom:
This builds atop #886

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -1,7 +1,7 @@
-use std::fmt::Debug;
 use std::{convert::Infallible, num::NonZeroU32};
 
 use shardtree::{error::ShardTreeError, store::ShardStore, ShardTree};
+use zcash_primitives::transaction::TxId;
 use zcash_primitives::{
     consensus::{self, BlockHeight, NetworkUpgrade},
     memo::MemoBytes,
@@ -203,7 +203,7 @@ pub fn create_spend_to_address<DbT, ParamsT>(
     ovk_policy: OvkPolicy,
     min_confirmations: NonZeroU32,
 ) -> Result<
-    DbT::TxRef,
+    TxId,
     Error<
         <DbT as WalletRead>::Error,
         <DbT as WalletCommitmentTrees>::Error,
@@ -306,7 +306,7 @@ pub fn spend<DbT, ParamsT, InputsT>(
     ovk_policy: OvkPolicy,
     min_confirmations: NonZeroU32,
 ) -> Result<
-    DbT::TxRef,
+    TxId,
     Error<
         <DbT as WalletRead>::Error,
         <DbT as WalletCommitmentTrees>::Error,
@@ -317,7 +317,6 @@ pub fn spend<DbT, ParamsT, InputsT>(
 >
 where
     DbT: WalletWrite + WalletCommitmentTrees,
-    DbT::TxRef: Copy + Debug,
     DbT::NoteRef: Copy + Eq + Ord,
     ParamsT: consensus::Parameters + Clone,
     InputsT: InputSelector<DataSource = DbT>,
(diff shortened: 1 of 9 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -441,7 +440,7 @@ pub fn create_proposed_transaction<DbT, ParamsT, InputsErrT, FeeRuleT>(
     min_confirmations: NonZeroU32,
     change_memo: Option<MemoBytes>,
 ) -> Result<
-    DbT::TxRef,
+    TxId,
```
pacu, line 443: [Non - Blocking] Although small, this is a great improvement!

I'm mostly sure it will break the ffi layer so it will be good to track this hiccup in the ffi repository on iOS and the android sdk r

pacu (commented):
UTACK

Left non blocking comment on possible breaking API

zcash_client_sqlite/src/lib.rs
```diff
@@ -230,17 +216,17 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
-            NoteId::ReceivedNoteId(id_note) => {
-                wallet::get_received_memo(self.conn.borrow(), id_note)
-            }
+    fn get_memo(&self, note_id: NoteId) -> Result<Option<Memo>, Self::Error> {
+        let sent_memo = wallet::get_sent_memo(self.conn.borrow(), note_id)?;
+        if sent_memo.is_some() {
+            Ok(sent_memo)
+        } else {
+            wallet::get_received_memo(self.conn.borrow(), note_id)
         }
```
str4d, line 225: This method is now concretely less efficient than before. I don't know by how much, but it is now `O(num_received + num_sent + 2*num_internal)`, because we have two separate table joins against `transactions` (instead of a direct query of an indexed column), and for received notes we always query the sent notes table first. We'll know more once we get this into the mobile SDKs, so this is non-blocking, but we should track its effect on the history-rendering app views.

More non-blocking thoughts: Could we gain back or amortize some of the performance loss by combining the new queries into a single query against all three tables at once? AFAICT this is the only place that `wallet::get_sent_memo` and `wallet::get_received_memo` are used (they were removed from the public API in 0.7.0), so we can just have a single backing function. This does become slightly more annoying given the `sapling_received_notes` vs `sent_notes` split, but we can put off dealing with that to the Orchard work.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -507,16 +514,16 @@ pub(crate) fn get_received_memo(
 pub(crate) fn get_transaction<P: Parameters>(
     conn: &rusqlite::Connection,
     params: &P,
-    id_tx: i64,
+    txid: TxId,
```
str4d, line 517: The documentation for this function needs to be updated.

str4d (changes requested):
Reviewed ef0b27c657514e26b2f2d44303ed833cbe4a6ad0.

str4d:
Needs a rebase to fix merge conflicts after #889 was merged.

str4d (approved):
utACK f602ec125d4de4b32e79ba583113ec57ff01c56d

zcash_client_sqlite/src/lib.rs
```diff
@@ -98,22 +98,9 @@ pub(crate) const VERIFY_LOOKAHEAD: u32 = 10;
...
-/// A newtype wrapper for sqlite primary key values for the notes
-/// table.
+/// A newtype wrapper for received note identifiers.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub enum NoteId {
-    SentNoteId(i64),
-    ReceivedNoteId(i64),
-}
-
-impl fmt::Display for NoteId {
```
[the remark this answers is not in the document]
str4d: Oops, I missed that the `NoteId` generic parameter on `Error` was actually just renamed to `NoteRef`, so we did still need a `Display` impl. Fixed in #890.

PR #891, opened 2023-08-08
Add protobuf representation for transaction proposals

nuttycom:
In order to make complete information about proposed transactions readily available to wallets, it's necessary to provide a mechanism to pass a `TransactionProposal` across the FFI. This PR adds a Protobuf representation for proposed transactions and implements conversions back and forth between the protobuf representation and the native type.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -15,6 +15,7 @@ and this library adheres to Rust's notion of
   - `ScannedBlock::sapling_tree_size`.
   - `ScannedBlock::orchard_tree_size`.
   - `wallet::propose_standard_transfer_to_address`
+  - `wallet::input_selection::Proposal::from_parts`
   - `wallet::input_selection::SaplingInputs`
   - `wallet::input_selection::ShieldingSelector` has been
     factored out from the `InputSelector` trait to separate out transparent
@@ -23,6 +24,22 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::wallet`:
   - `ReceivedSaplingNote::from_parts`
   - `ReceivedSaplingNote::{txid, output_index, diversifier, rseed, note_commitment_tree_position}`
+- `zcash_client_backend::zip321::TransactionRequest::total`
+- `zcash_client_backend::proto::`
+  - `PROPOSAL_SER_V1`
+  - `ProposalError`
+  - `proposal::Proposal::{from_standard_proposal, try_into_standard_proposal}`
+  - `proposal::ProposedInput::parse_txid`
+- `impl Clone for zcash_client_backend::{
+     zip321::{Payment, TransactionRequest, Zip321Error, parse::Param, parse::IndexedParam},
+     wallet::{ReceivedSaplingNote, WalletTransparentOutput},
+     wallet::input_selection::{Proposal, SaplingInputs},
+   }`
+- `impl {PartialEq, Eq} for zcash_client_backend::{
+     zip321::{Zip321Error, parse::Param, parse::IndexedParam},
+     wallet::{ReceivedSaplingNote, WalletTransparentOutput},
+     wallet::input_selection::{Proposal, SaplingInputs},
+   }`
 
 ### Changed
 - `zcash_client_backend::data_api`:
@@ -42,11 +59,11 @@ and this library adheres to Rust's notion of
       backend-specific note identifier. The related `NoteRef` type parameter has
       been removed from `error::Error`.
     - A new variant `UnsupportedPoolType` has been added.
-  - `wallet::shield_transparent_funds` no longer
-    takes a `memo` argument; instead, memos to be associated with the shielded
-    outputs should be specified in the construction of the value of the
-    `input_selector` argument, which is used to construct the proposed shielded
-    values as internal "change" outputs.
+  - `wallet::shield_transparent_funds` no longer takes a `memo` argument;
+    instead, memos to be associated with the shielded outputs should be
+    specified in the construction of the value of the `input_selector`
+    argument, which is used to construct the proposed shielded values as
+    internal "change" outputs.
   - `wallet::create_proposed_transaction` no longer takes a
     `change_memo` argument; instead, change memos are represented in the
     individual values of the `proposed_change` field of the `Proposal`'s
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,8 +28,7 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::proto::`
   - `PROPOSAL_SER_V1`
   - `ProposalError`
-  - `proposal::Proposal::{from_standard_proposal, try_into_standard_proposal}`
-  - `proposal::ProposedInput::parse_txid`
+  - `proposal` module, for parsing and serializing transaction proposals.
 - `impl Clone for zcash_client_backend::{
      zip321::{Payment, TransactionRequest, Zip321Error, parse::Param, parse::IndexedParam},
      wallet::{ReceivedSaplingNote, WalletTransparentOutput},
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -25,6 +25,7 @@ and this library adheres to Rust's notion of
   - `ReceivedSaplingNote::from_parts`
   - `ReceivedSaplingNote::{txid, output_index, diversifier, rseed, note_commitment_tree_position}`
 - `zcash_client_backend::zip321::TransactionRequest::total`
+- `zcash_client_backend::zip321::parse::Param::name`
 - `zcash_client_backend::proto::`
   - `PROPOSAL_SER_V1`
   - `ProposalError`
--- a/zcash_client_backend/proto/proposal.proto
+++ b/zcash_client_backend/proto/proposal.proto
@@ -1,0 +1,86 @@
+// Copyright (c) 2023 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+syntax = "proto3";
+package cash.z.wallet.sdk.ffi;
+
+// A data structure that describes the inputs to be consumed and outputs to
+// be produced in a proposed transaction.
+message Proposal {
+    uint32 protoVersion = 1;
+    // ZIP 321 serialized transaction request
+    string transactionRequest = 2;
+    // The transparent UTXOs to use as inputs to the transaction.
+    repeated ProposedInput transparentInputs = 3;
+    // The Sapling input notes and anchor height to be used in creating the transaction.
+    SaplingInputs saplingInputs = 4;
+    // The total value, fee amount, and change outputs of the proposed
+    // transaction
+    TransactionBalance balance = 5;
+    // The fee rule used in constructing this proposal
+    FeeRule feeRule = 6;
+    // The target height for which the proposal was constructed
+    //
+    // The chain must contain at least this many blocks in order for the proposal to
+    // be executed.
+    uint32 minTargetHeight = 7;
+    // A flag indicating whether the proposal is for a shielding transaction,
+    // used for determining which OVK to select for wallet-internal outputs.
+    bool isShielding = 8;
+}
+
+message SaplingInputs {
+    // Returns the Sapling anchor height to be used in creating the transaction.
+    uint32 anchorHeight = 1;
+    // The unique identifier and amount for each proposed Sapling input
+    repeated ProposedInput inputs = 2;
+}
+
+// The unique identifier and amount for each proposed input.
+message ProposedInput {
+    bytes txid = 1;
+    uint32 index = 2;
+    uint64 value = 3;
+}
+
+// The fee rule used in constructing a Proposal
+enum FeeRule {
+    // Protobuf requires that enums have a zero descriminant as the default
+    // value. However, we need to require that a known fee rule is selected,
+    // and we do not want to fall back to any default, so sending the
+    // FeeRuleNotSpecified value will be treated as an error.
+    FeeRuleNotSpecified = 0;
+    // 10000 ZAT
+    PreZip313 = 1;
+    // 1000 ZAT
+    Zip313 = 2;
+    // MAX(10000, 5000 * logical_actions) ZAT
+    Zip317 = 3;
+}
+
+// The proposed change outputs and fee amount.
+message TransactionBalance {
+    repeated ChangeValue proposedChange = 1;
+    uint64 feeRequired = 2;
+}
+
+// An enumeration of change value types.
+message ChangeValue {
+    oneof value {
+        SaplingChange saplingValue = 1;
+    }
+}
+
+// An object wrapper for memo bytes, to facilitate representing the
+// `change_memo == None` case.
+message MemoBytes {
+    bytes value = 1;
+}
+
+// The amount and memo for a proposed Sapling change output.
+message SaplingChange {
+    uint64 amount = 1;
+    MemoBytes memo = 2;
+}
+
--- a/zcash_client_backend/proto/proposal.proto
+++ b/zcash_client_backend/proto/proposal.proto
@@ -31,7 +31,7 @@ message Proposal {
 }
 
 message SaplingInputs {
-    // Returns the Sapling anchor height to be used in creating the transaction.
+    // The Sapling anchor height to be used in creating the transaction
     uint32 anchorHeight = 1;
     // The unique identifier and amount for each proposed Sapling input
     repeated ProposedInput inputs = 2;
@@ -46,7 +46,7 @@ message ProposedInput {
 
 // The fee rule used in constructing a Proposal
 enum FeeRule {
-    // Protobuf requires that enums have a zero descriminant as the default
+    // Protobuf requires that enums have a zero discriminant as the default
     // value. However, we need to require that a known fee rule is selected,
     // and we do not want to fall back to any default, so sending the
     // FeeRuleNotSpecified value will be treated as an error.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -17,7 +17,10 @@ use zcash_primitives::{
     memo::{Memo, MemoBytes},
     sapling::{self, Node, NOTE_COMMITMENT_TREE_DEPTH},
     transaction::{
-        components::amount::{Amount, NonNegativeAmount},
+        components::{
+            amount::{Amount, NonNegativeAmount},
+            OutPoint,
+        },
         Transaction, TxId,
     },
     zip32::AccountId,
@@ -30,9 +33,6 @@ use crate::{
     proto::service::TreeState,
     wallet::{ReceivedSaplingNote, WalletTransparentOutput, WalletTx},
 };
-
-#[cfg(feature = "transparent-inputs")]
-use zcash_primitives::transaction::components::OutPoint;
 
 use self::chain::CommitmentTreeRoot;
 use self::scanning::ScanRange;
@@ -222,6 +222,14 @@ pub trait SaplingInputSource {
     /// or a UUID.
     type NoteRef: Copy + Debug + Eq + Ord;
 
+    /// Returns a received Sapling note, or Ok(None) if the note is not known to belong to the
+    /// wallet or if the note is not spendable.
+    fn get_spendable_sapling_note(
+        &self,
+        txid: &TxId,
+        index: u32,
+    ) -> Result<Option<ReceivedSaplingNote<Self::NoteRef>>, Self::Error>;
+
     /// Returns a list of spendable Sapling notes sufficient to cover the specified target value,
     /// if possible.
     fn select_spendable_sapling_notes(
@@ -235,10 +243,16 @@ pub trait SaplingInputSource {
 
 /// A trait representing the capability to query a data store for unspent transparent UTXOs
 /// belonging to a wallet.
-#[cfg(feature = "transparent-inputs")]
 pub trait TransparentInputSource {
     /// The type of errors produced by a wallet backend.
     type Error;
+
+    /// Returns a received transparent UTXO, or Ok(None) if the UTXO is not known to belong to the
+    /// wallet or is not spendable.
+    fn get_unspent_transparent_output(
+        &self,
+        outpoint: &OutPoint,
+    ) -> Result<Option<WalletTransparentOutput>, Self::Error>;
 
     /// Returns a list of unspent transparent UTXOs that appear in the chain at heights up to and
     /// including `max_height`.
@@ -1008,6 +1022,14 @@ pub mod testing {
         type Error = ();
         type NoteRef = u32;
 
+        fn get_spendable_sapling_note(
+            &self,
+            _txid: &TxId,
+            _index: u32,
+        ) -> Result<Option<ReceivedSaplingNote<Self::NoteRef>>, Self::Error> {
+            Ok(None)
+        }
+
         fn select_spendable_sapling_notes(
             &self,
             _account: AccountId,
@@ -1022,6 +1044,13 @@ pub mod testing {
     #[cfg(feature = "transparent-inputs")]
     impl TransparentInputSource for MockWalletDb {
         type Error = ();
+
+        fn get_unspent_transparent_output(
+            &self,
+            _outpoint: &OutPoint,
+        ) -> Result<Option<WalletTransparentOutput>, Self::Error> {
+            Ok(None)
+        }
 
         fn get_unspent_transparent_outputs(
             &self,
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -222,8 +222,11 @@ pub trait SaplingInputSource {
     /// or a UUID.
     type NoteRef: Copy + Debug + Eq + Ord;
 
-    /// Returns a received Sapling note, or Ok(None) if the note is not known to belong to the
-    /// wallet or if the note is not spendable.
+    /// Fetches a spendable Sapling note by indexing into the specified transaction's
+    /// [`sapling::Bundle::shielded_outputs`].
+    ///
+    /// Returns `Ok(None)` if the note is not known to belong to the wallet or if the note
+    /// is not spendable.
     fn get_spendable_sapling_note(
         &self,
         txid: &TxId,
@@ -247,8 +250,10 @@ pub trait TransparentInputSource {
     /// The type of errors produced by a wallet backend.
     type Error;
 
-    /// Returns a received transparent UTXO, or Ok(None) if the UTXO is not known to belong to the
-    /// wallet or is not spendable.
+    /// Fetches a spendable transparent output.
+    ///
+    /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or is not
+    /// spendable.
     fn get_unspent_transparent_output(
         &self,
         outpoint: &OutPoint,
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -90,10 +90,25 @@ pub struct Proposal<FeeRuleT, NoteRef> {
     is_shielding: bool,
 }
 
+/// Errors that
+#[derive(Debug, Clone)]
+pub enum ProposalError {
+    RequestInvalid,
+    Overflow,
+    BalanceError {
+        input_total: NonNegativeAmount,
+        output_total: NonNegativeAmount,
+    },
+    ShieldingFlagInvalid,
+}
+
 impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
-    /// Constructs a [`Proposal`] from its constituent parts.
+    /// Constructs a validated [`Proposal`] from its constituent parts.
+    ///
+    /// This operation validaes the proposal for balance consistency and agreement between
+    /// the `is_shielding` flag and the structure of the proposal.
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn from_parts(
+    pub fn from_parts(
         transaction_request: TransactionRequest,
         transparent_inputs: Vec<WalletTransparentOutput>,
         sapling_inputs: Option<SaplingInputs<NoteRef>>,
@@ -101,19 +116,35 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
         fee_rule: FeeRuleT,
         min_target_height: BlockHeight,
         is_shielding: bool,
-    ) -> Result<Self, ()> {
-        let transparent_total = transparent_inputs
+    ) -> Result<Self, ProposalError> {
+        let transparent_input_total = transparent_inputs
             .iter()
             .map(|out| out.txout().value)
-            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| (acc? + a).ok_or(()))?;
-        let sapling_total = sapling_inputs
+            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| {
+                (acc? + a).ok_or(ProposalError::Overflow)
+            })?;
+        let sapling_input_total = sapling_inputs
             .iter()
             .flat_map(|s_in| s_in.notes().iter())
             .map(|out| out.value())
-            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| (acc? + a).ok_or(()))?;
-        let input_total = (transparent_total + sapling_total).ok_or(())?;
+            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| {
+                (acc? + a).ok_or(ProposalError::Overflow)
+            })?;
+        let input_total =
+            (transparent_input_total + sapling_input_total).ok_or(ProposalError::Overflow)?;
 
-        let output_total = (transaction_request.total()? + balance.total()).ok_or(())?;
+        let request_total = transaction_request
+            .total()
+            .map_err(|_| ProposalError::RequestInvalid)?;
+        let output_total = (request_total + balance.total()).ok_or(ProposalError::Overflow)?;
+
+        if is_shielding
+            && (transparent_input_total == NonNegativeAmount::ZERO
+                || sapling_input_total > NonNegativeAmount::ZERO
+                || request_total > NonNegativeAmount::ZERO)
+        {
+            return Err(ProposalError::ShieldingFlagInvalid);
+        }
 
         if input_total == output_total {
             Ok(Self {
@@ -126,7 +157,10 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
                 is_shielding,
             })
         } else {
-            Err(())
+            Err(ProposalError::BalanceError {
+                input_total,
+                output_total,
+            })
         }
     }
 
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -79,6 +79,7 @@ impl<DE: fmt::Display, SE: fmt::Display> fmt::Display for InputSelectorError<DE,
 }
 
 /// The inputs to be consumed and outputs to be produced in a proposed transaction.
+#[derive(Clone, PartialEq, Eq)]
 pub struct Proposal<FeeRuleT, NoteRef> {
     transaction_request: TransactionRequest,
     transparent_inputs: Vec<WalletTransparentOutput>,
@@ -90,6 +91,45 @@ pub struct Proposal<FeeRuleT, NoteRef> {
 }
 
 impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
+    /// Constructs a [`Proposal`] from its constituent parts.
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn from_parts(
+        transaction_request: TransactionRequest,
+        transparent_inputs: Vec<WalletTransparentOutput>,
+        sapling_inputs: Option<SaplingInputs<NoteRef>>,
+        balance: TransactionBalance,
+        fee_rule: FeeRuleT,
+        min_target_height: BlockHeight,
+        is_shielding: bool,
+    ) -> Result<Self, ()> {
+        let transparent_total = transparent_inputs
+            .iter()
+            .map(|out| out.txout().value)
+            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| (acc? + a).ok_or(()))?;
+        let sapling_total = sapling_inputs
+            .iter()
+            .flat_map(|s_in| s_in.notes().iter())
+            .map(|out| out.value())
+            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| (acc? + a).ok_or(()))?;
+        let input_total = (transparent_total + sapling_total).ok_or(())?;
+
+        let output_total = (transaction_request.total()? + balance.total()).ok_or(())?;
+
+        if input_total == output_total {
+            Ok(Self {
+                transaction_request,
+                transparent_inputs,
+                sapling_inputs,
+                balance,
+                fee_rule,
+                min_target_height,
+                is_shielding,
+            })
+        } else {
+            Err(())
+        }
+    }
+
     /// Returns the transaction request that describes the payments to be made.
     pub fn transaction_request(&self) -> &TransactionRequest {
         &self.transaction_request
@@ -147,12 +187,24 @@ impl<FeeRuleT, NoteRef> Debug for Proposal<FeeRuleT, NoteRef> {
 }
 
 /// The Sapling inputs to a proposed transaction.
+#[derive(Clone, PartialEq, Eq)]
 pub struct SaplingInputs<NoteRef> {
     anchor_height: BlockHeight,
     notes: NonEmpty<ReceivedSaplingNote<NoteRef>>,
 }
 
 impl<NoteRef> SaplingInputs<NoteRef> {
+    /// Constructs a [`SaplingInputs`] from its constituent parts.
+    pub fn from_parts(
+        anchor_height: BlockHeight,
+        notes: NonEmpty<ReceivedSaplingNote<NoteRef>>,
+    ) -> Self {
+        Self {
+            anchor_height,
+            notes,
+        }
+    }
+
     /// Returns the anchor height for Sapling inputs that should be used when constructing the
     /// proposed transaction.
     pub fn anchor_height(&self) -> BlockHeight {
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -145,7 +145,7 @@ impl std::error::Error for ProposalError {}
 impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
     /// Constructs a validated [`Proposal`] from its constituent parts.
     ///
-    /// This operation validaes the proposal for balance consistency and agreement between
+    /// This operation validates the proposal for balance consistency and agreement between
     /// the `is_shielding` flag and the structure of the proposal.
     #[allow(clippy::too_many_arguments)]
     pub fn from_parts(
@@ -167,9 +167,8 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
             .iter()
             .flat_map(|s_in| s_in.notes().iter())
             .map(|out| out.value())
-            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| {
-                (acc? + a).ok_or(ProposalError::Overflow)
-            })?;
+            .fold(Some(NonNegativeAmount::ZERO), |acc, a| (acc? + a))
+            .ok_or(ProposalError::Overflow)?;
         let input_total =
             (transparent_input_total + sapling_input_total).ok_or(ProposalError::Overflow)?;
 
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -1,7 +1,7 @@
 //! Types related to the process of selecting inputs to be spent given a transaction request.
 
 use core::marker::PhantomData;
-use std::fmt::{self, Debug};
+use std::fmt::{self, Debug, Display};
 
 use nonempty::NonEmpty;
 use zcash_primitives::{
@@ -90,17 +90,57 @@ pub struct Proposal<FeeRuleT, NoteRef> {
     is_shielding: bool,
 }
 
-/// Errors that
+/// Errors that can occur in construction of a [`Proposal`].
 #[derive(Debug, Clone)]
 pub enum ProposalError {
-    RequestInvalid,
+    /// The total output value of the transaction request is not a valid Zcash amount.
+    RequestTotalInvalid,
+    /// The total of transaction inputs overflows the valid range of Zcash values.
     Overflow,
+    /// The input total and output total of the payment request are not equal to one another. The
+    /// sum of transaction outputs, change, and fees is required to be exactly equal to the value
+    /// of provided inputs.
     BalanceError {
         input_total: NonNegativeAmount,
         output_total: NonNegativeAmount,
     },
-    ShieldingFlagInvalid,
+    /// The `is_shielding` flag may only be set to `true` under the following conditions:
+    /// * The total of transparent inputs is nonzero
+    /// * There exist no Sapling inputs
+    /// * There provided transaction request is empty; i.e. the only output values specified
+    ///   are change and fee amounts.
+    ShieldingInvalid,
 }
+
+impl Display for ProposalError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ProposalError::RequestTotalInvalid => write!(
+                f,
+                "The total requested output value is not a valid Zcash amount."
+            ),
+            ProposalError::Overflow => write!(
+                f,
+                "The total of transaction inputs overflows the valid range of Zcash values."
+            ),
+            ProposalError::BalanceError {
+                input_total,
+                output_total,
+            } => write!(
+                f,
+                "Balance error: the output total {} was not equal to the input total {}",
+                u64::from(*output_total),
+                u64::from(*input_total)
+            ),
+            ProposalError::ShieldingInvalid => write!(
+                f,
+                "The proposal violates the rules for a shielding transaction."
+            ),
+        }
+    }
+}
+
+impl std::error::Error for ProposalError {}
 
 impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
     /// Constructs a validated [`Proposal`] from its constituent parts.
@@ -135,7 +175,7 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
 
         let request_total = transaction_request
             .total()
-            .map_err(|_| ProposalError::RequestInvalid)?;
+            .map_err(|_| ProposalError::RequestTotalInvalid)?;
         let output_total = (request_total + balance.total()).ok_or(ProposalError::Overflow)?;
 
         if is_shielding
@@ -143,7 +183,7 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
                 || sapling_input_total > NonNegativeAmount::ZERO
                 || request_total > NonNegativeAmount::ZERO)
         {
-            return Err(ProposalError::ShieldingFlagInvalid);
+            return Err(ProposalError::ShieldingInvalid);
         }
 
         if input_total == output_total {
--- a/zcash_client_backend/src/proto.rs
+++ b/zcash_client_backend/src/proto.rs
@@ -3,15 +3,31 @@
 use std::io;
 
 use incrementalmerkletree::frontier::CommitmentTree;
+
+use nonempty::NonEmpty;
 use zcash_primitives::{
     block::{BlockHash, BlockHeader},
-    consensus::BlockHeight,
+    consensus::{self, BlockHeight, Parameters},
+    memo::{self, MemoBytes},
     merkle_tree::read_commitment_tree,
     sapling::{self, note::ExtractedNoteCommitment, Node, Nullifier, NOTE_COMMITMENT_TREE_DEPTH},
-    transaction::TxId,
+    transaction::{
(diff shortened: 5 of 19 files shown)


zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,89 @@
...
+    bytes txid = 1;
+    ValuePool pool = 2;
+    uint32 index = 3;
+    int64 value = 4;
+}
+
+// The fee rule used in constructing a Proposal
+enum FeeRule {
+    // 10000 ZAT
+    PreZip313 = 0;
```
str4d, line 57: Use `Auto` for discriminant 0, so that if the proposer doesn't specify a fee rule, the "default" becomes "use whatever the wallet believes is the best fee rule".

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,89 @@
...
+    Transparent = 0;
+    Sapling = 2;
+}
+
+// The unique identifier, value pool, and amount for each proposed input
+message ProposedInput {
+    bytes txid = 1;
+    ValuePool pool = 2;
+    uint32 index = 3;
+    int64 value = 4;
```
str4d, line 51: - Consumers of a proposal will need to validate `value` against the input that was actually selected by `(txid, pool, index)`, and reject proposals as invalid if there is a mismatch.
- Input values are by definition unsigned, and we want to ensure we allow the full range (for when ZSAs need to be expressed in these inputs).
```suggestion
    uint64 value = 4;
```

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,89 @@
...
+    Zip313 = 1;
+    // MAX(10000, 5000 * logical_actions) ZAT
+    Zip317 = 2;
+}
+
+// The total value, fee amount, and change outputs of the proposed
+// transaction
+message TransactionBalance {
+    repeated ChangeValue proposedChange = 1;
+    int64 feeRequired = 2;
```
str4d, line 68: Fee is by definition unsigned.
```suggestion
    uint64 feeRequired = 2;
```

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,89 @@
...
+
+// An object wrapper for memo bytes, to facilitate representing the
+// `change_memo == None` case.
+message MemoBytes {
+    bytes value = 1;
+}
+
+// The amount and memo for a proposed Sapling change output.
+message SaplingChange {
+    int64 amount = 1;
```
str4d, line 86: Change is by definition unsigned.
```suggestion
    uint64 amount = 1;
```

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,89 @@
...
+enum FeeRule {
+    // 10000 ZAT
+    PreZip313 = 0;
+    // 1000 ZAT
+    Zip313 = 1;
+    // MAX(10000, 5000 * logical_actions) ZAT
+    Zip317 = 2;
+}
+
+// The total value, fee amount, and change outputs of the proposed
```
str4d, line 64: `TransactionBalance` doesn't include the total value.

zcash_client_backend/proto/proposal.proto
str4d: Hmm, except that this then leaves the change and fee fields overconstrained. Instead let's call this `Unspecified`, and then the proposal consumer can treat it as an invalid state.

zcash_client_backend/proto/proposal.proto
nuttycom: Ah, right; I removed the total transfer value because it's not really a piece of information that's relevant to the user. I'll fix the documentation.

zcash_client_backend/proto/proposal.proto
nuttycom: These were both int64 values because they're represented as `Amount` values in the Rust code. These should probably be made `NonNegativeAmount`s instead.

pacu:
Folks, do you have an expected deadline for accepting comments on this?
Just prioritizing here 😅

nuttycom:
I would really like to get this in to an SDK release immediately following
Spend-before-Sync, but the proposal to allow using ZIP-317 fees without
displaying the fees to the user up front takes the pressure off a little
bit. But ideally wallet devs would be able to adopt SbS and the proposal
system simultaneously.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,99 @@
...
+    string transactionRequest = 2;
+    // The unique identifier and amount for each proposed input
+    repeated ProposedInput inputs = 3;
+    // The total value, fee amount, and change outputs of the proposed
+    // transaction
+    TransactionBalance balance = 4;
+    // The fee rule used in constructing this proposal
+    FeeRule feeRule = 6;
+    // The minimum number of confirmations required for transaction inputs to be selected.
+    // Each of the proposed inputs is know to have at least this many confirmations.
```
daira, line 22:
```suggestion
    // Each of the proposed inputs must have at least this many confirmations.
```

daira: The draft ZIP 315 [proposes](https://github.com/zcash/zips/pull/607/files#diff-81a3ab3f008443ec864f316692e2e6a15808c87d22f0032a930ce76019a02a05R134-R147) a different number of confirmations for "trusted" and "untrusted" TXOs:
> ## Trusted and untrusted TXOs
> 
> A wallet SHOULD treat received TXOs that are outputs of transactions created
by the same wallet, as trusted TXOs. Wallets MAY enable users to mark specific
external transactions as trusted, allowing their received TXOs also to be
classified as trusted TXOs.
> 
> A wallet SHOULD have a policy that is clearly communicated to the user for
the number of confirmations needed to spend untrusted and trusted TXOs
respectively. The following confirmation policy is RECOMMENDED:
> 
> * 10 confirmations, for untrusted TXOs;
> * 3 confirmations, for trusted TXOs.

So this should be moved to `ProposedInput`, I think.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,99 @@
...
+
+syntax = "proto3";
+package cash.z.wallet.sdk.ffi;
+
+// A data structure that describes the inputs to be consumed and outputs to
+// be produced in a proposed transaction.
+message Proposal {
+    uint32 protoVersion = 1;
+    // ZIP 321 serialized transaction request
+    string transactionRequest = 2;
```
daira, line 13: What is the relation of this to the other fields? It seems redundant, and could be inconsistent with other fields.

nuttycom: This is not at all redundant; the transaction request is what encapsulates all of the information about transaction recipients.

zcash_client_backend/proto/proposal.proto
nuttycom: Instead of this being on `ProposedInput` it should probably be an object with `trusted_min_confirmations` and `untrusted_min_confirmations`. Implementing the ZIP 315 recommendation will require changes to the note selection APIs, because at present all of them only support a single `min_confirmations` value and don't reflect the ZIP 315 distinction.

daira: Actually I'm wrong; it is sufficient to use a single `minConfirmations` here. There are two reasons why:
1. Draft ZIP 315 is wrong in its conclusion that a higher confirmation depth should be used for untrusted shielded inputs. (["A wallet MUST NOT attempt to spend a TXO in a user-initiated transaction that is not confirmed-spendable."](https://github.com/zcash/zips/pull/607/files#r982891435), where "confirmed-spendable" enforces a higher confirmation depth for untrusted TXOs, including shielded ones.)
2. Even if draft ZIP 315 were right, `minConfirmations` is only a double-check that all of the proposed inputs have at least the given confirmation depth. It wouldn't stop a transaction creation algorithm from enforcing a greater confirmation depth on some inputs if that were useful (which it can be *for fully transparent transactions*).


## Why draft ZIP 315 is wrong

First let's consider a purely transparent, purely UTXO-based blockchain. The term "confirmation depth" is used in two related but distinct ways:

:one: When a wallet reports how much money has been received (either in total or from a given source), counting only TXOs with a given confirmation depth avoids relying on money that the wallet holder might not actually have in the case of a rollback.

:two: When creating a transaction, enforcing a confirmation depth on input selection reduces the risk of the transaction becoming invalid because the chain has rolled back past an input.

There are accidental rollbacks (a.k.a. orphan blocks), but we can reasonably assume that the confirmation depth for *untrusted* TXOs is enough to mitigate transaction unreliability due to those. So the question of whether we need to have separate confirmation depths for trusted vs untrusted inputs depends only on the possibility of rollback *attacks*.

A rollback attack is this: if an adversary has sent coins and obtained benefit from them being understood as received, it has an incentive to roll back the chain in order to also have the coins (instead of the recipient).

This incentive applies whenever a recipient has understood the coins to have been received and provided a benefit. It does not *directly* depend on the confirmation depth at which those inputs are ever used. However, a complicating factor is that if the recipient spent those coins just before the attack, then rolling back the chain far enough for the adversary to benefit will also roll back those spends by the recipient. This may affect the reputation of the recipient with counterparties to whom they spend those coins, and they will likely want to avoid that problem. As I understand it, that is the motivation for using the same confirmation depth for :one: and :two: in a transparent cryptocurrency (and may be a reason why :one: and :two: tend to be conflated in discussions).

----

Now consider how this changes in the case of a Zerocash-style cryptocurrency that uses anchors to identify commitment treestates. :one: is the same, and :two: still applies to transparent inputs. However, for transactions that are at-least-partly shielded, we *also* have:

:three: When creating an at-least-partly shielded transaction, enforcing a minimum depth on anchor selection reduces the risk of the transaction becoming invalid because the chain has rolled back past the anchor.

Is it ever beneficial to enforce a higher confirmation depth when choosing shielded inputs than the anchor depth [*]? No, because this will not change whether or not the transaction succeeds: that depends only on anchor invalidation (under the assumption that the probability of getting the same commitment treestate after a rollback is negligible [**]). Indeed, the consensus cannot tell how far back the shielded notes were output. And so the argument linking :one: and :two: above does not work to link :one: and :three:.

> [*] [In v5 transactions there is at most one Sapling anchor and at most one Orchard anchor. In v4 transactions there are separate anchors for each Sapling input.](https://zips.z.cash/protocol/protocol.pdf#txnencoding) By "anchor depth" I mean the minimum depth of any anchor in the transaction. In any case, we never want to choose different anchors for different Sapling Spends, because that leaks information :— potentially a lot of information if the reason we are doing it is to allow including an input that could not otherwise be included. Similarly, there is never any reason to choose Sapling and Orchard anchors at different depths in a mixed-pool transaction.

> [**] We can't actually infer that it is negligible because an attack might only affect fully transparent transactions, leaving all updates to commitment trees the same. However, that doesn't really affect the argument because we can't *rely* on the commitment treestates remaining the same.

Note that in transparent chains, when a transaction is in one of the rolled-back blocks in a rollback attack, it still has a good chance of getting into the new chain quickly (if not already included by the attacker), as long as its inputs are still present in the new chain. This doesn't apply to at-least-partly shielded transactions, which will have to be recreated if any anchor is invalidated.

And so, it's sufficient to have a single `minConfirmations` field. (Actually, `minTargetHeight` would be sufficient on its own, if we trust the transaction creator to calculate it correctly to enforce all of the confirmation depths. On the other hand, explicitly specifying `minConfirmations` is arguably less prone to error.)

We should also change the ZIP 315 draft to reflect the above reasoning (using the concept of "trusted" and "untrusted" TXOs for balance reporting but not for creation of at-least-partially shielded transactions).

pacu (commented):
Left non-blocking comments on PR. 

1. possible feature request: create proposals from ZIP-321 string-serialized requests
2. Evaluate ways to make it hard (or better impossible) for `Proposal`'s payment request property to be out of sync of a `Proposal` instance.
3. Consider adding "privacy policy" to the transaction proposal similar as the ones documented in `z_sendmany` and `z_shieldcoinbase` on Zcashd.

zcash_client_backend/CHANGELOG.md
```diff
@@ -85,26 +122,39 @@ and this library adheres to Rust's notion of
...
+- `zcash_client_backend::fees::ChangeValue::Sapling` is now a structured variant.
+  In addition to the existing change value, it now also carries an optional memo
+  to be associated with the change output.
+- `zcash_client_backend::fees::fixed::SingleOutputChangeStrategy::new` and
+  `zcash_client_backend::fees::zip317::SingleOutputChangeStrategy::new` each now
+  accept an additional `change_memo` argument.
 
 ### Removed
 - `zcash_client_backend::data_api`:
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
nuttycom: Oops, I forgot about this conflict, thanks!

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,99 @@
...
+    //
+    // The chain must contain at least this many blocks in order for the proposal to
+    // be executed.
+    uint32 minTargetHeight = 8;
+    // Returns the anchor height used in preparing the proposal.
+    //
+    // If, at the time that the proposal is executed, the anchor height
+    // required to satisfy / the minimum confirmation depth is less than this
+    // height, the proposal execution / API will return an error.
+    uint32 minAnchorHeight = 9;
```
daira, line 34: I considered saying that might be an overspecification to give both these heights (i.e. that it might be sufficient to specify the anchor depth at the time the proposal is executed, as long as the confirmation depth constraints are still met). However, on reflection I think that might leak information and also be more complicated.

zcash_client_backend/proto/proposal.proto
daira: The above argument has been reproduced and refined at https://github.com/zcash/zips/pull/607#issuecomment-1707243952 ; please read that version since there are subtle differences (in particular, there is no also benefit in choosing a greater confirmation depth for *transparent* inputs of a partly-shielded transaction than the anchor depth).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -384,6 +361,77 @@ where
...
+///   received note must have in the blockchain in order to be considered for being
+///   spent. A value of 10 confirmations is recommended and 0-conf transactions are
+///   not supported.
+/// * `to`: The address to which `amount` will be paid.
+/// * `amount`: The amount to send.
+/// * `memo`: A memo to be included in the output to the recipient.
+/// * `change_memo`: A memo to be included in any change output that is created.
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::type_complexity)]
+pub fn propose_standard_transfer_to_address<DbT, ParamsT, CommitmentTreeErrT>(
```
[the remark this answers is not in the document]
nuttycom: This can be done at the FFI layer; the caller simply passes the string they want across and then on the Rust side it gets deserialized and then the normal `propose_transfer` function. `propose_standard_transfer_to_address` is just a convenience method anyway.

zcash_client_backend/src/proto.rs
```diff
@@ -142,3 +164,226 @@ impl compact_formats::CompactSaplingSpend {
...
+    /// protobuf representation.
+    pub fn try_into_standard_proposal<P: consensus::Parameters, DbT: WalletRead>(
+        &self,
+        params: &P,
+        wallet_db: &DbT,
+    ) -> Result<Proposal<StandardFeeRule, DbT::NoteRef>, ProposalError<DbT::Error>> {
+        match self.proto_version {
+            PROPOSAL_SER_V1 => {
+                let transaction_request =
+                    TransactionRequest::from_uri(params, &self.transaction_request)?;
```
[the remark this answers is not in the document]
nuttycom: At present there's not any integrity checking, because we don't keep track of proposals that have been returned to the caller. That seems like a reasonable feature to add; we could add an integrity check and store the proposal or at least its hash, but I think that should be done in a separate PR to this one. If that's a feature that you want, please add an issue for it and we can discuss designs.

zcash_client_backend/proto/proposal.proto
nuttycom: Yeah, I would like the proposals to be essentially stable.

zcash_client_backend/src/proto.rs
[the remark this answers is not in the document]
pacu: I think I was not specific, I was referring to this comment https://github.com/zcash/librustzcash/pull/891/files#r1301913308, where the proposal protobuf can be changed and the Request representation can be out of sync with the protobuf itself.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,99 @@
...
+    // A flag indicating whether the proposal is for a shielding transaction,
+    // used for determining which OVK to select for wallet-internal outputs.
+    bool isShielding = 10;
```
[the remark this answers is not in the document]
nuttycom: The intent is that the proposal contain enough information to determine what privacy policies may or may not be satisfied, but this doesn't need to be a part of the proposal as the data within the proposal is sufficient to figure this out. The privacy policies used by `zcashd` don't cover all the axes of privacy that one might be interested in, and what information is leaked when using a light wallet depends upon a lot apart from what is revealed on-chain, so it doesn't make sense for `zcash_client_backend` to make claims about privacy that might might be violated by the light wallet.

zcash_client_backend/src/proto.rs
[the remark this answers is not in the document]
nuttycom: This is a good catch; parsing the `Proposal` should do balance integrity checks.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,99 @@
...
+    // 10000 ZAT
+    PreZip313 = 1;
+    // 1000 ZAT
+    Zip313 = 2;
+    // MAX(10000, 5000 * logical_actions) ZAT
+    Zip317 = 3;
+}
+
+// The total value, fee amount, and change outputs of the proposed
+// transaction
```
nuttycom, line 75:
```suggestion
// The proposed change outputs and fee amount.
```

zcash_client_backend/src/proto.rs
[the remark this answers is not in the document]
nuttycom: Fixed in a5d2a61624873161032e58fc4b5b44d2f46eea4e

zcash_client_backend/src/proto/proposal.rs
str4d, line 1: This file needs to be marked in `.gitattributes` as generated.

zcash_client_backend/src/fees/standard.rs
```diff
@@ -0,0 +1,101 @@
+//! Change strategies designed for use with a standard fee.
```
daira, line 1: Why are we adding new strategies that are called "standard"? The standard is ZIP 317.

I see that we might want names that don't assume that we won't change the applicable ZIP again. However, for the time being I think this just obscures the fact that we need everyone to move to ZIP 317 ASAP. (We needed them to do that months ago. I'm grumpy about it, as you might be able to tell.)

nuttycom: The main reason is that we need an enumerated type that defines the existing strategies, so we can correctly deserialize them when passed across the FFI boundary. Fully implementing ZIP 317 support requires UI changes that go beyond the scope of the change to the SDK, so we want to make it possible for wallet partners to make those UI changes on their own schedule.

tw0po1nt:
> Left non-blocking comments on PR.
> 
> 1. possible feature request: create proposals from ZIP-321 string-serialized requests
> 2. Evaluate ways to make it hard (or better impossible) for `Proposal`'s payment request property to be out of sync of a `Proposal` instance.
> 3. Consider adding "privacy policy" to the transaction proposal similar as the ones documented in `z_sendmany` and `z_shieldcoinbase` on Zcashd.

@nuttycom I think my only add here is that I second point 1 in pacu's comment here. I think that being able to pass a ZIP-321 string-serialized request will satisfy our requirements and delegate the more security-sensitive nature of parsing these URIs further down the stack, away from client app code.

nuttycom:
> > Left non-blocking comments on PR.
> > 
> > 1. possible feature request: create proposals from ZIP-321 string-serialized requests

According to the principal of least power, it's the responsibility of the caller to parse a string to a `TransactionRequest` that can then be provided to the `propose_transfer` function. Note that in `zcash-light-client-ffi` the method at the _ffi boundary_ can perform both operations, but we don't need the Rust API to do both in a single call.

> > 2. Evaluate ways to make it hard (or better impossible) for `Proposal`'s payment request property to be out of sync of a `Proposal` instance.

This is already the case? The transaction request that is serialized in the proposal protobuf is the sole place where the recipients of the payments are defined; parsing from the protobuf representation to the Rust `Proposal` type verifies that the proposal balances.

> > 3. Consider adding "privacy policy" to the transaction proposal similar as the ones documented in `z_sendmany` and `z_shieldcoinbase` on Zcashd.

This is out of scope for this particular PR; it's possible to use the information in a `Proposal`'s protobuf implementation to determine the privacy implications of the proposed transaction. I think that it would be better to leave such interpretation to the Kotlin/Swift side of the SDK, at least.

> @nuttycom I think my only add here is that I second point 1 in pacu's comment here. I think that being able to pass a ZIP-321 string-serialized request will satisfy our requirements and delegate the more security-sensitive nature of parsing these URIs further down the stack, away from client app code.

pacu:
> This is already the case? The transaction request that is serialized in the proposal protobuf is the sole place where the recipients of the payments are defined; parsing from the protobuf representation to the Rust `Proposal` type verifies that the proposal balances.

Great! Just wanted to point that out to have that clarified. I didn't see it considered at the moment. Thanks Kris!

>> Consider adding "privacy policy" to the transaction proposal similar as the ones documented in z_sendmany and z_shieldcoinbase on Zcashd.

>This is out of scope for this particular PR; it's possible to use the information in a Proposal's protobuf implementation to determine the privacy implications of the proposed transaction. I think that it would be better to leave such interpretation to the Kotlin/Swift side of the SDK, at least.

That's fair. I don't think that duplicating that logic in the native code of each SDK is a good idea though.

>According to the principal of least power, it's the responsibility of the caller to parse a string to a TransactionRequest that can then be provided to the propose_transfer function. Note that in zcash-light-client-ffi the method at the ffi boundary can perform both operations, but we don't need the Rust API to do both in a single call.

What I understand, please correct me if I'm wrong, is that it is already possible to do this by doing two API calls. First going from ZIP-321 URI to a `TransactionRequest`, and then secondly calling a function that receives that `TransactionRequest` and returns (if such proposal is possible) a Transaction Proposal.

nuttycom:
> > This is out of scope for this particular PR; it's possible to use the information in a Proposal's protobuf implementation to determine the privacy implications of the proposed transaction. I think that it would be better to leave such interpretation to the Kotlin/Swift side of the SDK, at least.
> 
> That's fair. I don't think that duplicating that logic in the native code of each SDK is a good idea though.

It'd be possible to create an enumeration similar to what we use in zcashd, but I'm not even sure how useful it would be given the spending restrictions imposed by the wallet SDK; we don't implement T->T, and with a shielding transaction there aren't any choices to be made, you either want to shield your transparent funds or not. Essentially the only meaningful policy would be `AllowRevealedAmounts` vs. `FullPrivacy`, and that won't become relevant until Orchard support is merged. 
 
> > According to the principal of least power, it's the responsibility of the caller to parse a string to a TransactionRequest that can then be provided to the propose_transfer function. Note that in zcash-light-client-ffi the method at the ffi boundary can perform both operations, but we don't need the Rust API to do both in a single call.
> 
> What I understand, please correct me if I'm wrong, is that it is already possible to do this by doing two API calls. First going from ZIP-321 URI to a `TransactionRequest`, and then secondly calling a function that receives that `TransactionRequest` and returns (if such proposal is possible) a Transaction Proposal.

It depends upon what you mean by "API calls" here. At the level of the FFI boundary, we pass only serialized data, which is the ZIP-321 URI; in Rust code, the parsing step to create a `TransactionRequest` from the URI is required in order to transform that URI to the type that the `propose_transfer` function requires. If the client wants to be able to inspect the transaction request, we should have parsers in Swift and Kotlin, so that those parsers can return data structures that are usable by client code; using the Rust parser will give you a pointer to a Rust `TransactionRequest` but then we'd have to convert all that data to a Swift or Kotlin object via a bunch of mess pulling that data back across the FFI boundary, and that's both a big hassle and totally unnecessary; my hope is that ZIP 321 parsers will one day be implemented in a whole bunch of languages; they're really pretty simple to write.

zcash_client_backend/src/proto.rs
```diff
@@ -177,3 +196,249 @@ impl service::TreeState {
...
+        params: &P,
+        value: &Proposal<StandardFeeRule, NoteRef>,
+    ) -> Option<Self> {
+        let transaction_request = value.transaction_request().to_uri(params)?;
+        let inputs = value
+            .sapling_inputs()
+            .iter()
+            .map(|rec_note| proposal::ProposedInput {
+                txid: rec_note.txid().as_ref().to_vec(),
+                pool: 2,
```
daira, line 239: We should have constants for these.

daira: Is this a pool code (`zcash_client_sqlite::wallet::pool_code`) or a typecode (`u32::from(zcash_address::kind::unified::Typecode))`)?

daira: Is this a pool code (`zcash_client_sqlite::wallet::pool_code`) or a typecode (`u32::from(zcash_address::kind::unified::Typecode))`)? It doesn't appear to be a `proposal::ValuePool`.

Too many similar concepts. Let's just use ZIP 316 typecodes everywhere.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,98 @@
...
+// The value pool from which a particular proposed input is drawn
+enum ValuePool {
+    // Protobuf requires that enums have a zero descriminant as the default
+    // value. However, we need to require that a known value pool is selected,
+    // and we do not want to fall back to any default, so sending the
+    // ValuePoolNotSpecified value will be treated as an error.
+    ValuePoolNotSpecified = 0;
+    Transparent = 1;
+    Sapling = 2;
+}
```
daira, line 49: Do we need to define a new enumeration of pools that is not a ZIP 316 typecode?

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,98 @@
...
+    // height, the proposal execution / API will return an error.
+    uint32 minAnchorHeight = 9;
+    // A flag indicating whether the proposal is for a shielding transaction,
+    // used for determining which OVK to select for wallet-internal outputs.
+    bool isShielding = 10;
+}
+
+// The value pool from which a particular proposed input is drawn
+enum ValuePool {
+    // Protobuf requires that enums have a zero descriminant as the default
```
daira, line 42:
```suggestion
    // Protobuf requires that enums have a zero discriminant as the default
```

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,98 @@
...
+message ProposedInput {
+    bytes txid = 1;
+    ValuePool pool = 2;
+    uint32 index = 3;
+    uint64 value = 4;
+}
+
+// The fee rule used in constructing a Proposal
+enum FeeRule {
+    // Protobuf requires that enums have a zero descriminant as the default
```
daira, line 61:
```suggestion
    // Protobuf requires that enums have a zero discriminant as the default
```

zcash_client_backend/src/proto.rs
```diff
@@ -177,3 +196,249 @@ impl service::TreeState {
...
+            min_confirmations: value.min_confirmations().into(),
+            min_target_height: value.min_target_height().into(),
+            min_anchor_height: value.min_anchor_height().into(),
+            is_shielding: value.is_shielding(),
+        })
+    }
+
+    /// Attempts to parse a [`Proposal`] based upon a supported [`StandardFeeRule`] from its
+    /// protobuf representation.
+    pub fn try_into_standard_proposal<P: consensus::Parameters, DbT: WalletRead>(
```
daira, line 298: It looks like this is not tested.

zcash_client_backend/src/zip321.rs
```diff
@@ -152,6 +152,21 @@ impl TransactionRequest {
         &self.payments[..]
     }
 
+    /// Returns the total value of payments to be made.
+    ///
+    /// Returns `Err` in the case of overflow, if payment values are negative, or the value is
+    /// outside the valid range of Zcash values. .
```
daira, line 158:
```suggestion
    /// outside the range `0..=MAX_MONEY` zatoshis.
```
"The valid range of Zcash values" is ambiguous.

zcash_client_sqlite/src/lib.rs
```diff
@@ -320,6 +328,19 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
+        &self,
+        _outpoint: &OutPoint,
+    ) -> Result<Option<WalletTransparentOutput>, Self::Error> {
+        #[cfg(feature = "transparent-inputs")]
+        return wallet::get_unspent_transparent_output(self.conn.borrow(), _outpoint);
+
+        #[cfg(not(feature = "transparent-inputs"))]
+        panic!(
+            "The wallet must be compiled with the transparent-inputs feature to use this method."
+        );
```
daira, line 341: This seems unfortunate. Why can't the existence of the method be conditional?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1298,45 +1358,18 @@ pub(crate) fn get_unspent_transparent_outputs<P: consensus::Parameters>(
...
-            SqliteClientError::CorruptedData(format!("Negative utxo value: {}", value_raw))
-        })?;
-        let height: u32 = row.get(4)?;
-
-        let outpoint = OutPoint::new(txid_bytes, index);
-        if excluded.contains(&outpoint) {
+        let output = to_unspent_transparent_output(row)?;
+        if excluded.contains(output.outpoint()) {
             continue;
         }
```
daira, line 1371: There must be a more functional way to express this, probably something like:
```rust
let utxos: Vec<_> = rows.filter_map(|row| {
    let output = to_unspent_transparent_output(row)?;
    if excluded.contains(output.outpoint()) {
        None
    } else {
        Some(output)
    }
}).collect();
```

zcash_primitives/CHANGELOG.md
```diff
@@ -79,7 +83,7 @@ and this library adheres to Rust's notion of
 - `impl From<zcash_primitive::components::transaction::Amount> for u64`
 
 ### Added
-- `transaction::fees::StandardFeeRule`
+>>>>>>> 89f0d17e7 (zcash_client_backend: Add serialization & parsing for protobuf Proposal representation.)
```
daira, line 86: This is a leftover conflict marker.

daira (changes requested):
Some comments are blocking, e.g. the leftover conflict marker.

zcash_client_backend/proto/proposal.proto
nuttycom: This is purely a protobuf-specific serialization; I don't think it makes sense to have a corresponding Rust type.

zcash_client_sqlite/src/lib.rs
nuttycom: Trait methods are not supposed to have their presence or absence determined by feature flags. The correct thing will be to factor this out into the `TransparentInputSource` type introduced by #1003.

zcash_client_sqlite/src/wallet.rs
nuttycom: `Rows` isn't a proper iterator. Unfortunately this is probably the best way, given the restrictions of rusqlite.

zcash_primitives/CHANGELOG.md
nuttycom, line 86:
```suggestion
```

zcash_client_backend/src/proto.rs
nuttycom: That's puzzling, it should be covered in the roundtrip tests. I'll double-check.

zcash_client_backend/src/data_api.rs
```diff
@@ -379,6 +387,13 @@ pub trait WalletRead {
         account: AccountId,
     ) -> Result<HashMap<TransparentAddress, AddressMetadata>, Self::Error>;
 
+    /// Returns a received transparent UTXO, or Ok(None) if the UTXO is not known to belong to the
+    /// wallet or is not spendable.
```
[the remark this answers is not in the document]
nuttycom: The important thing here is whether or not the output is available to be spent, i.e. is a proposal that contains that utxo as an input valid?

zcash_client_backend/src/proto.rs
nuttycom: This has been removed in favor of storing separate input data structures, which is necessary if Sapling and Orchard parts of the proposal use different anchors anyway.

zcash_client_backend/proto/proposal.proto
nuttycom: This has now been removed.

zcash_client_sqlite/src/lib.rs
nuttycom: Actually, it can't be conditional, because bounds can't vary based on feature flags.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,86 @@
...
+    // The chain must contain at least this many blocks in order for the proposal to
+    // be executed.
+    uint32 minTargetHeight = 7;
+    // A flag indicating whether the proposal is for a shielding transaction,
+    // used for determining which OVK to select for wallet-internal outputs.
+    bool isShielding = 8;
+}
+
+message SaplingInputs {
+    // Returns the Sapling anchor height to be used in creating the transaction.
```
daira, line 34:
```suggestion
    // The Sapling anchor height to be used in creating the transaction
```
Non-blocking.

zcash_client_backend/src/proto/proposal.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[prost(uint64, tag = "1")]
+    pub amount: u64,
+    #[prost(message, optional, tag = "2")]
+    pub memo: ::core::option::Option<MemoBytes>,
+}
+/// The fee rule used in constructing a Proposal
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+#[repr(i32)]
+pub enum FeeRule {
+    /// Protobuf requires that enums have a zero descriminant as the default
```
daira, line 102: Note to self: check this is regenerated when the comment typo (should be "discriminant") is fixed.

daira (approved):
utACK with minor comments.

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +204,223 @@ impl service::TreeState {
...
+                                    })
+                                })
+                        })
+                        .collect::<Result<Vec<_>, _>>()
+                        .map(|notes| {
+                            NonEmpty::from_vec(notes).map(|notes| {
+                                SaplingInputs::from_parts(s_in.anchor_height.into(), notes)
+                            })
+                        })
+                        .transpose()
```
str4d, line 377: This map is a bit weird. We already map over `self.sapling_inputs` which represents the optionality of all Sapling inputs. This `map` introduces a non-canonical encoding that can contain an anchor height but no notes, and the parser will drop the anchor height (meaning that this non-canonical encoding is not round-trip compatible). I feel like this case should instead be rejected with something like `ProposalError::NonCanonical`. I know that protobuf itself does not have canonical encodings (mainly due to unconstrained field ordering and backwards compatibility support), but I feel that in this particular case it is a structurally invalid representation (so maybe we don't call this "non-canonical" and instead some other kind of error).

Non-blocking thought prompted by this: we should set up proptests and/or a fuzzer for this proposal format.

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +204,223 @@ impl service::TreeState {
...
+
+                let balance = self.balance.as_ref().ok_or(ProposalError::BalanceInvalid)?;
+                let balance = TransactionBalance::new(
+                    balance
+                        .proposed_change
+                        .iter()
+                        .filter_map(|cv| {
+                            cv.value
+                                .as_ref()
+                                .map(|cv| -> Result<ChangeValue, ProposalError<_>> {
```
str4d, line 388: When can `cv.value` be `None`? That would imply a protobuf encoding of a non-empty sequence of default enums? I feel like this should also be rejected.

Non-blocking because this is purely a representational non-canonicity, and thus is closer to the non-canonicity in the rest of protobuf (vs my earlier comment which points to a more structural issue).

zcash_client_sqlite/src/lib.rs
```diff
@@ -196,6 +204,19 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> TransparentInput
...
+    }
```
str4d, line 218: This makes me think that the removal of the feature flag on the trait definition in `zcash_client_backend` is a rebase bug. Please fix if so.

In any case, this impl should be adjusted (because the entire trait impl is guarded by the feature flag):
```suggestion
    fn get_unspent_transparent_output(
        &self,
        outpoint: &OutPoint,
    ) -> Result<Option<WalletTransparentOutput>, Self::Error> {
        wallet::get_unspent_transparent_output(self.conn.borrow(), outpoint)
    }
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -1061,3 +1070,15 @@ pub(crate) fn input_selector(
     let change_strategy = standard::SingleOutputChangeStrategy::new(fee_rule, change_memo);
     GreedyInputSelector::new(change_strategy, DustOutputPolicy::default())
 }
+
+pub(crate) fn check_proposal_serialization_roundtrip(
```
str4d, line 1074: Add documentation to this method noting that this is not actually checking the on-the-wire serialization (as protobuf [does not provide a canonical encoding](https://github.com/protocolbuffers/protobuf/issues/3417)), but is just checking the round-trip of the "structural encoding" to the protobuf encoder's types.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -135,6 +135,38 @@ fn to_spendable_note(row: &Row) -> Result<ReceivedSaplingNote<ReceivedNoteId>, S
     ))
 }
 
+// The `clippy::let_and_return` lint is explicitly allowed here because a bug in the Rust compiler
+// (https://github.com/rust-lang/rust/issues/114633) fails to identify that the `result` temporary
```
str4d, line 139:
```suggestion
// The `clippy::let_and_return` lint is explicitly allowed here because a bug in Clippy
// (https://github.com/rust-lang/rust-clippy/issues/11308) means it fails to identify that the `result` temporary
```

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +204,223 @@ impl service::TreeState {
...
+impl<E> From<Zip321Error> for ProposalError<E> {
+    fn from(value: Zip321Error) -> Self {
+        Self::Zip321(value)
+    }
+}
+
+impl proposal::ProposedInput {
+    pub fn parse_txid(&self) -> Result<TxId, Vec<u8>> {
+        Ok(TxId::from_bytes(self.txid.clone().try_into()?))
+    }
```
str4d, line 232: Could use at least token documentation. Also instead of cloning, use the `TryFrom<&[T]>` impl that returns [TryFromSliceError](https://doc.rust-lang.org/std/array/struct.TryFromSliceError.html) instead of the vec:
```suggestion
    pub fn parse_txid(&self) -> Result<TxId, TryFromSliceError> {
        Ok(TxId::from_bytes(self.txid[..].try_into()?))
    }
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this library adheres to Rust's notion of
   - `ScannedBlock::sapling_tree_size`.
   - `ScannedBlock::orchard_tree_size`.
   - `wallet::propose_standard_transfer_to_address`
+  - `wallet::input_selection::Proposal::from_parts`
```
str4d, line 18: This is `pub(crate)` so isn't part of the public API.
```suggestion
```

str4d (changes requested):
Reviewed as of 33169719cee44242f29095c9a8d52e338f3b9f9d.

zcash_client_backend/src/data_api.rs
```diff
@@ -358,6 +358,14 @@ pub trait WalletRead {
         query: NullifierQuery,
     ) -> Result<Vec<(AccountId, sapling::Nullifier)>, Self::Error>;
 
+    /// Returns a received Sapling note, or Ok(None) if the note is not known to belong to the
+    /// wallet or if the note is not spendable.
```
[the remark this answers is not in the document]
str4d: This was from my in-progress review back at the beginning of October; IDK if it is still relevant.

zcash_client_backend/src/data_api.rs
```diff
@@ -386,6 +394,13 @@ pub trait WalletRead {
         account: AccountId,
     ) -> Result<HashMap<TransparentAddress, AddressMetadata>, Self::Error>;
 
+    /// Returns a received transparent UTXO, or Ok(None) if the UTXO is not known to belong to the
+    /// wallet or is not spendable.
```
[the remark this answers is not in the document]
str4d: This was from my in-progress review back at the beginning of October; IDK if it is still relevant.

zcash_client_backend/src/data_api.rs
```diff
@@ -235,11 +243,17 @@ pub trait SaplingInputSource {
 
 /// A trait representing the capability to query a data store for unspent transparent UTXOs
 /// belonging to a wallet.
-#[cfg(feature = "transparent-inputs")]
```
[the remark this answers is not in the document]
nuttycom: It hasn't moved; it has to be removed for the deserialization to be able to deserialize the transparent bits of a proposal.

It can be implemented as a no-op, but the interface has to be public.

zcash_client_backend/proto/proposal.proto
```diff
@@ -0,0 +1,86 @@
...
+    // The fee rule used in constructing this proposal
+    FeeRule feeRule = 6;
+    // The target height for which the proposal was constructed
+    //
+    // The chain must contain at least this many blocks in order for the proposal to
+    // be executed.
+    uint32 minTargetHeight = 7;
+    // A flag indicating whether the proposal is for a shielding transaction,
+    // used for determining which OVK to select for wallet-internal outputs.
+    bool isShielding = 8;
```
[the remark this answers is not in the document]
nuttycom: The essence of a shielding transaction is that its associated `TransactionRequest` specifies no recipients, and we need some way of indicating that the proposal is to send all of the input value to the wallet's internal change address. I think that this is better as an explicit flag than as an implicit detail of the transaction's structure; however, we could impose an additional validation criterion that the `shielding` flag must be `false` when the `TransactionRequest` has recipients, and that when it is `true` that the `TransactionRequest` be empty.

I guess that if we wanted a structural solution, then we could look at the value of the change, but that gets complicated.

zcash_client_backend/CHANGELOG.md
nuttycom: Hm, perhaps it should be, though? Otherwise the only way to construct a proposal is via parsing, which is pretty inconvenient if you're trying to implement your own fee/change strategy.

zcash_client_backend/src/proto.rs
nuttycom: Yeah, this seems okay; we can treat a `SaplingInputs` that has an anchor height but no inputs as an error.

zcash_client_backend/src/proto.rs
nuttycom: All object-valued types are `Option`al in protobuf encodings, so we have to handle that case. If there's no proposed change value for an entry, we can also treat that as an encoding error.

zcash_client_sqlite/src/lib.rs
nuttycom: The removal of the feature flag is not a bug; the alternative here to a panic is to return `Ok(None)` but I think that this would obscure the fact that the user is doing something wrong with the way that they're using the library.

zcash_client_backend/src/proto.rs
str4d: Yep, that was my non-blocking conclusion above.

zcash_client_sqlite/src/lib.rs
str4d: Removing the feature flag on the trait definition seems fine (per your rationale in the other comment). But this PR keeps the feature flag on the whole trait implementation, meaning that my suggestion above is entirely correct in that context.

zcash_client_backend/CHANGELOG.md
str4d: That seems fine, as long as `Proposal::from_parts` is enforcing all necessary internal consistency relations.

zcash_client_backend/proto/proposal.proto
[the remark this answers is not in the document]
str4d: > The essence of a shielding transaction is that its associated `TransactionRequest` specifies no recipients

And also that the proposal's `transparentInputs` are non-empty, and (for the current design decisions in `zcash_client_backend`) `saplingInputs` is empty.

If you check these as well in the validation criteria then I'm okay with this flag.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -89,7 +90,120 @@ pub struct Proposal<FeeRuleT, NoteRef> {
...
+
+        let request_total = transaction_request
+            .total()
+            .map_err(|_| ProposalError::RequestTotalInvalid)?;
+        let output_total = (request_total + balance.total()).ok_or(ProposalError::Overflow)?;
+
+        if is_shielding
+            && (transparent_input_total == NonNegativeAmount::ZERO
+                || sapling_input_total > NonNegativeAmount::ZERO
+                || request_total > NonNegativeAmount::ZERO)
```
str4d, line 184: Should we constrain the inverse as well, to match the `zcash_client_backend` design decisions? i.e. `if !is_shielding && transparent_input_total > NonNegativeAmount::ZERO`?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -89,7 +90,120 @@ pub struct Proposal<FeeRuleT, NoteRef> {
...
+            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| {
+                (acc? + a).ok_or(ProposalError::Overflow)
+            })?;
+        let sapling_input_total = sapling_inputs
+            .iter()
+            .flat_map(|s_in| s_in.notes().iter())
+            .map(|out| out.value())
+            .fold(Ok(NonNegativeAmount::ZERO), |acc, a| {
+                (acc? + a).ok_or(ProposalError::Overflow)
+            })?;
```
str4d, line 172: Nit: this would be equivalent:
```suggestion
            .fold(Some(NonNegativeAmount::ZERO), |acc, a| (acc? + a))
            .ok_or(ProposalError::Overflow)?;
```

zcash_client_backend/src/zip321.rs
```diff
@@ -426,8 +489,22 @@ mod parse {
         Other(String, String),
     }
 
+    impl Param {
+        /// Returns the name of the parameter from which this value was parsed.
+        pub fn name(&self) -> String {
```
str4d, line 494: Changelog addition (or make it crate-private if you only want to use it for nice errors).

str4d (dismissed):
utACK 7aab6fd7a7ea4248919b0571138a983945de1575 with a few nits.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: I don't think so; the `shielding` flag _adds_ a constraint, but setting the `shielding` flag to `false` means that the transaction is unconstrained except for that it must balance.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -89,7 +90,120 @@ pub struct Proposal<FeeRuleT, NoteRef> {
...
+        }
+    }
+}
+
+impl std::error::Error for ProposalError {}
+
 impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
+    /// Constructs a validated [`Proposal`] from its constituent parts.
+    ///
+    /// This operation validaes the proposal for balance consistency and agreement between
```
daira, line 148:
```suggestion
    /// This operation validates the proposal for balance consistency and agreement between
```

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +208,300 @@ impl service::TreeState {
...
+
+/// Errors that can occur in the process of decoding a [`Proposal`] from its protobuf
+/// representation.
+#[derive(Debug, Clone)]
+pub enum ProposalDecodingError<DbError> {
+    /// The ZIP 321 transaction request URI was invalid.
+    Zip321(Zip321Error),
+    /// A transaction identifier string did not decode to a valid transaction ID.
+    TxIdInvalid(TryFromSliceError),
+    /// A failure occurred trying to retrievean unspent note or UTXO from the wallet database.
```
daira, line 223:
```suggestion
    /// A failure occurred trying to retrieve an unspent note or UTXO from the wallet database.
```

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +208,300 @@ impl service::TreeState {
...
+    InputRetrieval(DbError),
+    /// The unspent note or UTXO corresponding to a proposal input was not found in the wallet
+    /// database.
+    InputNotFound(TxId, PoolType, u32),
+    /// The transaction balance, or a component thereof, failed to decode correctly.
+    BalanceInvalid,
+    /// Failed to decode a ZIP-302 compliant memo from the provided memo bytes.
```
daira, line 230:
```suggestion
    /// Failed to decode a ZIP-302-compliant memo from the provided memo bytes.
```

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +208,300 @@ impl service::TreeState {
...
+impl<E: Display> Display for ProposalDecodingError<E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ProposalDecodingError::Zip321(err) => write!(f, "Transaction request invalid: {}", err),
+            ProposalDecodingError::TxIdInvalid(err) => {
+                write!(f, "Invalid transaction id: {:?}", err)
+            }
+            ProposalDecodingError::InputRetrieval(err) => write!(
+                f,
+                "An error occurred retrieving a transaction intput: {}",
```
daira, line 257:
```suggestion
                "An error occurred retrieving a transaction input: {}",
```

zcash_client_backend/src/proto.rs
```diff
@@ -183,3 +208,300 @@ impl service::TreeState {
...
+    /// Failed to decode a ZIP-302 compliant memo from the provided memo bytes.
+    MemoInvalid(memo::Error),
+    /// The serialization version returned by the protobuf was not recognized.
+    VersionInvalid(u32),
+    /// The proposal did not correctly specify a standard fee rule.
+    FeeRuleNotSpecified,
+    /// The proposal violated balance or structural constraints.
+    ProposalInvalid(ProposalError),
+    /// A `sapling_inputs` field was present, but contained no input note references.
+    EmptySaplingInputs,
```
daira, line 239: Should this be `EmptyShieldedInputs(PoolType)`? Non-blocking.

zcash_client_backend/src/zip321.rs
```diff
@@ -45,6 +48,51 @@ pub enum Zip321Error {
     ParseError(String),
 }
 
+impl Display for Zip321Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Zip321Error::InvalidBase64(err) => {
+                write!(f, "Memo value was not correctly base64 encoded: {:?}", err)
```
daira, line 55:
```suggestion
                write!(f, "Memo value was not correctly base64-encoded: {:?}", err)
```

zcash_client_backend/src/zip321.rs
```diff
@@ -45,6 +48,51 @@ pub enum Zip321Error {
...
+            }
+            Zip321Error::MemoBytesError(err) => write!(
+                f,
+                "Memo exceeded maximum length or violated utf-8 encodiding restrictions: {:?}.",
```
daira, line 59:
```suggestion
                "Memo exceeded maximum length or violated UTF-8 encoding restrictions: {:?}",
```

zcash_client_backend/src/zip321.rs
```diff
@@ -45,6 +48,51 @@ pub enum Zip321Error {
...
+            ),
+            Zip321Error::DuplicateParameter(param, idx) => write!(
+                f,
+                "There is a duplicate {} parameter at index {}",
+                param.name(),
+                idx
+            ),
+            Zip321Error::TransparentMemo(idx) => write!(
+                f,
+                "Payment {} is invalid: cannot send a memo to a transparent recipient address.",
```
daira, line 75:
```suggestion
                "Payment {} is invalid: cannot send a memo to a transparent recipient address",
```

zcash_client_backend/src/zip321.rs
```diff
@@ -152,6 +200,21 @@ impl TransactionRequest {
         &self.payments[..]
     }
 
+    /// Returns the total value of payments to be made.
+    ///
+    /// Returns `Err` in the case of overflow, or the value is
```
daira, line 205:
```suggestion
    /// Returns `Err` in the case of overflow, or if the value is
```

daira (approved):
utACK with typo fix suggestions and a non-blocking comment.

zcash_client_backend/CHANGELOG.md
```diff
@@ -23,6 +24,21 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::wallet`:
   - `ReceivedSaplingNote::from_parts`
   - `ReceivedSaplingNote::{txid, output_index, diversifier, rseed, note_commitment_tree_position}`
+- `zcash_client_backend::zip321::TransactionRequest::total`
```
nuttycom, line 27:
```suggestion
- `zcash_client_backend::zip321::TransactionRequest::total`
- `zcash_client_backend::zip321::parse::Param::name`
```

zcash_client_backend/src/proto.rs
nuttycom: Hm, yeah, Orchard will probably have the same issue, so I should generalize it now.

str4d (approved):
Re-utACK c5f48f50e1aa8c48fe236db623391e9e1121d351

daira (commented):
Post-hoc ACK

PR #893, opened 2023-08-09
Scanning bugfixes

str4d:


zcash_client_sqlite/src/chain.rs
```diff
@@ -58,8 +58,18 @@ where
...
         .map_err(to_chain_error)?;
 
+    // Only look for the `from_height` in the scanned blocks if it is set.
+    let mut from_height_found = from_height.is_none();
     while let Some(row) = rows.next().map_err(to_chain_error)? {
         let height = BlockHeight::from_u32(row.get(0).map_err(to_chain_error)?);
+        if !from_height_found {
+            // We will only perform this check on the first row.
+            if from_height != Some(height) {
+                break;
```
nuttycom, line 68: nonblocking: I slightly prefer to return the error from here directly, and put the `from_height_found = true;` into an `else`.

zcash_client_sqlite/src/lib.rs
```diff
@@ -882,6 +882,7 @@ pub enum FsBlockDbError {
     MissingBlockPath(PathBuf),
     InvalidBlockstoreRoot(PathBuf),
     InvalidBlockPath(PathBuf),
+    InvalidParameter(String),
```
nuttycom, line 885: blocking: The intent of this particular error is that it be specifically identifiable (and recoverable). So a generic `InvalidParameter` is not the right approach, because you'd have to parse the string to identify why it failed. So this should be something explicit like `CacheMiss(BlockHeight)` so that the caller can then go and retrieve the block.

zcash_client_sqlite/src/chain.rs
```diff
@@ -260,8 +280,18 @@ where
...
         .map_err(to_chain_error)?;
 
+    // Only look for the `from_height` in the scanned blocks if it is set.
+    let mut from_height_found = from_height.is_none();
     for row_result in rows {
         let cbr = row_result.map_err(to_chain_error)?;
+        if !from_height_found {
+            // We will only perform this check on the first row.
+            if from_height != Some(cbr.height) {
+                break;
```
nuttycom, line 290: nonblocking ditto

nuttycom (changes requested):
One blocking change requested.

zcash_client_sqlite/src/chain.rs
str4d: We need to return an error in two places: here, and in the case where there are no rows at all (i.e. `from_height` was set to a height above any block in the database). To avoid a second SQL query (and the transactional logic associated with it), while keeping the error logic simple, I wrote it this way. But with the other proposed change, the inline errors might be straightforward enough that we can duplicate them.

nuttycom: Ah, okay, avoiding the duplication is better then.

zcash_client_sqlite/src/lib.rs
str4d: A downside to this is that it's less clear to the caller that the problem is the `from_height` parameter, rather than "any" cache miss. So this change trades off one issue for another, but assuming that's acceptable I'm fine with going fo the option that can be matched on.

nuttycom (approved):
utACK

str4d:
Tested with the iOS SDK; this PR combined with a change over there gets us past the main scanning section, and into a different set of bugs! If there are more changes necessary over here, they can go into another PR.

PR #894, opened 2023-08-09
zcash_client_sqlite: Only select notes for which witnesses can be constructed.

nuttycom:
This change modifies the implementation of `get_spendable_sapling_notes` and `select_spendable_sapling_notes` to only return notes at positions where the associated note commitment tree shard has been fully scanned. This is slightly more conservative than it needs to be, because there could be cases where witnesses imported into the tree in the `shardtree_support` migration cover the complete range of a subtree (and hence that subtree doesn't need to be re-scanned). However, we can't detect or depend upon that condition in general without attempting to create a witness for each note retrieved.

A possible alternative to this approach would be to not bound our query results on the requested total, and instead attempt to construct a witness for each note we retrieve, skipping the note if we cannot construct a witness. However, given that accessing the note commitment tree can be a costly operation requiring nontrivial deserialization costs, the more conservative database-oriented approach is perhaps better.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -353,14 +353,14 @@ mod tests {
 
     use zcash_client_backend::{
         address::RecipientAddress,
-        data_api::WalletRead,
+        data_api::{scanning::ScanPriority, WalletRead},
         encoding::{encode_extended_full_viewing_key, encode_payment_address},
         keys::{sapling, UnifiedFullViewingKey, UnifiedSpendingKey},
     };
 
     use zcash_primitives::{
         block::BlockHash,
-        consensus::{BlockHeight, BranchId, Parameters},
+        consensus::{BlockHeight, BranchId, NetworkUpgrade, Parameters},
         transaction::{TransactionData, TxVersion},
         zip32::sapling::ExtendedFullViewingKey,
     };
@@ -368,6 +368,7 @@ mod tests {
     use crate::{
         error::SqliteClientError,
         tests::{self, network},
+        wallet::scanning::priority_code,
         AccountId, WalletDb,
     };
 
@@ -558,6 +559,33 @@ mod tests {
         }
 
         let expected_views = vec![
+            // v_sapling_shard_unscanned_ranges
+            format!(
+                "CREATE VIEW v_sapling_shard_unscanned_ranges AS
+                SELECT
+                    shard.shard_index,
+                    shard.shard_index << 16 AS start_position,
+                    (shard.shard_index + 1) << 16 AS end_position_exclusive,
+                    IFNULL(prev_shard.subtree_end_height, {}) AS subtree_start_height,
+                    shard.subtree_end_height AS subtree_end_height,
+                    shard.contains_marked,
+                    scan_queue.block_range_start,
+                    scan_queue.block_range_end,
+                    scan_queue.priority
+                FROM sapling_tree_shards shard
+                LEFT OUTER JOIN sapling_tree_shards prev_shard
+                    ON shard.shard_index = prev_shard.shard_index + 1
+                INNER JOIN scan_queue ON 
+                    (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    ((scan_queue.block_range_end - 1) BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    (
+                        scan_queue.block_range_start <= prev_shard.subtree_end_height
+                        AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
+                    )
+                WHERE scan_queue.priority != {}",
+                u32::from(tests::network().activation_height(NetworkUpgrade::Sapling).unwrap()),
+                priority_code(&ScanPriority::Scanned),
+            ),
             // v_transactions
             "CREATE VIEW v_transactions AS
             WITH
@@ -649,7 +677,7 @@ mod tests {
             LEFT JOIN sent_note_counts
                       ON sent_note_counts.account_id = notes.account_id
                       AND sent_note_counts.id_tx = notes.id_tx
-            GROUP BY notes.account_id, transactions.id_tx",
+            GROUP BY notes.account_id, transactions.id_tx".to_owned(),
             // v_tx_outputs
             "CREATE VIEW v_tx_outputs AS
             SELECT sapling_received_notes.tx           AS id_tx,
@@ -693,7 +721,7 @@ mod tests {
                       ON (sent_notes.tx, sent_notes.output_pool, sent_notes.output_index) =
                          (sapling_received_notes.tx, 2, sapling_received_notes.output_index)
             WHERE  sapling_received_notes.is_change IS NULL
-               OR  sapling_received_notes.is_change = 0"
+               OR  sapling_received_notes.is_change = 0".to_owned(),
         ];
 
         let mut views_query = db_data
@@ -706,7 +734,7 @@ mod tests {
             let sql: String = row.get(0).unwrap();
             assert_eq!(
                 re.replace_all(&sql, " "),
-                re.replace_all(expected_views[expected_idx], " ")
+                re.replace_all(&expected_views[expected_idx], " ")
             );
             expected_idx += 1;
         }
@@ -971,7 +999,7 @@ mod tests {
             wdb.conn.execute(
                 "INSERT INTO transactions (block, id_tx, txid, raw) VALUES (0, 0, :txid, :tx_bytes)",
                 named_params![
-                    ":txid": tx.txid().as_ref(), 
+                    ":txid": tx.txid().as_ref(),
                     ":tx_bytes": &tx_bytes[..]
                 ],
             )?;
--- a/zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
@@ -5,7 +5,7 @@ use std::collections::HashSet;
 
 use schemer_rusqlite::RusqliteMigration;
 use uuid::Uuid;
-use zcash_client_backend::data_api::scanning::ScanPriority;
+use zcash_client_backend::data_api::{scanning::ScanPriority, SAPLING_SHARD_HEIGHT};
 use zcash_primitives::consensus;
 
 use crate::wallet::{init::WalletMigrationError, scanning::priority_code};
@@ -41,8 +41,8 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                 "CREATE VIEW v_sapling_shard_unscanned_ranges AS
                 SELECT
                     shard.shard_index,
-                    shard.shard_index << 16 AS start_position,
-                    (shard.shard_index + 1) << 16 AS end_position_exclusive,
+                    shard.shard_index << {} AS start_position,
+                    (shard.shard_index + 1) << {} AS end_position_exclusive,
                     IFNULL(prev_shard.subtree_end_height, {}) AS subtree_start_height,
                     shard.subtree_end_height AS subtree_end_height,
                     shard.contains_marked,
@@ -60,6 +60,8 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                         AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
                     )
                 WHERE scan_queue.priority != {}",
+                SAPLING_SHARD_HEIGHT,
+                SAPLING_SHARD_HEIGHT,
                 u32::from(self.params.activation_height(consensus::NetworkUpgrade::Sapling).unwrap()),
                 priority_code(&ScanPriority::Scanned),
             )
--- a/zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
@@ -1,0 +1,75 @@
+//! This migration adds a view that returns the un-scanned ranges associated with each sapling note
+//! commitment tree shard.
+
+use std::collections::HashSet;
+
+use schemer_rusqlite::RusqliteMigration;
+use uuid::Uuid;
+use zcash_client_backend::data_api::scanning::ScanPriority;
+use zcash_primitives::consensus;
+
+use crate::wallet::{init::WalletMigrationError, scanning::priority_code};
+
+use super::shardtree_support;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0xfa934bdc_97b6_4980_8a83_b2cb1ac465fd);
+
+pub(super) struct Migration<P> {
+    pub(super) params: P,
+}
+
+impl<P> schemer::Migration for Migration<P> {
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        [shardtree_support::MIGRATION_ID].into_iter().collect()
+    }
+
+    fn description(&self) -> &'static str {
+        "Adds a view that returns the un-scanned ranges associated with each sapling note commitment tree shard."
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        transaction.execute_batch(
+            &format!(
+                "CREATE VIEW v_sapling_shard_unscanned_ranges AS
+                SELECT
+                    shard.shard_index,
+                    shard.shard_index << 16 AS start_position,
+                    (shard.shard_index + 1) << 16 AS end_position_exclusive,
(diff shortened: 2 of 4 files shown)


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -558,6 +559,33 @@ mod tests {
...
+                    scan_queue.priority
+                FROM sapling_tree_shards shard
+                LEFT OUTER JOIN sapling_tree_shards prev_shard
+                    ON shard.shard_index = prev_shard.shard_index + 1
+                INNER JOIN scan_queue ON 
+                    (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    (scan_queue.block_range_end BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    (
+                        scan_queue.block_range_start <= prev_shard.subtree_end_height
+                        AND scan_queue.block_range_end >= shard.subtree_end_height
```
daira, line 583: This could be `>` because the `==` case is caught by the previous `BETWEEN` clause. It's correct as-is (assuming inclusive ranges) though.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -132,14 +132,33 @@ pub(crate) fn get_spendable_sapling_notes(
...
          INNER JOIN transactions ON transactions.id_tx = sapling_received_notes.tx
          WHERE account = :account
          AND spent IS NULL
          AND transactions.block <= :anchor_height
-         AND id_note NOT IN rarray(:exclude)",
+         AND id_note NOT IN rarray(:exclude)
+         AND NOT EXISTS (
+            SELECT 1 FROM v_sapling_shard_unscanned_ranges unscanned
+            WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
+            AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
```
daira, line 159: Consider `WHERE sapling_received_notes.commitment_tree_position BETWEEN unscanned.start_position AND unscanned.end_position_exclusive - 1`.

(`BETWEEN` does the correct thing, i.e. return false, when its end value is less than its start value.)

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -164,10 +183,21 @@ pub(crate) fn select_spendable_sapling_notes(
...
+        // if the tip shard has unscanned ranges below the anchor height, none of our notes can be
+        // spent
+        return Ok(vec![]);
+    }
+
     // The goal of this SQL statement is to select the oldest notes until the required
-    // value has been reached, and then fetch the witnesses at the desired height for the
-    // selected notes. This is achieved in several steps:
-    //
+    // value has been reached.
```
daira, line 200:
```suggestion
    // value has been reached.
    //
```

daira: "up to" on lines 204 and 206 mean "up to and including", correct? Please make that explicit.

daira: Point 3 does not take into account that a note can have zero value. Thus, taking notes with cumulative value less than the required value and then one extra note, does not necessarily satisfy the required value. It is necessary to take notes with cumulative value **less than or equal to** the required value, and then one extra note (which necessarily then has a non-zero value) if needed.

daira (approved):
utACK modulo the corner-case bug in https://github.com/zcash/librustzcash/pull/894/files#r1291221657 (which could be addressed in a separate PR).

zcash_client_sqlite/src/wallet/sapling.rs
daira: Also, taking a note may require an increase in the fee. I'm not sure whether that is handled at a higher level.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -558,6 +559,33 @@ mod tests {
...
+                    shard.subtree_end_height AS subtree_end_height,
+                    shard.contains_marked,
+                    scan_queue.block_range_start,
+                    scan_queue.block_range_end,
+                    scan_queue.priority
+                FROM sapling_tree_shards shard
+                LEFT OUTER JOIN sapling_tree_shards prev_shard
+                    ON shard.shard_index = prev_shard.shard_index + 1
+                INNER JOIN scan_queue ON 
+                    (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
```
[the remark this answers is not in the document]
nuttycom: The naming problem is hard here. It could be that `subtree_last_position_height` is a better name, but the query is inclusive on both ends because both the first and last position of notes within a subtree may be in the middle of blocks. It's a wart that this view is needed at all, really; had we waited longer to release the updated `lightwalletd` we might have decided to include the start height information in the shard metadata directly, and so wouldn't have to compute it and could have just done a normal join against the `sapling_tree_shards` table.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: There are three cases that we're concerned with: the block range overlaps the start of the subtree range, the block range overlaps the end of the subtree range, or the block range fully contains the subtree range. I thought expressing it this way was clearest, because you can verify that each case is independently correct.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: On further inspection, I don't think this is a bug:

https://github.com/zcash/librustzcash/pull/894/files#diff-9cea07a2b90189d927a2737a964494dfea6ab1bc5a51342e481210b93586716eR218-R219

creates a running sum over the rows, and then 

https://github.com/zcash/librustzcash/pull/894/files#diff-9cea07a2b90189d927a2737a964494dfea6ab1bc5a51342e481210b93586716eR237-R238

ensures that a row is returned (if it exists) such that the running sum is greater than or equal to the required amount.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -132,14 +132,34 @@ pub(crate) fn get_spendable_sapling_notes(
     anchor_height: BlockHeight,
```
daira, line 132: This should be taking a transaction since it needs to be atomic.

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -0,0 +1,75 @@
...
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        transaction.execute_batch(
+            &format!(
+                "CREATE VIEW v_sapling_shard_unscanned_ranges AS
+                SELECT
+                    shard.shard_index,
+                    shard.shard_index << 16 AS start_position,
+                    (shard.shard_index + 1) << 16 AS end_position_exclusive,
```
daira, line 45: This is hard-coding the shard depth; we use a constant elsewhere. We are already using a `format!` macro here.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -132,14 +132,34 @@ pub(crate) fn get_spendable_sapling_notes(
...
     exclude: &[ReceivedNoteId],
 ) -> Result<Vec<ReceivedSaplingNote<ReceivedNoteId>>, SqliteClientError> {
+    let mut stmt_unscanned_tip = conn.prepare_cached(
+        "SELECT 1 FROM v_sapling_shard_unscanned_ranges unscanned
+         WHERE unscanned.shard_index IN (SELECT MAX(shard_index) FROM sapling_tree_shards)
+         AND unscanned.subtree_start_height <= :anchor_height",
```
str4d, line 138: Bug: this will cause no notes to be selected in an edge case where witnesses could have been built.

Consider the most recent subtree that has start height `X`. The wallet has `update_chain_tip(Y)` called, and then the wallet is given blocks to scan in the range `X..=Y`. After that, `update_chain_tip(Z)` is called, but before any more scanning happens, the wallet tries to create a transaction with `X < anchor_height < Y` and `Z - anchor_height < 100`.

It _should_ be possible to generate witnesses in this case, because `Y < anchor_height` is false, but because this conditional is checking for `X <= anchor_height` it fails. I note that the comment below is written assuming that `Y < anchor_height` is being checked.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -132,14 +132,34 @@ pub(crate) fn get_spendable_sapling_notes(
...
+         AND commitment_tree_position IS NOT NULL
          AND spent IS NULL
          AND transactions.block <= :anchor_height
-         AND id_note NOT IN rarray(:exclude)",
+         AND id_note NOT IN rarray(:exclude)
+         AND NOT EXISTS (
+            SELECT 1 FROM v_sapling_shard_unscanned_ranges unscanned
+            WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
+            AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
+            AND unscanned.block_range_start <= :anchor_height
```
daira, line 161:
```suggestion
```
This should be redundant because `AND transactions.block <= :anchor_height` already excludes notes received after the anchor height.

If there are any unscanned ranges between the position of the note and the anchor position, then we won't be able to build a witness for that note. So we need some other check that is checking whether the range from commitment_tree_position to the anchor position overlaps any unscanned range.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: I think that this can be fixed by changing `unscanned.subtree_start_height` to `unscanned.block_range_start`.

zcash_client_sqlite/src/wallet/sapling.rs
str4d, line 161: I think the correct condition here is:
```suggestion
            WHERE (
                sapling_received_notes.commitment_tree_position >= unscanned.start_position
                AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
            ) OR (
                sapling_received_notes.commitment_tree_position < unscanned.start_position
                AND unscanned.block_range_start <= :anchor_height
            )
```

nuttycom: @daira this is incorrect; we may have complete unscanned subtrees but we can still create a witness because we have the note commitment tree roots. We only need the subtree containing the note and the subtree containing the anchor to be fully scanned.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -164,10 +184,21 @@ pub(crate) fn select_spendable_sapling_notes(
     anchor_height: BlockHeight,
     exclude: &[ReceivedNoteId],
 ) -> Result<Vec<ReceivedSaplingNote<ReceivedNoteId>>, SqliteClientError> {
+    let mut stmt_unscanned_tip = conn.prepare_cached(
+        "SELECT 1 FROM v_sapling_shard_unscanned_ranges unscanned
+         WHERE unscanned.shard_index IN (SELECT MAX(shard_index) FROM sapling_tree_shards)
+         AND unscanned.subtree_start_height <= :anchor_height",
```
daira, line 190: Same issue as https://github.com/zcash/librustzcash/pull/894/files#r1294004615

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -188,11 +219,19 @@ pub(crate) fn select_spendable_sapling_notes(
...
              WHERE account = :account
+             AND commitment_tree_position IS NOT NULL
              AND spent IS NULL
              AND transactions.block <= :anchor_height
              AND id_note NOT IN rarray(:exclude)
+             AND NOT EXISTS (
+                SELECT 1 FROM v_sapling_shard_unscanned_ranges unscanned
+                WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
+                AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
+                AND unscanned.block_range_start <= :anchor_height
```
daira, line 233: Same issue as above. Actually can we factor out the duplicated code?

zcash_client_sqlite/src/wallet/sapling.rs
daira: Yes you came to the same conclusion as we did, which is reassuring.

daira (changes requested):
Reviewed by @str4d and I with comments.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom:
```
sapling_received_notes.commitment_tree_position < unscanned.start_position
```
This seems to select too many ranges; it can include unscanned subtrees for which we have the root, and therefore we don't need the range to be scanned.

str4d: Ah, I was misreading the new view in that case. That means AFAICT we cannot filter out the second conditional at all, and instead have to manually check whether we can obtain witnesses after the notes have been selected.

It remains the case that the first conditional above is valid, i.e. `AND transactions.block <= :anchor_height` at the note level implies that `AND unscanned.block_range_start <= :anchor_height` is redundant at the `NOT EXISTS` conditional.

nuttycom: In order for a note to be spendable, two subtrees have to be fully scanned: the subtree containing the note, and the portion of the subtree at the chain tip up to and including the anchor height. The latter is the same for all notes, so we ensure that it has been scanned using the query on lines 136..139; if that range hasn't been scanned, then no notes are spendable and we can exit early.

The new view has one entry per scan range (not one entry per shard) and so we want the `EXISTS` subquery to find any unscanned range that overlaps with the range of blocks associated with the subtree containing the note being selected. The existing implementation, combined with the check related to 136..139 gives us exactly the notes where we have enough information to spend them.

Given this, I don't see why you say we would need to manually check whether we can construct the witnesses after retrieval?

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: This does not have the same issue; this has the correct check.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: `unscanned.block_range_start` turns out to be not quite correct; the problem is that `MAX(shard_index)` doesn't take into account the anchor height; the anchor could exist in the previous block.

nuttycom:
This is now based on #896 to avoid conflicts.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: Comments have been added to the query to more clearly describe what's going on here.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: Hmmm... that presents a bit of a problem, because it means that anything accessing this needs at least &mut reference to the wallet database in order to construct the transaction, which is a more far-reaching change. I think it's okay for this to be non-atomic, because if the initial filter (on checking whether the subtree containing the anchor height is complete to the anchor height) passes, then the only way that subsequent results could be inconsistent with that would require a rollback, and the error that this would be preventing (a transaction constructed using the results of this query is rejected by the network because it attempts a double-spend) exists regardless.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/0f69b71d6033dac6adb78f78f21961be59f6a97b..b25aaf5533324a4878cf15636652494cda2df0cc) to address review comments & fix formatting & linting issues.

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -0,0 +1,77 @@
...
+                    shard.contains_marked,
+                    scan_queue.block_range_start,
+                    scan_queue.block_range_end,
+                    scan_queue.priority
+                FROM sapling_tree_shards shard
+                LEFT OUTER JOIN sapling_tree_shards prev_shard
+                    ON shard.shard_index = prev_shard.shard_index + 1
+                INNER JOIN scan_queue ON 
+                    (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    (scan_queue.block_range_end BETWEEN subtree_start_height AND shard.subtree_end_height) OR
```
str4d, line 57:
```suggestion
                    (scan_queue.block_range_end - 1 BETWEEN subtree_start_height AND shard.subtree_end_height) OR
```

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -0,0 +1,77 @@
...
+                    (
+                        scan_queue.block_range_start <= prev_shard.subtree_end_height
+                        AND scan_queue.block_range_end >= shard.subtree_end_height
```
str4d, line 60:
```suggestion
                        AND scan_queue.block_range_end > shard.subtree_end_height
```

str4d (approved):
utACK 1f8108e8dabb95b218952eb33f31d38a1441de94

daira (commented):
Post-hoc utACK

Issue #860, opened 2023-06-19
zcash_primitives 0.13-rc1, zcash_client_backend 0.10-rc1 and zcash_client_sqlite 0.8-rc1


str4d:

str4d:
Missing changelog entries: https://github.com/zcash/librustzcash/pull/857#pullrequestreview-1486261868

str4d:
Also, I missed in my review of #857 that `WalletDb::get_update_ops` was removed without mention in the changelog; I immediately ran into this while updating `zec-sqlite-cli`.

str4d:
https://github.com/zcash/librustzcash/pull/869#discussion_r1254953856

str4d:
https://github.com/zcash/librustzcash/pull/878#discussion_r1271574978

PR #896, opened 2023-08-11
zcash_client_sqlite: Ensure that target and anchor heights are relative to the chain tip.

nuttycom:
Prior to the scan-before-sync changes, the wallet was able to assume that the maximum scanned block height at the time of the spend was within a few blocks of the chain tip. However, under linear scanning after the spend-before-sync changes this invariant no longer holds, resulting in a situation where in linear sync conditions the wallet could attempt to create transactions with already-past expiry heights.

This change separates the notion of "chain tip" from "max scanned height", relying upon the `scan_queue` table to maintain the wallet's view of the consensus chain height and using information from the `blocks` table only in situations where the latest and/or earliest scanned height is required.

As part of this change, the `WalletRead` interface is also modified to disambiguate these concepts.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -17,7 +17,7 @@ and this library adheres to Rust's notion of
   - `ScannedBlock`
   - `ShieldedProtocol`
   - `WalletCommitmentTrees`
-  - `WalletRead::{block_metadata, block_fully_scanned, suggest_scan_ranges}`
+  - `WalletRead::{chain_height, block_metadata, block_fully_scanned, suggest_scan_ranges}`
   - `WalletWrite::{put_blocks, update_chain_tip}`
   - `chain::CommitmentTreeRoot`
   - `scanning` A new module containing types required for `suggest_scan_ranges`
@@ -89,10 +89,13 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::data_api`:
-  - `WalletRead::get_all_nullifiers`
-  - `WalletRead::{get_commitment_tree, get_witnesses}` have been removed 
-    without replacement. The utility of these methods is now subsumed
-    by those available from the `WalletCommitmentTrees` trait.
+  - `WalletRead::block_height_extrema` has been removed. Use `chain_height`
+    instead to obtain the wallet's view of the chain tip instead, or
+    `suggest_scan_ranges` to obtain information about blocks that need to be
+    scanned.
+  - `WalletRead::{get_all_nullifiers, get_commitment_tree, get_witnesses}` have
+    been removed without replacement. The utility of these methods is now
+    subsumed by those available from the `WalletCommitmentTrees` trait.
   - `WalletWrite::advance_by_block` (use `WalletWrite::put_blocks` instead).
   - `PrunedBlock` has been replaced by `ScannedBlock`
   - `testing::MockWalletDb`, which is available under the `test-dependencies`
@@ -107,6 +110,9 @@ and this library adheres to Rust's notion of
   have been removed as individual incremental witnesses are no longer tracked on a
   per-note basis. The global note commitment tree for the wallet should be used
   to obtain witnesses for spend operations instead.
+- Default implementations of `zcash_client_backend::data_api::WalletRead::{
+    get_target_and_anchor_heights, get_max_height_hash
+  }` have been removed. These should be implemented in a backend-specific fashion.
 
 
 ## [0.9.0] - 2023-04-28
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1,6 +1,5 @@
 //! Interfaces for wallet data persistence & low-level wallet utilities.
 
-use std::cmp;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::num::NonZeroU32;
@@ -60,10 +59,11 @@ pub trait WalletRead {
     /// or a UUID.
     type NoteRef: Copy + Debug + Eq + Ord;
 
-    /// Returns the minimum and maximum block heights for stored blocks.
+    /// Returns the height of the chain as known to the wallet as of the most recent call to
+    /// [`WalletWrite::update_chain_tip`].
     ///
-    /// This will return `Ok(None)` if no block data is present in the database.
-    fn block_height_extrema(&self) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error>;
+    /// This will return `Ok(None)` if the height of the current consensus chain tip is unknown.
+    fn chain_height(&self) -> Result<Option<BlockHeight>, Self::Error>;
 
     /// Returns the available block metadata for the block at the specified height, if any.
     fn block_metadata(&self, height: BlockHeight) -> Result<Option<BlockMetadata>, Self::Error>;
@@ -98,22 +98,7 @@ pub trait WalletRead {
     fn get_target_and_anchor_heights(
         &self,
         min_confirmations: NonZeroU32,
-    ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error> {
-        self.block_height_extrema().map(|heights| {
-            heights.map(|(min_height, max_height)| {
-                let target_height = max_height + 1;
-
-                // Select an anchor min_confirmations back from the target block,
-                // unless that would be before the earliest block we have.
-                let anchor_height = BlockHeight::from(cmp::max(
-                    u32::from(target_height).saturating_sub(min_confirmations.into()),
-                    u32::from(min_height),
-                ));
-
-                (target_height, anchor_height)
-            })
-        })
-    }
+    ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error>;
 
     /// Returns the minimum block height corresponding to an unspent note in the wallet.
     fn get_min_unspent_height(&self) -> Result<Option<BlockHeight>, Self::Error>;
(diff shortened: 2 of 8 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -98,22 +98,7 @@ pub trait WalletRead {
...
-                let anchor_height = BlockHeight::from(cmp::max(
-                    u32::from(target_height).saturating_sub(min_confirmations.into()),
-                    u32::from(min_height),
-                ));
-
-                (target_height, anchor_height)
-            })
-        })
-    }
+    ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error>;
```
str4d, line 101: `WalletRead::get_target_and_anchor_heights` no longer being a provided trait method is a change that needs to be documented in the changelog.

zcash_client_backend/CHANGELOG.md
```diff
@@ -128,10 +128,14 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::data_api`:
-  - `WalletRead::get_all_nullifiers`
-  - `WalletRead::{get_commitment_tree, get_witnesses}` have been removed
-    without replacement. The utility of these methods is now subsumed
-    by those available from the `WalletCommitmentTrees` trait.
+  - `WalletRead::block_height_extrema` has been removed. Use `chain_height`
```
str4d, line 131: Why does this method need to be removed? It is still a valid query to ask "what are the minimum and maximum heights for blocks we have seen?" even if there is a separate new method for the wallet's view of the chain tip. The changelog entry seems to imply that this data now needs to be fetched directly from the wallet DB's SQLite tables, but we've been generally moving in the direction of having callers access data via Rust methods instead.

If this method were left in place, then `WalletRead::get_target_and_anchor_heights` could remain a provided trait method (because the minimum height would be accessible).

zcash_client_backend/src/data_api.rs
```diff
@@ -123,22 +108,10 @@ pub trait WalletRead {
     /// is not found in the database.
     fn get_block_hash(&self, block_height: BlockHeight) -> Result<Option<BlockHash>, Self::Error>;
 
-    /// Returns the block hash for the block at the maximum height known
-    /// in stored data.
+    /// Returns the block height and hash for the block at the maximum scanned block height.
     ///
-    /// This will return `Ok(None)` if no block data is present in the database.
-    fn get_max_height_hash(&self) -> Result<Option<(BlockHeight, BlockHash)>, Self::Error> {
```
str4d, line 130: Ditto: either this no longer being a provided method needs to be documented in the changelog, or it can be reverted to a provided method if `block_height_extrema` is kept.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -771,11 +737,7 @@ pub(crate) mod tests {
                 None,
                 None
             ),
-            Err(data_api::error::Error::InsufficientFunds {
-                available,
-                required
-            })
-            if available == Amount::zero() && required == Amount::from_u64(10001).unwrap()
+            Err(data_api::error::Error::ScanRequired)
```
str4d, line 740: This change suggests that we currently have a bug in `init_blocks_table` (either in the code or documentation).

Before SbS the `blocks` table was the authoritative source of truth for the wallet's view of the chain tip, so by definition `init_blocks_table` was setting the wallet's initial view of the chain tip, and `InsufficientFunds` is the correct error.

With SbS, we now have two scenarios:
- We are doing linear scanning, in which case the `blocks` table's maximum block height will be equal to the highest scanned height in the scan queue, so either can be used.
- We are doing SbS, in which case we expect the scan queue to have heights greater than the `blocks` table as a result of calling `update_chain_tip`, and to be the authoritative source of truth for the wallet's view of the chain tip.

In the SbS case, `ScanRequired` makes sense as an error that can occur when there is a gap between the `blocks` table and the scan queue (or equivalently, when there are non-`Scanned` ranges in the scan queue above the maximum height in the `blocks` table). But if all ranges are scanned, then we should see `InsufficientFunds`, and in the test above where we never call `update_chain_tip`, the _only_ source of information is the `blocks` table, and what it was initialised to; hence in the wallet's view it is fully-synchronised, and should return `InsufficientFunds`.

The question is what to do about this. I see two mutually-exclusive pathways:
- Alter `init_blocks_table` to ensure we preserve its current semantics: that it fully-initialises a new wallet's view of the chain tip to be at the given block height. Existing wallets continue to call it without changes, and then those wallets that additionally use SbS will naturally call `update_chain_tip` to move the view forward.
- Keep the implementation of `init_blocks_table` as-is, and change its documentation to indicate that it is now an error to _not_ call `update_chain_tip` afterwards. This should be coupled with pervasive error condition checking to ensure that wallets don't miss the fact that they need to change their code. That could reuse the `ScanRequired` error, but it feels to me like this should be a separate `NotFullyInitialised` error or similar (because the fix is _not_ to call `suggest_scan_ranges` and scan them - that won't fix the problem - but to call `update_chain_tip`).

str4d (changes requested):
I only reviewed the new commit in this PR (f7080a54497c01c8b9c127ae643be771068a8622).

zcash_client_backend/CHANGELOG.md
nuttycom: My rationale for removing it is that while it's valid to ask what the maximum and minimum heights are that the wallet has seen, it's not really very meaningful because of all the gaps that may exist, and the minimum block height no longer corresponds to the wallet birthday; instead, it corresponds to the start of the shard containing the wallet birthday height. If we want to provide a single method, it should probably return a struct with `min_block_scanned`, `max_block_scanned`, `chain_tip`, and maybe `wallet_birthday` fields, and it should be named something different in order to ensure users switch away from the old method, which has muddled semantics.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: > We are doing linear scanning, in which case the blocks table's maximum block height will be equal to the highest scanned height in the scan queue, so either can be used.

This is incorrect. When we call `update_chain_tip` we will create a new scan entry with the range `previous_max_scanned+1..new_chain_tip` and these two values will not be in agreement. Linear scanning is still implemented in terms of `suggest_scan_ranges`. 

I don't like the inconsistency of having `init_blocks_table` add metadata for a block, when we haven't scanned that block, while in all other cases we add block metadata _after_ having scanned the block. Instead, one possibility is that `init_blocks_table` (which should probably be renamed) does two things: first, it adds the note commitment tree frontier that it is provided to the note commitment tree, making it possible to scan from that point forward, and then secondly it should create the 1-block scan range that starts with that block. 

This would also fix an issue that the wallet developers have been noticing that affects testnet wallets, which is that since there are so few Sapling notes on testnet, the block range that must be scanned in order to fill out the current-tip shard is enormous. What we should do is explicitly add a method to set the wallet birthday that takes the birthday height and the note commitment tree frontier as of that height; we should save the wallet birthday explicitly (perhaps in a different table) and then ensure that the view of unscanned ranges that is created in #894 uses this birthday height as the lower bound for the scan ranges, rather than Sapling activation.

str4d: > This is incorrect. When we call `update_chain_tip` we will create a new scan entry with the range `previous_max_scanned+1..new_chain_tip` and these two values will not be in agreement. Linear scanning is still implemented in terms of `suggest_scan_ranges`.

Hmm, I did not pick this up across the multiple PRs that implemented SbS. My understanding from our previous discussions was that they were being implemented such that an existing wallet doing linear scanning could keep doing what they were before in terms of their scanning life cycle (after updating for the breaking changes to the APIs they were using, per the changelog), and it would continue to work.

The changelog entries for this release definitely need to explain this semantic change (and the fact that pure linear scanning has been removed, replaced by "linear scanning SbS"). If I were to follow the instructions of the changelog entries as currently written, but go no further to implement SbS, my wallet would be broken. We also need a test for the error(s) that these wallets would see if they did not change their scanning life cycle to be "linear scanning SbS".

> I don't like the inconsistency of having `init_blocks_table` add metadata for a block, when we haven't scanned that block, while in all other cases we add block metadata _after_ having scanned the block.

It wasn't inconsistent with my view of how SbS was being integrated (per above). The definition of `init_blocks_table` is that the block you are initializing with _is_ fully scanned, because it has nothing of interest to the wallet (there are no notes in that block or any prior one, and you're providing a Sapling frontier that means this and prior blocks do not need to be scanned for `ShardTree` purposes). So it forms a valid fully-scanned start point for both old linear scanning, and new SbS.

I'm glad we've uncovered this discrepancy in how you and I were viewing the SbS changes; I had previously said that we needed to be very careful to look out for breaking changes to the assumptions inherent in the `blocks` table, and this is another such assumption-break. I'm worried how much extra work this might be to resolve though, compared to preserving the old linear scanning workflow (though maybe this far into the SbS PRs it's now more work to do the latter, IDK). Let's chat about this offline.

nuttycom: I don't see a use case for "update to this new version of the SDK, which has major breaking changes that I need to address in a bunch of places, but don't update to the new scan logic." If someone wants to keep with linear scanning, they just don't have to update.

str4d: The problem is that the fact that the crate releases completely breaks the old scanning logic _isn't documented anywhere_. So users can rightfully believe that they have done everything they need to, based on the current changelog (and even after the missing changelog entries from previous PRs are added as tracked by #860), their code would compile, and would likely work for some test scenarios - and then they deploy a broken wallet update.

If the old scanning logic is going to remain broken, then (in addition to the various code and test changes we still need to make due to breaking more assumptions about the `blocks` table) the crate releases are blocked on documenting this in a way that people upgrading can easily identify and follow.

str4d: In an offline meeting, @nuttycom and I decided that what we should do is replace `init_blocks_table` with a dedicated initialization method that handles the `blocks` table and scan queue, and sets up any relevant other state. Kris thinks it makes sense for this to also be tied to account initialization, but I'm not yet convinced and am waiting on the PR to review 🙂

nuttycom: This issue is now tracked as #901

str4d (approved):
utACK dee43854cb69fe428d4779eda7202d7744de0fb5

Issue #901, opened 2023-08-17
zcash_client_backend: Add an "account birthday" parameter to the `create_account` API.


str4d:
As part of this implementation, the `init_blocks_table` and `init_accounts_table` APIs should be removed in favor of initialization using the normal methods of the data access API, `add_account` and `update_chain_tip`. By performing these removals, we can then establish the invariant that the `blocks` table only contains entries for blocks that the wallet has scanned, which will be a useful property.

str4d:
This issue will need to result in an additional error state indicating that the caller needs to first call `update_chain_tip` so that newly-created accounts can have birthdays assigned to them (as opposed to `ScanRequired` which indicates the caller needs to call `suggest_scan_ranges`).

nuttycom:
Fixed by #907

PR #903, opened 2023-08-19
zcash_client_sqlite: Initialize the scan queue as part of `init_blocks_table`

### LINKED ISSUE #898: Send fails: Tree does not contain a root at address
In the Android SDK, which is operating upon the new SbS sync and the sync is in progress, we can reproduce `RuntimeException: Error while creating transaction: An error occurred in querying or updating a note commitment tree: Tree does not contain a root at address Address { level: Level(0), index: 127765 }`, which comes from the Rust layer of the SDK.

Full log:
```
TransactionEncoderImpl.createSpend(): creating transaction to spend Zatoshi(value=50) zatoshi to ****rgd9 with memo: test send
SaplingParamTool.validate$zcash_android_sdk_1_20_0_beta01_release(): Param files  both exist!
TransactionEncoderImpl.createSpend(): params exist! attempting to send...
TransactionEncoderImpl.createSpend(): Caught exception while creating transaction.
java.lang.RuntimeException: Error while creating transaction: An error occurred in querying or updating a note commitment tree: Tree does not contain a root at address Address { level: Level(0), index: 127765 }
	at cash.z.ecc.android.sdk.internal.jni.RustBackend.createToAddress(Native Method)
	at cash.z.ecc.android.sdk.internal.jni.RustBackend.access$createToAddress(RustBackend.kt:19)
	at cash.z.ecc.android.sdk.internal.jni.RustBackend$Companion.createToAddress(Unknown Source:0)
	at cash.z.ecc.android.sdk.internal.jni.RustBackend$Companion.access$createToAddress(RustBackend.kt:369)
	at cash.z.ecc.android.sdk.internal.jni.RustBackend$createToAddress$2.invokeSuspend(RustBackend.kt:303)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
```

Here are a few more hints:
- reproducible on the Testnet
- with SbS sync in progress
- old/new wallet makes no difference

### LINKED ISSUE #902: Insert a "Scanned" range prior to the wallet birthday
A beta tester of the iOS wallet reported that a brand new wallet appeared to attempt to scan from Sapling activation. This is likely because a "Scanned" range isn't being inserted for `SaplingActivation..=WalletBirthday`. We should insert this.

nuttycom:
This is a partial solution to #902 and #898. The complete fix, which avoids the need to fully scan the subtree containing the wallet's birthday height, will be achievable once #901 is complete. 

Fixes #902
Fixes #898

### DIFF
--- a/zcash_client_backend/src/data_api/scanning.rs
+++ b/zcash_client_backend/src/data_api/scanning.rs
@@ -6,7 +6,9 @@ use zcash_primitives::consensus::BlockHeight;
 /// Scanning range priority levels.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum ScanPriority {
-    /// Block ranges that have already been scanned have lowest priority.
+    /// Block ranges that are ignored have lowest priority.
+    Ignored,
+    /// Block ranges that have already been scanned will not be re-scanned.
     Scanned,
     /// Block ranges to be scanned to advance the fully-scanned height.
     Historic,
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -4,7 +4,7 @@ use incrementalmerkletree::Retention;
 use std::{collections::HashMap, fmt};
 use tracing::debug;
 
-use rusqlite::{self, types::ToSql};
+use rusqlite::{self, named_params};
 use schemer::{Migrator, MigratorError};
 use schemer_rusqlite::RusqliteAdapter;
 use secrecy::SecretVec;
@@ -13,18 +13,27 @@ use uuid::Uuid;
 
 use zcash_primitives::{
     block::BlockHash,
-    consensus::{self, BlockHeight},
+    consensus::{self, BlockHeight, NetworkUpgrade},
     merkle_tree::read_commitment_tree,
     sapling,
     transaction::components::amount::BalanceError,
     zip32::AccountId,
 };
 
-use zcash_client_backend::{data_api::SAPLING_SHARD_HEIGHT, keys::UnifiedFullViewingKey};
+use zcash_client_backend::{
+    data_api::{
+        scanning::{ScanPriority, ScanRange},
+        SAPLING_SHARD_HEIGHT,
+    },
+    keys::UnifiedFullViewingKey,
+};
 
 use crate::{error::SqliteClientError, wallet, WalletDb, PRUNING_DEPTH, SAPLING_TABLES_PREFIX};
 
-use super::commitment_tree::{self, SqliteShardStore};
+use super::{
+    commitment_tree::{self, SqliteShardStore},
+    scanning::insert_queue_entries,
+};
 
 mod migrations;
 
@@ -309,14 +318,27 @@ pub fn init_blocks_table<P: consensus::Parameters>(
 
         wdb.conn.0.execute(
             "INSERT INTO blocks (height, hash, time, sapling_tree)
-            VALUES (?, ?, ?, ?)",
-            [
-                u32::from(height).to_sql()?,
-                hash.0.to_sql()?,
-                time.to_sql()?,
-                sapling_tree.to_sql()?,
+             VALUES (:height, :hash, :time, :sapling_tree)",
+            named_params![
+                ":height": u32::from(height),
+                ":hash": hash.0,
+                ":time": time,
+                ":sapling_tree": sapling_tree,
             ],
         )?;
+
+        if let Some(sapling_activation) = wdb.params.activation_height(NetworkUpgrade::Sapling) {
+            let scan_range_start = std::cmp::min(sapling_activation, height);
+            let scan_range_end = height + 1;
+            insert_queue_entries(
+                wdb.conn.0,
+                Some(ScanRange::from_parts(
+                    scan_range_start..scan_range_end,
+                    ScanPriority::Ignored,
+                ))
+                .iter(),
+            )?;
+        }
 
         if let Some(nonempty_frontier) = block_end_tree.to_frontier().value() {
             debug!("Inserting frontier into ShardTree: {:?}", nonempty_frontier);
@@ -582,7 +604,7 @@ mod tests {
                         scan_queue.block_range_start <= prev_shard.subtree_end_height
                         AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
                     )
-                WHERE scan_queue.priority != {}",
+                WHERE scan_queue.priority > {}",
                 u32::from(tests::network().activation_height(NetworkUpgrade::Sapling).unwrap()),
                 priority_code(&ScanPriority::Scanned),
             ),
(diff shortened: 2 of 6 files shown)


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -309,15 +315,27 @@ pub fn init_blocks_table<P: consensus::Parameters>(
...
+            wdb.conn.0.execute(
+                "INSERT INTO scan_queue (block_range_start, block_range_end, priority)
+                 VALUES (:block_range_start, :block_range_end, :priority)",
+                named_params![
+                   ":block_range_start": u32::from(std::cmp::min(sapling_activation, height)),
+                   ":block_range_end": u32::from(height + 1),
+                   ":priority": priority_code(&ScanPriority::Ignored)
+                ],
+            )?;
+        }
```
daira, line 337: What about existing databases? Doesn't this need to be a migration?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -309,15 +315,27 @@ pub fn init_blocks_table<P: consensus::Parameters>(
...
+            named_params![
+                ":height": u32::from(height),
+                ":hash": hash.0,
+                ":time": time,
+                ":sapling_tree": sapling_tree,
             ],
         )?;
 
+        if let Some(sapling_activation) = wdb.params.activation_height(NetworkUpgrade::Sapling) {
+            wdb.conn.0.execute(
```
daira, line 328: Consider adding a `debug!` here.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -309,15 +315,27 @@ pub fn init_blocks_table<P: consensus::Parameters>(
...
+                "INSERT INTO scan_queue (block_range_start, block_range_end, priority)
+                 VALUES (:block_range_start, :block_range_end, :priority)",
+                named_params![
+                   ":block_range_start": u32::from(std::cmp::min(sapling_activation, height)),
```
daira, line 332:
```suggestion
                   ":block_range_start": u32::from(start_height),
```
(if the above suggestion is taken).

zcash_client_backend/src/data_api/scanning.rs
```diff
@@ -6,7 +6,9 @@ use zcash_primitives::consensus::BlockHeight;
 /// Scanning range priority levels.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum ScanPriority {
-    /// Block ranges that have already been scanned have lowest priority.
+    /// Block ranges that are ignored have lowest priority.
+    Ignored,
```
daira, line 10: There are a bunch of places where we use the Sapling activation height as the first block height that should potentially be scanned. If I understand correctly, this PR is marking the priority of the blocks between the Sapling activation height and the wallet's birthday as ignored, with the intent that they stay permanently ignored.

Wouldn't it be simpler to explicitly store the birthday and use that in place of the Sapling activation height? That would be a simpler data model with fewer states.

(I know that it is possible to have a rollback just after the wallet is initialised that effectively rolls back its birthday. If that ends up being too complex to handle in the approach I'm suggesting then I can see the rationale for doing it this way.)

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1212,4 +1220,102 @@ mod tests {
             ]
         );
     }
+
+    #[test]
+    fn init_blocks_table_creates_scanned_range() {
```
daira, line 1225:
```suggestion
    fn init_blocks_table_creates_ignored_range() {
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1212,4 +1220,102 @@ mod tests {
...
+            &mut db_data,
+            birthday_height,
+            BlockHash([1; 32]),
+            1,
+            &[0x0, 0x0, 0x0],
+        )
+        .unwrap();
+
+        let expected = vec![
+            // The range below the wallet's birthday height is ignored
```
daira, line 1252: This is not the range below the wallet's birthday height; it is the range up to and including the birthday height. Is that the intent?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1212,4 +1220,102 @@ mod tests {
...
+            scan_range(
+                u32::from(sap_active)..u32::from(birthday_height - 1000),
+                Ignored,
+            ),
+        ];
+
+        assert_matches!(
+            suggest_scan_ranges(&db_data.conn, Ignored),
+            Ok(scan_ranges) if scan_ranges == expected        );
```
daira, line 1319: Extraneous whitespace.

daira (dismissed):
utACK with comments.

zcash_client_backend/src/data_api/scanning.rs
daira: I see the #901 is about doing precisely that. But the issue is that if you do this change first and #901 later, then you've already complicated the data model by adding the "ignored" priority, and can't undo that.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Yes; the birthday height is the height of the current chain tip at the time that the wallet is created; by definition, the wallet can receive no transactions within that block.

zcash_client_backend/src/data_api/scanning.rs
nuttycom: I think that the notion of having an ignored range could actually be generally useful; one hypothetical use case is that someone who knows that their wallet has been offline for a long period of time (say, for example, a paper/cold wallet for which they never gave out any addresses) could explicitly inform the wallet that it should ignore the range since the last height that they interacted with it, and thereby avoid scanning costs. So even after #901 I think it's good to keep the `Ignored` case; I just hadn't really thought about it before or I might have included the behavior I'm implementing here in the first draft.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/8dc851f3d488a7a8ea93c2a63bb3c976945a87db..b3be0318c69d129599f7bf94845251cc5d58f9a3) to address comments from review.

zcash_client_sqlite/src/wallet/scanning.rs
daira:
```suggestion
            // The range up to and including the wallet's birthday height is ignored.
```

daira (dismissed):
utACK with minor non-blocking suggestion.

daira (approved):
utACK

PR #907, opened 2023-08-22
zcash_client_backend: Add account birthday management to the Data Access API

### LINKED ISSUE #897: Too wide suggested ranges for Testnet
Hi, when testing the new SbS sync algorithm with our mobile SDKs, we noticed that the Testnet server ([lightwalletd.testnet.electriccoin.co](http://lightwalletd.testnet.electriccoin.co/)) replies to `suggestScanRanges` besides others with also very old blocks for a new wallet: `ScanRange(range=BlockHeight(value=490074)..BlockHeight(value=2459999), priority=ChainTip)`. My guess is the `getSubtreeRoots`. Here is my getSubtreeRoot args:
```
downloader.getSubtreeRoots(
   startIndex = 0,
   maxEntries = if (network.isTestnet()) {
      65536
   } else {
      0
   },
   shieldedProtocol = ShieldedProtocolEnum.SAPLING
)
```
Here is a log for the SbS syncing:
```
CompactBlockProcessor.start$suspendImpl(): Fetched SubTreeRoot result: UseSbS(subTreeRootList=[SubtreeRoot(completingBlockHeight=490074)]), with preferred sync algorithm: SPEND_BEFORE_SYNC
CompactBlockProcessor.processNewBlocksInSbSOrder(): Beginning to process new blocks with Spend-before-Sync approach (with roots: [SubtreeRoot(completingBlockHeight=490074)], and lowerbound: BlockHeight(value=2459900))...
CompactBlockProcessor.putSaplingSubtreeRoots$zcash_android_sdk_1_20_0_beta01_release(): Sapling subtree roots put successfully with startIndex: 0 and roots: 1
CompactBlockProcessor.updateChainTip$zcash_android_sdk_1_20_0_beta01_release(): Chain tip updated successfully with height: BlockHeight(value=2469481)
CompactBlockProcessor.suggestScanRanges$zcash_android_sdk_1_20_0_beta01_release(): Successfully got newly suggested ranges: [ScanRange(range=BlockHeight(value=2460000)..BlockHeight(value=2460009), priority=Verify), ScanRange(range=BlockHeight(value=2460010)..BlockHeight(value=2469481), priority=ChainTip), ScanRange(range=BlockHeight(value=490074)..BlockHeight(value=2459999), priority=ChainTip)]
CompactBlockProcessor.processNewBlocksInSbSOrder(): Check for verification of ranges resulted with: ShouldVerify(scanRange=ScanRange(range=BlockHeight(value=2460000)..BlockHeight(value=2460009), priority=Verify))
CompactBlockProcessor.processNewBlocksInSbSOrder(): Starting verification of range: ShouldVerify(scanRange=ScanRange(range=BlockHeight(value=2460000)..BlockHeight(value=2460009), priority=Verify))
CompactBlockProcessor.invokeSuspend(): Syncing blocks in range BlockHeight(value=2460000)..BlockHeight(value=2460009)
```
And here is a log of also a new wallet, which operates upon the Linear sync algorithm:
```
CompactBlockProcessor.processNewBlocksInLinearOrder(): Beginning to process new blocks with Linear approach (with lower bound: BlockHeight(value=2459900))...
CompactBlockProcessor.invokeSuspend(): Syncing blocks in range BlockHeight(value=2459901)..BlockHeight(value=2469486)
```

You can see that the Linear one syncs newer blocks than the SbS one, which goes much deeper. This seems to be Testnet-only related.

nuttycom:
Closes #897.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -6,6 +6,17 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Notable Changes
+
+- `zcash_client_backend` now supports out-of-order scanning of blockchain history.
+- This release of `zcash_client_backend` defines the concept of an account
+  birthday. The account birthday is defined as the minimum height among blocks
+  to be scanned when recovering an account.
+- Account creation now requires the caller to provide account birthday information,
+  including the state of the note commitment tree at the end of the block prior to
+  the birthday height.
+
 ### Added
 - `impl Eq for zcash_client_backend::address::RecipientAddress`
 - `impl Eq for zcash_client_backend::zip321::{Payment, TransactionRequest}`
@@ -39,6 +50,7 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::data_api`:
   - `WalletRead::TxRef` has been removed in favor of consistently using `TxId` instead.
   - `WalletRead::get_transaction` now takes a `TxId` as its argument.
+  - `WalletRead::create_account` now takes an additional `birthday` argument.
   - `WalletWrite::{store_decrypted_tx, store_sent_tx}` now return `Result<(), Self::Error>`
     as the `WalletRead::TxRef` associated type has been removed. Use
     `WalletRead::get_transaction` with the transaction's `TxId` instead.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -29,7 +29,8 @@ and this library adheres to Rust's notion of
   - `ScannedBlock`
   - `ShieldedProtocol`
   - `WalletCommitmentTrees`
-  - `WalletRead::{chain_height, block_metadata, block_fully_scanned, suggest_scan_ranges}`
+  - `WalletRead::{chain_height, block_metadata, block_fully_scanned, suggest_scan_ranges,
+    get_wallet_birthday, get_account_birthday}`
   - `WalletWrite::{put_blocks, update_chain_tip}`
   - `chain::CommitmentTreeRoot`
   - `scanning` A new module containing types required for `suggest_scan_ranges`
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -119,6 +119,16 @@ pub trait WalletRead {
     /// transaction is not in the main chain.
     fn get_tx_height(&self, txid: TxId) -> Result<Option<BlockHeight>, Self::Error>;
 
+    /// Returns the birthday height for the wallet.
+    ///
+    /// This returns earliest birthday height among accounts maintained by this wallet, or
+    /// `Ok(None)` if the wallet has no initialized accounts.
+    fn get_wallet_birthday(&self) -> Result<Option<BlockHeight>, Self::Error>;
+
+    /// Returns the birthday height for the given account, or an error if the account is not known
+    /// to the wallet.
+    fn get_account_birthday(&self, account: AccountId) -> Result<BlockHeight, Self::Error>;
+
     /// Returns the most recently generated unified address for the specified account, if the
     /// account identifier specified refers to a valid account for this wallet.
     ///
@@ -799,6 +809,14 @@ pub mod testing {
 
         fn get_tx_height(&self, _txid: TxId) -> Result<Option<BlockHeight>, Self::Error> {
             Ok(None)
+        }
+
+        fn get_wallet_birthday(&self) -> Result<Option<BlockHeight>, Self::Error> {
+            Ok(None)
+        }
+
+        fn get_account_birthday(&self, _account: AccountId) -> Result<BlockHeight, Self::Error> {
+            Err(())
         }
 
         fn get_current_address(
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1,8 +1,9 @@
 //! Interfaces for wallet data persistence & low-level wallet utilities.
 
-use std::collections::HashMap;
 use std::fmt::Debug;
+use std::io;
 use std::num::NonZeroU32;
+use std::{collections::HashMap, num::TryFromIntError};
 
 use incrementalmerkletree::{frontier::Frontier, Retention};
 use secrecy::SecretVec;
@@ -24,6 +25,7 @@ use crate::{
     address::{AddressMetadata, UnifiedAddress},
     decrypt::DecryptedOutput,
     keys::{UnifiedFullViewingKey, UnifiedSpendingKey},
+    proto::service::TreeState,
     wallet::{ReceivedSaplingNote, WalletTransparentOutput, WalletTx},
 };
 
@@ -472,23 +474,75 @@ impl SentTransactionOutput {
 pub struct AccountBirthday {
     height: BlockHeight,
     sapling_frontier: Frontier<Node, NOTE_COMMITMENT_TREE_DEPTH>,
+    recover_until: Option<BlockHeight>,
+}
+
+/// Errors that can occur in the construction of an [`AccountBirthday`] from a [`TreeState`]
+pub enum BirthdayError {
+    HeightInvalid(TryFromIntError),
+    Decode(io::Error),
+}
+
+impl From<TryFromIntError> for BirthdayError {
+    fn from(value: TryFromIntError) -> Self {
+        Self::HeightInvalid(value)
+    }
+}
+
+impl From<io::Error> for BirthdayError {
+    fn from(value: io::Error) -> Self {
+        Self::Decode(value)
+    }
 }
 
 impl AccountBirthday {
     /// Constructs a new [`AccountBirthday`] from its constituent parts.
     ///
-    /// * `height`: The birthday height of the account. This is defined as the height of the last
-    /// block that is known to contain no transactions sent to addresses belonging to the account.
+    /// * `height`: The birthday height of the account. This is defined as the height of the first
+    ///    block block to be scanned in wallet recovery.
     /// * `sapling_frontier`: The Sapling note commitment tree frontier as of the end of the block
-    /// at `height`.
+    ///    prior to the birthday height.
+    /// * `recover_until`: An optional height at which the wallet should exit "recovery mode". In
+    ///    order to avoid confusing shifts in wallet balance and spendability that may temporarily be
+    ///    visible to a user during the process of recovering from seed, wallets may optionally set a
+    ///    "recover until" height. The wallet is considered to be in "recovery mode" until there
+    ///    exist no unscanned ranges between the wallet's birthday height and the provided
+    ///    `recover_until` height, exclusive.
+    ///
+    /// This API is intended primarily to be used in testing contexts; under normal circumstances,
+    /// [`AccountBirthday::from_treestate`] should be used instead.
+    #[cfg(feature = "test-dependencies")]
     pub fn from_parts(
         height: BlockHeight,
         sapling_frontier: Frontier<Node, NOTE_COMMITMENT_TREE_DEPTH>,
+        recover_until: Option<BlockHeight>,
     ) -> Self {
         Self {
             height,
             sapling_frontier,
+            recover_until,
         }
+    }
+
+    /// Constructs a new [`AccountBirthday`] from a [`TreeState`] returned from `lightwalletd`.
+    ///
+    /// * `treestate`: The tree state corresponding to the last block prior to the wallet's
+    ///    birthday height.
+    /// * `recover_until`: An optional height at which the wallet should exit "recovery mode". In
+    ///    order to avoid confusing shifts in wallet balance and spendability that may temporarily be
+    ///    visible to a user during the process of recovering from seed, wallets may optionally set a
+    ///    "recover until" height. The wallet is considered to be in "recovery mode" until there
+    ///    exist no unscanned ranges between the wallet's birthday height and the provided
+    ///    `recover_until` height, exclusive.
+    pub fn from_treestate(
+        treestate: TreeState,
+        recover_until: Option<BlockHeight>,
+    ) -> Result<Self, BirthdayError> {
+        Ok(Self {
+            height: BlockHeight::try_from(treestate.height + 1)?,
+            sapling_frontier: treestate.sapling_tree()?.to_frontier(),
+            recover_until,
+        })
     }
 
     /// Returns the Sapling note commitment tree frontier as of the end of the block at
@@ -501,6 +555,24 @@ impl AccountBirthday {
     pub fn height(&self) -> BlockHeight {
         self.height
     }
+
+    /// Returns the height at which the wallet should exit "recovery mode".
+    pub fn recover_until(&self) -> Option<BlockHeight> {
+        self.recover_until
+    }
+
+    #[cfg(feature = "test-dependencies")]
+    pub fn from_sapling_activation<P: zcash_primitives::consensus::Parameters>(
+        params: &P,
+    ) -> AccountBirthday {
+        use zcash_primitives::consensus::NetworkUpgrade;
+
+        AccountBirthday::from_parts(
+            params.activation_height(NetworkUpgrade::Sapling).unwrap(),
+            Frontier::empty(),
+            None,
+        )
+    }
 }
 
 /// This trait encapsulates the write capabilities required to update stored
@@ -509,24 +581,36 @@ pub trait WalletWrite: WalletRead {
     /// The type of identifiers used to look up transparent UTXOs.
     type UtxoRef;
 
-    /// Tells the wallet to track the next available account-level spend authority, given
-    /// the current set of [ZIP 316] account identifiers known to the wallet database.
+    /// Tells the wallet to track the next available account-level spend authority, given the
+    /// current set of [ZIP 316] account identifiers known to the wallet database.
     ///
-    /// Returns the account identifier for the newly-created wallet database entry, along
-    /// with the associated [`UnifiedSpendingKey`].
+    /// Returns the account identifier for the newly-created wallet database entry, along with the
+    /// associated [`UnifiedSpendingKey`].
     ///
-    /// If `seed` was imported from a backup and this method is being used to restore a
-    /// previous wallet state, you should use this method to add all of the desired
-    /// accounts before scanning the chain from the seed's birthday height.
+    /// If a birthday height is having a height is below the current chain tip, this operation will
+    /// trigger a re-scan of the blocks at and above the provided height. The birthday height is
+    /// defined as the minimum block height that will be scanned for funds belonging to the wallet.
     ///
-    /// By convention, wallets should only allow a new account to be generated after funds
-    /// have been received by the currently-available account (in order to enable
-    /// automated account recovery).
+    /// For new wallets, callers should construct the [`AccountBirthday`] using
+    /// [`AccountBirthday::from_treestate`] for the block at height `chain_tip_height - PRUNING_DEPTH`.
+    /// Setting the birthday height to a tree state below the pruning depth ensures that reorgs
+    /// cannot cause funds intended for the wallet to be missed; otherwise, if the chain tip height
+    /// were used for the wallet birthday, a transaction targeted at a height greater than the
+    /// chain tip could be mined at a height below that tip as part of a reorg.
+    ///
+    /// If `seed` was imported from a backup and this method is being used to restore a previous
+    /// wallet state, you should use this method to add all of the desired accounts before scanning
+    /// the chain from the seed's birthday height.
+    ///
+    /// By convention, wallets should only allow a new account to be generated after funds have
+    /// been received by the currently-available account (in order to enable automated account
+    /// recovery).
     ///
     /// [ZIP 316]: https://zips.z.cash/zip-0316
     fn create_account(
         &mut self,
         seed: &SecretVec<u8>,
+        birthday: AccountBirthday,
     ) -> Result<(AccountId, UnifiedSpendingKey), Self::Error>;
 
     /// Generates and persists the next available diversified address, given the current
@@ -645,9 +729,9 @@ pub mod testing {
     };
 
     use super::{
-        chain::CommitmentTreeRoot, scanning::ScanRange, BlockMetadata, DecryptedTransaction,
-        NoteId, NullifierQuery, ScannedBlock, SentTransaction, WalletCommitmentTrees, WalletRead,
-        WalletWrite, SAPLING_SHARD_HEIGHT,
+        chain::CommitmentTreeRoot, scanning::ScanRange, AccountBirthday, BlockMetadata,
+        DecryptedTransaction, NoteId, NullifierQuery, ScannedBlock, SentTransaction,
+        WalletCommitmentTrees, WalletRead, WalletWrite, SAPLING_SHARD_HEIGHT,
     };
 
     pub struct MockWalletDb {
@@ -818,6 +902,7 @@ pub mod testing {
         fn create_account(
             &mut self,
             seed: &SecretVec<u8>,
+            _birthday: AccountBirthday,
         ) -> Result<(AccountId, UnifiedSpendingKey), Self::Error> {
             let account = AccountId::from(0);
             UnifiedSpendingKey::from_seed(&self.network, seed.expose_secret(), account)
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -121,8 +121,8 @@ pub trait WalletRead {
 
     /// Returns the birthday height for the wallet.
     ///
-    /// This returns earliest birthday height among accounts maintained by this wallet, or
-    /// `Ok(None)` if the wallet has no initialized accounts.
+    /// This returns the earliest birthday height among accounts maintained by this wallet,
+    /// or `Ok(None)` if the wallet has no initialized accounts.
     fn get_wallet_birthday(&self) -> Result<Option<BlockHeight>, Self::Error>;
 
     /// Returns the birthday height for the given account, or an error if the account is not known
@@ -487,7 +487,7 @@ pub struct AccountBirthday {
     recover_until: Option<BlockHeight>,
 }
 
-/// Errors that can occur in the construction of an [`AccountBirthday`] from a [`TreeState`]
+/// Errors that can occur in the construction of an [`AccountBirthday`] from a [`TreeState`].
 pub enum BirthdayError {
     HeightInvalid(TryFromIntError),
     Decode(io::Error),
@@ -509,7 +509,7 @@ impl AccountBirthday {
     /// Constructs a new [`AccountBirthday`] from its constituent parts.
     ///
     /// * `height`: The birthday height of the account. This is defined as the height of the first
-    ///    block block to be scanned in wallet recovery.
+    ///    block to be scanned in wallet recovery.
     /// * `sapling_frontier`: The Sapling note commitment tree frontier as of the end of the block
     ///    prior to the birthday height.
     /// * `recover_until`: An optional height at which the wallet should exit "recovery mode". In
@@ -572,6 +572,12 @@ impl AccountBirthday {
     }
 
     #[cfg(feature = "test-dependencies")]
+    /// Constructs a new [`AccountBirthday`] at Sapling activation, with no
+    /// "recover until" height.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the Sapling activation height is not set.
     pub fn from_sapling_activation<P: zcash_primitives::consensus::Parameters>(
         params: &P,
     ) -> AccountBirthday {
@@ -597,12 +603,12 @@ pub trait WalletWrite: WalletRead {
     /// Returns the account identifier for the newly-created wallet database entry, along with the
     /// associated [`UnifiedSpendingKey`].
     ///
-    /// If a birthday height is having a height is below the current chain tip, this operation will
+    /// If `birthday.height()` is below the current chain tip, this operation will
     /// trigger a re-scan of the blocks at and above the provided height. The birthday height is
     /// defined as the minimum block height that will be scanned for funds belonging to the wallet.
     ///
     /// For new wallets, callers should construct the [`AccountBirthday`] using
-    /// [`AccountBirthday::from_treestate`] for the block at height `chain_tip_height - PRUNING_DEPTH`.
+    /// [`AccountBirthday::from_treestate`] for the block at height `chain_tip_height - 100`.
     /// Setting the birthday height to a tree state below the pruning depth ensures that reorgs
     /// cannot cause funds intended for the wallet to be missed; otherwise, if the chain tip height
     /// were used for the wallet birthday, a transaction targeted at a height greater than the
@@ -612,9 +618,9 @@ pub trait WalletWrite: WalletRead {
     /// wallet state, you should use this method to add all of the desired accounts before scanning
     /// the chain from the seed's birthday height.
     ///
-    /// By convention, wallets should only allow a new account to be generated after funds have
-    /// been received by the currently-available account (in order to enable automated account
-    /// recovery).
+    /// By convention, wallets should only allow a new account to be generated after confirmed
+    /// funds have been received by the currently-available account (in order to enable automated
+    /// account recovery).
     ///
     /// [ZIP 316]: https://zips.z.cash/zip-0316
     fn create_account(
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -28,6 +28,7 @@ and this library adheres to Rust's notion of
 - `zcash_client_sqlite::error::SqliteClientError` has new error variants:
   - `SqliteClientError::BlockConflict`
   - `SqliteClientError::CacheMiss`
+  - `SqliteClientError::ChainHeightUnknown`
 - `zcash_client_backend::FsBlockDbError` has a new error variant:
   - `FsBlockDbError::CacheMiss`
 - `zcash_client_sqlite::FsBlockDb::write_block_metadata` now overwrites any
@@ -38,6 +39,12 @@ and this library adheres to Rust's notion of
 - `zcash_client_sqlite::NoteId` has been replaced with `zcash_client_sqlite::ReceivedNoteId`
   as the `SentNoteId` variant of is now unused following changes to 
   `zcash_client_backend::data_api::WalletRead`.
+- `zcash_client_sqlite::wallet::init::{init_blocks_table, init_accounts_table}`
+  have been removed. `zcash_client_backend::data_api::WalletWrite::create_account`
+  should be used instead; the initialization of the note commitment tree
+  previously performed by `init_blocks_table` is now handled by passing an
+  `AccountBirthday` containing the note commitment tree frontier as of the
+  end of the birthday height block to `create_account` instead.
 
 ### Fixed
 - Fixed an off-by-one error in the `BlockSource` implementation for the SQLite-backed
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -26,6 +26,7 @@ and this library adheres to Rust's notion of
   wallet did not contain enough observed blocks to satisfy the `min_confirmations`
   value specified; this situation is now treated as an error.
 - `zcash_client_sqlite::error::SqliteClientError` has new error variants:
+  - `SqliteClientError::AccountUnknown`
   - `SqliteClientError::BlockConflict`
   - `SqliteClientError::CacheMiss`
   - `SqliteClientError::ChainHeightUnknown`
@@ -37,7 +38,7 @@ and this library adheres to Rust's notion of
 ### Removed
 - The empty `wallet::transact` module has been removed.
 - `zcash_client_sqlite::NoteId` has been replaced with `zcash_client_sqlite::ReceivedNoteId`
-  as the `SentNoteId` variant of is now unused following changes to 
+  as the `SentNoteId` variant of is now unused following changes to
   `zcash_client_backend::data_api::WalletRead`.
 - `zcash_client_sqlite::wallet::init::{init_blocks_table, init_accounts_table}`
   have been removed. `zcash_client_backend::data_api::WalletWrite::create_account`
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -38,7 +38,7 @@ and this library adheres to Rust's notion of
 ### Removed
 - The empty `wallet::transact` module has been removed.
 - `zcash_client_sqlite::NoteId` has been replaced with `zcash_client_sqlite::ReceivedNoteId`
-  as the `SentNoteId` variant of is now unused following changes to
+  as the `SentNoteId` variant is now unused following changes to
   `zcash_client_backend::data_api::WalletRead`.
 - `zcash_client_sqlite::wallet::init::{init_blocks_table, init_accounts_table}`
   have been removed. `zcash_client_backend::data_api::WalletWrite::create_account`
--- a/zcash_client_sqlite/src/chain.rs
+++ b/zcash_client_sqlite/src/chain.rs
@@ -324,8 +324,6 @@ where
 mod tests {
     use std::num::NonZeroU32;
 
-    use secrecy::Secret;
-
     use zcash_primitives::{
         block::BlockHash,
         transaction::{components::Amount, fees::zip317::FeeRule},
@@ -335,8 +333,8 @@ mod tests {
     use zcash_client_backend::{
         address::RecipientAddress,
         data_api::{
-            chain::error::Error, wallet::input_selection::GreedyInputSelector, WalletRead,
-            WalletWrite,
+            chain::error::Error, wallet::input_selection::GreedyInputSelector, AccountBirthday,
+            WalletRead,
         },
         fees::{zip317::SingleOutputChangeStrategy, DustOutputPolicy},
         scanning::ScanError,
@@ -345,7 +343,7 @@ mod tests {
     };
 
     use crate::{
-        testing::{birthday_at_sapling_activation, AddressType, TestBuilder},
+        testing::{AddressType, TestBuilder},
         wallet::{get_balance, truncate_to_height},
         AccountId,
     };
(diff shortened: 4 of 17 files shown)


zcash_client_sqlite/src/wallet/init/migrations/addresses_table.rs
```diff
@@ -152,13 +152,17 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                 }
             }
 
-            add_account_internal::<P, WalletMigrationError>(
-                transaction,
-                &self.params,
-                "accounts_new",
-                account,
-                &ufvk,
+            transaction.execute(
```
nuttycom, line 155: This change is necessary because `add_account_internal` is no longer able to correctly function at this point given the state of the database at this point in the migration graph. Instead, the previous functionality of `add_account_internal` is reimplemented here, and `add_account_internal` is removed. This is also something to note for future migrations; it's a good idea to not rely upon functions whose implementation might change because it would be possible to silently change the behavior of old migrations with very surprising results.

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -32,18 +29,24 @@ impl<P> schemer::Migration for Migration<P> {
     }
 }
 
-impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+impl RusqliteMigration for Migration {
     type Error = WalletMigrationError;
 
     fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        transaction.execute_batch(
```
nuttycom, line 36: This should probably be in a separate migration; there are nightly Android builds of the SDK that are depending upon the existing migration. Also, the account birthday change is really a semantically separate thing from adding the view.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,89 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
     key: &UnifiedFullViewingKey,
+    birthday: Option<AccountBirthday>,
 ) -> Result<(), SqliteClientError> {
-    add_account_internal(conn, params, "accounts", account, key)
-}
+    // Set the wallet birthday, falling back to the chain tip if not specified
+    let chain_tip = scan_queue_extrema(conn)?.map(|(h, _)| h);
+
+    let birthday_height = birthday.as_ref().map_or_else(
+        || chain_tip.ok_or(SqliteClientError::ChainHeightUnknown),
```
nuttycom, line 172: I'm not certain whether this should be an error or whether we should default to Sapling activation. Strictly speaking it's not an error to create an account in a state where the chain tip is unknown and no wallet birthday is specified.

HonzaR (commented):
From the Android SDK consumer view, it looks good. Thank you! The only question I have is that I see both `init_blocks_table` and `init_accounts_table` are replaced by `create_account`, which expects `AccountBirthday`, which is optional. Can you please describe what we should use in case of new or restored wallets for this parameter? Or should we leave it empty?

nuttycom:
> From the Android SDK consumer view, it looks good. Thank you! The only question I have is that I see both `init_blocks_table` and `init_accounts_table` are replaced by `create_account`, which expects `AccountBirthday`, which is optional. Can you please describe what we should use in case of new or restored wallets for this parameter? Or should we leave it empty?

There are two options:
(1) is to first call `updateChainTip` and then call this without specifying an `AccountBirthday`.
(2) call this with the arguments that you would have previously provided to `init_blocks_table` (block height and note commitment tree frontier at the chain tip).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -577,32 +365,30 @@ mod tests {
...
-                    (
-                        scan_queue.block_range_start <= prev_shard.subtree_end_height
-                        AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
-                    )
-                WHERE scan_queue.priority > {}",
-                u32::from(db_data.params.activation_height(NetworkUpgrade::Sapling).unwrap()),
-                priority_code(&ScanPriority::Scanned),
-            ),
+            "CREATE VIEW v_sapling_shard_unscanned_ranges AS
+            WITH wallet_birthday AS (SELECT MIN(birthday_height) AS height FROM accounts)
```
str4d, line 369: After the change to `AccountBirthday`, we are effectively changing the definition of the `wallet_birthday` column (to match its semantics for new accounts, i.e. last height we don't need to scan). Check that this works for the view, and adjust if necessary.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -577,32 +365,30 @@ mod tests {
...
+            SELECT
+                shard.shard_index,
+                shard.shard_index << 16 AS start_position,
+                (shard.shard_index + 1) << 16 AS end_position_exclusive,
+                IFNULL(prev_shard.subtree_end_height, wallet_birthday.height) AS subtree_start_height,
```
str4d, line 374: This should still be Sapling activation, because that is definitionally correct for the shard at index 0, whereas this change would cause the shard at index 0 to have a start height after its end height for almost all accounts. `wallet_birthday.height` should be filtering somewhere on `scan_queue.block_range_end` (or whichever is the correct bound).

str4d: Additionally, I think we should add continuity checks for places that cause the `sapling_tree_shards` table to be modified. If we end up in a situation where the table has discontinuities in `shard_index`, this line will create rows that have incorrect and overlapping shard tree ranges, which may cause subsequent logic (that assumes correct ranges) to fail in weird and annoying ways. Functionally, this means that anyone calling `update_subtree_roots` must do so in a way that doesn't add discontinuities (as adding a root causes the entire shard to be serialized if it didn't exist).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -577,32 +365,30 @@ mod tests {
...
+                ON shard.shard_index = prev_shard.shard_index + 1
+            INNER JOIN scan_queue ON 
+                (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                ((scan_queue.block_range_end - 1) BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                (
+                    scan_queue.block_range_start <= prev_shard.subtree_end_height
+                    AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
+                )
+            INNER JOIN wallet_birthday
+            WHERE scan_queue.priority > 10",
```
str4d, line 391: If you're putting the `format!` back for Sapling activation, revert to substituting in `ScanPriority::Scanned` here as well.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -577,32 +365,30 @@ mod tests {
...
-                    ON shard.shard_index = prev_shard.shard_index + 1
-                INNER JOIN scan_queue ON 
-                    (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
-                    ((scan_queue.block_range_end - 1) BETWEEN subtree_start_height AND shard.subtree_end_height) OR
-                    (
-                        scan_queue.block_range_start <= prev_shard.subtree_end_height
-                        AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
-                    )
-                WHERE scan_queue.priority > {}",
-                u32::from(db_data.params.activation_height(NetworkUpgrade::Sapling).unwrap()),
```
str4d: Note for future maintainers (hi us!) - hard-coding activation heights into the database is fine for settled network upgrades, but for ones in development (e.g. NU6 for us) this can lead to a problem if an activation height is picked, then crates published (and used) with that height, and then we discover a bug that requires us to change the consensus branch ID and activation height. In this case the hard-coded height would need to be fixed via a migration.

zcash_client_sqlite/src/wallet/init/migrations/add_account_birthdays.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        transaction.execute_batch(
+            "ALTER TABLE accounts ADD COLUMN birthday_height INTEGER;
+            UPDATE accounts SET birthday_height = MIN(blocks.height) FROM blocks;",
```
str4d, line 39: I'm happy with this migration, because `init_blocks_table` would insert a row corresponding to the last height that wasn't scanned for the wallet (matching the intended semantics of `birthday_height`), and as of this migration being written we haven't had any production wallets inserting blocks in any other way.

str4d (commented):
Flushing my review for the first two commits (up to 12ddc7d9f189eb0fc3c184ebbda3885df3b1b9fd).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,92 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+    }
+
+    let sapling_activation_height = params
+        .activation_height(NetworkUpgrade::Sapling)
+        .expect("Sapling activation height must be available.");
+
+    // Add the ignored range up to the birthday height. We always re-scan the birthday block, even
+    // though it should never contain funds for the wallet, just to avoid off-by-one errors that
+    // might occur if someone were to use the block height of their first received transaction as
+    // the birthday.
```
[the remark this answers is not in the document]
nuttycom: I think that this is fine, given that we're going to be suggesting that the birthday height be set to `chain_tip - PRUNING_DEPTH` to avoid potential errors related to reorg.

zcash_client_backend/src/data_api.rs
```diff
@@ -472,22 +482,31 @@ impl SentTransactionOutput {
...
+    /// * `recover_until`: An optional height at which the wallet should exit "recovery mode". In
+    ///    order to avoid confusing shifts in wallet balance and spendability that may temporarily be
+    ///    visible to a user during the process of recovering from seed, wallets may optionally set a
+    ///    "recover until" height. The wallet is considered to be in "recovery mode" until there
+    ///    exist no unscanned ranges between the wallet's birthday height and the provided
+    ///    `recover_until` height, exclusive.
     pub fn from_parts(
         height: BlockHeight,
         sapling_frontier: Frontier<Node, NOTE_COMMITMENT_TREE_DEPTH>,
+        recover_until: Option<BlockHeight>,
```
nuttycom, line 504: This will be used in #914 to provide a "recovery status" endpoint in addition to (or as part of) the "sync status" support.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -612,6 +679,40 @@ pub(crate) fn get_sent_memo(
...
+        },
+    )
+}
+
+pub(crate) fn account_birthday(
+    conn: &rusqlite::Connection,
+    account: AccountId,
+) -> Result<BlockHeight, SqliteClientError> {
+    conn.query_row(
+        "SELECT birthday_height AS wallet_birthday
```
nuttycom, line 705:
```suggestion
        "SELECT birthday_height
```

zcash_client_sqlite/src/wallet/init.rs
nuttycom: This hasn't been done yet, but since that's independent I think that it should be done in a separate ticket.

nuttycom: Added #924

zcash_client_sqlite/src/chain.rs
```diff
@@ -501,8 +502,13 @@ mod tests {
         let mut st = TestBuilder::new().with_block_cache().build();
 
         // Add an account to the wallet
+        let birthday =
+            AccountBirthday::from_parts(st.sapling_activation_height(), Frontier::empty());
```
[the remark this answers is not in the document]
nuttycom: Added #925

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Added #924

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -159,8 +168,10 @@ pub(crate) fn get_spendable_sapling_notes(
             -- select all the unscanned ranges involving the shard containing this note
             WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
             AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
-            -- exclude unscanned ranges above the anchor height which don't affect spendability
+            -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
             AND unscanned.block_range_start <= :anchor_height
+            -- exclude unscanned ranges that end below or on the wallet birthday
+            AND unscanned.block_range_end > :wallet_birthday
```
str4d, line 174: For the same reason as before:
```suggestion
            AND (unscanned.block_range_end - 1) > :wallet_birthday
```

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -233,8 +254,10 @@ pub(crate) fn select_spendable_sapling_notes(
                 -- select all the unscanned ranges involving the shard containing this note
                 WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
                 AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
-                -- exclude unscanned ranges above the anchor height which don't affect spendability
+                -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
                 AND unscanned.block_range_start <= :anchor_height
+                -- exclude unscanned ranges that end below or on the wallet birthday
+                AND unscanned.block_range_end > :wallet_birthday
```
str4d, line 260: Ditto:
```suggestion
                AND (unscanned.block_range_end - 1) > :wallet_birthday
```

str4d (changes requested):
Reviewed as of 497bceb217c861634da8e14a0fda06566383f325.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,90 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+            },
+        )?;
+    }
+
+    let sapling_activation_height = params
+        .activation_height(NetworkUpgrade::Sapling)
+        .expect("Sapling activation height must be available.");
+
+    // Add the ignored range up to and including the birthday height.
+    if sapling_activation_height < birthday_height {
```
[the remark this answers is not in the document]
nuttycom: A zero-width range doesn't make sense here; these ranges are end-exclusive, so `<` is correct.

zcash_client_sqlite/src/wallet/init/migrations/add_account_birthdays.rs
```diff
@@ -0,0 +1,75 @@
...
+                UPDATE accounts SET birthday_height = {} WHERE birthday_height IS NULL;
+
+                CREATE TABLE accounts_new (
+                    account INTEGER PRIMARY KEY,
+                    ufvk TEXT NOT NULL,
+                    birthday_height INTEGER NOT NULL,
+                    recover_until_height INTEGER
+                );
+                INSERT INTO accounts_new (account, ufvk, birthday_height)
+                SELECT account, ufvk, birthday_height FROM accounts;
```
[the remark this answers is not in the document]
nuttycom: It populates `accounts_new`

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -52,17 +53,19 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
                     (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
                     ((scan_queue.block_range_end - 1) BETWEEN subtree_start_height AND shard.subtree_end_height) OR
                     (
                         scan_queue.block_range_start <= prev_shard.subtree_end_height
                         AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
                     )
-                WHERE scan_queue.priority > {}",
+                INNER JOIN wallet_birthday
+                WHERE scan_queue.priority > {}
+                AND scan_queue.block_range_end > wallet_birthday.height;",
```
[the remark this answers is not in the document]
nuttycom: No, this is correct as it stands. Consider if we modified your suggestion to account for underflow:
```
AND scan_queue.block_range_end >= wallet_birthday.height + 1
```
This is clearly incorrect, because this would mean that a range that has scan_queue.block_range_end == wallet.birthday_height + 1 would be selected, meaning that the birthday height would be scanned.

zcash_client_backend/src/data_api.rs
```diff
@@ -509,24 +536,34 @@ pub trait WalletWrite: WalletRead {
...
-    /// Returns the account identifier for the newly-created wallet database entry, along
-    /// with the associated [`UnifiedSpendingKey`].
+    /// If a birthday height is provided and that height is below the current chain tip, this
+    /// operation will trigger a re-scan of the blocks at and above the provided height; otherwise,
+    /// the account's birthday will be set to the current chain tip. The birthday height is defined
+    /// as the maximum block height that is known to contain no funds for the wallet. Ordinarily,
+    /// callers should set this to `chain_tip_height - PRUNING_DEPTH` for new wallets to ensure
+    /// that chain reorgs cannot cause loss of funds as a consequence of a transaction targeted at
+    /// a height greater than the chain tip instead being mined at a height below that chain tip as
+    /// part of a reorg.
```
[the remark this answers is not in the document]
nuttycom: Hm, this is tricky because the `chain_tip_height - PRUNING_DEPTH` might be below Sapling activation. I guess that's okay, it just makes things weird.

zcash_client_backend/src/data_api.rs
```diff
@@ -501,6 +510,24 @@ impl AccountBirthday {
...
+    }
+
+    #[cfg(feature = "test-dependencies")]
+    pub fn from_sapling_activation<P: zcash_primitives::consensus::Parameters>(
+        params: &P,
+    ) -> AccountBirthday {
+        use zcash_primitives::consensus::NetworkUpgrade;
+
+        AccountBirthday::from_parts(
+            params.activation_height(NetworkUpgrade::Sapling).unwrap() - 1,
```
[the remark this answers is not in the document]
nuttycom: Since it's test-only code, I think this is fine.

zcash_client_backend/src/data_api.rs
```diff
@@ -509,24 +536,34 @@ pub trait WalletWrite: WalletRead {
...
-    /// the current set of [ZIP 316] account identifiers known to the wallet database.
+    /// Tells the wallet to track the next available account-level spend authority, given the
+    /// current set of [ZIP 316] account identifiers known to the wallet database.
+    ///
+    /// Returns the account identifier for the newly-created wallet database entry, along with the
+    /// associated [`UnifiedSpendingKey`].
     ///
-    /// Returns the account identifier for the newly-created wallet database entry, along
-    /// with the associated [`UnifiedSpendingKey`].
+    /// If a birthday height is provided and that height is below the current chain tip, this
```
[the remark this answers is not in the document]
nuttycom: I agree; a separate `recover_wallet` is probably a good idea, as it can correctly set scan ranges. However, since support for more than a single account is technically out of scope for the v2.0.0 SDKs, I don't think we should worry about it right now.

zcash_client_sqlite/src/wallet/init/migrations/add_account_birthdays.rs
```diff
@@ -0,0 +1,75 @@
...
+                UPDATE accounts SET birthday_height = MIN(blocks.height) FROM blocks;
+                -- if the blocks table is empty, set the birthday height to Sapling activation - 1
+                UPDATE accounts SET birthday_height = {} WHERE birthday_height IS NULL;
+
+                CREATE TABLE accounts_new (
+                    account INTEGER PRIMARY KEY,
+                    ufvk TEXT NOT NULL,
+                    birthday_height INTEGER NOT NULL,
+                    recover_until_height INTEGER
+                );
```
nuttycom, line 51:
```suggestion
                );

```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1200,42 +1324,46 @@ mod tests {
...
+
+        let expected = vec![
+            // The range up to the wallet's birthday height is ignored.
+            scan_range(
+                u32::from(sap_active)..u32::from(birthday.height() + 1),
+                Ignored,
+            ),
+        ];
+        let actual = suggest_scan_ranges(&st.wallet().conn, Ignored).unwrap();
+        assert_eq!(actual, expected);
```
nuttycom, line 1366:
```suggestion
```

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
nuttycom: Actually, we need to make sure that we have the note commitment tree frontier when a wallet is created. Maybe make the birthday entirely non-optional?

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
str4d: It's not zero-width: the line below uses `sapling_activation_height..(birthday_height + 1)`, so if these are equal then the range has length 1.

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
str4d: When `birthday = None`, the wallet historically set the birthday to "whatever height it has the most recent frontier for", rather than "chain tip".
- For the first account (previously added via `init_accounts_table`), that was the initial frontier (added via `init_blocks_table`).
- For subsequent accounts, it happened to be chain tip, because previously we used linear scanning so we'd always end up making the frontier at that height accessible.

Now that both of those methods are being replaced by `create_account`, and we aren't assuming linear scanning anymore, we need to ensure both cases are handled.
- For the first call to `create_account`, we want to have a frontier to bound the scanning work necessary.
  - If `birthday = Some(_)` then we're fine, as the caller provides a frontier.
  - If `birthday = None`, then the internally-selected birthday height will be missing a frontier, and we need to ensure that the scan queue is updated in a way that will produce it.
- For subsequent calls to `create_account`:
  - If `birthday = Some(_)` then we're fine, even if that birthday is earlier than any known to the wallet (as the provided frontier moves the bound on necessary scanning work a bit earlier, but doesn't otherwise break the assumption that we don't need to scan the tree earlier than the earlierst birthday height).
  - If `birthday = None` then we can just use the height we want (ensuring the scan queue will rescan after that height), and we don't need a frontier at all. We can instead rely on the fact that we will encounter a note for the account after this height, and all we actually need is a path at that note, not a frontier at the account's birthday height.

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
nuttycom: Ah right, I was remembering the rationale from before we changed the definition of "Wallet birthday" back.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/be79a1c5cc0f7847db70516529cd876e7d3422bb..b845947e66082b70e6619558e2483a85374ff6f7) to switch back to "wallet birthday is the first block to scan" semantics after a detailed discussion with @str4d.

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
nuttycom: After an offline discussion, we decided to make the birthday non-optional. This means that wallet developers can get predictable results, and we don't have to worry about a bunch of different cases here.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/b845947e66082b70e6619558e2483a85374ff6f7..7a532e93f272463e3bda1a4bbaadace78adca46d) for rebase to `main`.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/b845947e66082b70e6619558e2483a85374ff6f7..1bd64b2) for rebase to `main`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,86 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+                is_marked: false,
+            },
+        )?;
+    }
+
+    let sapling_activation_height = params
+        .activation_height(NetworkUpgrade::Sapling)
+        .expect("Sapling activation height must be available.");
+
+    // Add the ignored range up to and including the birthday height.
```
str4d, line 210:
```suggestion
    // Add the ignored range up to the birthday height.
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,86 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+                ignored_range.clone(),
+                ScanPriority::Ignored,
+            ))
+            .into_iter(),
+            false,
+        )?;
+    };
+
+    // Rewrite the scan ranges above the birthday height so that we'll ensure we re-scan to find
+    // any notes that might belong to the newly added account.
```
str4d, line 227:
```suggestion
    // Rewrite the scan ranges starting from the birthday height so that we'll ensure we
    // re-scan to find any notes that might belong to the newly added account.
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1274,42 +1325,33 @@ mod tests {
...
-            &[0x0, 0x0, 0x0],
-        )
-        .unwrap();
-
-        let seed = Secret::new(vec![0u8; 32]);
-        let (_, usk) = st.wallet_mut().create_account(&seed).unwrap();
-        let _dfvk = usk.to_unified_full_viewing_key().sapling().unwrap().clone();
-
         let expected = vec![
             // The range up to and including the wallet's birthday height is ignored.
```
str4d, line 1350:
```suggestion
            // The range up to the wallet's birthday height is ignored.
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1322,14 +1364,26 @@ mod tests {
                     // complete the left-hand side of the required shard; this can be fixed once we
                     // have proper account birthdays.
```
str4d, line 1365: The comment above should now be out-of-date, given that we now have proper account birthdays? Can this test be fixed?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1343,25 +1397,25 @@ mod tests {
             // The birthday height was "last scanned" (as the wallet birthday) so we verify 10
             // blocks starting at that height.
```
str4d, line 1398:
```suggestion
            // The birthday height is the "first to be scanned" (as the wallet birthday),
            // so we verify 10 blocks starting at that height.
```

zcash_client_sqlite/src/wallet/init/migrations/add_account_birthdays.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        transaction.execute_batch(&format!(
+            "ALTER TABLE accounts ADD COLUMN birthday_height INTEGER;
+                ALTER TABLE accounts ADD COLUMN recover_until_height INTEGER;
```
str4d, line 39: Is this `ALTER TABLE` to add `recover_until_height` necessary? We don't migrate any data into it, and we don't insert anything from it into `accounts_new` below, so I think we can skip this and just create it directly in `accounts_new`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -675,6 +675,40 @@ pub(crate) fn get_sent_memo(
         .transpose()
 }
 
+/// Returns the minimum birthday height for accounts in the wallet.
+///
+/// TODO ORCHARD: we should consider whether we want to permit protocol-restricted accounts; if so,
+/// we would then want this method to take a protocol identifier to be able to learn the wallet's
+/// "Orchard birthday" which might be different from the overall wallet birthday.
```
str4d, line 682: Don't expose the TODO in the public documentation:
```suggestion
//
// TODO ORCHARD: we should consider whether we want to permit protocol-restricted accounts; if so,
// we would then want this method to take a protocol identifier to be able to learn the wallet's
// "Orchard birthday" which might be different from the overall wallet birthday.
```

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -159,9 +168,9 @@ pub(crate) fn get_spendable_sapling_notes(
             -- select all the unscanned ranges involving the shard containing this note
             WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
             AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
-            -- exclude unscanned ranges above the anchor height which don't affect spendability
+            -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
             AND unscanned.block_range_start <= :anchor_height
-            -- exclude unscanned ranges that end below the wallet birthday
+            -- exclude unscanned ranges that end below or on the wallet birthday
```
str4d, line 173:
```suggestion
            -- exclude unscanned ranges that end below the wallet birthday
```

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -235,9 +254,9 @@ pub(crate) fn select_spendable_sapling_notes(
                 -- select all the unscanned ranges involving the shard containing this note
                 WHERE sapling_received_notes.commitment_tree_position >= unscanned.start_position
                 AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
-                -- exclude unscanned ranges above the anchor height which don't affect spendability
+                -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
                 AND unscanned.block_range_start <= :anchor_height
-                -- exclude unscanned ranges that end below the wallet birthday
+                -- exclude unscanned ranges that end below or on the wallet birthday
```
str4d, line 259:
```suggestion
                -- exclude unscanned ranges that end below the wallet birthday
```

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -161,6 +161,8 @@ pub(crate) fn get_spendable_sapling_notes(
             AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
             -- exclude unscanned ranges above the anchor height which don't affect spendability
             AND unscanned.block_range_start <= :anchor_height
+            -- exclude unscanned ranges that end below the wallet birthday
+            AND unscanned.block_range_end > :wallet_birthday
```
str4d: 11fc78c7e80f6d0d6bd83a2f684b9741e86de26e does not currently run, because these two lines should be part of 1bd64b2c4afeb1acc9131627ba94f378a3671041.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -235,6 +237,8 @@ pub(crate) fn select_spendable_sapling_notes(
                 AND sapling_received_notes.commitment_tree_position < unscanned.end_position_exclusive
                 -- exclude unscanned ranges above the anchor height which don't affect spendability
                 AND unscanned.block_range_start <= :anchor_height
+                -- exclude unscanned ranges that end below the wallet birthday
+                AND unscanned.block_range_end > :wallet_birthday
```
str4d: 11fc78c7e80f6d0d6bd83a2f684b9741e86de26e does not currently run, because these two lines should be part of 1bd64b2c4afeb1acc9131627ba94f378a3671041.

str4d (dismissed):
utACK 1bd64b2c4afeb1acc9131627ba94f378a3671041 modulo fixing the second commit to run (by moving the noted SQL query changes to the fourth commit). Everything else is non-blocking documentation fixes or cleanups.

zcash_client_backend/CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Notable Changes
+
+- `zcash_client_backend` now supports out-of-order scanning of blockchain history.
+- This release of `zcash_client_backend` defines the concept of an account
+  birthday. The account birthday is defined as the minimum height among blocks
+  to be scanned when recovering an account.
```
daira, line 15:
```suggestion
  to be scanned when recovering an account. A wallet's birthday is the earliest
  birthday height among accounts maintained by the wallet.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -117,6 +119,16 @@ pub trait WalletRead {
     /// transaction is not in the main chain.
     fn get_tx_height(&self, txid: TxId) -> Result<Option<BlockHeight>, Self::Error>;
 
+    /// Returns the birthday height for the wallet.
+    ///
+    /// This returns earliest birthday height among accounts maintained by this wallet, or
+    /// `Ok(None)` if the wallet has no initialized accounts.
```
daira, line 125:
```suggestion
    /// This returns the earliest birthday height among accounts maintained by this wallet,
    /// or `Ok(None)` if the wallet has no initialized accounts.
```

str4d (dismissed):
Re-utACK 3a1d4a4161af07f02e3750553ebad088692c444d

zcash_client_backend/src/data_api.rs
```diff
@@ -472,25 +484,77 @@ impl SentTransactionOutput {
...
+    }
 }
 
 impl AccountBirthday {
     /// Constructs a new [`AccountBirthday`] from its constituent parts.
     ///
-    /// * `height`: The birthday height of the account. This is defined as the height of the last
-    /// block that is known to contain no transactions sent to addresses belonging to the account.
+    /// * `height`: The birthday height of the account. This is defined as the height of the first
+    ///    block block to be scanned in wallet recovery.
```
daira, line 512:
```suggestion
    ///    block to be scanned in wallet recovery.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -501,6 +565,24 @@ impl AccountBirthday {
...
         self.height
     }
+
+    /// Returns the height at which the wallet should exit "recovery mode".
+    pub fn recover_until(&self) -> Option<BlockHeight> {
+        self.recover_until
+    }
+
+    #[cfg(feature = "test-dependencies")]
+    pub fn from_sapling_activation<P: zcash_primitives::consensus::Parameters>(
```
daira, line 575:
```suggestion
    /// Constructs a new [`AccountBirthday`] at Sapling activation, with no
    /// "recover until" height.
    ///
    /// # Panics
    ///
    /// Panics if the Sapling activation height is not set.
    pub fn from_sapling_activation<P: zcash_primitives::consensus::Parameters>(
```

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: The comment here is just wrong; we don't actually scan from below the account birthday in the test below. Removing.

zcash_client_backend/src/data_api.rs
```diff
@@ -509,24 +591,36 @@ pub trait WalletWrite: WalletRead {
...
+    /// If `seed` was imported from a backup and this method is being used to restore a previous
+    /// wallet state, you should use this method to add all of the desired accounts before scanning
+    /// the chain from the seed's birthday height.
     ///
-    /// By convention, wallets should only allow a new account to be generated after funds
-    /// have been received by the currently-available account (in order to enable
-    /// automated account recovery).
+    /// By convention, wallets should only allow a new account to be generated after funds have
+    /// been received by the currently-available account (in order to enable automated account
+    /// recovery).
```
daira, line 617:
```suggestion
    /// By convention, wallets should only allow a new account to be generated after confirmed
    /// funds have been received by the currently-available account (in order to enable automated
    /// account recovery).
```

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -36,8 +38,14 @@ and this library adheres to Rust's notion of
 ### Removed
 - The empty `wallet::transact` module has been removed.
 - `zcash_client_sqlite::NoteId` has been replaced with `zcash_client_sqlite::ReceivedNoteId`
-  as the `SentNoteId` variant of is now unused following changes to 
+  as the `SentNoteId` variant of is now unused following changes to
```
daira, line 41:
```suggestion
  as the `SentNoteId` variant of it is now unused following changes to
```

str4d (approved):
Re-utACK 5b3f54410261d610d9090edb8dcddd5b4b26ddda.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,90 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+
+        replace_queue_entries::<SqliteClientError>(
+            conn,
+            &rescan_range,
+            Some(ScanRange::from_parts(
+                rescan_range.clone(),
+                ScanPriority::Historic,
+            ))
+            .into_iter(),
+            true,
```
daira, line 243:
```suggestion
            true, // force rescan
```

daira (commented):
Flushing comments.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -157,29 +163,90 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     params: &P,
     account: AccountId,
     key: &UnifiedFullViewingKey,
+    birthday: AccountBirthday,
 ) -> Result<(), SqliteClientError> {
-    add_account_internal(conn, params, "accounts", account, key)
-}
+    // Set the wallet birthday, falling back to the chain tip if not specified
```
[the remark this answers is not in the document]
nuttycom: Oh, this comment is now wrong.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1628,11 +1732,15 @@ mod tests {
...
     fn put_received_transparent_utxo() {
+        use crate::testing::TestBuilder;
+
         let mut st = TestBuilder::new().build();
 
         // Add an account to the wallet
         let seed = Secret::new([0u8; 32].to_vec());
-        let (account_id, _usk) = st.wallet_mut().create_account(&seed).unwrap();
+        let birthday =
+            AccountBirthday::from_parts(st.sapling_activation_height(), Frontier::empty(), None);
```
daira, line 1742: Is this intentionally not `AccountBirthday::from_sapling_activation`? It seems equivalent.

zcash_client_sqlite/src/wallet/init/migrations/add_account_birthdays.rs
```diff
@@ -0,0 +1,74 @@
...
+            PRAGMA legacy_alter_table = ON;
+            DROP TABLE accounts;
+            ALTER TABLE accounts_new RENAME TO accounts;
+            PRAGMA legacy_alter_table = OFF;
+            PRAGMA foreign_keys=ON;",
+            u32::from(
+                self.params
+                    .activation_height(NetworkUpgrade::Sapling)
+                    .unwrap()
+            )
```
daira, line 65: I know this is guaranteed to be an integer, nevertheless I would have used a named parameter.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -186,6 +198,15 @@ pub(crate) fn select_spendable_sapling_notes(
     anchor_height: BlockHeight,
     exclude: &[ReceivedNoteId],
 ) -> Result<Vec<ReceivedSaplingNote<ReceivedNoteId>>, SqliteClientError> {
```
daira, line 200: The partially duplicated code between this and `get_spendable_sapling_notes` seems like a maintenance hazard. I see that the SQL statement is different and has an extra `:target_value` parameter; can the rest of these methods be abstracted out?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -443,14 +483,66 @@ pub(crate) fn insert_queue_entries<'a>(
...
-) -> Result<(), SqliteClientError> {
+    force_rescans: bool,
+) -> Result<(), E> {
     let (to_create, to_delete_ends) = {
-        let mut suggested_stmt = conn.prepare_cached(
-            "SELECT block_range_start, block_range_end, priority
+        let mut suggested_stmt = conn
+            .prepare_cached(
+                "SELECT block_range_start, block_range_end, priority
             FROM scan_queue
```
daira, line 546: rustfmt has messed up the indentation of the SQL statement.

daira (commented):
Post-hoc ACK with comments.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: The fact that we have two separate such methods is a historical artifact. We should clean it up some day.

zcash_client_sqlite/src/wallet/init/migrations/add_account_birthdays.rs
nuttycom: I did this mostly for the sake of using `execute_batch`; as far as I know that is the only mechanism available for executing _multiple_ SQL statements in a single call.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/935

zcash_client_sqlite/src/wallet.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/935

zcash_client_sqlite/src/wallet.rs
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/935

zcash_client_sqlite/src/error.rs
```diff
@@ -124,13 +135,16 @@ impl fmt::Display for SqliteClientError {
             SqliteClientError::BlockConflict(h) => write!(f, "A block hash conflict occurred at height {}; rewind required.", u32::from(*h)),
             SqliteClientError::NonSequentialBlocks => write!(f, "`put_blocks` requires that the provided block range be sequential"),
             SqliteClientError::DiversifierIndexOutOfRange => write!(f, "The space of available diversifier indices is exhausted"),
+            SqliteClientError::AccountUnknown(id) => write!(f, "Account {} does not belong to this wallet.", u32::from(*id)),
```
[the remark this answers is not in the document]
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/935

Issue #899, opened 2023-08-16
`zcash_client_sqlite`: Double-entry accounting not working for recover-from-seed


str4d:
I have a mainnet wallet in which I have a bunch of transactions, generating a bunch of change, back when we didn't have internal viewing keys. I recovered from seed to set up a new instance of a mobile wallet, and then made a transaction. My balance went up by the value of the change output, instead of decreasing by the value of the sent amount.

Looking at the SQLite database, I see that my `sent_notes` table contains an output for the new recipient, and the new change note (that my balance increased by), but does not contain any entries for the old change notes (violating the double-entry accounting assumptions now baked into `zcash_client_sqlite`).

nuttycom:
After a brief look over the code, I think what's happening is that the change notes that were created under the old rules where the note was sent back to the sending address are note correctly being added to the `received_notes` table in the same fashion as are notes that are now sent to the account's internal change address.

daira:
> Looking at the SQLite database, I see that my sent_notes table contains an output for the new recipient, and the new change note (that my balance increased by), but does not contain any entries for the old change notes (violating the double-entry accounting assumptions now baked into zcash_client_sqlite).

I can reproduce the absence of entries in `sent_notes` for notes sent prior to the wallet recovery, but I cannot reproduce the incorrect balance symptom.

nuttycom:
This may now be fixed by #1306

nuttycom:
Closing as complete.

PR #908, opened 2023-08-22
Improve wallet tests

### LINKED ISSUE #936: sent_notes table is not fully populated after a recovery from seed
This is separated out of #899 since it is a better-understood issue, and can be reproduced by the test in #908.

When a wallet is recovered from seed and fully rescanned, the `sent_notes` table does not contain all of the notes that were in that table before the recovery.

daira:
This was originally going to test #899, and then it was going to test and fix #936. These tasks ended up being more complicated than we thought given the inconsistent use of the `sent_notes` table, and the introduction of out-of-order scanning (which can cause us not to know which spent notes are ours during the process of trying to categorize outputs). Now the PR just improves the existing wallet tests, provides support for resetting a wallet in the test framework, and checks that resetting a wallet works at all.

### DIFF
--- a/zcash_client_sqlite/src/wallet/sapling.rs
+++ b/zcash_client_sqlite/src/wallet/sapling.rs
@@ -434,6 +434,7 @@ pub(crate) mod tests {
     use std::{convert::Infallible, num::NonZeroU32};
 
     use incrementalmerkletree::Hashable;
+    use secrecy::Secret;
     use zcash_proofs::prover::LocalTxProver;
 
     use zcash_primitives::{
@@ -464,6 +465,7 @@ pub(crate) mod tests {
             error::Error,
             wallet::input_selection::{GreedyInputSelector, GreedyInputSelectorError},
             AccountBirthday, Ratio, ShieldedProtocol, WalletCommitmentTrees, WalletRead,
+            WalletWrite,
         },
         decrypt_transaction,
         fees::{fixed, zip317, DustOutputPolicy},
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1325,6 +1326,161 @@ pub(crate) mod tests {
...
+        );
+
+        assert_eq!(
+            get_balance(&db_restore.conn, AccountId::from(1)).unwrap(),
+            amount_sent
+        );
+        assert_eq!(
+            get_balance(&db_restore.conn, AccountId::from(0)).unwrap(),
+            amount_left
+        );
```
daira, line 1481: These checks currently fail. I believe that's detecting the bug, but we'll have to fix the bug to be sure.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1325,6 +1326,161 @@ pub(crate) mod tests {
...
+        ])
+        .unwrap();
+
+        let fee_rule = FixedFeeRule::standard();
+        let input_selector = GreedyInputSelector::new(
+            fixed::SingleOutputChangeStrategy::new(fee_rule),
+            DustOutputPolicy::default(),
+        );
+
+        let txid = spend(
```
nuttycom, line 1416: This will send change back to the internal address; is that intended for this test?

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: You've initialized `db_restore`, but not added the account to the wallet, so this won't detect anything.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: Oh, right, the self-send should be detected as legacy change.

daira:
I want to get the test improvements that are already finished merged, so that I don't have to keep rebasing them.

nuttycom (approved):
utACK https://github.com/zcash/librustzcash/pull/908/commits/101c9c0ea491853b67dda9fdfbf8ae35948df202

PR #909, opened 2023-08-24
Fix API doc comment on UnifiedSpendingKey

AArnott:
This fixes a simple copy/paste error of the docs from the unified viewing key.


nuttycom (approved):
Doc-only ACK

PR #914, opened 2023-08-24
Add `get_wallet_summary` to `WalletRead`

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -44,6 +47,129 @@ pub enum NullifierQuery {
...
+    progress: ScanProgress,
+    progress_since_fully_scanned: ScanProgress,
+}
+
+impl WalletSummary {
+    pub fn new(
+        account_balances: BTreeMap<AccountId, AccountBalance>,
+        chain_tip_height: BlockHeight,
+        fully_scanned_height: BlockHeight,
+        progress: ScanProgress,
```
nuttycom, line 140: Overall scan progress is probably not a valuable metric, and will likely be removed.

str4d (commented):
Reviewed 1e5eae87a4412bb66789f563e64aa9905e512261

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -54,6 +54,7 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                 LEFT OUTER JOIN sapling_tree_shards prev_shard
                     ON shard.shard_index = prev_shard.shard_index + 1
                 INNER JOIN scan_queue ON
+                    (scan_queue.block_range_start >= subtree_start_height AND shard.subtree_end_height IS NULL) OR
```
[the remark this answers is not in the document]
nuttycom: Correct.

zcash_client_sqlite/src/wallet/init/migrations/wallet_summaries.rs
```diff
@@ -0,0 +1,122 @@
...
+                LEFT OUTER JOIN sapling_tree_shards prev_shard
+                    ON shard.shard_index = prev_shard.shard_index + 1
+                INNER JOIN scan_queue ON
+                    (scan_queue.block_range_start >= subtree_start_height AND shard.subtree_end_height IS NULL) OR
+                    (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    ((scan_queue.block_range_end - 1) BETWEEN subtree_start_height AND shard.subtree_end_height) OR
+                    (
+                        scan_queue.block_range_start <= prev_shard.subtree_end_height
+                        AND (scan_queue.block_range_end - 1) >= shard.subtree_end_height
+                    )
```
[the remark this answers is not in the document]
nuttycom: So, we could potentially create a base view that doesn't filter on spentness, and then select from that with spentness conditions in one place, and use the `GROUP BY` in the other. I'll try it out.

zcash_client_backend/src/data_api.rs
```diff
@@ -44,6 +47,95 @@ pub enum NullifierQuery {
...
+}
+
+/// A type representing the potentially-spendable value of unspent outputs in the wallet.
+///
+/// The balances reported using this data structure may overestimate the total spendable value of
+/// the wallet, in the case that the spend of a previously received shielded note has not yet been
+/// detected by the process of scanning the chain. The balances reported using this data structure
+/// can only be certain to be unspent in the case that [`Self::chain_tip_height`] is equal to
+/// [`Self::fully_scanned_height]`, and even in this circumstance it is possible that a newly
+/// created transaction could conflict with a not-yet-mined transaction in the mempool.
```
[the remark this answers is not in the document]
nuttycom: Yeah, adding a method for that to the struct is trivial, no reason not to do it.

zcash_client_backend/src/data_api.rs
```diff
@@ -44,6 +47,95 @@ pub enum NullifierQuery {
     All,
 }
 
+#[derive(Debug, Clone)]
+pub struct AccountBalance {
+    pub spendable_value: NonNegativeAmount,
+    pub pending_change_value: NonNegativeAmount,
+    pub value_pending_spendability: NonNegativeAmount,
+    pub unshielded: NonNegativeAmount,
```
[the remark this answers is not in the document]
nuttycom: The primary reason that I made the struct fields public is that it's convenient to be able to mutate them in https://github.com/zcash/librustzcash/pull/914/commits/1e5eae87a4412bb66789f563e64aa9905e512261#diff-a1724f7547ecabc613a07520eaf7917d9084d4277219d186698b0c8e352495a4R643-R649, and since the balances themselves are only available immutably via reference from `WalletSummary`, I figured that was okay.

zcash_client_backend/src/data_api.rs
```diff
@@ -44,6 +47,95 @@ pub enum NullifierQuery {
     All,
 }
 
+#[derive(Debug, Clone)]
+pub struct AccountBalance {
+    pub spendable_value: NonNegativeAmount,
+    pub pending_change_value: NonNegativeAmount,
+    pub value_pending_spendability: NonNegativeAmount,
```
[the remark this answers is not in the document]
nuttycom: Naming is hard. I didn't want to actually expose change as separate from value pending spendability, because I don't think the difference between them is all that important, but the current Zashi wallet designs have them separated out (though recently received non-change notes with fewer than `min_confirmations` and old change notes that are awaiting subtree scanning are not distinguished from one another.)

zcash_client_backend/src/data_api.rs
```diff
@@ -44,6 +47,95 @@ pub enum NullifierQuery {
...
+    ) -> Self {
+        Self {
+            account_balances,
+            chain_tip_height,
+            fully_scanned_height,
+            sapling_scan_progress,
+        }
+    }
+
+    pub fn account_balances(&self) -> &BTreeMap<AccountId, AccountBalance> {
```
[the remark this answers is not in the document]
nuttycom: I subscribe to the philosophy that one should be abstract in inputs and concrete in results.

zcash_client_backend/src/data_api.rs
```diff
@@ -155,15 +247,12 @@ pub trait WalletRead {
...
-    /// that they are not yet deemed spendable.
-    fn get_balance_at(
+    /// Returns the wallet balances and sync status for an account as of the specified block
+    /// height, or `Ok(None)` if the wallet has no balance data available.
+    fn get_wallet_summary(
         &self,
-        account: AccountId,
-        anchor_height: BlockHeight,
-    ) -> Result<Amount, Self::Error>;
+        anchor_height: Option<BlockHeight>,
```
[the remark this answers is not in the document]
nuttycom: Resolved by requiring `min_confirmations: u32`.

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
nuttycom: Naming has been improved, unresolve if necessary.

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
nuttycom: I'll add extra documentation to the `unshielded` field to the following effect:

Unshielded balances are not subject to confirmation-depth constraints, because the only possible operation on a transparent balance is to shield it, it is possible to create a zero-conf transaction to perform that shielding, and the resulting shielded notes will be subject to normal confirmation rules. So there's currently no reason to make transparent funds subject to confirmation requirements at all.

Some of the other existing APIs should probably be updated to reflect this realization.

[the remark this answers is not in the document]
nuttycom: Right, we don't have any automatic shielding functionality at all, so at present this isn't a concern.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,201 @@ pub(crate) fn get_balance(
...
+                    Ok(min_idx.zip(max_idx).map(|(min, max)| {
+                        let max_tree_size = (max + 1) << SAPLING_SHARD_HEIGHT;
+                        let min_tree_size = min << SAPLING_SHARD_HEIGHT;
+                        Ratio::new(fully_scanned, max_tree_size - min_tree_size)
+                    }))
+                },
+            )
+            .optional()
+            .map(|opt| opt.flatten())
+        }
```
nuttycom, line 587: This whole thing is a bit of a nightmare, because we don't have any source of note commitment tree sizes at the wallet birthday or at the chain tip. If we could reliably get those pieces of information, then this could be much better but this is a reasonable minimum-viable implementation.

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
     All,
 }
 
+/// Balance information for a value within a single shielded pool in an account.
```
daira, line 54:
```suggestion
/// Balance information for a value within a single shielded pool in an account.
```
Will this only ever be used for shielded pools? It seems applicable to the transparent pool as well.

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
daira: Right, but defining this to include zero-conf funds means that an adversary will be able to arbitrarily inflate it above the realistic balance, as soon as we implement detection of zero-conf funds.

If I send you funds in a bunch of transactions that each have 50 unpaid actions and expire at the next block, I can be pretty confident that at most one (probably zero) of them will be mined.

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+/// A polymorphic ratio type, usually used for rational numbers.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Ratio<T> {
+    numerator: T,
+    denominator: T,
+}
+
+impl<T> Ratio<T> {
+    /// Constructs a new Ratio from a numerator and a denominator.
+    pub fn new(numerator: T, denominator: T) -> Self {
```
daira, line 121: If the numerator and denominator were separate types, then the denominator could be defined to be a non-zero type.

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+        &self.account_balances
+    }
+
+    /// Returns the height of the current chain tip.
+    pub fn chain_tip_height(&self) -> BlockHeight {
+        self.chain_tip_height
+    }
+
+    /// Returns the height below which all blocks wallet have been scanned, ignoring blocks below
+    /// the wallet birthday.
```
daira, line 182:
```suggestion
    /// Returns the height below which all blocks have been scanned by the wallet, ignoring blocks
    /// below the wallet birthday.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+    pub fn fully_scanned_height(&self) -> BlockHeight {
+        self.fully_scanned_height
+    }
+
+    /// Returns the progress of scanning Sapling outputs, in terms of the ratio between notes
+    /// scanned and the total number of notes added to the chain since the wallet birthday.
+    ///
+    /// This ratio should only be used to compute progress percentages, and the numerator and
+    /// denominator should not be treated as authoritative note counts. Returns `None` if the
+    /// wallet is unable to determine the size of the note commitment tree.
```
daira, line 192: Does this need to be Sapling-specific? Once we support Orchard scanning, if we follow the pattern established here (adding an `orchard_scan_progress` method) then the wallet UI would have to somehow merge the progress reports for Sapling and Orchard. It could do that by calculating `(sapling_num + orchard_num) / (sapling_denom + orchard_denom)`, but that would only give a sensible result if we assume that the numerators and denominators represent note counts, which the wallet is not supposed to do. Therefore we might as well define a single `scan_progress` method and make things simpler.

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+    pub fn sapling_scan_progress(&self) -> Option<Ratio<u64>> {
```
daira, line 193: Does this need to be Sapling-specific? Once we support Orchard scanning, if we follow the pattern established here (adding an `orchard_scan_progress` method) then the wallet UI would have to somehow merge the progress reports for Sapling and Orchard. It could do that by calculating `(sapling_num + orchard_num) / (sapling_denom + orchard_denom)`, but that would only give a sensible result if we assume that the numerators and denominators represent note counts, which the wallet is not supposed to assume. Therefore we might as well define a single `scan_progress` method and make things simpler.

```suggestion
    /// Returns the progress of scanning outputs, in terms of the ratio between notes scanned
    /// and the total number of notes added to the chain since the wallet birthday.
    ///
    /// This ratio should only be used to compute progress percentages, and the numerator and
    /// denominator should not be treated as authoritative note counts. Returns `None` if the
    /// wallet is unable to determine the size of the note commitment tree(s).
    pub fn scan_progress(&self) -> Option<Ratio<u64>> {
```

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+    sapling_scan_progress: Option<Ratio<u64>>,
+}
+
+impl WalletSummary {
+    /// Constructs a new [`WalletSummary`] from its constituent parts.
+    pub fn new(
+        account_balances: BTreeMap<AccountId, AccountBalance>,
+        chain_tip_height: BlockHeight,
+        fully_scanned_height: BlockHeight,
+        sapling_scan_progress: Option<Ratio<u64>>,
```
daira, line 161:
```suggestion
        scan_progress: Option<Ratio<u64>>,
```
Change this also if taking the suggestion [below](https://github.com/zcash/librustzcash/pull/914/files#r1314220128).

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+    ) -> Self {
+        Self {
+            account_balances,
+            chain_tip_height,
+            fully_scanned_height,
+            sapling_scan_progress,
```
daira, line 167:
```suggestion
            scan_progress,
```

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+    }
+
+    /// Returns the progress of scanning Sapling outputs, in terms of the ratio between notes
+    /// scanned and the total number of notes added to the chain since the wallet birthday.
+    ///
+    /// This ratio should only be used to compute progress percentages, and the numerator and
+    /// denominator should not be treated as authoritative note counts. Returns `None` if the
+    /// wallet is unable to determine the size of the note commitment tree.
+    pub fn sapling_scan_progress(&self) -> Option<Ratio<u64>> {
+        self.sapling_scan_progress
```
daira, line 194:
```suggestion
        self.scan_progress
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -286,6 +289,57 @@ where
...
+
+    pub(crate) fn get_total_balance(&self, account: AccountId) -> NonNegativeAmount {
+        get_wallet_summary(&self.wallet().conn, 0)
+            .unwrap()
+            .unwrap()
+            .account_balances()
+            .get(&account)
+            .unwrap()
+            .total()
+    }
```
daira, line 301:
```suggestion
    /// Returns the sum of balances in all pools with the given `min_confirmations`, for the given account.
    pub(crate) fn get_total_balance_at_depth(&self, account: AccountId, min_confirmations: u32) -> NonNegativeAmount {
        get_wallet_summary(&self.wallet().conn, min_confirmations)
            .unwrap()
            .unwrap()
            .account_balances()
            .get(&account)
            .unwrap()
            .total()
    }

    /// Returns the sum of balances in all pools with no constraint on confirmations, for the
    /// given account.
    pub(crate) fn get_total_balance(&self, account: AccountId) -> NonNegativeAmount {
        self.get_total_balance_at_depth(account, 0)
    }
```
(may need reformatting).

daira: Note to self: when rebasing #908 on this, don't add `TestState::get_balance`, and replace uses of it with `TestState::get_total_balance`. Keep `TestState::get_balance_at`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+        if let Some(tip) = tip_row {
+            let tip_delta = u32::from(chain_tip_height) - tip.subtree_start_height;
+
+            if let Some(prev) = prev_row {
+                // use the previous row to interpolate the recent rate of output creation
+                let prev_end = prev
+                    .subtree_end_height
+                    .expect("End height is known for the second-to-last shard");
+                let block_delta = prev_end - prev.subtree_start_height;
+                let rate = f64::from(0x1 << SAPLING_SHARD_HEIGHT) / f64::from(block_delta);
```
daira, line 588:
```suggestion
                assert!(block_delta > 0, "a shard spans at least one block");
                let rate = f64::from(0x1 << SAPLING_SHARD_HEIGHT) / f64::from(block_delta);
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+                let tip_notes = (f64::from(tip_delta) * rate) as u64;
```
daira, line 589: This rounds toward zero.
```suggestion
                let tip_notes = (f64::from(tip_delta) * rate).round() as u64;
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+                    .map(|opt| opt.unwrap_or(0)),
+                    Ok
+                )?.into();
+
+                Ok(Some(Ratio::new(
+                    scanned,
+                    (tip.start_position - incomplete_start) + tip_notes
+                )))
+            } else {
+                // the chain tip shard is the only one
```
daira, line 619: I'm confused that this compiles. How does Rust infer what the value of the `if let Some(prev) = prev_row { ... } else {}` construct should be in the `else` branch?

daira: Oh, it does not compile. Now I'm unconfused, lol.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+    let summary_height = chain_tip_height + 1 - min_confirmations;
+
+    // If any of the shard below the anchor (or chain tip if no anchor is specified) is unscanned
+    // above the wallet birthday height, none of our balance is currently spendable.
+    let any_spendable = conn
+        .query_row(
+            "SELECT 1 FROM v_sapling_shard_unscanned_ranges
+             WHERE :summary_height
+                BETWEEN subtree_start_height
+                AND IFNULL(subtree_end_height, :summary_height)
```
daira, line 660: If `min_confirmations` is zero, then the chain could be fully scanned up to and including a complete shard at `chain_tip_height`, and this query would return no results because `summary_height` would be one greater than `subtree_end_height`. This seems not to be the intent.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+    let birthday_height =
+        wallet_birthday(conn)?.expect("If a scan range exists, we know the wallet birthday.");
+
+    let fully_scanned_height =
+        block_fully_scanned(conn)?.map_or(birthday_height - 1, |m| m.block_height());
+
+    let sapling_scan_progress =
+        sapling_scan_progress(conn, fully_scanned_height, chain_tip_height)?;
+
+    let summary_height = chain_tip_height + 1 - min_confirmations;
```
daira, line 651:
```suggestion
    let summary_height = (chain_tip_height + 1).saturating_sub(std::cmp::max(min_confirmations, 1));
```
I think this is the correct fix for the bug below, and another bug if you pass a very large `min_confirmations`. (Might need reformatting.)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+/// Returns the spendable balance for the account at the specified height.
+///
+/// This may be used to obtain a balance that ignores notes that have been detected so recently
+/// that they are not yet spendable, or for which it is not yet possible to construct witnesses.
```
daira, line 630:
```suggestion
/// that they are not yet spendable, or for which it is not yet possible to construct witnesses.
///
/// `min_confirmations` can be 0, but that case is currently treated identically to
/// `min_confirmations == 1`. This behaviour may change in future.
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+            "SELECT u.received_by_account, SUM(u.value_zat)
+             FROM utxos u
+             LEFT OUTER JOIN transactions tx
+             ON tx.id_tx = u.spent_in_tx
+             WHERE u.height <= :max_height
+             AND tx.block IS NULL
+             GROUP BY u.received_by_account",
+        )?;
+        let mut rows = stmt_transparent_balances
+            .query(named_params![":max_height": u32::from(summary_height)])?;
```
daira, line 755: Re: the fix I suggested, this query does not currently appear to rely on `summary_height` being `chain_tip_height + 1` if `min_confirmations` is given as zero.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -365,16 +367,15 @@ mod tests {
         }
 
         let expected_views = vec![
-            // v_sapling_shard_unscanned_ranges
+            // v_sapling_shard_scan_ranges
             format!(
-                "CREATE VIEW v_sapling_shard_unscanned_ranges AS
-                WITH wallet_birthday AS (SELECT MIN(birthday_height) AS height FROM accounts)
+                "CREATE VIEW v_sapling_shard_scan_ranges AS
```
daira, line 372: What happens to the old view? Or did we not release any version with the old view?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -383,18 +384,53 @@ mod tests {
                 LEFT OUTER JOIN sapling_tree_shards prev_shard
                     ON shard.shard_index = prev_shard.shard_index + 1
                 INNER JOIN scan_queue ON
+                    (scan_queue.block_range_start >= subtree_start_height AND shard.subtree_end_height IS NULL) OR
                     (scan_queue.block_range_start BETWEEN subtree_start_height AND shard.subtree_end_height) OR
                     ((scan_queue.block_range_end - 1) BETWEEN subtree_start_height AND shard.subtree_end_height) OR
```
daira, line 389: Is it better not to rely on the behaviour of `BETWEEN x and NULL`? As I read the sqlite docs this is always false (as intended), and this appears to be consistent with at least the [X/Open version of the SQL standard](https://pubs.opengroup.org/onlinepubs/9695959099/toc.pdf).

https://www.sqlite.org/lang_expr.html#the_between_operator
> The BETWEEN operator is logically equivalent to a pair of comparisons. "x BETWEEN y AND z" is equivalent to "x>=y AND x<=z" except that with BETWEEN, the x expression is only evaluated once.

https://www.sqlite.org/lang_expr.html#operators_and_parse_affecting_attributes
> All operators generally evaluate to NULL when any operand is NULL, with specific exceptions as stated below. This is in accordance with the SQL92 standard.
>
> When paired with NULL:
> * AND evaluates to 0 (false) when the other operand is false; and
> * OR evaluates to 1 (true) when the other operand is true.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -383,18 +384,53 @@ mod tests {
...
-                    )
-                INNER JOIN wallet_birthday
-                WHERE scan_queue.priority > {}
-                AND scan_queue.block_range_end > wallet_birthday.height",
+                    )",
                 u32::from(st.network().activation_height(NetworkUpgrade::Sapling).unwrap()),
+            ),
+            // v_sapling_shard_unscanned_ranges
+            format!(
+                "CREATE VIEW v_sapling_shard_unscanned_ranges AS
```
daira, line 398: What happens to existing wallets that already contain this view (defined differently)? In general are we okay to arbitrarily change existing views in unreleased versions?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+        let max_priority = parse_priority_code(max_priority_raw).ok_or_else(|| {
+            SqliteClientError::CorruptedData(format!(
+                "Priority code {} not recognized.",
+                max_priority_raw
+            ))
+        })?;
+
+        let received_height = row
+            .get::<_, Option<u32>>(4)
+            .map(|opt| opt.map(BlockHeight::from))?;
```
daira, line 700:
```suggestion
            .get::<_, Option<u32>>(4)?
            .map(BlockHeight::from);
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+
+        let mut spendable_value = NonNegativeAmount::ZERO;
+        let mut change_pending_confirmation = NonNegativeAmount::ZERO;
+        let mut value_pending_spendability = NonNegativeAmount::ZERO;
+
+        let is_spendable = any_spendable
+            && received_height.iter().any(|h| h <= &summary_height)
```
daira, line 707: I continue to prefer `received_height.map_or(false, |h| h <= &summary_height)`, but I won't press the issue.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+            && max_priority <= ScanPriority::Scanned;
+
+        let is_pending_change = is_change && received_height.iter().all(|h| h > &summary_height);
```
daira, line 710: And this would be `received_height.map_or(true, |h| h > &summary_height)

daira: Is the `block` field in `sapling_received_notes` reset to NULL if the block is rolled back?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+
+    #[cfg(feature = "transparent-inputs")]
+    {
+        let mut stmt_transparent_balances = conn.prepare(
+            "SELECT u.received_by_account, SUM(u.value_zat)
+             FROM utxos u
+             LEFT OUTER JOIN transactions tx
+             ON tx.id_tx = u.spent_in_tx
+             WHERE u.height <= :max_height
+             AND tx.block IS NULL
```
daira, line 751: This is transparent TXOs that are received below or at the summary height, but if spent in any tx then that tx must be unmined, correct?

Are the `spent_in_tx` field in `utxos` and/or the `block` field in `transactions` set to NULL if the block containing a spending tx is rolled back?

Also, is the `height` field in `utxos` reset to NULL (or the row for that UTXO removed) if the block containing the output is rolled back?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -519,28 +521,264 @@ pub(crate) fn get_balance(
...
+        while let Some(row) = rows.next()? {
+            let account = AccountId::from(row.get::<_, u32>(0)?);
+            let raw_value = row.get(1)?;
+            let value = NonNegativeAmount::from_nonnegative_i64(raw_value).map_err(|_| {
+                SqliteClientError::CorruptedData(format!("Negative UTXO value {:?}", raw_value))
+            })?;
+
+            account_balances
+                .entry(account)
+                .and_modify(|bal| bal.unshielded = value)
```
daira, line 766:
```suggestion
                .and_modify(|bal| bal.unshielded += value)
```

zcash_client_sqlite/src/wallet/init/migrations/wallet_summaries.rs
```diff
@@ -0,0 +1,89 @@
...
+                MAX(priority) AS max_priority
+            FROM v_sapling_shard_scan_ranges
+            GROUP BY
+                shard_index,
+                start_position,
+                end_position_exclusive,
+                subtree_start_height,
+                subtree_end_height,
+                contains_marked;",
+        )?;
```
daira, line 81: `v_sapling_shard_scan_ranges` and `v_sapling_shard_unscanned_ranges` are both also changed.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -693,47 +691,37 @@ pub(crate) mod tests {
...
-            .unwrap()
-            .unwrap();
-        assert_eq!(
-            get_balance(&st.wallet().conn, AccountId::from(0)).unwrap(),
-            value
-        );
         assert_eq!(
-            get_balance_at(&st.wallet().conn, AccountId::from(0), anchor_height).unwrap(),
-            value
+            get_balance(&st.wallet().conn, account).unwrap(),
```
daira, line 704: Does this get verified balance?

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -693,47 +691,37 @@ pub(crate) mod tests {
...
-            .unwrap();
+        let total = (value + value).unwrap();
         assert_eq!(
-            get_balance(&st.wallet().conn, AccountId::from(0)).unwrap(),
-            (value + value).unwrap()
-        );
-        assert_eq!(
-            get_balance_at(&st.wallet().conn, AccountId::from(0), anchor_height2).unwrap(),
-            value
+            get_balance(&st.wallet().conn, account).unwrap(),
```
daira, line 719: Again, does this get verified balance?

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1049,27 +1023,25 @@ pub(crate) mod tests {
...
         let (h, _, _) = st.generate_next_block(&dfvk, AddressType::Internal, value);
         st.scan_cached_blocks(h, 1);
 
         // Verified balance matches total balance
-        let (_, anchor_height) = st
-            .wallet()
-            .get_target_and_anchor_heights(NonZeroU32::new(10).unwrap())
-            .unwrap()
-            .unwrap();
+        assert_eq!(get_balance(&st.wallet().conn, account).unwrap(), value);
```
daira, line 1035: Ditto.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1116,18 +1088,10 @@ pub(crate) mod tests {
...
-        let (_, anchor_height) = st
-            .wallet()
-            .get_target_and_anchor_heights(NonZeroU32::new(1).unwrap())
-            .unwrap()
-            .unwrap();
-        assert_eq!(
-            get_balance(&st.wallet().conn, AccountId::from(0)).unwrap(),
-            total
-        );
+        assert_eq!(get_balance(&st.wallet().conn, account).unwrap(), total);
```
daira, line 1091: Ditto.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1377,6 +1388,15 @@ mod tests {
         );
         st.scan_cached_blocks(birthday.height(), 1);
 
+        // We have scanned a the birthday block, so our scan is complete, showing 1/1
```
daira, line 1391:
```suggestion
        // We have scanned the birthday block, so our scan is complete, showing 1/1
```

daira, line 1391:
```suggestion
        // We have scanned the birthday block, so our scan is complete, showing 1/1
```

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -54,22 +53,39 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+                    )",
                 SAPLING_SHARD_HEIGHT,
                 SAPLING_SHARD_HEIGHT,
                 u32::from(self.params.activation_height(NetworkUpgrade::Sapling).unwrap()),
-                priority_code(&ScanPriority::Scanned),
             )
         )?;
 
+        transaction.execute_batch(&format!(
+            "CREATE VIEW v_sapling_shard_unscanned_ranges AS
```
daira, line 70: `v_sapling_shard_scan_ranges` is added in this PR, and `v_sapling_shard_unscanned_ranges` is changed.

daira (dismissed):
Currently does not compile; several blocking comments.

str4d:
Test failure:
```
---- wallet::scanning::tests::create_account_creates_ignored_range stdout ----
thread 'wallet::scanning::tests::create_account_creates_ignored_range' panicked at 'assertion failed: `(left == right)`
  left: `Some(Ratio { numerator: 1, denominator: 64300 })`,
 right: `Some(Ratio { numerator: 1, denominator: 65536 })`', zcash_client_sqlite/src/wallet/scanning.rs:1444:9
```

zcash_client_sqlite/src/wallet.rs
nuttycom: Yeah, that was the WIP commit that I pushed. Now removed.

zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
nuttycom: Since these views are unreleased and not part of the public API, there's no need to document this.

nuttycom:
> Currently does not compile; several blocking comments.

The second commit was WIP at the time of this review.

zcash_client_sqlite/src/wallet.rs
nuttycom: Yes

zcash_client_sqlite/src/wallet.rs
nuttycom: Yes, I believe so. If height information is not set to `NULL` when a block is rolled back, that would be a preexisting bug.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: We can't rely on the behavior of `OR` with respect to `NULL` here, because we explicitly want to return `TRUE` in the case that `shard.subtree_end_height IS NULL` and none of the other clauses evaluate to `TRUE`.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Yes.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: No, this gets total balance.

zcash_client_backend/src/data_api.rs
```diff
@@ -46,6 +51,155 @@ pub enum NullifierQuery {
...
+    /// The value of unspent Sapling outputs belonging to the account.
+    pub sapling_balance: Balance,
+
+    /// The value of all unspent transparent outputs belonging to the account, irrespective of
+    /// confirmation depth.
+    ///
+    /// Unshielded balances are not subject to confirmation-depth constraints, because the only
+    /// possible operation on a transparent balance is to shield it, it is possible to create a
+    /// zero-conf transaction to perform that shielding, and the resulting shielded notes will be
+    /// subject to normal confirmation rules.
```
daira, line 100: Defining this to include all zero-conf funds means that an adversary will be able to arbitrarily inflate it above the realistic balance, as soon as we implement detection of zero-conf funds.

If I send you funds in a bunch of transactions that each have 50 unpaid actions and expire at the next block, I can be pretty confident that at most one (probably zero) of them will be mined. So, using this value as defined for anything opens the wallet up to the transparent equivalent of a Faerie Gold attack (and to various social engineering attacks, such as refund scams, if it is ever shown to a user). This is also a potential DoS attack on auto-shielding.

I agree that we haven't yet implemented the things that would enable these attacks, but still, it should not be defined that way.

```suggestion
    /// Unshielded balance, with a minimum confirmation depth of 1.
    ///
    /// The only possible operation on a transparent balance is to shield it, and the resulting
    /// shielded notes will be subject to normal confirmation rules. Nevertheless, the unshielded
    /// balance should not be over-reported, and including zero-conf UTXOs without restriction
    /// would make it too easy for an adversary to cause such over-reporting. This policy might
    /// be refined in future.
```

zcash_client_backend/src/data_api.rs
daira: I disagree, because that is complexity that we know we won't need: neither a user nor a wallet cares about the progress of pools individually. In my opinion the best course when we introduce Orchard will be to rename `sapling_scan_progress` to `scan_progress` and have it do the merging. If we do that now instead then the API is both conceptually simpler, and avoids a future compatibility break.

str4d (changes requested):
Reviewed in a pairing with @nuttycom, so the PR kept changing to reflect these as we went. Marking as "changes requested" and then we can check they all were addressed.

zcash_client_backend/src/data_api.rs
nuttycom: Zero-conf balances have essentially become the standard among zcash (and Bitcoin and Ethereum) wallets, and the functionality to provide zero-conf balance is highly in demand. I don't know how we get around that; users in general would prefer to immediately see incoming funds as being recognized, at the cost of having to be diligent about waiting for important transactions to be mined.

zcash_client_backend/src/data_api.rs
nuttycom: Okay, I've made this change.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/7ba20e3c..6cbb107c71124b78cde1a19dc52173fb447045b17ba20e3) to address comments from code review

str4d (dismissed):
utACK 6cbb107c71124b78cde1a19dc52173fb447045b1

zcash_client_sqlite/src/testing.rs
nuttycom: This can be added when it's needed for a test.

zcash_client_sqlite/src/wallet.rs
nuttycom: This applies to https://github.com/zcash/librustzcash/pull/939

zcash_client_sqlite/src/wallet.rs
nuttycom: This applies to https://github.com/zcash/librustzcash/pull/939

zcash_client_sqlite/src/wallet.rs
```diff
@@ -491,56 +493,266 @@ pub(crate) fn is_valid_account_extfvk<P: consensus::Parameters>(
...
+            return Ok(None);
+        }
+    };
+
+    let birthday_height =
+        wallet_birthday(conn)?.expect("If a scan range exists, we know the wallet birthday.");
+
+    let fully_scanned_height =
+        block_fully_scanned(conn)?.map_or(birthday_height - 1, |m| m.block_height());
+    let summary_height = (chain_tip_height + 1).saturating_sub(std::cmp::max(min_confirmations, 1));
```
str4d, line 604: There is a slight issue here regarding notes that are just about to expire. Without the `max(_, 1)`, with `min_confirmations == 0` this would set `summary_height` to the "mempool height" (`chain_height + 1`), which is precisely what we want for any transaction in the mempool (that still has a chance of being mined, and thus mined before expiry). Down below in the note selection code, we exclude transactions that have expired as of `summary_height` (because `expiry_height < summary_height`) from the balance, which means if someone calls this API with `min_confirmations == 0`, then about-to-expire funds will disappear from the wallet balance, only to reappear again if the transaction does actually get mined in the last block it can.

I'm fine with this being an edge case we don't handle yet, and instead merging @daira's suggested comment change that we treat `min_confirmations == 0` as `min_confirmations == 1` for shielded notes.

str4d (approved):
Re-utACK 1e12e9d0e6ccb4583a173e4e8e295b9cab38663a

zcash_client_backend/src/data_api.rs
daira: We spent a huge amount of time, effort, and protocol complexity preventing Faerie Gold attacks. I don't see why we would just throw up our hands and accept the equivalent transparent attack because not caring about it is "the standard".

zcash_client_sqlite/src/wallet.rs
daira: Filed as #940.

PR #916, opened 2023-08-28
`zcash_client_sqlite`: Introduce `TestBuilder` and `TestRunner`

### LINKED ISSUE #917: `zcash_client_sqlite`: Refactor and simplify tests
It's currently quite laborious to write tests for `zcash_client_sqlite`. We should have a testing harness that does all the heavy lifting for us (like the RPC test harness does for `zcashd`), so that the tests themselves are more straightforward, and easier to write.

str4d:
These provide a standard way to set up tests, and helper methods for performing common test actions.

Closes #917.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -1100,20 +1100,17 @@ mod tests {
 
     #[test]
     pub(crate) fn get_next_available_address() {
-        let mut test = TestBuilder::new().with_test_account().build();
+        let mut st = TestBuilder::new().with_test_account().build();
 
         let account = AccountId::from(0);
-        let current_addr = test.wallet().get_current_address(account).unwrap();
+        let current_addr = st.wallet().get_current_address(account).unwrap();
         assert!(current_addr.is_some());
 
-        let addr2 = test
-            .wallet_mut()
-            .get_next_available_address(account)
-            .unwrap();
+        let addr2 = st.wallet_mut().get_next_available_address(account).unwrap();
         assert!(addr2.is_some());
         assert_ne!(current_addr, addr2);
 
-        let addr2_cur = test.wallet().get_current_address(account).unwrap();
+        let addr2_cur = st.wallet().get_current_address(account).unwrap();
         assert_eq!(addr2, addr2_cur);
     }
 
@@ -1122,17 +1119,17 @@ mod tests {
     fn transparent_receivers() {
         use secrecy::Secret;
 
-        let test = TestBuilder::new()
+        let st = TestBuilder::new()
             .with_block_cache()
             .with_seed(Secret::new(vec![]))
             .with_test_account()
             .build();
 
         // Add an account to the wallet.
-        let (ufvk, taddr) = test.test_account().unwrap();
+        let (ufvk, taddr) = st.test_account().unwrap();
         let taddr = taddr.unwrap();
 
-        let receivers = test.wallet().get_transparent_receivers(0.into()).unwrap();
+        let receivers = st.wallet().get_transparent_receivers(0.into()).unwrap();
 
         // The receiver for the default UA should be in the set.
         assert!(receivers.contains_key(ufvk.default_address().0.transparent().unwrap()));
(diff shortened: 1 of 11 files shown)


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/87ddeae1b36d43d50deda76928237cb2d5195067..84a5d5ca94dfb733b57d799d949b466cb27e7d6c) to fix clippy lints.

zcash_client_sqlite/src/testing.rs
```diff
@@ -25,19 +46,452 @@ use zcash_primitives::{
...
+        propose_transfer::<_, _, _, Infallible>(
+            &mut self.db_data,
+            &self.params,
+            spend_from_account,
+            input_selector,
+            request,
+            min_confirmations,
+        )
+    }
+
```
nuttycom, line 424: `propose_shielding` should also be supported here.

zcash_client_sqlite/src/testing.rs
```diff
@@ -0,0 +1,814 @@
...
+    }
+}
+
+impl<Cache> TestBuilder<Cache> {
+    /// Gives the test knowledge of the wallet seed for initialization.
+    pub(crate) fn with_seed(mut self, seed: SecretVec<u8>) -> Self {
+        // TODO remove
+        self.seed = Some(seed);
+        self
+    }
```
str4d, line 125: I have a feeling that this can be removed, but I included it because some tests didn't provide a seed for `init_wallet_db` while others explicitly provided a `vec![]`, and I don't remember why the latter are doing that (there's probably some migration-related reason, though that seems odd for blank databases).

zcash_client_sqlite/src/testing.rs
```diff
@@ -25,19 +46,452 @@ use zcash_primitives::{
...
+            latest_cached_block: None,
+            _data_file: data_file,
+            db_data,
+            test_account,
+        }
+    }
+}
+
+/// A `zcash_client_sqlite` test runner.
+pub(crate) struct TestRunner<Cache> {
```
nuttycom, line 159: This seems somewhat misnamed, because it doesn't actually run a test. Perhaps `TestContext` instead?

zcash_client_sqlite/src/lib.rs
```diff
@@ -1087,77 +1087,52 @@ extern crate assert_matches;
...
-    use super::BlockDb;
+    use crate::testing::AddressType;
 
     #[test]
     pub(crate) fn get_next_available_address() {
-        use tempfile::NamedTempFile;
-
-        let data_file = NamedTempFile::new().unwrap();
-        let mut db_data = WalletDb::for_path(data_file.path(), network()).unwrap();
+        let mut test = TestBuilder::new().with_test_account().build();
```
nuttycom, line 1103: If my `TestContext` suggestion is accepted, I'd like these variables to be named `ctx`.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1451,63 +1162,47 @@ pub(crate) mod tests {
...
         let dfvk = usk.sapling().to_diversifiable_full_viewing_key();
-        let uaddr = db_data.get_current_address(account_id).unwrap().unwrap();
+        let uaddr = test
+            .wallet()
+            .get_current_address(account_id)
+            .unwrap()
+            .unwrap();
         let taddr = uaddr.transparent().unwrap();
 
+        // Ensure that the wallet has at least one block
```
nuttycom, line 1178: ~This change in behavior of the test seems innocuous, but the comment should elaborate a little on why it's necessary.~ realized later that this is a move.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1082,317 +1085,33 @@ extern crate assert_matches;
...
     #[cfg(feature = "unstable")]
-    use super::{
-        chain::{init::init_blockmeta_db, BlockMeta},
+    use crate::{
+        chain::init::init_blockmeta_db,
+        testing::{fake_compact_block, store_in_fsblockdb, AddressType},
         FsBlockDb,
     };
 
-    #[cfg(feature = "mainnet")]
```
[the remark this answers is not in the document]
nuttycom: Never mind; it'll be easier to just rewrite anyway.

zcash_client_sqlite/src/testing.rs
nuttycom: Account initialization will also change with the `AccountBirthday` addition in #915, so it may be able to be removed after that.

zcash_client_sqlite/src/lib.rs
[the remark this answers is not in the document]
str4d: When I initially performed the move, I thought I'd end up moving more of those methods _into_ `TestBuilder` / `TestRunner`. But it ended up being simpler to just call the existing methods (except for `insert_into_cache` etc. that were very small). I could probably remove the first commit and still have things work (it would be some effort, but doable).

zcash_client_sqlite/src/testing.rs
str4d: Yeah; I only added methods that corresponded to things I was migrating out of tests, which indicates that we don't currently have a test for this.

zcash_client_sqlite/src/testing.rs
str4d: I initially thought it might end up being more of a test runner / driver (with the wallet available through a closure similar to what you did in #915), but it ended up being a much simpler change to have it be a "test oracle". I'd weakly prefer `TestOracle` over `TestContext` myself, as it's not a passive context but an active participant in the test.

zcash_client_sqlite/src/wallet/sapling.rs
str4d: Yeah, AFAICT the test was not relying on the ordering here; it was just pretending that the UTXO was received in the fake block without actually doing so. This makes the height of the latter accessible to the former.

zcash_client_sqlite/src/testing.rs
str4d: Going with `TestState`.

zcash_client_sqlite/src/lib.rs
str4d: Going with `st` following the `TestState` rename.

nuttycom (approved):
utACK

PR #926, opened 2023-08-30
Debugging improvements

str4d:


ebfull (approved):

ebfull:
Might want to list it as a change in the changelog on the crate since adding variants to public enums is not backwards compatible. Can be done in another PR or during the release though.

str4d:
> Might want to list it as a change in the changelog on the crate since adding variants to public enums is not backwards compatible. Can be done in another PR or during the release though.

The `enum` being modified is not public, so that's not an issue here. And AFAICT the other changes are just improvements to what `Debug` prints out, not changing whether or not it exists (and the precise format of `Debug` is not part of the public API).

PR #928, opened 2023-08-31
zcash_client_backend: Reduce scope of `lightwalletd-tonic` feature flag

str4d:
The `zcash_client_backend::proto::service::compact_tx_streamer_client` is now the only module controlled by that feature flag, exposing the service types for use by parsers.

### DIFF
--- a/zcash_client_backend/build.rs
+++ b/zcash_client_backend/build.rs
@@ -38,45 +38,43 @@ fn build() -> io::Result<()> {
         "src/proto/compact_formats.rs",
     )?;
 
-    {
-        // Build the gRPC types and client.
-        tonic_build::configure()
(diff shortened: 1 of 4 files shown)


nuttycom (dismissed):
utACK

zcash_client_backend/build.rs
```diff
@@ -40,11 +38,14 @@ fn build() -> io::Result<()> {
         "src/proto/compact_formats.rs",
     )?;
 
-    #[cfg(feature = "lightwalletd-tonic")]
     {
```
nuttycom, line 41: nit: these braces are now superfluous.

nuttycom (approved):
re-utACK 3272f12

PR #932, opened 2023-09-02
Fix `update_chain_tip` bugs

str4d:


zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -748,82 +758,115 @@ pub(crate) fn update_chain_tip<P: consensus::Parameters>(
...
+            // range to something else.
+            wallet_birthday.map_or_else(
+                // We don't have a wallet birthday, which means we have no accounts yet.
+                // We can therefore ignore all blocks up to the chain tip.
+                || ScanRange::from_parts(sapling_activation..chain_end, ScanPriority::Ignored),
+                // We have a wallet birthday, so mark all blocks between that and the
+                // chain tip as `Historic` (performing wallet recovery).
+                |wallet_birthday| {
+                    ScanRange::from_parts(wallet_birthday..chain_end, ScanPriority::Historic)
+                },
```
daira, line 779: Shouldn't we also be marking blocks between Sapling activation and the wallet birthday as `Ignored` when we have a birthday?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -715,8 +717,17 @@ pub(crate) fn update_chain_tip<P: consensus::Parameters>(
     params: &P,
     new_tip: BlockHeight,
 ) -> Result<(), SqliteClientError> {
+    // If the caller provided a chain tip that is before Sapling activation, do nothing.
+    let sapling_activation = match params.activation_height(NetworkUpgrade::Sapling) {
+        Some(h) => h,
+        None => return Ok(()),
+    };
+
```
daira, line 725: This doesn't directly check whether `new_tip` is before Sapling activation because it doesn't use `new_tip`, so the comment is confusing.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -748,82 +758,115 @@ pub(crate) fn update_chain_tip<P: consensus::Parameters>(
...
+            wallet_birthday.map_or_else(
+                // We don't have a wallet birthday, which means we have no accounts yet.
+                // We can therefore ignore all blocks up to the chain tip.
+                || ScanRange::from_parts(sapling_activation..chain_end, ScanPriority::Ignored),
+                // We have a wallet birthday, so mark all blocks between that and the
+                // chain tip as `Historic` (performing wallet recovery).
+                |wallet_birthday| {
+                    ScanRange::from_parts(wallet_birthday..chain_end, ScanPriority::Historic)
+                },
+            )
```
daira, line 780: Shouldn't we also be marking blocks between Sapling activation and the wallet birthday as `Ignored` when we have a birthday?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -748,82 +758,115 @@ pub(crate) fn update_chain_tip<P: consensus::Parameters>(
...
+                    // This overlaps the `tip_shard_entry` range and so will be coalesced
+                    // with it.
+                    ScanRange::from_parts(min_unscanned..chain_end, ScanPriority::ChainTip)
+                } else {
+                    // In this case, the max scanned height is considered stable relative
+                    // to the chain tip. However, it may be stable or unstable relative to
+                    // the prior chain tip, which we could determine by looking up the
+                    // prior chain tip height from the scan queue. For simplicity we merge
+                    // these two cases together, and pretend that the max scanned block is
+                    // always unstable relative to the prior chain tip.
```
daira, line 814:
```suggestion
                    // these two cases together, and proceed as though the max scanned
                    // block is unstable relative to the prior chain tip.
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -748,82 +758,115 @@ pub(crate) fn update_chain_tip<P: consensus::Parameters>(
...
+                    //
+                    // To confirm its stability, prioritize the `VERIFY_LOOKAHEAD` blocks
+                    // above the max scanned height as `Verify`:
+                    //
+                    // - We use `Verify` to ensure this connectivity check is performed,
```
daira, line 819:
```suggestion
                    // - We use `Verify` to ensure that a connectivity check is performed,
```

daira (approved):
utACK.

Did you also want to address my late comments on #907 in this PR, or will that be a separate PR?

zcash_client_sqlite/src/wallet/scanning.rs
str4d: That happens when the account is created.

nuttycom (approved):
utACK modulo @daira's suggested comment corrections

zcash_client_sqlite/src/wallet/scanning.rs
str4d: This is a bug: the old code did check this condition, but in the refactor the comment was moved without fully moving the logic.

str4d:
> Did you also want to address my late comments on #907 in this PR, or will that be a separate PR?

We should do those in a separate PR (as this PR is fixing specific bugs).

nuttycom (approved):
re-utACK f132af1f3350813f185a2c5c1eaf31a659785fe5

nuttycom:
> > Did you also want to address my late comments on #907 in this PR, or will that be a separate PR?
> 
> We should do those in a separate PR (as this PR is fixing specific bugs).

Added #935

PR #935, opened 2023-09-05
Address unresovled code review comments from #907

nuttycom:


zcash_client_sqlite/src/error.rs
```diff
@@ -135,9 +135,9 @@ impl fmt::Display for SqliteClientError {
             SqliteClientError::BlockConflict(h) => write!(f, "A block hash conflict occurred at height {}; rewind required.", u32::from(*h)),
             SqliteClientError::NonSequentialBlocks => write!(f, "`put_blocks` requires that the provided block range be sequential"),
             SqliteClientError::DiversifierIndexOutOfRange => write!(f, "The space of available diversifier indices is exhausted"),
-            SqliteClientError::AccountUnknown(id) => write!(f, "Account {} does not belong to this wallet.", u32::from(*id)),
+            SqliteClientError::AccountUnknown(acct_id) => write!(f, "Account {:?} does not belong to this wallet.", u32::from(*acct_id)),
```
str4d, line 138: Why the `{:?}` change? This prints the `Debug` impl for `u32`, not for `acct_id`.

zcash_client_sqlite/src/error.rs
```diff
@@ -135,9 +135,9 @@ impl fmt::Display for SqliteClientError {
...
 
-            SqliteClientError::KeyDerivationError(acct_id) => write!(f, "Key derivation failed for account {:?}", acct_id),
+            SqliteClientError::KeyDerivationError(acct_id) => write!(f, "Key derivation failed for account {:?}", u32::from(*acct_id)),
```
str4d, line 140: Probably the same issue here: moved from `Debug` impl of `acct_id` to that of `u32`. If the `u32` change is intentional, then use the `Display` impl for both:
```suggestion
            SqliteClientError::KeyDerivationError(acct_id) => write!(f, "Key derivation failed for account {}", u32::from(*acct_id)),
```

PR #938, opened 2023-09-05
Check for potential discontinuities when inserting into the `*_tree_shards` table

### LINKED ISSUE #924: zcash_client_sqlite: Ensure shard index continuity in `put_shard_roots`
> Additionally, I think we should add continuity checks for places that cause the `sapling_tree_shards` table to be modified. If we end up in a situation where the table has discontinuities in `shard_index`, this line will create rows that have incorrect and overlapping shard tree ranges, which may cause subsequent logic (that assumes correct ranges) to fail in weird and annoying ways. Functionally, this means that anyone calling `update_subtree_roots` must do so in a way that doesn't add discontinuities (as adding a root causes the entire shard to be serialized if it didn't exist).

_Originally posted by @str4d in https://github.com/zcash/librustzcash/pull/907#discussion_r1309402735_

ebfull:
Closes #924

### DIFF
--- a/zcash_client_sqlite/src/wallet/commitment_tree.rs
+++ b/zcash_client_sqlite/src/wallet/commitment_tree.rs
@@ -4,6 +4,7 @@ use std::{
     error, fmt,
     io::{self, Cursor},
     marker::PhantomData,
+    ops::Range,
     sync::Arc,
 };
 use zcash_client_backend::data_api::chain::CommitmentTreeRoot;
@@ -35,6 +36,13 @@ pub enum Error {
         extant_tree_state: TreeState,
         extant_marks_removed: Option<BTreeSet<Position>>,
     },
+    /// Raised when attempting to add shard roots to the database that
+    /// are discontinuous with the existing roots in the database.
+    SubtreeDiscontinuity {
+        attempted_insertion_range: Range<u64>,
+        min: u64,
+        max: u64,
+    },
 }
 
 impl fmt::Display for Error {
@@ -54,6 +62,17 @@ impl fmt::Display for Error {
                     checkpoint_id, checkpoint, extant_tree_state, extant_marks_removed
                 )
             }
+            Error::SubtreeDiscontinuity {
+                attempted_insertion_range,
+                min,
+                max,
+            } => {
+                write!(
+                    f,
+                    "Attempted to write subtree roots with indicies {:?} which is discontinuous with existing subtree range {:?}",
+                    attempted_insertion_range, *min..(*max+1)
+                )
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -54,6 +62,17 @@ impl fmt::Display for Error {
...
             }
+            Error::SubtreeDiscontinuity {
+                attempted_insertion_range,
+                min,
+                max,
+            } => {
+                write!(
+                    f,
+                    "Attempted to write subtree roots with indicies {:?} which is discontinuous with existing subtree range {:?}",
+                    attempted_insertion_range, *min..(*max+1)
```
str4d, line 73: Nit: we could just store `min..(max+1)` as a range.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -367,6 +387,46 @@ pub(crate) fn last_shard<H: HashSer>(
...
+    {
+        if !proposed_insertion_range.contains(&min) && !proposed_insertion_range.contains(&max) {
+            // The proposed insertion range does not overlap with the existing shard indicies.
+            // This means a discontinuity is introduced if the proposed insertion range's start
+            // is not `max + 1` _and_ `min` is not the proposed insertion range's end.
+            if max + 1 != proposed_insertion_range.start && min != proposed_insertion_range.end {
+                return Err(Error::SubtreeDiscontinuity {
+                    attempted_insertion_range: proposed_insertion_range,
+                    min,
+                    max,
```
str4d, line 421: If my nit were acted upon:
```suggestion
                    existing_range: min..(max+1),
```

str4d (approved):
utACK 976b0f093153ff8f72aaae00066105c9ce6f2025 with tiny nit.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -367,6 +387,46 @@ pub(crate) fn last_shard<H: HashSer>(
...
+            [],
+            |row| {
+                let min = row.get::<_, Option<u64>>(0)?;
+                let max = row.get::<_, Option<u64>>(1)?;
+                Ok((min, max))
+            },
+        )
+        .map_err(Error::Query)
+    {
+        if !proposed_insertion_range.contains(&min) && !proposed_insertion_range.contains(&max) {
```
str4d, line 413: There was a bug here, that my nit was pointing towards (but I reviewed this at midnight so I didn't spot it then): This checks for the proposed insertion range containing the existing range, but not for the reverse (which is far more common). We need to actually check that the ranges do not overlap before proceeding to the adjacency check below.

PR #942, opened 2023-09-06
zcash_client_backend: Add `WalletRead::block_max_scanned`

### LINKED ISSUE #941: zcash_client_backend: Add `WalletRead::max_scanned_height`
When a wallet detects that it is necessary to perform a reorg by observing that scanning has failed, it's necessary for the wallet to be able to determine the last scanned height in order to figure out where they need to attempt to reorg to.

nuttycom:
Fixes #941

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -1086,6 +1086,33 @@ pub(crate) fn block_fully_scanned(
     }
 }
 
+pub(crate) fn block_max_scanned(
+    conn: &rusqlite::Connection,
+) -> Result<Option<BlockMetadata>, SqliteClientError> {
+    conn.query_row(
+        "SELECT blocks.height, hash, sapling_commitment_tree_size, sapling_tree
+         FROM blocks
+         JOIN (SELECT MAX(height) AS height FROM blocks) blocks_max
+         ON blocks.height = blocks_max.height",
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -1086,6 +1086,33 @@ pub(crate) fn block_fully_scanned(
...
 }
 
+pub(crate) fn block_max_scanned(
+    conn: &rusqlite::Connection,
+) -> Result<Option<BlockMetadata>, SqliteClientError> {
+    conn.query_row(
+        "SELECT blocks.height, hash, sapling_commitment_tree_size, sapling_tree
+         FROM blocks
+         JOIN (SELECT MAX(height) AS height FROM blocks) blocks_max
+         ON blocks.height = blocks_max.height",
```
str4d, line 1096: Wouldn't it be more performant to use `ORDER BY height DESC LIMIT 1`?

str4d (approved):
utACK with question.

zcash_client_sqlite/src/wallet.rs
nuttycom: I honestly have no idea about the relative performance profiles of these two queries. It depends upon whether `ORDER BY` is lazy, and how `MAX` interacts with the index on `height`.

PR #943, opened 2023-09-06
zcash_client_sqlite: Fix bug in `check_shard_discontinuity`

str4d:
The previous code was checking that the existing shard range was at least partially inside the proposed insertion range, but this didn't handle the case where the proposed insertion range was fully contained by the existing shard range.


ebfull (approved):

nuttycom (approved):
utACK

PR #944, opened 2023-09-06
Measure scan progress from wallet birthday, not the fully scanned height

nuttycom:


zookozcash:
Not to go into SDK 2.0 Release Candidate, at least pending my imminent phone call with Kris.

str4d:
The motivation for this change is that the current approach looks exponential, which gives the impression to users during recovery from seed or after long periods offline, of very little progress. Here are some snapshots from my mainnet node:
```
    chain_tip_height: BlockHeight(
        2217761,
    ),
    fully_scanned_height: BlockHeight(
        1373035,
    ),
    scan_progress: Some(
        Ratio {
            numerator: 48807,
            denominator: 72279746,
        },
    ),
     Synced: 0.068%
[...]
    chain_tip_height: BlockHeight(
        2217766,
    ),
    fully_scanned_height: BlockHeight(
        1722754,
    ),
    scan_progress: Some(
        Ratio {
            numerator: 48818,
            denominator: 71762038,
        },
    ),
     Synced: 0.068%
[...]
    chain_tip_height: BlockHeight(
        2217771,
    ),
    fully_scanned_height: BlockHeight(
        1732754,
    ),
    scan_progress: Some(
        Ratio {
            numerator: 48831,
            denominator: 67203704,
        },
    ),
     Synced: 0.073%
[...]
    chain_tip_height: BlockHeight(
        2217785,
    ),
    fully_scanned_height: BlockHeight(
        1772754,
    ),
    scan_progress: Some(
        Ratio {
            numerator: 48867,
            denominator: 39559642,
        },
    ),
     Synced: 0.124%
```

zookozcash:
Update: Per phone call with Kris and str4d’s notes above in this PR, current implementation does _not_ satisfy Criterion 3, because under the (probably common and important use case) that the user hasn’t let their wallet sync in months or years, it would show no visible progress for a long time (like at least minutes, potentially hours). So I’m reversing the above decision to exclude this change to progress measure for 2.0.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -524,21 +524,22 @@ impl ScanProgress for SubtreeScanProgress {
...
-            // Compute the number of fully scanned notes directly from the blocks table
-            let fully_scanned_size = conn.query_row(
+            let start_height = birthday_height;
+            // Compute the starting number of notes directly from the blocks table
+            let start_size = conn.query_row(
                 "SELECT MAX(sapling_commitment_tree_size)
                  FROM blocks
-                 WHERE height <= :fully_scanned_height",
-                named_params![":fully_scanned_height": u32::from(fully_scanned_height)],
+                 WHERE height <= :start_height",
```
str4d, line 532: Nit: now that `start_height = birthday_height` there's an off-by-one here, because the birthday height is one of the blocks that gets scanned, not the base block.

Non-blocking, the error in the output is minimal, and it isn't worth re-running CI.

str4d (approved):
ACK f56c5d8f4b191cc365eb753d87d37574529b8ce7. Output appears to be more comprehensible for the catching-up-to-chain-tip use case.

str4d:
Output 10k blocks later:
```
    chain_tip_height: BlockHeight(
        2217817,
    ),
    fully_scanned_height: BlockHeight(
        1782754,
    ),
    scan_progress: Some(
        Ratio {
            numerator: 38544695,
            denominator: 72853741,
        },
    ),
     Synced: 52.907%
```

daira (commented):
Post-hoc ACK

PR #946, opened 2023-09-06
zcash_client_sqlite: Add a test demonstrating chain tip fund unavailability

nuttycom:
This also factors out and cleans up repeated code that checks for anchor shard completeness.

### DIFF
--- a/zcash_client_sqlite/src/wallet/sapling.rs
+++ b/zcash_client_sqlite/src/wallet/sapling.rs
@@ -126,6 +126,30 @@ fn to_spendable_note(row: &Row) -> Result<ReceivedSaplingNote<ReceivedNoteId>, S
     })
 }
 
+/// Utility method for determining whether we have any spendable notes
+///
+/// If the tip shard has unscanned ranges below the anchor height and greater than or equal to
+/// the wallet birthday, none of our notes can be spent because we cannot construct witnesses at
+/// the provided anchor height.
+fn unscanned_tip_exists(
+    conn: &Connection,
+    anchor_height: BlockHeight,
+) -> Result<bool, rusqlite::Error> {
+    // v_sapling_shard_unscanned_ranges only returns ranges ending on or after wallet birthday, so
+    // we don't need to refer to the birthday in this query.
+    conn.query_row(
+        "SELECT EXISTS (
+             SELECT 1 FROM v_sapling_shard_unscanned_ranges range
+             WHERE range.block_range_start <= :anchor_height
+             AND :anchor_height BETWEEN 
+                range.subtree_start_height 
+                AND IFNULL(range.subtree_end_height, :anchor_height)
+         )",
+        named_params![":anchor_height": u32::from(anchor_height),],
+        |row| row.get::<_, bool>(0),
+    )
+}
+
 pub(crate) fn get_spendable_sapling_notes(
     conn: &Connection,
     account: AccountId,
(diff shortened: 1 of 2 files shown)


zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -126,6 +126,28 @@ fn to_spendable_note(row: &Row) -> Result<ReceivedSaplingNote<ReceivedNoteId>, S
     })
 }
 
+/// Utility method for determining whether we have any spendable notes
+///
+/// If the tip shard has unscanned ranges below the anchor height and greater than or equal to
+/// the wallet birthday, none of our notes can be spent because we cannot construct witnesses at
```
str4d, line 132: This function does not query or take as a parameter the wallet birthday. Why is it referenced here?

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1203,4 +1214,91 @@ pub(crate) mod tests {
...
+    fn birthday_in_anchor_shard() {
+        let (mut st, dfvk, birthday, _) = test_with_canopy_birthday();
+
+        // Set up the following situation:
+        //
+        //        |<------ 500 ------->|<--- 10 --->|<--- 10 --->|
+        // last_shard_start   wallet_birthday  received_tx  anchor_height
+        //
+        // Set up some shard root history before the wallet birthday.
+        let second_to_last_shard_start = birthday.height() - 500;
```
str4d, line 1228:
```suggestion
        let last_shard_start = birthday.height() - 500;
```

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1203,4 +1214,91 @@ pub(crate) mod tests {
...
+        st.wallet_mut()
+            .put_sapling_subtree_roots(
+                0,
+                &[CommitmentTreeRoot::from_parts(
+                    second_to_last_shard_start,
```
str4d, line 1233:
```suggestion
                    last_shard_start,
```

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1203,4 +1214,91 @@ pub(crate) mod tests {
...
+            Amount::const_from_i64(500000),
+        );
+
+        // Generate some more blocks to get above our anchor height
+        for _ in 0..15 {
+            st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+        }
+
+        // Scan all the blocks since the wallet birthday
+        st.scan_cached_blocks(birthday.height() + 5, 25);
```
str4d, line 1275:
```suggestion
        // Scan a block range that includes our received note, but skips some blocks we
        // need to make it spendable.
        st.scan_cached_blocks(birthday.height() + 5, 20);
```

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: This is where I *thought* a bug might be, but `v_sapling_shard_unscanned_ranges` is what is actually bounding on the wallet birthday. I'll add a comment in the body to that effect.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/2a5c12cc8b1c2c9b060107bca9b43b2678a53082..0937c6a38407a9c98f546c43dc96ffdc6dece48e) to address comments from code review.

str4d (approved):
utACK 0937c6a38407a9c98f546c43dc96ffdc6dece48e

PR #951, opened 2023-09-07
zcash_client_sqlite: Return a balance for every account from `get_wallet_summary`

### LINKED ISSUE #948: zcash_client_sqlite: Ensure that `WalletSummary` has a balance for every account in the wallet.

nuttycom:
Fixes #948

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -73,7 +73,7 @@ use std::convert::TryFrom;
 use std::io::{self, Cursor};
 use std::num::NonZeroU32;
 use tracing::debug;
-use zcash_client_backend::data_api::{AccountBalance, Balance, Ratio, WalletSummary};
+use zcash_client_backend::data_api::{AccountBalance, Ratio, WalletSummary};
 use zcash_primitives::transaction::components::amount::NonNegativeAmount;
 
 use zcash_client_backend::data_api::{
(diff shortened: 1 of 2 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -689,27 +697,17 @@ pub(crate) fn get_wallet_summary(
...
-                    + spendable_value)
-                    .expect("Spendable value cannot overflow");
-                bal.sapling_balance.change_pending_confirmation =
-                    (bal.sapling_balance.change_pending_confirmation + change_pending_confirmation)
-                        .expect("Pending change value cannot overflow");
-                bal.sapling_balance.value_pending_spendability =
-                    (bal.sapling_balance.value_pending_spendability + value_pending_spendability)
-                        .expect("Value pending spendability cannot overflow");
-            })
-            .or_insert(AccountBalance {
```
str4d, line 705: Non-blocking: we could probably also remove the `or_insert` from the transparent part.

str4d (approved):
utACK 0771cb26d818263165e614415a24ddd9b441bfd4

zcash_client_sqlite/src/wallet.rs
nuttycom: Fixed.

str4d (approved):
utACK 2131cd547e56b9de836d431c72890d41e7851837

PR #953, opened 2023-09-08
Fix a veritable cornucopia of bugs

str4d:


nuttycom (approved):
utACK 504efcfab77bf65583f2d88daf2447c2df5547b5. This has been tested using @str4d's mainnet wallet.

str4d:
`rustfmt` appears to be toggling a bit between formatting states across a few PRs. We'll try fixing it uniformly in a subsequent PR.

daira (commented):
Post-hoc utACK.

Shouldn't we have regression tests for bugs like this, though? (Don't worry, I'm not saying yous should have written one in the 4.5-hour debugging session! Sympathy.)

PR #963, opened 2023-09-11
zcash_client_backend: Fix an error that can occur due to sparse `shardtree` checkpoints.

nuttycom:


zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1304,4 +1304,67 @@ pub(crate) mod tests {
...
+    fn checkpoint_gaps() {
+        let mut st = TestBuilder::new()
+            .with_block_cache()
+            .with_test_account(AccountBirthday::from_sapling_activation)
+            .build();
+
+        let (account, usk, birthday) = st.test_account().unwrap();
+        let dfvk = st.test_account_sapling().unwrap();
+
+        // Generate a block that with funds belonging to our wallet.
```
str4d, line 1318:
```suggestion
        // Generate a block with funds belonging to our wallet.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -867,6 +867,13 @@ pub trait WalletCommitmentTrees {
         Error = Self::Error,
     >;
 
+    /// Returns the depth of the checkpoint in the tree that can be used to create
+    /// a witness at the anchor having the given number of confirmations.
```
str4d, line 871: An artifact of this method is that some notes that might previously have been spendable may become not so, if the notes were mined between the chosen checkpoint depth and the desired anchor height, and the checkpoint heights are for whatever reason non-contiguous in terms of tree updates (rather than just non-contiguous in checkpoint IDs). Perhaps the checkpoint ID corresponding to this depth should be returned as well?

That does complicate the API a bit though, so maybe instead we just document that this method relies on the caller ensuring that checkpoints exist for every change to the tree, even if there aren't checkpoints for every block.

zcash_client_backend/src/data_api.rs
```diff
@@ -1182,5 +1189,12 @@ pub mod testing {
 
             Ok(())
         }
+
+        fn get_checkpoint_depth(
+            &self,
+            min_confirmations: NonZeroU32,
+        ) -> Result<usize, ShardTreeError<Self::Error>> {
+            Ok(usize::try_from(u32::from(min_confirmations) - 1).unwrap())
```
str4d, line 1197: Is the `- 1` here correct? I think `min_confirmations` is actually equal to the checkpoint depth, as they are both 1-indexed relative to the chain (`min_confirmations = 0` is "unmined, in the mempool").
```suggestion
            Ok(usize::try_from(u32::from(min_confirmations)).unwrap())
```

(The `- 1` in `get_checkpoint_depth` is correct, because there we're deriving an absolute height relative to the chain tip, so it is for adjusting from 1-indexing to 0-indexing.)

zcash_client_sqlite/src/lib.rs
```diff
@@ -835,6 +847,18 @@ impl<'conn, P: consensus::Parameters> WalletCommitmentTrees for WalletDb<SqlTran
...
+
+    fn get_checkpoint_depth(
+        &self,
+        min_confirmations: NonZeroU32,
+    ) -> Result<usize, ShardTreeError<Self::Error>> {
+        get_checkpoint_depth(&self.conn.0, min_confirmations)
+            .map_err(|e| ShardTreeError::Storage(commitment_tree::Error::Query(e)))?
+            // `CheckpointPruned` is perhaps a little misleading; in this case it's that
+            // the chain tip is unknown, but if that were the case we should never have been
+            // calling this anyway.
```
str4d, line 859: Does this comment apply to the other `WalletCommitmentTrees` impl as well?

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -747,6 +750,29 @@ pub(crate) fn get_checkpoint(
...
+        .map(|chain_tip| {
+            let max_checkpoint_height =
+                u32::from(chain_tip).saturating_sub(u32::from(min_confirmations) - 1);
+
+            // We exclude from consideration all checkpoints having heights greater than the maximum
+            // checkpoint height. The checkpoint depth is the number of excluded checkpoints + 1.
+            conn.query_row(
+                "SELECT COUNT(*) 
+             FROM sapling_tree_checkpoints
+             WHERE checkpoint_id > :max_checkpoint_height",
```
str4d, line 768:
```suggestion
                 FROM sapling_tree_checkpoints
                 WHERE checkpoint_id > :max_checkpoint_height",
```

str4d (changes requested):
utACK after fixes.

zcash_client_backend/src/data_api.rs
nuttycom: If the note was mined in between, then there'll be a checkpoint for it that is less than the anchor height, and there's not a problem, right?

zcash_client_sqlite/src/lib.rs
nuttycom: Ah, yeah, it's the same.

zcash_client_backend/src/data_api.rs
str4d: I mean mined between the anchor height and checkpoint ID:
```
chain_tip
 |
 | (10 blocks)
 |
anchor_height
 |
note_mined
 |
checkpoint_id
```
Put another way, the note is selected assuming an anchor will be built to `anchor_height`, while this code now causes it to be built at `checkpoint_id`. If those two are not the same height, there's a potential for errors. We can get around this problem by essentially documenting that we assume no notes ever get mined between `checkpoint_id` and `anchor_height`.

str4d (approved):
utACK e5af0175104b4e76c5c67e971dfc82a1828836e2

zcash_client_backend/src/data_api.rs
nuttycom: If `checkpoint_id` is at a height less than `note_mined`, then the note isn't spendable because the wallet has never observed it as having been mined.

zcash_client_sqlite/src/lib.rs
```diff
@@ -795,6 +798,18 @@ impl<P: consensus::Parameters> WalletCommitmentTrees for WalletDb<rusqlite::Conn
...
+
+    fn get_checkpoint_depth(
+        &self,
+        min_confirmations: NonZeroU32,
+    ) -> Result<usize, ShardTreeError<Self::Error>> {
+        get_checkpoint_depth(&self.conn, min_confirmations)
+            .map_err(|e| ShardTreeError::Storage(commitment_tree::Error::Query(e)))?
+            // `CheckpointPruned` is perhaps a little misleading; in this case it's that
+            // the chain tip is unknown, but if that were the case we should never have been
+            // calling this anyway.
```
daira, line 810: This seems like a poor argument. Just add another error case, if that's not too difficult.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -1304,4 +1304,70 @@ pub(crate) mod tests {
...
+            Amount::const_from_i64(300000),
+            birthday.height() + 5,
+            &[],
+        )
+        .unwrap();
+        assert_eq!(spendable.len(), 1);
+
+        // Attempt to spend the note with 5 confirmations
+        let to = not_our_key.default_address().1.into();
+        assert_matches!(
```
daira, line 1361: In #908 I change tests like this to more thoroughly check that the spend succeeds (e.g. see https://github.com/zcash/librustzcash/pull/908/files#diff-9cea07a2b90189d927a2737a964494dfea6ab1bc5a51342e481210b93586716eR812-R835 ). I'd really like to get that PR in soon, so that we don't keep adding tests in the less thorough style.

daira (dismissed):
utACK

zcash_client_sqlite/src/lib.rs
nuttycom: It would require a new release of `shardtree`

zcash_client_backend/src/data_api.rs
str4d: In a pairing, we clarified that this is assuming that when a note is witnessed, the block it was witnessed in is being checkpointed. Under that assumption, I agree.

daira: Yes that is an assumption I knew we were making. It should be explicitly documented if it isn't already, though.

zcash_client_sqlite/src/lib.rs
daira: Ok that's a good reason not to.

str4d (approved):
utACK af653519a69529932f0903121ea301b50a8e9363

daira (approved):
utACK, since the current commit is the one I had already reviewed (f2dcba3a34bb5f00ec670f478029621201529b2d).

str4d (approved):
utACK f2dcba3a34bb5f00ec670f478029621201529b2d. Please open an issue for the backwards-sliding problem that is not being addressed here (due to testnet compatibility problems).

nuttycom:
> utACK [f2dcba3](https://github.com/zcash/librustzcash/commit/f2dcba3a34bb5f00ec670f478029621201529b2d). Please open an issue for the backwards-sliding problem that is not being addressed here (due to testnet compatibility problems).

Added https://github.com/zcash/librustzcash/issues/968 to track this.

PR #964, opened 2023-09-11
Add `Cargo.lock` to repository

str4d:
Going forward will be the lockfile we use for testing our MSRV. See https://blog.rust-lang.org/2023/08/29/committing-lockfiles for rationale on this change.


nuttycom (approved):
utACK

str4d:
Checking in a lockfile means we don't see non-MSRV breakages in otherwise-compatible updates. In a subsequent PR I will change CI to also compile with "stable Rust and latest dependencies", so we catch these.

PR #965, opened 2023-09-11
zcash_client_sqlite: `serialization` module should not have been public.

nuttycom:


str4d (approved):
utACK 771e4013c3f7affcfc05f351cde4a2f9cf20058a

PR #966, opened 2023-09-11
CI: Add job to build with latest Rust and dependencies

str4d:
This ensures that we catch non-MSRV compilation breakages caused by SemVer updates.


nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK

PR #969, opened 2023-09-12
zcash_client_backend 0.10.0-rc.2 and zcash_client_sqlite 0.8.0-rc.2

nuttycom:


str4d (approved):
utACK e6b34eeaa2861a7fb3e7f6cdc1748481877e6fff. Technically the changelog versions should also be rewritten, but I'm not going to block on that (as they will be rewritten anyway in the final release).

PR #973, opened 2023-09-16
zcash_client_backend: make the `SpanningTree` type usable outside of `zcash_client_sqlite`

nuttycom:
This adds the `data_api::scanning::spanning_tree` module under a new `unstable-spanning-tree` feature flag, making it available to other implementations who want to be able to write their own storage backends without having to reinvent the spanning tree logic.


daira (approved):
utACK. I did not check that the moved code is (essentially) unchanged.

str4d (approved):

str4d:
I used `git diff --color-moved=dimmed-zebra release-lwsdk-2.0.0..nuttycom/refactor/backend_spanning_tree` to check that the move parts were move-only.

PR #974, opened 2023-09-18
zcash_client_sqlite: Remove inner join between `transactions` and `utxos` from `v_transactions`

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -1461,11 +1460,12 @@ pub(crate) fn put_block(
...
 
 /// Inserts information about a mined transaction that was observed to
-/// contain a note related to this wallet into the database.
-pub(crate) fn put_tx_meta<N>(
+/// contain a note or utxo related to this wallet into the database.
+pub(crate) fn put_tx_meta(
     conn: &rusqlite::Connection,
-    tx: &WalletTx<N>,
     height: BlockHeight,
+    tx_index: Option<u32>,
```
str4d, line 1467: This is a technically incorrect change: mined transactions always have a `tx_index`, so it is a type error for `height` to be set and `tx_index` to be `None`.

`tx_index` is exposed via the `v_transactions` view, so it is part of the public API. This bugfix is therefore incomplete: it changes the `v_transactions` bug from "never returns transparent transactions" to "returns transparent transactions with an incorrect `tx_index`".

However, we have a limitation here for UTXOs in that `tx_index` is not needed to create spends, so when transparent receives were added, that field was never passed across from `lightwalletd`:
https://github.com/zcash/librustzcash/blob/ba8cb2eff5afdc40d3f4ccaf58443ae3d0a9235c/zcash_client_backend/proto/service.proto#L144-L151

(We only happen to know `height` because it is necessary to determine whether a UTXO has sufficient confirmations, but we don't actually use it for that because we zero-conf spend all UTXOs, so we're just _very_ lucky in that regard.)

The correct fix would be to:
- Change `lightwalletd` to provide `tx_index`.
- Change the mobile SDKs to pass `tx_index` across the FFI.
- Change `WalletTransparentOutput` to hold `tx_index`.
- Pass it into `put_tx_meta`.

That set of changes is too complex to fit into the mobile SDK 2.0.0 release window. So we now need to ask ourselves several questions:
- Is `v_transactions.tx_index` used _anywhere_ that we support? Main places to look are the Android and iOS SDKs.
- Will it cause any non-transient problems for us to expose the wrong value of `tx_index` via a public API?
- Will it cause any non-transient problems for us to store `(height, NULL)` in the wallet DB, in addition to `(NULL, NULL)` and `(height, tx_index)`?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1478,9 +1478,9 @@ pub(crate) fn put_tx_meta<N>(
     )?;
 
     let tx_params = named_params![
-        ":txid": &tx.txid.as_ref()[..],
+        ":txid": &txid.as_ref()[..],
         ":block": u32::from(height),
```
str4d, line 1482: A test is failing because this insert fails due to no corresponding block existing in the `blocks` table. This presents a much larger problem, because we don't have any of the rest of the data needed to fill in the `blocks` table at the time we insert UTXOs (because none of it was necessary for spending UTXOs).

In a pairing, @nuttycom @daira and I decided that the current approach of this PR won't be workable for the mobile SDK 2.0.0 release, and instead we should alter the `v_transactions` view to make a bunch of its columns optional / best-effort (which is fine if the downstream users of it are only using it to display transaction history, and are not making any assumptions otherwise about the data).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -452,32 +453,36 @@ mod tests {
...
+                       2                                     AS pool,
                        -sapling_received_notes.value         AS value,
                        0                             AS is_change,
                        0                             AS received_count,
                        0                             AS memo_present
-                FROM   sapling_received_notes
+                FROM sapling_received_notes
+                JOIN transactions
+                     ON transactions.id_tx = sapling_received_notes.spent
                 WHERE  sapling_received_notes.spent IS NOT NULL
```
str4d, line 481: I think this `WHERE` is now unnecessary, given that `JOIN` is an inner join? We could make the latter explicit here to clarify we only want spent notes:
```suggestion
                INNER JOIN transactions
                     ON transactions.id_tx = sapling_received_notes.spent
```

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ and this library adheres to Rust's notion of
   - `FsBlockDbError::CacheMiss`
 - `zcash_client_sqlite::FsBlockDb::write_block_metadata` now overwrites any
   existing metadata entries that have the same height as a new entry.
+- The `v_transactions` and `v_tx_outputs` views now return the 32-byte
+  transaction identifier for transactions instead of the internal database
+  id.
```
str4d, line 47: `v_transactions` already returned the txid; the change there is only to remove `id_tx`.

Additionally, was the bug that this PR is fixing always present in `v_transactions`? I suspect so (and that users just weren't seeing it before because the mobile SDKs were still using direct DB access to provide history), in which case there should also be an entry under `### Fixed`.

zcash_client_sqlite/src/wallet/init.rs
daira, line 481:
```suggestion
```
This `WHERE` clause is unnecessary because `transactions.id_tx` is not nullable ([it's an `INTEGER PRIMARY KEY`](https://www.sqlite.org/autoinc.html)), so the inner join on `transactions` will exclude entries for which `sapling_received_notes.spent` is null.

str4d (approved):
utACK 14401cb275e813dce03aa28c8f9312e38dfcc6dc once the changelog is fixed. SQL changes look correct; my comment there is non-blocking.

zcash_client_sqlite/CHANGELOG.md
nuttycom: Yup, I think that's correct.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -487,21 +492,22 @@ mod tests {
                          END
                        ) AS memo_count
                 FROM sent_notes
+                JOIN transactions
+                     ON transactions.id_tx = sent_notes.tx
                 LEFT JOIN sapling_received_notes
                           ON (sent_notes.tx, sent_notes.output_pool, sent_notes.output_index) =
                              (sapling_received_notes.tx, 2, sapling_received_notes.output_index)
                 WHERE  sapling_received_notes.is_change IS NULL
                    OR  sapling_received_notes.is_change = 0
```
daira, line 501:
```suggestion
                WHERE COALESCE(sapling_received_notes.is_change, 0) = 0
```
Non-blocking, just a style nit.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -487,21 +492,22 @@ mod tests {
...
+                GROUP BY account_id, txid
             ),
             blocks_max_height AS (
                 SELECT MAX(blocks.height) as max_height FROM blocks
             )
             SELECT notes.account_id                  AS account_id,
-                   transactions.id_tx                AS id_tx,
-                   transactions.block                AS mined_height,
+                   notes.block                       AS mined_height,
+                   notes.txid                        AS txid,
```
daira, line 508:
```suggestion
                   notes.txid                        AS txid,
                   notes.block                       AS mined_height,
```
So that we more closely follow the release notes' "The `v_transactions` and `v_tx_outputs` views now return the 32-byte transaction identifier for transactions instead of the internal database id." (The ordering is visible.)

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ and this library adheres to Rust's notion of
   - `FsBlockDbError::CacheMiss`
 - `zcash_client_sqlite::FsBlockDb::write_block_metadata` now overwrites any
   existing metadata entries that have the same height as a new entry.
+- The `v_transactions` and `v_tx_outputs` views no longer return the
+  internal database identifier for the transaction. The `txid` column
+  should be used instead.
```
daira, line 47: Explain that some columns of `v_transactions` are now nullable.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -515,17 +519,18 @@ mod tests {
                         blocks.height IS NULL
                         AND transactions.expiry_height <= blocks_max_height.max_height
                    ) AS expired_unmined
```
daira, line 521: Comment on line 516: I think this may double-count memos for notes that are in both `sapling_received_notes` and `sent_notes`. (Pre-existing bug.)

str4d (approved):
utACK 0e0e79733de4ae91135baece06ec8c425fff095d

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -515,17 +519,18 @@ mod tests {
                         blocks.height IS NULL
                         AND transactions.expiry_height <= blocks_max_height.max_height
```
daira, line 520:
```suggestion
                        AND transactions.expiry_height <= blocks_max_height.max_height
                        AND transactions.expiry_height > 0
```
`put_tx_data` in `wallet.rs` does not map a zero expiry height ([meaning "does not expire"](https://zips.z.cash/zip-0203#specification)) to null, so this would incorrectly include unmined transactions with expiry height 0.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_transparent_history.rs
```diff
@@ -0,0 +1,187 @@
...
+                   SUM(notes.value)                  AS account_balance_delta,
+                   transactions.fee                  AS fee_paid,
+                   SUM(notes.is_change) > 0          AS has_change,
+                   MAX(COALESCE(sent_note_counts.sent_notes, 0))  AS sent_note_count,
+                   SUM(notes.received_count)         AS received_note_count,
+                   SUM(notes.memo_present) + MAX(COALESCE(sent_note_counts.memo_count, 0)) AS memo_count,
+                   blocks.time                       AS block_time,
+                   (
+                        blocks.height IS NULL
+                        AND transactions.expiry_height <= blocks_max_height.max_height
```
daira, line 124: Make sure the fix at https://github.com/zcash/librustzcash/pull/974/files#r1329258653 is copied here.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_transparent_history.rs
```diff
@@ -0,0 +1,187 @@
...
+                   notes.txid                        AS txid,
+                   transactions.tx_index             AS tx_index,
+                   transactions.expiry_height        AS expiry_height,
+                   transactions.raw                  AS raw,
+                   SUM(notes.value)                  AS account_balance_delta,
+                   transactions.fee                  AS fee_paid,
+                   SUM(notes.is_change) > 0          AS has_change,
+                   MAX(COALESCE(sent_note_counts.sent_notes, 0))  AS sent_note_count,
+                   SUM(notes.received_count)         AS received_note_count,
+                   SUM(notes.memo_present) + MAX(COALESCE(sent_note_counts.memo_count, 0)) AS memo_count,
```
daira, line 120: Make sure the fix for https://github.com/zcash/librustzcash/pull/974/files#r1329243000 (if needed) is copied here.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_transparent_history.rs
```diff
@@ -0,0 +1,187 @@
...
+                   false                          AS is_change,
+                   sent_notes.memo                AS memo
+            FROM sent_notes
+            JOIN transactions
+                 ON transactions.id_tx = sent_notes.tx
+            LEFT JOIN sapling_received_notes
+                      ON (sent_notes.tx, sent_notes.output_pool, sent_notes.output_index) =
+                         (sapling_received_notes.tx, 2, sapling_received_notes.output_index)
+            WHERE COALESCE(sapling_received_notes.is_change, 0) = 0;"
+        ).map_err(WalletMigrationError::from)
```
daira, line 181: I have not checked that this is identical to the SQL in `wallet/init.rs`.

daira (commented):
See comments.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_transparent_history.rs
str4d: I didn't either; I'm relying on the test in `wallet/init.rs` to already be doing that.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: Resolved by https://github.com/zcash/librustzcash/pull/974/files#diff-54ec5e42408ac28d55641fbe89818c7a8ea21c666b6e4b79b059776661c4bce9R488

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -513,19 +517,20 @@ mod tests {
...
-                        AND transactions.expiry_height <= blocks_max_height.max_height
+                        AND transactions.expiry_height BETWEEN 1 AND blocks_max_height.max_height
                    ) AS expired_unmined
-            FROM transactions
-            JOIN notes ON notes.id_tx = transactions.id_tx
+            FROM notes
+            LEFT OUTER JOIN transactions
+                 ON notes.txid = transactions.txid
             JOIN blocks_max_height
             LEFT JOIN blocks ON blocks.height = transactions.block
```
str4d, line 526: This join is still being done between `blocks` and `transactions`, which means that for UTXOs, even if we have scanned the block the UTXO was received in, we will report it as having `block_time = null` (which ends up being set to 0 somewhere in some of the downstream SDK usages). If we join `blocks` to `notes` instead, we avoid this problem.
```suggestion
            LEFT JOIN blocks ON blocks.height = notes.block
```

daira (approved):
utACK

str4d (approved):
utACK c0e29ad103ef774b9aa3af68b80cdb164ed649e1

PR #978, opened 2023-09-20
zcash_client_sqlite: Remove use of boolean constants from `v_tx_outputs`

nuttycom:
At present, [zcash/zcash-android-wallet-sdk] supports Android API 27, which bundles SQLite 3.19. SQLite support for the `TRUE` and `FALSE` constants were introduced in SQLite 3.23, so we cannot currently use these constants and retain support for Android API 27.

This version support limitation applies only to the `v_transactions` and `v_tx_outputs` views, which are considered part of the public API of this crate; other use of more recent SQLite features is fine because they rely upon the SQLite bundled via our use of the `rusqlite` crate and feature compatibility is verified via the unit tests of this crate.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -553,7 +553,7 @@ mod tests {
                    utxos.received_by_account   AS to_account,
                    utxos.address               AS to_address,
                    utxos.value_zat             AS value,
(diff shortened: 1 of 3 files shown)


str4d (approved):
utACK. I did not check for additional occurrences of `true` or `false` in the views, but the instances changed here are done so correctly.

PR #979, opened 2023-09-20
zcash_client_backend: Update API documentation.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -468,34 +473,53 @@ impl<Nf> ScannedBlock<Nf> {
...
     pub fn height(&self) -> BlockHeight {
         self.metadata.block_height
     }
 
+    /// Returns the block hash of the block that was scanned.
     pub fn block_hash(&self) -> BlockHash {
         self.metadata.block_hash
     }
 
+    /// Returns the block time of the block that was scanned.
```
daira, line 486:
```suggestion
    /// Returns the block time of the block that was scanned, as a Unix timestamp in seconds.
```

daira (approved):
utACK modulo the intra-doc links failure.

PR #984, opened 2023-09-21
zcash_client_sqlite: Add a test for cross-shard truncation.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -490,7 +490,7 @@ impl<Nf> ScannedBlock<Nf> {
     }
 
     /// Returns the metadata describing the state of the note commitment trees as of the end of the
-    /// scanned block. 
+    /// scanned block.
```
daira, line 493: #984 and #985 both include this change.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1208,7 +1208,16 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
             params: params.clone(),
         };
         wdb.with_sapling_tree_mut(|tree| {
-            tree.truncate_removing_checkpoint(&block_height).map(|_| ())
+            let truncate_succeeded = tree.truncate_removing_checkpoint(&block_height)?;
```
daira, line 1211: The odd `Result<bool, _>` return type from `ShardTree::truncate_removing_checkpoint` probably contributed to this bug.

https://github.com/zcash/incrementalmerkletree/blob/2ce18583d482d3868a2218657882c521e08f9c6d/shardtree/src/lib.rs#L562-L570

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1200,4 +1211,85 @@ mod tests {
...
+        for _ in 0..9 {
+            st.generate_next_block(
+                &dfvk,
+                AddressType::DefaultExternal,
+                Amount::const_from_i64(10000),
+            );
+        }
+        st.scan_cached_blocks(start_height, 10);
+
+        // Verify that the wallet has a checkpoint at the height we're goint to rewind to:
```
daira, line 1249:
```suggestion
        // Verify that the wallet has a checkpoint at the height we're going to rewind to:
```

daira (approved):
utACK

PR #985, opened 2023-09-22
`zcash_client_sqlite`: Fix transparent balance APIs

str4d:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -2048,4 +2049,138 @@ mod tests {
...
+        check_balance(&st, 2, NonNegativeAmount::ZERO);
+
+        // Expire the shielding transaction.
+        let expiry_height = st.wallet().get_transaction(txid).unwrap().expiry_height();
+        st.wallet_mut().update_chain_tip(expiry_height).unwrap();
+
+        // The transparent output should be spendable again, with more confirmations.
+        check_balance(&st, 0, value);
+        check_balance(&st, 1, value);
+        check_balance(&st, 2, value);
```
str4d, line 2184: I added this to the test because it occurred to me that this is something that can happen to user wallets, particularly for SDK 2.0.0 where transactions will be using a fixed fee. However, this part of the test is failing because we don't have any handling for this situation.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2048,4 +2049,138 @@ mod tests {
...
+
+        // The wallet should still have zero transparent balance.
+        check_balance(&st, 0, NonNegativeAmount::ZERO);
+        check_balance(&st, 1, NonNegativeAmount::ZERO);
+        check_balance(&st, 2, NonNegativeAmount::ZERO);
+
+        // Unmine the shielding transaction via a reorg.
+        st.wallet_mut()
+            .truncate_to_height(mined_height - 1)
+            .unwrap();
```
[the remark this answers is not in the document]
nuttycom: Hm, yeah, that `Verify` range insertion was from pretty early in my implementation, and I don't think it necessarily makes sense any more. Instead, after the rewind the next `update_chain_tip` call should establish the scanning ranges.

zcash_client_sqlite/src/wallet.rs
nuttycom: Let's talk tomorrow to figure out where it makes sense to handle this. Up until now transaction expiry has been treated as sort of a passive thing. I think that the correct solution here is going to be for the queries that are computing transparent balance to ensure that the spending transactions have not expired unmined. So, a UTXO might have a "spent in tx" field that is non-null, but that UTXO should still be treated as spendable if the referenced transaction has expired unmined, and future spends of that UTXO should then overwrite the spent-in-tx field.

daira: I suspect this should be a post-2.0 cleanup.

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
str4d: I think there is still a benefit to having a `Verify` range at the post-truncation tip, because the caller may not have rewound far enough, and if the new chain tip is scanned before we check the connection then the subsequent truncation will undo that new scan as well (causing more duplication of scanning work).

But it should be a verify range that doesn't extend past the truncated tip, which means we need at least one block of overlap (unlike the `Verify` range that `update_chain_tip` creates which sits adjacent to the prior tip, sufficient for connectivity checking but avoiding duplicate scanning). Probably just adding a 1-block range rescanning the truncated tip is sufficient; we already tell the caller to pick some rewind distance themselves, so there's no need to bake in additional duplicate scanning unless we want to change that interaction pattern (e.g. by having APIs that tell the caller what height to rewind to).

zcash_client_sqlite/src/wallet.rs
str4d: Opened #986.

zcash_client_sqlite/src/testing.rs
```diff
@@ -234,6 +234,35 @@ where
         (res, nf)
     }
 
+    /// Creates a fake block at the expected next height including the wallet transaction
```
nuttycom, line 237: This overlaps somewhat with @daira's changes in https://github.com/zcash/librustzcash/pull/908

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
nuttycom: What if the `Verify` range was just one block long?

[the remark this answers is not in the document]
str4d: Hmm, yeah. It's probably reasonable to assume here that the caller rewound to a block before a problematic scan, which implies that there exists at least one block after the truncation height. So instead of my current fix (verify the truncation height), we could verify the block after the truncation height (but no further). Still slightly unintuitive in terms of the method name though, as `chain_height` would not match the truncation.

zcash_client_sqlite/src/testing.rs
str4d: Gah, it's exactly duplicative. But I don't have access to those changes because they were merged into `main`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -716,10 +716,8 @@ pub(crate) fn get_wallet_summary(
         let mut stmt_transparent_balances = conn.prepare(
             "SELECT u.received_by_account, SUM(u.value_zat)
              FROM utxos u
-             LEFT OUTER JOIN transactions tx
-             ON tx.id_tx = u.spent_in_tx
              WHERE u.height <= :max_height
-             AND tx.block IS NULL
+             AND u.spent_in_tx IS NULL
```
nuttycom, line 720: I think the `expired unmined` could be handled by:
```suggestion
             LEFT OUTER JOIN transactions tx
             ON tx.id_tx = u.spent_in_tx
             WHERE u.height <= :max_height
             AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height < :max_height))
```

zcash_client_sqlite/src/testing.rs
str4d: Yep, this directly conflicts with `main`. AFAICT the only PRs that `main` has which the release branch doesn't are #908, #972 (no-op), and #954, so I propose that we merge `main` into the release branch (highly unconventional for a release branch, but AFAICT safe) and then I rebase on it.

nuttycom: Okay, agreed.

zcash_client_sqlite/src/wallet.rs
str4d: There's maybe an off-by-one here (this might need to be `tx.expiry_height <= max_height`, but I'd have to test to confirm). But per our thread in Slack, this is insufficient and the full fix would need more wide-scale changes (complicated by the difference in data model between transparent and shielded).

str4d:
[Rebased](https://github.com/zcash/librustzcash/compare/60063a4cbd34aa3fa0d043cbbf1af87b080fb54d..6a9ad39bde992ddea9374a74c0a219c4a7a8bf88) to bring in #987 and rewrite my `TestState` changes on top of it.

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
str4d: I'm going to leave my current fix in place (rescan the truncated-to height, so the chain height doesn't differ from that height until `update_chain_tip` is called) unless review asks otherwise.

zcash_client_sqlite/src/testing.rs
```diff
@@ -728,35 +755,42 @@ pub(crate) fn fake_compact_block<P: consensus::Parameters>(
 }
 
 /// Create a fake CompactBlock at the given height containing only the given transaction.
```
daira, line 757:
```suggestion
/// Create a fake CompactBlock at the given height containing only the given transaction.
/// This assumes that the transaction has no transparent components.
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -270,22 +270,49 @@ where
...
+            .expect("TxId should exist in the wallet");
+
+        // Index 0 is by definition a coinbase transaction, and the wallet doesn't
+        // construct coinbase transactions. So we pretend here that the block has a
+        // coinbase transaction that does not have shielded coinbase outputs.
+        self.generate_next_block_from_tx(1, &tx)
+    }
+
     /// Creates a fake block at the expected next height containing only the given
     /// transaction, and inserts it into the cache.
```
daira, line 294:
```suggestion
    /// transaction, and inserts it into the cache.
    /// This assumes that the transaction has only Sapling and Orchard components.
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -270,22 +270,49 @@ where
         (height, res)
     }
 
+    /// Creates a fake block at the expected next height containing only the wallet
+    /// transaction with the given txid, and inserts it into the cache.
```
daira, line 274:
```suggestion
    /// transaction with the given txid, and inserts it into the cache.
    /// This assumes that the transaction has only Sapling and Orchard components.
```

daira (approved):
utACK

zcash_client_backend/src/data_api.rs
```diff
@@ -490,7 +490,7 @@ impl<Nf> ScannedBlock<Nf> {
     }
 
     /// Returns the metadata describing the state of the note commitment trees as of the end of the
-    /// scanned block. 
+    /// scanned block.
```
daira, line 493: #984 and #985 both include this change.

zcash_client_sqlite/src/wallet.rs
nuttycom: It actually needs to compare to the chain tip, not :max_height

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/6a9ad39bde992ddea9374a74c0a219c4a7a8bf88..b76b028b3a84d58318a9f6433f0d3ff95468ac6b) to bring in changes from #988.

zcash_client_sqlite/src/wallet.rs
str4d: Changes from #988 have been brought into this PR.

nuttycom (approved):
utACK https://github.com/zcash/librustzcash/pull/985/commits/b76b028b3a84d58318a9f6433f0d3ff95468ac6b

daira (commented):
Post-hoc ACK

PR #987, opened 2023-09-22
Merge test harness fixes to the release candidate branch

str4d:
Brings in #908.


nuttycom (approved):
utACK; I verified that this merges 101c9c0ea491853b67dda9fdfbf8ae35948df202 which is already in `main` so no need to wait for CI.

str4d:
This is the same commit that was approved in #908, and I verified locally that tests pass after merging. The rustfmt issue is being fixed in #985.

daira (commented):
utACK, looks identical to the commits on `main`.

PR #988, opened 2023-09-22
zcash_client_sqlite: Consider utxos unspent if the spending transaction expired.

nuttycom:
If a shielding transaction is expired as of the current chain tip, its input utxos should be considered unspent.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -716,12 +716,15 @@ pub(crate) fn get_wallet_summary(
         let mut stmt_transparent_balances = conn.prepare(
             "SELECT u.received_by_account, SUM(u.value_zat)
              FROM utxos u
+             LEFT OUTER JOIN transactions tx ON tx.id_tx = u.spent_in_tx
              WHERE u.height <= :max_height
-             AND u.spent_in_tx IS NULL
+             AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height < :stable_height))
              GROUP BY u.received_by_account",
         )?;
-        let mut rows = stmt_transparent_balances
-            .query(named_params![":max_height": u32::from(zero_conf_height)])?;
+        let mut rows = stmt_transparent_balances.query(named_params![
+            ":max_height": u32::from(zero_conf_height),
+            ":stable_height": u32::from(chain_tip_height - PRUNING_DEPTH)
+        ])?;
 
         while let Some(row) = rows.next()? {
             let account = AccountId::from(row.get::<_, u32>(0)?);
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -716,12 +716,15 @@ pub(crate) fn get_wallet_summary(
         let mut stmt_transparent_balances = conn.prepare(
             "SELECT u.received_by_account, SUM(u.value_zat)
              FROM utxos u
+             LEFT OUTER JOIN transactions tx ON tx.id_tx = u.spent_in_tx
              WHERE u.height <= :max_height
-             AND u.spent_in_tx IS NULL
+             AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height < :chain_tip_height))
```
daira, line 721: Off-by-one: if the expiry height is equal to the the current tip height and the tx is not yet mined, it has expired.
```suggestion
             AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height <= :chain_tip_height))
```
https://zips.z.cash/zip-0203#specification

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1300,23 +1303,25 @@ pub(crate) fn get_unspent_transparent_outputs<P: consensus::Parameters>(
...
     let mut stmt_blocks = conn.prepare(
-        "SELECT u.prevout_txid, u.prevout_idx, u.script,
-                u.value_zat, u.height
+        "SELECT u.prevout_txid, u.prevout_idx, u.script, u.value_zat, u.height
          FROM utxos u
+         LEFT OUTER JOIN transactions tx ON tx.id_tx = u.spent_in_tx
          WHERE u.address = :address
          AND u.height <= :max_height
-         AND u.spent_in_tx IS NULL",
+         AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height < :chain_tip_height))",
```
daira, line 1313: Same off-by-one.
```suggestion
         AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height <= :chain_tip_height))",
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1363,19 +1368,22 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
 ) -> Result<HashMap<TransparentAddress, Amount>, SqliteClientError> {
+    let chain_tip_height = scan_queue_extrema(conn)?.map(|(_, max)| max);
     let mut stmt_blocks = conn.prepare(
         "SELECT u.address, SUM(u.value_zat)
          FROM utxos u
+         LEFT OUTER JOIN transactions tx ON tx.id_tx = u.spent_in_tx
          WHERE u.received_by_account = :account_id
          AND u.height <= :max_height
-         AND u.spent_in_tx IS NULL
+         AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height < :chain_tip_height))
```
daira, line 1378: Same off-by-one.
```suggestion
         AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height <= :chain_tip_height))
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2176,15 +2184,11 @@ mod tests {
 
         // Expire the shielding transaction.
         let expiry_height = st.wallet().get_transaction(txid).unwrap().expiry_height();
-        st.wallet_mut().update_chain_tip(expiry_height).unwrap();
+        st.wallet_mut().update_chain_tip(expiry_height + 1).unwrap();
```
daira, line 2187:
```suggestion
        st.wallet_mut().update_chain_tip(expiry_height).unwrap();
```
This was correct. If we update the chain tip height to the expiry height then the tx is expired, because that was the last block height it could be mined at.

daira (changes requested):
Blocking off-by-one error.

zcash_client_sqlite/src/wallet.rs
daira: If we make this change first it should fail, and then it will be fixed by correcting the off-by-ones in the SQL queries.

str4d:
utACK the general changes. I'm going to adopt these changes directly into #985 instead of merging this PR.

PR #991, opened 2023-09-25
zcash_client_sqlite: Include spent utxos in v_transactions net value.

nuttycom:
This fixes a bug in v_transactions whereby shielding transactions displayed an incorrect balance.


str4d (approved):
utACK b90f285359ebc691278e451d47ce6ce576bccc20

daira (approved):
utACK

PR #995, opened 2023-09-26
Move common package and dependency configs into workspace config

str4d:
The MSRV for the main crates is 1.65, which is higher than the Rust version that stabilised workplace dependencies (1.64). The implicit MSRV for the component crates is still lower than this, so we don't migrate these crates.


nuttycom (approved):
tested ACK; I like having all the dependency versions handled in one place and avoiding that redundancy.

Is there any real reason we shouldn't bump the components MSRV soon? Do we know of any use cases that require the lower MSRV?

str4d:
> Is there any real reason we shouldn't bump the components MSRV soon? Do we know of any use cases that require the lower MSRV?

My reasoning is the reverse: is there any reason we _should_ bump the components MSRV?
- `equihash`, `f4jumble`, and `zcash_encoding` have not been touched in a year.
- `zcash_address` updates in the last year have only been to de-duplicate dependencies.
- `zcash_note_encryption` has seen the most activity, and is the most complex of the `components/` crates to maintain, so is probably the best candidate for a higher MSRV. But it's explicitly the crate we need to move out of this repository for cycle-breaking purposes, so integrating it into the workspace dependencies here is unhelpful.

The answer here might be that we end up having _multiple_ component repositories, rather than moving all of `components/` into the same separate repo.

PR #996, opened 2023-09-27
zcash_client_backend: Return summary information from `scan_cached_blocks`

### LINKED ISSUE #918: zcash_client_backend: Return scan summary from `scan_cached_blocks`
It is useful for the wallet to be able to skip updating the UI when scanning a range of blocks results in no changes to wallet state. The easiest way to do this is for `scan_cached_blocks` to report this to the caller so that the caller can choose whether to update the UI.

nuttycom:
When scanning, a wallet only needs to update balance and transaction information shown to users when the scan has resulted in a change to wallet state. This modifies `scan_cached_blocks` to return the range of block heights actually scanned, along with the counts of notes spent and received by the wallet in that range.

Fixes #918

### DIFF
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -144,6 +144,8 @@
 //! # }
 //! ```
 
+use std::ops::Range;
+
 use zcash_primitives::{
     block::BlockHash,
     consensus::{self, BlockHeight},
@@ -199,8 +201,6 @@ pub trait BlockSource {
     /// `from_height`, applying the provided callback to each block. If `from_height`
     /// is `None` then scanning will begin at the first available block.
     ///
-    /// Returns an error if `from_height` is set to a block that does not exist in `self`.
-    ///
     /// * `WalletErrT`: the types of errors produced by the wallet operations performed
     ///   as part of processing each row.
     /// * `NoteRefT`: the type of note identifiers in the wallet data store, for use in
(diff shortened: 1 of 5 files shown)


zcash_client_backend/src/data_api/chain.rs
```diff
@@ -209,12 +213,45 @@ pub trait BlockSource {
         &self,
         from_height: Option<BlockHeight>,
         limit: Option<usize>,
-        with_row: F,
-    ) -> Result<(), error::Error<WalletErrT, Self::Error>>
+        with_block: F,
+    ) -> Result<Option<Range<BlockHeight>>, error::Error<WalletErrT, Self::Error>>
```
str4d, line 217: Change needs to be documented in the changelog.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -199,7 +201,9 @@ pub trait BlockSource {
     /// `from_height`, applying the provided callback to each block. If `from_height`
     /// is `None` then scanning will begin at the first available block.
     ///
-    /// Returns an error if `from_height` is set to a block that does not exist in `self`.
+    /// Returns the range of block heights for blocks passed to `with_block` if any blocks were
+    /// available, `Ok(None)` if no blocks were provided, or an error if `from_height` is set to a
+    /// block that does not exist in `self`.
```
str4d, line 206: This reads weirdly to me, but I can't think of a nicer way to state it. "No blocks provided" and "blocks passed" are the phrases that are tripping me up. I think what it's trying to convey is that the range contains all blocks that `with_block` was called on, and `Ok(None)` means that `with_block` was never called?

Separately, what should this return if `from_height` is `Some(h)` and `limit` is `Some(0)`? I think that semantically, this should return `Ok(Some(Range(h, h)))` (i.e. the empty range starting at `h`), even though `with_block` was never called, but the answer to this somewhat depends on how the return value is documented (per above). The empty range is not fully duplicative of `Ok(None)`, due to `from_height` being optional (an empty range still has a defined start height, and `from_height` being `None` is documented as "use an undefined start height" - although we could improve that definition by returning the empty range starting at Sapling activation, for instance).

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -209,12 +213,45 @@ pub trait BlockSource {
...
-        with_row: F,
-    ) -> Result<(), error::Error<WalletErrT, Self::Error>>
+        with_block: F,
+    ) -> Result<Option<Range<BlockHeight>>, error::Error<WalletErrT, Self::Error>>
     where
         F: FnMut(CompactBlock) -> Result<(), error::Error<WalletErrT, Self::Error>>;
 }
 
+#[derive(Debug)]
+pub struct ScanSummary {
```
str4d, line 223: - Needs documentation (both the struct, and its methods).
- Needs to be mentioned as an addition in the changelog.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -209,12 +213,45 @@ pub trait BlockSource {
...
         limit: Option<usize>,
-        with_row: F,
-    ) -> Result<(), error::Error<WalletErrT, Self::Error>>
+        with_block: F,
+    ) -> Result<Option<Range<BlockHeight>>, error::Error<WalletErrT, Self::Error>>
     where
         F: FnMut(CompactBlock) -> Result<(), error::Error<WalletErrT, Self::Error>>;
 }
 
+#[derive(Debug)]
```
str4d, line 222:
```suggestion
#[derive(Clone, Debug)]
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -240,7 +277,7 @@ pub fn scan_cached_blocks<ParamsT, DbT, BlockSourceT>(
     data_db: &mut DbT,
     from_height: BlockHeight,
     limit: usize,
-) -> Result<(), Error<DbT::Error, BlockSourceT::Error>>
+) -> Result<Option<ScanSummary>, Error<DbT::Error, BlockSourceT::Error>>
```
str4d, line 280: This change needs to be documented in the changelog.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -341,7 +378,9 @@ where
     batch_runner.flush();
 
     let mut scanned_blocks = vec![];
-    block_source.with_blocks::<_, DbT::Error>(
+    let mut received_note_count = 0;
+    let mut spent_note_count = 0;
+    let scanned_range = block_source.with_blocks::<_, DbT::Error>(
```
str4d, line 383: We're using the fact that `with_blocks` takes an `FnMut` to collect information about the blocks that the closure was called with. That means we could _also_ detect and collect `scanned_range` in the same way, without altering `BlockSource`. If we can't adequately explain what the return value of `with_blocks` is meant to be (per my previous comment), then an alternative is to not alter `BlockSource` and just collect the data in this context where we know what we want to report.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -355,6 +394,15 @@ where
             )
             .map_err(Error::Scan)?;
 
+            let (s, r) = scanned_block
+                .transactions
+                .iter()
+                .fold((0, 0), |(s, r), wtx| {
+                    (s + wtx.sapling_spends.len(), r + wtx.sapling_outputs.len())
+                });
+            spent_note_count += s;
```
str4d, line 403: This undercounts during reverse scanning, so it isn't a reliable metric. We either need to exclude it, document it as such, or additionally modify `put_blocks` to report how many spends it detects for the reverse case (so that this metric overall reports all spends, just not necessarily while scanning the block ranges that actually perform the spends).

zcash_client_backend/src/data_api/chain.rs
nuttycom: I struggled here with whether I should return `Option<ScanSummary>` or just `ScanSummary`, but my conclusion was that I couldn't actually produce a useful `ScanSummary` if no height information was provided and the `BlockSource` didn't pick up anything.

zcash_client_backend/src/data_api/chain.rs
nuttycom: Yeah, I thought about adding a `fold_blocks` method to `BlockSource` and then implementing `fold_blocks` in terms of it.

zcash_client_backend/src/data_api/chain.rs
nuttycom: What we're trying to determine here is the answer to the question "do I need to update my user's view of the wallet state". We could just return a boolean, but that seems a bit impoverished. When we're scanning in reverse order and discover one of our notes for which the nullifier has already been revealed in a spend in a later block, we're actually discovering a received note, and we always want to update the user's view of the wallet state when we find a received not, irrespective of whether that note has been spent. So, I see two possibilities. We can keep this as it is implemented here, or we can modify `put_blocks` as you suggest, counting the received notes in two separate buckets, one containing those that we know have already been spent, and the other where we're not sure whether they've already been spent.

Since the "received, but we're not sure if it's been spent" bucket would have uncertainty more or less equivalent to the undercounting of spends you describe in your comment above, I think that we should just keep this implementation.

nuttycom: I've updated the documentation of `ScanSummary` to clarify the meaning of these values.

zcash_client_backend/src/data_api/chain.rs
nuttycom: Change was reverted.

zcash_client_backend/src/data_api/chain.rs
nuttycom: I realized that it actually _is_ possible to provide a non-Optional `ScanSummary`, because `from_height` is now non-optional.

zcash_client_backend/src/data_api/chain.rs
str4d: The problem here is that because `ScanSummary` is completely undocumented, I have no way of knowing what `ScanSummary::spent_sapling_note_count` is meant to mean, beyond what the method name says. And what the method name says, is not what it actually counts.

If these numbers are intended to be "something happened, and here is a measure of how much of that something happened in this range", then we should convey that in the output (similarly to how the progress metric is an opaque numerator and denominator, and does not indicate to the caller that it happens to be counting notes).

str4d (approved):
utACK

zcash_client_backend/src/data_api/chain.rs
str4d: Changes post-documentation LGTM.

PR #997, opened 2023-09-28
zcash_client_sqlite: Replace internal height tuples with `RangeInclusive`

str4d:
We don't need to iterate over them, but the `*_extrema` internal methods are semantically returning inclusive ranges, and using `RangeInclusive` avoids bugs where the wrong half of the tuple is used (instead moving the location of the tuple handling inside the `*_extrema` methods, which cuts the number of occurrences from linear in the number of function calls to constant).


nuttycom (approved):
utACK, this is a good change and I wish I'd thought of it.

daira (commented):
Post-hoc ACK

PR #1000, opened 2023-09-29
add Debug to transaction builder-reported types

zancas:
@nuttycom here're some debug attributes

### DIFF
--- a/zcash_primitives/src/transaction/components/transparent/fees.rs
+++ b/zcash_primitives/src/transaction/components/transparent/fees.rs
@@ -9,7 +9,7 @@ use crate::{
 
 /// This trait provides a minimized view of a transparent input suitable for use in
 /// fee and change computation.
-pub trait InputView {
+pub trait InputView: std::fmt::Debug {
     /// The outpoint to which the input refers.
     fn outpoint(&self) -> &OutPoint;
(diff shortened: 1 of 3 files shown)


nuttycom (approved):
utACK

zcash_primitives/src/transaction/components/transparent/fees.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
 
 /// This trait provides a minimized view of a transparent input suitable for use in
 /// fee and change computation.
-pub trait InputView {
+pub trait InputView: std::fmt::Debug {
```
str4d, line 12: As a public trait, this change needs to be documented in the changelog.

zcash_primitives/src/transaction/components/transparent/fees.rs
```diff
@@ -18,7 +18,7 @@ pub trait InputView {
 
 /// This trait provides a minimized view of a transparent output suitable for use in
 /// fee and change computation.
-pub trait OutputView {
+pub trait OutputView: std::fmt::Debug {
```
str4d, line 21: Ditto.

str4d (commented):
Post-hoc utACK with necessary follow-ups.

PR #1005, opened 2023-10-06
Add a MARGINAL_FEE constant for the ZIP 317 marginal fee

zancas:


daira (approved):
utACK modulo comment, but please add a release notes entry, since this is a public API addition. Also please change the commit message to match the PR title (and squash all commits), since there are no uses in tests.

(I'm not sure we want clients to be calculating ZIP 317 fees themselves rather than using the `FeeRule`, but it seems relatively harmless to add the constant, and we already have a constant for the minimum fee.)

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -19,6 +19,7 @@ use crate::{
 ///
 /// [ZIP 317]: https//zips.z.cash/zip-0317
 pub const MINIMUM_FEE: Amount = Amount::const_from_i64(10_000);
+pub const MARGINAL_FEE: Amount = Amount::const_from_i64(5_000);
```
daira, line 22: Remember that each public item must be separately documented.

```suggestion

/// The marginal fee using the standard [ZIP 317] constants. Equivalent to
/// `FeeRule::standard().marginal_fee()`, but as a constant.
///
/// [ZIP 317]: https//zips.z.cash/zip-0317
pub const MARGINAL_FEE: Amount = Amount::const_from_i64(5_000);
```

nuttycom: Oh hey, I wanted to make this change just yesterday!

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -20,6 +20,12 @@ use crate::{
 /// [ZIP 317]: https//zips.z.cash/zip-0317
 pub const MINIMUM_FEE: Amount = Amount::const_from_i64(10_000);
 
+/// The marginal fee using the standard [ZIP 317] constants. Equivalent to
+/// `FeeRule::standard().marginal_fee()`, but as a constant.
```
nuttycom, line 24:
```suggestion
/// The marginal fee using the standard [ZIP 317] constants. 
```

nuttycom: `FeeRule::standard` could change at some point and this would then be incorrect.

nuttycom (approved):
utACK 0e3b0e6d213c0eb02b8ef085f95fe2daddfc271d

nuttycom (approved):
re-utACK

Issue #741, opened 2022-12-13
Refactor `SaplingBuilder` to have a similar API to `orchard::builder::Builder`


str4d:
We should produce an unproven, unsigned `Bundle`, and then use state transitions to create the proofs and apply the signatures.

PR #1007, opened 2023-10-06
Changes to Sapling bundle mapping

str4d:
Also includes a few new `Clone` and `Debug` impls we need.

Part of #741.

### DIFF
--- a/zcash_primitives/src/transaction/components/sapling.rs
+++ b/zcash_primitives/src/transaction/components/sapling.rs
@@ -99,6 +99,33 @@ impl MapAuth<Authorized, Authorized> for () {
     }
 }
 
+/// A helper for implementing `MapAuth` with a set of closures.
+impl<A, B, F, G, H, I> MapAuth<A, B> for (F, G, H, I)
+where
+    A: Authorization,
+    B: Authorization,
+    F: FnMut(A::SpendProof) -> B::SpendProof,
+    G: FnMut(A::OutputProof) -> B::OutputProof,
+    H: FnMut(A::AuthSig) -> B::AuthSig,
+    I: FnMut(A) -> B,
+{
+    fn map_spend_proof(&mut self, p: A::SpendProof) -> B::SpendProof {
+        self.0(p)
+    }
+
+    fn map_output_proof(&mut self, p: A::OutputProof) -> B::OutputProof {
+        self.1(p)
+    }
+
+    fn map_auth_sig(&mut self, s: A::AuthSig) -> B::AuthSig {
+        self.2(s)
+    }
+
+    fn map_authorization(&mut self, a: A) -> B {
+        self.3(a)
+    }
+}
+
 /// A fallible map from one bundle authorization to another.
 ///
 /// For use with [`Bundle::try_map_authorization`].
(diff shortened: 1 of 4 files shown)


str4d:
The tuple impls let me write code like:
```rust
impl Bundle<InProgress<Proven, PartiallyAuthorized>> {
    /// Finalizes this bundle, enabling it to be included in a transaction.
    ///
    /// Returns an error if any signatures are missing.
    pub fn finalize(self) -> Result<Bundle<Authorized>, Error> {
        self.try_map_authorization((
            |p| Ok(p),
            |p| Ok(p),
            |maybe: MaybeSigned| maybe.finalize(),
            |partial: InProgress<Proven, PartiallyAuthorized>| {
                Ok(Authorized {
                    binding_sig: partial.sigs.binding_signature,
                })
            },
        ))
    }
}
```
which has the same usability as the `orchard` crate's implementations, while still allowing the maps to be defined as structs if desired.

str4d:
Meanwhile, having the `MapAuth` trait methods take `&mut self` means I can write code like:
```rust

impl<P: InProgressProofs> Bundle<InProgress<P, Unsigned>> {
    /// Loads the sighash into this bundle, preparing it for signing.
    ///
    /// This API ensures that all signatures are created over the same sighash.
    pub fn prepare<R: RngCore + CryptoRng>(
        self,
        mut rng: R,
        sighash: [u8; 32],
    ) -> Bundle<InProgress<P, PartiallyAuthorized>> {
        self.map_authorization((
            |proof| proof,
            |proof| proof,
            |parts| MaybeSigned::SigningMetadata(parts),
            |auth: InProgress<P, Unsigned>| InProgress {
                sigs: PartiallyAuthorized {
                    binding_signature: auth.sigs.bsk.sign(
                        &sighash,
                        &mut rng,
                        VALUE_COMMITMENT_RANDOMNESS_GENERATOR,
                    ),
                    sighash,
                },
                _proof_state: PhantomData::default(),
            },
        ))
    }
}
```
which requires `FnMut` closures in order to access `rng` (or equivalently `&mut self` to update mutable fields of struct-based maps).

daira (approved):
utACK

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -56,11 +56,14 @@ impl Authorization for Authorized {
...
+/// For use with [`Bundle::map_authorization`].
 pub trait MapAuth<A: Authorization, B: Authorization> {
-    fn map_spend_proof(&self, p: A::SpendProof) -> B::SpendProof;
-    fn map_output_proof(&self, p: A::OutputProof) -> B::OutputProof;
-    fn map_auth_sig(&self, s: A::AuthSig) -> B::AuthSig;
-    fn map_authorization(&self, a: A) -> B;
+    fn map_spend_proof(&mut self, p: A::SpendProof) -> B::SpendProof;
+    fn map_output_proof(&mut self, p: A::OutputProof) -> B::OutputProof;
+    fn map_auth_sig(&mut self, s: A::AuthSig) -> B::AuthSig;
+    fn map_authorization(&mut self, a: A) -> B;
```
nuttycom, line 66: `map` with a `&mut self` reference seems increasingly misnamed, and the commit text doesn't explain the rationale for this change, so I'm assuming I'll be able to figure it out from later commits.

nuttycom (approved):
utACK modulo documentation; it's not clear at all why it's necessary to have the transformation algebra itself be mutable.

EDIT: I see now the PR comments.

zcash_primitives/src/transaction/components/sapling.rs
str4d: I gave an example of why I needed this in a comment on the PR.

PR #1010, opened 2023-10-09
Update error messages that referred to the fetch-params script

daira:


nuttycom (commented):
I think there are a couple of extra things we need here:
* use wagyu-zcash-parameters for the tests
* It's not clear what the "default location" is for mobile platforms; perhaps we should get rid of the notion of there being defaults entirely at this point.

nuttycom:
Made irrelevant by #1017

PR #1011, opened 2023-10-09
Make `ReceivedSaplingNote` internals private, and use (TxId, output_index) for note identifiers.

nuttycom:


zcash_client_backend/CHANGELOG.md
```diff
@@ -7,9 +7,21 @@ and this library adheres to Rust's notion of
...
+
+- Added methods to `zcash_client_backend::wallet::ReceivedSaplingNote`:
+  `{from_parts, txid, output_index, diversifier, rseed, note_commitment_tree_position}`.
+
 ### Changed
 - `zcash_client_backend::data_api::chain::scan_cached_blocks` now returns
   a `ScanSummary` containing metadata about the scanned blocks on success.
+- The fields of `zcash_client_backend::wallet::ReceivedSaplingNote` are now
+  private. Use `ReceivedSaplingNote::from_parts` for construction instead.
+  Accessor methods are provided for each previously-public field.
```
daira, line 20:
```suggestion
  Accessor methods are provided for each previously public field.
```
(It's only compound adjectives that need a hyphen. An adverb modifying an adjective doesn't count as a compound adjective, even though it's logically similar.)

daira (approved):
utACK

PR #1012, opened 2023-10-09
zcash_client_backend: Move change memos into the `ChangeValue` components of `Proposal`s

nuttycom:
The existing API limited change outputs to having only a single memo repeated across each change output. This change makes it so that each proposed change output can have its own associated memo, and leaves it up to the input selector to determine how requested change memos are associated with change outputs.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -118,6 +118,7 @@ where
 ///   received note must have in the blockchain in order to be considered for being
 ///   spent. A value of 10 confirmations is recommended and 0-conf transactions are
 ///   not supported.
+/// * `change_memo`: A memo to be included in the change output
 ///
 /// # Examples
 ///
@@ -174,7 +175,8 @@ where
 ///     Amount::from_u64(1).unwrap(),
 ///     None,
 ///     OvkPolicy::Sender,
-///     10
+///     10,
+///     None
 /// )
 ///
 /// # }
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -196,6 +196,7 @@ pub fn create_spend_to_address<DbT, ParamsT>(
     memo: Option<MemoBytes>,
     ovk_policy: OvkPolicy,
     min_confirmations: NonZeroU32,
+    change_memo: Option<MemoBytes>,
 ) -> Result<
     TxId,
     Error<
@@ -224,7 +225,7 @@ where
 
     #[allow(deprecated)]
     let fee_rule = fixed::FeeRule::standard();
-    let change_strategy = fees::fixed::SingleOutputChangeStrategy::new(fee_rule);
+    let change_strategy = fees::fixed::SingleOutputChangeStrategy::new(fee_rule, change_memo);
     spend(
         wallet_db,
         params,
@@ -335,7 +336,6 @@ where
         ovk_policy,
         proposal,
         min_confirmations,
-        None,
     )
 }
 
@@ -418,7 +418,6 @@ pub fn create_proposed_transaction<DbT, ParamsT, InputsErrT, FeeRuleT>(
     ovk_policy: OvkPolicy,
     proposal: Proposal<FeeRuleT, DbT::NoteRef>,
     min_confirmations: NonZeroU32,
-    change_memo: Option<MemoBytes>,
 ) -> Result<
     TxId,
     Error<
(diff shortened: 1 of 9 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -199,6 +199,7 @@ pub fn create_spend_to_address<DbT, ParamsT>(
     memo: Option<MemoBytes>,
     ovk_policy: OvkPolicy,
     min_confirmations: NonZeroU32,
+    change_memo: Option<MemoBytes>,
```
daira, line 202: The doc comment needs to be updated in the Parameters section and the example (lines 124 and 181).

zcash_client_backend/src/fees.rs
```diff
@@ -19,13 +20,20 @@ pub mod zip317;
...
+        Self::Sapling { value, memo }
+    }
+
     pub fn value(&self) -> Amount {
         match self {
-            ChangeValue::Sapling(value) => *value,
+            ChangeValue::Sapling { value, .. } => *value,
         }
     }
 }
```
daira, line 39:
```suggestion
    pub fn memo(&self) -> MemoBytes {
        match self {
            ChangeValue::Sapling { memo, .. } =>
                memo.as_ref().map_or_else(MemoBytes::empty, |m| m.clone())
        }
    }
}
```
(or however rustfmt formats that). Alternatively keep the return value as `Option<MemoBytes>`.

zcash_client_backend/src/fees/fixed.rs
```diff
@@ -21,12 +22,16 @@ use super::{
 /// shielded pool and delegates fee calculation to the provided fee rule.
 pub struct SingleOutputChangeStrategy {
     fee_rule: FixedFeeRule,
+    change_memo: Option<MemoBytes>,
 }
 
 impl SingleOutputChangeStrategy {
     /// Constructs a new [`SingleOutputChangeStrategy`] with the specified fee rule.
```
daira, line 29:
```suggestion
    /// Constructs a new [`SingleOutputChangeStrategy`] with the specified fee rule
    /// and change memo.
```

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -28,13 +29,17 @@ use super::{
 /// shielded pool and delegates fee calculation to the provided fee rule.
 pub struct SingleOutputChangeStrategy {
     fee_rule: Zip317FeeRule,
+    change_memo: Option<MemoBytes>,
 }
 
 impl SingleOutputChangeStrategy {
     /// Constructs a new [`SingleOutputChangeStrategy`] with the specified ZIP 317
     /// fee parameters.
```
daira, line 37:
```suggestion
    /// fee parameters and change memo.
```

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -28,13 +29,17 @@ use super::{
...
-    pub fn new(fee_rule: Zip317FeeRule) -> Self {
-        Self { fee_rule }
+    pub fn new(fee_rule: Zip317FeeRule, change_memo: Option<MemoBytes>) -> Self {
+        Self {
+            fee_rule,
+            change_memo,
+        }
     }
```
daira, line 43: Alternatively, if we wanted to make this a less disruptive API change, we could do:
```suggestion
    pub fn new(fee_rule: Zip317FeeRule) -> Self {
        Self {
            fee_rule,
            None,
        }
    }

    /// Constructs a new [`SingleOutputChangeStrategy`] with the specified ZIP 317
    /// fee parameters and change memo.
    pub fn new_with_memo(fee_rule: Zip317FeeRule, change_memo: Option<MemoBytes>) -> Self {
        Self {
            fee_rule,
            change_memo,
        }
    }
```

daira (approved):
utACK with suggestions.

zcash_client_backend/src/fees/zip317.rs
nuttycom: I prefer the more-disruptive change because it forces people to adapt to the change, and won't then silently do a thing they don't expect.

daira (approved):
utACK

daira:
The wasi and clippy failures were due to failed downloads from crates.io, so I restarted them.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/f5b5849310270f7aff5b20613da298888909fcc5..bcea060) to resolve conflicts after the merge of #1013

daira (approved):
utACK

PR #1013, opened 2023-10-09
zcash_client_backend: Use `NonNegativeAmount` for fee and change amounts.

nuttycom:
It's incorrect to use the possibly-negative `Amount` type in places where values must be nonnegative.

This updates fee and change data structures to use `NonNegativeAmount` where possible.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -22,6 +22,12 @@ and this library adheres to Rust's notion of
   - The `NoteMismatch` variant of `data_api::error::Error` now wraps a
     `data_api::NoteId` instead of a backend-specific note identifier. The
     related `NoteRef` type parameter has been removed from `data_api::error::Error`.
+  - `wallet::create_spend_to_address` now takes a `NonNegativeAmount` rather than
+    an `Amount`.
+  - All uses of `Amount` in `data_api::wallet::input_selection` have been replaced
+    with `NonNegativeAmount`.
+- All uses of `Amount` in `zcash_client_backend::fees` have been replaced
+  with `NonNegativeAmount`.
 
 ## [0.10.0] - 2023-09-25
 
--- a/zcash_primitives/src/sapling/note_encryption.rs
+++ b/zcash_primitives/src/sapling/note_encryption.rs
@@ -27,7 +27,7 @@ use crate::{
         Diversifier, Note, PaymentAddress, Rseed,
     },
     transaction::components::{
-        amount::Amount,
+        amount::NonNegativeAmount,
         sapling::{self, OutputDescription},
     },
 };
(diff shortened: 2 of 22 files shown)


zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -83,7 +83,12 @@ where
 
     // The unwraps below are guaranteed to succeed by the assertion above
     let diversifier = Diversifier(plaintext[1..12].try_into().unwrap());
-    let value = Amount::from_u64_le_bytes(plaintext[12..20].try_into().unwrap()).ok()?;
+    let value = Amount::from_u64_le_bytes(
+        plaintext[12..20]
+            .try_into()
+            .expect("Note plaintext should contain a valid ZEC value."),
```
daira, line 89: This message is incorrect; it's only expecting that the slice is 8 bytes.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -345,7 +347,11 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
             &mut self.rng,
             ovk,
             to,
-            NoteValue::from_raw(value.into()),
+            NoteValue::from_raw(
+                value
+                    .try_into()
+                    .expect("Cannot create Sapling outputs with negative note values."),
```
daira, line 353:
```suggestion
                    .expect("value is checked to be non-negative"),
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this library adheres to Rust's notion of
   - The `NoteMismatch` variant of `data_api::error::Error` now wraps a
     `data_api::NoteId` instead of a backend-specific note identifier. The
     related `NoteRef` type parameter has been removed from `data_api::error::Error`.
+  - `wallet::create_spend_to_address` now takes a `NonNegativeAmount` rather than
+    an `Amount`.
```
daira, line 26: There are other public API changes.

zcash_primitives/src/sapling/note_encryption.rs
```diff
@@ -82,9 +82,20 @@ where
     }
 
     // The unwraps below are guaranteed to succeed by the assertion above
-    let diversifier = Diversifier(plaintext[1..12].try_into().unwrap());
-    let value = Amount::from_u64_le_bytes(plaintext[12..20].try_into().unwrap()).ok()?;
-    let r: [u8; 32] = plaintext[20..COMPACT_NOTE_SIZE].try_into().unwrap();
+    let diversifier = Diversifier(
+        plaintext[1..12]
+            .try_into()
+            .expect("Note plaintext is checked to have length > COMPACT_NOTE_SIZE."),
```
daira, line 88: Should be `>=` (also below).

zcash_primitives/src/transaction/fees.rs
```diff
@@ -31,7 +31,7 @@ pub trait FeeRule {
         sapling_input_count: usize,
         sapling_output_count: usize,
         orchard_action_count: usize,
-    ) -> Result<Amount, Self::Error>;
+    ) -> Result<NonNegativeAmount, Self::Error>;
```
daira, line 34: These are public API changes that need a CHANGELOG entry.

daira (approved):
utACK modulo comments.

zcash_primitives/src/transaction/builder.rs
nuttycom: This is being made infallible in #1014

zcash_primitives/src/transaction/components/amount.rs
```diff
@@ -38,6 +40,14 @@ impl Amount {
         Amount(amount)
     }
 
+    /// Creates a constant Amount from a u64.
+    ///
+    /// Panics: if the amount is outside the range `{0..MAX_MONEY}`.
+    const fn const_from_u64(amount: u64) -> Self {
```
daira, line 46: Why not `pub`?

daira (approved):
utACK with minor comment.

zcash_primitives/src/transaction/components/amount.rs
nuttycom: I want to reconsider these types; `Amount` will become something like `ZatoshiDelta` and be used exclusively in those few cases where negative values make sense, and so I don't want to expose more public APIs. Ideally, we'll eliminate all of the conversions from raw types and make it so that these value deltas can only be produced by subtraction of `Zatoshi` amounts; right now there are some serialization boundaries that still expect i64s but I think that in almost all those cases, the negative part of the range is actually in error.

PR #1014, opened 2023-10-10
Use `NonNegativeAmount` for note and UTXO value fields

nuttycom:
Builds atop #1012 

Review of this PR is slightly easier if you ignore whitespace.

PR #1015, opened 2023-10-10
zcash_primitives: Use sapling::NoteValue instead of bare u64

nuttycom:
Builds upon #1014

### DIFF
--- a/zcash_extensions/src/transparent/demo.rs
+++ b/zcash_extensions/src/transparent/demo.rs
@@ -815,7 +815,10 @@ mod tests {
         // create some inputs to spend
         let extsk = ExtendedSpendingKey::master(&[]);
         let to = extsk.default_address().1;
-        let note1 = to.create_note(110000, Rseed::BeforeZip212(jubjub::Fr::random(&mut rng)));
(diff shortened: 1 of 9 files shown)


zancas:
Hi @nuttycom we're having [under/over]-flow panics in `zingolib` so we're quite interested in this upgrade!  Can @AloeareV or I help move it along?

AloeareV:
Does `sapling::NoteValue` need to be a distinct type from `NonNegativeAmount`?

nuttycom:
> Does `sapling::NoteValue` need to be a distinct type from `NonNegativeAmount`?

The intent is that the `sapling` module is soon going to be factored out into a separate `sapling-crypto` crate, following much the same design as `orchard`. Sapling note values are not internally (within the ZK circuit) bounded above by `MAX_MONEY`; the value balance rules are enforced at the consensus level, so semantically these _are_ different types (and the dependency graph will reflect that better once `sapling-crypto` is factored out.)

str4d (approved):
utACK caee90dce9b33f81d99369c5410d3aa9cb8d0215

PR #1017, opened 2023-10-11
Test with bundled prover instead of local prover.

nuttycom:


.github/workflows/ci.yml
```diff
@@ -12,21 +12,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Fetch path to Zcash parameters
-        working-directory: ./zcash_proofs
-        shell: bash
-        run: echo "ZCASH_PARAMS=$(cargo run --release --example get-params-path --features directories)" >> $GITHUB_ENV
```
str4d, line 18: We can delete the `get-params-path` example now; it was only added for use in our CI. (However, we should verify that zebrad's CI isn't also using it; I think they might have been at one stage.)

.github/workflows/ci.yml
```diff
@@ -12,21 +12,6 @@ jobs:
...
-      - name: Cache Zcash parameters
-        id: cache-params
-        uses: actions/cache@v3.3.2
-        with:
-          path: ${{ env.ZCASH_PARAMS }}
-          key: ${{ runner.os }}-params
-      - name: Fetch Zcash parameters
-        if: steps.cache-params.outputs.cache-hit != 'true'
-        working-directory: ./zcash_proofs
-        run: cargo run --release --example download-params --features download-params
```
str4d, line 28: We can delete the `download-params` example now; it was only added for use in our CI. (However, we should verify that zebrad's CI isn't also using it; I think they might have been at one stage.)

str4d (approved):
ACK 3cdb0054f7f8ef1bf873cfcd2d4e7e113619d1fb

.github/workflows/ci.yml
nuttycom: `git grep` shows no usage of `get-params-path` in the zebrad repo.

.github/workflows/ci.yml
nuttycom: git grep shows no usage of download-params in the zebrad repo.

.github/workflows/ci.yml
str4d: Aha, what they actually did is copied those files into their CI. They removed their copies in https://github.com/ZcashFoundation/zebra/pull/7800.

str4d: So yes, we can just remove `zcash_proofs/examples/` entirely (I don't think we need to offer Sprout parameter downloading examples either).

nuttycom: Removed.

str4d (approved):
utACK c0a0ef1653e4806ae1e3475e10c1ede0ebaa3992.

PR #1020, opened 2023-10-13
zcash_client_sqlite: Fix incorrect note deduplication in `v_transactions`

nuttycom:
The `v_transactions` view is built upon the set of received notes, received note values being added to the balance for the transaction and spent notes being deducted from this balance. This fixes an error wherein if multiple identically-valued notes were spent in a transaction, only one of those notes' values was being counted as having been spent.


daira (approved):
This change looks correct. However, I do think it should have a test.

PR #1023, opened 2023-10-23
Refactor Sapling builder to separate out proof generation

str4d:
Closes zcash/librustzcash#741.

### DIFF
--- a/zcash_primitives/src/sapling/value/sums.rs
+++ b/zcash_primitives/src/sapling/value/sums.rs
@@ -123,6 +123,14 @@ impl Sub<&ValueCommitTrapdoor> for ValueCommitTrapdoor {
     }
 }
 
+impl Sub<TrapdoorSum> for TrapdoorSum {
+    type Output = TrapdoorSum;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        TrapdoorSum(self.0 - rhs.0)
+    }
+}
+
 impl SubAssign<&ValueCommitTrapdoor> for TrapdoorSum {
     fn sub_assign(&mut self, rhs: &ValueCommitTrapdoor) {
         self.0 -= rhs.0;
@@ -204,6 +212,14 @@ impl Sub<&ValueCommitment> for ValueCommitment {
 impl SubAssign<&ValueCommitment> for CommitmentSum {
     fn sub_assign(&mut self, rhs: &ValueCommitment) {
         self.0 -= rhs.0;
+    }
+}
+
+impl Sub<CommitmentSum> for CommitmentSum {
+    type Output = CommitmentSum;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        CommitmentSum(self.0 - rhs.0)
     }
 }
 
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -12,7 +12,11 @@ use crate::{
     keys::OutgoingViewingKey,
     legacy::TransparentAddress,
     memo::MemoBytes,
-    sapling::{self, prover::TxProver, Diversifier, Note, PaymentAddress},
+    sapling::{
+        self,
+        prover::{OutputProver, SpendProver},
+        redjubjub, Diversifier, Note, PaymentAddress,
+    },
     transaction::{
         components::{
             amount::{Amount, BalanceError},
@@ -162,6 +166,7 @@ pub struct Builder<'a, P, R> {
     // `add_sapling_spend` or `add_orchard_spend`, we will build an unauthorized, unproven
     // transaction, and then the caller will be responsible for using the spending keys or their
     // derivatives for proving and signing to complete transaction creation.
+    sapling_asks: Vec<redjubjub::PrivateKey>,
     orchard_saks: Vec<orchard::keys::SpendAuthorizingKey>,
     #[cfg(feature = "zfuture")]
     tze_builder: TzeBuilder<'a, TransactionData<Unauthorized>>,
@@ -266,6 +271,7 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
             transparent_builder: TransparentBuilder::empty(),
             sapling_builder: SaplingBuilder::new(params, target_height),
             orchard_builder,
+            sapling_asks: vec![],
             orchard_saks: Vec::new(),
             #[cfg(feature = "zfuture")]
             tze_builder: TzeBuilder::empty(),
@@ -329,7 +335,12 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
         merkle_path: sapling::MerklePath,
     ) -> Result<(), sapling_builder::Error> {
         self.sapling_builder
-            .add_spend(&mut self.rng, extsk, diversifier, note, merkle_path)
+            .add_spend(&mut self.rng, &extsk, diversifier, note, merkle_path)?;
+
+        self.sapling_asks
+            .push(redjubjub::PrivateKey(extsk.expsk.ask));
+
+        Ok(())
     }
 
     /// Adds a Sapling address to send funds to.
@@ -432,13 +443,14 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
     ///
     /// Upon success, returns a tuple containing the final transaction, and the
     /// [`SaplingMetadata`] generated during the build process.
-    pub fn build<FR: FeeRule>(
+    pub fn build<SP: SpendProver, OP: OutputProver, FR: FeeRule>(
         self,
-        prover: &impl TxProver,
+        spend_prover: &SP,
+        output_prover: &OP,
         fee_rule: &FR,
     ) -> Result<(Transaction, SaplingMetadata), Error<FR::Error>> {
         let fee = self.get_fee(fee_rule).map_err(Error::Fee)?;
-        self.build_internal(prover, fee.into())
+        self.build_internal(spend_prover, output_prover, fee.into())
     }
 
     /// Builds a transaction from the configured spends and outputs.
@@ -446,9 +458,10 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
     /// Upon success, returns a tuple containing the final transaction, and the
     /// [`SaplingMetadata`] generated during the build process.
     #[cfg(feature = "zfuture")]
-    pub fn build_zfuture<FR: FutureFeeRule>(
+    pub fn build_zfuture<SP: SpendProver, OP: OutputProver, FR: FutureFeeRule>(
         self,
-        prover: &impl TxProver,
+        spend_prover: &SP,
+        output_prover: &OP,
         fee_rule: &FR,
     ) -> Result<(Transaction, SaplingMetadata), Error<FR::Error>> {
         let fee = fee_rule
@@ -464,12 +477,13 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
             )
             .map_err(Error::Fee)?;
 
-        self.build_internal(prover, fee.into())
+        self.build_internal(spend_prover, output_prover, fee.into())
     }
 
-    fn build_internal<FE>(
+    fn build_internal<SP: SpendProver, OP: OutputProver, FE>(
         self,
-        prover: &impl TxProver,
+        spend_prover: &SP,
+        output_prover: &OP,
         fee: Amount,
     ) -> Result<(Transaction, SaplingMetadata), Error<FE>> {
         let consensus_branch_id = BranchId::for_height(&self.params, self.target_height);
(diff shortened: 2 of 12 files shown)


zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -336,22 +429,18 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
         Ok(())
     }
 
-    pub fn build<Pr: TxProver, R: RngCore>(
+    pub fn build<SP: SpendProver, OP: OutputProver, R: RngCore>(
```
nuttycom, line 432: I don't see how these type parameters are being used; they don't show up either in the inputs or outputs to this function?

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -373,204 +462,401 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+        };
+
+        // Create the unauthorized Spend and Output descriptions.
+        let shielded_spends = spend_infos
+            .into_iter()
+            .map(|a| a.build::<SP>(self.anchor))
+            .collect::<Result<Vec<_>, _>>()?;
+        let shielded_outputs = output_infos
+            .into_iter()
+            .map(|a| a.build::<P, OP, _>(&mut rng))
```
nuttycom, line 506: Okay, so I see that we're propagating the type ascription in to here. I guess that this `build` calls uses methods on `OutputProver` that don't take a `self` reference, and are instead strictly determined by the static type.

nuttycom (commented):
Flushing first-pass comments. I will need to make another pass to better verify how the new behavior compares to the old.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -439,23 +452,25 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
     ///
     /// Upon success, returns a tuple containing the final transaction, and the
     /// [`SaplingMetadata`] generated during the build process.
-    pub fn build<FR: FeeRule>(
+    pub fn build<SP: SpendProver, OP: OutputProver, FR: FeeRule>(
```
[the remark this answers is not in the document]
str4d: > I feel like these should be qualified, and the corresponding variables named, to reflect their being specific to Sapling, given that this API is part of the top-level builder.
> [...]
> Is this PR just an intermediate step?

The type names _are_ qualified, with the `sapling::prover` module prefix. And when these eventually move into the `sapling-crypto` crate, we wouldn't want to include redundant prefixes on the type or variable names. At the top-level transaction builder however (that is staying in `zcash_primitives`) I'd be fine with a prefix.

Also note that the presence of these traits is an artifact of how the Sapling transaction builder was originally constructed to allow remote proving (in environments where local proving could be too costly), as well as different types to represent loading the parameters in different ways. Orchard doesn't have these intermediate traits; it instead has concrete types for its prover. But keeping with the trait-based approach kept this (and the prior PRs) simpler to implement and review.

So in this regard I do consider this PR to be an intermediate step: the `TxBuilder` trait is unused after this PR and can be removed, and similarly we could remove the ability to do remote proving (which AFAIK has never been done beyond a demo I put together years ago) and simplify the types to be more concrete like Orchard.

> Also, if the objective is to align the Sapling and Orchard designs, why does this need the Sapling prover(s) passed in but not an Orchard prover?

For the current transaction builder implementation, the Orchard proving parameters are being constructed on-the-fly. That adds overhead to the transaction building process that we won't want in wallets, but we haven't noticed it yet because we don't have Orchard support in `zcash_client_backend`. So yes, eventually we will want to pass in the Orchard proving parameters and let the caller decide when to build them.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -504,17 +520,27 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
...
-                self.target_height,
-                self.progress_notifier.as_ref(),
-            )
-            .map_err(Error::SaplingBuild)?;
+            .build::<SP, OP, _>(&mut rng, self.target_height)
+            .map_err(Error::SaplingBuild)?
+            .map(|(bundle, tx_metadata)| {
+                // We need to create proofs before signatures, because we still support
+                // creating V4 transactions, which commit to the Sapling proofs in the
+                // transaction digest.
```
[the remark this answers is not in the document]
str4d: If you look a few lines lower, you'll see:
```rust
b.create_proof(&orchard::circuit::ProvingKey::build(), &mut rng)
```
which is where the Orchard parameters are being constructed on-the-fly like I mentioned above.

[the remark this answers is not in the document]
str4d: Separately, yes I think we should at some point deprecate creating v4 transactions. But that will require additional refactoring work elsewhere, because currently the type system forces us to always build proofs before constructing `TransactionData`, so that we are able to derive the txid no matter what version is used. We will likely need to introduce a type-level distinction between pre-v5 and v5+ transactions to address this (instead of always resolving versions at runtime like we currently do).

zcash_primitives/src/transaction/components/sapling/builder.rs
str4d: Correct. This is not necessary for Orchard because it uses a concrete type to hold its proof inputs, whereas here we need to indirectly construct that type via the trait. The aforementioned simplification to not allow remote proving would remove this need.

zcash_primitives/src/sapling/value/sums.rs
```diff
@@ -123,6 +123,14 @@ impl Sub<&ValueCommitTrapdoor> for ValueCommitTrapdoor {
     }
 }
 
+impl Sub<TrapdoorSum> for TrapdoorSum {
+    type Output = TrapdoorSum;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        TrapdoorSum(self.0 - rhs.0)
```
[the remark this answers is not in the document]
str4d: Indeed these Sapling types have been somewhat ad-hoc constructed, and we should do a full pass over both these Sapling types (some of which pre-date the `orchard` crate, and others were partly duplicated from `orchard::value` after we figured out a nice set of structs there), as well as the Orchard types (e.g. `TrapdoorSum` here is not present in `orchard`, but maybe should be).

> What's the algebraic relationship between `TrapdoorSum` and `ValueCommitTrapdoor`?

A `ValueCommitTrapdoor` is a specific `rcv` value for a specific Sapling Spend or Output, corresponding to its `cv` value in the transaction.

A `TrapdoorSum` is the sum or difference of one or more `ValueCommitTrapdoor`s, corresponding to the sum or difference of one or more Sapling Spends or Outputs. As such, it doesn't necessarily correspond to any particular `cv` value, but instead is a valid `bsk` value for that set of Spends and Outputs.

In the `orchard` crate we have `ValueSum` (a signed 64-bit integer) to distinguish from `NoteValue` (an unsigned 64-bit integer) or `ValueBalance` (a signed 63-bit integer). `TrapdoorSum` fills a similar spot in the type graph for `bsk`, and IIRC arose here from how we previously did value balancing in the `SaplingProvingContext` and `SaplingVerificationContext` types (which become obsolete after this PR, as their job is now done inside the Sapling builder). Currently in the `orchard` crate we just allow summing `ValueCommitTrapdoor`s directly into the same type, but we could choose to introduce this same distinction (or alternatively remove `TrapdoorSum`).

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -373,204 +462,401 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+            spend_prover,
+            output_prover,
+            rng,
+            progress_notifier,
+            total_progress,
+        ))
+    }
+}
+
+/// M<arker for an unauthorized bundle with no signatures.
```
nuttycom, line 684:
```suggestion
/// Marker for an unauthorized bundle with no signatures.
```

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -373,204 +462,401 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+            .iter()
+            .fold(self.prepare(&mut rng, sighash), |partial, ask| {
+                partial.sign(&mut rng, ask)
+            })
+            .finalize()
+    }
+}
+
+impl<P: InProgressProofs> Bundle<InProgress<P, PartiallyAuthorized>> {
+    /// Signs this bundle with the given [`SpendAuthorizingKey`].
```
nuttycom, line 787: The comment here says `SpendAuthorizingKey` but the type of the argument is `PrivateKey`; this is confusing.

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -373,204 +462,401 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+    fn append_signature(self, signature: &Signature) -> Result<Self, Error> {
+        let sighash = self.authorization.sigs.sighash;
+        let mut signature_valid_for = 0usize;
+        let bundle = self.map_authorization((
+            |proof| proof,
+            |proof| proof,
+            |maybe| match maybe {
+                MaybeSigned::SigningMetadata(parts) => {
+                    let rk = parts.ak.randomize(parts.alpha, SPENDING_KEY_GENERATOR);
+                    // TODO: Fix msg
```
nuttycom, line 825: I don't understand this TODO; can it be fixed now, or if not can what it's trying to say be clarified?

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -373,204 +462,401 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+}
+
+impl Bundle<InProgress<Proven, PartiallyAuthorized>> {
+    /// Finalizes this bundle, enabling it to be included in a transaction.
+    ///
+    /// Returns an error if any signatures are missing.
+    pub fn finalize(self) -> Result<Bundle<Authorized>, Error> {
+        self.try_map_authorization((
+            |p| Ok(p),
+            |p| Ok(p),
```
nuttycom, line 853:
```suggestion
            Ok,
            Ok,
```

zcash_primitives/src/transaction/components/sapling/builder.rs
```diff
@@ -373,204 +462,401 @@ impl<P: consensus::Parameters> SaplingBuilder<P> {
...
+    /// This API ensures that all signatures are created over the same sighash.
+    pub fn prepare<R: RngCore + CryptoRng>(
+        self,
+        mut rng: R,
+        sighash: [u8; 32],
+    ) -> Bundle<InProgress<P, PartiallyAuthorized>> {
+        self.map_authorization((
+            |proof| proof,
+            |proof| proof,
+            |parts| MaybeSigned::SigningMetadata(parts),
```
nuttycom, line 750:
```suggestion
            MaybeSigned::SigningMetadata,
```

nuttycom (approved):
utACK with minor suggestions & clippy fixes.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/c4d9de2effe2a054f1d10ea947c2477364c8572e..082b70af2a369fc2cd5c49b6508e820724a8a6f6) to fix merge conflicts.

zcash_primitives/src/transaction/components/sapling/builder.rs
str4d: Copy-pasta from `orchard`.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/082b70af2a369fc2cd5c49b6508e820724a8a6f6..d43cc778f7e0532bb401a31840accc0fce8c389f) to address review comments.

zcash_primitives/src/transaction/components/sapling/builder.rs
str4d: This was a TODO I forgot I hadn't done (because COVID). I've implemented a temporary fix; the real fix is to finally migrate to the `redjubjub` crate.

nuttycom (dismissed):
re-utACK https://github.com/zcash/librustzcash/pull/1023/commits/d43cc778f7e0532bb401a31840accc0fce8c389f

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/d43cc778f7e0532bb401a31840accc0fce8c389f..0989147787770557865883ff8ec316721b32f0db) to fix clippy lint.

This PR doesn't have changelog entries, but there are other changes that need to be made to the changelog that overlap / conflict with that this PR needs, so I'll make them in a separate PR.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/0989147787770557865883ff8ec316721b32f0db..927d32b63a902479d003db55d4dff09ccc4e870f) to fix copy-pasta bugs that resulted in broken intra-doc links.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/927d32b63a902479d003db55d4dff09ccc4e870f..b2ff29db78bceb1eb31c1eaed96ebbf0601471d0) to fix merge conflict.

nuttycom (approved):
utACK https://github.com/zcash/librustzcash/pull/1023/commits/b2ff29db78bceb1eb31c1eaed96ebbf0601471d0

PR #1024, opened 2023-10-25
zcash_client_backend: Add Orchard note commitment tree size to block metadata.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -437,11 +439,17 @@ impl BlockMetadata {
         self.block_hash
     }
 
-    /// Returns the size of the Sapling note commitment tree as of the block that this
+    /// Returns the size of the Sapling note commitment tree as of the end of the block that this
     /// [`BlockMetadata`] describes.
```
daira, line 443:
```suggestion
    /// Returns the size of the Sapling note commitment tree for the final treestate of the block
    /// that this [`BlockMetadata`] describes.
```
(This is the [terminology used in the protocol spec](https://zips.z.cash/protocol/protocol.pdf#transactions).)

zcash_client_backend/src/data_api.rs
```diff
@@ -437,11 +439,17 @@ impl BlockMetadata {
...
     pub fn sapling_tree_size(&self) -> u32 {
         self.sapling_tree_size
     }
+
+    /// Returns the size of the Orchard note commitment tree as of the end of the block that this
+    /// [`BlockMetadata`] describes.
```
daira, line 449:
```suggestion
    /// Returns the size of the Orchard note commitment tree for the final treestate of the block
    /// that this [`BlockMetadata`] describes.
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -363,6 +364,17 @@ where
...
+                    block
+                        .chain_metadata
+                        .as_ref()
+                        .map(|m| m.orchard_commitment_tree_size)
+                        .unwrap_or_else(|| {
+                            m.orchard_tree_size()
+                                + u32::try_from(
+                                    block.vtx.iter().map(|tx| tx.actions.len()).sum::<usize>(),
+                                )
+                                .unwrap()
```
daira, line 376:
```suggestion
                                .expect("the number of Orchard actions in a block cannot overflow u32")
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -363,6 +364,17 @@ where
                                 )
                                 .unwrap()
```
daira, line 365:
```suggestion
                                .expect("the number of Sapling outputs in a block cannot overflow u32")
```

zcash_client_backend/src/scanning.rs
```diff
@@ -365,22 +365,57 @@ pub(crate) fn scan_block_with_runner<
             if m.sapling_commitment_tree_size == 0 {
                 None
```
daira, line 366: This will be `None` if we are after Sapling activation and haven't had any Sapling outputs yet. The `or_else` below will then keep the result at `None` which will report a `ScanError::TreeSizeUnknown`. That seems wrong; should it be this?
```suggestion
                Some(0)
```

zcash_client_backend/src/scanning.rs
```diff
@@ -365,22 +365,57 @@ pub(crate) fn scan_block_with_runner<
...
             protocol: ShieldedProtocol::Sapling,
             at_height: cur_height,
         })?;
 
+    let mut orchard_commitment_tree_size = block
+        .chain_metadata
+        .as_ref()
+        .and_then(|m| {
+            if m.orchard_commitment_tree_size == 0 {
+                None
```
daira, line 395: Same comment as above; should it be this?
```suggestion
                Some(0)
```

zcash_client_backend/src/data_api/chain.rs
daira: Also, we have at least two places where the number of outputs/actions in a block is computed. Should it be a method returning `u32` (well, two methods: `output_count` and `action_count`)?

zcash_client_backend/src/scanning.rs
```diff
@@ -365,22 +365,57 @@ pub(crate) fn scan_block_with_runner<
...
-                    .sum();
-                Some(m.sapling_commitment_tree_size - block_note_count)
+                    .map(|tx| tx.outputs.len())
+                    .sum::<usize>()
+                    .try_into()
+                    .expect("Sapling output count cannot exceed a u32");
+                Some(m.sapling_commitment_tree_size - block_output_count)
             }
         })
         .or_else(|| prior_block_metadata.map(|m| m.sapling_tree_size()))
```
str4d, line 378: While looking into the off-by-one bug, I realised that this code is backwards. If we have a `prior_block_metadata`, then (because `BlockMetadata` stores non-optional tree sizes, and we migrated the SQLite database to fill in Sapling data) we are guaranteed to have the correct value, so we should just use it directly instead of recomputing from the block's `chain_metadata` (which ends up performing a no-op computation, subtracting all the output counts and then re-adding them incrementally). That would amount to swapping the order of using `prior_block_metadata` and `block.chain_metadata`.

We should separately consider checking at the end of the scan that `sapling_commitment_tree_size` is equal to whatever is in `block.chain_metadata` (if present), as another arm of the consistency check (if we didn't have `prior_block_metadata` and thus weren't doing consistency checks, then this will be a check of the no-op-ness of the `block.chain_metadata` usage here).

zcash_client_backend/src/scanning.rs
```diff
@@ -365,22 +365,57 @@ pub(crate) fn scan_block_with_runner<
...
+                    .vtx
+                    .iter()
+                    .map(|tx| tx.actions.len())
+                    .sum::<usize>()
+                    .try_into()
+                    .expect("Orchard action count cannot exceed a u32");
+                Some(m.orchard_commitment_tree_size - block_action_count)
+            }
+        })
+        .or_else(|| prior_block_metadata.map(|m| m.orchard_tree_size()))
```
str4d, line 407: Same thing: if we have `prior_block_metadata` then we should use it (because `BlockMetadata` stores the Orchard tree size non-optionally), and fall back to `block.chain_metadata`.

zcash_client_backend/src/scanning.rs
```diff
@@ -365,22 +365,57 @@ pub(crate) fn scan_block_with_runner<
...
+        .or_else(|| {
+            // If we're below Nu5 activation, or Nu5 activation is not set, the tree size is zero
+            params
+                .activation_height(NetworkUpgrade::Nu5)
+                .map_or(Some(0), |h| if h > cur_height { Some(0) } else { None })
```
str4d, line 412: Same off-by-one bug here.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -977,15 +979,33 @@ fn parse_block_metadata(
...
+        {
+            orchard_tree_size_opt.ok_or_else(|| {
+                SqliteClientError::from(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!(
+                        "Missing final Orchard note commitment tree size at height: {}",
+                        block_height
+                    ),
+                ))
+            })?
```
str4d, line 995: Isn't this going to immediately fail for existing wallets? Given that we haven't been writing Orchard commitment tree sizes into the wallet since NU5 activation, there will be a long stretch of `None`s between NU5 and whatever height the wallet was last synced to before upgrading to this code.

I currently see three pathways to resolving this issue:
- Make `orchard_tree_size` optional in `BlockMetadata`.
  - This makes the Orchard logic in the rest of the wallet less clean compared to the Sapling logic. The current non-optional approach is definitely what we would have done had we succeeded in deploying Orchard support prior to NU5 activation.
  - The "Orchard wallet birthday" is then effectively "the height at which the `orchard_commitment_tree_size` column of the `blocks` table is first non-null". (We separately need to decide how to handle pool-specific wallet birthdays in general within the wallet structure.)
- Add a required initial online migration scan, which wallets must complete before they do any other wallet operations.
  - We don't have any established API that we could hook this into, so wallets would be entirely reliant on the changelog to notice that they need to start doing this (along with occurrences of this error message, which we could enhance to point developers to the new API).
  - We could maybe detect this issue in the wallet migration code (see below), but it is not an error condition in this pathway (because any online step necessarily occurs after wallet migration, not before).
- Require wallets to force every user to recover from seed in the upgrade that deploys this code.
  - We could do this by adding a check in the wallet migration code, that checks if the most recently-scanned block is at or beyond NU5 activation and has `orchard_commitment_tree_size = NULL`.
  - This would cause expected but maybe unnecessary data loss to users (specifically of data in the `sent_notes` table that is not recoverable from chain).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -947,10 +948,11 @@ pub(crate) fn get_target_and_anchor_heights(
...
-    row: (BlockHeight, Vec<u8>, Option<u32>, Vec<u8>),
+fn parse_block_metadata<P: consensus::Parameters>(
+    params: &P,
+    row: (BlockHeight, Vec<u8>, Option<u32>, Vec<u8>, Option<u32>),
 ) -> Result<BlockMetadata, SqliteClientError> {
-    let (block_height, hash_data, sapling_tree_size_opt, sapling_tree) = row;
+    let (block_height, hash_data, sapling_tree_size_opt, sapling_tree, orchard_tree_size_opt) = row;
     let sapling_tree_size = sapling_tree_size_opt.map_or_else(|| {
         if sapling_tree == BLOCK_SAPLING_FRONTIER_ABSENT {
             Err(SqliteClientError::CorruptedData("One of either the Sapling tree size or the legacy Sapling commitment tree must be present.".to_owned()))
```
str4d, line 958: I do not remember why the `sapling_commitment_tree_size` column in the `blocks` table was made nullable, given that we ran a migration that parsed the legacy commitment tree data into that column (the same data we use here to make `sapling_tree_size` non-optional). Do we still insert the legacy Sapling commitment tree into the database even after shardtree deployment? @nuttycom if you remember why you implemented it this way, a code comment would be very useful.

str4d (changes requested):
Reviewed c10670b0291927dbfeba74c6a977857e6278887b.

zcash_client_backend/src/scanning.rs
```diff
@@ -365,22 +365,57 @@ pub(crate) fn scan_block_with_runner<
...
+        .or_else(|| {
+            // If we're below Sapling activation, or Sapling activation is not set, the tree size is zero
+            params
+                .activation_height(NetworkUpgrade::Sapling)
+                .map_or(Some(0), |h| if h > cur_height { Some(0) } else { None })
+        })
         .ok_or(ScanError::TreeSizeUnknown {
             protocol: ShieldedProtocol::Sapling,
             at_height: cur_height,
         })?;
```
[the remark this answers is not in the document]
nuttycom: Hm, I wonder if it would be better to make the comment valid again?

zcash_client_sqlite/src/wallet.rs
nuttycom: It was left nullable so that wallets connecting to older lightwalletd servers would be able to continue to function, although they would have been restricted to scanning in linear order. However, since the decision to implement it that way, we decided that connecting to an older lightwalletd is an unsupported use case; however, I'm not sure that it's worth the effort of writing a migration, particularly since Orchard tree sizes being null _is_ a supported use case, and I prefer to have the symmetry. That potentially means that we could get rid of the `CorruptedData` error here, and instead just return `None` in this case, since we have to handle the absence of the tree size at the use site (so that the error path is the same as for Orchard) anyway.

zcash_client_sqlite/src/wallet.rs
nuttycom: I took a two-pronged approach: I made `orchard_tree_size` and `sapling_tree_size` both optional in `BlockMetadata`, but both required in `ScannedBlock`.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/bb7f14fda005e949909ad894b1b3d6b15d3576bf..2b87156bb679861f71680c83242f6b27acd98b2f) to address comments from code review.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/2b87156bb679861f71680c83242f6b27acd98b2f..bb0162f10738b784dcc76f015dc61c3a24dec630) to fix Clippy complaint.

daira:
Reviewing now.

zcash_client_backend/src/data_api.rs
```diff
@@ -409,21 +409,23 @@ pub trait WalletRead {
...
 }
 
 impl BlockMetadata {
     /// Constructs a new [`BlockMetadata`] value from its constituent parts.
     pub fn from_parts(
         block_height: BlockHeight,
         block_hash: BlockHash,
-        sapling_tree_size: u32,
+        sapling_tree_size: Option<u32>,
+        orchard_tree_size: Option<u32>,
```
daira, line 422: Strictly speaking this should have a changelog entry.

zcash_client_backend/src/data_api.rs
```diff
@@ -437,37 +439,53 @@ impl BlockMetadata {
         self.block_hash
     }
 
-    /// Returns the size of the Sapling note commitment tree as of the block that this
-    /// [`BlockMetadata`] describes.
-    pub fn sapling_tree_size(&self) -> u32 {
+    /// Returns the size of the Sapling note commitment tree for the final treestate of the block
+    /// that this [`BlockMetadata`] describes.
```
daira, line 443:
```suggestion
    /// that this [`BlockMetadata`] describes, if available.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -437,37 +439,53 @@ impl BlockMetadata {
...
+    pub fn sapling_tree_size(&self) -> Option<u32> {
         self.sapling_tree_size
     }
+
+    /// Returns the size of the Orchard note commitment tree for the final treestate of the block
+    /// that this [`BlockMetadata`] describes.
```
daira, line 449:
```suggestion
    /// that this [`BlockMetadata`] describes, if available.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -437,37 +439,53 @@ impl BlockMetadata {
...
 impl<Nf> ScannedBlock<Nf> {
     /// Constructs a new `ScannedBlock`
+    #[allow(clippy::too_many_arguments)]
     pub fn from_parts(
-        metadata: BlockMetadata,
+        block_height: BlockHeight,
+        block_hash: BlockHash,
         block_time: u32,
+        sapling_tree_size: u32,
+        orchard_tree_size: u32,
```
daira, line 478: Needs a changelog entry.

zcash_client_backend/src/scanning.rs
```diff
@@ -345,42 +327,91 @@ pub(crate) fn scan_block_with_runner<
...
-                    .iter()
-                    .map(|tx| {
-                        u32::try_from(tx.outputs.len()).expect("output count cannot exceed a u32")
-                    })
-                    .sum();
-                Some(m.sapling_commitment_tree_size - block_note_count)
-            }
+    let initial_sapling_tree_size = prior_block_metadata.and_then(|m| m.sapling_tree_size());
+    // If we're below Sapling activation, or Sapling activation is not set, the
+    // tree size is zero
```
daira, line 339: This comment doesn't seem to match the code. The code sets `nu_sapling_tree_size` to `Some(0)` only when Sapling activation is set and `cur_height` is before it.

zcash_client_backend/src/scanning.rs
```diff
@@ -345,42 +327,91 @@ pub(crate) fn scan_block_with_runner<
...
+                }
+            });
+    let mut sapling_commitment_tree_size = initial_sapling_tree_size
+        .or_else(|| {
+            block
+                .chain_metadata
+                .as_ref()
+                .and_then(|m| {
+                    if m.sapling_commitment_tree_size == 0 {
+                        nu_sapling_tree_size
```
daira, line 357:
```suggestion
                        None
```
This is equivalent because of the `.or(nu_sapling_tree_size)`. But then `nu_sapling_tree` size is only used in one place, so it can be replaced by its definition (see suggestion below), which I think will be clearer.

zcash_client_backend/src/scanning.rs
nuttycom: Ah, right. I think that the code is correct; if Sapling activation is not set, then essentially any operation that attempts to affect the Sapling tree should result in an error, so the comment needs to change.

zcash_client_backend/src/scanning.rs
nuttycom: Thinking aloud: 0 is the default value for unsigned integer fields in protobuf, but before Sapling activation the size of the tree is known to be zero. However, now that I think about it, it doesn't make sense to set the tree size to Some(0) in `nu_sapling_tree_size` because what we actually want is for the code that ultimately uses this value to return an error in any case where the tree size is unknown _or_ where it would be invalid to attempt to manipulate the tree. So the `nu_` variables don't need to exist; their uses can all be replaced by `None` and then the calling code will have to change so that it doesn't prematurely force the observation of a tree size.

nuttycom: So, I went down the road of implementation described above, and realized that this won't work correctly without the tree sizes in `ScannedBlock` being non-optional. The whole check against the block metadata having a commitment tree size of zero is wrong; if we're given block metadata at all, we have to treat its given commitment tree sizes as correct because we need the case where we're after activation but the note commitment tree size is actually zero to be treated correctly.

zcash_client_backend/src/scanning.rs
```diff
@@ -345,41 +340,106 @@ pub(crate) fn scan_block_with_runner<
...
+                        .sum::<usize>()
+                        .try_into()
+                        .expect("Sapling output count cannot exceed a u32");
+
+                    if m.sapling_commitment_tree_size >= sapling_output_count {
+                        Ok(m.sapling_commitment_tree_size - sapling_output_count)
+                    } else {
+                        // The default for m.sapling_commitment_tree_size is zero, so we need to check
+                        // that the subtraction will not underflow; if it would do so, we given invalid
+                        // chain metadata for a block with Sapling outputs.
```
daira, line 386: Should this comment be before the `if`?

daira, line 386:
```suggestion
                    // The default for m.sapling_commitment_tree_size is zero, so we need to check
                    // that the subtraction will not underflow; if it would do so, we given invalid
                    // chain metadata for a block with Sapling outputs.
                    if m.sapling_commitment_tree_size >= sapling_output_count {
                        Ok(m.sapling_commitment_tree_size - sapling_output_count)
                    } else {
```

nuttycom: Ah, yeah, I forgot to move it.

zcash_client_backend/src/scanning.rs
```diff
@@ -345,41 +340,106 @@ pub(crate) fn scan_block_with_runner<
...
+                        .vtx
+                        .iter()
+                        .map(|tx| tx.outputs.len())
+                        .sum::<usize>()
+                        .try_into()
+                        .expect("Sapling output count cannot exceed a u32");
+
+                    // The default for m.sapling_commitment_tree_size is zero, so we need to check
+                    // that the subtraction will not underflow; if it would do so, we given invalid
+                    // chain metadata for a block with Sapling outputs.
```
daira, line 383:
```suggestion
                    // that the subtraction will not underflow; if it would do so, we were given
                    // invalid chain metadata for a block with Sapling outputs.
```

zcash_client_backend/src/scanning.rs
```diff
@@ -345,41 +340,106 @@ pub(crate) fn scan_block_with_runner<
...
+                        .vtx
+                        .iter()
+                        .map(|tx| tx.actions.len())
+                        .sum::<usize>()
+                        .try_into()
+                        .expect("Orchard action count cannot exceed a u32");
+
+                    // The default for m.orchard_commitment_tree_size is zero, so we need to check
+                    // that the subtraction will not underflow; if it would do so, we given invalid
+                    // chain metadata for a block with Orchard actions.
```
daira, line 429:
```suggestion
                    // that the subtraction will not underflow; if it would do so, we were given
                    // invalid chain metadata for a block with Orchard actions.
```

daira (approved):
utACK 09e37bcc8dac34c58fad8742a061b3fb62a891dd

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -326,46 +325,10 @@ where
         None
     };
 
-    let mut continuity_check_metadata = prior_block_metadata;
     block_source.with_blocks::<_, DbT::Error>(
         Some(from_height),
         Some(limit),
         |block: CompactBlock| {
-            // check block continuity
```
str4d, line 334: Huh, did we move the continuity check information into the scanner, but forget to remove it from the loader? I guess that checking it here means we avoid some unnecessary work if there's a connection problem, but I think in general it is fine integrated into the scanner.

str4d (approved):
utACK 09e37bcc8dac34c58fad8742a061b3fb62a891dd with suggestions.

str4d (approved):
utACK 7720a4fc589375a9eafd1b488ebf9c609c203b8c

PR #1033, opened 2023-11-03
Introduce newtypes for Sapling circuit parameters

str4d:


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/e96ed5bcadc693251a0974db1195d65278b19abb..6edaa178f6c66c5f7a944575e555c993f4de95e5) to fix intra-doc link.

zcash_primitives/src/sapling/circuit.rs
```diff
@@ -551,6 +553,82 @@ impl Circuit<bls12_381::Scalar> for Output {
...
+pub struct PreparedSpendVerifyingKey(pub(crate) groth16::PreparedVerifyingKey<Bls12>);
+
+/// The parameters for the Sapling Output circuit.
+pub struct OutputParameters(pub(crate) groth16::Parameters<Bls12>);
+
+impl OutputParameters {
+    /// Reads the parameters from their encoding.
+    ///
+    /// Only set `verify_point_encodings` to false if you are verifying the parameters in
+    /// another way (such as checking the hash of the parameters file on disk).
```
daira, line 601: `verify_point_encodings` just does on-curve and field element canonicity checks, right? Those should be really cheap; it's point decompression that is expensive (because it requires a square root).

str4d: `bellman::groth16::Parameters` uses uncompressed encodings.

daira: Exactly, so `verify_point_encodings` is always cheap because it only checks field element canonicity and the curve equation ($-u^2 + v^2 = 1 + d_{\mathbb{J}} \cdot u^2 \cdot v^2$). Have we measured the overhead for the Sapling spend and output parameters if we just always set it to `true`?

nuttycom (approved):
utACK c56ab0001ebe43ccf0a22e8f2597a674ec18a6ef

daira (approved):
utACK. Requires 3 ACKs since (for the verifier) this is consensus-critical code — even if it is just a refactoring.

zcash_primitives/src/sapling/circuit.rs
str4d: The curve equation in this case is for BLS12-381, as these are the proving system parameters.

The checks that `G1Affine::from_uncompressed` does (in constant time) are `is_on_curve` and `is_torsion_free`. The latter is more expensive because of multiplying the point by $x^2$.

`G2Affine::from_uncompressed` does the same checks. This time the point is only multiplied by $x$ in `is_torsion_free`, but the rest of the field arithmetic is more expensive (being in `Fp2`).

I haven't benchmarked the difference, but I presume @ebfull did when originally implementing it.

daira: Oh of course, sorry. Yeah that's a significant cost.

ebfull (approved):

PR #1035, opened 2023-11-03
CI: Build `wasm32-wasi` in a synthetic crate

str4d:
This target is incompatible with various crates that we use for testing. However, when those crates are enabled via dev-dependencies, they become part of the build graph and prevent us from building the regular code.

Downstream crates don't inherit dev-dependencies, so we simulate this directly in CI.


nuttycom (approved):
ACK, builds pass.

daira (commented):
Post-hoc ACK

PR #1036, opened 2023-11-04
zcash_primitives: Implement `Clone` on bundle marker types

str4d:
This enables unauthorized Sapling bundles to be cloned, for example when loading them into a `TransactionData` for signature hashing.


nuttycom (approved):
utACK

PR #1037, opened 2023-11-09
Update `Cargo.lock` file with upgraded dependency versions.

nuttycom:


Cargo.lock
```diff
@@ -965,7 +955,7 @@ checksum = "5a03ba7d4c9ea41552cd4351965ff96883e629693ae85005c501bb4b9e1c48a7"
 dependencies = [
  "lazy_static",
  "rand_core",
- "ring",
+ "ring 0.16.20",
```
str4d, line 958: Hmm, it's not great that we end up with two copies of `ring` in the dependency tree, but this is blocked on https://github.com/jjyr/hdwallet/pull/21.

str4d (approved):
utACK

daira (commented):
Post-hoc ACK. New duplicate dependencies are likely to be temporary.

PR #1039, opened 2023-11-09
Refactor Sapling components and builder into `zcash_primitives::sapling`

### LINKED ISSUE #736: Internally refactor Sapling tx components and builder to `zcash_primitives::sapling`
As part of the effort to recreate the `sapling-crypto` crate, we want to collect all of its code under `zcash_primitives::sapling`. Some of it currently lives under `zcash_primitives::transaction::components::sapling`, and we should refactor it to match the `orchard::{bundle, builder, ...}` layout. We can do this by moving them into `pub(crate)` submodules, and re-exporting the types in their current locations (so the `zcash_primitives` API is unaffected).

str4d:
Closes #736.

### DIFF
--- a/zcash_primitives/src/sapling/bundle.rs
+++ b/zcash_primitives/src/sapling/bundle.rs
@@ -1,0 +1,736 @@
+use core::fmt::Debug;
+
+use memuse::DynamicUsage;
+
+use zcash_note_encryption::{
+    EphemeralKeyBytes, ShieldedOutput, COMPACT_NOTE_SIZE, ENC_CIPHERTEXT_SIZE,
+};
+
+use crate::{
+    consensus,
+    sapling::{
+        note::ExtractedNoteCommitment,
+        note_encryption::{CompactOutputDescription, SaplingDomain},
+        redjubjub::{self, PublicKey, Signature},
+        value::ValueCommitment,
+        Nullifier,
+    },
+    transaction::components::{Amount, GROTH_PROOF_SIZE},
+};
+
+pub type GrothProofBytes = [u8; GROTH_PROOF_SIZE];
+
+/// Defines the authorization type of a Sapling bundle.
+pub trait Authorization: Debug {
+    type SpendProof: Clone + Debug;
+    type OutputProof: Clone + Debug;
+    type AuthSig: Clone + Debug;
+}
+
+/// Authorizing data for a bundle of Sapling spends and outputs, ready to be committed to
+/// the ledger.
+#[derive(Debug, Copy, Clone)]
+pub struct Authorized {
+    // TODO: Make this private.
+    pub binding_sig: redjubjub::Signature,
+}
+
+impl Authorization for Authorized {
+    type SpendProof = GrothProofBytes;
+    type OutputProof = GrothProofBytes;
+    type AuthSig = redjubjub::Signature;
+}
+
+/// A map from one bundle authorization to another.
+///
+/// For use with [`Bundle::map_authorization`].
+pub trait MapAuth<A: Authorization, B: Authorization> {
+    fn map_spend_proof(&mut self, p: A::SpendProof) -> B::SpendProof;
+    fn map_output_proof(&mut self, p: A::OutputProof) -> B::OutputProof;
+    fn map_auth_sig(&mut self, s: A::AuthSig) -> B::AuthSig;
+    fn map_authorization(&mut self, a: A) -> B;
+}
+
+/// The identity map.
+///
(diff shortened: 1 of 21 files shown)


str4d:
Best reviewed locally with `git show COMMIT --color-moved=zebra`.

str4d:
I haven't moved `zcash_primitives::transaction::components::sapling::fees` yet, because I'm not entirely sure how to handle it (as the traits directly use amount types that will not be moving into `sapling-crypto`).

nuttycom:
> I haven't moved `zcash_primitives::transaction::components::sapling::fees` yet, because I'm not entirely sure how to handle it (as the traits directly use amount types that will not be moving into `sapling-crypto`). I suspect it will stay where it is?

Yeah, I think keeping this where it is makes sense for now. TBH these traits could probably be renamed and just moved (along with the sibling traits for transparent and TZE) into `zcash_primitives::transaction::fees` itself.

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -477,196 +103,39 @@ impl SpendDescription<Authorized> {
...
-impl<A> std::fmt::Debug for OutputDescription<A> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
-        write!(
-            f,
-            "OutputDescription(cv = {:?}, cmu = {:?}, ephemeral_key = {:?})",
-            self.cv, self.cmu, self.ephemeral_key
-        )
+        Ok(SpendDescriptionV5::from_parts(cv, nullifier, rk))
     }
 }
```
nuttycom, line 141: It's weird to have this `impl OutputDescription<GrothProofBytes>` here when neither `OutputDescription` nor `GrothProofBytes` are defined in this module. I'm surprised it doesn't result in an orphan rule violation now, but it certainly will once the `sapling` module is moved out to the `sapling_crypto` crate, so I think another commit fixing this putative orphan rule violation would be good to include with this PR.

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -694,45 +163,34 @@ impl OutputDescription<GrothProofBytes> {
...
-    cv: ValueCommitment,
-    cmu: ExtractedNoteCommitment,
-    ephemeral_key: EphemeralKeyBytes,
-    enc_ciphertext: [u8; 580],
-    out_ciphertext: [u8; 80],
-}
-
-memuse::impl_no_dynamic_usage!(OutputDescriptionV5);
-
 impl OutputDescriptionV5 {
```
nuttycom, line 194: Same comment with this impl.

nuttycom (changes requested):
I think that the soon-to-be orphan-rule-violating `impl`s in `transaction::components::sapling` should be fixed as part of this PR.

zcash_primitives/src/transaction/components/sapling.rs
str4d: I was planning to change these in a separate PR, to keep the diff on this PR smaller.

nuttycom (approved):
utACK 043ad55

nuttycom (approved):
utACK c6dd9ad858f3efadde50f30edf9072ee7a6882f6

zcash_primitives/src/sapling/bundle.rs
```diff
@@ -0,0 +1,736 @@
...
+    pub(crate) fn cv_mut(&mut self) -> &mut ValueCommitment {
+        &mut self.cv
+    }
+    pub(crate) fn cmu_mut(&mut self) -> &mut ExtractedNoteCommitment {
+        &mut self.cmu
+    }
+    pub(crate) fn ephemeral_key_mut(&mut self) -> &mut EphemeralKeyBytes {
+        &mut self.ephemeral_key
+    }
+    pub(crate) fn enc_ciphertext_mut(&mut self) -> &mut [u8; 580] {
```
daira, line 518: Could this be `[u8; ENC_CIPHERTEXT_SIZE]`?

zcash_primitives/src/sapling/bundle.rs
```diff
@@ -0,0 +1,736 @@
...
+        &mut self.enc_ciphertext
+    }
+    pub(crate) fn out_ciphertext_mut(&mut self) -> &mut [u8; 80] {
```
daira, line 521: Could this be `[u8; OUT_CIPHERTEXT_SIZE]`?

daira: (and similarly for other instances of these constants above)

PR #1040, opened 2023-11-10
Use `{ENC,OUT}_CIPHERTEXT_SIZE` constants in `zcash_primitives::sapling::bundle` and `zcash_primitives::transaction::components::sapling`

daira:


str4d (approved):
utACK

PR #1042, opened 2023-11-21
Fix generated protobuf file to match source

str4d:
A comment change to the source file was made in zcash/librustzcash#891 before merging, but the generated file was not updated to match.


nuttycom (approved):
utACK, doc-only

Issue #738, opened 2022-12-13
Extract Sapling protocol into `sapling-crypto` crate


str4d:
Back when the Sapling protocol was first developed, it lived in two places:
- A `sapling-crypto` crate for the circuit logic.
- A `librustzcash` crate that exposed the circuit logic to `zcashd` via FFI methods that contained a bunch of logic.

We refactored these to produce the `zcash_*` crates, and eventually got rid of both of the above crates (`sapling-crypto` entirely, `librustzcash` moved into `zcashd` as a pure FFI layer).

However, with the introduction of the `orchard` crate (which is kept in a separate repository for licensing reasons), we now run into a dependency problem: the `orchard` crate depends on `zcash_note_encryption` which is in this workspace, and `zcash_primitives` depends on `orchard`. This makes updates very tricky.

Instead, we want to fix this by moving the `components/*` libraries (everything that exists "below" the shielded protocols) out of this workspace, and also move the Sapling protocol into a separate crate to emulate the `orchard` crate positioning in the dependency graph. We can use this as an opportunity to resurrect the `sapling-crypto` crate!

This issue is not for moving `sapling-crypto` back into its old GitHub repository (the dependency issue only requires moving the component crates). We may still choose to do that, but it would be after this issue.

PR #1043, opened 2023-11-21
Refactor ZIP 32 code

str4d:
The Sapling-specific code moves to `zcash_primitives::sapling::zip32`, to be part of the `sapling-crypto` extraction.

The protocol-independent types are refactored to support this refactor, and to enable themselves to be refactored into a protocol component crate that can be used by both `sapling-crypto` and `orchard`.

Part of #738.


zcash_primitives/src/zip32.rs
```diff
@@ -43,41 +48,148 @@ impl ConditionallySelectable for AccountId {
...
+            n => Self::non_hardened(n),
         }
     }
+
+    /// Constructs a non-hardened `ChildIndex` from the given value.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `value >= (1 << 31)`.
+    pub fn non_hardened(value: u32) -> Self {
```
nuttycom, line 173: Why have this panic instead of return `Option<Self>`?

zcash_primitives/src/zip32.rs
```diff
@@ -43,41 +48,148 @@ impl ConditionallySelectable for AccountId {
...
-    fn value(&self) -> u32 {
-        match *self {
-            ChildIndex::Hardened(i) => i + (1 << 31),
-            ChildIndex::NonHardened(i) => i,
+    /// Constructs a hardened `ChildIndex` from the given value.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `value >= (1 << 31)`.
+    pub fn hardened(value: u32) -> Self {
```
nuttycom, line 143: Why panic instead of return `Option<Self>`?

zcash_primitives/src/zip32.rs
```diff
@@ -43,41 +48,148 @@ impl ConditionallySelectable for AccountId {
...
+impl private::Sealed for Hard {
+    fn master() -> Self {
+        // Hardened-only derivation still uses the non-hardened child index of 0 for the
+        // master key, for historic reasons. This only appears in the extended spending
+        // key encoding format, and is not used anywhere.
+        Self(0)
+    }
+
+    fn hardened(index: u32) -> Self {
+        Self(index)
```
nuttycom, line 77: This can only be implemented in this module, but it can be invoked from anywhere; shouldn't this method be fallible with a restricted range of possible inputs?

str4d:
In a pairing with @nuttycom, we decided to simplify this by just removing non-hardened derivation support from our Sapling implementation (which we have never used, and to our knowledge no one else widely does either).

daira:
I thought I remembered a dependency of [ZIP 316](https://zips.z.cash/zip-0316) on non-hardened derivation. Indeed there is, but only *transparent* non-hardened derivation, not Sapling.

zcash_primitives/src/zip32.rs
str4d: This is for simplicity, as its use case is basically always with constants. The API now has `ChildIndex::from_index` again, which is now fallible.

zcash_primitives/src/zip32.rs
str4d: This method was on a sealed trait. Users outside the crate can't name the trait, so they can't use the method.

The trait has now been removed.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/1f6c0a786233aa4da07f5eed7a8814c6a6186b39..f40337a7b600f0a9f0e0518b2ffdc3be352f4220) to remove non-hardened derivation and simplify the resulting refactor.

zcash_primitives/src/sapling/zip32.rs
```diff
@@ -1193,506 +1151,431 @@ mod tests {
             },
             TestVector {
                 ask: Some([
-                    0x28, 0x2b, 0xc1, 0x97, 0xa5, 0x16, 0x28, 0x7c, 0x8e, 0xa8, 0xf6, 0x8c, 0x42,
```
str4d, line 1196: These updated test vectors come from https://github.com/zcash/zcash-test-vectors/pull/94.

nuttycom: I was just about to ask about that. I also noticed that the URL at the top of this block of test vectors should now change.

zcash_primitives/CHANGELOG.md
```diff
@@ -131,6 +139,9 @@ and this library adheres to Rust's notion of
     - `fees::fixed::FeeRule::fixed_fee`
     - `fees::zip317::FeeRule::marginal_fee`
     - `sighash::TransparentAuthorizingContext::input_amounts`
+- `zcash_primitives::zip32`:
+  - `ChildIndex` has been changed from an enum to an opaque struct, and no
+    longer supports non-hardened indices.
```
daira, line 144: Do you want to add a note to ZIP 32 that discourages or deprecates use of non-hardened indices with Sapling?

I feel like this requires more community discussion. I'll raise it at the ZIP sync meeting today.

zcash_primitives/src/sapling/zip32.rs
```diff
@@ -231,12 +229,34 @@ impl DiversifierKey {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum KeyIndex {
+    Master,
+    Child(ChildIndex),
+}
```
daira, line 236: I'm confused by this type, which doesn't seem to correspond to anything in the ZIP. The path to a key is a sequence of child indices, and so the master key corresponds to the empty path, no?

str4d: `KeyIndex::Master` corresponds to the middle paragraph of [this section of ZIP 32](https://zips.z.cash/zip-0032#sapling-extended-spending-keys):

> For the master extended spending key, `depth` is 0, `parent_fvk_tag` is 4 zero bytes, and `i` is 0.

Its only purpose is to enable the key corresponding to the path `m` (i.e. the empty path) to be serialized.

zcash_primitives/src/sapling/zip32.rs
str4d: Yeah, it would have changed automatically had the test vector repo been updated when the repo was moved.

zcash_primitives/src/sapling/zip32.rs
daira: Got it. I know this is internal, but please add a comment explaining its usage.

zcash_primitives/CHANGELOG.md
str4d: ZIP 32's existing language already somewhat discourages it, particularly the language that specifies Orchard. We only specify non-hardened derivation for an optional `address_index` path extension for Sapling with MAY, and then immediately note that `zcashd` did this path extension with hardened derivation instead (which made sense for its use case, and also means I still am unaware of anyone ever exercising this MAY).

I also note that while we do still use a non-hardened-like derivation process to produce Sapling internal keys, it is domain-separated from non-hardened derivation and thus distinct (so we don't use the functionality being removed here).

zcash_primitives/src/sapling/zip32.rs
```diff
@@ -329,7 +350,14 @@ impl ExtendedSpendingKey {
         let depth = reader.read_u8()?;
```
daira, line 350: Pre-existing typo in comment: s/define/defined/

zcash_primitives/src/sapling/keys.rs
```diff
@@ -33,6 +33,8 @@ pub enum DecodingError {
     InvalidAsk,
     /// Could not decode the `nsk` bytes to a jubjub field element.
     InvalidNsk,
+    /// A non-hardened extended spending key was found, which is unsupported.
```
daira, line 36:
```suggestion
    /// An extended spending key had an unsupported child index: either a non-hardened
    /// index, or a non-zero index at depth 0.
```

zcash_primitives/src/sapling/zip32.rs
```diff
@@ -231,12 +229,34 @@ impl DiversifierKey {
...
+enum KeyIndex {
+    Master,
+    Child(ChildIndex),
+}
+
+impl KeyIndex {
+    fn new(depth: u8, i: u32) -> Option<Self> {
+        match i {
+            0 if depth == 0 => Some(KeyIndex::Master),
+            _ => ChildIndex::from_index(i).map(KeyIndex::Child),
```
daira, line 242: This allows non-zero indices at depth 0, which I think is invalid. (If it were not invalid then the child key at depth 0 would be ambiguous with the master key.)

zcash_primitives/src/sapling/zip32.rs
```diff
@@ -329,7 +350,14 @@ impl ExtendedSpendingKey {
         let depth = reader.read_u8()?;
         let mut tag = [0; 4];
         reader.read_exact(&mut tag)?;
-        let i = reader.read_u32::<LittleEndian>()?;
+        let child_index = reader.read_u32::<LittleEndian>().and_then(|i| {
+            KeyIndex::new(depth, i).ok_or_else(|| {
+                io::Error::new(
+                    io::ErrorKind::Unsupported,
+                    "Non-hardened keys are not supported",
```
daira, line 357: This error message might not be correct if we make the depth 0, index non-zero case an error.

zcash_primitives/src/sapling/zip32.rs
```diff
@@ -231,12 +229,34 @@ impl DiversifierKey {
...
+
+impl KeyIndex {
+    fn new(depth: u8, i: u32) -> Option<Self> {
+        match i {
+            0 if depth == 0 => Some(KeyIndex::Master),
+            _ => ChildIndex::from_index(i).map(KeyIndex::Child),
+        }
+    }
+
+    fn index(&self) -> u32 {
```
daira, line 246: Document that this index has the $2^{31}$ hardened bit set. (ZIP 32, following BIP 32, uses a poorly chosen notation that obscures whether that bit is part of the index or not. If you read carefully it is for child indices, but not for account indices, so it's easy to get confused.)

str4d: In this PR, I clean up the types and methods to consistently use "index" to mean the ZIP 32 notion, and "value" to mean the 31-bit value without the hardening bit.

zcash_primitives/src/zip32.rs
```diff
@@ -43,41 +48,54 @@ impl ConditionallySelectable for AccountId {
...
 
 impl ChildIndex {
-    pub fn from_index(i: u32) -> Self {
-        match i {
-            n if n >= (1 << 31) => ChildIndex::Hardened(n - (1 << 31)),
-            n => ChildIndex::NonHardened(n),
+    /// Parses the given ZIP 32 child index.
+    ///
+    /// Returns `None` if the hardened bit is not set.
+    pub fn from_index(i: u32) -> Option<Self> {
```
nuttycom, line 61: I would almost prefer that this operation be called `decode` or something of the sort, pursuant to my previous comment.

nuttycom (dismissed):
utACK modulo discussing the naming confusion around "index"

zcash_primitives/src/zip32.rs
nuttycom: Can this now be `pub const`?

zcash_primitives/src/zip32.rs
```diff
@@ -43,41 +48,54 @@ impl ConditionallySelectable for AccountId {
...
+        Self(value + (1 << 31))
     }
 
-    fn value(&self) -> u32 {
-        match *self {
-            ChildIndex::Hardened(i) => i + (1 << 31),
-            ChildIndex::NonHardened(i) => i,
-        }
+    /// Returns the index as a 32-bit integer, including the hardened bit.
+    pub fn index(&self) -> u32 {
```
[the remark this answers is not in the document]
nuttycom: Hm. That's unfortunate naming in the ZIP then, but I suppose we're probably stuck with it and this should agree with the ZIP.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/a86299c4b4b71cf69fef57b0cebc7657e4e561f8..b81b8ce8440fb3031c1fcccf595dd05d02bc7dc0) to fix merge conflicts.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/b81b8ce8440fb3031c1fcccf595dd05d02bc7dc0..7064efe697688b5d1912be8a134e36e56a581936) to make `ChildIndex::hardened` a `const fn`.

nuttycom (approved):
utACK 7064efe697688b5d1912be8a134e36e56a581936

zcash_primitives/src/zip32.rs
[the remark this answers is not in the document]
str4d: The "index" terminology is [inherited from BIP 32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys).

Issue #1044, opened 2023-11-21
Refactor `zcash_primitives::sapling` module to not depend on anything else in the crate


str4d:

daira:
Was this done? It is marked as blocking #738 which is closed.

str4d:
This was completed, yes. The last PR working on this issue was #1064.

PR #1047, opened 2023-11-23
zcash_primitives: Make `value_balance` generic in `sapling::Bundle`

str4d:
This removes the dependency on `Amount`, and matches how we handle this in the `orchard` crate.


nuttycom (dismissed):
utACK cdd20e85831dd454c6e9e6a656e15ec39bc0e256

nuttycom (approved):
re-utACK cdd20e8

PR #1048, opened 2023-11-23
Add Network Upgrade 6 to `consensus::NetworkUpgrade`

tomekpiotrowski:
I've started working on adding Network Upgrade 6 to Zebra. I need to add it to `zcash_primitives` as well as `zebra-chain` and `zebra-scan` both depend on it.

In this code NU6 references Tx V5. This will need to be updated to reference Tx V6 and ZIP-230 once Tx V6 is implemented.

Let me know if this is not the right way to proceed here.

cc @teor2345

### DIFF
--- a/zcash_primitives/src/consensus.rs
+++ b/zcash_primitives/src/consensus.rs
@@ -212,6 +212,7 @@ impl Parameters for MainNetwork {
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_000)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_046_400)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_687_104)),
+            #[cfg(feature = "unstable-nu6")]
             NetworkUpgrade::Nu6 => None,
             #[cfg(feature = "zfuture")]
             NetworkUpgrade::ZFuture => None,
@@ -264,6 +265,7 @@ impl Parameters for TestNetwork {
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_800)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_028_500)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_842_420)),
+            #[cfg(feature = "unstable-nu6")]
             NetworkUpgrade::Nu6 => None,
             #[cfg(feature = "zfuture")]
             NetworkUpgrade::ZFuture => None,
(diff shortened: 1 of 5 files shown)


zcash_primitives/src/consensus.rs
```diff
@@ -212,6 +212,7 @@ impl Parameters for MainNetwork {
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_000)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_046_400)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_687_104)),
+            NetworkUpgrade::Nu6 => Some(BlockHeight(2_726_400)),
```
str4d, line 215: NU6 has no activation height on mainnet.
```suggestion
            NetworkUpgrade::Nu6 => None,
```

zcash_primitives/src/consensus.rs
```diff
@@ -263,6 +264,7 @@ impl Parameters for TestNetwork {
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_800)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_028_500)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_842_420)),
+            NetworkUpgrade::Nu6 => Some(BlockHeight(2_726_400)),
```
str4d, line 267: NU6 has no activation height on testnet.
```suggestion
            NetworkUpgrade::Nu6 => None,
```

zcash_primitives/src/consensus.rs
```diff
@@ -498,6 +509,7 @@ impl TryFrom<u32> for BranchId {
             0xf5b9_230b => Ok(BranchId::Heartwood),
             0xe9ff_75a6 => Ok(BranchId::Canopy),
             0xc2d6_d0b4 => Ok(BranchId::Nu5),
+            0xc8e71055 => Ok(BranchId::Nu6),
```
str4d, line 512: For clarity, any branch ID we set here is just a placeholder ID, because we only set the final ID (which is always chosen randomly) once the final feature set is implemented and a mainnet activation height is committed to. Also note that if we commit to a mainnet activation height, and then later change it, the branch ID changes (because we change it any time the consensus rules change).

str4d (changes requested):
We should consider placing all this behind an `unstable-nu6` feature flag, so we aren't accidentally affecting the consensus rules of production nodes.

PR #1050, opened 2023-11-28
Avoid passing duplicate diversifier information to Sapling builder.

nuttycom:
The note provided to `add_sapling_spend` contains the recipient address, and we can extract the diversifier from this address, so we should not pass it separately.

### DIFF
--- a/zcash_primitives/src/sapling/builder.rs
+++ b/zcash_primitives/src/sapling/builder.rs
@@ -74,7 +74,6 @@ impl fmt::Display for Error {
 #[derive(Debug, Clone)]
 pub struct SpendDescriptionInfo {
     proof_generation_key: ProofGenerationKey,
-    diversifier: Diversifier,
     note: Note,
     alpha: jubjub::Fr,
     merkle_path: MerklePath,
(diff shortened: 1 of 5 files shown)


zcash_primitives/src/sapling/builder.rs
```diff
@@ -370,7 +367,6 @@ impl SaplingBuilder {
         &mut self,
         mut rng: R,
         extsk: &ExtendedSpendingKey,
-        diversifier: Diversifier,
```
str4d, line 373: Needs changelog entry.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -328,12 +328,11 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
     pub fn add_sapling_spend(
         &mut self,
         extsk: sapling::zip32::ExtendedSpendingKey,
-        diversifier: Diversifier,
```
str4d, line 331: Needs changelog entry.

str4d (approved):
utACK 85b09ec412b9cf096341e09e36d18d12c86d818b after changelog update.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/85b09ec412b9cf096341e09e36d18d12c86d818b..28174e1ba30b8a43906c8f1193dd0a12ef953105) to add a changelog entry.

str4d (approved):
utACK 28174e1ba30b8a43906c8f1193dd0a12ef953105

PR #1053, opened 2023-11-29
Fix dependencies to repair `cargo -p` subproject builds.

nuttycom:


daira (approved):
utACK

daira (approved):
Tested ACK for all crates (`zcash_{address,client_backend,client_sqlite,encoding,extensions,history,primitives,proofs}`, `equihash`, `f4jumble`).

PR #1054, opened 2023-11-30
Add regtest support to `Parameter` for Mobile SDKs

### LINKED ISSUE #1051: REGTEST Support for Mobile SDKs
In order to get REGTEST support on the mobile SDKs, certain changes must be accommodated.  I don't have a 100% certainty of what they are but I'll start pouring some of them here.
#### Add an enum that represents Regtest distinctively
```
pub enum Network {
    MainNetwork,
    TestNetwork,
    RegTestNetwork,
}
```
#### Provide Regtest variants to Parameters

`pub trait Parameters: Clone `

#### Make NetworkUpgrade Parameters variable.

REGTEST can be set up with different activation heights.

A RegTestNetwork struct should be able to provide these values according to the
environment that the caller is expecting.

````Rust
pub const MAIN_NETWORK: MainNetwork = MainNetwork;

impl Parameters for MainNetwork {
    fn activation_height(&self, nu: NetworkUpgrade) -> Option<BlockHeight> {
        match nu {
            NetworkUpgrade::Overwinter => Some(BlockHeight(347_500)),
            NetworkUpgrade::Sapling => Some(BlockHeight(419_200)),
            NetworkUpgrade::Blossom => Some(BlockHeight(653_600)),
            NetworkUpgrade::Heartwood => Some(BlockHeight(903_000)),
            NetworkUpgrade::Canopy => Some(BlockHeight(1_046_400)),
            NetworkUpgrade::Nu5 => Some(BlockHeight(1_687_104)),
            #[cfg(feature = "zfuture")]
            NetworkUpgrade::ZFuture => None,
        }
    }
````

#### Provide Regtest-specific`Parameters` trait values  

the following values are constant and part of the `Parameters` trait

````Rust
 fn coin_type(&self) -> u32 {
        constants::mainnet::COIN_TYPE
    }

    fn address_network(&self) -> Option<zcash_address::Network> {
        Some(zcash_address::Network::Regtest)
    }

    fn hrp_sapling_extended_spending_key(&self) -> &str {
        constants::regtest::HRP_SAPLING_EXTENDED_SPENDING_KEY
    }

    fn hrp_sapling_extended_full_viewing_key(&self) -> &str {
        constants::regtest::HRP_SAPLING_EXTENDED_FULL_VIEWING_KEY
    }

    fn hrp_sapling_payment_address(&self) -> &str {
        constants::maregtestinnet::HRP_SAPLING_PAYMENT_ADDRESS
    }

    fn b58_pubkey_address_prefix(&self) -> [u8; 2] {
        constants::regtest::B58_PUBKEY_ADDRESS_PREFIX
    }

    fn b58_script_address_prefix(&self) -> [u8; 2] {
        constants::regtest::B58_SCRIPT_ADDRESS_PREFIX
    }
````

### `zcash_client_backend`

Add regtest encoding tests to `encoding.rs` 
add regtest examples to `zip321.rs`

### `zcash_client_sqlite`
no changes

### `zcash_extensions`
no changes

### `zcash_history`
no changes

### `zcash_address`

I don't think changes are needed *but* this assertion on `zcash_address::Network::Regtest` 
might not hold after the changes in `zcash_primitives`

````Rust
    /// For some address types there is no distinction between test and regtest encodings;
    /// those will always be parsed as `Network::Test`.
    Regtest,
 ````

### `zcash_encoding`

No changes

### 
## Related GitHub Issues:
https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/issues/1233
https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/issues/1314

### PR DESCRIPTION
**Pull request by pacu**
closes #1051

### DIFF
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -54,6 +54,16 @@ and this library adheres to Rust's notion of
 - `impl {Clone, PartialEq, Eq} for zcash_primitives::memo::Error`
 - `impl {PartialEq, Eq} for zcash_primitives::sapling::note::Rseed`
 - `impl From<TxId> for [u8; 32]`
+- Added feature `local-consensus` to crate `zcash_primitives`
+  - Adds `pub mod local_consensus`
+  - `local_consensus::FullNodeParameters` adds functionality to refer to Full Node
+    parameters that can be either (public) Consensus nodes like `Network::MainNetwork`
+    or `Network::TestNetwork`, or a `Local(P)` with a set of Parameters `P`. `Regtest` 
+    is considered a `LocalConsensus`.
+  - `zcash_primitives::local_consensus::LocalNetwork` provides a type for specifying 
+    network upgrade activation heights for a given `FullNodeParameters::Local(P)`
+  - `impl Parameters for LocalNetwork` that use provided activation heights and
+    defaults to `constants::regtest::` for everything else. 
 
 ### Changed
 - `zcash_primitives::transaction`:
@@ -112,6 +122,9 @@ and this library adheres to Rust's notion of
 - `zcash_primitives::zip32`:
   - `ChildIndex` has been changed from an enum to an opaque struct, and no
     longer supports non-hardened indices.
+- `zcash_client_backend` changes related to `local-consensus` feature:
+  - added tests that verify `zip321` supports Payment URIs with `Local(P)`
+  network parameters.
 
 ### Removed
 - `zcash_primitives::constants`:
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -56,12 +56,10 @@ and this library adheres to Rust's notion of
 - `impl From<TxId> for [u8; 32]`
 - Added feature `local-consensus` to crate `zcash_primitives`
   - Adds `pub mod local_consensus`
-  - `local_consensus::FullNodeParameters` adds functionality to refer to Full Node
-    parameters that can be either (public) Consensus nodes like `Network::MainNetwork`
-    or `Network::TestNetwork`, or a `Local(P)` with a set of Parameters `P`. `Regtest` 
-    is considered a `LocalConsensus`.
   - `zcash_primitives::local_consensus::LocalNetwork` provides a type for specifying 
-    network upgrade activation heights for a given `FullNodeParameters::Local(P)`
+    network upgrade activation heights for a local or specific configuration of a full
+    node. Developers can make use of this type when connecting to a Regtest node by
+    replicating the activation heights used on their node configuration.
   - `impl Parameters for LocalNetwork` that use provided activation heights and
     defaults to `constants::regtest::` for everything else. 
 
--- a/zcash_primitives/Cargo.toml
+++ b/zcash_primitives/Cargo.toml
@@ -132,6 +132,8 @@ unstable-nu6 = []
 ## Exposes early in-development features that are not yet planned for any network upgrade.
 zfuture = []
 
+## Exposes support for working with a local consensus (e.g. regtest
+local-consensus = []
 [lib]
 bench = false
 
--- a/zcash_primitives/src/consensus.rs
+++ b/zcash_primitives/src/consensus.rs
@@ -688,11 +688,10 @@ pub mod testing {
 
 #[cfg(test)]
 mod tests {
-    use std::convert::TryFrom;
-
     use super::{
         BlockHeight, BranchId, NetworkUpgrade, Parameters, MAIN_NETWORK, UPGRADES_IN_ORDER,
     };
+    use std::convert::TryFrom;
 
     #[test]
     fn nu_ordering() {
--- a/zcash_primitives/src/lib.rs
+++ b/zcash_primitives/src/lib.rs
@@ -26,7 +26,8 @@ pub mod merkle_tree;
 use sapling;
 pub mod transaction;
 pub use zip32;
-pub mod zip339;
-
 #[cfg(feature = "zfuture")]
 pub mod extensions;
+#[cfg(feature = "local-consensus")]
+pub mod local_consensus;
+pub mod zip339;
--- a/zcash_primitives/src/local_consensus.rs
+++ b/zcash_primitives/src/local_consensus.rs
@@ -1,16 +1,7 @@
 use crate::{
-    consensus::{BlockHeight, Network, NetworkUpgrade, Parameters},
+    consensus::{BlockHeight, NetworkUpgrade, Parameters},
     constants,
 };
-
-/// Represents the Network Parameters of the Full Node the caller connects to.
-pub enum FullNodeParameters<P: Parameters> {
-    /// Variant indicating a Zcash Full Node running on public Mainnet or Testnet
-    Consensus(Network),
-    /// The consensus parameters of a local or non-public network like one or more nodes
-    /// in `regtest` mode.
-    Local(P),
-}
 
 /// a `LocalNetwork` setup should define the activation heights
 /// of network upgrades. `None` is considered as "not activated"
@@ -60,15 +51,10 @@ pub struct LocalNetwork {
 
 /// Parameters default implementation for `LocalNetwork`
 /// Important note:
-/// The following functions return `constants::regtest` values
-/// ```
-/// fn coin_type()
-/// fn address_network()
-/// fn hrp_sapling_extended_spending_key()
-/// fn hrp_sapling_extended_full_viewing_key()
-/// fn hrp_sapling_payment_address()
-/// fn b58_script_address_prefix()
-/// ```
+/// The functions `coin_type()`, `address_network()`,  
+/// `hrp_sapling_extended_spending_key()`, `hrp_sapling_extended_full_viewing_key()`,
+/// `hrp_sapling_payment_address()`, `b58_script_address_prefix()` return 
+/// `constants::regtest` values
 impl Parameters for LocalNetwork {
     fn activation_height(&self, nu: NetworkUpgrade) -> Option<BlockHeight> {
         match nu {
--- a/zcash_primitives/src/local_consensus.rs
+++ b/zcash_primitives/src/local_consensus.rs
@@ -53,7 +53,7 @@ pub struct LocalNetwork {
 /// Important note:
 /// The functions `coin_type()`, `address_network()`,  
 /// `hrp_sapling_extended_spending_key()`, `hrp_sapling_extended_full_viewing_key()`,
-/// `hrp_sapling_payment_address()`, `b58_script_address_prefix()` return 
+/// `hrp_sapling_payment_address()`, `b58_script_address_prefix()` return
 /// `constants::regtest` values
 impl Parameters for LocalNetwork {
     fn activation_height(&self, nu: NetworkUpgrade) -> Option<BlockHeight> {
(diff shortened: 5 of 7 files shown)

### CODE REVIEW

File: zcash_primitives/src/consensus.rs [hunk 1]
```diff
@@ -301,6 +301,14 @@ impl Parameters for TestNetwork {
...
     MainNetwork,
     TestNetwork,
+    RegtestNetwork {
+        overwinter: Option<BlockHeight>,
+        sapling: Option<BlockHeight>,
+        blossom: Option<BlockHeight>,
+        heartwood: Option<BlockHeight>,
+        canopy: Option<BlockHeight>,
+        nu5: Option<BlockHeight>,
+    },
```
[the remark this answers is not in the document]
pacu: Thanks @str4d for your review. I agree with this quote below.
> I think regtest support (or more generally, user-configurable test networks) should be provided with a separate type that downstreams can opt into, similar to what we do in zcashd, and that lives somewhere other than zcash_primitives::consensus (maybe zcash_primitives::testing)

The only thing I would differ is that we need to be able to use regtest from clients to have the integration tests like the advanced ReOrg tests. So we would need to access this RegtestNetwork types from the FFI.

I chose consensus::Network mostly for two reasons:
1. other crates were referring to it 
2. [The FFI uses that](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/blob/310c1ec43d59fa7b6a2962e30fd344c028bd1291/rust/src/lib.rs#L54-L68)

And no others, so I'm definitely ok with refactoring this so that it's more convenient for today's reality.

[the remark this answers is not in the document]
str4d: > The FFI uses it 

To clarify, the FFI itself uses numeric constants for the kinds of networks it knows about, and then [converts those into a `consensus::Network`](https://github.com/Electric-Coin-Company/zcash-light-client-ffi/blob/310c1ec43d59fa7b6a2962e30fd344c028bd1291/rust/src/lib.rs#L2686). So the way for the iOS SDK to opt in to testing with a configurable local network is to change that helper function to use the new type that supports local testing. That change makes sense because the iOS SDK won't be able to do local testing without a bunch of other changes on its Swift side.

> other crates were referring to it

If we change `consensus::Network` then some other crates using it may be forced to do that same bunch of work (depending on how they use the enum) that they don't want to do (perhaps because they are only concerned with the consensus networks).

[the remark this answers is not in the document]
pacu: @str4d I think the changes I pushed might be heading into the right direction

zcash_primitives/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub mod consensus;
 pub mod constants;
 pub mod keys;
 pub mod legacy;
+pub mod local_consensus;
```
nuttycom, line 17: Should this whole module be feature-flagged off by default?

nuttycom: This feature flag will need to be added to Cargo.toml.

nuttycom:
@pacu sorry that we didn't get this in earlier; it's now conflicted, so can you please rebase and I'll re-review?

pacu:
> @pacu sorry that we didn't get this in earlier; it's now conflicted, so can you please rebase and I'll re-review?

No worries. I'll rebase it! Thanks

zcash_primitives/src/local_consensus.rs
```diff
@@ -0,0 +1,242 @@
+use crate::{
+    consensus::{BlockHeight, Network, NetworkUpgrade, Parameters},
+    constants,
+};
+
+pub enum FullNodeParameters<P: Parameters> {
```
nuttycom, line 6: This enum should be documented.

zcash_primitives/src/local_consensus.rs
```diff
@@ -0,0 +1,242 @@
...
+    Consensus(Network),
+    Local(P),
+}
+
+#[derive(Clone, PartialEq, Eq, Copy, Debug)]
+pub struct LocalNetwork {
```
nuttycom, line 12: Document

zcash_primitives/src/lib.rs
nuttycom: Note the addition of the feature flag and the module in the `CHANGELOG`

zcash_primitives/src/consensus.rs
```diff
@@ -365,6 +365,10 @@ impl Parameters for Network {
             Network::TestNetwork => TEST_NETWORK.b58_script_address_prefix(),
         }
     }
+
+    fn is_nu_active(&self, nu: NetworkUpgrade, height: BlockHeight) -> bool {
```
nuttycom, line 369: Why is this override of the default impl necessary?

nuttycom (changes requested):
Sorry I missed this before, but the additions need rustdoc & CHANGELOG entries, and the feature flag needs to be added to Cargo.toml

zcash_primitives/src/consensus.rs
pacu: uhmmm this doesn't seem to be in the right place!  great catch! I'll look into it

zcash_primitives/CHANGELOG.md
```diff
@@ -54,6 +54,15 @@ and this library adheres to Rust's notion of
 - `impl {Clone, PartialEq, Eq} for zcash_primitives::memo::Error`
 - `impl {PartialEq, Eq} for zcash_primitives::sapling::note::Rseed`
 - `impl From<TxId> for [u8; 32]`
+- Added feature `local-consensus` to crate `zcash_primitives`
+  - Adds `pub mod local_consensus`
+  - `local_consensus::FullNodeParameters` adds functionality to refer to Full Node 
+  parameters that can be either (public) Consensus nodes like `Network::MainNetwork` or `Network::TestNetwork`, or a `Local(P)` with a set of Parameters `P`. `Regtest` 
+  is considered a `LocalConsensus`.
```
nuttycom, line 61: nit: please wrap.

zcash_primitives/src/consensus.rs
```diff
@@ -688,11 +688,10 @@ pub mod testing {
 
 #[cfg(test)]
 mod tests {
-    use std::convert::TryFrom;
```
nuttycom, line 691: Superfluous change?

zcash_primitives/src/local_consensus.rs
```diff
@@ -0,0 +1,284 @@
...
+/// running on public Mainnet
+/// `Consensus(Network::TestNetwork)` refers to a Zcash Full Node
+/// running on public Testnet
+/// `Local(P)` can be used to define the consensus parameters of
+/// a local or non-public network like one or more nodes on
+/// `regtest` mode.
+pub enum FullNodeParameters<P: Parameters> {
+    Consensus(Network),
+    Local(P),
+}
```
nuttycom, line 18:
```suggestion
/// Represents the Network Parameters of the Full Node the caller connects to.
pub enum FullNodeParameters<P: Parameters> {
    /// Variant indicating a Zcash Full Node running on public Mainnet or Testnet
    Consensus(Network),
    /// The consensus parameters of a local or non-public network like one or more nodes 
    /// in `regtest` mode.
    Local(P),
}
```

Who is "the caller" here?

zcash_primitives/src/local_consensus.rs
```diff
@@ -0,0 +1,284 @@
...
+    pub blossom: Option<BlockHeight>,
+    pub heartwood: Option<BlockHeight>,
+    pub canopy: Option<BlockHeight>,
+    pub nu5: Option<BlockHeight>,
+    pub nu6: Option<BlockHeight>,
+    #[cfg(feature = "zfuture")]
+    pub z_future: Option<BlockHeight>,
+}
+
+impl Parameters for LocalNetwork {
```
nuttycom, line 66: Document that this `Parameters` impl hardcodes network constants other than activation heights to the `regtest` parameters.

nuttycom (approved):
utACK modulo a few additional documentation suggestions.

zcash_primitives/src/local_consensus.rs
pacu: how do you refer to those who are using librustzcash in the documentation?

zcash_primitives/src/consensus.rs
pacu: that probably came from `cargo fmt`

nuttycom:
@pacu builds are currently failing?

zcash_primitives/src/local_consensus.rs
nuttycom: Hm, actually, `FullNodeParameters` is a bit misleading; this is basically just `NetworkParameters` or something. It doesn't really pertain to full nodes specifically, just whatever network the client is connecting to. 

I'm now actually a little puzzled why a concrete enum is needed for this type at all; since everything is bounded on `consensus::Parameters`, don't you just create a new concrete type for your network parameters and use that in the caller? Under what circumstances do you need to switch between parameters at runtime (as this enum lets you do)?

pacu:
> @pacu builds are currently failing?

They don't fail for me locally but I'm not running an ubuntu target. It seemed intermittent to me. 

do we have some script to build all the different targets or is that something you folks delegate to CI?

zcash_primitives/src/local_consensus.rs
pacu: Going back in the conversation of this change the original idea was to follow the current enum approach were one would specify MainNetwork or TestNetwork and add a Regtest, that didn't make much sense because Regtest is not consensus but a particular Node setup. So FullNodeParameters came up as a way to still group the possible node setups a client could connect to without changing the existing API fo `consensus::`.

Particularly, the SDKs' `RustBackend` would receive this FullNodeParameters that could be `Consensus` or `Local` (regtest or other kind of setup with different activation heights, etc.). What you suggest is another way of achieving the same. The mobile SDKs would need update anyway so we could do that as well.

nuttycom (approved):
utACK 09bbc79

zcash_primitives/Cargo.toml
```diff
@@ -132,6 +132,8 @@ unstable-nu6 = []
 ## Exposes early in-development features that are not yet planned for any network upgrade.
 zfuture = []
 
+## Exposes support for working with a local consensus (e.g. regtest
```
daira, line 135:
```suggestion
## Exposes support for working with a local consensus (e.g. regtest).
```

zcash_primitives/src/local_consensus.rs
```diff
@@ -0,0 +1,276 @@
+use crate::{
+    consensus::{BlockHeight, NetworkUpgrade, Parameters},
+    constants,
+};
+
+/// a `LocalNetwork` setup should define the activation heights
+/// of network upgrades. `None` is considered as "not activated"
+/// These heights are not validated. Callers shall initialized
```
daira, line 8:
```suggestion
/// A `LocalNetwork` setup should define the activation heights
/// of network upgrades. `None` is considered as "not activated"
/// These heights are not validated. Callers should initialize
```

zcash_primitives/src/local_consensus.rs
```diff
@@ -0,0 +1,276 @@
...
+///         sapling: Some(BlockHeight::from_u32(1)),
+///         blossom: Some(BlockHeight::from_u32(1)),
+///         heartwood: Some(BlockHeight::from_u32(1)),
+///         canopy: Some(BlockHeight::from_u32(1)),
+///         nu5: Some(BlockHeight::from_u32(1)),
+///         #[cfg(feature = "unstable-nu6")]
+///         nu6: Some(BlockHeight::from_u32(1)),
+///         #[cfg(feature = "zfuture")]
+///         z_future: Some(BlockHeight::from_u32(1)),
+///         };
```
daira, line 36:
```suggestion
///     };
```
Non-blocking.

daira (approved):
utACK, non-blocking doc-only suggestions.

PR #1055, opened 2023-11-30
Replace `SaplingReceivedNote` with `ReceivedNote`

nuttycom:
This generalizes the data structure used to represent spendable notes to handle both Sapling and Orchard notes.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -30,8 +30,8 @@ and this library adheres to Rust's notion of
     functionality and move it behind the `transparent-inputs` feature flag.
 - `zcash_client_backend::fees::standard`
 - `zcash_client_backend::wallet`:
-  - `ReceivedSaplingNote::from_parts`
-  - `ReceivedSaplingNote::{txid, output_index, diversifier, rseed, note_commitment_tree_position}`
+  - `WalletNote`
+  - `ReceivedNote`
 - `zcash_client_backend::zip321::TransactionRequest::total`
 - `zcash_client_backend::zip321::parse::Param::name`
 - `zcash_client_backend::proto::`
@@ -60,6 +60,13 @@ and this library adheres to Rust's notion of
   - Arguments to `BlockMetadata::from_parts` have changed to include Orchard.
   - `BlockMetadata::sapling_tree_size` now returns an `Option<u32>` instead of
     a `u32` for consistency with Orchard.
+  - `WalletShieldedOutput` has an additiona type parameter which is used for
+    key scope. `WalletShieldedOutput::from_parts` now takes an additional
+    argument of this type.
+  - `WalletTx` has an additional type parameter as a consequence of the 
+    `WalletShieldedOutput` change.
+  - `ScannedBlock` has an additional type parameter as a consequence of the 
+    `WalletTx` change.
   - Arguments to `ScannedBlock::from_parts` have changed.
   - `ScannedBlock::metadata` has been renamed to `to_block_metadata` and now
     returns an owned value rather than a reference.
@@ -174,6 +181,8 @@ and this library adheres to Rust's notion of
     - `WalletTransparentOutput::value`
 
 ### Removed
+- `zcash_client_backend::wallet::ReceivedSaplingNote` has been replaced by
+  `zcash_client_backend::ReceivedNote.
 - `zcash_client_backend::data_api::WalletRead::is_valid_account_extfvk` has been
   removed; it was unused in the ECC mobile wallet SDKs and has been superseded by
   `get_account_for_ufvk`.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -49,6 +49,12 @@ and this library adheres to Rust's notion of
      wallet::input_selection::{Proposal, SaplingInputs},
    }`
 
+### Moved
+- `zcash_client_backend::data_api::{PoolType, ShieldedProtocol}` have
+  been moved into the `zcash_client_backend` root module.
+- `zcash_client_backend::data_api::{NoteId, Recipient}` have
+  been moved into the `zcash_client_backend::wallet` module.
+
 ### Changed
 - `zcash_client_backend::data_api`:
   - Arguments to `BlockMetadata::from_parts` have changed to include Orchard.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -32,6 +32,7 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::wallet`:
   - `WalletNote`
   - `ReceivedNote`
+  - `WalletSaplingOutput::recipient_key_scope`
 - `zcash_client_backend::zip321::TransactionRequest::total`
 - `zcash_client_backend::zip321::parse::Param::name`
 - `zcash_client_backend::proto::`
@@ -60,7 +61,7 @@ and this library adheres to Rust's notion of
   - Arguments to `BlockMetadata::from_parts` have changed to include Orchard.
   - `BlockMetadata::sapling_tree_size` now returns an `Option<u32>` instead of
     a `u32` for consistency with Orchard.
-  - `WalletShieldedOutput` has an additiona type parameter which is used for
+  - `WalletShieldedOutput` has an additional type parameter which is used for
     key scope. `WalletShieldedOutput::from_parts` now takes an additional
     argument of this type.
   - `WalletTx` has an additional type parameter as a consequence of the 
@@ -182,7 +183,7 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::wallet::ReceivedSaplingNote` has been replaced by
-  `zcash_client_backend::ReceivedNote.
+  `zcash_client_backend::ReceivedNote`.
 - `zcash_client_backend::data_api::WalletRead::is_valid_account_extfvk` has been
   removed; it was unused in the ECC mobile wallet SDKs and has been superseded by
   `get_account_for_ufvk`.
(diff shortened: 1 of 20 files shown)


zcash_client_backend/CHANGELOG.md
```diff
@@ -157,6 +170,8 @@ and this library adheres to Rust's notion of
     - `WalletTransparentOutput::value`
 
 ### Removed
+- `zcash_client_backend::wallet::ReceivedSaplingNote` has been replaced by
+  `zcash_client_backend::ReceivedNote.
```
daira, line 174:
```suggestion
  `zcash_client_backend::ReceivedNote`.
```

zcash_client_backend/src/wallet.rs
```diff
@@ -15,17 +15,66 @@ use zcash_primitives::{
...
+    }
+
+    /// Returns the index of this note within its transaction's corresponding list of
+    /// shielded outputs.
+    pub fn output_index(&self) -> u16 {
+        self.output_index
+    }
+}
+
+/// A type that represents the recipient of a transaction output; a recipient address (and, for
```
daira, line 58:
```suggestion
/// A type that represents the recipient of a transaction output: a recipient address (and, for
```

zcash_client_backend/src/wallet.rs
```diff
@@ -15,17 +15,66 @@ use zcash_primitives::{
...
+/// unified addresses, the pool to which the payment is sent) in the case of outgoing output, or an
```
daira, line 59:
```suggestion
/// unified addresses, the pool to which the payment is sent) in the case of an outgoing output, or an
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -248,6 +248,7 @@ mod tests {
                 memo BLOB,
                 spent INTEGER,
                 commitment_tree_position INTEGER,
+                recipient_key_scope INTEGER NOT NULL DEFAULT 0,
```
daira, line 251: * Add a comment that this is External scope.
* Do we actually need this DEFAULT?

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        // Add commitment tree sizes to block metadata.
+        debug!("Adding new columns");
+        transaction.execute_batch(
+            &format!(
+                "ALTER TABLE sapling_received_notes ADD COLUMN recipient_key_scope INTEGER NOT NULL DEFAULT {};",
```
daira, line 53:
```suggestion
                "ALTER TABLE sapling_received_notes ADD COLUMN recipient_key_scope INTEGER NOT NULL DEFAULT 0;",
```
I think that if we specify the default here then it doesn't need to be specified as a property of the column.

zcash_client_sqlite/src/wallet/init.rs
daira: According to @str4d this is just a side effect of how Sqlite implements the ALTER COLUMN.

daira (dismissed):
utACK

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -590,26 +585,24 @@ where
...
-                })?;
-
-                let key = match scope {
-                    Scope::External => usk.sapling().clone(),
-                    Scope::Internal => usk.sapling().derive_internal(),
-                };
-
-                builder.add_sapling_spend(key, note, merkle_path)?;
+                match selected.note() {
+                    WalletNote::Sapling(note) => {
```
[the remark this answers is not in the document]
nuttycom: That's because `proposal` needs to change. But doing so was making this PR too big, so this is an incremental step in that direction.

zcash_client_backend/src/wallet.rs
```diff
@@ -119,9 +168,10 @@ pub struct WalletSaplingOutput<N> {
     is_change: bool,
     note_commitment_tree_position: Position,
     nf: N,
+    recipient_key_scope: S,
```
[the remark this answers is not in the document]
str4d: The reason for that polymorphism (added in 8f03208439db1613739bb5bcf84e6791445088c5) was specifically to allow scanning with just `ivk`s (rather than `dfvk`s from which `ivk`s of known scope could be derived) while still enabling performance improvements. Put another way, the axis of polymorphism was whether we knew the scopes at all, not what protocol the scopes belonged to. So:
- I still think we should only use `zip32::Scope` for both Sapling and Orchard.
- Given that you want to report here the scope the output was received in, we could instead make this `recipient_key_scope: Option<zip32::Scope>` to allow for the case where we scanned with just `ivk`s with unknown scope (represented as `None`). But we can also leave in this genericism here, and just document the axis (i.e. that this will either be `zip32::Scope` or `()` depending on what kind of scanning was done).

zcash_client_backend/src/data_api/wallet.rs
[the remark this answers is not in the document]
str4d: If this isn't going to be in the next PR, then please add a doc comment with a `TODO` etc.

zcash_client_backend/src/wallet.rs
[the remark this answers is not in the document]
nuttycom: Using `Option<zip32::Scope>` complicates things because then we can't have the storage/retrieval methods typed to guarantee that a scope is always present. I don't mind documenting the scope of the variation, but TBH I think it's unnecessary, because the use sites all explicitly type it to `zip32::Scope` where a scope value is known to exist. Such documentation is also misleading, because it claims something that isn't enforced by the type, and therefor isn't an invariant - the caller can use whatever type they personally want for scope, so long as they "map" it to the desired type for the methods that require it to be so refined.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/cad4f25..8bba26bcbc2e150340218ee2c34a34737a324d0d)  to accept suggested changes.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -82,9 +88,19 @@ impl ReceivedSaplingOutput for DecryptedOutput<Note> {
     fn note_commitment_tree_position(&self) -> Option<Position> {
         None
     }
+    fn recipient_key_scope(&self) -> Scope {
+        if self.transfer_type == TransferType::WalletInternal {
```
str4d, line 92: There was a prior issue where `TransferType::WalletInternal` does not directly correlate with either `is_change` or the decryption key scope. Now that this PR is introducing a migration to explicitly store the recipient key scope, how is the way `TransferType::WalletInternal` gets handled being altered to fix the prior issue?

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        [shardtree_support::MIGRATION_ID].into_iter().collect()
+    }
+
+    fn description(&self) -> &'static str {
+        "Add support for receiving storage of note commitment tree data using the `shardtree` crate."
```
str4d, line 41: This description doesn't match the migration. Copy-pasta?

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        // Add commitment tree sizes to block metadata.
```
str4d, line 49: 📋🍝

str4d (changes requested):
Reviewed 8bba26bcbc2e150340218ee2c34a34737a324d0d.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: I checked that all of the places that `TransferType::WalletInternal` can arise are, at least at present, a consequence of funds being sent to the internal-scoped IVK. So I think that this is fine.

`is_change` might be false for shielding transactions while still being wallet-internal. But that's fine.

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -0,0 +1,111 @@
+//! This migration adds support for the Orchard protocol to `zcash_client_sqlite`
```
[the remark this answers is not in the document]
nuttycom: Ah, yeah, originally I had thought that I would do all of the changes required for Orchard in one go, but decided that this could be factored out.

zcash_client_sqlite/src/wallet/sapling.rs
str4d: What about for historic change outputs created before the introduction of internal-scoped keys? Are these correctly detected as `is_change = true`?

str4d: In Slack, @nuttycom told me that there is no change to the `is_change` problem for older wallets. This might be what #936 is tracking.

str4d (approved):
utACK ca54b3489d48ff6b9d93c81e39307b57ac0fece7

PR #1056, opened 2023-11-30
zcash_primitives: Replace `sapling::redjubjub` with `redjubjub` crate

str4d:
As a side-effect, we remove the ability to verify individual transactions with pre-ZIP 216 rules (which we already removed from `zcashd` consensus nodes in zcash/zcash#6000 and zcash/zcash#6399, as all pre-ZIP 216 transactions on mainnet and testnet are also valid under ZIP 216).


zcash_primitives/src/sapling/builder.rs
```diff
@@ -752,11 +742,7 @@ impl<P: InProgressProofs, V> Bundle<InProgress<P, Unsigned>, V> {
             MaybeSigned::SigningMetadata,
             |auth: InProgress<P, Unsigned>| InProgress {
                 sigs: PartiallyAuthorized {
-                    binding_signature: auth.sigs.bsk.sign(
-                        &sighash,
-                        &mut rng,
-                        VALUE_COMMITMENT_RANDOMNESS_GENERATOR,
-                    ),
+                    binding_signature: auth.sigs.bsk.sign(&mut rng, &sighash),
```
str4d, line 745: There was a bug here previously in that this didn't include key prefixing, because we never implemented that inside our `redjubjub` module (as the original RedDSA draft spec lacked it). I should have used the `sapling::spend_sig` method here when I added this in #1023, but as we haven't cut a release including this API yet, the bug is not in deployed code. The `redjubjub` crate includes key prefixing.

zcash_primitives/src/sapling/verifier/batch.rs
```diff
@@ -68,21 +68,11 @@ impl BatchValidator {
...
                 spend.rk(),
-                &sighash,
-                spend.spend_auth_sig(),
                 zkproof,
                 self,
-                |this, rk, _, spend_auth_sig| {
-                    let rk = redjubjub::VerificationKeyBytes::<redjubjub::SpendAuth>::from(
-                        rk.0.to_bytes(),
-                    );
-                    let spend_auth_sig = {
```
nuttycom, line 79: This is a nice simplification. The name overloading (while referring to a different type) confused me a bit at the use site.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -334,8 +335,11 @@ impl<'a, P: consensus::Parameters, R: RngCore + CryptoRng> Builder<'a, P, R> {
         self.sapling_builder
             .add_spend(&mut self.rng, &extsk, note, merkle_path)?;
 
-        self.sapling_asks
-            .push(redjubjub::PrivateKey(extsk.expsk.ask));
+        // TODO: store a `redjubjub::SigningKey` inside `extsk`.
```
nuttycom, line 338: I'm not sure that I agree with this TODO; it seems like it would be storing duplicate data. If the cost of recomputing the signing key is a problem then I guess that caching it is okay, but we should motivate such caching with a benchmark.

nuttycom (dismissed):
utACK de1ed210517524ad63adfd0a6ab08fa9d201ef76

zcash_primitives/src/transaction/builder.rs
str4d: It wouldn't be storing duplicate data; it would be replacing the `ask` currently stored there. This is equivalent to how the `orchard` key structs work.

zcash_primitives/src/sapling/builder.rs
```diff
@@ -117,10 +114,11 @@ impl SpendDescriptionInfo {
         // Construct the value commitment.
         let cv = ValueCommitment::derive(self.note.value(), self.rcv.clone());
 
-        let ak = PublicKey(self.proof_generation_key.ak.into());
+        let ak = redjubjub::VerificationKey::try_from(self.proof_generation_key.ak.to_bytes())
+            .expect("valid points are valid verification keys");
```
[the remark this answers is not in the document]
str4d: The introduction of `SpendValidatingKey` now creates this distinction.

str4d:
There's a somewhat concerning test vector failure in `zcashd` (https://github.com/zcash/zcash/pull/6784#pullrequestreview-1759696989) that I need to debug.

nuttycom (approved):
utACK ded09f99b56914d60a70f9a8de02ea714dc2818d

zcash_primitives/src/sapling/keys.rs
```diff
@@ -196,7 +367,7 @@ impl FullViewingKey {
     pub fn from_expanded_spending_key(expsk: &ExpandedSpendingKey) -> Self {
         FullViewingKey {
             vk: ViewingKey {
-                ak: SPENDING_KEY_GENERATOR * expsk.ask,
+                ak: (&expsk.ask).into(),
```
daira, line 370: I would prefer this to be an explicit method, not a `From`/`Into` impl.

zcash_primitives/src/sapling/keys.rs
```diff
@@ -38,14 +42,175 @@ pub enum DecodingError {
...
+
+/// A key used to validate spend authorization signatures.
+///
+/// Defined in [Zcash Protocol Spec § 4.2.2: Sapling Key Components][saplingkeycomponents].
+///
+/// [saplingkeycomponents]: https://zips.z.cash/protocol/protocol.pdf#saplingkeycomponents
+#[derive(Clone, Debug)]
+pub struct SpendValidatingKey(redjubjub::VerificationKey<SpendAuth>);
+
+impl From<&SpendAuthorizingKey> for SpendValidatingKey {
```
daira, line 138: Please change this to an explicitly named method.

zcash_primitives/src/sapling/test_vectors/signatures.rs
```diff
@@ -0,0 +1,476 @@
...
+    pub(crate) alpha: [u8; 32],
+    pub(crate) rsk: [u8; 32],
+    pub(crate) rvk: [u8; 32],
+    pub(crate) m: [u8; 32],
+    pub(crate) sig: [u8; 64],
+    pub(crate) rsig: [u8; 64],
+}
+
+pub(crate) fn make_test_vectors() -> Vec<TestVector> {
+    // From https://github.com/zcash/zcash-test-vectors/blob/master/zcash_test_vectors/sapling/redjubjub.py
```
daira, line 13: I have not checked these.

daira: I will check them when I review https://github.com/zcash/zcash-test-vectors/pull/95 .

zcash_primitives/src/sapling/keys.rs
daira: Filed https://github.com/zcash/librustzcash/pull/1056 .

daira (approved):
utACK

PR #1057, opened 2023-11-30
Expose Orchard value in balance API & add Orchard components to ScannedBlock

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -95,6 +95,9 @@ pub struct AccountBalance {
     /// The value of unspent Sapling outputs belonging to the account.
     pub sapling_balance: Balance,
 
+    /// The value of unspent Orchard outputs belonging to the account.
+    pub orchard_balance: Balance,
```
str4d, line 99: This change needs a changelog entry.

zcash_client_backend/src/data_api.rs
```diff
@@ -109,14 +112,43 @@ impl AccountBalance {
...
     /// Returns the total value of funds belonging to the account.
     pub fn total(&self) -> NonNegativeAmount {
-        (self.sapling_balance.total() + self.unshielded)
+        (self.sapling_balance.total() + self.orchard_balance.total() + self.unshielded)
+            .expect("Account balance cannot overflow MAX_MONEY")
+    }
+
+    /// Returns the total value of shielded (Sapling and Orchard) funds that may immediately be
+    /// spent.
+    pub fn spendable_value(&self) -> NonNegativeAmount {
```
str4d, line 127: This addition needs a changelog entry.

zcash_client_backend/src/data_api.rs
```diff
@@ -109,14 +112,43 @@ impl AccountBalance {
...
+        (self.sapling_balance.spendable_value + self.orchard_balance.spendable_value)
+            .expect("Account balance cannot overflow MAX_MONEY")
+    }
+
+    /// Returns the total value of change and/or shielding transaction outputs that are awaiting
+    /// sufficient confirmations for spendability.
+    pub fn change_pending_confirmation(&self) -> NonNegativeAmount {
```
str4d, line 134: This addition needs a changelog entry.

zcash_client_backend/src/data_api.rs
```diff
@@ -109,14 +112,43 @@ impl AccountBalance {
...
+        (self.sapling_balance.change_pending_confirmation
+            + self.orchard_balance.change_pending_confirmation)
+            .expect("Account balance cannot overflow MAX_MONEY")
+    }
+
+    /// Returns the value of shielded funds that are not yet spendable because additional scanning
+    /// is required before it will be possible to derive witnesses for the associated notes.
+    pub fn value_pending_spendability(&self) -> NonNegativeAmount {
```
str4d, line 142: This addition needs a changelog entry.

zcash_client_backend/src/data_api.rs
```diff
@@ -109,14 +112,43 @@ impl AccountBalance {
...
+        (self.sapling_balance.value_pending_spendability
+            + self.orchard_balance.value_pending_spendability)
             .expect("Account balance cannot overflow MAX_MONEY")
     }
+
+    /// Returns the total value of unspent transparent transaction outputs belonging to the wallet.
+    pub fn unshielded(&self) -> NonNegativeAmount {
```
str4d, line 149: This addition needs a changelog entry (and is duplicative of the public field, which further implies that `AccountBalance` should be opaque).

str4d (changes requested):
Reviewed 9aec53eec9cfb3cdbc488b8e1b6068f745439f45.

zcash_client_backend/src/data_api.rs
nuttycom: The reason that `AccountBalance` was originally made with public fields was to facilitate its incremental computation by mutation of the struct, but I guess that maybe just a local mutable tuple (or local struct) is better.

str4d (approved):
utACK 214a3750c5a5079ef57a84fa2a664418e15c7cda

PR #1061, opened 2023-12-07
zcash_primitives: Introduce type-safe `PRF^expand`

str4d:


nuttycom (approved):
utACK - neat macro technique for the appearance of arity-polymorphic `with`.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/c5ce4756f1b3e2fe3dcf021e4a03e87b87a34c82..ce7b7df0cca40d91910545926d4d3d5863dc7a27) to use the `zcash_spec` crate.

zcash_primitives/src/legacy/keys.rs
```diff
@@ -111,11 +112,8 @@ impl AccountPubKey {
     ///
     /// [transparent-ovk]: https://zips.z.cash/zip-0316#deriving-internal-keys
     pub fn ovks_for_shielding(&self) -> (InternalOvk, ExternalOvk) {
-        let i_ovk = prf_expand_vec(
-            &self.0.chain_code,
-            &[&[0xd0], &self.0.public_key.serialize()],
-        );
-        let i_ovk = i_ovk.as_bytes();
+        let i_ovk = PrfExpand::TRANSPARENT_ZIP316_OVK
+            .with(&self.0.chain_code, &self.0.public_key.serialize());
```
nuttycom, line 116: One minor complaint about the `with` macro: all the macro noise makes it somewhat difficult to understand the difference here, in terms of how the domain separators are included. It would be nice for the source documentation around the macro to give some examples of the code it generates.

nuttycom (approved):
utACK with request for additional documentation in `zcash_spec`.

zcash_primitives/src/legacy/keys.rs
str4d: What the macro is generating is:
```rust
fn with(self, sk: &[u8], a: &[u8; A], b: &[u8; B] -> [u8; 64] {
    self.apply(sk, &[a, b])
}
```
i.e. `PrfExpand::with` is a type safety wrapper (that takes fixed-length inputs) around `PrfExpand::apply` (the latter allows you to use arbitrary-length inputs, so we don't expose it publicly).

daira (commented):
Post-hoc ACK

PR #1062, opened 2023-12-07
zcash_primitives: Remove and relocate `InputView` traits.

nuttycom:


zcash_primitives/src/sapling/builder.rs
```diff
@@ -154,7 +143,7 @@ impl SpendDescriptionInfo {
 /// A struct containing the information required in order to construct a
 /// Sapling output to a transaction.
 #[derive(Clone)]
-struct SaplingOutputInfo {
+pub struct SaplingOutputInfo {
```
nuttycom, line 146: nit: I'd love to come up with better names for these before making them public. The `Info` suffix is just meaningless.

str4d (approved):
utACK 4eaa41affa61c9412c7419e1ea8e8f750d3a3b9b

PR #1063, opened 2023-12-07
zcash_client_backend: Rename `RecipientAddress` to `Address`

nuttycom:
Also, rename `Address::Shielded` to `Address::Sapling`

This addresses an old `TODO`, now's as good a time to do it as any.


pacu:
I'm super curious about the rationale behind this change. Is it because `RecipientAddress`  could only be created from an input that had to be parsed? How if you derive from keys? Would that be a different struct?  🤯

daira (dismissed):
utACK

str4d (approved):
utACK 9ac3594bbb1aef29bb084eade61180a4d2f0296f

PR #1064, opened 2023-12-08
zcash_primitives: Move tx builder progress type into `sapling::builder`

str4d:


zcash_primitives/src/sapling/builder.rs
```diff
@@ -571,11 +568,39 @@ impl InProgressProofs for Proven {
...
+        self.end
+    }
+}
+
 struct CreateProofs<'a, SP: SpendProver, OP: OutputProver, R: RngCore> {
     spend_prover: &'a SP,
     output_prover: &'a OP,
     rng: R,
-    progress_notifier: Option<&'a Sender<Progress>>,
+    progress_notifier: Option<&'a Sender<ProverProgress>>,
```
nuttycom, line 603: Instead of taking a concrete `Sender` struct, could this take a type bounded on a trait that can be implemented for `Sender`? That way it can be implemented as a noop, and if `Sender` is unavailable due to feature flagging or something of the sort, the type constraints can still be satisfied.

nuttycom (changes requested):
Question about the use of `Sender` versus a generic capability is blocking

zcash_primitives/src/sapling/builder.rs
nuttycom: For example, someone might want to use an implementation of this capability that simply aggregated progress information in memory, rather than sending it over a channel.

nuttycom (approved):
utACK fa8de4f

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/fa8de4fba08d12c0154d74aa3ab7db13880f46df..b6ee98ed46f6fbba97dbe741edd706b6787a52b7) to fix trait impls to not break after `sapling-crypto` extraction.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -152,7 +157,7 @@ impl Progress {
 }
 
 /// Generates a [`Transaction`] from its inputs and outputs.
-pub struct Builder<'a, P, R> {
+pub struct Builder<'a, P, R, U: sapling::builder::ProverProgress> {
```
nuttycom, line 160: I presume that when/if we implement a progress notifier for Orchard, that we'd bound on both those traits?

One question, how does this end up working if we feature-flag Sapling off?

nuttycom (approved):
re-utACK b6ee98e with a future-feature-flagging question.

zcash_primitives/src/transaction/builder.rs
str4d: That's gonna start getting messy very quickly, which is partly why I did not go down the trait pathway to begin with, and only did so at your request.

It's probably doable by replacing this bound with a top-level insulatory bound, and then having an intermediate top-level struct that wraps the Sapling and Orchard type impls and applies relevant feature-flagging.

PR #1065, opened 2023-12-08
zcash_primitives: Add explicit control over padding to the transaction builder.

nuttycom:

PR #1066, opened 2023-12-08
Factor out common logic from zip317 and fixed-fee single output change strategies.

nuttycom:


zcash_client_backend/src/fees/fixed.rs
```diff
@@ -52,115 +52,19 @@ impl ChangeStrategy for SingleOutputChangeStrategy {
...
-                    .num_outputs(sapling_outputs.len() + 1),
-                //Orchard is not yet supported in zcash_client_backend
-                0,
-            )
-            .unwrap(); // fixed::FeeRule::fee_required is infallible.
-
-        let total_in = (t_in + sapling_in).ok_or(BalanceError::Overflow)?;
-
-        if (!transparent_inputs.is_empty() || !sapling_inputs.is_empty()) && fee_amount > total_in {
-            // For the fixed-fee selection rule, the only time we consider inputs dust is when the fee
```
nuttycom, line 91: This case has been eliminated in this refactoring, because it doesn't matter: under a circumstance where the fee amount is greater than the total of the inputs, the proposed change calculation would return a negative value and thus return an insufficient funds error.

zcash_client_backend/src/fees/common.rs
```diff
@@ -0,0 +1,122 @@
...
+            params,
+            target_height,
+            transparent_inputs,
+            transparent_outputs,
+            sapling_inputs.len(),
+            if sapling_inputs.is_empty() {
+                sapling_outputs.len() + 1
+            } else {
+                std::cmp::max(sapling_outputs.len() + 1, 2)
+            },
```
str4d, line 68: I presume this will change to use `sapling::builder::BundleType::num_outputs(sapling_inputs.len(), sapling_outputs.len() + 1)?`

str4d (approved):
utACK d74f635d9dd3165f9c0e393d2643cf24602bbc62 with suggestion.

zcash_client_backend/src/fees/common.rs
nuttycom: That's correct.

zcash_client_backend/src/fees.rs
```diff
@@ -12,6 +12,7 @@ use zcash_primitives::{
     },
 };
 
+pub mod common;
```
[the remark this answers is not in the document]
nuttycom: I'll fix this in a follow-on PR so that I can build atop this more quickly.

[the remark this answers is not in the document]
nuttycom: Fixed in #1065

PR #1067, opened 2023-12-11
Extract `sapling-crypto` crate again

### LINKED ISSUE #738: Extract Sapling protocol into `sapling-crypto` crate

str4d:
Closes #738.


sapling-crypto/src/bundle.rs
```diff
@@ -331,19 +319,8 @@ impl<A: Authorization> std::fmt::Debug for SpendDescription<A> {
...
-        rk: redjubjub::VerificationKey<SpendAuth>,
-        zkproof: A::SpendProof,
-        spend_auth_sig: A::AuthSig,
-    ) -> Self {
-        Self::from_parts(cv, anchor, nullifier, rk, zkproof, spend_auth_sig)
-    }
-
-    pub(crate) fn from_parts(
+    /// Constructs a v4 `SpendDescription` from its constituent parts.
+    pub fn from_parts(
```
nuttycom: Now that these functions are public, rather than just pub(crate), are there invariants that should be enforced in them? It's fine to do this as a follow-up in the `sapling-crypto` crate, but if such work should be done then please open an issue for it.

sapling-crypto/src/bundle.rs
```diff
@@ -410,7 +387,8 @@ pub struct SpendDescriptionV5 {
 }
 
 impl SpendDescriptionV5 {
-    pub(crate) fn from_parts(
+    /// Constructs a v5 `SpendDescription` from its constituent parts.
+    pub fn from_parts(
```
nuttycom: Same question about invariants.

sapling-crypto/src/bundle.rs
```diff
@@ -475,26 +453,8 @@ impl<Proof> OutputDescription<Proof> {
...
-            ephemeral_key,
-            enc_ciphertext,
-            out_ciphertext,
-            zkproof,
-        )
-    }
-
-    pub(crate) fn from_parts(
+    /// Constructs a v4 `OutputDescription` from its constituent parts.
+    pub fn from_parts(
```
nuttycom: Ditto.

sapling-crypto/src/bundle.rs
```diff
@@ -578,7 +538,8 @@ pub struct OutputDescriptionV5 {
 memuse::impl_no_dynamic_usage!(OutputDescriptionV5);
 
 impl OutputDescriptionV5 {
-    pub(crate) fn from_parts(
+    /// Constructs a v5 `OutputDescription` from its constituent parts.
```
nuttycom: Ditto.

nuttycom (approved):
utACK. I have verified that [zcash/sapling-crypto@df6681c1046bd28073befd7d42224e4a4c70032a] is equivalent to the `sapling-crypto` subdirectory as of 6acc64e

sapling-crypto/src/bundle.rs
str4d: I added the necessary invariant to `Bundle::from_parts`, but I couldn't at the time think of any other invariants that were necessary for `SpendDescription` and `OutputDescription`. I'll think it over before we cut a release of `sapling-crypto`.

str4d:
`wasm32-wasi` build is temporarily broken because of the patch directive; this will be fixed once `sapling-crypto 0.1.0` is published and we switch to it.

PR #1071, opened 2023-12-19
Remove notices about the BOSL license

### LINKED ISSUE #1072: Remove notices about the BOSL license after updating orchard dependency
fixed by #1071 which depends on #1108.

daira:
fixes #1072

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,17 +16,6 @@ All code in this workspace is licensed under either of
 
 at your option.
 
-Downstream code forks should note that some (but not all) of these crates
-and components depend on the 'orchard' crate, which is licensed under the
-[Bootstrap Open Source License](https://github.com/zcash/orchard/blob/main/LICENSE-BOSL).
-A license exception is provided allowing some derived works that are linked or
-combined with the 'orchard' crate to be copied or distributed under the original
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -49,7 +49,7 @@ jubjub = "0.10"
 
 # - Orchard
 nonempty = "0.7"
-orchard = { version = "0.6", default-features = false }
+orchard = { version = "0.6.1", default-features = false }
```
str4d, line 52: This PR will be blocked on @nuttycom's other PRs.
```suggestion
orchard = { version = "0.7", default-features = false }
```

str4d (approved):
Changes LGTM. However, I double-checked, and in `main` we currently patch in https://github.com/zcash/orchard/commit/189257391a5726d7f3f2568d547a5ef8603c7156 which is the tip of a branch that does not include the license change. Its child merge commit zcash/orchard@ba70c32c280aa2dc23e4caa82252adc2f28cf17a does include the license change. If we wait until #1108 merges then this PR's changes will be correct.

str4d:
#1108 has merged, so this can now be rebased.

PR #1077, opened 2023-12-29
Enable unstable features when editing in VS Code

AArnott:
This improves the editing experience by 'enabling' all the code that otherwise is turned off because of the `unstable` flag.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,3 @@
 {
-    "rust-analyzer.cargo.features": ["unstable"]
+    "rust-analyzer.cargo.features": "all"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "rust-analyzer.cargo.features": ["unstable"]
+}


nuttycom (commented):
In general we don't commit editor-specific config to this respository. I'm not strongly opposed to it in this case, but maybe @str4d or another vscode user can chime in here.

AArnott:
It took me a while to figure out how to get a good editing experience by adding this file. As it doesn't hurt any _other_ IDE to add it, IMO we should add it so that any VS Code user doesn't have to independently discover how to do it, and maintain a file like this without checking it in by accident.
Either way, it's not blocking me. It's in my own fork of the repo where I do most of my work.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "rust-analyzer.cargo.features": ["unstable"]
```
str4d, line 2: I develop locally with:
```suggestion
    "rust-analyzer.cargo.features": "all"
```

str4d (changes requested):
I'm fine with including this in the repo to help VS Code users, but we shouldn't special-case the `unstable` flag; there are other flags that are also relevant.

str4d (approved):
ACK (haven't tested this PR but I have the identical setting locally)

PR #1078, opened 2023-12-30
Implement Display trait for `DecodingError`

AArnott:


zcash_primitives/src/transaction/components/amount.rs
```diff
@@ -22,6 +22,24 @@ pub const MAX_MONEY: i64 = 21_000_000 * COIN;
...
+            (false, self.0)
+        };
+
+        let zec = value / COIN;
+        let zat = value % COIN;
+
+        if negative {
+            write!(f, "-")?;
+        }
+        write!(f, "{}.{:08}", zec, zat)
```
nuttycom, line 39: While `zat` is correctly named here given that it's padding the value to 8 decimal places, it's very common for ZEC display values to be truncated and not show trailing zeros. Whether or not one wants to do so is contextual, so I don't think there should be a `Display` impl at all, and instead we should just have rendering functions for the different human-facing use cases.

In general I don't think that `Display` is a good trait, despite how common it is in the ecosystem, for exactly this reason: the notion of a "canonical human-facing rendering" is ill-formed. For example, in a lot of localizations, a comma is used instead of a decimal point; you can't have a `Display` impl that localizes correctly.

nuttycom (changes requested):
While the `Display` impl for `DecodingError` makes sense to me (because of the way that `Display` is required in the standard `Error` system) I don't think we should have a canonical `Display` implementation for `Amount` as I explain in the comments here.

zcash_primitives/src/transaction/components/amount.rs
AArnott: Good insights. I'm still new to rust, but my expectation was that the Display trait was not guaranteed to be good for humans. In my case, this was useful for error messages because some error types include an `Amount` field and making amounts able to print themselves, it makes it easier for the error type to print itself. In fact, IIRC I _think_ this became transitively necessary in order to make an error enum display-able, which in turn was necessary at the UniFFI layer to be able to reduce zcash-related error enums into just an error message that could be sent across the FFI layer.

Ultimately, error codes rather than error messages are probably better in most cases for the FFI layer, so the app can render its own messages (localized, referencing UI elements that the user can relate to, etc.). So I expect I can ultimately live without this.

AArnott:
I've removed the change for `Amount`.

zcash_primitives/src/transaction/components/amount.rs
str4d: The [`std::fmt` module states](https://doc.rust-lang.org/stable/std/fmt/index.html#fmtdisplay-vs-fmtdebug):
> - [`fmt::Display`](https://doc.rust-lang.org/stable/std/fmt/trait.Display.html) implementations assert that the type can be faithfully represented as a UTF-8 string at all times. It is **not** expected that all types implement the [`Display`](https://doc.rust-lang.org/stable/std/fmt/trait.Display.html) trait.
> - [`fmt::Debug`](https://doc.rust-lang.org/stable/std/fmt/trait.Debug.html) implementations should be implemented for all public types. Output will typically represent the internal state as faithfully as possible. The purpose of the [`Debug`](https://doc.rust-lang.org/stable/std/fmt/trait.Debug.html) trait is to facilitate debugging Rust code. In most cases, using `#[derive(Debug)]` is sufficient and recommended.

We can't reliably represent `Amount` as a UTF-8 string at all times, because at a minimum its units depend on what network the client is connected to. So we will likely need a dedicated method for this. As Kris noted above, its localization is also runtime-dependent, but this _can_ be handled at runtime (for example, [here is how I localize error messages in the `age` crate](https://github.com/str4d/rage/blob/4bac30b7c3587f4524cb506294fb54debf2751c3/age/src/error.rs#L57-L106)).

nuttycom (approved):
utACK

PR #1060, opened 2023-12-06
Add Orchard support to fees & transaction proposals.

nuttycom:
Now builds atop #1090

Part of #403.
Closes #1094.
Closes #1097.

PR #1080, opened 2024-01-03
zcash_client_backend: Introduce an "orchard" feature flag.

nuttycom:
We plan to also introduce a similar flag to gate access to Sapling functionality. Since introduction of Orchard functionality is still nascent, it's the correct time to introduce this isolation, before there's more functionality that needs to be isolated in this fashion.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -9,13 +9,14 @@ and this library adheres to Rust's notion of
 
 ### Added
 - `zcash_client_backend::data_api`:
-  - `BlockMetadata::orchard_tree_size`.
+  - `BlockMetadata::orchard_tree_size` (when the `orchard` feature is enabled).
   - `TransparentInputSource`
   - `SaplingInputSource`
-  - `ScannedBlock::{
-      sapling_tree_size, orchard_tree_size, orchard_nullifier_map,
-      orchard_commitments, into_commitments
-    }`
+  - `ScannedBlock::{into_commitments, sapling}`
+  - `ScannedBlock::orchard` (when the `orchard` feature is enabled.)
+  - `ScannedBlockSapling`
+  - `ScannedBlockOrchard` (when the `orchard` feature is enabled.)
+  - `ScannedBlockCommitments`
   - `Balance::{add_spendable_value, add_pending_change_value, add_pending_spendable_value}`
   - `AccountBalance::{
       with_sapling_balance_mut, 
@@ -49,12 +50,17 @@ and this library adheres to Rust's notion of
      wallet::{ReceivedSaplingNote, WalletTransparentOutput},
      wallet::input_selection::{Proposal, SaplingInputs},
    }`
+- A new `orchard` feature flag has been added to make it possible to
+  build client code without `orchard` dependendencies.
 
 ### Moved
 - `zcash_client_backend::data_api::{PoolType, ShieldedProtocol}` have
   been moved into the `zcash_client_backend` root module.
 - `zcash_client_backend::data_api::{NoteId, Recipient}` have
   been moved into the `zcash_client_backend::wallet` module.
+- `ScannedBlock::{sapling_tree_size, sapling_nullifier_map, sapling_commitments}`
+  have been moved to `ScannedBlockSapling` and in that context are now 
+  named `{tree_size, nullifier_map, commitments}` respectively.
 
 ### Changed
 - `zcash_client_backend::data_api`:
@@ -68,7 +74,6 @@ and this library adheres to Rust's notion of
     `WalletShieldedOutput` change.
   - `ScannedBlock` has an additional type parameter as a consequence of the 
     `WalletTx` change.
-  - Arguments to `ScannedBlock::from_parts` have changed.
   - `ScannedBlock::metadata` has been renamed to `to_block_metadata` and now
     returns an owned value rather than a reference.
   - `ShieldedProtocol` has a new variant for `Orchard`, allowing for better
@@ -184,6 +189,18 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::address`:
   - `RecipientAddress` has been renamed to `Address`
   - `Address::Shielded` has been renamed to `Address::Sapling`
+  - `UnifiedAddress::from_receivers` no longer takes an Orchard receiver 
+    argument uless the `orchard` feature is enabled.
+  - `UnifiedAddress::orchard` is now only available when the `orchard` feature
+    is enabled.
+
+- `zcash_client_backend::keys`:
+  - `DerivationError::Orchard` is now only available when the `orchard` feature
+    is enabled.
+  - `UnifiedSpendingKey::orchard` is now only available when the `orchard`
+    feature is enabled.
+  - `UnifiedFullViewingKey::new` no longer takes an Orchard full viewing key
+    argument uless the `orchard` feature is enabled.
 
 ### Removed
 - `zcash_client_backend::wallet::ReceivedSaplingNote` has been replaced by
@@ -195,6 +212,7 @@ and this library adheres to Rust's notion of
   removed without replacement as it was unused, and its functionality will be
   fully reproduced by `SaplingInputSource::select_spendable_sapling_notes` in a future
   change.
+- `zcash_client_backend::data_api::ScannedBlock::from_parts` has been made crate-private.
 - `zcash_client_backend::data_api::ScannedBlock::into_sapling_commitments` has been
   replaced by `into_commitments` which returns both Sapling and Orchard note commitments
   and associated note commitment retention information for the block.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -14,8 +14,7 @@ and this library adheres to Rust's notion of
   - `SaplingInputSource`
   - `ScannedBlock::{into_commitments, sapling}`
   - `ScannedBlock::orchard` (when the `orchard` feature is enabled.)
-  - `ScannedBlockSapling`
-  - `ScannedBlockOrchard` (when the `orchard` feature is enabled.)
+  - `ScannedBundles`
   - `ScannedBlockCommitments`
   - `Balance::{add_spendable_value, add_pending_change_value, add_pending_spendable_value}`
   - `AccountBalance::{
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -189,7 +189,7 @@ and this library adheres to Rust's notion of
   - `RecipientAddress` has been renamed to `Address`
   - `Address::Shielded` has been renamed to `Address::Sapling`
   - `UnifiedAddress::from_receivers` no longer takes an Orchard receiver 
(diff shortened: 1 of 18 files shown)


zcash_client_backend/CHANGELOG.md
```diff
@@ -184,6 +189,18 @@ and this library adheres to Rust's notion of
...
+  - `UnifiedAddress::orchard` is now only available when the `orchard` feature
+    is enabled.
+
+- `zcash_client_backend::keys`:
+  - `DerivationError::Orchard` is now only available when the `orchard` feature
+    is enabled.
+  - `UnifiedSpendingKey::orchard` is now only available when the `orchard`
+    feature is enabled.
+  - `UnifiedFullViewingKey::new` no longer takes an Orchard full viewing key
+    argument uless the `orchard` feature is enabled.
```
str4d, line 202:
```suggestion
    argument unless the `orchard` feature is enabled.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -593,11 +595,115 @@ impl BlockMetadata {
...
+    pub fn final_tree_size(&self) -> u32 {
+        self.final_tree_size
+    }
+
+    /// Returns the vector of Sapling nullifiers for each transaction in the block.
+    ///
+    /// The returned tuple is keyed by both transaction ID and the index of the transaction within
+    /// the block, so that either the txid or the combination of the block hash available from
+    /// [`Self::block_hash`] and returned transaction index may be used to uniquely identify the
+    /// transaction, depending upon the needs of the caller.
```
str4d, line 638:
```suggestion
    /// [`ScannedBlock::block_hash`] and returned transaction index may be used to uniquely
    /// identify the transaction, depending upon the needs of the caller.
```

zcash_client_backend/src/keys.rs
```diff
@@ -678,19 +716,28 @@ mod tests {
...
         let _encoded_no_t = "uview12z384wdq76ceewlsu0esk7d97qnd23v2qnvhujxtcf2lsq8g4hwzpx44fwxssnm5tg8skyh4tnc8gydwxefnnm0hd0a6c6etmj0pp9jqkdsllkr70u8gpf7ndsfqcjlqn6dec3faumzqlqcmtjf8vp92h7kj38ph2786zx30hq2wru8ae3excdwc8w0z3t9fuw7mt7xy5sn6s4e45kwm0cjp70wytnensgdnev286t3vew3yuwt2hcz865y037k30e428dvgne37xvyeal2vu8yjnznphf9t2rw3gdp0hk5zwq00ws8f3l3j5n3qkqgsyzrwx4qzmgq0xwwk4vz2r6vtsykgz089jncvycmem3535zjwvvtvjw8v98y0d5ydwte575gjm7a7k";
-        #[cfg(feature = "transparent-inputs")]
-        assert_eq!(encoded, encoded_with_t);
-        #[cfg(not(feature = "transparent-inputs"))]
-        assert_eq!(encoded, _encoded_no_t);
+
+        // We test the full roundtrip only with the `orchard` feature enabled, because we will
+        // not generate the `orchard` part of the encoding if the UFVK does not have an Orchard
+        // part.
```
str4d, line 733: Non-blocking: we should instead update these test vectors to cover the matrix of flags.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -997,12 +998,13 @@ fn parse_block_metadata<P: consensus::Parameters>(
         block_height,
         block_hash,
         Some(sapling_tree_size),
-        if params
+        #[cfg(feature = "orchard")]
+        if _params
```
str4d, line 1002: Per my comment above, if we're still going to have the database tables for Orchard in the schema, then for values that don't depend on `orchard` crate types and aren't wallet-specific (like the Orchard tree size for a given block), we probably don't want to feature flag this off.

The main reason for doing so would be if the presence of an Orchard tree size is used as some kind of indicator or switch for later logic. Do we need to use it as such?

zcash_primitives/src/transaction/builder.rs
```diff
@@ -201,36 +201,40 @@ impl Progress {
 #[derive(Clone, Copy)]
 pub enum BuildConfig {
     Standard {
-        sapling_anchor: sapling::Anchor,
-        orchard_anchor: orchard::Anchor,
+        sapling_anchor: Option<sapling::Anchor>,
+        orchard_anchor: Option<orchard::Anchor>,
```
str4d, line 205: (Asking so I can keep track of the design:) Why are the anchors now optional again? (And why the change in this PR - what about the `orchard` flag necessitates this?)

zcash_primitives/src/transaction/builder.rs
```diff
@@ -522,20 +528,22 @@ impl<'a, P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<
                 transparent_inputs,
                 self.transparent_builder.outputs(),
                 sapling_spends,
```
str4d, line 530: Non-blocking: I think we need to alter this to call `bundle_type.num_spends()` after the changes in https://github.com/zcash/sapling-crypto/pull/114. This should happen in a separate PR.

str4d (changes requested):
Reviewed as of 857e44a11c1db5d2139f39092e212a366fd3f0f9.

zcash_client_backend/src/address.rs
```diff
@@ -103,12 +111,18 @@ impl UnifiedAddress {
     /// Returns `None` if the receivers would produce an invalid Unified Address (namely,
     /// if no shielded receiver is provided).
     pub fn from_receivers(
-        orchard: Option<orchard::Address>,
+        #[cfg(feature = "orchard")] orchard: Option<orchard::Address>,
```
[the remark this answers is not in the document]
nuttycom: If we don't have it public, then the only way that someone can construct a `UnifiedAddress` is from the string encoding, which I think is too restrictive. For example, the stuff that @AArnott is doing would require plumbing through strings, and I think that we should be supportive of his use cases.

zcash_client_backend/src/keys.rs
```diff
@@ -324,13 +339,19 @@ impl UnifiedSpendingKey {
                 }
             }
 
+            #[cfg(feature = "orchard")]
+            let has_orchard = orchard.is_some();
+            #[cfg(not(feature = "orchard"))]
+            let has_orchard = true;
```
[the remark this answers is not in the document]
nuttycom: After having done this exercise, I think that we should go ahead with adding the `sapling` feature flag to `zcash_client_backend` in the near future; it's not that much work now that the pattern is established and it really does help tease out coupling and improve modularity.

zcash_client_sqlite/Cargo.toml
```diff
@@ -80,6 +80,10 @@ multicore = ["maybe-rayon/threads", "zcash_primitives/multicore"]
 ## client is configured for use with the Zcash testnet.
 mainnet = []
 
+## Enables support for storing data related to the sending and receiving of 
+## Orchard funds.
+orchard = ["zcash_client_backend/orchard"]
```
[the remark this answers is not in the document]
nuttycom: Yes, ultimately, but I want it to be default-disabled until there are no more `FIXME` or `unimplemented()` sections.

zcash_client_sqlite/src/testing.rs
```diff
@@ -841,6 +841,8 @@ pub(crate) fn fake_compact_block_from_tx(
             ctx.outputs.push(output.into());
         }
     }
+
+    #[cfg(feature = "orchard")]
```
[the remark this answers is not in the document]
nuttycom: Right now I think they'll always contain both.

zcash_client_backend/src/data_api.rs
```diff
@@ -593,11 +595,115 @@ impl BlockMetadata {
 
     /// Returns the size of the Orchard note commitment tree for the final treestate of the block
     /// that this [`BlockMetadata`] describes, if available.
+    #[cfg(feature = "orchard")]
     pub fn orchard_tree_size(&self) -> Option<u32> {
```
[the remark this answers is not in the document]
nuttycom: Hmmm... what are you going to do with the Orchard tree size if Orchard is generally flagged off?

zcash_client_backend/src/keys.rs
nuttycom: Created https://github.com/zcash/librustzcash/issues/1084

zcash_client_sqlite/src/wallet.rs
nuttycom: I don't want to use the presence or absence of data to be used as a logic switch, for sure. However, as I stated above, I think that we should try to fully segregate protocol-specific functionality so that as much `orchard` stuff as possible is removed from the API when the `orchard` flag is disabled; a smaller API surface is going to be less error-prone.

zcash_primitives/src/transaction/builder.rs
nuttycom: So, the anchors are optional so that we don't have to depend on the `orchard` crate in `zcash_client_backend`, since the feature flagging doesn't yet extend to `zcash_primitives`. We will likely want to do that work in `zcash_primitives`, but that's a bigger job that I didn't want to tackle yet. So making these optional is the compromise.

zcash_primitives/src/transaction/builder.rs
nuttycom: That is happening as part of #1060

zcash_client_backend/src/keys.rs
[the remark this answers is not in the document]
AArnott: Given transparent support is also under a feature flag, what will happen when _no_ feature flag is specified? Do we expect that to build, or would the error be helpful in directing the consumer to specify at least one?

str4d (approved):
utACK 967fd36d390ecfc1659439b44d192ed1ce769681

zcash_client_backend/src/data_api.rs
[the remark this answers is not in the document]
str4d: Conclusion from chatting after standup: we may re-expose this if this kind of block-level metadata is useful to track even when Orchard support is disabled, but we'll figure that out once we start working out the modifications to the scan queue.

zcash_client_backend/src/keys.rs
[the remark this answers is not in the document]
daira: We can just have that cause a compilation error.

zcash_client_backend/src/address.rs
```diff
@@ -273,8 +293,12 @@ mod tests {
 
         let transparent = { None };
 
+        #[cfg(feature = "orchard")]
         let ua = UnifiedAddress::from_receivers(orchard, sapling, transparent).unwrap();
 
+        #[cfg(not(feature = "orchard"))]
+        let ua = UnifiedAddress::from_receivers(sapling, transparent).unwrap();
```
daira, line 300: This could just be
```rust
let ua = UnifiedAddress::from_receivers(
    #[cfg(feature = "orchard")] orchard,
    sapling,
    transparent,
).unwrap();
```
(which will be simpler once we add the "sapling" feature).

daira (commented):
Post-hoc ACK

PR #1083, opened 2024-01-04
equihash: Add Rust API for Tromp solver

str4d:


components/equihash/tromp/equi_miner.c
```diff
@@ -0,0 +1,699 @@
...
+#error not implemented
+#endif
+        const u32 slot = getslot(eq, 0, bucketid);
+        if (slot >= NSLOTS) {
+          eq->bfull++;
+          continue;
+        }
+        slot0 *s = &eq->hta.trees0[0][bucketid][slot];
+        s->attr = tree_from_idx(block * HASHESPERBLAKE + i);
+        memcpy(s->hash->bytes+htl.nextbo, ph+WN/8-hashbytes, hashbytes);
```
str4d, line 512: Currently the solver segfaults unless I comment out the second and third of these lines (and similarly comment out any accesses to `htlayout` data). I suspect that means my porting of the code from C++ to C is buggy.

components/equihash/src/lib.rs
```diff
@@ -26,3 +26,6 @@ mod verify;
 mod test_vectors;
 
 pub use verify::{is_valid_solution, Error};
+
+mod blake2b;
+pub mod tromp;
```
str4d, line 31: TODO: Gate this off behind a feature flag, so anyone who doesn't need it won't need to depend on `cc` or include any of the native code.

teor2345: I implemented this feature (and got the code producing valid solutions) in PR #1088.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/8e66e0d952240ea304251fe78873a19a97f30770..d33ec1d5c0f0d34150691ca25ea90df47fcc42da) to clean up the commit history; no changes to @teor2345's changes, just regrouping them (and squashing some of the fixes into the original commits that didn't make sense to keep in standalone commits).

components/equihash/src/tromp.rs
```diff
@@ -175,10 +175,16 @@ pub fn solve_200_9_compressed<const N: usize>(
...
     let solutions = solve_200_9(input, next_nonce);
 
-    solutions
+    let mut solutions: Vec<Vec<u8>> = solutions
         .iter()
         .map(|solution| get_minimal_from_indices(solution, DIGIT_BITS))
-        .collect()
+        .collect();
+
+    // Just in case the solver returns solutions that become the same when compressed.
```
str4d, line 201: I don't think this should ever be the case. The compressed representation should not be lossy, it just packs the indices together to remove the effective padding bits that exist in the `u32` representation.

components/equihash/src/tromp.rs
```diff
@@ -70,13 +71,21 @@ unsafe fn worker(eq: *mut CEqui, p: verify::Params, curr_state: &State) -> Vec<V
         let nsols = equi_nsols(eq);
         let sols = equi_sols(eq);
         let solution_len = 1 << p.k;
+        //println!("{nsols} solutions of length {solution_len} at {sols:?}");
```
str4d, line 74: I'm inclined to remove this and the other introduced debugging print statements, unless anyone else disagrees.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/d33ec1d5c0f0d34150691ca25ea90df47fcc42da..9dbdf1880e6a47f071d821aac48fd9d9cb1d78a8) to GitHub-linkify an imported source link in a commit message.

str4d:
And somehow I managed to click two wrong buttons in a row 🤦

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/9dbdf1880e6a47f071d821aac48fd9d9cb1d78a8..b201c1012e2332543f66636f830974ec38815904) to clean up the feature flagging in the build script, and add C `#ifdef`s to prevent some warnings.

str4d:
[Rebased on current `main`](https://github.com/zcash/librustzcash/compare/b201c1012e2332543f66636f830974ec38815904..be16706e326a63165660e38a32b7b1b9eebdf06d). I will rebase this PR again once #1357 merges.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/ed22417ade401b2e98ffcb3a165a666336fc7fd1..6571aa980b0dff722d84317361844c30eea036be) to move the new minimal-form compression methods into the new submodule from #1357.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/6571aa980b0dff722d84317361844c30eea036be..76131db25a2772c5e139d066ec68975f26d988b5) to adjust the minimal-form compression method APIs and integrate them into the existing tests.

components/equihash/src/tromp.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    solutions
+}
+
+/// Performs multiple equihash solver runs with equihash parameters `200, 9`, initialising the hash with
+/// the supplied partial `input`. Between each run, generates a new nonce of length `N` using the
+/// `next_nonce` function.
+///
+/// Returns zero or more unique solutions.
+pub fn solve_200_9<const N: usize>(
```
str4d, line 129: I don't think there is value in exposing the API for uncompressed solutions; block headers only use compressed solutions, and this crate's other existing API is only for verifying compressed solutions. So I'll probably just make this internal.

What I do want to think about before merging is how to handle configurable parameters. We will need at least two parameter sets (for mainnet and regtest), and depending on how that will be enabled in the C code, we might end up supporting any parameter set. But if we don't then we should probably instead have an enum or a generic parameter for selecting which Equihash parameters to solve with.

arya2: > I don't think there is value in exposing the API for uncompressed solutions; block headers only use compressed solutions, and this crate's other existing API is only for verifying compressed solutions. So I'll probably just make this internal.

This is currently true for [Zebra's use case](https://github.com/ZcashFoundation/zebra/blob/v1.5.1/zebra-chain/src/work/equihash.rs#L115-L166), though I see no harm in exposing the API for uncompressed solutions either so solutions could be compressed outside this crate more easily.

> What I do want to think about before merging is how to handle configurable parameters. We will need at least two parameter sets (for mainnet and regtest), and depending on how that will be enabled in the C code, we might end up supporting any parameter set. But if we don't then we should probably instead have an enum or a generic parameter for selecting which Equihash parameters to solve with.

It may be useful to allow for configurable parameters for custom testnets to allow for picking even easier parameters instead of disabling PoW checks altogether for ephemeral testnets, or harder ones than those used by regtest for testnets that are intended to be used for longer periods.

components/equihash/src/tromp.rs
arya2: This was added in the same [commit](https://github.com/zcash/librustzcash/pull/1088/commits/d2c4765d16f20d86f87431920baa6f1d2ef516c4) as the `.dedup()` call on the list of uncompressed solutions, if you're confident that the C code only removes effective padding, it may not have been observed, or there could be a subtle difference between the C code and the Rust code in `compress_array()`.

components/equihash/src/tromp.rs
```diff
@@ -0,0 +1,256 @@
...
+use std::slice;
+
+use blake2b_simd::State;
+
+use crate::{blake2b, minimal::minimal_from_indices, params::Params, verify};
+
+#[repr(C)]
+struct CEqui {
+    _f: [u8; 0],
+    _m: PhantomData<(*mut u8, PhantomPinned)>,
```
arya2, line 13: Could we add a short explanation here?

components/equihash/tromp/equi_miner.c
```diff
@@ -0,0 +1,737 @@
...
+    BLAKE2bState* state;
+    htlayout htl = htlayout_new(eq, 0);
+    const u32 hashbytes = hashsize(0);
+    for (u32 block = id; block < NBLOCKS; block++) {
+      state = eq->blake2b_clone(eq->blake_ctx);
+      u32 leb = htole32(block);
+      eq->blake2b_update(state, (uchar *)&leb, sizeof(u32));
+      eq->blake2b_finalize(state, hash, HASHOUT);
+      eq->blake2b_free(state);
+      // Avoid use-after-free and double-free
```
arya2, line 534: It may be a good idea to add this warning to the `blake2b_free()` function docs too.

arya2 (commented):
This looks like it should work so far, and none of my suggestions are blockers, feel free to resolve any/all of them without making changes.

I'm still reviewing `equi_miner.c` in detail, but everything else looks equivalent to the logic in zcashd, at least for the Mainnet parameters.

components/equihash/Cargo.toml
```diff
@@ -9,9 +9,21 @@ license = "MIT OR Apache-2.0"
 edition = "2021"
 rust-version = "1.56.1"
 
+[features]
+default = []
+
+## Builds the C++ tromp solver and Rust FFI layer.
```
arya2, line 15:
```suggestion
# Experimental tromp solver support, builds the C++ tromp solver and Rust FFI layer.
```

arya2 (commented):
@str4d Testing indicates that there are still some issues in `equi_miner.c`, but I think we should document this as an experimental feature and otherwise merge it as-is, because it usually works, the occasional unexpected errors are handled gracefully, and the `internal-miner` feature in Zebra that uses it is already documented as experimental.

We could try porting it over to Rust or fixing the occasional issues in the current C code once the resource-constrained steps needed to deploy NU7 are complete. Running this code via Zebra in CI could produce some helpful debug output if we keep the debug print statements and perhaps add more of them.

If we don't want to document it as an experimental feature, a pair review of `equi_miner.c` could be helpful for figuring out what's causing the `Collision` and `DuplicateIdxs`errors.

components/equihash/src/tromp.rs
arya2: I think we should keep the debug print statements and consider adding more.

arya2 (approved):

components/equihash/src/minimal.rs
```diff
@@ -5,6 +5,49 @@ use byteorder::{BigEndian, ReadBytesExt};
...
+        // When we have fewer than 8 bits left in the accumulator, read the next
+        // input element.
+        if acc_bits < 8 {
+            acc_value <<= bit_len;
+            for x in byte_pad..in_width {
+                acc_value |= (
+                    // Apply bit_len_mask across byte boundaries
+                    (array[j + x] & ((bit_len_mask >> (8 * (in_width - x - 1))) as u8)) as u32
+                )
+                    .wrapping_shl(8 * (in_width - x - 1) as u32); // Big-endian
```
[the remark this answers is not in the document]
arya2: Is it possible this is a problem if it's different than what the original code was doing?

mpguerra:
Do we want to merge this one? It's blocking us from restoring the `internal-miner` feature in zebra

components/equihash/src/tromp.rs
str4d: For now I'm going to remove the uncompressed solution API as it does not exist anywhere else, and not try to solve the configurable parameters question in this PR.

str4d (commented):
utACK f3d5a5cfdfc23af73ff0cb221b6005545a395d36. I've made the changes I wanted for the release; the remaining issues are non-blocking and can be fixed over time.

arya2 (approved):
Thank you!!

PR #1088, opened 2024-01-05
equihash: Fix some Rust API issues

teor2345:
This is a fully-functional draft, but it contains some extra lint handling and review comments.


components/equihash/src/tromp.rs
```diff
@@ -30,23 +32,30 @@ extern "C" {
     fn equi_digiteven(eq: *mut CEqui, r: u32, id: u32);
     fn equi_digitK(eq: *mut CEqui, id: u32);
     fn equi_nsols(eq: *const CEqui) -> usize;
-    fn equi_sols(eq: *const CEqui) -> *const *const u32;
+    /// Returns `equi_nsols()` solutions of length `2^K`, in a single memory allocation.
+    fn equi_sols(eq: *const CEqui) -> *const u32;
```
teor2345, line 36: Despite its type in the C code, this is actually a single allocation containing multiple solutions. This fix avoids a segmentation fault.

https://github.com/zcash/librustzcash/blob/f3ea80e2abb1eef5cc38a554e7a60a90958262a4/components/equihash/tromp/equi_miner.c#L256

https://github.com/zcash/librustzcash/blob/f3ea80e2abb1eef5cc38a554e7a60a90958262a4/components/equihash/tromp/equi_miner.c#L208-L213

components/equihash/src/tromp.rs
```diff
@@ -30,23 +32,30 @@ extern "C" {
...
 
-unsafe fn worker(p: verify::Params, curr_state: &State) -> Vec<Vec<u32>> {
-    // Create solver and initialize it.
-    let eq = equi_new(
-        1,
-        blake2b::blake2b_clone,
-        blake2b::blake2b_free,
-        blake2b::blake2b_update,
-        blake2b::blake2b_finalize,
-    );
```
teor2345, line 44: Moving the allocation out of the loop gives me a 5-10% performance boost, but it's optional.

components/equihash/src/tromp.rs
```diff
@@ -55,25 +64,64 @@ unsafe fn worker(p: verify::Params, curr_state: &State) -> Vec<Vec<u32>> {
...
+        let solutions = (&mut chunks)
+            .map(|solution| solution.to_vec())
+            .collect::<Vec<_>>();
+
+        assert_eq!(chunks.remainder().len(), 0);
 
         solutions
-            .iter()
-            .map(|solution| slice::from_raw_parts(*solution, solution_len).to_vec())
-            .collect::<Vec<_>>()
```
teor2345, line 69: This is the other part of the fix that avoids a segmentation fault.

components/equihash/src/tromp.rs
```diff
@@ -82,41 +130,163 @@ pub fn solve_200_9<const N: usize>(
...
+) -> Vec<Vec<u8>> {
+    // https://github.com/zcash/zcash/blob/6fdd9f1b81d3b228326c9826fa10696fc516444b/src/pow/tromp/equi.h#L34
+    const DIGIT_BITS: usize = 200 / (9 + 1);
+    let solutions = solve_200_9(input, next_nonce);
+
+    solutions
+        .iter()
+        .map(|solution| get_minimal_from_indices(solution, DIGIT_BITS))
+        .collect()
+}
```
teor2345, line 197: This provides the format used by the block header and verifier.

components/equihash/src/tromp.rs
```diff
@@ -125,6 +295,17 @@ mod tests {
...
+                    |error| {
+                        panic!(
+                            "unexpected invalid equihash 200, 9 solution:\n\
+                             error: {error:?}\n\
+                             input: {input:?}\n\
+                             nonce: {nonce:?}\n\
+                             solution: {solution:?}"
+                        )
+                    },
+                );
```
teor2345, line 308: I haven't got this far yet in the tests, but if the code I've added works, we will be able to verify that the generated solutions are valid.

teor2345 (commented):
Here are the major changes I've made to the code. The solution allocation type fix is required, everything else is optional.

(There are also a bunch of lints and comments that can be ignored, I'll clean them up when we get a valid solution from the code.)

components/equihash/Cargo.toml
```diff
@@ -9,12 +9,21 @@ license = "MIT OR Apache-2.0"
 edition = "2021"
 rust-version = "1.56.1"
 
+[features]
+default = []
+
+## Builds the C++ tromp solver and Rust FFI layer.
+solver = ["dep:cc"]
```
teor2345, line 16: Whatever you want to name the feature is fine with me.

components/equihash/tromp/equi_miner.c
```diff
@@ -281,7 +290,7 @@ typedef struct equi equi;
   u32 getnslots(equi *eq, const u32 r, const u32 bid) { // SHOULD BE METHOD IN BUCKET STRUCT
     au32 *nslot = &eq->nslots[r&1][bid];
     const u32 n = min(*nslot, NSLOTS);
-    nslot = 0;
+    *nslot = 0;
```
teor2345, line 293: This was the bug preventing valid solutions, in the original code, `nslot` was a reference.

teor2345:
It looks like Windows and macOS need a portable `endian.h`, so I added a trivial permissively licensed header.

Also Windows didn't like the thread support, which we're not using anyway, so I deleted it.

components/equihash/tromp/equi_miner.c
```diff
@@ -228,42 +245,58 @@ struct equi {
...
     free(eq);
   }
   void equi_setstate(equi *eq, const BLAKE2bState *ctx) {
+    if (eq->blake_ctx) {
+      eq->blake2b_free(eq->blake_ctx);
+    }
+
     eq->blake_ctx = eq->blake2b_clone(ctx);
     memset(eq->nslots, 0, NBUCKETS * sizeof(au32)); // only nslots[0] needs zeroing
     eq->nsols = 0;
```
teor2345, line 302: This allows us to re-use an `eq` instance, and we've just cleared the slot memory.

```suggestion
    equi_clearslots(eq);
    eq->nsols = 0;
```

components/equihash/tromp/equi_miner.c
```diff
@@ -649,50 +693,33 @@ nc++,       candidate(eq, tree_from_bid(bucketid, s0, s1));
...
-
 void *worker(void *vp) {
   thread_ctx *tp = (thread_ctx *)vp;
   equi *eq = tp->eq;
 
-  if (tp->id == 0)
+//  if (tp->id == 0)
 //    printf("Digit 0\n");
-  barrier(&eq->barry);
   equi_digit0(eq, tp->id);
```
teor2345, line 705: Just keeping this code in sync
```suggestion
  if (tp->id == 0) {
    equi_clearslots(eq);
  }
  equi_digit0(eq, tp->id);
```

components/equihash/src/tromp.rs
```diff
@@ -30,23 +31,30 @@ extern "C" {
     fn equi_digiteven(eq: *mut CEqui, r: u32, id: u32);
     fn equi_digitK(eq: *mut CEqui, id: u32);
```
teor2345, line 32: Since we're not using threads, we could remove `id` from all these functions and replace it with zero in C.

str4d (approved):
Changes generally LGTM. I'm going to merge this and then do some refactoring / squashing in #1083, to in particular ensure that this is also usable by `zcashd` (so we can deduplicate the code therein).

PR #1102, opened 2024-01-07
Various documentation improvements

str4d:


nuttycom (dismissed):
utACK 72c427487be8d2164e2d0bcf505fd9a1173c89e7

It might be helpful to add some more text that describes how the Download/Scan/Delete/Enhance box can relate to the overall cycle; it's not strictly necessary to exit to the `Scan ranges updated -> Update chain tip` flow on any given schedule; for example, in wallet recovery, once you've obtained a set of scan ranges, it may not make sense to perform chain tip updates, or even respond to continuity errors until all of the blocks in "stable" ranges have been scanned, depending upon your use case. 

It should also be documented that there's no need to exit the download/scan loop to update scan ranges at all unless either transactions belonging to the wallet are discovered, or the chain tip changes. For example, if a higher-level process monitors changes to the chain tip, when the chain tip changes that higher-level process could then "manually" add a single-block range to their local cache of the suggested scan ranges. 

In general I think that the download/scan loop should be well-isolated from the higher-level process; a set of scan ranges is never invalidated below the max reorg depth, though it can be reprioritized by newly-received information.

nuttycom (approved):
re-utACK

PR #1103, opened 2024-01-07
Doc improvements

AArnott:


zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -44,9 +44,9 @@ pub const MINIMUM_FEE: NonNegativeAmount = NonNegativeAmount::const_from_u64(10_
 
 /// A [`FeeRule`] implementation that implements the [ZIP 317] fee rule.
 ///
-/// This fee rule supports only P2pkh transparent inputs; an error will be returned if a coin
-/// containing a non-p2pkh script is provided as an input.  This fee rule may slightly overestimate
-/// fees in case where the user is attempting to spend more than ~150 transparent inputs.
+/// This fee rule supports inputs from orchard, sapling, and transparent pools (P2PKH only).
+/// An error will be returned if a coin containing a non-p2pkh script is provided as an input.
```
str4d, line 48:
```suggestion
/// This fee rule supports Orchard, Sapling, and (P2PKH only) transparent inputs.
/// Returns an error if a coin containing a non-p2pkh script is provided as an input.
```

zcash_keys/src/encoding.rs
```diff
@@ -76,13 +76,24 @@ where
     }
 }
 
+/// A trait for encoding and decoding Zcash addresses.
 pub trait AddressCodec<P>
```
str4d, line 80: I'm happy to merge the PR with these docs, but this trait is almost certainly going away (as the address encoding and decoding logic here is duplicative of `zcash_address`).

str4d (changes requested):

nuttycom (approved):
ACK, doc-only.

str4d (approved):
ACK 296e770ec827c9a500e6751f681e1d79371734a9

PR #1104, opened 2024-01-08
Drop unused `mainnet` feature

AArnott:
Nothing seems to have depended on it, and my testing suggests the same build of this crate can work on both networks.


nuttycom (approved):
utACK, this was vestigial

Issue #1087, opened 2024-01-04
zcash_client_sqlite: Update `InputSource` implementation for `WalletDb` to support Orchard


nuttycom:
Noted in https://github.com/zcash/librustzcash/pull/1085/files/d4aa2d6f066d9757b6bf79c4b18a1b48bc1a44f8#r1442284049 and https://github.com/zcash/librustzcash/pull/1085/files/d4aa2d6f066d9757b6bf79c4b18a1b48bc1a44f8#r1442284758

PR #1105, opened 2024-01-09
zcash_client_backend: add Orchard spends and outputs to transaction construction

nuttycom:
Builds upon #1187


zcash_client_backend/src/data_api.rs
```diff
@@ -1091,12 +1102,48 @@ pub trait WalletCommitmentTrees {
...
+                { ORCHARD_SHARD_HEIGHT * 2 },
+                ORCHARD_SHARD_HEIGHT,
+            >,
+        ) -> Result<A, E>,
+        E: From<ShardTreeError<Self::Error>>;
+
+    /// Adds a sequence of Orchard note commitment tree subtree roots to the data store.
+    ///
+    /// Each such value should be the Merkle root of a subtree of the Orchard note commitment tree
+    /// containing 2^[`SAPLING_SHARD_HEIGHT`] note commitments.
```
daira, line 1140:
```suggestion
    /// containing 2^[`ORCHARD_SHARD_HEIGHT`] note commitments.
```

zcash_client_backend/src/data_api/error.rs
```diff
@@ -120,8 +121,8 @@ where
             Error::ScanRequired => write!(f, "Must scan blocks first"),
             Error::Builder(e) => write!(f, "An error occurred building the transaction: {}", e),
             Error::MemoForbidden => write!(f, "It is not possible to send a memo to a transparent address."),
-            Error::UnsupportedPoolType(t) => write!(f, "Attempted to send to an unsupported pool: {}", t),
-            Error::NoSupportedReceivers(t) => write!(f, "Unified address contained only unsupported receiver types: {:?}", &t[..]),
+            Error::UnsupportedChangeType(t) => write!(f, "Attempted to send change to an unsupported pool type: {}", t),
+            Error::NoSupportedReceivers(_) => write!(f, "A recipient's unified address not contain any receivers to which the wallet can send funds."),
```
daira, line 125:
```suggestion
            Error::NoSupportedReceivers(_) => write!(f, "A recipient's unified address does not contain any receivers to which the wallet can send funds."),
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -695,6 +668,62 @@ where
...
+        #[cfg(feature = "transparent-inputs")]
+        return if proposal.is_shielding() {
+            Some(orchard::keys::OutgoingViewingKey::from(
+                usk.transparent()
+                    .to_account_pubkey()
+                    .internal_ovk()
+                    .as_bytes(),
+            ))
+        } else {
+            Some(orchard_fvk.to_ovk(orchard::keys::Scope::Internal))
```
daira, line 711:
```suggestion
            Some(orchard_fvk.to_ovk(orchard::keys::Scope::Internal))
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -661,6 +687,62 @@ where
...
+        };
+
+        #[cfg(not(feature = "transparent-inputs"))]
```
daira, line 714:
```suggestion
        if proposal.is_shielding() {
            return Some(orchard::keys::OutgoingViewingKey::from(
                usk.transparent()
                    .to_account_pubkey()
                    .internal_ovk()
                    .as_bytes(),
            ))
        }

```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -239,6 +239,75 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
...
         self.is_shielding
     }
+
+    /// Returns whether or not this proposal requires interaction with the specified pool
+    //
+    // TODO: the `TransactionRequest` type doesn't contain quite enough information about
+    // recipients, because it does not resolve each output payment address to a single receiver.
+    // In the presence of source restriction metadata or pool-crossing prohibitions, we need to
+    // resolve receiving pools prior to constructing the proposal; otherwise we don't have
+    // sufficient information to fully determine the privacy implications of a transaction.
```
daira, line 249: Since this is resolved by #1187, let's make sure it doesn't remain after both are merged.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -239,6 +239,75 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
...
+        match pool_type {
+            PoolType::Transparent => {
+                self.is_shielding
+                    || !self.transparent_inputs.is_empty()
+                    || self.transaction_request.payments().iter().any(|p| {
+                        match &p.recipient_address {
+                            Address::Sapling(_) => false,
+                            Address::Transparent(_) => true,
+                            Address::Unified(addr) => {
+                                !addr.has_orchard() && !addr.has_sapling() && addr.has_transparent()
```
daira, line 260: Please add a more specific comment about this line (if it's not fixed before this merges).

nuttycom: What needs to happen here is that after #1187, this should check that the transparent component of the UA is used.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -239,6 +239,75 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
...
+                let sapling_in = self.shielded_inputs.iter().any(|s_in| {
+                    s_in.notes()
+                        .iter()
+                        .any(|note| matches!(note.note(), Note::Sapling(_)))
+                });
+                let sapling_out = self.transaction_request.payments().iter().any(|p| {
+                    match &p.recipient_address {
+                        Address::Sapling(_) => true,
+                        Address::Transparent(_) => false,
+                        Address::Unified(addr) => !addr.has_orchard() && addr.has_sapling(),
```
daira, line 275: and here.

nuttycom, line 275: And the same should be adjusted here. It looks like I should probably at this point rebase this whole PR atop #1187

daira (commented):
I believe this is going to be rebased on top of #1187, but looks good so far.

AArnott:
Accrues toward #403

zcash_client_backend/src/data_api/error.rs
```diff
@@ -66,11 +67,11 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError> {
     /// It is forbidden to provide a memo when constructing a transparent output.
     MemoForbidden,
 
-    /// Attempted to create a spend to an unsupported pool type (currently, Orchard).
-    UnsupportedPoolType(PoolType),
+    /// Attempted to create a send change to an unsupported pool.
```
daira, line 70:
```suggestion
    /// Attempted to create a send with change to an unsupported pool.
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -842,24 +936,32 @@ where
...
                     }
-                } else {
-                    return Err(Error::NoSupportedReceivers(
-                        ua.unknown().iter().map(|(tc, _)| *tc).collect(),
-                    ));
                 }
+
+                return Err(Error::NoSupportedReceivers(Box::new(ua.clone())));
             }
             Address::Sapling(addr) => {
```
daira, line 954: Maybe we could define a closure to add an output of each type, so that the Sapling and transparent code here would be deduplicated? It's not a lot of code, so non-blocking.

zcash_client_backend/src/fees/common.rs
```diff
@@ -86,7 +87,6 @@ where
 
     // TODO: implement a less naive strategy for selecting the pool to which change will be sent.
```
daira, line 88:
```suggestion
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -465,6 +468,7 @@ impl<Cache> TestState<Cache> {
             ovk_policy,
             min_confirmations,
             change_memo,
+            ShieldedProtocol::Sapling,
```
daira, line 471: This might create a bias toward tests only testing change sent to Sapling.

daira (approved):
utACK

str4d (commented):
Flushing comments from my in-progress review.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: This had previously been implemented using closures as you suggest, but additional changes that will be coming in #1143 make this approach a bit cleaner. I refactored it this way in #1087 in order to eventually avoid conflicts with #1143.

zcash_client_backend/src/fees/common.rs
nuttycom: This TODO still applies; for example, by sending change and/or recipient values in multiple outputs we could potentially avoid pool-crossing entirely.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -955,27 +1058,28 @@ where
...
-                                &bundle.shielded_outputs()[output_index],
-                                consensus::sapling_zip212_enforcement(params, min_target_height),
-                            )
-                            .map(|(note, _, _)| (account, note))
-                        })
-                } else {
-                    None
-                };
+                let recipient = recipient
+                    .map(|pool| {
```
str4d, line 1073: I was confused by this for at least five minutes, trying to figure out why `recipient` was optional and why this code was being significantly altered, until I realised that this is the new `Recipient::map` method that only affects `Recipient::InternalAccount`. If we're going to use this approach, please rename it for clarity (to e.g. `Recipient::map_internal_account`).

zcash_client_backend/CHANGELOG.md
```diff
@@ -232,6 +233,11 @@ and this library adheres to Rust's notion of
   - The fields of `ReceivedSaplingNote` are now private. Use
     `ReceivedSaplingNote::from_parts` for construction instead. Accessor methods
     are provided for each previously public field.
+  - `Recipient` is now polymorphic in the type of the payload for wallet-internal
+    recipients. This simplifies the handling of wallet-internal outputs.
+  - `SentTransactionOutput::from_parts` now takes a `Recipient<Note>`
+  - `SentTransactionOutput::recipient` now returns a `Recipient<Note>`
```
str4d, line 241:
```suggestion
  - `SentTransactionOutput::from_parts` now takes a `Recipient<Note>`.
  - `SentTransactionOutput::recipient` now returns a `Recipient<Note>`.
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -232,6 +233,11 @@ and this library adheres to Rust's notion of
...
+  - `OvkPolicy::Custom` now wraps a bare `[u8; 32]` instead of a Sapling `OutgoingViewingKey`.
```
str4d, line 242: If the "document `OvkPolicy::Custom` very carefully" option is taken, this changelog entry needs to be correspondingly expanded.

zcash_client_backend/CHANGELOG.md
```diff
@@ -154,8 +154,9 @@ and this library adheres to Rust's notion of
       the database identifiers for its contained notes by universally quantifying
       the `NoteRef` type parameter.
     - It returns a `NonEmpty<TxId>` instead of a single `TxId` value.
-  - `wallet::create_spend_to_address` now takes an additional `change_memo`
-    argument. It also returns its result as a `NonEmpty<TxId>` instead of a
+  - `wallet::create_spend_to_address` now takes additional `change_memo` and
+    `fallback_change_pool` arguments. It also returns its result as a
+    `NonEmpty<TxId>` instead of a
     single `TxId`.
```
str4d, line 160:
```suggestion
    `NonEmpty<TxId>` instead of a single `TxId`.
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -984,7 +1002,6 @@ where
                     Some(memo),
                 ))
             }
-            #[cfg(zcash_unstable = "orchard")]
```
str4d, line 897: Why is this removed?

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -644,73 +652,83 @@ where
...
 
-    let (sapling_anchor, sapling_inputs) = proposal_step.shielded_inputs().map_or_else(
-        || Ok((sapling::Anchor::empty_tree(), vec![])),
-        |inputs| {
-            wallet_db.with_sapling_tree_mut::<_, _, Error<_, _, _, _>>(|sapling_tree| {
-                let anchor = sapling_tree
-                    .root_at_checkpoint_id(&inputs.anchor_height())?
-                    .into();
+    let (sapling_anchor, sapling_inputs) =
+        if proposal_step.involves(PoolType::Shielded(ShieldedProtocol::Sapling)) {
```
str4d, line 656: I don't fully understand what this extra conditional has to do with allowing the proposer to specify a fallback change pool. But I think it's an optimisation to not construct a protocol's builder if we know we won't be generating any outputs (real or padded) for it, which we now know deterministically after providing a fallback change pool?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -455,9 +455,9 @@ where
                 Err(other) => return Err(other.into()),
             }
 
-            #[cfg(not(zcash_unstable = "orchard"))]
+            #[cfg(not(feature = "orchard"))]
             let selectable_pools = &[ShieldedProtocol::Sapling];
-            #[cfg(zcash_unstable = "orchard")]
+            #[cfg(feature = "orchard")]
```
str4d, line 460: Hmm: we now don't attempt to select from the Orchard pool if `orchard` is disabled. What was the previous behaviour that made this change necessary, given that the `wallet_db` method is publicly reachable and thus other people could also run into it?

zcash_client_backend/src/lib.rs
```diff
@@ -97,7 +97,6 @@ pub enum ShieldedProtocol {
     /// The Sapling protocol
     Sapling,
     /// The Orchard protocol
-    #[cfg(zcash_unstable = "orchard")]
```
str4d, line 100: Why is this being removed before the release that adds Orchard support?

zcash_client_sqlite/src/testing.rs
str4d, line 471: Make this an argument of the test helper method (it is intended to only provide the common internal parameters).

zcash_client_sqlite/src/testing.rs
```diff
@@ -567,6 +571,7 @@ impl<Cache> TestState<Cache> {
             amount,
             memo,
             change_memo,
+            ShieldedProtocol::Sapling,
```
str4d, line 574: Ditto.

zcash_client_sqlite/src/testing.rs
```diff
@@ -1058,7 +1063,8 @@ pub(crate) fn input_selector(
     standard::SingleOutputChangeStrategy,
 > {
     let change_memo = change_memo.map(|m| MemoBytes::from(m.parse::<Memo>().unwrap()));
-    let change_strategy = standard::SingleOutputChangeStrategy::new(fee_rule, change_memo);
+    let change_strategy =
+        standard::SingleOutputChangeStrategy::new(fee_rule, change_memo, ShieldedProtocol::Sapling);
```
str4d, line 1067: I think this also needs to be an argument of the test helper, if the others are?

str4d (changes requested):
Reviewed as of 8c78d7f2a01d4a4dfd6530cf98c6cf31b527e525.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -851,6 +824,62 @@ where
...
+        #[cfg(feature = "transparent-inputs")]
+        return if proposal_step.is_shielding() {
+            Some(orchard::keys::OutgoingViewingKey::from(
+                usk.transparent()
+                    .to_account_pubkey()
+                    .internal_ovk()
+                    .as_bytes(),
+            ))
+        } else {
+            Some(orchard_fvk.to_ovk(orchard::keys::Scope::Internal))
```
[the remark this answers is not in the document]
str4d: This line was later removed, leaving the code further down which already does this.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -851,6 +824,62 @@ where
         utxos_spent
     };
 
+    #[cfg(feature = "orchard")]
+    let orchard_fvk: orchard::keys::FullViewingKey = usk.orchard().into();
+
+    #[cfg(feature = "orchard")]
+    let orchard_external_ovk = match ovk_policy {
+        OvkPolicy::Sender => Some(orchard_fvk.to_ovk(orchard::keys::Scope::External)),
```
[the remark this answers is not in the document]
nuttycom: Yes, they are now, which is why I removed the qualifier.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: It's just a functor! But fine.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: It's an optimization to avoid constructing the ShardTree.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: The objective here is to make the zcash_unstable only be a gate for the 'orchard' feature, which is then the ultimate gate (that will remain in place permanently) for Orchard functionality. Removing the other uses of the config flag means that we have less variation in compilation to account for.

zcash_client_backend/src/lib.rs
nuttycom: Same reason as previous comment.

zcash_client_sqlite/src/testing.rs
nuttycom: This will change when we start testing Orchard functionality, because I want to refactor the tests to make our existing Sapling pool tests reusable for Orchard; making this change right now is more invasive than necessary.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -851,6 +824,62 @@ where
...
+        OvkPolicy::Custom(ovk) => Some(orchard::keys::OutgoingViewingKey::from(ovk)),
```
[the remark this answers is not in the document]
nuttycom: I think that unless we were to go with allowing the user to provide a different OVK for each output, it doesn't make sense to distinguish by pool. That being said, perhaps we should consistently use the most-preferred OVK for all kinds of outputs.

zcash_client_backend/src/lib.rs
str4d: In that case I need to re-review this PR with an eye for everything that was behind this config flag being ready for stabilisation, because I was under the impression that _everything_ Orchard-specific was only being stabilised in the release after next.

zcash_client_sqlite/src/testing.rs
str4d: In that case, open an issue for this.

zcash_client_backend/src/data_api/wallet.rs
str4d: Mmk. Still not clear what necessitated it to happen in the fallback change pool commit, but the optimisation makes sense.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -140,8 +141,8 @@ where
             Error::ScanRequired => write!(f, "Must scan blocks first"),
             Error::Builder(e) => write!(f, "An error occurred building the transaction: {}", e),
             Error::MemoForbidden => write!(f, "It is not possible to send a memo to a transparent address."),
-            Error::UnsupportedPoolType(t) => write!(f, "Attempted to send to an unsupported pool: {}", t),
-            Error::NoSupportedReceivers(t) => write!(f, "Unified address contained only unsupported receiver types: {:?}", &t[..]),
```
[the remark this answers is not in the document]
nuttycom: We can do so for receiver types that are recognized, but the `Unknown` receivers will be a little ugly.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Since the Orchard code is being added, this was to make the two implementations consistent with one another.

zcash_client_sqlite/src/lib.rs
```diff
@@ -591,12 +591,13 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
             for output in d_tx.sapling_outputs {
                 match output.transfer_type {
                     TransferType::Outgoing | TransferType::WalletInternal => {
+                        let value = output.note.value();
```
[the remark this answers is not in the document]
nuttycom: The issue is with the lifetime of `output.note`.

zcash_client_backend/src/fees/fixed.rs
```diff
@@ -22,15 +24,21 @@ use super::orchard as orchard_fees;
 pub struct SingleOutputChangeStrategy {
     fee_rule: FixedFeeRule,
     change_memo: Option<MemoBytes>,
+    fallback_change_pool: ShieldedProtocol,
 }
 
 impl SingleOutputChangeStrategy {
     /// Constructs a new [`SingleOutputChangeStrategy`] with the specified fee rule
     /// and change memo.
```
str4d, line 32:
```suggestion
    /// and change memo.
    ///
    /// `fallback_change_pool` is used when more than one shielded pool is enabled via
    /// feature flags, and the transaction has no shielded inputs.
```

zcash_client_backend/src/fees/standard.rs
```diff
@@ -27,15 +29,21 @@ use super::orchard as orchard_fees;
 pub struct SingleOutputChangeStrategy {
     fee_rule: StandardFeeRule,
     change_memo: Option<MemoBytes>,
+    fallback_change_pool: ShieldedProtocol,
 }
 
 impl SingleOutputChangeStrategy {
     /// Constructs a new [`SingleOutputChangeStrategy`] with the specified ZIP 317
     /// fee parameters.
```
str4d, line 37:
```suggestion
    /// fee parameters.
    ///
    /// `fallback_change_pool` is used when more than one shielded pool is enabled via
    /// feature flags, and the transaction has no shielded inputs.
```

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -26,15 +28,21 @@ use super::orchard as orchard_fees;
 pub struct SingleOutputChangeStrategy {
     fee_rule: Zip317FeeRule,
     change_memo: Option<MemoBytes>,
+    fallback_change_pool: ShieldedProtocol,
 }
 
 impl SingleOutputChangeStrategy {
     /// Constructs a new [`SingleOutputChangeStrategy`] with the specified ZIP 317
     /// fee parameters and change memo.
```
str4d, line 36:
```suggestion
    /// fee parameters and change memo.
    ///
    /// `fallback_change_pool` is used when more than one shielded pool is enabled via
    /// feature flags, and the transaction has no shielded inputs.
```

zcash_client_backend/src/proto.rs
```diff
@@ -319,7 +319,6 @@ fn pool_type<T>(pool_id: i32) -> Result<PoolType, ProposalDecodingError<T>> {
     match proposal::ValuePool::try_from(pool_id) {
         Ok(proposal::ValuePool::Transparent) => Ok(PoolType::Transparent),
         Ok(proposal::ValuePool::Sapling) => Ok(PoolType::Shielded(ShieldedProtocol::Sapling)),
-        #[cfg(zcash_unstable = "orchard")]
         Ok(proposal::ValuePool::Orchard) => Ok(PoolType::Shielded(ShieldedProtocol::Orchard)),
         _ => Err(ProposalDecodingError::ValuePoolNotSupported(pool_id)),
```
str4d, line 323: Bug: After removing `#[cfg(zcash_unstable = "orchard")]` here, proposals that contain Orchard spends will be parsed incorrectly. `WalletDb::get_spendable_note` is called with `ShieldedProtocol::Orchard`, but internally it ignores that parameter and always calls `get_spendable_sapling_note`. This may fail with a "note missing" error, but it also may (in the case of an Orchard note received in the same transaction as a Sapling note) return the wrong note. I have not determined what happens after this; I expect something will probably break during transaction creation, but in any case the `Proposal` returned by `zcash_client_backend::proto::proposal::Proposal::try_into_standard_proposal` will be in an invalid state.

Proposals with no Orchard spends, but that request sending to Orchard receivers of UAs, will AFAICT be successfully parsed instead of being rejected with `ProposalDecodingError::ValuePoolNotSupported` (this helper function is called in `zcash_client_backend::proto::proposal::Proposal::try_into_standard_proposal`). I believe what happens instead is that once the user attempts to create transactions from the proposal, the Orchard inputs are ignored, and the resulting transaction fails to balance, which is an unintuitive error.
- EDIT: Nope, currently what happens is that `payment_pools` is being ignored, and the UA's Sapling or transparent receiver would be used, which is a bug. Once that bug is fixed, this will be an unintuitive error.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -827,9 +902,28 @@ where
                     .as_ref()
                     .map_or_else(MemoBytes::empty, |m| m.clone());
 
+                #[cfg(feature = "orchard")]
+                if let Some(orchard_receiver) = ua.orchard() {
```
str4d, line 906: Bug: this (and the code below) is not using the `payment_pools` metadata now-encoded in the proposal, and is creating UA outputs to the preferred pool of this wallet version (instead of the wallet version that made the proposal).

zcash_client_backend/src/proposal.rs
```diff
@@ -490,6 +490,59 @@ impl<NoteRef> Step<NoteRef> {
...
+                    .values()
+                    .any(|pool| matches!(pool, PoolType::Shielded(ShieldedProtocol::Orchard)));
+                let orchard_change = self
+                    .balance
+                    .proposed_change()
+                    .iter()
+                    .any(|c| c.output_pool() == ShieldedProtocol::Orchard);
+
+                orchard_in || orchard_out || orchard_change
+            }
```
str4d, line 543: After removing `#[cfg(zcash_unstable = "orchard")]` elsewhere, I think this should also be ungated. Now that (per my earlier comment) parsing an Orchard-output-only-containing proposal will succeed, we need to report whether any such outputs are requested.
```suggestion
            PoolType::Shielded(ShieldedProtocol::Orchard) => {
                #[cfg(not(feature = "orchard"))]
                let orchard_in = false;
                #[cfg(feature = "orchard")]
                let orchard_in = self.shielded_inputs.iter().any(|s_in| {
                    s_in.notes()
                        .iter()
                        .any(|note| matches!(note.note(), Note::Orchard(_)))
                });
                let orchard_out = self
                    .payment_pools()
                    .values()
                    .any(|pool| matches!(pool, PoolType::Shielded(ShieldedProtocol::Orchard)));
                let orchard_change = self
                    .balance
                    .proposed_change()
                    .iter()
                    .any(|c| c.output_pool() == ShieldedProtocol::Orchard);

                orchard_in || orchard_out || orchard_change
            }
```

str4d (changes requested):
Reviewed as of 050a124cb600cae93161d9eea0a07376c209ee18.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: `ShieldedProtocol::Orchard` is stabilized, so this is then required for the match to be total.

zcash_client_backend/src/proto.rs
nuttycom: Okay, so this is actually a revealed bug in `zcash_client_sqlite`, not a bug here.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -817,49 +842,142 @@ where
...
-                    ));
-                } else if let Some(taddr) = ua.transparent() {
-                    if payment.memo.is_some() {
-                        return Err(Error::MemoForbidden);
-                    } else {
-                        builder.add_transparent_output(taddr, payment.amount)?;
+                match output_pool {
+                    #[cfg(not(feature = "orchard"))]
+                    PoolType::Shielded(ShieldedProtocol::Orchard) => {
+                        return Err(Error::ProposalNotSupported);
```
str4d, line 921: The documentation for this error variant is now out of date.

str4d (approved):
utACK 184286c43092be121a0d5bb2d52c6479a54551e1

PR #1106, opened 2024-01-10
Add `WalletRead::get_account_ids` function

AArnott:
As per our [discord discussion](https://discord.com/channels/809218587167293450/972649509651906711/1193920166598287421), this needs to be exposed via the API since the sqlite 'public API' is said to not include most of the tables, including the accounts table.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -140,7 +140,7 @@ and this library adheres to Rust's notion of
     - `get_spendable_sapling_notes`, `select_spendable_sapling_notes`, and
       `get_unspent_transparent_outputs` have been removed; use
       `data_api::InputSource` instead.
-    - Added `get_account_ids`
+    - Added `get_account_ids`.
   - `wallet::{propose_shielding, shield_transparent_funds}` now takes their
     `min_confirmations` arguments as `u32` rather than a `NonZeroU32` to permit
     implmentations to enable zero-conf shielding.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -140,6 +140,7 @@ and this library adheres to Rust's notion of
     - `get_spendable_sapling_notes`, `select_spendable_sapling_notes`, and
       `get_unspent_transparent_outputs` have been removed; use
       `data_api::InputSource` instead.
+    - Added `get_account_ids`
   - `wallet::{propose_shielding, shield_transparent_funds}` now takes their
     `min_confirmations` arguments as `u32` rather than a `NonZeroU32` to permit
     implmentations to enable zero-conf shielding.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -560,6 +560,9 @@ pub trait WalletRead {
         account: AccountId,
         max_height: BlockHeight,
     ) -> Result<HashMap<TransparentAddress, Amount>, Self::Error>;
+
+    /// Returns a vector with the IDs of all accounts known to this wallet.
+    fn get_account_ids(&self) -> Result<Vec<AccountId>, Self::Error>;
 }
 
 /// Metadata describing the sizes of the zcash note commitment trees as of a particular block.
@@ -1282,6 +1285,10 @@ pub mod testing {
             _max_height: BlockHeight,
         ) -> Result<HashMap<TransparentAddress, Amount>, Self::Error> {
             Ok(HashMap::new())
+        }
+
+        fn get_account_ids(&self) -> Result<Vec<AccountId>, Self::Error> {
+            Ok(Vec::new())
         }
     }
 
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -1458,6 +1458,20 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
     Ok(res)
 }
 
+/// Returns a vector with the IDs of all accounts known to this wallet.
+pub(crate) fn get_account_ids(conn: &rusqlite::Connection) -> Result<Vec<AccountId>, SqliteClientError> {
+    let mut stmt = conn.prepare("SELECT account FROM accounts")?;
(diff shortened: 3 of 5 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -1458,6 +1458,20 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
     Ok(res)
 }
 
+/// Returns a vector with the IDs of all accounts known to this wallet.
+pub fn get_accounts(conn: &rusqlite::Connection) -> Result<Vec<AccountId>, SqliteClientError> {
```
str4d, line 1462:
```suggestion
pub(crate) fn get_accounts(conn: &rusqlite::Connection) -> Result<Vec<AccountId>, SqliteClientError> {
```

zcash_client_backend/src/data_api.rs
```diff
@@ -560,6 +560,9 @@ pub trait WalletRead {
         account: AccountId,
         max_height: BlockHeight,
     ) -> Result<HashMap<TransparentAddress, Amount>, Self::Error>;
+
+    /// Returns a vector with the IDs of all accounts known to this wallet.
+    fn get_account_ids(&self) -> Result<Vec<AccountId>, Self::Error>;
```
nuttycom, line 565: Since we're starting to be a bit forward-looking, I'm wondering now whether we shouldn't come up with a better abstraction than `AccountId`; the `AccountId` itself is really a low-level implementation detail, and it assumes a single seed. One could alternately use a tuple `(SeedFingerprint, AccountId)` but that doesn't necessarily support FVK-only or IVK-only use cases very well because in those cases you probably don't have a seed fingerprint available, and so would potentially have to fake it; realistically, you may also not have the account ID for zip32 derivation, so we probably need a different approach here. 

Perhaps a better approach would be to use the UIVK for the account identifier everywhere we need it?

nuttycom (approved):
utACK with suggestion for a potentially more comprehensive change. If we're going to be generalizing to new ways of using `zcash_client_backend` in the near future, I'm not sure whether it makes sense to introduce this API and then remove it, or whether we should pursue a more aggressive change.

zcash_client_backend/src/data_api.rs
str4d, line 565: Document this change to the `WalletRead` trait in the changelog (add it to the existing list of changes there in the "Unreleased" section).

str4d (changes requested):

zcash_client_backend/src/data_api.rs
str4d: The UIVK is way heavier than we need, and it also makes identifying the corresponding account much more complex (because the UIVK doesn't necessarily contain all the same receiver types as the wallet supports, and then we'd have a whole extra axis to handle when a UIVK contains some receivers that do correspond to an account and other receivers that don't, or a UIVK that contains receivers from two accounts under the same or different seeds).

nuttycom: The other possible approach, which I like less, is using a database-internal identifier, a UUID or some such, but we've been moving away from the wallet interfaces leaking internal identifiers.

nuttycom: Another, even better, possibility: how about we make `AccountId` an associated type on the `WalletRead` trait? Then, for the sqlite implementation for now, we can continue to use the `zip32::AccountId` type, but in the future we can change that type without breaking code for which that type is existentially quantified.

AArnott: Really good questions. I agree this addition can't stand as-is given the use cases we'd like to add later. But I suspect supporting all that will be a far more sweeping change that is likely to be months off, at least. This addresses a need that we have now and later, as the `AccountId` should be replaceable (across all the API both new and pre-existing) across the board with whatever the new shape takes.
So my vote would be in favor of merging the PR (after addressing the feedback), and I look forward to contributing to the additional use cases soon after.

FWIW, my current idea for the future of the accounts table would be one column for the IVK of each individual pool, so that no question of which receivers a UIVK comes up as a cause for an account lookup failure. The primary key of the table shouldn't be any one of these columns because no one column is guaranteed to be non-null. A multi-key PK would be both heavy and not be friendly to lookups either. 
As my understanding of db schema best practices suggests that PKs should _not_ be user data at all, I'd then add a auto-increment PK, that in the rust crate we can expose as a struct with a private field. Callers can use an API to exchange an IVK (or any other kind of key) for the new PK that represents the account, and keep that around for as long as they need to interact with that account, or lookup repeatedly -- whatever is convenient and performant for them.

str4d: For this PR, we'll leave account IDs as-is. I've opened https://github.com/zcash/librustzcash/issues/1107 for looking into generalising account IDs.

AArnott: Done

AArnott: I also added a small `.editorconfig` file to preserve the space style indentation in that file, since my editor defaulted to tabs there.

zcash_client_sqlite/src/wallet.rs
AArnott: Done. I also fixed the name of this function to match the others (`get_account_ids`).

zcash_client_backend/CHANGELOG.md
```diff
@@ -140,6 +140,7 @@ and this library adheres to Rust's notion of
     - `get_spendable_sapling_notes`, `select_spendable_sapling_notes`, and
       `get_unspent_transparent_outputs` have been removed; use
       `data_api::InputSource` instead.
+    - Added `get_account_ids`
```
daira, line 143:
```suggestion
    - Added `get_account_ids`.
```

daira (dismissed):
utACK with a non-blocking nit.

nuttycom (approved):
utACK 28ed048c7ea9954a3c69b699c2e752a946b1d4a9

daira (approved):
utACK

str4d (approved):
utACK 28ed048c7ea9954a3c69b699c2e752a946b1d4a9

PR #1113, opened 2024-01-12
zcash_address: Remove private type aliases for raw data sizes

str4d:


nuttycom (dismissed):
utACK bc14ee63f48549640203afe55dcd5bad7dee5b76

A possible alternative would have been to make these type aliases public, but maybe that's just increasing the API surface area that a user has to inspect without adding safety?

nuttycom:
Another alternative would be to make these type aliases public newtype wrappers, but that would be a breaking change.

str4d:
If we added the type aliases to the public API, all we get is nicer naming of these for the downstream users, not type safety. If we add type-safe wrappers, then we get a modicum type safety in the trait method impls (and e.g. the same type in `unified::address::Receiver` enum cases), but that's only really useful for people parsing the addresses and then doing nothing with the data (not even crossing an FFI). For downstream users that need to parse the data itself, all we're doing is forcing them to unwrap an extra layer.

nuttycom (approved):
re-utACK c69169ba1ae7c3d63b8624655d1e51e3dcd0bd67

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -83,6 +83,30 @@ impl SealedItem for Ivk {
...
+///
+/// let (network, uivk) = unified::Uivk::decode(example_uivk)?;
+///
+/// // We can obtain the pool-specific Incoming Viewing Keys for the UIVK in
+/// // preference order (the order in which wallets should prefer to use their
+/// // corresponding address receivers):
+/// let ivks = uivk.items();
+///
+/// // And we can create the UIVK from a list of IVKs:
+/// let new_uivk = unified::Uivk::try_from_items(ivks)?;
```
nuttycom, line 105: Ideally the documentation here would make the `Item` type more evident; all that this documentation provides is the knowledge that you can roundtrip, but usually you'll be either obtaining the items (and will need to understand their type) or constructing a new IVK yourself from more primitive parts, and will need to wrap them.

components/zcash_address/src/lib.rs
```diff
@@ -1,3 +1,116 @@
...
+//! ## I just need to validate Zcash addresses
+//!
+//! ```
+//! # use zcash_address::ZcashAddress;
+//! fn is_valid_zcash_address(addr_string: &str) -> bool {
+//!     addr_string.parse::<ZcashAddress>().is_ok()
+//! }
+//! ```
+//!
+//! ## I want to parse mainnet Zcash addresses in a language that supports C FFI
```
nuttycom, line 32: Minor prioritization nit, I think that this should probably be the last item in the list; FFI usage is _uncommon_.

components/zcash_address/src/lib.rs
```diff
@@ -1,3 +1,116 @@
...
+//! # Using this crate
+//!
+//! ## I just need to validate Zcash addresses
+//!
+//! ```
+//! # use zcash_address::ZcashAddress;
+//! fn is_valid_zcash_address(addr_string: &str) -> bool {
+//!     addr_string.parse::<ZcashAddress>().is_ok()
+//! }
+//! ```
```
nuttycom, line 30: The critical thing that I think this documentation is lacking is clarity as to how `ZcashAddress` relates to the protocol types, specifically that if you want to actually _use_ the addresses and/or keys parsed using this crate, you need to use something like the types in `zcash_client_backend`. To that point, I think that we ought to factor those out into their own crate at this point; it shouldn't be necessary to pull in all of `zcash_client_backend` just to get access to the address types that you end up wanting to use with `zcash_primitives`. 

Alternately, we could promote the unified address types up into `zcash_primitives`.

nuttycom (dismissed):
utACK 7248dcbeff2b754d1b987541e211ae4d06194c8a with additional documentation suggestions; I think that the fact that this crate _only_ handles parsing and encoding (and doesn't for example provide anything for deriving addresses from keys) is going to catch folks off guard. It's not intended to be a complete solution, and that's fine, but it's a bit of an idiosyncratic line that we've drawn.

components/zcash_address/src/lib.rs
nuttycom: Ideally I'd want examples that use the `sapling-crypto`, `orchard` and/or `zcash_client_backend` types here, because that's what users will actually need.

components/zcash_address/src/kind/unified/ivk.rs
str4d: That's easy to add: `let ivks: Vec<unified::Ivk> = ...`.

components/zcash_address/src/lib.rs
str4d: Indeed, the thing this crate is missing is another crate we can point people to when they are writing a Rust app and _don't_ need their own types, and are happy to use ours. This crate isn't intended to be that crate.

str4d: Added a few more notes about this.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/7248dcbeff2b754d1b987541e211ae4d06194c8a..29dc5d1d5c789d32bee1e08a22332bb9aa08900f) to address review comments.

nuttycom (approved):
re-utACK https://github.com/zcash/librustzcash/pull/1113/commits/29dc5d1d5c789d32bee1e08a22332bb9aa08900f

PR #1120, opened 2024-01-22
zcash_client_backend: Enable `zcash_keys/unstable` feature flag

str4d:
Due to `zcash_keys` being in `zcash_client_backend`'s public API, their unstable APIs are a common API surface (as they were before `zcash_keys` was extracted).

This can be reverted once `zcash_keys` types with unstable APIs are not being re-exported from `zcash_client_backend`.


nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK

PR #1122, opened 2024-01-22
Improve light client performance

str4d:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -727,13 +741,37 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
                 balances.add_unshielded_value(value)?;
             }
         }
+        drop(transparent_trace);
     }
 
+    let next_sapling_subtree_index = {
+        // This will return a runtime error if we call `get_wallet_summary` from two
+        // threads at the same time, as transactions cannot nest.
+        let tx = conn.unchecked_transaction()?;
```
nuttycom, line 750: Why not have this whole method take a `rusqlite::Transaction` rather than a `rusqlite::Connection`?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -727,13 +741,37 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
+        let shard_store =
+            SqliteShardStore::<_, ::sapling::Node, SAPLING_SHARD_HEIGHT>::from_connection(
+                &tx,
+                SAPLING_TABLES_PREFIX,
+            )?;
+
+        // The last shard will be incomplete, and we want the next range to overlap with
+        // the last complete shard, so return the index of the second-to-last shard root.
+        shard_store
+            .get_shard_roots()
```
nuttycom, line 760: Instead of constructing a `SqliteShardStore` and calling `get_shard_roots`, I think it would be better to just query the `sapling_shard_roots` table directly. It _should_ be equivalent, but querying `sapling_shard_roots` directly ensures that we _only_ get data downloaded from lightwallet, and not internally derived roots.

str4d: I'm confused; I cannot find any `sapling_shard_roots` table. There is only `sapling_tree_shards`, which is where `WalletCommitmentTrees::put_sapling_subtree_roots` inserts into, and what `shard_store.get_shard_roots()` fetches from.

nuttycom: Oops, yes, `sapling_tree_shards` is what I meant. 

One minor improvement would be to use https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/commitment_tree.rs#L481 directly, rather than having to instantiate a full `SqliteShardStore`.

str4d: Eh, `SqliteShardStore::from_connection` is literally just returning a container without any extra data storage or logic going on, so it's likely that the compiler will optimise this out. I'm fine with the more explicit type-ing.

zcash_client_sqlite/src/testing.rs
```diff
@@ -734,7 +726,7 @@ impl<Cache> TestState<Cache> {
 
     pub(crate) fn get_wallet_summary(&self, min_confirmations: u32) -> Option<WalletSummary> {
         get_wallet_summary(
-            &self.wallet().conn,
+            &self.wallet().conn.unchecked_transaction().unwrap(),
```
nuttycom, line 729: This is fine, because it's test-only.

zcash_client_sqlite/src/wallet.rs
nuttycom: That's okay, I guess. It's a little less insulated against future changes to SaplingShardStore construction; I'd still slightly prefer it the way I suggested but I don't mind too much.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1052,42 +1089,42 @@ pub(crate) fn block_fully_scanned<P: consensus::Parameters>(
...
+                FROM scan_queue
+                WHERE priority = :priority
+                ORDER BY block_range_start ASC
+                LIMIT 1",
+                named_params![":priority": priority_code(&ScanPriority::Scanned)],
+                |row| {
+                    let block_range_start = BlockHeight::from_u32(row.get(0)?);
+                    let block_range_end = BlockHeight::from_u32(row.get(1)?);
+
+                    Ok(if block_range_start <= birthday_height {
```
nuttycom, line 1113:
```suggestion
                    // If the start of the earliest scanned range is greater than 
                    // the birthday height, then there is an unscanned range between 
                    // the wallet birthday and that range, so there is no fully 
                    // scanned height.
                    Ok(if block_range_start <= birthday_height {
```

nuttycom (dismissed):
utACK d07f6d43369082e6f5a23f2b1a6dc04e75f72831 with suggestions

zcash_client_sqlite/src/lib.rs
```diff
@@ -302,8 +302,10 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
         &self,
         min_confirmations: u32,
     ) -> Result<Option<WalletSummary>, Self::Error> {
+        // This will return a runtime error if we call `get_wallet_summary` from two
+        // threads at the same time, as transactions cannot nest.
         wallet::get_wallet_summary(
-            self.conn.borrow(),
+            &self.conn.borrow().unchecked_transaction()?,
```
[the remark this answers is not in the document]
str4d: Opened #1131.

nuttycom (dismissed):
utACK 25aef9aaaed0bd755a6fa12f72a9b75f05372ef1

nuttycom (approved):
utACK https://github.com/zcash/librustzcash/commit/fe902f74bcb46cafb431e6b199addf97f6bfcb20

PR #1127, opened 2024-01-24
receiving_key_scopes migration: Fall back to note commitment matching for scope determination

### LINKED ISSUE #1117: `zcash_client_sqlite`: `receiving_key_scopes` migration appears broken for existing wallets
I have a local test wallet DB working with revision 236cd569ee4a824d98920d1a61b8cc2b90ceba1d. I tried to upgrade it to latest `main` and got the following error:

```
2024-01-17T03:08:09.323554Z  INFO schemer: Migrating up to target: None    
2024-01-17T03:08:09.323897Z  INFO schemer: Migrating up to target: None    
2024-01-17T03:08:09.323928Z  INFO schemer: Applying migration ee89ed2b-c1c2-421e-9e98-c1e3e54a7fc2    
Error: An error occurred while applying migration ee89ed2b-c1c2-421e-9e98-c1e3e54a7fc2 (Add decryption key scope to persisted information about received notes.) up: Invalid column type Null at index: 2, name: raw.

Caused by:
    0: Invalid column type Null at index: 2, name: raw
    1: Invalid column type Null at index: 2, name: raw
```

The migration in question was added in #1055, and AFAICT no changes were made to existing migrations since 236cd569ee4a824d98920d1a61b8cc2b90ceba1d, so AFAICT this migration is intended to work (but of course could itself still change before the next crate release of `zcash_client_sqlite`).

nuttycom:
Fixes #1117

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
@@ -264,3 +264,472 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
         panic!("Cannot revert this migration.");
     }
 }
+
+#[cfg(feature = "transparent-inputs")]
+#[cfg(test)]
+mod tests {
+    use std::convert::Infallible;
+
+    use incrementalmerkletree::Position;
+    use maybe_rayon::{
+        iter::{IndexedParallelIterator, ParallelIterator},
+        slice::ParallelSliceMut,
+    };
+    use rand_core::OsRng;
+    use rusqlite::{named_params, params, Connection};
+    use tempfile::NamedTempFile;
+    use zcash_client_backend::{
+        data_api::{
+            BlockMetadata, DecryptedTransaction, WalletCommitmentTrees, SAPLING_SHARD_HEIGHT,
+        },
+        decrypt_transaction,
+        proto::compact_formats::{CompactBlock, CompactTx},
+        scanning::scan_block,
+        wallet::Recipient,
+        PoolType, ShieldedProtocol, TransferType,
+    };
+    use zcash_keys::keys::{UnifiedFullViewingKey, UnifiedSpendingKey};
+    use zcash_primitives::{
+        block::BlockHash,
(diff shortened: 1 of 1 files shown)


str4d:
This migration fails with a different error:
```
2024-01-24T18:12:08.307240Z  INFO schemer: Applying migration ee89ed2b-c1c2-421e-9e98-c1e3e54a7fc2    
Error: An error occurred while applying migration ee89ed2b-c1c2-421e-9e98-c1e3e54a7fc2 (Add decryption key scope to persisted information about received notes.) up: Invalid column type Null at index: 4, name: expiry_height.

Caused by:
    0: Invalid column type Null at index: 4, name: expiry_height
    1: Invalid column type Null at index: 4, name: expiry_height
```

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -111,13 +177,85 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+                let scope = select_note_scope(
+                    &mut commitment_tree,
+                    dfvk,
+                    &diversifier,
+                    &note_value.try_into().unwrap(),
+                    &rseed,
+                    note_commitment_tree_position,
                 )?;
+
+                if scope == Some(Scope::Internal) {
```
nuttycom, line 252: In the case where `scope == None`, it means that we have no way to spend the note, so the scope of the key is irrelevant. We _could_ add a sentinel for this, but I think that just leaving it as `External` is probably okay?

str4d (approved):
ACK 55a8668a3b45306e580114f2bf313ee891e58e7e with non-blocking nits.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/55a8668a3b45306e580114f2bf313ee891e58e7e..d0a27eda90ffbc9cdf5367aa782a91522eb7f030) to resolve my comments and fix clippy lints.

str4d (approved):
Re-ACK d0a27eda90ffbc9cdf5367aa782a91522eb7f030

PR #1129, opened 2024-01-24
Add `zcash_unstable` compiler flag to fully gate `orchard` functionality

### LINKED ISSUE #1123: Gate Orchard changes behind `rustc` config flag
In order to enable publishing releases of the crates in the near term, we need to gate the incomplete Orchard support (and related changes like the new migration). We have an `orchard` feature flag for most of it, but we don't want people to enable it accidentally with `--all-features`.

We will add a `--cfg zcash_unstable="orchard"` flag that completely prevents using the Orchard code without it. We will add that gate directly to the Orchard code that isn't behind a feature flag, and add a top-level directive that prevents compiling with the feature flag enabled if the `rustc` flag isn't set. The latter will prevent anyone from using `--all-features`, but that is desirable in this case.

nuttycom:
Closes #1123.

### DIFF
--- a/zcash_client_backend/src/data_api/error.rs
+++ b/zcash_client_backend/src/data_api/error.rs
@@ -61,6 +61,9 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError> {
     /// Attempted to create a spend to an unsupported pool type (currently, Orchard).
     UnsupportedPoolType(PoolType),
 
+    /// Attempted to create a spend to an unsupported Unified Address receiver
+    NoSupportedReceivers(Vec<u32>),
+
     /// A note being spent does not correspond to either the internal or external
     /// full viewing key for an account.
     NoteMismatch(NoteId),
@@ -117,7 +120,8 @@ where
             Error::ScanRequired => write!(f, "Must scan blocks first"),
             Error::Builder(e) => write!(f, "An error occurred building the transaction: {}", e),
             Error::MemoForbidden => write!(f, "It is not possible to send a memo to a transparent address."),
-            Error::UnsupportedPoolType(t) => write!(f, "Attempted to create spend to an unsupported pool type: {}", t),
+            Error::UnsupportedPoolType(t) => write!(f, "Attempted to send to an unsupported pool: {}", t),
+            Error::NoSupportedReceivers(t) => write!(f, "Unified address contained only unsupported receiver types: {:?}", &t[..]),
             Error::NoteMismatch(n) => write!(f, "A note being spent ({:?}) does not correspond to either the internal or external full viewing key for the provided spending key.", n),
 
             #[cfg(feature = "transparent-inputs")]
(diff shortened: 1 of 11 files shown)


zcash_client_backend/src/data_api/error.rs
```diff
@@ -61,6 +61,9 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError> {
     /// Attempted to create a spend to an unsupported pool type (currently, Orchard).
     UnsupportedPoolType(PoolType),
 
+    /// Attempted to create a spend to an unsupported Unified Address receiver
+    NoSupportedReceivers(Vec<u32>),
```
str4d, line 65: Document as a change in the changelog.

zcash_client_backend/src/proto.rs
```diff
@@ -222,7 +222,7 @@ pub enum ProposalDecodingError<DbError> {
     /// A transaction identifier string did not decode to a valid transaction ID.
     TxIdInvalid(TryFromSliceError),
     /// An invalid value pool identifier was encountered.
-    ValuePoolInvalid(i32),
+    ValuePoolNotSupported(i32),
```
str4d, line 225: This change doesn't need to be documented in the changelog because the enum is new in this release. However, in the PR the enum was introduced in (specifically in aeb405ef5d6fb9f18cdfaae3a9406d8798f0226a), it got renamed without the changelog entry being renamed. Fix the changelog addition entry to say `ProposalDecodingError` instead of `ProposalError`.

zcash_keys/src/address.rs
```diff
@@ -150,6 +150,11 @@ impl UnifiedAddress {
         self.transparent.as_ref()
     }
 
+    /// Returns the set of unknown receivers of the unified address.
+    pub fn unknown(&self) -> &[(u32, Vec<u8>)] {
```
str4d, line 154: Document as an addition in the changelog.

str4d (commented):
LGTM; a few changelog fixes needed.

Two questions before I ACK:
- Are you planning to do the changelog refactor (to move all the now-gated things into a separate new "Unreleased" area at the top of the changelog, so we'll just have two such areas temporarily until the crates are published) in a subsequent PR?
- This PR does not gate the new migration (that is broken at least until #1127 is reviewed and merged). Are you planning to gate this migration, or do you want to stabilise it before Orchard support?

zcash_keys/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The entries below are relative to the `zcash_client_backend` crate as of
   - `address`
   - `encoding`
   - `keys`
+- `zcash_keys::address::UnifiedAddress::unknown`:
```
str4d, line 17:
```suggestion
- `zcash_keys::address::UnifiedAddress::unknown`
```

str4d (approved):
utACK 0ae986cad00f5994dbaef94cc3d4bd1374b0e60e. We can do a pass over the changelogs during the release process to do any cleanups.

PR #1130, opened 2024-01-24
CI: Replace `--all-features` with common feature config

str4d:
Part of zcash/librustzcash#1123.


nuttycom (approved):
utACK 5187a2dafb49b4b82af7add7d309626508566a16

PR #1133, opened 2024-01-25
CI: Copy patch directives into synthetic crate

str4d:
This is a little fragile as it requires we always add patch directives at the bottom of `Cargo.toml`, but that's indeed what we do.


nuttycom (approved):
utACK c52a26c

PR #1135, opened 2024-01-25
zcash_address: Add handling for Unified Metadata Items

nuttycom:


nuttycom:
Returning to draft status to update the parsing of the HRP for ZIP 316, Revision 1

components/zcash_address/src/encoding.rs
```diff
@@ -54,34 +54,55 @@ impl FromStr for ZcashAddress {
...
                 return Err(ParseError::from(e));
             }
         }
 
-        // Try decoding as a Sapling address (Bech32)
-        if let Ok((hrp, data, Variant::Bech32)) = bech32::decode(s) {
-            // If we reached this point, the encoding is supposed to be valid Bech32.
+        // Try decoding as a Sapling or TEX address (Bech32/Bech32m)
+        if let Ok((hrp, data, variant)) = bech32::decode(s) {
+            // If we reached this point, the encoding is found to be valid Bech32.
```
daira, line 67:
```suggestion
            // If we reached this point, the encoding is found to be valid Bech32 or Bech32m.
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -39,7 +39,37 @@ pub enum Typecode {
...
+    type Error = ();
+
+    fn try_from(typecode: u32) -> Result<Self, Self::Error> {
+        match typecode {
+            0x00 => Ok(DataTypecode::P2pkh),
+            0x01 => Ok(DataTypecode::P2sh),
+            0x02 => Ok(DataTypecode::Sapling),
+            0x03 => Ok(DataTypecode::Orchard),
+            0x04..=0xDF | 0xFD..=0x02000000 => Ok(DataTypecode::Unknown(typecode)),
+            _ => Err(()),
```
daira, line 52:
```suggestion
            // 0xC0..=0xDF are metadata typecodes that do not need to be understood.
            0x04..=0xDF | 0xFD..=0x02000000 => Ok(DataTypecode::Unknown(typecode)),
            // 0xE0..=0xFC are MUST-understand metadata typecodes. Since this library
            // does not understand them, they cause an error.
            _ => Err(()),
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +99,212 @@ impl Typecode {
...
-        u32::from(*a).cmp(&u32::from(*b))
+/// The known Metadata Typecodes
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum MetadataTypecode {
+    /// Expiration height metadata as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
+    ExpiryHeight,
+    /// Expiration height metadata as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
+    ExpiryTime,
+    /// An unknown MUST-underatand metadata item as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
+    /// A parser encountering this typecode must halt with an error.
```
daira, line 112:
```suggestion
    /// An unknown MUST-understand metadata item as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
    /// A parser encountering this typecode MUST halt with an error.
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +99,212 @@ impl Typecode {
...
+            0xC0..=0xDF => Ok(MetadataTypecode::Unknown(typecode)),
+            0xE0 => Ok(MetadataTypecode::ExpiryHeight),
+            0xE1 => Ok(MetadataTypecode::ExpiryTime),
+            0xE2..=0xFC => Ok(MetadataTypecode::MustUnderstand(typecode)),
+            _ => Err(()),
+        }
+    }
+}
+
+impl From<MetadataTypecode> for u32 {
```
daira, line 132:
```suggestion
impl From<MetadataTypecode> for u64 {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +99,212 @@ impl Typecode {
...
+    /// A parser encountering this typecode must halt with an error.
+    MustUnderstand(u32),
+    /// An unknown metadata item as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
+    Unknown(u32),
+}
+
+impl TryFrom<u32> for MetadataTypecode {
+    type Error = ();
+
+    fn try_from(typecode: u32) -> Result<Self, Self::Error> {
```
daira, line 121:
```suggestion
impl TryFrom<u64> for MetadataTypecode {
    type Error = ();

    fn try_from(typecode: u64) -> Result<Self, Self::Error> {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -39,7 +39,37 @@ pub enum Typecode {
...
+            0x01 => Ok(DataTypecode::P2sh),
+            0x02 => Ok(DataTypecode::Sapling),
+            0x03 => Ok(DataTypecode::Orchard),
+            0x04..=0xDF | 0xFD..=0x02000000 => Ok(DataTypecode::Unknown(typecode)),
+            _ => Err(()),
+        }
+    }
+}
+
+impl From<DataTypecode> for u32 {
```
daira, line 57:
```suggestion
impl From<DataTypecode> for u64 {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -39,7 +39,37 @@ pub enum Typecode {
...
+impl TryFrom<u32> for DataTypecode {
+    type Error = ();
+
+    fn try_from(typecode: u32) -> Result<Self, Self::Error> {
+        match typecode {
+            0x00 => Ok(DataTypecode::P2pkh),
+            0x01 => Ok(DataTypecode::P2sh),
+            0x02 => Ok(DataTypecode::Sapling),
+            0x03 => Ok(DataTypecode::Orchard),
+            0x04..=0xDF | 0xFD..=0x02000000 => Ok(DataTypecode::Unknown(typecode)),
```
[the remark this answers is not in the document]
daira: See also https://github.com/zcash/zips/issues/787 .

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +99,212 @@ impl Typecode {
...
+                data: data.to_vec(),
+            }),
+            (Revision::R1, MetadataTypecode::MustUnderstand(tc)) => {
+                Err(ParseError::NotUnderstood(tc))
+            }
+            (Revision::R0, _) => {
+                // The entire R0 metadata range has now been reserved for MUST-understand metadata
+                // in R1, so no R0 metadata range now exists.
+                Err(ParseError::NotUnderstood(typecode.into()))
+            }
```
daira, line 250: We should just pretend that 0xC0..=0xDF was always defined as a metadata range. It's conformant because there isn't any difference in behaviour between data and metadata items in this range.

components/zcash_address/src/kind/unified.rs
```diff
@@ -370,9 +606,12 @@ pub trait Encoding: private::SealedContainer {
...
     /// * the item list may not contain only transparent items (or no items)
     /// * the item list may not contain both P2PKH and P2SH items.
-    fn try_from_items(mut items: Vec<Self::Item>) -> Result<Self, ParseError> {
-        items.sort_unstable_by(Self::Item::encoding_order);
-        Self::try_from_items_internal(items)
+    fn try_from_items(
+        revision: Revision,
+        mut items: Vec<Item<Self::DataItem>>,
+    ) -> Result<Self, ParseError> {
+        items.sort_unstable_by(Item::encoding_order);
```
daira, line 613: The documentation says:
```rust
    /// Constructs a value of a unified container type from a vector
    /// of container items, sorted according to typecode as specified
    /// in ZIP 316.
```
but the "vector of container items" need not be pre-sorted. Please remove that clause from the doc.

components/zcash_address/src/kind/unified.rs
```diff
@@ -411,20 +650,15 @@ pub trait Encoding: private::SealedContainer {
...
 
     /// Returns the items in the order they were parsed from the string encoding.
     ///
     /// This API is for advanced usage; in most cases you should use `Self::items`.
-    fn items_as_parsed(&self) -> &[Self::Item];
+    fn items_as_parsed(&self) -> &[Item<Self::DataItem>];
+
+    /// Returns the revision of the ZIP 316 standard that this unified container
+    /// conforms to.
+    fn revision(&self) -> Revision;
```
daira, line 663: This trait doesn't seem to have any way to get the metadata items.

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -94,51 +90,75 @@ impl SealedItem for Fvk {
 /// // We can obtain the pool-specific Full Viewing Keys for the UFVK in preference
 /// // order (the order in which wallets should prefer to use their corresponding
 /// // address receivers):
```
daira, line 92: This comment is no longer correct for `ufvk.items_as_parsed()`.

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -340,36 +364,8 @@ mod tests {
...
 
         assert_eq!(
-            Ufvk::parse_internal(Ufvk::MAINNET, &encoded[..]),
+            Ufvk::parse_internal(Ufvk::MAINNET_R0, &encoded[..]),
             Err(ParseError::OnlyTransparent)
         );
     }
-
-    #[test]
-    fn fvks_are_sorted() {
```
daira, line 349: Why is this test removed?

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -99,51 +95,75 @@ impl SealedItem for Ivk {
 /// // We can obtain the pool-specific Incoming Viewing Keys for the UIVK in
 /// // preference order (the order in which wallets should prefer to use their
 /// // corresponding address receivers):
```
daira, line 97: This comment is no longer correct for `uivk.items_as_parsed()`.

components/zcash_address/src/kind/unified.rs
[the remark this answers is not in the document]
nuttycom: Ah, yeah, I missed this spot.

zcash_keys/src/address.rs
```diff
@@ -124,9 +166,33 @@ impl UnifiedAddress {
...
+    pub fn unknown_data(&self) -> &[(u32, Vec<u8>)] {
+        self.unknown_data.as_ref()
+    }
+
+    /// Returns the expiration height for this address.
+    pub fn expiry_height(&self) -> Option<BlockHeight> {
+        self.expiry_height
+    }
+
+    /// Returns the expiration time for this address.
```
daira, line 179:
```suggestion
    /// Returns the expiration time for this address as a Unix timestamp.
```

daira, line 179:
```suggestion
    /// Returns the expiration time for this address as a Unix Epoch Time.
```

zcash_keys/src/address.rs
```diff
@@ -124,9 +166,33 @@ impl UnifiedAddress {
...
+    pub fn expiry_time(&self) -> Option<u64> {
+        self.expiry_time
+    }
+
+    /// Returns any unknown metadata items parsed from the encoded form of the address.
+    ///
+    /// Unknown metadata items are guaranteed by construction and parsing to not have keys in the
+    /// `must-understand` metadata typecode range.
```
daira, line 187:
```suggestion
    /// MUST-understand metadata typecode range.
```

zcash_keys/src/address.rs
```diff
@@ -124,9 +166,33 @@ impl UnifiedAddress {
...
+    pub fn unknown_metadata(&self) -> &[(u32, Vec<u8>)] {
+        self.unknown_metadata.as_ref()
+    }
+
+    /// Returns the string encoding of this `UnifiedAddress` for the given network.
+    pub fn encode<P: consensus::Parameters>(&self, params: &P) -> String {
+        self.to_address(params.address_network().expect("Unrecognized network"))
```
daira, line 194:
```suggestion
        self.to_address(params.address_network().expect("network should be recognized"))
```

zcash_keys/src/address.rs
```diff
@@ -140,59 +206,77 @@ impl UnifiedAddress {
         let orchard_receiver = None;
 
         let ua = unified::Address::try_from_items(
```
daira, line 208: Does `try_from_items` do the sorting?

zcash_keys/src/keys.rs
```diff
@@ -422,39 +429,53 @@ pub enum AddressGenerationError {
...
-    pub const fn unsafe_new(has_orchard: bool, has_sapling: bool, has_p2pkh: bool) -> Self {
-        if !(has_orchard || has_sapling) {
+    /// Panics: at least one of `has_orchard`, `has_sapling`, or `has_p2pkh` must be `true`.
+    pub const fn unsafe_new_without_expiry(
+        has_orchard: bool,
+        has_sapling: bool,
+        has_p2pkh: bool,
+    ) -> Self {
+        if !(has_orchard || has_sapling || has_p2pkh) {
             panic!("At least one shielded receiver must be requested.")
```
daira, line 470: This message is no longer accurate.

daira (changes requested):
Some changes needed &mdash; see the "(blocking)" comments.

components/zcash_address/src/kind/unified.rs
nuttycom: `items_as_parsed` returns a `Vec<Item<DataItem>>`. The type of data items varies between addresses, IVKs and FVKs, but the type of metadata items does not vary; `Item` is an enum that encompasses both data and metadata items.

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +99,212 @@ impl Typecode {
...
-            0x03 => Ok(Typecode::Orchard),
-            0x04..=0x02000000 => Ok(Typecode::Unknown(typecode)),
-            0x02000001..=u32::MAX => Err(ParseError::InvalidTypecodeValue(typecode as u64)),
-        }
+        DataTypecode::try_from(typecode)
+            .map_or_else(
+                |()| MetadataTypecode::try_from(typecode).map(Typecode::Metadata),
+                |t| Ok(Typecode::Data(t)),
+            )
+            .map_err(|()| ParseError::InvalidTypecodeValue(typecode as u64))
```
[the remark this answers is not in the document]
nuttycom: Using the constant is a good idea, but as to the other part it is already a parsing error: https://github.com/zcash/librustzcash/blob/main/components/zcash_address/src/kind/unified.rs#L264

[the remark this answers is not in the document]
nuttycom: :heavy_check_mark:

components/zcash_address/src/kind/unified.rs
nuttycom: :heavy_check_mark:

components/zcash_address/src/kind/unified.rs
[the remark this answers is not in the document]
nuttycom: :heavy_check_mark:

components/zcash_address/src/kind/unified.rs
nuttycom: :heavy_check_mark:

components/zcash_address/src/kind/unified/fvk.rs
nuttycom: :heavy_check_mark:

components/zcash_address/src/kind/unified/fvk.rs
nuttycom: The `items` function no longer exists.

components/zcash_address/src/kind/unified/ivk.rs
nuttycom: :heavy_check_mark:

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -322,36 +346,8 @@ mod tests {
...
 
         assert_eq!(
-            Uivk::parse_internal(Uivk::MAINNET, &encoded[..]),
+            Uivk::parse_internal(Uivk::MAINNET_R0, &encoded[..]),
             Err(ParseError::OnlyTransparent)
         );
     }
-
-    #[test]
-    fn ivks_are_sorted() {
```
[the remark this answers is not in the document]
nuttycom: The `items` method has been removed; preference order is only significant when considering unified address receivers, and users should prefer `Address::receivers` for this use case.

zcash_keys/src/address.rs
nuttycom: Yes.

components/zcash_address/src/kind/unified.rs
nuttycom: The one reason that it might make sense to keep this doc is that this _does_ sort, and so if you pass a different order of items, `items_as_parsed` will return the sorted order.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/7bace95cabb48f03ba864489d220cb757bdfefb9..3742501615b8729db64848539e0ccd9cd76cca2d) to fix conflicts with `main`.

components/zcash_address/src/encoding.rs
```diff
@@ -269,6 +313,40 @@ mod tests {
...
     }
 
+    #[test]
+    fn tex() {
+        let p2pkh_str = "t1VmmGiyjVNeCjxDZzg7vZmd99WyzVby9yC";
+        let tex_str = "tex1s2rt77ggv6q989lr49rkgzmh5slsksa9khdgte";
+
+        // Transcode P2PKH to TEX
+        let p2pkh_zaddr: ZcashAddress = p2pkh_str.parse().unwrap();
+        assert_matches!(p2pkh_zaddr.net, Network::Main);
```
daira, line 323: This only tests mainnet TEX addresses; add a test for testnet.

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +100,214 @@ impl Typecode {
...
+            (Revision::R1, MetadataTypecode::Unknown(typecode)) => Ok(MetadataItem::Unknown {
+                typecode,
+                data: data.to_vec(),
+            }),
+            (Revision::R1, MetadataTypecode::MustUnderstand(tc)) => {
+                Err(ParseError::NotUnderstood(tc))
+            }
+            (Revision::R0, _) => {
+                // The entire R0 metadata range has now been reserved for MUST-understand metadata
+                // in R1, so no R0 metadata range now exists.
```
daira, line 251: I disagree: on [line 126](https://github.com/zcash/librustzcash/pull/1135/files#diff-55c7c102f533ffc792f55a3517e036e5e41f08c1ee669b688a91d9ac90162b6fR126) we parse 0xC0..=0xDF as a metadata typecode, independent of revision. This range should be allowed without error, because either possible interpretation of the spec (that these are unknown data typecodes, or that they are retrospectively changed to unknown and *not* MUST-understand metadata typecodes) would lead to allowing them without error.

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +100,214 @@ impl Typecode {
...
+            (Revision::R1, MetadataTypecode::ExpiryTime) => data
+                .try_into()
+                .map(u64::from_le_bytes)
+                .map(MetadataItem::ExpiryTime)
+                .map_err(|_| {
+                    ParseError::InvalidEncoding(
+                        "Expiry time must be a 64-bit little-endian value.".to_string(),
+                    )
+                }),
+            (Revision::R1, MetadataTypecode::Unknown(typecode)) => Ok(MetadataItem::Unknown {
```
daira, line 242:
```suggestion
            (_, MetadataTypecode::Unknown(typecode)) => Ok(MetadataItem::Unknown {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +100,214 @@ impl Typecode {
...
+                typecode,
+                data: data.to_vec(),
+            }),
+            (Revision::R1, MetadataTypecode::MustUnderstand(tc)) => {
```
daira, line 246:
```suggestion
            (_, MetadataTypecode::MustUnderstand(tc)) => {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -332,47 +568,54 @@ pub(crate) mod private {
                     return Err(ParseError::InvalidTypecodeOrder);
                 } else if t_code == prev_code {
                     return Err(ParseError::DuplicateTypecode(t));
-                } else if t == Typecode::P2sh && prev_code == Some(u32::from(Typecode::P2pkh)) {
+                } else if t == Typecode::Data(DataTypecode::P2sh)
+                    && prev_code == Some(u32::from(DataTypecode::P2pkh))
+                {
```
daira, line 573:
```suggestion
                } else if t == Typecode::P2SH && prev_code == Some(u32::from(DataTypecode::P2pkh)) {
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -332,47 +568,54 @@ pub(crate) mod private {
...
+    /// Constructs a value of a unified container type from a vector of container
+    /// items. These items will be sorted according to typecode as specified in ZIP
+    /// 316, so this method is not necessarily round-trip compatible with
+    /// [`Container::items_as_parsed`].
     ///
     /// This function will return an error in the case that the following ZIP 316
     /// invariants concerning the composition of a unified container are
     /// violated:
     /// * the item list may not contain two items having the same typecode
     /// * the item list may not contain only transparent items (or no items)
```
daira, line 611: I thought we had decided to drop the prohibition on an R1 UA/UVK containing only transparent items? The motivation for doing that holds even though we are doing TEX addresses, because it is possible for an R1 UA/UVK to include metadata for a transparent address that is inexpressible using the bare t-addr/viewing key formats.

daira (dismissed):
utACK modulo comments.

components/zcash_address/src/kind/unified.rs
nuttycom: I have a slight preference to keep the match expressions explicit, because that's more resilient to future changes to the set of revisions (it will cause a compilation failure, rather than silently add default behavior for the new revision.)

components/zcash_address/src/kind/unified.rs
nuttycom: Ah, yeah, this doc is now wrong.

components/zcash_address/src/encoding.rs
nuttycom: We don't yet have test vectors or tests for _any_ testnet addresses, and while I agree that such tests would be useful we should probably approach this systematically. I'll add a test here, but it won't be robust; we should actually do proper test vectors for TEX addresses.

components/zcash_address/src/encoding.rs
```diff
@@ -54,34 +54,55 @@ impl FromStr for ZcashAddress {
...
+                        .map(AddressKind::Sapling)
+                        .map_err(|_| ParseError::InvalidEncoding)
+                        .map(|kind| ZcashAddress { net, kind });
                 }
-            };
+                Variant::Bech32m => {
+                    // Try decoding as a TEX address (Bech32m)
+                    let net = match hrp.as_str() {
+                        tex::MAINNET => Network::Main,
+                        tex::TESTNET => Network::Test,
```
daira, line 92:
```suggestion
                        tex::TESTNET => Network::Test,
                        tex::REGTEST => Network::Regtest,
```

components/zcash_address/src/kind/tex.rs
```diff
@@ -0,0 +1,7 @@
+/// The prefix for a Base58Check-encoded mainnet transparent
```
daira, line 1:
```suggestion
/// The prefix for a Bech32m-encoded mainnet transparent
```

components/zcash_address/src/kind/tex.rs
```diff
@@ -0,0 +1,7 @@
...
+/// [ZIP 320](https://zips.z.cash/zip-0320.html) TEX address.
+pub(crate) const MAINNET: &str = "tex";
+
+/// The prefix for a Base58Check-encoded testnet transparent
```
daira, line 5:
```suggestion
/// The prefix for a Bech32m-encoded testnet transparent
```

components/zcash_address/src/kind/tex.rs
```diff
@@ -0,0 +1,7 @@
...
+/// [ZIP 320](https://zips.z.cash/zip-0320.html) TEX address.
+pub(crate) const TESTNET: &str = "textest";
```
daira, line 7:
```suggestion
pub(crate) const TESTNET: &str = "textest";

/// The prefix for a Bech32m-encoded regtest transparent
/// [ZIP 320](https://zips.z.cash/zip-0320.html) TEX address.
pub(crate) const REGTEST: &str = "texregtest";
```

components/zcash_address/src/kind/unified.rs
```diff
@@ -279,12 +505,15 @@ pub(crate) mod private {
                         length
                     )));
                 }
-                let result = R::try_from((
-                    typecode,
-                    &buf[cursor.position() as usize..addr_end as usize],
-                ));
+                let data = &buf[cursor.position() as usize..addr_end as usize];
```
daira, line 508:
```suggestion
                // The "as usize" casts cannot change the values, because both
                // cursor.position() and addr_end are unsigned values <= buf.len()
                // which is usize.
                let data = &buf[cursor.position() as usize..addr_end as usize];
```

daira (dismissed):
utACK with comments.

components/zcash_address/CHANGELOG.md
```diff
@@ -7,6 +7,30 @@ and this library adheres to Rust's notion of
...
+### Changed
+- `zcash_address::unified`:
+  - `Typecode` has changed. Instead of having a variant for each receiver type,
+    it now has two variants, `Typecode::Data` and `Typecode::Metadata`.
+  - `Encoding::try_from_items` now takes an additional `Revision` argument.
+
+### Removed
+- `zcash_address::unified::Container::items` Preference order is only 
+  significant when considering unified address receivers; use
+  `Address::receivers` instead.
```
str4d, line 32: These changes prevent us from introducing TEX addresses in a point release, as they are not backwards-compatible. I would prefer that we separate out the TEX address parts from the ZIP 316 rev1 parts; the former can be deployed as 0.3.2 as intended (#1221), while the latter are required (for this current implementation) to be in 0.4.0.

str4d (changes requested):

components/zcash_address/CHANGELOG.md
str4d: I confirmed locally that the only conflict between 35ea09a16cce24b11c64249020e1ce9f8f557027 and the rest of the PR is in the changelog. Please extract that commit into a separate PR.

nuttycom: @str4d does this really matter? It's extra effort to perform the extraction, and I don't see the benefit.

daira (approved):
utACK 4992ff765f68871343b1ad8c25363486e78a0264

components/zcash_address/CHANGELOG.md
daira: I agree with @nuttycom. A breaking update of `zcash_address` is not likely to impose much burden on its current users.

str4d: I did the extraction locally before making my second comment; it was trivial (with a minor adjustment afterwards because @daira had suggested TEX changes here, that would have been instead made on #1226 had it been opened as a separate PR initially).

> A breaking update of zcash_address is not likely to impose much burden on its current users.

The entire point of the `zcash_address` crate's approach, that TEX addresses are being added to, is to ensure that even if the direct users of `zcash_address` don't perform the breaking change upgrade, _their_ downstream users (i.e. people using Zcash) benefit as a side-effect of the end binary app's `cargo update` process. It's never been about the direct users of `zcash_address`, but about improving the UX for end users: the user learns "hey, this address is unsupported, I should ask APP_NAME to upgrade", not "oh, this address is invalid, I must have done the wrong thing".

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/4992ff765f68871343b1ad8c25363486e78a0264..a509b72a7f1a6d7487e0419bedbf92930728be27) to fix merge conflicts after #1142 was merged and #1226 was extracted.

zcash_keys/src/keys.rs
```diff
@@ -498,22 +505,28 @@ pub struct UnifiedAddressRequest {
...
             None
         } else {
             Some(Self {
                 has_orchard,
                 has_sapling,
                 has_p2pkh,
+                expiry_height,
+                expiry_time,
             })
         }
```
daira, line 531:
```suggestion
        if has_sapling || has_orchard || (has_p2pkh && (expiry_height.is_some() || expiry_time.is_some())) {
            Some(Self {
                has_orchard,
                has_sapling,
                has_p2pkh,
                expiry_height,
                expiry_time,
            })
        } else {
            None
        }
```

zcash_keys/src/keys.rs
```diff
@@ -533,21 +546,27 @@ impl UnifiedAddressRequest {
...
+        Self::new(_has_orchard, _has_sapling, _has_p2pkh, None, None)
     }
 
     /// Construct a new unified address request from its constituent parts.
     ///
-    /// Panics: at least one of `has_orchard` or `has_sapling` must be `true`.
-    pub const fn unsafe_new(has_orchard: bool, has_sapling: bool, has_p2pkh: bool) -> Self {
-        if !(has_orchard || has_sapling) {
-            panic!("At least one shielded receiver must be requested.")
+    /// Panics: at least one of `has_orchard`, `has_sapling`, or `has_p2pkh` must be `true`.
```
daira, line 554:
```suggestion
    /// Panics: at least one of `has_orchard` or `has_sapling` must be `true`.
```

zcash_keys/src/address.rs
```diff
@@ -96,11 +110,12 @@ impl UnifiedAddress {
     ///
     /// Returns `None` if the receivers would produce an invalid Unified Address (namely,
     /// if no shielded receiver is provided).
```
daira, line 112:
```suggestion
    /// Returns `None` if the receivers would produce an invalid Unified Address (namely,
    /// if no receiver is provided, or if there is only a transparent receiver and no expiry
    /// height or time).
```

zcash_keys/src/address.rs
```diff
@@ -119,7 +134,10 @@ impl UnifiedAddress {
                 #[cfg(feature = "sapling")]
                 sapling,
                 transparent,
-                unknown: vec![],
+                unknown_data: vec![],
+                expiry_height,
+                expiry_time,
+                unknown_metadata: vec![],
             })
```
daira, line 141: Change this to allow creating a UA with only a transparent receiver plus expiration metadata, consistent with the suggested doc change.

zcash_keys/src/keys.rs
```diff
@@ -533,21 +546,27 @@ impl UnifiedAddressRequest {
...
+    pub const fn unsafe_new_without_expiry(
+        has_orchard: bool,
+        has_sapling: bool,
+        has_p2pkh: bool,
+    ) -> Self {
+        if !(has_orchard || has_sapling || has_p2pkh) {
+            panic!("At least one receiver must be requested.")
```
daira, line 561:
```suggestion
        if !(has_orchard || has_sapling) {
            panic!("At least one shielded receiver must be requested.")
```
We should still disallow creation of UAs that don't have a shielded receiver or metadata. As it's currently implemented, passing `(false, false, true)` will create a `UnifiedAddressRequest` that can't be satisfied, I think.

zcash_keys/src/keys.rs
```diff
@@ -533,21 +546,27 @@ impl UnifiedAddressRequest {
...
         }
 
         Self {
             has_orchard,
             has_sapling,
             has_p2pkh,
+            expiry_height: None,
+            expiry_time: None,
         }
```
daira, line 570:
```suggestion
        Self::unsafe_new(has_orchard, has_sapling, has_p2pkh, None, None)
```
This ensures that the validity checking is consistent between `unsafe_new` and `unsafe_new_without_expiry`, rather than duplicating it.

zcash_keys/src/keys.rs
```diff
@@ -807,18 +884,20 @@ impl UnifiedFullViewingKey {
...
+        UnifiedAddress::new(
             #[cfg(feature = "orchard")]
             orchard,
             #[cfg(feature = "sapling")]
             sapling,
             transparent,
+            std::cmp::min(self.expiry_height, request.expiry_height),
+            std::cmp::min(self.expiry_time, request.expiry_time),
         )
         .ok_or(AddressGenerationError::ShieldedReceiverRequired)
```
daira, line 902: We should make sure that an invalid combination isn't representable as a `UnifiedAddressRequest`, so that this error is impossible (and can be a panic if it does occur due to some inconsistency).

zcash_keys/src/keys.rs
```diff
@@ -619,57 +644,62 @@ impl UnifiedFullViewingKey {
         let mut sapling = None;
         #[cfg(feature = "transparent-inputs")]
         let mut transparent = None;
+        let mut unknown_data = vec![];
+        let mut expiry_height = None;
+        let mut expiry_time = None;
+        let mut unknown_metadata = vec![];
```
daira, line 650: It's unclear whether `unified::Ufvk::decode(encoding)` enforces the requirement on the allowed receivers and metadata.

zcash_keys/src/keys.rs
```diff
@@ -678,42 +708,69 @@ impl UnifiedFullViewingKey {
             sapling,
             #[cfg(feature = "orchard")]
             orchard,
-            unknown,
+            unknown_data,
+            expiry_height,
+            expiry_time,
+            unknown_metadata,
         })
```
daira, line 715: This does not check the requirement on the allowed receivers and metadata. It may be that an invalid combination can't occur here because of checks in `unified::Ufvk::decode(encoding)`, but if so it's not obvious.

zcash_keys/src/keys.rs
```diff
@@ -678,42 +708,69 @@ impl UnifiedFullViewingKey {
...
+                self.expiry_height
+                    .map(|h| unified::MetadataItem::ExpiryHeight(u32::from(h))),
+            )
+            .chain(self.expiry_time.map(unified::MetadataItem::ExpiryTime));
+
+        let ufvk = unified::Ufvk::try_from_items(
+            if self.expiry_height().is_some() || self.expiry_time().is_some() {
+                Revision::R1
+            } else {
+                Revision::R0
```
daira, line 766: This causes the construction to fail for a transparent-only UFVK without expiry height or time. That's fine, but we should enforce this policy consistently.

zcash_keys/src/address.rs
```diff
@@ -168,43 +186,84 @@ impl UnifiedAddress {
...
+                self.expiry_height
+                    .map(|h| unified::MetadataItem::ExpiryHeight(u32::from(h))),
+            )
+            .chain(self.expiry_time.map(unified::MetadataItem::ExpiryTime));
+
+        let ua = unified::Address::try_from_items(
+            if self.expiry_height().is_some() || self.expiry_time().is_some() {
+                Revision::R1
+            } else {
+                Revision::R0
```
daira, line 259: This causes the construction to fail for a transparent-only UA without expiry height or time. That's fine, but we should enforce this policy consistently.

daira (changes requested):
Changes requested.

zcash_keys/src/keys.rs
nuttycom: I didn't think that we were _requiring_ must-understand metadata to accompany p2pkh addresses? I don't see that anywhere in the current text of ZIP 316; https://zips.z.cash/zip-0316#requirements-for-both-unified-addresses-and-unified-viewing-keys just says that the "at least one shielded receiver" requirement is dropped.

zcash_keys/src/keys.rs
nuttycom: Also depends upon the resolution to https://github.com/zcash/librustzcash/pull/1135#discussion_r1515444301

zcash_keys/src/keys.rs
nuttycom: This is certainly the wrong error to return at this point, but what the correct error is depends upon https://github.com/zcash/librustzcash/pull/1135#discussion_r1515444301

zcash_keys/src/keys.rs
nuttycom: The decoding of Unified containers currently disallows transparent-only UAs; this needs to change but will depend upon the resolution of https://github.com/zcash/librustzcash/pull/1135#discussion_r1515444301.

zcash_keys/src/keys.rs
daira: I thought we decided to make the F4Jumble input size 40 bytes so that the original rationale for excluding bare P2PKH addresses as UAs — that they can be more compatibly expressed as t-addresses — would be maintained? 

https://github.com/zcash/zips/pull/797/files#diff-8ff1060cb8fa47eb429fb6c849f1364e28fd4b42868ae8f56dd20ea8b9028b2c

zcash_keys/src/keys.rs
nuttycom: Fixed in 078b4ffa4577c8cb0e1a325b6a1b2b512c20a18a

zcash_keys/src/address.rs
nuttycom: Fixed in 078b4ffa4577c8cb0e1a325b6a1b2b512c20a18a

components/zcash_address/src/kind/unified.rs
```diff
@@ -69,51 +100,213 @@ impl Typecode {
...
+}
 
-    pub fn encoding_order(a: &Self, b: &Self) -> cmp::Ordering {
-        u32::from(*a).cmp(&u32::from(*b))
+/// The known Metadata Typecodes
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum MetadataTypecode {
+    /// Expiration height metadata as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
+    ExpiryHeight,
+    /// Expiration height metadata as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
```
daira, line 110:
```suggestion
    /// Expiration time metadata as specified in [ZIP 316, Revision 1](https://zips.z.cash/zip-0316)
```

zcash_keys/src/keys.rs
nuttycom: Addressed in 673e1a21d28209103dce50ede358491e5978d007

zcash_keys/src/keys.rs
nuttycom: Resolved in 673e1a21d28209103dce50ede358491e5978d007

zcash_keys/src/keys.rs
nuttycom: Resolved in 673e1a21d28209103dce50ede358491e5978d007

zcash_keys/src/keys.rs
nuttycom: Addressed in 673e1a21d28209103dce50ede358491e5978d007, but I hope that it's clear how much less maintainable and error-prone this approach is.

zcash_keys/src/keys.rs
nuttycom: Resolved in 673e1a21d28209103dce50ede358491e5978d007

zcash_keys/src/keys.rs
nuttycom: 673e1a21d28209103dce50ede358491e5978d007

components/zcash_address/src/kind/unified.rs
```diff
@@ -538,12 +580,13 @@ pub(crate) mod private {
...
                 // All checks pass!
-                Ok(Self::from_inner(items))
+                Ok(Self::from_inner(revision, items))
             }
         }
 
         fn parse_internal<T: Into<Vec<u8>>>(hrp: &str, buf: T) -> Result<Self, ParseError> {
-            Self::parse_items(hrp, buf).and_then(Self::try_from_items_internal)
+            Self::parse_items(hrp, buf)
+                .and_then(|(revision, items)| Self::try_from_items_internal(revision, items))
```
str4d, line 587: I _think_ this line should work fine as it was before?

components/zcash_address/CHANGELOG.md
```diff
@@ -10,10 +10,18 @@ and this library adheres to Rust's notion of
 - `impl serde::{Serialize, Deserialize} for zcash_address::ZcashAddress` behind
   the `serde` feature flag.
 
+### Added
+- `zcash_address::unified`:
+  - `Container::revision`
```
str4d, line 15: This is a change to the `Container` trait, not an addition. `Container` is implementable by anyone outside the crate (as long as they reuse an existing `Container::DataItem` as that is not externally implementable due to the sealed trait bound).

components/zcash_address/src/kind/unified/address.rs
```diff
@@ -267,9 +294,12 @@ pub mod testing {
         arb_typecodes()
             .prop_flat_map(arb_unified_address_receivers)
             .prop_map(|rs| {
-                let mut items = rs.into_iter().map(Item::Data).collect::<Vec<_>>();
-                items.sort_unstable_by(Item::encoding_order);
-                Address(items)
+                let mut receivers = rs.into_iter().map(Item::Data).collect::<Vec<_>>();
+                receivers.sort_unstable_by(Item::encoding_order);
+                Address {
+                    revision: Revision::R0,
```
str4d, line 300: Given that this strategy doesn't generate metadata, it should be fine to generate an arbitrary revision?

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -199,9 +223,9 @@ mod tests {
             shielded in arb_shielded_fvk(),
             transparent in prop::option::of(arb_transparent_fvk()),
         ) -> Ufvk {
-            let mut items: Vec<_> = transparent.into_iter().chain(shielded).map(Item::Data).collect();
-            items.sort_unstable_by(Item::encoding_order);
-            Ufvk(items)
+            let mut fvks: Vec<_> = transparent.into_iter().chain(shielded).map(Item::Data).collect();
+            fvks.sort_unstable_by(Item::encoding_order);
+            Ufvk { revision: Revision::R0, fvks }
```
str4d, line 228: Ditto; this can be an arbitrary revision?

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -192,9 +216,9 @@ mod tests {
             shielded in arb_shielded_ivk(),
             transparent in prop::option::of(arb_transparent_ivk()),
         ) -> Uivk {
-            let mut items: Vec<_> = transparent.into_iter().chain(shielded).map(Item::Data).collect();
-            items.sort_unstable_by(Item::encoding_order);
-            Uivk(items)
+            let mut ivks: Vec<_> = transparent.into_iter().chain(shielded).map(Item::Data).collect();
+            ivks.sort_unstable_by(Item::encoding_order);
+            Uivk { revision: Revision::R0, ivks }
```
str4d, line 221: Ditto.

zcash_keys/CHANGELOG.md
```diff
@@ -22,6 +25,9 @@ and this library adheres to Rust's notion of
 - Migrated to `zcash_primitives 0.20.0`
 - MSRV is now 1.77.0.
 
+### Removed
+- `zcash_keys::keys::DerivationError` use `UnifiedKeyError` instead.
```
str4d, line 29:
```suggestion
- `zcash_keys::keys::DerivationError` (use `UnifiedKeyError` instead).
```

zcash_keys/src/keys.rs
```diff
@@ -91,30 +91,40 @@ fn to_transparent_child_index(j: DiversifierIndex) -> Option<NonHardenedChildInd
     }
 }
 
+/// Errors that can occur in the generation of Unified Spending Keys, Unified Viewing Keys, or
+/// Unified Addresses.
 #[derive(Debug)]
-pub enum DerivationError {
+pub enum UnifiedKeyError {
+    /// At least
```
str4d, line 98: Unfinished.

zcash_keys/src/keys.rs
```diff
@@ -147,8 +157,11 @@ pub enum DecodingError {
     LengthMismatch(Typecode, u32),
     #[cfg(feature = "unstable")]
     InsufficientData(Typecode),
-    /// The key data could not be decoded from its string representation to a valid key.
+    /// The key data for the given key type could not be decoded from its string representation to
+    /// a valid key.
     KeyDataInvalid(Typecode),
+    /// Decoding resulted in a value that would violate validity constraints.
+    ConstraintViolation(String),
```
str4d, line 164: Document this change to `DecodingError` variants in the changelog.

zcash_keys/src/keys.rs
```diff
@@ -1105,6 +1161,50 @@ impl UnifiedIncomingViewingKey {
         }
     }
 
+    fn from_checked_parts(
```
str4d, line 1164: `UnifiedIncomingViewingKey::new` should probably be altered to use this method, as currently (when `test-dependencies` is enabled`) it enables producing invalid states.

components/zcash_address/src/kind/unified/address.rs
```diff
@@ -380,13 +410,16 @@ mod tests {
...
+            revision: Revision::R0,
+            receivers: vec![
+                Item::Data(Receiver::Sapling([1; 43])),
+                Item::Data(Receiver::Sapling([2; 43])),
+            ],
+        };
+        let encoded = ua.to_jumbled_bytes(Address::MAINNET_R0);
         assert_eq!(
-            Address::parse_internal(Address::MAINNET, &encoded[..]),
+            Address::parse_internal(Address::MAINNET_R0, &encoded[..]),
```
str4d, line 422: (Just picking this test as an example)

I'm a little unhappy about the lack of Revision 1 tests. I think we'll be fine for now (particularly if my earlier suggestion to generate arbitrary revisions for proptests is accepted), but please open an issue for expanding testing to cover revision 1.

str4d (changes requested):
Reviewed 7852a319e69e4c291ea63e7c19b66d48f905b398.

zcash_keys/src/address.rs
```diff
@@ -168,43 +191,99 @@ impl UnifiedAddress {
...
+        self.unknown_data.as_ref()
+    }
+
+    /// Returns the expiration height for this address.
+    pub fn expiry_height(&self) -> Option<BlockHeight> {
+        self.expiry_height
+    }
+
+    /// Sets the expiry height of this address.
+    pub fn set_expiry_height(&mut self, height: BlockHeight) {
```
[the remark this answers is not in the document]
nuttycom: I will also add similar methods to UFVK/UIVK, as I noticed that they're missing.

zcash_keys/src/keys.rs
```diff
@@ -959,7 +1060,10 @@ impl UnifiedIncomingViewingKey {
         >,
         #[cfg(feature = "sapling")] sapling: Option<::sapling::zip32::IncomingViewingKey>,
         #[cfg(feature = "orchard")] orchard: Option<orchard::keys::IncomingViewingKey>,
-        // TODO: Implement construction of UIVKs with metadata items.
+        unknown_data: Vec<(u32, Vec<u8>)>,
+        expiry_height: Option<BlockHeight>,
+        expiry_time: Option<u64>,
+        unknown_metadata: Vec<(u32, Vec<u8>)>,
```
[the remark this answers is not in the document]
nuttycom: This is test-only code; it's intended for use in constructing test values with unknown metadata without having to go through serialization.

zcash_keys/src/keys.rs
```diff
@@ -1238,6 +1398,8 @@ impl UnifiedIncomingViewingKey {
                 Err(AddressGenerationError::InvalidSaplingDiversifierIndex(_)) => {
                     if j.increment().is_err() {
                         return Err(AddressGenerationError::DiversifierSpaceExhausted);
+                    } else {
+                        continue;
```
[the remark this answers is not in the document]
nuttycom: Just easier to read and reason about given the early returns.

zcash_client_backend/CHANGELOG.md
```diff
@@ -440,6 +440,9 @@ funds to those addresses. See [ZIP 320](https://zips.z.cash/zip-0320) for detail
   - Arguments to `ChangeStrategy::compute_balance` have changed.
   - `ChangeError::DustInputs` now has an `orchard` field behind the `orchard`
     feature flag.
+- `zcash_client_backend::wallet`:
+  - The address variants of `Recipient` now `Box` their contents to avoid large
+    discrepancies in enum variant sizing.
```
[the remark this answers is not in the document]
nuttycom: Aha, this is because the changes to use `ZcashAddress` were written after, this PR, but then merged before it.

zcash_keys/src/keys.rs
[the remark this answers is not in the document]
nuttycom: This is test-only code; it's intended for use in generating UAs that have unknown data or metadata for testing without having to go through an encoding.

components/zcash_address/src/kind/unified.rs
nuttycom: A tuple can't be converted to an argument pair.

components/zcash_address/src/kind/unified/address.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/1647

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/7852a319e69e4c291ea63e7c19b66d48f905b398..1251f959e0af14c6a1657263b40db927f8967c24) to address comments from code review.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/7852a319e69e4c291ea63e7c19b66d48f905b398..f1455af54ff10bb1e0f7ba4ef9bd27adfd3fa0c3) to address comments from code review.

components/zcash_address/src/kind/unified.rs
```diff
@@ -308,22 +545,27 @@ pub(crate) mod private {
...
 
         /// A private function that constructs a unified container with the
         /// specified items, which must be in ascending typecode order.
-        fn try_from_items_internal(items: Vec<Self::Item>) -> Result<Self, ParseError> {
-            assert!(u32::from(Typecode::P2sh) == u32::from(Typecode::P2pkh) + 1);
+        fn try_from_items_internal(
+            revision: Revision,
+            items: Vec<Item<Self::DataItem>>,
+        ) -> Result<Self, ParseError> {
+            assert!(u32::from(Typecode::P2SH) == u32::from(Typecode::P2PKH) + 1);
```
daira, line 567: This could be a [`const_assert_eq!`](https://docs.rs/static_assertions/latest/static_assertions/macro.const_assert_eq.html) I think. Non-blocking.

components/zcash_address/src/kind/unified.rs
```diff
@@ -332,47 +574,46 @@ pub(crate) mod private {
                     return Err(ParseError::InvalidTypecodeOrder);
                 } else if t_code == prev_code {
                     return Err(ParseError::DuplicateTypecode(t));
-                } else if t == Typecode::P2sh && prev_code == Some(u32::from(Typecode::P2pkh)) {
+                } else if t == Typecode::P2SH && prev_code == Some(u32::from(DataTypecode::P2pkh)) {
```
daira, line 577: Why is the case of `Typecode::P2sh` changed?

components/zcash_address/src/kind/unified.rs
```diff
@@ -332,47 +574,46 @@ pub(crate) mod private {
...
     /// violated:
     /// * the item list may not contain two items having the same typecode
-    /// * the item list may not contain only transparent items (or no items)
     /// * the item list may not contain both P2PKH and P2SH items.
-    fn try_from_items(mut items: Vec<Self::Item>) -> Result<Self, ParseError> {
-        items.sort_unstable_by(Self::Item::encoding_order);
-        Self::try_from_items_internal(items)
+    fn try_from_items(
+        revision: Revision,
+        mut items: Vec<Item<Self::DataItem>>,
```
daira, line 613: Consider using `NonEmpty`.

components/zcash_address/src/kind/unified.rs
```diff
@@ -308,22 +545,27 @@ pub(crate) mod private {
...
         }
 
         /// A private function that constructs a unified container with the
         /// specified items, which must be in ascending typecode order.
-        fn try_from_items_internal(items: Vec<Self::Item>) -> Result<Self, ParseError> {
-            assert!(u32::from(Typecode::P2sh) == u32::from(Typecode::P2pkh) + 1);
+        fn try_from_items_internal(
+            revision: Revision,
+            items: Vec<Item<Self::DataItem>>,
+        ) -> Result<Self, ParseError> {
```
daira, line 566: This function should enforce that the UA/UVK contains at least one non-Metadata item, and at least one shielded item in the case of an R0 UA/UVK.

components/zcash_address/src/kind/unified/address.rs
daira: I actually think it's essential to have tests covering the differences between R0 and R1.

daira (approved):
utACK with comments. (Note that I was tired by the end of the review so it might not be as thorough as usual.)

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/38cb9466aa00df8a2670782dbc447c7261ab9496..014e992bbfa955c940e12f0e9a1d2d5136c165a4) to rebase on `main` & fix conflicts with `no_std` changes.

Then [force-pushed](https://github.com/zcash/librustzcash/compare/014e992bbfa955c940e12f0e9a1d2d5136c165a4..154b0021292046782bc96640c934a161f7407708) to take into account review suggestions.

PR #1139, opened 2024-01-26
zcash_keys: Remove `UnifiedAddressRequest::DEFAULT`

nuttycom:
This default only made sense in the context of what was supported by `zcash_client_sqlite`, and not in any other context. Unified address requests no longer have their parts conditioned by what feature flags are available; instead, if a request is constructed for which the required key parts are not supported under a particular selection of feature flags, address generation will raise a runtime error.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
@@ -395,6 +395,7 @@ mod tests {
     #[test]
     #[cfg(feature = "transparent-inputs")]
     fn migrate_from_wm2() {
+        use zcash_client_backend::keys::UnifiedAddressRequest;
         use zcash_primitives::transaction::components::amount::NonNegativeAmount;
 
         let network = Network::TestNetwork;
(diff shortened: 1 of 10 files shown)


zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -437,7 +438,7 @@ mod tests {
 
         let usk = UnifiedSpendingKey::from_seed(&network, &[0u8; 32][..], AccountId::ZERO).unwrap();
         let ufvk = usk.to_unified_full_viewing_key();
-        let (ua, _) = ufvk.default_address();
+        let (ua, _) = ufvk.default_address(UnifiedAddressRequest::unsafe_new(false, true, true));
```
nuttycom, line 441: Since migrations are correct as of a particular point in time, here we manually construct the request given the state of the codebase at the time that the migration was created.

str4d (approved):
utACK 30f1abc34cfa0718fbc382da9160cbfa71eca68c. I will check this with the Swift SDK tests before merging.

zcash_keys/src/keys.rs
```diff
@@ -413,44 +419,41 @@ pub enum AddressGenerationError {
 /// Specification for how a unified address should be generated from a unified viewing key.
 #[derive(Clone, Copy, Debug)]
 pub struct UnifiedAddressRequest {
-    #[cfg(feature = "orchard")]
-    has_orchard: bool,
+    _has_orchard: bool,
```
daira, line 422: Is it necessary for this to have the leading underscore?

daira (approved):
utACK (reviewed with @str4d).

str4d (approved):
ACK 3f3efd2445ff7b5126d804d5f1ffdaf172356ace. I have confirmed that this fixes the Swift SDK's OfflineTests.

PR #1140, opened 2024-01-27
zcash_primitives: Factor out the `memo` module into the `zcash_memos` crate

nuttycom:


nuttycom:
Closing in favor of moving the `memo` module to the `zcash_protocol` crate in #1142

PR #1142, opened 2024-01-27
Factor out protocol constants and value types into a `zcash_protocol` crate.

nuttycom:
This is best reviewed commit-by-commit.

Closes #1157.

PR #1143, opened 2024-01-27
Extract `zip321` crate from `zcash_client_backend`

### LINKED ISSUE #1158: Use `zcash_address` instead of protocol-specific address types in ZIP 321 parsing.

nuttycom:
This depends upon ~#1140 and #1142~

Fixes #1158


zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -540,9 +556,16 @@ where
...
             };
 
-            match &payment.recipient_address {
+            let addr_network = params.address_network().ok_or(ConversionError::User(
+                "Unable to determine network for address.",
+            ))?;
+            let recipient_address = payment
+                .recipient_address
+                .clone()
+                .convert_if_network::<Address>(addr_network)?;
```
nuttycom, line 565: We shouldn't need to have to convert to a `zcash_keys::Address` at this point, but the alternative is creating a separate type just so that we can have a `TryFromAddress` impl that does the equivalent of this match.

components/zcash_address/src/kind/unified/address.rs
```diff
@@ -189,7 +210,11 @@ mod tests {
         })
     }
 
-    fn arb_unified_address_for_typecodes(
+    /// Generates an arbitrary Unified address containing receivers corresponding to the provided
+    /// set of typecodes.. The receivers of this address are ikely to not represent valid protocol
```
daira, line 214:
```suggestion
    /// set of typecodes. The receivers of this address are likely to not represent valid protocol
```

components/zcash_address/src/kind/unified/address.rs
```diff
@@ -206,11 +231,33 @@ mod tests {
             .collect::<Vec<_>>()
     }
 
-    fn arb_unified_address() -> impl Strategy<Value = Address> {
+    /// Generates an arbitrary Unified address. The receivers of this address are ikely to not
```
daira, line 234:
```suggestion
    /// Generates an arbitrary Unified address. The receivers of this address are likely to not
```

components/zcash_address/src/lib.rs
```diff
@@ -266,4 +268,114 @@ impl ZcashAddress {
...
+
+    /// Returns whether this address has the ability to receive transfers of the given pool type.
+    pub fn can_receive_as(&self, pool_type: PoolType) -> bool {
+        match &self.kind {
+            AddressKind::Sprout(_) => false,
+            AddressKind::Sapling(_) => pool_type == PoolType::Shielded(ShieldedProtocol::Sapling),
+            AddressKind::Unified(addr) => addr.has_receiver_of_type(pool_type),
+            AddressKind::P2pkh(_) => pool_type == PoolType::Transparent,
+            AddressKind::P2sh(_) => pool_type == PoolType::Transparent,
+            AddressKind::Tex(_) => pool_type == PoolType::Transparent,
```
daira, line 280: These three case could be combined with a `|` pattern.

components/zip321/LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2017-2019 Electric Coin Company
```
daira, line 3:
```suggestion
Copyright (c) 2017-2024 Electric Coin Company
```

components/zip321/src/lib.rs
```diff
@@ -110,29 +116,43 @@ pub fn memo_from_base64(s: &str) -> Result<MemoBytes, Zip321Error> {
...
-    /// purpose of this payment.
-    pub message: Option<String>,
-    /// A list of other arbitrary key/value pairs associated with this payment.
-    pub other_params: Vec<(String, String)>,
+    recipient_address: ZcashAddress,
+    amount: Zatoshis,
+    memo: Option<MemoBytes>,
+    label: Option<String>,
+    message: Option<String>,
+    other_params: Vec<(String, String)>,
```
daira, line 124: There's no need to remove documentation just because the fields are private.

zcash_client_sqlite/src/error.rs
```diff
@@ -33,15 +34,16 @@ pub enum SqliteClientError {
...
+    /// A Zcash key or address decoding error
+    DecodingError(ParseError),
 
     /// An error produced in legacy transparent address derivation
     #[cfg(feature = "transparent-inputs")]
     HdwalletError(hdwallet::error::Error),
 
     /// An error encountered in decoding a transparent address from its
     /// serialized form.
+    #[cfg(feature = "transparent-inputs")]
```
daira, line 46: Why is this specific to transparent-inputs? It seems as though it could also occur in principle due to support for transparent outputs, and the fact that it doesn't currently is just an implementation detail.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1126,20 +1144,22 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                                &wdb.params,
+                                wdb.conn.0,
+                                *output.account(),
+                                &receiver
+                            )?.unwrap_or_else(||
+                                receiver.to_zcash_address(wdb.params.network_type())
+                            );
+
+                            Recipient::External(wallet_address, PoolType::Shielded(ShieldedProtocol::Orchard))
+                        };
```
daira, line 1159: So much near-duplicate code 😢

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2366,6 +2370,49 @@ pub(crate) fn put_tx_meta(
...
+                    ":taddr": Address::Transparent(*taddr).encode(params)
+                },
+                |row| row.get::<_, String>(0),
+            )
+            .optional()?
+            .map(|addr_str| addr_str.parse::<ZcashAddress>())
+            .transpose()
+            .map_err(SqliteClientError::from),
+        receiver => {
+            // at present,
```
daira, line 2398: Expand on this comment.

zcash_keys/src/address.rs
```diff
@@ -233,6 +233,54 @@ impl UnifiedAddress {
     }
 }
 
+/// An enumeration of protocol-level receiver types. While these correspond to unified address
+/// receiver
```
daira, line 237: This comment seems incomplete.

daira (approved):
utACK with comments.

zcash_keys/src/address.rs
```diff
@@ -233,6 +233,58 @@ impl UnifiedAddress {
...
+            Receiver::Transparent(TransparentAddress::PublicKeyHash(data)) => {
+                ZcashAddress::from_transparent_p2pkh(net, *data)
+            }
+            Receiver::Transparent(TransparentAddress::ScriptHash(data)) => {
+                ZcashAddress::from_transparent_p2sh(net, *data)
+            }
+        }
+    }
+
+    pub fn corresponds(&self, addr: &ZcashAddress) -> bool {
```
daira, line 272: Please document this.

daira (approved):
utACK with comments.

zcash_keys/src/address.rs
```diff
@@ -233,6 +233,58 @@ impl UnifiedAddress {
     }
 }
 
+/// An enumeration of protocol-level receiver types. While these correspond to unified address
+/// receiver
```
daira, line 237:
```suggestion
/// An enumeration of protocol-level receiver types.
```

daira (approved):
re-utACK with comments.

zcash_keys/src/address.rs
```diff
@@ -291,8 +339,14 @@ impl TryFromRawAddress for Address {
 
 impl Address {
     pub fn decode<P: consensus::Parameters>(params: &P, s: &str) -> Option<Self> {
-        let addr = ZcashAddress::try_from_encoded(s).ok()?;
-        addr.convert_if_network(params.network_type()).ok()
+        Self::try_from_zcash_address(params, s.parse::<ZcashAddress>().ok()?).ok()
+    }
+
+    pub fn try_from_zcash_address<P: consensus::Parameters>(
```
str4d, line 354: Document this.

str4d (changes requested):
Reviewed aeac544aedb0b147b4cfa09f4abeee2c06ff1abe. The `dbg!` removal is blocking.

components/zcash_address/Cargo.toml
```diff
@@ -19,18 +19,18 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-bech32 = "0.9"
-bs58 = { version = "0.5", features = ["check"] }
+bech32.workspace = true
+bs58.workspace = true
```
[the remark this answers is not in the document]
nuttycom: Yeah, I think that it's better to have everything consistent to the workspace; that way the workspace-level deps can be updated as needed and we won't end up with multiple versions in the dependency tree, and if we factor the crates out we'll have no choice but to provide specific version information.

str4d (approved):
utACK d2aa6cfc7f1e17afa182470611c14cc5058fe0b3

PR #1149, opened 2024-02-02
zcash_primitives: Rename `TransparentAddress` variants.

nuttycom:
This resolves an old TODO.


str4d (approved):
utACK 2360609f1a377e0f87c42c3903e4f56d9c849864

PR #1151, opened 2024-02-05
Add `WalletWrite::insert_address_with_diversifier_index` function

AArnott:
This enables use cases where the diversifier index is prescribed instead of sequentially assigned.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -22,6 +22,7 @@ use zcash_primitives::{
     },
     zip32::{AccountId, Scope},
 };
+use zip32::DiversifierIndex;
 
 use crate::{
     address::{AddressMetadata, UnifiedAddress},
@@ -1019,6 +1020,27 @@ pub trait WalletWrite: WalletRead {
         request: UnifiedAddressRequest,
     ) -> Result<Option<UnifiedAddress>, Self::Error>;
 
+    /// Generates and persists a new address for the specified account, with the specified
+    /// diversifier index.
+    ///
+    /// Returns the new address, or an error if the account identifier does not correspond to a
+    /// known account.
+    /// A conflict with an existing row in the database is considered acceptable and no error is returned.
+    /// If the diversifier index cannot produce a valid Sapling address, no sapling receiver will
+    /// be included in the returned address.
+    /// If the diversifier is outside the range for transparent addresses, no transparent receiver
+    /// will be included in the returned address.
+    ///
+    /// This supports a more advanced use case than `get_next_available_address` where the caller
+    /// simply gets the next diversified address sequentially. Mixing use of the two functions
+    /// is not recommended because `get_next_available_address` will return the next address
+    /// after the highest diversifier index in the database, which may leave gaps in the sequence.
+    fn insert_address_with_diversifier_index(
+        &mut self,
+        account: AccountId,
+        diversifier_index: DiversifierIndex,
+    ) -> Result<UnifiedAddress, Self::Error>;
+
     /// Updates the state of the wallet database by persisting the provided block information,
     /// along with the note commitments that were detected when scanning the block for transactions
     /// pertaining to this wallet.
@@ -1105,6 +1127,7 @@ pub mod testing {
     use secrecy::{ExposeSecret, SecretVec};
     use shardtree::{error::ShardTreeError, store::memory::MemoryShardStore, ShardTree};
     use std::{collections::HashMap, convert::Infallible, num::NonZeroU32};
+    use zip32::DiversifierIndex;
 
     use zcash_primitives::{
         block::BlockHash,
@@ -1321,6 +1344,14 @@ pub mod testing {
             _request: UnifiedAddressRequest,
         ) -> Result<Option<UnifiedAddress>, Self::Error> {
             Ok(None)
+        }
+
+        fn insert_address_with_diversifier_index(
+            &mut self,
+            _account: AccountId,
+            _diversifier_index: DiversifierIndex,
+        ) -> Result<UnifiedAddress, Self::Error> {
+            todo!()
         }
 
         #[allow(clippy::type_complexity)]
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -42,6 +42,7 @@ use std::{
     borrow::Borrow, collections::HashMap, convert::AsRef, fmt, num::NonZeroU32, ops::Range,
     path::Path,
 };
+use zcash_keys::keys::AddressGenerationError;
 
 use incrementalmerkletree::Position;
 use shardtree::{error::ShardTreeError, ShardTree};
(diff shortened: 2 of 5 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -1019,6 +1020,27 @@ pub trait WalletWrite: WalletRead {
...
     ) -> Result<Option<UnifiedAddress>, Self::Error>;
 
+    /// Generates and persists a new address for the specified account, with the specified
+    /// diversifier index.
+    ///
+    /// Returns the new address, or an error if the account identifier does not correspond to a
+    /// known account.
+    /// A conflict with an existing row in the database is considered acceptable and no error is returned.
+    /// If the diversifier index cannot produce a valid Sapling address, no sapling receiver will
+    /// be included in the returned address.
```
nuttycom, line 1030: I think that this method should take a `UnifiedAddressRequest` instead of having this default behavior: https://github.com/zcash/librustzcash/blob/main/zcash_keys/src/keys.rs#L421-L425

zcash_client_backend/src/data_api.rs
```diff
@@ -1323,6 +1346,14 @@ pub mod testing {
             Ok(None)
         }
 
+        fn insert_address_with_diversifier_index(
+            &mut self,
+            _account: AccountId,
+            _diversifier_index: DiversifierIndex,
+        ) -> Result<UnifiedAddress, Self::Error> {
+            todo!()
```
nuttycom, line 1354: Please either implement this, or add an issue to track the TODO

zcash_client_sqlite/src/lib.rs
```diff
@@ -445,6 +446,63 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                &addr,
+            ) {
+                Ok(_) => Ok(addr),
+                Err(rusqlite::Error::SqliteFailure(
+                    libsqlite3_sys::Error {
+                        code: libsqlite3_sys::ErrorCode::ConstraintViolation,
+                        ..
+                    },
+                    _,
+                )) => Ok(addr), // conflicts are ignorable
```
nuttycom, line 500: Should we permit conflicts on insertion more generally, and just add an `ON CONFLICT DO NOTHING` clause to the SQL code in `wallet::insert_address`?

zcash_client_sqlite/src/lib.rs
```diff
@@ -445,6 +446,63 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+        account: AccountId,
+        diversifier_index: DiversifierIndex,
+    ) -> Result<UnifiedAddress, SqliteClientError> {
+        self.transactionally(|wdb| {
+            let keys = wdb.get_unified_full_viewing_keys()?;
+            let ufvk = keys
+                .get(&account)
+                .ok_or(SqliteClientError::AccountUnknown(account))?;
+
+            let has_orchard = true;
```
nuttycom, line 460: This can't be merged as-is, because this will cause Orchard receivers to begin being generated before the wallet has `orchard` support, which is a "the user can't see their received funds" risk. Instead, either use https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/lib.rs#L110 or have the request passed in by the caller.

nuttycom (changes requested):
This is looking good, but we should not hardcode Orchard receivers until Orchard is supported by the rest of `zcash_client_sqlite` and in general, I think we want to move in the direction of having explicit requests passed in (the current hardcoded default is temporary.)

zcash_client_backend/src/data_api.rs
AArnott: The reason I left it out was that elsewhere, it was used to influence or check the diversifier index would work for the required receivers. In this case, the caller requires the exact index given to be inserted, so compatibility checks are moot. I didn't really need to return a UnifiedAddress from the function either. In my case I drop it on the floor. 
My concern with taking `UnifiedAddressRequest` here is that it complicates the caller, as it now must describe what it expects to be the allowed receivers (and no more than that), when that isn't necessary.

nuttycom: The caller may be someone like the Brave browser, which will not be supporting the Sapling pool at all. So it _must_ be up to the caller what receivers are generated, and they must handle the error that can occur if the diversifier index does not correspond to an index that is valid for the request they provided.

zcash_client_backend/src/data_api.rs
AArnott: Would returning `Err(())` count? It's difficult to imagine a better implementation given this mock doesn't store anything at all.

zcash_client_sqlite/src/lib.rs
AArnott: That potentially presents a problem: if orchard cannot be guaranteed, and sapling may fail at a given diversifier index, that leaves me with just transparent in some cases. I can't construct a UA with just a transparent receiver. Yet the sqlite table I'm adding a row to has an `address` column which (at least so far) seems to only take UAs. Can I insert a transparent address there, or will code fall over?

AArnott: Looking at `get_next_available_address` I guess I don't see any insurance that T addresses aren't put into that table as it is, so maybe this isn't a new problem.

AArnott: Maybe I should just create a new table to track the transparent addresses. It's not in this PR, but I have a follow-up change in my fork that adds a column to the `addresses` table to record the last sync'd block for the transparent component of the address. If that isn't tolerable for this crate, then my creating a whole new table to store the data I need (and isn't wanted upstream) may be a better overall approach.

zcash_client_backend/src/data_api.rs
nuttycom: I was thinking that the mock could store a seed, but just adding an issue for this is fine.

zcash_client_sqlite/src/lib.rs
nuttycom: If a Sapling receiver is requested and no valid diversifier exists at the specified index, that should result in an error being returned; we should _never_ generate an address that does not have all of the requested components.

AArnott: That doesn't really answer my question. I need to be able to store addresses for which no valid sapling receiver exists. Since I can't store orchard receivers in this table without the caller's permission, I have to either store transparent addresses in the `address` column, or find another table to store this data. Do you have a preference?

nuttycom: Is the use case you're after here the "transparent address zero" problem, wherein older wallets may have generated a transparent address at index zero but there is no corresponding Sapling address, and so you can't generate a UA to correspond? 

That is a use case that we do need to handle, and I'd rather not special-case it, so we could potentially handle it by making the `address` column nullable, and using the `cached_transparent_receiver_address` for this purpose. However, for that, you'd need a different entrypoint, as at present you can't create UA requests that don't have a shielded component.

The alternative is that in ZIP 316 Revision 1, we could permit transparent-only UAs. Ultimately this might be what we do.

AArnott: No, this isn't directly related to the index 0 problem at all. This is simply about tracking which transparent addresses have been created/used so that sync knows which addresses to download transactions for.

nuttycom: So up until now, you could obtain this from the transparent parts of UAs saved to the `addresses` table, plus the 0 address (which I had forgotten I added handling for in https://github.com/zcash/librustzcash/pull/685). Is the thing that is changing is that you intend to generate transparent addresses that are _not_ correlated to any UA? Sorry if I'm being obtuse, but I'm still having a bit of trouble understanding the use case.

AArnott: Yes. *All* my transparent addresses are generated completely independently of any UA. So I need to be able to represent all of them, even if their sapling counterpart is invalid.

nuttycom:
This PR has been subsumed by #1673: https://github.com/zcash/librustzcash/pull/1673/files#diff-7f82231afa857f8e7562f9184e551564997e1f74aaad9c6ba17e348fc89abeadR2407-R2412

PR #1164, opened 2024-02-06
zcash_client_backend: Fix broken `zip321` tests.

nuttycom:
These tests appear to have been broken by #1139; it's not clear why CI passed on that at the time.


str4d (approved):
utACK 3acd6348aad7f87903eb38209e8710e609d510e7

PR #1168, opened 2024-02-07
zcash_keys: Add `sapling` and `transparent` feature flags.

### LINKED ISSUE #1160: `zcash_keys`: Add a `sapling` feature flag that can be used to eliminate the dependency on `sapling-crypto`
This was requested by the Brave folks to help limit the needed scope of their audit.

nuttycom:
Fixes #1160

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -140,7 +140,7 @@ where
 /// };
 /// use zcash_proofs::prover::LocalTxProver;
 /// use zcash_client_backend::{
-///     keys::UnifiedSpendingKey,
+///     keys::{UnifiedSpendingKey, UnifiedAddressRequest},
 ///     data_api::{wallet::create_spend_to_address, error::Error, testing},
 ///     wallet::OvkPolicy,
 /// };
@@ -166,8 +166,13 @@ where
 /// };
 ///
 /// let account = AccountId::from(0);
+/// let req = UnifiedAddressRequest {
+///     has_orchard: false,
+///     has_sapling: true,
+///     has_p2pkh: true
+/// };
 /// let usk = UnifiedSpendingKey::from_seed(&Network::TestNetwork, &[0; 32][..], account).unwrap();
-/// let to = usk.to_unified_full_viewing_key().default_address().0.into();
+/// let to = usk.to_unified_full_viewing_key().default_address(req).0.into();
 ///
 /// let mut db_read = testing::MockWalletDb {
 ///     network: Network::TestNetwork
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -166,11 +166,7 @@ where
 /// };
 ///
 /// let account = AccountId::from(0);
-/// let req = UnifiedAddressRequest {
-///     has_orchard: false,
-///     has_sapling: true,
-///     has_p2pkh: true
-/// };
+/// let req = UnifiedAddressRequest::new(false, true, true);
 /// let usk = UnifiedSpendingKey::from_seed(&Network::TestNetwork, &[0; 32][..], account).unwrap();
 /// let to = usk.to_unified_full_viewing_key().default_address(req).0.into();
 ///
--- a/zcash_keys/src/address.rs
+++ b/zcash_keys/src/address.rs
@@ -1,17 +1,20 @@
 //! Structs for handling supported address types.
 
-use sapling::PaymentAddress;
 use zcash_address::{
     unified::{self, Container, Encoding, Typecode},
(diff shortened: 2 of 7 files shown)


zcash_keys/src/keys.rs
```diff
@@ -628,39 +657,52 @@ impl UnifiedFullViewingKey {
     /// Returns `None` if the specified index does not produce a valid diversifier.
     pub fn address(
         &self,
-        j: DiversifierIndex,
-        request: UnifiedAddressRequest,
+        _j: DiversifierIndex,
+        _request: UnifiedAddressRequest,
     ) -> Result<UnifiedAddress, AddressGenerationError> {
         #[cfg(feature = "orchard")]
-        let orchard = if request._has_orchard {
```
str4d, line 635: Hmm, IIRC this is a bugfix to how `UnifiedAddressRequest` works. We probably want to get this PR in before `zcash_keys 0.1.0` in that case, which means the earlier feature flag issue I raised in the PR needs to be fixed.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -168,8 +168,13 @@ where
 /// };
 ///
 /// let account = AccountId::from(0);
+/// let req = UnifiedAddressRequest {
+///     has_orchard: false,
+///     has_sapling: true,
+///     has_p2pkh: true
+/// };
```
str4d, line 175: This won't compile, because the fields of `UnifiedAddressRequest` are private. I suspect it's not being detected in CI however because of the required feature flags.
```suggestion
/// let req = UnifiedAddressRequest::new(false, true, true);
```
The required feature flags for this doctest will also need to be updated to include `sapling`.

zcash_keys/src/address.rs
```diff
@@ -199,6 +222,7 @@ impl From<UnifiedAddress> for Address {
 impl TryFromRawAddress for Address {
     type Error = &'static str;
 
+    #[cfg(feature = "sapling")]
```
str4d, line 225: Yay, this is exactly the kind of use case that the `zcash_address` crate conversion traits are designed to handle 😄

zcash_keys/src/address.rs
```diff
@@ -293,20 +339,24 @@ mod tests {
...
-        #[cfg(feature = "orchard")]
+        #[cfg(all(feature = "orchard", feature = "sapling"))]
         let ua = UnifiedAddress::from_receivers(orchard, sapling, transparent).unwrap();
 
-        #[cfg(not(feature = "orchard"))]
+        #[cfg(all(not(feature = "orchard"), feature = "sapling"))]
         let ua = UnifiedAddress::from_receivers(sapling, transparent).unwrap();
 
+        #[cfg(all(not(feature = "orchard"), not(feature = "sapling")))]
+        let ua = UnifiedAddress::from_receivers(transparent).unwrap();
```
str4d, line 358: This test case should be failing at the unwrap. I suspect it doesn't fail CI because we don't (and can't really) test all feature flag combinations. Move it into a separate test (or early-return after checking you get `None`).

zcash_keys/src/address.rs
```diff
@@ -293,20 +339,24 @@ mod tests {
...
 
-        let transparent = { None };
+        let transparent = None;
 
-        #[cfg(feature = "orchard")]
+        #[cfg(all(feature = "orchard", feature = "sapling"))]
         let ua = UnifiedAddress::from_receivers(orchard, sapling, transparent).unwrap();
 
-        #[cfg(not(feature = "orchard"))]
+        #[cfg(all(not(feature = "orchard"), feature = "sapling"))]
```
str4d, line 354: We need an additional case for `all(feature = "orchard", not(feature = "sapling"))`.

str4d (changes requested):
Reviewed 56da01f51388f41535d0ff89091e803bd75745ac

zcash_keys/src/keys.rs
```diff
@@ -831,85 +889,129 @@ mod tests {
...
+            ))]
+            assert_eq!(decoded_with_t.unknown.len(), 1);
+
+            // Orchard enabled
+            #[cfg(all(
+                feature = "orchard",
+                not(feature = "sapling"),
+                feature = "transparent-inputs"
+            ))]
+            assert_eq!(decoded_with_t.unknown.len(), 2);
```
str4d, line 993: If I'm counting the `not`s in the `cfg` correctly, this should be:
```suggestion
            assert_eq!(decoded_with_t.unknown.len(), 1);
```

str4d (dismissed):
utACK 2a6330f2ea903897cd355c23eff9abb0f93cae14

str4d (approved):
Re-utACK 635bc3158b1885ad5b6f9f75009d0fb9bb5ce281

PR #1172, opened 2024-02-07
zcash_client_sqlite: Do not request transparent UA components if "transparent-inputs" feature is disabled.

nuttycom:


zcash_client_sqlite/src/lib.rs
```diff
@@ -107,8 +107,13 @@ pub(crate) const VERIFY_LOOKAHEAD: u32 = 10;
 
 pub(crate) const SAPLING_TABLES_PREFIX: &str = "sapling";
 
-pub const DEFAULT_UA_REQUEST: UnifiedAddressRequest =
-    UnifiedAddressRequest::unsafe_new(false, true, true);
+#[cfg(not(feature = "transparent-inputs"))]
+pub(crate) const UA_TRANSPARENT: bool = false;
+#[cfg(feature = "transparent-inputs")]
+pub(crate) const UA_TRANSPARENT: bool = true;
```
str4d, line 113: In hindsight it makes sense to me why the introduction of `UnifiedAddressRequest` interacts this poorly with protocol feature flagging. Previously we relied on the feature flagging for determining what receiver kinds the caller cares about. Now, the caller explicitly tells us, and so there's a new error condition that we were previously not accounting for (that we might require a transparent receiver when the code to do so is flagged off).

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -438,7 +440,11 @@ mod tests {
 
         let usk = UnifiedSpendingKey::from_seed(&network, &[0u8; 32][..], AccountId::ZERO).unwrap();
         let ufvk = usk.to_unified_full_viewing_key();
-        let (ua, _) = ufvk.default_address(UnifiedAddressRequest::unsafe_new(false, true, true));
+        let (ua, _) = ufvk.default_address(UnifiedAddressRequest::unsafe_new(
+            false,
+            true,
+            UA_TRANSPARENT,
```
str4d, line 446: This and the other migrations are going to need to be handled carefully in future when we allow Sapling to be disabled, because the migrations are written assuming that Sapling is always available.

str4d (approved):
Tested ACK 7c0b9da9b7e362bc3b498abf04a07499d1391317

zcash_client_sqlite/src/lib.rs
nuttycom: So, another possible design for `UnifiedAddressRequest` is to have it as a struct with all public members, each of which is guarded by the associated feature flag. We didn't want to do that, because it would allow transparent-only requests to be constructed; however, we've been talking about relaxing that requirement in ZIP 316. In the case we do that, I think it would make sense to use a struct-with-flagged-members approach.

PR #1173, opened 2024-02-07
zcash_client_backend: Extend `BatchRunner` to support full outputs

### LINKED ISSUE #1171: zcash_client_backend: Extend `BatchRunner` to support full outputs
`BatchRunner` and its associated logic was copied from `zcashd` and modified to work on compact outputs instead of full outputs. This means we have basically-identical code duplicated between `zcashd` and `zcash_client_backend`. We should generalise the code in `zcash_client_backend` to support both kinds of outputs (which just changes whether or not the memo is available for decrypted outputs). This would let us remove the code in `zcashd`, and also provide a foundation for a future Rust full node wallet's scanning logic.

str4d:
Closes #1171.

### DIFF
--- a/zcash_client_backend/src/scan.rs
+++ b/zcash_client_backend/src/scan.rs
@@ -524,9 +524,9 @@ where
     /// batch, or the all-zeros hash to indicate that no block triggered it (i.e. it was a
     /// mempool change).
     ///
-    /// If after adding the given outputs, the accumulated batch size is at least
-    /// `BATCH_SIZE_THRESHOLD`, `Self::flush` is called. Subsequent calls to
-    /// `Self::add_outputs` will be accumulated into a new batch.
+    /// If after adding the given outputs, the accumulated batch size is at least the size
+    /// threshold that was set via `Self::new`, `Self::flush` is called. Subsequent calls
+    /// to `Self::add_outputs` will be accumulated into a new batch.
     pub(crate) fn add_outputs(
         &mut self,
         block_tag: BlockHash,
--- a/zcash_client_backend/src/scan.rs
+++ b/zcash_client_backend/src/scan.rs
@@ -12,28 +12,31 @@ use zcash_note_encryption::{batch, BatchDomain, Domain, ShieldedOutput, COMPACT_
 use zcash_primitives::{block::BlockHash, transaction::TxId};
 
 /// A decrypted transaction output.
-pub(crate) struct DecryptedOutput<A, D: Domain> {
+pub(crate) struct DecryptedOutput<A, D: Domain, M> {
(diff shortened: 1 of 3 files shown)


zcash_client_backend/src/scan.rs
```diff
@@ -8,36 +8,107 @@ use std::sync::{
...
+    // Once we reach MSRV 1.75.0, this can return `impl Iterator`.
+    fn batch_decrypt(
+        tags: &[A],
+        ivks: &[D::IncomingViewingKey],
+        outputs: &[(D, Output)],
+    ) -> Vec<Option<DecryptedOutput<A, D, Self::Memo>>>;
+}
+
+/// A decryptor of outputs as encoded in transactions.
+pub(crate) struct FullDecryptor;
```
nuttycom, line 59: If this type is only pub(crate), doesn't that limit its utility somewhat? Same question about the trait. It's fine for now, but I expect eventually we'll want to expose them?

zcash_client_backend/src/scan.rs
```diff
@@ -8,36 +8,107 @@ use std::sync::{
...
             .finish()
     }
 }
 
+/// A decryptor of transaction outputs.
+pub(crate) trait Decryptor<A, D: BatchDomain, Output> {
+    type Memo;
+
+    // Once we reach MSRV 1.75.0, this can return `impl Iterator`.
+    fn batch_decrypt(
```
nuttycom, line 51: It seems like `batch_decrypt` could be polymorphic in `A`, rather than it having to be fixed at the `Decryptor` level? It would probably then need the `: Clone` bound here.

nuttycom (approved):
utACK 9eb671ef8d7a5564ed5b4d17559163c79f0283bc with nonblocking comments

zcash_client_backend/src/scan.rs
str4d: Yep, the intention is that `BatchRunner` will become public, along with everything that is required for that. But I want to ensure that what is exposed is actually usable first. I've already made some adjustments locally while testing this with `zcashd`.

zcash_client_backend/src/scan.rs
str4d: Yeah, I'll try something like that.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/9eb671ef8d7a5564ed5b4d17559163c79f0283bc..f6224e7e8056039a3fc0e2812139086092ba7d24) to address comment.

nuttycom (approved):
re-utACK f6224e7e8056039a3fc0e2812139086092ba7d24

PR #1181, opened 2024-02-08
zcash_client_backend: Add per-pool flagging to `ScanRange`.

### LINKED ISSUE #1119: `zcash_client_backend`: Update `SpanningTree` and related types to track Orchard and Sapling scanning separately.
The approach here is to add a `ScanFlags` type to `ScanRange`, where the flags indicate in which pools, if any, scanning must be performed.

In ranges having `Scanned` and `Ignored` priority, no flags may be set. In all other ranges, at least one flag must be set.

nuttycom:
Fixes #1119


nuttycom:
This is ultimately not the direction that is best to go in for this, since it doesn't handle spendability when rescanning with imported keys.

Issue #1169, opened 2024-02-07
Extract sync state machine from `zec-client-sqlite` into `zcash_client_backend`


str4d:
Currently the canonical state machines for using the `zcash_client_backend` APIs live in the Android and Swift mobile SDKs. This would provide a Rust implementation of the state machine, that could be used by pure- or primarily-Rust wallets as well as web wallets, and could subsequently be improved.

PR #1184, opened 2024-02-09
zcash_client_backend: Implement async wallet synchronization function

str4d:
This implements the necessary state machine for taking a wallet in some arbitrary synchronization status, and fully scanning (the remainder of) the chain.

Closes zcash/librustzcash#1169.

### DIFF
--- a/zcash_client_backend/Cargo.toml
+++ b/zcash_client_backend/Cargo.toml
@@ -66,7 +66,7 @@ tracing.workspace = true
 # - Protobuf interfaces and gRPC bindings
 hex.workspace = true
 prost.workspace = true
-tonic = { workspace = true, optional = true, features = ["prost", "codegen"]}
+tonic = { workspace = true, optional = true, features = ["prost", "codegen"] }
 
 # - Secret management
 secrecy.workspace = true
@@ -78,6 +78,10 @@ group.workspace = true
 orchard = { workspace = true, optional = true }
 sapling.workspace = true
 
+# - Sync engine
+async-trait = { version = "0.1", optional = true }
+futures-util = { version = "0.3", optional = true }
+
 # - Note commitment trees
 incrementalmerkletree.workspace = true
 shardtree.workspace = true
@@ -88,9 +92,6 @@ jubjub = { workspace = true, optional = true }
 
 # - ZIP 321
 nom = "7"
-
-# - Asychronous
-async-trait = "0.1.78"
 
 # Dependencies used internally:
 # (Breaking upgrades to these are usually backwards-compatible, but check MSRVs.)
@@ -106,7 +107,7 @@ crossbeam-channel.workspace = true
 rayon.workspace = true
 
 [build-dependencies]
-tonic-build = { workspace = true, features = ["prost"]}
+tonic-build = { workspace = true, features = ["prost"] }
 which = "4"
 
 [dev-dependencies]
@@ -140,6 +141,13 @@ transparent-inputs = [
 
 ## Enables receiving and spending Orchard funds.
 orchard = ["dep:orchard", "zcash_keys/orchard"]
+
+## Exposes a wallet synchronization function that implements the necessary state machine.
+sync = [
+    "lightwalletd-tonic",
+    "dep:async-trait",
+    "dep:futures-util",
+]
 
 ## Exposes APIs that are useful for testing, such as `proptest` strategies.
 test-dependencies = [
--- a/zcash_client_backend/src/sync.rs
+++ b/zcash_client_backend/src/sync.rs
@@ -1,0 +1,491 @@
+//! Implementation of the synchronization flow described in the crate root.
+//!
+//! This is currently a simple implementation that does not yet implement a few features:
+//!
+//! - Block batches are not downloaded in parallel with scanning.
+//! - Transactions are not enhanced once detected (that is, after an output is detected in
+//!   a transaction, the full transaction is not downloaded and scanned).
+//! - There is no mechanism for notifying the caller of progress updates.
+//! - There is no mechanism for interrupting the synchronization flow, other than ending
+//!   the process.
+
+use std::fmt;
+
+use futures_util::TryStreamExt;
+use shardtree::error::ShardTreeError;
+use subtle::ConditionallySelectable;
+use tonic::{
+    body::BoxBody,
+    client::GrpcService,
+    codegen::{Body, Bytes, StdError},
+};
+use tracing::{debug, info};
+use zcash_primitives::{
+    consensus::{BlockHeight, Parameters},
+    merkle_tree::HashSer,
+};
+
+use crate::{
+    data_api::{
+        chain::{
+            error::Error as ChainError, scan_cached_blocks, BlockCache, ChainState,
+            CommitmentTreeRoot,
+        },
+        scanning::{ScanPriority, ScanRange},
+        WalletCommitmentTrees, WalletRead, WalletWrite,
+    },
+    proto::service::{self, compact_tx_streamer_client::CompactTxStreamerClient, BlockId},
+    scanning::ScanError,
+};
+
+#[cfg(feature = "orchard")]
+use orchard::tree::MerkleHashOrchard;
+
+/// Scans the chain until the wallet is up-to-date.
+pub async fn run<P, ChT, CaT, DbT>(
+    client: &mut CompactTxStreamerClient<ChT>,
+    params: &P,
+    db_cache: &CaT,
+    db_data: &mut DbT,
+    batch_size: u32,
+) -> Result<(), Error<CaT::Error, <DbT as WalletRead>::Error, <DbT as WalletCommitmentTrees>::Error>>
+where
+    P: Parameters + Send + 'static,
+    ChT: GrpcService<BoxBody>,
+    ChT::Error: Into<StdError>,
+    ChT::ResponseBody: Body<Data = Bytes> + Send + 'static,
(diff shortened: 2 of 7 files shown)


zcash_client_backend/src/sync.rs
```diff
@@ -0,0 +1,399 @@
...
+                .with_blocks(Some(rewind_height + 1), None, |block| {
+                    let meta = BlockMeta {
+                        height: block.height(),
+                        block_hash: block.hash(),
+                        block_time: block.time,
+                        // These values don't matter for deletion.
+                        sapling_outputs_count: 0,
+                        orchard_actions_count: 0,
+                    };
+                    std::fs::remove_file(get_block_path(fsblockdb_root, &meta))
```
fluidvanadium, line 375: why do we individually remove files instead of relying on truncate_to_height?

str4d: `WalletWrite::truncate_to_height` is truncating the _wallet database_; it has no knowledge about the internal details of the block cache. This file deletion is part of the concrete behaviour required to use `FsBlockDb`, and as such it should move behind the new trait (i.e. there should be an equivalent `BlockCache::truncate_to_height` method).

zcash_client_backend/src/sync.rs
```diff
@@ -0,0 +1,399 @@
...
+        WalletCommitmentTrees, WalletRead, WalletWrite,
+    },
+    proto::service::{self, compact_tx_streamer_client::CompactTxStreamerClient},
+};
+
+pub(crate) fn get_block_path(fsblockdb_root: &Path, meta: &BlockMeta) -> PathBuf {
+    meta.block_file_path(&fsblockdb_root.join("blocks"))
+}
+
+/// Scans the chain until the wallet is up-to-date.
```
AArnott, line 33: Why stop there? Why not add a parameter to indicate that the caller wishes the method to run indefinitely, keeping up with the blockchain?

zcash_client_backend/src/sync.rs
```diff
@@ -0,0 +1,399 @@
...
+                deletion.await?;
+            }
+            return Ok(true);
+        }
+    }
+
+    info!("Waiting for cached blocks to be deleted...");
+    for deletion in block_deletions {
+        deletion.await?;
+    }
```
AArnott, line 183: In my perf testing, at least on Windows, the disk I/O associated with storing and soon after deleting the blocks from disk _dominated_ the sync time. By keeping the batch in memory, sync went _much_ faster. And I have a fast SSD by the way. The downside of this approach is with a fixed batch size, the memory requirement for this varies widely, as sandblasted blocks can cause a batch that typically requires small memory to take >2GB of memory. So I have in mind another approach that somewhat does away with batches altogether. We can chat about it if you're interested. Essentially though, it's the idea that we'll ask the server to give us _all_ the blocks we're interested in, and we'll throttle the download speed by reading them no faster than we have space for in a deliberately limited queue that feeds into the scan_blocks fn. This way, we can keep the scan_blocks function busy, memory usage under control, and not worry about some somewhat arbitrary value for 'batch size'.

zcash_client_backend/src/sync.rs
```diff
@@ -0,0 +1,399 @@
...
+            }
+        })
+    }) {
+        // Download the blocks in `scan_range` into the block source.
+        let block_meta = download_blocks(client, fsblockdb_root, db_cache, &scan_range).await?;
+
+        // Scan the downloaded blocks.
+        let scan_ranges_updated =
+            scan_blocks(params, fsblockdb_root, db_cache, db_data, &scan_range)?;
+
```
AArnott, line 165: I have this pattern in my sync implementation as well, and it's painfully obvious when looking at the resource manager that sync time takes about twice as long as it needs to because resource utilization keeps switching between bandwidth and CPU. If we could be downloading the next batch _while_ scanning the prior batch, we could better utilize both resources in parallel and get the job done much faster.

zcash_client_backend/src/sync.rs
```diff
@@ -0,0 +1,399 @@
...
+    },
+    proto::service::{self, compact_tx_streamer_client::CompactTxStreamerClient},
+};
+
+pub(crate) fn get_block_path(fsblockdb_root: &Path, meta: &BlockMeta) -> PathBuf {
+    meta.block_file_path(&fsblockdb_root.join("blocks"))
+}
+
+/// Scans the chain until the wallet is up-to-date.
+pub async fn run<P, ChT, DbT>(
```
AArnott, line 34: This method only scans CompactBlocks, which contain only shielded transaction data. Transparent data is totally omitted. I had to fill in the gap in my sync routine, which is mostly complete, although zcash_client_backend itself doesn't yet support recording a *spend* of a transparent output, so while I download transparent transactions, the recorded history is incorrect if UTXOs are spent, but it works if they are shielded. 
In any case, I think the transparent support either needs to be added here, or the function should be very prominently documented as to its limitations.

zcash_client_backend/src/sync.rs
AArnott: Other parameter ideas from my own implementation: 
- take a callback parameter that receives % progress updates as well as newly found transactions.
- take a cancellation token so the caller can gracefully abort the sync where it's at (such that it can be resumed at roughly the same location later).

zcash_client_backend/Cargo.toml
```diff
@@ -68,6 +68,11 @@ group.workspace = true
 orchard = { workspace = true, optional = true }
 sapling.workspace = true
 
+# - Sync engine
+anyhow = { version = "1", optional = true }
```
AArnott, line 72: I hope you'll avoid `anyhow` errors. Use an enum type with members for each error condition. That way the app can reason about what went wrong and make a decision about how to recover, or what to communicate with the user.

zcash_client_backend/src/sync.rs
```diff
@@ -0,0 +1,399 @@
...
+    client: &mut CompactTxStreamerClient<ChT>,
+    params: &P,
+    fsblockdb_root: &Path,
+    db_cache: &mut FsBlockDb,
```
AArnott, line 38: Why couple this sync implementation to `FsBlockDb`? I can't use it because it's too slow. Can we use a trait for it instead?

AArnott (commented):
Thanks for getting this started. zcash_client_backend will be much better for having its own sync engine in it. This code looks very much like what I saw in str4d's console wallet prototype, which I took and [built up considerably](https://github.com/nerdcash/Nerdbank.Cryptocurrencies/blob/main/src/nerdbank-zcash-rust/src/sync.rs). Would you consider reviewing that implementation to see what you can glean from it to build up your PR?

zcash_client_backend/src/sync.rs
Oscar-Pepper: hi arnott, I am working on a trait here https://github.com/zcash/librustzcash/pull/1192

I also have a branch in my fork which updates this sync engine to build and run with the block cache trait so we dont need to use the SQLite or store blocks on disk

zcash_client_backend/src/sync.rs
str4d: Indeed, this is exactly what the Android and Swift mobile SDKs do. The approach here was intentionally simplistic for the codebase I wrote it for, but I would want this to also be improved to follow the pipelined approach of the mobile SDKs.

zcash_client_backend/Cargo.toml
str4d: This PR is a draft for a reason: I threw it together having extracted it from an _application_, where `anyhow` is perfectly acceptable to use. Part of the work to do to make this ready for review will indeed be to have error types that are useful for this in a library context.

zcash_client_backend/src/sync.rs
str4d: Indeed, this is the TODO I noted at the very top of this PR: https://github.com/zcash/librustzcash/pull/1184#discussion_r1484594940

zcash_client_backend/src/sync.rs
AArnott: I've updated my own sync implementation to be pipelined, limiting resource usage while keeping the CPU busy. I'll be happy to contribute to this PR (or after it) if those changes are welcome.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/c83b443770ac01caa33c2539b715058637cfde76..2d098342dd5c86414e645a8f85ed6816ae756a5e) to use the new `BlockCache` trait. I still need to replace the `anyhow` usage with typed errors, and migrate the previous `FsBlockDb`-specific code into `zcash_client_sqlite`.

zcash_client_backend/src/sync.rs
str4d: As I have said in multiple places, this code is extracted from the existing private code that I have (and that I shared previously with you, on top of which you built your implementation). It is intended as a starting point for further development in-repo, not the perfect sync engine up-front.

So the reason why it stops there is that it was simpler for me to implement initially, and also how my CLI wallet works (which has no long-running backend process). Please move these suggestions into separate issues for future discussion and implementation.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/2d098342dd5c86414e645a8f85ed6816ae756a5e..24277a6ba4013ead9a385acceff6db3de7a0c198) to fix failing tests, and implement an `Error` enum to replace `anyhow` usage.

nuttycom (approved):
utACK. Since this is feature-flagged at this point, I'm happy to go ahead and merge this and have its API revised and improved over time.

PR #1185, opened 2024-02-10
Declare and use NonHardenedChildIndex for transparent

AArnott:
- Declare `NonHardenedChildIndex` struct
- Change `WalletRead::get_transparent_receivers` signature to use the new struct
It needn't return the account id that was given as an input, and it shouldn't return an 11-byte diversifier index when a 31-bit child index is more appropriate.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -52,6 +52,7 @@ and this library adheres to Rust's notion of
   - `wallet::input_selection::ShieldingSelector` has been
     factored out from the `InputSelector` trait to separate out transparent
     functionality and move it behind the `transparent-inputs` feature flag.
+  - `TransparentAddressMetadata` (which replaces `zcash_keys::address::AddressMetadata`).
 - `zcash_client_backend::fees::{standard, sapling}`
 - `zcash_client_backend::fees::ChangeValue::new`
 - `zcash_client_backend::wallet`:
@@ -114,6 +115,7 @@ and this library adheres to Rust's notion of
       been removed from `error::Error`.
     - A new variant `UnsupportedPoolType` has been added.
     - A new variant `NoSupportedReceivers` has been added.
+    - A new variant `NoSpendingKey` has been added.
     - Variant `ChildIndexOutOfRange` has been removed.
   - `wallet::shield_transparent_funds` no longer takes a `memo` argument;
     instead, memos to be associated with the shielded outputs should be
@@ -163,8 +165,8 @@ and this library adheres to Rust's notion of
       `data_api::InputSource` instead.
     - Added `get_account_ids`.
     - `get_transparent_receivers` now returns
-      `zcash_primitives::legacy::NonHardenedChildIndex` instead of
-      `zcash_client_backend::address::AddressMetadata`.
+      `zcash_client_backend::data_api::TransparentAddressMetadata` instead of
+      `zcash_keys::address::AddressMetadata`.
   - `wallet::{propose_shielding, shield_transparent_funds}` now takes their
     `min_confirmations` arguments as `u32` rather than a `NonZeroU32` to permit
     implmentations to enable zero-conf shielding.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,7 +28,6 @@ and this library adheres to Rust's notion of
   - Arguments to `ScannedBlock::from_parts` have changed.
   - Changes to the `WalletRead` trait:
     - Added `get_orchard_nullifiers`
-    - Changed `get_transparent_receivers` return type.
   - `ShieldedProtocol` has a new `Orchard` variant.
 - `zcash_client_backend::fees`:
   - Arguments to `ChangeStrategy::compute_balance` have changed.
@@ -115,6 +114,7 @@ and this library adheres to Rust's notion of
       been removed from `error::Error`.
     - A new variant `UnsupportedPoolType` has been added.
     - A new variant `NoSupportedReceivers` has been added.
+    - Variant `ChildIndexOutOfRange` has been removed.
   - `wallet::shield_transparent_funds` no longer takes a `memo` argument;
     instead, memos to be associated with the shielded outputs should be
     specified in the construction of the value of the `input_selector`
@@ -162,6 +162,9 @@ and this library adheres to Rust's notion of
       `get_unspent_transparent_outputs` have been removed; use
       `data_api::InputSource` instead.
     - Added `get_account_ids`.
+    - `get_transparent_receivers` now returns
+      `zcash_primitives::legacy::NonHardenedChildIndex` instead of
+      `zcash_client_backend::address::AddressMetadata`.
   - `wallet::{propose_shielding, shield_transparent_funds}` now takes their
     `min_confirmations` arguments as `u32` rather than a `NonZeroU32` to permit
     implmentations to enable zero-conf shielding.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,6 +28,7 @@ and this library adheres to Rust's notion of
   - Arguments to `ScannedBlock::from_parts` have changed.
   - Changes to the `WalletRead` trait:
     - Added `get_orchard_nullifiers`
+    - Changed `get_transparent_receivers` return type.
   - `ShieldedProtocol` has a new `Orchard` variant.
 - `zcash_client_backend::fees`:
   - Arguments to `ChangeStrategy::compute_balance` have changed.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -56,6 +56,30 @@ pub enum NullifierQuery {
     All,
 }
 
+/// Describes the derivation of a transparent address.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TransparentAddressMetadata {
+    scope: zip32::Scope,
+    address_index: NonHardenedChildIndex,
+}
+
+impl TransparentAddressMetadata {
+    pub fn new(scope: zip32::Scope, address_index: NonHardenedChildIndex) -> Self {
+        Self {
+            scope,
+            address_index,
+        }
+    }
+
+    pub fn scope(&self) -> zip32::Scope {
+        self.scope
+    }
+
+    pub fn address_index(&self) -> NonHardenedChildIndex {
+        self.address_index
+    }
+}
+
 /// Balance information for a value within a single pool in an account.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Balance {
@@ -560,7 +584,7 @@ pub trait WalletRead {
     fn get_transparent_receivers(
         &self,
         account: AccountId,
-    ) -> Result<HashMap<TransparentAddress, NonHardenedChildIndex>, Self::Error>;
+    ) -> Result<HashMap<TransparentAddress, Option<TransparentAddressMetadata>>, Self::Error>;
 
     /// Returns a mapping from transparent receiver to not-yet-shielded UTXO balance,
     /// for each address associated with a nonzero balance.
@@ -1109,7 +1133,7 @@ pub mod testing {
     use zcash_primitives::{
         block::BlockHash,
         consensus::{BlockHeight, Network},
-        legacy::{NonHardenedChildIndex, TransparentAddress},
+        legacy::TransparentAddress,
         memo::Memo,
         transaction::{components::Amount, Transaction, TxId},
         zip32::{AccountId, Scope},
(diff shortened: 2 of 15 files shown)


AArnott:
I don't know why the new type isn't found by `cargo check`. It works in the rust language server, and my own inspection of the code doesn't lead to any explanation. Any ideas?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -380,16 +381,17 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
             })?;
 
         if let Some(taddr) = ua.transparent() {
-            ret.insert(
-                *taddr,
-                AddressMetadata::new(account, DiversifierIndex::from(di_be)),
-            );
+            let di_short = DiversifierIndex::from(di).try_into();
```
daira, line 386:
```suggestion
            let di_short: Result<u32, _> = DiversifierIndex::from(di).try_into();

File: zcash_client_sqlite/src/wallet.rs [hunk 2]
```diff
@@ -380,16 +381,17 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
...
+            if let Ok(di_short) = di_short {
+                if let Some(index) = NonHardenedChildIndex::from_index(di_short) {
+                    ret.insert(*taddr, index);
+                }
+            }
```
nuttycom, line 391: These two `if let` expressions could silently hide database corruption; neither of them should fail if the transparent address exists, so rather than just filtering out addresses for which there is no result, a failure here must return an error.

Also, I think it would be useful to provide a `TryFrom<DiversifierIndex> for NonHardenedChildIndex` so that you don't have to plumb through the u32 manually here.

nuttycom: Or, just use the `to_transparent_child_index` function from https://github.com/zcash/librustzcash/pull/1185/commits/64cd5444af771ea7da91dcbea5d99ac65221e172#diff-190066eb72ebab2517bed56c3fb391dd4ac2b10d9a3283615d57b667f5a784c4R72

zcash_primitives/src/legacy/keys.rs
```diff
@@ -187,30 +185,31 @@ pub trait IncomingViewingKey: private::SealedChangeLevelKey + std::marker::Sized
...
+    fn default_address(&self) -> (TransparentAddress, NonHardenedChildIndex) {
+        let mut child_index = NonHardenedChildIndex::ZERO;
+        loop {
             match self.derive_address(child_index) {
                 Ok(addr) => {
                     return (addr, child_index);
                 }
                 Err(_) => {
-                    child_index += 1;
+                    child_index = child_index.next().unwrap_or_else(|| {
```
nuttycom, line 206: :+1:

zcash_primitives/src/legacy/keys.rs
```diff
@@ -320,6 +321,27 @@ impl NonHardenedChildIndex {
     pub fn index(&self) -> u32 {
         self.0
     }
+
+    pub fn next(&self) -> Option<Self> {
+        Self::from_index(self.0 + 1)
```
daira:
```suggestion
        // self.0 + 1 cannot overflow as a u32
        Self::from_index(self.0 + 1)
```

nuttycom: It would be useful to have a comment here to note that `self.0 + 1` doesn't need to use `checked_add` because this addition can't overflow, and would instead flow into the hardened range, which is then excluded by `Self::from_index`. Just to help casual inspection.

zcash_client_backend/src/data_api.rs
nuttycom, line 1: Since this is fixing a compilation error, I'd prefer that it would be squashed into the previous commit; otherwise it's not clear exactly what was causing the compilation failure.

zcash_primitives/src/legacy.rs
```diff
@@ -561,4 +563,38 @@ mod tests {
...
+            NonHardenedChildIndex::ZERO,
+            NonHardenedChildIndex::ZERO
+        ));
+        assert!(!check(
+            NonHardenedChildIndex::ZERO,
+            NonHardenedChildIndex::ZERO.next().unwrap()
+        ));
+
+        fn check<T: ConstantTimeEq>(v1: T, v2: T) -> bool {
+            v1.ct_eq(&v2).unwrap_u8() == 1
```
daira, line 588:
```suggestion
            v1.ct_eq(&v2).into()
```

nuttycom (changes requested):
Overall this looks excellent, thank you very much for the contribution! The only blocking comment is for the error handling in `zcash_client_sqlite`.

zcash_client_backend/src/data_api.rs
AArnott: Squashing will complicate my other PR that builds on this, but OK.

zcash_client_sqlite/src/wallet.rs
AArnott: If I read this correctly, it's just adding a type annotation that matches what rust was already inferring, is that right? If so, what's your general guidance for when to add optional type annotations?

zcash_client_sqlite/src/wallet.rs
AArnott: I thought of reusing that fn, but it isn't public.

zcash_client_sqlite/src/wallet.rs
AArnott: This code was rewritten anyway.

AArnott:
All comments addressed. Thanks for reviewing.

nuttycom (approved):
utACK 9f221f869d296ab18ed65ebf942e7b9e98681020, thank you!

zcash_primitives/src/legacy/keys.rs
```diff
@@ -50,11 +48,11 @@ impl AccountPrivKey {
     /// `m/44'/<coin_type>'/<account>'/0/<child_index>`.
     pub fn derive_external_secret_key(
         &self,
-        child_index: u32,
+        child_index: NonHardenedChildIndex,
```
str4d, line 51: Change to a public API needs to be documented in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -560,7 +560,7 @@ pub trait WalletRead {
     fn get_transparent_receivers(
         &self,
         account: AccountId,
-    ) -> Result<HashMap<TransparentAddress, AddressMetadata>, Self::Error>;
+    ) -> Result<HashMap<TransparentAddress, NonHardenedChildIndex>, Self::Error>;
```
str4d, line 563: If all we wanted to return from this API is the transparent address and the child index it was derived from, then I argue that `NonHardenedChildIndex` should be the key: it is trivial to derive the address from the full viewing key given the index (and providing the address is thus only an optimization), whereas the reverse requires brute forcing, so logically the child index is the unique key here.

However, given the existing usage of this API, the child index is being treated as metadata, so we should keep `AddressMetadata` as the return type here. This would allow for e.g. marking whether a receiver is external or internal (currently this API only returns external receivers because `zcash_client_backend` wallets never create transparent change, but if a third party wallet seed phrase is imported that did, we'd need to support scanning at least the default internal transparent receiver to detect it). If we keep `AddressMetadata` removed then in future we'd also need to change the map to be `HashMap<(zip32::Scope, NonHardenedChildIndex), TransparentAddress>`.

Separately, I note that `AddressMetadata` is currently only used for transparent addresses, and is really a detail of this specific `WalletRead` method. It should probably not have been moved into `zcash_keys`, and should be moved back into `zcash_client_backend` (and maybe renamed to be specific to transparent addresses, since we **do not** expose internal addresses for shielded keys).

---

After a rather deep rabbit-hole session with @nuttycom, our conclusion is we should:
- Move `AddressMetadata` back into `zcash_client_backend::address`.
- Change `AddressMetadata` to replace `diversifier_index` with `NonHardenedChildIndex`, remove `account_id`, and add `zip32::Scope`.

These changes don't need to happen in this PR, however this PR _should_ be changed to not remove `AddressMetadata` from the return type of this method. So if you do want to make the other changes here, you are welcome to.

str4d (changes requested):
Reviewed as of 9f221f869d296ab18ed65ebf942e7b9e98681020. The changelog changes are non-blocking (we can fix them in a separate PR before the crate releases, if they aren't fixed here).

zcash_client_backend/src/data_api.rs
nuttycom: More specifically, we realized as part of the discussion that there is probably an existing bug related to the import of seeds from transparent-only wallets, where we are not at present scanning any of the internal key tree (which is where Bitcoin wallets regularly send change) so if you import a seed that was previously used with a transparent-only wallet, you're likely to be missing funds.

Then again, if you do that, then you're also not doing gap limit scanning at the moment, so _both_ of these should be addressed at once. I'll open an issue for that and it can get stuck in the backlog somewhere.

zcash_client_sqlite/src/wallet.rs
daira: Yes the suggestion was for what Rust was already inferring. To answer your question, I typically add type annotations when:
* inferring the type would require a non-local search (in this case, for implementors of `From<DiversifierIndex>`); *and*
* the type is useful to know in order to understand the code.

Uses of `into()`, `try_into()`, and `collect()` are particularly common cases (but not all of them).

zcash_client_sqlite/src/wallet.rs
daira: You can make it `pub(crate)` if you need to.

zcash_client_backend/src/data_api.rs
daira: Please add an issue for that @nuttycom.

daira (dismissed):
9f221f869d296ab18ed65ebf942e7b9e98681020 looks good but @nuttycom and @str4d's other comments are blocking.

zcash_client_backend/src/data_api.rs
daira: Is it all of the metadata that may be known or unknown, or is it each field of `AddressMetadata` somewhat independently? If the former then use `Option<AddressMetadata>`, otherwise make the relevant accessors on `AddressMetadata` return `Option`s.

zcash_client_backend/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this library adheres to Rust's notion of
   - Arguments to `ScannedBlock::from_parts` have changed.
   - Changes to the `WalletRead` trait:
     - Added `get_orchard_nullifiers`
+    - Changed `get_transparent_receivers` return type.
```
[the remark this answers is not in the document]
AArnott: Done.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -70,9 +70,6 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError> {
 
     #[cfg(feature = "transparent-inputs")]
     AddressNotRecognized(TransparentAddress),
-
-    #[cfg(feature = "transparent-inputs")]
-    ChildIndexOutOfRange(DiversifierIndex),
```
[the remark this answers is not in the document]
AArnott: Done.

zcash_primitives/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 ### Added
+- `legacy::keys::NonHardenedChildIndex`
```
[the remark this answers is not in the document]
AArnott: Fixed.

zcash_client_backend/src/data_api.rs
AArnott: Addressed. Please take another look.
I left the key=value order as it was because in the future when we support importing transparent private keys, we won't have derivation data.

nuttycom (approved):
utACK f1b6dd0636abb8aee2e66e75e04da164dd4a077f

It turns out that I need this to address problems in #1187 so I have marked this as blocking that PR.

zcash_client_backend/src/data_api.rs
str4d: > Although now that I think about it, I guess we're not even guaranteed to know the `NonHardenedChildIndex`.

If we can't even guarantee that, then we cannot spend any funds received to that address, because:

> I left the key=value order as it was because in the future when we support importing transparent private keys, we won't have derivation data.

We **do not** import raw transparent spending keys; everything needs to be at a minimum tied back to a transparent full viewing key (which we define in ZIP 316 as a particular node in the BIP 44 key derivation path).

Importing arbitrary non-HD transparent private keys will break a bunch of assumptions the current wallet architecture makes, and no one has committed to supporting and maintaining that use case in the long term. From a user perspective, it is only necessary in the context of a full node wallet that explicitly supports importing an arbitrary `zcashd` wallet, and even in that case we should preferably instead have a migration pathway away from those transparent keys. Transparent-only mobile wallets have for basically all of Zcash's existence used HD-derived keys.

str4d: While I am strongly opposed to trait changes that point us in the direction of adding support for long-since-deprecated `zcashd` behaviours, I am not going to block this PR being merged. I have not fully re-reviewed the PR after the recent changes, but the trait returning `Option<Metadata>` is the least intrusive thing that could be done, and I will almost certainly just be `.unwrap()`ing everywhere downstream.

nuttycom:
I will fix the formatting errors in #1187

PR #1186, opened 2024-02-11
Use mermaid for migrations graph

AArnott:
The migrations ASCII art previously used was somewhat hard to read and much harder to maintain over time. Mermaid markdown provides a very easy way to add new nodes to the graph, and the result is far more visually parseable.

Mermaid can be rendered on GitHub or via a local extension such as in VS Code. I've added an extension recommendation for VS Code as part of this change.

Here is what it looks like in VS Code:
image

And right here in the PR, github will render the graphic with the click of a button on the introduced file.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+    "recommendations": [
+        "bierner.markdown-mermaid"
+    ]
+}
--- a/zcash_client_sqlite/src/wallet/init/migrations.md
+++ b/zcash_client_sqlite/src/wallet/init/migrations.md
@@ -1,0 +1,44 @@
+This graph can be rendered on github.com or with the `bierner.markdown-mermaid` extension in VS Code.
+
+```mermaid
+graph TD;
+    initial_setup-->utxos_table;
+
+    initial_setup-->ufvk_support;
+
+    ufvk_support-->addresses_table;
+
+    ufvk_support-->sent_notes_to_internal;
+
+    utxos_table-->add_utxo_account;
+    addresses_table-->add_utxo_account;
+
+    sent_notes_to_internal-->add_transaction_views;
+    add_utxo_account-->add_transaction_views;
+
+    add_transaction_views-->v_transactions_net;
+
+    v_transactions_net-->received_notes_nullable_nf;
+
+    received_notes_nullable_nf-->shardtree_support;
+
+    received_notes_nullable_nf-->nullifier_map;
+
+    received_notes_nullable_nf-->sapling_memo_consistency;
+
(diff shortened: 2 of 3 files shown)


AArnott:
Other ideas I have for improving on this is to make it more clear _which_ nodes I need to derive on when I change a particular database entity. So maybe elsewhere in the new markdown file, a list of each db entity (e.g. table, view, index) with a sub-list of each migration that impacted it, and a description of what the migration did to it. That way if I need to change an entity, I can quickly find the list of dependencies I'll need. Then by referring to the graph, I can weed out the redundant dependencies.

nuttycom:
Can `mermaid` render to a textual DAG? I find it really helpful to have the graph in `migrations.rs`, but I would not at all object to doing the rendering with a tool and then copy/pasting in the result.

AArnott:
@nuttycom: Copilot says no.

> [Mermaid is a JavaScript-based diagramming and charting tool that uses Markdown-inspired text definitions to create and modify complex diagrams](https://github.com/mermaid-js/mermaid)². However, it does not inherently support rendering to a textual Directed Acyclic Graph (DAG). Mermaid's primary function is to render graphical representations of diagrams based on the provided textual input².

>
> If you want to convert a Mermaid diagram to a textual DAG, you would need to implement a custom solution. This could involve parsing the Mermaid syntax and converting it into a textual representation of a DAG. Please note that this would require a good understanding of both the Mermaid syntax and the desired textual format for the DAG. 

>
> If you're looking for a tool that can generate textual DAGs, you might want to consider other libraries or tools that are designed for this purpose. Always choose the tool that best fits your specific needs.

([1](https://github.com/mermaid-js/mermaid)) GitHub - mermaid-js/mermaid: Generation of diagrams like flowcharts or .... https://github.com/mermaid-js/mermaid.
(2) How to render a mermaid flowchart dynamically? - Stack Overflow. https://stackoverflow.com/questions/65212332/how-to-render-a-mermaid-flowchart-dynamically.
(3) Include diagrams in your Markdown files with Mermaid. https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/.
(4) undefined. https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js.

AArnott:
I'd rather see only the .md file. But given @nuttycom  prefers the ascii approach, what if I modify the PR to retain the ASCII graphic and simply add the .md file? That way folks who prefer to refer the image can do so, at the cost of maintaining two sources of truth, although the .md file is much easier to maintain than the .rs comments, so it doesn't add a lot of effort to have to add maintenance of the .md file.

PR #1187, opened 2024-02-12
`zcash_client_backend`: Updates to the `Proposal` data structure

nuttycom:
Best reviewed commit-by-commit and with whitespace changes hidden.

* Add the chosen output pool for each payment to the proposal data.
* Move proposal-related types to their own module
* Make proposals able to represent multiple linked transactions, for ZIP 320

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,3 +115,6 @@ zip32 = "0.1"
 lto = true
 panic = 'abort'
 codegen-units = 1
+
+[patch.crates-io]
+sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto", rev = "346a6507413eb2388b2967c014672cd83199648a" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ bitvec = "1"
 blake2s_simd = "1"
 bls12_381 = "0.8"
 jubjub = "0.10"
-sapling = { package = "sapling-crypto", version = "0.1" }
+sapling = { package = "sapling-crypto", version = "0.1.1" }
 
 # - Orchard
 nonempty = "0.7"
@@ -115,6 +115,3 @@ zip32 = "0.1"
 lto = true
 panic = 'abort'
 codegen-units = 1
-
-[patch.crates-io]
-sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto", rev = "346a6507413eb2388b2967c014672cd83199648a" }
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -35,6 +35,7 @@ and this library adheres to Rust's notion of
 ## [0.11.0-pre-release] Unreleased
 
 ### Added
+- `zcash_client_backend::PoolType::is_receiver`
 - `zcash_client_backend::data_api`:
   - `InputSource`
   - `ScannedBlock::{into_commitments, sapling}`
@@ -206,6 +207,11 @@ and this library adheres to Rust's notion of
     it possible to represent multi-step transactions, such as a deshielding
     transaction followed by a zero-conf transfer as required by ZIP 320. Individual
     transaction proposals are now represented by the `proposal::Step` type.
+  - `ProposalError` has new variants:
+    - `ReferenceError`
+    - `StepDoubleSpend`
+    - `ChainDoubleSpend`
+    - `PaymentPoolsMismatch`
 
 - `zcash_client_backend::fees`:
   - `ChangeStrategy::compute_balance` arguments have changed.
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -47,7 +47,6 @@ and this library adheres to Rust's notion of
     }`
   - `WalletSummary::next_sapling_subtree_index`
   - `wallet::propose_standard_transfer_to_address`
-  - `wallet::input_selection::Proposal::{from_parts, shielded_inputs, payment_pools}`
   - `wallet::input_selection::ShieldedInputs`
   - `wallet::input_selection::ShieldingSelector` has been
     factored out from the `InputSelector` trait to separate out transparent
@@ -59,6 +58,10 @@ and this library adheres to Rust's notion of
   - `ReceivedNote`
   - `WalletSaplingOutput::recipient_key_scope`
   - `wallet::TransparentAddressMetadata` (which replaces `zcash_keys::address::AddressMetadata`).
+- `zcash_client_backend::zip321::TransactionRequest::total`
+- `zcash_client_backend::zip321::parse::Param::name`
+- `zcash_client_backend::proposal`
+  - `Proposal::{from_parts, shielded_inputs, payment_pools}`
 - `zcash_client_backend::proto::`
   - `PROPOSAL_SER_V1`
   - `ProposalDecodingError`
@@ -66,12 +69,12 @@ and this library adheres to Rust's notion of
 - `impl Clone for zcash_client_backend::{
      zip321::{Payment, TransactionRequest, Zip321Error, parse::Param, parse::IndexedParam},
      wallet::{ReceivedSaplingNote, WalletTransparentOutput},
-     wallet::input_selection::{Proposal, SaplingInputs},
+     proposal::{Proposal, SaplingInputs},
    }`
 - `impl {PartialEq, Eq} for zcash_client_backend::{
      zip321::{Zip321Error, parse::Param, parse::IndexedParam},
      wallet::{ReceivedSaplingNote, WalletTransparentOutput},
-     wallet::input_selection::{Proposal, SaplingInputs},
+     proposal::{Proposal, SaplingInputs},
    }`
 - `zcash_client_backend::zip321
   ` `TransactionRequest::{total, from_indexed}`
@@ -85,6 +88,8 @@ and this library adheres to Rust's notion of
 - `ScannedBlock::{sapling_tree_size, sapling_nullifier_map, sapling_commitments}`
   have been moved to `ScannedBlockSapling` and in that context are now
   named `{tree_size, nullifier_map, commitments}` respectively.
+- `zcash_client_backend::::wallet::input_selection::{Proposal, ShieldedInputs, ProposalError}`
+  have been moved to `zcash_client_backend::proposal`.
 
 ### Changed
 - `zcash_client_backend::data_api`:
@@ -115,6 +120,7 @@ and this library adheres to Rust's notion of
     - A new variant `UnsupportedPoolType` has been added.
     - A new variant `NoSupportedReceivers` has been added.
     - A new variant `NoSpendingKey` has been added.
+    - A new variant `Proposal` has been added.
     - Variant `ChildIndexOutOfRange` has been removed.
   - `wallet::shield_transparent_funds` no longer takes a `memo` argument;
     instead, memos to be associated with the shielded outputs should be
@@ -148,6 +154,7 @@ and this library adheres to Rust's notion of
       `min_confirmations` argument as `u32` instead of `NonZeroU32`.
   - The `wallet::input_selection::InputSelector::DataSource`
     associated type has been renamed to `InputSource`.
+  - `wallet::input_selection::InputSelectorError` has added variant `Proposal`
   - The signature of `wallet:input_selection::InputSelector::propose_transaction`
     has been altered such that it longer takes `min_confirmations` as an
     argument, instead taking explicit `target_height` and `anchor_height`
@@ -175,11 +182,14 @@ and this library adheres to Rust's notion of
   - `wallet::create_proposed_transaction` now forces implementations to ignore
     the database identifiers for its contained notes by universally quantifying
     the `NoteRef` type parameter.
-  - Arguments to `wallet::input_selection::Proposal::from_parts` have changed.
-  - `wallet::input_selection::Proposal::min_anchor_height` has been removed in
-    favor of storing this value in `SaplingInputs`.
   - `wallet::input_selection::GreedyInputSelector` now has relaxed requirements
     for its `InputSource` associated type.
+
+- `zcash_client_backend::proposal`:
+  - Arguments to `Proposal::from_parts` have changed.
+  - `Proposal::min_anchor_height` has been removed in favor of storing this
+    value in `SaplingInputs`.
+  - `Proposal::sapling_inputs` has been replaced by `Proposal::shielded_inputs`
 
 - `zcash_client_backend::fees`:
   - `ChangeStrategy::compute_balance` arguments have changed.
@@ -230,8 +240,7 @@ and this library adheres to Rust's notion of
 ### Removed
 - `zcash_client_backend::wallet::ReceivedSaplingNote` has been replaced by
   `zcash_client_backend::ReceivedNote`.
-- `zcash_client_backend::wallet::input_selection::Proposal::sapling_inputs` has
-  been replaced by `Proposal::shielded_inputs`
+- `zcash_client_backend::wallet::input_selection::Proposal`
 - `zcash_client_backend::data_api`
 - `zcash_client_backend::data_api::ScannedBlock::from_parts` has been made crate-private.
 - `zcash_client_backend::data_api::ScannedBlock::into_sapling_commitments` has been
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -47,7 +47,7 @@ and this library adheres to Rust's notion of
     }`
   - `WalletSummary::next_sapling_subtree_index`
   - `wallet::propose_standard_transfer_to_address`
-  - `wallet::input_selection::Proposal::{from_parts, shielded_inputs}`
+  - `wallet::input_selection::Proposal::{from_parts, shielded_inputs, payment_pools}`
   - `wallet::input_selection::ShieldedInputs`
   - `wallet::input_selection::ShieldingSelector` has been
     factored out from the `InputSelector` trait to separate out transparent
@@ -59,8 +59,6 @@ and this library adheres to Rust's notion of
   - `ReceivedNote`
   - `WalletSaplingOutput::recipient_key_scope`
   - `wallet::TransparentAddressMetadata` (which replaces `zcash_keys::address::AddressMetadata`).
-- `zcash_client_backend::zip321::TransactionRequest::total`
-- `zcash_client_backend::zip321::parse::Param::name`
 - `zcash_client_backend::proto::`
   - `PROPOSAL_SER_V1`
   - `ProposalDecodingError`
@@ -75,8 +73,9 @@ and this library adheres to Rust's notion of
      wallet::{ReceivedSaplingNote, WalletTransparentOutput},
      wallet::input_selection::{Proposal, SaplingInputs},
    }`
-- `zcash_client_backend::zip321::to_uri` now returns a `String` rather than an
-  `Option<String>` and provides canonical serialization for the empty proposal.
+- `zcash_client_backend::zip321
+  ` `TransactionRequest::{total, from_indexed}`
+  - `parse::Param::name`
 
 ### Moved
 - `zcash_client_backend::data_api::{PoolType, ShieldedProtocol}` have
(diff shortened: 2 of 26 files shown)


zcash_client_backend/src/proposal.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Returns the map from payment index to the pool that has been selected
+    /// for the output that will fulfill that payment.
+    pub fn payment_pools(&self) -> &BTreeMap<usize, PoolType> {
+        &self.payment_pools
+    }
+    /// Returns the transparent inputs that have been selected to fund the transaction.
+    pub fn transparent_inputs(&self) -> &[WalletTransparentOutput] {
+        &self.transparent_inputs
+    }
+    /// Returns the Sapling inputs that have been selected to fund the transaction.
```
daira, line 301: How will this change for Orchard?

zcash_client_backend/src/proposal.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Returns the target height for which the proposal was prepared.
+    ///
+    /// The chain must contain at least this many blocks in order for the proposal to
+    /// be executed.
+    pub fn min_target_height(&self) -> BlockHeight {
+        self.min_target_height
+    }
+    /// Returns a flag indicating whether or not the proposed transaction
+    /// is exclusively wallet-internal (if it does not involve any external
+    /// recipients).
```
daira, line 311: While it may be the case currently that only shielding transactions are fully internal, that is not true by definition (and not likely to remain true if we implement note management). Please decide which is meant and make the documentation and the method name consistent.

zcash_client_backend/CHANGELOG.md
```diff
@@ -121,23 +123,32 @@ and this library adheres to Rust's notion of
...
-    internal "change" outputs.
-  - `wallet::create_proposed_transaction` no longer takes a
-    `change_memo` argument; instead, change memos are represented in the
-    individual values of the `proposed_change` field of the `Proposal`'s
-    `TransactionBalance`.
-  - `wallet::create_proposed_transaction` now takes its `proposal` argument
-    by reference instead of as an owned value.
-  - `wallet::create_proposed_transaction` no longer takes a `min_confirmations`
-    argument. Instead, it uses the anchor height from its `proposal` argument.
+    internal "change" outputs. Also, it returns its result as a `Vec<TxId>`
```
daira, line 126: How about a NonEmptyVec?

daira, line 126: How about a `NonEmpty` vec?

daira (approved):
utACK

zcash_client_backend/src/proposal.rs
nuttycom: Change is also fully wallet-internal, but in the case of change we don't have any confusion about which OVK should be used; it's only for shielding transactions that we need to pick the correct OVK. But I agree that this comment can be better.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -521,14 +521,59 @@ where
...
+where
+    DbT: WalletWrite + WalletCommitmentTrees,
+    ParamsT: consensus::Parameters + Clone,
+    FeeRuleT: FeeRule,
+{
+    proposal
+        .steps()
+        .iter()
+        .map(|step| {
+            create_proposed_transaction(
```
str4d, line 550: This will need to be altered once the `Proposal` protobuf is fixed to allow chained transactions, because as-implemented the proposal steps are not steps, but are completely independent set elements that can be created in arbitrary order.

zcash_client_backend/src/proposal.rs
```diff
@@ -98,21 +98,127 @@ impl<NoteRef> ShieldedInputs<NoteRef> {
...
+    /// Parameters:
+    /// * `fee_rule`: The fee rule observed by the proposed transaction.
+    /// * `min_target_height`: The minimum block height at which the transaction may be created.
+    /// * `steps`: A vector of steps that make up the proposal.
+    pub fn multi_step(
+        fee_rule: FeeRuleT,
+        min_target_height: BlockHeight,
+        steps: Vec<Step<NoteRef>>,
+    ) -> Result<Self, ProposalError> {
+        // TODO: actually perform the validation described in the documentation.
```
str4d, line 129: If being left as a TODO, please open a downstream issue for this that blocks the appropriate release.

zcash_client_backend/src/proposal.rs
```diff
@@ -98,21 +98,127 @@ impl<NoteRef> ShieldedInputs<NoteRef> {
...
+    }
+}
+
+impl<FeeRuleT: Debug, NoteRef> Debug for Proposal<FeeRuleT, NoteRef> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Proposal")
+            .field("fee_rule", &self.fee_rule)
+            .field("min_target_height", &self.min_target_height)
+            .field("steps", &self.steps)
+            .finish_non_exhaustive()
```
str4d, line 205: This is not non-exhaustive (for `Proposal`) as all of its fields are included in the `Debug` output:
```suggestion
            .finish()
```
Also, I note that `fee_rule` was not included in the previous `Debug` output for `Proposal`; why was that? (Or alternatively, why is it now being included, which is what makes this impl now not non-exhaustive?)

zcash_client_backend/src/proto.rs
```diff
@@ -359,72 +359,84 @@ impl proposal::Proposal {
...
+                            value_pool: proposal::ValuePool::from(change.output_pool()).into(),
+                            memo: change.memo().map(|memo_bytes| proposal::MemoBytes {
+                                value: memo_bytes.as_slice().to_vec(),
+                            }),
+                        })
+                        .collect(),
+                    fee_required: step.balance().fee_required().into(),
+                });
+
+                #[allow(deprecated)]
```
str4d, line 417: Is this annotation needed on `ProposalStep`? I think it was only here because of the non-ZIP 317 fee rules, which are not on `ProposalStep`.

zcash_client_backend/src/wallet.rs
```diff
@@ -375,6 +375,7 @@ impl<NoteRef> orchard_fees::InputView<NoteRef> for ReceivedNote<NoteRef, orchard
 /// viewing key, refer to [ZIP 310].
 ///
 /// [ZIP 310]: https://zips.z.cash/zip-0310
+#[derive(Debug, Clone)]
```
str4d, line 378: Nit: should document as an addition.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -316,7 +316,7 @@ pub fn spend<DbT, ParamsT, InputsT>(
     ovk_policy: OvkPolicy,
     min_confirmations: NonZeroU32,
 ) -> Result<
-    TxId,
+    Vec<TxId>,
```
str4d, line 319: Update the documentation for this function to not assume a single transaction is created.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -207,7 +207,7 @@ pub fn create_spend_to_address<DbT, ParamsT>(
     min_confirmations: NonZeroU32,
     change_memo: Option<MemoBytes>,
 ) -> Result<
-    TxId,
+    Vec<TxId>,
```
str4d, line 210: Update the documentation for this function to not assume a single transaction is created.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -877,7 +919,7 @@ pub fn shield_transparent_funds<DbT, ParamsT, InputsT>(
     from_addrs: &[TransparentAddress],
     min_confirmations: u32,
 ) -> Result<
-    TxId,
+    Vec<TxId>,
```
str4d, line 922: Update the documentation for this function to not assume a single transaction is created.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -525,46 +548,46 @@ where
         let mut sapling_outputs = vec![];
         #[cfg(feature = "orchard")]
         let mut orchard_outputs = vec![];
-        for payment in transaction_request.payments() {
-            let mut push_transparent = |taddr: TransparentAddress| {
```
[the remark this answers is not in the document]
nuttycom: This is to reduce conflicts with others of my open PRs.

zcash_client_backend/src/zip321.rs
```diff
@@ -275,12 +250,14 @@ impl TransactionRequest {
...
-            [payment] => {
-                let query_params = payment_params(payment, None)
-                    .into_iter()
-                    .collect::<Vec<String>>();
+        match self.payments.len() {
+            0 => "zcash:".to_string(),
+            1 => {
+                let (idx, payment) = self.payments.iter().next().unwrap();
+                let query_params =
+                    payment_params(payment, if *idx == 0 { None } else { Some(*idx) })
```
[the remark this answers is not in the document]
nuttycom: I think the correct fix here is to check the index in a guard on the match, and only apply this if the index is zero.

zcash_client_backend/proto/proposal.proto
```diff
@@ -5,34 +5,41 @@
...
     uint32 protoVersion = 1;
+    // The fee rule used in constructing this proposal
+    FeeRule feeRule = 2;
+    // The target height for which the proposal was constructed
+    //
+    // The chain must contain at least this many blocks in order for the proposal to
+    // be executed.
+    uint32 minTargetHeight = 3;
+    // The series of transactions to be created.
+    repeated ProposalStep steps = 4;
```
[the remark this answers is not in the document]
nuttycom: The validity of a series of proposal steps will now be checked via the inter-step references. Since different wallets may support different subsets of inter-step references (for example, `zcash_client_backend` will support only transparent inter-step references after this PR) I don't think it makes sense to add restrictions at the encoding layer.

zcash_client_backend/proto/proposal.proto
```diff
@@ -5,34 +5,41 @@
 syntax = "proto3";
 package cash.z.wallet.sdk.ffi;
 
-// A data structure that describes the inputs to be consumed and outputs to
-// be produced in a proposed transaction.
+// A data structure that describes the a series of transactions to be created.
 message Proposal {
+    // The version of this serialization format.
     uint32 protoVersion = 1;
```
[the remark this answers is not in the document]
nuttycom: We haven't actually published a `zcash_client_backend` release with `proposal.proto`, and in the Zeboot tag we explicitly decided that this format wasn't stable, so I don't think we need to change this. The Rust `Proposal` type has been in the public API before now, but this is the first actual release with a stable protobuf encoding.

zcash_client_backend/proto/proposal.proto
```diff
@@ -5,34 +5,41 @@
...
-    repeated PaymentOutputPool paymentOutputPools = 3;
+    repeated PaymentOutputPool paymentOutputPools = 2;
     // The anchor height to be used in creating the transaction, if any.
     // Setting the anchor height to zero will disallow the use of any shielded
     // inputs.
-    uint32 anchorHeight = 4;
+    uint32 anchorHeight = 3;
     // The inputs to be used in creating the transaction.
-    repeated ProposedInput inputs = 5;
+    repeated ProposedInput inputs = 4;
```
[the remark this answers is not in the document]
nuttycom: This has now been fixed.

zcash_client_backend/src/proposal.rs
nuttycom: I don't know why it wasn't included; it's simple enough to add the `Debug` constraint and I don't see any downside to requiring that. It was probably just an oversight. And for a debug impl, including the fee rule is useful information.

zcash_client_backend/CHANGELOG.md
```diff
@@ -52,13 +52,13 @@ and this library adheres to Rust's notion of
...
     factored out from the `InputSelector` trait to separate out transparent
     functionality and move it behind the `transparent-inputs` feature flag.
-  - `TransparentAddressMetadata` (which replaces `zcash_keys::address::AddressMetadata`).
 - `zcash_client_backend::fees::{standard, sapling}`
 - `zcash_client_backend::fees::ChangeValue::new`
 - `zcash_client_backend::wallet`:
   - `Note`
   - `ReceivedNote`
   - `WalletSaplingOutput::recipient_key_scope`
+  - `wallet::TransparentAddressMetadata` (which replaces `zcash_keys::address::AddressMetadata`).
```
AArnott, line 63: You added `wallet::`, but it's under a `::wallet` parent bullet, and is different from the rest in this list.

zcash_client_backend/src/proposal.rs
nuttycom: Fixed in 063c472

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -709,33 +710,93 @@ where
...
+                    let outpoint = OutPoint::new(
+                        result.transaction().txid().into(),
+                        u32::try_from(
+                            prior_step
+                                .payment_pools()
+                                .iter()
+                                .filter(|(_, pool)| pool == &&PoolType::Transparent)
+                                .take_while(|(j, _)| j <= &&i)
+                                .count()
+                                - 1,
```
str4d, line 792: Hmm, I'm not a huge fan of this dead reckoning, given how we randomize shielded output orders and could in theory also randomize transparent output orders, but it's fine for now.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -513,14 +516,16 @@ where
         .map_err(Error::from)
 }
 
-/// Construct, prove, and sign a transaction using the inputs supplied by the given proposal,
-/// and persist it to the wallet database.
+/// Construct, prove, and sign a transaction or series of transactions using the inputs supplied by
+/// the given proposal, and persist it to the wallet database.
 ///
-/// Returns the database identifier for the newly constructed transaction, or an error if
+/// Returns the database identifier for eacy newly constructed transaction, or an error if
```
str4d, line 525:
```suggestion
/// Returns the database identifier for each newly constructed transaction, or an error if
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -513,14 +516,16 @@ where
...
 /// an error occurs in transaction construction, proving, or signing.
 ///
-/// Note: If the payment includes a recipient with an Orchard-only UA, this will attempt
-/// to fall back to the transparent receiver until full Orchard support is implemented.
+/// When evaluating multi-step proposals, only transparent outputs of any given step may be spent
+/// in later steps; attempting to spend a shielded note (including change) output by an earlier
+/// step is not supported, because the ultimate positions of those notes cannot be known and
```
str4d, line 530: "the ultimate positions of those notes in the global commitment tree" (which requires mining, as opposed to their position in the prior transaction).

Cargo.toml
```diff
@@ -115,3 +115,6 @@ zip32 = "0.1"
 lto = true
 panic = 'abort'
 codegen-units = 1
+
+[patch.crates-io]
+sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto", rev = "346a6507413eb2388b2967c014672cd83199648a" }
```
str4d, line 120: Remove this and bump `sapling-crypto` to `0.1.1`.

zcash_client_backend/src/proposal.rs
```diff
@@ -39,6 +40,10 @@ pub enum ProposalError {
     ShieldingInvalid,
     /// A reference to the output of a prior step is invalid.
     ReferenceError(StepOutput),
+    /// An attempted double-spend of a prior step output was detected.
+    StepDoubleSpend(StepOutput),
+    /// An attempted double-spend of an output belonging to the wallet was detected.
+    ChainDoubleSpend(PoolType, TxId, u32),
```
str4d, line 46: These two changes should also go in the changelog.

zcash_client_backend/src/proposal.rs
```diff
@@ -131,7 +148,64 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
...
+                        if prior_step.balance().proposed_change().len() <= idx {
+                            return Err(ProposalError::ReferenceError(*prior_ref));
+                        }
+                    }
+                }
+                // check that there are no double-spends
+                if consumed_prior_inputs.contains(prior_ref) {
+                    return Err(ProposalError::StepDoubleSpend(*prior_ref));
+                }
+                consumed_prior_inputs.insert(*prior_ref);
```
str4d, line 178: Minor optimization:
```suggestion
                if !consumed_prior_inputs.insert(*prior_ref) {
                    return Err(ProposalError::StepDoubleSpend(*prior_ref));
                }
```

zcash_client_backend/src/proposal.rs
```diff
@@ -131,7 +148,64 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
...
+            for t_out in step.transparent_inputs() {
+                let key = (
+                    PoolType::Transparent,
+                    TxId::from_bytes(*t_out.outpoint().hash()),
+                    t_out.outpoint().n(),
+                );
+                if consumed_chain_inputs.contains(&key) {
+                    return Err(ProposalError::ChainDoubleSpend(key.0, key.1, key.2));
+                }
+                consumed_chain_inputs.insert(key);
```
str4d, line 190: Minor optimization:
```suggestion
                if !consumed_chain_inputs.insert(key) {
                    return Err(ProposalError::ChainDoubleSpend(key.0, key.1, key.2));
                }
```

zcash_client_backend/src/proposal.rs
```diff
@@ -131,7 +148,64 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
...
+                        #[cfg(feature = "orchard")]
+                        Note::Orchard(_) => PoolType::Shielded(ShieldedProtocol::Orchard),
+                    },
+                    *s_out.txid(),
+                    s_out.output_index().into(),
+                );
+                if consumed_chain_inputs.contains(&key) {
+                    return Err(ProposalError::ChainDoubleSpend(key.0, key.1, key.2));
+                }
+                consumed_chain_inputs.insert(key);
```
str4d, line 206:
```suggestion
                if !consumed_chain_inputs.insert(key) {
                    return Err(ProposalError::ChainDoubleSpend(key.0, key.1, key.2));
                }
```

str4d (changes requested):
Reviewed as of 6aabe60d2138671435585df445fd70845799fffa. The only blocking change is the validation in `Step::from_parts`.

zcash_client_backend/src/data_api.rs
```diff
@@ -1314,6 +1300,7 @@ pub mod testing {
             Ok(HashMap::new())
         }
 
+        #[cfg(feature = "transparent-inputs")]
         fn get_transparent_balances(
```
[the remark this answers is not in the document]
nuttycom: Yeah, I'm aware, but the hope is that eventually we'll implement this mock as a simple in-memory database, so I left it in.

str4d (approved):
utACK e814cdcad72a76aedfa9506e08f55f1b0580df44

str4d (approved):
Re-utACK 74b487e4c9d6e85ed24b2bd7d3620d8e8206cb51

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -355,8 +357,8 @@ where
     )
 }
 
-/// Select transaction inputs, compute fees, and construct a proposal for a transaction
-/// that can then be authorized and made ready for submission to the network with
+/// Select transaction inputs, compute fees, and construct a proposal for a transaction or series
+/// of transactions that can then be authorized and made ready for submission to the network with
 /// [`create_proposed_transaction`].
```
daira, line 362:
```suggestion
/// [`create_proposed_transactions`].
```
(I know this is already merged.)

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -399,9 +401,13 @@ where
         .map_err(Error::from)
 }
 
-/// Proposes a transaction paying the specified address from the given account.
+/// Proposes making a payment to the specified address from the given account.
 ///
-/// Returns the proposal, which may then be executed using [`create_proposed_transaction`]
+/// Returns the proposal, which may then be executed using [`create_proposed_transaction`].
```
daira, line 406:
```suggestion
/// Returns the proposal, which may then be executed using [`create_proposed_transactions`].
```

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Fixed in #1105

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Fixed in #1105

PR #1192, opened 2024-02-20
Add block cache trait

### LINKED ISSUE #1170: zcash_client_backend: Add cache insertion and deletion APIs to `BlockSource` trait or a new trait
Currently the `BlockSource` trait only exposes a generic way to access cached blocks. However, if we want to provide a generic state machine (#1169), we will need a generic way to perform the following operations:
- Insert downloaded `CompactBlock`s into the cache.
- Mark previously-cached blocks as scanned (to trigger cache removal at some point).
- Remove all cached blocks above some height (for rollbacks).

We might want to do this as part of `BlockSource`, but more likely we will have a new `BlockCache: BlockSource` trait providing these capabilities.

Oscar-Pepper:
Adds a `BlockCache` trait for `BlockSource` which aims to generalise the block source for consumers who are not implementing an SQLite database. This lays some of the groundwork for #1184 by removing cyclic dependencies and generalising the sync engine for wallets with different implementations

Closes #1170


zcash_client_backend/src/data_api/chain.rs
```diff
@@ -213,6 +216,41 @@ pub trait BlockSource {
...
 
+pub trait BlockCache: BlockSource {
+    /// Returns a range of compact blocks from the cache.
+    fn read<WalletErrT>(
+        &self,
+        range: &ScanRange,
+    ) -> Result<Vec<CompactBlock>, error::Error<WalletErrT, Self::Error>>;
+
+    /// Returns the height of highest block known to the block cache, within a specified range.
+    /// If `range` is `None`, returns the tip of the entire cache.
```
zancas, line 227: Can you add a doc-test here?

Oscar-Pepper: good idea

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -213,6 +213,168 @@ pub trait BlockSource {
...
+/// block_cache.truncate(BlockHeight::from_u32(1)).unwrap();
+/// assert_eq!(block_cache.cached_blocks.len(), 1);
+/// assert_eq!(
+///     block_cache.cache_tip(None).unwrap(),
+///     Some(BlockHeight::from_u32(1))
+/// );
+/// ```
+pub trait BlockCache: BlockSource + Send + Sync {
+    /// Returns a range of compact blocks from the cache.
+    fn read(&self, range: &ScanRange) -> Result<Vec<CompactBlock>, Self::Error>;
```
nuttycom, line 354: What are the semantics  that are required if not all blocks in the scan range exist in the cache? Are partial reads allowed, or does that need to be represented as an error?

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.65.0"
+channel = "1.75.0"
```
nuttycom, line 2: We can't update the Rust toolchain without also updating zcashd, this will need to be reverted.

fluidvanadium: that sounds like a quest, my good fellow!

Oscar-Pepper: OK I think I can change the async functions that @str4d used in the FsBlockDb implementation to sync versions. I will just return a join handle for delete scanned. Damn I wish I knew this sooner... I was wondering why the tool chain was fixed!

zcash_client_backend/src/data_api/chain.rs
Oscar-Pepper: I imagined that the error types and conditions would be decided on implementation so it is flexible for the consumer

Oscar-Pepper: On second thought partial reads probably should be allowed in the case that a block cache is storing compact blocks on disk and therefore can resume if the whole scan range had not be stored yet instead of downloading all the blocks in the scan range. This would be a nice addition to the sync implementation but would be an optimisation as oppose to a key component and im not sure how often this would occur. As other wallets (i.e. probably zingo) will store the block cache in memory for performance, this will not apply in those implemenations. The other option would be to define an additional block cache error enum where one of the variants is associated with not reading all blocks in scan range so the sync implementation can mathc on this error instead of checking the Vec<CompactBlock> matches the size of the scan range

zancas: Isn't the availability of blocks in the cache orthogonal to the backing store?  If the cache is in-memory it still must go through a non-deterministic population process?  e.g. downloading the blocks from a chain-attached server? Apologies if I am misunderstanding something fundamental..   perhaps that concern is handled elsewhere?

Oscar-Pepper:
#1217 shows how this BlockCache trait is implemented into the #1184 sync engine

Cargo.toml
```diff
@@ -114,6 +114,9 @@ aes = "0.8"
 fpe = "0.6"
 zip32 = "0.1"
 
+# Runtime
+tokio = { version = "1.21.0", features = ["rt-multi-thread"] }
```
str4d, line 118: We cannot enable this flag at a library level; it needs to be up to the application.
```suggestion
tokio = "1.21"
```

zcash_client_backend/Cargo.toml
```diff
@@ -75,10 +75,14 @@ shardtree.workspace = true
 
 # - Test dependencies
 proptest = { workspace = true, optional = true }
+jubjub = { workspace = true, optional = true }
 
 # - ZIP 321
 nom = "7"
 
+# - Runtime
+tokio.workspace = true
```
str4d, line 84: Add two separate entries for this in `[dev-dependencies]` that enable either `rt` or `rt-multi-thread` depending on whether the build target is WASM or not.

str4d (changes requested):

Cargo.toml
Oscar-Pepper: I have moved the tokio dependency to zcash_client_backend an optional and behind a "sync" feature for BlockCache. rt-multi-thread is only a dev-dependency now for the doc-test

zcash_client_backend/Cargo.toml
Oscar-Pepper: see comment above

zcash_client_backend/src/data_api/chain.rs
nuttycom: Okay. Please document this decision in the method documentation.

daira: It seems reasonable to me to allow short reads (but not reads with gaps). It is fairly easy for consumers to handle that, and it still fits the current `Result<Vec<CompactBlock>, Self::Error>` return type.

daira (approved):
utACK modulo documentation for short reads, if they are allowed.

zancas:
Oh hi!  What needs to be done to approve this workflow?

daira (dismissed):
utACK

zancas:
I think this is ready to land.  I think @Oscar-Pepper has already addressed @str4d 's requested changes.

zcash_client_backend/CHANGELOG.md
```diff
@@ -38,6 +39,8 @@ and this library adheres to Rust's notion of
   - `Nullifiers::{orchard, extend_orchard, retain_orchard}`
   - `TaggedOrchardBatch`
   - `TaggedOrchardBatchRunner`
+  - `testing` module
+  - `testing::{'fake_compact_block`, `random_compact_tx`} (moved from `tests` module).
```
str4d, line 43: The `testing` module is new (so the addition of `fake_compact_block` is implicit), the `tests` module is private, and `random_compact_tx` is private. Thus none of this needs to go into the changelog (which is for downstream consumers of the crate's public API):
```suggestion
```
or
```suggestion
  - `testing::fake_compact_block`
```
(Also note the Markdown syntax errors; the backticks should go around the entire name fragment, not the individual function names.)

zcash_client_backend/CHANGELOG.md
```diff
@@ -81,6 +84,8 @@ and this library adheres to Rust's notion of
     constraint on its `<AccountId>` parameter has been strengthened to `Copy`.
 - `zcash_client_backend::fees`:
   - Arguments to `ChangeStrategy::compute_balance` have changed.
+- `zcash_client_backend::scanning`:
+  - `testing::fake_compact_block` is now public.
```
str4d, line 88: Duplicated from the "Added" section (the fact that it was crate-internal previously is irrelevant to consumers of the public API; this is an addition).
```suggestion
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -220,6 +223,198 @@ pub trait BlockSource {
...
+
+    /// Removes all cached blocks above a specified block height.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
+    fn truncate(&self, block_height: BlockHeight) -> Result<(), Self::Error>;
+
+    /// Deletes a range of compact blocks from the block cache.
+    ///
+    /// Returns a `JoinHandle` from a `tokio::spawn` task.
+    fn delete(&self, range: &ScanRange) -> JoinHandle<Result<(), Self::Error>>;
```
str4d, line 415: `tokio::task::JoinHandle` requires the `tokio/rt` feature flag, which seems excessive for the library (requiring us to at a library level depend on a specific runtime). I think all we actually need here is a `Box<dyn Future>` we can then await? `JoinHandle` implements `Future`, so a concrete implementation of this trait should be able to satisfy it with `tokio`.

We might also want to use the `async-trait` crate here to enable effectively returning `impl Future` instead, by making this:
```suggestion
    async fn delete(&self, range: &ScanRange) -> Result<(), Self::Error>;
```

zcash_client_backend/Cargo.toml
```diff
@@ -78,12 +78,16 @@ group.workspace = true
 orchard = { workspace = true, optional = true }
 sapling.workspace = true
 
+# - Sync engine
+tokio = { version = "1.21.0", optional = true, features = ["fs", "macros"] }
```
str4d, line 82: I do not know where these two features are depended on. The only feature I _do_ see used is `rt`, which is not enabled here; therefore I believe this PR is currently broken for downstream users (it won't appear broken when testing locally, because `rt-multi-thread` is enabled in dev-dependencies).

str4d (changes requested):
Reviewed c62b5b2fba3199cfc740362ccf7ff6bd56d8d781

zcash_client_backend/src/data_api/chain.rs
Oscar-Pepper: Ok yes I prefer this to be async and was how I implemented originally. I will put this back but use the `aync-trait` crate so its compatible with rust v1.65

zcash_client_backend/Cargo.toml
Oscar-Pepper: I agree "rt" makes more sense in the context of this PR and will make this change. The reason why this is here is because I am developping this PR alongside #1217 which builds upon your 1169-sync engine PR. In your PR you use these features along with a JoinHandle and i wanted to keep in line with your decisions as much as possible. You are right... now you bring this up I wonder how this is building as JoinHandle is not part of "fs" or "macros"

str4d: To clarify, after my other comment is addressed (i.e. not using `JoinHandle`), we _shouldn't_ enable `rt` here (i.e. addressing the other comment will also fix the breakage). I also think we can remove `fs` and `macros` (at least from this PR; they can be added later if there is a later PR that needs them).

zancas:
It looks like this workflow is ready to be kicked...

Oscar-Pepper:
Hi @str4d I've made the requested changes and it's ready for review again

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -224,6 +225,195 @@ pub trait BlockSource {
...
+    /// If `range` is `None`, returns the tip of the entire cache.
+    ///
+    /// Returns `Ok(Some(BlockHeight))` on success, otherwise returns an error.
+    /// If no blocks are found in the cache, returns Ok(`None`).
+    fn get_tip_height(&self, range: Option<&ScanRange>)
+        -> Result<Option<BlockHeight>, Self::Error>;
+
+    /// Inserts a vec of compact blocks into the block cache.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
```
str4d, line 403: Document whether this should return an error if the blocks aren't contiguous. It probably doesn't matter here if they aren't, but we should make sure devs are clear on what the expected behaviour is here.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -224,6 +225,195 @@ pub trait BlockSource {
...
+    fn insert(&self, compact_blocks: Vec<CompactBlock>) -> Result<(), Self::Error>;
+
+    /// Removes all cached blocks above a specified block height.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
+    fn truncate(&self, block_height: BlockHeight) -> Result<(), Self::Error>;
+
+    /// Deletes a range of compact blocks from the block cache.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
```
str4d, line 413: This kind of documentation is not useful, as it is entirely self-evident from the return type.
```suggestion
```

What is more useful is to document what the expectations are for the state of `self` when an error occurs. For example, saying that some cached blocks may remain in the database, or (if this needs to occur atomically) that no change occurs.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -224,6 +225,195 @@ pub trait BlockSource {
...
+
+    /// Inserts a vec of compact blocks into the block cache.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
+    fn insert(&self, compact_blocks: Vec<CompactBlock>) -> Result<(), Self::Error>;
+
+    /// Removes all cached blocks above a specified block height.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
+    fn truncate(&self, block_height: BlockHeight) -> Result<(), Self::Error>;
```
str4d, line 409: Hmm, do we need this as a separate synchronous method? It seems functionally identical to `Self::delete` except with an open-ended range. If we make `truncate` async, it could be implemented as a default method with an implementation something like:
```rust
if let Some(latest) = self.get_tip_height(None)? {
    self.delete(ScanRange(block_height, latest + 1)).await?;
}
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -224,6 +225,195 @@ pub trait BlockSource {
...
+    ///
+    /// Returns `Ok(Some(BlockHeight))` on success, otherwise returns an error.
+    /// If no blocks are found in the cache, returns Ok(`None`).
+    fn get_tip_height(&self, range: Option<&ScanRange>)
+        -> Result<Option<BlockHeight>, Self::Error>;
+
+    /// Inserts a vec of compact blocks into the block cache.
+    ///
+    /// Returns `Ok(())` on success, otherwise returns an error.
+    fn insert(&self, compact_blocks: Vec<CompactBlock>) -> Result<(), Self::Error>;
```
str4d, line 404: Should the mutating methods take `&mut`? That seems morally correct, even if the inner cache uses `Rc` or `Arc` to manage the database, but is there an issue preventing this that I'm not aware of?

Also, looking at this trait again, this method and `read` need to be async. The implementation of this trait cannot just modify the block metadata database; it also needs to do the disk / memory writes. So any mutations need to be async in the same way that deletions are, to support whatever the cache needs to do. (The previous join handle approach meant that I wasn't thinking about the overall async process, and was just thinking about the "deletion deferrment", but the operations themselves can also be async.)

str4d (changes requested):
Reviewed as of 7301c0c76e8dddd965b4705097b7a97144aebe9b.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -224,6 +225,195 @@ pub trait BlockSource {
...
+///    assert_eq!(block_cache.get_tip_height(None).unwrap(), None);
+/// ```
+#[async_trait]
+pub trait BlockCache: BlockSource + Send + Sync {
+    /// Retrieves contiguous compact blocks specified by the given `range` from the block cache.
+    ///
+    /// Returns `Ok(Vec<CompactBlock>)` on success, otherwise returns an error.
+    ///
+    /// Short reads are allowed, meaning that returning fewer blocks than requested should not
+    /// return an error as long as all blocks are sequentially continuous in height.
```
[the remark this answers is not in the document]
Oscar-Pepper: I have added your suggestions to clarify the first block should be returned

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -224,6 +225,195 @@ pub trait BlockSource {
...
+    ///
+    /// # Errors
+    ///
+    /// This method should return an error if there are gaps in the requested range of blocks,
+    /// indicating there are blocks missing from the cache.
```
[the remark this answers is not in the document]
Oscar-Pepper: I have changed this to hopefully be more clear. I believe this was suggested by @daira to not allow gaps in reads where the blocks read fro the cache are not contiguous

zcash_client_backend/src/data_api/chain.rs
Oscar-Pepper: I did not want to constrain an insert to a contiguous range of blocks and i will make that clear in the doc comment. I dont fully recall off the top of my head but there was a case where I realised, either from your current implemenation or a potential future optimisation, where it may be beneficial to be able to insert non-sequential ranges in one call.

zcash_client_backend/src/data_api/chain.rs
Oscar-Pepper: Yeah this is a great idea. It will change the way truncate is implemented in 1169-sync-engine but for the better

zcash_client_backend/src/data_api/chain.rs
Oscar-Pepper: I will remove these

Oscar-Pepper: When this is implemented for FsBlockDb it will be removing each block from disk sequentially so I think its best to say that in the case of an error blocks may remain in the database here

zcash_client_backend/src/data_api/chain.rs
Oscar-Pepper: I agree with making more functions async. I don't recall needing async for reads in your implementation but for inserts I did change for a sync version when I was trying to find a solution for not using async traits. Either way, making all these async will support a wider range of implemetations and sync optimisations. That being said, making insert async will also require some test framework such as `generate_blocks_at` to also become async or be refactored, although this is not a problem to implement at all.

Regarding the `&mut` I think it would be good to have quick call soon to discuss the sync implementation especially when it comes to concurrency. For example, maybe im missing something, but would there not be issues with running several `delete` tasks that take a `&mut` as in your 1169-sync-engine? Or would there be issues for example reading from the cache for scanning while concurrently downloading and inserting into the cache which would be a good sync optimisation?

Oscar-Pepper:
> Reviewed as of [7301c0c](https://github.com/zcash/librustzcash/commit/7301c0c76e8dddd965b4705097b7a97144aebe9b).

Hi @str4d, thanks for the review, I have replied to and made all requested changes except for adding `&mut` which we should discuss if its felt this is neccessary as its not clear to me how to implement 1169-sync-engine this way or further concurrent optimisations.

I have sent for review again.

str4d (approved):
utACK. There are some further changes I want to make (like placing this behind a feature flag), but I can do that when I rebase #1184 on top of this.

PR #1194, opened 2024-02-22
Changelog cleanups

str4d:


daira (approved):
ACK, doc-only.

nuttycom (approved):
doc-only ACK

PR #1207, opened 2024-02-29
WIP: Create an in-memory wallet backend.

nuttycom:


zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -0,0 +1,459 @@
...
+    addresses: BTreeMap<DiversifierIndex, UnifiedAddressRequest>,
+    notes: HashSet<NoteId>,
+}
+
+pub struct MemoryWalletDb {
+    network: Network,
+    accounts: BTreeMap<u32, MemoryWalletAccount>,
+    blocks: BTreeMap<BlockHeight, MemoryWalletBlock>,
+    tx_idx: HashMap<TxId, BlockHeight>,
+    sapling_spends: HashMap<sapling::Nullifier, (TxId, bool)>,
```
oxarbitrage, line 86: How do you populate a field like this? I made some experiments and i always end up with an error:

```
the trait bound `sapling_crypto::Nullifier: Hash` is not satisfied
the trait `Hash` is not implemented for `sapling_crypto::Nullifier`
```

nuttycom: Ah... apparently we need to add some impl's to the `sapling-crypto` crate. :sweat_smile:

nuttycom: This is definitely a sketch at present.

oxarbitrage: Ah, ok ,it will be ok to just derive it? https://github.com/zcash/sapling-crypto/pull/133

nuttycom: Since `orchard` only derives `PartialOrd` and `Ord`, let's use BTreeMap instead (and add those to `sapling-crypto`).

PR #1213, opened 2024-03-01
zcash_client_backend: Add `WalletRead::validate_seed`

### LINKED ISSUE #1189: zcash_client_backend: Add a `validate_seed` method to `WalletRead`
It is necessary that the wallet be able to verify that a provided seed corresponds to the existing UFVK(s) in the wallet database for which there are associated ZIP 32 account IDs. 

See https://github.com/Electric-Coin-Company/zashi/issues/20#issuecomment-1946285981 for additional detail.

nuttycom:
This adds a mechanism that allows a caller to verify that a given seed generates the viewing key that is stored in the wallet for a specified account.

Fixes #1189

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -492,6 +492,17 @@ pub trait WalletRead {
     /// will be interpreted as belonging to that account.
     type AccountId: Copy + Debug + Eq + Hash;
 
+    /// Verifies that the given seed corresponds to the viewing key for the specified account.
+    ///
+    /// Returns `Ok(true)` if the viewing key for the specified account can be derived from
+    /// the provided seed, `Ok(false)` if the derived seed does not match or the specified
+    /// account is not present in the database, or an error in the case
+    fn validate_seed(
+        &self,
+        account_id: Self::AccountId,
+        seed: &SecretVec<u8>,
+    ) -> Result<bool, Self::Error>;
+
     /// Returns the height of the chain as known to the wallet as of the most recent call to
     /// [`WalletWrite::update_chain_tip`].
     ///
@@ -1317,14 +1328,15 @@ pub mod testing {
         type Error = ();
         type AccountId = u32;
 
-        fn chain_height(&self) -> Result<Option<BlockHeight>, Self::Error> {
-            Ok(None)
+        fn validate_seed(
+            &self,
+            _account_id: Self::AccountId,
+            _seed: &SecretVec<u8>,
+        ) -> Result<bool, Self::Error> {
+            Ok(false)
         }
 
-        fn get_target_and_anchor_heights(
-            &self,
-            _min_confirmations: NonZeroU32,
-        ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error> {
+        fn chain_height(&self) -> Result<Option<BlockHeight>, Self::Error> {
             Ok(None)
         }
 
@@ -1345,6 +1357,13 @@ pub mod testing {
 
         fn suggest_scan_ranges(&self) -> Result<Vec<ScanRange>, Self::Error> {
             Ok(vec![])
+        }
+
+        fn get_target_and_anchor_heights(
+            &self,
+            _min_confirmations: NonZeroU32,
+        ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error> {
+            Ok(None)
         }
 
         fn get_min_unspent_height(&self) -> Result<Option<BlockHeight>, Self::Error> {
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -494,9 +494,13 @@ pub trait WalletRead {
 
     /// Verifies that the given seed corresponds to the viewing key for the specified account.
     ///
-    /// Returns `Ok(true)` if the viewing key for the specified account can be derived from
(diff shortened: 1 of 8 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -492,6 +492,17 @@ pub trait WalletRead {
     /// will be interpreted as belonging to that account.
     type AccountId: Copy + Debug + Eq + Hash;
 
+    /// Verifies that the given seed corresponds to the viewing key for the specified account.
+    ///
+    /// Returns `Ok(true)` if the viewing key for the specified account can be derived from
+    /// the provided seed, `Ok(false)` if the derived seed does not match or the specified
+    /// account is not present in the database, or an error in the case
```
daira, line 499: The last sentence appears to be incomplete.

zcash_client_sqlite/src/lib.rs
```diff
@@ -245,6 +245,32 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
+    fn validate_seed(
+        &self,
+        account_id: Self::AccountId,
+        seed: &SecretVec<u8>,
+    ) -> Result<bool, Self::Error> {
+        self.get_unified_full_viewing_keys().and_then(|keys| {
+            keys.get(&account_id).map_or(Ok(false), |ufvk| {
+                let usk = UnifiedSpendingKey::from_seed(
+                    &self.params,
+                    &seed.expose_secret()[..],
```
daira, line 257: Should this be passed further down the call chain as a `&SecretVec<u8>`?

zcash_client_sqlite/src/lib.rs
```diff
@@ -245,6 +245,32 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
+                    account_id,
+                )
+                .map_err(|_| SqliteClientError::KeyDerivationError(account_id))?;
+
+                // Keys are not comparable with `Eq`, but addresses are, so we derive what should
+                // be equivalent addresses for each key and use those to check for key equality.
```
daira, line 263: This should work but is quite ugly. I don't see any reason why keys shouldn't implement `ConstantTimeEq`.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1217,6 +1243,20 @@ mod tests {
...
+            .with_test_account(AccountBirthday::from_sapling_activation)
+            .build();
+
+        assert!({
+            let account = st.test_account().unwrap().0;
+            st.wallet()
+                .validate_seed(account, st.test_seed().unwrap())
+                .unwrap()
+        })
+    }
```
daira, line 1258: Add a test for the case where `validate_seed` should return false.

zcash_keys/src/keys.rs
```diff
@@ -450,6 +450,9 @@ pub struct UnifiedAddressRequest {
 }
 
 impl UnifiedAddressRequest {
+    /// Construct a new unified address request from its constituent parts.
+    ///
+    /// Returns `None` if the resulting unified address would not include at least one shielded receiver.
```
daira, line 455: I realise this is just documenting the current behaviour, but it should be creating an R1 UA if there are only transparent receivers.

daira (approved):
utACK modulo comments.

zcash_keys/src/keys.rs
nuttycom: The revision of the UA is not tracked at this level; that's an encoding concern for `zcash_address`

zcash_client_sqlite/src/lib.rs
nuttycom: We can do that; I was being a bit hasty trying to get this in so that the guys could use it in Zashi, and I didn't want to wait for `sapling-crypto` and `orchard` releases. I guess the change is purely additive, though, so we can do it in a point release.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/1e7e0d9e157fb94c17471df2fc2be8d46f6fa841..e4c0b0e63fe335656e8d29890a21271b2e429871) to address comments from review.

zcash_client_backend/src/data_api.rs
str4d, line 499:
```suggestion
    /// Returns:
    /// - `Ok(true)` if the viewing key for the specified account can be derived from the
    ///   provided seed.
    /// - `Ok(false)` if the derived seed does not match, or the specified account is not
    ///   present in the database.
    /// - `Err(_)` if a Unified Spending Key cannot be derived from the seed for the
    ///   specified account.
```

str4d (dismissed):
utACK e4c0b0e63fe335656e8d29890a21271b2e429871 after doc changes.

zcash_client_sqlite/src/lib.rs
nuttycom: If we want to do so, it should be in a separate PR.

str4d: All this practically does is force the caller to use `SecretVec` for its own memory management; given that we do not copy or persist the seed inside `UnifiedSpendingKey::from_seed`, and don't perform any comparisons or branching on it, passing the data around as a slice is fine.

str4d (approved):
utACK 0d430ba62c1ffe097136c9dfe388c8e9683af4a4

daira (commented):
Post-hoc ACK

PR #1214, opened 2024-03-04
zcash_keys: Fix a few problems with no-flags compilation.

nuttycom:


zcash_keys/src/address.rs
```diff
@@ -143,7 +143,11 @@ impl UnifiedAddress {
 
     /// Returns whether this address has a Sapling receiver.
     pub fn has_sapling(&self) -> bool {
-        self.sapling.is_some()
```
str4d, line 146: Document this in a `Fixed` section, as we will be pushing this out in a `0.1.1`.

str4d (approved):
utACK 39045342df8d236108dce2cb251597d03bb20691 after changelog update.

zcash_keys/src/address.rs
nuttycom: Fixed.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/39045342df8d236108dce2cb251597d03bb20691..f26ea5d40138e8471e37e3864943cf56e4f82a7e) to note the fix in the CHANGELOG.

str4d (dismissed):
Tested ACK f26ea5d40138e8471e37e3864943cf56e4f82a7e

str4d (approved):
Re-ACK 836d88b9c305e04e969f0d8f73a3cf25b14928ab

daira (commented):
Post-hoc ACK

PR #1228, opened 2024-03-07
Fixed cargo doc build failure

Oscar-Pepper:
This fixes cargo doc build when transparent-inputs feature is not specified


nuttycom (approved):
Tested ACK

PR #1233, opened 2024-03-08
zcash_client_sqlite: Add Orchard support to test harness

str4d:


zcash_client_sqlite/src/testing.rs
```diff
@@ -815,6 +841,79 @@ impl TestFvk for DiversifiableFullViewingKey {
...
 
+#[cfg(feature = "orchard")]
+impl TestFvk for orchard::keys::FullViewingKey {
+    type Nullifier = orchard::note::Nullifier;
+
+    fn sapling_ovk(&self) -> Option<sapling::keys::OutgoingViewingKey> {
+        None
+    }
+
+    fn orchard_ovk(&self, scope: zip32::Scope) -> Option<orchard::keys::OutgoingViewingKey> {
```
nuttycom, line 875: We probably only need external scope; for internal scope we will always decrypt with the ivk?

zcash_client_sqlite/src/testing.rs
```diff
@@ -1099,24 +1152,22 @@ fn fake_compact_block_spending<P: consensus::Parameters, Fvk: TestFvk>(
     let mut rng = OsRng;
```
nuttycom, line 1152: The `PaymentAddress` argument of this function should probably become an `Address`.

zcash_client_sqlite/src/testing.rs
```diff
@@ -1099,24 +1152,22 @@ fn fake_compact_block_spending<P: consensus::Parameters, Fvk: TestFvk>(
...
-    fvk.add_compact_spend(&mut ctx, nf, &mut rng);
-
-    // Create a fake Note for the payment
-    ctx.outputs
-        .push(compact_sapling_output(params, height, to, value, fvk.sapling_ovk(), &mut rng).0);
-
-    // Create a fake Note for the change
-    fvk.add_compact_output(
+    // Create a fake spend and a fake Note for the change
+    fvk.add_compact_action(
```
nuttycom, line 1156: Hm, this seems too restrictive; isn't `add_compact_action` only available when the `orchard` feature is enabled? I think that the change should go to whatever pool the output is targeting.

zcash_client_sqlite/src/testing.rs
```diff
@@ -1099,24 +1152,22 @@ fn fake_compact_block_spending<P: consensus::Parameters, Fvk: TestFvk>(
...
         &mut ctx,
         params,
         height,
+        nf,
         AddressType::Internal,
         (in_value - value).unwrap(),
         initial_sapling_tree_size,
```
nuttycom, line 1163: Why does this take the Sapling tree size, and not the Orchard tree size?

nuttycom (changes requested):
I'm a little confused by the `add_compact_action` function - is that adding an Orchard action, or a logical action?

zcash_client_sqlite/src/testing.rs
str4d: Sailing nullifiers depend on tree position, Orchard nullifiers don't.

zcash_client_sqlite/src/testing.rs
str4d: This is the trait method, not the feature gated helper. I will name this `add_logical_action` to clarify.

zcash_client_sqlite/src/testing.rs
str4d: I didn't write `TestFvk` with the intention of it being used directly in tests; it provides the functionality needed by `TestState`.

Separately, you do not want to use external ovks for internal change, otherwise giving out an external ivk ovk pair leaks internal wallet actions.

zcash_client_sqlite/src/testing.rs
str4d: That change is out of scope for what I was able to finish by 1am, because it requires a completely separate generalization, not `TestFvk`

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/495cae82209be50923c067ccc40c95eaa053d448..cc90c4c4059a6d122e54160474d6a0b01758ac0f) to silence Clippy lints, and rename `TestFvk` methods for clarity.

zcash_client_sqlite/src/testing.rs
str4d: Now that it's 1pm instead, indeed this particular generalization is trivial now that I've done the `TestFvk` part. Added it to the PR.

nuttycom (approved):
utACK c8d03f69efbc5574ba81a2e60f531a845f0cf016

PR #1234, opened 2024-03-08
zcash_client_backend: Update `DecryptedTransaction` to support Orchard

nuttycom:


components/zcash_protocol/src/memo.rs
```diff
@@ -197,13 +197,25 @@ impl TryFrom<MemoBytes> for Memo {
...
+    /// example, if the slice is not 512 bytes, or the encoded `Memo` is non-canonical).
+    fn try_from(bytes: &MemoBytes) -> Result<Self, Self::Error> {
         match bytes.0[0] {
             0xF6 if bytes.0.iter().skip(1).all(|&b| b == 0) => Ok(Memo::Empty),
             0xFF => Ok(Memo::Arbitrary(Box::new(bytes.0[1..].try_into().unwrap()))),
             b if b <= 0xF4 => str::from_utf8(bytes.as_slice())
                 .map(|r| Memo::Text(TextMemo(r.to_owned())))
                 .map_err(Error::InvalidUtf8),
-            _ => Ok(Memo::Future(bytes)),
+            _ => Ok(Memo::Future(bytes.clone())),
```
str4d, line 218: I'm somewhat less fond on the fact that we now take reference to an owned value and then clone it, but I'm also less fond of duplicating all of this, and the other option (always cloning the reference) also induces unnecessary clones. So I guess changing this back later is an implementation detail that we can fix in a point release if we want to later.

zcash_client_backend/CHANGELOG.md
```diff
@@ -69,6 +82,8 @@ and this library adheres to Rust's notion of
 ### Removed
 - `zcash_client_backend::PoolType::is_receiver`: use
   `zcash_keys::Address::has_receiver` instead.
+- `zcash_client_backend::DecryptedTransaction::sapling_outputs` use
+  the `DecryptedTransaction::outputs` method instead.
```
str4d, line 86:
```suggestion
- `zcash_client_backend::DecryptedTransaction::sapling_outputs` (use
  the `DecryptedTransaction::outputs` method instead).
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -631,76 +628,100 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+            for output in d_tx.outputs() {
+                match output.transfer_type() {
                     TransferType::Outgoing | TransferType::WalletInternal => {
-                        let value = output.note.value();
-                        let recipient = if output.transfer_type == TransferType::Outgoing {
-                            Recipient::Sapling(output.note.recipient())
+                        let recipient = if output.transfer_type() == TransferType::Outgoing {
+                            match output.note() {
+                                Note::Sapling(n) => {
+                                    //TODO: Recover the UA, if possible.
```
str4d, line 640: Open an issue for this.

zcash_client_backend/src/data_api.rs
```diff
@@ -872,13 +872,29 @@ impl<A> ScannedBlock<A> {
...
+    /// Constructs a new [`DecryptedTransaction`] from its constituent parts.
+    pub fn new(tx: &'a Transaction, outputs: &'a Vec<DecryptedOutput<Note, AccountId>>) -> Self {
+        Self { tx, outputs }
+    }
+
+    /// Returns the raw transaction data.
+    pub fn tx(&self) -> &Transaction {
+        self.tx
+    }
+    /// Returns the Sapling outputs that were decrypted from the transaction.
```
str4d, line 894:
```suggestion
    /// Returns the outputs that were decrypted from the transaction.
```

str4d (approved):
utACK 9de4f1f1b7eef7dfef04b7d77c4c805ee315bf1a

zcash_client_backend/CHANGELOG.md
```diff
@@ -65,6 +82,8 @@ and this library adheres to Rust's notion of
 ### Removed
 - `zcash_client_backend::PoolType::is_receiver`: use
   `zcash_keys::Address::has_receiver` instead.
```
daira, line 84:
```suggestion
- `zcash_client_backend::PoolType::is_receiver` (use
  the `zcash_keys::Address::has_receiver` method instead).
```
for consistency with Str4d's suggestion.

components/zcash_protocol/src/memo.rs
nuttycom: I figured that the unnecessary clone here is less common.

nuttycom: I figured that the `Future` case is relatively uncommon.

zcash_client_backend/CHANGELOG.md
nuttycom: I restored the `sapling_outputs` method; it ultimately didn't make sense to unify the output vectors at this level.

zcash_client_sqlite/src/lib.rs
nuttycom: https://github.com/zcash/librustzcash/issues/1236

zcash_client_backend/src/data_api.rs
nuttycom: Resolved in a slightly different fashion; Orchard and Sapling are kept separate at this level.

zcash_client_backend/CHANGELOG.md
nuttycom: This can be handled in PR cleanup, as it's not affecting changes of this PR.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/9de4f1f1b7eef7dfef04b7d77c4c805ee315bf1a..0911d34a4762b24551acbdcae1a0092651cca61c) to take a slightly cleaner approach to the `DecryptedTransaction` changes.

zcash_client_sqlite/src/lib.rs
```diff
@@ -703,12 +696,12 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                     TransferType::Incoming => {
                         match spending_account_id {
                             Some(id) => {
-                                if id != output.account {
+                                if id != *output.account() {
                                     panic!("Unable to determine a unique account identifier for z->t spend.");
                                 }
```
nuttycom, line 701: We need to figure out why we're enforcing this "one account only for outputs" rule but aren't doing anything with the account ID that we resolve on.

nuttycom: Opened https://github.com/zcash/librustzcash/issues/1238

zcash_client_backend/src/decrypt.rs
```diff
@@ -29,43 +31,91 @@ pub enum TransferType {
...
+
+    /// Returns a [`TransferType`] value that is determined based upon what type of key was used to
+    /// decrypt the transaction.
+    pub fn transfer_type(&self) -> TransferType {
+        self.transfer_type
+    }
+}
+
+impl<A> DecryptedOutput<sapling::Note, A> {
+    pub fn note_value(&self) -> NonNegativeAmount {
```
str4d, line 87: Document.

zcash_client_backend/src/decrypt.rs
```diff
@@ -29,43 +31,91 @@ pub enum TransferType {
...
+        NonNegativeAmount::from_u64(self.note.value().inner())
+            .expect("Sapling note value is expected to have been validated by consensus.")
+    }
+}
+
+#[cfg(feature = "orchard")]
+impl<A> DecryptedOutput<orchard::note::Note, A> {
+    pub fn note_value(&self) -> NonNegativeAmount {
```
str4d, line 95: Document.

zcash_client_backend/src/decrypt.rs
```diff
@@ -29,43 +31,91 @@ pub enum TransferType {
...
-    pub index: usize,
+    index: usize,
+    note: Note,
+    account: AccountId,
+    memo: MemoBytes,
+    transfer_type: TransferType,
+}
+
+impl<Note, AccountId: Copy> DecryptedOutput<Note, AccountId> {
+    pub fn new(
```
str4d, line 42: Document.

str4d (approved):
utACK 78e491689946c62452088685e79576b2612e7891

zcash_client_sqlite/src/lib.rs
```diff
@@ -665,35 +662,31 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
-            for output in d_tx.sapling_outputs {
-                match output.transfer_type {
+            for output in d_tx.sapling_outputs() {
+                match output.transfer_type() {
                     TransferType::Outgoing | TransferType::WalletInternal => {
-                        let value = output.note.value();
-                        let recipient = if output.transfer_type == TransferType::Outgoing {
-                            Recipient::Sapling(output.note.recipient())
+                        let recipient = if output.transfer_type() == TransferType::Outgoing {
+                            //TODO: Recover the UA, if possible.
```
AArnott, line 672: If you're going to do something clever to build a UA from a transaction recipient, please only do so into some new Option field for that purpose. At least, there should be a straightforward way to determine the _actual_ recipient address (read: _one_ receiver) for a given output as a consumer of this API.

zcash_client_sqlite/src/lib.rs
```diff
@@ -717,24 +710,80 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                        if matches!(recipient, Recipient::InternalAccount(_, _)) {
+                            todo!();
+                            //wallet::orchard::put_received_note(wdb.conn.0, output, tx_ref, None)?;
+                        }
+                    }
+                    TransferType::Incoming => {
+                        match spending_account_id {
+                            Some(id) => {
+                                if id != *output.account() {
+                                    panic!("Unable to determine a unique account identifier for z->t spend.");
```
AArnott, line 755: This isn't a z->t spend at least because we're in orchard (not sapling).
I also don't see anything suggesting this is an unshielding transaction. We're just looping over orchard outputs here.
Why is the local variable called `spending_account_id` when in fact it is set to the value of the _receiving_ account id?
And why must every output in a transaction have the same receiving account? This looks like a panic will occur for any transaction that sends outputs to more than one account tracked by the sqlite db. 
And in fact, I found this code due to a panic when doing account recovery for a testnet account I have, and a multi-account transaction is very likely to be what I have here (though I haven't confirmed this.
See also [discord post](https://discord.com/channels/809218587167293450/876655911790321684/1219086198513860659).

PR #1175, opened 2024-02-08
Give zcash_client_sqlite its own account ID

AArnott:
This is work toward supporting arbitrary imported keys, which may not be based on seeds (as far as we know) and may consist only of a full viewing key or incoming viewing key.

Closes #1218.

## Additional work
- [ ] *New* tests?
- [ ] Bring in `UnifiedIncomingViewingKey` as a type and then a `UnifiedViewingKey` enum that will allow code to *start* preparing for the possibility of UIVKs being in the `accounts` table instead of it always being UFVKs.
- [x] Update the CHANGELOGs
- [ ] Are there other SQL queries for which no automated test exists and therefore may still need to be discovered and updated given the new db schema?
- [x] Either update the ASCII art graph in migrations.rs or merge in #1186 and then update the `migrations.md` file.

## Interesting things of note

- Several tests needed to be updated because they assumed the result of `create_account` was a ZIP-32 account index where it was really only documented to return an account id or that the first account id would be 0. In fact in the sqlite implementation these two values are no longer the same. I've updated the docs and CHANGELOG to call out this behavior.
- In addition to adding a few columns to the `accounts` table, I also *renamed* a couple of existing columns to avoid their being used with their old semantics. I also added indexes for the FK columns of each table that had to be migrated.
- The `add_utxo_account` migration was improperly relying on internal crate code to query the database as part of migration. The problem with this is the internal crate code assumes the _latest_ db schema, whereas the migration must be able to run with an earlier schema. With this PR, a later migration occurs that breaks the schema in ways that required internal crate updates, and a migration that runs _before_ those updates then failed when calling into the 'newer' crate code. The fix is to remove the dependency: migrations should never make indirect sql calls through non-migration code. I copied the two functions from the regular code into the migration and reverted the changes I had made to them so that the migration works again. This code in the migration may be trimmable to something smaller to fit what the migration needs, but I didn't want to rewrite it.
- The new db schema renames the `accounts.ufvk` column to `uvk` to allow for UFVKs today, and UIVKs tomorrow. Code remains in the crate itself that only tries to parse the value of this column as a UFVK because no UIVK type exists yet. But I've tried to model the APIs to allow for adding support for that with no further breaking changes.

This contributes toward #1075.

PR #1235, opened 2024-03-08
zcash_client_sqlite: Rework accounts to support imported viewing keys

str4d:
Squashed version of #1175.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -339,6 +339,7 @@ pub(crate) fn ufvk_to_uivk<P: consensus::Parameters>(
         divk[32..].copy_from_slice(&ivk[..]);
         ivks.push(Ivk::Sapling(divk));
     }
+    #[cfg(feature = "transparent-inputs")]
     if let Some(tfvk) = ufvk.transparent() {
         let tivk = tfvk.derive_external_ivk()?;
         ivks.push(Ivk::P2pkh(tivk.serialize().try_into().map_err(|_| {
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -53,8 +53,8 @@
 //! This view exposes the history of transaction outputs received by and sent from the wallet,
 //! keyed by transaction ID, pool type, and output index. The contents of this view are useful for
 //! producing a detailed report of the effects of a transaction. Each row of this view contains:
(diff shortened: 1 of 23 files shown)


str4d:
The current commit in this PR (0c80893ce0f8b88d5329157dfb583af14cd2b3aa) [is identical to](https://github.com/nerdcash/librustzcash/compare/cc6e7153676bf6db93eef91a3466de9623f7e51e..zcash:librustzcash:0c80893ce0f8b88d5329157dfb583af14cd2b3aa) the tip commit in #1175 (nerdcash/librustzcash@cc6e7153676bf6db93eef91a3466de9623f7e51e).

AArnott:
@str4d did you see @nuttycom's comment here: https://github.com/zcash/librustzcash/pull/1175#pullrequestreview-1925276413

nuttycom (approved):
utACK bfc2278bfe38b792dc494ffd72c1dda3294dd8aa

This is identical to #1175, which was previously approved, with the addition of bugfixes and updating to use sapling-crypto 0.1.2 for IVK encoding.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -332,12 +332,8 @@ pub(crate) fn ufvk_to_uivk<P: consensus::Parameters>(
...
     }
     if let Some(sapling) = ufvk.sapling() {
-        let dk = sapling.to_bytes();
-        let ivk = sapling.to_ivk(Scope::External).to_repr();
-        let mut divk = [0u8; 64];
-        divk[..32].copy_from_slice(&dk[96..]);
-        divk[32..].copy_from_slice(&ivk[..]);
-        ivks.push(Ivk::Sapling(divk));
+        let ivk = sapling.to_external_ivk();
+        ivks.push(Ivk::Sapling(ivk.to_bytes()));
```
AArnott, line 336: Oh ya, that's much nicer. I had planned to add the missing function as part of my UIVK work, but I'm glad it's already been done.

nuttycom:
> @str4d did you see @nuttycom's comment here: [#1175 (review)](https://github.com/zcash/librustzcash/pull/1175#pullrequestreview-1925276413)

This isn't something that needs to be fixed right now, because at present `zcash_client_sqlite` depends upon the `orchard` feature being enabled for `zcash_keys` (because we initially created the key types before we make things feature-flaggable, and so existing wallets will already have Orchard keys.). When in the future we do our comprehensive feature-flagging pass, then this will end up as a compilation error, and we'll correctly flag it off.

PR #1241, opened 2024-03-08
zcash_client_sqlite: Generalize internal operations to simplify Orchard support

nuttycom:


zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -323,36 +324,25 @@ pub(crate) fn select_spendable_sapling_notes<P: consensus::Parameters>(
...
-        let nf_bytes: Vec<u8> = row.get(2)?;
-        Ok::<_, rusqlite::Error>((account, sapling::Nullifier::from_slice(&nf_bytes).unwrap()))
-    })?;
-
-    let res: Vec<_> = nullifiers.collect::<Result<_, _>>()?;
-    Ok(res)
-}
+    let mut stmt_fetch_nullifiers = match query {
+        NullifierQuery::Unspent => conn.prepare(
+            "SELECT rn.id, rn.account_id, rn.nf
```
str4d, line 332: Do we need to select `rn.id` anymore? IIRC that was because we used to pass around the note ID in an opaque struct, but we don't do that anymore. Is the ID select serving any remaining purpose?

str4d (approved):
utACK e257896462e7c7bcb45c742a27dfb1ba06db0d1b with question.

zcash_client_sqlite/src/wallet/sapling.rs
nuttycom: Looks like not; we don't query that index in the returned row.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e257896462e7c7bcb45c742a27dfb1ba06db0d1b..e2ac746e9d2e5f1fbe7cd2b97c38b6e83c165853) to address Clippy complaint and review comment.

str4d (approved):
Re-utACK e2ac746e9d2e5f1fbe7cd2b97c38b6e83c165853

PR #1243, opened 2024-03-09
zcash_client_sqlite: Only consider notes spendable when both seed fingerprint & key scope are available.

### LINKED ISSUE #1203: zcash_client_sqlite: Handle missing recipient key metadata.
An expectation is introduced in #1195 that will need to be removed and/or reworked as part of #1175.
 
_Originally posted by @str4d in https://github.com/zcash/librustzcash/pull/1195#discussion_r1505045842_

nuttycom:
Closes #1203

This PR is best viewed with whitespace changes hidden.

### DIFF
--- a/zcash_client_sqlite/src/wallet/sapling.rs
+++ b/zcash_client_sqlite/src/wallet/sapling.rs
@@ -8,10 +8,10 @@ use std::rc::Rc;
 use sapling::{self, Diversifier, Nullifier, Rseed};
 use zcash_client_backend::{
     data_api::NullifierQuery,
-    keys::UnifiedFullViewingKey,
     wallet::{Note, ReceivedNote, WalletSaplingOutput},
     DecryptedOutput, ShieldedProtocol, TransferType,
 };
+use zcash_keys::keys::UnifiedFullViewingKey;
 use zcash_primitives::transaction::{components::amount::NonNegativeAmount, TxId};
 use zcash_protocol::{
     consensus::{self, BlockHeight},
@@ -96,7 +96,7 @@ impl ReceivedSaplingOutput for DecryptedOutput<sapling::Note, AccountId> {
 fn to_spendable_note<P: consensus::Parameters>(
     params: &P,
     row: &Row,
-) -> Result<ReceivedNote<ReceivedNoteId, Note>, SqliteClientError> {
+) -> Result<Option<ReceivedNote<ReceivedNoteId, Note>>, SqliteClientError> {
     let note_id = ReceivedNoteId(ShieldedProtocol::Sapling, row.get(0)?);
     let txid = row.get::<_, [u8; 32]>(1).map(TxId::from_bytes)?;
     let output_index = row.get(2)?;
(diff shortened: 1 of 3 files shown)


zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -136,37 +136,51 @@ fn to_spendable_note<P: consensus::Parameters>(
...
-    ))
+    let seed_fingerprint = row
+        .get::<_, Option<[u8; 32]>>(7)?
+        .map(HdSeedFingerprint::from_bytes);
+    let ufvk_str: Option<String> = row.get(8)?;
+    let scope_code: Option<i64> = row.get(9)?;
+
+    // If we don't have information about the recipient key scope or the seed fingerprint, we can't
+    // determine which spending key to use. This may be because the received note was associated
+    // with an imported viewing key, so we treat such notes as not spendable.
```
str4d, line 147: Hmm: at the moment, _we_ can't determine that, but in theory the seed phrase holder might have a mapping from `InputSelector::AccountId` to USK. `create_proposed_transactions` only takes a USK, not a seed, so we can't usefully use the seed fingerprint.

However, when we eventually _are_ the Rust wallet (as part of `zcashd` deprecation), we will definitely need such a mapping ourselves, which we can use the seed FP for (caller will fetch the seed FP and ZIP 32 account index, and use those to determine which USK to derive from their seed). So I can see the rationale that for simplicity we treat imported UFVKs as view-only. If this is the direction we take, then when UFVK import is exposed we need to document it as such.

nuttycom: In the future, I think what we should probably do is take the seed fingerprint(s) as an argument when we go to initially request spendable inputs, so that we know what seed(s) we'll be spending from?

AArnott: I'm perplexed at the thought that a seed is assumed to have even been seed by the database in order to spend funds. An imported account won't provide a seed or fingerprint, yet it can absolutely be spendable. Rather than use the fingerprint and zip32 account index to ask for the USK, could we instead use the UFVK as the 'key' by which we ask for the USK?

str4d: UFVKs are not suitable for use as "map keys", due to being containers themselves. Once arbitrary UFVK import is allowed, we have no way of enforcing that its component FVKs are at all tied to the same ZIP 32 account or seed phrase. And when additional receivers are added in future, or metadata is added to a UFVK, the matching becomes more complex. In general, the result is that instead of using the UFVK as a key, you would need to do an arbitrary search for seed or key material that can reproduce each of the known component FVKs within a given UFVK, which is not great UX.

str4d: But as I said in my original comment, as long as we know the UFVK for an account (so we are acting as a wallet), and the surrounding app knows all spending key material for that UFVK, then we can create transactions. (Conversely, we _cannot_ ever do this for UIVKs, because we cannot track balance, so we cannot know whether any given note is spent or not.)

So if we can figure out an approach with reasonable UX for mapping arbitrary imported UFVKs to their key material, then we could support that. But I don't think we have an answer for this at the moment (at least, I do not have time to think of one on the Orchard implementation timeline)

nuttycom: I've removed the seed fingerprint check.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -136,37 +136,50 @@ fn to_spendable_note<P: consensus::Parameters>(
...
+    // If we don't have information about the recipient key scope or the ufvk we can't determine
+    // which spending key to use. This may be because the received note was associated with an
+    // imported viewing key, so we treat such notes as not spendable. Although this method is
+    // presently only called using the results of queries where both the ufvk and
+    // recipient_key_scope columns are checked to be non-null, this is method is written
+    // defensively to account for the fact that both of these are nullable columns in case it
+    // is used elsewhere in the future.
+    ufvk_str
+        .zip(scope_code)
+        .map(|(ufvk_str, scope_code)| {
```
str4d, line 151: We can always determine the scope when we have a UFVK _and are scanning with it_, so it could be considered an `SqliteClientError::CorruptedData` if one is null and the other is non-null, but it could also be either a programming error on the part of the caller, or a database that had a UIVK imported and then later a corresponding UFVK imported (which is not a use case I am advocating we support). So from a defensive implementation perspective I think this code is fine (we just ignore these notes instead of erroring).

str4d (approved):
utACK 5511bacf2504c42f928fda648a8d6b0c86dc4f49

PR #1244, opened 2024-03-09
Spending pool selection

nuttycom:
Builds on #1243


zcash_client_sqlite/src/wallet.rs
```diff
@@ -665,21 +694,49 @@ pub(crate) fn get_unified_full_viewing_keys<P: consensus::Parameters>(
...
+        )?
+        .collect::<Result<Vec<_>, _>>()?;
+
+    if accounts.len() > 1 {
+        Err(SqliteClientError::CorruptedData(
+            "Mutiple account records correspond to a single UFVK".to_owned(),
+        ))
+    } else {
+        Ok(accounts.first().copied())
+    }
```
nuttycom, line 739: This section is the important bit.

PR #1245, opened 2024-03-09
Add `UnifiedIncomingViewingKey` struct

AArnott:
- [x] Remove some fn implementations in the UFVK type and have it instead simply forward to the UIVK type. For example address generation needn't be implemented twice.
- [x] Update zcash_client_sqlite where it left placeholders for using the new UIVK type.
- [x] ~~Consider adding ZIP-316 rev. 1 support to both UFVK and UIVK types.~~
- [x] Add tests

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -74,7 +74,6 @@ use std::io::{self, Cursor};
 use std::num::NonZeroU32;
 use std::ops::RangeInclusive;
 use tracing::debug;
-use zcash_address::unified::{Encoding, Uivk};
 use zcash_keys::keys::{
     AddressGenerationError, HdSeedFingerprint, UnifiedAddressRequest, UnifiedIncomingViewingKey,
 };
@@ -120,7 +119,7 @@ use {
     crate::UtxoId,
     rusqlite::Row,
     std::collections::BTreeSet,
-    zcash_address::unified::Ivk,
+    zcash_address::unified::{Encoding, Ivk, Uivk},
     zcash_client_backend::wallet::{TransparentAddressMetadata, WalletTransparentOutput},
     zcash_primitives::{
         legacy::{
@@ -347,7 +346,7 @@ pub(crate) fn add_account<P: consensus::Parameters>(
             ":hd_seed_fingerprint": hd_seed_fingerprint.as_ref().map(|fp| fp.as_bytes()),
             ":hd_account_index": hd_account_index.map(u32::from),
             ":ufvk": viewing_key.ufvk().map(|ufvk| ufvk.encode(params)),
-            ":uivk": viewing_key.uivk().to_uivk().encode(&params.network_type()),
+            ":uivk": viewing_key.uivk().encode(params),
             ":orchard_fvk_item_cache": orchard_item,
             ":sapling_fvk_item_cache": sapling_item,
             ":p2pkh_fvk_item_cache": transparent_item,
@@ -1503,18 +1502,9 @@ pub(crate) fn get_account<P: Parameters>(
                 ))
             } else {
                 let uivk_str: String = row.get("uivk")?;
-                let (network, uivk) = Uivk::decode(&uivk_str).map_err(|e| {
-                    SqliteClientError::CorruptedData(format!("Failure to decode UIVK: {e}"))
-                })?;
-                if network != params.network_type() {
-                    return Err(SqliteClientError::CorruptedData(
-                        "UIVK network type does not match wallet network type".to_string(),
-                    ));
-                }
                 ViewingKey::Incoming(Box::new(
-                    UnifiedIncomingViewingKey::from_uivk(&uivk).map_err(|e| {
-                        SqliteClientError::CorruptedData(format!("Failure to decode UIVK: {e}"))
-                    })?,
+                    UnifiedIncomingViewingKey::decode(params, &uivk_str[..])
+                        .map_err(SqliteClientError::BadAccountData)?,
                 ))
             };
 
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -208,7 +208,7 @@ impl Account {
     ) -> Result<(UnifiedAddress, DiversifierIndex), AddressGenerationError> {
         match &self.viewing_key {
             ViewingKey::Full(ufvk) => ufvk.default_address(request),
-            ViewingKey::Incoming(_uivk) => todo!(),
+            ViewingKey::Incoming(uivk) => uivk.default_address(request),
         }
     }
 }
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -235,12 +235,10 @@ impl ViewingKey {
         }
     }
 
-    fn uivk(&self) -> Result<UnifiedIncomingViewingKey, SqliteClientError> {
+    fn uivk(&self) -> UnifiedIncomingViewingKey {
         match self {
-            ViewingKey::Full(ufvk) => Ok(ufvk
-                .to_unified_incoming_viewing_key()
-                .map_err(|e| SqliteClientError::CorruptedData(e.to_string()))?),
-            ViewingKey::Incoming(uivk) => Ok((**uivk).to_owned()),
+            ViewingKey::Full(ufvk) => ufvk.as_ref().to_unified_incoming_viewing_key(),
+            ViewingKey::Incoming(uivk) => uivk.as_ref().clone(),
         }
     }
 }
@@ -349,7 +347,7 @@ pub(crate) fn add_account<P: consensus::Parameters>(
             ":hd_seed_fingerprint": hd_seed_fingerprint.as_ref().map(|fp| fp.as_bytes()),
             ":hd_account_index": hd_account_index.map(u32::from),
             ":ufvk": viewing_key.ufvk().map(|ufvk| ufvk.encode(params)),
-            ":uivk": viewing_key.uivk()?.to_uivk().encode(&params.network_type()),
+            ":uivk": viewing_key.uivk().to_uivk().encode(&params.network_type()),
             ":orchard_fvk_item_cache": orchard_item,
             ":sapling_fvk_item_cache": sapling_item,
             ":p2pkh_fvk_item_cache": transparent_item,
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -74,7 +74,7 @@ use std::io::{self, Cursor};
 use std::num::NonZeroU32;
 use std::ops::RangeInclusive;
 use tracing::debug;
-use zcash_address::unified::{Encoding, Ivk, Uivk};
+use zcash_address::unified::{Encoding, Uivk};
 use zcash_keys::keys::{
     AddressGenerationError, HdSeedFingerprint, UnifiedAddressRequest, UnifiedIncomingViewingKey,
 };
@@ -120,6 +120,7 @@ use {
     crate::UtxoId,
     rusqlite::Row,
     std::collections::BTreeSet,
+    zcash_address::unified::Ivk,
     zcash_client_backend::wallet::{TransparentAddressMetadata, WalletTransparentOutput},
     zcash_primitives::{
         legacy::{
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -298,14 +298,13 @@ struct AccountSqlValues<'a> {
     hd_seed_fingerprint: Option<&'a [u8]>,
     hd_account_index: Option<u32>,
     ufvk: Option<&'a UnifiedFullViewingKey>,
-    uivk: String,
+    uivk: UnifiedIncomingViewingKey,
 }
 
 /// Returns (account_type, hd_seed_fingerprint, hd_account_index, ufvk, uivk) for a given account.
-fn get_sql_values_for_account_parameters<'a, P: consensus::Parameters>(
-    account: &'a Account,
-    params: &P,
-) -> Result<AccountSqlValues<'a>, SqliteClientError> {
+fn get_sql_values_for_account_parameters(
+    account: &Account,
+) -> Result<AccountSqlValues, SqliteClientError> {
     Ok(match account {
         Account::Zip32(hdaccount) => AccountSqlValues {
             account_type: AccountType::Zip32.into(),
@@ -315,9 +314,7 @@ fn get_sql_values_for_account_parameters<'a, P: consensus::Parameters>(
             uivk: hdaccount
                 .ufvk()
                 .to_unified_incoming_viewing_key()
-                .map_err(|e| SqliteClientError::CorruptedData(e.to_string()))?
-                .to_uivk()
-                .encode(&params.network_type()),
+                .map_err(|e| SqliteClientError::CorruptedData(e.to_string()))?,
         },
         Account::Imported(ImportedAccount::Full(ufvk)) => AccountSqlValues {
             account_type: AccountType::Imported.into(),
@@ -326,16 +323,14 @@ fn get_sql_values_for_account_parameters<'a, P: consensus::Parameters>(
             ufvk: Some(ufvk),
             uivk: ufvk
                 .to_unified_incoming_viewing_key()
-                .map_err(|e| SqliteClientError::CorruptedData(e.to_string()))?
-                .to_uivk()
-                .encode(&params.network_type()),
+                .map_err(|e| SqliteClientError::CorruptedData(e.to_string()))?,
         },
         Account::Imported(ImportedAccount::Incoming(uivk)) => AccountSqlValues {
             account_type: AccountType::Imported.into(),
             hd_seed_fingerprint: None,
             hd_account_index: None,
             ufvk: None,
-            uivk: uivk.to_uivk().encode(&params.network_type()),
+            uivk: (**uivk).clone(),
         },
     })
 }
@@ -346,7 +341,7 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     account: Account,
     birthday: AccountBirthday,
 ) -> Result<AccountId, SqliteClientError> {
-    let args = get_sql_values_for_account_parameters(&account, params)?;
+    let args = get_sql_values_for_account_parameters(&account)?;
 
     let orchard_item = args
         .ufvk
@@ -382,7 +377,7 @@ pub(crate) fn add_account<P: consensus::Parameters>(
             ":hd_seed_fingerprint": args.hd_seed_fingerprint,
             ":hd_account_index": args.hd_account_index,
             ":ufvk": args.ufvk.map(|ufvk| ufvk.encode(params)),
-            ":uivk": args.uivk,
+            ":uivk": args.uivk.to_uivk().encode(&params.network_type()),
             ":orchard_fvk_item_cache": orchard_item,
             ":sapling_fvk_item_cache": sapling_item,
             ":p2pkh_fvk_item_cache": transparent_item,
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -75,7 +75,9 @@ use std::num::NonZeroU32;
 use std::ops::RangeInclusive;
 use tracing::debug;
 use zcash_address::unified::{Encoding, Ivk, Uivk};
-use zcash_keys::keys::{AddressGenerationError, HdSeedFingerprint, UnifiedAddressRequest};
+use zcash_keys::keys::{
+    AddressGenerationError, HdSeedFingerprint, UnifiedAddressRequest, UnifiedIncomingViewingKey,
+};
 
 use zcash_client_backend::{
     address::{Address, UnifiedAddress},
@@ -200,7 +202,7 @@ pub(crate) enum ImportedAccount {
     /// An account that was imported via its full viewing key.
     Full(Box<UnifiedFullViewingKey>),
     /// An account that was imported via its incoming viewing key.
-    Incoming(Uivk),
+    Incoming(Box<UnifiedIncomingViewingKey>),
 }
 
 /// Describes an account in terms of its UVK or ZIP-32 origins.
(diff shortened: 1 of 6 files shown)


zcash_keys/src/keys.rs
```diff
@@ -928,6 +928,351 @@ impl UnifiedFullViewingKey {
...
+    transparent: Option<zcash_primitives::legacy::keys::ExternalIvk>,
+    #[cfg(feature = "sapling")]
+    sapling: Option<::sapling::zip32::IncomingViewingKey>,
+    #[cfg(feature = "orchard")]
+    orchard: Option<orchard::keys::IncomingViewingKey>,
+    /// Stores the unrecognized elements of the unified encoding.
+    unknown: Vec<(u32, Vec<u8>)>,
+}
+
+impl TryFrom<UnifiedFullViewingKey> for UnifiedIncomingViewingKey {
```
nuttycom, line 944: Instead of `TryFrom`, this should just be a method on `UnifiedFullViewingKey`.

zcash_keys/src/keys.rs
```diff
@@ -928,6 +928,351 @@ impl UnifiedFullViewingKey {
...
+            })
+        } else {
+            None
+        }
+    }
+
+    /// Parses a `UnifiedIncomingViewingKey` from its [ZIP 316] string encoding.
+    ///
+    /// [ZIP 316]: https://zips.z.cash/zip-0316
+    pub fn decode<P: consensus::Parameters>(params: &P, encoding: &str) -> Result<Self, String> {
```
nuttycom, line 1002: Instead of `String`, please create a new error enum.

nuttycom: Also, instead of encoding and decoding functions that go to and from strings, these conversions should go to and from the `zcash_address::unified::Uivk` type. That then provides the conversion to and from strings.

zcash_keys/src/keys.rs
```diff
@@ -928,6 +928,351 @@ impl UnifiedFullViewingKey {
...
+    #[cfg(feature = "orchard")]
+    fn orchard(&self) -> &Option<orchard::keys::IncomingViewingKey> {
+        &self.orchard
+    }
+
+    /// Attempts to derive the Unified Address for the given diversifier index and
+    /// receiver types.
+    ///
+    /// Returns `None` if the specified index does not produce a valid diversifier.
+    pub fn address(
```
nuttycom, line 1137: In order to prevent duplication, the `UnifiedFullViewingKey` methods that this is derived from should now delegate to this method.

zcash_keys/src/keys.rs
```diff
@@ -928,6 +928,351 @@ impl UnifiedFullViewingKey {
...
+    }
+
+    /// Searches the diversifier space starting at diversifier index `j` for one which will
+    /// produce a valid diversifier, and return the Unified Address constructed using that
+    /// diversifier along with the index at which the valid diversifier was found.
+    ///
+    /// Returns an `Err(AddressGenerationError)` if no valid diversifier exists or if the features
+    /// required to satisfy the unified address request are not properly enabled.
+    #[allow(unused_mut)]
+    pub fn find_address(
```
nuttycom, line 1227: Same thing here, modify `UnifiedFullViewingKey`'s impl to construct the uivk and delegate to this method.

AArnott: Great. That was the plan (see PR description).

zcash_keys/src/keys.rs
AArnott: Heh heh. Ya, I wrote this before reading your feelings on the From trait on discord last night. I figured you'd comment on this if I didn't change it first. That's fine.

zcash_keys/src/keys.rs
AArnott: Ya, I feel string error types are gross. I only did this in the copy, but I was hoping you'd be ok with me changing it. Can I change in on the UFVK type as well?

zcash_keys/src/keys.rs
AArnott: I absolutely agree, as per my PR description's checklist.

AArnott:
@nuttycom What does `#[doc(hidden)]` do? I see it on the UFVK struct and I didn't copy it down to UIVK.

zcash_keys/src/keys.rs
```diff
@@ -652,16 +685,7 @@ impl UnifiedFullViewingKey {
     /// Parses a `UnifiedFullViewingKey` from its [ZIP 316] string encoding.
     ///
     /// [ZIP 316]: https://zips.z.cash/zip-0316
-    pub fn decode<P: consensus::Parameters>(params: &P, encoding: &str) -> Result<Self, String> {
```
nuttycom, line 655: Since this is an existing API, we probably don't want to entirely remove it; it can remain as a convenience wrapper around the `zcash_address` based approach. We have such convenience methods in a number of places; the address generation methods on this type will be of that sort after this change.

zcash_keys/src/keys.rs
```diff
@@ -732,7 +756,7 @@ impl UnifiedFullViewingKey {
     }
 
     /// Returns the string encoding of this `UnifiedFullViewingKey` for the given network.
-    pub fn encode<P: consensus::Parameters>(&self, params: &P) -> String {
```
nuttycom, line 735: Same thing here, keep `encode` but just have it implemented in terms of `to_ufvk`.

zcash_keys/src/keys.rs
```diff
@@ -791,43 +833,262 @@ impl UnifiedFullViewingKey {
     /// Returns `None` if the specified index does not produce a valid diversifier.
     pub fn address(
         &self,
-        _j: DiversifierIndex,
-        _request: UnifiedAddressRequest,
+        j: DiversifierIndex,
+        request: UnifiedAddressRequest,
+    ) -> Result<UnifiedAddress, AddressGenerationError> {
+        Ok(self
+            .to_unified_incoming_viewing_key()?
```
nuttycom, line 840: :+1:

zcash_keys/src/keys.rs
```diff
@@ -791,43 +833,262 @@ impl UnifiedFullViewingKey {
...
-        if _request.has_sapling {
+        if request.has_sapling {
             #[cfg(not(feature = "sapling"))]
             return Err(AddressGenerationError::ReceiverTypeNotSupported(
                 Typecode::Sapling,
             ));
 
             #[cfg(feature = "sapling")]
-            if let Some(extfvk) = &self.sapling {
+            if let Some(extivk) = &self.sapling {
```
nuttycom, line 1083: nonblocking nit: the `ext` here is unnecessary (and was actually kind of incorrect previously, as it should have been named `divk`).

zcash_keys/src/keys.rs
```diff
@@ -1053,7 +1317,7 @@ mod tests {
         #[cfg(any(feature = "orchard", feature = "sapling"))]
         {
             let ufvk = ufvk.expect("Orchard or Sapling fvk is present.");
-            let encoded = ufvk.encode(&MAIN_NETWORK);
+            let encoded = ufvk.to_ufvk().encode(&Network::Main);
```
nuttycom, line 1320: This change won't be necessary if the convenience methods are left in place.

zcash_keys/src/keys.rs
```diff
@@ -492,6 +495,29 @@ pub enum AddressGenerationError {
...
+}
+
+impl From<hdwallet::error::Error> for AddressGenerationError {
+    fn from(e: hdwallet::error::Error) -> Self {
+        match e {
+            hdwallet::error::Error::KeyIndexOutOfRange => todo!(),
+            hdwallet::error::Error::ChainPath(_) => todo!(),
+            hdwallet::error::Error::Secp(_) => todo!(),
+            hdwallet::error::Error::Rng(_) => todo!(),
+        }
```
nuttycom, line 519: Instead of this, just wrap the entire `hdwallet::error::Error` in a new `AddressGenerationError` variant.

zcash_keys/src/keys.rs
AArnott: I like the idea of the convenience methods. But you _don't_ want them added to the new type?

nuttycom (dismissed):
This looks great, it looks to me like the hdwallet AddressGenerationError and keeping the `decode` and `encode` convenience methods (to limit API breakage slightly) are the only blocking issues.

zcash_keys/src/keys.rs
AArnott: I can't, because the hdwallet Error enum doesn't derive the same traits as `AddressGenerationError`. I'd love to though. Time for an official fork of hdwallet?

nuttycom:
Also, add tests that verify against https://github.com/zcash/zcash-test-vectors/blob/master/test-vectors/rust/unified_incoming_viewing_keys.rs

nuttycom:
> @nuttycom Why is `#[doc(hidden)]` applied to the existing `UnifiedFullViewingKey` struct? Should I remove it at this point, or add it to the new `UnifiedIncomingViewingKey` struct?

Oh, yeah, we should definitely remove that and have both of these publicly documented.

zcash_keys/src/keys.rs
nuttycom: I'm indifferent; if you want them on the new type then adding them there is fine.

zcash_keys/src/keys.rs
nuttycom: What traits would be restricted to? We may not be using all of the traits we derive.

AArnott: `hdwallet` derives `Error` and `Debug`. `AddressGenerationError` adds `Clone`.

zcash_keys/src/keys.rs
AArnott: Great. I like consistency. Folks who switch from UFVK to IFVK for good hygiene ideally don't have to then figure out which functions on those types to switch to.

AArnott: Although, then I would have to replicate the `string` error result if the goal was really consistency, and I feel gross doing that. So maybe I'll leave them off.

zcash_keys/src/keys.rs
AArnott: Removing the `Clone` trait didn't break anything within librustzcash itself. So as it sounds like you were suggesting that approach, I'm taking it.

zcash_keys/src/keys.rs
```diff
@@ -785,48 +881,262 @@ impl UnifiedFullViewingKey {
...
+                    })
+                    .transpose(),
+                #[cfg(not(feature = "transparent-inputs"))]
+                unified::Ivk::P2pkh(data) => Some(Ok::<_, UnifiedError>((
+                    u32::from(unified::Typecode::P2pkh),
+                    data.to_vec(),
+                ))),
+                unified::Ivk::Unknown { typecode, data } => Some(Ok((*typecode, data.clone()))),
+            })
+            .collect::<Result<_, _>>()?;
```
nuttycom, line 1034:
```suggestion

        let mut unknown = vec![];

        // We can use as-parsed order here for efficiency, because we're breaking out the
        // receivers we support from the unknown receivers.
        for receiver in uivk.items_as_parsed() {
            match receiver {
                unified::Ivk::Orchard(data) => {
                    #[cfg(feature = "orchard")]
                    {
                        orchard = Some(
                            Option::from(orchard::keys::IncomingViewingKey::from_bytes(data))
                                .ok_or(UnifiedError::InvalidShieldedKey(
                                    ShieldedProtocol::Orchard,
                                ))?,
                        );
                    }

                    #[cfg(feature = "orchard")]
                    unknown.push((u32::from(unified::Typecode::Orchard), data.to_vec()));
                }
                unified::Ivk::Sapling(data) => {
                    #[cfg(feature = "sapling")]
                    {
                        sapling = Some(
                            Option::from(::sapling::zip32::IncomingViewingKey::from_bytes(data))
                                .ok_or(UnifiedError::InvalidShieldedKey(
                                    ShieldedProtocol::Sapling,
                                ))?,
                        );
                    }

                    #[cfg(not(feature = "sapling"))]
                    unknown.push((u32::from(unified::Typecode::Sapling), data.to_vec()));
                }
                unified::Ivk::P2pkh(data) => {
                    #[cfg(feature = "transparent-inputs")]
                    {
                        transparent = Some(
                            legacy::ExternalIvk::deserialize(data)
                                .map_err(UnifiedError::InvalidTransparentKey)?,
                        );
                    }

                    #[cfg(not(feature = "transparent-inputs"))]
                    unknown.push((u32::from(unified::Typecode::P2pkh), data.to_vec()));
                }
                unified::Ivk::Unknown { typecode, data } => {
                    unknown.push((*typecode, data.clone()));
                }
            }
        }
```

The style that this was done in for the ufvk doesn't really make much sense; the direct mutation approach is much clearer. I had made a similar change to the ufvk handling in #1135 so we shouldn't reproduce the ugly approach here.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -310,50 +312,34 @@ fn get_sql_values_for_account_parameters<'a, P: consensus::Parameters>(
             hd_seed_fingerprint: Some(hdaccount.hd_seed_fingerprint().as_bytes()),
             hd_account_index: Some(hdaccount.account_index().into()),
             ufvk: Some(hdaccount.ufvk()),
-            uivk: ufvk_to_uivk(hdaccount.ufvk(), params)?,
+            uivk: hdaccount
+                .ufvk()
+                .to_unified_incoming_viewing_key()
+                .map_err(|e| SqliteClientError::CorruptedData(e.to_string()))?
+                .to_uivk()
+                .encode(&params.network_type()),
```
nuttycom, line 320: For consistency with the `ufvk`, should this encoding be done at the use site?

zcash_keys/src/keys.rs
```diff
@@ -601,9 +633,43 @@ impl UnifiedAddressRequest {
     }
 }
 
+/// Errors that may be returned from the [UnifiedFullViewingKey] and [UnifiedIncomingViewingKey] types.
+#[derive(Debug)]
+pub enum UnifiedError {
```
nuttycom, line 638: Naming bikeshed: `UnifiedError` doesn't mean very much to me, maybe `DerivationError` would be better? Or `KeyDerivationError`?

zcash_keys/src/keys.rs
```diff
@@ -492,6 +496,26 @@ pub enum AddressGenerationError {
     /// A Unified address cannot be generated without at least one shielded receiver being
     /// included.
     ShieldedReceiverRequired,
+
+    // An error occurred while deriving a key or address from an HD wallet.
+    UnifiedError(UnifiedError),
```
nuttycom, line 501: I would prefer this be `AddressGenerationError::Derivation`

zcash_keys/src/keys.rs
```diff
@@ -492,6 +496,26 @@ pub enum AddressGenerationError {
...
+
+    // An error occurred while deriving a transparent key or address from an HD wallet.
+    #[cfg(feature = "transparent-inputs")]
+    HDWalletError(hdwallet::error::Error),
```
nuttycom, line 505: Given that (with the renaming) `DerivationError::HdWallet` exists, delete this variant and use `Derivation` variant instead.

zcash_keys/src/keys.rs
```diff
@@ -601,9 +633,43 @@ impl UnifiedAddressRequest {
...
 }
 
+/// Errors that may be returned from the [UnifiedFullViewingKey] and [UnifiedIncomingViewingKey] types.
+#[derive(Debug)]
+pub enum UnifiedError {
+    InvalidShieldedKey(ShieldedProtocol),
+    #[cfg(feature = "transparent-inputs")]
+    InvalidTransparentKey(hdwallet::error::Error),
+    #[cfg(feature = "transparent-inputs")]
+    HDWallet(hdwallet::error::Error),
```
nuttycom, line 643:
```suggestion
    HdWallet(hdwallet::error::Error),
```
By Rust convention.

nuttycom (changes requested):
One minor naming fix, otherwise looks good!

zcash_keys/src/keys.rs
AArnott: I'm glad you brought it up. I wasn't settled on it either. There is another enum error type in this file--is it worth combining them?

nuttycom: Oh, right, yes, the existing `DerivationError` type should just be modified.

AArnott:
The second round of merge conflicts with main caused quite a bit of churn. But the PR is still pretty simple.

nuttycom (approved):
utACK 4d9927b993b10ae407952efd1bcbc423eb532b22

nuttycom (dismissed):
utACK b4171caaf43514b5a51d94c4b6b764a70bbec805

zcash_client_sqlite/src/wallet.rs
```diff
@@ -232,10 +235,10 @@ impl ViewingKey {
         }
     }
```
daira, line 236: The `todo!()` on line 211 should be implemented. It should be straightforward given that you already have `UnifiedIncomingViewingKey::default_address(UnifiedAddressRequest)` in this PR:
```rust
    pub(crate) fn default_address(
        &self,
        request: UnifiedAddressRequest,
    ) -> Result<(UnifiedAddress, DiversifierIndex), AddressGenerationError> {
        match &self.viewing_key {
            ViewingKey::Full(ufvk) => ufvk.default_address(request),
            ViewingKey::Incoming(uivk) => uivk.default_address(request),
        }
    }
```
And then it might as well be public, I think.

zcash_keys/src/address.rs
```diff
@@ -342,7 +342,14 @@ impl Address {
...
 
-#[cfg(any(test, feature = "test-dependencies"))]
+#[cfg(all(
+    any(
+        feature = "orchard",
+        feature = "sapling",
+        feature = "transparent-inputs"
+    ),
+    any(test, feature = "test-dependencies")
+))]
```
daira, line 352: I don't think we support enabling neither `sapling` nor `orchard`, do we?

nuttycom: These are feature flags, so we don't have a way (that I know of) to enforce that at least one be enabled at the build system level.

AArnott: I suspect a `#[cfg(not(any(` type attribute _might_ allow you to introduce a compile break if you were interested in that, but I haven't tried it.

zcash_client_sqlite/src/wallet.rs
AArnott: Quite right. That was one of the main points of this PR and I'm shocked I missed it. 
Maybe in the merge conflicts, I lost it.

nuttycom (approved):
re-utACK 25045b3914ed539cab375509741e1037445a3f64

zcash_keys/src/address.rs
nuttycom: In any case, with ZIP 316 Revision 1, transparent-only UFVKs will be allowed (for use with metadata.)

nuttycom (approved):
re-utACK 6102e83b07a3c158891b85bbd34f2c7563116559

zcash_client_sqlite/src/wallet.rs
```diff
@@ -206,7 +206,7 @@ impl Account {
     ) -> Result<(UnifiedAddress, DiversifierIndex), AddressGenerationError> {
```
daira, line 206: This method can be public.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -233,10 +233,10 @@ impl ViewingKey {
         }
     }
 
-    fn uivk_str(&self, params: &impl Parameters) -> Result<String, SqliteClientError> {
+    fn uivk(&self) -> UnifiedIncomingViewingKey {
```
daira, line 236:
```suggestion
    pub fn uivk(&self) -> UnifiedIncomingViewingKey {
```
(with changelog entry).

zcash_client_sqlite/src/wallet.rs
nuttycom: In general we would like to avoid encouraging users to use the `default_address` method. In any case, this is not a change that should be made in this PR.

zcash_keys/src/address.rs
daira: They will, but there's still no point in *none* of "orchard", "sapling", or "transparent-inputs" being enabled. That should be a compilation error IMHO.

nuttycom: You could still use this crate to parse `TransparentAddress` values with metadata, so that you can send to them properly. `transparent-inputs` only applies to key and address _generation_, not parsing.

str4d: The `transparent-inputs` flag is not any different to the`sapling` and `orchard` flags. All three have the purpose of removing the need for certain dependencies; in the case of `transparent-inputs` it is the `secp256k1` crate. It's only because we solely support P2PKH and P2SH outputs that we only flag off transparent inputs; if we had to support arbitrary transparent output generation then we'd likely need `secp256k1` for that, and we'd need a; more general transparent flag.

zcash_keys/src/keys.rs
```diff
@@ -999,105 +1323,277 @@ mod tests {
...
+        let encoded = uivk.render().encode(&NetworkType::Main);
+
+        // Test encoded form against known values; these test vectors contain Orchard receivers
+        // that will be treated as unknown if the `orchard` feature is not enabled.
+        let encoded_with_t = "uivk1z28yg638vjwusmf0zc9ad2j0mpv6s42wc5kqt004aaqfu5xxxgu7mdcydn9qf723fnryt34s6jyxyw0jt7spq04c3v9ze6qe9gjjc5aglz8zv5pqtw58czd0actynww5n85z3052kzgy6cu0fyjafyp4sr4kppyrrwhwev2rr0awq6m8d66esvk6fgacggqnswg5g9gkv6t6fj9ajhyd0gmel4yzscprpzduncc0e2lywufup6fvzf6y8cefez2r99pgge5yyfuus0r60khgu895pln5e7nn77q6s9kh2uwf6lrfu06ma2kd7r05jjvl4hn6nupge8fajh0cazd7mkmz23t79w";
+        let _encoded_no_t = "uivk1020vq9j5zeqxh303sxa0zv2hn9wm9fev8x0p8yqxdwyzde9r4c90fcglc63usj0ycl2scy8zxuhtser0qrq356xfy8x3vyuxu7f6gas75svl9v9m3ctuazsu0ar8e8crtx7x6zgh4kw8xm3q4rlkpm9er2wefxhhf9pn547gpuz9vw27gsdp6c03nwlrxgzhr2g6xek0x8l5avrx9ue9lf032tr7kmhqf3nfdxg7ldfgx6yf09g";
+
+        // We test the full roundtrip only with the `sapling` and `orchard` features enabled,
+        // because we will not generate these parts of the encoding if the UIVK does not have an
```
daira, line 1513:
```suggestion
        // because we will not generate these parts of the encoding if the UIVK does not have
```
Non-blocking.

daira (approved):
utACK

PR #1250, opened 2024-03-09
zcash_client_sqlite: Run single-shielded-pool tests on Orchard

str4d:
Almost all of these tests currently fail, as expected.

This PR also makes some fixes and improvements to the test harness, that arose from debugging the new Orchard code.


nuttycom (approved):
utACK

PR #1251, opened 2024-03-10
CI: Rework the test states

str4d:


.github/workflows/ci.yml
```diff
@@ -82,6 +77,63 @@ jobs:
...
+        include:
+          - target: Linux
+            os: ubuntu-latest
+          - target: macOS
+            os: macOS-latest
+          - target: Windows
+            os: windows-latest
+
+          - state: ZFuture
+            extra_flags: zfuture
```
nuttycom: Weird formatting?

nuttycom (approved):
utACK

.github/workflows/ci.yml
str4d: Intentional: "ZFuture" is the state name (shown in CI status name), `zfuture` is (what was) the feature flag.

PR #1252, opened 2024-03-10
Various fixes to Orchard tests

str4d:


zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1178,10 +1198,18 @@ pub(crate) mod tests {
 
         // We've crossed a subtree boundary, and so still only have one scanned note but have two
         // shards worth of notes to scan.
+        let expected_denom = expected_denom
+            + match T::SHIELDED_PROTOCOL {
+                ShieldedProtocol::Sapling => 1 << SAPLING_SHARD_HEIGHT,
+                #[cfg(feature = "orchard")]
+                ShieldedProtocol::Orchard => 1 << ORCHARD_SHARD_HEIGHT,
+                #[cfg(not(feature = "orchard"))]
+                ShieldedProtocol::Orchard => unreachable!(),
```
nuttycom, line 1207: Why is this `unreachable` instead of 0?

str4d: If `T::SHIELDED_PROTOCOL` is `ShieldedProtocol::Orchard`, then that means we are using `OrchardPoolTester` which is only available when the `orchard` feature is enabled. It should be a programming error to reach here.

nuttycom (approved):
utACK

PR #1182, opened 2024-02-08
`zcash_client_sqlite`: Add Orchard wallet support

nuttycom:
Closes #1087.
Closes #1177.

PR #1254, opened 2024-03-11
zcash_client_sqlite: Add database tables for Orchard

### LINKED ISSUE #1227: zcash_client_sqlite: Add database tables for Orchard

str4d:
Closes #1227.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
@@ -120,6 +120,26 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
             priority_code(&ScanPriority::Scanned),
         ))?;
 
+        transaction.execute_batch(
+            "CREATE VIEW v_orchard_shards_scan_state AS
+            SELECT
+                shard_index,
+                start_position,
+                end_position_exclusive,
+                subtree_start_height,
+                subtree_end_height,
+                contains_marked,
+                MAX(priority) AS max_priority
+            FROM v_orchard_shard_scan_ranges
+            GROUP BY
+                shard_index,
+                start_position,
+                end_position_exclusive,
+                subtree_start_height,
+                subtree_end_height,
+                contains_marked;",
+        )?;
+
         // Treat the current best-known chain tip height as the height to use for Orchard
         // initialization, bounded below by NU5 activation.
         if let Some(orchard_init_height) = scan_queue_extrema(transaction)?.and_then(|r| {
--- a/zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
@@ -1,0 +1,196 @@
+//! This migration adds tables to the wallet database that are needed to persist Orchard note
+//! commitment tree data using the `shardtree` crate.
+
+use std::collections::HashSet;
+
+use rusqlite::{named_params, OptionalExtension};
+use schemer_rusqlite::RusqliteMigration;
+use tracing::debug;
+use uuid::Uuid;
+use zcash_client_backend::data_api::scanning::ScanPriority;
+use zcash_protocol::consensus::{self, BlockHeight, NetworkUpgrade};
+
+use super::shardtree_support;
+use crate::wallet::{init::WalletMigrationError, scan_queue_extrema, scanning::priority_code};
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x3a6487f7_e068_42bb_9d12_6bb8dbe6da00);
+
+pub(super) struct Migration<P> {
+    pub(super) params: P,
+}
+
+impl<P> schemer::Migration for Migration<P> {
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        [shardtree_support::MIGRATION_ID].into_iter().collect()
+    }
+
+    fn description(&self) -> &'static str {
+        "Add support for storage of Orchard note commitment tree data using the `shardtree` crate."
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        // Add shard persistence
+        debug!("Creating tables for Orchard shard persistence");
+        transaction.execute_batch(
+            "CREATE TABLE orchard_tree_shards (
+                shard_index INTEGER PRIMARY KEY,
+                subtree_end_height INTEGER,
+                root_hash BLOB,
+                shard_data BLOB,
+                contains_marked INTEGER,
+                CONSTRAINT root_unique UNIQUE (root_hash)
+            );
+            CREATE TABLE orchard_tree_cap (
+                -- cap_id exists only to be able to take advantage of `ON CONFLICT`
+                -- upsert functionality; the table will only ever contain one row
+                cap_id INTEGER PRIMARY KEY,
+                cap_data BLOB NOT NULL
+            );",
+        )?;
+
+        // Add checkpoint persistence
+        debug!("Creating tables for checkpoint persistence");
+        transaction.execute_batch(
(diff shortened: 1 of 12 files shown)


nuttycom:
nonblocking: We will need to add CHANGELOG entries describing the new views.

nuttycom (approved):
utACK; we will need to add CHANGELOG entries for the views

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
...
+                    ],
+                )?;
+                // Rewrite the remainder of the range to have priority `Historic`
+                transaction.execute(
+                    "INSERT INTO scan_queue (block_range_start, block_range_end, priority)
+                     VALUES (:block_range_start, :block_range_end, :priority)",
+                    named_params![
+                        ":block_range_start": u32::from(orchard_init_height),
+                        ":block_range_end": u32::from(end),
+                        ":priority": priority_code(&ScanPriority::Historic),
```
str4d, line 190: This should not decrease the existing priority (as the range we select could have any priority):
```suggestion
                        ":priority": priority_code(&range_priority.max(ScanPriority::Historic)),
```
(or whatever syntax works).

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
...
+                transaction.execute(
+                    "INSERT INTO scan_queue (block_range_start, block_range_end, priority)
+                     VALUES (:block_range_start, :block_range_end, :priority)",
+                    named_params![
+                        ":block_range_start": u32::from(start),
+                        ":block_range_end": u32::from(orchard_init_height),
+                        ":priority": range_priority,
+                    ],
+                )?;
+                // Rewrite the remainder of the range to have priority `Historic`
```
str4d, line 183:
```suggestion
                // Rewrite the remainder of the range to have at least priority `Historic`
```

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
...
+                transaction.execute(
+                    "INSERT INTO scan_queue (block_range_start, block_range_end, priority)
+                     VALUES (:block_range_start, :block_range_end, :priority)",
+                    named_params![
+                        ":block_range_start": u32::from(orchard_init_height),
+                        ":block_range_end": u32::from(end),
```
str4d, line 189: This is fine, because we gate the select on `block_range_end > :orchard_init_height`, meaning that this range is guaranteed to have length `>= 1`.

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
...
+                // Rewrite any scanned ranges above the end of the first Orchard
+                // range to have priority `Historic`
+                transaction.execute(
+                    "UPDATE scan_queue SET priority = :historic
+                     WHERE :block_range_start >= :orchard_initial_range_end
+                     AND priority = :scanned",
+                    named_params![
+                        ":historic": priority_code(&ScanPriority::Historic),
+                        ":orchard_initial_range_end": u32::from(end),
+                        ":scanned": priority_code(&ScanPriority::Scanned),
```
str4d, line 202: Is it okay to only bump `Scanned` ranges, and not `Ignored` ranges? I think so, because in this migration we are assuming that the "Orchard birthday" is after the wallet birthday, and IIRC `Ignored` ranges are only created between Sapling activation and the wallet birthday.

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
...
+        // Treat the current best-known chain tip height as the height to use for Orchard
+        // initialization, bounded below by NU5 activation.
+        if let Some(orchard_init_height) = scan_queue_extrema(transaction)?.and_then(|r| {
+            self.params
+                .activation_height(NetworkUpgrade::Nu5)
+                .map(|orchard_activation| std::cmp::max(orchard_activation, *r.end()))
+        }) {
+            // If a scan range exists that contains the Orchard init height, split it in two at the
+            // init height.
+            if let Some((start, end, range_priority)) = transaction
```
str4d, line 152: If I'm reading these correctly, then AFAICT we will _never_ see any scan range above the range that contains `orchard_init_height`, because the range we select will always be the last range. So the code below to "[r]ewrite any scanned ranges above the end of the first Orchard range to have priority `Historic`" could instead just verify that no such ranges exist (and return `CorruptedData` if they do).

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
+//! This migration adds tables to the wallet database that are needed to persist Orchard note
+//! commitment tree data using the `shardtree` crate.
+
+use rusqlite::{self, named_params, OptionalExtension};
```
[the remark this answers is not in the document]
nuttycom: For some reason I thought that this didn't compile? Anyway, if it works this is fine.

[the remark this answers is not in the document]
str4d: `rusqlite` is in the global namespace, as are all dependencies.

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
nuttycom: This is a result of the evolution of the migration over time. Initially, I thought that the migration should be constructed with an Orchard start height, defaulting to `max(orchard_start_height, nu5_activation)` so that existing keys with Orchard funds could have them become spendable. However, I realized that we don't actually have any existing source for such a start height other than the wallet birthday, which would trigger a full rescan, and I didn't want to do that. I came to the same conclusion that you did about the actual effect of this bit, but left it in because I figured it was safe and that we might decide to improve the decision about whether or not we're able to rescan without dropping spendability for existing notes between now and the release.

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
```diff
@@ -0,0 +1,214 @@
...
+                // change.
+                transaction.execute(
+                    "INSERT INTO scan_queue (block_range_start, block_range_end, priority)
+                     VALUES (:block_range_start, :block_range_end, :priority)",
+                    named_params![
+                        ":block_range_start": u32::from(start),
+                        ":block_range_end": u32::from(orchard_init_height),
+                        ":priority": range_priority,
+                    ],
+                )?;
```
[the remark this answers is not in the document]
nuttycom: I think that for an empty range we can probably just skip this block?

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/fe0b6c7c109f6a784277cdb6527cba831058e886..ae9dd25525432421c5927b3fced8c73851117f61) to address review comments.

zcash_client_sqlite/src/wallet/init/migrations/orchard_shardtree.rs
str4d: For future-proofing, I encoded this as `priority < :historic` (as the priority codes are by design ordered).

nuttycom (approved):
re-utACK ae9dd25525432421c5927b3fced8c73851117f61

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -415,9 +445,21 @@ pub(crate) fn update_chain_tip<P: consensus::Parameters>(
     // `ScanRange` uses an exclusive upper bound.
     let chain_end = new_tip + 1;
 
-    // Read the maximum height from the shards table.
+    // Read the maximum height from each of the the shards tables. The minimum of the two
```
daira, line 448:
```suggestion
    // Read the maximum height from each of the shards tables. The minimum of the two
```
Non-blocking.

daira (commented):
I have reviewed the last force-push with @str4d.

zcash_client_sqlite/src/wallet/scanning.rs
str4d: Fixed in #1182.

Issue #1176, opened 2024-02-08
zcash_client_sqlite: Update scanning queue to support Orchard


nuttycom:

str4d:
Closed enough in #1255. There's more we could do to get fancy with the scan queues, but for Zashi 1.0 this is sufficient.

PR #1255, opened 2024-03-11
zcash_client_sqlite: Implement Orchard scanning

### LINKED ISSUE #1198: zcash_client_sqlite: implement `WalletRead::get_orchard_nullifiers`

str4d:
Part of #1176.
Closes #1198.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -550,6 +550,10 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                     for spend in tx.sapling_spends() {
                         wallet::sapling::mark_sapling_note_spent(wdb.conn.0, tx_row, spend.nf())?;
                     }
+                    #[cfg(feature = "orchard")]
+                    for spend in tx.orchard_spends() {
+                        wallet::orchard::mark_orchard_note_spent(wdb.conn.0, tx_row, spend.nf())?;
+                    }
 
                     for output in tx.sapling_outputs() {
                         // Check whether this note was spent in a later block range that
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/lib.rs
```diff
@@ -776,21 +869,31 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                #[cfg(feature = "orchard")]
+                let orchard_from_account = wdb.get_orchard_nullifiers(NullifierQuery::All)?.into_iter().find(
+                    |(_, nf)|
+                        d_tx.tx().orchard_bundle().iter().flat_map(|b| b.actions().iter())
+                        .any(|input| nf == input.nullifier())
+                ).map(|(account_id, _)| account_id);
+                #[cfg(not(feature = "orchard"))]
+                let orchard_from_account = None;
+
+                if let Some(account_id) = sapling_from_account.or(orchard_from_account) {
```
daira, line 890:
```suggestion
                if let Some(account_id) = orchard_from_account.or(sapling_from_account) {
```
Although we do not currently create transactions that shield from multiple accounts simultaneously in such a way that `sapling_from_account` would differ from `orchard_from_account`, this is possible and could be done by other wallets or a future version of a zcash_client_sqlite-based wallet. In that case, ZIP 316 would recommend to associate the transaction with the most preferred protocol [for the purpose of choosing an outgoing ovk](https://zips.z.cash/zip-0316#usage-of-outgoing-viewing-keys), when creating the transaction. So that is also the most consistent thing to do here.

daira (approved):
utACK (reviewed with @str4d).

daira (approved):
utACK 33be548f082ea9e8085b96ddb2a9733fc2e87f2b

zcash_client_sqlite/src/lib.rs
```diff
@@ -603,36 +665,70 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                            shardtree::LocatedTree::from_iter(
+                                start..end,
+                                ORCHARD_SHARD_HEIGHT.into(),
+                                chunk.iter_mut().map(|n| n.take().expect("always Some")),
+                            )
+                        })
+                        .map(|res| (res.subtree, res.checkpoints))
+                        .collect::<Vec<_>>();
+
+                    // Update the Sapling note commitment tree with all newly read note commitments
```
ebfull, line 722:
```suggestion
                    // Update the Orchard note commitment tree with all newly read note commitments
```

ebfull (approved):
utACK, though I didn't look very closely to some of the SQL.

zcash_client_sqlite/src/lib.rs
str4d: Fixed in #1182.

PR #1258, opened 2024-03-12
Return partial matches when using `WalletRead::get_account_for_ufvk`.

nuttycom:
Also, return the backend database's view of the matching UFVK.

PR #1259, opened 2024-03-12
Add `WalletRead::get_seed_account`

nuttycom:
Builds atop #1258

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -779,6 +779,46 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
     }
 }
 
+/// Returns the account id corresponding to a given [`HdSeedFingerprint`]
+/// and [`zip32::AccountId`], if any.
+pub(crate) fn get_seed_account<P: consensus::Parameters>(
+    conn: &rusqlite::Connection,
+    params: &P,
+    seed: &HdSeedFingerprint,
+    account_id: zip32::AccountId,
+) -> Result<Option<(AccountId, Option<UnifiedFullViewingKey>)>, SqliteClientError> {
+    let mut stmt = conn.prepare(
+        "SELECT id, ufvk
+        FROM accounts
+        WHERE hd_seed_fingerprint = :hd_seed_fingerprint
+          AND hd_account_index = :account_id",
+    )?;
+
+    let mut accounts = stmt.query_and_then::<_, SqliteClientError, _, _>(
+        named_params![
+            ":hd_seed_fingerprint": seed.as_bytes(),
+            ":hd_account_index": u32::from(account_id),
+        ],
+        |row| {
+            let account_id = row.get::<_, u32>(0).map(AccountId)?;
+            Ok((
+                account_id,
+                row.get::<_, Option<String>>(1)?
+                    .map(|ufvk_str| UnifiedFullViewingKey::decode(params, &ufvk_str))
+                    .transpose()
+                    .map_err(|e| {
+                        SqliteClientError::CorruptedData(format!(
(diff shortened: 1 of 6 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -778,6 +778,46 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
...
+    conn: &rusqlite::Connection,
+    params: &P,
+    seed: &HdSeedFingerprint,
+    account_id: zip32::AccountId,
+) -> Result<Option<(AccountId, Option<UnifiedFullViewingKey>)>, SqliteClientError> {
+    let mut stmt = conn.prepare(
+        "SELECT id, ufvk
+        FROM accounts
+        WHERE hd_seed_fingerprint = :hd_seed_fingerprint
+          AND hd_account_index = :account_id",
```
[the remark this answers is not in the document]
nuttycom: Fixed in https://github.com/zcash/librustzcash/pull/1259/commits/328e854202584be3055df4c9e1de915b48f4f1b5

zcash_client_sqlite/src/wallet/init/migrations/full_account_ids.rs
```diff
@@ -69,6 +69,7 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                     (account_type = {account_type_imported} AND hd_seed_fingerprint IS NULL AND hd_account_index IS NULL)
                 )
             );
+            CREATE UNIQUE INDEX hd_account ON accounts_new (hd_seed_fingerprint, hd_account_index);
```
str4d, line 72: I don't understand the full subtleties of `UNIQUE INDEX` vs `CONSTRAIN UNIQUE` and why you'd choose one over the other. AFAICT:
- They both produce an index.
- `UNIQUE INDEX` can be removed and added without altering the table (and I guess `CREATE UNIQUE INDEX` fails if the table contains non-unique data?)
- `UNIQUE INDEX` can include expressions
- `CONSTRAIN UNIQUE` has greater support in `ON CONFLICT` logic (whereas `UNIQUE INDEX` only supports `INSERT ... ON CONFLICT` I think).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -436,6 +436,54 @@ mod tests {
             expected_idx += 1;
         }
 
+        let expected_indices = vec![
+            r#"CREATE UNIQUE INDEX accounts_ufvk ON "accounts" (ufvk)"#,
+            r#"CREATE UNIQUE INDEX accounts_uivk ON "accounts" (uivk)"#,
+            r#"CREATE UNIQUE INDEX hd_account ON "accounts" (hd_seed_fingerprint, hd_account_index)"#,
+            r#"CREATE INDEX "addresses_accounts" ON "addresses" (
```
str4d, line 443: Thanks for adding this; I was unaware that uniqueness constraints were starting to be embedded in the indices (which prior to now I only thought were present for performance). I rely heavily on changes to `init.rs` to track changes to the database structure and determine what constraints are applied.

str4d (approved):
utACK 79f5bb411ccc1baa4b0424b20ee66453ecb31016

zcash_client_sqlite/src/wallet/init/migrations/full_account_ids.rs
nuttycom: Yeah, I was looking at this today. I think that the `ON CONFLICT` capabilities are actually the same, from what I was reading, at least for recent releases? Given that the separate indices can be dropped and added independently, I think I have a slight preference for them, but I don't think it makes much difference one way or the other.

PR #1261, opened 2024-03-12
update bip0039

zancas:
https://github.com/koushiro/rust-bips/blob/master/bip0039/CHANGELOG.md#version-0110-2022-11-22

https://github.com/koushiro/rust-bips/blob/master/bip0039/CHANGELOG.md#version-0120-2024-02-19

bip0039 has been updated twice since the version was set for librustzcash.

zingolib has dependencies that require the newer version of bip0039

### DIFF
--- a/zcash_primitives/Cargo.toml
+++ b/zcash_primitives/Cargo.toml
@@ -64,7 +64,7 @@ hdwallet = { workspace = true, optional = true }
 secp256k1 = { workspace = true, optional = true }
 
 # - ZIP 339
-bip0039 = { version = "0.10", features = ["std", "all-languages"] }
+bip0039 = { version = "0.12", features = ["std", "all-languages"] }
 
 # Dependencies used internally:
 # (Breaking upgrades to these are usually backwards-compatible, but check MSRVs.)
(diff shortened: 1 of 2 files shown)


zcash_primitives/Cargo.toml
```diff
@@ -64,7 +64,7 @@ hdwallet = { workspace = true, optional = true }
 secp256k1 = { workspace = true, optional = true }
 
 # - ZIP 339
-bip0039 = { version = "0.10", features = ["std", "all-languages"] }
+bip0039 = { version = "0.12", features = ["std", "all-languages"] }
```
str4d, line 67: We attempted to upgrade from `bip0039 0.10` in #808, but had to downgrade in #812 because the change in `bip0039 0.11` to use generics resulted in me spending several hours fighting and losing to the C++ template system in `zcashd`. While `zcashd` depends on `bip0039 0.10` via `zcash_primitives::zip339`, we will not be upgrading this dependency. (If someone else wants to fight the C++ template system to enable this, let me know!)

str4d (changes requested):
NACK, blocked on `zcashd`.

PR #1266, opened 2024-03-13
APIs for mobile SDKs

str4d:
This PR only adds one of the needed APIs for the mobile SDKs; due to also refactoring the traits, I decided to split the other changes (which are themselves non-trivial) to a follow-on PR.


nuttycom (approved):
utACK, essentially move-only.

PR #1267, opened 2024-03-13
Account APIs needed for the mobile SDKs

str4d:
This exposes the kind of account (derived or imported, from which the ZIP 32 account index can be obtained), and enables the `WalletRead::Account` type to be looked up via `WalletRead::AccountId` instead of solely via the derivation parameters or UFVK.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -576,7 +576,7 @@ pub trait WalletRead {
 
     /// Returns the account corresponding to a given [`HdSeedFingerprint`] and
     /// [`zip32::AccountId`], if any.
-    fn get_seed_account(
+    fn get_derived_account(
         &self,
         seed: &HdSeedFingerprint,
         account_id: zip32::AccountId,
@@ -1553,7 +1553,7 @@ pub mod testing {
             Ok(Vec::new())
         }
 
-        fn get_seed_account(
+        fn get_derived_account(
             &self,
             _seed: &HdSeedFingerprint,
             _account_id: zip32::AccountId,
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -574,6 +574,12 @@ pub trait WalletRead {
     /// Returns a vector with the IDs of all accounts known to this wallet.
     fn get_account_ids(&self) -> Result<Vec<Self::AccountId>, Self::Error>;
 
+    /// Returns the account corresponding to the given ID, if any.
+    fn get_account(
+        &self,
+        account_id: Self::AccountId,
+    ) -> Result<Option<Self::Account>, Self::Error>;
+
     /// Returns the account corresponding to a given [`HdSeedFingerprint`] and
     /// [`zip32::AccountId`], if any.
     fn get_derived_account(
@@ -1551,6 +1557,13 @@ pub mod testing {
 
         fn get_account_ids(&self) -> Result<Vec<Self::AccountId>, Self::Error> {
             Ok(Vec::new())
+        }
+
+        fn get_account(
+            &self,
+            _account_id: Self::AccountId,
+        ) -> Result<Option<Self::Account>, Self::Error> {
+            Ok(None)
         }
 
         fn get_derived_account(
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -315,18 +315,42 @@ impl AccountBalance {
     }
 }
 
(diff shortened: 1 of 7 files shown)


str4d:
The changes to `zcash_client_sqlite/src/wallet.rs` are best reviewed by-commit with whitespace hidden.

zcash_client_backend/src/data_api.rs
```diff
@@ -315,12 +315,32 @@ impl AccountBalance {
     }
 }
 
+/// The kinds of accounts supported by `zcash_client_backend`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum AccountKind {
```
str4d, line 320: I named this `AccountKind` instead of `AccountType` because `type` is a Rust keyword (so precludes nice method names like `account.type()`), and there's precedent here with things like `std::io::ErrorKind`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -140,82 +140,71 @@ pub(crate) const BLOCK_SAPLING_FRONTIER_ABSENT: &[u8] = &[0x0];
...
+    /// This is available to derived accounts, as well as accounts directly imported as
+    /// full viewing keys.
     Full(Box<UnifiedFullViewingKey>),
-    /// An account that was imported via its incoming viewing key.
+
+    /// An incoming viewing key.
+    ///
+    /// Accounts that have this kind of viewing key cannot be used in wallet contexts,
+    /// because they are unable to maintain an accurate balance.
     Incoming(Uivk),
```
nuttycom, line 185: Note that this will change in https://github.com/zcash/librustzcash/pull/1245

zcash_client_sqlite/src/wallet.rs
```diff
@@ -357,20 +333,27 @@ pub(crate) fn ufvk_to_uivk<P: consensus::Parameters>(
 pub(crate) fn add_account<P: consensus::Parameters>(
     conn: &rusqlite::Transaction,
     params: &P,
-    account: Account,
+    kind: AccountType,
+    viewing_key: ViewingKey,
```
nuttycom: Hm, doing it this way means that the `(Derived, IVK)` combination is supported, which doesn't make much sense.

nuttycom (dismissed):
utACK https://github.com/zcash/librustzcash/pull/1245 modulo Clippy complaints. Doing the column rename in this PR would also be good.

zcash_client_sqlite/src/wallet.rs
str4d: Yep that's fine; this is still a private enum.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1516,51 +1506,36 @@ pub(crate) fn get_account<C: Borrow<rusqlite::Connection>, P: Parameters>(
     let row = result.next()?;
     match row {
         Some(row) => {
-            let account_type: AccountType =
-                row.get::<_, u32>("account_type")?.try_into().map_err(|_| {
-                    SqliteClientError::CorruptedData("Unrecognized account_type".to_string())
-                })?;
+            let kind = parse_account_kind(
+                row.get("account_type")?,
```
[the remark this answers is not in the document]
str4d: That was introduced in #1235 and I don't quite get how it is implemented (or when support was added). It might be nice to do a pass over the crate to use that instead of column indexing, if it is not inefficient, but for now I've stuck with whatever style the local code used in each place.

zcash_client_sqlite/src/wallet.rs
str4d: It's not an _invalid_ combination: one could import a UIVK for which you know its derivation pathway, in order to make future identification of the corresponding UFVK and USK easier.

nuttycom: Yeah, I decided that that would be a reasonable use case, hence "this is fine."

str4d: What this boils down to is that the thing that makes an account "usable as a wallet" is that it has a UFVK (and can track balance), not whether we know the derivation path; as long as the caller can provide the corresponding USK when needed, that's all that matters.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/1e1378596acb31b13d7a0de1e46b26d88a4d2f33..b161472cc0a551d7dfacf33517d4c5d7a0a6521b) to address review comments.

nuttycom (approved):
utACK https://github.com/zcash/librustzcash/pull/1245

PR #1270, opened 2024-03-14
Disable the default features of the `tonic` and `tonic-build` crates to allow downstream consumers to build in Wasm

### LINKED ISSUE #1269: zcash_client_backend: unused `tonic` crate features prevent Wasm builds with feature `lightwalletd-tonic`
As a consumer of this crate we want to use the generated gRPC client within a Wasm environment. Currently this is not possible due to the `transport` feature of tonic being enabled by default and being incompatible with Wasm.

Currently there is no reason for having this feature enabled for the tonic and tonic-build dependencies for `zcash_client_backend` and it could be safely removed.

Removing it does change the public API of the crate though as it alters the code produced by the build script.

willemolding:
Closes #1269 

Note this regenerates the code-generated `proto/service.rs` file and the newer version is included in this PR. This is a minor change to the `zcash_client_backend` public API as these are exposed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
         run: cargo add --no-default-features --path ../crates/zcash_proofs
       - name: Add zcash_client_backend as a dependency of the synthetic crate
         working-directory: ./ci-build
-        run: cargo add --no-default-features --path ../crates/zcash_client_backend
+        run: cargo add --features lightwalletd-tonic --path ../crates/zcash_client_backend
       - name: Copy pinned dependencies into synthetic crate
         run: cp crates/Cargo.lock ci-build/
       - name: Add target
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
         run: cargo add --no-default-features --path ../crates/zcash_proofs
       - name: Add zcash_client_backend as a dependency of the synthetic crate
         working-directory: ./ci-build
-        run: cargo add --features lightwalletd-tonic-transport --path ../crates/zcash_client_backend
+        run: cargo add --features lightwalletd-tonic --path ../crates/zcash_client_backend
       - name: Copy pinned dependencies into synthetic crate
         run: cp crates/Cargo.lock ci-build/
       - name: Add target
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
         run: cargo add --no-default-features --path ../crates/zcash_proofs
       - name: Add zcash_client_backend as a dependency of the synthetic crate
         working-directory: ./ci-build
-        run: cargo add --features lightwalletd-tonic --path ../crates/zcash_client_backend
+        run: cargo add --features lightwalletd-tonic-transport --path ../crates/zcash_client_backend
       - name: Copy pinned dependencies into synthetic crate
         run: cp crates/Cargo.lock ci-build/
       - name: Add target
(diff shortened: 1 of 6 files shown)


str4d (changes requested):
Thanks for the PR!

str4d:
Also, in addition to making the above change, add the following change to `.github/workflows/ci.yml` so we are testing this change:
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 6263019a6..ebf418cfe 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
         run: cargo add --no-default-features --path ../crates/zcash_proofs
       - name: Add zcash_client_backend as a dependency of the synthetic crate
         working-directory: ./ci-build
-        run: cargo add --no-default-features --path ../crates/zcash_client_backend
+        run: cargo add --features lightwalletd-tonic --path ../crates/zcash_client_backend
       - name: Copy pinned dependencies into synthetic crate
         run: cp crates/Cargo.lock ci-build/
       - name: Add target
```

zcash_client_backend/src/proto/service.rs
```diff
@@ -284,17 +284,6 @@ pub mod compact_tx_streamer_client {
...
-        /// Attempt to create a new client by connecting to a given endpoint.
-        pub async fn connect<D>(dst: D) -> Result<Self, tonic::transport::Error>
-        where
-            D: TryInto<tonic::transport::Endpoint>,
-            D::Error: Into<StdError>,
-        {
-            let conn = tonic::transport::Endpoint::new(dst)?.connect().await?;
-            Ok(Self::new(conn))
-        }
-    }
```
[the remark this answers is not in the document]
zancas: I can confirm that we're depending on this`impl`.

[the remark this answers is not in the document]
ec2: > What we ideally want is for this impl to not be present when the target is WASM, but available otherwise. 

Curious to understand why you think that. Seems like pulling in the provided this should be irrespective of what `target_arch` you're in. Maybe the `transport` feature will be supported by `wasm32` at some point (obviously not), but still. I'm of the opinion the less stuff you need to pull into `zcash_client_backend/src/proto.rs`, the better.

willemolding:
Thanks for the review @str4d !

I moved the impl of `CompactTxStreamerClient<tonic::transport::Channel>` to `proto.rs` as suggested but put it behind just a feature flag without a target_arch check.

I think this makes more sense as it it feels a little weird to have the API change depending on the build target. Instead builds will fail if the feature is enabled and then Wasm builds can disable it to have it work. I think more useful information is communicated this way.

.github/workflows/ci.yml
```diff
@@ -195,7 +195,7 @@ jobs:
         run: cargo add --no-default-features --path ../crates/zcash_proofs
       - name: Add zcash_client_backend as a dependency of the synthetic crate
         working-directory: ./ci-build
-        run: cargo add --no-default-features --path ../crates/zcash_client_backend
+        run: cargo add --features lightwalletd-tonic-transport --path ../crates/zcash_client_backend
```
str4d, line 198: I suspect this actually needs to be what I originally said, otherwise this will just always fail, instead of checking that the old flag now passes.
```suggestion
        run: cargo add --features lightwalletd-tonic --path ../crates/zcash_client_backend
```
I've fired off a CI run to confirm this.

Cargo.lock
str4d, line 1: Please revert the changes to this file (now that the dependencies in the lockfile should be unaltered by the changes in this PR).

str4d (changes requested):
Reviewed 61447d94381f61c50da5fd9278d70422fb190275

.github/workflows/ci.yml
str4d: Confirmed, this fails CI as-is: https://github.com/zcash/librustzcash/actions/runs/8277319191/job/22684643018?pr=1270

willemolding: Ah yep I misunderstood what that CI task was doing. Should have wasm builds in CI now

str4d (approved):

PR #1271, opened 2024-03-14
zcash_client_backend: Pure refactoring for note selection ergonomics.

nuttycom:
This is a pure refactoring that should result in no functional changes.


daira (approved):
utACK

PR #1275, opened 2024-03-14
zcash_client_backend: Make `WalletRead::get_transaction` return `Result<Option<Transaction>, _>`

nuttycom:
This should never have had the behavior of returning an error on a missing txid in the first place; doing so conflates database corruption or connectivity errors with the ordinary case where data may not be available.

Best reviewed hiding whitespace changes.


str4d (approved):
utACK 46fd6ab0fe5eab3b440f9d6bba75c4d8671c9b0a

daira (commented):
Post-hoc ACK

PR #1276, opened 2024-03-15
Add local network to network enum

Oscar-Pepper:
Adds LocalNetwork variant to Network enum and adds methods for LocalNetwork for defining common network upgrade activation height patterns.

This allows production code of consumers to be tested in a regtest environment by easily switching out network parameters without needing to create their own version of the Network enum with a regtest variant.

### DIFF
--- a/components/zcash_protocol/CHANGELOG.md
+++ b/components/zcash_protocol/CHANGELOG.md
@@ -10,6 +10,12 @@ and this library adheres to Rust's notion of
 ### Added
 - `zcash_protocol::memo`:
   - `impl TryFrom<&MemoBytes> for Memo`
+- `zcash_protocol::local_consensus`:
+  - `new`, `all_upgrades_active` and `canopy_active` methods to `LocalNetwork`
+
+### Changed
+- `zcash_protocol::consensus`:
+  - Added `LocalNetwork` variant to `Network` enum behind "local-consensus" feature
 
 ### Removed
 - `unstable-nu6` and `zfuture` feature flags (use `--cfg zcash_unstable=\"nu6\"`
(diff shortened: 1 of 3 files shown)


nuttycom (changes requested):
NACK. This is not how the `Network` type is intended to be used; in circumstances where a local network is necessary, you should use a bespoke type that implements the `Parameters` trait. See how `LocalNetwork` is used in https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/testing.rs for an example.

If there are places where we are depending upon the `Network` type that we shouldn't be, I'm happy to change those. There may be places where instead we should be taking a `NetworkType` parameter, rather than a `Network` (where we only need the network constants), otherwise everything should be a polymorphic type having the `consensus::Parameters` bound.

Oscar-Pepper:
> NACK. This is not how the `Network` type is intended to be used; in circumstances where a local network is necessary, you should use a bespoke type that implements the `Parameters` trait. See how `LocalNetwork` is used in https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/testing.rs for an example.
> 
> If there are places where we are depending upon the `Network` type that we shouldn't be, I'm happy to change those. There may be places where instead we should be taking a `NetworkType` parameter, rather than a `Network` (where we only need the network constants), otherwise everything should be a polymorphic type having the `consensus::Parameters` bound.

Ok thanks for the info, our wallet struct contains an enum similar to NetworkType instead of a generic like WalletDb in `testing.rs`. It makes sense for us to eventually restructure this so we can be more aligned with librustzcash but for now I will just send the additional methods to LocalNetwork so we can use it in place of our own LocalNetwork struct which is essentially the same.

Oscar-Pepper:
closed for #1281

PR #1277, opened 2024-03-15
rename the build.rs such that it's not run by default

zancas:
This script is solely for the purpose of producing Rust source code from protobuf spec files, to support lightwalletd gRPC services.

It should only be run when a developer explicitly intends to *change* [add/remove/modify] those bindings.

It's unlikely this will be the case for the majority of builds, and probably shouldn't be done implicitly.


zcash_client_backend/build_grpc_rust_bindings.rs
str4d, line 1: - This rename silently breaks CI: the protobuf consistency check will now pass even if the `*.proto` files and generated files are out-of-sync.
- This rename silently breaks the workflows of developers maintaining the protobuf files, which combined with the CI break could have led to `*.proto` changes not correctly being reflected in the generated files.
- There is now no way to trigger the generator without renaming the build script back to `build.rs`. The rename will inevitably get accidentally checked in along with a future protobuf update, undoing the change.

str4d (changes requested):
I'm not currently convinced that this change has better trade-offs overall than the existing approach. The outer build script might get run in all builds, but it does nothing unless:
- the protobuf files are present (which they are not in published crate versions).
- `protoc` is available in `PATH`.

We could perhaps exclude the build script from published crate versions in addition to excluding the `*.proto` files; I don't recall if I tried that when I switched from `protobuf` to `prost` in #691.

For historical context, we previously _always_ generated code from `*.proto` files on-demand using `protobuf-codegen-pure`, so we didn't need to deal with keeping the `*.proto` files and generated code in sync. `prost` codegen isn't implemented in pure Rust and instead calls out to the `protoc` binary, so we compromised by checking in the generated files and adding machinery to always keep things in-sync.

daira:
I just don't think we should make this change. The `build.rs` is doing precisely what a `build.rs` should do: the source is the `.proto` files, not the generated `.rs` files. It doesn't cause any problem if you don't have `protoc` on the path. It's common to take this approach in build systems that rely on tools that might not be installed.

I think we should close this PR.

PR #1279, opened 2024-03-15
zcash_client_sqlite: Fix ambiguities in transaction views

nuttycom:


zcash_client_sqlite/src/testing.rs
```diff
@@ -958,6 +959,105 @@ impl<Cache> TestState<Cache> {
...
     }
+
+    /// Returns a vector of transaction summaries
+    pub(crate) fn get_tx_history(
+        &self,
+    ) -> Result<Vec<TransactionSummary<AccountId>>, SqliteClientError> {
+        let mut stmt = self.wallet().conn.prepare_cached(
+            "SELECT * 
+             FROM v_transactions 
+             ORDER BY mined_height, tx_index DESC",
```
ebfull, line 970: Sorry for the dumb question; haven't written much SQL since I was a teenager. You don't have to do `ORDER BY mined_height DESC, tx_index DESC` here?

ebfull (approved):
utACK

zcash_client_sqlite/src/testing.rs
nuttycom: Good catch!

ebfull: Oh, that's actually the intentional ordering here to ascend on mined height and descend on tx_index.

ebfull (approved):
re-ACK

PR #1281, opened 2024-03-18
zcash_protocol: Add constructors to `LocalNetwork`

Oscar-Pepper:
Adds commonly used constructors to `LocalNetwork` struct.

### DIFF
--- a/components/zcash_protocol/CHANGELOG.md
+++ b/components/zcash_protocol/CHANGELOG.md
@@ -10,6 +10,8 @@ and this library adheres to Rust's notion of
 ### Added
 - `zcash_protocol::memo`:
   - `impl TryFrom<&MemoBytes> for Memo`
+- `zcash_protocol::local_consensus`:
+  - `new`, `all_upgrades_active` and `canopy_active` constructors to `LocalNetwork`
 
 ### Removed
 - `unstable-nu6` and `zfuture` feature flags (use `--cfg zcash_unstable=\"nu6\"`
--- a/components/zcash_protocol/src/local_consensus.rs
+++ b/components/zcash_protocol/src/local_consensus.rs
@@ -43,6 +43,37 @@ pub struct LocalNetwork {
     pub z_future: Option<BlockHeight>,
 }
 
+impl LocalNetwork {
+    pub fn new(
+        overwinter: u64,
+        sapling: u64,
+        blossom: u64,
+        heartwood: u64,
+        canopy: u64,
+        nu5: u64,
+    ) -> Self {
+        LocalNetwork {
+            overwinter: Some(BlockHeight::from_u32(overwinter as u32)),
+            sapling: Some(BlockHeight::from_u32(sapling as u32)),
+            blossom: Some(BlockHeight::from_u32(blossom as u32)),
+            heartwood: Some(BlockHeight::from_u32(heartwood as u32)),
+            canopy: Some(BlockHeight::from_u32(canopy as u32)),
+            nu5: Some(BlockHeight::from_u32(nu5 as u32)),
+        }
+    }
+
+    /// Creates a `LocalNetwork` with all network upgrades initially active.
+    pub fn all_upgrades_active() -> Self {
+        Self::new(1, 1, 1, 1, 1, 1)
+    }
+
+    /// Creates a `LocalNetwork` with all network upgrades up to and including canopy
+    /// initally active.
+    pub fn canopy_active(nu5_activation_height: u64) -> Self {
+        Self::new(1, 1, 1, 1, 1, nu5_activation_height)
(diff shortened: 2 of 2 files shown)


nuttycom (approved):
utACK

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -43,6 +43,37 @@ pub struct LocalNetwork {
     pub z_future: Option<BlockHeight>,
 }
 
+impl LocalNetwork {
+    pub fn new(
```
str4d, line 47: - Document the constructor.
- The arguments should be checked for ordering. It is fine to activate two sequential network upgrades at the same height, but it is not okay to change the order in which NUs activate. Either make the constructor family, or just document that it panics (which is probably fine for something only used for development).

str4d (changes requested):

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -43,6 +43,37 @@ pub struct LocalNetwork {
...
+        canopy: u64,
+        nu5: u64,
+    ) -> Self {
+        LocalNetwork {
+            overwinter: Some(BlockHeight::from_u32(overwinter as u32)),
+            sapling: Some(BlockHeight::from_u32(sapling as u32)),
+            blossom: Some(BlockHeight::from_u32(blossom as u32)),
+            heartwood: Some(BlockHeight::from_u32(heartwood as u32)),
+            canopy: Some(BlockHeight::from_u32(canopy as u32)),
+            nu5: Some(BlockHeight::from_u32(nu5 as u32)),
```
str4d, line 61: Constructor is currently broken because it needs to include `nu6` and `z_future` default values behind their respective cfg flags. (Do not expose these in the constructor arguments; they are not necessary for local testing against existing consensus rules, and should be handled separately.)
```suggestion
            nu5: Some(BlockHeight::from_u32(nu5 as u32)),
            #[cfg(zcash_unstable = "nu6")]
            nu6: None,
            #[cfg(zcash_unstable = "zfuture")]
            z_future: None,
```

str4d (changes requested):

components/zcash_protocol/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this library adheres to Rust's notion of
 ### Added
 - `zcash_protocol::memo`:
   - `impl TryFrom<&MemoBytes> for Memo`
+- `zcash_protocol::local_consensus`:
+  - `new`, `all_upgrades_active` and `canopy_active` constructors to `LocalNetwork`
```
str4d, line 14: PR needs a rebase because this now merge-conflicts with the 0.1.1 release (in that it will end up editing the 0.1.1 changelog).

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -43,6 +43,54 @@ pub struct LocalNetwork {
...
+    ///
+    /// Sequential network upgrades may share the same activation height but this constructor will
+    /// panic if the order in which network upgrades activate is changed.
+    pub fn new(
+        overwinter: u64,
+        sapling: u64,
+        blossom: u64,
+        heartwood: u64,
+        canopy: u64,
+        nu5: u64,
```
str4d, line 57: I just realised that all of these arguments are `u64`. Two questions:
- Why not `BlockHeight`?
- If not `BlockHeight`, why `u64`? The internal casts to `u32` will silently truncate the heights, which is undocumented and confusing. Additionally, these truncations occur after the assertion, which means that a caller can bypass the ordering requirement by setting the high bits of the heights.

The latter question is blocking; these should be `u32` at a minimum, and I don't see a good reason for these to not be `BlockHeight`.

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -43,6 +43,54 @@ pub struct LocalNetwork {
...
+    }
+
+    /// Construct a new `LocalNetwork` with all network upgrades initially active.
+    pub fn all_upgrades_active() -> Self {
+        Self::new(1, 1, 1, 1, 1, 1)
+    }
+
+    /// Construct a new `LocalNetwork` with all network upgrades up to and including canopy
+    /// initally active.
+    pub fn canopy_active(nu5_activation_height: u64) -> Self {
```
str4d, line 89: Ditto: this should at least be `u32`, and probably a `BlockHeight`.

str4d (changes requested):
Reviewed 68d2add2a62c879465adc6658e1b672ffec64582.

str4d:
Closing due to no response. If you still need these APIs, please reopen and address the review comments.

PR #1287, opened 2024-03-19
Distinguish seed relevance when no derived accounts are present

str4d:
During wallet migration in particular, the absence of _any_ accounts is expected, and all seeds should be treated as relevant (because accounts cannot be added before a wallet is initialized).

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -64,6 +64,7 @@ use std::{
 };
 
 use incrementalmerkletree::{frontier::Frontier, Retention};
+use nonempty::NonEmpty;
 use secrecy::SecretVec;
 use shardtree::{error::ShardTreeError, store::ShardStore, ShardTree};
 use zip32::fingerprint::SeedFingerprint;
@@ -742,10 +743,13 @@ pub trait WalletRead {
 
     /// Checks whether the given seed is relevant to any of the derived accounts (where
     /// [`Account::source`] is [`AccountSource::Derived`]) in the wallet.
-    fn is_seed_relevant_to_any_derived_accounts(
+    ///
+    /// This API does not check whether the seed is relevant to any imported account,
+    /// because that would require brute-forcing the ZIP 32 account index space.
+    fn seed_relevance_to_derived_accounts(
         &self,
         seed: &SecretVec<u8>,
-    ) -> Result<bool, Self::Error>;
+    ) -> Result<SeedRelevance<Self::AccountId>, Self::Error>;
 
     /// Returns the account corresponding to a given [`UnifiedFullViewingKey`], if any.
     fn get_account_for_ufvk(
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -289,11 +292,26 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
...
+        // We ignore `SeedRelevance::NoDerivedAccounts`;
+        match wdb
             .is_seed_relevant_to_any_derived_accounts(&seed)
             .map_err(sqlite_client_error_to_wallet_migration_error)?
         {
-            return Err(WalletMigrationError::SeedNotRelevant.into());
+            SeedRelevance::Relevant => (),
+            // Every seed is relevant to a wallet with no accounts; this is most likely a
+            // new wallet database being initialized for the first time.
+            SeedRelevance::NoDerivedAccounts
```
str4d, line 303: I'm considering adding a `NoAccounts` variant to simplify this case, as I think the mobile SDKs will just want to do the exact same logic as here.

nuttycom (dismissed):
utACK

zcash_client_backend/src/data_api.rs
```diff
@@ -907,6 +907,16 @@ pub trait WalletRead {
     }
 }
 
+/// The relevance of a seed to a given wallet.
+pub enum SeedRelevance {
```
nuttycom, line 911: This will need documentation in the CHANGELOG

zcash_client_sqlite/src/wallet/init.rs
nuttycom: I think having the enum is good; usually if you have the option to use an enum instead of a boolean it's a good idea to do so, just for better type safety.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/adaa2b564cbb4a1b8e123300ecd8fbb688598309..cbfc3b8087b82829165487fb8a10944dd96ae4f3) to address comments from review and a subsequent pairing.

zcash_client_sqlite/src/lib.rs
```diff
@@ -337,11 +338,16 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
 
-    fn is_seed_relevant_to_any_derived_accounts(
+    fn seed_relevance_to_derived_accounts(
         &self,
         seed: &SecretVec<u8>,
-    ) -> Result<bool, Self::Error> {
+    ) -> Result<SeedRelevance<Self::AccountId>, Self::Error> {
+        let mut has_accounts = false;
+        let mut has_derived = false;
+        let mut relevant_account_ids: Option<NonEmpty<Self::AccountId>> = None;
```
nuttycom, line 347: nonblocking: Instead of this, just make it a `Vec`, and then at the end use `NonEmpty::from_vec`.

nuttycom, line 347:
```suggestion
        let mut relevant_account_ids = vec![];
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -353,21 +359,34 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
                     &account.uivk(),
                 )? {
-                    // The seed is relevant to this account. No need to check any others.
-                    return Ok(true);
+                    // The seed is relevant to this account.
+                    if let Some(account_ids) = &mut relevant_account_ids {
+                        account_ids.push(account_id);
+                    } else {
+                        relevant_account_ids = Some(NonEmpty::new(account_id));
+                    }
```
nuttycom, line 376:
```suggestion
                    relevant_account_ids.push(account_id);
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -353,21 +359,34 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
                 }
             }
         }
 
-        // The seed was not relevant to any of the accounts in the wallet.
-        Ok(false)
+        Ok(if let Some(account_ids) = relevant_account_ids {
```
nuttycom, line 381:
```suggestion
        Ok(if let Some(account_ids) = NonEmpty::from_vec(relevant_account_ids) {
```

nuttycom (dismissed):
utACK with nonblocking suggestions.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/cbfc3b8087b82829165487fb8a10944dd96ae4f3..3c1e82a0c87ba4e6b9b69014d3fc218881ff2fe3) to address second review comments.

nuttycom (approved):
re-utACK

PR #1288, opened 2024-03-19
Remove `orchard` feature flag from behind `zcash_unstable` cfg flag

str4d:


nuttycom:
Will CI still do the right thing here before we fix the tests?

str4d:
> Will CI still do the right thing here before we fix the tests?

The "continue on error" behaviour is currently configured for every non-default feature configuration. So yes, the Orchard tests failing won't cause the regular tests to not be run.

What we should eventually do is alter CI to be more specific about only having "continue on error" behaviour for test states that require `RUSTFLAGS`, but that's not what is currently configured.

nuttycom (approved):
utACK

PR #1289, opened 2024-03-19
zcash_client_sqlite: Fix `scan_complete` tests.

nuttycom:


zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -79,6 +79,7 @@ pub(crate) type OutputRecoveryError = Error<
 pub(crate) trait ShieldedPoolTester {
     const SHIELDED_PROTOCOL: ShieldedProtocol;
     const TABLES_PREFIX: &'static str;
+    // const MERKLE_TREE_DEPTH: u8;
```
str4d, line 82: Is this coming back in a future PR?

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -172,6 +175,7 @@ use super::WalletRead;
 ///
 /// This stores the block height at which the leaf that completed the subtree was
 /// added, and the root hash of the complete subtree.
+#[derive(Debug)]
```
str4d, line 178: Document addition in changelog.

zcash_client_backend/src/proto.rs
```diff
@@ -299,6 +299,15 @@ impl service::TreeState {
             self.height
                 .try_into()
                 .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "Invalid block height"))?,
+            BlockHash::try_from_slice(&hex::decode(&self.hash).map_err(|e| {
```
str4d, line 302: Because `self.hash` is a protobuf `string`, not a `bytes`, it means it is directly conveying the hex encoding from `zcashd`'s JSON-RPC. Hence it needs to be byte-reversed as part of being parsed (unlike `CompactBlock` where I handled this for the caller).

zcash_client_sqlite/src/testing.rs
```diff
@@ -23,6 +25,8 @@ use sapling::{
     zip32::DiversifiableFullViewingKey,
     Note, Nullifier,
 };
+use zcash_client_backend::data_api::chain::CommitmentTreeRoot;
+use zcash_client_backend::data_api::WalletCommitmentTrees;
```
str4d, line 29: Merge both of these into the `zcash_client_backend` import block below.

zcash_client_sqlite/src/testing.rs
```diff
@@ -323,6 +319,48 @@ where
...
+        &mut self,
+        state: ChainState,
+        prior_sapling_roots: &[CommitmentTreeRoot<sapling::Node>],
+        #[cfg(feature = "orchard")] prior_orchard_roots: &[CommitmentTreeRoot<MerkleHashOrchard>],
+    ) -> Result<(), ShardTreeError<commitment_tree::Error>> {
+        self.wallet_mut()
+            .put_sapling_subtree_roots(0, prior_sapling_roots)?;
+        #[cfg(feature = "orchard")]
+        self.wallet_mut()
+            .put_orchard_subtree_roots(0, prior_orchard_roots)?;
```
str4d, line 332: Given that this method is named `initialize_from_chain_state`, it should either assert that the trees are empty, or clear them before adding anything. If neither of those should be done, then rename the method.

zcash_client_sqlite/src/testing.rs
```diff
@@ -348,6 +386,22 @@ where
         (height, res, nf)
     }
 
+    pub(crate) fn insert_cached_block(
```
str4d, line 389: Should this be `pub(crate)`?

str4d (changes requested):
Reviewed 0d4a3009945bb4bd66ace3de03b46a90236139a2.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/0d4a3009945bb4bd66ace3de03b46a90236139a2..4f7c5bd722f6592a2fcd111adaf443bbdd10cd08) to address review comments.

str4d (approved):
utACK 4f7c5bd722f6592a2fcd111adaf443bbdd10cd08

PR #1290, opened 2024-03-19
zcash_client_backend: Treat protobuf default as the empty `CommitmentTree`.

### LINKED ISSUE #1280: Add account fails with orchard feature enabled for birthday heights prior to orchard activation
The "failed to fill whole buffer" error is returned from `AccountBirthday::from_treestate` when the `TreeState` is for a block height that has an `orchard_tree` string. This is easy to reproduce when using a pre-orchard activation height, such as 1101500, which is the one I was using.

nuttycom:
Fixes #1280


str4d (dismissed):
utACK 7da3a43e4a3f56be649a3399bbf8f005d83741de

str4d (approved):
Re-utACK 5f1d75937b84cdc51ad81d7eb2a9b0123c57af3e

PR #1291, opened 2024-03-20
zcash_client_sqlite: Include Orchard receiver in default UA for new accounts

str4d:


str4d:
I noticed this while adding Orchard to the Swift SDK. We might need to consider whether we need to add a migration to upgrade previously-generated addresses to include Orchard receivers, or else the default UA post-upgrade probably won't include Orchard support.

nuttycom (approved):
utACK

PR #1293, opened 2024-03-20
Feature: Add an in-memory wallet backend.

nuttycom:


oxarbitrage:
I think it can be a good to create a list of the minimal requirements we consider needed to merge this PR while leave the rest of the functions to make individual pull requests against the main repo ?

str4d:
Per the meeting today, this PR needs to be rebased on `main` and squashed down. Then once we're happy with it as a framework, we can merge it, and then implement its internals by migrating tests over (#1415) after the test framework has been migrated {#1412).

PR #1295, opened 2024-03-20
Fix the memory wallet build by hiding orchard fields in `new()` behind a flag

oxarbitrage:
Needed for building the branch locally.


nuttycom (approved):
utACK

PR #1296, opened 2024-03-20
remove doc hidden from some keys

AloeareV:


zcash_keys/src/keys.rs
```diff
@@ -624,7 +622,6 @@ pub struct UnifiedFullViewingKey {
     unknown: Vec<(u32, Vec<u8>)>,
 }
 
-#[doc(hidden)]
```
AArnott, line 627: This should have been done as part of #1245 but it was overlooked.

AArnott (commented):
I'm fully in favor of this. I know USK is not a spec'd format. And for that matter, IMO it's a terrible structure to reverse engineer. But it is nonetheless necessary to be able to cross the interop layer, and pretending that it isn't needed is silly.

nuttycom (approved):
utACK

PR #1217, opened 2024-03-04
1169 sync engine with block cache trait

Oscar-Pepper:
- Builds on #1184, implementing an async wallet synchronization function
- Replaces `FsBlockDb` with `BlockCache` for generic implementation of block caches, solving cyclic dependencies.
- Implements `BlockCache` for `FsBlockDb` and `TestBlockCache` (formerly `BlockCache`)

PR #1298, opened 2024-03-21
Basic `put_block` function for the memory wallet

oxarbitrage:
I want to create the most basic version of a `put_blocks` function for the memory wallet, one where we assume that the blocks will be in order. 

This PR is for @nuttycom to take a look, it has some comments and questions in the code and also review comments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,8 +52,7 @@ bitvec = "1"
 blake2s_simd = "1"
 bls12_381 = "0.8"
 jubjub = "0.10"
-#sapling = { package = "sapling-crypto", version = "0.1.2" }
-sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto", rev = "54fc7d3d9c8eeb8e6ef6f64619deed43b0b681a6" }
+sapling = { package = "sapling-crypto", version = "0.1.3" }
 
 # - Orchard
 nonempty = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,8 @@ bitvec = "1"
 blake2s_simd = "1"
 bls12_381 = "0.8"
 jubjub = "0.10"
-sapling = { package = "sapling-crypto", version = "0.1.2" }
+#sapling = { package = "sapling-crypto", version = "0.1.2" }
+sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto", rev = "54fc7d3d9c8eeb8e6ef6f64619deed43b0b681a6" }
 
 # - Orchard
 nonempty = "0.7"
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -992,7 +992,7 @@ pub struct ScannedBundles<NoteCommitment, NF> {
 }
 
 impl<NoteCommitment, NF> ScannedBundles<NoteCommitment, NF> {
-    pub fn new(
+    pub(crate) fn new(
         final_tree_size: u32,
         commitments: Vec<(NoteCommitment, Retention<BlockHeight>)>,
         nullifier_map: Vec<(TxId, u16, Vec<NF>)>,
@@ -1053,7 +1053,7 @@ pub struct ScannedBlock<A> {
 
 impl<A> ScannedBlock<A> {
     /// Constructs a new `ScannedBlock`
-    pub fn from_parts(
+    pub(crate) fn from_parts(
         block_height: BlockHeight,
         block_hash: BlockHash,
         block_time: u32,
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -992,7 +992,7 @@ pub struct ScannedBundles<NoteCommitment, NF> {
 }
 
 impl<NoteCommitment, NF> ScannedBundles<NoteCommitment, NF> {
-    pub(crate) fn new(
+    pub fn new(
         final_tree_size: u32,
         commitments: Vec<(NoteCommitment, Retention<BlockHeight>)>,
         nullifier_map: Vec<(TxId, u16, Vec<NF>)>,
@@ -1053,7 +1053,7 @@ pub struct ScannedBlock<A> {
 
 impl<A> ScannedBlock<A> {
     /// Constructs a new `ScannedBlock`
-    pub(crate) fn from_parts(
+    pub fn from_parts(
         block_height: BlockHeight,
         block_hash: BlockHash,
         block_time: u32,
--- a/zcash_client_backend/src/data_api/mem_wallet.rs
+++ b/zcash_client_backend/src/data_api/mem_wallet.rs
@@ -1,5 +1,6 @@
 #![allow(unused)]
 use incrementalmerkletree::{Address, Retention};
+use sapling::NullifierDerivingKey;
 use secrecy::{ExposeSecret, SecretVec};
 use shardtree::{error::ShardTreeError, store::memory::MemoryShardStore, ShardTree};
 use std::{
@@ -31,7 +32,7 @@ use crate::{
 };
 
 use super::{
-    chain::CommitmentTreeRoot, scanning::ScanRange, AccountBirthday, BlockMetadata,
+    chain::CommitmentTreeRoot, scanning::ScanRange, Account, AccountBirthday, BlockMetadata,
     DecryptedTransaction, NullifierQuery, ScannedBlock, SentTransaction, WalletCommitmentTrees,
     WalletRead, WalletSummary, WalletWrite, SAPLING_SHARD_HEIGHT,
 };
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -52,7 +52,8 @@ bitvec = "1"
 blake2s_simd = "1"
 bls12_381 = "0.8"
 jubjub = "0.10"
-sapling = { package = "sapling-crypto", version = "0.1.2" }
+#sapling = { package = "sapling-crypto", version = "0.1.2" }
+sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto", rev = "54fc7d3d9c8eeb8e6ef6f64619deed43b0b681a6" }
```
oxarbitrage, line 56: Changed to the sapling version with the derives we need until a release is done there.

zcash_client_backend/src/data_api.rs
```diff
@@ -992,7 +992,7 @@ pub struct ScannedBundles<NoteCommitment, NF> {
 }
 
 impl<NoteCommitment, NF> ScannedBundles<NoteCommitment, NF> {
-    pub(crate) fn new(
```
oxarbitrage, line 995: This might not be needed yet but i was wondering if it is a possibility.

zcash_client_backend/src/data_api.rs
```diff
@@ -1053,7 +1053,7 @@ pub struct ScannedBlock<A> {
 
 impl<A> ScannedBlock<A> {
     /// Constructs a new `ScannedBlock`
-    pub(crate) fn from_parts(
+    pub fn from_parts(
```
oxarbitrage, line 1056: Same as above.

zcash_client_backend/src/wallet.rs
```diff
@@ -98,6 +98,7 @@ impl<AccountId, N> Recipient<AccountId, Option<N>> {
 /// The shielded subset of a [`Transaction`]'s data that is relevant to a particular wallet.
 ///
 /// [`Transaction`]: zcash_primitives::transaction::Transaction
+#[derive(Clone)]
```
oxarbitrage, line 101: I had to do this  for my version of `put_blocks` as i was not able to make it otherwise, it might be not needed and not wanted.

oxarbitrage (commented):
I am pretty new to the codebase so i have a lot of unknowns and questions.

zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -379,12 +381,62 @@ impl WalletWrite for MemoryWalletDb {
...
+
+            let memory_block = MemoryWalletBlock {
+                height: block.block_height,
+                hash: block.block_hash,
+                block_time: block.block_time,
+                transactions,
+                // TODO: Add memos
+                memos: HashMap::new(),
+            };
+
```
nuttycom, line 435: This will have several additional responsibilities related to what will need to happen in the wallet overall:
* Note commitment trees will need to be updated.
* Received notes need to be made available for note selection & balance calculation
* Spent notes need to be made unavailable for note selection & balance calculation

zcash_client_backend/src/data_api.rs
nuttycom: If these are needed, they should be made as a separate PR outside of the context of this one.

zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -236,7 +235,10 @@ impl WalletRead for MemoryWalletDb {
     }
 
     fn chain_height(&self) -> Result<Option<BlockHeight>, Self::Error> {
-        todo!()
+        match self.blocks.last_key_value() {
```
nuttycom, line 238: Something to take into consideration here is how this is handled in `zcash_client_sqlite`: in that context, the chain tip may not yet have been scanned, so we get the current chain height as the maximum height known to the scan queue, rather than the blocks table.

zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -379,12 +381,76 @@ impl WalletWrite for MemoryWalletDb {
         todo!()
     }
 
+    /// Adds a sequence of blocks to the data store.
+    ///
+    /// Assumes blocks will be here in order.
     fn put_blocks(
         &mut self,
+        // TODO: Figure out what to do with this field.
         _from_state: &super::chain::ChainState,
```
nuttycom, line 390: This is used to ensure we have a correct "starting point" for adding to the note commitment trees in https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/lib.rs#L928-L934 and https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/lib.rs#L977-L983

zcash_client_backend/src/data_api/mem_wallet.rs
nuttycom: With respect to adding to the note commitment tree, it looks like you're assuming linear scanning here. That's probably okay for a v0, because it saves a bunch of complexity; I just haven't been thinking about that use case in a long time; if that's the case though, you won't be able to use the sync engine from https://github.com/zcash/librustzcash/pull/1217 

With respect to the other two comments:
Depending upon how balance is calculated, it may be that only the second comment is relevant; when a note is spent, we need to update our record for that note so that it's no longer included as having value belonging to the wallet.

nuttycom (commented):
Sorry it's taken me a bit to get to commenting on this; it's been a busy time getting Zashi 1.0 out. :)

zcash_client_backend/src/data_api/mem_wallet.rs
oxarbitrage: Blocks are added to the memory wallet database by `put_blocks` which currently receives a vector of `ScannedBlock`. Do we plan to change that ?

oxarbitrage: Whatever is the case, ill open a separated PR for this.

nuttycom: > Blocks are added to the memory wallet database by `put_blocks` which currently receives a vector of `ScannedBlock`. Do we plan to change that ?

No; the main distinction here is that the chain tip height is used in computing how to prioritize the scanning queue; we scan blocks out-of-order in many cases to improve time to spendability. The important related operation is `update_chain_tip`; basically, `chain_height` should return whatever is correct as of the last call to `update_chain_tip`, not the last call to `put_blocks`.

oxarbitrage: In `update_chain_tip` where do you store the height that is passed by argument so you can use it in `chain_height`?

zcash_client_backend/src/data_api/mem_wallet.rs
oxarbitrage: Thanks, i added that at https://github.com/zcash/librustzcash/pull/1298/commits/a4323c2d60d6696ad867bed63183206a1a874c47

Cargo.toml
oxarbitrage: We can use v0.1.3 here now. https://github.com/zcash/librustzcash/pull/1298/commits/a3b4de0459e2ee443385b7a062910644db6ac400

zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -379,12 +378,98 @@ impl WalletWrite for MemoryWalletDb {
...
+                hash: block.block_hash,
+                block_time: block.block_time,
+                transactions,
+                // TODO: Add memos
+                memos: HashMap::new(),
+            };
+
+            self.blocks.insert(block.block_height, memory_block);
+
+            // Add the sapling commitments to the sapling tree.
```
oxarbitrage, line 459: Is this still needed now that we added the frontier?

nuttycom: Yes; the frontier just provides a starting point, but the note commitments for each note in the block, and the commitment for the block-end checkpoint, must also be added.

oxarbitrage: Thanks, i added the orchard version of this at https://github.com/zcash/librustzcash/pull/1298/commits/370a4aa7e522faca1b3bc56ae9c7087fa7a6ac91

zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -379,12 +380,142 @@ impl WalletWrite for MemoryWalletDb {
...
+            };
+
+            self.blocks.insert(block.block_height, memory_block);
+
+            // Add the Sapling commitments to the sapling tree.
+            let block_commitments = block.into_commitments();
+            if let Some(value) = from_state.final_sapling_tree().value() {
+                self.sapling_tree
+                    .batch_insert(value.position(), block_commitments.sapling.into_iter());
+            }
```
nuttycom, line 508: This is a bug, in that it doesn't handle constructing the tree starting from Sapling activation properly. I think something like this is necessary:

```suggestion
            let start_position = from_state.final_sapling_tree().value().map_or(0.into(), |t| t.position() + 1);
            self.sapling_tree.batch_insert(start_position, block_commitments.sapling.into_iter());
```

zcash_client_backend/src/data_api/mem_wallet.rs
```diff
@@ -379,12 +380,142 @@ impl WalletWrite for MemoryWalletDb {
...
+
+            #[cfg(feature = "orchard")]
+            // Add the Orchard commitments to the orchard tree.
+            if let Some(value) = from_state.final_orchard_tree().value() {
+                self.orchard_tree
+                    .batch_insert(value.position(), block_commitments.orchard.into_iter());
+            }
```
nuttycom, line 515:
```suggestion
            {
            // Add the Orchard commitments to the orchard tree.
            let start_position = from_state.final_orchard_tree().value().map_or(0.into(), |t| t.position() + 1);
            self.orchard_tree.batch_insert(start_position, block_commitments.orchard.into_iter());
```

nuttycom (changes requested):
The incorrect start to the note commitment tree position is blocking.

oxarbitrage:
I fixed the clippy warnings in https://github.com/zcash/librustzcash/pull/1298/commits/c9a236e90ef16cebdcebcba2566613f42da9236f

The orchard tests failures are unrelated to this PR as they are also happening in https://github.com/zcash/librustzcash/pull/1293

nuttycom (approved):
utACK c9a236e90ef16cebdcebcba2566613f42da9236f

PR #1300, opened 2024-03-22
zcash_client_sqlite: Fix balance in expired unmined transaction history.

### LINKED ISSUE #1292: zcash_client_sqlite: Value of expired transactions is reported incorrectly from the `v_transactions` view.
At present, the `sapling_received_notes` and `orchard_received_notes` tables both have a `spent` column, which references the transaction in which the note is spent, if any. This works fine under ordinary circumstances, but if a transaction expires without being mined, the reference to the transaction is removed from the `spent` column. This then causes a problem for the `v_transactions` view, which can no longer correctly represent the original balance of the transaction.

The fix for this is to use a junction table between the `*received_notes` tables and the spending transactions, and for queries that use this association to determine spentness to verify whether or not a transaction has expired instead of simply checking whether or not the `spent` column is null.

### LINKED ISSUE #1299: zcash_client_sqlite: Do not delete received notes in chain rollback.
We should not discard "failed received notes" - they may have memo data associated with them that we can still read, and we just need to be sure that they do not contribute to balance or spendability.

nuttycom:
Fixes #1292
Fixes #1299

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -1128,12 +1128,20 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
              LEFT OUTER JOIN v_{table_prefix}_shards_scan_state scan_state
                 ON n.commitment_tree_position >= scan_state.start_position
                 AND n.commitment_tree_position < scan_state.end_position_exclusive
-             WHERE n.spent IS NULL
-             AND (
-                 t.expiry_height IS NULL
-                 OR t.block IS NOT NULL
-                 OR t.expiry_height >= :summary_height
-             )",
+             WHERE (
+                t.block IS NOT NULL -- the receiving tx is mined
+                OR t.expiry_height IS NULL -- the receiving tx will not expire
+                OR t.expiry_height >= :summary_height -- the receiving tx is unexpired
+             )
+             -- and the received note is unspent
+             AND n.id NOT IN (
+               SELECT {table_prefix}_received_note_id
+               FROM {table_prefix}_received_note_spends
+               JOIN transactions t ON t.id_tx = transaction_id
+               WHERE t.block IS NOT NULL -- the spending transaction is mined
+               OR t.expiry_height IS NULL -- the spending tx will not expire
+               OR t.expiry_height > :summary_height -- the spending tx is unexpired
+             )"
         ))?;
 
         let mut rows =
@@ -1245,10 +1253,17 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
         let mut stmt_transparent_balances = tx.prepare(
             "SELECT u.received_by_account_id, SUM(u.value_zat)
              FROM utxos u
-             LEFT OUTER JOIN transactions tx
-             ON tx.id_tx = u.spent_in_tx
              WHERE u.height <= :max_height
-             AND (u.spent_in_tx IS NULL OR (tx.block IS NULL AND tx.expiry_height <= :stable_height))
+             -- and the received txo is unspent
+             AND u.id NOT IN (
+               SELECT transparent_received_output_id
+               FROM transparent_received_output_spends txo_spends
+               JOIN transactions tx
+                 ON tx.id_tx = txo_spends.transaction_id
+               WHERE tx.block IS NOT NULL -- the spending tx is mined
+               OR tx.expiry_height IS NULL -- the spending tx will not expire
+               OR tx.expiry_height > :stable_height -- the spending tx is unexpired
+             )
              GROUP BY u.received_by_account_id",
         )?;
         let mut rows = stmt_transparent_balances.query(named_params![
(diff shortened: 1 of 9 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -1913,36 +1956,15 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
-            );",
-            [u32::from(block_height)],
-        )?;
-
         // Do not delete sent notes; this can contain data that is not recoverable
         // from the chain. Wallets must continue to operate correctly in the
         // presence of stale sent notes that link to unmined transactions.
+        // Also, do not delete received notes; they may contain memo data that is
+        // not recoverable; balance APIs must ensure that un-mined received notes
+        // do not count towards spendability or transaction balalnce.
```
daira, line 1964:
```suggestion
        // do not count towards spendability or transaction balance.
```
Non-blocking.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1245,10 +1253,17 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
+             -- and the received txo is unspent
+             AND u.id NOT IN (
+               SELECT transparent_received_output_id
+               FROM transparent_received_output_spends txo_spends
+               JOIN transactions tx
+                 ON tx.id_tx = txo_spends.transaction_id
+               WHERE tx.block IS NOT NULL -- the spending tx is mined
+               OR tx.expiry_height IS NULL -- the spending tx will not expire
+               OR tx.expiry_height > :stable_height -- the spending tx is unexpired
+             )
```
daira, line 1266: Can some of the duplication in the SQL be avoided by defining a view (or views for Sapling and Orchard) of notes that are considered spent? Non-blocking.

daira: Oh, the PR does create such views. I don't understand why there is so much SQL duplication, then. It looks pretty error-prone to maintain :-/

daira (approved):
utACK with comments.

zcash_client_sqlite/src/wallet.rs
nuttycom: I want to make a pass after 1.0 is out to do the refactoring to eliminate a whole bunch of the duplication, but I want that to be a pure refactoring rather than part of fixing bugs.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -584,26 +611,19 @@ mod tests {
             r#"CREATE INDEX orchard_received_notes_account ON orchard_received_notes (
                 account_id ASC
             )"#,
-            r#"CREATE INDEX orchard_received_notes_spent ON orchard_received_notes (
```
str4d, line 587: Should these index removals be replaced with index creations for the foreign keys of the new junction tables?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1898,7 +1923,25 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+        "DELETE FROM scan_queue
+        WHERE block_range_start >= :new_end_height",
+        named_params![":new_end_height": u32::from(block_height + 1)],
+    )?;
+    conn.execute(
+        "UPDATE scan_queue
+        SET block_range_end = :new_end_height
+        WHERE block_range_end > :new_end_height",
+        named_params![":new_end_height": u32::from(block_height + 1)],
+    )?;
```
str4d, line 1940: Why was this moved to earlier in the truncation process?

str4d (approved):
Concept ACK fada46e6383ccdb78eff56f7709c9246426a80db. I reviewed `zcash_client_sqlite/src/wallet/init.rs` for the effects of the migration changes on the tables, and skimmed the rest of the PR. I did not review any changes to SQL queries.

zcash_client_sqlite/src/wallet.rs
nuttycom: It was previously guarded by the 'if' statement below, meaning that the scan queue was not being correctly updated in all cases. This was one of the bugs fixed by this PR.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2037,11 +2033,15 @@ pub(crate) fn get_unspent_transparent_output(
     let mut stmt_select_utxo = conn.prepare_cached(
         "SELECT u.prevout_txid, u.prevout_idx, u.script, u.value_zat, u.height
          FROM utxos u
-         LEFT OUTER JOIN transactions tx
-         ON tx.id_tx = u.spent_in_tx
          WHERE u.prevout_txid = :txid
          AND u.prevout_idx = :output_index
-         AND tx.block IS NULL",
+         AND u.id NOT IN (
+            SELECT txo_spends.id
```
str4d, line 2039: Bug: this should be:
```suggestion
            SELECT txo_spends.transparent_received_output_id
```

Bug found by Lukas while testing the Zashi shield funds button.

PR #1301, opened 2024-03-22
zcash_client_sqlite: Fix tests that broke with the addition of Orchard UA receivers.

nuttycom:
These tests were broken by #1291

### DIFF
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -326,7 +326,7 @@ mod tests {
         address::Address,
         data_api::scanning::ScanPriority,
         encoding::{encode_extended_full_viewing_key, encode_payment_address},
-        keys::{sapling, UnifiedFullViewingKey, UnifiedSpendingKey},
+        keys::{sapling, UnifiedAddressRequest, UnifiedFullViewingKey, UnifiedSpendingKey},
     };
 
     use ::sapling::zip32::ExtendedFullViewingKey;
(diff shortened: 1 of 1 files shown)


str4d (dismissed):
utACK 72c346c8ddeabba1f3fdd212fc35863d637b23d2

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -1482,10 +1484,13 @@ mod tests {
                 assert_eq!(DiversifierIndex::from(tv.diversifier_index), di);
                 assert_eq!(tvua.transparent(), ua.transparent());
                 assert_eq!(tvua.sapling(), ua.sapling());
+                #[cfg(not(feature = "orchard"))]
                 assert_eq!(tv.unified_addr, ua.encode(&Network::MainNetwork));
 
+                // hardcoded with knowledge of what's coming next
+                let ua_request = UnifiedAddressRequest::unsafe_new(false, true, true);
```
str4d, line 1491: Should this also be `UA_TRANSPARENT` like before?

str4d (approved):
Re-utACK 6f1e4aff1acceedd55244857901678ae1fc900a7 with question.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: No; the whole test is under the `transparent-inputs` feature.

daira (commented):
Post-hoc ACK 6f1e4aff1acceedd55244857901678ae1fc900a7

Issue #1302, opened 2024-03-22
zcash_client_sqlite: Inconsistency between tree size & checkpoint position updates.


nuttycom:
This selection of mainnet checkpoint data illustrates the issue: 
```
height  | s_tree_sz | position | o_tree_sz | position
--------|-----------|----------|-----------|---------
2399872 | 73205447  | 73205446 | 48705967  | 48705960
2399871 | 73205445  |          | 48705967  |  
2399870 | 73205445  |          | 48705967  |  
2399869 | 73205445  |          | 48705967  |  
2399868 | 73205445  | 73205444 | 48705967  | 48705960
2399867 | 73205443  |          | 48705967  |  
2399866 | 73205443  |          | 48705967  |  
2399865 | 73205443  | 73205442 | 48705967  | 48705960 <-- problem here
2399864 | 73205441  | 73205440 | 48705961  | 48705960
2399863 | 73205439  |          | 48705961  |  
2399862 | 73205439  | 73205438 | 48705961  | 48705960
2399861 | 73205437  | 73205436 | 48705959  | 48705958
2399860 | 73205437  | 73205436 | 48705957  | 48705956
2399859 | 73205437  | 73205436 | 48705955  | 48705954
```
In this trace, it appears that an Orchard checkpoint is inserted at height `2399865` that does not reflect the additions to the Orchard note commitment tree that appear in that block. None of the later checkpoints then observe that adjustment.

nuttycom:
Fixed by #1303

nuttycom:
This does not reproduce when using a batch size of 10,000 blocks, but with a batch size of 100 blocks it reproduces reliably.

Issue #1294, opened 2024-03-20
Frequent panics during `scan_cached_blocks`: "Tree state inconsistent with checkpoints."


AArnott:
I'm seeing _very_ frequent panics during account recovery or even incremental syncs. Here is the output:

```
thread 'tokio-runtime-worker' panicked at C:\Users\andre\.cargo\git\checkouts\incrementalmerkletree-220c7a5aa40ed638\fa147c8\shardtree\src\prunable.rs:865:33:
Tree state inconsistent with checkpoints.
stack backtrace:
   0:     0x7ff66dd54182 - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\sys_common\backtrace.rs:44
   1:     0x7ff66dd7982d - core::fmt::rt::Argument::fmt
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\core\src\fmt\rt.rs:142
   2:     0x7ff66dd7982d - core::fmt::write
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\core\src\fmt\mod.rs:1120
   3:     0x7ff66dd4f3f1 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\io\mod.rs:1810
   4:     0x7ff66dd53faa - std::sys_common::backtrace::_print
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\sys_common\backtrace.rs:47
   5:     0x7ff66dd53faa - std::sys_common::backtrace::print
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\sys_common\backtrace.rs:34
   6:     0x7ff66dd56a49 - std::panicking::default_hook::closure$1
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\panicking.rs:272
   7:     0x7ff66dd56705 - std::panicking::default_hook
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library\std\src\panicking.rs:292
   8:     0x7ff66dd56f74 - std::panicking::rust_panic_with_hook
[...]

PR #1303, opened 2024-03-22
zcash_client_sqlite: Fix a problem related to checkpoint pruning.

nuttycom:
This fixes, but does not yet reproduce the bug in #1302, which is likely the same as the bug in #1294.

The error occurs as a consequence of checkpoints that were added for "companion" heights in the other tree being pruned after the nodes that were associated with those checkpoints had been previously pruned. This could occur because updates to the note commitment trees were done in two stages:
* In the initial stage, the tree state is updated for a scanned range, which may trigger pruning of previous checkpoints.
* In the second stage, the "missed" companion checkpoints are re-added to the tree, even if the leaf positions that they were referring to had already been pruned.
* When _those_ checkpoints are later pruned, multiple checkpoints are associated with a single internal node in the note commitment tree whose children have been pruned. This causes the panic observed in #1294.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -838,7 +838,7 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                 >(
                     // An iterator of checkpoints heights for which we wish to ensure that
                     // checkpoints exists.
-                    checkpoint_heights: I,
+                    ensure_heights: I,
                     // The map of checkpoint positions from which we will draw note commitment tree
                     // position information for the newly created checkpoints.
                     existing_checkpoint_positions: &BTreeMap<BlockHeight, Position>,
@@ -846,15 +846,15 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                     // there is no preceding checkpoint in existing_checkpoint_positions.
                     state_final_tree: &Frontier<H, DEPTH>,
                 ) -> Vec<(BlockHeight, Checkpoint)> {
-                    checkpoint_heights
-                        .flat_map(|from_checkpoint_height| {
+                    ensure_heights
+                        .flat_map(|ensure_height| {
                             existing_checkpoint_positions
-                                .range::<BlockHeight, _>(..=*from_checkpoint_height)
+                                .range::<BlockHeight, _>(..=*ensure_height)
                                 .last()
                                 .map_or_else(
                                     || {
                                         Some((
-                                            *from_checkpoint_height,
+                                            *ensure_height,
                                             state_final_tree
                                                 .value()
                                                 .map_or_else(Checkpoint::tree_empty, |t| {
@@ -862,10 +862,10 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                                                 }),
                                         ))
                                     },
-                                    |(to_prev_height, position)| {
-                                        if *to_prev_height < *from_checkpoint_height {
+                                    |(existing_checkpoint_height, position)| {
+                                        if *existing_checkpoint_height < *ensure_height {
                                             Some((
-                                                *from_checkpoint_height,
+                                                *ensure_height,
                                                 Checkpoint::at_position(*position),
                                             ))
                                         } else {
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -612,7 +612,7 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
         Ok(())
     }
 
-    #[tracing::instrument(skip_all, fields(height = blocks.first().map(|b| u32::from(b.height()))))]
+    #[tracing::instrument(skip_all, fields(height = blocks.first().map(|b| u32::from(b.height())), count = blocks.len()))]
     #[allow(clippy::type_complexity)]
     fn put_blocks(
         &mut self,
@@ -881,16 +881,17 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                 }
 
                 #[cfg(feature = "orchard")]
-                let missing_sapling_checkpoints = ensure_checkpoints(
-                    orchard_checkpoint_positions.keys(),
-                    &sapling_checkpoint_positions,
-                    from_state.final_sapling_tree(),
-                );
-                #[cfg(feature = "orchard")]
-                let missing_orchard_checkpoints = ensure_checkpoints(
-                    sapling_checkpoint_positions.keys(),
-                    &orchard_checkpoint_positions,
-                    from_state.final_orchard_tree(),
+                let (missing_sapling_checkpoints, missing_orchard_checkpoints) = (
+                    ensure_checkpoints(
+                        orchard_checkpoint_positions.keys(),
+                        &sapling_checkpoint_positions,
+                        from_state.final_sapling_tree(),
+                    ),
+                    ensure_checkpoints(
+                        sapling_checkpoint_positions.keys(),
+                        &orchard_checkpoint_positions,
+                        from_state.final_orchard_tree(),
+                    ),
                 );
 
                 // Update the Sapling note commitment tree with all newly read note commitments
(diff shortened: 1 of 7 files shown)


zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1599,6 +1613,212 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
     );
 }
 
+#[cfg(feature = "orchard")]
+pub(crate) fn multi_pool_checkpoint<P0: ShieldedPoolTester, P1: ShieldedPoolTester>() {
```
daira, line 1617: The PR description says it does not reproduce the bug; is that out-of-date? There appears to be uncovered code according to codecov.

daira (dismissed):
utACK with comment.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1599,6 +1613,212 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
...
+        usize::try_from(u32::from(max_height) - u32::from(next_to_scan) + 1).unwrap(),
+    );
+
+    let expected_final =
+        (mid_balance + note_value - (transfer_amount * 3).unwrap() - (expected_fee * 3).unwrap())
+            .unwrap();
+    assert_eq!(st.get_total_balance(acct_id), expected_final);
+
+    use incrementalmerkletree::Position;
+    let expected_checkpoints_p0: Vec<(BlockHeight, ShieldedProtocol, Option<Position>)> = [
```
str4d, line 1728: It would be useful to add some inline comments to the test above that describe how the tree state is being manipulated by each choice of scan range, which lead to these expected checkpoints.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1599,6 +1613,212 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
...
+        // activation after Sapling
+        .build();
+
+    let account = st.test_account().cloned().unwrap();
+
+    let p0_fvk = P0::random_fvk(&mut st.rng);
+    let p1_fvk = P1::random_fvk(&mut st.rng);
+
+    let note_value = NonNegativeAmount::const_from_u64(10000);
+    // Generate 100 P0 blocks, then 100 P1 blocks, then another 100 P0 blocks.
```
str4d, line 1805: Bug: this documentation does not match what the code is actually doing (which is much more tightly interleaving P0 and P1 blocks).

zcash_client_sqlite/src/lib.rs
```diff
@@ -910,12 +911,37 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                         }
 
                         // Ensure we have a Sapling checkpoint for each checkpointed Orchard block height
```
str4d, line 913:
```suggestion
                        // Ensure we have a Sapling checkpoint for each checkpointed Orchard block height.
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -939,13 +965,38 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                             orchard_tree.insert_tree(tree, checkpoints)?;
                         }
 
-                        for (height, checkpoint) in &missing_orchard_checkpoints {
-                            orchard_tree
-                                .store_mut()
-                                .add_checkpoint(*height, checkpoint.clone())
+                        // Ensure we have an Orchard checkpoint for each checkpointed Sapling block height
```
str4d, line 968:
```suggestion
                        // Ensure we have an Orchard checkpoint for each checkpointed Sapling block height.
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -910,12 +911,37 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
                         #[cfg(feature = "orchard")]
-                        for (height, checkpoint) in &missing_sapling_checkpoints {
-                            sapling_tree
-                                .store_mut()
-                                .add_checkpoint(*height, checkpoint.clone())
+                        {
+                            let min_checkpoint_height = sapling_tree
+                                .store()
+                                .min_checkpoint_id()
                                 .map_err(ShardTreeError::Storage)?;
```
str4d, line 921: Will this ever return `None` if called here? We insert an explicit checkpoint for `from_state`, so surely we will always have a checkpoint ID by now. I think for the intended effect, we need to call this before we make any modifications to the Sapling tree.

zcash_client_sqlite/src/lib.rs
```diff
@@ -910,12 +911,37 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+
+                            if let Some(h) = min_checkpoint_height {
+                                missing_sapling_checkpoints
+                                    .iter()
+                                    .filter(|(ch, _)| ch > &h)
+                                    .try_for_each(add_checkpoint)?;
+                            } else {
+                                let skip_len = missing_sapling_checkpoints
+                                    .len()
+                                    .saturating_sub(usize::try_from(PRUNING_DEPTH).unwrap());
```
str4d, line 939: This skips checkpoints that are verifiably going to be pruned, but AFAICT it doesn't factor in any checkpoints we added from `from_state` or the scanned blocks. If that is intentional (i.e. we are still going to insert more checkpoints than will exist post-pruning, and are just calculating in this way for simplicity), add a comment.

zcash_client_sqlite/src/lib.rs
```diff
@@ -939,13 +965,38 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
-                                .store_mut()
-                                .add_checkpoint(*height, checkpoint.clone())
+                        // Ensure we have an Orchard checkpoint for each checkpointed Sapling block height
+                        // We skip all checkpoints below the minimum retained checkpoint in the
+                        // Orchard tree, because branches below this height may be pruned.
+                        {
+                            let min_checkpoint_height = orchard_tree
+                                .store()
+                                .min_checkpoint_id()
                                 .map_err(ShardTreeError::Storage)?;
```
str4d, line 975: Ditto: I think this needs to be called before we modify the Orchard tree.

zcash_client_sqlite/src/lib.rs
```diff
@@ -939,13 +965,38 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+
+                            if let Some(h) = min_checkpoint_height {
+                                missing_orchard_checkpoints
+                                    .iter()
+                                    .filter(|(ch, _)| ch > &h)
+                                    .try_for_each(add_checkpoint)?;
+                            } else {
+                                let skip_len = missing_orchard_checkpoints
+                                    .len()
+                                    .saturating_sub(usize::try_from(PRUNING_DEPTH).unwrap());
```
str4d, line 993: Ditto re: question.

str4d (changes requested):
Reviewed a5f2682ece1d2fef4bbd74463473834e2486d73d

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1599,6 +1613,212 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
...
+    // Generate several empty blocks
+    for _ in 0..10 {
+        st.generate_empty_block();
+    }
+
+    // Scan into the middle of the empty range
+    let scanned = st.scan_cached_blocks(next_to_scan, 5);
+    let next_to_scan = scanned.scanned_range().end;
+
+    let mid_balance = (note_value * 3).unwrap();
```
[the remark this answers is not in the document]
nuttycom: The possible bug is not a bug. :) "In the middle" just meant "somewhere in the span"

zcash_client_sqlite/src/testing/pool.rs
nuttycom: This is the bit where I am _still_ unable to actually reproduce the error except in a mainnet wallet, and I'm kind of at wit's end.

zcash_client_sqlite/src/lib.rs
nuttycom: This has to be called after `insert_tree`, because that is the call that triggers the pruning that then may make this insertion invalid.

zcash_client_sqlite/src/lib.rs
nuttycom: As noted, this whole clause can go away.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1599,6 +1613,212 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
...
+    let p0_fvk = P0::test_account_fvk(&st);
+    let p1_fvk = P1::test_account_fvk(&st);
+
+    // Add some funds to the wallet; we add two notes to allow successive spends. Also,
+    // we will generate a note in the P1 pool to ensure that we have some tree state.
+    let note_value = NonNegativeAmount::const_from_u64(500000);
+    st.generate_next_block(&p0_fvk, AddressType::DefaultExternal, note_value);
+    st.generate_next_block(&p0_fvk, AddressType::DefaultExternal, note_value);
+    st.generate_next_block(&p1_fvk, AddressType::DefaultExternal, note_value);
+    let scanned = st.scan_cached_blocks(account.birthday().height(), 2);
```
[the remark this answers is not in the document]
nuttycom: Scanning only the first two blocks wasn't a bug, but the issue I was trying to trigger by scanning the third block separately wasn't actually an issue so I'll remove the third block anyway.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/a5f2682ece1d2fef4bbd74463473834e2486d73d..635057d86faa257f103eda8d75736c7b8dd972c4) to address review comments.

AArnott:
I merged these changes into my branch and no panics occurred within a few minutes (where they usually show up within several seconds. 🎉🎉

str4d (approved):
utACK 635057d86faa257f103eda8d75736c7b8dd972c4. Even though the test fails to reproduce the bug (please open a new issue to track this reproduction), this change can't make the bug more likely as it undoes some of the checkpoint insertion changes made in #1262.

There's a chance that it makes the bug that #1262 was fixing more likely to reoccur, but if it does we can address that in a follow-up PR.

nuttycom:
Opened https://github.com/zcash/librustzcash/issues/1308 for the follow-up test fixes.

Issue #1282, opened 2024-03-18
Panic when a transaction has payments into more than one account tracked by sqlite


AArnott:
The latest librustzcash in main, with the orchard feature enabled, panics during account recovery for me, on this line: https://github.com/zcash/librustzcash/blob/e0227ed175d8e1c6d1aaffb2e69ba930069a4c10/zcash_client_sqlite/src/lib.rs#L1047

When adding a transaction to the sqlite database, if the tx has multiple outputs, destined for more than one account that is tracked by the sqlite database, a panic results.

Panic text:
> This isn't a z->t spend at least because we're in orchard (not sapling).

I also don't see anything suggesting this is an unshielding transaction. We're just looping over orchard outputs here.
Why is the local variable called `spending_account_id` when in fact it is set to the value of the _receiving_ account id?
And why must every output in a transaction have the same receiving account? This looks like a panic will occur for any transaction that sends outputs to more than one account tracked by the sqlite db. 
And in fact, I found this code due to a panic when doing account recovery for a testnet account I have, and a multi-account transaction is very likely to be what I have here (though I haven't confirmed this.)
See also [discord post](https://discord.com/channels/809218587167293450/876655911790321684/1219086198513860659).

_Originally posted by @AArnott in https://github.com/zcash/librustzcash/pull/1234#discussion_r1527664053_

nuttycom:
The code that causes this problem was originally introduced in 665c4c7aff3743fce38e7527b553197c51ffa65e in 2021, very early on in the development of this code. I think that the original rationale was perhaps related to the fact that when using decrypted transaction data, we can only consistently obtain the account ID for incoming transactions. Since there was only ever one account in the wallet, the incoming and outgoing account were assumed to be the same; that assumption was broken when we later added the ability to create additional accounts, but we never ran into it (because the mobile wallets still only use a single account.)

What needs to be done to resolve this is to determine the sending account by checking against the nullifiers of the notes that were spent instead.

nuttycom:
Fixed by #1306

Issue #1238, opened 2024-03-08
zcash_client_sqlite: Figure out why we're enforcing "one account for all outputs" rule


nuttycom:
> We need to figure out why we're enforcing this "one account only for outputs" rule but aren't doing anything with the account ID that we resolve on.

_Originally posted by @nuttycom in https://github.com/zcash/librustzcash/pull/1234#discussion_r1518285492_

nuttycom:
Fixed in #1306

Issue #1305, opened 2024-03-23
zcash_client_backend: Correctly track accounts that fund each transaction output.


nuttycom:
>The complete solution here would be something like:
>* Add a method to `WalletRead` that is something like `detect_funding_accounts` that checks whether any of the wallet's accounts contributed value to the transaction.
>* Remove the `from_account_id` column from the `sent_notes` table, and replace it with a `funding_accounts` junction table between `sent_notes` and `accounts`. 

_Originally posted by @nuttycom in https://github.com/zcash/librustzcash/issues/1304#issuecomment-2016543402_

nuttycom:
Also consider the remaining comments on #1306 when implementing these changes.

AArnott:
I have another case like #1304 where `from_account_id` is incorrectly NULL. But as that issue was closed and this one is active and seems to convey the idea of fixing "the whole problem", is this the right place to report it?

The following shows a _de_shielding transaction. Both rows were funded and directed at account_id 1, yet the top row (the deshielded value) claims to be novel income by leaving `from_account_id` with a NULL value:
![image](https://github.com/zcash/librustzcash/assets/3548/5ae8c84c-79c7-4d72-9e65-0cc482297235)

AArnott:
I guess it's a hard-coded limitation. The `v_tx_outputs` view literally hard-codes `from_account_id` to always be `NULL`:

```sql
SELECT utxos.prevout_txid           AS txid,
       0                            AS output_pool,
       utxos.prevout_idx            AS output_index,
       NULL                         AS from_account_id,
       utxos.received_by_account_id AS to_account_id,
       utxos.address                AS to_address,
       utxos.value_zat              AS value,
       0                            AS is_change,
       NULL                         AS memo
FROM utxos
```

AArnott:
I have my own view that is based on `v_tx_outputs` and joins with a few other views/tables. I was able to recover the proper `from_account_id` value by replacing the column in the original view with this:

```sql
coalesce(
   txo.from_account_id,
   (SELECT account_id FROM sapling_received_notes srn WHERE srn.id = (SELECT sapling_received_note_id FROM sapling_received_note_spends srns WHERE transaction_id = tx.id_tx)),
   (SELECT account_id FROM orchard_received_notes srn WHERE srn.id = (SELECT orchard_received_note_id FROM orchard_received_note_spends srns WHERE transaction_id = tx.id_tx))
) AS from_account_id,
```

[...]

PR #1306, opened 2024-03-23
zcash_client_sqlite: Use consistent criteria for detecting what accounts fund a transaction.

### LINKED ISSUE #1304: v_tx_outputs `from_account_id` column left NULL for notes sent to self
This issue screws up running balance calculations in my app, by interfering with the ability to detect 'change' line items in a transaction.  I expect change rows in `v_tx_outputs` to have `from_account_id` and `to_account_id` set to the same value, but instead, `from_account_id` is set to NULL, suggesting that all change is novel _income_ for the account, causing running balance to skyrocket.

I believe this is a recent regression in librustzcash because I had running balance working until I merged recently to get the orchard changes.

Repro steps:

Execute account recovery using this UFVK:
```
uviewtest1tcygrtut692vqlx9nlyknx0fqq59am5vhaf97gxncqnfn87qrgey68777tumstc2lcp4r9yxd3fknkpmxtgw8awhcg40cw00ahvtaeqmpfqvjz6e3v234zsfvdvt6dm8dpzxv970wkdv2jrfm3t2m9cde9ry8mrxr286ns4yqwmcx3k4netqqhgldthnhzhlpg0lk00eruy4tf3fx3k9xn7fywppj8wyzjjc3dcrqe6kxnc6zxpfly9e2uk3k7jyy3n70zpj5zfheedzz0sw2pp96rvy9xt2dw94nplfx0usrwtshrmf5xwq84qcq459kvks5g28gvkxrjpujgc9gkjt5np5m4afruk0z8zlyd65hfqqu9pg3u9lkk26r7ad4l59yy3tn2xlmad42a6kee8l92ddj7tgf4fhv4x9thx7kf28jc6gvf3xr3lhtegd8ly590595g6dfh3w7nalmt8zx6yestgfqu9uvg8gkwkwmuc4u5jsecqu
```
(or, if you prefer, use the same seed phrase that I privately sent to @nuttycom over Signal).

Open the sqlite data file in DB Browser for SQLite and inspect the `v_tx_outputs` table:

image

The selection in the above screenshot show the two outputs of a particular orchard transaction. The top one is change, but notice how `from_account_id` is `NULL`.
It's the same for _all_ change in this table. In fact you can see not a single row has both `from_account_id` and `to_account_id` set to non-NULL values, which is of course highly suspicious, as that would require every one of 21 transactions to have exactly no change.

In fact it isn't just change. One transaction in this account's history records an explicit 'send to self' transaction. Both the user amount and the change amount rows have `from_account_id` left at NULL.
image

nuttycom:
Fixes #1304, #1282, #1238

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -1062,16 +1062,27 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
 
             for output in d_tx.sapling_outputs() {
                 match output.transfer_type() {
-                    TransferType::Outgoing | TransferType::WalletInternal => {
-                        let recipient = if output.transfer_type() == TransferType::Outgoing {
-                            //TODO: Recover the UA, if possible.
-                            Recipient::Sapling(output.note().recipient())
-                        } else {
-                            Recipient::InternalAccount {
-                                receiving_account: *output.account(),
-                                external_address: None,
-                                note: Note::Sapling(output.note().clone()),
-                            }
+                    TransferType::Outgoing => {
+                        //TODO: Recover the UA, if possible.
+                        let recipient = Recipient::Sapling(output.note().recipient());
+                        wallet::put_sent_output(
+                            wdb.conn.0,
+                            &wdb.params,
+                            *output.account(),
+                            tx_ref,
+                            output.index(),
+                            &recipient,
+                            output.note_value(),
+                            Some(output.memo()),
+                        )?;
+                    }
+                    TransferType::WalletInternal => {
+                        wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
+
+                        let recipient = Recipient::InternalAccount {
+                            receiving_account: *output.account(),
+                            external_address: None,
+                            note: Note::Sapling(output.note().clone()),
                         };
 
                         wallet::put_sent_output(
@@ -1084,10 +1095,6 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                             output.note_value(),
                             Some(output.memo()),
                         )?;
-
-                        if matches!(recipient, Recipient::InternalAccount { .. }) {
-                            wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
-                        }
                     }
                     TransferType::Incoming => {
                         wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
@@ -1118,24 +1125,36 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
             #[cfg(feature = "orchard")]
             for output in d_tx.orchard_outputs() {
                 match output.transfer_type() {
-                    TransferType::Outgoing | TransferType::WalletInternal => {
-                        let recipient = if output.transfer_type() == TransferType::Outgoing {
-                            // TODO: Recover the actual UA, if possible.
-                            Recipient::Unified(
-                                UnifiedAddress::from_receivers(
-                                    Some(output.note().recipient()),
-                                    None,
-                                    None,
-                                )
-                                .expect("UA has an Orchard receiver by construction."),
-                                PoolType::Shielded(ShieldedProtocol::Orchard),
+                    TransferType::Outgoing => {
+                        // TODO: Recover the actual UA, if possible.
+                        let recipient = Recipient::Unified(
+                            UnifiedAddress::from_receivers(
+                                Some(output.note().recipient()),
+                                None,
+                                None,
                             )
-                        } else {
-                            Recipient::InternalAccount {
-                                receiving_account: *output.account(),
-                                external_address: None,
-                                note: Note::Orchard(*output.note()),
-                            }
+                            .expect("UA has an Orchard receiver by construction."),
+                            PoolType::Shielded(ShieldedProtocol::Orchard),
+                        );
+
+                        wallet::put_sent_output(
+                            wdb.conn.0,
+                            &wdb.params,
+                            *output.account(),
+                            tx_ref,
+                            output.index(),
+                            &recipient,
+                            output.note_value(),
+                            Some(output.memo()),
+                        )?;
+                    }
+                    TransferType::WalletInternal => {
+                        wallet::orchard::put_received_note(wdb.conn.0, output, tx_ref, None)?;
+
+                        let recipient = Recipient::InternalAccount {
+                            receiving_account: *output.account(),
+                            external_address: None,
+                            note: Note::Orchard(*output.note()),
                         };
 
                         wallet::put_sent_output(
@@ -1148,10 +1167,6 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                             output.note_value(),
                             Some(output.memo()),
                         )?;
-
-                        if matches!(recipient, Recipient::InternalAccount { .. }) {
-                            wallet::orchard::put_received_note(wdb.conn.0, output, tx_ref, None)?;
-                        }
                     }
                     TransferType::Incoming => {
                         wallet::orchard::put_received_note(wdb.conn.0, output, tx_ref, None)?;
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -65,6 +65,7 @@ use zcash_client_backend::{
     wallet::{Note, NoteId, ReceivedNote, Recipient, WalletTransparentOutput},
     DecryptedOutput, PoolType, ShieldedProtocol, TransferType,
 };
+use zcash_keys::address::Address;
 use zcash_primitives::{
     block::BlockHash,
     consensus::{self, BlockHeight},
@@ -1066,10 +1067,11 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                             //TODO: Recover the UA, if possible.
                             Recipient::Sapling(output.note().recipient())
                         } else {
-                            Recipient::InternalAccount(
-                                *output.account(),
-                                Note::Sapling(output.note().clone()),
-                            )
+                            Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                external_address: None,
+                                note: Note::Sapling(output.note().clone()),
+                            }
                         };
 
                         wallet::put_sent_output(
@@ -1083,7 +1085,7 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                             Some(output.memo()),
                         )?;
 
-                        if matches!(recipient, Recipient::InternalAccount(_, _)) {
+                        if matches!(recipient, Recipient::InternalAccount { .. }) {
                             wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
                         }
                     }
@@ -1091,11 +1093,12 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                         wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
 
                         if let Some(account_id) = funding_account {
-                            // Even if the recipient address is external, record the send as internal.
-                            let recipient = Recipient::InternalAccount(
-                                *output.account(),
-                                Note::Sapling(output.note().clone()),
-                            );
+                            let recipient = Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                // TODO: recover the actual UA, if possible
+                                external_address: Some(Address::Sapling(output.note().recipient())),
+                                note: Note::Sapling(output.note().clone()),
+                            };
 
                             wallet::put_sent_output(
                                 wdb.conn.0,
@@ -1128,10 +1131,11 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                                 PoolType::Shielded(ShieldedProtocol::Orchard),
                             )
                         } else {
-                            Recipient::InternalAccount(
-                                *output.account(),
-                                Note::Orchard(*output.note()),
-                            )
+                            Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                external_address: None,
+                                note: Note::Orchard(*output.note()),
+                            }
                         };
 
                         wallet::put_sent_output(
@@ -1145,7 +1149,7 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                             Some(output.memo()),
                         )?;
 
-                        if matches!(recipient, Recipient::InternalAccount(_, _)) {
+                        if matches!(recipient, Recipient::InternalAccount { .. }) {
                             wallet::orchard::put_received_note(wdb.conn.0, output, tx_ref, None)?;
                         }
                     }
@@ -1154,10 +1158,17 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
 
                         if let Some(account_id) = funding_account {
                             // Even if the recipient address is external, record the send as internal.
-                            let recipient = Recipient::InternalAccount(
-                                *output.account(),
-                                Note::Orchard(*output.note()),
-                            );
+                            let recipient = Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                // TODO: recover the actual UA, if possible
+                                external_address: Some(Address::Unified(
+                                    UnifiedAddress::from_receivers(
+                                    Some(output.note().recipient()),
+                                    None,
+                                    None,
+                                ).expect("UA has an Orchard receiver by construction."))),
+                                note: Note::Orchard(*output.note()),
+                            };
 
                             wallet::put_sent_output(
                                 wdb.conn.0,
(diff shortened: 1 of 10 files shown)


zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -319,6 +321,27 @@ pub(crate) fn get_orchard_nullifiers(
...
 }
 
+pub(crate) fn detect_spending_accounts<'a>(
+    conn: &Connection,
+    nfs: impl Iterator<Item = &'a Nullifier>,
+) -> Result<HashSet<AccountId>, rusqlite::Error> {
+    let mut account_q = conn.prepare_cached(
+        "SELECT rn.account_id
+        FROM orchard_received_notes rn
+        WHERE rn.nf IN :nf_ptr",
```
nuttycom, line 331: @AArnott one question about this: should an account be detected as funding the transaction even if it's an imported account? Or should this only apply to derived accounts? Same question for Sapling and transparent.

AArnott: My initial answer is "yes, of course", which makes me wonder if there's a downside that you see that I'm missing. Isn't scanning the same in only requiring the UFVK whether the account is derived or imported? I would expect downloaded or transmitted transactions to be recorded in the db exactly the same whether the account derivation information is in the db or not (as in the imported case).

AArnott:
I applied this change to my branch (combined with your tree panic fix), and early in sync, it failed with:

```
near ":nf_ptr": syntax error in SELECT rn.account_id
        FROM sapling_received_notes rn
        WHERE rn.nf IN :nf_ptr at offset 83
```

nuttycom:
> I applied this change to my branch (combined with your tree panic fix), and early in sync, it failed with:
> 
> ```
> near ":nf_ptr": syntax error in SELECT rn.account_id
>         FROM sapling_received_notes rn
>         WHERE rn.nf IN :nf_ptr at offset 83
> ```

Oops, fixed, thanks.

AArnott:
That's much better. Everything in the sqlite table looks correct now. There is one remaining error in the balance impact calculation for the send-to-self transaction, but I believe that is in my own code.

AArnott:
I just confirmed the remaining discrepancy was in my code and I have a fix for that too. Thanks for fixing this!

zcash_client_sqlite/src/wallet/orchard.rs
str4d: For imported UFVK accounts, we definitely should do this tracking, because it's just metadata between accounts for which we have complete information. I don't believe we will ever select imported UIVK accounts here, as we don't have nullifier-deriving keys for those accounts, and thus can never populate `rn.nf`.

We should separately ensure that we set the "to account" correctly on the send side for any source account that we would be fetching here (to ensure symmetry in the reported data), but that's likely going to require a migration to the `sent_notes` table (as we currently use the account column there as a proxy for "account-internal send", which would be invalid in this context), so that does not need to happen in this PR (or even in the next release).

zcash_client_sqlite/src/lib.rs
```diff
@@ -1052,8 +1052,16 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
     ) -> Result<(), Self::Error> {
         self.transactionally(|wdb| {
             let tx_ref = wallet::put_tx_data(wdb.conn.0, d_tx.tx(), None, None)?;
+            let funding_accounts = wallet::get_funding_accounts(wdb.conn.0, d_tx.tx())?;
+            let funding_account = funding_accounts.iter().next().copied();
+            if funding_accounts.len() > 1 {
+                warn!(
+                    "More than one wallet account detected as funding transaction {:?}, selecting {:?}",
```
str4d, line 1059: I believe that this behaviour will lead to inconsistent views for the non-selected accounts, in exactly the way that @AArnott originally raised. It won't occur _often_ because multi-account funded transactions should be rare (as the approach is generally frowned upon), but it could happen. I think it's okay to choose this behaviour for now however, because any change we'd make to enable full funding-account tracking would require a database migration, at which point we'd just run `get_funding_accounts` again inside the migration.

Separately, do we need to pair this change with a migration to change how we recorded past notes? AFAICT the change in this PR will only affect future notes in the wallet, and the views will remain invalid for past notes. The data migration in that case would be the same as I mention above (except selecting a single account in the migration), so I think it's fine to omit this change in this PR (as newly recovered-from-seed wallets will be fine with just this change) and then add a migration later. If no migration is added to this PR, open an issue to do so.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1083,18 +1091,26 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
-                            }
-                            None => {
-                                spending_account_id = Some(*output.account());
-                            }
-                        }
-
                         wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
+
+                        if let Some(account_id) = funding_account {
+                            // Even if the recipient address is external, record the send as internal.
```
str4d, line 1097: Blocking:

This is the exact misuse of the `sent_notes` account column that I mentioned earlier. The enum is named `WalletInternal`, but prior to now AFAICT this has always been used as "account-internal". Given that _any_ account in the wallet can be the funding account here, "wallet-internal" is ill-defined; the two accounts may have no common controller of spend authority (e.g. not be derived from the same seed). For example, it would make no sense for a note sent from a wallet account to ZecPages to be considered a "wallet-internal" transaction just because the ZecPages full viewing key is imported.

As an additional datapoint for this column currently meaning "account-internal", if we detect a note as going to an external address, and then later call `create_account` such that the external address is learned to actually be a wallet address, we do not update the `sent_notes` table to correct the information.

If we are going to make this change to how funding accounts are tracked, we need to do so in a way that ensures data consistency. That means either writing a migration to distinguish "account-internal" from "wallet-internal" in the `sent_notes` table, or writing a migration to redefine the `to_account_id` column as "amongst the accounts derived within or imported into this wallet database" and add code to rescan the _entire_ wallet to update relationships whenever a new account is created or imported.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1136,50 +1153,69 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
-                            }
-                            None => {
-                                spending_account_id = Some(*output.account());
-                            }
-                        }
-
                         wallet::orchard::put_received_note(wdb.conn.0, output, tx_ref, None)?;
+
+                        if let Some(account_id) = funding_account {
+                            // Even if the recipient address is external, record the send as internal.
```
str4d, line 1159: Ditto.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1136,50 +1153,69 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
-                        .any(|input| nf == input.nullifier())
-                ).map(|(account_id, _)| account_id);
-                #[cfg(not(feature = "orchard"))]
-                let orchard_from_account = None;
+                let funding_accounts = wallet::get_funding_accounts(wdb.conn.0, d_tx.tx())?;
+                let funding_account = funding_accounts.iter().next().copied();
+                if let Some(account_id) = funding_account {
+                    if funding_accounts.len() > 1 {
+                        warn!(
+                            "More than one wallet account detected as funding transaction {:?}, selecting {:?}",
```
str4d, line 1206: There's a slight difference in the new logic: before if a transaction spent Sapling funds from one account and Orchard funds from another, the correct data would be recorded. Now, one of the accounts will be used for both. This is probably fine for the same reason as above: a future migration could insert the other accounts after changing the tables.

str4d (changes requested):
Reviewed 9b4e9ea10d97594b59f5de355996cdb9cfe8ccdf. The `sent_notes` table migration is blocking, because the data inconsistency introduced in this PR will affect Zashi 1.0; the other comments would not because Zashi 1.0 will not support multiple accounts.

If the `sent_notes` table migration is not feasible on the Zashi 1.0 timeline, then for this PR to be merged as part of Zashi 1.0 the new `WalletInternal` parts must be gated on the wallet only containing a single account (which ensures that the cross-account data inconsistency cannot occur). This restriction could then be lifted in a subsequent post-Zashi-1.0 crate release when the necessary migrations are introduced.

daira (approved):
utACK; no additional comments beyond @str4d's. Please at least file tickets for remaining issues.

zcash_client_sqlite/src/lib.rs
nuttycom: The easy fix here is to restrict this check to verify that the funding account is the receiving account. When I refer to "external" here I'm simply meant the zip32 external scope.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -288,6 +293,11 @@ pub(crate) fn send_single_step_proposed_transfer<T: ShieldedPoolTester>() {
 
     let tx_history = st.get_tx_history().unwrap();
     assert_eq!(tx_history.len(), 2);
+
+    assert_matches!(
+        decrypt_and_store_transaction(&st.network(), st.wallet_mut(), &tx),
+        Ok(_)
+    );
```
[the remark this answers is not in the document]
nuttycom: It's here to ensure that the view query is valid. When I have more time I will verify the return values.

zcash_client_sqlite/src/lib.rs
nuttycom: Wait a minute though, this is just populating the "from_account" column of the sent_notes table correctly; the to_account column should already be getting the correct data. So I don't think there is a problem here at all.

nuttycom: The recipient value being constructed in this fashion ensures that the to_account field gets set correctly. You're correct though that reusing the "WalletInternal" variant in this fashion is a change in semantics. It may be that we just need an additional Recipient variant, that can carry both the address (for external addresses) and the recipient account ID.

zcash_client_sqlite/src/lib.rs
AArnott: To be clear, since I don't fully follow @str4d's comment, I don't really care about transactions that are funded from multiple accounts, as to my knowledge no wallet enabled users to do that, so it's unlikely to come up in account recovery scenarios. 
What I _do_ care about is that in a multi-account wallet, transactions are correctly attributed as having been funded by the _one_ account that applies.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1049,19 +1050,28 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
                             //TODO: Recover the UA, if possible.
                             Recipient::Sapling(output.note().recipient())
                         } else {
-                            Recipient::InternalAccount(
-                                *output.account(),
-                                Note::Sapling(output.note().clone()),
-                            )
+                            Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                external_address: None,
```
str4d, line 1072: How can we ensure this is set correctly for cross-account transfers instead of setting it to `None`?

zcash_client_sqlite/src/lib.rs
```diff
@@ -1075,23 +1085,32 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
-                            }
-                        }
-
                         wallet::sapling::put_received_note(wdb.conn.0, output, tx_ref, None)?;
+
+                        if let Some(account_id) = funding_account {
+                            let recipient = Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                // TODO: recover the actual UA, if possible
+                                external_address: Some(Address::Sapling(output.note().recipient())),
```
str4d, line 1099: How can we ensure we distinguish between account-internal and cross-account transfers, if we always set `external_address`?

zcash_client_sqlite/src/lib.rs
```diff
@@ -1106,15 +1125,17 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                                PoolType::Shielded(ShieldedProtocol::Orchard),
                             )
                         } else {
-                            Recipient::InternalAccount(
-                                *output.account(),
-                                Note::Orchard(*output.note()),
-                            )
+                            Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                external_address: None,
```
str4d, line 1136: Ditto: how do we ensure this is set correctly for cross-account transfers?

zcash_client_sqlite/src/lib.rs
```diff
@@ -1128,55 +1149,81 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                            // Even if the recipient address is external, record the send as internal.
+                            let recipient = Recipient::InternalAccount {
+                                receiving_account: *output.account(),
+                                // TODO: recover the actual UA, if possible
+                                external_address: Some(Address::Unified(
+                                    UnifiedAddress::from_receivers(
+                                    Some(output.note().recipient()),
+                                    None,
+                                    None,
+                                ).expect("UA has an Orchard receiver by construction."))),
```
str4d, line 1169: Ditto: how do we distinguish account-internal from cross-account?

str4d (commented):
Reviewed as of 151e6e526e52f96f2c686516daf8cc7cecf18c1a

zcash_client_sqlite/src/lib.rs
nuttycom: Same answer as above, this is the contract of `TransferType`.

zcash_client_sqlite/src/lib.rs
nuttycom: Same, this is down to `TransferType`

zcash_client_sqlite/src/lib.rs
str4d: What I mean is, don't we need a separate `TransferType` case to distinguish between wallet-internal (cross-account) and account-internal?

str4d: Oh, I guess that for correct cross-account transfers (i.e. correctly treated as separate accounts, so that each account's external FVK sees the correct view), they should show up here as `TransferType::Outgoing`. In that case, we should have `external_address` available already for cross-account transfers and can set it here?

str4d: Ah, and now going back into the PR diff, we're currently in the `TransferType::WalletInternal` branch. Okay, so the change I think is missing here is that the other side of this branch should be tracking the receiving account for cross-account transfers.

zcash_client_sqlite/src/lib.rs
str4d: Here I now agree that this is doing the correct thing.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1049,19 +1050,39 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                    TransferType::Outgoing => {
+                        //TODO: Recover the UA, if possible.
+                        let recipient = Recipient::Sapling(output.note().recipient());
+                        wallet::put_sent_output(
+                            wdb.conn.0,
+                            &wdb.params,
+                            *output.account(),
+                            tx_ref,
+                            output.index(),
+                            &recipient,
```
str4d, line 1074: Okay, the 3-way match now makes clear that there's an asymmetry currently between `TransferType::Outgoing` and `TransferType::Incoming` when it comes to cross-account transfers. Both will put an entry into `sent_notes`, but only `Incoming` fills in both the funding and receiving accounts; `Outgoing` just sets the funding account but not the receiving account.

In general I think we are "probably okay" here, because we scan with IVKs before OVKs, so in the case of a cross-account transfer we will see `Incoming` here instead of `Outgoing`. However, in the future case of actually supporting imported accounts, there will be a bug here depending on the order in which accounts are imported. `wallet::put_sent_output` overwrites the contents of the `to_account_id` column, so if the receiving account gets handled before the sending account, then the collected data will be erased.

This is not simple to recover from: we cannot fix the data inconsistency in a migration if we don't have the raw transaction (to re-decrypt), and we are not guaranteed to have that if the caller is not running enhancement. So I think there are two possible options here:
- Track the receiving account in `TransferType::Outgoing`, as I had previously suggested.
- Change `wallet::put_sent_output` to not overwrite known `to_account_id` values.

zcash_client_sqlite/src/lib.rs
nuttycom: Discussion from offline: the received_notes tables would then duplicate exactly information that's available from the sent_notes table, and such duplication would be a potential source of error. So this should be fine.

str4d (changes requested):
Reviewed 4ba7fbd977c05dd3269ca5f263ca65e3915a7d8c

zcash_client_sqlite/src/lib.rs
nuttycom: When we scan, we scan with _all_ of the available IVKs, so in the case of a cross-account transfer it will always be detected as `TransferType::Incoming` and will get handled by that branch, which will correctly find and set the source account. And the `Outgoing` case is already handled correctly.

str4d: There's nothing that guarantees that (from the perspective of an implementation of the `WalletWrite` trait), but I agree that's what the current code does in `zcash_client_backend` which is what callers are most likely using, as I said in my subsequent review. It's just brittle.

zcash_client_sqlite/src/lib.rs
nuttycom: I prefer the latter solution, because in general for outgoing transfers we don't know anything about the recipient (and it would be a pain to recover that in the send flow), whereas our usual upsert functionality is to not overwrite known data with nulls.

zcash_client_sqlite/src/lib.rs
nuttycom: Oh, also, I had already opened https://github.com/zcash/librustzcash/issues/1305 to track the remaining work here.

zcash_client_sqlite/src/lib.rs
nuttycom: Within the lifetime of an a wallet database, I can't see how account IDs could ever become unstable; if you swapped them around, you'd have to update all the foreign keys everywhere and this is no exception.

zcash_client_sqlite/src/testing/pool.rs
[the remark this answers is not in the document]
nuttycom: Opened #1315

str4d (approved):
utACK 1719b0186cd2accabdf6e8f8058614afe8fe7702. Please open issues for the remaining multi-account comments.

PR #1309, opened 2024-03-24
zcash_client_backend: Ensure checkpoint at the end of each block.

### LINKED ISSUE #1302: zcash_client_sqlite: Inconsistency between tree size & checkpoint position updates.

nuttycom:
This fixes an error wherein a note commitment in the last note commitment position in a block was not being correctly marked as a checkpoint.

This would occur when a block contained both Sapling and Orchard note commitments, but the final transaction in the block contained only either Sapling or Orchard note commitments, but not both.

Fixes #1302

### DIFF
--- a/zcash_client_backend/src/scanning.rs
+++ b/zcash_client_backend/src/scanning.rs
@@ -12,6 +12,7 @@ use sapling::{
 };
 use subtle::{ConditionallySelectable, ConstantTimeEq, CtOption};
 
+use tracing::{debug, trace};
 use zcash_keys::keys::UnifiedFullViewingKey;
 use zcash_note_encryption::{batch, BatchDomain, Domain, ShieldedOutput, COMPACT_NOTE_SIZE};
 use zcash_primitives::{
@@ -666,6 +667,11 @@ where
     ) -> Option<ScanError> {
         if let Some(prev) = prior_block_metadata {
             if block.height() != prev.block_height() + 1 {
+                debug!(
+                    "Block height discontinuity at {:?}, previous was {:?} ",
+                    block.height(),
+                    prev.block_height()
+                );
                 return Some(ScanError::BlockHeightDiscontinuity {
                     prev_height: prev.block_height(),
                     new_height: block.height(),
@@ -673,6 +679,7 @@ where
             }
 
             if block.prev_hash() != prev.block_hash() {
+                debug!("Block hash discontinuity at {:?}", block.height());
                 return Some(ScanError::PrevHashMismatch {
                     at_height: block.height(),
                 });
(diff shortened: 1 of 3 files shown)


zcash_client_backend/src/scanning.rs
```diff
@@ -736,6 +745,12 @@ where
             },
             Ok,
         )?;
+    let sapling_final_tree_size = sapling_commitment_tree_size
+        + block
+            .vtx
+            .iter()
+            .map(|tx| u32::try_from(tx.outputs.len()).unwrap())
+            .sum::<u32>();
```
str4d, line 753: Nit: if we're now going to compute this for every block (instead of just for blocks where we don't have prior block metadata), we should instead extract `sapling_output_count` from inside the `sapling_commitment_tree_size` definition, and use the result both there and here.

zcash_client_backend/src/scanning.rs
```diff
@@ -782,8 +797,14 @@ where
             },
             Ok,
         )?;
+    #[cfg(feature = "orchard")]
+    let orchard_final_tree_size = orchard_commitment_tree_size
+        + block
+            .vtx
+            .iter()
+            .map(|tx| u32::try_from(tx.actions.len()).unwrap())
+            .sum::<u32>();
```
str4d, line 806: Ditto: reuse `orchard_action_count`.

str4d (approved):
utACK 874d6b608b827c0cf2e628cfe949d150663bd3e7

PR #1310, opened 2024-03-25
zcash_protocol release version 0.1.1

nuttycom:


str4d (approved):
utACK. The feature flag removal is technically breaking, but it was only unstable flags so I'm fine with doing this in a point release.

PR #1312, opened 2024-03-25
zcash_client_sqlite: Disallow selection of dust notes.

nuttycom:
At present, we don't take builder padding rules into account in the determination of whether or not we can include spends of dust notes in grace actions. It's too complex to attempt a complete fix for Zashi 1.0, but we can have a tolerable workaround for launch by just never selecting dust-valued notes.

### DIFF
--- a/zcash_client_sqlite/src/wallet/common.rs
+++ b/zcash_client_sqlite/src/wallet/common.rs
@@ -158,6 +158,7 @@ where
                  INNER JOIN transactions
                     ON transactions.id_tx = {table_prefix}_received_notes.tx
                  WHERE {table_prefix}_received_notes.account_id = :account
+                 AND value >= 5000 -- FIXME #1016, allow selection of a dust inputs
                  AND accounts.ufvk IS NOT NULL
                  AND recipient_key_scope IS NOT NULL
                  AND nf IS NOT NULL
--- a/zcash_client_sqlite/src/wallet/orchard.rs
+++ b/zcash_client_sqlite/src/wallet/orchard.rs
@@ -596,6 +596,7 @@ pub(crate) mod tests {
     }
 
     #[test]
+    #[ignore] // FIXME: #1316 This requires support for dust outputs.
     fn zip317_spend() {
         testing::pool::zip317_spend::<OrchardPoolTester>()
     }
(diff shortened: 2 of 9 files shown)


zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -157,7 +157,8 @@ where
                     ON accounts.id = {table_prefix}_received_notes.account_id
                  INNER JOIN transactions
                     ON transactions.id_tx = {table_prefix}_received_notes.tx
-                 WHERE {table_prefix}_received_notes.account_id = :account
+                 WHERE value >= 5000 -- FIXME, allow seslection of a dust input
```
str4d, line 160:
```suggestion
                 WHERE value >= 5000 -- FIXME, allow selection of a dust input
```
Also small nit: if this were done in the first `AND` clause, it would have been a one-line change instead of a two-line change.

str4d (changes requested):
Reviewed d6f841d6b86166b5b7253132fd499490a20a0ae9. Concept ACK, but a test is failing.

str4d (dismissed):
utACK 0d8f5692df046da3d80290f244423b45dbd86941

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -597,6 +597,7 @@ pub(crate) mod tests {
 
     #[test]
     #[ignore] // FIXME: #1316 This requires support for dust outputs.
+    #[cfg(not(feature = "expensive-tests"))]
```
ebfull, line 600: Why are these `cfg(not(...))` instead of just `cfg(...)`?

nuttycom: So, it's annoying. CI uses the -- --ignored flag to run _just_ the ignored tests. So I want this test to _actually_ be ignored, so I made the slow tests CI workflow use the `expensive-tests` feature, and then I opt this out of the slow tester by adding this cfg attr.

ebfull (approved):

str4d (approved):
utACK bda72e36eb93713000bdbe107c87554dfc3d8632

daira (commented):
Post-hoc ACK

PR #1321, opened 2024-03-26
zcash_client_sqlite: Fix sent note recipient constraint

### LINKED ISSUE #1320: CHECK constraint failed: note_recipient
In the latest librustzcash (a9c4bc17), I'm getting this error very quickly after starting an account recovery:

> CHECK constraint failed: note_recipient

This is defined by:

https://github.com/zcash/librustzcash/blob/ad4bc84098b8a5edb4b6ee50cdf1f83f1d5ee449/zcash_client_sqlite/src/wallet/init/migrations/full_account_ids.rs#L331-L333

This seems like an odd condition to me in the first place, that a recipient address and a receiving account id could not be both specified.

nuttycom:
Fixes #1320

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3055,7 +3055,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_client_sqlite"
-version = "0.10.0"
+version = "0.10.1"
 dependencies = [
  "assert_matches",
  "bls12_381",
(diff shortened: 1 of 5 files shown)


daira (approved):
utACK

str4d (approved):
utACK

PR #1333, opened 2024-03-27
zcash_client_sqlite: Fix incorrect input selection filtering when sending to transparent.

nuttycom:
The "avoid pool crossing" conditions in note selection were erroneously not taking into account the need to pay transparent outputs.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -394,20 +394,25 @@ where
         // of funds selected is strictly increasing. The loop will either return a successful
         // result or the wallet will eventually run out of funds to select.
         loop {
-            #[cfg(feature = "orchard")]
-            let (sapling_input_total, orchard_input_total) = (
-                shielded_inputs.sapling_value()?,
-                shielded_inputs.orchard_value()?,
-            );
-
             #[cfg(not(feature = "orchard"))]
-            let orchard_input_total = NonNegativeAmount::ZERO;
-
-            let use_sapling =
-                !(sapling_outputs.is_empty() && orchard_input_total >= amount_required);
+            let use_sapling = true;
             #[cfg(feature = "orchard")]
-            let use_orchard =
-                !(orchard_outputs.is_empty() && sapling_input_total >= amount_required);
+            let (use_sapling, use_orchard) = {
+                let (sapling_input_total, orchard_input_total) = (
+                    shielded_inputs.sapling_value()?,
+                    shielded_inputs.orchard_value()?,
+                );
+
+                // Use Sapling inputs if there are no Orchard outputs or there are not sufficient
+                // Orchard outputs to cover the amount required.
+                let use_sapling =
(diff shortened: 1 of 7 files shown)


str4d (approved):
utACK 1cb33c8b06411f13977da0002e30149efeb5edbf

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -394,20 +394,25 @@ where
...
+                    shielded_inputs.orchard_value()?,
+                );
+
+                // Use Sapling inputs if there are no Orchard outputs or there are not sufficient
+                // Orchard outputs to cover the amount required.
+                let use_sapling =
+                    orchard_outputs.is_empty() || amount_required > orchard_input_total;
+                // Use Orchard inputs if there are insufficient Sapling funds to cover the amount
+                // reqiuired.
+                let use_orchard = !use_sapling || amount_required > sapling_input_total;
```
daira, line 412: I know this is out-of-scope for this release, but just to document it as an issue: this code might set either `use_sapling` or `use_orchard` to `false` because it decides that the amount required is satisfiable from the other pool, but actually it is not because of fees (even though the amount could potentially have been satisfied from both pills). The correct way to handle this will be to subtract the marginal fee from each note value when calculating balance, as @nuttycom suggested some time ago.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1634,6 +1634,97 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
...
+    let proposal0 = st
+        .propose_transfer(
+            account.account_id(),
+            &input_selector,
+            p0_to_p1,
+            NonZeroU32::new(1).unwrap(),
+        )
+        .unwrap();
+
+    let _min_target_height = proposal0.min_target_height();
```
daira, line 1690: Was this intended to be unused?

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1634,6 +1634,97 @@ pub(crate) fn fully_funded_fully_private<P0: ShieldedPoolTester, P1: ShieldedPoo
...
+    st.scan_cached_blocks(h, 1);
+
+    assert_eq!(
+        st.get_total_balance(account.account_id()),
+        (initial_balance - transfer_amount - expected_fee).unwrap()
+    );
+    assert_eq!(
+        st.get_spendable_balance(account.account_id(), 1),
+        (initial_balance - transfer_amount - expected_fee).unwrap()
+    );
```
daira, line 1725: Do we want to check the balances of each pool here?

daira (approved):
utACK

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: This isn't a current bug, because this whole thing is in a loop that continues selecting value until it either has fees covered or runs out of notes.

PR #1334, opened 2024-03-28
zcash_client_sqlite: Add a migration to ensure that default addresses…

nuttycom:
… have an Orchard receiver.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
@@ -305,7 +305,8 @@ mod tests {
         let network = Network::TestNetwork;
         let data_file = NamedTempFile::new().unwrap();
         let mut db_data = WalletDb::for_path(data_file.path(), network).unwrap();
-        init_wallet_db_internal(&mut db_data, None, &[addresses_table::MIGRATION_ID]).unwrap();
+        init_wallet_db_internal(&mut db_data, None, &[addresses_table::MIGRATION_ID], false)
+            .unwrap();
         let usk = UnifiedSpendingKey::from_seed(&network, &[0u8; 32][..], AccountId::ZERO).unwrap();
         let ufvk = usk.to_unified_full_viewing_key();
 
@@ -336,7 +337,7 @@ mod tests {
             VALUES (0, 4, 0, '', 7, '', 'c', true, X'63');",
         ).unwrap();
 
-        init_wallet_db_internal(&mut db_data, None, &[super::MIGRATION_ID]).unwrap();
+        init_wallet_db_internal(&mut db_data, None, &[super::MIGRATION_ID], false).unwrap();
 
         let mut q = db_data
             .conn
@@ -407,6 +408,7 @@ mod tests {
             &mut db_data,
             None,
             &[utxos_table::MIGRATION_ID, ufvk_support::MIGRATION_ID],
+            false,
         )
         .unwrap();
 
@@ -479,7 +481,7 @@ mod tests {
             )
             .unwrap();
 
-        init_wallet_db_internal(&mut db_data, None, &[super::MIGRATION_ID]).unwrap();
+        init_wallet_db_internal(&mut db_data, None, &[super::MIGRATION_ID], false).unwrap();
 
         let fee = db_data
             .conn
--- a/zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
@@ -1,0 +1,200 @@
+//! This migration ensures that an Orchard receiver exists in the wallet's default Unified address.
+use std::collections::HashSet;
+
+use rusqlite::named_params;
+use schemer_rusqlite::RusqliteMigration;
+use uuid::Uuid;
+
+use zcash_client_backend::keys::{
+    UnifiedAddressRequest, UnifiedFullViewingKey, UnifiedIncomingViewingKey,
+};
+use zcash_primitives::consensus;
+
+use super::orchard_received_notes;
+use crate::{wallet::init::WalletMigrationError, UA_ORCHARD, UA_TRANSPARENT};
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x604349c7_5ce5_4768_bea6_12d106ccda93);
+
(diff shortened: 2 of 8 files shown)


zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
```diff
@@ -0,0 +1,200 @@
...
+            } else {
+                let uivk_str: String = row.get("uivk")?;
+                UnifiedIncomingViewingKey::decode(&self.params, &uivk_str[..]).map_err(|_| {
+                    WalletMigrationError::CorruptedData("Unable to decode UIVK".to_string())
+                })?
+            };
+
+            let (default_addr, diversifier_index) = uivk.default_address(
+                UnifiedAddressRequest::unsafe_new(UA_ORCHARD, true, UA_TRANSPARENT),
+            )?;
```
str4d, line 74: What is the rationale for only adding Orchard to the default address?

- `WalletWrite::get_next_available_address` was introduced in #632, at which point the behaviour was "generate a UA with all of the wallet's receiver types".
  - Released in 0.6.0 on 2022-11-12.
- #829 added receiver type selection to this API. After this PR, the caller would have needed to explicitly specify which receiver types they wanted.
  - Released in 0.11.0 on 2024-03-01, less than one month ago.

Given the above, I think the vast majority of addresses (if not _all_ of them) were generated assuming implicit addressing (i.e. the default address is not special in that regard), and we should add Orchard receivers to all of them.

We might also want to distinguish addresses generated before 0.11.0 from addresses generated after (i.e. distinguish them _here_), so we can keep track of which ones should continue to be bumped. Or maybe we just distinguish which addresses were generated with a _subset_ of the wallet's supported receivers, so that callers who just want "all of them" get them bumped as well in future, and only UAs created with a specific subset are left alone by future migrations.

zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
```diff
@@ -0,0 +1,200 @@
...
+               WHERE account_id = :account_id
+               AND diversifier_index_be = :j
+            "#,
+        )?;
+
+        let mut accounts = get_accounts.query([])?;
+        while let Some(row) = accounts.next()? {
+            let account_id = row.get::<_, u32>("id")?;
+            let ufvk_str: Option<String> = row.get("ufvk")?;
+            let uivk = if let Some(ufvk_str) = ufvk_str {
```
str4d, line 59: Non-blocking: I'm a little confused by this, as the prior migrations should have ensured that all accounts with a UFVK have a corresponding UIVK, so querying `uivk` should be sufficient, but this is a migration so I don't particularly mind.

zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
```diff
@@ -0,0 +1,200 @@
...
+                    .to_unified_incoming_viewing_key()
+            } else {
+                let uivk_str: String = row.get("uivk")?;
+                UnifiedIncomingViewingKey::decode(&self.params, &uivk_str[..]).map_err(|_| {
+                    WalletMigrationError::CorruptedData("Unable to decode UIVK".to_string())
+                })?
+            };
+
+            let (default_addr, diversifier_index) = uivk.default_address(
+                UnifiedAddressRequest::unsafe_new(UA_ORCHARD, true, UA_TRANSPARENT),
```
str4d, line 73: A fun side-effect of the Sapling behaviour is that if it were Sapling we were introducing here, `diversifier_index` might be different than the previous entry, and we wouldn't be replacing it. Irrelevant because we're not doing that, and we shouldn't ever introduce a new receiver kind that has this behaviour, but something to keep in mind for these kinds of migrations.

str4d (changes requested):
Reviewed c34b68204d39ee12704964f64c1c0a53afdd5144. Question about which addresses to update must be resolved.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
```diff
@@ -305,7 +305,8 @@ mod tests {
         let network = Network::TestNetwork;
         let data_file = NamedTempFile::new().unwrap();
         let mut db_data = WalletDb::for_path(data_file.path(), network).unwrap();
-        init_wallet_db_internal(&mut db_data, None, &[addresses_table::MIGRATION_ID]).unwrap();
+        init_wallet_db_internal(&mut db_data, None, &[addresses_table::MIGRATION_ID], false)
```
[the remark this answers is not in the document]
nuttycom: They all pass `None` for the seed.

zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
nuttycom: Ah, yeah, I guess that's so.

zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
nuttycom: Hm, I weakly disagree; my expectation is that if people are generating multiple addresses, that they're also not reusing addresses and so we shouldn't retroactively change things; the new addresses being generated with Orchard receivers going forward would fix the problem.

str4d: Hmmm, okay I'll buy this.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_views.rs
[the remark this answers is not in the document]
str4d: Ah yep 😄

str4d (approved):
utACK c34b68204d39ee12704964f64c1c0a53afdd5144

zcash_client_sqlite/src/wallet/init/migrations/ensure_orchard_ua_receiver.rs
daira: If this ends up removing a Sapling receiver from the UA it's migrating from, isn't that bad?

daira (commented):
Post-hoc ACK

PR #1341, opened 2024-04-05
zcash_client_sqlite: Fix broken `account_birthday` method.

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3201,4 +3203,15 @@ mod tests {
...
+
+    #[test]
+    fn test_account_birthday() {
+        let st = TestBuilder::new()
+            .with_block_cache()
+            .with_account_from_sapling_activation(BlockHash([0; 32]))
+            .build();
+
+        let account_id = st.test_account().unwrap().account_id();
+        assert_matches!(account_birthday(&st.wallet().conn, account_id), Ok(_))
```
str4d, line 3215: Check that it actually returns the correct birthday.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1591,7 +1591,7 @@ pub(crate) fn account_birthday(
     conn.query_row(
         "SELECT birthday_height
          FROM accounts
-         WHERE account = :account_id",
+         WHERE id = :account_id",
```
str4d, line 1594: The account ID refactor, and the large-scale column name overhauling that was performed as a side-effect, has caused multiple of these bugs. Some of them we noticed and fixed in a hurry before 1.0, but like this one here, I do not doubt there are more.

What can we do to systematically check and fix all the queries, and ensure we don't have this problem in future? Defining a bunch of column name constants becomes unwieldy when it forces us to use string formatting, but if we start building wrappers to make that easier, we'd basically be building a custom DAO.

str4d (changes requested):

str4d:
Also, the fix needs to be documented in the changelog, as the bug made it into a published crate release.

zcash_client_sqlite/src/wallet.rs
nuttycom: Basically this just points to the fact that we actually need tests for all of the SQL queries; given that, such bugs become shallow.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/8b390b988cd18632561e162c70fbfa09b48ddf74..f7c29f0) to address comments from code review.

str4d (approved):
utACK

zcash_client_sqlite/src/wallet.rs
daira: Is there some automated tool to lint / statically analyse SQL queries given a dump of the schema? It seems feasible.

daira (commented):
Post-hoc ACK

PR #1357, opened 2024-04-18
Refactor `equihash` code

str4d:
Move-only; my main motivation is moving the minimal-encoding functions into a separate submodule, so that the inverse functions introduced in #1083 can also be placed there (with round-trip tests and proptests).

### DIFF
--- a/components/equihash/src/params.rs
+++ b/components/equihash/src/params.rs
@@ -1,0 +1,37 @@
+#[derive(Clone, Copy)]
+pub(crate) struct Params {
+    pub(crate) n: u32,
+    pub(crate) k: u32,
+}
+
+impl Params {
+    /// Returns `None` if the parameters are invalid.
+    pub(crate) fn new(n: u32, k: u32) -> Option<Self> {
+        // We place the following requirements on the parameters:
+        // - n is a multiple of 8, so the hash output has an exact byte length.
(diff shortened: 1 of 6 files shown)


components/equihash/src/params.rs
```diff
@@ -0,0 +1,37 @@
+#[derive(Clone, Copy)]
+pub(crate) struct Params {
+    pub(crate) n: u32,
+    pub(crate) k: u32,
+}
+
+impl Params {
+    /// Returns `None` if the parameters are invalid.
```
arya2, line 8: Optional/Nitpick: More verbose docs, maybe with a reference to where and/or how the parameters are used? 

```suggestion
    /// Accepts `n` and `k` values to use as Equihash parameters.
    ///
    /// Returns a new instance of [`Params`] if the provided values are valid, or
    /// `None` if the parameters are invalid.
```

arya2 (approved):
Looks good!

components/equihash/src/params.rs
str4d: The doc change here was just to clarify the change from `Error` (there is only one "error" case here, and `Error` is about verification errors). I've opened #1358 for the general case of documenting the crate-private methods.

nuttycom (approved):

daira (commented):
Post-hoc ACK

PR #1359, opened 2024-04-23
Set up Cargo audit infrastructure

str4d:
This starts the process of removing our dependency on `zcash/zcash` for our `cargo-vet` audits. We import audits from there and other upstreams to bootstrap audits here; once this PR merges, we'll include this repo into our aggregated audit set, and then we can use audits here (for the MSRV-compatible dependencies pinned in this repo's `Cargo.lock`) to augment the audits done in our end binary repos (`zcashd` and the mobile SDKs, which use as close to stable Rust as we can).

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -7,6 +7,9 @@ version = "0.9"
 [imports.google]
 url = "https://raw.githubusercontent.com/google/supply-chain/main/audits.toml"
 
+[imports.isrg]
+url = "https://raw.githubusercontent.com/divviup/libprio-rs/main/supply-chain/audits.toml"
+
 [imports.mozilla]
 url = "https://raw.githubusercontent.com/mozilla/supply-chain/main/audits.toml"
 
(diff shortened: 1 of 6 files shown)


nuttycom (approved):
ACK. I did not carefully review the audit claims from other organizations.

str4d:
The other orgs are all ones we already depend on for audits in `zcash/zcash` except for Fermyon, and their repo only introduces one additional audit (for `oorandom 11.1.3`).

supply-chain/config.toml
```diff
@@ -0,0 +1,881 @@
...
+
+[[exemptions.wasm-bindgen-backend]]
+version = "0.2.88"
+criteria = "safe-to-deploy"
+
+[[exemptions.wasm-bindgen-macro]]
+version = "0.2.88"
+criteria = "safe-to-deploy"
+
+[[exemptions.wasm-bindgen-macro-support]]
```
daira, line 831: I actually reviewed this completely: https://github.com/zcash/zcash/pull/6864/files#diff-03405b160ee10541a38d19f2c8d1947bbc63d55a17ff1c73843697e84b33475aL622-L624

daira (commented):
Post-hoc ACK

PR #1376, opened 2024-04-26
CI: Add explicit token to GitHub CLI usage

str4d:
Fixes https://github.com/zcash/librustzcash/actions/runs/8853688391/job/24315095768.


daira (approved):
utACK

PR #1377, opened 2024-04-27
Update bip0039 for Arti support in Zebra

emersonian:
Upgrading bip0039 to v0.12 in zcash_primitives resolves Zebra's ability to build with Arti (Tor) support.

Patch versions will need to be released of the following crates to simplify integration with Zebra:

- zcash_primitives: v0.13.1
- zcash_client_backend: v0.10.1 (depending upon zcash_primitives v0.13.1)
- zcash_proofs: v0.13.1 (depending upon zcash_primitives v0.13.1)
- zcash_script: v0.1.16 (depending upon zcash_primitives v0.13.1)

[Demonstration of a working build](https://github.com/emersonian/zebra/tree/arti-deps)

See ZcashFoundation/zebra#8328

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,9 +247,9 @@ dependencies = [
 
 [[package]]
 name = "bip0039"
-version = "0.10.1"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef0f0152ec5cf17f49a5866afaa3439816207fd4f0a224c0211ffaf5e278426"
+checksum = "568b6890865156d9043af490d4c4081c385dd68ea10acd6ca15733d511e6b51c"
 dependencies = [
  "hmac",
  "pbkdf2",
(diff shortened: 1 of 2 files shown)


str4d (changes requested):
NACK for the same reason as when this was proposed in #1261.

We attempted to upgrade from `bip0039 0.10` in #808, but had to downgrade in #812 because the change in `bip0039 0.11` to use generics resulted in me spending several hours fighting and losing to the C++ template system in `zcashd`.

While `zcashd` depends on `bip0039 0.10` via `zcash_primitives::zip339`, we will not be upgrading this dependency. If someone else wants to fight the C++ template system to enable this, let me know!

str4d:
Also, as a side-note:

> Patch versions will need to be released

As documented in the `Cargo.toml` that you edited, `bip0039` is exposed in our API, and thus changing its left-most non-negative version integer is a SemVer-breaking change in Rust.

nuttycom:
@str4d `zip339` is simply a reexport of the `bip0039` crate, and nothing in `librustzcash` depends upon it. Also, given the extraction of `zcash_keys`, I don't think that it's appropriate for this to live in `zcash_primitives` any longer anyway. I recommend that we simply remove the `zip339` module entirely, which then decouples the use of `bip0039` in `zcashd` from any potential use in `zebra`.

nuttycom:
@emersonian instead of this, please just revert https://github.com/zcash/librustzcash/pull/424 instead. This will require a major version bump to `zcash_primitives` and downstream crates, so make sure to note the `zip339` module as removed in the CHANGELOG so that we remember that a major version bump is required.

daira:
Superceded by #1391.

PR #1388, opened 2024-05-11
zcash_primitives: Refactor `Transaction` to permit omitting protocol-specific bundles

str4d:
Draft because there are still some things that need doing:
- [x] I want some feedback on the overall approach to check the direction makes sense.
- [ ] I haven't documented much yet (spent my time getting the refactor structured correctly so the commits make sense).
- [ ] I haven't checked all the various feature flag combinations yet.
- [x] The `Authorization`-ish traits are still tied to the protocol-specific crates. I think I need to introduce intermediate traits with blanket impls.
- [x] Factor out Sprout for completeness (even though we don't depend on anything for Sprout that we don't inherently need for basic Zcash support).
- [x] Factor out `ZFuture`.
- [ ] Needs changelog updates.

### DIFF
--- a/devtools/src/bin/inspect/transaction.rs
+++ b/devtools/src/bin/inspect/transaction.rs
@@ -19,7 +19,9 @@ use zcash_primitives::{
     legacy::{keys::pubkey_to_address, Script, TransparentAddress},
     memo::{Memo, MemoBytes},
     transaction::{
-        components::{amount::NonNegativeAmount, sapling as sapling_serialization, transparent},
+        components::{
+            amount::NonNegativeAmount, sapling as sapling_serialization, transparent, AllBundles,
+        },
         sighash::{signature_hash, SignableInput, TransparentAuthorizingContext},
         txid::TxIdDigester,
         Authorization, Transaction, TransactionData, TxId, TxVersion,
@@ -207,7 +209,7 @@ pub(crate) fn inspect(
         tx.write(&mut buf).unwrap();
         let tx = Transaction::read(&buf[..], tx.consensus_branch_id()).unwrap();
 
-        let tx: TransactionData<PrecomputedAuth> = tx.into_data().map_authorization(
+        let tx: TransactionData<AllBundles<PrecomputedAuth>> = tx.into_data().map_authorization(
             f_transparent,
             (),
             (),
--- a/zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
@@ -149,7 +149,10 @@ mod tests {
     use tempfile::NamedTempFile;
     use zcash_primitives::{
         legacy::{Script, TransparentAddress},
-        transaction::{components::transparent, Authorized, TransactionData, TxVersion},
+        transaction::{
+            components::{transparent, AllBundles},
+            Authorized, TransactionData, TxVersion,
+        },
     };
     use zcash_protocol::{
         consensus::{BranchId, Network},
@@ -185,7 +188,7 @@ mod tests {
         .unwrap();
 
         // Add transactions to the wallet that exercise the data migration.
-        let add_tx_to_wallet = |tx: TransactionData<Authorized>| {
+        let add_tx_to_wallet = |tx: TransactionData<AllBundles<Authorized>>| {
             let tx = tx.freeze().unwrap();
             let txid = tx.txid();
             let mut raw_tx = vec![];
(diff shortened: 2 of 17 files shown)


daira (commented):
LGTM so far. I was a little bit alarmed about the name `UnauthorizedTransactionDataWithSaplingProofs` (i.e. why so unorthogonal?), but then I saw the motivation and it's fine.

daira:
> I haven't done anything with `ZFuture`, and am not quite sure what to do there.

Can we just delete TZE support? I don't care about them and I very much doubt they are ever going to be deployed. We should ask on #libraries in the R&D Discord obviously. Deleting them should be easy given that they're feature flagged.

str4d:
> I was a little bit alarmed about the name `UnauthorizedTransactionDataWithSaplingProofs` (i.e. why so unorthogonal?), but then I saw the motivation and it's fine.

My hope is that this type goes away once I land the refactor that follows after this one (to restructure the `zcash_primitives` transaction builder to match Sapling and Orchard, by separating the building of the unauthorized bundle and its authorization). At that point, this type will be a well-defined intermediate state, instead of an ill-defined internal detail of `Builder::build` that requires type-pinning to help the compiler.

str4d:
In 1322447f34c6c508630fbeab87991c960b1671c1 I've added Yet Another Intermediate Trait `Bundles`, which I think solves the "conditional compilation of bundles" that is necessary for merging code that is not currently part of consensus (as well as making the enabling of these bundles in consensus less of a breaking change to downstream code). If this approach looks good, I'll rebase the PR to merge the `Bundles` changes back into the earlier commits (simplifying them).

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/e3a29b86cd473324b3a8bd480e6640b6fac03d84..d185e6fd592a8541989a018ddd8772dff1099fad) to merge the commit introducing the `Bundles` trait back into the rest of the commits, simplifying them.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/d185e6fd592a8541989a018ddd8772dff1099fad..65df348cf11b83815134d10194f3055de7212ed0) to address some initial review comments from a pairing with @nuttycom.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/65df348cf11b83815134d10194f3055de7212ed0..6da0ba6b072273333daaacf3fdfd685b9425dae9) to move Sprout onto the `Bundles` trait. It has no authorization traits because we only support already-authorized Sprout bundles (due to not supporting Sprout in transaction building).

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/6da0ba6b072273333daaacf3fdfd685b9425dae9..43d2820f19bd330d1405b31b25633cc2e0895d87) to move TZEs onto the `Bundles` trait, as a proof-of-concept for other conditionally-compiled transaction format changes that we want to merge into the codebase before inclusion in consensus (e.g. ZSAs).

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/443855e9e9821e43dc9d937cc1516769d73fc675..47189947d2b57a8ca72e2beaf0d4c104c45adc19) to adjust how the new traits work in order to eliminate the bounds on the protocol-specific authorization traits.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/47189947d2b57a8ca72e2beaf0d4c104c45adc19..7477ff1604afded995f66a5709892a77885d822d) with some further decoupling adjustments and trait renames. It also removes the `NoTransparent` etc. changes (and the `TransactionWith` refactor) as not currently necessary, because the dependency removal will be done by altering the `Transparent<A>` etc. types (in a way that is technically a breaking change, but in a way that should only be observable as purely additive in our API).

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/7477ff1604afded995f66a5709892a77885d822d..16988bea5d44e5752cc5c21e93ac4e29a020f048) to fix a small merge conflict.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/16988bea5d44e5752cc5c21e93ac4e29a020f048..822cd3677362e555660889cf32d6031a5a1bb52d) to fix merge conflicts.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/822cd3677362e555660889cf32d6031a5a1bb52d..2a33c0ee7b87238389deb9c389c05f97d03d1411) to bring in recent changes.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/2a33c0ee7b87238389deb9c389c05f97d03d1411..52bb40d7209a02adcfe06ed13a74d25da8382c71) to fix formatting.

PR #1391, opened 2024-05-14
Remove zcash_primitives::zip339 module

daira:
Supercedes #1377. This helps with Arti support in Zebra. See https://github.com/zcash/librustzcash/pull/1377#issuecomment-2098664351 for discussion.

### DIFF
--- a/zcash_primitives/Cargo.toml
+++ b/zcash_primitives/Cargo.toml
@@ -63,9 +63,6 @@ hdwallet = { workspace = true, optional = true }
 #   - `SecretKey` and `PublicKey` types exposed
 secp256k1 = { workspace = true, optional = true }
 
-# - ZIP 339
-bip0039 = { version = "0.10", features = ["std", "all-languages"] }
-
 # Dependencies used internally:
 # (Breaking upgrades to these are usually backwards-compatible, but check MSRVs.)
 # - Documentation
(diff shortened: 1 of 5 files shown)


zcash_primitives/Cargo.toml
```diff
@@ -63,9 +63,6 @@ hdwallet = { workspace = true, optional = true }
 #   - `SecretKey` and `PublicKey` types exposed
 secp256k1 = { workspace = true, optional = true }
 
-# - ZIP 339
-bip0039 = { version = "0.10", features = ["std", "all-languages"] }
```
str4d, line 67: This removal results in changes to `Cargo.lock` that were not included in the commit. Please include those changes in the commit removing this dependency.

str4d (changes requested):

zcash_primitives/Cargo.toml
daira: Are you sure it should be in the same commit? That commit is a pure reversion of d1443e5049c32615155eaed0e36915dbe5222f6b.

str4d: Given that it is a pure reversion, fine to be in a follow-on commit.

str4d (approved):
utACK 031b416b75bb7e654504b4dbdeeec3262dfc8e10

AArnott:
This was a breaking change for me, and IMO deserves a mention in the CHANGELOG.

PR #1392, opened 2024-05-14
Dependency updates

daira:
Update dependencies and audits, and bump MSRV to 1.66.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,9 +29,9 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
  "getrandom",
@@ -53,18 +53,18 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.16"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
+checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "anes"
@@ -74,9 +74,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anyhow"
-version = "1.0.75"
+version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
+checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
 
 [[package]]
 name = "arrayref"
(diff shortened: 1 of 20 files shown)


daira:
Audit updates are not complete yet.

Cargo.lock
```diff
@@ -1536,25 +1507,25 @@ dependencies = [
...
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c42a9226546d68acdd9c0a280d17ce19bfe27a46bf68784e4066115788d008e"
+checksum = "1e401f977ab385c9e4e3ab30627d6f26d00e2c73eef317493c4ec6d468726cf8"
 dependencies = [
  "cfg-if",
  "libc",
  "redox_syscall",
  "smallvec",
- "windows-targets",
+ "windows-targets 0.52.5",
```
daira, line 1528: The duplication between 0.52.0 and 0.52.5 could potentially be avoided I guess. Not sure why this is requiring 0.52.5.

Cargo.toml
```diff
@@ -96,8 +96,8 @@ rayon = "1.5"
 
 # Protobuf and gRPC
 prost = "0.12"
-tonic = { version = "0.10", default-features = false }
-tonic-build = { version = "0.10", default-features = false }
+tonic = { version = "0.11", default-features = false }
+tonic-build = { version = "0.11", default-features = false }
```
str4d, line 100: [`tonic 0.11` bumps MSRV to 1.70](https://github.com/hyperium/tonic/blob/master/CHANGELOG.md#0110-2024-02-08). That also happens to be the next MSRV I wanted to target for `zcash_primitives` and its downstream crates (I'm more cautious about bumping MSRV for crates like `pasta_curves`, but for the crates in this workspace 1.70 should be fine). We were blocked from doing so by `zcashd` being stuck on 1.69, but that was finally fixed last month in https://github.com/zcash/zcash/pull/6879.

I note that in the first commit you avoided certain other dependency updates that would require 1.70; we can upgrade those now. Given that `tonic 0.11` only forces the case for `zcash_client_backend` with the relevant feature flag enabled, I'm fine with merging this PR as-is, and then landing a follow-up PR that uniformly bumps MSRV to 1.70.

components/equihash/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- MSRV is now 1.66.0.
```
str4d, line 10: Most of the `components/` crates use their own MSRVs as they are much more stable and rarely updated (note they do not use `rust-version.workspace = true`, so the top-level workspace MSRV change did not alter them). Revert these changes.
```suggestion
```

components/f4jumble/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- MSRV is now 1.66.0.
```
str4d, line 10:
```suggestion
```

components/zcash_address/CHANGELOG.md
```diff
@@ -6,14 +6,16 @@ and this library adheres to Rust's notion of
...
 ### Added
 - `zcash_address::ZcashAddress::{can_receive_memo, can_receive_as, matches_receiver}`
 - `zcash_address::unified::Address::{can_receive_memo, has_receiver_of_type, contains_receiver}`
 - Module `zcash_address::testing` under the `test-dependencies` feature.
 - Module `zcash_address::unified::address::testing` under the 
   `test-dependencies` feature.
 
+### Changed
+- MSRV is now 1.66.0.
+
```
str4d, line 18:
```suggestion
```

components/zcash_encoding/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- MSRV is now 1.66.0.
```
str4d, line 10:
```suggestion
```

components/zcash_protocol/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this library adheres to Rust's notion of
 ### Added
 - `zcash_protocol::PoolType::{TRANSPARENT, SAPLING, ORCHARD}`
 
+### Changed
+- MSRV is now 1.66.0.
```
str4d, line 13: This component crate is an exception, as it was recently extracted from `zcash_primitives` and thus shares a previous MSRV (and is less of a component crate and more of a crate we will regularly be updating, so a synced MSRV makes sense).

components/zip321/CHANGELOG.md
```diff
@@ -32,3 +32,4 @@ The entries below are relative to the `zcash_client_backend` crate as of
   - `zip321::parse::{lead_addr, zcashparam}`
 - `zip321::Param::Memo` now boxes its argument.
 - `zip321::Param::Addr` now wraps a `zcash_address::ZcashAddress`
+- MSRV is now 1.66.0.
```
str4d, line 35: This crate was also recently extracted from another crate with a higher MSRV than other component crates, and currently is synced with the workspace MSRV. It doesn't necessarily need to be though as I expect it to be a bit more stable, but fine to keep it synced for now.

zcash_client_backend/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this library adheres to Rust's notion of
   wrapping protocol-address types, the `Recipient` type now wraps a
   `zcash_address::ZcashAddress`. This simplifies the process of tracking the
   original address to which value was sent.
+- MSRV is now 1.66.0.
```
str4d, line 32: I don't actually know if `tonic-build` also had its MSRV bumped (they are developed in the same workspace, but neither have explicit `rust-version`s), but given that CI is passing I think that means it currently isn't using post-1.66 features. So the only thing we should note here is that the `lightwalletd-tonic` feature has MSRV 1.70.0 (until we change workspace MSRV to 1.70 in a subsequent PR):
```suggestion
- MSRV is now 1.66.0 (except for the `lightwalletd-tonic` feature flag which
  requires 1.70.0).
```

supply-chain/audits.toml
```diff
@@ -7,7 +7,354 @@ description = "The cryptographic code in this crate has been reviewed for correc
...
+
+[[audits.rustls]]
+who = "Daira-Emma Hopwood <daira@jacaranda.org>"
+criteria = "safe-to-deploy"
+delta = "0.21.8 -> 0.21.12"
+notes = """
+A comment in get_sni_extension asks whether the behaviour of parsing an IPv4 or IPv6 address
+in a host_name field of a server_name extension, but then ignoring the extension (because
+'Literal IPv4 and IPv6 addresses are not permitted in \"HostName\"'), as the server, is
+compliant with RFC 6066. As an original author of RFC 3546 which has very similar wording,
```
str4d, line 204: Neat!

str4d (approved):
utACK after changelog fixes.

str4d (approved):
Re-utACK 851e20347f0ea304f3083830ad65bc3fa389365b

daira:
Post-hoc ACK 851e20347f0ea304f3083830ad65bc3fa389365b.

PR #1396, opened 2024-05-20
`zcash_client_sqlite` Fix error related to commitment tree frontier insertions.

### LINKED ISSUE #1398: Error in note commitment tree maintenance at the bounds of the chain tip shard.
A Zashi user encountered the following error:

> Error while sending funds: An error occurred in querying or updating a note commitment tree: Unable to compute root; missing values for nodes `[...]` 

where the node in question corresponds to the last level-1 subtree of a `shardtree` shard. 

What happens is this:

We insert the frontier at the start of the block range being inserted. This gives us a `CHECKPOINT` leaf. Then, we insert the subtree fragment covering the range for notes in the subsequent block, which crosses the shard boundary, and then prune checkpoints.

Because the frontier was inserted for a not-yet-scanned block, we don't actually know (at the application level) whether the sibling nodes to the frontier should be pruned; we need to scan the block corresponding to the frontier to determine this. However, in `shardtree`, we don't have any distinction between leaves that were inserted as part of a frontier and leaves that were inserted as part of a scan. And so, `shardtree` is completely willing to prune nodes when we remove the `CHECKPOINT` - nodes that are later discovered to be necessary ommers when we discover our wallet's note at the tip of the blindly-inserted frontier.

The solution is to make the distinction between "known prunable" and "not known to be prunable" in the `Retention` type, and use not-known-prunable retention when inserting frontiers.

nuttycom:
Fixes #1398.

### DIFF
--- a/zcash_client_backend/src/data_api/chain.rs
+++ b/zcash_client_backend/src/data_api/chain.rs
@@ -188,6 +188,10 @@ pub struct CommitmentTreeRoot<H> {
 
 impl<H> CommitmentTreeRoot<H> {
     /// Construct a new `CommitmentTreeRoot` from its constituent parts.
+    ///
+    /// - `subtree_end_height`: The height of the block containing the note commitment that
+    ///   completed the subtree.
+    /// - `root_hash`: The Merkle root of the completed subtree.
     pub fn from_parts(subtree_end_height: BlockHeight, root_hash: H) -> Self {
         Self {
             subtree_end_height,
--- a/zcash_client_backend/src/serialization/shardtree.rs
+++ b/zcash_client_backend/src/serialization/shardtree.rs
@@ -13,6 +13,7 @@ const SER_V1: u8 = 1;
 const NIL_TAG: u8 = 0;
 const LEAF_TAG: u8 = 1;
 const PARENT_TAG: u8 = 2;
+const PRUNED_TAG: u8 = 3;
 
 /// Writes a [`PrunableTree`] to the provided [`Write`] instance.
 ///
@@ -41,6 +42,10 @@ pub fn write_shard<H: HashSer, W: Write>(writer: &mut W, tree: &PrunableTree<H>)
             }
             Node::Nil => {
                 writer.write_u8(NIL_TAG)?;
+                Ok(())
+            }
+            Node::Pruned => {
+                writer.write_u8(PRUNED_TAG)?;
                 Ok(())
             }
         }
@@ -74,6 +79,7 @@ fn read_shard_v1<H: HashSer, R: Read>(mut reader: &mut R) -> io::Result<Prunable
             Ok(Tree::leaf((value, flags)))
         }
         NIL_TAG => Ok(Tree::empty()),
+        PRUNED_TAG => Ok(Tree::empty_pruned()),
         other => Err(io::Error::new(
             io::ErrorKind::InvalidData,
             format!("Node tag not recognized: {}", other),
--- a/zcash_client_sqlite/src/testing.rs
+++ b/zcash_client_sqlite/src/testing.rs
@@ -461,25 +461,24 @@ where
     pub(crate) fn generate_next_block<Fvk: TestFvk>(
         &mut self,
         fvk: &Fvk,
-        req: AddressType,
+        address_type: AddressType,
         value: NonNegativeAmount,
     ) -> (BlockHeight, Cache::InsertResult, Fvk::Nullifier) {
         let pre_activation_block = CachedBlock::none(self.sapling_activation_height() - 1);
         let prior_cached_block = self.latest_cached_block().unwrap_or(&pre_activation_block);
         let height = prior_cached_block.height() + 1;
 
-        let (res, nf) = self.generate_block_at(
+        let (res, nfs) = self.generate_block_at(
             height,
             prior_cached_block.chain_state.block_hash(),
             fvk,
-            req,
-            value,
+            &[FakeCompactOutput::new(address_type, value)],
             prior_cached_block.sapling_end_size,
             prior_cached_block.orchard_end_size,
             false,
         );
 
-        (height, res, nf)
+        (height, res, nfs[0])
     }
 
     /// Adds an empty block to the cache, advancing the simulated chain height.
@@ -528,12 +527,11 @@ where
         height: BlockHeight,
         prev_hash: BlockHash,
         fvk: &Fvk,
-        req: AddressType,
-        value: NonNegativeAmount,
+        outputs: &[FakeCompactOutput],
         initial_sapling_tree_size: u32,
         initial_orchard_tree_size: u32,
         allow_broken_hash_chain: bool,
-    ) -> (Cache::InsertResult, Fvk::Nullifier) {
+    ) -> (Cache::InsertResult, Vec<Fvk::Nullifier>) {
         let mut prior_cached_block = self
             .latest_cached_block_below_height(height)
             .cloned()
@@ -587,13 +585,12 @@ where
                 .insert(height - 1, prior_cached_block.clone());
         }
 
-        let (cb, nf) = fake_compact_block(
+        let (cb, nfs) = fake_compact_block(
             &self.network(),
             height,
             prev_hash,
             fvk,
-            req,
-            value,
+            outputs,
             initial_sapling_tree_size,
             initial_orchard_tree_size,
             &mut self.rng,
@@ -603,7 +600,7 @@ where
         let res = self.cache_block(&prior_cached_block, cb);
         self.latest_block_height = Some(height);
 
-        (res, nf)
+        (res, nfs)
     }
 
     /// Creates a fake block at the expected next height spending the given note, and
@@ -796,6 +793,15 @@ impl<Cache> TestState<Cache> {
             .params
             .activation_height(NetworkUpgrade::Sapling)
             .expect("Sapling activation height must be known.")
+    }
+
+    /// Convenience method for obtaining the NU5 activation height for the network under test.
+    #[allow(dead_code)]
+    pub(crate) fn nu5_activation_height(&self) -> BlockHeight {
+        self.db_data
+            .params
+            .activation_height(NetworkUpgrade::Nu5)
+            .expect("NU5 activation height must be known.")
     }
 
     /// Exposes the test seed, if enabled via [`TestBuilder::with_test_account`].
@@ -1158,8 +1164,8 @@ impl<Cache> TestState<Cache> {
         &self,
     ) -> Result<Vec<TransactionSummary<AccountId>>, SqliteClientError> {
         let mut stmt = self.wallet().conn.prepare_cached(
-            "SELECT * 
-             FROM v_transactions 
+            "SELECT *
+             FROM v_transactions
              ORDER BY mined_height DESC, tx_index DESC",
         )?;
 
@@ -1283,7 +1289,7 @@ impl<AccountId> TransactionSummary<AccountId> {
 
 /// Trait used by tests that require a full viewing key.
 pub(crate) trait TestFvk {
-    type Nullifier;
+    type Nullifier: Copy;
 
     fn sapling_ovk(&self) -> Option<sapling::keys::OutgoingViewingKey>;
 
@@ -1306,6 +1312,8 @@ pub(crate) trait TestFvk {
         req: AddressType,
         value: NonNegativeAmount,
         initial_sapling_tree_size: u32,
+        // we don't require an initial Orchard tree size because we don't need it to compute
+        // the nullifier.
         rng: &mut R,
     ) -> Self::Nullifier;
 
@@ -1319,19 +1327,10 @@ pub(crate) trait TestFvk {
         req: AddressType,
         value: NonNegativeAmount,
         initial_sapling_tree_size: u32,
+        // we don't require an initial Orchard tree size because we don't need it to compute
+        // the nullifier.
         rng: &mut R,
-    ) -> Self::Nullifier {
-        self.add_spend(ctx, nf, rng);
-        self.add_output(
-            ctx,
-            params,
-            height,
-            req,
-            value,
-            initial_sapling_tree_size,
-            rng,
-        )
-    }
+    ) -> Self::Nullifier;
 }
 
 impl TestFvk for DiversifiableFullViewingKey {
@@ -1379,6 +1378,30 @@ impl TestFvk for DiversifiableFullViewingKey {
         ctx.outputs.push(cout);
 
         note.nf(&self.fvk().vk.nk, position as u64)
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn add_logical_action<P: consensus::Parameters, R: RngCore + CryptoRng>(
+        &self,
+        ctx: &mut CompactTx,
+        params: &P,
+        height: BlockHeight,
+        nf: Self::Nullifier,
+        req: AddressType,
+        value: NonNegativeAmount,
+        initial_sapling_tree_size: u32,
+        rng: &mut R,
+    ) -> Self::Nullifier {
+        self.add_spend(ctx, nf, rng);
+        self.add_output(
+            ctx,
+            params,
+            height,
+            req,
+            value,
+            initial_sapling_tree_size,
+            rng,
+        )
     }
 }
 
@@ -1428,10 +1451,10 @@ impl TestFvk for orchard::keys::FullViewingKey {
         _: BlockHeight,
         req: AddressType,
         value: NonNegativeAmount,
-        _: u32,
+        _: u32, // the position is not required for computing the Orchard nullifier
         mut rng: &mut R,
     ) -> Self::Nullifier {
-        // Generate a dummy nullifier
+        // Generate a dummy nullifier for the spend
         let revealed_spent_note_nullifier =
             orchard::note::Nullifier::from_bytes(&pallas::Base::random(&mut rng).to_repr())
                 .unwrap();
@@ -1461,12 +1484,12 @@ impl TestFvk for orchard::keys::FullViewingKey {
         _: &P,
         _: BlockHeight,
         revealed_spent_note_nullifier: Self::Nullifier,
-        req: AddressType,
+        address_type: AddressType,
         value: NonNegativeAmount,
-        _: u32,
+        _: u32, // the position is not required for computing the Orchard nullifier
         rng: &mut R,
     ) -> Self::Nullifier {
-        let (j, scope) = match req {
+        let (j, scope) = match address_type {
             AddressType::DefaultExternal => (0u32.into(), zip32::Scope::External),
             AddressType::DiversifiedExternal(idx) => (idx, zip32::Scope::External),
             AddressType::Internal => (0u32.into(), zip32::Scope::Internal),
@@ -1486,9 +1509,10 @@ impl TestFvk for orchard::keys::FullViewingKey {
     }
 }
 
-#[allow(dead_code)]
+#[derive(Clone, Copy)]
 pub(crate) enum AddressType {
     DefaultExternal,
+    #[allow(dead_code)]
     DiversifiedExternal(DiversifierIndex),
     Internal,
 }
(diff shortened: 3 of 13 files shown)


zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1498,307 @@ pub(crate) mod tests {
...
+
+    #[cfg(feature = "orchard")]
+    fn prepare_orchard_block_spanning_test(update_subtree_roots: bool) -> TestState<BlockCache> {
+        // Set up a scanning state such that:
+        // * The wallet birthday is in the shard prior to the chain tip
+        // * The user receives funds in the last complete block in the birthday shard,
+        //   in the last note in that block.
+        // * The next block crosses the shard boundary, with two notes in the prior
+        //   shard and two blocks in the subsequent shard
+        // * An additional 100 blocks are scanned, to ensure that
```
daira, line 1510: Incomplete sentence.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1498,307 @@ pub(crate) mod tests {
...
+            for c in res.orchard_note_commitments() {
+                final_orchard_tree.append(*c);
+            }
+
+            // one block spanning the shard boundary
+            let (spanning_block_height, res, _) =
+                st.generate_next_block_multi(&vec![fake_output(false); 4]);
+
+            // We only add two note commitments to the Orchard frontier, which should
+            // complete a 2^16 subtree. We can then add that subtree root to the
```
daira, line 1599: Incomplete sentence.

daira (approved):
utACK with minor comments on comments.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1374,8 +1377,10 @@ pub(crate) fn checkpoint_gaps<T: ShieldedPoolTester>() {
         account.birthday().height() + 10,
         BlockHash([0; 32]),
         &not_our_key,
-        AddressType::DefaultExternal,
-        not_our_value,
+        &[FakeCompactOutput {
+            address_type: AddressType::DefaultExternal,
+            value: not_our_value,
+        }],
```
str4d: You defined a `FakeCompactOutput::new` function, but are not using it here:
```suggestion
        &[FakeCompactOutput::new(
            AddressType::DefaultExternal,
            not_our_value,
        )],
```

EDIT: Looks like this was fixed in the second commit. I'd prefer it to be done in the first commit if you're making other changes to that commit.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1956,8 +1961,10 @@ pub(crate) fn invalid_chain_cache_disconnected<T: ShieldedPoolTester>() {
         disconnect_height,
         BlockHash([1; 32]),
         &dfvk,
-        AddressType::DefaultExternal,
-        NonNegativeAmount::const_from_u64(8),
+        &[FakeCompactOutput {
+            address_type: AddressType::DefaultExternal,
+            value: NonNegativeAmount::const_from_u64(8),
+        }],
```
str4d: Ditto:
```suggestion
        &[FakeCompactOutput::new(
            AddressType::DefaultExternal,
            NonNegativeAmount::const_from_u64(8),
        )],
```

EDIT: Looks like this was fixed in the second commit. I'd prefer it to be done in the first commit if you're making other changes to that commit.

zcash_client_sqlite/src/testing.rs
```diff
@@ -471,8 +471,31 @@ where
...
+        value: NonNegativeAmount,
+    ) -> (BlockHeight, Cache::InsertResult, Fvk::Nullifier) {
+        let pre_activation_block = CachedBlock::none(self.sapling_activation_height() - 1);
+        let prior_cached_block = self.latest_cached_block().unwrap_or(&pre_activation_block);
+        let height = prior_cached_block.height() + 1;
+
+        let (res, nf) = self.generate_block_at(
+            height,
+            prior_cached_block.chain_state.block_hash(),
+            &[FakeCompactOutput::new(fvk, address_type, value)],
```
str4d: This method makes no sense to me. Its name (`generate_next_block_multi`) and documentation indicates that it should include multiple outputs, but its code is identical to `generate_next_block`; the code here explicitly only adds one output. Either fix the method to take multiple outputs, or remove it.

EDIT: Looks like this was fixed in the third commit. This fix should definitely be moved into the second commit, as otherwise that commit makes no sense.

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -188,6 +188,10 @@ pub struct CommitmentTreeRoot<H> {
 
 impl<H> CommitmentTreeRoot<H> {
     /// Construct a new `CommitmentTreeRoot` from its constituent parts.
+    ///
+    /// - subtree_end_height The height of the block that produced the note commitment that
+    ///   completed the subtree.
+    /// - root_hash The Merkle root of the completed subtree.
```
str4d, line 194:
```suggestion
    /// - `subtree_end_height`: The height of the block containing the note commitment that
    ///   completed the subtree.
    /// - `root_hash`: The Merkle root of the completed subtree.
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -1938,22 +1937,62 @@ impl BlockCache {
...
 
+pub(crate) struct NoteCommitments {
+    sapling_note_commitments: Vec<sapling::Node>,
+    #[cfg(feature = "orchard")]
+    orchard_note_commitments: Vec<MerkleHashOrchard>,
+}
+
+impl NoteCommitments {
+    #[allow(dead_code)]
+    pub(crate) fn sapling_note_commitments(&self) -> &[sapling::Node] {
```
str4d, line 1932: The type is already called `NoteCommitments`, so just name this method `NoteCommitments::sapling`:
```suggestion
    pub(crate) fn sapling(&self) -> &[sapling::Node] {
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -1938,22 +1937,62 @@ impl BlockCache {
...
+        self.sapling_note_commitments.as_ref()
+    }
+
+    #[cfg(feature = "orchard")]
+    pub(crate) fn orchard_note_commitments(&self) -> &[MerkleHashOrchard] {
```
str4d, line 1937: Ditto:
```suggestion
    pub(crate) fn orchard(&self) -> &[MerkleHashOrchard] {
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -1938,22 +1937,62 @@ impl BlockCache {
...
+        self.orchard_note_commitments.as_ref()
+    }
+}
+
 impl TestCache for BlockCache {
     type BlockSource = BlockDb;
-    type InsertResult = ();
+    type InsertResult = NoteCommitments;
```
str4d, line 1944: Why are we only returning `NoteCommitments` from `impl TestCache for BlockCache`, and not `impl TestCache for FsBlockCache`?

Also, why is this even necessary? `TestCache::insert` takes a `&CompactBlock`, so the caller still has the block and can fetch these commitments from it directly (e.g. with the `NoteCommitments::from_compact_block` constructor I suggest below, or by us adding a helper method to `CompactBlock` in `zcash_client_backend/src/proto.rs`).

zcash_client_sqlite/src/testing.rs
```diff
@@ -1938,22 +1937,62 @@ impl BlockCache {
...
+            orchard_note_commitments: cb
+                .vtx
+                .iter()
+                .flat_map(|tx| {
+                    tx.actions
+                        .iter()
+                        .map(|act| MerkleHashOrchard::from_cmx(&act.cmx().unwrap()))
+                })
+                .collect(),
+        };
```
str4d, line 1972: Move this into a private constructor `NoteCommitments::from_compact_block` (enabling its reuse as necessary), or move its parts onto `CompactBlock` itself and remove `NoteCommitments`.

str4d (changes requested):
Partial review of 349722aefe4d905314ead4028b283f1ac2252d8a. I have not reviewed the new test, the bugfix, or the clippy lint fixes.

zcash_client_sqlite/src/testing.rs
nuttycom: `InsertResult` for `FsBlockCache` was already returning `BlockMeta` and I saw no reason to change it, because all of the relevant tests are just using `BlockCache`.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/349722aefe4d905314ead4028b283f1ac2252d8a..62f3abf300c856fe0bc95c2e538b542387b26db8) to address comments from code review.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1481,4 +1492,307 @@ pub(crate) mod tests {
...
+                                .value()
+                                .unwrap()
+                                .root(Some(Level::from(16))),
+                        )],
+                    )
+                    .unwrap();
+            }
+
+            // Add blocks up to the current chain tip.
+            for _ in 12..123 {
```
str4d, line 1624: This starting at 12 is confusing to me. AFAICT it's meant to indicate the block height offset relative to the birthday height (given how `final_orchard_tree` is mutated), but there's an off-by-one error here (13 blocks are generated prior to this point). Perhaps this is the source of the 111-block bug I mentioned above?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1481,4 +1492,307 @@ pub(crate) mod tests {
...
+        let new_tip = birthday.height() + 123;
+        st.wallet_mut().update_chain_tip(new_tip).unwrap();
+
+        // Verify that the suggested scan ranges includes only the chain-tip range with ChainTip
+        // priority, and that the range from the wallet birthday to the end of the birthday shard
+        // has Historic priority.
+        let birthday_height = birthday.height().into();
+        let expected = vec![
+            scan_range(
+                (birthday_height + 12)..(new_tip + 1).into(),
```
str4d, line 1652: How should this 12 correspond to the (possibly-erroneous) 12 in `prepare_orchard_block_spanning_test`? As in an earlier comment, a diagram of how the chain is being set up would be helpful.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1481,4 +1492,307 @@ pub(crate) mod tests {
...
+            &GreedyInputSelector::new(change_strategy, DustOutputPolicy::default());
+
+        let proposal = st.propose_transfer(
+            account.account_id(),
+            input_selector,
+            request,
+            NonZeroU32::new(10).unwrap(),
+        );
+
+        assert_matches!(proposal, Err(_));
```
str4d, line 1803: Would it also be useful to test that after scanning the spanning block, we can make a proposal?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1481,4 +1492,307 @@ pub(crate) mod tests {
...
+        let new_tip = birthday.height() + 123;
+        st.wallet_mut().update_chain_tip(new_tip).unwrap();
+
+        // Verify that the suggested scan ranges includes only the chain-tip range with ChainTip
+        // priority, and that the range from the wallet birthday to the end of the birthday shard
+        // has Historic priority.
+        let birthday_height = birthday.height().into();
+        let expected = vec![
+            scan_range(
+                birthday_height..(new_tip + 1).into(),
```
str4d, line 1743: This range not being split in two is clearly due to not updating the subtree roots in `prepare_orchard_block_spanning_test`, but it's not clear to me why this is necessary in addition to skipping the spanning block (to provoke the error), and why the other two possible combinations are not tested. While my earlier comment about documenting `prepare_orchard_block_spanning_test` would help, a comment here (or in `orchard_block_spanning_tip_boundary`, wherever is clearer) would also be useful.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1010,17 +1010,11 @@ pub(crate) fn put_shard_roots<
...
-                    LevelShifter(r.root_hash().clone()),
-                    Retention::Checkpoint {
-                        id: (),
-                        is_marked: false,
-                    },
-                )
-            }),
+            roots
+                .iter()
+                .map(|r| (LevelShifter(r.root_hash().clone()), Retention::Reference)),
```
str4d, line 1017: Does the fact that we no longer checkpoint in the cap tree matter for backwards compatibility (serialization or otherwise)?

zcash_client_sqlite/src/wallet/init/migrations/shardtree_support.rs
```diff
@@ -173,7 +173,7 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                                 nonempty_frontier.clone(),
                                 Retention::Checkpoint {
                                     id: block_height,
-                                    is_marked: false,
+                                    marking: Marking::Reference,
```
str4d, line 176: This change will only affect unmigrated DBs. It is a correct change to the migration, but (separately from this particular migration) is there any reasonable way to detect frontiers that were already inserted into the tree that have this issue (such as the ones that led to the bug report this PR is addressing)? Or will affected users need to recover from seed?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1501,7 +1507,8 @@ pub(crate) mod tests {
         //   in the last note in that block.
         // * The next block crosses the shard boundary, with two notes in the prior
         //   shard and two blocks in the subsequent shard
-        // * An additional 100 blocks are scanned, to ensure that
+        // * An additional 100 blocks are scanned, to ensure that the checkpoint
+        //   is pruned.
```
str4d, line 1511: Nit: this change, and the two below it, can safely be moved to the third commit that added the test (as they are not changes forced by the `shardtree` bump).

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -791,46 +791,65 @@ pub(crate) mod tests {
...
-
+            .with_initial_chain_state(|rng, network| {
                 // We set the Sapling and Orchard frontiers at the birthday height to be
                 // 1234 notes into the second shard.
                 let frontier_position = Position::from((0x1 << 16) + 1234);
-                let birthday_height =
-                    network.activation_height(NetworkUpgrade::Nu5).unwrap() + birthday_offset;
+                let initial_shard_end =
+                    network.activation_height(NetworkUpgrade::Nu5).unwrap() + initial_shard_blocks;
+                let birthday_height = initial_shard_end + birthday_offset;
```
str4d, line 807: Document on `test_with_nu5_birthday_offset` that `birthday_offset` is an offset from `initial_shard_blocks`, as this is a semantic change to the function argument. Also document that `initial_shard_blocks` is an offset from NU5.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -791,46 +791,65 @@ pub(crate) mod tests {
     }
 
     pub(crate) fn test_with_nu5_birthday_offset<T: ShieldedPoolTester>(
+        initial_shard_blocks: u32,
         birthday_offset: u32,
         prior_block_hash: BlockHash,
+        insert_prior_roots: bool,
```
str4d, line 797: Document what this does and in what kind of tests it should be used.

str4d (changes requested):
Reviewed 62f3abf300c856fe0bc95c2e538b542387b26db8

zcash_client_sqlite/src/testing.rs
```diff
@@ -528,8 +527,7 @@ where
         height: BlockHeight,
         prev_hash: BlockHash,
         fvk: &Fvk,
-        req: AddressType,
-        value: NonNegativeAmount,
+        outputs: &[FakeCompactOutput],
```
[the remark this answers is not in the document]
str4d: AFAICT this is still unresolved (the documentation for `generate_block_at` is now wrong).

zcash_client_sqlite/src/wallet/init/migrations/shardtree_support.rs
nuttycom: The affected users will need to recover from seed, but due to the nature of the bug there should be very few affected users.

zcash_client_sqlite/src/wallet/commitment_tree.rs
nuttycom: It shouldn't. Checkpoint retention was possibly actually a bug, because it implied that pruning was permitted within the cap, which shouldn't be allowed to happen (it's not a correctness problem, because the cap nodes can always be recomputed, but losing that caching kind of defeats the purpose of the cap.)

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: This test does not actually reproduce the error; I added it as an initial attempt to reproduce, and then decided that it was useful independently and so decided to keep it. I can split it out of this PR, but it's closely related to the behavior under test (for obvious reasons) so I thought it kind of made sense to keep here?

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: `birthday_height + 12` is the height of the block that spans the shard boundary. Remember that the birthday height is the height of the first block that may contain wallet funds, not the maximum height known _not_ to contain wallet funds. So the birthday block is the first of the 11 blocks initially generated.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/62f3abf300c856fe0bc95c2e538b542387b26db8..e3956c325155d73a469f8fd59730f6ce40668269?w=1) to address comments from review.

zcash_client_backend/src/serialization/shardtree.rs
```diff
@@ -43,6 +44,10 @@ pub fn write_shard<H: HashSer, W: Write>(writer: &mut W, tree: &PrunableTree<H>)
                 writer.write_u8(NIL_TAG)?;
                 Ok(())
             }
+            Node::Pruned => {
+                writer.write_u8(PRUNED_TAG)?;
+                Ok(())
```
daira, line 49: This will not be readable by prior implementations of `read_shard_v1`. Should the version number be bumped, or would that just cause more problems?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1508,352 @@ pub(crate) mod tests {
         let actual = suggest_scan_ranges(&st.wallet().conn, Ignored).unwrap();
         assert_eq!(actual, expected);
     }
+
+    /// This tests the case wherein:
+    /// * The wallet birthday is in the shard prior to the chain tip
+    /// * The user receives funds in the last complete block in the birthday shard,
+    ///   in the last note in that block.
+    /// * The next block crosses the shard boundary, with two notes in the prior
+    ///   shard and two blocks in the subsequent shard
```
daira, line 1517:
```suggestion
    ///   shard and two notes in the subsequent shard.
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1508,352 @@ pub(crate) mod tests {
...
+            // root of the completed subtree.
+            for c in res.orchard().iter().take(2) {
+                final_orchard_tree.append(*c);
+            }
+
+            assert_eq!(final_orchard_tree.tree_size(), 0x1 << 17);
+            assert_eq!(spanning_block_height, birthday.height() + 12);
+
+            // Insert the root of the completed subtree if `with_birthday_subtree_root` is set.
+            // This simulates the situation where the subtree roots have all bee inserted prior
```
daira, line 1641:
```suggestion
            // This simulates the situation where the subtree roots have all been inserted prior
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1508,352 @@ pub(crate) mod tests {
...
+        let new_tip = birthday.height() + 122;
+        st.wallet_mut().update_chain_tip(new_tip).unwrap();
+
+        // Verify that the suggested scan ranges includes only the chain-tip range with ChainTip
+        // priority, and that the range from the wallet birthday to the end of the birthday shard
+        // has Historic priority.
+        let birthday_height = birthday.height().into();
+        let expected = vec![
+            scan_range(
+                (birthday_height + 12)..(new_tip + 1).into(),
```
daira, line 1691:
```suggestion
                (birthday_height + 12)..=new_tip.into(),
```
Non-blocking.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1508,352 @@ pub(crate) mod tests {
...
+        let new_tip = birthday.height() + 122;
+        st.wallet_mut().update_chain_tip(new_tip).unwrap();
+
+        // Verify that the suggested scan ranges includes only the chain-tip range with ChainTip
+        // priority, and that the range from the wallet birthday to the end of the birthday shard
+        // has Historic priority.
+        let birthday_height = birthday.height().into();
+        let expected = vec![
+            scan_range(
+                birthday_height..(new_tip + 1).into(),
```
daira, line 1784:
```suggestion
                birthday_height..=new_tip.into(),
```
Non-blocking.

daira (dismissed):
utACK

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1508,352 @@ pub(crate) mod tests {
         let actual = suggest_scan_ranges(&st.wallet().conn, Ignored).unwrap();
         assert_eq!(actual, expected);
     }
+
+    /// This tests the case wherein:
```
nuttycom, line 1512:
```suggestion
    /// This sets up the case wherein:
```

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1475,4 +1508,352 @@ pub(crate) mod tests {
...
+    ///   shard and two blocks in the subsequent shard
+    /// * An additional 110 blocks are scanned, to ensure that the checkpoint
+    ///   is pruned.
+    ///
+    /// The diagram below shows the arrangement. the position of the X indicates the
+    /// note commitment for the note belonging to our wallet.
+    /// ```
+    /// blocks:      |<---- 5000 ---->|<----- 10 ---->|<--- 11 --->|<- 1  ->|<- 1 ->|<----- 110 ----->|
+    ///       nu5_activation                       birthday                                       chain_tip
+    /// commitments: |<---- 2^16 ---->|<--(2^16-50)-->|<--- 44 --->|<-___X->|<- 4 ->|<----- 114 ------|
```
nuttycom, line 1526:
```suggestion
    /// commitments: |<---- 2^16 ---->|<--(2^16-50)-->|<--- 44 --->|<-___X->|<- 4 ->|<----- 110 ------|
```

zcash_client_backend/src/serialization/shardtree.rs
nuttycom: I think this is fine; we can treat version changes that are backwards-compatible as part of the same version.

str4d (approved):
utACK 0c1ac8a629aaa4a9d8fe95460a3965554ee4d8d9

str4d (approved):
Re-utACK 3e090dca6695040b1a446c130b864f230a3e8408

daira (commented):
Post-hoc ACK

PR #1399, opened 2024-05-23
Disable `sapling-crypto` default features by default

str4d:
The only default-enabled feature flag in `sapling-crypto` is the `multicore` feature flag, which we re-export in each crate that includes proof creation. We need to disable it as a default feature of our dependency in order to enable it to be correctly disabled when a user of e.g. `zcash_primitives` disables its default features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3176,7 +3176,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_primitives"
-version = "0.15.0"
+version = "0.15.1"
 dependencies = [
  "aes",
  "assert_matches",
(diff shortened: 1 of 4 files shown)


nuttycom (approved):
utACK

str4d:
The `cargo vet` issue I'm not going to resolve in this PR, as I expect it would cause conflicts with #1392.

nuttycom (approved):
utACK

str4d:
`8b2188a0f34fea2cebbf48a9945c75715f92992e` published as `zcash_primitives-0.15.1`.

PR #1400, opened 2024-05-29
zcash_client_backend: Allow sharing instances of `ScanningKeys` between threads

upbqdn:
## Motivation

It would be handy if Zebra could share instances of `ScanningKeys` between threads.

## Solution

- Bound `ScanningKeyOps` by `Send + Sync` so `ScanningKeys` auto-implements `Send` and `Sync` instead of `!Send` and `!Sync`.

### DIFF
--- a/zcash_client_backend/src/scanning.rs
+++ b/zcash_client_backend/src/scanning.rs
@@ -52,7 +52,7 @@ use std::marker::PhantomData;
 /// [`CompactSaplingOutput`]: crate::proto::compact_formats::CompactSaplingOutput
 /// [`CompactOrchardAction`]: crate::proto::compact_formats::CompactOrchardAction
 /// [`scan_block`]: crate::scanning::scan_block
-pub trait ScanningKeyOps<D: Domain, AccountId, Nf> {
+pub trait ScanningKeyOps<D: Domain, AccountId, Nf>: Send + Sync {
     /// Prepare the key for use in batch trial decryption.
     fn prepare(&self) -> D::IncomingViewingKey;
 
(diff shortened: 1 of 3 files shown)


zcash_client_backend/src/scanning.rs
```diff
@@ -120,7 +120,7 @@ pub struct ScanningKey<Ivk, Nk, AccountId> {
     key_scope: Option<Scope>,
 }
 
-impl<AccountId> ScanningKeyOps<SaplingDomain, AccountId, sapling::Nullifier>
+impl<AccountId: Send + Sync> ScanningKeyOps<SaplingDomain, AccountId, sapling::Nullifier>
```
str4d, line 123: Is this bound (repeated everywhere) avoidable if we add the equivalent bound on `AccountId` in `ScanningKeyOps`?

upbqdn: Unfortunately, to be able to move an instance of `ScannningKeys` to `tokio::task::spawn_blocking` for example, the compiler requires `Send` and `Sync` to be implemented for `ScanningKeyOps`. Otherwise, I get this error:

```md
`(dyn ScanningKeyOps<sapling_crypto::note_encryption::SaplingDomain, AccountId, sapling_crypto::note::nullifier::Nullifier> + 'static)` cannot be sent between threads safely
```

I tried bounding `AccountId` by `Send` and `Sync` in `ScanningKeyOps`, but that didn't help.

Bounding `ScanningKeyOps` itself eliminates the error, but once I do that, I have to make the rest of the changes in this PR to fix subsequent bound requirements.

upbqdn: This PR is not important though, feel free to close it without merging. The workaround was to create an instance of `ScanningKeys` in the scanning thread itself. Moreover, we have no plans to work on the scanner any time soon.

upbqdn: We'll be removing the scanner from Zebra entirely: https://github.com/ZcashFoundation/zebra/issues/9428, so I'm going to close this PR.

PR #1257, opened 2024-03-12
Add support for transparent-source-only (TEX) addresses

daira:
fixes #1167

PR #1402, opened 2024-05-31
zcash_client_sqlite: Align handling of transparent UTXOs with that of shielded notes.

nuttycom:
This PR is best reviewed hiding whitespace changes.


zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -745,4 +745,88 @@ mod tests {
         }
         assert_eq!(row_count, 2);
     }
+
+    /// This is a copy of [`crate::wallet::put_block`] as of the expected database
+    /// state corresponding to this migration. It is duplicated here as later
+    /// updates to the database schema require incomplatible changes to `put_block`.
```
daira, line 751:
```suggestion
    /// updates to the database schema require incompatible changes to `put_block`.
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2338,6 +2350,21 @@ pub(crate) fn put_block(
         ":orchard_action_count": orchard_action_count,
     ])?;
 
+    // If we now have a block corresponding to a received transparent output that
+    // had not been scanned at the time the UTXO was discovered, update the
+    // associated transaction record to refer to that block.
```
nuttycom, line 2355: There's a small data corruption hazard here, in that we're relying exclusively upon the block height to associate the transaction to the block. This is because `CompactBlock` values only contain `CompactTx` entries for transactions that contain shielded inputs or outputs, and the `GetAddressUtxosReply` data does not contain the block hash. As such, it's necessary to ensure that any chain rollback to below the received height causes that height to be set to `NULL`.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -574,35 +574,43 @@ mod tests {
...
                     ON DELETE CASCADE,
                 FOREIGN KEY (transaction_id)
                     -- We do not delete transactions, so this does not cascade
                     REFERENCES transactions(id_tx),
                 UNIQUE (transparent_received_output_id, transaction_id)
+            )"#,
+            "CREATE TABLE transparent_received_outputs (
+                id INTEGER PRIMARY KEY,
+                transaction_id INTEGER,
+                output_index INTEGER NOT NULL,
```
daira, line 591: `transaction_id` is nullable but `output_index` is not. Do we ever receive a transparent output without having its transaction?

In the `utxos` table we had `prevout_txid BLOB NOT NULL`, but we have no equivalent of that here. So if it were the case that we can see a transparent output with knowledge of only the outpoint and no corresponding row in `transactions`, this would be a loss of information relative to `utxos`.

Also, I don't think that `CONSTRAINT transparent_output_unique UNIQUE (transaction_id, output_index)` does the intended thing if `transaction_id` can be null.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -574,35 +574,43 @@ mod tests {
...
+                script BLOB NOT NULL,
+                value_zat INTEGER NOT NULL,
+                received_height INTEGER,
+                max_observed_unspent_height INTEGER,
+                FOREIGN KEY (transaction_id) REFERENCES transactions(id_tx),
+                FOREIGN KEY (account_id) REFERENCES accounts(id),
+                CONSTRAINT transparent_output_unique UNIQUE (transaction_id, output_index),
+                CONSTRAINT last_observed_unspent CHECK (
+                    -- it is intended that if received_height is null, then
+                    -- last_observed_unspent_height must also be null
```
daira, line 603: What's the intended relation between nullness of `received_height` and `transaction_id`?

nuttycom: `transaction_id`, once set, will not be changed; the record for the transaction with that txid will exist, but it may not be mined.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -574,35 +574,43 @@ mod tests {
...
+                    REFERENCES transparent_received_outputs(id)
                     ON DELETE CASCADE,
                 FOREIGN KEY (transaction_id)
                     -- We do not delete transactions, so this does not cascade
                     REFERENCES transactions(id_tx),
                 UNIQUE (transparent_received_output_id, transaction_id)
+            )"#,
+            "CREATE TABLE transparent_received_outputs (
+                id INTEGER PRIMARY KEY,
+                transaction_id INTEGER,
```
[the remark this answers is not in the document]
nuttycom: I have actively been attempting to move us to the normal SQL naming conventions:
* tables are named `<plural_noun>`
* the primary key is always named `id`
* foreign key columns are always named `<singular_noun>_id`

zcash_client_sqlite/src/wallet/init.rs
nuttycom: `transaction_id` should not be nullable, this is an oversight.

zcash_client_sqlite/src/wallet/init.rs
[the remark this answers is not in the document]
daira: Oh this is analogous to `{sapling,orchard}_received_note_spends(transaction_id)`. Never mind, this is fine as-is.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -840,107 +913,79 @@ mod tests {
...
-                    SELECT v_received_notes.account_id  AS account_id,
-                           transactions.block           AS block,
-                           transactions.txid            AS txid,
-                           v_received_notes.pool        AS pool,
+                    -- Outputs spent in this transaction
+                    SELECT ro.account_id            AS account_id,
+                           transactions.block       AS block,
+                           transactions.txid        AS txid,
+                           ro.pool                  AS pool,
                            id_within_pool_table,
```
daira, line 944: This is hazardous because it mixes together ids from several tables.

daira: @str4d points out https://github.com/zcash/librustzcash/commit/8b8757ce659f981eb6a3e0aa69c27d9115f3a783 which was a fix to an even worse situation.

zcash_client_sqlite/src/wallet/init.rs
str4d: @daira and I noted that the act of fetching UTXOs _is_ scanning for a wallet, so we should require the caller to always call `getTransaction` for each UTXO afterwards. That means we will always have the full transaction to insert into the wallet.

We do also need to be able to set `received_height`, which exists as a temporary crutch until we can rework the `blocks` table to not only contain fully-scanned blocks, so rather than directly using `decrypt_and_store_tx` we may want to rework `put_utxo` to instead take the full transaction.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: We could avoid this if we were using UUIDs for primary keys (as has been the best practice for a lot of years now) but for here I don't think we can avoid this.

zcash_client_sqlite/src/wallet/init.rs
daira: Whenever we call `put_legacy_transparent_utxo` we have the tx_ref (i.e. `transaction_id`) anyway, so we can just add it as a parameter to that call. Then `transaction_id` can be non-nullable.

daira: Oh! `WalletTransparentOutput` should include `transaction_id`. We only construct that type in these situations:
* from a row of the `utxos` / `transparent_received_outputs` table (in `to_unspent_transparent_output`);
* in `store_sent_tx` for ephemeral outputs (after #1257), in which case we have `transaction_id` because we've just obtained it from `wallet::put_tx_data`;
* in tests where we could easily make a fake transaction (which would improve the realism of the test).

zcash_client_sqlite/src/wallet/init.rs
daira: The `transactions` table has `block` which is the block height if known. So storing this in `transparent_received_outputs` as well is unnormalized. We should just join on `transactions` whenever we need it. In fact I think we do already join on `transactions` in all the cases where we access `received_height` (or we're setting both `received_height` and `block` to null), so there is no loss of efficiency.

This does make the constraint that `max_observed_unspent_height IS NULL OR max_observed_unspent_height >= transactions(block)` harder to automatically enforce, but it's still an improvement, because we already had an unenforced implicit constraint on consistency of `received_height` and `transactions(block)`.

zcash_client_sqlite/src/wallet/init.rs
daira: It has a BLOB type that we could use to encode the pool and id. Not sure that is worth the complexity of encoding and decoding.

You'll love [this gem from the SQLite docs](https://www.sqlite.org/datatype3.html#affinity_name_examples):
> Note that a declared type of "FLOATING POINT" would give INTEGER affinity, not REAL affinity, due to the "INT" at the end of "POINT".

:sob:

daira: In the cold light of day, using BLOB sounds like a bad idea because you then either have to encode a pool field as well, which is a redundant encoding that could be inconsistent, or you have no way to obtain the pool for SQL queries. And there's no way of saying "pair for which it's okay to destructure the first element but not the second" in SQL.

zcash_client_sqlite/src/wallet/init.rs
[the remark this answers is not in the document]
str4d: I explicitly did not use the normal SQL naming convention when originally writing the tables, because (for instance) `transaction_id` collides semantically with "transaction ID". But Kris spent enough time convincing me that I'm now not opposed to this migration.

zcash_client_sqlite/src/wallet/init.rs
daira: @str4d and/or @nuttycom pointed out that `wallet::put_legacy_transparent_utxo` is also called from `wallet::put_received_transparent_utxo` which is exposed via the data API. (E.g. see [Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putUtxo](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/159de09b2bf1cc1c752260c76d444c11dfabebc9/backend-lib/src/main/rust/lib.rs#L1518) in the Android SDK.)

In that case, we should obtain the txid from the `WalletTransparentOutput`'s `output.hash`, and use it to look up the transaction in case we have it. If we don't have it, then we might have to synthesize a new entry in `transactions` with mostly null fields. However, we do always know the height in that case, because we created the `WalletTransparentOutput` from a `GetAddressUtxosReply` which has the height. So we either create the transaction entry with a known height, or we upsert that height to the existing entry.

If, on the other hand, we are adding an entry to `transparent_received_outputs` in order to record the ephemeral output of the first transaction in a ZIP 320 pair (once #1257 has merged), then we have the tx_ref so can record that immediately. We will only have the height once the transaction is mined (if it is), and so joining on `transactions` is the correct way to get the height for that output at any point in time.

In neither case do we need a `received_height` field directly in `transparent_received_outputs`, and it violates normalization to have it.

nuttycom: We can't set the height of the transaction unless we've already scanned the corresponding block, because the height is a foreign key to the `blocks` table (which uses the block height as its primary key.).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -563,46 +563,49 @@ mod tests {
...
+                mined_height INTEGER,
                 tx_index INTEGER,
                 expiry_height INTEGER,
                 raw BLOB,
                 fee INTEGER,
-                FOREIGN KEY (block) REFERENCES blocks(height)
-            )",
-            "CREATE TABLE transparent_received_output_spends (
+                FOREIGN KEY (block) REFERENCES blocks(height),
+                CHECK (block IS NULL OR mined_height = block)
```
daira, line 577: SQLite will apparently only report which constraint failed if it is named.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2115,10 +2123,13 @@ pub(crate) fn get_unspent_transparent_output(
...
+        "SELECT t.txid, u.output_index, u.script,
+                u.value_zat, t.mined_height AS received_height
+         FROM transparent_received_outputs u
+         JOIN transactions t ON t.id_tx = u.transaction_id
+         WHERE t.txid = :txid
+         AND t.mined_height IS NOT NULL
+         AND u.output_index = :output_index
          AND u.id NOT IN (
             SELECT txo_spends.transparent_received_output_id
             FROM transparent_received_output_spends txo_spends
```
daira:
```
OR tx.expiry_height IS NULL -- the spending tx will not expire
```
It could be that the spending tx is unmined and will not expire, but is dependent on another transaction that is unmined and *will* expire.

zcash_client_sqlite/src/wallet.rs
daira: Add something about that to the comment (maybe referencing `truncate_to_height`).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2350,10 +2376,11 @@ pub(crate) fn put_tx_meta(
 ) -> Result<i64, SqliteClientError> {
     // It isn't there, so insert our transaction into the database.
     let mut stmt_upsert_tx_meta = conn.prepare_cached(
-        "INSERT INTO transactions (txid, block, tx_index)
-        VALUES (:txid, :block, :tx_index)
+        "INSERT INTO transactions (txid, block, mined_height, tx_index)
+        VALUES (:txid, :block, :block, :tx_index)
         ON CONFLICT (txid) DO UPDATE
         SET block = :block,
+            mined_height = :mined_height,
```
daira, line 2047:
```suggestion
            mined_height = :block,
```
There is no `:mined_height` named param.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2527,29 +2555,56 @@ pub(crate) fn put_legacy_transparent_utxo<P: consensus::Parameters>(
...
+         RETURNING id_tx",
+        named_params![
+           ":txid": &output.outpoint().hash().to_vec(),
+           ":block": block,
+           ":mined_height": u32::from(output.height())
+        ],
+        |row| row.get::<_, i64>(0),
+    )?;
+
     let mut stmt_upsert_legacy_transparent_utxo = conn.prepare_cached(
```
daira:
```suggestion
    let mut stmt_upsert_legacy_transparent_output = conn.prepare_cached(
```

zcash_client_sqlite/src/wallet/init.rs
daira: For completeness: I suggested adding a `mined_height` field to `transactions`, but adding a constraint that it is equal to `block` when the latter is non-null. This avoids problems with a height recorded in `transparent_received_outputs` potentially being inconsistent between outputs in the same transaction. The PR currently takes my suggested approach, and `transaction_id` is no longer nullable, so this is resolved.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -32,3 +51,711 @@ pub(crate) fn detect_spending_accounts<'a>(
...
+         WHERE t.mined_height <= :summary_height
+         -- and the received txo is unspent
+         AND u.id NOT IN (
+           SELECT transparent_received_output_id
+           FROM transparent_received_output_spends txo_spends
+           JOIN transactions tx
+             ON tx.id_tx = txo_spends.transaction_id
+           WHERE tx.block IS NOT NULL -- the spending tx is mined
+           OR tx.expiry_height IS NULL -- the spending tx will not expire
+           OR tx.expiry_height > :summary_height -- the spending tx is unexpired
```
daira, line 341: It could be that the spending tx is unmined and will not expire, but is dependent on another transaction that is unmined and *will* expire.

This is a pre-existing issue. I don't think there's any way to fix it from this side of the lightwalletd protocol. A full node will keep only transactions that *could* be mined in its mempool (subject also to [ZIP 401](https://zips.z.cash/zip-0401) if the mempool becomes full), but a light client doesn't have enough information to do that.

daira: Filed as #1426.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -563,46 +563,49 @@ mod tests {
...
+            )"#,
+            "CREATE TABLE transparent_received_outputs (
+                id INTEGER PRIMARY KEY,
+                transaction_id INTEGER NOT NULL,
+                output_index INTEGER NOT NULL,
+                account_id INTEGER NOT NULL,
+                address TEXT NOT NULL,
+                script BLOB NOT NULL,
+                value_zat INTEGER NOT NULL,
+                max_observed_unspent_height INTEGER,
```
daira, line 598: Is it clear what this means? I'm concerned about off-by-one errors.

From how it's used, it seems to mean that the TXO was either created in this block or unspent at the *start* of the block. Either that or we are incorrectly assuming that UTXOs are not spent in the block where they're created.

zcash_client_sqlite/src/wallet/init/migrations/utxos_to_txos.rs
```diff
@@ -0,0 +1,322 @@
...
+                u.id,
+                t.id_tx,
+                prevout_idx,
+                received_by_account_id,
+                address,
+                script,
+                value_zat,
+                NULL
+            FROM utxos u
+            JOIN transactions t ON t.txid = u.prevout_txid;
```
daira, line 97: I think if you write this as `LEFT OUTER JOIN`, there will be a constraint error if the transactions table hasn't been updated to include entries for every `prevout_txid` (as should have happened above on lines 55-64), because `t.id_tx` will be NULL and `transaction_id` is declared as `NOT NULL`. That would make sure that if there is a bug in the migration then we can't inadvertently drop any TXOs.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/0eb0b3f07f353f48b027ee3ccf5561bcc4a43fb4..ddbe9936f7e78feb0233a5a3bb025dd6ed15af7b) to rebase on updated `main`.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/0eb0b3f07f353f48b027ee3ccf5561bcc4a43fb4..4a91a2cf411646a0d4c2cf82df651a20154083d9) to rebase on updated `main`.

zcash_client_sqlite/src/wallet/init.rs
nuttycom: The is the maximum block height at which we have observed that the TXO appears in the UTXO set.

str4d (changes requested):
Flushing comments

zcash_client_sqlite/src/wallet/init/migrations/utxos_to_txos.rs
```diff
@@ -0,0 +1,322 @@
...
+                blocks.height
+            FROM utxos
+            LEFT OUTER JOIN blocks ON blocks.height = utxos.height
+            WHERE utxos.prevout_txid NOT IN (
+                SELECT txid FROM transactions
+            );
+
+            INSERT INTO transactions_new
+            SELECT id_tx, txid, created, block, block, tx_index, expiry_height, raw, fee
+            FROM transactions;
```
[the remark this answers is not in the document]
str4d: I missed the `WHERE utxos.prevout_txid NOT IN` while writing my comment, but the renumbering of `id_tx` is definitely a problem (it breaks all existing foreign keys), so swapping the order of insertions here to preserve them is still something we should do.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -216,54 +218,68 @@ CREATE TABLE orchard_received_note_spends (
...
 ///   (downloading the full transaction containing the transparent output spend).
-pub(super) const TABLE_UTXOS: &str = r#"
-CREATE TABLE "utxos" (
+///
+/// ### Columns:
+/// - `id`: Primary key
+/// - `transaction_id`: Reference to the transaction in which this TXO was created
+/// - `output_index`: The output index of this TXO in the transaction referred to by `transaction_id`
+/// - `account_id`: The account that controls spend authority for this TXO
+/// - `address`: The address to which this TXO was sent
```
str4d, line 240: We don't store the address a shielded note was received to in the `*_received_notes` tables directly, because we don't need to: all we need to know is the recipient account and the note plaintext, and we can spend.

For transparent outputs, due to how our "transparent viewing keys" work, the specific address a transparent output was sent to is relevant in order to derive its corresponding spending key. However, we also already have that information in the `script` column (the address is extracted from it), and thus the seemingly-obvious reason for this column to exist is for joins. Where do we do these joins?

Another question: is `address` the right thing to store here semantically, or should we instead be storing the `(scope, address_index)` tuple that can be used to derive the address from the corresponding transparent IVK? We should be determining this at "transparent scan" time.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -216,54 +223,68 @@ CREATE TABLE orchard_received_note_spends (
...
+/// - `script`: The full txout script
+/// - `value_zat`: The value of the TXO in zatoshis
+/// - `max_observed_unspent_height`: The maximum block height at which this TXO was observed
+///   to be a member of the UTXO set. This is intended to be used to determine when the TXO
+///   is no longer a part of the UTXO set, in the case that the transaction that spends it is
+///   not detected by the wallet.
```
daira, line 251: This is still unclear about off-by-one issues.
```suggestion
/// - `max_observed_unspent_height`: The maximum block height at which this TXO was either
///   observed to be a member of the UTXO set at the start of the block, or observed to be an
///   output of a transaction mined in the block. This is intended to be used to determine when
///   the TXO is no longer a part of the UTXO set, in the case that the transaction that spends it
///   is not detected by the wallet.
```
I'm still worried by the fact that we currently never seem to use the contents of this field, and so I don't know whether these semantics will be what we actually want.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -216,54 +223,68 @@ CREATE TABLE orchard_received_note_spends (
     UNIQUE (orchard_received_note_id, transaction_id)
```
daira, line 223: Lines 211-212: "It plays the same role for Orchard notes as does [`TABLE_SAPLING_RECEIVED_NOTE_SPENDS`] for Sapling notes."

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -216,54 +223,68 @@ CREATE TABLE orchard_received_note_spends (
...
-/// This is identical to [`TABLE_SAPLING_RECEIVED_NOTE_SPENDS`]; see its documentation for
-/// details. Note however that [`TABLE_UTXOS`] and [`TABLE_SAPLING_RECEIVED_NOTES`] are
-/// not equivalent, and care must be taken when interpreting the result of joining this
-/// table to [`TABLE_UTXOS`].
-pub(super) const TABLE_TRANSPARENT_RECEIVED_OUTPUT_SPENDS: &str = "
-CREATE TABLE transparent_received_output_spends (
+/// This is identical to [`TABLE_SAPLING_RECEIVED_NOTE_SPENDS`]; see its documentation for details.
+/// Note however that [`TABLE_TRANSPARENT_RECEIVED_OUTPUTS`] and [`TABLE_SAPLING_RECEIVED_NOTES`]
+/// are not equivalent, and care must be taken when interpreting the result of joining this table
+/// to [`TABLE_TRANSPARENT_RECEIVED_OUTPUTS`].
```
daira, line 275: This comment just left me anxious that I was missing something, without increasing my understanding. :sweat_smile:

```suggestion
/// This plays the same role for transparent TXOs as does [`TABLE_SAPLING_RECEIVED_NOTE_SPENDS`]
/// for Sapling notes. However, [`TABLE_TRANSPARENT_RECEIVED_OUTPUTS`] and [`TABLE_SAPLING_RECEIVED_NOTES`] differ in that [fill in here precisely *how* they differ, and *what* care must be taken in joining them].
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -100,21 +100,28 @@ CREATE TABLE blocks (
...
+///   containing the transaction. It is present to allow the block height for a retrieved transaction
+///   to be stored without requiring that the entire block containing the transaction be scanned; the
+///   foreign key constraint on `block` prevents that column from being populated prior to complete
+///   scanning of the block. This is constrained to be equal to the `block` column if `block` is
+///   non-null.
+pub(super) const TABLE_TRANSACTIONS: &str = r#"
+CREATE TABLE "transactions" (
     id_tx INTEGER PRIMARY KEY,
     txid BLOB NOT NULL UNIQUE,
     created TEXT,
```
daira, line 115: Please document the `created` field.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: We need to be able to detect when one of our UTXOs is no longer part of the UTXO set; since this table is being updated, I think that now is a good time to start populating this field so that we can later add the functionality to identify when we need to go searching for a fully-transparent spend of one of our transparent UTXOs, which we can't otherwise detect.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: I don't have time at the moment to try to figure out what the right thing is to do here. We've had the address here all along, I don't see a reason to change it right now.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -497,219 +518,201 @@ pub(super) const TABLE_SQLITE_SEQUENCE: &str = "CREATE TABLE sqlite_sequence(nam
...
+       (orchard_received_notes.tx, 3, orchard_received_notes.action_index)
+UNION
+    SELECT
+        u.id AS id_within_pool_table,
+        u.transaction_id,
+        0 AS pool,
+        u.output_index,
+        u.account_id,
+        u.value_zat AS value,
+        0 AS is_change,
```
str4d, line 564: How is `is_change` used? `sapling_received_notes` and `orchard_received_notes` have columns `is_change INTEGER NOT NULL`; is it currently a glorified boolean (i.e. `0` specifically means "is not change"), or can it take other values? It would be useful to document this (either on the `sapling_received_notes` and `orchard_received_notes` tables, or on this view).

(I know that we set `0` for this in `v_transactions` below, but it would be good to document this.)

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -497,219 +518,201 @@ pub(super) const TABLE_SQLITE_SEQUENCE: &str = "CREATE TABLE sqlite_sequence(nam
...
 JOIN transactions
-    ON transactions.id_tx = v_received_notes.tx
+    ON transactions.id_tx = ro.transaction_id
 LEFT JOIN sent_notes
-    ON sent_notes.id = v_received_notes.sent_note_id
-UNION
-SELECT utxos.prevout_txid           AS txid,
-        0                            AS output_pool,
-        utxos.prevout_idx            AS output_index,
-        NULL                         AS from_account_id,
```
str4d, line 690: AFAICT (at least as of the first commit in the PR), this `NULL AS from_account_id` is preserved, because there is still an asymmetry between transparent and shielded (we are not inserting transactions into both sent and received tables for transparent, in the same way we do for shielded). I think that asymmetry needs fixing prior to the next release (if not in this PR).

str4d (changes requested):
Finished reviewing the first commit (e9ff54ac840a4d08f2d02b3d3a8e744c8360865f).

zcash_client_sqlite/src/wallet/db.rs
str4d: Specifically, there is an asymmetry between transparent and shielded in the following places:
- `WalletDb::put_blocks`
  - Likely non-blocking: this does nothing with transparent details, but that makes sense for now as we don't currently have transparent logic in `scan_cached_blocks`. However, we do need to ensure that any transparent aspects influenced by block scanning do get updated (the one that comes to mind is `transactions.block` which I think _is_ correctly handled in this PR).
- `WalletDb::store_decrypted_tx`
  - Calls both `put_received_note` and `put_sent_output` for both Sapling and Orchard depending on detection code, but only ever calls `put_sent_output` for transparent.
- `WalletDb::store_decrypted_tx`
  - Calls `put_received_note` for `sent_tx.outputs()` that are shielded, but not for transparent outputs.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2020,9 +2022,29 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    // considered to have been returned to the mempool; it _might_ be spendable in this state, but
+    // we must also set its max_observed_unspent_height field to NULL because the transaction may
+    // be rendered entirely invalid by a reorg that alters anchor(s) used in constructing shielded
+    // spends in the transaction.
+    conn.execute(
+        "UPDATE transparent_received_outputs
+         SET max_observed_unspent_height = NULL
+         WHERE max_observed_unspent_height > :height",
+        named_params![":height": u32::from(block_height)],
+    )?;
```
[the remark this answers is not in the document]
nuttycom: The output might be invalid at the truncation height; i.e. it might have been created in a transaction where the anchor used to spend notes in order to create the output is not valid. So we can't say anything about the max observed unspent height; we have to re-observe the output in order to set it.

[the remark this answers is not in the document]
str4d: By construction, `tx.anchor_height < tx.mined_height`. Ergo, per my comment, if `tx.mined_height <= :height` then `tx.anchor_height < :height` and thus the anchor is still valid.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -331,19 +319,15 @@ pub(crate) fn get_transparent_address_balances<P: consensus::Parameters>(
 
 pub(crate) fn add_transparent_account_balances(
     conn: &rusqlite::Connection,
-    chain_tip_height: BlockHeight,
-    min_confirmations: u32,
+    summary_height: BlockHeight,
     account_balances: &mut HashMap<AccountId, AccountBalance>,
 ) -> Result<(), SqliteClientError> {
-    let zero_conf_height = (chain_tip_height + 1).saturating_sub(min_confirmations);
-    let stable_height = chain_tip_height.saturating_sub(PRUNING_DEPTH);
```
str4d: I don't think that this change fixes the bug I outlined in https://github.com/zcash/librustzcash/pull/1257#discussion_r1637994605.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2020,9 +2022,29 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+
+    // Un-mine transactions. This must be done outside of the last_scanned_height check because
+    // transaction entries may be created as a consequence of receiving transparent TXOs.
+    conn.execute(
+        "UPDATE transactions
+         SET block = NULL, mined_height = NULL, tx_index = NULL
+         WHERE block > ?",
```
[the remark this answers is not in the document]
nuttycom: Hm, yeah, I guess that given the constraints on `mined_height` it's okay to use it in this fashion. This makes it a bit more load-bearing than I had originally anticipated.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: There is no boolean type supported by sqlite, so we have to use an integer.

zcash_client_sqlite/src/wallet.rs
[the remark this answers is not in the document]
nuttycom: Wait a minute: you're reading documentation about sort order, not about binary operations involving NULL:
```
sqlite> select NULL < 1;

sqlite> select 2 < 1;
0
sqlite> select 1 < 2;
1
```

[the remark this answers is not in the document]
str4d: After a long Slack thread discussion, we determined that the SQLite documentation I linked above is Just Wrong, because it is _very clearly_ documenting binary operations, _not_ sort order. But the binary operations actually behave like mathematical operations (if an argument is NULL, the result is NULL), despite the separate mathematical operations section very specifically limiting itself to not apply to binary operations.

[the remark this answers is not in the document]
str4d: None of this materially affects my bug reports; it just means that instead of `height = NULL` being treated as "infinite confirmations", any row matching `height = NULL` is ignored completely. The latter behaviour is incorrect when `max_height = chain_height + 1` (targeting the mempool height), so it still comes down to "unmined transactions are not correctly treated as being in the mempool by the SQL queries".

zcash_client_sqlite/src/wallet/db.rs
str4d: That's what I figured; please document this.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -33,3 +51,694 @@ pub(crate) fn detect_spending_accounts<'a>(
...
+                *taddr,
+                Some(TransparentAddressMetadata::new(
+                    Scope::External.into(),
+                    index,
+                )),
+            );
+        }
+    }
+
+    if let Some((taddr, child_index)) = get_legacy_transparent_address(params, conn, account)? {
```
str4d, line 112: The rebase of the first move-only commit undid the `s/child_index/address_index` rename from #1423 (41c30ef44da6e2fdbc3938148e0816a0cc704a56). Fix up the first move-only commit (and the second if necessary) to preserve the rename.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -33,3 +51,694 @@ pub(crate) fn detect_spending_accounts<'a>(
...
+        let height_1 = BlockHeight::from_u32(12345);
+        let bal_absent = st
+            .wallet()
+            .get_transparent_balances(account_id, height_1)
+            .unwrap();
+        assert!(bal_absent.is_empty());
+
+        // Create a fake transparent output.
+        let value = NonNegativeAmount::const_from_u64(100000);
+        let outpoint = OutPoint::new([1u8; 32], 1);
```
str4d, line 545: The rebase of the first move-only commit undid the `OutPoint::fake()` helper introduction from #1423 (6d35583a4b7eff04c094a82a16be83a7e97727a1). Fix up the first move-only commit (and the second if necessary) to preserve the change.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1167,47 +1167,12 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
     drop(sapling_trace);
 
     #[cfg(feature = "transparent-inputs")]
-    {
-        let transparent_trace = tracing::info_span!("stmt_transparent_balances").entered();
```
str4d, line 1297: Preserve this trace by adding a `#[tracing::instrument]` annotation to the new `transparent::add_transparent_account_balances` method.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Okay, the reason that we store the address explicitly is simply that we query against it - `get_unspent_transparent_outputs` takes the address as an argument. This is because when shielding we want to avoid shielding from multiple transparent addresses in a single transaction, because that would link them.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/1434

zcash_client_sqlite/src/wallet/db.rs
daira: You can add a constraint: `CONSTRAINT foo_is_boolean CHECK (foo = FALSE OR foo = TRUE)`.

`FALSE` and `TRUE` are synonyms for 0 and 1, but using them makes the intent clear, and the constraint enforces it.

```suggestion
        FALSE AS is_change,
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -525,7 +523,12 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
         account: AccountId,
         max_height: BlockHeight,
     ) -> Result<HashMap<TransparentAddress, NonNegativeAmount>, Self::Error> {
-        wallet::get_transparent_balances(self.conn.borrow(), &self.params, account, max_height)
+        wallet::transparent::get_transparent_address_balances(
```
daira, line 526: Why is this renaming the function in `wallet::transparent` so that it doesn't match the trait method? I think that for all of the other functions there that correspond to trait methods, the names do match.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2020,9 +1849,31 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+         SET block = NULL, mined_height = NULL, tx_index = NULL
+         WHERE mined_height > :height",
+        named_params![":height": u32::from(block_height)],
+    )?;
+
+    // If we're removing scanned blocks, we need to truncate the note commitment tree and remove
+    // affected block records from the database.
     if block_height < last_scanned_height {
         // Truncate the note commitment trees
         let mut wdb = WalletDb {
```
daira, line 1879: Aside: I don't understand why we're creating a new `WalletDb` wrapper when we could instead pass in `wdb: &mut WalletDb<SqlTransaction<'_>, P>`.

If I understand correctly, the point of the `SqlTransaction` wrapper is to mark that it was constructed by `WalletDb::transactionally`. This does several useful things:
* It guarantees that the `rusqlite::Transaction` has `DEFERRED` behaviour, because `WalletDb::transactionally` calls `self.conn.transaction()` which is implemented to call `rusqlite::Transaction::new(conn, TransactionBehavior::Deferred)`. The meaning of `DEFERRED` is that the transaction rolls back unless `commit()` is called explicitly (given that we never call `rusqlite::Connection::set_drop_behavior`). `WalletDb::transactionally` calls `commit()` iff the closure returns `Ok(_)`, which is the proper behaviour of a Rust database API that we should always want.
* It also guarantees that we can't accidentally commit earlier than we intended to. SQLite doesn't support nested transactions (it does support [savepoints](https://www.sqlite.org/lang_savepoint.html) which can be used to simulate them, but we don't use that). If some called function `f` that needs to be transactional were responsible for calling `commit()`, that would be a hazard because we couldn't reuse `f` in a context that also needs to be transactional: `f` would call `commit()` too early. This is avoided with the `transactionally` API because the transactional code never needs to call `commit()`.
  * Note that the fact that `rusqlite::Connection::transaction` takes a `&mut` reference does not prevent this hazard: it prevents *creating* nested transactions at the SQLite API layer (if you don't use the `Transaction::new_unchecked` escape hatch), but it can't prevent user code from committing too early.
* It enforces single-threaded use of the connection when it is in a transaction, which (even if SQLite didn't internally use locks in a way that effectively prevents any useful concurrency) is necessary for the transaction API to even make sense: you have to do all the things *and then* call `commit()`.

Here we don't call `commit()` in `truncate_to_height`, which is correct because we take a `rusqlite::Transaction` as argument, but that means:
* we don't know what the transaction behaviour is set to without tracing where `conn` came from;
* we don't have any guarantee that the caller has taken responsibility for calling `commit()`.

Please let's refactor (not in this PR) so that we never need to explicitly create a `WalletDb` wrapper like this outside `WalletDb::transactionally`.

zcash_client_sqlite/src/lib.rs
nuttycom: When moving the methods, I got confused about the meaning of `get_transparent_balances` vs `get_transparent_account_balance` and I wanted to clarify. I didn't want to change the top-level trait method, but since this is crate-private I thought that it was a good start.

zcash_client_sqlite/src/wallet.rs
nuttycom: `SqlTransaction` was added to the codebase much later on, to deal with some problems around how to implement the `WalletCommitmentTrees` trait. I can't recall all the problems that I ran into, but I remember spending several days and multiple pairings with @str4d figuring out an approach that could work here; there's some problem if the connection type is the same in `WalletWrite` as in `WalletCommitmentTrees.`

zcash_client_sqlite/src/lib.rs
daira: Do you want me to undo renaming it back?

zcash_client_sqlite/src/wallet.rs
daira: Filed #1435.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -216,54 +226,75 @@ CREATE TABLE orchard_received_note_spends (
...
+/// Stores the transparent outputs received by the wallet.
 ///
 /// Originally this table only stored the current UTXO set (as of latest refresh), and the
 /// table was cleared prior to loading in the latest UTXO set. We now upsert instead of
 /// insert into the database, meaning that spent outputs are left in the database. This
-/// makes it similar to the `*_received_notes` tables in that it can store history, but
-/// has several downsides:
-/// - The table has incomplete contents for recovered-from-seed wallets.
-/// - The table can have inconsistent contents for seeds loaded into multiple wallets
+/// makes it similar to the `*_received_notes` tables in that it can store history.
```
daira, line 234:
```suggestion
/// Originally this table (when it was named `utxos`) only stored the current UTXO set as
/// of the latest refresh, and the table was cleared prior to loading in the latest UTXO set.
/// We now upsert instead of insert into the database, meaning that spent outputs are left in
/// the database. This makes it similar to the `*_received_notes` tables in that it can store
/// history.
///

Comment (daira) [inline zcash_client_sqlite/src/wallet/db.rs:234]: "Originally this table, when it was called `utxos`, ..."

File: zcash_client_sqlite/src/wallet/db.rs [hunk 31]
```diff
@@ -497,219 +528,201 @@ pub(super) const TABLE_SQLITE_SEQUENCE: &str = "CREATE TABLE sqlite_sequence(nam
...
-            transactions.block              AS block,
-            transactions.txid               AS txid,
-            v_received_notes.pool           AS pool,
-            id_within_pool_table,
-            v_received_notes.value          AS value,
-            CASE
-                WHEN v_received_notes.is_change THEN 1
+    -- Outputs received in this transaction
+    SELECT ro.account_id              AS account_id,
+           transactions.mined_height  AS mined_height,
```
daira, line 604: Document this schema change in the changelog.

daira (approved):
utACK

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -32,3 +50,755 @@ pub(crate) fn detect_spending_accounts<'a>(
...
+    Ok(res)
+}
+
+#[tracing::instrument(skip(conn, account_balances))]
+pub(crate) fn add_transparent_account_balances(
+    conn: &rusqlite::Connection,
+    mempool_height: BlockHeight,
+    account_balances: &mut HashMap<AccountId, AccountBalance>,
+) -> Result<(), SqliteClientError> {
+    let transparent_trace = tracing::info_span!("stmt_transparent_balances").entered();
```
str4d, line 368: Now that this method is a separate function with a `#[tracing::instrument]` annotation, this manual span is unnecessary:
```suggestion
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -32,3 +50,755 @@ pub(crate) fn detect_spending_accounts<'a>(
...
+        "SELECT u.account_id, SUM(u.value_zat)
+         FROM transparent_received_outputs u
+         JOIN transactions t
+         ON t.id_tx = u.transaction_id
+         -- the transaction that created the output is mined or is definitely unexpired
+         WHERE (
+            t.mined_height < :mempool_height -- tx is mined
+            OR t.expiry_height = 0 -- tx will not expire
+            OR t.expiry_height >= :mempool_height
+         )
```
str4d, line 379: Observation: The other  minconf-0 instances of this pattern use `t.mined_height <= SOME_HEIGHT OR t.expiry_height >= SOME_HEIGHT`. I think it's okay to use `< :mempool_height` here however, solely because we never actually encounter the off-by-one here (because we encode mempool heights in the database as `NULL`).

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -32,3 +50,755 @@ pub(crate) fn detect_spending_accounts<'a>(
...
+        let value = NonNegativeAmount::from_nonnegative_i64(raw_value).map_err(|_| {
+            SqliteClientError::CorruptedData(format!("Negative UTXO value {:?}", raw_value))
+        })?;
+
+        account_balances
+            .entry(account)
+            .or_insert(AccountBalance::ZERO)
+            .add_unshielded_value(value)?;
+    }
+    drop(transparent_trace);
```
str4d, line 407: Ditto, replaced by `#[tracing::instrument]`:
```suggestion
```

str4d (approved):
utACK 10258b8443f92726bc0ce514e63898d8168e53fc with nit about tracing spans. I did not review the codebase outside of what is already changed in this PR to look for other places in e.g. the SQL queries that might need to be altered.

str4d (dismissed):
Re-utACK 5c8f585091160d76575b052a8de03ddd38efc05f

zcash_client_sqlite/src/lib.rs
nuttycom: I don't feel strongly about it.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Where we use this column, we actually use it as a count, so probably the better naming would be `change_count` or something: 
```        
SUM(notes.is_change) > 0     AS has_change,
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -497,219 +528,201 @@ pub(super) const TABLE_SQLITE_SEQUENCE: &str = "CREATE TABLE sqlite_sequence(nam
...
-        sent_notes.value             AS value,
-        0                            AS is_change,
-        sent_notes.memo              AS memo
+       sent_notes.output_pool       AS output_pool,
+       sent_notes.output_index      AS output_index,
+       sent_notes.from_account_id   AS from_account_id,
+       ro.account_id                AS to_account_id,
+       sent_notes.to_address        AS to_address,
+       sent_notes.value             AS value,
+       0                            AS is_change,
```
nuttycom, line 718: It just occurs to me, will this be correct behavior after ZIP 320 integration?

daira: I don't see why not, but we should test it. Note that `put_transparent_output` is called from `store_sent_tx` for ephemeral outputs.

str4d (approved):
utACK 9716617b558bec40f937fbe5c82b84ed3645e229

Issue #1408, opened 2024-06-12
`zcash_primitives::legacy::keys::AccountPrivKey::{to_bytes, from_bytes}` does not use the `xprv` encoding


str4d:
When we defined a wallet-internal-only serialization for Unified Spending Keys, we wanted the transparent key part to use the same byte encoding as the `xprv` format from BIP 32 (i.e. without Base58 encoding, and without the leading version bytes that act like a Bech32 HRP). In the PR that made the change, I asked this question during review (https://github.com/zcash/librustzcash/pull/625#discussion_r964233132) and was told that `hdwallet` provided that encoding. As part of working on #1407, I have discovered this to be false.

The bytes used by the `xprv` encoding are:

    1 byte: depth: 0x00 for master nodes, 0x01 for level-1 derived keys, ....
    4 bytes: the fingerprint of the parent's key (0x00000000 if master key)
    4 bytes: child number. This is ser32(i) for i in xi = xpar/i, with xi the key being serialized. (0x00000000 if master key)
    32 bytes: the chain code
    33 bytes: the public key or private key data (serP(K) for public keys, 0x00 || ser256(k) for private keys)

The `hdwallet::ExtendedPrivKey::serialize` encoding appears to be:

    32 bytes: the private key data (ser256(k))
    32 bytes: the chain code

This causes us a problem, because by omitting the extended metadata, we cannot parse an existing USK encoding with a transparent part into a `bip32::ExtendedPrivateKey` type (which requires the metadata).

str4d:
Some of the missing information can be reconstructed from contextual information:
- ✅ `depth` is always 3, as `AccountPrivKey` is derived using the path `m/44'/coin_type'/account'`.
- ❌ We cannot determine the parent key fingerprint correctly without re-deriving the parent from the seed.
- ❓`child number` is `account'` (I believe including the hardening flag), which we "know" in the places we currently use the USK encoding (specifically for derived accounts in `zcash_client_*`), but don't necessarily have at the spot we are doing the decoding (in particular for imported UFVKs, support for which was recently added to `zcash_client_*`).

[...]

Issue #1407, opened 2024-06-12
Migrate from `hdwallet` to `bip32`


str4d:
The `hdwallet` crate depends on `ring 0.16`, while most of the rest of the crates we depend on use `ring 0.17`. We have been waiting for eight months for `hdwallet` to merge the PR updating dependencies (https://github.com/jjyr/hdwallet/pull/21), without response.

I noticed that the only reason `hdwallet` depends on `ring` is for SHA-256 and HMAC. These could instead be sourced from the Rust Crypto crates `sha2` and `hmac`. While looking into that, I found the `bip32` crate, which provides a very similar API to `hwallet` but using pure-Rust crates (or optionally the `secp256k1` crate, which is what we want).

We should migrate from `hdwallet` to `bip32`, both to consolidate our dependencies and consolidate our dependency maintainers.

PR #1421, opened 2024-06-14
Migrate from `hdwallet` to `bip32`

str4d:
As part of this, we migrate to `secp256k1 0.27`. This version does not bump `secp256k1-sys`, so remains compatible with the `libsecp256k1` revision used in `zcashd`.

The `zcash_primitives::legacy::keys::AccountPrivKey` encoding also changes to preserve the transparent extended key metadata. Previously the type was documented as such, but only encoded the private key and chain code; the new encoding now matches the documentation. As a side effect, the unstable encoding of `zcash_keys::keys::UnifiedSpendingKey` also changes.

Closes zcash/librustzcash#1407.
Closes zcash/librustzcash#1408.

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -8,6 +8,9 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 ### Changed
 - MSRV is now 1.70.0.
+- `zcash_client_sqlite::error::SqliteClientError` has changed variants:
+  - Removed `HdwalletError`.
+  - Added `TransparentDerivation`.
 
 ## [0.10.3] - 2024-04-08
 
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -6,13 +6,39 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `zcash_primitives::legacy::keys`:
+  - `impl From<TransparentKeyScope> for bip32::ChildNumber`
+  - `impl From<NonHardenedChildIndex> for bip32::ChildNumber`
+  - `impl TryFrom<bip32::ChildNumber> for NonHardenedChildIndex`
+
 ### Changed
 - MSRV is now 1.70.0.
+- Bumped dependencies to `secp256k1 0.27`.
+- `zcash_primitives::legacy::keys`:
+  - `AccountPrivKey::{from_bytes, to_bytes}` now use the byte encoding from the
+    inside of a `xprv` Base58 string encoding from BIP 32, excluding the prefix
+    bytes (i.e. starting with `depth`).
+  - `AccountPrivKey::from_extended_privkey` now takes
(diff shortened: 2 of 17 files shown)


zcash_primitives/CHANGELOG.md
```diff
@@ -6,13 +6,39 @@ and this library adheres to Rust's notion of
...
+    bytes (i.e. starting with `depth`).
+  - `AccountPrivKey::from_extended_privkey` now takes
+    `bip32::ExtendedPrivateKey<secp256k1::SecretKey>`.
+  - The following methods now return `Result<_, bip32::Error>`:
+    - `AccountPrivKey::from_seed`
+    - `AccountPrivKey::derive_secret_key`
+    - `AccountPrivKey::derive_external_secret_key`
+    - `AccountPrivKey::derive_internal_secret_key`
+    - `AccountPubKey::derive_external_ivk`
+    - `AccountPubKey::derive_internal_ivk`
```
nuttycom, line 30: Would this be a good time to move the transparent keys to `zcash_keys`? I think that if we did so, `zcash_keys` would no longer need to depend upon `zcash_primitives`.

str4d: Hmm, I kinda see it, but also the other protocol-specific key types exist in their own crates, so `zcash-keys` may not be where we want this code.

str4d: It may instead make more sense to plan a move to some `zcash_transparent` crate that involves not just the key material, but the transparent builder logic etc. as well.

nuttycom: In some sense bip32 *is* that protocol-specific key derivation crate; we just have to add some more layers for Zcash, so I think this going in zcash_keys would be reasonable.

str4d: Moving these as-is may mean that `zcash_keys` no longer depends on `zcash_primitives`, but it would mean that `zcash_primitives` now needs to depends on `zcash_keys`, because the rest of the transparent logic is still there. So the move would not decouple the crates, it would just invert their relationship.

nuttycom (approved):
utACK f54ee4abde1ae3bc5ab187529f781294acaebdf1 - it might have been nice to have the bugfix in a separate commit, but this is fine.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 ### Changed
 - MSRV is now 1.70.0.
+- `zcash_client_sqlite::error::SqliteClientError` has changed variants:
+  - Removed `HdwalletError`.
+  - Added `TransparentDerivation`.
```
daira, line 13: Note that we also currently do an `.unwrap()` after calling `derive_secret_key` in `hdwallet`, in `zcash_client_backend` here:
https://github.com/zcash/librustzcash/blob/03fc64cfd03547483d448f60797545ce3ae5b147/zcash_client_backend/src/data_api/wallet.rs#L826-L829

It would be good to be able to report a useful error there, but `zcash_client_backend` has no error type to do so.

daira (approved):
utACK

PR #1422, opened 2024-06-14
zcash_client_backend: Add a Tor client that can fetch USD/ZEC exchange rates

### LINKED ISSUE #1416: zcash_client_backend: Add Arti-based Tor client for fetching USD/ZEC exchange rates
We want wallets to be able to fetch the latest USD/ZEC exchange rate without revealing their IP address to the data sources.

In testing, it's been possible to do this using Arti, which is now usable as a client library. We should add this to `zcash_client_backend` behind a feature flag, so it can be shared across any wallets that use the backend.

str4d:
Closes #1416.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3350,6 +3350,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "rust_decimal"
+version = "1.35.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1790d1c4c0ca81211399e0e0af16333276f375209e71a37b67698a373db5b47a"
+dependencies = [
+ "arrayvec",
+ "num-traits",
+ "serde",
+]
+
+[[package]]
 name = "rustc-demangle"
 version = "0.1.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5615,6 +5626,7 @@ dependencies = [
  "rand 0.8.5",
  "rand_core 0.6.4",
  "rayon",
+ "rust_decimal",
  "sapling-crypto",
  "secrecy",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1910,9 +1910,9 @@ dependencies = [
 
 [[package]]
 name = "hyper"
-version = "1.4.1"
+version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50dfd22e0e76d0f662d429a5f80fcaf3855009297eab6a0a9f8543834744ba05"
+checksum = "fe575dd17d0862a9a33781c8c4696a55c320909004a67a00fb286ba8b1bc496d"
 dependencies = [
  "bytes",
  "futures-channel",
@@ -1944,9 +1944,9 @@ dependencies = [
 
 [[package]]
 name = "hyper-util"
-version = "0.1.6"
+version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ab92f4f49ee4fb4f997c784b7a2e0fa70050211e0b6a287f898c3c9785ca956"
+checksum = "7b875924a60b96e5d7b9ae7b066540b1dd1cbd90d1828f54c92e02a283351c56"
 dependencies = [
  "bytes",
  "futures-channel",
@@ -4137,6 +4137,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "tokio-rustls"
+version = "0.24.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0d409377ff5b1e3ca6437aa86c1eb7d40c134bfec254e44c830defa92669db5"
+dependencies = [
+ "rustls",
+ "tokio",
+]
+
+[[package]]
 name = "tokio-stream"
 version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5590,6 +5600,9 @@ dependencies = [
  "group",
  "gumdrop",
  "hex",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
  "incrementalmerkletree",
  "jubjub",
  "memuse",
@@ -5599,18 +5612,23 @@ dependencies = [
  "percent-encoding",
  "proptest",
  "prost",
+ "rand 0.8.5",
  "rand_core 0.6.4",
  "rayon",
  "sapling-crypto",
  "secrecy",
+ "serde",
+ "serde_json",
  "shardtree",
  "subtle",
  "time",
  "tokio",
+ "tokio-rustls",
  "tonic",
  "tonic-build",
  "tor-rtcompat",
  "tracing",
+ "webpki-roots",
  "which",
  "zcash_address",
  "zcash_encoding",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,7 @@ dependencies = [
  "cfg-if",
  "cipher",
  "cpufeatures",
+ "zeroize",
 ]
 
 [[package]]
@@ -45,7 +46,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -65,6 +66,65 @@ name = "allocator-api2"
 version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
+
+[[package]]
+name = "amplify"
+version = "4.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e711289a6cb28171b4f0e6c8019c69ff9476050508dc082167575d458ff74d0"
+dependencies = [
+ "amplify_derive",
+ "amplify_num",
+ "ascii",
+ "wasm-bindgen",
+]
+
+[[package]]
+name = "amplify_derive"
+version = "4.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "759dcbfaf94d838367a86d493ec34ccc8aa6fe365cb7880d6bf89006de24d9c1"
+dependencies = [
+ "amplify_syn",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
+name = "amplify_num"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "04c009c5c4de814911b177e2ea59e4930bb918978ed3cce4900d846a6ceb0838"
+dependencies = [
+ "wasm-bindgen",
+]
+
+[[package]]
+name = "amplify_syn"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7736fb8d473c0d83098b5bac44df6a561e20470375cd8bcae30516dc889fd62a"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "anes"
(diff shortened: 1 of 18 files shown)


Cargo.toml
```diff
@@ -113,6 +132,13 @@ proptest = "1"
 rand_chacha = "0.3"
 rand_xorshift = "0.3"
 
+# Tor
+# - `arti-client` depends on `rusqlite`, and a version mismatch there causes a compilation
+#   failure due to incompatible `libsqlite3-sys` versions.
+arti-client = { version = "0.11", default-features = false, features = ["compression", "rustls", "tokio"] }
```
str4d, line 138: After #1421 we are no longer constrained by `ring`, but we are still constrained here by `rusqlite` which is why I only bumped this from `0.9` to `0.11`, not to the latest `0.19`. We should separately look at bumping `rusqlite`, subject to the CocoaPods issue above.

str4d: Ugh, we can't actually avoid cross-version `ring` dependencies, because `tor-rtcompat` depends on both a `rustls` version that uses `ring 0.17`, and the `x509-signature` crate that uses `ring 0.16`. We'd need to use the older `arti-client` version etc. that uses the older `rustls`. Given that the most recent release of Arti is also still in this state, I'm inclined to just leave it as-is (we're at least down to a single version of `rustls`) and see how it goes.

str4d: I confirmed that this still happens to work on iOS, so I think we should just leave it as-is for now, and if I have time I'll try to contribute a change to Arti that migrates away from `x509-signature`.

zcash_client_backend/src/tor/http/exchange_rate.rs
```diff
@@ -0,0 +1,150 @@
...
+            )
+            .map(|res| res.map(|resp| resp.into_body()));
+
+        let coinbase_ticker = self
+            .get_json::<CoinbaseTicker>(
+                "https://api.exchange.coinbase.com/products/ZEC-USD/ticker"
+                    .parse()
+                    .unwrap(),
+            )
+            .map(|res| res.map(|resp| resp.into_body()));
```
str4d, line 34: I have yet to see this ever succeed; so far in testing it always returns 403 Forbidden. Coinbase must be fetching a list of Tor exit nodes to block. I haven't checked whether providing an API key enables bypassing this.

If we can find an alternate ticket to use, we may want to consider dropping the Coinbase source.

zcash_client_backend/src/tor/http/exchange_rate.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        let gemini_ticker = self
+            .get_json::<GeminiTicker>("https://api.gemini.com/v2/ticker/zecusd".parse().unwrap())
+            .map(|res| res.map(|resp| resp.into_body()));
```
str4d, line 38: I have now observed Gemini returning 451 Unavailable for Legal Reasons. I've only seen it once though; I'm guessing a specific Tor exit node was blocked, and on a subsequent query the circuit used a different exit.

Cargo.lock
```diff
@@ -194,9 +363,9 @@ dependencies = [
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
- "http",
- "http-body",
- "hyper",
+ "http 0.2.12",
+ "http-body 0.4.6",
+ "hyper 0.14.28",
```
str4d, line 368: The upcoming `tonic 0.12` migrates to `http 1`, so once that is released almost all of these particular duplicate dependencies will disappear (except for one crate that Arti depends on at the version we're running).

zcash_client_backend/src/tor/http/exchange_rate.rs
str4d: I've replaced Coinbase with MEXC; they appear to have similar volumes over the last 24 hours.

str4d: I've now re-added Coinbase to the mix; we'll query them along with every other data source, just in case they start working.

zcash_client_backend/src/tor/http/exchange_rate.rs
str4d: It's still the most reliable data source, so Gemini is now excluded from being ignored.

zcash_client_backend/src/tor.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        let mut config_builder = TorClientConfigBuilder::from_directories(
+            tor_dir.join("arti-data"),
+            tor_dir.join("arti-cache"),
+        );
+        config_builder
+            .storage()
+            .permissions()
+            .dangerously_trust_everyone();
```
nuttycom, line 38: Can you leave a comment explaining why this option is chosen?

zcash_client_backend/src/tor.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        Ok(Self { inner })
+    }
+}
+
+/// Errors that can occur while creating or using a Tor [`Client`].
+#[derive(Debug)]
+pub enum Error {
+    /// The directory passed to [`Client::create`] does not exist.
+    MissingTorDirectory,
```
nuttycom, line 57: Does it make sense to return the path?

zcash_client_backend/src/tor/http.rs
```diff
@@ -0,0 +1,207 @@
...
+    // Spawn a task to poll the connection and drive the HTTP state.
+    PreferredRuntime::current()?
+        .spawn(async move {
+            if let Err(e) = connection.await {
+                error!("Connection failed: {}", e);
+            }
+        })
+        .map_err(HttpError::from)?;
+
+    let req = h(Request::builder()
```
nuttycom, line 115: I don't understand the reason for using a callback here. Also, the `.body` call explicitly passing the empty bytes value seems potentially restrictive. 

Is it not possible to just take a `Request` value and the validate it before sending?

zcash_client_backend/src/tor/http/exchange_rate.rs
```diff
@@ -0,0 +1,427 @@
...
+
+        // Fetch the data in parallel.
+        let res = join!(
+            Binance::query(self, pair),
+            Coinbase::query(self, pair),
+            GateIo::query(self, pair),
+            Gemini::query(self, pair),
+            KuCoin::query(self, pair),
+            Mexc::query(self, pair),
+        );
```
nuttycom, line 35: Why not have a trait that defines the `query` method, so that users of the library can provide their own currency retrieval? Hardcoding this set of endpoints seems pretty limiting.

nuttycom (approved):
utACK 361653b55ea6366cbe1bbfa002f68cd37736bede

I'm surprised that `exchange_rate.rs` hardcodes the set of permitted handlers. That seems pretty inflexible for third-party users.

zcash_client_backend/src/tor.rs
str4d: I don't think so? The caller directly provided the path as the only argument, and therefore has it. If they want to augment the error response with that path as context, they can do so.

The other direction we could go here is to just create the folder ourselves, but I feel like it should be a conscious choice of the caller to use a particular directory, especially given that there's no difference between creating a new Arti client, and loading an existing one.

zcash_client_backend/src/tor/http.rs
str4d: The callback was an easy way for `get_json` to be self-contained with only JSON-relevant details.

`.body` passing empty bytes is indeed restrictive, but it is sufficient for what we need, and the methods here are all private so we can change these details at any time. I don't know `hyper` well enough to know what the best API is for supporting generic bodies (this changed between `hyper 0.*` and `hyper 1.*`, so my prior experience doesn't help), and I didnd't have time to figure it out.

zcash_client_backend/src/tor.rs
str4d: Oops, this _was_ commented on, but in a PR comment on the Swift FFI PR before I extracted the changes into this PR. https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/142#discussion_r1628700953
> The permissions here need testing. I was having trouble getting Arti to work in the iOS simulator, and one of the things I did to get around it was enable this setting. I eventually also figured out that on the Swift side we needed to pass over `URL.path` instead of `URL.absoluteString`, but I haven't gone back and experimented with other permissions here yet.

zcash_client_backend/src/tor/http/exchange_rate.rs
str4d: My goal here was for one new API that was straightforward to use, and then we could expand later as needed. This part has gone through multiple complete redesigns, and as of late Friday evening was at a point where this might be feasible (it only just gained API support for multi-currency, though I could not enable it as almost every exchange only offers `USD(T)/ZEC`).

I don't think a `query` trait is sufficient on its own, because an important part of this API is how the individual query results are combined, and I do not want that to be up to the caller. The caller would instead need a way to provide a heterogeneous set of queryable exchanges over which we can operate, which gets into "implement this trait for all tuple lengths up to 32" territory. I can do that, but not in the currently-available time, and it would make the API more complex.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/361653b55ea6366cbe1bbfa002f68cd37736bede..426e29eae4a2ed1b4cad144307fc42707bcffc04) to fix merge conflict in changelog.

zcash_client_backend/src/tor.rs
str4d: I've now tested this branch on both Android and iOS without this setting, and they both appear to work. I'll remove it.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/426e29eae4a2ed1b4cad144307fc42707bcffc04..ced6316d3017ebbd0bc527dd08487f1764a234f7) to remove unnecessary `tor_dir` permissions changes, and address CI lints.

nuttycom (dismissed):
re-utACK ced6316d3017ebbd0bc527dd08487f1764a234f7

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/ced6316d3017ebbd0bc527dd08487f1764a234f7..ee6902b5346d080ffa3a72fe4c7979077e0ae2e5) to not pass `ExchangePair` through to the individual queriers. Only two exchanges offer actual `USD/ZEC` pairs (the rest are `USDT/ZEC`), and even if we could get data for e.g. `GBP/ZEC` that wouldn't help with arbitrary fiat currencies.

Instead, we will always query for `USD(T)/ZEC`, and then use a forex rate for `TARGET/USD` to adjust, once I can find a suitable free source of forex data.

zcash_client_backend/src/tor/http/exchange_rate.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Fetches the latest `currency`/ZEC exchange rate, derived from several exchanges.
+    ///
+    /// Supported currencies:
+    /// - USD
+    ///
+    /// Returns:
+    /// - `Ok(Some(rate))` if at least one exchange request succeeds.
+    /// - `Ok(None)` if the given currency is unsupported.
+    /// - `Err(_)` if none of the exchange queries succeed.
+    pub async fn get_exchange_rate(&self, currency: Currency) -> Result<Option<Decimal>, Error> {
```
str4d, line 21: Hmm, now that I'm always fetching `USD(T)/ZEC`, maybe this API should go back to being single-currency:
```suggestion
    pub async fn get_latest_zec_to_usd_rate(&self) -> Result<Decimal, Error> {
```
and then if the caller wants something other than `USD/ZEC`, it can join this itself with a call to a (to-be-implemented) `get_latest_usd_rate(&self, currency: Currency)` function? That would keep the inner implementation of this method simpler.

str4d: @nuttycom I have this change implemented locally; if you agree with it, I'll force-push the last commit.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/ee6902b5346d080ffa3a72fe4c7979077e0ae2e5..8fb7d772a482e1bf4661f461bb51573424e76263) after TEX address support landed.

zcash_client_backend/src/tor/http/exchange_rate.rs
nuttycom: I'm somewhat indifferent, as I don't think that this should be a concrete method on `Client` but should instead be a trait method that allows others to implement their own preferences for what data source(s) to connect to. IMO without that sort of extensibility available to the third-party users of the crate, this functionality is basically a proof of concept; anyone wishing to use it will likely end up forking.

zcash_client_backend/src/tor/http/exchange_rate.rs
```diff
@@ -0,0 +1,373 @@
...
+        let pair = match ExchangePair::get(currency) {
+            Some(pair) => pair,
+            None => return Ok(None),
+        };
+        assert!(matches!(pair, ExchangePair::Usd));
+
+        // Fetch the data in parallel.
+        let res = join!(
```
nuttycom, line 29: I somewhat object that the user of this library has no option to exclude any of these data sources from consideration.

zcash_client_backend/src/tor/http/exchange_rate.rs
```diff
@@ -0,0 +1,373 @@
...
+            assert!(prices.len() % 2 != 0);
+            prices.sort();
+            let median = prices.len() / 2;
+            Ok(Some(prices[median]))
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+enum ExchangePair {
```
nuttycom, line 89: I'm skeptical of this type. Why isn't it effectively a tuple of `Currency` values, so that it can represent both halves of the exchange rate? `ExchangePair` is a misleading name for it.

nuttycom (changes requested):
I think that the implementation of `get_exchange_rate` is excessively restrictive - it does not permit either the exclusion of data providers or the addition of providers without forking the crate. If the API is released as-is, it will require a breaking API change to remedy that, and I don't see a good reason for not adding the necessary abstraction layer when introducing the API to begin with.

zcash_client_backend/src/tor/http/exchange_rate.rs
str4d: This type is left over from when I mistakenly though that there would be _any_ exchange data sources that reliably gave ZEC pairs in anything other than USD(T). It is deleted if I make the change to `get_latest_zec_to_usd_rate` I suggested above.

str4d: Also, we _do not want_ an API that allows for arbitrary cryptocurrency exchange rate lookups; this API is specifically for looking up ZEC/`something` pairs, and if I make the change per above, then specifically ZEC/USD pairs.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/8fb7d772a482e1bf4661f461bb51573424e76263..675490e9a02f672cacc235829842e030a2965400) to rework the exchange rate API to allow callers to customise the exchanges (or define their own data sources).

zcash_client_backend/src/tor/http/cryptex.rs
```diff
@@ -0,0 +1,174 @@
...
+        trace!(?res, "Data results");
+        let (trusted_res, other_res) = res;
+
+        // Split into successful queries and errors.
+        fn split(s: &mut Vec<Decimal>, e: &mut Vec<Error>, res: Result<ExchangeData, Error>) {
+            match res {
+                Ok(d) => s.push(d.exchange_rate()),
+                Err(error) => e.push(error),
+            }
+        }
```
nuttycom, line 137: nonblocking nit: this could now be inlined, and doesn't need to be a separate function

zcash_client_backend/src/tor/http/cryptex.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut rates: Vec<Decimal> = vec![];
+        let mut errors = vec![];
+        for res in other_res {
+            split(&mut rates, &mut errors, res);
+        }
+
+        // "Never go to sea with two chronometers; take one or three."
+        // Randomly drop one rate if necessary to have an odd number of rates.
```
nuttycom, line 145: Should an error be returned if exactly two successful responses are returned? If only one successful response is returned, should it be used if it's not the trusted exchange? Should we even permit using only a single source of truth?

nuttycom (commented):
utACK 675490e9a02f672cacc235829842e030a2965400 with nonblocking nit & questions.

zcash_client_backend/src/tor/http/cryptex.rs
str4d: Now that we allow the caller to configure the exchanges, we _have_ to permit a single source of truth, because the builder allows them to construct this (it requires 1 but does not require others, and the one required source may be set to a source of data that they themselves control). However, we might want to prohibit configuring only two sources of truth: if any other source is provided, require providing at least two other sources. That means making the builder fallible in a way that would force almost every normal usage to `unwrap`, which feels wrong.

If exactly two successful responses are returned, _and we have more than two sources_, then I would be instead inclined to retry the others if there's a chance it was a transient failure. Determining what is a transient failure is hard however, because "the exit node used for this specific query is blocked / rate limited" is in our case a transient failure, whereas "all Tor exit nodes are blocked" is not.

nuttycom: Okay, I agree that since we only require one source to be configured, we should allow single-source responses. I think that we shouldn't contort the API to prohibit configuring only two sources of truth; however, maybe if three or more sources are configured, then we should require at least three responses, and return an error otherwise?

str4d: > maybe if three or more sources are configured, then we should require at least three responses, and return an error otherwise?

At least 3 responses, or should 1 also be acceptable and we only error on 2?

For this PR I'm not going to stick in retries, but we can add that later if we determine that "error on 2" is causing too many problems.

> It's still not clear to me what the behavior should be if two sources are configured; probably take the trusted source if available, and treat the second source as a backup?

That seems reasonable and comprehensible.

nuttycom: If 3 are configured, require 3 is best I think. If you're content with a single source, just configure 2. It's a small hole in the API, but I think it's sensible.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/675490e9a02f672cacc235829842e030a2965400..ac5bac9969ff3890240a281ac784bb705af7019f?w=1) to address review comments.

nuttycom (approved):
re-utACK ac5bac9969ff3890240a281ac784bb705af7019f

Issue #1075, opened 2023-12-24
Design for importing FVKs, IVKs, and even Unified spending keys


AArnott:
As [discussed on discord](https://discord.com/channels/809218587167293450/972649509651906711/1186343152286707784), use cases exist for importing accounts with spending or viewing keys into a Zcash application (e.g. a wallet, or POS machine) that are unrelated to a seed phrase, or possibly related to a seed phrase that does not equal the seed phrase the wallet was initialized with.

Some example use cases:
1. **Importing a UFVK** so a user who wants to monitor a cold wallet via a full viewing key from his/her hot wallet app.
2. **Multiple seed phrases** so a user that already has multiple seed phrases across various wallets can control them all from one app.
3. **Import unified spending keys** for an enterprise can create a seed phrase which the CFO keeps confidential, but uses the seed phrase to create accounts that he/she assigns to various departments or individual stores. These accounts would be shared via unified spending key so as to keep the seed phrase confidential.
4. **Importing an IVK** at a Point of Sale machine at a store needs to generate diversified addresses and see incoming transactions that it can associated with those generated addresses so the clerk can confirm products are paid for.

The `zcash_client_sqlite` crate's table schema assumes exactly one seed phrase, and all accounts are spending accounts based on that seed phrase. According to the discord discussion, this assumption may have impacted the design and code of the `zcash_client_backend` crate as well. 

Let's start the discussion by identifying which of the above use cases *may* fit into these crates named above, and which ones are definitely out of scope such that that use case can only be addressed by forking or relying only upon the lower-level crates and recreating some of the same functionality that the `zcash_client_*` crates have.

nuttycom:
Related to Electric-Coin-Company/zashi#55

AArnott:
I'm developing a change to send as a pull request that will add the ability to import an account based on its UFVK.

PR #1425, opened 2024-06-17
Add support for importing accounts by UFVK

AArnott:
Add `import_account_hd` and `import_account_ufvk` methods to the `WalletWrite` trait and implements them for the `zcash_client_sqlite` crate.

Also fix `wallet::add_account` to not fail when the UFVK contains fewer keys than librustzcash was compiled to support. For example, a UFVK that lacks a transparent key will no longer be rejected.

Partially fulfills #1075

## Open questions

- [x] The `spending_key_available` parameter declared by `import_account_ufvk` function isn't used at present. It's a future-proofing idea for an optimization that we'll surely want to leverage down the road. The open question is should we go to the trouble to _record_ its value in the `accounts` table _now_, or only when we do the additional work to actually optimize that path?
- [x] Eventually I'd like to add an API to import an account with only an UIVK. With that in mind, should we add _another_ function later to import such an account, or should we promote the sqlite crate's `ViewingKey` enum for use by a more general import account function that takes the `ViewingKey` enum? Consider also that we may want the ability to import accounts by pool-specific keys (e.g. transparent private key, sapling private key, sapling viewing key, etc.) eventually too.
- [x] Not entirely related to this PR, but it came up in my review: when `zcash_client_sqlite::wallet::add_account` rewrites scan ranges for the new account, could it possibly make the range to scan _smaller_ than it already was? Say, for example, one account was fully synced. Then two new accounts were added together, the first with an old birthday and the second with a young birthday height. Would the older birthday with the broader scan range survive such that the next sync would scan enough blocks for the longer of the two?
- [ ] Whichever merges later (this or #1135), the error raised by code added in this PR if no UA can be created should be updated to _not_ return a code indicating that a shielded receiver is required.

## TODO

- Add tests
  - [x] Success: importing an account with a UFVK that does not conflict
  - [x] Failure: importing an account with an index that is already defined in the database.
  - [x] Failure: importing an account with a viewing key that collides with an existing view-only account
  - [x] Failure: importing an account with a viewing key that collides with an existing HD derived account


AArnott:
I've manually tested this in my app and it works. I'd prefer to add unit tests to this PR only after I get a thumbs up on the API changes so that I know I'm testing the right thing.

zcash_client_backend/src/data_api.rs
```diff
@@ -1468,6 +1471,72 @@ pub trait WalletWrite: WalletRead {
...
+    ///
+    /// Panics if the length of the seed is not between 32 and 252 bytes inclusive.
+    ///
+    /// [ZIP 316]: https://zips.z.cash/zip-0316
+    fn import_account_hd(
+        &mut self,
+        seed: &SecretVec<u8>,
+        account_index: zip32::AccountId,
+        birthday: &AccountBirthday,
+    ) -> Result<(Self::Account, UnifiedSpendingKey), Self::Error>;
```
nuttycom, line 1511: While I think that UFVK import is a necessary feature, I'm much less convinced about multi-seed support. The whole idea of a seed phrase is that there's a single piece of information that is needed in order to provide recoverability for the funds in a wallet. On the other hand, I guess that this can be useful for recovery scenarios, but I really don't care much for the idea of a wallet having to manage multiple spend authority roots.

zcash_client_sqlite/src/error.rs
```diff
@@ -72,6 +73,9 @@ pub enum SqliteClientError {
     /// The account for which information was requested does not belong to the wallet.
     AccountUnknown,
 
+    /// The account being added collides with an existing account in the wallet with the given ID.
+    AccountCollision(AccountId),
+
```
nuttycom, line 78: I don't see where errors of this type are being created, and it doesn't make sense to me; account IDs are opaque autoincrementing values, there should never be a collision between them.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -218,6 +218,27 @@ impl Account {
...
+        let uivk = self.viewing_key.uivk();
+
+        #[cfg(feature = "orchard")]
+        let has_orchard = uivk.orchard().is_some();
+        #[cfg(not(feature = "orchard"))]
+        let has_orchard = false;
+
+        let has_sapling = uivk.sapling().is_some();
+        let has_p2pkh = uivk.transparent().is_some();
+
```
nuttycom, line 235: `Account` seems like the wrong place for this behavior. I think that a method `UnifiedIncomingViewingKey::to_address_request` would be better; it's not clear how this is intended to be used without going and looking at the call site.

nuttycom (changes requested):
Reviewed a273970b78b6e51c9abf2a47257f73b98212abd2. While I'm a bit leery about multi-seed support, it might be okay for a low-level API like this one. The removal of the unused enum variant and rework of the `intersect_ua_request` method are blocking.

zcash_client_backend/src/data_api.rs
nuttycom: Is `AccountCollision` intended for the case where `create_account` attempts to create an account for a seed and zip32 account ID that were previously created via `import_account_hd`?

AArnott: I don't think `import_account_id` introduced multi-seed support. We did that a while back when seed fingerprinting was added. 
The new thing that this method adds is direct control over the account index. It doesn't have to be a blind +1 over whatever the highest index was before. The app can actually be aware of and control it. And while users will typically be blissfully unaware of it, from an API level I strongly believe this should be exposed. 
In my .NET library, I expose a `ZcashAccount` type that you initialize with a seed and index (or a viewing key) and you can pass that around. When it's time to hook it up to a lightwallet to start syncing, I just need to be able to get that account created, even if the index is 5 and 4 other accounts weren't created first.

`AccountCollision` is for a couple cases: a colliding account index, or importing an account by viewing key that happens to collide with pre-existing account (which might be another viewing key imported account or an HD derived account).

zcash_client_sqlite/src/error.rs
AArnott: This isn't about an account ID colliding. It's about the _account_ colliding, as detectable by (seed fingerprint, index) or viewing key collisions.

The `AccountId` provided as an argument _is_ the opaque account ID value, and is provided so that the caller can match the account up with one shown in the app, so the app can tell the user "oops, you already imported this account. It's _this one_.".

zcash_client_sqlite/src/wallet.rs
AArnott: I wasn't confident about where this fn should go either. Happy to move it.

zcash_client_backend/src/data_api.rs
AArnott: And I have my own personal use case for explicit account indexes: for my seed phrase, account indexes 0-2 are used and _very_ old. I don't want to sync them _at all_ going forward. I want to keep my seed phrase by using index 3 for a new account. Given a new wallet, I can see myself using an account import feature, specifying my seed phrase and account index 3 as the _starting_ index, with a much higher birthday height. No reason to sync 0-2 at all.

zcash_client_sqlite/src/wallet.rs
AArnott: I've removed it as you suggested it.

zcash_client_sqlite/src/error.rs
AArnott: I will add code that returns this error as part of adding tests that exercise the failure cases

zcash_client_backend/src/data_api.rs
nuttycom: I think that adding an explicit `sunset_height` to the `accounts` table would be a good way to serve this. Sunsetted accounts won't have their keys used for scanning.

zcash_client_sqlite/src/error.rs
nuttycom: Can you add the code to detect this collision in the implementation of account insertion, and return it from there? My primary complaint was that the enum variant was added without it ever being created. If the check for collision were implemented, I would have been better able to understand its use. The comment here should also be expanded to describe under what conditions (fingerprint/zip32 index, viewing key conflict) the collisions can happen.

In the case of viewing key import, what happens if you attempt to import an FVK where you currently have an IVK for the same account? The reverse should be a collision, but does FVK import constitute an "upgrade" that is allowed to silently occur?

AArnott: I just pushed the first few tests, and actual use of the `AccountCollision` variant. I'll add another test or two later today.

As for IVK/FVK differences, I agree with you. But the table doesn't store IVKs at all at the moment, so when we add support for IVKs, we should include handling of IVK collisions and upgrades.

zcash_client_backend/src/data_api.rs
AArnott: Yes, if that feature existed, that might work. Still, IMO limiting the API in this way is tying people's hands in order to try to force a particular wallet UX. I see account management as something that the app (or a higher level library as is in my case) to be something that can and should occur separately from the engine that does sync. My mental model is that I jump across the FFI layer to sync the blockchain and create transactions. The fact that it tracks accounts is only nice if it _helps_ me. 
But I guess I can always just use the UFVK import function and bypass seeds and account indexes at this layer entirely so that I can manage them at my preferred level. So if you feel strongly against this API I'll remove it.

AArnott:
I rebased my commits and removed the merges (which brought back merge conflicts) so that I can use these same commits in my fork without bringing in the #1431 regression. Once approved, I can do a final merge with main to resolve conflicts.

zcash_client_backend/src/data_api.rs
nuttycom: After some additional thought, I'm okay with this API. Also, I agree that returning `Self::Account` instead of `Self::AccountId` is a nicer API, and would support changing the existing method (in a separate PR).

nuttycom (commented):
utACK modulo that this will need re-review after merge conflicts are addressed.

AArnott:
Merge pushed.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -375,40 +379,63 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
-        |row| Ok(AccountId(row.get(0)?)),
-    )?;
+        .map_err(|e| match e {
+            rusqlite::Error::SqliteFailure(f, s)
+                if f.code == rusqlite::ErrorCode::ConstraintViolation =>
+            {
+                // An account conflict occurred.
+                // Make a best effort to determine the AccountId of the pre-existing row
+                // and provide that to our caller.
+                if s.clone().is_some_and(|s| s.contains(".ufvk")) {
```
nuttycom, line 425: This check is incompatible with test or regtest keys; instead, one would want to take the prefix of the serialized key up to the character `1` . Also, this depends upon the stability of the text encoding of rusqlite error messages, which I'm not comfortable with; instead, simply query for the ufvk and if no results are returned, then return the original error.

nuttycom (changes requested):
One minor blocking comment related to detection of the `AccountCollision` error.

zcash_client_sqlite/src/wallet.rs
AArnott: I don't know how this fails with test or regtest keys. The substring I'm testing for here is for the name of the column in the table which violated the constraint. So as long as the error mentions the column, it should continue to work. 
But if you want me to drop that check and just go straight to the next level check that queries the db to see if there is a conflicting ufvk (which I already do next), I can drop this first check. It was just an attempt to avoid another db hit when possible.

nuttycom: Yeah, please go ahead and drop the check; this should be a rare error case anyway, and I'd prefer not to rely on error strings that we don't control.

AArnott: Done.

nuttycom (dismissed):
utACK 36b3fc866adde5038cf677062abd99476c88d909

nuttycom:
There's a clippy lint that needs to be fixed (just a duplicate import); also, please check the intra-doc linking error.

nuttycom:
Also it appears that there are some test failures that need to be resolved.

AArnott:
I'll take a look. I ran `cargo test` at the root and that passed, but I'll see what the command line is to run tests that are failing. And run clippy.

AArnott:
It looks like the test failure was really just a compile failure due to the duplicate import.

nuttycom:
> I'll take a look. I ran `cargo test` at the root and that passed, but I'll see what the command line is to run tests that are failing. And run clippy.

Ah, yeah, usually you want `cargo test --release --all-features` for the final test run.

nuttycom (approved):
re-utACK fa5248def04a57baa6a5c67b5eb2bb4e6a774d3a

zcash_client_sqlite/src/wallet.rs
```diff
@@ -202,9 +202,13 @@ impl Account {
         &self,
         request: UnifiedAddressRequest,
     ) -> Result<(UnifiedAddress, DiversifierIndex), AddressGenerationError> {
+        self.uivk().default_address(request)
+    }
+
+    pub(crate) fn uivk(&self) -> UnifiedIncomingViewingKey {
```
str4d, line 208: Delete this method; it is entirely duplicative of the existing `uivk()` method that `Account` has via its implementation of `zcash_client_backend::data_api::Account`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -202,9 +202,13 @@ impl Account {
         &self,
         request: UnifiedAddressRequest,
     ) -> Result<(UnifiedAddress, DiversifierIndex), AddressGenerationError> {
+        self.uivk().default_address(request)
```
str4d, line 205: This refactor of `Account::default_address` OTOH is fine; its previous internals were somewhat duplicative and this ends up cleaner.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -510,14 +535,25 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
 
     // Initialize the `ephemeral_addresses` table.
     #[cfg(feature = "transparent-inputs")]
     transparent::ephemeral::init_account(conn, params, account_id)?;
 
-    Ok(account_id)
+    // Initialize the `ephemeral_addresses` table.
+    #[cfg(feature = "transparent-inputs")]
+    transparent::ephemeral::init_account(conn, params, account_id)?;
+
```
str4d, line 555: Bug: this has been duplicated, which will cause the ephemeral addresses table to be initialized twice for the account. I presume this was introduced by an incorrect merge conflict resolution.
```suggestion
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -510,14 +535,25 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     }
 
     // Always derive the default Unified Address for the account.
```
str4d, line 537:
```suggestion
    // Always derive the default Unified Address for the account. If the account's viewing
    // key has fewer components than the wallet supports (most likely due to this being an
    // imported viewing key), derive an address containing the common subset of receivers.
```

zcash_client_sqlite/src/testing.rs
```diff
@@ -211,6 +216,12 @@ impl<Cache> TestBuilder<Cache> {
         self
     }
 
+    pub(crate) fn with_account_having_index(mut self, index: zip32::AccountId) -> Self {
```
str4d, line 219: This method name is misleading: it only works if either `with_account_having_current_birthday` or `with_account_from_sapling_activation` was called, and its effect is to _modify_ the account index of the account that those methods would create. Rename the method to something clearer and document it.

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+    pub(crate) fn import_account_hd_1_then_2() {
+        let mut st = TestBuilder::new().build();
+
+        let birthday = AccountBirthday::from_parts(
+            ChainState::empty(st.wallet().params.sapling.unwrap() - 1, BlockHash([0; 32])),
+            None,
+        );
+
+        let seed = Secret::new(vec![0u8; 32]);
+        let zip32_index = zip32::AccountId::ZERO.next().unwrap();
```
str4d, line 2098:
```suggestion
        let zip32_index_1 = zip32::AccountId::ZERO.next().unwrap();
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+
+        let first = st
+            .wallet_mut()
+            .import_account_hd(&seed, zip32_index, &birthday)
```
str4d, line 2102:
```suggestion
            .import_account_hd(&seed, zip32_index_1, &birthday)
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+            .unwrap();
+        assert_matches!(
+            first.0.source(),
+            AccountSource::Derived { seed_fingerprint: _, account_index } if account_index == zip32_index);
```
str4d, line 2106:
```suggestion
            AccountSource::Derived { seed_fingerprint: _, account_index } if account_index == zip32_index_1);
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+
+        let second = st
```
str4d, line 2108:
```suggestion
        let zip32_index_2 = zip32_index_1.next().unwrap();
        let second = st
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+            .wallet_mut()
+            .import_account_hd(&seed, zip32_index.next().unwrap(), &birthday)
```
str4d, line 2110:
```suggestion
            .import_account_hd(&seed, zip32_index_2, &birthday)
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+            .unwrap();
+        assert_matches!(
+            second.0.source(),
+            AccountSource::Derived { seed_fingerprint: _, account_index } if account_index == zip32_index.next().unwrap());
```
str4d, line 2114:
```suggestion
            AccountSource::Derived { seed_fingerprint: _, account_index } if account_index == zip32_index_2);
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+    pub(crate) fn import_account_hd_1_twice() {
+        let mut st = TestBuilder::new().build();
+
+        let birthday = AccountBirthday::from_parts(
+            ChainState::empty(st.wallet().params.sapling.unwrap() - 1, BlockHash([0; 32])),
+            None,
+        );
+
+        let seed = Secret::new(vec![0u8; 32]);
+        let zip32_index = zip32::AccountId::ZERO.next().unwrap();
```
str4d, line 2127:
```suggestion
        let zip32_index_1 = zip32::AccountId::ZERO.next().unwrap();
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+
+        let first = st
+            .wallet_mut()
+            .import_account_hd(&seed, zip32_index, &birthday)
```
str4d, line 2131:
```suggestion
            .import_account_hd(&seed, zip32_index_1, &birthday)
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+            .unwrap();
+
+        assert_matches!(
+            st.wallet_mut().import_account_hd(&seed, zip32_index, &birthday),
```
str4d, line 2135:
```suggestion
            st.wallet_mut().import_account_hd(&seed, zip32_index_1, &birthday),
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -2020,6 +2074,111 @@ mod tests {
...
+        let ufvk = usk.to_unified_full_viewing_key();
+
+        let account = st
+            .wallet_mut()
+            .import_account_ufvk(&ufvk, &birthday, true)
+            .unwrap();
+        assert_eq!(
+            ufvk.encode(&st.wallet().params),
+            account.ufvk().unwrap().encode(&st.wallet().params)
+        );
```
str4d, line 2160: Check that the resulting `account` is imported, not derived (like the earlier tests check the reverse).

str4d (changes requested):
Reviewed fa5248def04a57baa6a5c67b5eb2bb4e6a774d3a.

zcash_client_sqlite/src/lib.rs
```diff
@@ -611,7 +611,58 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+
+            Ok((account, usk))
+        })
+    }
+
+    fn import_account_ufvk(
+        &mut self,
+        ufvk: &UnifiedFullViewingKey,
+        birthday: &AccountBirthday,
+        _spending_key_available: bool,
```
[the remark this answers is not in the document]
str4d: This corresponds to one of @AArnott's open questions:
> - The `spending_key_available` parameter declared by `import_account_ufvk` function isn't used at present. It's a future-proofing idea for an optimization that we'll surely want to leverage down the road. The open question is should we go to the trouble to _record_ its value in the `accounts` table _now_, or only when we do the additional work to actually optimize that path?

My inclination is towards recording its value now.

str4d:
Re: the other open questions:

> - Eventually I'd like to add an API to import an account with only an UIVK. With that in mind, should we add _another_ function later to import such an account, or should we promote the sqlite crate's `ViewingKey` enum for use by a more general import account function that takes the `ViewingKey` enum? Consider also that we may want the ability to import accounts by pool-specific keys (e.g. transparent private key, sapling private key, sapling viewing key, etc.) eventually too.

I do not want to make this decision at this time; we can alter `WalletWrite` later if we decide it becomes necessary. I will note however that the Data Access APIs are intended to be _granular_, with backend-agnostic logic living in `zcash_client_backend`. The more high-level we make `WalletWrite` APIs, the more logic needs to be duplicated across alternative backends (and kept in sync as future changes to the Data Access API are made.

> - Not entirely related to this PR, but it came up in my review: when `zcash_client_sqlite::wallet::add_account` rewrites scan ranges for the new account, could it possibly make the range to scan _smaller_ than it already was? Say, for example, one account was fully synced. Then two new accounts were added together, the first with an old birthday and the second with a young birthday height. Would the older birthday with the broader scan range survive such that the next sync would scan enough blocks for the longer of the two?

The scan queue priority logic is implemented such that new ranges won't incorrectly clobber existing ones. When these two new accounts are added:
- The first's older birthday would introduce a `Historic` priority scan range that replaces any existing `Scanned` priority range covering it from its birthday.
- The second's newer birthday would attempt to introduce an `Ignored` priority range extending prior to its birthday, that would not replace the `Historic` priority that is currently there.

zcash_client_sqlite/src/lib.rs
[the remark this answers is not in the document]
AArnott: We now record the data.

AArnott:
There is one test failure: `wallet::init::migrations::ephemeral_addresses::tests::initialize_table`

This test fails because it executes the `create_account` function in the product, but with a database that hasn't had the new migration applied (since it's a migration test). In general, it's incorrect to run the latest product as part of a migration test which executes code in a db state that isn't as current as the product. 
Should we copy the older version of the product functions into the migration test? 
The failure occurs here:

```rs
        // Creating a new account should initialize `ephemeral_addresses` for that account.
        let seed1 = vec![0x01; 32];
        let (account1_id, _usk) = db_data
            .create_account(&Secret::new(seed1), &birthday)
            .unwrap();
        assert_ne!(account0_id, account1_id);
        check(&db_data, account1_id);
```

I wonder based on the comment if the migration test is testing a little too much. It appears that it's testing product functionality beyond the migration itself. Maybe this part of the test should be moved to an ordinary non-migration test so that it always runs against the *current* db schema.
If you agree, perhaps @daira as the test owner can push an update? (or rather, since this PR comes from an org instead of a personal account, either push a commit that fixes the test as an independent PR or push the commit on top of this branch somewhere I can fetch it and push it to my source branch to update _this_ PR).

nuttycom:
> There is one test failure: `wallet::init::migrations::ephemeral_addresses::tests::initialize_table`
> 
> This test fails because it executes the `create_account` function in the product, but with a database that hasn't had the new migration applied (since it's a migration test). In general, it's incorrect to run the latest product as part of a migration test which executes code in a db state that isn't as current as the product. Should we copy the older version of the product functions into the migration test?

The normal way we handle this is that we copy the source of the method that is used in the migration test, as of the creation of that migration, into the migration test sources itself. We do this incrementally when we discover breakage; it's often a good canary for other things that one needs to look at when a migration test fails in this fashion.

nuttycom:
@AArnott I have opened #1459 with the fix for your test errors; I apparently can't even open a PR against the `nerdcash/librustzcash` repo.

nuttycom:
Alternately we can just merge #1459 if that's more convenient for you; that will close this PR as well.

AArnott:
Thanks for fixing the failing test, @nuttycom. I've merged it into this PR.

AArnott:
> I apparently can't even open a PR against the nerdcash/librustzcash repo. 

That's _very_ odd.

nuttycom:
Argh, looks like there's one more clippy error.

nuttycom:
@AArnott I have added the clippy fix to my branch, if you want to pull it across.

nuttycom (approved):
re-utACK 31f8e644837d643c1823a8adb82c08abae5bc36d

str4d (approved):
utACK 31f8e644837d643c1823a8adb82c08abae5bc36d modulo documentation issues, which we will fix in a separate PR because we can't make edits to this one.

zcash_client_backend/src/data_api.rs
```diff
@@ -1600,6 +1642,55 @@ pub trait WalletWrite: WalletRead {
...
+    ///
+    /// Returns details about the imported account, including the unique account  identifier for
+    /// the newly-created wallet database entry, along with the associated  [`UnifiedSpendingKey`].
+    /// Note that the unique account identifier should *not* be  assumed equivalent to the ZIP 32
+    /// account index. It is an opaque identifier for a  pool of funds or set of outputs controlled
+    /// by a single spending authority.  
+    ///
+    /// Import accounts with indexes that are exactly one greater than the highest existing account
+    /// index to ensure account indexes are contiguous, thereby facilitating automated account
+    /// recovery.
```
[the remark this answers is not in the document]
AArnott: the latter.

zcash_client_backend/src/data_api.rs
```diff
@@ -1556,38 +1556,80 @@ impl AccountBirthday {
...
 
     /// Tells the wallet to track the next available account-level spend authority, given the
     /// current set of [ZIP 316] account identifiers known to the wallet database.
     ///
+    /// The "next available account" is defined as the first unused ZIP-32 account index (counting
+    /// from 0) among all accounts that share the given seed. When [`Self::import_account_hd`] is
+    /// used to import an account with a specific index, account indexes *may* become fragmented
+    /// instead of one contiguous range. Where fragmentation occurs, the implementations *may*
+    /// choose to find the first unused account index or add 1 to the highest existing account
+    /// index.
```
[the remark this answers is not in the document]
AArnott: Great. As you say, your preferred behavior is what the sqlite crate does. I was divided on what the *best* behavior was, so I'm happy to consent to your suggestion.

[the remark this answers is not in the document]
daira: This documentation doesn't appear to have been changed before the PR was merged.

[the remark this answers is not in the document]
AArnott: Correct. It was merged after @str4d's comment, before I had even read it.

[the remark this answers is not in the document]
str4d: I opened #1465 with the changes because I could not push them to this PR.

PR #1429, opened 2024-06-18
zcash_client_sqlite: Start documenting the table structures

str4d:
The existing pins for the current database structure are moved out of the `verify_schema` test, so they can serve the dual purpose of documentating the database structure.

The documentation needs significantly more detail than this PR currently includes; this is intended as a starting point from which we can continually improve our documentation going forward.

### DIFF
--- a/zcash_client_sqlite/src/wallet/db.rs
+++ b/zcash_client_sqlite/src/wallet/db.rs
@@ -13,6 +13,11 @@
 // from showing up in `cargo doc --document-private-items`.
 #![allow(dead_code)]
 
+use zcash_client_backend::data_api::scanning::ScanPriority;
+use zcash_protocol::consensus::{NetworkUpgrade, Parameters};
+
+use crate::wallet::scanning::priority_code;
+
 /// Stores information about the accounts that the wallet is tracking.
 pub(super) const TABLE_ACCOUNTS: &str = r#"
 CREATE TABLE "accounts" (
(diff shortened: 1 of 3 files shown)


zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -236,12 +318,20 @@ CREATE TABLE sapling_tree_cap (
...
+/// [`ShardTree`]: shardtree::ShardTree
 pub(super) const TABLE_SAPLING_TREE_CHECKPOINTS: &str = "
 CREATE TABLE sapling_tree_checkpoints (
     checkpoint_id INTEGER PRIMARY KEY,
     position INTEGER
 )";
 
+/// TODO: Summary of what this stores.
+///
+/// TODO: Document this.
```
nuttycom, line 334:
```suggestion
/// Stores metadata about the positions of Sapling notes that have been spent but for which
/// witness information as not yet been removed from the note commitment tree.
///
/// In the process of updating the note commitment tree in response to the addition of 
/// a block, it is necessary to temporarily continue to store witness information for
/// each note so that a spent note can be made spendable again after a rollback of the
/// spending block. This table caches the metadata needed for that restoration.
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -269,12 +369,22 @@ CREATE TABLE orchard_tree_cap (
...
+/// details.
+///
+/// [`ShardTree`]: shardtree::ShardTree
 pub(super) const TABLE_ORCHARD_TREE_CHECKPOINTS: &str = "
 CREATE TABLE orchard_tree_checkpoints (
     checkpoint_id INTEGER PRIMARY KEY,
     position INTEGER
 )";
 
+/// TODO: Summary of what this stores.
```
nuttycom, line 384:
```suggestion
/// Stores metadata about the positions of Orchard notes that have been spent but for which
/// witness information as not yet been removed from the note commitment tree.
```

nuttycom (changes requested):
Added suggestions for documentation to replace `TODO`s.

nuttycom (approved):
utACK cf1922cdee5a72633099bd89472c01793f7d7919

daira (approved):
utACK. So much better! <3

PR #1430, opened 2024-06-18
Improvements to the strategy implemented by `{fixed,standard,zip317}::SingleOutputChangeStrategy`

daira:
`zcash_client_backend::{fixed,standard,zip317}::SingleOutputChangeStrategy` now implement a different strategy for choosing whether there will be any change, and its value. The aims are:

* Ensure that it is still possible to create fully transparent transactions with no change (this will be needed for ZIP 320). The `ChangeError::InsufficientFunds` error in this case should have a `required` field that reflects the additional amount needed, according to the fee calculated without an extra change output.
* Avoid leaking information about note amounts in some cases: an adversary that knew the number of external recipients and the sum of their outputs was able to learn the sum of the inputs if no change output was present.
* Defend against losing money by using `DustAction::AddDustToFee` with a too-high dust threshold.
* Ensure that if a "change memo" is requested, there will always be a shielded change output in which to put it. Previously, this would not be the case when using `DustAction::AddDustToFee`.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -20,6 +20,13 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.70.0.
+- `zcash_client_backend::{fixed,standard,zip317}::SingleOutputChangeStrategy`
+  now implement a different strategy for choosing whether there will be any
+  change, and its value. This can avoid leaking information about note amounts
+  in some cases. It also ensures that there will be a change output whenever a
+  `change_memo` is given, and defends against losing money by using
+  `DustAction::AddDustToFee` with a too-high dust threshold.
+  See [#1430](https://github.com/zcash/librustzcash/pull/1430) for details.
 - `zcash_client_backend::zip321` has been extracted to, and is now a reexport 
   of the root module of the `zip321` crate. Several of the APIs of this module
   have changed as a consequence of this extraction; please see the `zip321`
--- a/zcash_client_backend/src/fees/zip317.rs
+++ b/zcash_client_backend/src/fees/zip317.rs
@@ -212,7 +212,6 @@ impl ChangeStrategy for SingleOutputChangeStrategy {
 
 #[cfg(test)]
 mod tests {
-
     use std::convert::Infallible;
 
     use zcash_primitives::{
@@ -229,7 +228,7 @@ mod tests {
         data_api::wallet::input_selection::SaplingPayment,
         fees::{
             tests::{TestSaplingInput, TestTransparentInput},
-            ChangeError, ChangeStrategy, ChangeValue, DustOutputPolicy,
+            ChangeError, ChangeStrategy, ChangeValue, DustAction, DustOutputPolicy,
         },
         ShieldedProtocol,
     };
@@ -323,7 +322,19 @@ mod tests {
     }
 
     #[test]
-    fn change_with_transparent_payments() {
+    fn change_with_transparent_payments_implicitly_allowing_zero_change() {
+        change_with_transparent_payments(&DustOutputPolicy::default())
+    }
+
+    #[test]
+    fn change_with_transparent_payments_explicitly_allowing_zero_change() {
+        change_with_transparent_payments(&DustOutputPolicy::new(
+            DustAction::AllowDustChange,
+            Some(NonNegativeAmount::ZERO),
+        ))
+    }
+
+    fn change_with_transparent_payments(dust_output_policy: &DustOutputPolicy) {
         let change_strategy = SingleOutputChangeStrategy::new(
             Zip317FeeRule::standard(),
             None,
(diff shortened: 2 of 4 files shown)


zcash_client_backend/src/fees/zip317.rs
```diff
@@ -351,13 +350,149 @@ mod tests {
             ),
             #[cfg(feature = "orchard")]
             &orchard_fees::EmptyBundleView,
+            &DustOutputPolicy::new(DustAction::AllowDustChange, Some(NonNegativeAmount::ZERO)),
```
str4d, line 353: Why was the `change_with_transparent_payments` test altered to use this dust output policy (instead of `DustOutputPolicy::default()`)? AFAICT the test is not transparent-only (it has a Sapling input), and therefore the change to the fee code should not have affected it.

If this isn't a necessary change, revert it. If we don't have a test with `DustAction::AllowDustChange`, then we should add that separately.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -443,7 +453,7 @@ pub(crate) fn send_multi_step_proposed_transfer<T: ShieldedPoolTester>() {
 
     assert_eq!(
         confirmed_sent.get(0).and_then(|v| v.get(0)),
-        Some(&(&txids[0], 50000))
+        Some(&(&txids[0], 35000))
```
str4d, line 456: It took me a while to realise that this edit was not due to changes in the existing output amount or how fees were changed, but just checking the value of the new change output instead.

This test now has an undocunmented dependency on implementation details about the order in which the two outputs of the first transaction are added to the database. Change the test to instead sort the outputs by value and check both of them (the existing 50,000 zatoshi non-change output, and the new 35,000 zatoshi change output).

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -904,7 +914,7 @@ pub(crate) fn spend_succeeds_to_t_addr_zero_change<T: ShieldedPoolTester>() {
     let dfvk = T::test_account_fvk(&st);
 
     // Add funds to the wallet in a single note
-    let value = NonNegativeAmount::const_from_u64(60000);
+    let value = NonNegativeAmount::const_from_u64(70000);
```
str4d, line 917: Why is this amount increased?

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -951,7 +961,7 @@ pub(crate) fn change_note_spends_succeed<T: ShieldedPoolTester>() {
     let dfvk = T::test_account_fvk(&st);
 
     // Add funds to the wallet in a single note owned by the internal spending key
-    let value = NonNegativeAmount::const_from_u64(60000);
+    let value = NonNegativeAmount::const_from_u64(70000);
```
str4d, line 964: Why is this amount increased?

str4d (changes requested):

str4d (approved):
utACK 977e6c802a0a530c46ad1608cc8a3594c7a53d19 modulo questions about tests. Any that are not addressed here, should have issues opened.

zcash_client_backend/src/fees/zip317.rs
daira: The behaviour intentionally changes in the situation of the original test. Specifically, one of the dummy outputs that was originally present would become a change output instead, because we provided a change memo. (This would be observable to someone with the corresponding internal IVK, although not otherwise.) However, the default dust policy disallows creating a transaction with zero-valued change (the default dust threshold is the ZIP 317 marginal fee of 5000 zat), and so the test would fail without changing the dust policy.

I don't know whether this is the intent of the default policy, but it is strictly speaking what we asked for: we specified a change memo and said to reject if the change is less than the marginal fee. So what I did to make the test pass is explicitly specify a policy that allowed this case (which might not be the correct solution).

In more detail:

The original test has a Sapling input of 55000 zat (so, the flows are not fully transparent), and a transparent output of 40000 zat. The selected change pool is Sapling. There is no explicitly specified Sapling output (and the `bundle_required` setting of the builder defaults to false), but there is at least one Sapling input and so the outputs are padded to 2; [this is the intended behaviour of the Sapling builder](https://github.com/zcash/sapling-crypto/blob/d23547f0d3c753b706c6a4fd64456855e153d6b7/src/builder.rs#L36..L37). There is no Sapling change output. Therefore the ZIP 317 fee is, correctly, 15000 zat because we have 3 logical actions: two due to Sapling (max of one input and two dummy outputs), and one transparent output.

Before the latest changes from our pairing, the PR was incorrect because it was calculating the fee as 20000 zat.

The new behaviour is that we create a zero-valued change output with `change_memo`. Because of the output padding, this makes no difference to the fee of 15000 zat, and so the balance is calculated as 55000 - 40000 - 15000 = 0, i.e. we are in case 2, but the flows are not transparent so this becomes case 3b. However, the default dust policy disallows the change output. Therefore, we reject the transaction as having insufficient funds.

I'm open to either:
* renaming the test;
* having this test explicitly fail and adding another test that allows zero-valued change;
* changing the interpretation of the default dust policy to allow this case. (Should it also allow change less than the marginal fee in general if a change memo is given, or only zero-valued change?)

daira: An argument for allowing zero-valued change but not allowing other change values less than the threshold, is that we will never need to spend a zero-valued note (and in principle shouldn't be keeping track of its witness, although I don't know whether that is currently the case). And this case may occur in practice for the intended "send max" feature in Zashi.

str4d: > Because of the output padding, this makes no difference to the fee of 15000 zat, and so the balance is calculated as 55000 - 40000 - 15000 = 0, i.e. we are in case 2, **but the flows are not transparent so this becomes case 3b**.

This is the bit of logic that was missing when I originally wrote this comment in the previous PR. I think this test is therefore fine as-is now.

daira: I've implemented this now; please re-review.

zcash_client_backend/CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.70.0.
+- `zcash_client_backend::{fixed,standard,zip317}::SingleOutputChangeStrategy`
+  now implement a slightly different strategy for choosing whether there will
+  be any change. Transactions that have any shielded non-change flows will
+  always have a change output (possibly zero-valued). This can avoid leaking
+  information about note amounts in some cases.
```
daira, line 27: I will update this changelog entry.

zcash_client_sqlite/src/testing/pool.rs
daira: This change is no longer necessary. I think it was because we were previously calculating the fee incorrectly.

zcash_client_sqlite/src/testing/pool.rs
daira: This change is no longer necessary. I think it was because we were previously calculating the fee incorrectly.

PR #1432, opened 2024-06-19
Update tests in testing/pool.rs to use ZIP 317 fees

daira:


zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -921,18 +912,15 @@ pub(crate) fn spend_succeeds_to_t_addr_zero_change<T: ShieldedPoolTester>() {
     let dfvk = T::test_account_fvk(&st);
 
     // Add funds to the wallet in a single note
-    let value = NonNegativeAmount::const_from_u64(60000);
+    let value = NonNegativeAmount::const_from_u64(70000);
```
str4d, line 915: Is this bump necessary for the test to work under ZIP 317, or is it left over from your other PR (that you subsequently undid)?

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -968,7 +956,7 @@ pub(crate) fn change_note_spends_succeed<T: ShieldedPoolTester>() {
     let dfvk = T::test_account_fvk(&st);
 
     // Add funds to the wallet in a single note owned by the internal spending key
-    let value = NonNegativeAmount::const_from_u64(60000);
+    let value = NonNegativeAmount::const_from_u64(70000);
```
str4d, line 959: Ditto

zcash_client_sqlite/src/testing/pool.rs
daira: No, this is necessary because the ZIP 317 fee is higher than 10000 zats. (Actually a `value` of 65000 zats would be sufficient, but this is not intended to be testing the zero-change corner case.)

zcash_client_sqlite/src/testing/pool.rs
daira: Same response.

str4d (approved):
utACK f0cf40543a92cd98829b2a6c8462b2f0d0e6da46

PR #1433, opened 2024-06-19
zcash_client_sqlite: Fix handling of PRAGMA directives.

nuttycom:
The `foreign_keys` pragma has no effect when used within a transaction, so it should only be set at the top level. The `legacy_alter_table` pragma should only be used in cases where its effect is explicitly intended.

Best reviewed hiding whitespace changes.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -262,7 +262,7 @@ pub fn init_wallet_db<P: consensus::Parameters + 'static>(
     init_wallet_db_internal(wdb, seed, &[], true)
 }
 
-fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
+pub(crate) fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
     wdb: &mut WalletDb<rusqlite::Connection, P>,
     seed: Option<SecretVec<u8>>,
     target_migrations: &[Uuid],
@@ -270,13 +270,13 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
 ) -> Result<(), MigratorError<WalletMigrationError>> {
     let seed = seed.map(Rc::new);
 
-    // Turn off foreign keys, and ensure that table replacement/modification
-    // does not break views
+    // Turn off foreign key enforcement, to ensure that table replacement does not break foreign
+    // key references in table definitions.
+    //
+    // It is necessary to perform this operation globally using the outer connection because this
+    // pragma has no effect when set or unset within a transaction.
     wdb.conn
-        .execute_batch(
-            "PRAGMA foreign_keys = OFF;
-             PRAGMA legacy_alter_table = TRUE;",
-        )
+        .execute_batch("PRAGMA foreign_keys = OFF;")
         .map_err(|e| MigratorError::Adapter(WalletMigrationError::from(e)))?;
     let adapter = RusqliteAdapter::new(&mut wdb.conn, Some("schemer_migrations".to_string()));
     adapter.init().expect("Migrations table setup succeeds.");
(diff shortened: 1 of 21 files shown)


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -390,20 +399,12 @@ mod tests {
...
-            .wallet()
-            .conn
-            .prepare("SELECT sql FROM sqlite_schema WHERE type = 'table' ORDER BY tbl_name")
-            .unwrap();
-        let mut rows = tables_query.query([]).unwrap();
-        let mut expected_idx = 0;
-        while let Some(row) = rows.next().unwrap() {
-            let sql: String = row.get(0).unwrap();
+        let rows = describe_tables(&st.wallet().conn).unwrap();
+        for (actual, expected) in rows.iter().zip(expected_tables.iter()) {
```
str4d, line 403: `zip` short-circuits, so we no longer detect whether the real database contains tables that aren't listed in `expected_tables`. I also note that we weren't detecting the reverse case previously (`expected_tables` containing tables that are not in the real database).

Simple fix is to add an assertion that the two are the same length, prior to the `zip`:
```suggestion
        assert_eq!(rows.len(), expected_tables.len());
        for (actual, expected) in rows.iter().zip(expected_tables.iter()) {
```

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -116,3 +116,29 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
+#[cfg(test)]
+mod tests {
+    use secrecy::Secret;
+    use tempfile::NamedTempFile;
+    use uuid::Uuid;
+    use zcash_protocol::consensus::Network;
+
+    use crate::{wallet::init::init_wallet_db_internal, WalletDb};
+
+    pub(crate) fn test_migrate(migrations: &[Uuid]) {
```
str4d, line 129:
```suggestion
    /// Tests that we can migrate from a completely empty wallet database to the target
    /// migrations.
    pub(crate) fn test_migrate(migrations: &[Uuid]) {
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -262,21 +262,21 @@ pub fn init_wallet_db<P: consensus::Parameters + 'static>(
     init_wallet_db_internal(wdb, seed, &[], true)
 }
 
-fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
+pub(crate) fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
```
str4d, line 265: AFAICT this is unnecessary; the new places we use this are all in sub-modules. So to ensure we don't use this elsewhere in the crate, undo this:
```suggestion
fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
```

str4d (changes requested):
Reviewed a452cf966972339cc042906a53897607fb6ef163.

str4d (approved):
utACK 1110f5d8fb22650e2f0e10300be2ce62d8036d78

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -270,13 +270,13 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
...
+    // pragma has no effect when set or unset within a transaction.
     wdb.conn
-        .execute_batch(
-            "PRAGMA foreign_keys = OFF;
-             PRAGMA legacy_alter_table = TRUE;",
-        )
+        .execute_batch("PRAGMA foreign_keys = OFF;")
         .map_err(|e| MigratorError::Adapter(WalletMigrationError::from(e)))?;
     let adapter = RusqliteAdapter::new(&mut wdb.conn, Some("schemer_migrations".to_string()));
     adapter.init().expect("Migrations table setup succeeds.");
```
daira, line 282: Pre-existing issue, but the `?` early returns on lines 289 and 292 can bypass resetting `PRAGMA foreign_keys = ON`, leaving the database in a less-safe state.

daira (commented):
Post-hoc ACK with comment about a pre-existing issue.

PR #1437, opened 2024-06-25
zcash_client_backend: Re-add period that was removed

str4d:


str4d:
Closing as this has been fixed in #1257, and fixing it separately will create more merge conflicts than it solves.

PR #1443, opened 2024-07-07
Set the default optimization level to 3 for testing

daira:
Use `--profile=dev` to disable this. (As well as the fact that much of our code essentially requires optimization to run at all reasonably, this ensures that we are testing the code as it is compiled for release builds.)

An advantage of this over using `--release` when running tests is that it builds with full debug information (as far as possible given that some things will be optimized out).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,6 +123,10 @@ lto = true
 panic = 'abort'
 codegen-units = 1
 
+[profile.test]
+opt-level = 3
+debug = true
+
 [patch.crates-io]
 incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree", rev = "337f59179eda51261e9ddfc6b18e8fb84ea277c9" }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -123,6 +123,10 @@ lto = true
 panic = 'abort'
 codegen-units = 1
 
+[profile.test]
+opt-level = 3
```
str4d, line 127: We should experiment with Cargo profile overrides to just optimize the math-heavy dependencies: https://docs.rust-embedded.org/book/unsorted/speed-vs-size.html#optimizing-dependencies

daira:
Superceded by #1603.

Cargo.toml
daira: It's not necessary to block making this change (which is also in #1603) on doing that. We can do so *as well* to further improve build+test times, but this is *already* a strict improvement on both the time and debugging QoL axes over `--release`, and you can override it with `--profile=dev` when you need to.

daira: Filed as #1606.

PR #1444, opened 2024-07-07
Add utility methods on `TestState` to dump the contents of a database table, or run an sqlite3 command

daira:
These are only usable if both `#[cfg(test)]` and the "unstable" feature are enabled.


nuttycom (approved):
utACK 615f6254f872c2673127afa795ed709f227b6b0a; test-only

PR #1447, opened 2024-07-15
`zcash_extensions 0.1.0`

str4d:
We are forced to publish this in order to secure the `zcash_extensions` crate name. crates.io changed their Usage Policy in November 2023 to prohibit name-squatting; my account was temporarily locked, and this crate deleted, because it only had a `0.0.0` reservation release with no code.


nuttycom (approved):
utACK

PR #1454, opened 2024-07-23
Make `nu6` rust feature stable

oxarbitrage:
We want to add the [testnet network activation to Zebra](https://github.com/ZcashFoundation/zebra/issues/8689) but we need the same in `zcash_primitives`. We also need to remove the `zcash_unstable = "nu6"` feature.

### DIFF
--- a/components/zcash_protocol/src/consensus.rs
+++ b/components/zcash_protocol/src/consensus.rs
@@ -354,7 +354,6 @@ impl Parameters for MainNetwork {
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_000)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_046_400)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_687_104)),
-            #[cfg(zcash_unstable = "nu6")]
             NetworkUpgrade::Nu6 => None,
             #[cfg(zcash_unstable = "zfuture")]
             NetworkUpgrade::ZFuture => None,
@@ -384,8 +383,7 @@ impl Parameters for TestNetwork {
             NetworkUpgrade::Heartwood => Some(BlockHeight(903_800)),
             NetworkUpgrade::Canopy => Some(BlockHeight(1_028_500)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_842_420)),
-            #[cfg(zcash_unstable = "nu6")]
-            NetworkUpgrade::Nu6 => None,
+            NetworkUpgrade::Nu6 => Some(BlockHeight(2_942_000)),
             #[cfg(zcash_unstable = "zfuture")]
             NetworkUpgrade::ZFuture => None,
         }
@@ -452,7 +450,6 @@ pub enum NetworkUpgrade {
     /// The [Nu6] network upgrade.
     ///
     /// [Nu6]: https://z.cash/upgrade/nu6/
-    #[cfg(zcash_unstable = "nu6")]
     Nu6,
     /// The ZFUTURE network upgrade.
     ///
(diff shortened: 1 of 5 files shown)


oxarbitrage:
- Need to remove Testnet height and leave it to `None` for now.

components/zcash_protocol/src/consensus.rs
```diff
@@ -590,7 +582,6 @@ impl From<BranchId> for u32 {
             BranchId::Heartwood => 0xf5b9_230b,
             BranchId::Canopy => 0xe9ff_75a6,
             BranchId::Nu5 => 0xc2d6_d0b4,
-            #[cfg(zcash_unstable = "nu6")]
             BranchId::Nu6 => 0xc8e7_1055,
```
oxarbitrage, line 585: We can't put `None` here, should we go with `0xffff_ffff` or a random one that is not `0xc8e7_1055`?

`0xc8e7_1055` is the one in the proposed ZIP: https://github.com/zcash/zips/pull/878/files#diff-7912352c4bcac5fd61ea274ab3d1f0d53bf48f2bbb7426ede47b5982d8dc564bR44

components/zcash_protocol/src/consensus.rs
```diff
@@ -657,15 +648,16 @@ impl BranchId {
             BranchId::Canopy => params
                 .activation_height(NetworkUpgrade::Canopy)
                 .map(|lower| (lower, params.activation_height(NetworkUpgrade::Nu5))),
-            BranchId::Nu5 => params.activation_height(NetworkUpgrade::Nu5).map(|lower| {
+            BranchId::Nu5 => params
+                .activation_height(NetworkUpgrade::Nu5)
+                .map(|lower| (lower, params.activation_height(NetworkUpgrade::Nu6))),
+            BranchId::Nu6 => params.activation_height(NetworkUpgrade::Nu6).map(|lower| {
```
oxarbitrage, line 654: Codecov complains about this code has not being tested, which is correct but i think we can't add tests for this until we have the Nu6 activation height ?

components/zcash_protocol/src/consensus.rs
str4d: As long as we haven't set any activation heights for mainnet or testnet, we can use the random one from the current ZIP draft while getting the consensus rules implemented. Once we set an activation height, any subsequent consensus rule change would force us to change this.

oxarbitrage: Ok, maybe is safer to introduce a new branch id here?

```suggestion
            // TODO: Change this to ZIP-253 value when setting activation heights.
            BranchId::Nu6 => 0xffff_fffe,
```

daira (approved):
utACK

components/zcash_protocol/src/consensus.rs
oxarbitrage: We discussed this between the ECC and the ZF engineers and decided to use the same branch ID as the current open ZIP(https://github.com/zcash/zips/pull/878/files), which is `0xc8e7_1055`.

components/zcash_protocol/src/consensus.rs
oxarbitrage: Can be resolved as in this case the failure is informative.

str4d: > we can't add tests for this until we have the Nu6 activation height ?

That's not true; this code just needs _some_ `P: Parameters`, not necessarily the mainnet or testnet parameters.

That being said, we only use codecov for informational purposes.

str4d (approved):
utACK. Changelog was not updated, but I can do that in a separate PR.

PR #1457, opened 2024-07-24
devtools: Add block hash lookups

str4d:
Curently doesn't work because `lightwalletd` does not have support for looking up blocks by hash.


str4d:
Moved to https://github.com/zcash/zcash-devtool/pull/80.

Issue #1046, opened 2023-11-21
zcash_client_sqlite: Identify shielding transactions in the output of the v_transactions view.


nuttycom:
The specific ask here is to identify transactions that were created as a result of clicking the "shield funds" button in a wallet (i.e. by calling `propose_shielding`). These transactions have the following characteristics:
- Only transparent inputs.
- Only shielded outputs.
- Inputs are all from the wallet.
- Outputs are all to the wallet.
  - i.e. this is a wallet-internal transaction, though for historic wallets the outputs might not have been sent to wallet-internal addresses.

str4d:
We can't implement exactly the above in `v_transactions`:
- We do not record any details about other inputs beyond what we ourselves have spent.
- We do record details about other outputs in the `sent_notes` table, but probably not as reliably as we'd need.

However, we _can_ implement the following check:
- Only transparent inputs from our wallet.
- Only shielded outputs to our wallet.

This can be done by adding a count of `does_not_match_shielding` that we increment when we observe a spent note that is not from the transparent pool, or a received note that is to the transparent pool. Then sum that counter and `is_shielding` is when the counter is zero.

Issue #1434, opened 2024-06-22
Ensure symmetry of transparent output handling between `put_blocks`, `store_decrypted_tx` and `store_sent_tx`


nuttycom:
> AFAICT (at least as of the first commit in the PR), this `NULL AS from_account_id` is preserved, because there is still an asymmetry between transparent and shielded (we are not inserting transactions into both sent and received tables for transparent, in the same way we do for shielded). I think that asymmetry needs fixing prior to the next release (if not in this PR).

_Originally posted by @str4d in https://github.com/zcash/librustzcash/pull/1402#discussion_r1648028219_

_See also https://github.com/zcash/librustzcash/pull/1402#discussion_r1648037109_

daira:
Did merging #1402 fix this, or is there more to do?

str4d:
There is still more to do; e.g. `store_decrypted_tx` does not update the received transparent outputs table. This causes an issue with ZIP 320 transactions, but it is also one that we might decide to stick with for the next release. I've marked this issue as blocking #1231 for us to decide whether we do finish this or bump it.

PR #1461, opened 2024-07-26
zcash_client_sqlite: Add `is_shielding` to `v_transactions` view

str4d:
Closes zcash/librustzcash#1046.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/utxos_to_txos.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/utxos_to_txos.rs
@@ -1,4 +1,5 @@
-//! A migration that brings transparent UTXO handling into line with that for shielded outputs.
+//! A migration that brings transparent UTXO handling into line with that for shielded
+//! outputs, and adds `spent_note_count` and `is_shielding` to `v_transactions`.
 use std::collections::HashSet;
 
 use rusqlite;
@@ -22,7 +23,7 @@ impl schemer::Migration for Migration {
     }
 
     fn description(&self) -> &'static str {
-        "Updates transparent UTXO handling to be similar to that for shielded notes."
+        "Updates transparent UTXO handling to be similar to that for shielded notes, and adds spent_note_count and is_shielding to v_transactions."
     }
 }
 
@@ -197,6 +198,7 @@ impl RusqliteMigration for Migration {
                        ro.pool                    AS pool,
                        id_within_pool_table,
                        ro.value                   AS value,
+                       0                          AS spent_note_count,
                        CASE
                             WHEN ro.is_change THEN 1
                             ELSE 0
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations/utxos_to_txos.rs
```diff
@@ -209,7 +210,13 @@ impl RusqliteMigration for Migration {
                          WHEN (ro.memo IS NULL OR ro.memo = X'F6')
                            THEN 0
                          ELSE 1
-                       END AS memo_present
+                       END AS memo_present,
+                       -- The wallet cannot receive transparent outputs in shielding transactions.
+                       CASE
```
nuttycom, line 215: Won't this treat all T->Z transfers from another wallet to one of this wallet's shielded receivers as shielding transactions?

str4d: Yep, I forgot to add this. We have `sent_note_count` and `received_note_count` that should suffice here I think?

str4d: I've figured out why `received_note_count` is not working:
- We hard-code `is_change` to 0 for transparent received outputs:
  https://github.com/zcash/librustzcash/blob/001f2785313c2ac987ae6bac53ed8e1bce7f571c/zcash_client_sqlite/src/wallet/db.rs#L632-L635
- We compute both `has_change` and `received_note_count` from `is_change`:
  https://github.com/zcash/librustzcash/blob/001f2785313c2ac987ae6bac53ed8e1bce7f571c/zcash_client_sqlite/src/wallet/db.rs#L671-L678
  https://github.com/zcash/librustzcash/blob/001f2785313c2ac987ae6bac53ed8e1bce7f571c/zcash_client_sqlite/src/wallet/db.rs#L737-L739

We need to fix this before the crate release, as this is a bug in the transparent changes.

str4d: And I've also realised that `sent_note_count` won't work until #1434 is finished, as we aren't filling in the `sent_notes` table at all when shielding funds.

str4d: Hmm, actually `received_count` _should_ work with a hard-coded `is_change`, so there must be something else in that logic we're missing, given that we see _both_ `has_change = false` and `received_note_count = 0`.

str4d: The logic is now fixed.

nuttycom (approved):
utACK b42deab8572e4b8108a2dc93b12250cbd4c5d6e4

PR #1462, opened 2024-07-26
zcash_client_sqlite: Store received UTXOs in `store_decrypted_tx`

### LINKED ISSUE #1434: Ensure symmetry of transparent output handling between `put_blocks`, `store_decrypted_tx` and `store_sent_tx`

nuttycom:
Part of #1434

### DIFF
--- a/zcash_primitives/src/legacy.rs
+++ b/zcash_primitives/src/legacy.rs
@@ -1,6 +1,7 @@
 //! Support for legacy transparent addresses and scripts.
 
 use byteorder::{ReadBytesExt, WriteBytesExt};
+use zcash_address::TryFromRawAddress;
 
 use std::fmt;
 use std::io::{self, Read, Write};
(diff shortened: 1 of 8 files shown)


zcash_primitives/src/legacy.rs
```diff
@@ -407,6 +408,24 @@ impl TransparentAddress {
     }
 }
 
+impl TryFromAddress for TransparentAddress {
```
str4d, line 411: This should implement `TryFromRawAddress`, as `TransparentAddress` is a network-agnostic type.

str4d (approved):
utACK 3cec9ee4a7de366d20586ad8ee4da101055824f2

PR #1464, opened 2024-07-27
zcash_primitives: Use `TxId` type inside `transparent::OutPoint`

str4d:
This doesn't affect the public API, but it improves the `Debug` impl for `OutPoint` because `TxId` renders its `[u8; 32]` as byte-reversed hex.


nuttycom (approved):
utACK, I've wanted to do this forever.

PR #1465, opened 2024-07-27
Post #1425 fixes

str4d:
Todo:
- [x] Doc comment fixes left over from #1425.
- [x] Revert redefinition of `WalletWrite::create_account`.
- [x] Rewrite `WalletWrite` trait documentation to clarify account creation.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1636,7 +1636,7 @@ pub trait WalletWrite: WalletRead {
     /// The ZIP-32 account index may be obtained by calling [`WalletRead::get_account`]
     /// with the returned account identifier.
     ///
-    /// Learn more about account creation and import in the [`WalletWrite`] trait documentation.
+    /// The [`WalletWrite`] trait documentation has more details about account creation and import.
     ///
     /// # Panics
     ///
@@ -1657,11 +1657,11 @@ pub trait WalletWrite: WalletRead {
     /// account index. It is an opaque identifier for a pool of funds or set of outputs controlled
     /// by a single spending authority.
     ///
-    /// Import accounts with indexes that are exactly one greater than the highest existing account
-    /// index to ensure account indexes are contiguous, thereby facilitating automated account
+    /// Import accounts with indices that are exactly one greater than the highest existing account
+    /// index to ensure account indices are contiguous, thereby facilitating automated account
     /// recovery.
     ///
-    /// Learn more about account creation and import in the [`WalletWrite`] trait documentation.
+    /// The [`WalletWrite`] trait documentation has more details about account creation and import.
     ///
     /// # Panics
     ///
@@ -1686,7 +1686,7 @@ pub trait WalletWrite: WalletRead {
     /// `spending_key_available` is `false`, the wallet may choose to optimize for this case, in
     /// which case any attempt to spend funds from the account will result in an error.
     ///
-    /// Learn more about account creation and import in the [`WalletWrite`] trait documentation.
+    /// The [`WalletWrite`] trait documentation has more details about account creation and import.
     ///
     /// # Panics
     ///
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1460,7 +1460,7 @@ impl AccountBirthday {
     /// Constructs a new [`AccountBirthday`] from its constituent parts.
     ///
     /// * `prior_chain_state`: The chain state prior to the birthday height of the account. The
-    ///    birthday height  is defined as the height of the first block to be scanned in wallet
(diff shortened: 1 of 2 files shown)


zcash_client_sqlite/src/wallet/init/migrations/ephemeral_addresses.rs
```diff
@@ -81,21 +81,36 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+            data_api::{AccountBirthday, AccountSource},
+            wallet::TransparentAddressMetadata,
+        },
+        zcash_keys::keys::UnifiedSpendingKey,
+        zcash_primitives::{block::BlockHash, legacy::keys::NonHardenedChildIndex},
+        zcash_protocol::consensus::Network,
+        zip32::{fingerprint::SeedFingerprint, AccountId as Zip32AccountId},
+    };
+
+    #[cfg(feature = "transparent-inputs")]
```
daira, line 101: I made the same changes in the first commit of #1466. It's fine, we'll just rebase whichever lands second.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/5dcf83b86b7b94343b06a7be986c4213e9d226c0..9fea40811bde5317f9bcb0753b0d0d7494daa49c) to fix merge conflict with #1466.

zcash_client_sqlite/src/wallet/init/migrations/ephemeral_addresses.rs
```diff
@@ -85,13 +85,27 @@ mod tests {
...
         zcash_protocol::consensus::Network,
-        zip32::fingerprint::SeedFingerprint,
+        zip32::{fingerprint::SeedFingerprint, AccountId as Zip32AccountId},
+    };
+
+    #[cfg(feature = "transparent-inputs")]
+    use crate::{
+        error::SqliteClientError,
+        wallet::{
+            self, account_kind_code, init::init_wallet_db_internal, transparent,
```
daira, line 105: Please fix the clippy warning.

daira (dismissed):
utACK modulo the clippy unused import warning.

nuttycom (approved):
utACK 5b3c87db460f5e5cb565be1cb923bc93d0b470b2

zcash_client_backend/src/data_api.rs
```diff
@@ -1565,10 +1565,42 @@ impl AccountBirthday {
...
+///
+/// The order in which you call these methods will affect the resulting wallet structure:
+/// - If only [`WalletWrite::create_account`] is used, the resulting accounts will have
+///   sequential [ZIP 32] account indices within each given seed.
+/// - If [`WalletWrite::import_account_hd`] is used to import accounts with non-sequential
+///   ZIP 32 account indices from the same seed, a call to [`WalletWrite::create_account`]
+///   will use the ZIP 32 account index just after the highest-numbered existing account.
+/// - If an account is imported via [`WalletWrite::import_account_ufvk`], and then a later
+///   call to either [`WalletWrite::create_account`] or [`WalletWrite::import_account_hd`]
+///   would produce the same UFVK, an error will be returned.
```
daira, line 1588: Should this say "If an account created via [`WalletWrite::create_account`] or imported via [`WalletWrite::import_account_ufvk`] or [`WalletWrite::import_account_hd`] already exists in the wallet with the same UFVK (which will also be the case for accounts with the same seed), an error will be returned." ?

This differs from the current wording because (I think) it is not the case that the creation or import has to be in any particular order; any of these three methods will fail if it produces a duplicate UFVK, regardless of which of them was used to create the existing account. At least, if that's not the case then it is a bug.

zcash_client_backend/src/data_api.rs
```diff
@@ -1651,17 +1675,17 @@ pub trait WalletWrite: WalletRead {
...
+    /// by a single spending authority.
     ///
-    /// Import accounts with indexes that are exactly one greater than the highest existing account
-    /// index to ensure account indexes are contiguous, thereby facilitating automated account
+    /// Import accounts with indices that are exactly one greater than the highest existing account
+    /// index to ensure account indices are contiguous, thereby facilitating automated account
     /// recovery.
     ///
-    /// Learn more about account creation and import in the [`WalletWrite`] trait documentation.
+    /// The [`WalletWrite`] trait documentation has more details about account creation and import.
```
daira, line 1688: I think the doc for each of the three account creation/import methods should call out that it fails when there is a duplicate UFVK, even if that's redundant with the `WalletWrite` doc. Note that it is possible to hit this case even if you only use one of the three methods, and there might be implementors who only read the documentation of the methods they are using (because they haven't reached [this page](https://hmpg.net/) yet).

daira (commented):
Post-hoc ACK with some comments on the docs.

daira:
Filed #1474 for the post-hoc comments.

PR #1466, opened 2024-07-27
Don't store transactions when creation of a subsequent transaction fails

### LINKED ISSUE #1463: `create_proposed_transactions` can incorrectly produce transactions
https://github.com/zcash/librustzcash/blob/4f95bd22d47f6997a3dd5017c01c40769cdab3db/zcash_client_backend/src/data_api/wallet.rs#L611-L646

We are looping over the proposal steps, and calling `create_proposed_transaction` for each step. Then we check whether any ephemeral outputs were created but not spent, and return an error.

The bug is, this is not done transactionally. `create_proposed_transaction` alters the wallet:
https://github.com/zcash/librustzcash/blob/4f95bd22d47f6997a3dd5017c01c40769cdab3db/zcash_client_backend/src/data_api/wallet.rs#L1281-L1294

If the caller is separately scanning the wallet for unmined transactions to resubmit, it will see those created transactions, and "resubmit" them bypassing the intended error state.

It is currently necessary to add the earlier transactions to the wallet, in order to make their outputs selectable by the next step. We therefore need to either enter a database transaction for this entire call, or we need to manually remove the inserted transactions (and any of their effects).

daira:
fixes #1463

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -574,6 +574,13 @@ where
         .map_err(Error::from)
 }
 
+type StepResult<DbT> = (
+    BuildResult,
+    Vec<SentTransactionOutput<<DbT as WalletRead>::AccountId>>,
+    NonNegativeAmount,
+    Vec<OutPoint>,
+);
+
 /// Construct, prove, and sign a transaction or series of transactions using the inputs supplied by
 /// the given proposal, and persist it to the wallet database.
 ///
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -574,6 +574,13 @@ where
         .map_err(Error::from)
 }
 
+type StepResult<DbT> = (
```
nuttycom, line 577: nit: this would be nicer as a struct

nuttycom (approved):
utACK 163425add6d68e0588033a39f972f5452065b0a0

zcash_client_backend/src/data_api/wallet.rs
daira: You may be right. Shall I make it one in #1468?

str4d (commented):
Post-hoc utACK 163425add6d68e0588033a39f972f5452065b0a0

PR #1470, opened 2024-07-29
Add a `read_transactionally` method for transactional read-only operations

### LINKED ISSUE #1131: Remove `rusqlite::Connection`-based impl of `WalletRead`
> This seems pretty unsafe to me, as `get_wallet_summary` is likely to be called from multiple threads; I think that we need to remove this `impl` entirely; that should be a separate PR (or an additional commit to this PR) before release.

_Originally posted by @nuttycom in https://github.com/zcash/librustzcash/pull/1122#discussion_r1466576780_

daira:
`read_transactionally` does not require a `mut` reference to the wallet db. It works by catching `DatabaseBusy` errors from Sqlite and doing jittered exponential backoff.

fixes #1131

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -34,12 +34,13 @@
 
 use incrementalmerkletree::{Marking, Position, Retention};
 use nonempty::NonEmpty;
+use rand::{thread_rng, Rng};
 use rusqlite::{self, Connection};
 use secrecy::{ExposeSecret, SecretVec};
 use shardtree::{error::ShardTreeError, ShardTree};
 use std::{
-    borrow::Borrow, collections::HashMap, convert::AsRef, fmt, num::NonZeroU32, ops::Range,
-    path::Path,
+    borrow::Borrow, collections::HashMap, convert::AsRef, fmt, iter::repeat, num::NonZeroU32,
+    ops::Range, path::Path, thread::sleep, time::Duration,
 };
 use subtle::ConditionallySelectable;
 use tracing::{debug, trace, warn};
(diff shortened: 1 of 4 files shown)


daira:
This is currently untested; I am not certain that it is catching the right error.

zcash_client_sqlite/src/lib.rs
```diff
@@ -224,6 +225,39 @@ impl<P: consensus::Parameters + Clone> WalletDb<Connection, P> {
     }
 }
 
+impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletDb<C, P> {
+    pub fn read_transactionally<F, A, E: From<rusqlite::Error>>(&self, f: F) -> Result<A, E>
+    where
+        F: FnOnce(&WalletDb<SqlTransaction<'_>, P>) -> Result<A, E>,
+    {
+        let mut rng = thread_rng();
+        for backoff_ms in [0u64, 1, 2, 4, 8, 16, 32].into_iter().chain(repeat(64)) {
```
daira, line 234: If this somehow gets into a state where it loops forever, it will sleep an average of 32 ms per iteration, which hopefully should not peg the CPU.

zcash_client_sqlite/src/lib.rs
```diff
@@ -224,6 +225,39 @@ impl<P: consensus::Parameters + Clone> WalletDb<Connection, P> {
...
+            match self.conn.borrow().unchecked_transaction() {
+                Ok(tx) => {
+                    let wdb = WalletDb {
+                        conn: SqlTransaction(&tx),
+                        params: self.params.clone(),
+                    };
+                    let result = f(&wdb)?;
```
daira, line 241: Huh, why did the compiler allow `f` to be called in a loop when its type is only known to be `FnOnce`? (I intended `Fn`, but the question stands as to how this currently compiles.)

zcash_client_sqlite/src/lib.rs
```diff
@@ -224,6 +225,39 @@ impl<P: consensus::Parameters + Clone> WalletDb<Connection, P> {
     }
 }
 
+impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletDb<C, P> {
+    pub fn read_transactionally<F, A, E: From<rusqlite::Error>>(&self, f: F) -> Result<A, E>
```
daira, line 229: TODO: do we also need `transactionally` to retry? If we only had `transactionally` then we would not, but I think it is possible for a call to `transaction()` to fail with `ErrorCode::DatabaseBusy` when there is an outstanding transaction opened by `unchecked_transaction` (this is not prevented by the type system).

daira:
I can't seem to provoke a conflict in practice, even when I'm sure that a write transaction is currently open while I call `unchecked_transaction` in another thread. Sqlite seems to behave as though it is using snapshot isolation. But as far as I understand [the docs](https://www.sqlite.org/isolation.html), it *should not* be using snapshot isolation unless it is in [WAL mode](https://www.sqlite.org/wal.html). It shouldn't be in WAL mode, because that is not the default and we do nothing to change the default. :confused:

daira:
I think that means that the problem that @str4d was worried about in #1131 cannot occur. The Sqlite interface observes snapshot isolation across connections. It does *not* implement any form of isolation within a connection, but since `rusqlite::Connection` is not `Sync`, I don't think it's possible to share a connection between threads. Here's what I get when I try (this code is from a test that is not currently in the PR):

```
error[E0277]: `RefCell<rusqlite::inner_connection::InnerConnection>` cannot be shared between threads safely
    --> zcash_client_sqlite/src/lib.rs:2445:41
     |
2445 |           let join_handle = thread::spawn(|| {
     |  ___________________________-------------_^
     | |                           |
     | |                           required by a bound introduced by this call
2446 | |             wdb1.read_transactionally(|wdb| -> Result<_, rusqlite::Error> {
2447 | |                 // The default transaction behaviour is Deferred, so we need to write to the database
2448 | |                 // to create a db transaction (otherwise the test will hang).
...    |
2464 | |             .unwrap();
2465 | |         });
     | |_________^ `RefCell<rusqlite::inner_connection::InnerConnection>` cannot be shared between threads safely
     |
     = help: within `WalletDb<Connection, LocalNetwork>`, the trait `Sync` is not implemented for `RefCell<rusqlite::inner_connection::InnerConnection>`
     = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock` instead
     = note: required because it appears within the type `Connection`
note: required because it appears within the type `WalletDb<Connection, LocalNetwork>`
[...]
For more information about this error, try `rustc --explain E0277`.
```

daira:
Closing this because my current mental model is that *either*:
* #1131 can't occur, in which case we don't need this; or
* #1131 can occur but only in circumstances that make a fix essentially untestable, absent further ideas.

PR #1473, opened 2024-07-30
Add queues to track the need for transaction enhancement and/or verification of mined status.

nuttycom:


zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "CREATE TABLE tx_retrieval_queue (
+                txid BLOB NOT NULL UNIQUE,
+                dependent_transaction_id INTEGER NOT NULL,
+                dependent_transaction_height INTEGER,
```
daira: This is not normalized; we should JOIN on `transactions` and use `mined_height`.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,70 @@
...
+                txid BLOB NOT NULL UNIQUE,
+                dependent_transaction_id INTEGER NOT NULL
+                FOREIGN KEY (dependent_transaction_id) REFERENCES transactions(id_tx),
+            );
+            CREATE TABLE transparent_spend_search_queue (
+                address TEXT NOT NULL,
+                received_output_height INTEGER,
+                received_txid BLOB NOT NULL,
+                received_index INTEGER NOT NULL
+                CONSTRAINT value_received_height UNIQUE (address, received_output_height)
```
daira, line 45: This is wrong.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,70 @@
...
+        "Adds a table for tracking transactions to be downloaded for transparent output and/or memo retrieval."
+    }
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "CREATE TABLE tx_retrieval_queue (
```
daira, line 35: We need a `WalletRead` API to query this table.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,70 @@
...
+                txid BLOB NOT NULL UNIQUE,
+                dependent_transaction_id INTEGER NOT NULL
+                FOREIGN KEY (dependent_transaction_id) REFERENCES transactions(id_tx),
+            );
+            CREATE TABLE transparent_spend_search_queue (
+                address TEXT NOT NULL,
+                received_output_height INTEGER,
```
daira, line 42: This is unnormalized.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2300,6 +2300,67 @@ pub(crate) fn put_tx_data(
...
+
+    stmt_insert_tx.execute(named_params! {
+        ":txid": txid.as_ref(),
+        ":query_type": query_type.code(),
+    })?;
+
+    Ok(())
+}
+
+pub(crate) fn mark_tx_enriched(
```
str4d, line 2438:
```suggestion
pub(crate) fn notify_tx_retrieved(
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -102,11 +103,13 @@ where
...
-        .or(data.chain_height()?.map(|max_height| max_height + 1))
-        .or_else(|| params.activation_height(NetworkUpgrade::Sapling))
-        .expect("Sapling activation height must be known.");
+    let height = mined_height.map(Ok).unwrap_or_else(|| {
+        Ok(data
+            .get_tx_height(tx.txid())?
+            .or(data.chain_height()?.map(|max_height| max_height + 1))
+            .or_else(|| params.activation_height(NetworkUpgrade::Sapling))
+            .expect("Sapling activation height must be known."))
+    })?;
```
str4d, line 112: If `mined_height.is_some() && data.get_tx_height(_).is_some()`, and they are not equal, then per the `lightwalletd` data model, `mined_height` is correct, and a reorg has occurred. We therefore would need to signal somewhere that the wallet should proceed with reorg logic.

zcash_client_backend/proto/service.proto
```diff
@@ -34,6 +34,10 @@ message TxFilter {
 // RawTransaction contains the complete transaction data. It also optionally includes 
 // the block height in which the transaction was included, or, when returned
 // by GetMempoolStream(), the latest block height.
+//
+// FIXME: the documentation here about mempool status contradicts the documentation
+// for the `height` field (which I hope is correct, and this doesn't give the latest
+// block height for mempool transactions.)
```
str4d, line 40: It turns out that the `GetTransaction` method added to `lightwalletd` is horribly broken, because it does not correctly propagate the output of the `getrawtransaction` JSON-RPC method. `getrawtransaction` returns:
- A non-negative height for transactions mined in blocks in the current main chain.
- `-1` for transactions mined in blocks that are not in the main chain (i.e. conflicted).
- Omitted for unmined transactions in the mempool.

This is further muddied by the fact that the Go parser is parsing the height field as an `int`, which is machine-dependent (and thus could differ from `zcashd`), and also I don't know how it handles omitted values (because it isn't to my knowledge a nullable type). So the unmined transactions might be getting their height fields coerced to some default (0?) - this needs further investigation.

`RawTransaction` was then further misused by the addition of the `GetMempoolStream` RPC. Transactions in the mempool should **never** be associated with the latest block height; if any height is associated with them, it should be the "mempool height" which is `latest_height + 1` (which is the height at which the consensus rules are applied to the mempool transactions). So that RPC is broken as well.

This is going to be Not Fun™️ to fix.

zcash_client_backend/src/data_api.rs
```diff
@@ -541,6 +541,51 @@ pub struct SpendableNotes<NoteRef> {
...
 
+/// A request for transaction data enhancement, spentness check, or discovery
+/// of spends from a given transparent address within a specific block range.
+#[derive(Clone, Debug)]
+pub enum TransactionDataRequest {
+    /// Transaction enhancement (download of complete raw transaction data) is requested.
+    ///
+    /// The caller should respond to this request by calling [`wallet::decrypt_and_store_transaction`]
+    /// to provide data for the requested transaction to the wallet backend. If no data is
+    /// available for the specified transaction,
```
str4d, line 552: Complete this comment, and clarify that "the caller" is the caller of `WalletRead::transaction_data_requests`.

zcash_client_backend/src/data_api.rs
```diff
@@ -541,6 +541,51 @@ pub struct SpendableNotes<NoteRef> {
...
+    /// or shielded outputs belonging to the wallet, may not be discovered by the process of chain
+    /// scanning; as a consequence, the wallet must actively query to find transactions that spend
+    /// such funds. Ideally we'd be able to query by [`OutPoint`] but this is not currently
+    /// functionality that is supported by the light wallet server.
+    #[cfg(feature = "transparent-inputs")]
+    SpendsFromAddress {
+        address: TransparentAddress,
+        block_range_start: BlockHeight,
+        block_range_end: Option<BlockHeight>,
+    },
```
str4d, line 572: Move this into the commit that introduces the query concept.

zcash_client_backend/src/data_api.rs
```diff
@@ -541,6 +541,51 @@ pub struct SpendableNotes<NoteRef> {
...
+}
+
+/// Metadata about the status of a transaction obtained by inspecting the chain state.
+#[derive(Clone, Copy, Debug)]
+pub enum TransactionStatus {
+    /// The requested transaction ID was not recognized as belonging to a transaction
+    /// either in the main chain or in the mempool.
+    TxidNotRecognized,
```
str4d, line 580: If the intention is that the caller calls the `GetTransaction` gRPC, then there is a third state the transaction can be in: recognised as mined, but in a non-main chain block (i.e. `Conflicted`, a subset of which would be "mined, then reorg occurred and it wasn't remined before it expired"). See the earlier comment about the mess we need to clean up wrt `GetTransaction`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1969,6 +1970,89 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+                "UPDATE transactions
+                 SET block = blocks.height
+                 FROM blocks
+                 WHERE txid = :txid
+                 AND blocks.height = :height",
+                sql_args,
+            )?;
+
+            conn.execute(
+                "DELETE FROM tx_retrieval_queue WHERE txid = :txid",
```
str4d, line 2048: This does not check `query_status`, which is fine while `txid` remains `UNIQUE`, but if the previous comment is implemented, this deletion (and the equivalent one for `notify_tx_retrieved`) need to be gated on matching `query_status`.

str4d (commented):
Flushing comments up to 40d94d0ae2fa7c126341fccfb932b661e151bbbb (just before looking at `db.rs`).

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -390,6 +390,25 @@ pub(super) const INDEX_SENT_NOTES_TO_ACCOUNT: &str =
...
+/// - `query_type`:
+///     - `0` for raw transaction (enhancement) data,
+///     - `1` for transaction mined-ness information.
+/// - `dependent_transaction_id`: If the transaction data request is searching for information
+///   about transparent inputs to a transaction, this is a reference to that transaction record.
+///   NULL for transactions where the request for enhancement data is based on discovery due
+///   to blockchain scanning.
+pub(super) const TABLE_TX_RETRIEVAL_QUEUE: &str = r#"
+CREATE TABLE tx_retrieval_queue (
+    txid BLOB NOT NULL UNIQUE,
```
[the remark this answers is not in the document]
nuttycom: Thinking through this a bit further, I think that the semantics of `query_type` should be:
* If status is requested, report status.
* If enhancement is requested, call `decrypt_and_store` with the transaction data if available; otherwise, report status (which will then always be `TxidNotRecognized`). 

If a transaction exists in a block that is not in the main chain or is in the mempool, we still want to decrypt and store it, and then it may or may  not eventually be mined or expire. For the purpose of reporting status, we can have an additional `ExpiredUnmined` state for transactions that were mined in an orphaned block or evicted from the mempool but were never mined in the main chain. If a transaction is accepted to the mempool at any point, we should essentially consider it to be in the mempool until it expires, because some other node might have a reference to it in the mempool even if the node our lightwalletd is connected to does not.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: I don't think we need to separately trigger this; the ordinary scanning process should catch the need for the reorg, and having two pathways into triggering that seems more error-prone than having just one.

zcash_client_sqlite/src/wallet.rs
nuttycom: `txid` will remain unique, and enhancement requests will subsume status requests.

zcash_client_backend/src/data_api.rs
str4d: > if a transaction is mined on a fork but not on the main chain, it should be considered to be in the mempool somewhere until it expires.

This fails in the case of double-spends (same seed in two wallets, and each wallet creates a transaction that spends the same input, and gets mined in a different fork).

zcash_client_sqlite/src/wallet/db.rs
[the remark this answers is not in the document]
daira: Let's make sure that `decrypt_and_store` doesn't have undesired side-effects when called for transactions in the mempool (or transactions that don't and possibly can't get mined) before we make that change.

zcash_client_backend/proto/service.proto
pacu: >This is going to be Not Fun™️ to fix. 


I've pinged wallet developers that might affected by this to bring up awareness.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -479,11 +492,19 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
...
 ) -> Result<Vec<TransactionDataRequest>, SqliteClientError> {
     let mut tx_retrieval_stmt =
         conn.prepare_cached("SELECT txid, query_type FROM tx_retrieval_queue")?;
+    let mut address_request_stmt = conn.prepare_cached(
+        "SELECT ssq.address, t.target_height
+         FROM transparent_spend_search_queue ssq
+         JOIN transactions t ON t.id_tx = ssq.transaction_id
+         WHERE t.mined_height IS NULL
+         AND t.target_height IS NOT NULL",
+    )?;
```
str4d, line 507: I think that the `WHERE` conditions here are incorrect. In particular, `t.target_height IS NOT NULL` is going to prevent us looking up the second transaction in a ZIP 320 pair, because we don't set `target_height` for the first transaction when we detect it via `store_decrypted_transaction`. Similarly, on recovery from seed, the first ZIP 320 transaction will have non-null `mined_height` which would _also_ currently prevent us from looking up the second transaction.

str4d (changes requested):
Reviewed 1cbced3f819f9b583943231810ae7905159e3a09.

zcash_client_backend/src/data_api.rs
```diff
@@ -1811,9 +1860,32 @@ pub trait WalletWrite: WalletRead {
...
         &mut self,
-        account_id: Self::AccountId,
-        n: usize,
-    ) -> Result<Vec<(TransparentAddress, TransparentAddressMetadata)>, Self::Error>;
+        _account_id: Self::AccountId,
+        _n: usize,
+    ) -> Result<Vec<(TransparentAddress, TransparentAddressMetadata)>, Self::Error> {
+        // Default impl is required for feature-flagged trait methods to prevent
+        // breakage due to inadvertent activation of features by transitive dependencies
+        // of the implementing crate.
```
daira, line 1868: Ah. The existing pattern seemed to be that `WalletRead` methods had a default implementation but `WalletWrite` methods did not. Note that `WalletRead::get_orchard_nullifiers` is feature-flagged "orchard" but does not have a default implementation (just as the first example I found; there might be others).

zcash_client_backend/src/data_api.rs
```diff
@@ -1811,9 +1860,32 @@ pub trait WalletWrite: WalletRead {
...
+    /// scanning; as a consequence, the wallet must actively query to determine whether such
+    /// transactions have been mined.
+    fn set_transaction_status(
+        &mut self,
+        _txid: TxId,
+        _status: TransactionStatus,
+    ) -> Result<(), Self::Error> {
+        // Default impl is required for feature-flagged trait methods to prevent
+        // breakage due to inadvertent activation of features by transitive dependencies
+        // of the implementing crate.
```
daira, line 1886: This is not feature-flagged.

zcash_client_sqlite/src/lib.rs
```diff
@@ -587,6 +591,18 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
...
+    fn transaction_data_requests(&self) -> Result<Vec<TransactionDataRequest>, Self::Error> {
+        let iter = std::iter::empty();
+
+        #[cfg(feature = "transparent-inputs")]
+        let iter = iter.chain(
+            wallet::transparent::transaction_data_requests(self.conn.borrow(), &self.params)?
+                .into_iter(),
+        );
+
+        Ok(iter.collect())
```
daira, line 604:
```suggestion
        #[cfg(not(feature = "transparent-inputs"))]
        let reqs = vec![];

        #[cfg(feature = "transparent-inputs")]
        let reqs = wallet::transparent::transaction_data_requests(self.conn.borrow(), &self.params)?
            .into_iter(),
            .collect();

        Ok(reqs)
```
Non-blocking.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1180,7 +1202,24 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                let mut detectable_via_scanning = d_tx.tx().sapling_bundle().is_some();
+                #[cfg(feature = "orchard")] {
+                    detectable_via_scanning |= d_tx.tx().orchard_bundle().is_some();
+                }
+
+                detectable_via_scanning
+            };
+
             for output in d_tx.sapling_outputs() {
+                tx_has_wallet_outputs = true;
```
daira, line 1222: Does this compile if "transparent-inputs" is disabled?

zcash_client_sqlite/src/lib.rs
```diff
@@ -1265,6 +1304,7 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
 
             #[cfg(feature = "orchard")]
             for output in d_tx.orchard_outputs() {
+                tx_has_wallet_outputs = true;
```
daira, line 1307: Ditto.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1972,90 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+                    .iter()
+                    .any(|expiry| *expiry + VERIFY_LOOKAHEAD < chain_tip)
+                {
+                    conn.execute(
+                        "DELETE FROM tx_retrieval_queue WHERE txid = :txid",
+                        named_params![
+                            ":txid": txid.as_ref()
+                        ],
+                    )?;
+                }
```
daira, line 2025: I think this could probably be expressed more simply as a single SQL statement, i.e. something like
```sql
DELETE FROM tx_retrieval_queue q
JOIN transactions t
ON q.txid = t.txid
WHERE t.txid = :txid AND t.expiry_height < :chain_tip_minus_lookahead
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1972,90 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+            )?;
+
+            conn.execute(
+                "UPDATE transactions
+                 SET block = blocks.height
+                 FROM blocks
+                 WHERE txid = :txid
+                 AND blocks.height = :height",
+                sql_args,
+            )?;
```
daira, line 2050: I believe this may fix https://github.com/zcash/librustzcash/pull/1473/files#diff-14c1d11caabdb0c9a159f3a3ae58987177180bb9739872c7ea93cfd4952e9d37R667-R700 . If it does, please remove the workaround from that test.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2292,13 +2385,86 @@ pub(crate) fn put_tx_data(
...
     stmt_upsert_tx_data
-        .query_row(tx_params, |row| row.get::<_, i64>(0))
+        .query_row(tx_params, |row| row.get::<_, i64>(0).map(TxRef))
         .map_err(SqliteClientError::from)
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum TxQueryType {
+    Enhancement,
+    Status,
```
daira, line 2399:
```suggestion
    Status,
    Enhancement,
```
to follow the order of the `match`es below.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1972,90 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+                 FROM blocks
+                 WHERE txid = :txid
+                 AND blocks.height = :height",
+                sql_args,
+            )?;
+
+            conn.execute(
+                "DELETE FROM tx_retrieval_queue WHERE txid = :txid",
+                sql_args,
+            )?;
```
daira, line 2055:
```suggestion
            notify_tx_retrieved(conn, txid)?;
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -186,6 +186,7 @@ CREATE TABLE "transactions" (
     raw BLOB,
     fee INTEGER,
     confirmed_unmined INTEGER,
+    target_height INTEGER,
```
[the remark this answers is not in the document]
daira: Also, are there any constraints that could be applied to it? When can it be null?

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -390,6 +391,42 @@ pub(super) const INDEX_SENT_NOTES_TO_ACCOUNT: &str =
...
+    query_type INTEGER NOT NULL,
+    dependent_transaction_id INTEGER,
+    FOREIGN KEY (dependent_transaction_id) REFERENCES transactions(id_tx)
+)"#;
+
+/// Stores the set of transaction outputs received by the wallet for which spend information
+/// (if any) should be retrieved.
+///
+/// This table is populated in the process of wallet recovery when a deshielding transaction
+/// with transparent outputs belonging to the wallet (i.e., the deshielding half of a ZIP 320
```
daira, line 417:
```suggestion
/// with transparent outputs belonging to the wallet (e.g., the deshielding half of a ZIP 320
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -390,6 +391,42 @@ pub(super) const INDEX_SENT_NOTES_TO_ACCOUNT: &str =
...
+/// transaction pair) is discovered. It is expected that such a transparent output will be
+/// spent soon after it is received in a purely-transparent transaction, which the wallet
```
daira, line 419:
```suggestion
/// spent soon after it is received in a purely transparent transaction, which the wallet
```

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        Ok(())
+    }
+
+    fn down(&self, transaction: &Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "DROP TABLE transparent_spend_search_queue;
+             ALTER TABLE transactions DROP COLUMN target_height;
+             DROP TABLE tx_retrieval_queue;",
+        )?;
```
daira, line 69: When do we care about being able to revert a migration, and when do we not?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -468,13 +483,57 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
...
+            let txid = row.get(0).map(TxId::from_bytes)?;
+            let query_type = row.get(1).map(TxQueryType::from_code)?.ok_or_else(|| {
+                SqliteClientError::CorruptedData(
+                    "Unrecognized transaction data request type.".to_owned(),
+                )
+            })?;
+
+            Ok::<TransactionDataRequest, SqliteClientError>(match query_type {
+                TxQueryType::Enhancement => TransactionDataRequest::Enhancement(txid),
+                TxQueryType::Status => TransactionDataRequest::GetStatus(txid),
```
daira, line 520:
```suggestion
                TxQueryType::Status => TransactionDataRequest::GetStatus(txid),
                TxQueryType::Enhancement => TransactionDataRequest::Enhancement(txid),
```
Non-blocking.

zcash_primitives/CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ and this library adheres to Rust's notion of
   - `InputSize`
   - `InputView::serialized_size`
   - `OutputView::serialized_size`
+- `zcash_primitives::transaction::component::transparent::OutPoint::txid`
+- `zcash_primitives::transaction::builder::DEFAULT_TX_EXPIRY_DELTA
```
daira, line 24:
```suggestion
- `zcash_primitives::transaction::builder::DEFAULT_TX_EXPIRY_DELTA`
```

daira (commented):
Reviewed current state.

zcash_client_sqlite/src/wallet/db.rs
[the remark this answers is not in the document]
nuttycom: This has been resolved as I described above: the `Enhancement` query type now subsumes `Status`, and so the unique constraint on txid is correct.

zcash_client_backend/proto/service.proto
nuttycom: We have discovered additional underlying issues: https://github.com/ZcashFoundation/zebra/issues/8744

nuttycom: Lightwalletd issues documented here: https://github.com/zcash/lightwalletd/issues/493

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: The second transaction in a ZIP 320 pair, when created by this wallet, _will_ correctly have `target_height` set because it will ordinarily be populated by `store_transaction_to_be_sent`. However, you're correct that this will fail to achieve the desired goal for transactions discovered via the wallet recovery process.

zcash_client_sqlite/src/lib.rs
nuttycom: This must have been a victim of a bad rebase; the ordinary request for enhancement and status information is missing. :stuck_out_tongue:

zcash_client_sqlite/src/wallet.rs
nuttycom: Ah, yeah, this can be simplified now that it's no longer setting the `last_observed_unmined_height` field that had been added to the `transactions` table in a previous version of this PR but which has been removed as it ultimately was not needed.

zcash_client_sqlite/src/wallet.rs
nuttycom: At present, `set_transaction_status` is not called by any path in existing tests; it is explicitly introduced as a method for the wallet app to invoke once it has satisfied a transaction data request. It may be possible to update the test to simulate that interaction.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
nuttycom: We don't currently have any code paths that exercise down migrations. We should perhaps universally not implement them; we would always implement a fix to a previous migration as a forward migration, rather than trying to down-migrate.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: This is obsoleted by other changes.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/1cbced3..a3109cfb784579ee89f77b670df236657bcb8f2a) to address comments from code review.

zcash_client_backend/proto/service.proto
daira: Filed as https://github.com/zcash/librustzcash/issues/1484

daira, line 40:
```suggestion
// FIXME: the documentation here about mempool status contradicts the documentation
// for the `height` field. See https://github.com/zcash/librustzcash/issues/1484
```

zcash_client_backend/src/data_api.rs
```diff
@@ -541,6 +541,71 @@ pub struct SpendableNotes<NoteRef> {
...
+/// A request for transaction data enhancement, spentness check, or discovery
+/// of spends from a given transparent address within a specific block range.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TransactionDataRequest {
+    /// Information about the chain's view of a transaction is requested.
+    ///
+    /// The caller evaluating this request on behalf of the wallet backend should respond to this
+    /// request by determining the status of the specified transaction with respect to the main
+    /// chain; if using `lightwalletd` for access to chain data, this may be performed interpreting
+    /// the results of the [`GetTransaction`] RPC method. It should then call
```
daira, line 553:
```suggestion
    /// chain; if using `lightwalletd` for access to chain data, this may be achieved by
    /// interpreting the results of the [`GetTransaction`] RPC method. It should then call
```

zcash_client_backend/src/data_api.rs
```diff
@@ -541,6 +541,71 @@ pub struct SpendableNotes<NoteRef> {
...
+    /// [`WalletWrite::set_transaction_status`]. A [`TransactionDataRequest::Enhancement`] request
+    /// subsumes any previously existing [`TransactionDataRequest::GetStatus`] request.
+    ///
+    /// [`GetTransaction`]: crate::proto::service::compact_tx_streamer_client::CompactTxStreamerClient::get_transaction
+    Enhancement(TxId),
+    /// Information about transactions that receive or spend funds belonging to the specified
+    /// transparent address is requested.
+    ///
+    /// Fully transparent transactions, and transactions that do not contain either shielded inputs
+    /// or shielded outputs belonging to the wallet, may not be discovered by the process of chain
```
daira, line 575:
```suggestion
    /// or shielded outputs belonging to the wallet, will not be discovered by the process of chain
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1811,9 +1890,27 @@ pub trait WalletWrite: WalletRead {
...
+        // breakage due to inadvertent activation of features by transitive dependencies
+        // of the implementing crate.
+        Ok(vec![])
+    }
+
+    /// Updates the wallet backend with respect to the status of a specific transaction, from the
+    /// perspective of the main chain.
+    ///
+    /// Fully transparent transactions, and transactions that do not contain either shielded inputs
+    /// or shielded outputs belonging to the wallet, may not be discovered by the process of chain
```
daira, line 1906:
```suggestion
    /// or shielded outputs belonging to the wallet, will not be discovered by the process of chain
```

zcash_client_backend/src/proto/service.rs
```diff
@@ -38,6 +38,10 @@ pub struct TxFilter {
 /// RawTransaction contains the complete transaction data. It also optionally includes
 /// the block height in which the transaction was included, or, when returned
 /// by GetMempoolStream(), the latest block height.
+///
+/// FIXME: the documentation here about mempool status contradicts the documentation
+/// for the `height` field (which I hope is correct, and this doesn't give the latest
+/// block height for mempool transactions.)
```
daira, line 44:
```suggestion
// for the `height` field. See https://github.com/zcash/librustzcash/issues/1484
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -587,6 +591,19 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
             &address.encode(&self.params),
         )
     }
+
+    fn transaction_data_requests(&self) -> Result<Vec<TransactionDataRequest>, Self::Error> {
+        let iter = std::iter::empty();
+
+        let iter = iter.chain(wallet::transaction_data_requests(self.conn.borrow())?.into_iter());
```
daira, line 598:
```suggestion
        let iter = wallet::transaction_data_requests(self.conn.borrow())?.into_iter();

```

zcash_client_sqlite/src/lib.rs
```diff
@@ -1390,7 +1437,26 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                            // Since the wallet created the transparent output, we need to ensure
+                            // that any transparent inputs belonging to the wallet will be
+                            // discovered.
+                            tx_has_wallet_outputs = true;
+
+                            // When we receive transparent funds (particularly as ephemeral outputs
+                            // in transaction pairs sending to a ZIP 320 address) it becomes
+                            // possible that the spend of these outputs is not then later detected
+                            // if the transaction that spends them is purely transparent. This is
+                            // particularly a problem in wallet recovery.
```
daira, line 1453:
```suggestion
                            // especially a problem in wallet recovery.
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1974,70 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
         }
     }
 
+    // Add the transaction to the set to be queried for transaction status. This is only necessary
+    // at present for fully-transparent transactions, because any transaction with a shielded
```
daira, line 1978:
```suggestion
    // at present for fully transparent transactions, because any transaction with a shielded
```
The rule is that adverbs modifying an adjective do not require a hyphen.

zcash_client_sqlite/src/wallet.rs
daira: I will check that.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,87 @@
...
+                transaction_id INTEGER NOT NULL,
+                output_index INTEGER NOT NULL,
+                FOREIGN KEY (transaction_id) REFERENCES transactions(id_tx),
+                CONSTRAINT value_received_height UNIQUE (transaction_id, output_index)
+            );",
+        )?;
+
+        // Add estimated target height information for each transaction we know to
+        // have been created by the wallet; transactions that were discovered via
+        // chain scanning will have their `created` field set to `NULL`.
```
daira, line 56: What is the consequence of this being incorrect, e.g. when a non-default expiry delta was used?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -468,13 +480,50 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
...
+    conn: &rusqlite::Connection,
+    params: &P,
+) -> Result<Vec<TransactionDataRequest>, SqliteClientError> {
+    // We cannot construct transaction data requests for the case where we cannot determine the
+    // height at which to begin, so we require that either the target height or mined height be
+    // set. In practice, we should not encounter entries in `transparent_spend_search_queue`
+    // because under ordinary circumstances, it is populated via a call from
+    // `decrypt_and_store_transaction` on ordinary mined transaction data retrieved from the chain.
+    let mut address_request_stmt = conn.prepare_cached(
+        "SELECT ssq.address, IFNULL(t.target_height, t.mined_height + 1)
```
daira, line 503: Why is this `+ 1`?

daira (approved):
utACK with comments.

zcash_client_backend/src/data_api.rs
nuttycom: I used "may" because, with a future update to the lightwalletd protocol such as we have been discussing, this documentation will remain correct. And "may" is sufficient, because if there is any path by which such transactions would not be discovered, obtaining information for them via this path would still be necessary.

zcash_client_backend/src/data_api.rs
nuttycom: Same as above.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
nuttycom: This is a heuristic that is essentially targeted to catch _only_ ZIP-320 second transactions in wallet recovery:
* In the course of normal wallet operation, the wallet will have an entry in the `transactions` table already for that transaction, so we don't need to discover it.
* In the course of wallet recovery, there will ordinarily be change belonging to the wallet that can be used to further explore the transaction graph. The remaining case where history recovery will fail is if the full value of a transparent UTXO is spent outside of the expiry delta. To fix this, we still need to add functionality to compare "assumed UTXOs" to "actual UTXOs" but in order to do so correctly we need a lightwalletd API that looks up transactions by their input `OutPoint` values. When we have such an API, we can then fix this case properly. I think that there's a follow-up possible here that compares the actual UTXO set to the assumed UTXO set, and I've actually made a couple of attempts at this, but I don't think I have figured out a really good approach yet, so I would prefer not to block this PR on it.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: On reflection, I think this is an error - I was thinking that the output would not be spent in the same block it's created, but obviously that is not necessarily correct for spending transparent outputs.
```suggestion
        "SELECT ssq.address, IFNULL(t.target_height, t.mined_height)
```

zcash_client_sqlite/src/wallet.rs
daira: Filed #1485.

zcash_client_sqlite/src/lib.rs
```diff
@@ -587,6 +591,18 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletRead for W
             &address.encode(&self.params),
         )
     }
+
+    fn transaction_data_requests(&self) -> Result<Vec<TransactionDataRequest>, Self::Error> {
```
daira, line 595: This method doesn't seem to be covered by tests.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1180,7 +1202,28 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+            #[cfg(feature = "transparent-inputs")]
+            let detectable_via_scanning = {
+                #[allow(unused_mut)]
+                let mut detectable_via_scanning = d_tx.tx().sapling_bundle().is_some();
+                #[cfg(feature = "orchard")] {
+                    detectable_via_scanning |= d_tx.tx().orchard_bundle().is_some();
+                }
+
+                detectable_via_scanning
+            };
```
daira, line 1220:
```suggestion
            #[cfg(feature = "transparent-inputs")]
            let detectable_via_scanning = d_tx.tx().sapling_bundle().is_some();
            #[cfg(all(feature = "transparent-inputs", feature = "orchard"))] {
            let detectable_via_scanning = detectable_via_scanning | d_tx.tx().orchard_bundle().is_some();
```
This is slightly more elegant I think. Non-blocking.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2292,13 +2367,111 @@ pub(crate) fn put_tx_data(
...
+    let mut stmt_insert_tx = conn.prepare_cached(
+        "INSERT INTO tx_retrieval_queue (txid, query_type, dependent_transaction_id)
+        SELECT :txid, :query_type, :dependent_transaction_id
+        -- do not queue enhancement requests if we already have the raw transaction
+        WHERE NOT EXISTS (
+            SELECT 1 FROM transactions
+            WHERE txid = :txid
+            AND :query_type = :enhancement_type
+            AND raw IS NOT NULL
+        )
```
daira, line 2417:
```suggestion
        WHERE :query_type <> :enhancement_type OR NOT EXISTS (
            SELECT 1 FROM transactions
            WHERE txid = :txid
            AND raw IS NOT NULL
        )
```
(i.e. shortcut the `SELECT` if we don't need it). Non-blocking.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2292,13 +2367,111 @@ pub(crate) fn put_tx_data(
...
+    conn: &rusqlite::Transaction<'_>,
+    txids: impl Iterator<Item = TxId>,
+    query_type: TxQueryType,
+    dependent_tx_ref: Option<TxRef>,
+) -> Result<(), SqliteClientError> {
+    // Add an entry to the transaction retrieval queue if we don't already have raw transaction data.
+    let mut stmt_insert_tx = conn.prepare_cached(
+        "INSERT INTO tx_retrieval_queue (txid, query_type, dependent_transaction_id)
+        SELECT :txid, :query_type, :dependent_transaction_id
+        -- do not queue enhancement requests if we already have the raw transaction
```
daira, line 2411: Is it possible that we have the raw transaction but don't have its mined height?

daira: I think it is possible, because `wallet::put_tx_data` sets `raw` but not `mined_height`. Therefore this optimization is incorrect given that a Status request can be coalesced with an Enhancement request.

daira: Oh, I think it's fine. There are two nontrivial cases:
* If this is an Enhancement request and we have the raw tx, but there is an existing Status request for this txid, then the WHERE clause will be false and the SELECT won't match, so there will be no conflict and we will correctly leave the queue as-is.
* In any case where we already have an Enhancement request, this statement can't affect it. Even if it were possible to have an outstanding Enhancement request for a tx where `raw` is already set, it would not prevent the mined status from being updated by that request.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1974,70 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+            conn.execute(
+                "UPDATE transactions
+                 SET block = blocks.height
+                 FROM blocks
+                 WHERE txid = :txid
+                 AND blocks.height = :height",
+                sql_args,
+            )?;
+
+            notify_tx_retrieved(conn, txid)?;
```
daira, line 2037: There's a potential hazard here: `notify_tx_retrieved` will remove the txid from `tx_retrieval_queue` regardless of whether the request was for Status or Enhancement. But we haven't updated the `raw` field, so if it was Enhancement, then we'll forget that we need to enhance.

This could happen either with incorrect API usage by the wallet, or as a result of a race condition where the wallet is processing a Status request that gets upgraded to an Enhancement request, and the wallet doesn't notice the upgrade so still calls `set_transaction_status`. (This race condition can happen regardless of `set_transaction_status` and `store_decrypted_tx` being individually transactional.)

zcash_client_sqlite/src/lib.rs
nuttycom: I am working on an additional commit that adds more comprehensive testing.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1974,70 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+    conn: &rusqlite::Transaction,
+    txid: TxId,
+    status: TransactionStatus,
+) -> Result<(), SqliteClientError> {
+    match status {
+        TransactionStatus::TxidNotRecognized | TransactionStatus::NotInMainChain => {
+            // If the transaction is now expired, remove it from the retrieval queue.
+            if let Some(chain_tip) = chain_tip_height(conn)? {
+                conn.execute(
+                    "DELETE FROM tx_retrieval_queue WHERE txid IN (
```
daira, line 2002:
```suggestion
                    "DELETE FROM tx_retrieval_queue
                    WHERE query_type = :status_type AND txid IN (
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1974,70 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+                        SELECT txid FROM transactions
+                        WHERE txid = :txid AND expiry_height < :chain_tip_minus_lookahead
+                    )",
+                    named_params![
+                        ":txid": txid.as_ref(),
+                        ":chain_tip_minus_lookahead": u32::from(chain_tip).saturating_sub(VERIFY_LOOKAHEAD)
+                    ],
```
daira, line 2009:
```suggestion
                        ":status_type": TxQueryType::Status.code(),
                    ],
```

daira, line 2009: Same problem as [below](https://github.com/zcash/librustzcash/pull/1473/files#r1711616150)
```suggestion
                    "DELETE FROM tx_retrieval_queue
                     WHERE query_type = :status_type AND txid IN (
                         SELECT txid FROM transactions
                         WHERE txid = :txid AND expiry_height < :chain_tip_minus_lookahead
                     )",
                    named_params![
                        ":txid": txid.as_ref(),
                        ":chain_tip_minus_lookahead": u32::from(chain_tip).saturating_sub(VERIFY_LOOKAHEAD),
                        ":status_type": TxQueryType::Status.code(),
                    ],
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2292,13 +2367,110 @@ pub(crate) fn put_tx_data(
...
+                TxQueryType::Enhancement => TransactionDataRequest::Enhancement(txid),
+            })
+        })?
+        .collect::<Result<Vec<_>, _>>()?;
+
+    Ok(result)
+}
+
+pub(crate) fn notify_tx_retrieved(
+    conn: &rusqlite::Connection,
```
daira, line 2463: This should be required to be executed within a transaction.

zcash_client_sqlite/src/wallet.rs
nuttycom: *sigh* I had _just_ finished responding to your previous comment here when you deleted it, and now I've lost the contents of that response.

The status can be updated to "not recognized" via an enhancement request. This could occur if:
* A transaction for the wallet is detected via scanning.
* A rollback occurs, and the transaction is returned to the mempool.
* The transaction expires or is evicted without being re-mined.

In this situation, it is possible that an enhancement request _may never succeed_. We don't want to keep re-trying indefinitely.

zcash_client_sqlite/src/wallet.rs
nuttycom: Even though an `Enhancement` request subsumed a `Status` request, it should never be the case that such an upgrade occurs, because a request for `Enhancement` implies that we don't already have the raw transaction data.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1425,11 +1485,49 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                            // components, add it to the queue for status retrieval.
+                            #[cfg(feature = "transparent-inputs")]
+                            if d_tx.mined_height().is_none() && !detectable_via_scanning {
+                                wallet::queue_tx_retrieval(
+                                    wdb.conn.0,
+                                    std::iter::once(d_tx.tx().txid()),
+                                    TxQueryType::Status,
+                                    None
+                                )?;
+                            }
```
daira, line 1506: This can't work I think. At this point an Enhancement request for the txid (the one we're processing) is already on the queue. The Status request will be ignored and then the call to `notify_tx_retrieved` on line 1529 will remove the Enhancement request from the queue.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1425,11 +1485,49 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                    wallet::queue_tx_retrieval(
+                        wdb.conn.0,
+                        b.vin.iter().map(|txin| *txin.prevout.txid()),
+                        TxQueryType::Enhancement,
+                        Some(tx_ref)
+                    )?;
+                }
+            }
+
+            notify_tx_retrieved(wdb.conn.0, d_tx.tx().txid())?;
```
daira, line 1529: Since we're in a database transaction, I think [this problem](https://github.com/zcash/librustzcash/pull/1473/files#r1711669805) can probably be solved just by moving this call to be earlier.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2271,10 +2345,11 @@ pub(crate) fn put_tx_data(
...
+    target_height: Option<BlockHeight>,
+) -> Result<TxRef, SqliteClientError> {
     let mut stmt_upsert_tx_data = conn.prepare_cached(
-        "INSERT INTO transactions (txid, created, expiry_height, raw, fee)
-        VALUES (:txid, :created_at, :expiry_height, :raw, :fee)
+        "INSERT INTO transactions (txid, created, expiry_height, raw, fee, target_height)
+        VALUES (:txid, :created_at, :expiry_height, :raw, :fee, :target_height)
         ON CONFLICT (txid) DO UPDATE
         SET expiry_height = :expiry_height,
             raw = :raw,
```
daira, line 2355: Do we also need `target_height = IFNULL(:target_height, target_height)` in the ON CONFLICT clause?

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        [utxos_to_txos::MIGRATION_ID].into_iter().collect()
+    }
+
+    fn description(&self) -> &'static str {
+        "Adds a table for tracking transactions to be downloaded for transparent output and/or memo retrieval."
```
daira, line 27:
```suggestion
        "Adds tables for tracking transactions to be downloaded for transparent output and/or memo retrieval."
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -468,13 +480,50 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
...
+/// Returns the vector of [`TransactionDataRequest`]s that represents the information needed by the
+/// wallet backend in order to be able to present a complete view of wallet history and memo data.
+pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
+    conn: &rusqlite::Connection,
+    params: &P,
+) -> Result<Vec<TransactionDataRequest>, SqliteClientError> {
+    // We cannot construct transaction data requests for the case where we cannot determine the
+    // height at which to begin, so we require that either the target height or mined height be
+    // set. In practice, we should not encounter entries in `transparent_spend_search_queue`
+    // because under ordinary circumstances, it is populated via a call from
```
daira, line 500: Is there something missing here qualifying which entries we should not expect to encounter? Or does this mean that `transparent_spend_search_queue` is ordinarily empty?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -468,13 +480,50 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
...
+
+    let result = address_request_stmt
+        .query_and_then([], |row| {
+            let address = TransparentAddress::decode(params, &row.get::<_, String>(0)?)?;
+            let block_range_start = BlockHeight::from(row.get::<_, u32>(1)?);
+            Ok::<TransactionDataRequest, SqliteClientError>(
+                TransactionDataRequest::SpendsFromAddress {
+                    address,
+                    block_range_start,
+                    block_range_end: Some(block_range_start + DEFAULT_TX_EXPIRY_DELTA),
```
daira, line 518: Check for an off-by-one here. (Remember that the expiry height is the last block height at which the tx *can* be mined.)

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -640,14 +725,41 @@ pub(crate) fn put_transparent_output<P: consensus::Parameters>(
...
 
+pub(crate) fn queue_transparent_spend_detection<P: consensus::Parameters>(
+    conn: &rusqlite::Transaction<'_>,
+    params: &P,
+    receiving_address: TransparentAddress,
+    tx_ref: TxRef,
+    output_index: u32,
+) -> Result<(), SqliteClientError> {
+    // Add an entry to the transaction retrieval queue if we don't already have raw transaction
+    // data.
```
daira, line 744:
```suggestion
    // Add an entry to the transaction retrieval queue if we don't already
    // have one for this txout.
```
This doesn't optimize the case where we already have the raw transaction data. Should it?

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Remember that this is a heuristic upper bound; the expiry height of the transaction that created the output that we're looking for the spend of has no real bearing on when that output is actually spent; we don't know anything about the transaction that spends it (which is the transaction that we're looking for.)

As I've mentioned elsewhere, what we really need to be able to fix this correctly is to be able to query for spends of the outpoint, rather than just transactions related to the address it is sent to.

Something that just occurred to me is that there's another thing we should do here, which is explicitly check whether the outpoint is currently part of the UTXO set. If not, then we at least can definitely know that it's currently _not spendable_, which is the important thing.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Yes, good catch.

daira (changes requested):
NACK, there seem to be some bugs.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Oh, wait, no; this isn't a request for specific transaction data.

zcash_client_sqlite/src/wallet.rs
daira: I don't follow the argument. Why is it the case that we can't make a Status request and then later, before that is satisfied, see that we need an Enhancement request?

zcash_client_sqlite/src/lib.rs
nuttycom: Good catch; I think that the `notify_tx_retrieved` call needs to be guarded by `detectable_via_scanning`.

zcash_client_sqlite/src/wallet.rs
daira: > *sigh* I had *just* finished responding to your previous comment here when you deleted it, and now I've lost the contents of that response.

Sorry about that. I hadn't thought that deleting a comment deleted replies.

I don't understand your response. The suggestion adds "query_type = :status_type AND" to the query. This is clearly correct, because we only satisfied the Status request, not the Enhancement request. (We didn't update `raw`; that only happens via `put_tx_data` which is called from `store_decrypted_tx`.) So the question is only whether it's necessary, i.e. whether the situation I'm concerned about (a Status request being upgraded to an Enhancement request) can ever occur.

zcash_client_sqlite/src/wallet.rs
nuttycom: This should be set at the same time as `created_at`; I can't think of a situation where the target height would be set later.

zcash_client_sqlite/src/wallet.rs
nuttycom: We should only ever be making `Status` requests for transactions where we already have complete transaction data (i.e. an `Enhancement` request has already been satisfied.)

zcash_client_sqlite/src/wallet.rs
nuttycom: There is a unique constraint on `txid`; there can never be more than one entry for a given transaction in the retrieval queue.

zcash_client_backend/src/data_api.rs
```diff
@@ -541,6 +541,71 @@ pub struct SpendableNotes<NoteRef> {
...
+#[derive(Clone, Copy, Debug)]
+pub enum TransactionStatus {
+    /// The requested transaction ID was not recognized by the node.
+    TxidNotRecognized,
+    /// The requested transaction ID corresponds to a transaction that is recognized by the node,
+    /// but is in the mempool or is otherwise not mined in the main chain (but may have been mined
+    /// on a fork that was reorged away).
+    NotInMainChain,
+    /// The requested transaction ID corresponds to a transaction that has been included in the
+    /// block at the provided height.
```
daira, line 605:
```suggestion
    /// block at the provided height on the main chain.
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1811,9 +1898,27 @@ pub trait WalletWrite: WalletRead {
...
-        n: usize,
-    ) -> Result<Vec<(TransparentAddress, TransparentAddressMetadata)>, Self::Error>;
+        _account_id: Self::AccountId,
+        _n: usize,
+    ) -> Result<Vec<(TransparentAddress, TransparentAddressMetadata)>, Self::Error> {
+        // Default impl is required for feature-flagged trait methods to prevent
+        // breakage due to inadvertent activation of features by transitive dependencies
+        // of the implementing crate.
+        Ok(vec![])
+    }
```
daira, line 1908:
```suggestion
    ) -> Result<Vec<(TransparentAddress, TransparentAddressMetadata)>, Self::Error>;
```

zcash_client_backend/src/data_api.rs
nuttycom: Blocks only have a height when they're in the main chain.

zcash_client_sqlite/src/wallet.rs
daira: The comment added in d96bc110555cee26d405896e8e027e055061e654 explains why this is ok.

zcash_client_sqlite/src/wallet.rs
daira: The comment added in d96bc110555cee26d405896e8e027e055061e654 explains why this is ok.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Yes, `transparent_spend_search_queue` will be empty except in wallet recovery scenarios.

daira: I guess the comment doesn't really help me because I don't know what situations are "ordinary". The comment at https://github.com/zcash/librustzcash/blob/8752ad7e19a93b5b9dbc2976daa2f7a5b2ec5a29/zcash_client_sqlite/src/lib.rs#L1446-L1450 (in the only place where `queue_transparent_spend_detection` is called) is more helpful. I don't think it is restricted to wallet recovery; it can also happen if another wallet shares the spending key for the output.

zcash_client_sqlite/src/wallet/transparent.rs
daira: OK, but we might as well check for a potential off-by-one in the heuristic, anyway.

zcash_client_backend/src/data_api.rs
nuttycom: I would like to remove all of the noop default impls, but I think that this should be done as a separate commit once this has merged.

zcash_client_backend/src/data_api.rs
nuttycom: I will make this change as part of a wholesale removal of all the default impls.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: While I don't believe that wallets sharing a seed should be considered an ordinary circumstance, it can also occur with imported viewing capabilities, so I agree, I will update this comment.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: The block range start will ordinarily be the target height for the transaction, so if we want this to be end-exclusive we should add one.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1964,6 +1974,77 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+    conn: &rusqlite::Transaction,
+    txid: TxId,
+    status: TransactionStatus,
+) -> Result<(), SqliteClientError> {
+    // In order to have made a Status request, we must already have had the
+    // raw transaction (the only callers of `queue_tx_retrieval` are in contexts
+    // where we must have already called `put_tx_data`). Therefore, it is safe
+    // to unconditionally delete the request from `tx_retrieval_queue` below
+    // (both in the expired case and the case where it has been mined), because
+    // we already have all the data we need about this transaction.
```
str4d, line 2002: This is true for Status requests, but not true for Enhancement requests which subsume Status.

str4d (changes requested):
utACK 171cc3d18574ce7855bf4ce9c907d662244f3f1b once my concern about `Enhancement` results has been addressed (either by explaining why it's fine, or making a fix).

daira (approved):
utACK modulo suggested changes in https://github.com/nuttycom/librustzcash/pull/39 . We can either merge that PR into this one, or merge a copy of it separately. (I have checked that it passes tests locally.)

zcash_client_sqlite/src/wallet.rs
daira: Yes. The first commit of https://github.com/nuttycom/librustzcash/pull/39 fixes that.

zcash_client_backend/src/data_api.rs
```diff
@@ -1369,6 +1451,11 @@ impl<'a, AccountId> SentTransaction<'a, AccountId> {
     pub fn utxos_spent(&self) -> &[OutPoint] {
         self.utxos_spent
     }
+
+    /// Returns the block height that this transaction was created to target.
+    pub fn target_height(&self) -> BlockHeight {
```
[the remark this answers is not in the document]
nuttycom: `SentTransaction::new` is public, so it's possible for a third party to violate this, but we never make a "best guess."

zcash_client_sqlite/src/wallet.rs
nuttycom: It occurs to me that https://github.com/zcash/librustzcash/pull/1473/files#diff-a1724f7547ecabc613a07520eaf7917d9084d4277219d186698b0c8e352495a4R2425-R2427 is superfluous; we should never have a `Status` request for a transaction unless we have the raw data. So one way to fix this would be to remove the query type argument from the `queue_tx_retrieval` function, and to set it based upon the presence or absence of raw transaction data.

str4d: The comment change doesn't resolve my concern. `GetStatus` is indeed only generated when we have the raw transaction (as I noted), but `Enhancement` is generated by definition when we don't have the raw transaction, and its documentation states that when the caller is unable to look up the transaction, it must call this method.

The reason it is okay to delete for `Enhancement` requests is that:
- If a transaction can't be enhanced, we _assume_ that the caller calls this method with `TransactionStatus::TxidNotRecognized` (as the other two statuses mean that the chain data source knew about the transaction, and therefore must have observed it and could return its raw data), so we don't need to consider what happens in the other statuses.
- We only delete for `TransactionStatus::TxidNotRecognized` if the transaction is expired, at which point we don't _necessarily_ care about enhancement (though the user might still care about the memo), and it is very likely that we will be unable to ever enhance (as the tx will have been dropped from all mempools).

nuttycom: Simplified in https://github.com/zcash/librustzcash/pull/1473/commits/5a1645ac76199a571a5a9270e12114e553f019e1

nuttycom: It is always okay to delete _any_ sort of status request if a transaction is either (a) expired unmined, or (b) mined, in which case we know that we already have the transaction data, because we cannot generate a `GetStatus` request otherwise.

daira: I don't agree that 5a1645ac76199a571a5a9270e12114e553f019e1 is a simplification. It seems less explicit and requires more complicated reasoning about the sequence of states. The fact that we want to get the raw transaction when we don't have it and the status otherwise isn't fundamental, it's an implementation detail. We could have wanted to get both the status *and* the raw transaction in the case where we don't have the latter (which is what I originally thought we were doing until I looked closely at `store_decrypted_tx`).

nuttycom: The fact is that an `Enhancement` request _is_ a `Status` request. That's what it means for an `Enhancement` request to subsume a `Status` request.

str4d: @daira `Enhancement` _is_ getting the status; it subsumes / is a superset of `GetStatus`. That's why I raised my initial concern that this method was being written ostensibly assuming it would only be called via `GetStatus`, which was false.

As @nuttycom says, the actual implementation does correctly work with both `GetStatus` and `Enhancement`; the comment was just misleading.

zcash_client_sqlite/src/lib.rs
daira: Unresolved so that I can link to this thread in another comment.

zcash_client_sqlite/src/wallet.rs
nuttycom: I have now updated this comment again.

str4d (approved):
utACK b47c7bfe8d6b7c5f3442e1e77319b561dead8de6

zcash_client_sqlite/src/wallet.rs
daira: > The fact is that an Enhancement request *is* a Status request.

Not so. Otherwise it wouldn't have to requeue a Status request when `queue_status_retrieval` is true. It doesn't directly update the mined height in the case where `d_tx.mined_height().is_none()`.

nuttycom: > Otherwise it wouldn't have to requeue a Status request when queue_status_retrieval is true.

The re-queueing is so that the wallet continues to monitor for when the transaction is eventually mined. A relevant case here is that `store_decrypted_tx` was called with a transaction that was read from the mempool; we have complete transaction data, but we don't know yet whether the transaction will be mined, so we stick the status request back on the queue if we have no other way of detecting that it becomes mined.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1425,11 +1483,52 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
...
+                            // information for all of the transparent inputs to the transaction.
+                            #[cfg(feature = "transparent-inputs")]
+                            {
+                                tx_has_wallet_outputs = true;
+                            }
+
+                            // If the decrypted transaction is unmined and has no shielded
+                            // components, add it to the queue for status retrieval.
+                            #[cfg(feature = "transparent-inputs")]
+                            if d_tx.mined_height().is_none() && !detectable_via_scanning {
```
daira, line 1497: Why is this necessary? For an Enhancement request, the tx must have been retrieved via `GetTransaction`, which is supposed to return the current mined height (and despite https://github.com/zcash/lightwalletd/issues/493, we know that it will always return a height > 0 for a mined transaction). So we shouldn't need to make another query.

nuttycom: This method makes no assumptions about what has provided it with the transaction data. In particular, this is used in the tests, but it could also be used to introduce a purely-transparent transaction that was added from the mempool.

daira (commented):
Post-hoc ACK

PR #1475, opened 2024-07-31
corrected pool type for scan error

Oscar-Pepper:
small error I noticed while reading through scanning.rs


str4d (approved):
utACK 2929273ef196ebc3aca2438634c1612da683f3d2. Good catch, thanks!

PR #1478, opened 2024-07-31
[#1477] Allow Orchard-only UFVK be created from an Orchard FVK

### LINKED ISSUE #1476: Create `unstable-frost` feature
This feature will contain changes required for using FROST that will be feature-gated from the default features public API.

Related https://github.com/zcash/orchard/issues/430

### LINKED ISSUE #1477: Allow Orchard-only UFVK be created from an Orchard FVK
Allow creating a UnifiedFullViewingKey from a FullViewingKey. There is a new() method which could be used but it's gated under the test-dependencies feature.

uses `unstable-frost` #1476

related https://github.com/zcash/orchard/issues/430

pacu:
closes #1477
closes #1476

### DIFF
--- a/zcash_keys/CHANGELOG.md
+++ b/zcash_keys/CHANGELOG.md
@@ -6,8 +6,12 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 ### Notable changes
-- `zcash_keys` now supports TEX (transparent-source-only) addresses as specified
+- `zcash_keys`:
+  - Now supports TEX (transparent-source-only) addresses as specified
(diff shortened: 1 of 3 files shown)


conradoplg (commented):
LGTM, though if someone tries to implement FROST with Sapling they will need a `from_sapling_fvk` so it might make sense to just expose `new()`, but let's wait for ECC input

nuttycom (approved):
utACK, this is an unstable development API addition. Ordinarily we might use a compilation flag for this, but the change is trivial enough that I think it's fine.

Issue #1474, opened 2024-07-30
Precisely document same-UFVK errors on account creation and ensure they are per-component


daira:
This is an issue with #1465 found in post-hoc review.

_Originally posted by @daira in https://github.com/zcash/librustzcash/pull/1465#discussion_r1695972600_

> Should this say "If an account created via [`WalletWrite::create_account`] or imported via [`WalletWrite::import_account_ufvk`] or [`WalletWrite::import_account_hd`] already exists in the wallet with the same UFVK (which will also be the case for accounts with the same seed), an error will be returned." ?
>
> This differs from the current wording because (I think) it is not the case that the creation or import has to be in any particular order; any of these three methods will fail if it produces a duplicate UFVK, regardless of which of them was used to create the existing account. At least, if that's not the case then it is a bug.
>
> Also, do we actually mean "same UFVK" here, or do we mean that there will be a failure if *any component* of the UFVK matches that of an existing account? In the case of an account where we don't know the seed, the imported UFVK can't include components that are not known yet, and *may not* include some existing components (I don't know whether the latter is supported).

_Originally posted by @daira in https://github.com/zcash/librustzcash/pull/1465#discussion_r1695982562_

> I think the doc for each of the three account creation/import methods should call out that it fails when there is a duplicate UFVK, even if that's redundant with the `WalletWrite` doc. Note that it is possible to hit this case even if you only use one of the three methods, and there might be implementors who only read the documentation of the methods they are using (because they haven't reached [this page](https://hmpg.net/) yet).

str4d:
> Also, do we actually mean "same UFVK" here, or do we mean that there will be a failure if any component of the UFVK matches that of an existing account? In the case of an account where we don't know the seed, the imported UFVK can't include components that are not known yet, and may not include some existing components (I don't know whether the latter is supported).

Currently, the three uniqueness constraints that `zcash_client_sqlite` implements are:
- same `(seed, ZIP 32 account index)`
- same UFVK
- same UIVK (which we can consider equivalent to "same UFVK")
https://github.com/zcash/librustzcash/blob/7f7b685b99132505a0fe4ba588f329e3516e9669/zcash_client_sqlite/src/wallet/db.rs#L25-L34
https://github.com/zcash/librustzcash/blob/7f7b685b99132505a0fe4ba588f329e3516e9669/zcash_client_sqlite/src/wallet/db.rs#L55-L60
[...]

PR #1479, opened 2024-07-31
Make `WalletWrite` documentation for colliding accounts more precise

str4d:
Part of #1474.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1583,11 +1583,20 @@ impl AccountBirthday {
 /// - If [`WalletWrite::import_account_hd`] is used to import accounts with non-sequential
 ///   ZIP 32 account indices from the same seed, a call to [`WalletWrite::create_account`]
 ///   will use the ZIP 32 account index just after the highest-numbered existing account.
-/// - If an account is imported via [`WalletWrite::import_account_ufvk`], and then a later
-///   call to either [`WalletWrite::create_account`] or [`WalletWrite::import_account_hd`]
-///   would produce the same UFVK, an error will be returned.
-///   - A future change to this trait might introduce a method to "upgrade" an imported
-///     account with derivation information. See [zcash/librustzcash#1284] for details.
+/// - If an account is added to the wallet, and then a later call to one of the methods
+///   would produce the same UFVK, an error will be returned. While any combination of
+///   method calls can cause this, there are two main ways it can occur that users of the
+///   `WalletWrite` trait should watch out for:
+///   - An account is created via [`WalletWrite::create_account`] with an auto-selected
+///     ZIP 32 account index, and that index is later imported explicitly via either
+///     [`WalletWrite::import_account_hd`] or [`WalletWrite::import_account_ufvk`].
+///   - An account is imported via [`WalletWrite::import_account_ufvk`], and the ZIP 32
+///     account index corresponding to that account's UFVK is later imported either
+///     implicitly via [`WalletWrite::create_account`] or explicitly via
+///     [`WalletWrite::import_account_hd`].
+///
+/// A future change to this trait might introduce a method to "upgrade" an imported
+/// account with derivation information. See [zcash/librustzcash#1284] for details.
 ///
 /// Users of the `WalletWrite` trait should generally distinguish in their APIs and wallet
 /// UIs between creating a new account, and importing an account that previously existed.


zcash_client_backend/src/data_api.rs
```diff
@@ -1583,11 +1583,20 @@ impl AccountBirthday {
...
+///   would produce the same UFVK, an error will be returned. While any combination of
+///   method calls can cause this, there are two main ways it can occur that users of the
+///   `WalletWrite` trait should watch out for:
+///   - An account is created via [`WalletWrite::create_account`] with an auto-selected
+///     ZIP 32 account index, and that index is later imported explicitly via either
+///     [`WalletWrite::import_account_hd`] or [`WalletWrite::import_account_ufvk`].
+///   - An account is imported via [`WalletWrite::import_account_ufvk`], and the ZIP 32
+///     account index corresponding to that account's UFVK is later imported either
+///     implicitly via [`WalletWrite::create_account`] or explicitly via
+///     [`WalletWrite::import_account_hd`].
```
daira, line 1596: I don't understand why these two cases are being called out. Why is it less likely that you call the same method twice with the same arguments, for instance?

daira: Note that this is independent of the issue described in https://github.com/zcash/librustzcash/issues/1474#issuecomment-2261507844 that we don't currently detect collisions between non-identical UFVKs with shared components. The three cases where an account creation method collides with another invocation of itself, and the two remaining cases not covered above (`import_account_hd`, then `create_account` or `import_account_ufvk`) can all currently occur.

daira, line 1596:
```suggestion
///   would produce the same UFVK, an error will be returned. This can occur for any
///   combination of [`WalletWrite::create_account`], [`WalletWrite::import_account_hd`],
///   and [`WalletWrite::import_account_ufvk`] as the first and second calls. Note that
///   accounts having UFVKs that are not identical but have shared components (for
///   example, two accounts having the same Sapling FVK, one of which also has an
///   Orchard FVK while the other does not) are currently allowed. This may not be
///   allowed in future.
```

str4d: > I don't understand why these two cases (which represent 4 out of the 9 possible combinations of (method to create first account, method to create colliding account)) are being specifically called out.

Because those are (I think) the most likely cases that users of the `WalletWrite` trait may accidentally encounter without realising.

> Why is it less likely that you call the same method twice with the same arguments, for instance?

Because:
- this cannot happen with the `create_account` API.
- this is very obvious for the `import_account_*` APIs.

str4d: > Note that this is independent of the issue described in https://github.com/zcash/librustzcash/issues/1474#issuecomment-2261507844 that we don't currently detect collisions between non-identical UFVKs with shared components.

Yes I'm well aware of that, and this PR is not attempting to address that issue.

str4d: NACK, I think we _should_ decide how to handle this, and _not_ in this PR.

daira: > > Why is it less likely that you call the same method twice with the same arguments, for instance?
> 
> Because:
> 
> * this cannot happen with the `create_account` API.

That is not correct. Calling `create_account` with the same seed will result in a duplicate UFVK. You may say "the intended use of the API is to call it with a fresh random seed" (in which case a collision is obviously vanishingly unlikely), but nothing requires that and the API docs don't mention it.

> * this is very obvious for the `import_account_*` APIs, and thus I don't think needs to be called out here (for the complexity reason I give above).

The wording that avoids calling these 4 out of 9 cases as in any way special is *less* complex. As it is currently written, it immediately makes me wonder "what is different about these cases, that I am supposed to pay special attention to them?" And the answer is, nothing.

daira: Absolutely we should, but this documents the current behaviour correctly. Then in the PR where we fix it (which I agree, should be before the release), we can change the comment accordingly.

str4d: > > > Why is it less likely that you call the same method twice with the same arguments, for instance?
> > 
> > Because:
> > 
> > * this cannot happen with the `create_account` API.
>
> That is not correct. Calling `create_account` with the same seed will result in a duplicate UFVK. You may say "the intended use of the API is to call it with a fresh random seed" (in which case a collision is obviously vanishingly unlikely), but nothing requires that and the API docs don't mention it.

That is not correct. Calling `create_account` with the same seed will result in the next UFVK being generated, corresponding to the next unused ZIP 32 account index (or an error if you've hit the end of the account range). Sequential calls to `create_account` with the same arguments _never_ result in the same changes to the wallet database.

daira: The doc still omits cases. For example, you can import an account with `import_account_ufvk` first, and then `create_account` with the same seed can fail because `max_zip32_account_index` will not find the UFVK-imported account, since the wallet doesn't yet know the seed.

str4d: That's... literally the case documented in the bullet point that this thread is anchored to?
> - **An account is imported via `WalletWrite::import_account_ufvk`, and the ZIP 32 account index corresponding to that account's UFVK is later imported either implicitly via `WalletWrite::create_account`** or explicitly via `WalletWrite::import_account_hd`.

daira (approved):
Post-hoc ACK

zcash_client_backend/src/data_api.rs
```diff
@@ -1583,11 +1583,20 @@ impl AccountBirthday {
...
+///
```
daira, line 1597:
```suggestion
///
/// Note that accounts having UFVKs that are not identical but have shared
/// components (for example, two accounts having the same Sapling FVK, one
/// of which also has an Orchard FVK while the other does not) are currently
/// allowed. This will not be allowed in future.
```

@str4d wrote (in response to slightly different wording of the last sentence that said "may not" rather than "will not"):
> NACK, I think we should decide how to handle this, and not in this PR.

Absolutely we should, but this documents the current behaviour correctly. Then in the PR where we fix it (which I agree, should be before the release), we can and should change the comment accordingly.

daira (commented):
Updated my comments, and split the suggested changes for clarity.

zcash_client_backend/src/data_api.rs
daira: These suggestions are included in #1489, which then goes on to fix it and change the comment.

daira:
#1489 has been merged including these changes.

PR #286, opened 2020-09-01
Implementation for ZIP-222 Transparent Zcash Extensions

nuttycom:
This implements the core data structures and transaction builder function for TZEs. 

ZIP: https://github.com/zcash/zips/pull/248

PR #1480, opened 2024-08-01
zcash_protocol: Set the `ZFuture` consensus branch ID to agree with zcashd

nuttycom:


components/zcash_protocol/src/consensus.rs
```diff
@@ -593,7 +593,7 @@ impl From<BranchId> for u32 {
             #[cfg(zcash_unstable = "nu6")]
             BranchId::Nu6 => 0xc8e7_1055,
             #[cfg(zcash_unstable = "zfuture")]
-            BranchId::ZFuture => 0xffff_ffff,
+            BranchId::ZFuture => 0x7fff_ffff,
```
str4d, line 596: Looking at the git record:
- `0xffff_ffff` was used for ZFuture in the Rust code in #286, opened September 2020 and merged October 2020.
- `0x7fff_ffff` was used for ZFuture in `zcashd` in https://github.com/zcash/zcash/pull/4988, opened February 2021 and merged April 2021. However, it was factored out from https://github.com/zcash/zcash/pull/4480, opened April 2020 and not yet merged.

So for some reason, both consensus branch IDs came out of the same TZE work but ended up differing by a single bit.

daira (approved):
utACK. This has no effect on consensus for current mainnet and testnet, where "zfuture" must not be enabled.

nuttycom:
This is completely wrong; there was a confusion between `nProtocolVersion` for ZFuture and the consensus branch ID.

PR #1481, opened 2024-08-02
update doc-comment, to reflect existence of more shielded pools

zancas:


nuttycom (changes requested):
NACK, this change should be made in the compact_formats.proto file. This documentation is generated from that.

PR #1486, opened 2024-08-09
Add the ability to obtain the account type (spending vs view-only) from `AccountSource`

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -154,7 +155,12 @@ fn parse_account_source(
...
             })?,
         }),
-        (1, None, None) => Ok(AccountSource::Imported),
+        (1, None, None) => Ok(AccountSource::Imported {
+            account_type: AccountType::ViewOnly,
+        }),
+        (2, None, None) => Ok(AccountSource::Imported {
+            account_type: AccountType::Spending,
+        }),
         (0, None, None) | (1, Some(_), Some(_)) => Err(SqliteClientError::CorruptedData(
```
daira, line 164:
```suggestion
        (0, None, None) | (1, Some(_), Some(_)) | (2, Some(_), Some(_)) => Err(SqliteClientError::CorruptedData(
```

daira (dismissed):
utACK with minor suggestion.

zcash_client_sqlite/src/wallet.rs
nuttycom: Ah, crud, this actually means a database constraint (that I didn't really want in the first place) has to change. Grrr.

zcash_client_backend/CHANGELOG.md
```diff
@@ -86,8 +88,12 @@ funds to those addresses. See [ZIP 320](https://zips.z.cash/zip-0320) for detail
   - `DecryptedTransaction::new` takes an additional `mined_height` argument.
   - `SentTransaction` now stores its `outputs` and `utxos_spent` fields as
     references to slices, with a corresponding change to `SentTransaction::new`.
+<<<<<<< HEAD
   - `SentTransaction` takes an additional `target_height` argument, which is used
     to record the target height used in transaction generation.
+=======
+  - `AccountSource::Imported` is now a struct variant with an `account_type` field.
+>>>>>>> 0c46dd678 (zcash_client_backend: Add an `account_type` modifier for imported accounts.)
```
str4d, line 96: Merge conflict mis-resolution

zcash_client_backend/src/data_api.rs
```diff
@@ -321,6 +321,17 @@ impl AccountBalance {
     }
 }
 
+/// An enumeration used to control what information is tracked by the wallet for
+/// notes received by a given account.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum AccountType {
```
str4d, line 327: Naming is hard, but I don't like the name `AccountType`. We went through a similar discussion before to arrive at `AccountSource`, which is "the kinds of accounts supported". Kind and type are synonyms to users, so this new `AccountType` confuses the public API.

I think a better name would be something like `AccountUsage / AccountUse`, as this enum defines how the account can be used by the wallet user (due to it influencing what the wallet tracks). I also wondered about `AccountPurpose`, but that feels more inherent to the account, whereas this enum is about the account in the context of this particular wallet.

nuttycom: `AccountPurpose` seems good to me, let's go with that; we're talking about the wallet account, not the ZIP 32 account, and in the context of the wallet this is indeed what the purpose is for having the account in the wallet.

zcash_client_backend/src/data_api.rs
```diff
@@ -364,7 +385,9 @@ impl<A: Copy> Account<A> for (A, UnifiedFullViewingKey) {
     }
 
     fn source(&self) -> AccountSource {
-        AccountSource::Imported
+        AccountSource::Imported {
+            account_type: AccountType::ViewOnly,
+        }
     }
```
str4d, line 391: Should `AccountType` (or whatever it becomes named) be embedded in `AccountSource`? It feels more orthogonal: the source of an account does not necessarily constrain how it can be used within the wallet. Maybe a separate `Account::usage() -> AccountUsage` trait method would be preferable? (That would also have a closer correlation to the current `zcash_client_sqlite` database format, but that on its own should not be a deciding factor here.)

nuttycom: My thought here is that the purpose of the account is specified when describing its source, and it's natural to also retrieve it that way. An `Account::purpose` convenience method would be fine though.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/5a3a9ba82854a8372ab8f0304cbb43271a89f0b5..93809601bca5d978a9e77c7b48de3b3c9ea83807) to address comments from review.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/5a3a9ba82854a8372ab8f0304cbb43271a89f0b5..917f500) to address comments from review.

zcash_client_backend/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ funds to those addresses. See [ZIP 320](https://zips.z.cash/zip-0320) for detail
   - `DecryptedTransaction::mined_height`
   - `TransactionDataRequest`
   - `TransactionStatus`
+  - `AccountType`
+  - `AccountSource::account_type`
```
str4d, line 34: Changes to enums need to go under `Changed`, and this one is already there.
```suggestion
```

zcash_client_sqlite/src/wallet/init/migrations/full_account_ids.rs
```diff
@@ -53,7 +56,9 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
             seed_fingerprint: SeedFingerprint::from_bytes([0; 32]),
             account_index: zip32::AccountId::ZERO,
         });
-        let account_kind_imported = account_kind_code(AccountSource::Imported);
+        let account_kind_imported = account_kind_code(AccountSource::Imported {
+            purpose: AccountPurpose::ViewOnly,
```
str4d, line 60: Add a comment that the value of this `purpose` field is irrelevant to `account_kind_code`, and chosen arbitrarily.

zcash_client_backend/src/data_api.rs
```diff
@@ -343,6 +354,14 @@ pub trait Account<AccountId: Copy> {
     /// if applicable.
     fn source(&self) -> AccountSource;
 
+    /// Returns whether the account is a spending account or a view-only account.
+    fn purpose(&self) -> AccountPurpose {
```
str4d, line 358: Document this addition in the changelog.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -118,12 +116,29 @@ pub(crate) fn get_ephemeral_ivk<P: consensus::Parameters>(
     params: &P,
     account_id: AccountId,
 ) -> Result<EphemeralIvk, SqliteClientError> {
-    Ok(get_account(conn, params, account_id)?
-        .ok_or(SqliteClientError::AccountUnknown)?
-        .ufvk()
+    let ufvk = conn
+        .query_row(
+            "SELECT ufvk FROM accounts WHERE id = :account_id",
+            named_params![":account_id": account_id.0],
```
str4d, line 122: Why is this being altered to not use `get_account`? AFAICT none of the changes in this PR affect this code, and it should still work fine as it was before?

The refactored code is also subtly different. Previously this would return `Err(SqliteClientError::AccountUnknown)` for an unknown account; now it returns `Err(SqliteClientError::UnknownZip32Derivation)` (merging the two failure cases).

str4d (dismissed):
utACK 917f500069581b9f29ef0d93680911a4bb9160d0 after doc fixes.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
nuttycom: This is a bit annoying: the call here to `get_account` now breaks when used in migration tests for migrations prior to the addition of the `has_spending_key` field; now that we retrieve the value of that field in `get_account`, the migration tests break. Instead of having to copy a stack of business logic into the migration tests, minimizing this to query for only what it needs seemed like the better approach; the code here is pretty trivial.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/917f500069581b9f29ef0d93680911a4bb9160d0..ac7cbf9a415b5700f0241d44b4364d9324c43aa6) to address additional review comments.

str4d (approved):
utACK ac7cbf9a415b5700f0241d44b4364d9324c43aa6

daira (commented):
Post-hoc ACK

PR #1487, opened 2024-08-10
Improve the accuracy of a comment in `send_multi_step_proposed_transfer`

daira:
Trivial doc-only follow-up to #1473.


nuttycom (approved):
ACK, doc-only

str4d:
Post-hoc ACK

PR #1489, opened 2024-08-10
Check for colliding component FVKs

### LINKED ISSUE #1474: Precisely document same-UFVK errors on account creation and ensure they are per-component

daira:
When adding an account, check whether any component of its UFVK (if available) collides with an existing imported or derived FVK.

This does not check for collisions on IVK for `Incoming` accounts (#1490), which is okay because those are not yet exposed in the public API.

Builds on #1479, fixes #1474

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -2022,15 +2022,19 @@ extern crate assert_matches;
 
 #[cfg(test)]
 mod tests {
-    use secrecy::{Secret, SecretVec};
+    use secrecy::{ExposeSecret, Secret, SecretVec};
     use zcash_client_backend::data_api::{
         chain::ChainState, Account, AccountBirthday, AccountPurpose, AccountSource, WalletRead,
         WalletWrite,
     };
-    use zcash_keys::keys::UnifiedSpendingKey;
+    use zcash_keys::keys::{UnifiedFullViewingKey, UnifiedSpendingKey};
     use zcash_primitives::block::BlockHash;
 
-    use crate::{error::SqliteClientError, testing::TestBuilder, AccountId, DEFAULT_UA_REQUEST};
+    use crate::{
+        error::SqliteClientError,
+        testing::{TestBuilder, TestState},
+        AccountId, DEFAULT_UA_REQUEST,
+    };
 
     #[cfg(feature = "unstable")]
     use {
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -427,8 +435,9 @@ pub(crate) fn add_account<P: consensus::Parameters>(
             rusqlite::Error::SqliteFailure(f, s)
                 if f.code == rusqlite::ErrorCode::ConstraintViolation =>
             {
-                // An account conflict occurred.
-                // Make a best effort to determine the AccountId of the pre-existing row
+                // An account conflict occurred. This should already have been caught by
+                // the check using `get_account_for_ufvk` above, but in case it wasn't,
+                // make a best effort to determine the AccountId of the pre-existing row
                 // and provide that to our caller.
```
daira, line 441: This code will no longer be reached when a collision occurs, which explains the small reduction in test coverage.

zcash_client_sqlite/src/lib.rs
```diff
@@ -2198,13 +2260,16 @@ mod tests {
...
 
         let seed = Secret::new(vec![0u8; 32]);
+        let zip32_index_0 = zip32::AccountId::ZERO;
         let seed_based = st.wallet_mut().create_account(&seed, &birthday).unwrap();
         let seed_based_account = st.wallet().get_account(seed_based.0).unwrap().unwrap();
         let ufvk = seed_based_account.ufvk().unwrap();
 
         assert_matches!(
-            st.wallet_mut().import_account_ufvk(ufvk, &birthday, AccountPurpose::Spending),
+            st.wallet_mut().import_account_hd(&seed, zip32_index_0, &birthday),
```
daira, line 2269: The `import_account_ufvk` check is done in `check_collisions` below, so this is adding an extra check for collisions between `create_account` and `import_account_hd`.

nuttycom (approved):
utACK bd38645da54cfaa24935cd0001bef5a1a0054eb2

str4d:
Post-hoc utACK bd38645da54cfaa24935cd0001bef5a1a0054eb2

PR #1491, opened 2024-08-11
Ensure that `zcash_client_sqlite` only enables "orchard" for `zcash_keys` when its own "orchard" feature is enabled

### LINKED ISSUE #1488: zcash_client_sqlite always enables the "orchard" feature of zcash_keys
From `zcash_client_sqlite`'s `Cargo.toml`:

```toml
[dependencies]
...
zcash_keys = { workspace = true, features = ["orchard", "sapling"] }

[features]
...
orchard = ["dep:orchard", "zcash_client_backend/orchard", "zcash_keys/orchard"]
```
This will always enable the "orchard" feature of zcash_keys, because "orchard" is specified unconditionally in `[dependencies]` (so the conditional feature dependency in `[features]` is redundant). I've verified this using `cargo -v` to show the `rustc` command lines.

Intentional or bug?

daira:
fixes #1488

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -373,12 +373,17 @@ pub(crate) fn add_account<P: consensus::Parameters>(
         AccountSource::Imported { purpose } => (None, None, purpose == AccountPurpose::Spending),
     };
 
+    #[cfg(feature = "orchard")]
     let orchard_item = viewing_key
         .ufvk()
         .and_then(|ufvk| ufvk.orchard().map(|k| k.to_bytes()));
+    #[cfg(not(feature = "orchard"))]
+    let orchard_item: Option<Vec<u8>> = None;
+
     let sapling_item = viewing_key
         .ufvk()
         .and_then(|ufvk| ufvk.sapling().map(|k| k.to_bytes()));
+
     #[cfg(feature = "transparent-inputs")]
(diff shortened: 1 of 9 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -685,6 +690,13 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
...
+    #[cfg(feature = "orchard")]
+    let orchard_item = ufvk.orchard().map(|k| k.to_bytes());
+    #[cfg(not(feature = "orchard"))]
+    let orchard_item: Option<Vec<u8>> = None;
+
+    let sapling_item = ufvk.sapling().map(|k| k.to_bytes());
+
     #[cfg(feature = "transparent-inputs")]
     let transparent_item = ufvk.transparent().map(|k| k.serialize());
     #[cfg(not(feature = "transparent-inputs"))]
```
daira, line 702: I thought about writing a helper function for this. But given that it would only be used here and in two of the migrations, and migrations are supposed to be stable (so they should not be changed if, e.g., we add another shielded protocol), it didn't seem worthwhile.

nuttycom (approved):
utACK a021b9f2776e20f8971139d4f6bdb071ffbf5143

str4d:
Post-hoc utACK

PR #1496, opened 2024-08-15
zcash_client_sqlite: Track all transparent spends.

nuttycom:
Prior to this change, the `mark_transparent_utxo_spent` method assumed that the UTXO information for outputs belonging to the wallet would be known to exist before their spends could be detected. However, this is not true in transparent history recovery: the spends are detected first. We now cache the information about those spends so that we can then correctly record the spend when the output being spent is eventually detected.

At present, data from this spend cache is never deleted. This is because such deletions could undermine history recovery in some narrow cases related to chain reorgs.


daira (dismissed):
utACK

daira (approved):
utACK

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2311,6 +2314,371 @@ pub(crate) fn put_block(
...
+                        account_id,
+                        tx_ref,
+                        output.index(),
+                        &recipient,
+                        output.note_value(),
+                        Some(output.memo()),
+                    )?;
+                }
+            }
+        }
```
daira, line 2514: This method is way too long. How about factoring out this to `mark_orchard_output_spent`, analogous to `transparent::mark_transparent_utxo_spent`? (And similarly for Sapling.)

nuttycom: I agree that this could be factored out, although `mark_orchard_output_spent` is not at all the semantics. However, I don't think we need to worry about that now.

nuttycom: In any case, this just shows up as new here because of the move in  https://github.com/zcash/librustzcash/pull/1496/commits/75f236a148c936c34bfaa96f0986780695b7a231 - it shouldn't be moved and refactored at once.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -62,12 +79,44 @@ impl RusqliteMigration for Migration {
...
+                    WalletMigrationError::CorruptedData(
+                        "Could not read serialized transaction data.".to_owned(),
+                    )
+                })?;
+
+                wallet::store_decrypted_tx(
+                    transaction,
+                    &self.params,
+                    DecryptedTransaction::new(mined_height, &tx, vec![], vec![]),
+                )?;
```
daira, line 109: Hmm. This is a very complicated function to be calling in a migration. If we need to change it, will we duplicate it like we do for code called in other migrations?

daira (approved):
utACK

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
nuttycom: What I would like to do with `store_decrypted_tx` in general is to refactor it to abstract over the database operations (i.e. factor out the business logic) and perhaps even move the main logic to `zcash_client_backend` (relying upon some lower-level data access API trait.) That will make it easier to modularize it in a way that makes the migration maintenance less annoying.

daira (dismissed):
utACK

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -494,6 +517,11 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
     conn: &rusqlite::Connection,
     params: &P,
 ) -> Result<Vec<TransactionDataRequest>, SqliteClientError> {
+    // `lightwalletd` will return an error for `GetTaddressTxids` requests having an end height
+    // greater than the current chain tip height, so we take the chain tip height into account
+    // here in order to make this pothole easier for clients of the API to avoid.
+    let chain_tip_height = super::chain_tip_height(conn)?;
```
daira, line 523: Incidentally, why does the doc for `wallet::chain_tip_height` say
```rust
/// Returns the minimum and maximum heights of blocks in the chain which may be scanned.
```
when it actually only returns (optionally) a single height?

daira: Filed #1498.

daira (approved):
utACK

PR #1506, opened 2024-08-21
zcash_client_sqlite: Fix migration DAG edges

str4d:


zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -157,4 +216,46 @@ mod tests {
...
+            assert!(prev_state != new_state);
+            prev_state = new_state;
+        };
+
+        for migrations in super::PUBLIC_MIGRATION_STATES {
+            assert_matches!(
+                init_wallet_db_internal(
+                    &mut db_data,
+                    Some(Secret::new(seed.clone())),
+                    migrations,
```
nuttycom, line 257: Does this cause the migrations to be run in a particular order? Because if so, this test would still potentially miss this bug; in order to catch it, we would need to run all of the permutations of the leaf migrations.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue.rs
```diff
@@ -133,10 +135,89 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+        let data_file = NamedTempFile::new().unwrap();
+        let mut db_data = WalletDb::for_path(data_file.path(), Network::TestNetwork).unwrap();
+
+        let seed_bytes = vec![0xab; 32];
+
+        // Migrate to database state just prior to this migration.
+        init_wallet_db_internal(
+            &mut db_data,
+            Some(Secret::new(seed_bytes.clone())),
+            &DEPENDENCIES,
```
nuttycom, line 173: This is a good way to approach migration tests; it ensures that _only_ the nodes in the graph leading up to this node are evaluated, so anything that's missing should get caught. It does require inspecting what data is required in order to test all the possible branches within the migration though.

nuttycom (approved):
utACK modulo clippy failures.

nuttycom (dismissed):
ACK https://github.com/zcash/librustzcash/pull/1506/commits/e4b3fb27b599491126abc31bf13c2b60f626d107

zcash_client_sqlite/src/wallet/init/migrations.rs
str4d: The migration order permutation issue does need fixing, but not in this release.

daira (changes requested):
Just registering a block on merging because I'm doing an exhaustive analysis of whether `tx_retrieval_queue` depends semantically on any of the v0.11.1 leaf migrations (which the test will not necessarily catch). It shouldn't take too long.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e4b3fb27b599491126abc31bf13c2b60f626d107..498dfe5e174296e4c2b975e305930dd663c908d9) to make `tx_retrieval_queue` depend upon all migration leaves.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -131,8 +132,66 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
 
+/// All states of the migration DAG that have been exposed in a public crate release, in
+/// the order that crate users would have encountered them.
+///
+/// Omitted versions had the same migration state as the first prior version that is
+/// included.
+#[allow(dead_code)]
+const PUBLIC_MIGRATION_STATES: &[&[Uuid]] = &[
+    V_0_4_0, V_0_6_0, V_0_8_0, V_0_9_0, V_0_10_0, V_0_10_3, V_0_11_0,
+];
```
daira:
```suggestion
    V_0_3_0, V_0_4_0, V_0_6_0, V_0_8_0, V_0_9_0, V_0_10_0, V_0_10_3, V_0_11_0,
];

/// Leaf migrations in the 0.3.0 release.
/// (This release did not use the migration architecture, but this migration id
/// corresponds to its state.)
const V_0_3_0: &[Uuid] = &[initial_setup::MIGRATION_ID];

File: zcash_client_sqlite/src/wallet/init/migrations.rs [hunk 4]
```diff
@@ -131,8 +132,66 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
+
+/// Leaf migrations in the 0.4.0 release.
+const V_0_4_0: &[Uuid] = &[add_transaction_views::MIGRATION_ID];
```
daira, line 148: I verified that v0.4.2 did not add migrations relative to v0.4.0, and at that point (v0.4.2) the ASCII art was correct.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -131,8 +132,66 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
+
+/// Leaf migrations in the 0.6.0 release.
+const V_0_6_0: &[Uuid] = &[v_transactions_net::MIGRATION_ID];
```
daira, line 151: I verified that as of v0.6.0 the ASCII art was correct, and `v_transactions_net` was the only leaf.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -131,8 +132,66 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
+
```
daira, line 152: However, at v0.7.0 a [`received_notes_nullable_nf` migration](https://github.com/zcash/librustzcash/blob/zcash_client_sqlite-0.7.0/zcash_client_sqlite/src/wallet/init/migrations.rs#L48) was added that was not documented in the ASCII art.
```suggestion

/// Leaf migrations in the 0.7.0 release.
const V_0_7_0: &[Uuid] = &[received_notes_nullable_nf::MIGRATION_ID];

```
Non-blocking.

daira (approved):
utACK

PR #1509, opened 2024-08-22
Make the `{fixed,standard,zip317}::SingleOutputChangeStrategy` doc comments more accurate concerning the choice of output pool

daira:
Thanks to @AArnott for pointing out the inaccuracy.


nuttycom (approved):
ACK, doc-only.

PR #1511, opened 2024-08-22
Support older sqlite versions.

nuttycom:


zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -428,9 +428,8 @@ pub(crate) fn mark_ephemeral_address_as_seen<P: consensus::Parameters>(
          LEFT OUTER JOIN ephemeral_addresses e
          ON id_tx = e.seen_in_tx
          WHERE id_tx = :tx_ref OR e.address = :address
-         ORDER BY mined_height ASC NULLS LAST,
-                  tx_index ASC NULLS LAST,
-                  e.seen_in_tx ASC NULLS LAST
+         ORDER BY mined_height IS NULL, tx_index IS NULL, e.seen_in_tx IS NULL,
+                  mined_height ASC, tx_index, e.seen_in_tx ASC
```
daira, line 432: This isn't equivalent.
```suggestion
         ORDER BY mined_height IS NULL, mined_height ASC,
                  tx_index IS NULL, tx_index ASC,
                  e.seen_in_tx IS NULL, e.seen_in_tx ASC
```

nuttycom:
It turns out that Google is requiring Android API 34 anyway, so this PR is not needed: https://support.google.com/googleplay/android-developer/answer/11926878?hl=en

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/260c30254d5d65cb5e6b909f290d3af152fe7499..16db593fe540daaf8ca6e1e994b5a45ee67b8e22) to remove changes that are unnecessary as we now verify that we're using SQLite 3.35.0 or higher.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/260c30254d5d65cb5e6b909f290d3af152fe7499..d202d47db086176420092b1c8f1bd0a6121cb5e0) to remove changes that are unnecessary as we now verify that we're using SQLite 3.35.0 or higher.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/d202d47db086176420092b1c8f1bd0a6121cb5e0..2bb714d0e84831a217ae03bdd5113093bcad2995) to fix cargo fmt error.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -305,6 +320,9 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
 ) -> Result<(), MigratorError<WalletMigrationError>> {
     let seed = seed.map(Rc::new);
 
+    // Verify that the sqlite version in use supports the features required by this library
+    verify_sqlite_version_compatibility(&wdb.conn).map_err(MigratorError::Adapter)?;
```
daira, line 324: Since we've established that the issue was with the SQLite version used by the app to run queries, *not* the version used by `rusqlite`, is this even necessary any more? This query will always check only the version of the SQLite used by `rusqlite` (which is the bundled one, since we enable the `bundled` feature).

We know that the [required](https://github.com/zcash/librustzcash/blob/50cdf73dfdd13828995b06a7989c8c0fa10c27d4/Cargo.toml#L125) and pinned `rusqlite` (version 0.29.0) [bundles SQLite version 3.41.2](https://crates.io/crates/rusqlite/0.29.0#notes-on-building-rusqlite-and-libsqlite3-sys), so that is the version we are guaranteed to get currently. (Technically a `rusqlite` version 0.29.x could be released that bundles a different version and would be picked up by a `cargo update`, or we could increase the rusqlite version, but those would only be likely to bundle a newer SQLite.)

nuttycom: I added this essentially as a line of defense. I don't think that this query of constant data is an unreasonable guardrail, particularly given that it's only executed on wallet startup (and ideally would only be executed on wallet installation or upgrade).

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -357,6 +375,37 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
...
+            .captures(&sqlite_version)
+            .ok_or(WalletMigrationError::DatabaseNotSupported(
+                "Unknown".to_owned(),
+            ))?;
+    let parse_int = |value: &str| {
+        value.parse::<u32>().map_err(|_| {
+            WalletMigrationError::CorruptedData(format!(
+                "Cannot decode SQLite major version {}",
+                captures["major"].to_string()
+            ))
```
daira, line 396: Please fix the clippy lints if we keep this code.

daira (dismissed):
utACK, but I argue that we no longer need the version check. I don't object to keeping it; if we do, fix the clippy lints and please add a comment saying that what is being checked is the version used by `rusqlite`, which may be different than the version used to run view queries.

zcash_client_sqlite/src/wallet/init.rs
daira: Nod. In that case please log the version and add a comment about it being the version used by `rusqlite`. (Ideally I would also like to log the version used by the app, but that's an SDK issue.)

nuttycom:
@daira Comments addressed.

zcash_client_sqlite/src/wallet/init.rs
daira: We should also document what SQLite version we're relying on for the view queries, and then go through those checking that we're not relying on newer features. Is this PR expected to make it work on API level 27 (which provides at least SQLite 3.19.4 according to https://stackoverflow.com/a/4377116/393146 )?

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -357,6 +374,40 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
...
+/// Verify that the sqlite version in use supports the features required by this library.
+/// Note that the version of sqlite available to the database backend may be different
+/// from what is used to query the views that are part of the public API.
+fn verify_sqlite_version_compatibility(
+    conn: &rusqlite::Connection,
+) -> Result<(), WalletMigrationError> {
+    let sqlite_version =
+        conn.query_row("SELECT sqlite_version()", [], |row| row.get::<_, String>(0))?;
+
+    let version_re = Regex::new(r"^(?<major>[0-9]+)\.(?<minor>[0-9]+).+$").unwrap();
```
daira, line 386: This would not match a 2-component version number "x.y". We don't know that it couldn't be possible for a future version number to be of that form.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -357,6 +374,40 @@ fn init_wallet_db_internal<P: consensus::Parameters + 'static>(
...
+    let captures =
+        version_re
+            .captures(&sqlite_version)
+            .ok_or(WalletMigrationError::DatabaseNotSupported(
+                "Unknown".to_owned(),
+            ))?;
+    let parse_int = |value: &str| {
+        value.parse::<u32>().map_err(|_| {
+            WalletMigrationError::CorruptedData(format!(
+                "Cannot decode SQLite major version {}",
```
daira, line 396:
```suggestion
                "Cannot decode SQLite version component {}",
```

daira (approved):
utACK 5e6a8f02508893bce0410c9e3a941ea151d1d92f

daira (approved):
utACK f063e7ab7edfd279174795eaa8f106efcaca3f96

PR #1513, opened 2024-08-23
Transaction Table

ec2:
Unifies a bunch of stuff that was indexed by transaction id into a single struct


ec2:
Sorry... Github is really annoying about opening pull requests against repos that are forks of other repos...

PR #1514, opened 2024-08-25
zcash_protocol: Remove impl {Add, Sub} for BlockHeight

nuttycom:
These operations are unused, and it does not make sense to add block heights (or to subtract two block heights returning a block height).


daira (approved):
utACK

PR #1521, opened 2024-08-27
Add correct `BlockHeight` subtraction & remove panics.

nuttycom:
In contrast to the implementation of `Sub<BlockHeight> for BlockHeight` that was removed in version `0.3.0`, a saturating subtraction for block heights having a return type of `u32` makes sense for `BlockHeight`. Subtracting one block height from another yields the delta between them.

Other block height addition and subtraction operations have been made saturating, removing panics and the possibility of overflow.


daira (approved):
utACK

PR #1523, opened 2024-08-30
zcash_keys: Add `decode_extfvk_with_network`

nuttycom:


zcash_keys/src/encoding.rs
```diff
@@ -256,6 +265,39 @@ pub fn decode_extended_full_viewing_key(
     bech32_decode(hrp, s, |data| ExtendedFullViewingKey::read(&data[..]).ok())
 }
 
+/// Decodes an [`ExtendedFullViewingKey`] and the [`NetworkType`] that it is intended for use with
+/// from a Bech32-encoded string.
+#[cfg(feature = "sapling")]
+pub fn decode_extfvk_with_network(
```
daira, line 271: I know the other functions in this file don't explicitly say in their names that they're for Sapling, but that still seems odd.

zcash_keys/src/encoding.rs
```diff
@@ -256,6 +265,39 @@ pub fn decode_extended_full_viewing_key(
...
+    s: &str,
+) -> Result<(NetworkType, ExtendedFullViewingKey), Bech32DecodeError> {
+    use zcash_protocol::constants::{mainnet, regtest, testnet};
+
+    let (decoded_hrp, data, variant) = bech32::decode(s)?;
+    if variant != Variant::Bech32 {
+        Err(Bech32DecodeError::IncorrectVariant(variant))
+    } else {
```
daira, line 279: Consider using [`CheckedHrpstring::new::<Bech32>`](https://docs.rs/bech32/0.11.0/bech32/primitives/decode/struct.CheckedHrpstring.html) (although I'm not sure it really comes out any cleaner once error handling has been taken into account).

daira (approved):
utACK

PR #1524, opened 2024-08-30
zcash_client_sqlite: Do not require a transparent key for migrations.

nuttycom:
The `reserve_until` method for generating ephemeral addresses is used in database migrations, but it makes no sense to use this in the case that the UFVK for an account contains no transparent component; such accounts can never be used to make ZIP 320 spends.

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
+++ b/zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
@@ -113,7 +113,7 @@ pub(crate) fn get_ephemeral_ivk<P: consensus::Parameters>(
     conn: &rusqlite::Connection,
     params: &P,
     account_id: AccountId,
-) -> Result<EphemeralIvk, SqliteClientError> {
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -282,39 +282,40 @@ fn reserve_until<P: consensus::Parameters>(
...
-            .transpose()?;
-
-        stmt_insert_ephemeral_address.execute(named_params![
-            ":account_id": account_id.0,
-            ":address_index": raw_index,
-            ":address": address_str_opt,
-        ])?;
+        // used_in_tx and seen_in_tx are initially NULL
+        let mut stmt_insert_ephemeral_address = conn.prepare_cached(
+            "INSERT INTO ephemeral_addresses (account_id, address_index, address)
```
daira, line 294: Non-blocking typo below: "indicies"

daira (approved):
utACK

Issue #1412, opened 2024-06-13
Migrate test framework from `zcash_client_sqlite` to `zcash_client_backend`


str4d:
`zcash_client_sqlite::testing` contains a `TestBuilder` that can be used to produce a `TestState`, on which a test can operate. The test framework is generic over a block cache, that can optionally be provided with a `TestCache` implementation for tests that require block scanning. However, it currently hard-codes `zcash_client_sqlite::WalletDb` as the wallet that is tested.

In order to make the tests reusable with other wallet backends, we should rework `TestBuilder` etc to be generic over a `TestWallet` trait, that exposes the necessary operations that we currently depend on concrete `WalletDb` methods for. The new generic `TestBuilder` parameter should also be bound on `WalletRead`, `WalletWrite`, etc. as appropriate (i.e. only when the test framework requires a given bound).

Once that refactor is done, we can then move the test framework to `zcash_client_backend` behind the `test-dependencies` feature flag, and depend on it from `zcash_client_sqlite`. This will then free us up to subsequently genericise and migrate individual tests from `zcash_client_sqlite` to `zcash_client_backend` as we wish.

oxarbitrage:
FYI: I made some progress on this yesterday after discussing with @nuttycom that this is a good time. I will keep working on it and submit a PR soon (today or in the next few days). For easier review my current plan is to push 2 PRs, one with the all the refactoring needed inside `zcash_client_sqlite` and after we are happy with it, make another one with the actual move.

nuttycom:
Completed in #1530, #1539, #1543

PR #1526, opened 2024-09-02
Introduce a `Backend` type for tests

oxarbitrage:
This is a draft of what i have in mind for the refactor part of https://github.com/zcash/librustzcash/issues/1412

An new enum type `Backend` was added to `TestState` replacing the `db_data` with a `backend` field where the inner database type can be retrieved by methods. This should allow tests to use the framework with a `WalletDb` or a `MockWalletDb` backend.

Before making more changes, it will be good to get some feedback to see if i am in a good track or if i should rethink it.


nuttycom:
Closing in favor of #1530

PR #1529, opened 2024-09-04
zcash_client_backend: Fix broken --all-features build.

nuttycom:


zcash_client_backend/Cargo.toml
```diff
@@ -214,6 +214,14 @@ unstable-serialization = ["dep:byteorder"]
 ## Exposes the [`data_api::scanning::spanning_tree`] module.
 unstable-spanning-tree = []
 
+## Exposes access to the lightwalletd server via TOR
+tor-lightwalletd-tonic = [
```
str4d, line 218: This was placed at the end of the features section, which causes it to get categorised as an experimental feature in the crate docs.

zcash_client_backend/Cargo.toml
```diff
@@ -214,6 +214,14 @@ unstable-serialization = ["dep:byteorder"]
...
+  "tor",
+  "lightwalletd-tonic",
+  "tonic?/tls",
+  "tonic?/tls-webpki-roots"
```
str4d, line 222: I think this new feature is likely the wrong fix, and these two extra feature flags probably actually need to be part of the `tor` feature flag (because the currency conversion API uses HTTPS which requires TLS and the WebPKI roots as well).

str4d (commented):
Reviewed. I think this PR is the wrong fix, but because it is included in #1530 I'm just going to let this PR get merged when that one does, and then create a follow-up PR with my proposed fix.

zcash_client_backend/src/tor.rs
```diff
@@ -6,7 +6,7 @@ use arti_client::{config::TorClientConfigBuilder, TorClient};
 use tor_rtcompat::PreferredRuntime;
 use tracing::debug;
 
-#[cfg(feature = "lightwalletd-tonic")]
+#[cfg(feature = "tor-lightwalletd-tonic")]
```
str4d, line 9: This change turned out to be incomplete, and cause its own compilation errors. Fixed in #1532.

PR #1532, opened 2024-09-10
zcash_client_backend: Fix `tor::grpc` module feature flag

str4d:
We already depend on `rustls` and `webpki-roots` for `tor::http`, but `tonic` has its own feature flag that needs to be enabled for equivalent support in `tor::grpc`. We didn't need that feature flag enabled for the `proto::service::compact_tx_streamer_client` module because those constructors take a `D: TryInto<tonic::transport::Endpoint>`, which abstracts over TLS and leaves it up to the caller. By constrast, in `tor::grpc` we need to construct the `Endpoint` manually from a `Uri` and then configure TLS ourselves.


nuttycom (approved):
utACK

PR #1533, opened 2024-09-10
Migrate all `zcash_client_sqlite` pool tests to `zcash_client_backend` and make them generic over backend implementation

willemolding:
Moves the tests across, modifying where required and makes some additional changes to the `data_api` trait methods to support the test use-cases


zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1,36 +1,54 @@
...
     wallet::{Note, NoteId, OvkPolicy, ReceivedNote},
 };
 
-use super::{DataStoreFactory, TestCache, TestFvk, TestState};
+use super::{DataStoreFactory, Reset, TestCache, TestFvk, TestState};
+
+/// The number of ephemeral addresses that can be safely reserved without observing any
+/// of them to be mined. This is the same as the gap limit in Bitcoin.
+#[cfg(feature = "transparent-inputs")]
+pub(crate) const GAP_LIMIT: u32 = 20;
```
willemolding, line 51: Does this belong elsewhere maybe?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -238,3 +256,2321 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    let reservation_should_fail =
+        |st: &mut TestState<_, DSF::DataStore, _>, n, _expected_bad_index| {
+            assert_matches!(
+                st.wallet_mut()
+                    .reserve_next_n_ephemeral_addresses(account_id, n),
+                Err(..)
+            );
+            // previously it was hard-coded to a SqlClientError which could check the index
+            // not sure how best to handle this in a generic way
+            // Err(SqliteClientError::ReachedGapLimit(acct, bad_index))
```
willemolding, line 570: Any thoughts on how to handle checking for specific errors like this in the back-end agnostic case?

zcash_client_backend/src/data_api.rs
```diff
@@ -809,6 +809,15 @@ pub trait InputSource {
     ) -> Result<Vec<WalletTransparentOutput>, Self::Error> {
         Ok(vec![])
     }
+
+    /// Return the notes with a given value
+    #[cfg(any(test, feature = "test-dependencies"))]
+    fn get_notes(
```
str4d, line 829: In #1539 I'm moving these to a dedicated `WalletTest` trait so we can freely expose whatever internals are needed for testing purposes. (I'll handle any conflicts.)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3194,6 +3194,7 @@ pub mod testing {
 
         let results = stmt
             .query_and_then::<TransactionSummary<AccountId>, SqliteClientError, _, _>([], |row| {
+                let raw_tx: Vec<u8> = row.get("raw")?;
```
str4d: This will break (either loudly, or silently with an empty `Vec` which is a bad failure mode), as the `raw` column is nullable.
```suggestion
                let raw_tx: Option<Vec<u8>> = row.get("raw")?;
```

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -570,1247 +570,127 @@ pub(crate) fn change_note_spends_succeed<T: ShieldedPoolTester>() {
     )
 }
 
-pub(crate) fn external_address_change_spends_detected_in_restore_from_seed<
```
str4d, line 1152: Please fix this third commit (ad30ad6427034adf042ccf666a0ac6d705fea840):
- Commit message is `remove the raw field on get_tx_history as it broke tests` but the raw field is not removed.
- A bunch of test content is deleted, with no visible move.

In order to review this PR, I _need_ the code moves to be in the same commit (so that `git show COMMIT --color-moved=zebra --color-moved-ws=allow-indentation-change` works. Ideally the commit is as much a pure move as possible, with a fixup commit afterwards, but if the fixups are trivial then doing both at once is okayish.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -100,7 +100,6 @@ pub struct TransactionSummary<AccountId> {
     memo_count: usize,
     expired_unmined: bool,
     is_shielding: bool,
-    raw: Vec<u8>,
```
str4d: The `raw` removal is in the fourth commit (bb4dd304586f694395432a9e18df820dd0e2dc28) instead of the third, but the fix is probably to rewrite the fourth commit's message (and separately fixup or remove the third commit per my previous comment).

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1297,1247 +1304,1302 @@ pub fn change_note_spends_succeed<T: ShieldedPoolTester, DSF: DataStoreFactory>(
...
+    // Account balance should equal the change
+    assert_eq!(
+        st.get_total_balance(account.id()),
+        (value - value2).unwrap()
+    );
+
+    // Now scan the block in which we received the note that was spent.
+    st.scan_cached_blocks(received_height, 1);
+
+    // Account balance should be the same.
```
str4d, line 2571: This commit (bb4dd304586f694395432a9e18df820dd0e2dc28) appears to be uncommenting a large amount of test code? I cannot figure out whether any of the uncommented code is part of the `fix issue querying notes` that the commit message says is going on.

Given that there are thousands of lines changed as a result of doing so, please uncomment the code in one commit, and then in a second commit make changes to it.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -238,3 +240,2302 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
         Ok(_)
     );
 }
+
+// #[cfg(feature = "transparent-inputs")]
+// pub(crate) fn send_multi_step_proposed_transfer<T: ShieldedPoolTester>() {
```
str4d: Aha. Is what you did copy across all of the code, and comment it out at the destination in the same commit? Please do not do that, as the current approach makes it impossible to review the move.

Instead, either copy all the code as-is in one move (creating a commit that fails to compile) and then in subsequent commits apply the fixes, or move individual test functions over in one commit at a time.

str4d (changes requested):
The individual commits are structured as if they should be preserved, but they are currently unreviewable. Please fix the commits in the PR to be reviewable (which may amount to squashing all the commits together and then optionally splitting the resulting commit into functional parts).

zcash_client_sqlite/src/wallet.rs
str4d: And indeed this was reverted in a later commit as broken. I cannot tell from the current commits why this was initially considered necessary, and why it is no longer needed.

zcash_client_backend/src/data_api/testing/pool.rs
str4d: Yes; it can't live under `zcash_client_backend::data_api_testing` as that's only accessible with the `test-dependencies` feature flag.

nuttycom (commented):
Flushing a couple of comments from the review I started before @str4d's; I will re-review once the commit history is cleaned up to make it easier to see the moves.

str4d:
Replaced by #1543.

PR #1536, opened 2024-09-11
zcb::sync: Refresh UTXOs at the start of each scanning cycle

str4d:


zcash_client_backend/src/sync.rs
```diff
@@ -62,11 +72,27 @@ where
     <DbT as WalletRead>::Error: std::error::Error + Send + Sync + 'static,
     <DbT as WalletCommitmentTrees>::Error: std::error::Error + Send + Sync + 'static,
 {
+    #[cfg(feature = "transparent-inputs")]
+    let wallet_birthday = db_data
```
nuttycom, line 76: We should set Sapling activation (or maybe Sapling activation + 1) as the minimum height to use for block retrieval in scanning. IIRC there's currently a bug where lightwalletd won't return the frontiers prior to Sapling activation correctly.

nuttycom (approved):
utACK with a minor note about how to handle birthday heights (which may not be specifically applicable to this PR.)

zcash_client_backend/src/sync.rs
str4d: I think that already happens elsewhere, or if not then it should? Block retrieval is entirely dictated by the scan queue, so what matters is how the queue is filled; account creation (and the birthday heights provided therein) is what defines the earliest entries in the scan queue.

PR #1539, opened 2024-09-12
zcash_client_backend: Post-merge cleanups to test framework extraction

str4d:
See https://github.com/zcash/librustzcash/pull/1530#pullrequestreview-2289649712.


nuttycom (dismissed):
utACK, this seems like a good simplification. But should this wait to merge until after https://github.com/zcash/librustzcash/pull/1533?

str4d:
Yes it will; that's partly why this PR is a draft (the other part is that I have not finished it, but in any case my changes will be a much smaller diff).

daira (approved):
utACK

PR #1540, opened 2024-09-13
Change HashMaps in ScanningKeys to hold concrete ScanningKey 

ec2:
Removing the use of dynamic dispatch from the `ScanningKeys` struct. As per @nuttycom and @str4d, the original reason for this was to support both IVKs and FVKs for trial decryption. 

I'm of the opinion that allowing the mixing of the two makes things un-ergonomic. IVKs can be derived from FVKs anyways, so I don't really see the point in allowing mixing.

### DIFF
--- a/zcash_client_backend/src/scanning.rs
+++ b/zcash_client_backend/src/scanning.rs
@@ -187,11 +187,12 @@ impl<AccountId> ScanningKeyOps<OrchardDomain, AccountId, orchard::note::Nullifie
 
 /// A set of keys to be used in scanning for decryptable transaction outputs.
 pub struct ScanningKeys<AccountId, IvkTag> {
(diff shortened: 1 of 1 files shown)


nuttycom:
The problem that I see for this is that it then requires two calls to `scan_block_with_runners` for every block in a wallet that has both FVKs and IVKs. Instead of simply being able to query for the keys that are relevant to the wallet, and scan with them (taking advantage of batching) you have to do everything twice.

Something I'm not understanding with this change is: what use case is the `Box<dyn ScanningKeyOps>` interfering with? I can't see how this is a performance bottleneck; in general the number of keys will be small, and the cost of dynamic dispatch here should be negligible on a per-block basis. So what's the motivation for this change?

ec2:
I had a misunderstanding of requirements! Closing.

Issue #1070, opened 2023-12-14
zcash_client_backend: Use recover-until height in wallet summary


nuttycom:
This is important for distinguishing between restore mode and catch-up mode.

nuttycom:
Fixed in #1541

PR #1541, opened 2024-09-19
Improve progress representation in `WalletSummary`

str4d:
We now split progress into "scan progress" and "recover progress" based on the recover-until height (if set, otherwise recover progress is always 100%).

Part of #1070.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -471,6 +471,7 @@ pub struct WalletSummary<AccountId: Eq + Hash> {
     chain_tip_height: BlockHeight,
     fully_scanned_height: BlockHeight,
     scan_progress: Option<Ratio<u64>>,
+    recovery_progress: Option<Ratio<u64>>,
     next_sapling_subtree_index: u64,
     #[cfg(feature = "orchard")]
     next_orchard_subtree_index: u64,
@@ -483,6 +484,7 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
         chain_tip_height: BlockHeight,
         fully_scanned_height: BlockHeight,
         scan_progress: Option<Ratio<u64>>,
+        recovery_progress: Option<Ratio<u64>>,
         next_sapling_subtree_index: u64,
         #[cfg(feature = "orchard")] next_orchard_subtree_index: u64,
     ) -> Self {
@@ -491,6 +493,7 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
             chain_tip_height,
             fully_scanned_height,
             scan_progress,
+            recovery_progress,
             next_sapling_subtree_index,
             #[cfg(feature = "orchard")]
             next_orchard_subtree_index,
@@ -513,14 +516,45 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
         self.fully_scanned_height
     }
 
-    /// Returns the progress of scanning shielded outputs, in terms of the ratio between notes
-    /// scanned and the total number of notes added to the chain since the wallet birthday.
+    /// Returns the progress of scanning the chain to bring the wallet up to date.
     ///
-    /// This ratio should only be used to compute progress percentages, and the numerator and
-    /// denominator should not be treated as authoritative note counts. Returns `None` if the
-    /// wallet is unable to determine the size of the note commitment tree.
+    /// This progress metric is intended as an indicator of how close the wallet is to
+    /// general usability, including the ability to spend existing funds that were
+    /// previously spendable.
+    ///
+    /// The window over which progress is computed spans from the wallet's recovery height
+    /// to the current chain tip. This may be adjusted in future updates to better match
+    /// the intended semantics.
+    ///
+    /// Progress is represented in terms of the ratio between notes scanned and the total
+    /// number of notes added to the chain in the relevant window. This ratio should only
+    /// be used to compute progress percentages, and the numerator and denominator should
+    /// not be treated as authoritative note counts.
+    ///
+    /// Returns `None` if the wallet is unable to determine the size of the note
+    /// commitment tree.
     pub fn scan_progress(&self) -> Option<Ratio<u64>> {
         self.scan_progress
+    }
+
+    /// Returns the progress of recovering the wallet from seed.
+    ///
+    /// This progress metric is intended as an indicator of how close the wallet is to
+    /// having a complete history.
+    ///
+    /// The window over which progress is computed spans from the wallet birthday to the
+    /// wallet's recovery height. This may be adjusted in future updates to better match
+    /// the intended semantics.
+    ///
+    /// Progress is represented in terms of the ratio between notes scanned and the total
+    /// number of notes added to the chain in the relevant window. This ratio should only
+    /// be used to compute progress percentages, and the numerator and denominator should
+    /// not be treated as authoritative note counts.
+    ///
+    /// Returns `None` if the wallet is unable to determine the size of the note
+    /// commitment tree.
+    pub fn recovery_progress(&self) -> Option<Ratio<u64>> {
+        self.recovery_progress
     }
 
     /// Returns the Sapling subtree index that should start the next range of subtree
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -839,21 +839,23 @@ fn subtree_scan_progress(
     fully_scanned_height: BlockHeight,
     chain_tip_height: BlockHeight,
 ) -> Result<Option<Ratio<u64>>, SqliteClientError> {
+    let mut stmt_scanned_count_from = conn.prepare_cached(&format!(
+        "SELECT SUM({output_count_col})
+        FROM blocks
+        WHERE :start_height <= height",
+    ))?;
+
     if fully_scanned_height == chain_tip_height {
         // Compute the total blocks scanned since the wallet birthday
-        conn.query_row(
-            &format!(
-                "SELECT SUM({output_count_col})
-                FROM blocks
-                WHERE height >= :birthday_height",
-            ),
-            named_params![":birthday_height": u32::from(birthday_height)],
-            |row| {
-                let scanned = row.get::<_, Option<u64>>(0)?;
-                Ok(scanned.map(|n| Ratio::new(n, n)))
-            },
-        )
-        .map_err(SqliteClientError::from)
+        stmt_scanned_count_from
+            .query_row(
+                named_params![":start_height": u32::from(birthday_height)],
+                |row| {
+                    let scanned = row.get::<_, Option<u64>>(0)?;
+                    Ok(scanned.map(|n| Ratio::new(n, n)))
+                },
+            )
+            .map_err(SqliteClientError::from)
     } else {
         // Get the starting note commitment tree size from the wallet birthday, or failing that
         // from the blocks table.
@@ -887,13 +889,8 @@ fn subtree_scan_progress(
             .transpose()?;
 
         // Compute the total blocks scanned so far above the starting height
-        let scanned_count = conn.query_row(
-            &format!(
-                "SELECT SUM({output_count_col})
-                FROM blocks
-                WHERE height > :start_height",
-            ),
-            named_params![":start_height": u32::from(birthday_height)],
+        let scanned_count = stmt_scanned_count_from.query_row(
+            named_params![":start_height": u32::from(birthday_height + 1)],
             |row| row.get::<_, Option<u64>>(0),
         )?;
 
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -815,9 +815,10 @@ pub(crate) struct Progress {
 }
 
 pub(crate) trait ScanProgress {
-    fn sapling_scan_progress(
+    fn sapling_scan_progress<P: consensus::Parameters>(
         &self,
         conn: &rusqlite::Connection,
+        params: &P,
         birthday_height: BlockHeight,
         recover_until_height: Option<BlockHeight>,
         fully_scanned_height: BlockHeight,
@@ -825,9 +826,10 @@ pub(crate) trait ScanProgress {
     ) -> Result<Progress, SqliteClientError>;
 
     #[cfg(feature = "orchard")]
-    fn orchard_scan_progress(
+    fn orchard_scan_progress<P: consensus::Parameters>(
         &self,
         conn: &rusqlite::Connection,
+        params: &P,
         birthday_height: BlockHeight,
         recover_until_height: Option<BlockHeight>,
         fully_scanned_height: BlockHeight,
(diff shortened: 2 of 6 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -975,13 +980,183 @@ fn subtree_scan_progress(
...
+                    .optional()?;
+
+                if let Some((last_completed_subtree_index, last_completed_subtree_height)) =
+                    last_completed_subtree
+                {
+                    // If we know the tree size at the last scanned height, extrapolate.
+                    let tip_tree_size = last_scanned.zip(last_scanned_tree_size).and_then(
+                        |(last_scanned, last_scanned_tree_size)| {
+                            let scanned_notes = last_scanned_tree_size
+                                - (last_completed_subtree_index << shard_height);
```
nuttycom, line 1089: This using `last_completed_subtree_index` seems wrong to me, because we don't have any guarantee at this layer of scanning order; `last_scanned_tree_size` could be far smaller than the last completed subtree. Am I missing something that ties these two together? We need to extrapolate based upon the block position of the last scanned block _within its associated subtree_, but that's not what this is doing except by chance, AFAICT.

nuttycom (changes requested):
My review is not yet complete, but the last comment about the use of `last_completed_subtree_index` is blocking.

zcash_client_backend/src/data_api.rs
```diff
@@ -513,16 +516,47 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
...
-    /// This ratio should only be used to compute progress percentages, and the numerator and
-    /// denominator should not be treated as authoritative note counts. Returns `None` if the
-    /// wallet is unable to determine the size of the note commitment tree.
+    /// This progress metric is intended as an indicator of how close the wallet is to
+    /// general usability, including the ability to spend existing funds that were
+    /// previously spendable.
+    ///
+    /// The window over which progress is computed spans from the wallet's recovery height
+    /// to the current chain tip. This may be adjusted in future updates to better match
+    /// the intended semantics.
```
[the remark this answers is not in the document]
str4d: If we compute from max-fully-scanned height (presuming fully-recovered), then because we use linear scanning, the scan progress is _always_ 0% (as its start height keeps moving) until scan is complete, and then it jumps to 100%.

We need the boundary between "recovery" and "scanning" to be semi-persistent - it needs to be fixed across multiple calls to `get_wallet_summary` in the short term, while ideally moving towards the chain tip over time so that the scan progress remains in a useful percentage range.

For this PR, I'm just using the recover-until height as a proxy for this in order to get the initial split in place, and we can decide on an improved approach later.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -873,58 +918,113 @@ fn subtree_scan_progress(
...
+                        |row| row.get::<_, Option<u64>>(0),
+                    )
+                    .optional()
+                    .map(|opt| opt.flatten())
+            })
+            .transpose()?;
+
+        // Count the total outputs scanned so far on the birthday side of the
+        // recover-until height.
+        let recovered_count = recover_until_height
```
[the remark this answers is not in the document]
str4d: `recover_until_height` is optional (it can be set to `None` when constructing an `AccountBirthday`), so `recovered_count` is therefore necessarily also optional.

(Figuring out all the different bits of data in our database that may or may not be there is why this took me a solid day to implement.)

zcash_client_sqlite/src/wallet.rs
str4d: Indeed I think this is a bug caused by me spending an entire day trying to figure this out and finishing up at 5am.

However, I think the bug is different: if `last_scanned` is _not_ in the most recently-completed subtree, then I think we should extrapolate from the last completed subtree, and _not_ from `last_scanned` (which in that case would be older). We want to extrapolate from the most recently-available data, which is where I went astray here: I missed the condition that `last_scanned > last_completed_subtree_height` here.

nuttycom: It's not surprising that this took a long time to figure out; I have a branch where I started attempting to do this sort of interpolation and I abandoned it because it was taking too long. This PR overall is a better approach than I was taking, in that it's an incremental step in the right direction - I was trying to come up with a solution that didn't distinguish between wallet recovery and the steady state and it was a mess.

nuttycom: Extrapolating from the previous subtree seems fine; I really wish we just got the tree sizes when we got the chain tip information, that would make it trivial.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -873,58 +918,113 @@ fn subtree_scan_progress(
...
-                        |(min_tree_size, max_tree_size)| {
-                            Ratio::new(scanned_count.unwrap_or(0), max_tree_size - min_tree_size)
-                        },
-                    ))
+                    Ok(min_tree_size.zip(max_tree_size))
                 },
             )
-            .optional()?
-            .flatten())
+            .optional()?.flatten().unzip();
```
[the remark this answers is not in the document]
str4d: This was due to the long length of the `named_params!` line: if `rustfmt` can't linewrap down below the max line length, it doesn't format the code block. Manually wrapping that block enables it to fix this.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/d872202344efc7cd830ff7fcbb10c2302e161df9..8cc7566126c40d6e905b61ed889e3aa0e24d19c9?w=1) to address review comments.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,442 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+                    last_completed_subtree
+                {
+                    // If we know the tree size at the last scanned height, and that
+                    // height is within the incomplete subtree, extrapolate.
+                    let tip_tree_size =
+                        last_scanned.and_then(|(last_scanned, last_scanned_tree_size)| {
+                            (last_scanned > last_completed_subtree_height)
+                                .then(|| {
+                                    let scanned_notes = last_scanned_tree_size
+                                        - (last_completed_subtree_index << shard_height);
```
nuttycom, line 1077: Shouldn't this be `(last_completed_subtree_index + 1) << shard_height`? Without the `+1` this computes the start of a subtree, not the end as the indices are zero-based.

For a moment I thought you were averaging over all the data to the start of the previous subtree, but that's not the case as you're using the subtree end height.

nuttycom: It would be clearer in this code to name `last_completed_subtree_height` to `last_completed_subtree_end` or `incomplete_subtree_start_height`.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/8cc7566126c40d6e905b61ed889e3aa0e24d19c9..d29a9fdb159853308c75857519c9236305e2ecb6) to address more comments and fix a compilation bug.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,440 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+
+                        (subtree_notes * unscanned_range)
+                            .checked_div(subtree_range)
+                            .map(|extrapolated_incomplete_subtree_notes| {
+                                notes_in_complete_subtrees + extrapolated_incomplete_subtree_notes
+                            })
+                    } else {
+                        // There's only one completed subtree; its start height must
+                        // be the activation height for this shielded protocol.
+                        let scanned_notes = last_completed_subtree_index << shard_height;
```
nuttycom, line 1131: I think this also has the off-by-one error; if the completed subtree has index 0 (which is what you describe in the comment) then the value of `scanned_notes` will be zero.

nuttycom (changes requested):
A couple of off-by-one errors.

zcash_client_sqlite/src/wallet.rs
nuttycom: Another possibility is to use the `Address` for the root of the subtree. The whole reason that type exists is to make errors like this more obvious.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1332,15 +1345,20 @@ pub(crate) mod tests {
...
+            summary.as_ref().and_then(|s| s.recover_progress()),
+            fully_recovered,
+        );
+
         // Progress denominator depends on which pools are enabled (which changes the
-        // initial tree states). Here we compute the denominator based upon the fact that
-        // the trees are the same size at present.
-        let expected_denom = (1 << SAPLING_SHARD_HEIGHT) * 2 - frontier_tree_size;
+        // initial tree states), and is extrapolated from the scanned range.
+        let expected_denom: u64 = 2628;
```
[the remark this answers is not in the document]
str4d: After fixing the off-by-ones, I'm able to compute this now (and know why I wasn't able to compute it before)!

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1412,11 +1430,14 @@ pub(crate) mod tests {
         // We've crossed a subtree boundary, but only in one pool. We still only have one scanned
         // note but in the pool where we crossed the subtree boundary we have two shards worth of
         // notes to scan.
+        let expected_denom: u64 = 2823;
```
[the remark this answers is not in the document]
str4d: This one I was failing to calculate because the existing test was producing a broken shard tree state:
```
SELECT shard_index, subtree_end_height FROM sapling_tree_shards
[
    (0, Some(100066)),
    (1, None),
    (2, Some(100646)),
]
```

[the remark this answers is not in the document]
nuttycom: That's interesting, because in the absence of downloading subtree roots this *is* a valid state for the tree.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/d29a9fdb159853308c75857519c9236305e2ecb6..b4da98e510cfc035cd8d97bf48a7debe828004dd) to address remaining comments.

zcash_client_backend/src/data_api.rs
```diff
@@ -513,16 +516,47 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
...
+    ///
+    /// This progress metric is intended as an indicator of how close the wallet is to
+    /// having a complete history.
+    ///
+    /// The window over which progress is computed spans from the wallet birthday to the
+    /// wallet's recovery height. This may be adjusted in future updates to better match
+    /// the intended semantics.
+    ///
+    /// Progress is represented in terms of the ratio between notes scanned and the total
+    /// number of notes added to the chain in the relevant window. This ratio should only
```
daira, line 550: This sentence is the same between the two methods. Should it be?

str4d: Yes; "relevant window" is the parameter.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,449 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+    ))?;
+    let mut stmt_scanned_count_from = conn.prepare_cached(&format!(
+        "SELECT SUM({output_count_col})
+        FROM blocks
+        WHERE :start_height <= height",
+    ))?;
+    let mut stmt_start_tree_size = conn.prepare_cached(&format!(
+        "SELECT MAX({table_prefix}_commitment_tree_size - {output_count_col})
+        FROM blocks
+        WHERE height <= :start_height",
```
daira, line 869: Is it intended that `height <= :start_height` overlaps `:start_height <= height` (in the queries above) by one block?

str4d: Yes, because subtree boundaries don't fall at block boundaries.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,449 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+    output_count_col: &'static str,
+    shard_height: u8,
+    pool_activation_height: BlockHeight,
+    birthday_height: BlockHeight,
+    recover_until_height: Option<BlockHeight>,
+    fully_scanned_height: BlockHeight,
+    chain_tip_height: BlockHeight,
+) -> Result<Progress, SqliteClientError> {
+    let mut stmt_scanned_count_between = conn.prepare_cached(&format!(
+        "SELECT SUM({output_count_col})
```
daira, line 857: Why a separate `output_count_col` variable rather than using `{table_prefix}_output_count`?

str4d: In Orchard the column name says `action` instead of `output`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,449 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+                            (last_scanned > last_completed_subtree_end)
+                                .then(|| {
+                                    let scanned_notes = last_scanned_tree_size
+                                        - u64::from(last_completed_subtree.position_range_end());
+                                    let scanned_range =
+                                        u64::from(last_scanned - last_completed_subtree_end);
+                                    let unscanned_range =
+                                        u64::from(chain_tip_height - last_scanned);
+
+                                    (scanned_notes * unscanned_range)
```
daira, line 1091: saturating_mul? I think it can't overflow anyway, but that isn't particularly obvious.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,449 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+                            u64::from(last_completed_subtree.position_range_end());
+
+                        let subtree_notes = 1 << shard_height;
+                        let subtree_range = u64::from(
+                            last_completed_subtree_end - second_to_last_completed_subtree_end,
+                        );
+                        let unscanned_range =
+                            u64::from(chain_tip_height - last_completed_subtree_end);
+
+                        (subtree_notes * unscanned_range)
```
daira, line 1132: Ditto.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -808,213 +808,449 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
...
+                        // There's only one completed subtree; its start height must
+                        // be the activation height for this shielded protocol.
+                        let subtree_notes = 1 << shard_height;
+
+                        let subtree_range =
+                            u64::from(last_completed_subtree_end - pool_activation_height);
+                        let unscanned_range =
+                            u64::from(chain_tip_height - last_completed_subtree_end);
+
+                        (subtree_notes * unscanned_range)
```
daira, line 1147: Ditto.

daira (approved):
utACK with comments.

nuttycom (approved):
utACK b4da98e510cfc035cd8d97bf48a7debe828004dd

PR #1542, opened 2024-09-20
zcash_client_sqlite: Fix bad `WalletTest` import that breaks downstream builds.

nuttycom:


daira (approved):
utACK

str4d (changes requested):
NACK, this will conflict with #1543, where I (believe I) also fixed this.

nuttycom:
@str4d since the changed lines are identical, it shouldn't conflict. But I'm reviewing #1543 now.

str4d (approved):
ACK 176c0cb805c44ed85216ea6ce8f05b93bb3f3d42. I _think_ this should not conflict with #1543 as they do happen to be identical on the relevant lines, and this PR was an undeclared dependency of https://github.com/Electric-Coin-Company/zec-sqlite-cli/pull/46 so we need this commit to end up in `main`.

daira (commented):
Post-hoc ACK c0f4effdb6ea0c597a896c9817f44362899f7c9d

PR #1543, opened 2024-09-20
Move remaining pool tests from `zcash_client_sqlite` to `zcash_client_backend`

str4d:


zcash_client_sqlite/src/lib.rs
```diff
@@ -99,7 +99,7 @@ use maybe_rayon::{
 };
 
 #[cfg(any(test, feature = "test-dependencies"))]
-use zcash_client_backend::data_api::testing::TransactionSummary;
+use zcash_client_backend::data_api::{testing::TransactionSummary, WalletTest};
```
str4d, line 102: This is the same fix as in #1542.

zcash_client_backend/src/data_api.rs
```diff
@@ -1226,9 +1229,45 @@ pub trait WalletTest: WalletRead {
...
-    ) -> Result<Vec<NoteId>, Self::Error> {
+    ) -> Result<Vec<NoteId>, <Self as WalletRead>::Error> {
         Ok(vec![])
     }
+
+    #[allow(clippy::type_complexity)]
+    fn get_confirmed_sends(
+        &self,
+        txid: &TxId,
+    ) -> Result<Vec<(u64, Option<String>, Option<String>, Option<u32>)>, <Self as WalletRead>::Error>;
```
nuttycom, line 1240: Could the inner result type here be a struct? I have no idea what this tuple means.

str4d:
Third commit is best reviewed with:
```
git show 47cd7b8089921bb6d6f0648cfc8b9d8a6a1643dc --color-moved=zebra --color-moved-ws=allow-indentation-change
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -638,12 +641,102 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletTest for W
...
+                 ORDER BY value",
+            )?;
+
+        let sends = stmt_sent
+            .query_map(rusqlite::params![txid.as_ref()], |row| {
+                let v: u32 = row.get(0)?;
+                let to_address: Option<String> = row.get(1)?;
+                let ephemeral_address: Option<String> = row.get(2)?;
+                let address_index: Option<u32> = row.get(3)?;
+                Ok((u64::from(v), to_address, ephemeral_address, address_index))
```
nuttycom, line 679: Yeah, this would be much more usable if it were a struct.

zcash_client_sqlite/src/lib.rs
```diff
@@ -638,12 +641,102 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletTest for W
...
+    fn get_notes(
+        &self,
+        protocol: ShieldedProtocol,
+    ) -> Result<Vec<ReceivedNote<Self::NoteRef, Note>>, <Self as InputSource>::Error> {
+        let (table_prefix, index_col, _) = wallet::common::per_protocol_names(protocol);
+        let mut stmt_received_notes = self.conn.borrow().prepare(&format!(
+            "SELECT txid, {index_col}
+             FROM {table_prefix}_received_notes rn
+             INNER JOIN transactions ON transactions.id_tx = rn.tx
+             AND transactions.block IS NOT NULL
```
nuttycom, line 721: nit:
```suggestion
             WHERE transactions.block IS NOT NULL
```

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -1532,12 +1526,6 @@ pub(crate) fn checkpoint_gaps<T: ShieldedPoolTester>() {
     // Scan the block
     st.scan_cached_blocks(account.birthday().height() + 10, 1);
 
-    // Fake that everything has been scanned
```
nuttycom, line 1566: I'm a little bit concerned that this test still passes without this.

nuttycom (approved):
utACK a463f7dbc790a09d41675bcb2238f723e6fd0d93 with minor suggestion for improvement that can be addressed post-merge if necessary.

zcash_client_sqlite/src/testing/pool.rs
str4d: I verified that without the `scan_cached_blocks` it fails, so it needs that specific block to be scanned, but apparently not the rest of the chain.

zcash_client_backend/src/data_api.rs
str4d: Yeah, that would be nicer. I pulled this API as-is from #1533. We also need to add documentation etc. to the new APIs in this PR, which includes All Of The Tests (but I'd spent long enough on this PR).

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/a463f7dbc790a09d41675bcb2238f723e6fd0d93..e67e7abc522169fb1a70a33cff7a27337de5731a) to address the SQL review comment. I'll make the struct changes in a separate PR next week.

nuttycom (approved):
re-utACK https://github.com/zcash/librustzcash/pull/1543/commits/e67e7abc522169fb1a70a33cff7a27337de5731a

nuttycom:
Please add an issue to track the documentation & struct changes.

zcash_client_backend/src/data_api.rs
str4d: Opened #1544.

str4d:
Opened #1544 and #1545.

daira (commented):
Post-hoc ACK

PR #1554, opened 2024-10-02
Migrate to `sapling-crypto 0.3`, `orchard 0.10`

str4d:


Cargo.toml
```diff
@@ -163,8 +163,6 @@ codegen-units = 1
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(zcash_unstable, values("zfuture"))'] }
 
 [patch.crates-io]
-sapling-crypto = { git = "https://github.com/zcash/sapling-crypto", rev = "b1ad3694ee13a2fc5d291ad04721a6252da0993c" }
-orchard = { git = "https://github.com/zcash/orchard", rev = "55fb089a335bbbc1cda186c706bc037073df8eb7" }
 incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree", rev = "ffe4234788fd22662b937ba7c6ea01535fcc1293" }
 incrementalmerkletree-testing = { git = "https://github.com/zcash/incrementalmerkletree", rev = "ffe4234788fd22662b937ba7c6ea01535fcc1293" }
```
nuttycom, line 167: Can you drop the `incrementalmerkletree` patch dependency as well? You'll still need `incrementalmerkletree-testing`.

nuttycom (approved):
utACK with question

Cargo.toml
str4d: No, as I said in our pairing earlier this breaks while `shardtree` is a patch dependency.

str4d: Because you then have two versions of `incrementalmerkletree` in the dependency tree; you still get the git version via `shardtree`, but now it isn't unified with `sapling-crypto` and `orchard`, and you get trait incoherence errors.

str4d: https://github.com/zcash/incrementalmerkletree/pull/116 fixes this issue.

daira (dismissed):
utACK

nuttycom (approved):

nuttycom (approved):
re-utACK

zcash_primitives/CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ and this library adheres to Rust's notion of
 ## [0.17.0] - 2024-08-26
 
 ### Changed
-- Update dependencies to `zcash_protocol 0.3.0`, `zcash_address 0.5.0`
+- Update dependencies to `incrementalmerkletree 0.7`, `orchard 0.10`,
+  `sapling-crypto 0.3`, `zcash_protocol 0.3.0`, `zcash_address 0.5.0`.
```
str4d, line 14: I edited the wrong part of the changelog here and it wasn't caught in review. I'll fix this in the release train PR.

PR #1557, opened 2024-10-03
zcash_client_sqlite: Add unstable conversions between `AccountId` and `u32`.

nuttycom:
This is necessary for third parties using the `zcash_client_sqlite` crate directly to be able to refer to individual account IDs without having to always use FVKs to look them up.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -164,6 +164,26 @@ pub(crate) const DEFAULT_UA_REQUEST: UnifiedAddressRequest =
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default)]
 pub struct AccountId(u32);
 
+impl AccountId {
+    /// Constructs an `AccountId` from a bare `u32` value. The resulting identifier is not
+    /// guaranteed to correspond to any account stored in the database.
+    #[cfg(feature = "unstable")]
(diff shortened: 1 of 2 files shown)


zcash_client_sqlite/src/lib.rs
```diff
@@ -164,6 +164,26 @@ pub(crate) const DEFAULT_UA_REQUEST: UnifiedAddressRequest =
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default)]
 pub struct AccountId(u32);
 
+impl AccountId {
+    /// Constructs an `AccountId` from a bare `u32` value. The resulting identifier is not
+    /// guaranteed to correspond to any account stored in the database.
+    #[cfg(feature = "unstable")]
+    pub fn from_u32(value: u32) -> Self {
```
daira, line 171: Should this be `unsafe`?
```suggestion
    pub unsafe fn from_u32(value: u32) -> Self {
```

daira (approved):
utACK

zcash_client_sqlite/src/lib.rs
nuttycom: This isn't unsafe, by any meaning of the term?

daira: It's up to the caller to maintain the intended invariant that `AccountId`s are only constructed for accounts stored in the database, which isn't checked.

PR #1564, opened 2024-10-10
zcash_client_sqlite: Fix errors in progress computation.

nuttycom:
This ensures that we use `recover_until_height` as an exclusive end height, and makes sure that we are estimating tree size based upon consistent heights between the recovery and scan ranges.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -534,10 +534,11 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
     /// Progress is represented in terms of the ratio between notes scanned and the total
     /// number of notes added to the chain in the relevant window. This ratio should only
     /// be used to compute progress percentages, and the numerator and denominator should
-    /// not be treated as authoritative note counts.
+    /// not be treated as authoritative note counts. The denominator of this ratio is
+    /// guaranteed to be nonzero.
     ///
-    /// Returns `None` if the wallet is unable to determine the size of the note
-    /// commitment tree.
+    /// Returns `None` if the wallet has insufficient information to be able to determine
+    /// scan progress.
     pub fn scan_progress(&self) -> Option<Ratio<u64>> {
         self.scan_progress
     }
@@ -554,10 +555,12 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
     /// Progress is represented in terms of the ratio between notes scanned and the total
     /// number of notes added to the chain in the relevant window. This ratio should only
     /// be used to compute progress percentages, and the numerator and denominator should
-    /// not be treated as authoritative note counts.
+    /// not be treated as authoritative note counts. Note that both the numerator and the
+    /// denominator of this ratio may be zero in the case that there is no recovery range
+    /// that need be scanned.
     ///
-    /// Returns `None` if the wallet is unable to determine the size of the note
-    /// commitment tree.
+    /// Returns `None` if the wallet has insufficient information to be able to determine
+    /// progress in scanning between the wallet birthday and the wallet recovery height.
     pub fn recovery_progress(&self) -> Option<Ratio<u64>> {
         self.recovery_progress
     }
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1270,6 +1270,7 @@ pub trait WalletTest: InputSource + WalletRead {
 ///
 /// This type is opaque, and exists for use by tests defined in this crate.
 #[cfg(any(test, feature = "test-dependencies"))]
+#[allow(dead_code)]
 #[derive(Clone, Debug)]
 pub struct OutputOfSentTx {
     value: NonNegativeAmount,
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -898,10 +898,11 @@ fn subtree_scan_progress<P: consensus::Parameters>(
                 )
             })
             .transpose()?
-            // If none of the wallet's accounts have a recover-until height, then we can't
-            // (yet) distinguish general scanning from recovery, so treat the wallet as
-            // fully recovered.
-            .unwrap_or_else(|| Some(Ratio::new(1, 1)));
+            // If none of the wallet's accounts have a recover-until height, then there
+            // is no recovery phase for the wallet, and therefore the denominator in the
+            // resulting ratio (the number of notes in the recovery range) is zero.
+            .unwrap_or_else(|| Some(Ratio::new(0, 0)));
+
         let scan = stmt_scanned_count_from.query_row(
             named_params! {
                 ":start_height": u32::from(
@@ -915,9 +916,46 @@ fn subtree_scan_progress<P: consensus::Parameters>(
         )?;
         Ok(Progress { scan, recover })
     } else {
+        // In case we didn't have information about the tree size at the recover-until
+        // height, get the tree size from a nearby subtree. It's fine for this to be
+        // approximate; it just shifts the boundary between scan and recover progress.
+        let mut get_tree_size_near = |as_of: BlockHeight| {
+            stmt_start_tree_size
+                .query_row(
+                    named_params![":start_height": u32::from(as_of)],
+                    |row| row.get::<_, Option<u64>>(0),
+                )
+                .optional()
+                .map(|opt| opt.flatten())
+                .transpose()
+                .or_else(|| {
+                    conn.query_row(
+                        &format!(
+                            "SELECT MIN(shard_index)
+                    FROM {table_prefix}_tree_shards
+                    WHERE subtree_end_height >= :start_height
+                    OR subtree_end_height IS NULL",
+                        ),
+                        named_params! {
+                            ":start_height": u32::from(as_of),
+                        },
+                        |row| {
+                            let min_tree_size = row
+                                .get::<_, Option<u64>>(0)?
+                                .map(|min_idx| min_idx << shard_height);
+                            Ok(min_tree_size)
+                        },
+                    )
+                    .optional()
+                    .map(|opt| opt.flatten())
+                    .transpose()
+                })
+                .transpose()
+        };
+
         // Get the starting note commitment tree size from the wallet birthday, or failing that
         // from the blocks table.
-        let start_size = conn
+        let birthday_size = conn
             .query_row(
                 &format!(
                     "SELECT birthday_{table_prefix}_tree_size
@@ -930,29 +968,14 @@ fn subtree_scan_progress<P: consensus::Parameters>(
             .optional()?
             .flatten()
             .map(Ok)
-            .or_else(|| {
-                stmt_start_tree_size
-                    .query_row(
-                        named_params![":start_height": u32::from(birthday_height)],
-                        |row| row.get::<_, Option<u64>>(0),
-                    )
-                    .optional()
-                    .map(|opt| opt.flatten())
-                    .transpose()
-            })
+            // If we don't have an explicit birthday tree size, find something nearby.
+            .or_else(|| get_tree_size_near(birthday_height).transpose())
             .transpose()?;
 
         // Get the note commitment tree size as of the start of the recover-until height.
         let recover_until_size = recover_until_height
-            .map(|end_height| {
-                stmt_start_tree_size
-                    .query_row(
-                        named_params![":start_height": u32::from(end_height)],
-                        |row| row.get::<_, Option<u64>>(0),
-                    )
-                    .optional()
-                    .map(|opt| opt.flatten())
-            })
+            // Find a tree size near to the recover-until height
+            .map(get_tree_size_near)
             .transpose()?;
 
         // Count the total outputs scanned so far on the birthday side of the recover-until height.
@@ -967,31 +990,6 @@ fn subtree_scan_progress<P: consensus::Parameters>(
                 )
             })
             .transpose()?;
-
-        // In case we didn't have information about the tree size at the recover-until
-        // height, get the tree size from a nearby subtree. It's fine for this to be
-        // approximate; it just shifts the boundary between scan and recover progress.
-        let min_tree_size = conn
-            .query_row(
-                &format!(
-                    "SELECT MIN(shard_index)
-                    FROM {table_prefix}_tree_shards
-                    WHERE subtree_end_height >= :start_height
-                    OR subtree_end_height IS NULL",
-                ),
-                named_params! {
-                    ":start_height": u32::from(recover_until_height.unwrap_or(birthday_height)),
-                },
-                |row| {
-                    let min_tree_size = row
-                        .get::<_, Option<u64>>(0)?
-                        .map(|min_idx| min_idx << shard_height);
-                    Ok(min_tree_size)
-                },
-            )
-            .optional()?
-            .flatten();
-
         // If we've scanned the block at the chain tip, we know how many notes are
         // currently in the tree.
         let tip_tree_size = match stmt_end_tree_size_at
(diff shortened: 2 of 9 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -836,23 +836,199 @@ pub(crate) trait ScanProgress {
...
+                        row.get(0)?,
+                    ),
+                    BlockHeight::from_u32(row.get(1)?),
+                ))
+            },
+        )
+        // `None` if we have no subtree roots yet.
+        .optional()?;
+
+    if let Some((last_completed_subtree, last_completed_subtree_end)) = last_completed_subtree {
```
daira, line 944:
```suggestion
    if let Some((last_completed_shard, last_completed_subtree_end)) = last_completed_subtree {
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -836,23 +836,199 @@ pub(crate) trait ScanProgress {
...
+        // If we know the tree size at the last scanned height, and that
+        // height is within the incomplete subtree, extrapolate.
+        let tip_tree_size = last_scanned.and_then(|(last_scanned, last_scanned_tree_size)| {
+            (last_scanned > last_completed_subtree_end)
+                .then(|| {
+                    let scanned_notes = last_scanned_tree_size
+                        - u64::from(last_completed_subtree.position_range_end());
```
daira, line 951:
```suggestion
                        - u64::from(last_completed_shard.position_range_end());
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -836,23 +836,199 @@ pub(crate) trait ScanProgress {
...
+                        "SELECT subtree_end_height
+                         FROM {table_prefix}_tree_shards
+                         WHERE shard_index = :shard_index"
+                    ),
+                    named_params! {":shard_index": subtree_index},
+                    |row| Ok(row.get::<_, Option<_>>(0)?.map(BlockHeight::from_u32)),
+                )
+                .transpose()
+            })
+            .transpose()?
```
daira, line 981: Why do we have to use a query here rather than computing the end height directly from the shard index? (We know this shard is full because it precedes another one.)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -836,23 +836,199 @@ pub(crate) trait ScanProgress {
...
+    // block.
+
+    // Get the tree size at the last scanned height, if known.
+    let last_scanned = block_max_scanned(conn, params)?.and_then(|last_scanned| {
+        match shielded_protocol {
+            ShieldedProtocol::Sapling => last_scanned.sapling_tree_size(),
+            #[cfg(feature = "orchard")]
+            ShieldedProtocol::Orchard => last_scanned.orchard_tree_size(),
+            #[cfg(not(feature = "orchard"))]
+            ShieldedProtocol::Orchard => None
```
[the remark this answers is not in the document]
nuttycom: That relies upon non-local knowledge; I think that `None` here is a perfectly reasonable result.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -887,15 +1063,15 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
             })
             .transpose()?
-            // If none of the wallet's accounts have a recover-until height, then we can't
-            // (yet) distinguish general scanning from recovery, so treat the wallet as
-            // fully recovered.
-            .unwrap_or_else(|| Some(Ratio::new(1, 1)));
+            // If none of the wallet's accounts have a recover-until height, then there
+            // is no recovery phase for the wallet, and therefore the denominator in the
+            // resulting ratio (the number of notes in the recovery range) is zero.
+            .unwrap_or_else(|| Some(Ratio::new(0, 0)));
```
daira, line 1067: This is fine for this API, but is an incompatible change for the SDK-level API. It's fine if we intend to explicitly handle 0:0 at the wallet level now; otherwise we should map that to 1:1 to emulate the old API.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -915,9 +916,46 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
+                                .map(|min_idx| min_idx << shard_height);
+                            Ok(min_tree_size)
+                        },
+                    )
+                    .optional()
+                    .map(|opt| opt.flatten())
+                    .transpose()
+                })
+                .transpose()
+        };
```
daira: I personally prefer it this way:
```suggestion
        let mut get_tree_size_near = |as_of: BlockHeight| -> Result<_, SqliteClientError> {
            if let Some(tree_size) = stmt_start_tree_size
                .query_row(
                    named_params![":start_height": u32::from(birthday_height)],
                    |row| row.get::<_, Option<u64>>(0),
                )
                .optional()?
                .flatten()
            {
                return Ok(Some(tree_size));
            };
            Ok(conn
                .query_row(
                    &format!(
                        "SELECT MIN(shard_index)
                        FROM {table_prefix}_tree_shards
                        WHERE subtree_end_height >= :start_height
                        OR subtree_end_height IS NULL",
                    ),
                    named_params! {
                        ":start_height": u32::from(as_of),
                    },
                    |row| {
                        let min_tree_size = row
                            .get::<_, Option<u64>>(0)?
                            .map(|min_idx| min_idx << shard_height);
                        Ok(min_tree_size)
                    },
                )
                .optional()?
                .flatten())
        };
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -828,31 +842,219 @@ pub(crate) trait ScanProgress {
...
+    // block.
+
+    // Get the tree size at the last scanned height, if known.
+    let last_scanned = block_max_scanned(conn, params)?.and_then(|last_scanned| {
+        match shielded_protocol {
+            ShieldedProtocol::Sapling => last_scanned.sapling_tree_size(),
+            #[cfg(feature = "orchard")]
+            ShieldedProtocol::Orchard => last_scanned.orchard_tree_size(),
+            #[cfg(not(feature = "orchard"))]
+            ShieldedProtocol::Orchard => None,
```
daira, line 927:
```suggestion
            ShieldedProtocol::Orchard => unreachable!(),
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -828,31 +842,219 @@ pub(crate) trait ScanProgress {
...
+                    (scanned_notes * unscanned_range)
+                        .checked_div(scanned_range)
+                        .map(|extrapolated_unscanned_notes| {
+                            last_scanned_tree_size + extrapolated_unscanned_notes
+                        })
+                })
+                .flatten()
+        });
+
+        if let Some(tree_size) = tip_tree_size {
```
daira, line 976: Since every branch returns `Ok(...)`, the `Ok` can be pulled out, i.e. `Ok(if let ...)`.

zcash_client_sqlite/src/wallet.rs
str4d: I made the same comment to @nuttycom in mob today, that for this kind of highly-nested situation I find imperative code with early returns (`?`) clearer than functional code with type transposing (despite my general tendency to implement the latter, because in many cases it is great).

zcash_client_sqlite/src/wallet.rs
str4d: We _can_ cut an 0.13 as a hotfix if a breaking change is required. But I'm also inclined to _not_ make this breaking change, unless there's a bug-related reason for it.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -920,36 +1160,20 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
-                        named_params![":start_height": u32::from(birthday_height)],
-                        |row| row.get::<_, Option<u64>>(0),
-                    )
-                    .optional()
-                    .map(|opt| opt.flatten())
-                    .transpose()
-            })
+            // If we don't have an explicit birthday tree size, find something nearby.
+            .or_else(|| get_tree_size_near(birthday_height).transpose())
             .transpose()?;
```
daira, line 1165: Consider:
```rust
        let birthday_size = if let Some(tree_size) = conn
            .query_row(
                &format!(
                    "SELECT birthday_{table_prefix}_tree_size
                    FROM accounts
                    WHERE birthday_height = :birthday_height",
                ),
                named_params![":birthday_height": u32::from(birthday_height)],
                |row| row.get::<_, Option<u64>>(0),
            )
            .optional()?
            .flatten()
        {
            Some(tree_size)
        } else {
            // If we don't have an explicit birthday tree size, find something nearby.
            get_tree_size_near(birthday_height)?
        };
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -887,27 +1089,65 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
+                                .map(|min_idx| min_idx << shard_height);
+                            Ok(min_tree_size)
+                        },
+                    )
+                    .optional()
+                    .map(|opt| opt.flatten())
+                    .transpose()
+                })
+                .transpose()
+        };
```
daira, line 1146: I personally prefer it this way:
```suggestion
        let mut get_tree_size_near = |as_of: BlockHeight| -> Result<_, SqliteClientError> {
            Ok(
                if let Some(tree_size) = stmt_start_tree_size
                    .query_row(
                        named_params![":start_height": u32::from(birthday_height)],
                        |row| row.get::<_, Option<u64>>(0),
                    )
                    .optional()?
                    .flatten()
                {
                    Some(tree_size)
                } else {
                    conn.query_row(
                        &format!(
                            "SELECT MIN(shard_index)
                            FROM {table_prefix}_tree_shards
                            WHERE subtree_end_height >= :start_height
                            OR subtree_end_height IS NULL",
                        ),
                        named_params! {
                            ":start_height": u32::from(as_of),
                        },
                        |row| {
                            let min_tree_size = row
                                .get::<_, Option<u64>>(0)?
                                .map(|min_idx| min_idx << shard_height);
                            Ok(min_tree_size)
                        },
                    )
                    .optional()?
                    .flatten()
                },
            )
        };
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -920,36 +1160,20 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
-                    .map(|opt| opt.flatten())
-            })
+            // Find a tree size near to the recover-until height
+            .map(get_tree_size_near)
             .transpose()?;
 
-        // Count the total outputs scanned so far on the birthday side of the
-        // recover-until height.
+        // Count the total outputs scanned so far on the birthday side of the recover-until height.
         let recovered_count = recover_until_height
```
daira, line 1174: This is an `Option<Option<u64>>`. Are both levels of `Option` meaningful or can they be flattened?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -920,36 +1160,20 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
-                    .map(|opt| opt.flatten())
-                    .transpose()
-            })
+            // If we don't have an explicit birthday tree size, find something nearby.
+            .or_else(|| get_tree_size_near(birthday_height).transpose())
             .transpose()?;
 
-        // Get the note commitment tree size as of the end of the recover-until height.
+        // Get the note commitment tree size as of the start of the recover-until height.
         let recover_until_size = recover_until_height
```
daira, line 1168: This is an `Option<Option<u64>>`. Are both levels of `Option` meaningful or can they be flattened?

str4d: Both are meaningful:
- The outer optional means "do we have a recover-until height?", and is `unwrap_or`ed to the corresponding birthday height values as relevant.
- The inner optional means "have we scanned enough information from the chain to get tree size data for the recover-until height?", and is used for triggering fallback behaviour.

zcash_client_sqlite/src/wallet.rs
str4d: See https://github.com/zcash/librustzcash/pull/1564#discussion_r1796138798. Any nested optional derived from `recover_until_height` is meaningful in this way.

zcash_client_sqlite/src/wallet.rs
daira: Currently the mapping to the existing API is done in zcash-light-client-ffi. That's where it makes most sense to do the mapping from 0:0 to 1:1, *after* adding the numerators and denominators, IMHO (i.e. in https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/156).

daira (approved):
utACK with comments.

zcash_client_sqlite/src/wallet.rs
daira: Please add a comment about this.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -887,27 +1089,65 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
+                    )
+                    .optional()
+                    .map(|opt| opt.flatten())
+                    .transpose()
+                })
+                .transpose()
+        };
+
         // Get the starting note commitment tree size from the wallet birthday, or failing that
         // from the blocks table.
```
daira, line 1149: "or failing that from the blocks table" seems wrong.

zcash_client_sqlite/src/wallet.rs
nuttycom: *sigh* it's very frustrating that Rust doesn't have `traverse`, and therefore we have to make do with `.map(...).transpose()`. Personally I find the functional approach nicer, but I'll change it.

zcash_client_sqlite/src/wallet.rs
nuttycom: I honestly don't find this any clearer.

zcash_client_sqlite/src/wallet.rs
nuttycom: The scan progress denominator is guaranteed to always be nonzero, so adding the numerators and denominators always results in a valid fractional value.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1304,34 +1338,37 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
         })
-        .or(sapling_progress.recover)
-        .or(orchard_progress.recover);
+        .or(sapling_progress)
+        .or(orchard_progress);
+
+    let progress = match progress {
+        Some(p) => p,
+        None => return Ok(None),
+    };
```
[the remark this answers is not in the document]
str4d: In a pairing, @nuttycom and I determined that this was actually fine, as after these bugfixes these two states should essentially never diverge.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -870,86 +1074,114 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
-        Ok(Progress { scan, recover })
+
+        Ok(scan.map(|scan| Progress::new(scan, recover)))
     } else {
+        // In case we didn't have information about the tree size at the recover-until
+        // height, get the tree size from a nearby subtree. It's fine for this to be
+        // approximate; it just shifts the boundary between scan and recover progress.
+        let mut get_tree_size_near = |as_of: BlockHeight| {
+            let size_from_blocks = stmt_start_tree_size
+                .query_row(named_params![":start_height": u32::from(as_of)], |row| {
```
str4d, line 1118: Good catch on this! (`as_of` was incorrectly `birthday_height` before the most recent push.)

str4d (commented):
utACK 108e9530a2ecfd3247b120d792668fdfc690167d. My remaining comment feedback is non-blocking.

zcash_client_sqlite/src/wallet.rs
nuttycom: In pairing with @str4d, we determined that there may have already been uncommon but possible corner cases where the recovery ratio could have had a zero denominator: this would occur in circumstances where no notes for a given shielded protocol exist in blocks between the wallet's birthday height and the wallet recovery height. This is actually a relatively easily observable situation on testnet, since most of the time there's no shielded testnet activity. As such, `0:0` is not an API-breaking change; we just hadn't previously thought about or documented it adequately.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/a56de09..279479c19e8db8e97a0d618aee9be3dd4988181f) to address comments from code review and prepare for the hotfix release.

str4d (approved):
utACK 279479c19e8db8e97a0d618aee9be3dd4988181f

zcash_client_backend/src/data_api.rs
```diff
@@ -534,10 +534,11 @@ impl<AccountId: Eq + Hash> WalletSummary<AccountId> {
     /// Progress is represented in terms of the ratio between notes scanned and the total
     /// number of notes added to the chain in the relevant window. This ratio should only
     /// be used to compute progress percentages, and the numerator and denominator should
-    /// not be treated as authoritative note counts.
+    /// not be treated as authoritative note counts. The denominator of this ratio is
+    /// guaranteed to be nonzero.
```
daira, line 538: According to https://github.com/zcash/librustzcash/pull/1564#discussion_r1796339614 that is incorrect. That's okay for this release, but please file a PR to fix it.

zcash_client_sqlite/src/wallet.rs
daira: I was mistaken; `stmt_start_tree_size` does use the `blocks` table. `size_from_subtree_roots` doesn't though, so this comment is still not quite accurate.

daira (commented):
Post-hoc ACK 279479c19e8db8e97a0d618aee9be3dd4988181f. There is a minor inaccuracy in a comment, I think (https://github.com/zcash/librustzcash/pull/1564/files#r1797495621).

PR #1565, opened 2024-10-10
Derive Serialize/Deserialize for certain simple structs

ec2:
This PR introduces "non-controversial" derives of `Serialize` and `Deserialize` for some structs that we persist to storage in our memory wallet implementation. This is by no means all the structs that we want to be able to serialize, but they are the ones that should be stable or are simple internal representations.

### DIFF
--- a/zcash_client_backend/src/fees.rs
+++ b/zcash_client_backend/src/fees.rs
@@ -1,6 +1,5 @@
 use std::fmt;
 
-use serde::{Deserialize, Serialize};
 use zcash_primitives::{
     consensus::{self, BlockHeight},
     memo::MemoBytes,
@@ -25,10 +24,10 @@ pub mod zip317;
 /// `ChangeValue` represents either a proposed change output to a shielded pool
 /// (with an optional change memo), or if the "transparent-inputs" feature is
 /// enabled, an ephemeral output to the transparent pool.
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ChangeValue(ChangeValueInner);
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 enum ChangeValueInner {
     Shielded {
         protocol: ShieldedProtocol,
@@ -115,7 +114,7 @@ impl ChangeValue {
 /// The amount of change and fees required to make a transaction's inputs and
 /// outputs balance under a specific fee rule, as computed by a particular
 /// [`ChangeStrategy`] that is aware of that rule.
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct TransactionBalance {
     proposed_change: Vec<ChangeValue>,
     fee_required: NonNegativeAmount,
--- a/zcash_client_backend/src/fees.rs
+++ b/zcash_client_backend/src/fees.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 
+use serde::{Deserialize, Serialize};
 use zcash_primitives::{
     consensus::{self, BlockHeight},
     memo::MemoBytes,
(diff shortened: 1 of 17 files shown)


zcash_client_backend/src/fees.rs
```diff
@@ -24,10 +25,10 @@ pub mod zip317;
 /// `ChangeValue` represents either a proposed change output to a shielded pool
 /// (with an optional change memo), or if the "transparent-inputs" feature is
 /// enabled, an ephemeral output to the transparent pool.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ChangeValue(ChangeValueInner);
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
nuttycom, line 31: I'm surprised that serde serialization is needed for this type. Ordinarily, this type would only be serialized via a `Proposal`, in which case it should use the `protobuf` encoding. What is the reason that this serialization is being added?

zcash_client_backend/src/fees.rs
```diff
@@ -114,7 +115,7 @@ impl ChangeValue {
 /// The amount of change and fees required to make a transaction's inputs and
 /// outputs balance under a specific fee rule, as computed by a particular
 /// [`ChangeStrategy`] that is aware of that rule.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
nuttycom, line 118: Same here, `Proposal` components should be serialized using the protobuf encoding.

zcash_client_backend/src/proposal.rs
```diff
@@ -330,14 +331,14 @@ impl<FeeRuleT: Debug, NoteRef> Debug for Proposal<FeeRuleT, NoteRef> {
 }
 
 /// A reference to either a payment or change output within a step.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
```
nuttycom, line 334: Same, use protobuf instead.

zcash_primitives/src/transaction/fees.rs
```diff
@@ -60,7 +62,7 @@ pub trait FutureFeeRule: FeeRule {
 }
 
 /// An enumeration of the standard fee rules supported by the wallet.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
nuttycom, line 65: This should also only need to be serialized via the `Proposal` protobuf encoding.

zcash_primitives/src/transaction/fees/zip317.rs
```diff
@@ -53,7 +55,7 @@ pub const MINIMUM_FEE: NonNegativeAmount = NonNegativeAmount::const_from_u64(10_
 ///
 /// [`FeeRule`]: crate::transaction::fees::FeeRule
 /// [ZIP 317]: https//zips.z.cash/zip-0317
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
nuttycom, line 58: There should be no need for serde serialization for this type.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -67,7 +68,7 @@ const ZFUTURE_TX_VERSION: u32 = 0x0000FFFF;
 ///   that have been mined.
 /// - For v5 transactions onwards, this identifier is derived only from "effecting" data,
 ///   and is non-malleable in all contexts.
-#[derive(Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
nuttycom, line 71: This should use hand-written `Serialize` and `Deserialize` impls that serialize to the canonical txid text string representation.

nuttycom (changes requested):
It looks like there are a number of types here that already have canonical serialization, either to `protobuf` or string representations. Please reuse those serialized forms instead.

ec2:
@nuttycom w.r.t. Proposal and associated structs, we need it to be serializable for`wasm_bindgen` so that we can pass it between JS and Wasm.

zcash_primitives/src/transaction/mod.rs
ec2: Why? Also I dont see the definition for the canonical string rep - could you point me to it? I do see `Display` implemented so maybe that is it? Is there a `FromStr` for `TxId` anywhere?

ec2:
As per @daira 's suggestion, we will try using the protobuf encoding to do the passing back and forth.

zcash_primitives/src/transaction/mod.rs
nuttycom: We haven't needed to implement parsing, but yes the string representation used in the `Display` impl is canonical.

nuttycom:
I believe that the objectives of this PR were eventually satisfied by #1634

Issue #1413, opened 2024-06-13
`zcash_client_*`: Fix bug where transparent funds can't be shielded before the normal confirmation depth


str4d:
We've had Zashi user reports that they can't shield funds after 1 confirmation as they expect, and instead have to wait for 10 confirmations (used for normal spending). The mobile SDKs don't allow shielding funds if there are insufficient funds to shield in the wallet balance.

str4d:
I think I found the cause of the bug in `zcash_client_sqlite` here: https://github.com/zcash/librustzcash/pull/1257#discussion_r1637994605

https://github.com/zcash/librustzcash/blob/5a2a670ffca5d81e54c166d0f6b5cf5bcce84f32/zcash_client_sqlite/src/wallet.rs#L1299-L1305

The `:max_height` variable is set to `zero_conf_height`, but the latter is not actually zero-conf, it's the same as `summary_height` (except that it doesn't force `min_confirmations` to be at least 1).

str4d:
I think we have fixed this; we should check.

daira:
@true-jared wrote in #1572 which I'm merging into this issue:
> Even though SDK has 1 confirmation set let minimumShieldingConfirmations: UInt32 = 1 we still need to wait 10 confirmations.
>
> How to reproduce:
> - I tapped a button to shield my funds, it:
> - created a transaction in a pending state with is_shielding TRUE
> - after 1st confirmation, it flipped to sending instead of shielding, I checked the DB and there is no TRUE anymore (see the DB)
> - after 10 confirmations, it flipped to sent
> - when I restore the same wallet, I see shielded again
>
> Few more observations:
> - when I try to restore the wallet on a different simulator while it’s still waiting on 10 confirmations and should be shielding funds but it’s sending, I see correct state restored.. so restore is able to bring the TRUE flag back but where it matters it’s broken
>
> - db export shared via Slack here: https://zcash.slack.com/archives/C02C1GZ0R6F/p1724835111523729

daira:
Copying a comment I wrote at https://github.com/zcash/librustzcash/pull/1570/files#r1817869234 here. Note that it mainly concerns the fact that we should not expect to be able to do zero-conf shielding yet, but the information about where the number of confirmations is being set in each SDK may be useful.

[...]

PR #1570, opened 2024-10-13
[#1411] Refactor AccountBalance to use Balance for transparent funds

### LINKED ISSUE #1411: Refactor `AccountBalance` to use `Balance` for transparent funds
Currently `AccountBalance` stores all unspent transparent funds as a single value, irrespective of confirmation depth:
https://github.com/zcash/librustzcash/blob/5a2a670ffca5d81e54c166d0f6b5cf5bcce84f32/zcash_client_backend/src/data_api.rs#L211-L229

This is reasonable for the mobile SDKs, which explicitly do not support arbitrary spending of transparent funds (they must always be spent in a carefully-controlled manner, usually being shielded to a shielded address in the wallet first). However, for the `zcashd` wallet replacement we will need the ability to spend transparent funds, and thus to determine what transparent funds are suitable for selection. We will therefore need to distinguish to users between transparent funds that are available to the wallet for spending, and funds that still require confirmations.

The mobile SDKs can continue to behave as before, by adding up the separate fields to get the same value as they previously reported.

pacu:
fixes #1411

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -8,21 +8,36 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Changed
-- `zcash_client_backend::data_api::AccountBalance`:
-  - Refactored to use `Balance` for transparent funds (issue #1411).
-    `AccountBalance` now has an `unshielded_balance()` that uses `Balance` and replaces the
-    (now deleted) `unshielded` non-negative amount.
+- `zcash_client_backend::data_api::AccountBalance`: Refactored to use `Balance`
+  for transparent funds (issue #1411). It now has an `unshielded_balance()`
+  method that returns `Balance`, allowing the unshielded spendable, unshielded
+  pending change, and unshielded pending non-change values to be tracked
+  separately.
 - `zcash_client_backend::data_api::WalletRead`:
   - The `create_account`, `import_account_hd`, and `import_account_ufvk`
     methods now each take additional `account_name` and `key_source` arguments.
     These allow the wallet backend to store additional metadata that is useful
     to applications managing these accounts.
 - `zcash_client_backend::data_api::AccountSource`:
-  - Both the `Derived` and `Importants` now have an additional `key_source`
-    property that is used to convey application-specific key source metadata.
+  - Both `Derived` and `Imported` alternatives of `AccountSource` now have an
+    additional `key_source` field that is used to convey application-specific
+    key source metadata.
   - The `Copy` impl for this type has been removed.
 - `zcash_client_backend::data_api::Account` has an additional `name` method
   that returns the human-readable name of the account, if any.
+
+### Deprecated
+- `AccountBalance::unshielded`. Instead use `unshielded_balance` which
+  provides a `Balance` value. Its `total()` method can be used to obtain the
+  total of transparent funds.
+
+### Removed
+- `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
+  `AccountBalance::with_unshielded_balance_mut` with a closure that calls
+  the appropriate `add_*_value` method(s) of `Balance` on its argument.
+  Note that the appropriate method(s) depend on whether the funds are
+  spendable, pending change, or pending non-change (previously, only the
+  total unshielded value was tracked).
 
 ## [0.15.0] - 2024-11-14
 
@@ -109,15 +124,6 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::data_api`:
-  - `AccountBalance::unshielded`. `AccountBalance` no longer provides the `unshielded`
-    method returning a `NonNegativeAmount` for the total of transparent funds.
-    Instead use `unshielded_balance` which provides a `Balance` value.
-  - `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
-    `AccountBalance::with_unshielded_balance_mut` with a closure that calls
-    the appropriate `add_*_value` method(s) of `Balance` on its argument.
-    Note that the appropriate method(s) depend on whether the funds are
-    spendable, pending change, or pending non-change (previously, only the
-    total unshielded value was tracked).
   - `WalletSummary::scan_progress` and `WalletSummary::recovery_progress` have
     been removed. Use `WalletSummary::progress` instead.
   - `testing::input_selector` use explicit `InputSelector` constructors
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -8,6 +8,10 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Changed
+- `zcash_client_backend::data_api::AccountBalance`:
+  - Refactored to use `Balance` for transparent funds (issue #1411).
+    `AccountBalance` now has an `unshielded_balance()` that uses `Balance` and replaces the
+    (now deleted) `unshielded` non-negative amount.
 - `zcash_client_backend::data_api::WalletRead`:
   - The `create_account`, `import_account_hd`, and `import_account_ufvk`
     methods now each take additional `account_name` and `key_source` arguments.
@@ -105,6 +109,15 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::data_api`:
+  - `AccountBalance::unshielded`. `AccountBalance` no longer provides the `unshielded`
+    method returning a `NonNegativeAmount` for the total of transparent funds.
+    Instead use `unshielded_balance` which provides a `Balance` value.
+  - `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
+    `AccountBalance::with_unshielded_balance_mut` with a closure that calls
+    the appropriate `add_*_value` method(s) of `Balance` on its argument.
+    Note that the appropriate method(s) depend on whether the funds are
+    spendable, pending change, or pending non-change (previously, only the
+    total unshielded value was tracked).
   - `WalletSummary::scan_progress` and `WalletSummary::recovery_progress` have
     been removed. Use `WalletSummary::progress` instead.
   - `testing::input_selector` use explicit `InputSelector` constructors
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -231,14 +231,8 @@ pub struct AccountBalance {
     /// The value of unspent Orchard outputs belonging to the account.
     orchard_balance: Balance,
 
-    /// The value of all unspent transparent outputs belonging to the account, irrespective of
-    /// confirmation depth.
-    ///
-    /// Unshielded balances are not subject to confirmation-depth constraints, because the only
-    /// possible operation on a transparent balance is to shield it, it is possible to create a
-    /// zero-conf transaction to perform that shielding, and the resulting shielded notes will be
-    /// subject to normal confirmation rules.
-    unshielded: NonNegativeAmount,
+    /// The value of all unspent transparent outputs belonging to the account.
+    unshielded_balance: Balance,
 }
 
 impl AccountBalance {
@@ -246,12 +240,14 @@ impl AccountBalance {
     pub const ZERO: Self = Self {
         sapling_balance: Balance::ZERO,
         orchard_balance: Balance::ZERO,
-        unshielded: NonNegativeAmount::ZERO,
+        unshielded_balance: Balance::ZERO,
     };
 
     fn check_total(&self) -> Result<NonNegativeAmount, BalanceError> {
-        (self.sapling_balance.total() + self.orchard_balance.total() + self.unshielded)
-            .ok_or(BalanceError::Overflow)
+        (self.sapling_balance.total()
+            + self.orchard_balance.total()
+            + self.unshielded_balance.total())
+        .ok_or(BalanceError::Overflow)
     }
 
     /// Returns the [`Balance`] of Sapling funds in the account.
@@ -259,7 +255,7 @@ impl AccountBalance {
         &self.sapling_balance
     }
 
-    /// Provides a `mutable reference to the [`Balance`] of Sapling funds in the account
+    /// Provides a mutable reference to the [`Balance`] of Sapling funds in the account
     /// to the specified callback, checking invariants after the callback's action has been
     /// evaluated.
     pub fn with_sapling_balance_mut<A, E: From<BalanceError>>(
@@ -276,7 +272,7 @@ impl AccountBalance {
         &self.orchard_balance
     }
 
-    /// Provides a `mutable reference to the [`Balance`] of Orchard funds in the account
+    /// Provides a mutable reference to the [`Balance`] of Orchard funds in the account
     /// to the specified callback, checking invariants after the callback's action has been
     /// evaluated.
     pub fn with_orchard_balance_mut<A, E: From<BalanceError>>(
@@ -289,22 +285,36 @@ impl AccountBalance {
     }
 
     /// Returns the total value of unspent transparent transaction outputs belonging to the wallet.
+    #[deprecated(
+        note = "this function is deprecated. Please use [`AccountBalance::unshielded_balance`] instead."
+    )]
     pub fn unshielded(&self) -> NonNegativeAmount {
-        self.unshielded
+        self.unshielded_balance.total()
     }
 
-    /// Adds the specified value to the unshielded total, checking for overflow of
-    /// the total account balance.
-    pub fn add_unshielded_value(&mut self, value: NonNegativeAmount) -> Result<(), BalanceError> {
-        self.unshielded = (self.unshielded + value).ok_or(BalanceError::Overflow)?;
+    /// Returns the [`Balance`] of unshielded funds in the account.
+    pub fn unshielded_balance(&self) -> &Balance {
+        &self.unshielded_balance
+    }
+
+    /// Provides a mutable reference to the [`Balance`] of transparent funds in the account
+    /// to the specified callback, checking invariants after the callback's action has been
+    /// evaluated.
+    pub fn with_unshielded_balance_mut<A, E: From<BalanceError>>(
+        &mut self,
+        f: impl FnOnce(&mut Balance) -> Result<A, E>,
+    ) -> Result<A, E> {
+        let result = f(&mut self.unshielded_balance)?;
         self.check_total()?;
-        Ok(())
+        Ok(result)
     }
 
     /// Returns the total value of funds belonging to the account.
     pub fn total(&self) -> NonNegativeAmount {
-        (self.sapling_balance.total() + self.orchard_balance.total() + self.unshielded)
-            .expect("Account balance cannot overflow MAX_MONEY")
+        (self.sapling_balance.total()
+            + self.orchard_balance.total()
+            + self.unshielded_balance.total())
+        .expect("Account balance cannot overflow MAX_MONEY")
     }
 
     /// Returns the total value of shielded (Sapling and Orchard) funds that may immediately be
@@ -1204,8 +1214,9 @@ pub trait WalletRead {
     /// or `Ok(None)` if the wallet has no initialized accounts.
     fn get_wallet_birthday(&self) -> Result<Option<BlockHeight>, Self::Error>;
 
-    /// Returns the wallet balances and sync status for an account given the specified minimum
-    /// number of confirmations, or `Ok(None)` if the wallet has no balance data available.
+    /// Returns a [`WalletSummary`] that represents the sync status, and the wallet balances
+    /// given the specified minimum number of confirmations for all accounts known to the
+    /// wallet; or `Ok(None)` if the wallet has no summary data available.
     fn get_wallet_summary(
         &self,
         min_confirmations: u32,
--- a/zcash_client_backend/src/data_api/testing/transparent.rs
+++ b/zcash_client_backend/src/data_api/testing/transparent.rs
@@ -20,10 +20,10 @@ use zcash_protocol::local_consensus::LocalNetwork;
 
 use super::TestAccount;
 
-/// Checks whether the transparent balance of the given test [`account`] is as [`expected`]
-/// considering the [`min_confirmations`]. It is assumed that zero or one [`min_confirmations`]
-/// are treated the same so this function also checks the other case when 1 or 0 [`min_confirmations`]
-/// are passed as arguments.
+/// Checks whether the transparent balance of the given test `account` is as `expected`
+/// considering the `min_confirmations`. It is assumed that zero or one `min_confirmations`
+/// are treated the same, and so this function also checks the other case when
+/// `min_confirmations` is 0 or 1.
 fn check_balance<DSF>(
     st: &TestState<impl TestCache, <DSF as DataStoreFactory>::DataStore, LocalNetwork>,
     account: &TestAccount<<DSF as DataStoreFactory>::Account>,
--- a/zcash_client_backend/src/data_api/testing/transparent.rs
+++ b/zcash_client_backend/src/data_api/testing/transparent.rs
@@ -4,7 +4,7 @@ use crate::{
             AddressType, DataStoreFactory, ShieldedProtocol, TestBuilder, TestCache, TestState,
         },
         wallet::input_selection::GreedyInputSelector,
-        Account as _, InputSource, WalletRead, WalletWrite,
+        Account as _, Balance, InputSource, WalletRead, WalletWrite,
     },
     fees::{standard, DustOutputPolicy, StandardFeeRule},
     wallet::WalletTransparentOutput,
@@ -13,8 +13,75 @@ use assert_matches::assert_matches;
 use sapling::zip32::ExtendedSpendingKey;
 use zcash_primitives::{
     block::BlockHash,
+    legacy::TransparentAddress,
     transaction::components::{amount::NonNegativeAmount, OutPoint, TxOut},
 };
+use zcash_protocol::local_consensus::LocalNetwork;
+
+use super::TestAccount;
+/// checks whether the transparent balance of the given test [`account`] is as [`expected`]
+/// considering the [`min_confirmations`]. It is assumed that zero or one [`min_confirmations`]
+/// are treated the same so this function also checks the other case when 1 or 0 [`min_confirmations`]
+/// are passed as arguments.
+fn check_balance<DSF>(
+    st: &TestState<impl TestCache, <DSF as DataStoreFactory>::DataStore, LocalNetwork>,
+    account: &TestAccount<<DSF as DataStoreFactory>::Account>,
+    taddr: &TransparentAddress,
+    min_confirmations: u32,
+    expected: &Balance,
+) where
+    DSF: DataStoreFactory,
+{
+    // Check the wallet summary returns the expected transparent balance.
+    let summary = st
+        .wallet()
+        .get_wallet_summary(min_confirmations)
+        .unwrap()
+        .unwrap();
+    let balance = summary.account_balances().get(&account.id()).unwrap();
+
+    #[allow(deprecated)]
+    let old_unshielded_value = balance.unshielded();
+    assert_eq!(old_unshielded_value, expected.total());
+    assert_eq!(balance.unshielded_balance(), expected);
+
+    // Check the older APIs for consistency.
+    let mempool_height = st.wallet().chain_height().unwrap().unwrap() + 1;
+    assert_eq!(
+        st.wallet()
+            .get_transparent_balances(account.id(), mempool_height)
+            .unwrap()
+            .get(taddr)
+            .cloned()
+            .unwrap_or(NonNegativeAmount::ZERO),
+        expected.total(),
+    );
+    assert_eq!(
+        st.wallet()
+            .get_spendable_transparent_outputs(taddr, mempool_height, min_confirmations)
+            .unwrap()
+            .into_iter()
+            .map(|utxo| utxo.value())
+            .sum::<Option<NonNegativeAmount>>(),
+        Some(expected.spendable_value()),
+    );
+
+    // we currently treat min_confirmations the same regardless they are 0 (zero confirmations)
+    // or 1 (one block confirmation). We will check if this assumption holds until it's no
+    // longer made. If zero and one [`min_confirmations`] are treated differently in the future,
+    // this check should then be removed.
+    if min_confirmations == 0 || min_confirmations == 1 {
+        assert_eq!(
+            st.wallet()
+                .get_spendable_transparent_outputs(taddr, mempool_height, 1 - min_confirmations)
+                .unwrap()
+                .into_iter()
+                .map(|utxo| utxo.value())
+                .sum::<Option<NonNegativeAmount>>(),
+            Some(expected.spendable_value()),
+        );
+    }
+}
 
 pub fn put_received_transparent_utxo<DSF>(dsf: DSF)
 where
@@ -136,46 +203,8 @@ where
     }
     st.scan_cached_blocks(start_height, 10);
 
-    let check_balance = |st: &TestState<_, DSF::DataStore, _>, min_confirmations: u32, expected| {
-        // Check the wallet summary returns the expected transparent balance.
-        let summary = st
-            .wallet()
-            .get_wallet_summary(min_confirmations)
-            .unwrap()
-            .unwrap();
-        let balance = summary.account_balances().get(&account.id()).unwrap();
-        // TODO: in the future, we will distinguish between available and total
-        // balance according to `min_confirmations`
-        assert_eq!(balance.unshielded(), expected);
-
-        // Check the older APIs for consistency.
-        let mempool_height = st.wallet().chain_height().unwrap().unwrap() + 1;
-        assert_eq!(
-            st.wallet()
-                .get_transparent_balances(account.id(), mempool_height)
-                .unwrap()
-                .get(taddr)
-                .cloned()
-                .unwrap_or(NonNegativeAmount::ZERO),
-            expected,
-        );
-        assert_eq!(
-            st.wallet()
-                .get_spendable_transparent_outputs(taddr, mempool_height, 0)
-                .unwrap()
-                .into_iter()
-                .map(|utxo| utxo.value())
-                .sum::<Option<NonNegativeAmount>>(),
-            Some(expected),
-        );
-    };
-
     // The wallet starts out with zero balance.
-    // TODO: Once we have refactored `get_wallet_summary` to distinguish between available
-    // and total balance, we should perform additional checks against available balance;
-    // we use minconf 0 here because all transparent funds are considered shieldable,
-    // irrespective of confirmation depth.
-    check_balance(&st, 0, NonNegativeAmount::ZERO);
+    check_balance::<DSF>(&st, &account, taddr, 0, &Balance::ZERO);
 
     // Create a fake transparent output.
     let value = NonNegativeAmount::from_u64(100000).unwrap();
@@ -192,7 +221,12 @@ where
         .unwrap();
 
     // The wallet should detect the balance as available
-    check_balance(&st, 0, value);
+    let mut zero_or_one_conf_value = Balance::ZERO;
+
+    // add the spendable value to the expected balance
+    zero_or_one_conf_value.add_spendable_value(value).unwrap();
+
+    check_balance::<DSF>(&st, &account, taddr, 0, &zero_or_one_conf_value);
 
     // Shield the output.
     let input_selector = GreedyInputSelector::new();
@@ -216,14 +250,14 @@ where
 
     // The wallet should have zero transparent balance, because the shielding
     // transaction can be mined.
-    check_balance(&st, 0, NonNegativeAmount::ZERO);
+    check_balance::<DSF>(&st, &account, taddr, 0, &Balance::ZERO);
 
     // Mine the shielding transaction.
     let (mined_height, _) = st.generate_next_block_including(txid);
     st.scan_cached_blocks(mined_height, 1);
 
     // The wallet should still have zero transparent balance.
-    check_balance(&st, 0, NonNegativeAmount::ZERO);
+    check_balance::<DSF>(&st, &account, taddr, 0, &Balance::ZERO);
 
     // Unmine the shielding transaction via a reorg.
     st.wallet_mut()
@@ -232,7 +266,7 @@ where
     assert_eq!(st.wallet().chain_height().unwrap(), Some(mined_height - 1));
 
     // The wallet should still have zero transparent balance.
-    check_balance(&st, 0, NonNegativeAmount::ZERO);
+    check_balance::<DSF>(&st, &account, taddr, 0, &Balance::ZERO);
 
     // Expire the shielding transaction.
     let expiry_height = st
@@ -243,5 +277,91 @@ where
         .expiry_height();
     st.wallet_mut().update_chain_tip(expiry_height).unwrap();
 
-    check_balance(&st, 0, value);
+    check_balance::<DSF>(&st, &account, taddr, 0, &zero_or_one_conf_value);
+}
+
+/// This test attempts to verify that transparent funds spendability is
+/// accounted for properly given the different minimum confirmations values
+/// that can be set when querying for balances.
+pub fn transparent_balance_spendability<DSF>(dsf: DSF, cache: impl TestCache)
+where
+    DSF: DataStoreFactory,
+{
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(dsf)
+        .with_block_cache(cache)
+        .with_account_from_sapling_activation(BlockHash([0; 32]))
+        .build();
+
+    let account = st.test_account().cloned().unwrap();
+    let uaddr = st
+        .wallet()
+        .get_current_address(account.id())
+        .unwrap()
+        .unwrap();
+    let taddr = uaddr.transparent().unwrap();
+
+    // Initialize the wallet with chain data that has no shielded notes for us.
+    let not_our_key = ExtendedSpendingKey::master(&[]).to_diversifiable_full_viewing_key();
+    let not_our_value = NonNegativeAmount::const_from_u64(10000);
+    let (start_height, _, _) =
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    for _ in 1..10 {
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    }
+    st.scan_cached_blocks(start_height, 10);
+
+    // The wallet starts out with zero balance.
+    check_balance::<DSF>(
+        &st as &TestState<_, DSF::DataStore, _>,
+        &account,
+        taddr,
+        0,
+        &Balance::ZERO,
+    );
+
+    // Create a fake transparent output.
+    let value = NonNegativeAmount::from_u64(100000).unwrap();
+    let txout = TxOut {
+        value,
+        script_pubkey: taddr.script(),
+    };
+
+    // Pretend the output was received in the chain tip.
+    let height = st.wallet().chain_height().unwrap().unwrap();
+    let utxo = WalletTransparentOutput::from_parts(OutPoint::fake(), txout, Some(height)).unwrap();
+    st.wallet_mut()
+        .put_received_transparent_utxo(&utxo)
+        .unwrap();
+
+    // The wallet should detect the balance as available
+    let mut zero_or_one_conf_value = Balance::ZERO;
+
+    // add the spendable value to the expected balance
+    zero_or_one_conf_value.add_spendable_value(value).unwrap();
+
+    check_balance::<DSF>(&st, &account, taddr, 0, &zero_or_one_conf_value);
+
+    // now if we increase the number of confirmations our spendable balance should
+    // be zero and the total balance equal to `value`
+    let mut not_confirmed_yet_value = Balance::ZERO;
+
+    not_confirmed_yet_value
+        .add_pending_spendable_value(value)
+        .unwrap();
+
+    check_balance::<DSF>(&st, &account, taddr, 2, &not_confirmed_yet_value);
+
+    // Add one extra block
+    st.generate_empty_block();
+
+    // Scan that block
+    st.scan_cached_blocks(height, 1);
+
+    // now we generate one more block and the balance should be the same as when the
+    // check_balance function was called with zero or one confirmation.
+    st.generate_empty_block();
+    st.scan_cached_blocks(height + 1, 1);
+
+    check_balance::<DSF>(&st, &account, taddr, 2, &zero_or_one_conf_value);
 }
--- a/zcash_client_backend/src/data_api/testing/transparent.rs
+++ b/zcash_client_backend/src/data_api/testing/transparent.rs
@@ -19,7 +19,8 @@ use zcash_primitives::{
 use zcash_protocol::local_consensus::LocalNetwork;
 
 use super::TestAccount;
-/// checks whether the transparent balance of the given test [`account`] is as [`expected`]
+
+/// Checks whether the transparent balance of the given test [`account`] is as [`expected`]
 /// considering the [`min_confirmations`]. It is assumed that zero or one [`min_confirmations`]
 /// are treated the same so this function also checks the other case when 1 or 0 [`min_confirmations`]
 /// are passed as arguments.
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -43,6 +43,16 @@ and this library adheres to Rust's notion of
 - Migrated to `rusqlite 0.32`.
 - `error::SqliteClientError` has additional variant `NoteFilterInvalid`
 
+### Fixed
+- `zcash_client_sqlite::WalletDb`'s implementation of
+  `zcash_client_backend::data_api::WalletRead::get_wallet_summary` has been
+  fixed to take account of `min_confirmations` for transparent balances.
+  Note that this implementation treats `min_confirmations == 0` the same
+  as `min_confirmations == 1` for both shielded and transparent TXOs.
+  It also does not currently distinguish between pending change and
+  non-change; the pending value is all counted as non-change (issue
+  [#1592](https://github.com/zcash/librustzcash/issues/1592)).
+
 ## [0.12.2] - 2024-10-21
 
 ### Fixed
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -47,6 +47,8 @@ and this library adheres to Rust's notion of
 - `zcash_client_sqlite::WalletDb`'s implementation of
   `zcash_client_backend::data_api::WalletRead::get_wallet_summary` has been
   fixed to take account of `min_confirmations` for transparent balances.
+  (Previously, it would treat transparent balances as though
+  `min_confirmations` were `1` even if it was set to a higher value.)
   Note that this implementation treats `min_confirmations == 0` the same
   as `min_confirmations == 1` for both shielded and transparent TXOs.
   It also does not currently distinguish between pending change and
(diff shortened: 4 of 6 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -239,19 +239,19 @@ pub struct AccountBalance {
     /// possible operation on a transparent balance is to shield it, it is possible to create a
     /// zero-conf transaction to perform that shielding, and the resulting shielded notes will be
     /// subject to normal confirmation rules.
```
daira, line 241: Please also update the comment. It should just say "The value of unspent transparent outputs belonging to the account."

zcash_client_backend/src/data_api.rs
```diff
@@ -289,22 +289,45 @@ impl AccountBalance {
         Ok(result)
     }
 
+    
     /// Returns the total value of unspent transparent transaction outputs belonging to the wallet.
+    #[deprecated(
+        note = "this function is deprecated. Please use AccountBalance::unshielded_balance instead."
```
daira, line 295:
```suggestion
        note = "This function is deprecated. Please use AccountBalance::unshielded_balance instead."
```

zcash_client_backend/src/data_api.rs
```diff
@@ -290,22 +286,46 @@ impl AccountBalance {
     }
 
     /// Returns the total value of unspent transparent transaction outputs belonging to the wallet.
+    #[deprecated(
+        note = "this function is deprecated. Please use AccountBalance::unshielded_balance instead."
```
daira, line 290:
```suggestion
        note = "this function is deprecated. Please use [`AccountBalance::unshielded_balance`] instead."
```

zcash_client_backend/src/data_api.rs
```diff
@@ -290,22 +286,46 @@ impl AccountBalance {
...
+    ) -> Result<A, E> {
+        let result = f(&mut self.unshielded_balance)?;
+        self.check_total()?;
+        Ok(result)
     }
 
     /// Adds the specified value to the unshielded total, checking for overflow of
     /// the total account balance.
+    #[deprecated(
+        note = "this function is deprecated. Please use the `Balance::add_spendable_value` on the unshielded field instead instead."
```
daira, line 316:
```suggestion
        note = "this function is deprecated. Please use `account_balance.with_unshielded_balance_mut(|unshielded_balance| unshielded_balance.add_pending_spendable_value(value))` instead."
```

zcash_client_backend/src/data_api.rs
```diff
@@ -290,22 +286,46 @@ impl AccountBalance {
...
+    )]
     pub fn add_unshielded_value(&mut self, value: NonNegativeAmount) -> Result<(), BalanceError> {
-        self.unshielded = (self.unshielded + value).ok_or(BalanceError::Overflow)?;
-        self.check_total()?;
+        self.unshielded_balance.add_pending_spendable_value(value)?;
         Ok(())
```
daira, line 320:
```suggestion
        self.with_unshielded_balance_mut(|unshielded_balance| {
            unshielded_balance.add_pending_spendable_value(value)
        })
```
This fixes a bug: the PR's code did not check the overflow invariant. See discussion above for whether we should retain this method at all.

zcash_client_backend/src/data_api.rs
```diff
@@ -290,22 +286,46 @@ impl AccountBalance {
...
+        self.check_total()?;
+        Ok(result)
     }
 
     /// Adds the specified value to the unshielded total, checking for overflow of
     /// the total account balance.
+    #[deprecated(
+        note = "this function is deprecated. Please use the `Balance::add_spendable_value` on the unshielded field instead instead."
+    )]
     pub fn add_unshielded_value(&mut self, value: NonNegativeAmount) -> Result<(), BalanceError> {
```
daira, line 318: It is unclear whether we should retain this method at all. It may be better to make a breaking change and delete it. The problem is that we cannot know what the intent was, because previously only the total was tracked. The value to be added might be spendable, pending change, pending non-change, or some combination.

Option 1: assume that the value to be added is pending non-change, and document the assumption.
```suggestion
    /// Adds the specified value to the pending non-change unshielded balance,
    /// checking for overflow of the total account balance.
    ///
    /// Note: Previously only the *total* unshielded balance was tracked. In this
    /// implementation, we assume that the intent was to add the value to the
    /// unshielded balance as pending non-change. This assumption may or may not
    /// be correct.
    #[deprecated(
        note = "this function is deprecated. Please use `[AccountBalance::with_unshielded_balance_mut]` with a closure that calls the appropriate `add_*_value` method(s) of [`Balance`] on its argument."
    )]
    pub fn add_unshielded_value(&mut self, value: NonNegativeAmount) -> Result<(), BalanceError> {
```

Option 2 is to delete this method, which requires a breaking semver version update:
```suggestion
```

zcash_client_backend/src/data_api.rs
```diff
@@ -290,22 +286,46 @@ impl AccountBalance {
...
-        self.unshielded = (self.unshielded + value).ok_or(BalanceError::Overflow)?;
-        self.check_total()?;
+        self.unshielded_balance.add_pending_spendable_value(value)?;
         Ok(())
     }
```
daira, line 322: This has a bug: it does not check the overflow invariant.

Option 1 (as above):
```suggestion
        self.with_unshielded_balance_mut(|unshielded_balance| {
            unshielded_balance.add_pending_spendable_value(value)
        })
    }

```

Option 2 (as above):
```suggestion
```

zcash_client_backend/src/data_api.rs
```diff
@@ -290,22 +286,46 @@ impl AccountBalance {
...
```
daira, line 321: This has a bug: it does not check the overflow invariant.

Option 1 (as above):
```suggestion
        self.with_unshielded_balance_mut(|unshielded_balance| {
            unshielded_balance.add_pending_spendable_value(value)
        })
    }
```

Option 2 (as above):
```suggestion
```

zcash_client_backend/src/data_api.rs
daira, line 318: It is unclear whether we should retain this method at all. It may be better to make a breaking change and delete it. The problem is that we cannot know what the intent was, because previously only the total was tracked. The value to be added might be spendable, pending change, pending non-change, or some combination.

Option 1: assume that the value to be added is pending non-change, and document the assumption.
```suggestion

    /// Adds the specified value to the pending non-change unshielded balance,
    /// checking for overflow of the total account balance.
    ///
    /// Note: Previously only the *total* unshielded balance was tracked. In this
    /// implementation, we assume that the intent was to add the value to the
    /// unshielded balance as pending non-change. This assumption may or may not
    /// be correct.
    #[deprecated(
        note = "this function is deprecated. Please use `[AccountBalance::with_unshielded_balance_mut]` with a closure that calls the appropriate `add_*_value` method(s) of [`Balance`] on its argument."
    )]
    pub fn add_unshielded_value(&mut self, value: NonNegativeAmount) -> Result<(), BalanceError> {
```

Option 2 is to delete this method, which requires a breaking semver version update:
```suggestion
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -383,16 +383,61 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+            SqliteClientError::CorruptedData(format!("Negative UTXO value {:?}", raw_value))
+        })?;
+
+        account_balances
+            .entry(account)
+            .or_insert(AccountBalance::ZERO)
+            .with_unshielded_balance_mut::<_, SqliteClientError>(|bal| {
+                bal.add_spendable_value(value)?;
+                Ok(())
+            })?;
```
daira, line 429:
```suggestion
            .with_unshielded_balance_mut(|bal| bal.add_spendable_value(value))?;
```
This is equivalent (compile-tested and rustfmt'ed).

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -421,7 +469,10 @@ pub(crate) fn add_transparent_account_balances(
         account_balances
             .entry(account)
             .or_insert(AccountBalance::ZERO)
-            .add_unshielded_value(value)?;
+            .with_unshielded_balance_mut::<_, SqliteClientError>(|bal| {
+                bal.add_pending_spendable_value(value)?;
+                Ok(())
+            })?;
```
daira, line 475:
```suggestion
            .with_unshielded_balance_mut(|bal| bal.add_pending_spendable_value(value))?;
```
This is equivalent (compile-tested and rustfmt'ed).

zcash_primitives/CHANGELOG.md
```diff
@@ -6,11 +6,22 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `zcash_client_backend::AccountBalance::with_unshielded_balance_mut`
 
+### Deprecated
+- `zcash_client_backend::AccountBalance::add_unshielded_value`
+- `zcash_client_backend::AccountBalance::unshielded`
```
daira, line 14:
```suggestion
- `zcash_client_backend::AccountBalance::unshielded`.
  Use `account_balance.unshielded_balance().total()` instead.
```

zcash_primitives/CHANGELOG.md
```diff
@@ -6,11 +6,22 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `zcash_client_backend::AccountBalance::with_unshielded_balance_mut`
 
+### Deprecated
+- `zcash_client_backend::AccountBalance::add_unshielded_value`
```
daira, line 13:
```suggestion
- `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
  `AccountBalance::with_unshielded_balance_mut` with a closure that calls the
  appropriate `add_*_value` method(s) of `Balance` on its argument. Note that
  the appropriate method(s) depend on whether the funds are spendable,
  pending change, or pending non-change. The deprecated implementation
  now assumes that they are pending non-change, which may not be correct.
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -383,16 +383,61 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
 pub(crate) fn add_transparent_account_balances(
     conn: &rusqlite::Connection,
     mempool_height: BlockHeight,
+    min_confirmations: u32,
     account_balances: &mut HashMap<AccountId, AccountBalance>,
 ) -> Result<(), SqliteClientError> {
```
daira, line 388:
```suggestion
) -> Result<(), SqliteClientError> {
    // TODO (#issue number): distinguish between pending change and non-change.

```
(if this is not intended to be fixed in this PR).

zcash_primitives/CHANGELOG.md
```diff
@@ -6,11 +6,22 @@ and this library adheres to Rust's notion of
...
+- `zcash_client_backend::AccountBalance::unshielded`
 ## [0.19.0] - 2024-10-02
 
 ### Changed
 - Migrated to `zcash_address 0.6`.
+- Implemented changes to refactor AccountBalance to use Balance for transparent funds
+(see issue #1411). `AccountBalance` now has an `unshielded` value that uses Balance.
+The current implementation maintains retrocompatibility with the `unshielded` value
+represeted with a `NonNegativeAmount`. There are values that are pending to be 
+implemented such as change tracking.
```
daira, line 23:
```suggestion
- Refactored `AccountBalance` to use `Balance` for transparent funds (issue #1411).
  `AccountBalance` now has an `unshielded` value that uses `Balance`. This does not
  currently distinguish between pending change and non-change; the pending value
  is all counted as non-change.
```
Is the intention to fix the latter issue in this PR?

daira (changes requested):
Changes required.

zcash_primitives/CHANGELOG.md
daira, line 14: `zcash_client_backend::AccountBalance::unshielded` is removed, not deprecated.

Option 1 from above:
```suggestion
### Removed
- `zcash_client_backend::AccountBalance::unshielded`. Instead use
  `account_balance.unshielded_balance().total()`.

### Deprecated
- `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
  `AccountBalance::with_unshielded_balance_mut` with a closure that calls
  the appropriate `add_*_value` method(s) of `Balance` on its argument.
  Note that the appropriate method(s) depend on whether the funds are
  spendable, pending change, or pending non-change (previously, only the
  total unshielded value was tracked). The deprecated implementation now
  assumes that they are pending non-change, which may not be correct, and
  so all uses of this method should be checked and replaced.
```

Option 2 from above:
```suggestion
### Removed
- `zcash_client_backend::AccountBalance::unshielded`. Instead use
  `account_balance.unshielded_balance().total()`.
- `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
  `AccountBalance::with_unshielded_balance_mut` with a closure that calls
  the appropriate `add_*_value` method(s) of `Balance` on its argument.
  Note that the appropriate method(s) depend on whether the funds are
  spendable, pending change, or pending non-change (previously, only the
  total unshielded value was tracked).
```

pacu:
Thank you for the thorough review! I'll be addressing these comments shortly.

zcash_client_backend/src/data_api.rs
pacu: I'm good either way. I decided to maintain functionality (proved by passing the existing tests) because it's what a non maintainer contributor should do. But I can proceed with the option you folks consider more appropriate.

zcash_client_backend/src/data_api.rs
pacu: Thanks for catching this. I believe the `add_pending_spendable_value` does check a total 

```
/// Adds the specified value to the pending spendable total, checking for overflow.
    pub fn add_pending_spendable_value(
        &mut self,
        value: NonNegativeAmount,
    ) -> Result<(), BalanceError> {
        self.check_total_adding(value)?;
        self.value_pending_spendability = (self.value_pending_spendability + value).unwrap();
        Ok(())
    }
```

I think using the `with_unshielded_balance_mut()` helper does this check twice. Is that ok?

zcash_primitives/CHANGELOG.md
pacu: Yes! I'm figuring out how to do it 😅. Also I will add more tests.

zcash_client_backend/src/data_api.rs
daira: It's the invariant of `AccountBalance` that isn't checked. It's not implied by the one for `Balance` for any of `self.{sapling,orchard,unshielded}_balance` individually.

zcash_client_backend/src/data_api.rs
daira: I would prefer to delete it. @str4d or @nuttycom, what do you think?

zcash_primitives/CHANGELOG.md
pacu: I'm settling for option two (if possible) since it would be a better contribution overall

zcash_primitives/CHANGELOG.md
```diff
@@ -6,11 +6,29 @@ and this library adheres to Rust's notion of
...
+  `account_balance.unshielded_balance().total()`.
+  
+### Removed
+- `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
+  `AccountBalance::with_unshielded_balance_mut` with a closure that calls
+  the appropriate `add_*_value` method(s) of `Balance` on its argument.
+  Note that the appropriate method(s) depend on whether the funds are
+  spendable, pending change, or pending non-change (previously, only the
+  total unshielded value was tracked).
 ## [0.19.0] - 2024-10-02
```
daira, line 23:
```suggestion

## [0.19.0] - 2024-10-02
```
(There needs to be a blank line between sections in Markdown.)

zcash_primitives/CHANGELOG.md
```diff
@@ -6,11 +6,29 @@ and this library adheres to Rust's notion of
...
 
 ### Changed
 - Migrated to `zcash_address 0.6`.
+- Refactored `AccountBalance` to use `Balance` for transparent funds (issue #1411).
+  `AccountBalance` now has an `unshielded_balance()` that uses `Balance`. This does
+  not currently distinguish between pending change and non-change; the pending value
+  is all counted as non-change.
+
```
daira, line 31:
```suggestion
```
(just a nit)

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -105,6 +105,22 @@ where
         st.wallet().get_transparent_balances(account_id, height_2),
         Ok(h) if h.get(taddr) == Some(&value)
     );
+    // TODO: Find our why this does not work.
+    
+    // let unshielded_balance = st.wallet().get_wallet_summary(0)
```
pacu, line 110: @nuttycom @str4d  I naively thought this was a straightforward test change. But it results that `get_wallet_summary` return `None` instead while given the test's state,  it should have a balance. 

Bug or feature? 😅

zcash_primitives/CHANGELOG.md
pacu: Update: Given how `zcash_client_backend` and the sqlite crate accounts for transparent funds, this is a bigger issue that has dependencies that are out of scope for issue #1411. I created #1592 which describes the follow-up work and (some of) the related issues that are already tracked by the [Zcashd deprecation DAG](https://zcash.github.io/developers/zcash-zcashd-deprecation-dag)

zcash_client_backend/src/data_api.rs
pacu: I opted to remove it. :)

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -383,16 +383,59 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+
+    let mut stmt_account_unconfirmed_balances = conn.prepare(
+        "SELECT u.account_id, SUM(u.value_zat)
+         FROM transparent_received_outputs u
+         JOIN transactions t
+         ON t.id_tx = u.transaction_id
+         -- the transaction that created the output is mined with not enough confirmations or is definitely unexpired
+         WHERE (
+            t.mined_height < :mempool_height
+            AND (:mempool_height - t.mined_height + 1) < :min_confirmations -- tx is mined but not confirmed
```
daira, line 438: There seems to be an off-by-one error here. Suppose we've just mined a tx at height $x$, and the mempool height is $x+1$. Then the tx has 1 confirmation. This condition should then be true iff `min_confirmations > 1`, or more generally if `:min_confirmations > :mempool_height - t.mined_height` (equivalently, `:mempool_height - t.mined_height < :min_confirmations`).
```suggestion
            AND :mempool_height - t.mined_height < :min_confirmations -- tx is mined but not confirmed
```
Note that this will always exclude unmined UTXOs even if `min_confirmations == 0`, and I think that is intentional, but it should be documented.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -383,16 +383,59 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+    let mut stmt_account_spendable_balances = conn.prepare(
         "SELECT u.account_id, SUM(u.value_zat)
          FROM transparent_received_outputs u
          JOIN transactions t
          ON t.id_tx = u.transaction_id
-         -- the transaction that created the output is mined or is definitely unexpired
+         -- the transaction that created the output is mined and with enough confirmations
          WHERE (
             t.mined_height < :mempool_height -- tx is mined
+            AND (:mempool_height - t.mined_height + 1) >= :min_confirmations -- has at least min_confirmations
```
daira, line 398:
```suggestion
            AND :mempool_height - t.mined_height >= :min_confirmations -- has at least min_confirmations
```
Same as below.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1532,7 +1532,12 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
     drop(sapling_trace);
 
     #[cfg(feature = "transparent-inputs")]
-    transparent::add_transparent_account_balances(tx, chain_tip_height + 1, &mut account_balances)?;
+    transparent::add_transparent_account_balances(
+        tx,
+        chain_tip_height + 1,
+        min_confirmations,
+        &mut account_balances,
+    )?;
```
daira, line 1540: The doc comment for this method currently says:
```rust
/// `min_confirmations` can be 0, but that case is currently treated identically to
/// `min_confirmations == 1` for shielded notes. This behaviour may change in the future.
```
I think it should say:
```rust
/// `min_confirmations` can be 0, but that case is currently treated identically to
/// `min_confirmations == 1` for both shielded and transparent TXOs. This behaviour
/// may change in the future.
```

zcash_primitives/CHANGELOG.md
```diff
@@ -9,6 +9,26 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.77.0.
+- Refactored `AccountBalance` to use `Balance` for transparent funds (issue #1411).
+  `AccountBalance` now has an `unshielded_balance()` function that returns `Balance`. 
+  This does not currently distinguish between pending change and non-change; the 
+  pending value is all counted as non-change. Ability to distinguish such funds is 
+  tracked by issue #1592
```
daira, line 16:
```suggestion
  tracked by issue #1592.
```

daira (changes requested):
Changes needed.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -383,16 +383,59 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
 pub(crate) fn add_transparent_account_balances(
     conn: &rusqlite::Connection,
     mempool_height: BlockHeight,
+    min_confirmations: u32,
```
nuttycom, line 386: In the past, we have specifically chosen to use a minconf of 0 for transparent balances, because the mobile wallets don't permit spending transparent funds except to shield them. Given that the funds aren't available to the wallet until they've been shielded, we allow shielding at minconf 0 because it's the _shielding_ transaction whose number of confirmations we're concerned with.

pacu: Thanks for the context. I'm not clear what is the recommendation here. Let me put in words what I understand. from your comment. 

I assume that you are pointing out that, at the moment, `min_confirmations` is always **zero** for transparent funds. Given transparent funds are currently being shielded before they can be spent and shielding transactions use zero confirmations for the t -> 🛡️ part and non-zero `min_confirmations` for spending shielded funds. This would mean that the parameter `min_confirmations` for this particular functions is not correct in the current context and assumptions.

daira: Yes, `min_confirmations` is not being used for transparent TXOs, and that's a bug that this PR fixes. But @nuttycom is incorrect in asserting that "we allow shielding at minconf 0 [in the mobile wallets]". We do not, for two reasons:

1. `minimumShieldingConfirmations` is set to 1 in the SDKs, not 0.
2. None of the balances include unmined UTXOs. That also applies to `unshielded_balance` before this PR. The reason is this WHERE clause condition (line 395 before / line 397 after):
```sql
            t.mined_height < :mempool_height -- tx is mined
```
$\hspace{1.6em}$ And so we do not see unmined UTXOs as available to shield.

If we want to change this then we should do so separately.

daira (commented):
The changes to `AccountBalance` should be documented in the CHANGELOG for `zcash_client_backend`, not `zcash_primitives`.

Also, this fix to `get_wallet_summary` should be documented in the CHANGELOG for `zcash_client_sqlite`:
```markdown
### Fixed
- `zcash_client_sqlite::WalletDb::get_wallet_summary` has been fixed to take account of `min_confirmations` for transparent balances. Note that `min_confirmations == 0` is still treated the same as `min_confirmations == 1` for both shielded and transparent TXOs.
```

zcash_primitives/CHANGELOG.md
```diff
@@ -22,11 +23,18 @@ and this library adheres to Rust's notion of
   - `zip317::FeeRule::non_standard` has been moved behind the `non-standard-fees`
     feature flag. Using a non-standard fee may result in a transaction that cannot
     be mined on the current `Zcash` network.
+- Refactored `AccountBalance` to use `Balance` for transparent funds (issue #1411).
+  `AccountBalance` now has an `unshielded_balance()` function that returns `Balance`. 
+  This does not currently distinguish between pending change and non-change; the 
+  pending value is all counted as non-change. Ability to distinguish such funds is 
+  tracked by issue #1592
```
daira, line 30: This part is not part of the `WalletRead` API; it's a bug/limitation in `zcash_client_sqlite`'s implementation of that API for `WalletDb`. All other changes to this file should instead be to `zcash_client_backend/CHANGELOG.md`.

zcash_client_backend/src/data_api.rs
daira, line 1: I notice that the documentation for `WalletRead::get_wallet_summary` is:
```rust
    /// Returns the wallet balances and sync status for an account given the specified minimum
    /// number of confirmations, or `Ok(None)` if the wallet has no balance data available.
```

"for an account" is incorrect. It should be "for all accounts known to the wallet". `get_wallet_summary` does not take an `account` argument, and it returns `WalletSummary` which has a `HashMap` of `AccountBalance`s for all known accounts.

zcash_client_sqlite/src/wallet/transparent.rs
pacu: I think I may have addressed this as the test `transparent_balance_spendability` passes. I still need to make the test code non-repetitive (extract the `check_balance` function).

I enforced very obvious to the eye checks for zero and one confirmations with these kind of checks

```
    check_balance(&st, 0, &Balance::ZERO);
    // we currently treat min_confirmation the same regardless they are 0 or 1
    check_balance(&st, 1, &Balance::ZERO);
 ```
 
 and also verified that the balances hold when it comes to moving the min_confirmations up by one or mining an extra block.

File: zcash_primitives/CHANGELOG.md (hunk 23 above)
Comment (pacu) [reply zcash_primitives/CHANGELOG.md:30]: Oh, that's right. Thank you for pointing it out.

File: zcash_client_backend/src/data_api/testing/transparent.rs [hunk 24]
```diff
@@ -146,7 +149,8 @@ where
         let balance = summary.account_balances().get(&account.id()).unwrap();
         // TODO: in the future, we will distinguish between available and total
         // balance according to `min_confirmations`
-        assert_eq!(balance.unshielded(), expected);
+        // assert_eq!(balance.unshielded(), expected);
+        assert_eq!(balance.unshielded_balance(), expected);
```
daira, line 153:
```suggestion
        #[allow(deprecated)]
        assert_eq!(balance.unshielded_balance(), expected);
```

zcash_client_sqlite/src/wallet/transparent.rs
pacu: I think you were correct indeed and the `transparent_balance_spendability` test should exercise that case.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -146,7 +149,8 @@ where
         let balance = summary.account_balances().get(&account.id()).unwrap();
         // TODO: in the future, we will distinguish between available and total
         // balance according to `min_confirmations`
-        assert_eq!(balance.unshielded(), expected);
+        // assert_eq!(balance.unshielded(), expected);
```
daira, line 152:
```suggestion
        assert_eq!(balance.unshielded().total(), expected);
        #[allow(deprecated)]
```

zcash_client_backend/src/data_api/testing/transparent.rs
daira, line 153:
```suggestion
        assert_eq!(balance.unshielded().total(), expected);
        #[allow(deprecated)]
        assert_eq!(balance.unshielded_balance(), expected);
```

zcash_client_backend/src/data_api/testing/transparent.rs
daira, line 152:
```suggestion
        #[allow(deprecated)]
        assert_eq!(balance.unshielded(), expected.total());
```

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -146,7 +149,8 @@ where
         let balance = summary.account_balances().get(&account.id()).unwrap();
         // TODO: in the future, we will distinguish between available and total
         // balance according to `min_confirmations`
```
daira, line 151:
```suggestion
```

zcash_client_backend/src/data_api/testing/transparent.rs
daira, line 152:
```suggestion
        #[allow(deprecated)]
        assert_eq!(balance.unshielded(), expected.total());
```

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -175,8 +190,10 @@ where
     // and total balance, we should perform additional checks against available balance;
     // we use minconf 0 here because all transparent funds are considered shieldable,
     // irrespective of confirmation depth.
```
daira, line 192: Please update this comment. The tests now do distinguish between available and total (or pending) balances. They still don't distinguish between pending transparent change and pending transparent non-change.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -243,5 +272,128 @@ where
...
+    let not_our_key = ExtendedSpendingKey::master(&[]).to_diversifiable_full_viewing_key();
+    let not_our_value = NonNegativeAmount::const_from_u64(10000);
+    let (start_height, _, _) =
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    for _ in 1..10 {
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    }
+    st.scan_cached_blocks(start_height, 10);
+
+    // TODO: abstract this function so it's not duplicated among different tests.
```
daira, line 310: Since this PR is going from one to two uses of this code (i.e. it's introducing the duplication), and since it looks straightforward to move it out to a private function because it only closes over `account` and `taddr`, please do so now.

zcash_client_backend/src/data_api/testing/transparent.rs
pacu: the allow directive would be ignored because of the assertion macro so I did this in two separate LoCs so the warning went away

daira (dismissed):
utACK modulo the remaining changes I've suggested.

zcash_client_backend/src/data_api/testing/transparent.rs
pacu: the amount of generics that are involved in that apparently innocent function... oh dear FSM! 😅

I'll do my best!

zcash_client_backend/CHANGELOG.md
```diff
@@ -79,6 +82,10 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::data_api`:
+  - `AccountBalance::unshielded`. `AccountBalance` no longer provide the `unshielded` value returning a `NonNegativeAmount` for all transparent funds. Use`unshieded_balance` instead which provides `Balance` value.
```
daira, line 85:
```suggestion
  - `AccountBalance::unshielded`. `AccountBalance` no longer provides the `unshielded`
     method returning a `NonNegativeAmount` for the total of transparent funds.
     Instead use `unshielded_balance` which provides a `Balance` value.
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -79,6 +82,10 @@ and this library adheres to Rust's notion of
...
+  - `AccountBalance::add_unshielded_value` was removed with the `unshielded` value as
+  described above. Use the appropriate `add_*_value` method(s) of `Balance` on its argument
+  instead.
```
daira, line 88: These two lines should be indented another two spaces. (Markdown is picky about indentation.)

zcash_client_backend/CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ and this library adheres to Rust's notion of
 - MSRV is now 1.77.0.
 - Migrated to `arti-client 0.23`.
 - `zcash_client_backend::data_api`:
+  - `AccountBalance` was refactored to use `Balance` for transparent funds (issue #1411).
+  `AccountBalance` now has an `unshielded_balance()` that uses `Balance` and replaces the
+  (now deleted) `unshielded` non-negative amount.
```
daira, line 33: Indent by an extra 2 spaces.

zcash_primitives/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this library adheres to Rust's notion of
 - A new feature flag, `non-standard-fees`, has been added. This flag is now
   required in order to make use of any types or methods that enable non-standard
   fee calculation.
+- `zcash_client_backend::AccountBalance::with_unshielded_balance_mut`
```
daira, line 14: This is still in the wrong changelog.

zcash_primitives/CHANGELOG.md
```diff
@@ -38,6 +39,12 @@ and this library adheres to Rust's notion of
     ZIP 317 fee, use `zip317::FeeRule::standard()`. To preserve the current
     behaviour, use `fixed::FeeRule::non_standard(zip317::MINIMUM_FEE)`,
     but note that this is likely to result in transactions that cannot be mined.
+- `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
+  `AccountBalance::with_unshielded_balance_mut` with a closure that calls
+  the appropriate `add_*_value` method(s) of `Balance` on its argument.
+  Note that the appropriate method(s) depend on whether the funds are
+  spendable, pending change, or pending non-change (previously, only the
+  total unshielded value was tracked).
```
daira, line 47: This is still in the wrong changelog.

daira (approved):
utACK with minor comments on the changelogs.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -383,16 +383,59 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+    let mut stmt_account_spendable_balances = conn.prepare(
         "SELECT u.account_id, SUM(u.value_zat)
          FROM transparent_received_outputs u
          JOIN transactions t
          ON t.id_tx = u.transaction_id
-         -- the transaction that created the output is mined or is definitely unexpired
+         -- the transaction that created the output is mined and with enough confirmations
          WHERE (
             t.mined_height < :mempool_height -- tx is mined
+            AND :mempool_height - t.mined_height >= :min_confirmations -- has at least min_confirmations
```
nuttycom, line 398: This looks wrong to me. `mempool_height` is typically `chain_tip_height + 1`. `mined_height` is at most the chain tip height, or null, so changing the query in this fashion is introducing a footgun for zero-conf shielding.

Also, `:mempool_height - t.mined_height` will be `NULL` when `mined_height IS NULL`, which then will always cause this condition to fail.

Instead of using `:mempool_height` directly, compute a `balance_height` outside of the query from the mempool height and `min_confirmations`, and then query against that height here, accounting for `mined_height` being allowed to be null in the case that `min_confirmations == 0`.

pacu: > This looks wrong to me. `mempool_height` is typically `chain_tip_height + 1`. `mined_height` is at most the chain tip height, or null, so changing the query in this fashion is introducing a footgun for zero-conf shielding.

Is Zero-conf shielding something that should be address here? I understood it was not. 

> Also, `:mempool_height - t.mined_height` will be `NULL` when `mined_height IS NULL`, which then will always cause this condition to fail.

In this particular query a pre-condition is that the TX is mined so I guess desirable that the condition is evaluated to `FALSE`. 

> 
> Instead of using `:mempool_height` directly, compute a `balance_height` outside of the query from the mempool height and `min_confirmations`, and then query against that height here, accounting for `mined_height` being allowed to be null in the case that `min_confirmations == 0`.
This sounds like a good idea.

zcash_client_backend/src/data_api.rs
```diff
@@ -289,22 +285,36 @@ impl AccountBalance {
...
+    /// Returns the [`Balance`] of unshielded funds in the account.
+    pub fn unshielded_balance(&self) -> &Balance {
+        &self.unshielded_balance
     }
 
-    /// Adds the specified value to the unshielded total, checking for overflow of
-    /// the total account balance.
-    pub fn add_unshielded_value(&mut self, value: NonNegativeAmount) -> Result<(), BalanceError> {
-        self.unshielded = (self.unshielded + value).ok_or(BalanceError::Overflow)?;
+    /// Provides a `mutable reference to the [`Balance`] of transparent funds in the account
```
daira, line 300:
```suggestion
    /// Provides a mutable reference to the [`Balance`] of transparent funds in the account
```

zcash_client_sqlite/src/wallet/transparent.rs
daira: > Is Zero-conf shielding something that should be address here? I understood it was not.

No, we'll need another PR for that. The current code appears to work as documented: `min_confirmations == 0` behaves the same as `min_confirmations == 1`. I don't think we should complicate this PR to make zero-conf shielding work; it's obvious that this code will need to change.

daira (dismissed):
re-utACK (last commit was doc-only)

zcash_client_sqlite/src/wallet/transparent.rs
daira: Btw @nuttycom did you see this? https://github.com/zcash/librustzcash/issues/1413#issuecomment-2439717322

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -243,5 +279,90 @@ where
...
-    check_balance(&st, 0, value);
+    check_balance::<DSF>(&st, &account, taddr, 0, &zero_or_one_conf_value);
+    // we currently treat min_confirmation the same regardless they are 0 or 1
+    check_balance::<DSF>(&st, &account, taddr, 1, &zero_or_one_conf_value);
+}
+
+/// This test attempts to verify that transparent funds spendability is
+/// accounted for properly given the different minimum confirmations values
+/// that can be set when querying for balances.
+pub fn transparent_balance_spendability<DSF>(dsf: DSF, cache: impl TestCache)
```
daira, line 290: This does not seem to be called, and it should be.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -16,6 +16,16 @@ and this library adheres to Rust's notion of
 - Migrated from `schemer` to our fork `schemerz`.
 - Migrated to `rusqlite 0.32`.
 
+### Fixed
+- `zcash_client_sqlite::WalletDb`'s implementation of
+  `zcash_client_backend::data_api::WalletRead::get_wallet_summary` has been
+  fixed to take account of `min_confirmations` for transparent balances.
```
daira, line 22:
```suggestion
  fixed to take account of `min_confirmations` for transparent balances.
  (Previously, it would treat transparent balances as though `min_confirmations`
  were `1` even if it was set to a higher value.)
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -838,4 +884,12 @@ mod tests {
             BlockCache::new(),
         );
     }
+
+    #[test]
+    fn transparent_balance_spendability() {
+        zcash_client_backend::data_api::testing::transparent::transparent_balance_across_shielding(
```
daira, line 890:
```suggestion
        zcash_client_backend::data_api::testing::transparent::transparent_balance_spendability(
```

daira: Merging this so that CI runs again.

zcash_client_sqlite/src/wallet/transparent.rs
daira: It wasn't correct to conclude that `transparent_balance_spendability` passes, because the wrong test was being run. We'll see once CI has finished.

daira: It doesn't in fact pass:
```
---- wallet::transparent::tests::transparent_balance_spendability stdout ----
thread 'wallet::transparent::tests::transparent_balance_spendability' panicked at /home/runner/work/librustzcash/librustzcash/zcash_client_backend/src/data_api/testing/transparent.rs:56:5:
assertion `left == right` failed
  left: Some(Zatoshis(100000))
 right: Some(Zatoshis(0))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    wallet::transparent::tests::transparent_balance_spendability

test result: FAILED. 98 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 41.27s
```

daira: Filed #1602 for the missing traceback.

daira:
```
---- wallet::transparent::tests::transparent_balance_spendability stdout ----
thread 'wallet::transparent::tests::transparent_balance_spendability' panicked at /media/ssd/home/daira/research/librustzcash/zcash_client_backend/src/data_api/testing/transparent.rs:56:5:
assertion `left == right` failed
  left: Some(Zatoshis(100000))
 right: Some(Zatoshis(0))
stack backtrace:
   0: rust_begin_unwind
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/panicking.rs:72:14
   2: core::panicking::assert_failed_inner
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/panicking.rs:342:17
   3: core::panicking::assert_failed
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/panicking.rs:297:5
   4: zcash_client_backend::data_api::testing::transparent::check_balance
   5: zcash_client_backend::data_api::testing::transparent::transparent_balance_spendability
             at /media/ssd/home/daira/research/librustzcash/zcash_client_backend/src/data_api/testing/transparent.rs:361:5
   6: zcash_client_sqlite::wallet::transparent::tests::transparent_balance_spendability
             at ./src/wallet/transparent.rs:890:9
   7: zcash_client_sqlite::wallet::transparent::tests::transparent_balance_spendability::{{closure}}
             at ./src/wallet/transparent.rs:889:42
   8: core::ops::function::FnOnce::call_once
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -243,5 +279,90 @@ where
...
+
+    // now if we increase the number of confirmations our spendable balance should
+    // be zero and the total balance equal to `value`
+    let mut not_confirmed_yet_value = Balance::ZERO;
+
+    not_confirmed_yet_value
+        .add_pending_spendable_value(value)
+        .unwrap();
+
+    check_balance::<DSF>(&st, &account, taddr, 2, &not_confirmed_yet_value);
```
daira, line 361: In the current test failure, the spendable value at the mempool height is being calculated at 100000 zats here, but the test expects it to be 0 (because the call on line 358 is to `add_pending_spendable_value`, not `add_spendable_value`). The test looks correct to me: the funds should indeed not be spendable at two confirmations.

pacu: oh. I'll look into this!

pacu: This is interesting. I ran `cargo test --release --workspace --features ' bundled-prover download-params lightwalletd-tonic sync temporary-zcashd transparent-inputs unstable unstable-serialization unstable-spanning-tree  test-dependencies '` in MacOs and the failing test does not appear to be ran.

pacu: uuhmmm Interesting

Local on MacOS
```
test wallet::transparent::tests::transparent_balance_across_shielding ... ok
test wallet::transparent::tests::transparent_balance_spendability ... ok
```

Test on Linux CI
```
test wallet::transparent::tests::transparent_balance_across_shielding ... ok
test wallet::transparent::tests::transparent_balance_spendability ... FAILED
```

zcash_client_sqlite/src/wallet/transparent.rs
pacu: There were a couple of things here:
1. check_balance was not properly using `min_confirmations`. Anyway the test still failed. then,
2. the `transparent_balance_spendability` test was not advancing and scanning the chain so the balance was being reported properly by the backend (from it's PoV) and it was the test that assumed otherwise. I added the blocks and scanned them and the test passed.

zcash_client_backend/src/data_api/testing/transparent.rs
pacu: fixed that ... I'm a prisoner of my own copypasta

pacu:
I think I chewed a doc commit with this rebase from main and fix. I'll review all the changes and then ask for review

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -13,8 +13,78 @@ use assert_matches::assert_matches;
...
+            .sum::<Option<NonNegativeAmount>>(),
+        Some(expected.spendable_value()),
+    );
+
+    // we currently treat min_confirmations the same regardless they are 0 (zero confirmations) 
+    // or 1 (one block confirmation). We will check if this assumption holds until it's no 
+    // longer made. If zero and one [`min_confirmations`] are treated differently in the future,
+    // this check should then be removed.
+    if min_confirmations == 0 || min_confirmations == 1 {
+        let min_confirmations: u32 = if min_confirmations == 0 {
```
daira, line 71: I would just delete this and make the call below using `1 - min_confirmations`.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -192,7 +226,14 @@ where
...
     // The wallet should detect the balance as available
-    check_balance(&st, 0, value);
+    let mut zero_or_one_conf_value = Balance::ZERO;
+
+    // add the spendable value to the expected balance
+    zero_or_one_conf_value.add_spendable_value(value).unwrap();
+
+    check_balance::<DSF>(&st, &account, taddr, 0, &zero_or_one_conf_value);
+    // we currently treat min_confirmation the same regardless they are 0 or 1
+    check_balance::<DSF>(&st, &account, taddr, 1, &zero_or_one_conf_value);
```
daira, line 236: It seems redundant to both make two calls to `check_balance` here (and below), and to also have the `if min_confirmations == 0 || min_confirmations == 1 { ... }` code in `check_balance`. (Even though they are currently not quite equivalent.)

daira (dismissed):
utACK. (The complaint about redundancy in the test code is non-blocking.)

zcash_client_backend/src/data_api/testing/transparent.rs
pacu: Thanks! I was just trying stuff and got a bit verbose 🙏😅

zcash_client_backend/src/data_api/testing/transparent.rs
pacu: makes sense. I added some docs to the `check_balance` function instead so callers are warned that the assumption of the zero and one `min_confirmations` being treated equally will be checked

pacu:
oh no! a wild conflicting commit appeared!

pacu:
Squashed commits and rebased to main afterwards.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -13,8 +13,75 @@ use assert_matches::assert_matches;
 use sapling::zip32::ExtendedSpendingKey;
 use zcash_primitives::{
     block::BlockHash,
+    legacy::TransparentAddress,
     transaction::components::{amount::NonNegativeAmount, OutPoint, TxOut},
 };
+use zcash_protocol::local_consensus::LocalNetwork;
+
+use super::TestAccount;
+/// checks whether the transparent balance of the given test [`account`] is as [`expected`]
```
daira, line 22:
```suggestion

/// Checks whether the transparent balance of the given test [`account`] is as [`expected`]
```

daira (approved):
utACK with minor non-blocking suggestion.

daira (approved):
utACK (+ doc-only self-ACK for 19dedb75b3adf55e39c93595b042673df56758a6)

zcash_client_backend/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Changed
+- `zcash_client_backend::data_api::AccountBalance`:
+  - Refactored to use `Balance` for transparent funds (issue #1411).
+    `AccountBalance` now has an `unshielded_balance()` that uses `Balance` and replaces the
+    (now deleted) `unshielded` non-negative amount.
```
daira, line 14: `unshielded` wasn't public as a field (only as a method). So this would be better expressed as "and replaces the tracking of only total unshielded balance.".

zcash_client_backend/CHANGELOG.md
```diff
@@ -105,6 +109,15 @@ and this library adheres to Rust's notion of
 
 ### Removed
 - `zcash_client_backend::data_api`:
+  - `AccountBalance::unshielded`. `AccountBalance` no longer provides the `unshielded`
+    method returning a `NonNegativeAmount` for the total of transparent funds.
```
daira, line 113: This is not removed, only deprecated. Move it to the Deprecated section with the wording: "`AccountBalance::unshielded`. Instead use `unshielded_balance` which provides a `Balance` value. Its `total()` method can be used to obtain the total of transparent funds."

zcash_client_backend/CHANGELOG.md
```diff
@@ -105,6 +109,15 @@ and this library adheres to Rust's notion of
...
+    Instead use `unshielded_balance` which provides a `Balance` value.
+  - `zcash_client_backend::AccountBalance::add_unshielded_value`. Instead use
+    `AccountBalance::with_unshielded_balance_mut` with a closure that calls
+    the appropriate `add_*_value` method(s) of `Balance` on its argument.
+    Note that the appropriate method(s) depend on whether the funds are
+    spendable, pending change, or pending non-change (previously, only the
+    total unshielded value was tracked).
```
daira, line 120: This is in the section for the wrong release. (Also, the `unshielded()` method is deprecated, not removed.)
```suggestion
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
```
daira, line 9: (Will move this below "Changed" in a force-push.)

```suggestion

### Deprecated
- `AccountBalance::unshielded`. Instead use `unshielded_balance` which provides a `Balance`
  value. Its `total()` method can be used to obtain the total of transparent funds.

```

daira (approved):
utACK bc637678ec383c24cd3defb9a80510fc5521ce56 + doc-only self-ACK for 3f802121e7b8d7a8cf607ab1fcd4947024c91cc4
([changes in the force-push](https://github.com/zcash/librustzcash/compare/886f72f4278cf9a388dd5f70b56e18c7a1f2a486..3f802121e7b8d7a8cf607ab1fcd4947024c91cc4))

daira (approved):
doc-only self-ACK for bd45812d42c77931f4d8f09bba7c966b8216af34

PR #1573, opened 2024-10-14
zcash_keys: Add a test for parsing a problematic mainnet address.

nuttycom:


pacu:
shall this be ported to the GoLang version?

nuttycom:
The root cause of the failure that this was attempting to track down appears to have potentially have been use of an incorrect network constant.

PR #1575, opened 2024-10-15
zcash_client_sqlite: Migrate from `schemer` to `schemerz`

str4d:

PR #1578, opened 2024-10-16
Update the `rusqlite` and `arti-client` crates

str4d:
This was previously blocked on `schemer`, because we cannot have two different versions of the `libsqlite3-sys` crate in our dependency tree. After switching to `schemerz` in #1575, we can now update these in sync.


Cargo.toml
```diff
@@ -123,9 +123,9 @@ subtle = "2.2.3"
 # - Warning: One of the downstream consumers requires that SQLite be available through
 #   CocoaPods, due to being bound to React Native. We need to ensure that the SQLite
 #   version required for `rusqlite` is a version that is available through CocoaPods.
```
str4d, line 125: I checked and I believe our constraints here are met for all `rusqlite` versions existing in this PR:

rusqlite | Bundled SQLite | Matching CocoaPods sqlite3 tag
-----|--------|---
0.29 | 3.41.2 | :white_check_mark:
0.30 | 3.44.0 | :white_check_mark:
0.31 | 3.45.1 | :white_check_mark:
0.32 | 3.46.0 | :white_check_mark:

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/f15f8a8683b83fcb0be432ab7e6b6bdfbd12cc23..88d1736898eb39015e52b8074d550bb877ca3688) to back out `rusqlite 0.32` from this PR. It requires MSRV 1.77, which I think would be reasonable, but not in this PR (I'd make a separate PR for that, once we've determined that getting as far as `rusqlite 0.31` is fine).

str4d:
`cargo vet` will fail because I haven't updated the audits for the dependency changes. I'll force-push with those changes later.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/88d1736898eb39015e52b8074d550bb877ca3688..8ff0b1a26318dc42bae7de147c9ad3eb3ac6c154) to fix merge conflicts, and fixed lints.

nuttycom (approved):
utACK 8ff0b1a26318dc42bae7de147c9ad3eb3ac6c154 - it's really nice to see all that reduction in duplicated dependencies!

daira (commented):
Love to see those deduplications of crypto crates <3. Post-hoc ACK.

PR #1579, opened 2024-10-17
zcash_client_backend: Implement note management via change splitting.

nuttycom:
Best reviewed commit-by-commit, hiding whitespace changes.

### DIFF
--- a/components/zcash_protocol/src/value.rs
+++ b/components/zcash_protocol/src/value.rs
@@ -1,6 +1,7 @@
 use std::convert::{Infallible, TryFrom};
 use std::error;
 use std::iter::Sum;
+use std::num::NonZeroU64;
 use std::ops::{Add, Mul, Neg, Sub};
 
 use memuse::DynamicUsage;
@@ -229,6 +230,24 @@ impl Mul<usize> for ZatBalance {
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Eq, Ord)]
 pub struct Zatoshis(u64);
 
+/// A struct that provides both the quotient and remainder of a division operation.
+pub struct QuotRem<A> {
+    quotient: A,
+    remainder: A,
+}
+
+impl<A> QuotRem<A> {
+    /// Returns the quotient portion of the value.
+    pub fn quotient(&self) -> &A {
+        &self.quotient
+    }
+
+    /// Returns the remainder portion of the value.
+    pub fn remainder(&self) -> &A {
+        &self.remainder
+    }
+}
+
 impl Zatoshis {
     /// Returns the identity `Zatoshis`
     pub const ZERO: Self = Zatoshis(0);
@@ -297,6 +316,15 @@ impl Zatoshis {
     /// Returns whether or not this `Zatoshis` is positive.
     pub fn is_positive(&self) -> bool {
         self > &Zatoshis::ZERO
+    }
+
+    /// Divides this `Zatoshis` value by the given divisor and returns the quotient and remainder.
+    pub fn div_with_remainder(&self, divisor: NonZeroU64) -> QuotRem<Zatoshis> {
+        let divisor = u64::from(divisor);
+        QuotRem {
+            quotient: Zatoshis(self.0 / divisor),
+            remainder: Zatoshis(self.0 % divisor),
+        }
     }
 }
 
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -13,6 +13,8 @@ and this library adheres to Rust's notion of
   - `WalletSummary::progress`
   - `WalletMeta`
   - `impl Default for wallet::input_selection::GreedyInputSelector`
+- `zcash_client_backend::fees::SplitPolicy`
+- `zcash_client_backend::fees::zip317::MultiOutputChangeStrategy`
 
 ### Changed
 - `zcash_client_backend::data_api`:
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -11,6 +11,11 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::data_api`:
   - `Progress`
   - `WalletSummary::progress`
+  - `WalletMeta`
+
+### Changed
+- `zcash_client_backend::data_api`:
+  - `InputSource` has an added method `get_wallet_metadata`
 
 ### Changed
 - MSRV is now 1.77.0.
@@ -28,7 +33,7 @@ and this library adheres to Rust's notion of
   - `GAP_LIMIT`
   - `WalletSummary::recovery_progress`
   - `SpendableNotes::{take_sapling, take_orchard}`
-  - Tests and testing infrastructure have been migrated from the 
+  - Tests and testing infrastructure have been migrated from the
     `zcash_client_sqlite` internal tests to the `testing` module, and have been
     generalized so that they may be used for testing arbitrary implementations
     of the `zcash_client_backend::data_api` interfaces. The following have been
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -12,10 +12,55 @@ and this library adheres to Rust's notion of
   - `Progress`
   - `WalletSummary::progress`
   - `WalletMeta`
+  - `impl Default for wallet::input_selection::GreedyInputSelector`
 
 ### Changed
 - `zcash_client_backend::data_api`:
   - `InputSource` has an added method `get_wallet_metadata`
+  - `error::Error` has additional variant `Error::Change`. This necessitates
+    the addition of two type parameters to the `Error` type,
+    `ChangeErrT` and `NoteRefT`.
+  - The following methods each now take an additional `change_strategy`
+    argument, along with an associated `ChangeT` type parameter:
+    - `zcash_client_backend::data_api::wallet::spend`
+    - `zcash_client_backend::data_api::wallet::propose_transfer`
+    - `zcash_client_backend::data_api::wallet::propose_shielding`. This method
+      also now takes an additional `to_account` argument.
+    - `zcash_client_backend::data_api::wallet::shield_transparent_funds`. This
+      method also now takes an additional `to_account` argument.
+  - `wallet::input_selection::InputSelectionError` now has an additional `Change`
+    variant. This necessitates the addition of two type parameters.
+  - `wallet::input_selection::InputSelector::propose_transaction` takes an
+    additional `change_strategy` argument, along with an associated `ChangeT`
+    type parameter.
+  - The `wallet::input_selection::InputSelector::FeeRule` associated type has
+    been removed. The fee rule is now part of the change strategy passed to
+    `propose_transaction`.
+  - `wallet::input_selection::ShieldingSelector::propose_shielding` takes an
+    additional `change_strategy` argument, along with an associated `ChangeT`
+    type parameter. In addition, it also takes a new `to_account` argument
+    that identifies the destination account for the shielded notes.
+  - The `wallet::input_selection::ShieldingSelector::FeeRule` associated type
+    has been removed. The fee rule is now part of the change strategy passed to
+    `propose_shielding`.
+  - The `Change` variant of `wallet::input_selection::GreedyInputSelectorError`
+    has been removed, along with the additional type parameters it necessitated.
+  - The arguments to `wallet::input_selection::GreedyInputSelector::new` have
+    changed.
+- `zcash_client_backend::fees::ChangeStrategy` has changed. It has two new
+  associated types, `MetaSource` and `WalletMeta`, and its `FeeRule` associated
+  type now has an additional `Clone` bound. In addition, it defines a new
+  `fetch_wallet_meta` method, and the arguments to `compute_balance` have
+  changed.
+- `zcash_client_backend::fees::fixed::SingleOutputChangeStrategy::new`
+  now takes an additional `DustOutputPolicy` argument. It also now carries
+  an additional type parameter.
+- `zcash_client_backend::fees::standard::SingleOutputChangeStrategy::new`
+  now takes an additional `DustOutputPolicy` argument. It also now carries
+  an additional type parameter.
+- `zcash_client_backend::fees::zip317::SingleOutputChangeStrategy::new`
+  now takes an additional `DustOutputPolicy` argument. It also now carries
+  an additional type parameter.
 
 ### Changed
 - MSRV is now 1.77.0.
@@ -25,6 +70,8 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::data_api`:
   - `WalletSummary::scan_progress` and `WalletSummary::recovery_progress` have
     been removed. Use `WalletSummary::progress` instead.
+- `zcash_client_backend::fees`:
+  - `impl From<BalanceError> for ChangeError<...>`
 
 ## [0.14.0] - 2024-10-04
 
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -24,12 +24,12 @@ and this library adheres to Rust's notion of
     `ChangeErrT` and `NoteRefT`.
   - The following methods each now take an additional `change_strategy`
     argument, along with an associated `ChangeT` type parameter:
-    - `zcash_client_backend::data_api::wallet::spend`
-    - `zcash_client_backend::data_api::wallet::propose_transfer`
-    - `zcash_client_backend::data_api::wallet::propose_shielding`. This method
-      also now takes an additional `to_account` argument.
-    - `zcash_client_backend::data_api::wallet::shield_transparent_funds`. This
-      method also now takes an additional `to_account` argument.
+    - `wallet::spend`
+    - `wallet::propose_transfer`
+    - `wallet::propose_shielding`. This method also now takes an additional
+      `to_account` argument.
+    - `wallet::shield_transparent_funds`. This method also now takes an
+      additional `to_account` argument.
   - `wallet::input_selection::InputSelectionError` now has an additional `Change`
     variant. This necessitates the addition of two type parameters.
   - `wallet::input_selection::InputSelector::propose_transaction` takes an
@@ -49,20 +49,16 @@ and this library adheres to Rust's notion of
     has been removed, along with the additional type parameters it necessitated.
   - The arguments to `wallet::input_selection::GreedyInputSelector::new` have
     changed.
-- `zcash_client_backend::fees::ChangeStrategy` has changed. It has two new
-  associated types, `MetaSource` and `WalletMeta`, and its `FeeRule` associated
-  type now has an additional `Clone` bound. In addition, it defines a new
-  `fetch_wallet_meta` method, and the arguments to `compute_balance` have
-  changed.
-- `zcash_client_backend::fees::fixed::SingleOutputChangeStrategy::new`
-  now takes an additional `DustOutputPolicy` argument. It also now carries
-  an additional type parameter.
-- `zcash_client_backend::fees::standard::SingleOutputChangeStrategy::new`
-  now takes an additional `DustOutputPolicy` argument. It also now carries
-  an additional type parameter.
-- `zcash_client_backend::fees::zip317::SingleOutputChangeStrategy::new`
-  now takes an additional `DustOutputPolicy` argument. It also now carries
-  an additional type parameter.
+- `zcash_client_backend::fees`:
+  - `ChangeStrategy` has changed. It has two new associated types, `MetaSource`
+    and `WalletMeta`, and its `FeeRule` associated type now has an additional
+    `Clone` bound. In addition, it defines a new `fetch_wallet_meta` method, and
+    the arguments to `compute_balance` have changed.
+  - The following methods now take an additional `DustOutputPolicy` argument,
+    and carry an additional type parameter:
+    - `fixed::SingleOutputChangeStrategy::new`
+    - `standard::SingleOutputChangeStrategy::new`
+    - `zip317::SingleOutputChangeStrategy::new`
 
 ### Changed
 - MSRV is now 1.77.0.
@@ -72,6 +68,8 @@ and this library adheres to Rust's notion of
 - `zcash_client_backend::data_api`:
   - `WalletSummary::scan_progress` and `WalletSummary::recovery_progress` have
     been removed. Use `WalletSummary::progress` instead.
+  - `testing::input_selector` use explicit `InputSelector` constructors
+    directly instead.
 - `zcash_client_backend::fees`:
   - `impl From<BalanceError> for ChangeError<...>`
 
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -794,7 +794,7 @@ impl<NoteRef> SpendableNotes<NoteRef> {
     }
 }
 
-/// Metadata about the structure of the wallet for a particular account. 
+/// Metadata about the structure of the wallet for a particular account.
 ///
 /// At present this just contains counts of unspent outputs in each pool, but it may be extended in
 /// the future to contain note values or other more detailed information about wallet structure.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -848,12 +848,7 @@ impl WalletMeta {
     /// Returns the total number of unspent shielded notes belonging to the account for which this
     /// was generated.
     pub fn total_note_count(&self) -> usize {
-        #[cfg(feature = "orchard")]
-        let orchard_note_count = self.orchard_note_count;
-        #[cfg(not(feature = "orchard"))]
-        let orchard_note_count = 0;
-
-        self.sapling_note_count + orchard_note_count
+        self.sapling_note_count + self.note_count(ShieldedProtocol::Orchard)
     }
 }
 
@@ -903,8 +898,10 @@ pub trait InputSource {
 
     /// Returns metadata describing the structure of the wallet for the specified account.
     ///
-    /// The returned metadata value must exclude spent notes and unspent notes having value less
-    /// than the specified minimum value or identified in the given exclude list.
+    /// The returned metadata value must exclude:
+    /// - spent notes;
+    /// - unspent notes having value less than the specified minimum value;
+    /// - unspent notes identified in the given `exclude` list.
     fn get_wallet_metadata(
         &self,
         account: Self::AccountId,
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -794,6 +794,69 @@ impl<NoteRef> SpendableNotes<NoteRef> {
     }
 }
 
+/// Metadata about the structure of the wallet for a particular account. 
+///
+/// At present this just contains counts of unspent outputs in each pool, but it may be extended in
+/// the future to contain note values or other more detailed information about wallet structure.
+///
+/// Values of this type are intended to be used in selection of change output values. A value of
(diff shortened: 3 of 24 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -749,53 +787,54 @@ where
         return Err(Error::ProposalNotSupported);
     }
 
-    let (sapling_anchor, sapling_inputs) =
```
nuttycom, line 752: This is best viewed hiding whitespace changes.

components/zcash_protocol/src/value.rs
```diff
@@ -229,6 +230,24 @@ impl Mul<usize> for ZatBalance {
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Eq, Ord)]
 pub struct Zatoshis(u64);
 
+/// A struct that provides both the dividend and remainder of a division operation.
```
daira, line 233: We already have a dependency on num-integer which has https://docs.rs/num-integer/latest/num_integer/fn.div_rem.html

nuttycom: We would have to implement all the `num_integer::Integer` trait methods for `Zatoshis` to take advantage of that; honestly I don't think that's worth it.

daira: Is it worth following the signature of `div_rem` (or `div_floor_mod`) anyway?

nuttycom: A struct is better than a tuple.

nuttycom: Also, dividing a `Zatoshis` value by another `Zatoshis` makes no sense.

daira: Well it does, but the quotient would be dimensionless rather than `Zatoshis`, so yes it's not what we want here.

nuttycom: quotient! I had been trying to remember that word, that's what this struct should use, not dividend.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -117,6 +115,53 @@ where
     Ok(())
 }
 
+pub type ProposeTransferErrT<DbT, CommitmentTreeErrT, InputsT, ChangeT> = Error<
```
str4d, line 118: A short bit of documentation for each of these new type aliases would be useful (even just a single sentence linking to the method for which this error is used).

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e6c1fa308535d99d09e2dab111a1427a95bf52ee..6c95fd8bb0296959ead29dca6d6b1ab5ed4361db) to fix the name of `quotient` in `DivRem`.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -335,41 +348,41 @@ impl<DbT, ChangeT: ChangeStrategy> GreedyInputSelector<DbT, ChangeT> {
     /// attempting to construct a transaction proposal that requires such an output.
     ///
     /// [`EphemeralBalance::Output`]: crate::fees::EphemeralBalance::Output
-    pub fn new(change_strategy: ChangeT, dust_output_policy: DustOutputPolicy) -> Self {
+    pub fn new() -> Self {
```
str4d, line 351: I think this change is missing in the changelog.

str4d (commented):
Flushing comments from review up to 573899084e08fb7216a2e40719f773c3eb66ad0a (I'm still reviewing the last commit).

zcash_client_backend/src/data_api.rs
```diff
@@ -794,6 +794,69 @@ impl<NoteRef> SpendableNotes<NoteRef> {
...
+///
+/// At present this just contains counts of unspent outputs in each pool, but it may be extended in
+/// the future to contain note values or other more detailed information about wallet structure.
+///
+/// Values of this type are intended to be used in selection of change output values. A value of
+/// this type may represent filtered data, and may therefore not count all of the unspent notes in
+/// the wallet.
+pub struct WalletMeta {
+    sapling_note_count: usize,
+    #[cfg(feature = "orchard")]
```
daira, line 807: Do we really need to gate this on the `orchard` feature? It can just be zero whenever Orchard is not enabled.

zcash_client_backend/src/data_api.rs
```diff
@@ -822,6 +885,17 @@ pub trait InputSource {
         exclude: &[Self::NoteRef],
     ) -> Result<SpendableNotes<Self::NoteRef>, Self::Error>;
 
+    /// Returns metadata describing the structure of the wallet for the specified account.
+    ///
+    /// The returned metadata value must exclude spent notes and unspent notes having value less
+    /// than the specified minimum value or identified in the given exclude list.
```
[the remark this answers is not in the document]
daira: This is a good change, because I initially misinterpreted it as "(spent notes and unspent notes) having ..." and wondered why that was not just "notes having ...".

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -90,18 +93,118 @@ impl<I: InputSource> ChangeStrategy for SingleOutputChangeStrategy<I> {
...
+    /// fee parameters, change memo, and change splitting policy.
+    ///
+    /// This change strategy will fall back to creating a single change output if insufficient
+    /// change value is available to create notes with at least the minimum value dictated by the
+    /// split policy.
+    ///
+    /// `fallback_change_pool`: the pool to which change will be sent if when more than one
+    /// shielded pool is enabled via feature flags, and the transaction has no shielded inputs.
+    /// `split_policy`: A policy value describing how the change value should be returned as
+    /// multiple notes.
```
str4d, line 145:
```suggestion
    /// - `fallback_change_pool`: the pool to which change will be sent if when more than one
    ///   shielded pool is enabled via feature flags, and the transaction has no shielded inputs.
    /// - `split_policy`: A policy value describing how the change value should be returned as
    ///   multiple notes.
```

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -184,6 +288,132 @@ mod tests {
...
+                None,
+                Some(&WalletMeta::new(
+                    0,
+                    #[cfg(feature = "orchard")]
+                    0,
+                )),
+            );
+
+            // FIXME: The current splitting strategy may result in overpaying fees by a small amount.
+            // This may be acceptable for an initial implementation.
```
str4d, line 391: To fix this, can we recompute the fee once we have working outputs, and then add the excess to the first change output? That seems like a simple way to address this inside the transaction builder rather than in this fee calculator. I don't think we need to aim for perfectly-split notes (the first note already differs due to the remainder).

str4d (changes requested):
Reviewed 6c95fd8bb0296959ead29dca6d6b1ab5ed4361db.

zcash_client_backend/src/data_api/error.rs
```diff
@@ -33,6 +35,9 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError> {
     /// An error in note selection
     NoteSelection(SelectionError),
 
+    /// An error in change selection during transaction proposal construction
+    Change(ChangeError<ChangeErrT, NoteRefT>),
```
daira, line 39: The fact that this argument is of the form `crate::fees::ChangeError<ChangeErrT, NoteRefT>` does not appear to be used. Consider whether using just `ChangeError` (as a type parameter, not imported from `crate::fees`) would be sufficient.

zcash_client_backend/src/data_api.rs
daira: Oh, it's public. Yeah I guess all the other public APIs gate the Orchard bits on the `orchard` feature.

zcash_client_backend/src/data_api.rs
```diff
@@ -794,6 +794,69 @@ impl<NoteRef> SpendableNotes<NoteRef> {
...
+
+    /// Returns the total number of unspent shielded notes belonging to the account for which this
+    /// was generated.
+    pub fn total_note_count(&self) -> usize {
+        #[cfg(feature = "orchard")]
+        let orchard_note_count = self.orchard_note_count;
+        #[cfg(not(feature = "orchard"))]
+        let orchard_note_count = 0;
+
+        self.sapling_note_count + orchard_note_count
```
daira, line 856:
```suggestion
        self.sapling_note_count + self.note_count(ShieldedProtocol::Orchard)
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -225,3 +225,55 @@ where
...
+        .collect();
+    let excluded_ptr = Rc::new(excluded);
+
+    conn.query_row(
+        &format!(
+            "SELECT COUNT(*)
+             FROM {table_prefix}_received_notes
+             INNER JOIN accounts
+               ON accounts.id = {table_prefix}_received_notes.account_id
+             WHERE value >= :min_value
```
daira, line 256: Unlike `get_notes` or `select_spendable_notes`, this counts notes in unmined transactions. Is that intended? It doesn't take into account whether the transactions are likely to be mined. In general it makes me slightly more nervous of metadata leakage when transaction construction policy is dependent on mempool information (because that's less consistent across nodes).

To exclude notes in unmined transactions you would need something like:
```suggestion
             FROM {table_prefix}_received_notes rn
             INNER JOIN accounts ON accounts.id = rn.account_id
             INNER JOIN transactions ON transactions.id_tx = rn.tx
             WHERE transactions.block IS NOT NULL
             AND value >= :min_value
```

daira: Oh wait, is `commitment_tree_position` always NULL if the note is in an unmined transaction? I don't see how we could know it otherwise. In that case why does `get_notes` explicitly join on `transactions` and check for `transactions.block IS NOT NULL`? (Since `get_notes` also has `commitment_tree_position IS NOT NULL`, checking `transactions.block IS NOT NULL` there should be redundant by the same argument.)

zcash_client_backend/CHANGELOG.md
```diff
@@ -12,15 +12,62 @@ and this library adheres to Rust's notion of
...
+    the addition of two type parameters to the `Error` type,
+    `ChangeErrT` and `NoteRefT`.
+  - The following methods each now take an additional `change_strategy`
+    argument, along with an associated `ChangeT` type parameter:
+    - `zcash_client_backend::data_api::wallet::spend`
+    - `zcash_client_backend::data_api::wallet::propose_transfer`
+    - `zcash_client_backend::data_api::wallet::propose_shielding`. This method
+      also now takes an additional `to_account` argument.
+    - `zcash_client_backend::data_api::wallet::shield_transparent_funds`. This
+      method also now takes an additional `to_account` argument.
```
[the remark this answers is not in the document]
daira: Since lines 23-49 inclusive are all about `zcash_client_backend::data_api::wallet`, I suggest making that a section.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1227,15 +1221,22 @@ impl<Cache, DbT: WalletRead + Reset> TestState<Cache, DbT, LocalNetwork> {
 
 /// Helper method for constructing a [`GreedyInputSelector`] with a
 /// [`standard::SingleOutputChangeStrategy`].
-pub fn input_selector<DbT: InputSource>(
+pub fn input_selector<DbT: InputSource>() -> GreedyInputSelector<DbT> {
+    GreedyInputSelector::<DbT>::new()
+}
```
daira, line 1226: I cannot find any uses of this. Since we're changing its signature and since `GreedyInputSelector::new()` is already public, why not just delete it?

daira, line 1226: I cannot find any uses of this. Since we're changing its arguments and since `GreedyInputSelector::new()` is already public, why not just delete it?

In any case, if it is kept then the "with a [`standard::SingleOutputChangeStrategy`]" part of the comment no longer makes sense, because the input selector is now independent of the change strategy.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1227,15 +1221,22 @@ impl<Cache, DbT: WalletRead + Reset> TestState<Cache, DbT, LocalNetwork> {
 
 /// Helper method for constructing a [`GreedyInputSelector`] with a
 /// [`standard::SingleOutputChangeStrategy`].
```
daira, line 1223:
```suggestion
```

zcash_client_backend/src/data_api/testing.rs
daira, line 1226: I cannot find any uses of this. Since we're changing its arguments and since `GreedyInputSelector::new()` is already public, why not just delete it?

In any case, if it is kept then the "with a [`standard::SingleOutputChangeStrategy`]" part of the comment no longer makes sense, because the input selector is now independent of the change strategy.

```suggestion
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -310,6 +316,175 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    assert!(found_tx_empty_memo);
+
+    // Verify that the stored sent notes match what we're expecting
+    let sent_note_ids = st
+        .wallet()
+        .get_sent_note_ids(&sent_tx_id, T::SHIELDED_PROTOCOL)
+        .unwrap();
+    assert_eq!(sent_note_ids.len(), 3);
+
+    // The sent memo should be the empty memo for the sent output, and the
```
daira, line 425:
```suggestion
    // The sent memo should be the empty memo for the sent output, and each
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -310,6 +316,175 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+            }
+            Some(m) if m == &Memo::Empty => {
+                found_sent_empty_memo = true;
+            }
+            Some(other) => panic!("Unexpected memo value: {:?}", other),
+            None => panic!("Memo should not be stored as NULL"),
+        }
+    }
+    assert_eq!(change_memo_count, 2);
+    assert!(found_sent_empty_memo);
```
daira, line 447: I think nothing here checks that it is specifically the sent note (that is, the one with value 100_0000 zatoshis sent to the `to` address) that has the empty memo.

zcash_client_backend/src/fees/zip317.rs
nuttycom: Yup, I'm adding a commit to fix this.

components/zcash_protocol/src/value.rs
```diff
@@ -298,6 +317,15 @@ impl Zatoshis {
     pub fn is_positive(&self) -> bool {
         self > &Zatoshis::ZERO
     }
+
+    /// Divides this `Zatoshis` value by the given divisor and returns the quotient and remainder.
+    pub fn div_with_remainder(&self, divisor: NonZeroUsize) -> DivRem<Zatoshis> {
+        let divisor = u64::try_from(usize::from(divisor)).expect("divisor fits into a u64");
```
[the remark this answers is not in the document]
nuttycom: This makes the code in `zcash_client_backend` a bit more annoying, because there we're dealing mostly with `usize` values that are coming from `.len()` calls; `NonZeroUsize` still makes sense for `target_output_count` in the `SplitPolicy` type I think?

zcash_client_backend/src/data_api/error.rs
nuttycom: I don't understand this suggestion.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1396,14 +1583,18 @@ pub fn external_address_change_spends_detected_in_restore_from_seed<T: ShieldedP
 
     #[allow(deprecated)]
     let fee_rule = FixedFeeRule::standard();
-    let input_selector = GreedyInputSelector::new(
-        fixed::SingleOutputChangeStrategy::new(fee_rule, None, T::SHIELDED_PROTOCOL),
+    let change_strategy = fees::fixed::SingleOutputChangeStrategy::new(
```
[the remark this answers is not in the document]
nuttycom: If you want, we can change this, or delete the other change strategies, in a later PR. It should not be changed here though; this PR is already big enough.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: It's there: https://github.com/zcash/librustzcash/pull/1579/files#diff-1cc140d33c5912f6e98cc8bd07ee744a81aa7a70ee4becabe24617a5c9608f49R50-R51

zcash_client_sqlite/src/wallet/common.rs
nuttycom: Hm, I think the `commitment_tree_position` check is actually superfluous; it's the join against the `blocks` table that ensures that the transaction has been mined.

zcash_client_backend/src/fees.rs
```diff
@@ -336,6 +339,67 @@ impl Default for DustOutputPolicy {
...
+
+    /// Returns the number of output notes to produce from the given total change value, given the
+    /// number of existing unspent notes in the account and this policy.
+    pub fn split_count(
+        &self,
+        existing_notes: usize,
+        total_change: NonNegativeAmount,
+    ) -> NonZeroUsize {
+        let mut split_count =
+            NonZeroUsize::new(usize::from(self.target_output_count).saturating_sub(existing_notes))
```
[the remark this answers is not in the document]
nuttycom: I think it makes the most sense to perform this conversion just in the `div_with_remainder` call.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -225,3 +225,55 @@ where
...
+               WHERE stx.block IS NOT NULL -- the spending tx is mined
+               OR stx.expiry_height IS NULL -- the spending tx will not expire
+               OR stx.expiry_height > :anchor_height -- the spending tx is unexpired
+             )"
+        ),
+        named_params![
+            ":account_id": account.0,
+            ":min_value": u64::from(min_value),
+            ":exclude": &excluded_ptr
+        ],
```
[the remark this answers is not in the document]
nuttycom: I'm puzzled as to why this wasn't causing an error in tests.

nuttycom:
[https://github.com/zcash/librustzcash/compare/6c95fd8bb0296959ead29dca6d6b1ab5ed4361db..eb28bfd074fa71f75c226148226a6c744805504d] force-pushed to address comments from code review.

zcash_client_backend/src/data_api.rs
```diff
@@ -794,6 +794,69 @@ impl<NoteRef> SpendableNotes<NoteRef> {
...
+/// Values of this type are intended to be used in selection of change output values. A value of
+/// this type may represent filtered data, and may therefore not count all of the unspent notes in
+/// the wallet.
+pub struct WalletMeta {
+    sapling_note_count: usize,
+    #[cfg(feature = "orchard")]
+    orchard_note_count: usize,
+}
+
+impl WalletMeta {
```
daira, line 811: It bugs me a little that the concrete type here has the same name as the associated type on `ChangeStrategy`. I don't understand why they should have the same name. It's just confusing.

zcash_client_backend/src/fees.rs
```diff
@@ -368,13 +443,30 @@ impl EphemeralBalance {
...
+    type FeeRule: FeeRule + Clone;
     type Error;
 
+    /// The type of metadata source that this change strategy requires in order to be able to
+    /// retrieve required wallet metadata. If more capabilities are required of the backend than
+    /// are exposed in the [`InputSource`] trait, the implementer of this trait should define their
+    /// own trait that descends from [`InputSource`] and adds the required capabilities there, and
+    /// then implement that trait for their desired database backend.
+    type MetaSource: InputSource;
+    type WalletMeta;
```
daira, line 455: * Please document this type.
```suggestion
    type WalletMetaT;
```

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/5afea2e3b1aa65248496c2142d72b0bf3c3ea48c..4059a29b2b1615712d820428c0e6671122e9532d) to address additional comments from my pairing with @daira.

zcash_client_backend/src/data_api/error.rs
pacu: Maybe Daira-Emma means the NoteRef side of it?

IIUC the error is actually used on line 130 and 238 and as associated type of the enum `InputSelectorError`. Although I'm not really well versed on this advanced usage of generics in RustLang.

zcash_client_backend/src/fees/common.rs
```diff
@@ -410,13 +431,73 @@ where
...
+            });
+            let per_output_change = proposed_change.div_with_remainder(
+                NonZeroU64::new(
+                    u64::try_from(usize::from(split_count)).expect("usize fits into u64"),
+                )
+                .unwrap(),
+            );
+
+            // If we don't have as many change outputs as we expected, recompute the fee.
+            let (fee_with_change, excess_fee) =
```
pacu, line 443: [non-blocking] This section of the code from L. 450 to L. 526 is pretty hard to read and follow (even with the generous comments). I wonder if there's a way to reduce its complexity in terms of lamdas and if-else statements so not only is more friendly to follow and maintain, but also to test.

pacu (commented):
LGTM! 

left some non-blocking suggestions

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/4059a29b2b1615712d820428c0e6671122e9532d..47b1065db901e9e423e3b0cdb35fc94d9bd4b24c) to rebase on `main` to fix changelog conflicts.

str4d (approved):
utACK 47b1065db901e9e423e3b0cdb35fc94d9bd4b24c

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -598,6 +543,63 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                        &wallet_meta,
+                    )?;
+                    assert_eq!(tr1_balance.total(), tr1_balance.fee_required());
+
+                    (
+                        Some(EphemeralBalance::Output(tr1_required_input_value)),
+                        Some(tr1_balance),
+                    )
+                }
+            };
```
daira, line 601: Why does this need to be inside the loop? `ephemeral_balance` and `tr1_balance_opt` should not need to be recomputed on each loop iteration, I think.

daira: Oh I see, it's because `wallet_meta` may change.  I don't especially like that this requires redundant computation for change strategies that don't need it, but I guess that's ok. It also increases the volume of code in an already-too-long loop (which could be fixed with a closure). I'll see if I can refactor this code to make it more understandable overall.

zcash_client_backend/src/fees.rs
```diff
@@ -393,7 +488,11 @@ pub trait ChangeStrategy {
     /// - `ephemeral_balance`: if the transaction is to be constructed with either an
     ///   ephemeral transparent input or an ephemeral transparent output this argument
     ///   may be used to provide the value of that input or output. The value of this
-    ///   output should be `None` in the case that there are no such items.
+    ///   argument should be `None` in the case that there are no such items.
+    /// - `wallet_meta`: Additional wallet metadata that the change strategy may use
+    ///   in determining how to construct change outputs. This wallet metadata value
+    ///   should be computed excluding the inputs provided in the `transparent_inputs`,
+    ///   `sapling`, and `orchard` arguments.
```
daira, line 495: Oh I see, that's why the metadata needs to be recomputed in the loop.

zcash_client_backend/src/fees/common.rs
```diff
@@ -229,35 +298,36 @@ where
...
+            Err(ChangeError::BundleError(
+                "Nonzero Orchard change requested but the `orchard` feature is not enabled.",
+            ))
+        } else {
+            Ok(0)
+        }
+    };
 
     // Once we calculate the balance with and without change, there are five cases:
     //
```
daira, line 333: "change output is added" -> "change output(s) are added" on line 337.

zcash_client_backend/src/fees/common.rs
```diff
@@ -353,35 +424,96 @@ where
...
+            let split_count = wallet_meta.map_or(NonZeroUsize::MIN, |wm| {
+                cfg.split_policy
+                    .split_count(wm.total_note_count(), proposed_change)
+            });
+            let per_output_change = proposed_change.div_with_remainder(
+                NonZeroU64::new(
+                    u64::try_from(usize::from(split_count)).expect("usize fits into u64"),
+                )
+                .unwrap(),
+            );
```
daira, line 440: If `proposed_change` is not a multiple of `split_count`, this will create slightly less change than expected, presumably leaking the remainder publically in the fee. The remainder then gives a lower bound on `split_count`, which should have been private (if the adversary is uncertain about how many real outputs there are).

zcash_client_backend/src/fees/common.rs
```diff
@@ -353,35 +424,96 @@ where
...
+                match cfg.dust_output_policy.action() {
                     DustAction::Reject => {
                         // Always allow zero-valued change even for the `Reject` policy:
                         // * it should be allowed in order to record change memos and to improve
                         //   indistinguishability;
                         // * this case occurs in practice when sending all funds from an account;
                         // * zero-valued notes do not require witness tracking;
                         // * the effect on trial decryption overhead is small.
-                        if proposed_change.is_zero() {
+                        if per_output_change.quotient().is_zero() {
```
daira, line 512: The value of the first change output can be non-zero even when the quotient is zero. That should strictly speaking be rejected by the `DustAction::Reject` policy.
```suggestion
                        if proposed_change.is_zero() && excess_fee.is_zero() {
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -353,35 +424,96 @@ where
...
                             simple_case()
                         } else {
-                            let shortfall =
-                                (change_dust_threshold - proposed_change).ok_or_else(underflow)?;
+                            let shortfall = (change_dust_threshold - *per_output_change.quotient())
```
daira, line 515: This change also doesn't look right to me. It should be the overall shortfall.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -399,6 +399,11 @@ pub(crate) mod tests {
         testing::pool::send_single_step_proposed_transfer::<OrchardPoolTester>()
     }
 
+    #[test]
+    fn send_with_multiple_change_outputs() {
+        testing::pool::send_with_multiple_change_outputs::<SaplingPoolTester>()
```
daira, line 404:
```suggestion
        testing::pool::send_with_multiple_change_outputs::<OrchardPoolTester>()
```

daira (commented):
Post-hoc ACK with comments (please address them in #1590).

zcash_client_sqlite/src/wallet/orchard.rs
str4d: I spotted a similar bug in an existing test while rebasing another PR; we may want to check them all.

zcash_client_backend/src/fees/common.rs
nuttycom: Resolved in https://github.com/zcash/librustzcash/pull/1590/commits/d3b82ee51f7da5fdd3a0cb08cc81d328d58da0e3

zcash_client_backend/src/fees/common.rs
nuttycom: Resolved in https://github.com/zcash/librustzcash/pull/1590/commits/d3b82ee51f7da5fdd3a0cb08cc81d328d58da0e3

PR #1582, opened 2024-10-18
Bump MSRV to 1.77

str4d:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -2020,7 +2020,21 @@ pub(crate) fn block_fully_scanned<P: consensus::Parameters>(
         //
         // The fully-scanned height is therefore the last height that falls within the first range in
         // the scan queue with priority "Scanned".
-        // SQL query problems.
+        let calc_fully_scanned_height = |row: &rusqlite::Row| {
```
nuttycom, line 2023: This change is fine, but for my education what in the MSRV change prompted it?

nuttycom (approved):
utACK 9bf153d002e7729a19dd16c0c7b4a367921ac897

zcash_client_sqlite/src/wallet.rs
str4d: The new lint is "in a `match` scrutinee, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`".

https://rust-lang.github.io/rust-clippy/master/index.html#blocks_in_conditions

PR #1585, opened 2024-10-23
zcash_client_sqlite: Ensure that all shielded change outputs are correctly flagged.

### LINKED ISSUE #1571: Bug in is_shielding state going from TRUE to FALSE after the first confirmation
- the states of the DB and is_shielding changes from true to false

How to reproduce:
- I tapped a button to shield my funds, it:
- created a transaction in a pending state with is_shielding TRUE
- after 1st confirmation, it flipped to sending instead of shielding, I checked the DB and there is no TRUE anymore (see the DB)
- after 10 confirmations, it flipped to sent
- when I restore the same wallet, I see shielded again

Few more observations:
- when I try to restore the wallet on a different simulator while it’s still waiting on 10 confirmations and should be shielding funds but it’s sending, I see correct state restored.. so restore is able to bring the TRUE flag back but where it matters it’s broken

- db export shared via Slack here: https://zcash.slack.com/archives/C02C1GZ0R6F/p1724835111523729

nuttycom:
Fixes #1571

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -3223,9 +3223,38 @@ fn recipient_params<P: consensus::Parameters>(
     }
 }
 
+fn flag_previously_received_change(
+    conn: &rusqlite::Transaction,
+    tx_ref: TxRef,
+) -> Result<(), SqliteClientError> {
+    let flag_received_change = |table_prefix| {
+        conn.execute(
+            &format!(
+                "UPDATE {table_prefix}_received_notes 
+                 SET is_change = 1
+                 FROM sent_notes sn
+                 WHERE sn.tx = {table_prefix}_received_notes.tx
+                 AND sn.tx = :tx
+                 AND sn.from_account_id = {table_prefix}_received_notes.account_id
+                 AND {table_prefix}_received_notes.recipient_key_scope = :internal_scope"
+            ),
+            named_params! {
+                ":tx": tx_ref.0,
+                ":internal_scope": scope_code(Scope::Internal)
+            },
+        )
+    };
+
+    flag_received_change(SAPLING_TABLES_PREFIX)?;
+    #[cfg(feature = "orchard")]
+    flag_received_change(ORCHARD_TABLES_PREFIX)?;
+
+    Ok(())
+}
+
 /// Records information about a transaction output that your wallet created.
 pub(crate) fn insert_sent_output<P: consensus::Parameters>(
-    conn: &rusqlite::Connection,
+    conn: &rusqlite::Transaction,
     params: &P,
     tx_ref: TxRef,
     from_account: AccountId,
@@ -3253,6 +3282,7 @@ pub(crate) fn insert_sent_output<P: consensus::Parameters>(
     ];
 
     stmt_insert_sent_output.execute(sql_args)?;
+    flag_previously_received_change(conn, tx_ref)?;
 
     Ok(())
 }
@@ -3270,7 +3300,7 @@ pub(crate) fn insert_sent_output<P: consensus::Parameters>(
 ///   the transaction.
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn put_sent_output<P: consensus::Parameters>(
-    conn: &rusqlite::Connection,
+    conn: &rusqlite::Transaction,
     params: &P,
     from_account: AccountId,
     tx_ref: TxRef,
@@ -3307,6 +3337,7 @@ pub(crate) fn put_sent_output<P: consensus::Parameters>(
     ];
 
     stmt_upsert_sent_output.execute(sql_args)?;
+    flag_previously_received_change(conn, tx_ref)?;
 
     Ok(())
 }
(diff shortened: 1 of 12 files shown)


zcash_client_sqlite/src/wallet/init/migrations/fix_bad_change_flagging.rs
```diff
@@ -0,0 +1,198 @@
...
+        assert_eq!(tx.spent_note_count(), 1);
+        assert!(tx.has_change());
+        assert_eq!(tx.received_note_count(), 0);
+        assert_eq!(tx.sent_note_count(), 0);
+        assert!(tx.is_shielding());
+
+        // Complete the migration
+        init_wallet_db(st.wallet_mut().db_mut(), None).unwrap();
+
+        // Ensure that the transaction metadata is still correct after the update produced by scanning.
```
str4d, line 176: What scanning update? AFAICT this test doesn't do one; it just checks that the unscanned shielding transaction's metadata is not altered by the migration. It might indeed be useful to have two tests for this migration: one that ensures correct data isn't altered, and one that ensures incorrect data gets fixed.

zcash_client_sqlite/src/wallet/init/migrations/fix_bad_change_flagging.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        let fix_change_flag = |table_prefix| {
+            transaction.execute(
+                &format!(
+                    "UPDATE {table_prefix}_received_notes
+                     SET is_change = 1
+                     FROM sent_notes sn
+                     WHERE sn.tx = {table_prefix}_received_notes.tx
+                     AND sn.from_account_id = {table_prefix}_received_notes.account_id
```
str4d, line 52: This won't catch cross-account wallet-internal transfers, but I think that's okay.

zcash_client_sqlite/src/wallet/init/migrations/fix_bad_change_flagging.rs
```diff
@@ -0,0 +1,198 @@
...
+                     AND {table_prefix}_received_notes.recipient_key_scope = :internal_scope"
```
str4d, line 53: This restriction means that funds sent from an account to itself via its external address will not be updated here. For modern wallets that makes sense, but for historic wallets (when internal scope didn't exist) a "change" output was simply an output received by the account in a tx that the account also spent funds in. Do we have reason to believe that those earlier transactions are or are not affected by this bug?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3221,9 +3221,38 @@ fn recipient_params<P: consensus::Parameters>(
...
+    conn: &rusqlite::Transaction,
+    tx_ref: TxRef,
+) -> Result<(), SqliteClientError> {
+    let flag_received_change = |table_prefix| {
+        conn.execute(
+            &format!(
+                "UPDATE {table_prefix}_received_notes 
+                 SET is_change = 1
+                 FROM sent_notes sn
+                 WHERE sn.tx = {table_prefix}_received_notes.tx
```
str4d, line 3234: What kind of join does this implicitly do? (both here and in the migration which has an identical query)

It might be useful to add a comment describing the intent of this query.

str4d (commented):
Reviewed 0d82e1efc50aff89db715df1065f6a331701b361

zcash_client_sqlite/src/wallet/init/migrations/fix_bad_change_flagging.rs
nuttycom: That's actually intentional; a cross-account wallet-internal transfer output isn't change.

zcash_client_sqlite/src/wallet.rs
nuttycom: This is implicitly an inner join, the same as you get whenever you use a `FROM` clause with multiple tables and then use `WHERE` to perform the joins.

zcash_client_sqlite/src/wallet/init/migrations/fix_bad_change_flagging.rs
nuttycom: The legacy shielding transactions likely _will_ be affected by this problem in wallet recovery. However, I think that's acceptable; the wallets at the time those transactions were created did not show "shielding" for those transactions in wallet history, so this leaves that part of the transaction history essentially unmodified from the user's perspective.

zcash_client_sqlite/src/wallet/init/migrations/fix_bad_change_flagging.rs
nuttycom: Copy/paste error - this comment was correct in the test for the change from `IS_NULL` to `MAX`, and the test is the same.

nuttycom: [force-pushed](https://github.com/zcash/librustzcash/compare/0d82e1e..d75309080106639920f8302e555f482dfc35e6c2) to address comment.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/bca02d5c148d8e080f52ce32e1936fb55c8df62f..d75309080106639920f8302e555f482dfc35e6c2) to rebase on `main`.

str4d (approved):
utACK d75309080106639920f8302e555f482dfc35e6c2

daira (commented):
Post-hoc ACK

PR #1595, opened 2024-10-26
zcash_client_backend: Switch from `async-trait` to RPIT for `tor` module

str4d:
RPIT (return-position `impl Trait`) was stabilized in Rust 1.75, and the `dynosaur` crate providing dynamic dispatch support was published two weeks ago. Now that our MSRV is 1.77, we can use both of these.

### DIFF
--- a/zcash_client_backend/src/tor/http/cryptex/binance.rs
+++ b/zcash_client_backend/src/tor/http/cryptex/binance.rs
@@ -1,4 +1,3 @@
-use async_trait::async_trait;
 use rust_decimal::Decimal;
 use serde::Deserialize;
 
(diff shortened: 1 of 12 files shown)


zcash_client_backend/src/tor/http/cryptex/binance.rs
```diff
@@ -1,4 +1,3 @@
-use async_trait::async_trait;
```
daira, line 1: Can this also be used to remove librustzcash's one remaining dependency on `async_trait` for `data_api::chain::BlockCache`?

daira (approved):
utACK

zcash_client_backend/src/tor/http/cryptex/binance.rs
daira: Ah you're way ahead of me, lol. https://github.com/zcash/librustzcash/pull/1535#issuecomment-2439378806

nuttycom (commented):
post-hoc utACK

PR #1580, opened 2024-10-17
zcash_client_backend: Generalize & extend wallet metadata query API

nuttycom:

PR #1603, opened 2024-11-02
Dev quality-of-life improvements (better test backtraces in CI and locally; faster test builds; `RUST_BACKTRACE=1` by default)

### LINKED ISSUE #1602: CI tests should run with `RUST_BACKTRACE=1`
This would be more useful, especially for non-reproducible failures.

daira:
* Remove `.gitlab-ci.yml` which is not used (and probably bitrotted).
* Changes the test profile to compile with optimizations by default, but keep debug info.
  * This [improves both build+test times and backtraces](https://github.com/zcash/librustzcash/pull/1603/commits/9d1c898355875cc2228212ee5b26855e27f6773f) relative to `cargo test --release`, and does not require you to remember to pass `--release` (you still can, and the meaning of that has not changed).
  * You can use `--profile=dev` to get more precise stack traces in case information is optimized out. Remember to only run the desired subset of tests in that case (as before).
  * CI testing now uses this test profile instead of `--release`.
* `cargo` will now use `RUST_BACKTRACE=1` by default in this workspace, so that there is no need to set it manually.

In other words, `cargo test` now just does the Right Thing :tada: (you may still have to set features).

fixes #1602


daira:
See https://github.com/zcash/librustzcash/pull/1604#issuecomment-2453085057 for an example backtrace from CI.

daira:
This significantly improves CI latency for tests.

Before (#1580):
image

After (this PR):
image

nuttycom (approved):
utACK, looks great!

str4d:
Post-hoc ACK.

PR #1605, opened 2024-11-02
[Do not merge] Check CI backtraces 2

daira:
Just checking whether the `.cargo/config.toml` approach used in https://github.com/zcash/librustzcash/pull/1603/commits/5672b307b688e3375e509b29332801b3bddb952b also works.


daira:
It does work.

PR #1607, opened 2024-11-02
Change `opt-level` from 3 to 2 for the test profile

daira:


daira:
There is an improvement to build+test latency on CI, but it's small:
* "Test on Linux" took 5m 53s with this PR, vs 6m 2s before (a 2.5% improvement)
* "Test Orchard on Linux" took 7m 23s with this PR, vs 7m 39s before (a 3.5% improvement) 

The disadvantage is that testing would not see miscompilations that happen at opt-level 3 but not opt-level 2. On reflection, I think this is a significant downside. Testing at the same optimization level as the code we release (even without LTO, which admittedly could be a source of miscompilations) seems like a good thing to maintain.

PR #1608, opened 2024-11-04
[doc-only] Minor documentation fix and clarification for the data API

daira:
* Fix the documentation for `find_account_for_ephemeral_address` which included a use of `get_known_ephemeral_addresses` with incorrect types.
* Clarify that passing `None` for the `index_range` parameter of `get_known_ephemeral_addresses` is equivalent to passing `0..(1u32 << 31)`.


nuttycom (approved):
ACK, doc-only

PR #1614, opened 2024-11-14
Release zcash_protocol version 0.4.1

nuttycom:


components/zcash_protocol/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+## [0.4.1] - 2024-11-13
```
str4d, line 10: Hmm, we bump MSRV in this release, which I usually try to avoid doing in a point release (because there's no way to subdivide point releases, meaning if we ever needed to backport a bugfix we wouldn't be able to do so to pre-MSRV-bump downstreams).

str4d (approved):
utACK a1511cbb686570c4b2169c03c01a33442c5004b7

str4d (commented):
Re-utACK 20d00f6ecf8de13a6b340d435eaa541f524ed411 (just the cargo audit update).

PR #1615, opened 2024-11-14
Release zcash_primitives version 0.20.0

nuttycom:


zcash_primitives/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+## [0.20.0] - 2024-11-14
```
str4d, line 10: All of the downstream changelogs need to mention the migration to `zcash_primitives 0.20`. See e.g. the `zcash_primitives 0.19.0` release commit: c36c40818186484f435057019037f2fbc99fe705

str4d: @nuttycom has informed me that for this particular release, these are being added in the downstream crate releases. I would prefer we don't do that unless we're using automation to ensure we aren't missing any, but given that this release train has already been prepared, I'm not going to block on changing it.

str4d (dismissed):
utACK df0a42ef88a81b624e67e4b813f14128fe05dc32

str4d (approved):
utACK aa3ee689e1cc225e03a96fe001d2c6932d7a0b21

PR #1619, opened 2024-11-14
Release/zcash client sqlite 0.13.0

nuttycom:


zcash_client_backend/CHANGELOG.md
```diff
@@ -33,6 +35,7 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.77.0.
+- Migrated to `zcash_protocol 0.4.1`, `zcash_primitives 0.20.0`, `zcash_keys 0.5.0`.
```
str4d, line 38: I don't think we need to mention point release migrations, because all point releases are compatible in the sense that someone on `0.4.0` can trivially use `0.4.1` instead, but couldn't use `0.5`. This is also why I generally only include the first two digits of the version in "migrated to" messages (as those are the backwards-incompatible and/or MSRV-changing digits).

str4d (dismissed):
utACK 92c84618a4a16e3b2f7af67e9fd3e8912140fe4b with nit.

zcash_client_backend/CHANGELOG.md
nuttycom: In this case, we actually do require 0.4.1; that's the only reason I called it out.

str4d: We do, but the fact that we do is not a breaking change to downstream users of our crate.

str4d (dismissed):
utACK 484d0ef57181e1f0c9eb94a6a8458f0c91bb5134

str4d (approved):
utACK a240d49634d8b10b97904192d0a5761401138333

PR #1627, opened 2024-11-20
Add serde derives to primitive types behind feature flag

willemolding:
- Adds a serde feature to the `zcash_protocol`, `zcash_address`, and `zcash_primtives` crates. Enabling this feature adds serde derives to a number of types which are use in `zcash_client_memory` for serialization.

- Also adds a `wasm-bindgen` feature to `zcash_client_backend` crate to enable the require features in dependencies when building for wasm

- Also applied the toml formatter to some of the Cargo.toml files. This was unintentional but I figured it is nice to have these meet formatting standards so I left it in


components/zcash_protocol/Cargo.toml
```diff
@@ -11,7 +11,7 @@ repository.workspace = true
 readme = "README.md"
 license.workspace = true
 edition.workspace = true
-rust-version = "1.70"
+rust-version.workspace = true
```
nuttycom, line 14: Is the MSRV bump here required? We had opted not to bump past 1.70 in the previous update in order to avoid requiring a semver major version bump; since this looks like it should be additive, it'd be nice to retain an MSRV of 1.70 if we don't explicitly need features from the newer MSRV. Nonblocking, but the semver bump (and the additions) should be noted in the CHANGELOG(s).

components/zcash_protocol/src/lib.rs
```diff
@@ -25,7 +25,8 @@ pub mod memo;
 pub mod value;
 
 /// A Zcash shielded transfer protocol.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
nuttycom, line 29: What is the stability requirement for the serialized values here? If new enum variants are added, does the order matter in terms of being able to deserialize existing persistent data? Please document such requirements; if there's a way to assign stable identifiers for the serialization, that would be preferable instead of relying upon the ones that are implicitly assigned.

nuttycom (changes requested):
Please add documentation for the feature flag and the new `impl`s to the various CHANGELOGs as appropriate, and also see my question about the `zcash_protocol` MSRV. Otherwise LGTM.

Issue #1335, opened 2024-03-28
Support for multiple accounts


AArnott:
It seems we need a high-level issue tracking overall support for multiple accounts:

Here are some work items that occur to me:
- `WalletWrite` should expose a function that allows adding HD-derived accounts at a particular index. The existing `create_account` function assumes "last index + 1" which is a great policy, but doesn't provide the app with even an awareness of which index was created. I feel the app should also be able to control this, but with fn-level docs urging the app to only create one account index at a time and in-order. (see also #628)
- The ability to import accounts via UFVK and UIVK. A parameter may indicate whether the caller expects that a spending key is available for this account so that sync knows whether to maintain commitment trees (or whatever the marginal cost is for spendable accounts).
- Adding a second account should trigger partial re-scan. That is, the `scan_queue` should indicate that blocks must be re-scanned. Ideally, the rescan would _only_ run trial decryptions for recently added accounts that have not yet been scanned.
- Removing accounts

str4d:
In a meeting today, @daira @nuttycom and I discussed how wallet apps should identify accounts. We discussed several options, including deriving a persistent account-level identifier from the seed phrase, or deriving something from the Orchard IVK, but each of these ran into problems with various use cases (e.g. imporing the ZecPages viewing key, which has no Orchard component and for which we couldn't get a different fingerprint).

Instead, we've decided to continue with account identifiers that are local-only: they are only valid in the context of a specific wallet database, and MUST NOT be cached by wallet apps across wallet recreation events. Our plan is to:

- `zcash_client_backend`:
  - Augment `WalletRead::AccountId` with the ability to encode to, and parse from, opaque bytes. This encoding MUST NOT be parsed or otherwise interpreted by wallet apps.
  - Document on `AccountId` its local-only properties and how wallet apps must treat them.
- `zcash_client_sqlite` specifically:
  - Introduce a per-wallet UUID (in its own database as a singleton, that is generated via a migration)
  - Encode its account IDs (which are currently the account's primary key from the `accounts` table) as the tuple `(wallet_uuid, account_primary_key)`.

str4d:
> This doesn't need to be something that's exposed at the `zcash_client_backend` level

[...]

PR #1631, opened 2024-11-21
zcash_client_sqlite: Assign UUIDs to each account

### LINKED ISSUE #1629: zcash_client_sqlite: Implement the new multi-seed-compatible account ID

str4d:
This provides equivalent uniqueness to the `accounts` table primary key, but avoids collisions across wallet recreation events (to defend against downstream crate users who don't flush any persisted account IDs at those events).

Closes #1629.

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -9,8 +9,6 @@ and this library adheres to Rust's notion of
 
 ### Added
 - `zcash_client_sqlite::AccountUuid`
-- `zcash_client_sqlite::Account::uuid`
-- `zcash_client_sqlite::WalletDb::get_account_for_uuid`
 
 ### Changed
 - The `v_transactions` view has been modified:
@@ -18,10 +16,18 @@ and this library adheres to Rust's notion of
 - The `v_tx_outputs` view has been modified:
   - The `from_account_id` column has been replaced with `from_account_uuid`.
   - The `to_account_id` column has been replaced with `to_account_uuid`.
+- The `WalletRead` and `InputSource` impls for `WalletDb` now set the `AccountId`
+  associated type to `AccountUuid`.
+- Variants of `SqliteClientError` have changed:
+  - The `AccountCollision` and `ReachedGapLimit` now carry `AccountUuid` values
+    instead of `AccountId`s.
+  - `SqliteClientError::AccountIdDiscontinuity` has been removed as it is now
+    unused.
+  - `SqliteClientError::AccountIdOutOfRange` has been renamed to
+    `Zip32AccountIndexOutOfRange`.
 
 ### Removed
-- `zcash_client_sqlite::AccountId::{from_u32, as_u32}` (use `AccountUuid` and
-  its methods instead).
+- `zcash_client_sqlite::AccountId` (use `AccountUuid` instead).
 
 ## [0.13.0] - 2024-11-14
 
@@ -31,7 +37,7 @@ and this library adheres to Rust's notion of
 
 ### Changed
 - MSRV is now 1.77.0.
-- Migrated to `zcash_primitives 0.20`, `zcash_keys 0.5`, 
+- Migrated to `zcash_primitives 0.20`, `zcash_keys 0.5`,
   `zcash_client_backend 0.15`.
 - Migrated from `schemer` to our fork `schemerz`.
 - Migrated to `rusqlite 0.32`.
@@ -51,7 +57,7 @@ and this library adheres to Rust's notion of
   summary information is now only returned in the case that some note
   commitment tree size information can be determined, either from subtree root
   download or from downloaded block data. NOTE: The recovery progress ratio may
-  be present as `0:0` in the case that the recovery range contains no notes; 
+  be present as `0:0` in the case that the recovery range contains no notes;
   this was not adequately documented in the previous release.
 
 ## [0.12.0] - 2024-10-04
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -7,6 +7,13 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- The `v_transactions` view has been modified:
+  - The `account_id` column has been replaced with `account_uuid`.
+- The `v_tx_outputs` view has been modified:
+  - The `from_account_id` column has been replaced with `from_account_uuid`.
+  - The `to_account_id` column has been replaced with `to_account_uuid`.
+
 ## [0.13.0] - 2024-11-14
 
 ### Added
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -7,12 +7,21 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+- `zcash_client_sqlite::AccountUuid`
+- `zcash_client_sqlite::Account::uuid`
+- `zcash_client_sqlite::WalletDb::get_account_for_uuid`
+
 ### Changed
 - The `v_transactions` view has been modified:
   - The `account_id` column has been replaced with `account_uuid`.
 - The `v_tx_outputs` view has been modified:
   - The `from_account_id` column has been replaced with `from_account_uuid`.
   - The `to_account_id` column has been replaced with `to_account_uuid`.
+
+### Removed
+- `zcash_client_sqlite::AccountId::{from_u32, as_u32}` (use `AccountUuid` and
+  its methods instead).
 
 ## [0.13.0] - 2024-11-14
 
--- a/zcash_client_sqlite/src/error.rs
+++ b/zcash_client_sqlite/src/error.rs
@@ -11,8 +11,7 @@ use zcash_keys::keys::AddressGenerationError;
 use zcash_primitives::zip32;
 use zcash_primitives::{consensus::BlockHeight, transaction::components::amount::BalanceError};
 
-use crate::wallet::commitment_tree;
-use crate::AccountId;
+use crate::{wallet::commitment_tree, AccountUuid};
 
 #[cfg(feature = "transparent-inputs")]
 use {
@@ -79,7 +78,7 @@ pub enum SqliteClientError {
 
     /// The account being added collides with an existing account in the wallet with the given ID.
     /// The collision can be on the seed and ZIP-32 account index, or a shared FVK component.
-    AccountCollision(AccountId),
+    AccountCollision(AccountUuid),
 
     /// The account was imported, and ZIP-32 derivation information is not known for it.
     UnknownZip32Derivation,
@@ -90,12 +89,8 @@ pub enum SqliteClientError {
     /// An error occurred while processing an account due to a failure in deriving the account's keys.
     BadAccountData(String),
 
-    /// A caller attempted to initialize the accounts table with a discontinuous
-    /// set of account identifiers.
-    AccountIdDiscontinuity,
-
-    /// A caller attempted to construct a new account with an invalid account identifier.
-    AccountIdOutOfRange,
+    /// A caller attempted to construct a new account with an invalid ZIP 32 account identifier.
+    Zip32AccountIndexOutOfRange,
 
     /// The address associated with a record being inserted was not recognized as
     /// belonging to the wallet.
@@ -129,7 +124,7 @@ pub enum SqliteClientError {
     /// ephemeral address outputs have been mined. The parameters are the account id and
     /// the index that could not safely be reserved.
     #[cfg(feature = "transparent-inputs")]
-    ReachedGapLimit(AccountId, u32),
+    ReachedGapLimit(AccountUuid, u32),
 
     /// An ephemeral address would be reused. The parameters are the address in string
     /// form, and the txid of the earliest transaction in which it is known to have been
@@ -181,8 +176,7 @@ impl fmt::Display for SqliteClientError {
             SqliteClientError::UnknownZip32Derivation => write!(f, "ZIP-32 derivation information is not known for this account."),
             SqliteClientError::KeyDerivationError(acct_id) => write!(f, "Key derivation failed for account {}", u32::from(*acct_id)),
             SqliteClientError::BadAccountData(e) => write!(f, "Failed to add account: {}", e),
-            SqliteClientError::AccountIdDiscontinuity => write!(f, "Wallet account identifiers must be sequential."),
-            SqliteClientError::AccountIdOutOfRange => write!(f, "Wallet account identifiers must be less than 0x7FFFFFFF."),
+            SqliteClientError::Zip32AccountIndexOutOfRange => write!(f, "ZIP 32 account identifiers must be less than 0x7FFFFFFF."),
             SqliteClientError::AccountCollision(id) => write!(f, "An account corresponding to the data provided already exists in the wallet with internal identifier {}.", id.0),
             #[cfg(feature = "transparent-inputs")]
             SqliteClientError::AddressNotRecognized(_) => write!(f, "The address associated with a received txo is not identifiable as belonging to the wallet."),
--- a/zcash_client_sqlite/src/error.rs
+++ b/zcash_client_sqlite/src/error.rs
@@ -121,7 +121,7 @@ pub enum SqliteClientError {
     NoteFilterInvalid(NoteFilter),
 
     /// The proposal cannot be constructed until transactions with previously reserved
-    /// ephemeral address outputs have been mined. The parameters are the account id and
+    /// ephemeral address outputs have been mined. The parameters are the account UUID and
     /// the index that could not safely be reserved.
     #[cfg(feature = "transparent-inputs")]
     ReachedGapLimit(AccountUuid, u32),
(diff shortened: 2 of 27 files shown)


str4d:
I discovered that the Android SDK is misusing `v_tx_outputs` by putting `to_account_id` into a type that actually stores a ZIP 32 account index. These previously were indeed the same type, until #1235 was merged. I think we need to alter the views to expose the account UUIDs instead of the account IDs, particularly as after this PR the account IDs cannot be extracted from `AccountId` and are intended to be entirely internal.

zcash_client_sqlite/src/lib.rs
```diff
@@ -161,30 +162,48 @@ pub(crate) const UA_TRANSPARENT: bool = true;
...
         self.0
     }
 }
 
+/// The local identifier for an account.
+///
+/// This is an ephemeral value for efficiently and generically working with accounts in a
+/// [`WalletDb`]. To reference accounts in external contexts, use [`AccountUuid`].
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default)]
+pub struct AccountId(u32);
```
daira, line 205: Would it be useful to rename this to `AccountLocalId`? Doing so is obviously a more disruptive change, but it would force all code currently using `AccountId` to explicitly choose between `AccountUuid` and `AccountLocalId`. It would also, I think, reduce the chance of confusing these with ZIP 32 account indices.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -79,12 +80,12 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
...
     //             \                          \            |              /                       /
     //              ------------------------------ tx_retrieval_queue ----------------------------
-    //                                                     |
-    //                                            support_legacy_sqlite
-    //                                                     |
-    //                                         fix_broken_commitment_trees
-    //                                                     |
-    //                                          fix_bad_change_flagging
+    //                                               /           \
+    //                            support_legacy_sqlite         add_account_uuids
```
daira, line 84: Is there any need for this to be parallel with the branch up to `fix_bad_change_flagging`? I thought we decided that we would only use parallel migrations when we actually had a need for it, to simplify reasoning about the possible database states.

daira (approved):
utACK with comments.

zcash_client_sqlite/src/wallet/init/migrations.rs
str4d: It is explicitly required because the test for `fix_bad_change_flagging` creates an account, and the code now requires a `uuid` column. Preserving the old account creation code inside the test doesn't work in this case because it is indirectly used through the test harness. If we made `add_account_uuids` be a child of `fix_bad_change_flagging` then we could not run the test, because the test needs to initialise its state without `fix_bad_change_flagging` applied (which would then mean `add_account_uuids` is also not applied).

So yes, I would say that we have an actual need for parallel migrations here. (And also, we now have tracking of the leaf states as of various releases, so the possible database states are much easier to reason about now than they were when the recent global merge happened.)

zcash_client_sqlite/src/lib.rs
str4d: Such a rename serves little purpose, because it is named directly for the `WalletRead::AccountId` associated type that it gets set to, and the associated type is what actually gets used in most places.

I'm somewhat annoyed that we now have this split where our local account ID that needs to be in `WalletRead` is infallible concretely but fallible in the trait API, which was in part due to #1235 getting rushed into a crate release (something I expressed disagreement with at the time). As it is, we'd need Yet Another Large Refactor to fix it, and I explicitly do not want to do that at the same time as these changes; our changelogs get large enough as it is.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/c8aff5bd06ca7854175234dd88034ff392b36d23..bf42ec29a1792ce318bb38dd28c7731734de2b58) to replace `account_id` in the views with `account_uuid` (because `account_id` is now intended to be an internal-only value). cc @AArnott for awareness (though I've also documented it in the changelogs).

zcash_client_sqlite/src/lib.rs
daira: Confusion with the associated type is one reason I want to do this change, actually. It makes no sense to me to have the associated type be named the same as *one particular* possible instantiation of it. They are conceptually not the same thing.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ and this library adheres to Rust's notion of
...
+### Changed
+- The `v_transactions` view has been modified:
+  - The `account_id` column has been replaced with `account_uuid`.
+- The `v_tx_outputs` view has been modified:
+  - The `from_account_id` column has been replaced with `from_account_uuid`.
+  - The `to_account_id` column has been replaced with `to_account_uuid`.
+
+### Removed
+- `zcash_client_sqlite::AccountId::{from_u32, as_u32}` (use `AccountUuid` and
+  its methods instead).
```
daira, line 24: I'm not sure whether this is going to cause problems in the short term for clients that were using the account ids (e.g. to do a release that migrates to UUIDs for persisted data not directly controlled by `zcash_client_sqlite`). The alternative would be to deprecate these methods.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -779,8 +794,9 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
             ":hd_account_index": u32::from(account_index),
```
daira, line 794: I'm confused: `stmt` uses `:account_id` as a named parameter but the name used here is `:hd_account_index`. How was this working?

daira: Hmm. https://docs.rs/rusqlite/latest/rusqlite/trait.Params.html#named-parameters :
> Note: Unbound named parameters will be left to the value they previously were bound with, falling back to NULL for parameters which have never been bound.

zcash_client_sqlite/CHANGELOG.md
str4d: We added these behind an unstable feature flag in literally the last release. I do not believe anyone is actually depending on them yet. And anyone who is likely is also using SQL directly for the views, and can join on the `accounts` table for whatever migration they need.

str4d: We added these behind an unstable feature flag in literally the last release. I do not believe anyone is actually depending on them yet. And anyone who is likely is also using SQL directly for the views, and can join on the `accounts` table for whatever migration they need.

zcash_client_sqlite/src/wallet.rs
str4d: That's a bug, and we've had that kind of bug before. It would be nice to figure out a way to lint for this.

str4d: That's a bug, and we've had that kind of bug before. It would be nice to figure out a way to lint for this.

zcash_client_sqlite/CHANGELOG.md
daira: Ah I didn't realize they were unstable. That's fine then.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -767,7 +782,7 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
     account_index: zip32::AccountId,
 ) -> Result<Option<Account>, SqliteClientError> {
     let mut stmt = conn.prepare(
-        "SELECT id, ufvk
+        "SELECT id, uuid, ufvk
         FROM accounts
         WHERE hd_seed_fingerprint = :hd_seed_fingerprint
           AND hd_account_index = :account_id",
```
daira, line 788:
```suggestion
          AND hd_account_index = :hd_account_index",
```

zcash_client_sqlite/src/wallet.rs
daira: I'd settle for a runtime error. Substituting NULL is just terrible. :sob:

daira: I guess you could have a macro to generate a struct from the query string, and then you'd pass an instance of that struct rather than `named_params!`. That way the compiler would check whether you'd missed out fields or included undefined fields. Sounds like a lot of effort to get working, though, unless someone has already done something like it.

daira (approved):
utACK with this suggestion: https://github.com/zcash/librustzcash/pull/1631/files#r1855319372

zcash_client_sqlite/src/lib.rs
nuttycom: I have a different take here: `WalletRead::AccountId` should now be `AccountUuid`, and `AccountId` should be removed. The whole point is that the identifier is that there is a single, existentially typed identifier for the account. By adding `AccountUuid` instead of modifying `AccountId` to contain a UUID, you've created a confusing situation where there are two different ways to reference an account.

zcash_client_sqlite/src/lib.rs
```diff
@@ -219,6 +238,17 @@ pub struct WalletDb<C, P> {
     params: P,
 }
 
+impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters> WalletDb<C, P> {
+    /// Returns the account with the given "one-way stable" identifier, if it was created
+    /// by this wallet instance.
+    pub fn get_account_for_uuid(
```
nuttycom, line 244: While this approach is superficially less intrusive, I believe that it's a significantly worse API from the user perspective to have to retrieve a separate account identifier that they can't serialize every time they want to work with an account.

nuttycom (changes requested):
If we're going to make this change, we should do so wholesale: replace the contents of the `AccountId` type with the UUID, and change all the queries to take the UUID (and join to the `accounts` table if necessary).

zcash_client_sqlite/src/lib.rs
nuttycom: My preference is for the transition to instantiate `WalletRead::AccountId` as `AccountUuid` be part of this PR; that then obviates the need for a separate method to query for account data by UUID, and we don't really want anyone writing code that targets the intermediate API state that you're talking about.

str4d: @nuttycom Originally my plan _was_ to change `AccountId` to be the UUID-based thing (https://github.com/zcash/librustzcash/issues/1335#issuecomment-2477464017). But then you said the exact opposite of the above in https://github.com/zcash/librustzcash/issues/1335#issuecomment-2477484782: use two types. So that's what I did here.

zcash_client_sqlite/src/lib.rs
```diff
@@ -219,6 +238,17 @@ pub struct WalletDb<C, P> {
...
+        &self,
+        account_uuid: AccountUuid,
+    ) -> Result<Option<wallet::Account>, SqliteClientError> {
```
daira, line 247: This is returning (a Result of Option of) a [zcash_client_sqlite::wallet::Account](https://github.com/zcash/librustzcash/blob/2df44970b7510aa0b4d3e82578c374077e7f1f32/zcash_client_sqlite/src/wallet.rs#L199-L205). Why is it returning a `zcash_client_sqlite`-specific type rather than an associated type that impls `zcash_client_backend::data_api::Account`?

I think I'm with @nuttycom here: if for the `zcash_client_sqlite` backend `WalletRead::AccountId` is instantiated as `AccountUuid`, and `WalletRead::Account` is instantiated as `wallet::Account` (as it already is), then the existing `WalletRead::get_account` trait method *is* the required uuid → account mapping.

zcash_client_sqlite/src/lib.rs
daira: > > Augment WalletRead::AccountId with the ability to encode to, and parse from, opaque bytes. This encoding MUST NOT be parsed or otherwise interpreted by wallet apps.
>
> This doesn't need to be something that's exposed at the `zcash_client_backend` level, because the user of `zcash_client_sqlite` can use the knowledge that they're depending on that crate; they don't need this to be generic over data storage implementations.

I think this might be a miscommunication? I interpret it only as saying that the *serialization* shouldn't be exposed by `zcash_client_backend`. (FWIW I disagree and think that `WalletRead::AccountId` should be generically serializable, but it's only mild disagreement, because you should never mix serializations of it from different backends.)

> > Encode its account IDs (which are currently the account's primary key from the accounts table) as the tuple (wallet_uuid, account_primary_key).
>
> An alternative that we discussed to this was to assign a distinct UUID to each account, and map that to an account ID prior to performing queries. After some consideration, I think that I prefer this solution to the tuple-based approach, because when we add a persistent database reference, we can just have an in-memory map that acts as a cache of the `(account_uuid, account_primary_key)` mappings.

As I interpret this, it's primarily talking about the decision to use `AccountUuid` rather than `(WalletUuid, LocalAccountId)`. Using an in-memory map was secondary and was only one possibility.

zcash_client_sqlite/src/error.rs
```diff
@@ -129,7 +124,7 @@ pub enum SqliteClientError {
     /// ephemeral address outputs have been mined. The parameters are the account id and
```
daira, line 124:
```suggestion
    /// ephemeral address outputs have been mined. The parameters are the account UUID and
```

zcash_client_sqlite/src/error.rs
```diff
@@ -181,8 +176,7 @@ impl fmt::Display for SqliteClientError {
             SqliteClientError::UnknownZip32Derivation => write!(f, "ZIP-32 derivation information is not known for this account."),
             SqliteClientError::KeyDerivationError(acct_id) => write!(f, "Key derivation failed for account {}", u32::from(*acct_id)),
```
daira, line 177:
```suggestion
            SqliteClientError::KeyDerivationError(zip32_index) => write!(f, "Key derivation failed for ZIP 32 account index {}", u32::from(*zip32_index)),
```

zcash_client_sqlite/src/error.rs
```diff
@@ -181,8 +176,7 @@ impl fmt::Display for SqliteClientError {
...
             SqliteClientError::BadAccountData(e) => write!(f, "Failed to add account: {}", e),
-            SqliteClientError::AccountIdDiscontinuity => write!(f, "Wallet account identifiers must be sequential."),
-            SqliteClientError::AccountIdOutOfRange => write!(f, "Wallet account identifiers must be less than 0x7FFFFFFF."),
+            SqliteClientError::Zip32AccountIndexOutOfRange => write!(f, "ZIP 32 account identifiers must be less than 0x7FFFFFFF."),
             SqliteClientError::AccountCollision(id) => write!(f, "An account corresponding to the data provided already exists in the wallet with internal identifier {}.", id.0),
```
daira, line 180:
```suggestion
            SqliteClientError::AccountCollision(account_uuid) => write!(f, "An account corresponding to the data provided already exists in the wallet with UUID {:?}.", account_uuid.0),
```
(probably needs a rustfmt)

zcash_client_sqlite/src/lib.rs
```diff
@@ -818,7 +848,10 @@ impl<P: consensus::Parameters> WalletWrite for WalletDb<rusqlite::Connection, P>
                     )
                 })?;
             let account_index = wallet::max_zip32_account_index(wdb.conn.0, &seed_fingerprint)?
```
daira, line 850: Rename to `zip32_account_index`.

daira (approved):

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -324,12 +333,13 @@ pub(crate) fn mark_sapling_note_spent(
...
 /// - A transaction will not contain more than 2^63 shielded outputs.
 /// - A note value will never exceed 2^63 zatoshis.
-pub(crate) fn put_received_note<T: ReceivedSaplingOutput>(
+pub(crate) fn put_received_note<T: ReceivedSaplingOutput<AccountId = AccountUuid>>(
     conn: &Transaction,
     output: &T,
     tx_ref: TxRef,
     spent_in: Option<TxRef>,
 ) -> Result<(), SqliteClientError> {
+    let account_id = get_account_id(conn, output.account_id())?;
```
nuttycom, line 342:
```suggestion
    let account_id = get_account_ref(conn, output.account_id())?;
```

nuttycom (changes requested):
Additional changes & required accounts table functionality.

nuttycom (approved):
utACK 9bef4ee9ac1cc7395b2bcda58627f70cd18ba898

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3001,31 +2991,15 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
-        let detectable_via_scanning = d_tx.tx().sapling_bundle().is_some();
-        #[cfg(feature = "orchard")]
-        let detectable_via_scanning =
-            detectable_via_scanning | d_tx.tx().orchard_bundle().is_some();
-
-        if d_tx.mined_height().is_none() && !detectable_via_scanning {
-            queue_tx_retrieval(conn, std::iter::once(d_tx.tx().txid()), None)?;
-        }
-    }
+    queue_unmined_tx_retrieval(conn, &d_tx)?;
```
daira, line 3037: This is called whether or not the tx is mined. I suggest renaming it to something like `queue_tx_retrieval_if_needed`.

PR #1634, opened 2024-12-03
Adds zcash_client_memory crate and minor changes to zcash_client_backend to support

willemolding:
After a pretty big rewrite we have the memory wallet back-end in much better shape. In this large PR:

- Adds the `zcash_client_memory` crate
    - Passes all zcash_client_backend tests
    - Supports orchard pool and transparent inputs under different feature flag configurations
    - Supports building for Wasm (see https://github.com/ChainSafe/WebZjs)
    - Custom protobuf wire protocol for serialization of memory wallets with versioning
- Minor changes to `zcash_client_backend`
    - Adds some derives (PartialEq mostly) to types we used in our implementation so memory wallets can be compared with each other
    - Adds a `wasm-bindgen` feature flag which configures the time crate to support wasm builds
    - Adds an optional `finally` method to `WalletTest` which allows wallet backend developers to add an extra hook that runs after each test to tidy up. We use it to test roundtrip serialization of the wallet after each test

This is a very large PR but it is almost entirely additions but expecting this to be a difficult one to review. Thanks in advance for anyone who takes it on!


pacu:
There's no CHANGELOG.md update neither one is created for the new crate

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+}
+
+impl PartialEq for Account {
+    fn eq(&self, other: &Self) -> bool {
+        self.account_name == other.account_name
+            && self.account_id == other.account_id
+            && self.kind == other.kind
+            && self
+                .viewing_key
+                .encode(&zcash_primitives::consensus::MainNetwork)
```
pacu, line 230: [non-blocking] 
is this because `UnifiedFullViewingKey` does not implement Eq?  Isn't this a big performance hit in terms of what it seems to be an innocuous comparison that covertly does two encoding operations?

@nuttycom what's the proper way of telling if two UFVKs are the same?

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+            .ok_or_else(|| {
+                Error::AddressGeneration(AddressGenerationError::ShieldedReceiverRequired)
+            })?;
+        let (ua, diversifier_index) = acc.default_address(ua_request)?;
+        acc.addresses.insert(diversifier_index, ua);
+        #[cfg(feature = "transparent-inputs")]
+        acc.reserve_until(0)?;
+        Ok(acc)
+    }
+
```
pacu, line 275: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+    pub fn addresses(&self) -> &BTreeMap<DiversifierIndex, UnifiedAddress> {
+        &self.addresses
+    }
+
+    pub fn select_receiving_address(
```
pacu, line 280: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+        Ok(self
+            .uivk()
+            .transparent()
+            .as_ref()
+            .map(|tivk| tivk.default_address()))
+    }
+}
+#[cfg(feature = "transparent-inputs")]
+impl Account {
+    pub fn ephemeral_addresses(
```
pacu, line 363: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+                    addr.address,
+                    TransparentAddressMetadata::new(
+                        TransparentKeyScope::EPHEMERAL,
+                        NonHardenedChildIndex::from_index(*idx).unwrap(),
+                    ),
+                )
+            })
+            .collect())
+    }
+    pub fn ephemeral_ivk(&self) -> Result<Option<EphemeralIvk>, Error> {
```
pacu, line 380: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+        self.viewing_key
+            .transparent()
+            .map(AccountPubKey::derive_ephemeral_ivk)
+            .transpose()
+            .map_err(Into::into)
+    }
+
+    pub fn first_unstored_index(&self) -> Result<u32, Error> {
```
pacu, line 388: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+                unreachable!("violates constraint index_range_and_address_nullity")
+            } else {
+                Ok(idx.checked_add(1).unwrap())
+            }
+        } else {
+            Ok(0)
+        }
+    }
+
+    pub fn first_unreserved_index(&self) -> Result<u32, Error> {
```
pacu, line 400: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+        self.first_unstored_index()?
+            .checked_sub(GAP_LIMIT)
+            .ok_or(Error::CorruptedData(
+                "ephemeral_addresses corrupted".to_owned(),
+            ))
+    }
+
+    pub fn reserve_until(
```
pacu, line 408: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+        Ok(Vec::new())
+    }
+
+    #[cfg(feature = "transparent-inputs")]
+    pub fn mark_ephemeral_address_as_used(
+        &mut self,
+        address: &TransparentAddress,
+        tx_id: TxId,
+    ) -> Result<(), Error> {
+        // TODO: ephemeral_address_reuse_check
```
pacu, line 456: please create a github issue for this TODO and refer to it

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+                        .unwrap()
+                        .map_err(Into::into)
+                })
+                .collect::<Result<Vec<_>, _>>();
+        }
+        Ok(Vec::new())
+    }
+
+    #[cfg(feature = "transparent-inputs")]
+    pub fn mark_ephemeral_address_as_used(
```
pacu, line 451: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+                let next_to_reserve = idx.checked_add(1).expect("ensured by constraint");
+                self.reserve_until(next_to_reserve)?;
+                return Ok(());
+            }
+        }
+        Ok(())
+    }
+
+    #[cfg(feature = "transparent-inputs")]
+    pub fn mark_ephemeral_address_as_seen(
```
pacu, line 471: public function without documentation

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+        &mut self,
+        // txns: &TransactionTable,
```
pacu, line 473: remove commented code

zcash_client_memory/src/types/account.rs
```diff
@@ -0,0 +1,780 @@
...
+        address: &TransparentAddress,
+        tx_id: TxId,
+    ) -> Result<(), Error> {
+        for (idx, addr) in self.ephemeral_addresses.iter_mut() {
+            if addr.address == *address {
+                // TODO: this
```
pacu, line 479: Please create an issue for this TODO item if this is something out of the scope of this PR

zcash_client_memory/src/types/notes/sent.rs
```diff
@@ -0,0 +1,371 @@
...
+    pub fn txid(&self) -> &TxId {
+        match self {
+            SentNoteId::Shielded(note_id) => note_id.txid(),
+            SentNoteId::Transparent { txid, .. } => txid,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct SentNoteTable(pub(crate) BTreeMap<SentNoteId, SentNote>);
```
pacu, line 44: Why is the other table not a BTreeMap?

zcash_client_memory/src/types/transaction.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+/// A table of received notes. Corresponds to sapling_received_notes and orchard_received_notes tables.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct TransactionEntry {
+    // created: String,
```
pacu, line 31: delete commented field

zcash_client_memory/src/types/transparent.rs
```diff
@@ -0,0 +1,210 @@
...
+    type Target = BTreeMap<OutPoint, TxId>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+// transparent_received_outputs
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ReceivedTransparentOutput {
```
pacu, line 88: public struct missing documentation

zcash_client_memory/src/types/transparent.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// The maximum block height at which this TXO was either
+    /// observed to be a member of the UTXO set at the start of the block, or observed
+    /// to be an output of a transaction mined in the block. This is intended to be used to
+    /// determine when the TXO is no longer a part of the UTXO set, in the case that the
+    /// transaction that spends it is not detected by the wallet.
+    pub(crate) max_observed_unspent_height: Option<BlockHeight>,
+}
+
+impl ReceivedTransparentOutput {
+    pub fn new(
```
pacu, line 106: missing documentation

zcash_client_memory/src/types/transparent.rs
```diff
@@ -0,0 +1,210 @@
...
+        Self {
+            transaction_id,
+            account_id,
+            address,
+            txout,
+            max_observed_unspent_height: Some(max_observed_unspent_height),
+        }
+    }
+
+    pub fn to_wallet_transparent_output(
```
pacu, line 122: missing documentation

zcash_client_memory/src/types/transparent.rs
```diff
@@ -0,0 +1,210 @@
...
+/// A cache of the relationship between a transaction and the prevout data of its
+/// transparent inputs.
+///
+/// Output may be attempted to be spent in multiple transactions, even though only one will ever be mined
+/// which is why can cannot just rely on TransparentReceivedOutputSpends or implement this as as map
+#[derive(Debug, Default, PartialEq)]
+pub struct TransparentSpendCache(pub(crate) BTreeSet<(TxId, OutPoint)>);
+
+impl TransparentSpendCache {
+    pub fn new() -> Self {
```
pacu, line 140: missing documentation

pacu (commented):
Left some more comments

zcash_client_backend/Cargo.toml
```diff
@@ -154,6 +154,8 @@ zcash_proofs = { workspace = true, features = ["bundled-prover"] }
 zcash_protocol = { workspace = true, features = ["local-consensus"] }
 
 [features]
+default = ["time/default"]
+
```
str4d, line 158: Is `time/default` actually used anywhere by this crate? If not, we can remove this in the breaking change release immediately after this PR merges (which could potentially be the release this goes into).

zcash_client_backend/Cargo.toml
```diff
@@ -247,6 +249,11 @@ unstable-serialization = ["dep:byteorder"]
 ## Exposes the [`data_api::scanning::spanning_tree`] module.
 unstable-spanning-tree = []
 
+## feature to allow building for wasm using wasm-bindgen
+wasm-bindgen = [
+    "time/wasm-bindgen",
+]
+
```
str4d, line 256: Is this necessary or only a convenience? AFAICT downstream WASM crates can themselves add a dependency on `time` with the necessary feature flag enabled, without us needing to re-export the feature flag. I'm reluctant to re-export uncommon feature flags from crates we do not control.
```suggestion
```

components/zcash_protocol/src/lib.rs
```diff
@@ -35,7 +35,7 @@ mod txid;
 pub use txid::TxId;
 
 /// A Zcash shielded transfer protocol.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
str4d, line 38: Addition should be documented in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -752,7 +752,7 @@ pub enum TransactionDataRequest {
 }
 
 /// Metadata about the status of a transaction obtained by inspecting the chain state.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq)]
```
str4d, line 755: Addition should be documented in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -1576,6 +1576,10 @@ pub trait WalletTest: InputSource + WalletRead {
         &self,
         protocol: ShieldedProtocol,
     ) -> Result<Vec<ReceivedNote<Self::NoteRef, Note>>, <Self as InputSource>::Error>;
+
+    /// Optionally perform final checks at the conclusion of each test
+    /// Allows wallet backend developers to perform any necessary consistency checks or cleanup
+    fn finally(&self) {}
```
str4d, line 1582: Addition should be documented in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -2019,7 +2023,7 @@ impl<AccountId> SentTransactionOutput<AccountId> {
 
 /// A data structure used to set the birthday height for an account, and ensure that the initial
 /// note commitment tree state is recorded at that height.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
str4d, line 2026: Addition should be documented in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -2056,9 +2060,8 @@ impl AccountBirthday {
     ///    exist no unscanned ranges between the wallet's birthday height and the provided
     ///    `recover_until` height, exclusive.
     ///
-    /// This API is intended primarily to be used in testing contexts; under normal circumstances,
+    /// This API is intended primarily to be used in testing contexts and deserialization; under normal circumstances,
```
str4d, line 2063: Why is this API needed for deserialization, rather than `AccountBirthday::from_treestate`?

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -499,7 +499,7 @@ impl ScanSummary {
 }
 
 /// The final note commitment tree state for each shielded pool, as of a particular block height.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
str4d, line 502: This is one of those kinds of structs where I'm not entirely convinced that full struct comparison is what we want, and instead `block_hash` equivalence should be the appropriate proxy. Technically all other fields in the struct are committed to by `block_hash`; if `a.block_hash == b.block_hash` but the rest of their fields do not match, that means (except with cryptographically negligible probability) that either `ChainState::empty` or `ChainState::new` was called with invalid data.

But the real question here is, for what purpose is `ChainState` equality needed? I'm guessing the motivation here is a `HashMap` or something, and there we care about plain struct equality, for which the derive is fine.

In [the example from the stdlib docs](https://doc.rust-lang.org/stable/std/cmp/trait.PartialEq.html#how-can-i-implement-partialeq), `PartialEq` is used in a way that means "I do not care whether the other fields match or not", whereas for our environment we _do_ care whether the other fields match. We don't want two chain states with the same block hash but different final tree states to be treated interchangeably (one of them is invalid!) but we also don't want to end up with the same block hash occurring in multiple places (i.e. keeping around invalid state alongside potentially-valid state).

All this is to say, this is why I tend to avoid adding `PartialEq` to structs unless we know we want to use them in that way. So my question here is, how are you using this `ChainState`?

Also, additions must be documented in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -26,7 +26,7 @@ use crate::fees::orchard as orchard_fees;
 use ::transparent::keys::{NonHardenedChildIndex, TransparentKeyScope};
 
 /// A unique identifier for a shielded transaction output
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
str4d, line 29: Addition should be documented in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -67,7 +67,7 @@ impl NoteId {
 /// * for wallet-internal outputs, the internal account ID and metadata about the note.
 /// * if the `transparent-inputs` feature is enabled, for ephemeral transparent outputs, the
 ///   internal account ID and metadata about the outpoint;
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
str4d, line 70: Addition should be documented in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -89,6 +89,7 @@ pub enum Recipient<AccountId> {
 /// The shielded subset of a [`Transaction`]'s data that is relevant to a particular wallet.
 ///
 /// [`Transaction`]: zcash_primitives::transaction::Transaction
+#[derive(Clone)]
```
str4d, line 92: Addition should be documented in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -218,6 +219,7 @@ impl transparent_fees::InputView for WalletTransparentOutput {
 }
 
 /// A reference to a spent note belonging to the wallet within a transaction.
+#[derive(Clone)]
```
str4d, line 222: Addition should be documented in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -257,6 +259,7 @@ pub type WalletSaplingSpend<AccountId> = WalletSpend<sapling::Nullifier, Account
 pub type WalletOrchardSpend<AccountId> = WalletSpend<orchard::note::Nullifier, AccountId>;
 
 /// An output that was successfully decrypted in the process of wallet scanning.
+#[derive(Clone)]
```
str4d, line 262: Addition should be documented in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -2114,6 +2117,11 @@ impl AccountBirthday {
         self.recover_until
     }
 
+    /// Returns the [`ChainState`] corresponding to the last block prior to the wallet's birthday
+    pub fn prior_chain_state(&self) -> &ChainState {
```
str4d, line 2121: Why does this need to be exposed?

If kept: addition should be documented in the changelog.

zcash_keys/src/address.rs
```diff
@@ -176,7 +176,7 @@ impl UnifiedAddress {
         &self.unknown
     }
 
-    fn to_address(&self, net: NetworkType) -> ZcashAddress {
+    pub fn to_address(&self, net: NetworkType) -> ZcashAddress {
```
str4d, line 179: - Needs documentation now that this is in the public API.
- Addition should be documented in the changelog.

zcash_transparent/src/keys.rs
```diff
@@ -67,7 +67,7 @@ impl From<TransparentKeyScope> for ChildNumber {
 /// A child index for a derived transparent address.
 ///
 /// Only NON-hardened derivation is supported.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd)]
```
str4d, line 70: - Why is this needed?
- Addition should be documented in the changelog.
- This should derive `Ord` as well if we're deriving `PartialOrd`.

str4d (changes requested):
Partial review of 991cfd0460f554d8e29fde32c1101e96a1e09112. I reviewed everything outside of the `zcash_client_memory` crate.

Changelog entry comments are non-blocking; they can get out of sync rather quickly, so instead for this PR we should just note everything that needs changelog entries (like I've done) and then after this merges make a follow-up PR with the changelog entries.

zcash_client_memory/src/input_source.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    /// Returns the list of spendable transparent outputs received by this wallet at `address`
+    /// such that, at height `target_height`:
+    /// * the transaction that produced the output had or will have at least `min_confirmations`
+    ///   confirmations; and
+    /// * the output is unspent as of the current chain tip.
+    ///
+    /// An output that is potentially spent by an unmined transaction in the mempool is excluded
+    /// iff the spending transaction will not be expired at `target_height`.
```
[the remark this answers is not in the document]
zookoatshieldedlabs: I'll bet they mean "iff" as a short-hand for "if and only if".

NerdGGuy:
What's outstanding on this?

PR #1635, opened 2024-12-04
zcash_client_backend PCZT support

nuttycom:


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/c89eed5b95ff39038fa446e0a8d0e6a153a995e4..1764631e54dcd37d3cdde67c8edc055d99c266d1) to rebase on the `pczt-format` branch and fix the UFVK replacement commit to generate the necessary transparent spending keys.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/1764631e54dcd37d3cdde67c8edc055d99c266d1..125353018b6592958a6bb57762c4ff5284e341fa) to store ZIP 32 derivation paths for spent notes/coins in the PCZT when building for an account that has persisted derivation metadata.

pczt/src/roles/creator/mod.rs
```diff
@@ -106,6 +106,8 @@ impl Creator {
             | zcash_primitives::transaction::TxVersion::Overwinter => None,
             zcash_primitives::transaction::TxVersion::Sapling => Some(SAPLING_TX_VERSION),
             zcash_primitives::transaction::TxVersion::Zip225 => Some(V5_TX_VERSION),
+            #[cfg(zcash_unstable = "zfuture")]
+            zcash_primitives::transaction::TxVersion::ZFuture => None,
```
daira, line 110: As an aside, I don't think it makes sense to name transaction versions after NUs or ZIPs. With the exception of `TxVersion::ZFuture`, they are just numbers.

zcash_client_backend/CHANGELOG.md
```diff
@@ -8,20 +8,29 @@ and this library adheres to Rust's notion of
...
     These allow the wallet backend to store additional metadata that is useful
     to applications managing these accounts.
-- `zcash_client_backend::data_api::AccountSource`:
-  - Both the `Derived` and `Importants` now have an additional `key_source`
-    property that is used to convey application-specific key source metadata.
-  - The `Copy` impl for this type has been removed.
-- `zcash_client_backend::data_api::Account` has an additional `name` method
-  that returns the human-readable name of the account, if any.
+  - `AccountSource`:
+    - Both the `Derived` and `Importants` now have an additional `key_source`
```
daira, line 25: This will conflict with https://github.com/zcash/librustzcash/pull/1570/commits/3f802121e7b8d7a8cf607ab1fcd4947024c91cc4 btw.

zcash_client_backend/Cargo.toml
```diff
@@ -170,6 +172,22 @@ transparent-inputs = [
 ## Enables receiving and spending Orchard funds.
 orchard = ["dep:orchard", "dep:pasta_curves", "zcash_keys/orchard"]
 
+## Enables creating partially-signed transactions for use in hardware wallet and multisig scenarios.
```
daira, line 175:
```suggestion
## Enables creating partially-constructed transactions for use in hardware wallet and multisig scenarios.
```

zcash_client_backend/src/data_api/error.rs
```diff
@@ -79,11 +86,9 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError, C
     /// Attempted to create a spend to an unsupported Unified Address receiver
     NoSupportedReceivers(Box<UnifiedAddress>),
 
-    /// A proposed transaction cannot be built because it requires spending an input
-    /// for which no spending key is available.
-    ///
-    /// The argument is the address of the note or UTXO being spent.
-    NoSpendingKey(String),
+    /// A proposed transaction cannot be built because it requires spending an input of a type for
+    /// which a key that is required to construct the transaciton is not available.
```
daira, line 90:
```suggestion
    /// A proposed transaction cannot be built because it requires spending an input
    /// of a type for which a key required to construct the transaction is not available.
```

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1092,6 +1092,64 @@ where
         )
     }
 
+    /// Invokes [`create_proposed_transaction_pczt`] with the given arguments.
+    ///
+    /// [`create_proposed_transaction_pczt`]: super::wallet::create_proposed_transaction_pczt
+    #[cfg(feature = "pczt")]
+    #[allow(clippy::type_complexity)]
+    pub fn create_proposed_transaction_pczt<InputsErrT, FeeRuleT, ChangeErrT>(
```
daira, line 1100: "transaction_pczt" is redundant; consider renaming to `create_proposed_pczt`.

zcash_client_backend/CHANGELOG.md
```diff
@@ -8,20 +8,29 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Added
-- `zcash_client_backend::data_api::AccountSource::key_derivation`
+- `zcash_client_backend::data_api`
+  - `AccountSource::key_derivation`
+  - `error::PcztError`
+  - `wallet::ExtractErrT`
+  - `wallet::create_proposed_transaction_pczt`
```
daira, line 15: "transaction_pczt" is redundant; consider renaming to `create_proposed_pczt`.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3070,3 +3095,116 @@ pub fn metadata_queries_exclude_unwanted_notes<T: ShieldedPoolTester, DSF, TC>(
...
+{
+    use zcash_protocol::consensus::ZIP212_GRACE_PERIOD;
+
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(ds_factory)
+        .with_block_cache(cache)
+        .with_initial_chain_state(|_, network| {
+            // Initialize the chain state to after ZIP 212 became enforced.
+            let birthday_height =
+                network.activation_height(NetworkUpgrade::Nu5).unwrap() + ZIP212_GRACE_PERIOD;
```
daira, line 3115: The grace period was after Canopy, not NU5. Do you mean this?
```suggestion
            let birthday_height = std::cmp::max(
                network.activation_height(NetworkUpgrade::Nu5).unwrap(),
                network.activation_height(NetworkUpgrade::Canopy).unwrap() + ZIP212_GRACE_PERIOD,
            );
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3070,3 +3095,116 @@ pub fn metadata_queries_exclude_unwanted_notes<T: ShieldedPoolTester, DSF, TC>(
...
+    P0::apply_signatures_to_pczt(&mut signer, account.usk()).unwrap();
+    let pczt_authorized = signer.finish();
+
+    // Now we can extract the transaction.
+    let extract_and_store_result = st.extract_and_store_transaction_from_pczt(pczt_authorized);
+    assert_matches!(&extract_and_store_result, Ok(_));
+    let txid = extract_and_store_result.unwrap();
+
+    let (h, _) = st.generate_next_block_including(txid);
+    st.scan_cached_blocks(h, 1);
```
daira, line 3209: Do we want to check that the transaction includes the intended inputs and outputs?

zcash_client_backend/CHANGELOG.md
nuttycom:
```suggestion
    - Both the `Derived` and `Imported` now have an additional `key_source`
```

zcash_client_backend/src/data_api/error.rs
str4d: This error is being generated based on what kinds of keys are present in a USK; thus the "type" of the input here is really the "pool" of the input. In particular, no consideration is being given here to transparent multisig (a use case we do not currently support) where we might have some but not all of the necessary keys, because we'd never have all of the keys in the same USK (that would defeat the point).

zcash_client_backend/src/data_api/testing.rs
str4d: A more accurate name would be `create_proposed_transaction_via_pczt`. The PCZT is the mechanism by which the proposed transaction is being created.

zcash_client_backend/src/data_api/testing/pool.rs
str4d: Initially I just activated NU5, but that didn't work because the NUs are too close together in regtest mode (they are actually activated at the same height). What matters here is that a) Orchard is available (so the test is general) which requires NU5, and b) ZIP 212 is enforced, which requires Canopy + `ZIP212_GRACE_PERIOD`. The current code ensures both hold at once.

pczt/src/roles/creator/mod.rs
nuttycom: Agreed, we can fix `SAPLING_TX_VERSION` later.

zcash_client_backend/src/data_api/testing.rs
daira:
```suggestion
    pub fn create_pczt_from_proposal<InputsErrT, FeeRuleT, ChangeErrT>(
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -627,13 +732,28 @@ where
     #[cfg(all(feature = "transparent-inputs", feature = "orchard"))]
     let has_shielded_inputs = !(sapling_inputs.is_empty() && orchard_inputs.is_empty());
 
-    for (sapling_key, sapling_note, merkle_path) in sapling_inputs.into_iter() {
-        builder.add_sapling_spend(&sapling_key, sapling_note.clone(), merkle_path)?;
+    for (_sapling_key_scope, sapling_note, merkle_path) in sapling_inputs.into_iter() {
+        let key = match _sapling_key_scope {
```
daira, line 736:
```suggestion
    for (sapling_key_scope, sapling_note, merkle_path) in sapling_inputs.into_iter() {
        let key = match sapling_key_scope {
```

daira (approved):
utACK 388a64a475fdc625ee07384c52605a91304fdb53

str4d (dismissed):
utACK 0f4e01cd4a2e38bc8037aecc35ae901a7d6a005c. AFAICT the diff from @daira's review is just the rebase (it's not clear why GitHub doesn't recognise that the first two commits have already been merged, unless @nuttycom didn't alter this PR's target branch before merging the other): https://github.com/zcash/librustzcash/compare/388a64a475fdc625ee07384c52605a91304fdb53..0f4e01cd4a2e38bc8037aecc35ae901a7d6a005c

str4d (approved):
Re-utACK the diff from @daira's review to 0f4e01cd4a2e38bc8037aecc35ae901a7d6a005c (IDK why GitHub isn't retargeting correctly).

PR #1636, opened 2024-12-05
zcash_client_backend: Add optional derivation metadata when importing UFVKs with spending purpose.

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -150,28 +152,36 @@ fn parse_account_source(
...
-    match (account_kind, hd_seed_fingerprint, hd_account_index) {
-        (0, Some(seed_fp), Some(account_index)) => Ok(AccountSource::Derived {
-            seed_fingerprint: SeedFingerprint::from_bytes(seed_fp),
-            account_index: zip32::AccountId::try_from(account_index).map_err(|_| {
-                SqliteClientError::CorruptedData(
-                    "ZIP-32 account ID from wallet DB is out of range.".to_string(),
-                )
-            })?,
+    let derivation = hd_seed_fingerprint
+        .zip(hd_account_index)
```
str4d, line 156: I believe here we are now relying exclusively on the table constraint to ensure that these are both either `Some` or `None`, rather than catching a mixed-state corruption below.

str4d: Gah, I had a nice long comment here about how this now isn't as thoroughly checking for corrupt data, then a few lines below gave me the impression that we were instead relying on the table constraint so I deleted my comment, but now I see that the table constraint has been modified to remove that check. So yeah, now if one or other of these table columns gets corrupted and nulled, we will ignore whichever column is still non-null instead of detecting the corruption.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -41,18 +41,16 @@ CREATE TABLE "accounts" (
...
         AND ufvk IS NOT NULL
-        )
-        OR
-        (
+      )
+      OR
+      (
         account_kind = 1
-        AND hd_seed_fingerprint IS NULL
-        AND hd_account_index IS NULL
```
str4d, line 54: This is the removed constraint that now permits data corruption to go undetected (because we aren't requiring the columns to be either both null, or both not null).

zcash_client_sqlite/src/wallet/init/migrations/add_account_uuids.rs
```diff
@@ -77,8 +79,6 @@ impl RusqliteMigration for Migration {
                   OR
                   (
                     account_kind = {account_kind_imported}
-                    AND hd_seed_fingerprint IS NULL
-                    AND hd_account_index IS NULL
```
str4d, line 81: Document that this migration is now not just creating account UUIDs, but also allowing the imported account kind to have derivation metadata.

str4d (approved):
utACK 7fa87b57cfca82e946e005861c08d409957839fc with a small concern.

zcash_client_sqlite/src/wallet.rs
nuttycom: Yeah, I thought about adding a check to ensure that the nullability of the two columns was equal, but TBH I think that the parsing catching it is adequate.

str4d: The parsing doesn't catch it though. `Option::zip` is an AND; it maps mixed-optionals to `None`. The previous explicit matching on the two columns did catch it because we only "parsed" `(Some(_), Some(_))` or `(None, None`).

nuttycom: It doesn't catch explicit corruption, but what I mean is that it provides the desired interface. I can change it to ensure that they match though.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/7fa87b57cfca82e946e005861c08d409957839fc..f1724c436a99208e612d8dec2119613fd280492d) to address comments from review.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/7fa87b57cfca82e946e005861c08d409957839fc..cc2dfbf) to address comments from review.

str4d (approved):
utACK cc2dfbf7bf1f40a4d9d1aeb731537568bab61164

daira (commented):
Post-hoc ACK

Issue #1490, opened 2024-08-10
Cache incoming viewing keys and detect IVK collisions when adding accounts


daira:
In #1489 we added a check for collisions between FVK components when adding accounts.

As part of supporting Incoming accounts, we should:
* Use a migration to add `orchard_ivk_item_cache`, `sapling_ivk_item_cache`, and `p2pkh_ivk_item_cache` to the `addresses` table in addition to the existing `orchard_fvk_item_cache`, `sapling_fvk_item_cache`, and `p2pkh_fvk_item_cache`, and make all of these UNIQUE.
* Change the FVK collision check to an IVK collision check. (If two accounts collide on FVK then they necessarily also collide on IVK, and there is negligible chance that two different FVKs produce colliding IVKs.)

daira:
#1638 added similar logic to that needed to fix this issue.

PR #1638, opened 2024-12-07
zcash_keys: Make address generation use all items of the source key by default.

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -288,22 +288,39 @@ pub(crate) fn seed_matches_derived_account<P: consensus::Parameters>(
...
-                        == uivk.default_address(ua_request)?)
-                },
-            )?,
-        };
+    let uivk_match = {
+        let usk = UnifiedSpendingKey::from_seed(params, &seed.expose_secret()[..], account_index)
+            .map_err(|_| SqliteClientError::KeyDerivationError(account_index))?;
+
+        let (seed_addr, _) = usk.to_unified_full_viewing_key().default_address(Some(
+            UnifiedAddressRequest::all().expect("At least one supported pool feature ie enabled."),
```
daira, line 296:
```suggestion
            UnifiedAddressRequest::all().expect("At least one supported pool feature is enabled."),
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -288,22 +288,39 @@ pub(crate) fn seed_matches_derived_account<P: consensus::Parameters>(
 
     // Keys are not comparable with `Eq`, but addresses are, so we derive what should
     // be equivalent addresses for each key and use those to check for key equality.
```
daira, line 290: This comment doesn't make sense, I think. This function is supposed to check that the seed, fingerprint, and UIVK are consistent with each other. It has no other seed to compare against.
```suggestion
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -288,22 +288,39 @@ pub(crate) fn seed_matches_derived_account<P: consensus::Parameters>(
...
+            == Some(true);
+
+        let p2pkh_match = seed_addr
+            .transparent()
+            .zip(uivk_addr.transparent())
+            .map(|(a, b)| a == b)
+            == Some(true);
+
+        orchard_match || sapling_match || p2pkh_match
+    };
```
daira, line 323: This is pretty much the same logic needed to fix #1490 (see the TODO in `add_account`). Does it make sense to fix that at the same time?

zcash_keys/src/keys.rs
```diff
@@ -890,14 +890,15 @@ impl UnifiedFullViewingKey {
         self.orchard.as_ref()
     }
 
-    /// Attempts to derive the Unified Address for the given diversifier index and
-    /// receiver types.
+    /// Attempts to derive the Unified Address for the given diversifier index and receiver types.
+    /// If `request` is None, the address should be derived to contain a receiver for each item in
+    /// this UFVK.
```
daira, line 895: Add this to each of the other relevant methods.

zcash_keys/src/keys.rs
```diff
@@ -1506,7 +1509,10 @@ mod tests {
             }
 
             let ua = ufvk
-                .address(d_idx, UnifiedAddressRequest::unsafe_new(false, true, true))
+                .address(
+                    d_idx,
+                    Some(UnifiedAddressRequest::unsafe_new(false, true, true)),
```
daira, line 1514: Huh, I didn't realize that these test vectors don't include Orchard. Non-blocking.

daira (approved):
utACK. All comments are non-blocking.

daira (approved):
utACK

PR #1641, opened 2024-12-09
zcash_client_sqlite: Fix `add_account_uuids` migration

str4d:
`v_tx_outputs` gains new JOINs on `accounts.id` in order to migrate to exposing `accounts.uuid`, but as there are two table joins the tables are given aliases. The aliases were not being used in the JOIN's ON condition, leading to an error when querying the view.


daira (commented):
Post-hoc ACK

PR #1645, opened 2024-12-10
zcash_client_sqlite: Fix accounts joins in add_account_uuids migration

str4d:
Some views had account ID fields that were potentially NULL. As part of migrating to account UUIDs, we now join on the accounts table to access the UUID. We need to use a `LEFT JOIN` specifically to ensure that we propagate a NULL account ID into a NULL account UUID, instead of dropping the row.


nuttycom (approved):
utACK

PR #1649, opened 2024-12-11
Do not add key derivation metadata for dummy spends.

nuttycom:


str4d (approved):
utACK 702acff15593e6a3532436691205d633c28767fe

PR #1650, opened 2024-12-13
pczt: Add a Verifier role

str4d:
This isn't a real role per se; it's instead a way for accessing the parsed protocol-specific bundles for individual access and verification.

### DIFF
--- a/pczt/src/roles/verifier/mod.rs
+++ b/pczt/src/roles/verifier/mod.rs
@@ -1,0 +1,32 @@
+use crate::Pczt;
+
+#[cfg(feature = "orchard")]
+mod orchard;
+#[cfg(feature = "orchard")]
+pub use orchard::OrchardError;
+
+#[cfg(feature = "sapling")]
+mod sapling;
+#[cfg(feature = "sapling")]
+pub use sapling::SaplingError;
+
+#[cfg(feature = "transparent")]
+mod transparent;
+#[cfg(feature = "transparent")]
+pub use transparent::TransparentError;
+
+pub struct Verifier {
+    pczt: Pczt,
+}
+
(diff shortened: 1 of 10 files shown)


pczt/src/roles/verifier/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+#[cfg(feature = "transparent")]
+pub use transparent::TransparentError;
+
+pub struct Verifier {
+    pczt: Pczt,
+}
+
+impl Verifier {
+    /// Instantiates the Verifier role with the given PCZT.
+    pub fn new(pczt: Pczt) -> Self {
```
nuttycom, line 24: I'm surprised that this takes ownership of the value, instead of just taking a reference. Is this just for symmetry with how the other roles are implemented?

nuttycom (approved):
utACK f97f6fa8b3aa74dc02aae827ab5a417ed3b78c2c

pczt/src/roles/verifier/mod.rs
daira: If it takes ownership, you can no longer use a value that has failed to verify, which seems desirable. (You could copy the value beforehand, but it still seems as though this API reduces the chance of making a class of errors.)

daira (approved):
utACK

pczt/src/roles/verifier/mod.rs
str4d: The main reason is that the `into_parsed` methods consume the unparsed bundles, so if we took a reference here then we'd be internally cloning (and the caller can instead externally clone). The "lose access to the PCZT on detection of error" is a nice side-effect.

PR #1652, opened 2024-12-13
zcash_encoding: Only require `alloc` instead of `std`

str4d:


nuttycom (approved):
utACK 4d301c33ae58b17c8c76cdf91376472b8fe619e8 (in pairing)

nuttycom (approved):
utACK 79aae51974fe9faac06faf8044535a9e4f6d554c

PR #1654, opened 2024-12-13
pczt: Add output field for storing the user-facing address

str4d:
This is necessary in order for Signers to display the address encoding that a user is expecting to confirm.


nuttycom (approved):
utACK 0f1137b98b61003ddef4702569af8e0d11265a43 (in pairing)

nuttycom (approved):
utACK 3d1e32211daa1c3bbf65fc1099b74eb87eea980d

PR #1657, opened 2024-12-14
Move transparent protocol to `zcash_transparent` crate

str4d:


zcash_transparent/src/bundle.rs
```diff
@@ -188,11 +182,11 @@ impl OutPoint {
 
     /// Returns the byte representation of the txid of the transaction containing this `OutPoint`.
     pub fn hash(&self) -> &[u8; 32] {
-        self.hash.as_ref()
+        &self.hash
     }
 
     /// Returns the txid of the transaction containing this `OutPoint`.
-    pub fn txid(&self) -> &TxId {
+    pub fn txid(&self) -> &[u8; 32] {
```
str4d, line 189: I think this is the only change to the public API in this PR. A "txid" doesn't make sense in a protocol-specific crate, so I didn't want to move `TxId` into `zcash_transparent`. We could alter `OutPoint` to take a generic parameter, but that would result in _more_ changes to the public API.

zcash_transparent/src/builder.rs
```diff
@@ -323,12 +316,15 @@ impl TransparentAuthorizingContext for Unauthorized {
...
+    pub fn apply_signatures<F>(
         self,
-        #[cfg(feature = "transparent-inputs")] mtx: &TransactionData<tx::Unauthorized>,
-        #[cfg(feature = "transparent-inputs")] txid_parts_cache: &TxDigests<Blake2bHash>,
+        calculate_sighash: F,
         signing_set: &TransparentSigningSet,
-    ) -> Result<Bundle<Authorized>, Error> {
+    ) -> Result<Bundle<Authorized>, Error>
+    where
+        F: Fn(&SighashType, usize, &Script, &Script, NonNegativeAmount) -> [u8; 32],
```
nuttycom, line 326: Nonblocking nit: the type of this closure isn't very informative; could the argument to it be a struct so that the individual parts can be documented more easily? This is effectively how the `SignableInput::Transparent` variant used to work; it's basically just pulling out that variant as a struct up in `zcash_transparent`.

zcash_transparent/src/bundle.rs
nuttycom: I have an alternative suggestion: pull `TxId` up into `zcash_protocol`.

nuttycom (approved):
utACK with API suggestions (pulling TxId up into `zcash_protocol` is nonblocking but strongly recommended.)

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/98a06b3b793c172152a6c28907118f908280ac0e..ac7b66889a7076eab0fcc68cc6f8a050aea35b73) to address review comments.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/98a06b3b793c172152a6c28907118f908280ac0e..1767240be80532064a4a1883f79fc493354ef547) to address review comments.

nuttycom (dismissed):
utACK 1767240be80532064a4a1883f79fc493354ef547, looks great!

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/1767240be80532064a4a1883f79fc493354ef547..c85b12ce6500f27f94eed08fe49b441ddf6a8ad2) to fix Clippy complaints.

nuttycom (dismissed):
utACK c85b12ce6500f27f94eed08fe49b441ddf6a8ad2

nuttycom (approved):
re-utACK e4915f99c44e5b1651535441deba58d4df4e70c7

nuttycom (approved):
re-utACK https://github.com/zcash/librustzcash/pull/1657/commits/b1962ac396ca5e66d3ba2664ef99f45a32f3910b

PR #1663, opened 2024-12-17
pczt: Add no-std support

str4d:


nuttycom (approved):
utACK d11d428edf9f4c7a232dea616bfcb5e88f416879

PR #1672, opened 2024-12-20
Deprecate reexports of types that have been moved to other crates.

nuttycom:
The first commit of this PR removes the `pub use` reexports of types that have been moved to other crates and replaces the internal usage of these types with the types from their new crates. The second commit reintroduces those `pub use` reexports with `deprecated` pragmas.

### DIFF
--- a/zcash_primitives/src/transaction/mod.rs
+++ b/zcash_primitives/src/transaction/mod.rs
@@ -18,7 +18,7 @@ use std::fmt::Debug;
 use std::io::{self, Read, Write};
 use std::ops::Deref;
 
-use ::sapling::{self, builder as sapling_builder};
+use ::sapling::builder as sapling_builder;
 use ::transparent::bundle::{self as transparent, OutPoint, TxIn, TxOut};
 use zcash_encoding::{CompactSize, Vector};
 use zcash_protocol::{
--- a/zcash_primitives/src/transaction/mod.rs
+++ b/zcash_primitives/src/transaction/mod.rs
@@ -17,20 +17,19 @@ use std::convert::TryFrom;
 use std::fmt::Debug;
 use std::io::{self, Read, Write};
 use std::ops::Deref;
-use zcash_encoding::{CompactSize, Vector};
 
-use crate::{
+use ::sapling::{self, builder as sapling_builder};
+use ::transparent::bundle::{self as transparent, OutPoint, TxIn, TxOut};
+use zcash_encoding::{CompactSize, Vector};
+use zcash_protocol::{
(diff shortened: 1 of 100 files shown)


zcash_primitives/src/transaction/mod.rs
```diff
@@ -17,20 +17,19 @@ use std::convert::TryFrom;
 use std::fmt::Debug;
 use std::io::{self, Read, Write};
 use std::ops::Deref;
-use zcash_encoding::{CompactSize, Vector};
 
-use crate::{
+use ::sapling::{self, builder as sapling_builder};
```
str4d, line 21:
```suggestion
use ::sapling::builder as sapling_builder;
```
Or we could just use `sapling::builder` everywhere now, like we do `orchard::builder`.

zcash_primitives/src/transaction/sighash.rs
```diff
@@ -4,12 +4,24 @@ use super::{
     sighash_v4::v4_signature_hash, sighash_v5::v5_signature_hash, Authorization, TransactionData,
     TxDigests, TxVersion,
 };
-use crate::sapling::{self, bundle::GrothProofBytes};
+use ::sapling::{self, bundle::GrothProofBytes};
```
str4d, line 7:
```suggestion
use ::sapling::bundle::GrothProofBytes;
```

zcash_primitives/src/transaction/sighash_v4.rs
```diff
@@ -1,21 +1,19 @@
...
 use ff::PrimeField;
 
-use crate::{
-    consensus::BranchId,
-    sapling::{
-        self,
-        bundle::{GrothProofBytes, OutputDescription, SpendDescription},
-    },
+use ::sapling::{
+    self,
```
str4d, line 5:
```suggestion
```

zcash_primitives/src/transaction/txid.rs
```diff
@@ -5,21 +5,20 @@ use std::io::Write;
...
     consensus::{BlockHeight, BranchId},
-    sapling::{
-        self,
-        bundle::{OutputDescription, SpendDescription},
-    },
+    value::ZatBalance,
+};
+
+use ::sapling::{
+    self,
```
str4d, line 17:
```suggestion
```

str4d (dismissed):
utACK 4bdf39cb8f1428329b0ac5beb28caa2a271bff07 with nits (that will likely require a `cargo fmt`).

str4d (approved):
utACK 185538522c8a29878484e39f49450650d433b203

PR #1678, opened 2024-12-31
Fix crate vulnerabilities

rex4539:
https://rustsec.org/advisories/RUSTSEC-2024-0402
https://rustsec.org/advisories/RUSTSEC-2024-0376

Also updated `bytes`, `futures-util` and `libc` that were yanked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -610,9 +610,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.6.0"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


str4d:
Thanks for the PR! However, this kind of update does not fix any vulnerabilities.
- The only published artifacts from this repo are library crates; as such, the `Cargo.lock` file in this workspace is not used by any end binary, and exists only for testing MSRV in CI.
- The two linked advisories do not affect us:
  - We do not use borsh serialization.
  - We do not use `tonic::transport::Server` (we use `tonic` as a client).

We also don't update `Cargo.lock` in this repo to adapt to crates being yanked, unless those yanks affect APIs we are directly using. In those cases, we'd update our `Cargo.toml` to require the fixed version so that we can guarantee our own uses are safe against whatever the reason for the yank was. AFAICT the crates updated here are transitive dependencies, so we don't have any control over whether the relevant APIs are used (and it's up to the downstream binary usages of our libraries to decide on their dependencies).

PR #1679, opened 2025-01-09
CI: Check workflows and actions with zizmor

str4d:


str4d:
I can't figure out how to get a working CI with `${{ steps.prepare.outputs.feature-flags }}` expansion routed through an env variable. It seems that single quotes within an env variable are not preserved. For now I'm leaving those unaddressed.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,31 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["*"]
```
daira, line 7:
```suggestion
    branches: ["**"]
```
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
> `*`: Matches zero or more characters, but does not match the `/` character. For example, `Octo*` matches `Octocat`.
> `**`: Matches zero or more of any character.

daira (approved):
utACK with suggestion.

nuttycom (approved):
utACK

PR #1681, opened 2025-01-17
zcash_client_backend: Remove lightwalletd workaround

str4d:
The bug was fixed in zcash/lightwalletd@b805382ac34e91678ed0d06e9c5f84187ac630cc, and the workaround doesn't work with backed by zebrad.


nuttycom (approved):
utACK 2d59ef0c2ca7d96720ee413ffc848844b498a81e

PR #1688, opened 2025-01-28
`zcash_primitives`: Add `no_std` support

nuttycom:
This depends upon https://github.com/zcash/orchard/pull/453 and https://github.com/zcash/sapling-crypto/pull/152

Subsumes #1683

### DIFF
--- a/zcash_primitives/Cargo.toml
+++ b/zcash_primitives/Cargo.toml
@@ -44,7 +44,7 @@ subtle.workspace = true
 
 # - Shielded protocols
 ff.workspace = true
-group.workspace = true # = { workspace = true, features = ["wnaf-memuse"] }
+group.workspace = true
 jubjub.workspace = true
 nonempty.workspace = true
 orchard.workspace = true
--- a/zcash_primitives/Cargo.toml
+++ b/zcash_primitives/Cargo.toml
@@ -21,7 +21,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 equihash.workspace = true
 zcash_address.workspace = true
-zcash_encoding = { workspace = true, features = ["std"] }
+zcash_encoding.workspace = true
 zcash_protocol.workspace = true
 zip32.workspace = true
 
@@ -37,17 +37,17 @@ sha2.workspace = true
 
 # - Logging and metrics
 memuse.workspace = true
-tracing.workspace = true
+tracing = { workspace = true, default-features = false }
 
 # - Secret management
 subtle.workspace = true
 
 # - Shielded protocols
 ff.workspace = true
-group = { workspace = true, features = ["wnaf-memuse"] }
+group.workspace = true # = { workspace = true, features = ["wnaf-memuse"] }
 jubjub.workspace = true
 nonempty.workspace = true
-orchard = { workspace = true, features = ["circuit"] }
+orchard.workspace = true
 sapling.workspace = true
 zcash_spec.workspace = true
 
@@ -72,22 +72,23 @@ secp256k1 = { workspace = true, optional = true }
 getset.workspace = true
 
 # - Documentation
-document-features.workspace = true
+document-features = { workspace = true, optional = true }
 
 # - Encodings
 bs58.workspace = true
-byteorder.workspace = true
 hex.workspace = true
 
 # - Shielded protocols
-redjubjub = "0.7"
+redjubjub.workspace = true
 
 # - Transparent protocol
 ripemd.workspace = true
 
 # - ZIP 32
-aes.workspace = true
 fpe.workspace = true
+
+# No-std support
+core2.workspace = true
 
 [dependencies.zcash_note_encryption]
 workspace = true
@@ -100,6 +101,7 @@ incrementalmerkletree = { workspace = true, features = ["legacy-api", "test-depe
 proptest.workspace = true
 assert_matches.workspace = true
 rand_xorshift.workspace = true
+transparent = { workspace = true, features = ["test-dependencies"] }
 sapling = { workspace = true, features = ["test-dependencies"] }
 orchard = { workspace = true, features = ["test-dependencies"] }
 zcash_protocol = { workspace = true, features = ["test-dependencies"] }
@@ -108,7 +110,11 @@ zcash_protocol = { workspace = true, features = ["test-dependencies"] }
 pprof = { version = "0.13", features = ["criterion", "flamegraph"] }
 
 [features]
-default = ["multicore"]
+default = ["multicore", "std", "circuits"]
+std = ["document-features", "redjubjub/std"]
+
+## Enables creating proofs
+circuits = ["orchard/circuit", "sapling/circuit"]
 
 ## Enables multithreading support for creating proofs.
 multicore = ["orchard/multicore", "sapling/multicore"]
@@ -127,6 +133,7 @@ temporary-zcashd = []
 ## Exposes APIs that are useful for testing, such as `proptest` strategies.
 test-dependencies = [
     "dep:proptest",
+    "incrementalmerkletree/test-dependencies",
     "orchard/test-dependencies",
     "sapling/test-dependencies",
     "transparent/test-dependencies",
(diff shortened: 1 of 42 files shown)


zcash_primitives/Cargo.toml
```diff
@@ -37,17 +37,17 @@ sha2.workspace = true
...
-tracing.workspace = true
+tracing = { workspace = true, default-features = false }
 
 # - Secret management
 subtle.workspace = true
 
 # - Shielded protocols
 ff.workspace = true
-group = { workspace = true, features = ["wnaf-memuse"] }
+group.workspace = true # = { workspace = true, features = ["wnaf-memuse"] }
```
daira, line 47:
```suggestion
group.workspace = true
```

zcash_primitives/src/merkle_tree.rs
```diff
@@ -62,41 +64,48 @@ impl HashSer for MerkleHashOrchard {
...
 pub fn read_position<R: Read>(mut reader: R) -> io::Result<Position> {
-    reader.read_u64::<LittleEndian>().map(Position::from)
+    reader.read_u64_le().map(Position::from)
 }
 
 pub fn write_address<W: Write>(mut writer: W, addr: Address) -> io::Result<()> {
-    writer.write_u8(addr.level().into())?;
-    writer.write_u64::<LittleEndian>(addr.index())
+    writer.write_all(&[addr.level().into()])?;
+    writer.write_all(&addr.index().to_le_bytes())
```
daira, line 103: I checked that `Address::index` returns a `u64`.

zcash_primitives/src/merkle_tree.rs
```diff
@@ -266,25 +275,22 @@ pub fn merkle_path_from_slice<Node: HashSer, const DEPTH: u8>(
...
     // Read the position from the witness
-    let position = witness.read_u64::<LittleEndian>().map(Position::from)?;
+    let position = witness.read_u64_le().map(Position::from)?;
 
     // The witness should be empty now; if it wasn't, the caller would
     // have provided more information than they should have, indicating
     // a bug downstream
     if witness.is_empty() {
-        let path_len = auth_path.len();
+        let _path_len = auth_path.len();
```
daira, line 289:
```suggestion
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -398,9 +405,10 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, ()> {
     /// sent represents the total steps completed so far. It will eventually send number
     /// of spends + outputs. If there's an error building the transaction, the channel is
     /// closed.
+    #[cfg(feature = "std")]
     pub fn with_progress_notifier(
         self,
-        progress_notifier: Sender<Progress>,
+        _progress_notifier: Sender<Progress>,
```
daira, line 411: This is never unused.
```suggestion
        progress_notifier: Sender<Progress>,
```

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -1,8 +1,11 @@
...
-use std::io::{self, Read, Write};
+//! Functions for parsing & serialization of Orchard transaction components.
+use crate::encoding::ReadBytesExt;
 
-use byteorder::{ReadBytesExt, WriteBytesExt};
+use alloc::vec::Vec;
+use core::convert::TryFrom;
+use core2::io::{self, Read, Write};
+
+//use byteorder::{ReadBytesExt, WriteBytesExt};
```
daira, line 8:
```suggestion
```

zcash_primitives/src/transaction/components/sapling.rs
```diff
@@ -109,13 +110,13 @@ fn read_cmu<R: Read>(mut reader: R) -> io::Result<ExtractedNoteCommitment> {
...
 /// - Canonical encoding is enforced here
-pub fn read_base<R: Read>(mut reader: R, field: &str) -> io::Result<jubjub::Base> {
+pub fn read_base<R: Read>(mut reader: R, _field: &str) -> io::Result<jubjub::Base> {
     let mut f = [0u8; 32];
     reader.read_exact(&mut f)?;
     Option::from(jubjub::Base::from_repr(f)).ok_or_else(|| {
         io::Error::new(
             io::ErrorKind::InvalidInput,
-            format!("{} not in field", field),
+            "base value not a valid field element",
```
daira, line 119: Non-blocking: https://stackoverflow.com/questions/50200268/how-can-i-use-the-format-macro-in-a-no-std-environment has some alternatives to `format!` that can be used in a no_std environment.

zcash_primitives/src/transaction/builder.rs
nuttycom: This causes unused variable warnings, because `progress_notifier` is only used when the `std` feature is enabled.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -725,16 +728,13 @@ impl Transaction {
...
     fn read_v5_header_fragment<R: Read>(mut reader: R) -> io::Result<(BranchId, u32, BlockHeight)> {
-        let consensus_branch_id = reader.read_u32::<LittleEndian>().and_then(|value| {
-            BranchId::try_from(value).map_err(|e| {
-                io::Error::new(
-                    io::ErrorKind::InvalidInput,
-                    "invalid consensus branch id: ".to_owned() + e,
-                )
+        let consensus_branch_id = reader.read_u32_le().and_then(|value| {
+            BranchId::try_from(value).map_err(|_e| {
+                io::Error::new(io::ErrorKind::InvalidInput, "invalid consensus branch id")
```
daira, line 733: This one is particularly irritating, since converting a consensus branch id to a string should not even require allocation.

daira (approved):

zcash_primitives/src/transaction/components/sapling.rs
nuttycom: This change is because `core2::io::Error::new` takes a `&'static str` instead of a `String` here, and there's no way to format to a `&'static str`.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -411,7 +419,7 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, ()> {
             sapling_builder: self.sapling_builder,
             orchard_builder: self.orchard_builder,
             tze_builder: self.tze_builder,
-            progress_notifier,
+            _progress_notifier,
```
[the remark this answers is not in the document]
nuttycom: I think that the extra verbosity here isn't worth the argument name changing.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/4d0ce787fb055bf3e0c2f659dea65f396ddf9344..72d578f1013b93d40499046f3cea8101d6f5ad5e) to use versions of `orchard` and `sapling-crypto` that have been merged to their respective `main` branches.

zcash_primitives/src/transaction/mod.rs
daira: Actually this is not a regression because `BranchId::try_from` only returns an error with the fixed message "Unknown consensus branch ID". If we want to include the invalid id in the message, this works in no_std:

```suggestion
                let mut msg: [u8; 38] = *b"invalid consensus branch id 0x........";
                (&mut msg[30..38]).copy_from_slice(hex::encode(value.to_be_bytes()).as_bytes());
                io::Error::new(
                    io::ErrorKind::InvalidInput,
                    core::str::from_utf8(&msg).expect("valid utf8"),
                )
```
Non-blocking.

daira (approved):
utACK with non-blocking suggestion.

zcash_primitives/src/transaction/mod.rs
nuttycom: This still doesn't work because the argument must have `'static` lifetime.

daira (approved):
utACK d352b93e8538403fbc762828b7151cc90cceef7d

PR #1691, opened 2025-01-30
zcash_client_sqlite: Enable constructing `WalletDb` around references to `rusqlite::Connection`

str4d:
This enables downstream wallets such as Zallet to implement connection pooling, or other strategies to enable multi-threaded `WalletDb` access.


nuttycom (approved):
utACK bef5feef4ea141131946467ae70d34c58e397db9

Issue #1686, opened 2025-01-24
zcash_client_backend: Allow disabling Tor directory permissions tightening


str4d:
Arti does some custom permissions tightening on its Tor directory. We've been having some problems on mobile devices related to permissions, and we believe that it is related to this. Given that mobile OSs provide sandboxing, there's not a huge benefit to extra permissions shenanigans (unlike on a desktop or laptop). So we should alter the `create` method to allow disabling the permissions tightening.

PR #1697, opened 2025-02-07
zcash_client_backend: Allow disabling Tor directory permissions tightening

str4d:
Closes zcash/librustzcash#1686.

### DIFF
--- a/zcash_client_backend/src/tor.rs
+++ b/zcash_client_backend/src/tor.rs
@@ -24,20 +24,33 @@ impl Client {
     /// Preserving the contents of this directory will speed up subsequent calls to
     /// `Client::create`.
     ///
+    /// If `with_permissions` is `None`, the default from [`arti_client`] will be used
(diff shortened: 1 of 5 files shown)


zcash_client_backend/src/tor.rs
```diff
@@ -24,20 +24,33 @@ impl Client {
     /// Preserving the contents of this directory will speed up subsequent calls to
     /// `Client::create`.
     ///
+    /// If `with_permissions` is `None`, the default from [`arti_client`] will be used
+    /// (enable permissions checks unless the `ARTI_FS_DISABLE_PERMISSION_CHECKS` env
+    /// variable is set).
+    ///
```
daira, line 30: Perhaps include an example of using `with_permissions` other than `None`.

daira (approved):
utACK with suggestion.

zcash_client_backend/src/tor.rs
str4d: The type's methods have decent documentation, which is accessible through `rust-analyzer` in IDEs, so I think this should be fine as-is.

PR #1703, opened 2025-02-11
zcash_client_backend: Fix `tor::Client::create` argument

str4d:
The optional closure was requiring a type annotation when using `None`, which would create a more complex API than just passing an empty closure.

### DIFF
--- a/zcash_client_backend/src/tor.rs
+++ b/zcash_client_backend/src/tor.rs
@@ -24,9 +24,10 @@ impl Client {
     /// Preserving the contents of this directory will speed up subsequent calls to
     /// `Client::create`.
     ///
-    /// If the `with_permissions` closure does not make any changes, the default from
-    /// [`arti_client`] will be used (enable permissions checks unless the
-    /// `ARTI_FS_DISABLE_PERMISSION_CHECKS` env variable is set).
+    /// If the `with_permissions` closure does not make any changes (e.g. is
+    /// passed as `|_| {}`), the default from [`arti_client`] will be used.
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK b9fcefd057505610dc8b7ae37bcee210d153c394

zcash_client_backend/src/tor.rs
```diff
@@ -24,14 +24,14 @@ impl Client {
     /// Preserving the contents of this directory will speed up subsequent calls to
     /// `Client::create`.
     ///
-    /// If `with_permissions` is `None`, the default from [`arti_client`] will be used
-    /// (enable permissions checks unless the `ARTI_FS_DISABLE_PERMISSION_CHECKS` env
-    /// variable is set).
+    /// If the `with_permissions` closure does not make any changes, the default from
+    /// [`arti_client`] will be used (enable permissions checks unless the
+    /// `ARTI_FS_DISABLE_PERMISSION_CHECKS` env variable is set).
```
daira, line 29:
```suggestion
    /// If the `with_permissions` closure does not make any changes (e.g. is
    /// passed as `|_| {}`), the default from [`arti_client`] will be used.
    /// This default will enable permissions checks unless the
    /// `ARTI_FS_DISABLE_PERMISSION_CHECKS` env variable is set.
```

nuttycom (approved):
re-utACK https://github.com/zcash/librustzcash/pull/1703/commits/54c49a5c68c2a77b45b9d4a0896b682acae21d81

PR #1705, opened 2025-02-12
Add crate summaries to README

nuttycom:


nuttycom:
self-ACK; doc-only, summaries needed for librustzcash walkthrough presentation.

PR #1708, opened 2025-02-18
zcash_address: Add `unified::Item` trait to expose opaque typed encoding

str4d:


components/zcash_address/src/kind/unified.rs
```diff
@@ -446,3 +450,17 @@ pub trait Container {
     /// This API is for advanced usage; in most cases you should use `Self::items`.
     fn items_as_parsed(&self) -> &[Self::Item];
 }
+
+/// Trait for unified items, exposing specific methods on them.
+pub trait Item: SealedItem {
+    /// Returns the opaque typed encoding of this item.
+    ///
+    /// This is the same encoding used internally by [`Encoding::encode`].
```
daira, line 458:
```suggestion
    /// This is the same encoding used internally by [`Encoding::encode`].
    /// This API is for advanced usage; in most cases you should not depend
    /// on the raw encoding of items.
```

daira (approved):
utACK with non-blocking suggestion.

daira (approved):
utACK

nuttycom (approved):
utACK fb73395995a276f140de73189d6cb9a255a78368

PR #1714, opened 2025-02-24
deps: avoid duplicate crates

rex4539:


components/f4jumble/Cargo.toml
```diff
@@ -24,11 +24,11 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-blake2b_simd = { version = "1", default-features = false }
+blake2b_simd = { workspace = true }
```
str4d, line 27: This does not work because this crate's MSRV is older than when workspace support was added. Similarly for most of the rest of the changes in this PR.

PR #1730, opened 2025-03-04
Fix crate documentation

str4d:
This PR includes patch releases of `equihash` and `pczt`, for which their public documentation omits significant parts of their APIs (due to not being built with `--all-features`). The rest of the fixes don't appear to affect the public doc builds, they are just local issues I noticed while fixing the other two.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2911,7 +2911,7 @@ checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
 name = "pczt"
-version = "0.2.0"
+version = "0.2.1"
 dependencies = [
  "blake2b_simd",
  "bls12_381",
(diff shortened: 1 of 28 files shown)


daira (dismissed):
ACK, doc-only.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/0e247c6de8dfee198a9e2ceeaccd8eb4709da495..c1c020d7335c3ff745422afabc25a9665364c408) to fix intra-doc link failure (which I don't think would have broken docs.rs rendering, but I don't want to risk it).

nuttycom (approved):
doc-only ACK c1c020d7335c3ff745422afabc25a9665364c408

str4d:
Published:
- b649e544b5b773461e3c3a94d6bb95a9d26a88fe as `equihash-0.2.2`
- c1c020d7335c3ff745422afabc25a9665364c408 as `pczt-0.2.1`

PR #1732, opened 2025-03-05
zcash_client_sqlite: Migration fixes and API updates

str4d:


str4d:
This PR fixes the previous migration bug I found with a ZecPages wallet in https://github.com/zcash/wallet/pull/32#issuecomment-2697214094

However, I now see another one (after the migration thinks for a bit):
```
Failed to initialize zallet: An error occurred while applying migration c41dfc0e-e870-4859-be47-d2f572f5ca73 (Add support for general transparent gap limit handling, unifying the `addresses` and `ephemeral_addresses` tables.) up: Address generation error: KeyNotAvailable(P2pkh).
```

str4d:
I've confirmed I'm now able to migrate a Sapling-only wallet with this PR. The migration takes 15 seconds so it's not at all efficient for a database that contains no transparent addresses (but many Sapling notes).

nuttycom (approved):
utACK 4efef40246583a789db71e9105aca74cfd51af85

PR #1733, opened 2025-03-06
Implement send max feature

pacu:
This feature intends to provide a low level API to spend the max amount available on a given pool without pool crossing. 

This now builds atop https://github.com/zcash/librustzcash/pull/1888

complementary #1784 

This PR is best viewed with the "files" view. If you want to see how the featured evolved to what it is now, you can do a per-commit read but I wouldn't mind too much because we will squash this anyway.


components/zcash_protocol/src/value.rs
```diff
@@ -235,6 +235,22 @@ impl Mul<usize> for ZatBalance {
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Eq, Ord)]
 pub struct Zatoshis(u64);
 
+/// An intent of representing spendable value to reach a certain targeted
```
nuttycom, line 238: This type belongs in `zcash_client_backend`, not `zcash_protocol`.

nuttycom: Also its addition will need to be noted in the `zcash_client_backend` CHANGELOG

zcash_client_backend/CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- `zcash_client_backend::proposal::ProposalError`:
+  - `SendsMemoToTransparentRecipient`: signals an attempt to send a memo to a transparent recipient
+  - `ResultingPaymentInvalid(Zip321Error)`: details the ZIP-321 error that the proposal 
+  resulting causes.
+- `zcash_client_backend::data_api::wallet`:
+  - `SpendableNotes::total`: returns the sum of all spendable notes in `Zatoshis`
+  - `propose_send_max_transfer`: calls `propose_send_max` for a given `InputSelector`
```
nuttycom, line 15: The CHANGELOG entry for an addition should describe the purpose, not the implementation.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -341,6 +341,529 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    }
+
+    let network = *st.network();
+    assert_matches!(
+        decrypt_and_store_transaction(&network, st.wallet_mut(), &tx, None),
+        Ok(_)
+    );
+}
+
+// FIXME: Add a test that checks that Send Max to TEX address works as intended.
```
nuttycom, line 510: FIXME left in code.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -197,6 +202,43 @@ pub trait InputSelector {
     where
         ParamsT: consensus::Parameters,
         ChangeT: ChangeStrategy<MetaSource = Self::InputSource>;
+
+    /// Performs input selection and returns a proposal for the construction of a transaction
+    /// that sends the maximum amount possible from a given account to the specified recipient
+    /// ignoring notes that are below MARGINAL_FEE amount. This transaction will use all the
+    /// funds available minus the resulting fees that will vary according to ZIP-317 specifications.
```
nuttycom, line 209: The use of ZIP 317 is not determined at this layer; the fees are computed by the `change_strategy`.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -197,6 +202,43 @@ pub trait InputSelector {
...
+    ///
+    ///
+    /// Implementations should return the maximum possible number of economically useful inputs
+    /// required to supply at least the requested value, choosing only inputs received at the
+    /// specified source addresses. If insufficient funds are available to satisfy the required
+    /// outputs for the shielding request, this operation must fail and return
+    /// [`InputSelectorError::InsufficientFunds`].
```
nuttycom, line 216: This looks like copy-pasta? In any case it's not relevant documentation here.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -769,6 +811,164 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                .fee_required(
+                    params,
+                    target_height,
+                    [],
+                    [P2PKH_STANDARD_OUTPUT_SIZE],
+                    spendable_notes.sapling().len(),
+                    match spends_sapling > 0 && spends_orchard == 0 {
+                        true => 2,
+                        false => 0,
+                    },
```
nuttycom, line 923: Why is this written as a match rather than if/else? Also, I don't understand the `spends_orchard == 0` condition here. Finally, the hardcoded value `2` is probably not ideal; elsewhere we use the `BundleType::num_outputs` method to determine padding, for Sapling and Orchard independently. The same should probably be done here.

zcash_client_backend/src/fees.rs
```diff
@@ -488,7 +488,7 @@ impl EphemeralBalance {
 /// by a transaction having a specified set of inputs and outputs.
 pub trait ChangeStrategy {
     type FeeRule: FeeRule + Clone;
-    type Error;
+    type Error: From<<Self::FeeRule as FeeRule>::Error>;
```
nuttycom, line 491: This change needs to be noted in the CHANGELOG

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -769,6 +811,164 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                    )?;
+
+                    Ok((t0_fee + t1_fee).expect("fee is in range of valid Zatoshis values"))
+                }),
+        }
+        .map_err(|fee_error| {
+            InputSelectorError::Change(ChangeError::StrategyError(ChangeT::Error::from(fee_error)))
+        })?;
+
+        let amount = (input_total - fee_required).ok_or(BalanceError::Underflow)?;
```
nuttycom, line 948: This isn't an Underflow; this is `InsufficientFunds`.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -769,6 +811,164 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+
+        let payment = match zip321::Payment::new(recipient, amount, memo, None, None, vec![]) {
```
nuttycom, line 950: Firse, use `ok_or_else` instead of this `match`. Second, please open an issue to make `Payment::new` return a `Result` with an informative error type instead of just an `Option`.

zcash_client_backend/src/fees/orchard.rs
```diff
@@ -91,3 +91,9 @@ impl OutputView for Infallible {
         unreachable!()
     }
 }
+
+impl OutputView for Zatoshis {
```
nuttycom, line 95: Note this addition in the CHANGELOG

zcash_client_backend/src/fees/sapling.rs
```diff
@@ -111,3 +111,9 @@ impl OutputView for Infallible {
         unreachable!()
     }
 }
+
+impl OutputView for Zatoshis {
```
nuttycom, line 115: Note this addition in the CHANGELOG

zcash_client_backend/src/proposal.rs
```diff
@@ -48,6 +48,10 @@ pub enum ProposalError {
     PaymentPoolsMismatch,
     /// The proposal tried to spend a change output. Mark the `ChangeValue` as ephemeral if this is intended.
     SpendsChange(StepOutput),
+    /// The proposal attempts to send a memo to a transparent recipient.
+    SendsMemoToTransparentRecipient,
+    /// The proposal results in an invalid payment request.
+    ResultingPaymentInvalid(Zip321Error),
```
nuttycom, line 54:
```suggestion
    Zip321(Zip321Error),
```

We typically attempt to name error wrappers more like this.

zcash_client_backend/src/proposal.rs
```diff
@@ -48,6 +48,10 @@ pub enum ProposalError {
     PaymentPoolsMismatch,
     /// The proposal tried to spend a change output. Mark the `ChangeValue` as ephemeral if this is intended.
     SpendsChange(StepOutput),
+    /// The proposal attempts to send a memo to a transparent recipient.
+    SendsMemoToTransparentRecipient,
```
nuttycom, line 52: I'm surprised that this error variant is needed; the newly added paths at the high level shouldn't add any new error cases that have to be propagated out.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -116,6 +116,163 @@ where
...
+    let (table_prefix, index_col, note_reconstruction_cols) = per_protocol_names(protocol);
+    if unscanned_tip_exists(conn, anchor_height, table_prefix)? {
+        return Ok(vec![]);
+    }
+
+    // The goal of this SQL statement is to select all the spendable notes above the
+    // MARGINAL_FEE (5000 zats).
+    // 1) Use a window function to create a view of all notes, ordered from oldest to
+    //    newest
+    //
```
nuttycom, line 191: This comment seems wrong; I don't see any windowing happening here?

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -116,6 +116,163 @@ where
...
+            // the wallet birthday can only be unknown if there are no accounts in the wallet; in
+            // such a case, the wallet has no notes to spend.
+            return Ok(vec![]);
+        }
+    };
+
+    let (table_prefix, index_col, note_reconstruction_cols) = per_protocol_names(protocol);
+    if unscanned_tip_exists(conn, anchor_height, table_prefix)? {
+        return Ok(vec![]);
+    }
```
nuttycom, line 185: This seems like the wrong place to be performing this check, but I think it's copy-pasta from one of the other existing implementations. It is definitely in the wrong place for the case where we change how spendability is checked to allow different accounts in the wallet to have different spendability based upon the state of the note commitment tree; please open an issue so that this doesn't get missed when we go to make these changes.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -116,6 +116,163 @@ where
...
+                 INNER JOIN transactions
+                    ON transactions.id_tx = {table_prefix}_received_notes.tx
+                 WHERE accounts.uuid = :account_uuid
+                 AND {table_prefix}_received_notes.account_id = accounts.id
+                 AND value > 5000 -- FIXME #1316, allow selection of dust inputs
+                 AND accounts.ufvk IS NOT NULL
+                 AND recipient_key_scope IS NOT NULL
+                 AND nf IS NOT NULL
+                 AND commitment_tree_position IS NOT NULL
+                 AND transactions.block <= :anchor_height
```
nuttycom, line 212: It occurs to me that there's a race condition here: if someone makes a withdrawal from an exchange to the wallet, and then attempts to send max, the maximum balance of the wallet will not be sent but it would still allow the transaction to be constructed; the notes received in the last 10 blocks would simply not contribute to the balance. This is probably not what the user is intending; we should require that all notes apart from dust be outside the reorg window in order to allow sending max.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -116,6 +116,163 @@ where
...
+        .iter()
+        .filter_map(|ReceivedNoteId(p, n)| {
+            if *p == protocol {
+                Some(Value::from(*n))
+            } else {
+                None
+            }
+        })
+        .collect();
+    let excluded_ptr = Rc::new(excluded);
```
nuttycom, line 250: Exclusions don't make sense when sending max. The query above and call below, and the API of this method, need to be modified so that they don't take exclusions into account.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -116,6 +116,163 @@ where
...
+
+/// Selects the minimum set of spendable notes whose sum will be equal or greater that the
+/// specified ``target_value`` in Zatoshis from the specified shielded protocols excluding
+/// the ones present in the ``exclude`` slice.
+///
+/// - Implementation details
+///   - Notes with individual value *below* the ``MARGINAL_FEE`` will be ignored
+///   - Note spendability is determined using the `anchor_height`
+#[allow(clippy::too_many_arguments)]
+fn select_minimum_spendable_notes<P: consensus::Parameters, F, Note>(
```
nuttycom, line 275: This should probably still just be called "select_spendable_notes" - it's not actually minimal in its determination in any sense.

nuttycom (changes requested):
A variety of changes, cleanup, and documentation are needed.

zcash_client_backend/src/data_api.rs
```diff
@@ -1201,7 +1222,7 @@ pub trait InputSource {
     fn select_spendable_notes(
         &self,
         account: Self::AccountId,
-        target_value: Zatoshis,
+        target_value: TargetValue,
```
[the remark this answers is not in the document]
pacu: Yes, I omitted anything related to TargetValue given there's another PR that it's just that change

zcash_client_backend/src/data_api.rs
```diff
@@ -748,6 +748,27 @@ pub struct SpendableNotes<NoteRef> {
     orchard: Vec<ReceivedNote<NoteRef, orchard::note::Note>>,
 }
 
+impl<R> SpendableNotes<R> {
+    pub fn total(&self) -> Result<Zatoshis, BalanceError> {
+        let total = self.sapling.iter().try_fold(Zatoshis::ZERO, |acc, n| {
+            (acc + Zatoshis::try_from(n.note().value().inner())
+                .map_err(|_| BalanceError::Overflow)?)
```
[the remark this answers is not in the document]
pacu: I've seen many `expect` and I've always wondered whether panicking is the correct approach for a library with such wide range of clients like librustzcash that goes from rust native clients to mobile wallets through FFI and WASM.

zcash_client_backend/src/fees.rs
pacu: I don't recall doing this change to be honest. Maybe clippy  or compiler suggestion? I'll dig in

nuttycom: This was one of the things we did in our pairing; it just needs to be noted in the CHANGELOG

zcash_client_sqlite/src/wallet/common.rs
pacu: ohhhhhhh nice catch!!! I'll try to build a test that breaks this and work on the fix from it.

zcash_client_sqlite/src/wallet/common.rs
pacu: uhmm. Yes... but! the idea is that `select_spendable_notes` is the umbrella function that receives the `TargetValue` type and then calls the appropriate function.

zcash_client_sqlite/src/wallet/common.rs
pacu: I understand note exclusions as arbitrary decisions of an advanced user (or a wallet feature that makes very complex use of funds on behalf of a less experienced user) to **not spending** specific notes from their wallets. 

I think users that exclude notes would expect that all mechanics of the wallet hold when they exclude notes from being spent.  "I'd like to use the wallet as always but with the special consideration of it not spending this particular note".

If we remove note exclusion from this API, then you have to calculate the fees on your own when you choose to exclude a note and that doesn't make much sense.

nuttycom: The purpose of note exclusion is somewhat different from what you're intuiting here; it's there so that when you need to select _additional_ notes in order to make a transaction have enough funds to pay the fees, you can exclude dust notes that may have been selected by a previous invocation of the function. But that doesn't apply here; we're selecting as many notes as we can. If you want to keep funds in your wallet, you should do so by just sending whatever fraction of the balance that represents, not by selecting individual notes. I guess that maybe one could use exclusions to try to select value while skipping over more recent notes (favor spending oldest or something) but that seems so complicated as to not be worth the effort relative to just spending more notes and requesting "fresh" change.

zcash_client_backend/src/data_api/wallet/input_selection.rs
pacu: Created the issue https://github.com/zcash/librustzcash/issues/1794

zcash_client_sqlite/src/wallet/common.rs
pacu: Besides that currently no dust notes are picked (it's not implemented), wouldn't the possibility to exclude dust from "send max" be desirable as well? 
I've been developing "send max" to have the same intent as a "target value" send, with the exception that we don't know the total because fees vary depending on the composition of the wallet.

zcash_client_sqlite/src/wallet/common.rs
pacu: https://github.com/zcash/librustzcash/issues/1796 and added it to the code

zcash_client_sqlite/src/wallet/common.rs
pacu: If I make this fail from within this statement, I won't be able to catch the specific failure will I? I think that this particular failure needs to have its own error so wallets can inform that to the user specifically

nuttycom: A reasonable approach would be, for send-max, instead of passing the anchor height in to the query, instead return the minimum possible anchor height _from_ the query. Then the wallet can effectively schedule the transaction for the target height that is appropriate.

pacu: I added a test that produces this scenario, but I'm struggling with the error types, I think that a Sqlite error is too low level and specific to this implementation of the backend but couldn't manage to juggle generics so that a more clear ProposalError is thrown instead.

zcash_client_sqlite/src/wallet/common.rs
pacu: I think this is the last (but not least) outstanding thing for this PR.  I could remove exclusions but it would make the API less graceful in terms of select_all_notes would't be in the same call hierarchy as select_spendable_notes. Thoughts?

pacu:
rebased to latest main

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +343,675 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+                found_sent_change_memo = true;
+            }
+            Some(m) if m == &Memo::Empty => {
+                found_sent_empty_memo = true;
+            }
+            Some(m) if m == &send_max_memo => {
+                found_sent_max_memo = true;
+            }
+            Some(other) => panic!("Unexpected memo value: {:?}", other),
+            None => panic!("Memo should not be stored as NULL"),
```
pacu, line 740: @nuttycom is there a way to exclude these tests from the codecov reports?

zcash_client_backend/CHANGELOG.md
```diff
@@ -9,7 +9,41 @@ represent the transitive `semver` implications of changes within the enclosing
...
+  - `SpendableNotes::total`: returns the sum of all spendable notes in `Zatoshis`
+  - `propose_send_max_transfer`: Selects transaction inputs, computes fees, and 
+    constructs a proposal for a transaction (or series of them) that would spend all
+    the available funds from the given spend pools.
+  - `input_selection::propose_send_max`: Performs input selection and returns 
+    a proposal for the construction of a transaction that sends the maximum 
+    amount possible from a given account to the specified recipient ignoring
+    notes that are below MARGINAL_FEE amount.
+  - `zcash_client_sqlite::error::SqliteClientError`:
+    - `IneligibleNotes`: used `spendable_notes` is called with `TargetValue::MaxSpensable`
```
daira:
```suggestion
    - `IneligibleNotes`: used `spendable_notes` is called with `TargetValue::MaxSpendable` 
``

File: zcash_client_backend/CHANGELOG.md [hunk 23]
```diff
@@ -9,7 +9,41 @@ represent the transitive `semver` implications of changes within the enclosing
...
+  - `Client::set_dormant`
+  - `DormantMode`
 
+### Changed
+- `zcash_client_backend::data_api`:
+  - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. 
+    Existing calls to this function that used `Zatoshis` now use 
+    `TargetValue::AtLeast(Zatoshis)`
+  - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. 
+    Existing calls to this function that used `Zatoshis` now use `TargetValue::MinValue(Zatoshis)`
```
daira, line 39: `TargetValue::MinValue` does not exist.

daira:
```suggestion
```
This is a duplicate of the previous point. `TargetValue::AtLeast(Zatoshis)` is the correct spelling.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +342,527 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    // The sent memo should be the empty memo for the sent output, and the
+    // change output's memo should be as specified.
+    let mut found_sent_change_memo = false;
+    let mut found_sent_empty_memo = false;
+    let mut found_sent_max_memo = false;
+    for sent_note_id in sent_note_ids {
+        match st
+            .wallet()
+            .get_memo(sent_note_id)
+            .expect("Note id is valid")
```
nuttycom:
```suggestion
            .expect("Memo retrieval should succeed")
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +342,527 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(dsf)
+        .with_block_cache(cache)
+        .with_account_from_sapling_activation(BlockHash([0; 32]))
+        .build();
+
+    let account = st.test_account().cloned().unwrap();
+    let dfvk = T::test_account_fvk(&st);
+
+    // Add funds to the wallet in a single note
```
nuttycom, line 643: It's fine to have this test have just a single source note, but I think that it would be good to extend this test to cover the case where there are several notes in the wallet. Also, as part of that, it would be good to check the corner case where you have one or more notes that have exactly the marginal fee value.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +342,527 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    let tx = st
+        .wallet()
+        .get_transaction(sent_tx_id)
+        .unwrap()
+        .expect("Created transaction was stored.");
+    let ufvks = [(account.id(), account.usk().to_unified_full_viewing_key())]
+        .into_iter()
+        .collect();
+    let d_tx = decrypt_transaction(st.network(), None, Some(h), &tx, &ufvks);
+    assert_eq!(T::decrypted_pool_outputs_count(&d_tx), 2);
```
nuttycom: nit: A comment here as to why we expect to observe two outputs (I presume the sent output + the change output?) would be helpful.

zcash_client_backend/src/data_api.rs
```diff
@@ -156,6 +156,7 @@ pub enum NullifierQuery {
 #[derive(Debug, Clone, Copy)]
```
nuttycom, line 160: Fix documentation

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -385,6 +385,52 @@ where
...
+/// Select transaction inputs, compute fees, and construct a proposal for a transaction or series
+/// of transactions that would spend all available funds from the given `spend_pool`s that can then
+/// be authorized and made ready for submission to the network with [`create_proposed_transactions`].
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::type_complexity)]
+pub fn propose_send_max_transfer<DbT, ParamsT, InputsT, ChangeT, CommitmentTreeErrT>(
+    wallet_db: &mut DbT,
+    params: &ParamsT,
+    spend_from_account: <DbT as InputSource>::AccountId,
+    spend_pool: &[ShieldedProtocol],
```
nuttycom: TODO: check that there are tests covering the multi-pool use case.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -385,6 +385,52 @@ where
...
+    <DbT as InputSource>::NoteRef: Copy + Eq + Ord,
+    ParamsT: consensus::Parameters + Clone,
+    InputsT: InputSelector<InputSource = DbT>,
+    ChangeT: ChangeStrategy<MetaSource = DbT>,
+{
+    let (target_height, anchor_height) = wallet_db
+        .get_target_and_anchor_heights(min_confirmations)
+        .map_err(|e| Error::from(InputSelectorError::DataSource(e)))?
+        .ok_or_else(|| Error::from(InputSelectorError::SyncRequired))?;
+
```
nuttycom, line 680: Instead of delegating the check to the input selector, verify that we're not attempting to send a memo to a transparent recipient here, as is done in `propose_standard_transfer_to_address`

zcash_client_backend/src/data_api.rs
```diff
@@ -156,6 +156,7 @@ pub enum NullifierQuery {
...
 pub enum TargetValue {
     AtLeast(Zatoshis),
+    MaxSpendable,
```
nuttycom: We should probably have a third variant here, `Everything`; right now `MaxSpendable` will ignore notes above the anchor height.

We also need to document the behavior with respect to the behavior if dust notes are present in the wallet.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -385,6 +385,52 @@ where
...
+/// of transactions that would spend all available funds from the given `spend_pool`s that can then
+/// be authorized and made ready for submission to the network with [`create_proposed_transactions`].
+#[allow(clippy::too_many_arguments)]
+#[allow(clippy::type_complexity)]
+pub fn propose_send_max_transfer<DbT, ParamsT, InputsT, ChangeT, CommitmentTreeErrT>(
+    wallet_db: &mut DbT,
+    params: &ParamsT,
+    spend_from_account: <DbT as InputSource>::AccountId,
+    spend_pool: &[ShieldedProtocol],
+    input_selector: &InputsT,
```
nuttycom: We could make do with just an `InputSource` here; we don't need an input selector because there's no variation in the algorithm when the goal is "send everything."

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -193,6 +198,39 @@ pub trait InputSelector {
...
+        InputSelectorError<
+            <Self::InputSource as InputSource>::Error,
+            Self::Error,
+            ChangeT::Error,
+            <Self::InputSource as InputSource>::NoteRef,
+        >,
+    >
+    where
+        ParamsT: consensus::Parameters,
+        ChangeT: ChangeStrategy<MetaSource = Self::InputSource>;
```
nuttycom:
```suggestion
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -764,6 +802,149 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+    >
+    where
+        ParamsT: consensus::Parameters,
+        Self::InputSource: InputSource,
+        ChangeT: ChangeStrategy<MetaSource = DbT>,
+    {
+        let spendable_notes = wallet_db
+            .select_spendable_notes(
+                source_account,
+                TargetValue::MaxSpendable,
```
nuttycom: Add a boolean `force_sweep` and choose between `MaxSpendable` and `Everything` based on that.

zcash_client_backend/src/data_api.rs
nuttycom: Also, `Everything` should require that the wallet is fully synced, i.e. there are no unscanned ranges.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -764,6 +802,149 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                anchor_height,
+                &[],
+            )
+            .map_err(InputSelectorError::DataSource)?;
+
+        let input_total = spendable_notes.total_value()?;
+
+        let sapling_output_count = {
+            let spends_sapling = !spendable_notes.sapling().is_empty();
+            let requested_sapling_outputs: usize = if spends_sapling { 2 } else { 0 };
```
nuttycom: We need to determine first whether there will be pool-crossing involved; determine where we are being asked to send the funds. This shouldn't be based on Sapling spends; it should depend on the address we're sending to.

Also, do we want to avoid pool-crossing if possible, i.e. if we get a UA as the recipient, do we want to split things such that Sapling inputs go to a single Sapling output, and Orchard inputs go to a single Orchard output, such that we don't reveal the value being transferred?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -764,6 +802,149 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+            ::sapling::builder::BundleType::DEFAULT
+                .num_outputs(spendable_notes.sapling.len(), requested_sapling_outputs)
+                .map_err(|s| InputSelectorError::Change(ChangeError::BundleError(s)))?
+        };
+
+        #[cfg(feature = "orchard")]
+        let orchard_action_count = {
+            let spends_orchard = !spendable_notes.orchard().is_empty();
+            let requested_orchard_actions: usize = if spends_orchard { 2 } else { 0 };
```
nuttycom: See above.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -764,6 +802,149 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                sapling_output_count,
+                orchard_action_count,
+            ),
+            Address::Unified(addr) => change_strategy.fee_rule().fee_required(
+                params,
+                target_height,
+                [],
+                if addr.has_transparent() {
+                    vec![P2PKH_STANDARD_OUTPUT_SIZE]
+                } else {
```
nuttycom: This is incorrect; we don't want to use a transparent receiver unless the UA has no shielded receivers.

zcash_client_backend/src/fees.rs
```diff
@@ -487,7 +487,7 @@ impl EphemeralBalance {
...
```
nuttycom, line 490: This might go away with the removal of the logic from `InputSelector`.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -176,8 +185,138 @@ where
...
+                 INNER JOIN transactions
+                    ON transactions.id_tx = {table_prefix}_received_notes.tx
+                 WHERE accounts.uuid = :account_uuid
+                 AND {table_prefix}_received_notes.account_id = accounts.id
+                 AND value > 5000 -- FIXME #1316, allow selection of dust inputs
+                 AND accounts.ufvk IS NOT NULL
+                 AND recipient_key_scope IS NOT NULL
+                 AND nf IS NOT NULL
+                 AND commitment_tree_position IS NOT NULL
+                 AND transactions.block <= :anchor_height
```
nuttycom: This should change for the `Everything` case.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -176,8 +185,138 @@ where
...
+                 AND {table_prefix}_received_notes.id NOT IN rarray(:exclude)
+                 AND {table_prefix}_received_notes.id NOT IN (
+                   SELECT {table_prefix}_received_note_id
+                   FROM {table_prefix}_received_note_spends
+                   JOIN transactions stx ON stx.id_tx = transaction_id
+                   WHERE stx.block IS NOT NULL -- the spending tx is mined
+                   OR stx.expiry_height IS NULL -- the spending tx will not expire
+                   OR stx.expiry_height > :anchor_height -- the spending tx is unexpired
+                 )
+                 AND NOT EXISTS (
```
nuttycom: This needs to be adapted for the `Everything` case; if `Everything` is chosen, there must be no unscanned ranges. But we should do that in a separate query earlier.

zcash_client_backend/src/data_api.rs
pacu: +1 to the `E_very_THING` variant 😁

zcash_client_backend/src/data_api/testing/pool.rs
pacu: I replicated existing logic so I assume that's because the tx will have at least 2 outputs even if this is a send max tx because how dummy notes work. Is that assumption correct? could you link the ZIPs that will let me verify I'm testing this specific aspect properly?

zcash_client_backend/src/data_api/wallet/input_selection.rs
pacu: Probably we are lacking some "send policy" features in order to have an API that can express that kind of granularity without being a total mess. I think that for the purpose of most users, "send max" is by default a cross-pool YOLO kind of intent. 

I'm not opposing to this change -I indeed agree with your question- but I'm raising a "scope creep" flag.

zcash_client_backend/src/data_api/wallet/input_selection.rs
pacu: good catch!

zcash_client_sqlite/src/wallet/common.rs
pacu: How do you suggest to handle these differences in the SQL query logic at the Rust level? shall we juggle those in the SQL query or shall we create a separate rust function?

nuttycom: I think that I'm going to implement this as a separate method & query entirely.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: I agree that we don't want scope creep to avoid pool-crossing, but the requested output count still needs to take into account whether or not the recipient address can actually receive Sapling. And regardless, for bundle purposes the bundle policy will handle padding, so this can be set to a single output.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -693,12 +694,220 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+    let fee_required = (tr0_fee + tr1_fee.unwrap_or(Zatoshis::ZERO))
+        .expect("fee value addition does not overflow");
+
+    let amount = (input_total - fee_required).ok_or(InputSelectorError::InsufficientFunds {
+        available: input_total,
+        required: fee_required,
+    })?;
+
+    #[cfg(feature = "transparent-inputs")]
+    let ephemeral_output_value =
```
pacu, line 861: I think this makes no sense. Thanks to the the map operation  not using the mapped value made me look more carefully.

The ephemeral output value of a TEX address should be the total amount of the send _minus_ the fees to create that first transaction. 

in this case for a mere coincidence the test _kind of_ adds up because we are using a single note of 100_000 zats

This probably should be just `amount` which is already subtracting the fee for the first transaction

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +342,512 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+                NonZeroU32::new(1).unwrap(),
+            )
+            .unwrap();
+
+        let steps: Vec<_> = proposal.steps().iter().cloned().collect();
+        assert_eq!(steps.len(), 2);
+
+        assert_eq!(steps[0].balance().fee_required(), expected_step0_fee);
+        assert_eq!(steps[1].balance().fee_required(), expected_step1_fee);
+        assert_eq!(
```
pacu: This assertion doesn't match the code the documentation of the code that it's testing. according to what's implemented the proposed change of a send max to a TEX address should have no other change than the transparent funds that it's about to send to the recipient through an ephemeral address

pacu: This was fixed on commit a5f7730a77bb558f8bb6e8376dc51dc854ec90b1

zcash_client_backend/src/data_api/testing/pool.rs
pacu: I added a test that sends max from a wallet constituted by 10 notes of 100_000 zats and 10 notes of 5000 zats. The thing is that the assertions test the current logic (ignore maginal fee denominated notes). This test should break the moment we decide how to proceed with this.

nuttycom: That's great, I like having a test that's fragile to this, because it makes certain that we'll consider the marginal value case correctly here!

pacu: The one that uses both pools will be some challenge given that these test suites are divided by pool. 🥲

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e921a792faad3094ccfa0a6242e2266e57edb675..b271a66cf92744f3e91c48b6d66fead854c5d616) to fix `cargo fmt` errors.

zcash_client_backend/src/data_api/wallet.rs
pacu: Done!

zcash_client_sqlite/src/wallet/common.rs
nuttycom: Let's keep exclusions in; we don't have to use them.

nuttycom:
Argh, this PR was closed when it should have been retargeted to `main`. I will fix it.

zcash_transparent/src/builder.rs
```diff
@@ -459,7 +459,7 @@ impl Bundle<Unauthorized> {
         self,
         calculate_sighash: F,
         secp_ctx: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
-    ) -> Result<TransparentSignatureContext<secp256k1::VerifyOnly>, Error>
+    ) -> Result<TransparentSignatureContext<'_, secp256k1::VerifyOnly>, Error>
```
pacu: @nuttycom do you know where this change comes from?

zcash_client_backend/CHANGELOG.md
```diff
@@ -9,6 +9,40 @@ represent the transitive `semver` implications of changes within the enclosing
 workspace.
 
 ## [Unreleased]
+### Added
+- `zcash_client_backend::fees`:
+  - `impl sapling::OutputView for Zatoshis`
+  - `impl orchard::OutputView for Zatoshis`
+- `zcash_client_backend::data_api`:
+  - `MaxSpendMode` enum has been added to specify what means to spend "All" funds when specifying
+  a  `TargetValue` (see Changed section for more details and context).
```
pacu:
```suggestion
    a  `TargetValue` (see Changed section for more details and context).
```

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1002,6 +1003,37 @@ where
         )
     }
 
+    /// Invokes [`propose_transfer`] with the given arguments.
+    #[allow(clippy::type_complexity)]
+    pub fn propose_send_max_transfer<FeeRuleT>(
+        &mut self,
+        spend_from_account: <DbT as InputSource>::AccountId,
+        fee_rule: &FeeRuleT,
+        to: Address,
```
nuttycom: This parameter might be better as a `ZcashAddress` for API similarity to proposing transfers, which uses ZIP321 requests; I think that we'll have to reencode as `ZcashAddress` for inclusion in the proposal anyway?

zcash_client_backend/src/data_api/wallet.rs
nuttycom: Bump, this is still incomplete.

zcash_client_backend/src/proposal.rs
```diff
@@ -191,6 +197,7 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
         let mut consumed_prior_inputs: BTreeSet<StepOutput> = BTreeSet::new();
 
         for (i, step) in steps.iter().enumerate() {
+            dbg!(i);
```
nuttycom:
```suggestion
```

zcash_client_backend/src/proposal.rs
```diff
@@ -243,6 +250,8 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
             }
         }
 
+        dbg!(&steps);
+
```
nuttycom:
```suggestion
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -180,27 +183,210 @@ pub(crate) fn select_spendable_notes<P: consensus::Parameters, F, Note>(
...
+        },
+    )?;
+
+    let trusted_height = target_height.saturating_sub(u32::from(confirmations_policy.trusted()));
+    let untrusted_height =
+        target_height.saturating_sub(u32::from(confirmations_policy.untrusted()));
+    // when attempting to select max spendable funds it's assumed that there are unconfirmed funds
+    // whereas when the caller is intending to spend `Everything` the presence of unmined or
+    // unconfirmed funds must cause a failure.
+    let attempts_max_spendable = matches!(max_spend_mode, MaxSpendMode::MaxSpendable);
```
nuttycom: I find this documentation confusing. Nothing is "assumed" when the caller specified `MaxSpendable` - also, I think the way that this condition is being used in the `||` clause below is wrong.
```suggestion
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -180,27 +183,210 @@ pub(crate) fn select_spendable_notes<P: consensus::Parameters, F, Note>(
...
+
+    row_results
+        .map(|t| match t? {
+            (Some(note), Some(shard_priority))
+                if shard_priority <= ScanPriority::Scanned
+                    && (attempts_max_spendable
+                        || note.mined_height().iter().any(|h| h <= &anchor_height)) =>
```
nuttycom: Bug: the `expect` below relies on the second half of the `or` condition, and this will cause a panic if `max_spend_mode == MaxSpendable` and the note is unmined.

nuttycom (changes requested):

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1704 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+
+    // Spendable balance matches total balance
+    assert_eq!(
+        st.get_total_balance(account.id()),
+        Zatoshis::const_from_u64(120_000)
+    );
+    assert_eq!(
+        st.get_spendable_balance(account.id(), confirmation_policy),
+        value
+    );
```
nuttycom, line 502: This check looks to me like it should fail; the confirmations policy requires two confirmations, but the second 60000 ZAT chunk was received in the block at the chain tip, which only has one confirmation.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1704 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+
+    // The range of address indices that are safe to reserve now is
+    // 0..(gap_limits.ephemeral() * 2 - 1)`, and we have already reserved or used
+    // `gap_limits.ephemeral() + 1`, addresses, so trying to reserve another
+    // `gap_limits.ephemeral()` should fail.
+    reservation_should_fail(&mut st, gap_limits.ephemeral(), gap_limits.ephemeral() * 2);
+    reservation_should_succeed(&mut st, gap_limits.ephemeral() - 1);
+    // Now we've reserved everything we can, we can't reserve one more
+    reservation_should_fail(&mut st, 1, gap_limits.ephemeral() * 2);
```
nuttycom: @pacu why is this test checking all this ephemeral address gap limit stuff? It shouldn't be needed for what we want to test here.

pacu: oh I just replicated the send to text but with max

nuttycom: Removed in https://github.com/zcash/librustzcash/pull/1733/commits/0ad99b2815d605dc0537af102214f5600b7caacf

zcash_client_backend/CHANGELOG.md
```diff
@@ -9,6 +9,40 @@ represent the transitive `semver` implications of changes within the enclosing
...
+      `spendable_notes` is called with `TargetValue::MaxSpendable`
+      and there are funds that haven't been confirmed and all spendable notes
+      can't be selected.
+
+### Changed
+- `zcash_client_backend::data_api`:
+  - `TargetValue` has a new `AllFunds(MaxSpendMode)` variant to choose whether to consider "All"
+  funds being the `MaxSpendable` amount that it is currently in the wallet (there could be pending
+  funds or wallet could be unsynced) or if "All" means `Everything` where the presence of unconfirmed
+  funds or pending shards to be scanned would trigger an error.
```
nuttycom:
```suggestion
    funds being the `MaxSpendable` amount that it is currently in the wallet (there could be pending
    funds or wallet could be unsynced) or if "All" means `Everything` where the presence of unconfirmed
    funds or pending shards to be scanned would trigger an error.
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1191 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+            MaxSpendMode::Everything,
+            ConfirmationsPolicy::new_symmetrical_unchecked(
+                2,
+                #[cfg(feature = "transparent-inputs")]
+                true
+            )
+        ),
+        Err(data_api::error::Error::DataSource(_))
+    );
+}
```
nuttycom, line 921: Follow-up work: the setup here could also be used to test `MaxSpendMode::MaxSpendable` and eliminate some of the duplicate boilerplate.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1191 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+pub fn spend_everything_multi_step_single_note_proposed_transfer<T: ShieldedPoolTester, DSF>(
+    ds_factory: DSF,
+    cache: impl TestCache,
+) where
+    DSF: DataStoreFactory,
+    <DSF as DataStoreFactory>::AccountId: std::fmt::Debug,
+{
+    use crate::data_api::{testing::transparent::GapLimits, MaxSpendMode, OutputOfSentTx};
+
+    let gap_limits = GapLimits::new(10, 5, 3);
```
nuttycom, line 1108: nonblocking: gap limits are irrelevant to this test.

zcash_client_backend/src/data_api.rs
```diff
@@ -801,12 +817,104 @@ impl<NoteRef> NoteRetention<NoteRef> for SimpleNoteRetention {
...
+        }
+    }
+
+    /// Returns the set of spendable Sapling notes.
+    pub fn sapling(&self) -> &[ReceivedNote<NoteRef, sapling::Note>] {
+        self.sapling.as_ref()
+    }
+
+    /// Consumes this value and returns the Sapling notes contained within it.
+    pub fn take_sapling(self) -> Vec<ReceivedNote<NoteRef, sapling::Note>> {
```
nuttycom, line 878: This is only used for tests, and in fact seems like it should be unnecessary.

nuttycom: (This is a move, just noting.)

nuttycom (changes requested):
More changelog fixes need to be made.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/3dd47d1f0c3db574d09a0459492ad59617468d2f..e0c2ed7564a7f3d2b6c0c71a1eaffd1e12b4df28) to address conflicts with `main`.

nuttycom (approved):
ACK 5fe0e5234d6e8d0cee38ccc0dff6f32135aabedd

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1191 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+    assert!(!found_tx_empty_memo); // there's no empty memo in this case
+
+    // Verify that the stored sent notes match what we're expecting
+    let sent_note_ids = st
+        .wallet()
+        .get_sent_note_ids(&sent_tx_id, T::SHIELDED_PROTOCOL)
+        .unwrap();
+    assert_eq!(sent_note_ids.len(), 1);
+
+    // The sent memo should the specified memo for the sent output
```
daira, line 563:
```suggestion
    // The sent memo should be the specified memo for the sent output
```
(nit)

zcash_client_backend/src/data_api/testing/pool.rs
daira: https://docs.codecov.com/docs/ignoring-paths

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1191 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+    let expected_step1_fee = zip317::MINIMUM_FEE;
+    let expected_ephemeral_spend =
+        (non_marginal_notes_value - expected_step0_fee - expected_step1_fee).unwrap();
+    let expected_ephemeral_balance = (non_marginal_notes_value - expected_step0_fee).unwrap();
+    let expected_step0_change = (non_marginal_notes_value - expected_step0_fee).unwrap();
+
+    let total_sent = (expected_step0_fee + expected_step1_fee + expected_ephemeral_spend).unwrap();
+
+    // check that the napkin math is Ok. Total value send should be the whole
+    // value of the wallet
```
daira, line 1522:
```suggestion
    // value of the wallet's non-marginal notes
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1191 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
+    // Generate a ZIP 320 proposal, sending to the wallet's default transparent address
+    // expressed as a TEX address.
+    let tex_addr = match default_addr {
+        TransparentAddress::PublicKeyHash(data) => Address::Tex(data),
+        _ => unreachable!(),
+    };
+
+    let fee_rule = StandardFeeRule::Zip317;
+
+    // We use `st.propose_standard_transfer` here in order to also test round-trip
```
daira, line 1534:
```suggestion
    // We use `st.propose_send_max_transfer` here in order to also test round-trip
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -450,6 +450,1191 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
...
```
daira, line 1353:
```suggestion
    // We use `st.propose_send_max_transfer` here in order to also test round-trip
```

zcash_client_memory/src/input_source.rs
```diff
@@ -279,6 +279,7 @@ impl<P: consensus::Parameters> MemoryWalletDb<P> {
             .take_while(|note| {
                 let take = match target_value {
                     TargetValue::AtLeast(target) => value_acc <= target,
+                    TargetValue::AllFunds(_) => unimplemented!(),
```
daira, line 282: This is reachable so should return an Error rather than panicking.

nuttycom:
I did not attempt to implement any nontrivial changes to semantics in the
zcash_client_memory crate when bringing it in; ChainSafe has a maintenance
grant starting that will have them addressing these issues.

PR #1734, opened 2025-03-06
zcash_client_backend: Fix errors in change computation for wallet-emptying transactions.

nuttycom:
Best reviewed hiding whitespace. Split diff rather than unified helps also.

### DIFF
--- a/zcash_client_backend/src/fees/common.rs
+++ b/zcash_client_backend/src/fees/common.rs
@@ -254,7 +254,7 @@ where
     };
 
     // We don't create a fully-transparent transaction if a change memo is used.
-    let transparent = net_flows.is_transparent() && change_memo.is_none();
+    let fully_transparent = net_flows.is_transparent() && change_memo.is_none();
 
     // If we have a non-zero marginal fee, we need to check for uneconomic inputs.
     // This is basically assuming that fee rules with non-zero marginal fee are
@@ -265,7 +265,7 @@ where
         // including both possibilities.
         let possible_change = {
             // These are the situations where we might not have a change output.
-            if transparent
+            if fully_transparent
                 || (cfg.dust_output_policy.action() == DustAction::AddDustToFee
                     && change_memo.is_none())
             {
@@ -372,7 +372,7 @@ where
                 .map(|_| P2PKH_STANDARD_OUTPUT_SIZE),
         );
 
-    let fee_without_change = cfg
+    let min_fee = cfg
         .fee_rule
         .fee_required(
             cfg.params,
@@ -385,8 +385,8 @@ where
         )
         .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))?;
 
-    let fee_with_change = max(
-        fee_without_change,
+    let max_fee = max(
+        min_fee,
         cfg.fee_rule
             .fee_required(
                 cfg.params,
@@ -400,91 +400,74 @@ where
             .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))?,
     );
 
-    let total_out_plus_fee_without_change =
-        (total_out + fee_without_change).ok_or_else(overflow)?;
-    let total_out_plus_fee_with_change = (total_out + fee_with_change).ok_or_else(overflow)?;
+    let total_out_plus_min_fee = (total_out + min_fee).ok_or_else(overflow)?;
+    let total_out_plus_max_fee = (total_out + max_fee).ok_or_else(overflow)?;
 
     #[allow(unused_mut)]
     let (mut change, fee) = {
-        if transparent && total_in < total_out_plus_fee_without_change {
-            // Case 1 for a tx with all transparent flows.
+        if total_in < total_out_plus_min_fee {
+            // Insufficient inputs exist to pay the minimum fee; there's no way we can construct
+            // the transaction.
             return Err(ChangeError::InsufficientFunds {
                 available: total_in,
-                required: total_out_plus_fee_without_change,
+                required: total_out_plus_min_fee,
             });
-        } else if transparent && total_in == total_out_plus_fee_without_change {
-            // Case 2 for a tx with all transparent flows.
-            (vec![], fee_without_change)
-        } else if total_in < total_out_plus_fee_with_change {
-            // Case 3a, or case 1 or 2 with non-transparent flows.
-            return Err(ChangeError::InsufficientFunds {
-                available: total_in,
-                required: total_out_plus_fee_with_change,
-            });
+        } else if fully_transparent && total_in == total_out_plus_min_fee {
+            (vec![], min_fee)
         } else {
-            // Case 3b or 3c.
-            let proposed_change =
-                (total_in - total_out_plus_fee_with_change).expect("checked above");
-
             // We obtain a split count based on the total number of notes of sufficient size
             // available in the wallet, irrespective of pool. If we don't have any wallet metadata
             // available, we fall back to generating a single change output.
-            let split_count = wallet_meta.map_or(NonZeroUsize::MIN, |wm| {
+            let split_count = usize::from(wallet_meta.map_or(NonZeroUsize::MIN, |wm| {
                 cfg.split_policy.split_count(
                     wm.total_note_count(),
                     wm.total_value(),
-                    proposed_change,
+                    // We use a saturating subtraction here because there may be insufficient funds to pay
+                    // the fee, *if* the requested number of split outputs are created. If there is no
+                    // proposed change, the split policy should recommend only a single change output.
+                    (total_in - total_out_plus_max_fee).unwrap_or(Zatoshis::ZERO),
                 )
-            });
-            let per_output_change = proposed_change.div_with_remainder(
-                NonZeroU64::new(
-                    u64::try_from(usize::from(split_count)).expect("usize fits into u64"),
-                )
-                .unwrap(),
-            );
+            }));
 
             // If we don't have as many change outputs as we expected, recompute the fee.
-            let (fee_with_change, excess_fee) =
-                if usize::from(split_count) < target_change_counts.total_shielded() {
-                    let new_fee_with_change = cfg
-                        .fee_rule
-                        .fee_required(
-                            cfg.params,
-                            target_height,
-                            transparent_input_sizes,
-                            transparent_output_sizes,
-                            sapling_input_count,
-                            sapling_output_count(if change_pool == ShieldedProtocol::Sapling {
-                                usize::from(split_count)
-                            } else {
-                                0
-                            })?,
-                            orchard_action_count(if change_pool == ShieldedProtocol::Orchard {
-                                usize::from(split_count)
-                            } else {
-                                0
-                            })?,
-                        )
-                        .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))?;
-                    (
-                        new_fee_with_change,
-                        (fee_with_change - new_fee_with_change).unwrap_or(Zatoshis::ZERO),
+            let total_fee = if split_count < target_change_counts.total_shielded() {
+                cfg.fee_rule
+                    .fee_required(
+                        cfg.params,
+                        target_height,
+                        transparent_input_sizes,
+                        transparent_output_sizes,
+                        sapling_input_count,
+                        sapling_output_count(if change_pool == ShieldedProtocol::Sapling {
+                            split_count
+                        } else {
+                            0
+                        })?,
+                        orchard_action_count(if change_pool == ShieldedProtocol::Orchard {
+                            split_count
+                        } else {
+                            0
+                        })?,
                     )
-                } else {
-                    (fee_with_change, Zatoshis::ZERO)
-                };
+                    .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))
+            } else {
+                Ok(max_fee)
+            }?;
 
+            let total_change = (total_in - (total_out + total_fee).ok_or_else(overflow)?)
+                .unwrap_or(Zatoshis::ZERO);
+            let per_output_change = total_change.div_with_remainder(
+                NonZeroU64::new(u64::try_from(split_count).expect("usize fits into u64")).unwrap(),
+            );
             let simple_case = || {
                 (
-                    (0usize..split_count.into())
+                    (0usize..split_count)
                         .map(|i| {
                             ChangeValue::shielded(
                                 change_pool,
                                 if i == 0 {
                                     // Add any remainder to the first output only
-                                    (*per_output_change.quotient()
-                                        + *per_output_change.remainder()
-                                        + excess_fee)
+                                    (*per_output_change.quotient() + *per_output_change.remainder())
                                         .unwrap()
                                 } else {
                                     // For any other output, the change value will just be the
@@ -495,7 +478,7 @@ where
                             )
                         })
                         .collect(),
-                    fee_with_change,
+                    total_fee,
                 )
             };
 
@@ -504,7 +487,7 @@ where
                 .dust_threshold()
                 .unwrap_or(cfg.default_dust_threshold);
 
-            if proposed_change < change_dust_threshold {
+            if total_change < change_dust_threshold {
                 match cfg.dust_output_policy.action() {
                     DustAction::Reject => {
                         // Always allow zero-valued change even for the `Reject` policy:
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/fees/zip317.rs
```diff
@@ -425,14 +425,87 @@ mod tests {
                 result,
                 Ok(balance) if
                     balance.proposed_change() == [
-                        ChangeValue::sapling(Zatoshis::const_from_u64(124_7500), None),
-                        ChangeValue::sapling(Zatoshis::const_from_u64(124_2500), None),
-                        ChangeValue::sapling(Zatoshis::const_from_u64(124_2500), None),
-                        ChangeValue::sapling(Zatoshis::const_from_u64(124_2500), None),
+                        ChangeValue::sapling(Zatoshis::const_from_u64(124_3750), None),
```
nuttycom, line 428: The altered logic now evenly splits the change instead of adding excess to the first change output.

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,78 @@ where
...
+            // Insufficient inputs exist to pay the minimum fee; there's no way we can construct
+            // the transaction.
             return Err(ChangeError::InsufficientFunds {
                 available: total_in,
-                required: total_out_plus_fee_without_change,
-            });
-        } else if transparent && total_in == total_out_plus_fee_without_change {
-            // Case 2 for a tx with all transparent flows.
-            (vec![], fee_without_change)
-        } else if total_in < total_out_plus_fee_with_change {
```
nuttycom, line 418: This line was the primary cause of the error: we were computing the maximum fee, then checking that the total of the inputs was sufficient to satisfy that maximum fee plus the output total; however, the maximum fee potentially takes into account change outputs that are never actually created if there's insufficient change to perform splitting for note management purposes.

zcash_client_backend/src/fees/common.rs
```diff
@@ -372,7 +372,7 @@ where
                 .map(|_| P2PKH_STANDARD_OUTPUT_SIZE),
         );
 
-    let fee_without_change = cfg
+    let min_fee = cfg
```
daira, line 375: Please reconsider the wording of the comment on lines 333-356 in light of this change.

zcash_client_backend/src/fees/common.rs
```diff
@@ -385,8 +385,8 @@ where
         )
         .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))?;
```
daira, line 387: The comment on lines 333-356 above refers to these cases:

> 1. Insufficient funds even without change.
> 2. The fee amount without change exactly cancels out the net flow balance.
> 3. The fee amount without change is smaller than the change.
>    3a. Insufficient funds once the change output is added.
>    3b. The fee amount with change exactly cancels out the net flow balance.
>    3c. The fee amount with change leaves a non-zero change value.

The change made here merges the handling of cases 3{a,b,c}, so first delete the obsolete comment above and move it here:

```suggestion

    // Once we calculate the balance with minimum fee (i.e. with no change),
    // there are three cases:
    //
    // 1. Insufficient funds even with minimum fee.
    // 2. The minimum fee exactly cancels out the net flow balance.
    // 3. The minimum fee is smaller than the change.
    //
    // If (case 2, or case 3 when the fee amount with a *single* change note
    // exactly cancels out the net flow balance) happen for a transaction with
    // any shielded flows, we want there to be a zero-value shielded change
    // output anyway, because:
    // * being able to distinguish these cases potentially leaks too much
    //   information (an adversary that knows the number of external recipients
    //   and the sum of their outputs learns the sum of the inputs if no change
    //   output is present); and
    // * we will then always have an shielded output in which to put change_memo,
    //   if one is used.
    //
    // Note that using the `DustAction::AddDustToFee` policy inherently leaks
    // more information.
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
+    let total_out_plus_min_fee = (total_out + min_fee).ok_or_else(overflow)?;
+    let total_out_plus_max_fee = (total_out + max_fee).ok_or_else(overflow)?;
 
     #[allow(unused_mut)]
     let (mut change, fee) = {
-        if transparent && total_in < total_out_plus_fee_without_change {
-            // Case 1 for a tx with all transparent flows.
+        if total_in < total_out_plus_min_fee {
+            // Insufficient inputs exist to pay the minimum fee; there's no way we can construct
+            // the transaction.
```
daira, line 410:
```suggestion
            // Case 1. Insufficient input value exists to pay the minimum fee; there's no way
            // we can construct the transaction.
```

zcash_client_backend/src/fees/common.rs
daira: Yes, because there was no splitting when this code was written. So there was at most one change note and it was correct for this case to be an error, because neither adding nor omitting the change note would work. The bug was introduced in #1579, and we failed to catch it in review.

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
-            (vec![], fee_without_change)
-        } else if total_in < total_out_plus_fee_with_change {
-            // Case 3a, or case 1 or 2 with non-transparent flows.
-            return Err(ChangeError::InsufficientFunds {
-                available: total_in,
-                required: total_out_plus_fee_with_change,
+                required: total_out_plus_min_fee,
             });
+        } else if fully_transparent && total_in == total_out_plus_min_fee {
+            (vec![], min_fee)
```
daira, line 416:
```suggestion
            // Case 2 for a tx with all transparent flows and no change memo (e.g.
            // the second transaction of a ZIP 320 pair).
            (vec![], min_fee)
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
+                            0
+                        })?,
                     )
-                } else {
-                    (fee_with_change, Zatoshis::ZERO)
-                };
+                    .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))
+            } else {
+                Ok(max_fee)
+            }?;
```
daira, line 455:
```suggestion
                max_fee
            };
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
+                        orchard_action_count(if change_pool == ShieldedProtocol::Orchard {
+                            split_count
+                        } else {
+                            0
+                        })?,
                     )
-                } else {
-                    (fee_with_change, Zatoshis::ZERO)
-                };
+                    .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))
```
daira, line 452:
```suggestion
                    .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))?
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
-        } else if total_in < total_out_plus_fee_with_change {
-            // Case 3a, or case 1 or 2 with non-transparent flows.
-            return Err(ChangeError::InsufficientFunds {
-                available: total_in,
-                required: total_out_plus_fee_with_change,
+                required: total_out_plus_min_fee,
             });
+        } else if fully_transparent && total_in == total_out_plus_min_fee {
+            (vec![], min_fee)
         } else {
```
daira, line 417: `max_fee` and `total_out_plus_max_fee` are only used in this branch, so it might be clearer to compute them here. Non-blocking.

zcash_client_backend/src/fees/common.rs
```diff
@@ -254,7 +254,7 @@ where
     };
```
daira, line 254:
```suggestion
    };
    assert!(target_change_counts.total_shielded() == target_change_count);
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
-                                usize::from(split_count)
-                            } else {
-                                0
-                            })?,
-                        )
-                        .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))?;
-                    (
-                        new_fee_with_change,
-                        (fee_with_change - new_fee_with_change).unwrap_or(Zatoshis::ZERO),
+            let total_fee = if split_count < target_change_counts.total_shielded() {
```
daira, line 433:
```suggestion
            let total_fee = if split_count < target_change_count {
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -400,91 +400,74 @@ where
...
-                } else {
-                    (fee_with_change, Zatoshis::ZERO)
-                };
+                    .map_err(|fee_error| ChangeError::StrategyError(E::from(fee_error)))
+            } else {
+                Ok(max_fee)
+            }?;
 
+            let total_change = (total_in - (total_out + total_fee).ok_or_else(overflow)?)
+                .unwrap_or(Zatoshis::ZERO);
```
daira, line 480: Previously we checked for insufficient funds at line 418. Now we must do it here.

nuttycom: I think that the insufficient funds check on (new) line 428 is now sufficient? That should have checked the case where there are insufficient funds to satisfy the minimum fee case. I guess though that if the change strategy has a bug and picks an invalid number of splits, we would want to catch it here. Given that, maybe instead of this being an `unwrap_or`, it should just be `ok_or` and produce an error.

zcash_client_backend/src/fees/common.rs
```diff
@@ -533,10 +538,10 @@ where
                         let fee_with_dust = (total_in - total_out)
                             .expect("we already checked for sufficient funds");
```
daira, line 539: `total_out <= total_out_plus_fee`, so this is correct once we've added back the check at line 480.

daira, line 539: `total_out <= total_out_plus_fee`, so this is correct once we've added back the check at line 480. However, since `total_change == total_in - (total_out + total_fee)`, we can also calculate it like this:
```suggestion
                        let fee_with_dust = (total_change + total_fee).ok_or_else(overflow)?
```
This is better because it makes the assertion redundant.

zcash_client_backend/src/fees/common.rs
```diff
@@ -533,10 +538,10 @@ where
...
                         // We can add a change output if necessary.
-                        assert!(fee_with_change <= fee_with_dust);
+                        assert!(total_fee <= fee_with_dust);
```
daira, line 541:
```suggestion
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -533,10 +538,10 @@ where
                         let fee_with_dust = (total_in - total_out)
                             .expect("we already checked for sufficient funds");
                         // We can add a change output if necessary.
```
daira, line 540: At this point we already checked that there are sufficient funds with a change output, and we're removing the assertion this comment is attached to (I can't do it in one suggestion). So this comment is confusing and redundant.

zcash_client_backend/src/fees/common.rs
```diff
@@ -530,13 +519,13 @@ where
                         // Zero-valued change is also always allowed for this policy, but when
                         // no change memo is given, we might omit the change output instead.
 
-                        let fee_with_dust = (total_in - total_out)
+                        let fee_with_dust = (total_in - subtotal_out)
                             .expect("we already checked for sufficient funds");
```
daira, line 523: `subtotal_out <= total_out`, so this is correct once we've added back the check at lines 460-463. However, since `total_change == total_in - (subtotal_out + total_fee)`, we can also calculate it like this:
```suggestion
                        let fee_with_dust = (total_change + total_fee).ok_or_else(overflow)?
```
This is better because it makes the assertion redundant.

zcash_client_backend/src/fees/common.rs
```diff
@@ -385,106 +381,99 @@ where
...
-        } else {
-            // Case 3b or 3c.
-            let proposed_change =
-                (total_in - total_out_plus_fee_with_change).expect("checked above");
+        }
+        Ordering::Equal => {
+            // Case 2 for a tx with all transparent flows and no change memo
+            // (e.g. the second transaction of a ZIP 320 pair).
+            (vec![], min_fee)
+        }
```
daira, line 400: The refactoring to use `.cmp` here removed the `fully_transparent` check for case 2. This contradicts the comment on lines 358-366 and also on line 397. I think the behaviour before that refactor was correct.

zcash_client_backend/src/fees/common.rs
```diff
@@ -385,106 +381,99 @@ where
...
                 available: total_in,
-                required: total_out_plus_fee_with_change,
+                required: total_out_with_min_fee,
             });
-        } else {
-            // Case 3b or 3c.
-            let proposed_change =
-                (total_in - total_out_plus_fee_with_change).expect("checked above");
+        }
+        Ordering::Equal => {
```
daira, line 396:
```suggestion
        Ordering::Equal if fully_transparent => {
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -385,106 +381,99 @@ where
...
+            // Case 2 for a tx with all transparent flows and no change memo
+            // (e.g. the second transaction of a ZIP 320 pair).
+            (vec![], min_fee)
+        }
+        Ordering::Greater => {
```
daira, line 401: The refactoring to use `.cmp` here removed the `fully_transparent` check for case 2. This contradicts the comment on lines 358-366 and also on line 397. I think the behaviour before that refactor was correct.
```suggestion
        Ordering::Equal if fully_transparent => {
            // Case 2 for a tx with all transparent flows and no change memo
            // (e.g. the second transaction of a ZIP 320 pair).
            (vec![], min_fee)
        }
        _ => {
```

zcash_client_backend/src/fees/common.rs
```diff
@@ -385,106 +381,99 @@ where
...
+            } else {
+                max_fee
+            };
+
+            let total_out = (subtotal_out + total_fee).ok_or_else(overflow)?;
+            let total_change =
+                (total_in - total_out).ok_or_else(|| ChangeError::InsufficientFunds {
+                    available: total_in,
+                    required: total_out,
+                })?;
```
daira, line 463: Please write a test that would fail if this check was not done, i.e. when there are insufficient funds to add a single change note. (If that case already failed with existing tests, that's sufficient.)

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -527,7 +598,7 @@ mod tests {
         assert_matches!(
             result,
             Ok(balance) if
-                balance.proposed_change() == [ChangeValue::sapling(Zatoshis::ZERO, None)]
+                balance.proposed_change().is_empty()
```
daira, line 601: Please explain why this case changes.

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -728,7 +799,7 @@ mod tests {
         assert_matches!(
             result,
             Ok(balance) if
-                balance.proposed_change() == [ChangeValue::sapling(Zatoshis::ZERO, None)] &&
+                balance.proposed_change().is_empty() &&
```
daira, line 802: Please explain why this case changes.

daira (approved):
utACK modulo comments.

zcash_client_backend/src/fees/common.rs
nuttycom: Okay, yeah, I'm convinced.

zcash_client_backend/src/fees/zip317.rs
nuttycom: This was related to the change that I am reverting after your comment on https://github.com/zcash/librustzcash/pull/1734#discussion_r1985418687

zcash_client_backend/src/fees/common.rs
nuttycom: Done, added to the suggestions commit.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/b212dbd6e826869a7682a36250f2d179433f8726..b03f06e1e288c66115cb2fcfa319fb3bb7b159f5) to address final comments.

daira (approved):
utACK

PR #1736, opened 2025-03-11
zcash_client_sqlite: Enable decorrelated retrieval of transparent UTXOs sent to ephemeral addreses.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -740,10 +766,30 @@ pub enum TransactionDataRequest {
     ///
     /// [`GetTaddressTxids`]: crate::proto::service::compact_tx_streamer_client::CompactTxStreamerClient::get_taddress_txids
     #[cfg(feature = "transparent-inputs")]
-    SpendsFromAddress {
+    TransactionsInvolvingAddress {
+        /// The address to request transactions and/or UTXOs for.
         address: TransparentAddress,
+        /// Only transactions mined at heighs greater than or equal to this height should be
```
daira, line 772:
```suggestion
        /// Only transactions mined at heights greater than or equal to this height should be
```

zcash_client_backend/src/data_api.rs
```diff
@@ -740,10 +766,30 @@ pub enum TransactionDataRequest {
...
+        /// as practical to the specified instant, and in a fashion that decorrelates this request
+        /// to a light wallet server from other requests made by the same caller.
+        ///
+        /// This may be ignored by callers that are able to satisfy the request without exposing
+        /// correlations between addresses to untrusted parties; for example, a wallet application
+        /// that uses a private, trusted-for-privacy supplier of chain data can safely ignore this
+        /// field.
+        request_at: Option<SystemTime>,
+        /// The caller should respond to this request only with transactions that have conform to
+        /// the specified transaction status filter.
```
daira, line 788:
```suggestion
        /// The caller should respond to this request only with transactions that conform to the
        /// specified transaction status filter.
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1023,39 +1091,90 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+                block_range_end: Some(std::cmp::min(chain_tip_height + 1, max_end_height)),
+                request_at: None,
+                tx_status_filter: TransactionStatusFilter::Mined,
+                output_status_filter: OutputStatusFilter::All,
+            },
+        )
+    })?;
+
+    // Perform ephemeral address checks only for addresses that do not have an unexpired
+    // transaction associated with them in the database.
```
str4d, line 1127: This wording could be clarified; the expiry height is serving a seondary purpose of re-allowing checks on ephemeral addresses for which their unexpired transactions were mined prior to expiry, once those transactions are old enough to have expired if they were not mined. They are not "expired transactions", and thus can be interpreted as "an unexpired transaction associated with [the ephemeral address]", but that's not the kind of expiry we care about here.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1023,39 +1091,90 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+            cached_transparent_receiver_address,
+            transparent_receiver_next_check_time
+         FROM addresses
+         WHERE key_scope = :ephemeral_key_scope
+         AND NOT EXISTS (
+            SELECT 'x'
+            FROM transparent_received_outputs tro
+            JOIN transactions t ON t.id_tx = tro.transaction_id
+            WHERE tro.address_id = addresses.id
+            AND t.expiry_height >= :chain_tip_height
```
str4d, line 1139: If the expiry height is equal to the chain tip height, then the transaction has expired (because the next block to be created would have a height greater than the expiry height, and thus cannot include this transaction).
```suggestion
            AND t.expiry_height > :chain_tip_height
```
But per my above comment, if this check is intended to compare as such and is thus not actually checking for expiry, then please add documentation explaining it.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1023,39 +1091,90 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
             let address = TransparentAddress::decode(params, &row.get::<_, String>(0)?)?;
-            let block_range_start = BlockHeight::from(row.get::<_, u32>(1)?);
-            let max_end_height = block_range_start + DEFAULT_TX_EXPIRY_DELTA + 1;
+            let request_at = row
+                .get::<_, Option<i64>>(1)?
+                .map(|t| {
+                    let timestamp = u64::try_from(t).map_err(|_| {
+                        SqliteClientError::CorruptedData(format!("Invalid timestamp: {}", t))
+                    })?;
+                    Ok::<_, SqliteClientError>(SystemTime::UNIX_EPOCH + Duration::new(timestamp, 0))
```
str4d, line 1156: Should this use `decode_epoch_seconds`?

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -36,11 +39,13 @@ pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0xc41dfc0e_e870_4859_be47_
 
 const DEPENDENCIES: &[Uuid] = &[add_account_uuids::MIGRATION_ID];
 
-pub(super) struct Migration<P> {
+pub(super) struct Migration<P, C, R> {
     pub(super) params: P,
+    pub(super) _clock: C,
+    pub(super) _rng: Arc<Mutex<R>>,
```
str4d, line 45:
```suggestion
    pub(super) _rng: Rc<Mutex<R>>,
```

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -75,6 +80,7 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
             ALTER TABLE addresses ADD COLUMN transparent_child_index INTEGER;
             ALTER TABLE addresses ADD COLUMN exposed_at_height INTEGER;
             ALTER TABLE addresses ADD COLUMN receiver_flags INTEGER;
+            ALTER TABLE addresses ADD COLUMN transparent_receiver_next_check_time INTEGER;
```
str4d, line 83: We never fill anything into this column, so why add it to the original table? Is it present just to make existing updated queries work during this migration? If so, document this. If not, remove it.

str4d (changes requested):
Reviewed 759474477f06ac7ccf8db6065e83810a0bb2708a.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -448,8 +464,46 @@ fn verify_sqlite_version_compatibility(
...
+        P: consensus::Parameters + 'static,
+        CL: Clock + Clone + 'static,
+        R: RngCore + Clone + 'static,
+    >(
+        wdb: &mut WalletDb<rusqlite::Connection, P, CL, R>,
+        seed: Option<SecretVec<u8>>,
+        target_migrations: &[Uuid],
+        verify_seed_relevance: bool,
+    ) -> Result<(), MigratorError<Uuid, WalletMigrationError>> {
+        super::init_wallet_db_internal(wdb, seed, target_migrations, verify_seed_relevance)
```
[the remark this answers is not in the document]
nuttycom: In a previous version of this PR, these were more load-bearing; I think they can be removed now.

str4d (approved):
utACK f0c5bc919b47ffac4cc53bda4c37f4aab37d4247

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -90,6 +99,7 @@ pub(super) fn all_migrations<P: consensus::Parameters + 'static>(
     //                          fix_bad_change_flagging     v_transactions_additional_totals
     //                                                                     |
     //                                                       transparent_gap_limit_handling
```
daira, line 101: Reminder that we had decided that having a branched dag of migrations at releases makes for an unnecessary explosion of possible database states, so we wanted to bring together the dag into a single node (e.g. see `tx_retrieval_queue` above) whenever possible.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1023,39 +1098,90 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+            cached_transparent_receiver_address,
+            transparent_receiver_next_check_time
+         FROM addresses
+         WHERE key_scope = :ephemeral_key_scope
+         AND NOT EXISTS (
+            SELECT 'x'
+            FROM transparent_received_outputs tro
+            JOIN transactions t ON t.id_tx = tro.transaction_id
+            WHERE tro.address_id = addresses.id
+            AND t.expiry_height > :chain_tip_height
```
daira, line 1151: If `t` expires *at* the chain tip height then it has expired, so this test correctly excludes it.

daira (commented):
Post-hoc ACK

zcash_client_sqlite/src/wallet/init/migrations.rs
str4d: Recall that we cannot just tie disparate leaves together arbitrarily, because currently the order in which dependent migrations are applied is not guaranteed and could be any permutation; one of the permutations could have an undeclared data dependency (https://github.com/zcash/librustzcash/pull/1506#discussion_r1725697768).

What we could choose to do (which I had suggested at the time we were making the above PR to fix the migration graph bug) is introduce "dummy migrations" that perform no actual table migrations, but are used as the leaf nodes in our public releases. These could be added "arbitrarily" (i.e. just depend on everything that is currently a leaf), but it would be very important that we do not use one of these dummy migrations as a dependency itself of any internal migration other than other dummy release migrations, for the above permutation reason.

To sum up, it is always necessary to consider the interaction of two leaves in the migration graph when combining them as dependencies of a new migration. We need to do these merges at some point, and our current migration track record suggests that 5 leaf migrations is too many (prior to that we'd succeeded in maintaining the graph with at most 4 leaf migrations, and at the time of the bug we had 6 leaf migrations that we rearranged as 1 leaf with 5 dependencies).

So I'd suggest a policy that if we get to 3 leaf migrations, a subsequently-defined migration needs to have at least 2 dependencies (and fully analyse their merging interaction). This should mean we stick around 2 leaves most of the time, with short-term excursions above that depending on how many parallel branches are being worked on.

PR #1738, opened 2025-03-13
zcash_client_sqlite: Add a `WalletMigrator` that supports externally-defined migrations

str4d:


zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -324,16 +324,219 @@ pub fn init_wallet_db<
...
+/// When a seed is provided, it is checked against the database for _relevance_: if any
+/// account in the wallet for which [`Account::source`] is [`AccountSource::Derived`] can
+/// be derived from the given seed, the seed is relevant to the wallet. If the given seed
+/// is not relevant, [`Self::init_or_migrate`] returns
+/// `Err(schemerz::MigratorError::Migration { error: WalletMigrationError::SeedNotRelevant, .. })`
+/// or `Err(schemerz::MigratorError::Adapter(WalletMigrationError::SeedNotRelevant))`.
+///
+/// We do not check whether the seed is relevant to any imported account, because that
+/// would require brute-forcing the ZIP 32 account index space. Consequentially, imported
+/// accounts are not migrated.
```
nuttycom, line 363: "imported accounts are not migrated" is ambiguous. There are plenty of migrations that affect accounts that do not require the seed, and these will run without error.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -324,16 +324,219 @@ pub fn init_wallet_db<
...
+    /// Sets the external migration graph to apply alongside the internal migrations.
+    ///
+    /// From a data management perspective, it can be useful to store additional data
+    /// alongside the `zcash_client_sqlite` wallet database. This method enables you to
+    /// provide an external [`schemerz`] migration graph that the migrator will apply to
+    /// the wallet database.
+    ///
+    /// # WARNING
+    ///
+    /// **DO NOT** depend on internal details of the `zcash_client_sqlite` schema!
```
nuttycom, line 453:
```suggestion
    /// **DO NOT** depend on or modify internal details of the `zcash_client_sqlite` schema!
```

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -324,16 +324,219 @@ pub fn init_wallet_db<
...
+    ///
+    /// The internal migrations are written to take into account internal relationships
+    /// between the `zcash_client_sqlite` tables, but they will never take into account
+    /// external tables. In particular, this means that you **MUST NOT**:
+    /// - Modify the structure or contents of any internal table.
+    /// - Create foreign keys in your tables pointing to internal tables.
```
nuttycom, line 459: This is excessively restrictive; while I understand your desire to protect the external migrations from error, that's not our job here. We need to simply document that primary keys should not be expected to be stable, and external schema should take this into account; this means that if an external table references an internal table via a foreign key, if that key changes it may require a corresponding external migration to fix. This isn't a problem, though, since all migrations run with foreign key enforcement turned off, and an external migration can fix the issue.

nuttycom (commented):
concept ACK with comments. Note that this will conflict with https://github.com/zcash/librustzcash/pull/1736

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/6eebe35bdb50c01cbd3edf29d721dcb0015e5090..796912d8450c0cf1df692ece4a1608ba6a799d50) to fix merge conflicts.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/f0c542734ef0b71defd027a98d1ce58ede7e5c95..dfa231ee95c367969547af5058b2e7d0115ed555) to rebase on `main`.

zcash_client_sqlite/src/wallet/init.rs
str4d: We agreed that I will remove this bullet point.

nuttycom (approved):
utACK ef6b090af58e75d743f9b3526384639d318def5d

daira (commented):
Post-hoc ACK with minor nit.

PR #1739, opened 2025-03-14
zcash_client_backend: Add `WalletRead::list_addresses`

### LINKED ISSUE #1740: zcash_client_backend: Add an API for fetching all addresses in the wallet

str4d:
Closes #1740.


nuttycom (dismissed):
utACK

zcash_client_backend/src/data_api.rs
```diff
@@ -496,6 +500,51 @@ impl<A: Copy> Account for (A, UnifiedIncomingViewingKey) {
...
+    /// Returns the diversifier index the address was derived at.
+    pub fn diversifier_index(&self) -> DiversifierIndex {
+        self.diversifier_index
+    }
+
+    /// Returns the key scope if this is a transparent address.
+    #[cfg(feature = "transparent-inputs")]
+    pub fn transparent_key_scope(&self) -> Option<TransparentKeyScope> {
+        // Internally match to ensure we ignore `self.transparent_key_scope` if it was set
+        // to nonsense.
```
daira, line 540: This should be checked in `from_parts` (in which case it does not need to be checked here, because the fields are private and it will be an enforced invariant).

str4d: *sigh* I knew someone would ask for that

daira (dismissed):
utACK, comment is non-blocking.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/1dacab378b7b1c058e0c8c64e8e2614064baddfc..a3ad11f9a4ad30903720a5f813b788e0816d6c76) to fix the address comment addressing.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/a3ad11f9a4ad30903720a5f813b788e0816d6c76..698dbf134bd59d921e6c37258e9001f90cda8504) to fix clippy lint.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -761,6 +761,54 @@ pub(crate) fn get_next_available_address<P: consensus::Parameters, C: Clock>(
...
+
+    while let Some(row) = rows.next()? {
+        let addr_str: String = row.get(0)?;
+        let di_vec: Vec<u8> = row.get(1)?;
+        let _scope = KeyScope::decode(row.get(2)?)?;
+
+        let addr = Address::decode(params, &addr_str).ok_or_else(|| {
+            SqliteClientError::CorruptedData("Not a valid Zcash recipient address".to_owned())
+        })?;
+        let diversifier_index = decode_diversifier_index_be(&di_vec)?;
```
nuttycom, line 792: Nonblocking/not significantly related to this PR: something that occurs to (and annoys) me is that when we start to do integration of zcashd wallet addresses that were generated from randomness, we'll no longer be able to maintain the not-null constraint on the diversifier index.

nuttycom (approved):
utACK 698dbf134bd59d921e6c37258e9001f90cda8504

zcash_client_sqlite/src/wallet.rs
str4d: Yep, I raised that point here: https://github.com/zcash/wallet/pull/33#discussion_r1995488331

There's a question to answer of whether or not we support migrating purely-random addresses or whether we require wallets to migrate away from them.

daira (commented):
Post-hoc ACK

zcash_client_sqlite/src/wallet.rs
nuttycom: I don't think we can require wallets to migrate away: a very common usage pattern by exchanges is to assign each user a distinct transparent address, and credit deposits to that address to that user. These addresses are then very often deeply embedded in end-user workflows: they may be whitelisted for withdrawals, embedded in scripts or internet point-of-sale system configuration, and so forth. These addresses will continue to receive funds so long as the outputs to the transparent protocol are allowed on the network, and Zallet must be able to spend those funds.

PR #1745, opened 2025-03-20
Add Serde serialization to structs in zcash_history.

### LINKED ISSUE #1744: Add Serde Serialize and Deserialise to structs in `zcash_history`
This is required to open zebra's `ReadStateService` to remote use. The fist step in this process is implementing `Serde` `Serialize` and`Deserialize` for the `ReadReasponse` and `ReadRequest` structs in `zebra-state`, and therefore all primitives used.

The structs in zcash_history that need to implement serialisation are:
- `zcash_history::entry::Version`
- `zcash_history::{EntryLink, EntryKind}`
- `zcash_history::node_data::{NodeDate, V2}`
- `zcash_history::tree::Tree`
- `zcash_history::version::{V1, V2}`

- The Issue in Zebra tracking this task can be found here: https://github.com/ZcashFoundation/zebra/issues/9337.

idky137:
Add functionality detailed in https://github.com/zcash/librustzcash/issues/1744.

Adds all functionality under "remote_read_state_service" feature flag. This could be removed if this functionality should be included by default?

 - Closes #1744


nuttycom:
My primary question about this PR is: what is the expected lifetime of the serialized data? 

If the serialized data here are serialized for use as a wire format (i.e., the serialized data is never persisted for long-term storage), and there can be a guarantee that both ends of that wire upgrade their version of this library at the same time, then I'm generally okay with using bare `serde` derives. If, however, any of those are not expected to be the case (client and server don't upgrade in tandem, or data gets persisted for longer than a single request/response) then there needs to be some sort of versioning applied to the serialized data.

idky137:
The intended use for this data is purely as a wire format. All data structures in Zebra and Zaino's databases already have serialisation implementations. 

With respect to whether both ends of the wire will update at the same time, this will be used in the JsonRPC server in zebra so this cannot be guaranteed but it is intended to only be used over internal networks so there is an assumption that organisations would update client and server at the same time. 

I think the NodeData, history tree and Entry structs are already manually versioned in zcash_history? If this is correct then the only effected structs that does not already have versioning ire `EntryLink` and `EntryKind`? I would be happy to add versioning here if desired?

idky137:
One option, if adding versioning to these struct will be cumbersome is to change the feature name from "remote_read_state_service" to "wire_format_serialisation" and add an explicit warning? There are multiple structs in Zebra that are sent over wire that do not implement versioning.

zancas:
What do you think @nuttycom ?

nuttycom:
> One option, if adding versioning to these struct will be cumbersome is to change the feature name from "remote_read_state_service" to "wire_format_serialisation" and add an explicit warning? There are multiple structs in Zebra that are sent over wire that do not implement versioning.

My preference would be to add versioning from the outset - if there's one thing I've learned in 27 years of writing code, it's "always version your serialized data, no exceptions." It just saves so much pain later.

zancas:
OK, who in the Zebra team should we loop in on this?

If we're going to add versioning, then we probably ought to have one unified and coherent system for it across orgs/code-bases..  right?

nuttycom:
> If we're going to add versioning, then we probably ought to have one unified and coherent system for it across orgs/code-bases.. right?

All serialization in `librustzcash` is versioned, because I'm such a stickler about this. However, we don't generally use `serde`. Where we use derived serialization (i.e. with the Postcard format for PCZT) we add an explicit version byte to the binary string that serialization produces. I'm not sure what the best approach is for serde.

zancas:
Can we reuse LRZ versioning across the ecosystem?

idky137:
I think trying to rewrite all the serialisation in zebra now would be a big undertaking and I dont think this issue should hang on how serialisation in Zebra is handled

Im happy to manually implement serde::serialise and serde::deserialise for the structs in zcash_history, this would enable me to add a version byte to the serialised data.

zcash_history/src/tree.rs
```diff
@@ -22,36 +43,41 @@ pub struct Tree<V: Version> {
...
+where
+    V: Version,
+    V::EntryLink: Into<EntryLink> + From<EntryLink> + Copy,
+    V::EntryKind: From<EntryKind>,
+{
     /// Resolve link originated from this tree
-    pub fn resolve_link(&self, link: EntryLink) -> Result<IndexedNode<V>, Error> {
-        match link {
+    pub fn resolve_link(&self, link: V::EntryLink) -> Result<IndexedNode<V>, Error> {
+        match link.into() {
```
nuttycom, line 57:
```suggestion
        match EntryLink::from(link) {
```
This is kind of trivial, in this call site there's no ambiguity, but in general `into` is unsafe because it can cause builds to break at a distance if multiple `From` impls from different locations end up in scope. As a consequence, we prefer to use `X::from(y)` to `y.into()` in general; it makes it so that you don't have to do a complete analysis of the scope to decide whether or not using `into()` is risky in that particular context.

nuttycom (dismissed):
utACK with minor comment.

idky137:
I've made the `into()`s explicit but the only way the compiler would let me completely remove them was to add more constraints which i think could be hard to deal with downstream. 

Also, dding this versioning has also caused errors in zebra / zaino so I would like to fix those error before setting this PR as ready for review.

nuttycom:
> I've made the `into()`s explicit but the only way the compiler would let me completely remove them was to add more constraints which i think could be hard to deal with downstream.
> 
> Also, adding this versioning has also caused errors in zebra / zaino so I would like to fix those errors before setting this PR as ready for review.

I will see if I can figure out a way to make this nicer.

nuttycom:
@idky137 I'm not able to create a PR targeting your branch for some reason, so I'm attaching a patch here: 
[simplify_constraints.txt](https://github.com/user-attachments/files/19541505/simplify_constraints.txt)

nuttycom:
The above patch is semantically equivalent, because there's a blanket implementation from `From` that generates an `Into` for the inverse operation.

idky137:
Thank you for this!

idky137:
There are currently build errors in the downstream prs that were introduced by the versioning of EntryKind and EntyLink so I wanted to make sure that these changes are fully functional before committing any code. 

I think that this code is correct and the errors are coming from the new constraints introduced in zebra but I still haven't got to the bottom of the error.

It may be to do with zebra using an older version of librustzcash (cargo updating in zebra breaks the lrz dependency) and not to do with this new code.

idky137:
Errors are currently documented here https://github.com/zingolabs/zaino/pull/247

nuttycom:
Given that the motivation for this change was actually to be able to serialize the data structure inside of a `ReadStateService` response, after further investigation @str4d and I figured out that it shouldn't have been used in those responses to begin with; instead, just the hash of the tree should be included. Closing this as obviated by  https://github.com/ZcashFoundation/zebra/pull/9444

Issue #1746, opened 2025-03-20
zcash_client_backend: `decrypt_and_store_transaction` sets height incorrectly for mempool transactions


str4d:
The height at which a transaction is being decrypted is either the height it was mined at, or the "mempool height" (current chain tip height + 1, i.e. the height of the next block that could be mined).

In https://github.com/zcash/librustzcash/pull/258/commits/74b2f0a79e71e223b9f3615a536b79d6d4d346f0 we started passing the decryption height to `zcash_client_backend::decrypt::decrypt_transaction` because it was needed to correctly implement ZIP 212 rules. This code remains present:

https://github.com/zcash/librustzcash/blob/4130409eb04d30fb1dd9d516a7ce98dd1924c99a/zcash_client_backend/src/data_api/wallet.rs#L172-L198

https://github.com/zcash/librustzcash/blob/4130409eb04d30fb1dd9d516a7ce98dd1924c99a/zcash_client_backend/src/decrypt.rs#L104-L111

`DecryptedTransaction::mined_height` was added in https://github.com/zcash/librustzcash/pull/1462, released in `zcash_client_backend 0.TODO`. This took the decryption height passed into `decrypt_transaction` and mis-interpreted it as an unconditional mined height, making the implicit assumption that all transactions passed into it were mined:

https://github.com/zcash/librustzcash/blob/4130409eb04d30fb1dd9d516a7ce98dd1924c99a/zcash_client_backend/src/decrypt.rs#L219-L225

I think the fix here is that we need to either do the decryption height calculation inside `decrypt_transaction` (passing in both the optional mined height and the current chain tip height), or pass in both the decryption height and mined height (and check they are consistent).

daira:
> `DecryptedTransaction::mined_height` was added in [#1462](https://github.com/zcash/librustzcash/pull/1462), released in `zcash_client_backend 0.TODO`.

zcash_client_backend 0.13.0

https://github.com/zcash/librustzcash/blob/zcash_client_backend-0.13.0/zcash_client_backend/src/data_api/wallet.rs#L95

PR #1747, opened 2025-03-20
zcash_client_backend: Fix zcash/librustzcash#1746

nuttycom:
This fixes https://github.com/zcash/librustzcash/issues/1746, which made it possible for `zcash_client_backend::decrypt_and_store_transaction` to incorrectly set a `mined_height` value for a mempool transaction.

### DIFF
--- a/zcash_client_backend/src/decrypt.rs
+++ b/zcash_client_backend/src/decrypt.rs
@@ -7,7 +7,7 @@ use zcash_primitives::{
     transaction::components::sapling::zip212_enforcement, transaction::Transaction,
 };
 use zcash_protocol::{
-    consensus::{self, BlockHeight},
+    consensus::{self, BlockHeight, NetworkUpgrade},
     memo::MemoBytes,
     value::Zatoshis,
 };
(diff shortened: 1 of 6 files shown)


str4d (approved):
utACK ca18734642275819ba6744d272665141d7e4de22

zcash_client_backend/src/decrypt.rs
```diff
@@ -103,13 +103,36 @@ impl<A> DecryptedOutput<orchard::note::Note, A> {
...
+    chain_tip_height: Option<BlockHeight>,
     tx: &'a Transaction,
     ufvks: &HashMap<AccountId, UnifiedFullViewingKey>,
 ) -> DecryptedTransaction<'a, AccountId> {
-    let zip212_enforcement = zip212_enforcement(params, height);
+    let zip212_enforcement = zip212_enforcement(
+        params,
+        // Height is block height for mined transactions, and the "mempool height" (chain height + 1)
+        // for mempool transactions. We fall back to Sapling activation if we have no other
+        // information.
```
daira, line 128: Should we not fall back to a height after Canopy activation if we have no information? New wallets will have a birthday height after Canopy.

zcash_client_sqlite/src/wallet/init/migrations/sapling_memo_consistency.rs
```diff
@@ -108,7 +108,8 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
                     ))
                 })?;
 
-            let decrypted_outputs = decrypt_transaction(&self.params, block_height, &tx, &ufvks);
+            let decrypted_outputs =
+                decrypt_transaction(&self.params, Some(block_height), None, &tx, &ufvks);
```
daira, line 112: This decrypts the transaction according to its expiry height if it was not mined, which seems fine.

daira (commented):
Post-hoc ACK

zcash_client_backend/src/decrypt.rs
daira: Ah, I had misremembered how the spec (and hopefully the implementation?) handles decrypting a lead-byte 0x02 ciphertext before Canopy. It is accepted:
image
So this is "generous in what we accept", which is fine in this particular context.

PR #1748, opened 2025-03-21
Proposed contributor guidelines & PR template

shielded-nate:
I started looking into making a PR for a ticket [^1], then started wondering about some things (such as dependency policy), then decided that the most productive thing to do rather than asking multiple questions was submit a PR with my best guess as to various developer policies in the form of contributing guidelines and a PR template. ;-)

So, does this PR look like it matches the policies well enough in your collective understanding, or could it be changed to do so?

Please feel free to liberally edit/tweak this submission to match the actual maintainer choices (faster), or ping me to make edits (slower).

----

[^1]: The issue I started looking into is https://github.com/zcash/zcash-devtool/issues/95 and the first thought I had was "oh, I'll add `thiserror` like I do on personal projects", then I recalled how cautious this team is with vetting dependencies, which got me wondering about dev policies/practices.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,58 @@
+<!-- Thank you for your contribution! -->
+
+<!-- This template is intended to help you be most successful in contributing. While it may seem heavyweight, ensuring we cover all of these topics before PR reviews saves us all time and helps us maintain strong code safety and quality. -->
+
+# Summary
+
+<!-- Describe what this PR improves along with a description of how in a sentence or two. -->
+
+## Issue Tracking
+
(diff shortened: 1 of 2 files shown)


str4d:
Our development process and contributor guide is here: https://github.com/zcash/librustzcash/wiki

I don't have time to review this PR, but you should make sure anything here is consistent with that page.

nuttycom (changes requested):
Fixes to the incorrect statements about semver compatibility are blocking. Otherwise LGTM.

nuttycom:
Did you see we also just merged a CONTRIBUTING.md file?

PR #1752, opened 2025-03-26
zcash_client_sqlite: Fix invalid attempt to derive Sapling receivers in gap limit handling migration.

nuttycom:


zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -480,7 +480,7 @@ impl<P: consensus::Parameters, C: Clock, R: RngCore> RusqliteMigration for Migra
                                 .expect("previously generated an address"),
                         )
                         .expect("roundtrip");
-                    let ua = uivk.address(di, UnifiedAddressRequest::AllAvailableKeys)?;
+                    let ua = uivk.address(di, UnifiedAddressRequest::ALLOW_ALL)?;
```
nuttycom, line 483: This change is safe, because the migration will either have succeeded with `AllAvailableKeys`, or it will have failed in which case the migration will not have been applied and will re-run with the correction the next time the wallet starts up.

str4d (approved):
utACK 4b75b472838cec4a3a3ee9fe2f07b43e3bdf1a67

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
str4d: This change is also safe because it is only used in the context of diversifiers derived from either Sapling or Orchard receivers, ensuring that the resulting UA will contain a shielded receiver.

str4d:
Published 7dcdc4caf060b3aa45151562aacf2d1674b396d3 as `zcash_client_sqlite 0.16.1`.

PR #1755, opened 2025-04-01
zcash_client_sqlite: Fix migration error caused by missing transparent addresses prior to the index of the default address

nuttycom:
Earlier versions of the `zcash_client_sqlite` wallet did not store the transparent addresses at child indices prior to the diversifier index of the first unified address in the `addresses` table. This caused an error in migration for wallets that had received outputs at the address with child index 0, likely because they imported a seed generated by a wallet that had previously exposed that address.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -611,7 +611,7 @@ pub(crate) fn add_account<P: consensus::Parameters>(
         false,
     )?;
 
-    // Pre-generate transparent addresses prior to the index of the default address.
+    // Pre-generate external transparent addresses prior to the index of the default address.
     #[cfg(feature = "transparent-inputs")]
     if let Ok(default_addr_idx) = NonHardenedChildIndex::try_from(d_idx) {
         transparent::generate_address_range(
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -611,6 +611,20 @@ pub(crate) fn add_account<P: consensus::Parameters>(
         false,
     )?;
 
+    // Pre-generate transparent addresses prior to the index of the default address.
+    #[cfg(feature = "transparent-inputs")]
+    if let Ok(default_addr_idx) = NonHardenedChildIndex::try_from(d_idx) {
+        transparent::generate_address_range(
+            conn,
+            params,
+            account_id,
+            KeyScope::EXTERNAL,
+            UnifiedAddressRequest::ALLOW_ALL,
+            NonHardenedChildIndex::const_from_index(0)..default_addr_idx,
+            false,
+        )?
+    }
+
     // Pre-generate transparent addresses up to the gap limits for the external, internal,
     // and ephemeral key scopes.
     #[cfg(feature = "transparent-inputs")]
--- a/zcash_client_sqlite/src/wallet/init/migrations/ensure_default_transparent_address.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/ensure_default_transparent_address.rs
@@ -1,0 +1,50 @@
+//! Ensures that an external transparent address exists in the `addresses` table for each
+//! non-hardened child index starting at index 0 and ending at the index corresponding to default
+//! address for the account.
+
+use std::collections::HashSet;
+use uuid::Uuid;
+
+use rusqlite::Transaction;
+use schemerz_rusqlite::RusqliteMigration;
+use zcash_protocol::consensus;
+
+use super::transparent_gap_limit_handling;
+use crate::wallet::init::WalletMigrationError;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x702cf97b_8395_4edc_b584_5c9f87f0ef35);
+
+const DEPENDENCIES: &[Uuid] = &[transparent_gap_limit_handling::MIGRATION_ID];
+
+pub(super) struct Migration<P> {
+    pub(super) params: P,
+}
+
+impl<P> schemerz::Migration<Uuid> for Migration<P> {
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        DEPENDENCIES.iter().copied().collect()
+    }
+
+    fn description(&self) -> &'static str {
+        "Add support for general transparent gap limit handling, unifying the `addresses` and `ephemeral_addresses` tables."
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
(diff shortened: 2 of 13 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -611,6 +611,20 @@ pub(crate) fn add_account<P: consensus::Parameters>(
         false,
     )?;
 
+    // Pre-generate transparent addresses prior to the index of the default address.
```
daira, line 614:
```suggestion
    // Pre-generate external transparent addresses prior to the index of the default address.
```

zcash_client_sqlite/src/wallet/init/migrations/ensure_default_transparent_address.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn id(&self) -> Uuid {
+        MIGRATION_ID
+    }
+
+    fn dependencies(&self) -> HashSet<Uuid> {
+        DEPENDENCIES.iter().copied().collect()
+    }
+
+    fn description(&self) -> &'static str {
+        "Add support for general transparent gap limit handling, unifying the `addresses` and `ephemeral_addresses` tables."
```
daira, line 33: Correct this description (I think this is the description of `transparent_gap_limit_handling`).

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -60,6 +60,73 @@ impl<P, C, R> schemerz::Migration<Uuid> for Migration<P, C, R> {
...
+            .map(|b| decode_diversifier_index_be(&b[..]))
+            .transpose()?
+            .and_then(|di| NonHardenedChildIndex::try_from(di).ok());
+
+        // Ensure that there is an address for each possible external address index prior
+        // to the default UA for the account. We guarantee addresses up to the gap limit,
+        // but not beyond.
+        let start = NonHardenedChildIndex::const_from_index(0);
+        let end = std::cmp::min(
+            min_transparent_idx
```
daira, line 109: Is the assumption here that all necessary indices starting from the minimum transparent index will already have been generated?

zcash_client_sqlite/src/wallet/init/migrations/ensure_default_transparent_address.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, conn: &Transaction) -> Result<(), WalletMigrationError> {
+        #[cfg(feature = "transparent-inputs")]
+        transparent_gap_limit_handling::insert_initial_transparent_addrs(conn, &self.params)?;
```
daira, line 42: The code in the `transparent_gap_limit_handling` migration that updates `transparent_received_outputs.address_id` is dependent on the set of cached transparent addresses. Does it need to be re-run here?

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -60,6 +60,73 @@ impl<P, C, R> schemerz::Migration<Uuid> for Migration<P, C, R> {
...
+            start..end,
+            false,
+        )?;
+    }
+
+    Ok(())
+}
+
 impl<P: consensus::Parameters, C: Clock, R: RngCore> RusqliteMigration for Migration<P, C, R> {
     type Error = WalletMigrationError;
```
daira, line 132: Btw there is a typo in an error message on line 337.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -117,38 +121,36 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
...
+            )),
+        )?;
         let scope = KeyScope::decode(row.get(2)?)?;
 
-        let taddr = Address::decode(params, &ua_str)
+        let taddr = Address::decode(params, &addr_str)
             .ok_or_else(|| {
                 SqliteClientError::CorruptedData("Not a valid Zcash recipient address".to_owned())
             })?
             .to_transparent_address();
```
daira, line 138: Does this do the right thing when the decoded address is already a t-address? It should; just checking.

daira (changes requested):
Blocking comments.

zcash_client_sqlite/src/wallet/init/migrations/ensure_default_transparent_address.rs
nuttycom: No; if the `transparent_gap_limit_handling` migration ran successfully, then all of the `transparent_received_outputs` entries are correctly updated with the associated address ids, and we will not have inserted any records into `transparent_received_outputs` that could violate the relationship to the `addresses` table after that exactly because `address_id` is required to be non-null.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Yes; it returns the p2pkh address for bare taddrs, TEX addresses, and UAs with transparent receivers.

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -60,6 +60,73 @@ impl<P, C, R> schemerz::Migration<Uuid> for Migration<P, C, R> {
...
+            .and_then(|di| NonHardenedChildIndex::try_from(di).ok());
+
+        // Ensure that there is an address for each possible external address index prior to the
+        // default UA for the account. If the default address has a diversifier index greater than
+        // the gap limit, we generate transparent addresses up to that index but not beyond.
+        let start = NonHardenedChildIndex::const_from_index(0);
+        let end = std::cmp::min(
+            min_transparent_idx
+                .or(min_diversifier_idx)
+                .unwrap_or(NonHardenedChildIndex::const_from_index(1)),
```
str4d, line 111: Document that we only generate 1 in this case (instead of gap limit) because this is in essence inserting the "legacy default t-addr" that we previously were manually inserting into the output of `get_transparent_receivers`. (It also means that we have an entry at index 0, and thus when an address is actually generated we create a gap that the gap-handling logic will handle.)

str4d (changes requested):
Reviewed 28a64daa0bc72461252bbd31030ba0c72f30e07d.

str4d (approved):
utACK 9b24c2eff8cecdb30cbf76a7ee09aecdbd98bc03

str4d (approved):
utACK 3ff2a29efa553f3f97578406a81898b115745a76

daira:
Post-hoc ACK 3ff2a29efa553f3f97578406a81898b115745a76

PR #1753, opened 2025-03-28
Pin `time-core` to workaround `deranged` impl conflict.

nuttycom:
This pins `time-core` to version `0.1.2` to work around https://anonticket.torproject.org/user/projects/arti/issues/pending/4028/

PR #1757, opened 2025-04-04
Bump dependencies

str4d:


str4d:
I've not encountered the problem that #1753 fixes, so IDK if this PR also fixes it.

PR #1760, opened 2025-04-08
zcash_client_backend: Add helpers for parsing `LightdInfo` fields

str4d:


zcash_client_backend/src/proto.rs
```diff
@@ -261,6 +262,85 @@ impl<SpendAuth> From<&orchard::Action<SpendAuth>> for compact_formats::CompactOr
...
+        self.estimated_height
+            .try_into()
+            .expect("lightwalletd should provide in-range heights")
+    }
+
+    /// Returns the donation address for this server.
+    ///
+    /// Returns `None` if:
+    /// - no donation address was provided.
+    /// - the provided donation address is not a valid [`unified::Address`].
```
emersonian, line 332: Thanks for adding this. It would be nice to have an Error state rather than just "None" - silently failing when someone provides an invalid address in their config feels like bad CLI UX.

str4d: This is not CLI UX. This is the API that wallet libraries use; if the server configured an invalid address, there is nothing the wallet can do.

emersonian: If a server operator launches Zaino with an invalid UA in their config file I want Zaino to crash with an informative error. How can we do that if this method returns the same as an empty address? That's what I meant by "CLI UX" but was admittedly too vague of a term.

str4d: Zaino should be checking the value of this before putting it into a gRPC response, ideally by just using the `unified::Address` type in its in-memory representation. It would then error at that point which is far more helpful than hoping the operator double-checks the gRPC output after deploying their config into production.

str4d:
[Rebased on `main`](https://github.com/zcash/librustzcash/compare/f4185f4b5fbbbb09b3dc6ca7a51259f809d8739e..e7c2193117129d89dfb2e21b8b23c33d6b49cf60) to fix a changelog merge conflict.

nuttycom (approved):
utACK

PR #1764, opened 2025-04-09
zcash_client_sqlite migration and doc fixes; also CI: fix bugs related to pool feature flag combinations

str4d:
Previously, in CI we were:
- Not enabling the `orchard` feature flag for our regular builds and linters.
- Always enabling the `transparent-inputs` feature flag for our tests.

These led to bugs being missed, and deployed to production. We now always build/lint with all pool-specific feature flags enabled, and we test with more flag combinations.

This PR also corrects a problem in zcash_client_sqlite with the `transparent_gap_limit_handling` migration only having been run when the `transparent-inputs` feature flag enabled. It should have been run unconditionally. We also make some other doc and test fixes to zcash_client_sqlite.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations.rs
@@ -17,6 +17,7 @@ mod ensure_orchard_ua_receiver;
 mod ephemeral_addresses;
 mod fix_bad_change_flagging;
 mod fix_broken_commitment_trees;
+mod fix_transparent_received_outputs;
 mod full_account_ids;
 mod initial_setup;
 mod nullifier_map;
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -267,9 +269,19 @@ pub const V_0_15_0: &[Uuid] = &[
...
+    transparent_gap_limit_handling::MIGRATION_ID,
+];
 
-const V_0_16_2: &[Uuid] = &[ensure_default_transparent_address::MIGRATION_ID];
+/// Leaf migrations in the 0.16.2 release.
+const V_0_16_2: &[Uuid] = &[
+    fix_bad_change_flagging::MIGRATION_ID,
+    v_transactions_additional_totals::MIGRATION_ID,
+    ensure_default_transparent_address::MIGRATION_ID,
+];
```
str4d, line 284: These changes fix a bug where the text graph correctly showed there was more than one leaf migration (though incorrectly only two), but the added constants only declared one of them.

nuttycom: Okay, I wasn't sure what exactly to do here given that the same migration was updated in multiple releases.

nuttycom (approved):
utACK d53c02de5268b260f71a0f8be09d9bb6219e33d2

daira (approved):
utACK. Updated title and description.

PR #1765, opened 2025-04-15
Define `TargetValue` to select spendable notes

pacu:
Working towards enabling `send_max_funds`.

Said @nuttycom over discord:

> We can also just chat here: right now, the main impediment to the approach suggested above is that https://docs.rs/zcash_client_backend/latest/zcash_client_backend/data_api/trait.InputSource.html#tymethod.select_spendable_notes requires a target value. We'll likely need to change the API of this method to make the target value a bespoke enum with variants `pub enum TargetValue {MaxSpendable, MinValue(Zatoshis)}`. We could alternately add a separate method, but I think that changing to use an enum for the target value is better.

My idea is to start implementing this in the most non-breaking way as possible. 

``` rust
#[derive(Debug, Clone, Copy)]
pub enum TargetValue {
    MaxSpendable,
    MinValue(Zatoshis)
} 
```

The Send Max feature has a lot of parts. This is the smallest change that I could think of and it does break a public interface already :(

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -13,6 +13,12 @@ and this library adheres to Rust's notion of
   - `DormantMode`
 
 ### Changed
+- `zcash_client_backend::data_api`:
+  - `TargetValue`: An intent of representing spendable value to reach a certain 
+    targeted amount.
+  - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. 
+    Existing calls to this function that used `Zatoshis` now use 
+    `TargetValue::AtLeast(Zatoshis)`
 - Migrated to `arti-client 0.28`, `dynosaur 0.2`, `tonic 0.13`.
 - `zcash_client_backend::tor`:
   - `Client::{connect_to_lightwalletd, get_latest_zec_to_usd_rate}` now ensure
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -139,6 +139,21 @@ pub enum NullifierQuery {
     All,
 }
 
+/// An intent of representing spendable value to reach a certain targeted
+/// amount.  `AtLeast(Zatoshis)` refers to the amount of `Zatoshis` that can cover
+/// at minimum the given zatoshis that is conformed by the sum of spendable notes.
+///
+/// Discussion: why not just use ``Zatoshis``?
+///
+/// the `Zatoshis` value isn't enough to explain intent when seeking to match a
+/// given a given amount. Is the value expressed in `Zatoshis` the ceiling value
+/// or the minimum value of a given spend intent? How would you express that the
+/// value spend intent is "as much as possible" without knowing the value upfront?
+#[derive(Debug, Clone, Copy)]
+pub enum TargetValue {
+    AtLeast(Zatoshis),
+}
+
 /// Balance information for a value within a single pool in an account.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Balance {
@@ -1201,7 +1216,7 @@ pub trait InputSource {
     fn select_spendable_notes(
         &self,
         account: Self::AccountId,
-        target_value: Zatoshis,
+        target_value: TargetValue,
         sources: &[ShieldedProtocol],
         anchor_height: BlockHeight,
         exclude: &[Self::NoteRef],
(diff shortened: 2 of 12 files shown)


pacu:
I created the minimum change that would introduce the concept of a TargetValue.  I did so because otherwise I would have to implement the Max variant of the enum and the change would be huge and It would delay the discussion on whether we want this or not.

pacu:
Clippy beta is failing but appears to be crashing not actually objecting a format

zcash_client_backend/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this library adheres to Rust's notion of
   - `DormantMode`
 
 ### Changed
+- `zcash_client_backend::data_api`:
+  - `TargetValue`: An intent of representing spendable value to reach a certain 
+  targeted amount.
+  - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. Existing calls to this function that used `Zatoshis` now use `TargetValue::MinValue(Zatoshis)`
```
nuttycom, line 19: Fix the wrapping.

zcash_client_backend/src/data_api.rs
```diff
@@ -139,6 +139,21 @@ pub enum NullifierQuery {
...
+///
+/// Discussion: why not just use ``Zatoshis``?
+///
+/// the `Zatoshis` value isn't enough to explain intent when seeking to match a
+/// given a given amount. Is the value expressed in `Zatoshis` the ceiling value
+/// or the minimum value of a given spend intent? How would you express that the
+/// value spend intent is "as much as possible" without knowing the value upfront?
+#[derive(Debug, Clone, Copy)]
+pub enum TargetValue {
+    MinValue(Zatoshis),
```
nuttycom, line 154:
```suggestion
    AtLeast(Zatoshis),
```
nonblocking, but I think this would be a better name.

nuttycom (dismissed):
utACK 676402cd3eaf278a9d887bd5c6163a2e175e5d54 with suggestions

nuttycom (changes requested):
Lots of test failures now, rustfmt changes, and comments were left unaddressed. Please revise the additional commit to fix everything.

pacu:
> Lots of test failures now, rustfmt changes, and comments were left unaddressed. Please revise the additional commit to fix everything.

Yeah f'd up all the work 🥶. Fixed it.

zcash_client_backend/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this library adheres to Rust's notion of
   - `DormantMode`
 
 ### Changed
+- `zcash_client_backend::data_api`:
+  - `TargetValue`: An intent of representing spendable value to reach a certain 
+  targeted amount.
```
nuttycom, line 18:
```suggestion
    targeted amount.
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this library adheres to Rust's notion of
...
+  - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. 
+  Existing calls to this function that used `Zatoshis` now use 
+  `TargetValue::AtLeast(Zatoshis)`
```
nuttycom, line 21:
```suggestion
    Existing calls to this function that used `Zatoshis` now use 
    `TargetValue::AtLeast(Zatoshis)`
```

nuttycom (approved):
CHANGELOG wrapping must respect indentation rules or otherwise it won't render correctly, otherwise utACK.

nuttycom (approved):
utACK 6ecb508725c20ee5764788d40a9b2e677dec49e7

zcash_client_backend/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this library adheres to Rust's notion of
   - `DormantMode`
 
 ### Changed
+- `zcash_client_backend::data_api`:
+  - `TargetValue`: An intent of representing spendable value to reach a certain 
+    targeted amount.
+  - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. 
+    Existing calls to this function that used `Zatoshis` now use 
+    `TargetValue::AtLeast(Zatoshis)`
 - Migrated to `arti-client 0.28`, `dynosaur 0.2`, `tonic 0.13`.
```
str4d, line 22: Nit for the future @pacu: We generally put changelog entries for global things like dependency updates first, then go through the changes to modules in alphabetical order. So in this case, the spot to put your changelog entry would be between the two existing top-level bullet points.

PR #1768, opened 2025-04-15
zcash_client_backend, zcash_client_sqlite: Support zcashd wallet.dat import.

nuttycom:


zcash_client_backend/CHANGELOG.md
```diff
@@ -29,6 +33,8 @@ and this library adheres to Rust's notion of
   - `select_spendable_notes`: parameter `target_value` now is a `TargetValue`. 
     Existing calls to this function that used `Zatoshis` now use 
     `TargetValue::AtLeast(Zatoshis)`
+  - `WalletRead::get_derived_account` now takes an additional `legacy_address_index`
+    argument when the `zcashd-compat` feature flag is enabled.
```
nuttycom: Fix this comment; it now takes `Zip32Derivation` instead.

zcash_client_backend/src/data_api.rs
```diff
@@ -568,8 +578,8 @@ impl AddressInfo {
     }
 
     /// Returns the diversifier index the address was derived at.
-    pub fn diversifier_index(&self) -> DiversifierIndex {
-        self.diversifier_index
+    pub fn source(&self) -> AddressSource {
```
nuttycom, line 581: CHANGELOG

zcash_client_backend/src/data_api.rs
```diff
@@ -2499,6 +2507,21 @@ pub trait WalletWrite: WalletRead {
...
+    /// Imports the given pubkey into the account without key derivation information, and adds the
+    /// associated transparent p2pkh address.
+    ///
+    /// The imported address will contribute to the balance of the account, but spending funds held
+    /// by this address requires the associated spending keys to be provided explicitly when
+    /// calling [`create_proposed_transactions`].
+    ///
+    /// [`create_proposed_transactions`]: wallet::create_proposed_transactions
+    #[cfg(feature = "transparent-inputs")]
+    fn import_standalone_transparent_pubkey(
```
nuttycom, line 2519: CHANGELOG

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -439,6 +439,28 @@ struct StepResult<AccountId> {
     utxos_spent: Vec<OutPoint>,
 }
 
+pub struct SpendingKeys<'a> {
```
nuttycom, line 442: CHANGELOG

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -457,7 +479,7 @@ pub fn create_proposed_transactions<DbT, ParamsT, InputsErrT, FeeRuleT, ChangeEr
     params: &ParamsT,
     spend_prover: &impl SpendProver,
     output_prover: &impl OutputProver,
-    usk: &UnifiedSpendingKey,
+    spending_keys: &SpendingKeys,
```
nuttycom, line 482: CHANGELOG

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1202,7 +1227,7 @@ fn create_proposed_transaction<DbT, ParamsT, InputsErrT, FeeRuleT, ChangeErrT, N
...
```
nuttycom, line 1230: CHANGELOG

zcash_client_backend/src/wallet.rs
```diff
@@ -558,27 +558,36 @@ impl OvkPolicy {
 /// This is implicitly scoped to an account.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg(feature = "transparent-inputs")]
-pub struct TransparentAddressMetadata {
-    scope: TransparentKeyScope,
-    address_index: NonHardenedChildIndex,
+pub enum TransparentAddressMetadata {
```
nuttycom, line 561: CHANGELOG

zcash_client_sqlite/src/lib.rs
```diff
@@ -685,17 +681,15 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> WalletRea
 
     fn get_derived_account(
         &self,
-        seed: &SeedFingerprint,
-        account_id: zip32::AccountId,
-        #[cfg(feature = "zcashd-compat")] legacy_address_index: Option<zcashd::LegacyAddressIndex>,
+        derivation: &Zip32Derivation,
```
nuttycom, line 684: This change should occur instead of `legacy_address_index` addition in the previous commit.

zcash_keys/Cargo.toml
```diff
@@ -67,11 +69,18 @@ byteorder = { workspace = true, optional = true }
 # - Digests
 blake2b_simd = { workspace = true }
 
+# - Zcashd compatibility
+bip0039 = { workspace = true, optional = true }
```
daira: If I remember correctly, we previously had a lot of trouble with `bip0039` being a dependency of librustzcash crates that are also depended on by zebra. This might be okay now that we've upgraded past the `bip0039` 0.11 breaking change.

zcash_client_backend/src/data_api.rs
```diff
@@ -788,6 +814,65 @@ pub enum OutputStatusFilter {
...
+/// [`TransactionDataRequest::transactions_involving_address`].
+#[cfg(feature = "transparent-inputs")]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TransactionsInvolvingAddress {
+    /// The address to request transactions and/or UTXOs for.
+    address: TransparentAddress,
+    /// Only transactions mined at heights greater than or equal to this height should be
+    /// returned.
+    block_range_start: BlockHeight,
+    /// Only transactions mined at heights less than this height should be returned.
```
```suggestion
    /// If set, only transactions mined at heights less than this height should be returned.
```

zcash_client_backend/src/wallet.rs
```diff
@@ -558,27 +558,36 @@ impl OvkPolicy {
...
-        Self {
+        Self::Derived {
             scope,
             address_index,
         }
     }
 
-    pub fn scope(&self) -> TransparentKeyScope {
-        self.scope
+    pub fn scope(&self) -> Option<TransparentKeyScope> {
```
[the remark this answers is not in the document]
daira: This is still not in the CHANGELOG I think.

zcash_client_backend/src/wallet.rs
```diff
@@ -558,27 +558,36 @@ impl OvkPolicy {
...
+        match self {
+            TransparentAddressMetadata::Derived { scope, .. } => Some(*scope),
+            TransparentAddressMetadata::Standalone(_) => None,
+        }
     }
 
-    pub fn address_index(&self) -> NonHardenedChildIndex {
-        self.address_index
+    pub fn address_index(&self) -> Option<NonHardenedChildIndex> {
```
daira, line 667: Ditto, this is not in the CHANGELOG.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -161,24 +162,37 @@ pub(super) const INDEX_HD_ACCOUNT: &str =
...
     key_scope INTEGER NOT NULL,
-    diversifier_index_be BLOB NOT NULL,
+    diversifier_index_be BLOB,
     address TEXT NOT NULL,
     transparent_child_index INTEGER,
     cached_transparent_receiver_address TEXT,
     exposed_at_height INTEGER,
     receiver_flags INTEGER NOT NULL,
     transparent_receiver_next_check_time INTEGER,
+    cached_transparent_receiver_pubkey BLOB,
```
str4d: This shouldn't be `cached_*` because for some accounts it is the canonical data store, and we need to preserve it (vs other caches that could be reconstructed later from other data, which we cache solely for query efficiency).

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -161,24 +162,37 @@ pub(super) const INDEX_HD_ACCOUNT: &str =
...
     FOREIGN KEY (account_id) REFERENCES accounts(id),
     CONSTRAINT diversification UNIQUE (account_id, key_scope, diversifier_index_be),
+    CONSTRAINT transparent_pubkey_unique UNIQUE (cached_transparent_receiver_pubkey),
     CONSTRAINT transparent_index_consistency CHECK (
-        (transparent_child_index IS NOT NULL) == (cached_transparent_receiver_address IS NOT NULL)
+        (transparent_child_index IS NULL OR diversifier_index_be < x'0000000F00000000000000')
+        AND (cached_transparent_receiver_address IS NOT NULL) == (
+            transparent_child_index IS NOT NULL OR cached_transparent_receiver_pubkey IS NOT NULL
+        )
```
str4d: What we want to constrain here is:
- If `cached_transparent_receiver_address` is null then both `transparent_child_index` and `transparent_receiver_pubkey` must be null.
- If `cached_transparent_receiver_address` is not null then precisely one of `transparent_child_index` or `transparent_receiver_pubkey` must be non-null.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -161,24 +162,37 @@ pub(super) const INDEX_HD_ACCOUNT: &str =
...
+    ),
+    CONSTRAINT foreign_or_diversified CHECK (
+        (diversifier_index_be IS NULL) == (key_scope = -1)
```
str4d, line 204: - I think we'd actually want to enforce that `diversifier_index_be` is null when `transparent_receiver_pubkey` is non-null.
- In a pairing, @nuttycom clarified that the intention here is to be able to encode non-derived key scope information. I agree that this is better than making the column nullable, but in that case we need to update the `key_scope` column documentation to state that non-negative values are always path child numbers for derivation, and then specify what each allowed negative constant means.

zcash_client_backend/src/data_api.rs
```diff
@@ -2597,6 +2607,21 @@ pub trait WalletWrite: WalletRead {
...
+    /// Imports the given pubkey into the account without key derivation information, and adds the
+    /// associated transparent p2pkh address.
+    ///
+    /// The imported address will contribute to the balance of the account, but spending funds held
+    /// by this address requires the associated spending keys to be provided explicitly when
+    /// calling [`create_proposed_transactions`].
+    ///
+    /// [`create_proposed_transactions`]: wallet::create_proposed_transactions
+    #[cfg(feature = "transparent-inputs")]
+    fn import_standalone_transparent_pubkey(
```
str4d, line 2761: In a pairing with @nuttycom, we (believe that we) decided to add a new "constructor" API that allows for specifically creating legacy transparent-only accounts, and then this API should return an error if `account` is not one of those (to prevent "bleeding" of transparent addresses into the account semantics of existing accounts, and simplify note selection / tx construction for those accounts as we can continue to only need a single USK for them, vs the current state of this PR where you could end up with a partial-spending-key-set issue that leads to the tx builder being bricked due to always trying to select coins that it can't spend, which is a UX issue for the Proposal APIs).

We also discussed how P2SH multisig will work, and that will be another separate constructor for that which takes in the descriptor template (alongside the relevant viewing-level extpubkey for this wallet's spending authority), and creates an account that is transparent-only but otherwise has a consistent and diversifiable address pool (since the proposed ZIP 48 mechanism supports deriving a sequence of P2SH addresses).

nuttycom: The notion that we should add a separate constructor API for legacy transparent-only accounts is the most disputed point here; we've carried on the discussion and need to consider the implications if someone imports the keys from two separate `wallet.dat` files that both hold funds in legacy addresses. In this case, since we don't want to brick those funds, my proposal is that we should make the zcashd ZIP 32 legacy account index `0x7FFFFFFF` the sole determiner of whether legacy semantics are applied to an account. Then, for any RPC method that we choose to support that involves legacy semantics, we add an optional `account_uuid` parameter. If only a single account having ZIP 32 index `0x7FFFFFF` exists in the wallet, then legacy semantics apply to that account; if more than one such account exists, then the user MUST specify the account UUID to disambiguate.

Given this approach, a separate constructor for creating the legacy account is not necessary; instead, we would just perform a check against the ZIP 32 account identifier of the account being specified in this API call and return an error if it is not 0x7FFFFFF.

This _does_ mean that we would not support import of transparent spending keys into the wallet _unless_ a legacy account exists; I think that's an okay restriction for right now.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: I just realized when working on the changes here that the `diversification` constraint can't be properly maintained in conjunction with putting imported addresses into this table.

nuttycom: Actually, that's incorrect; all `NULL`s are considered distinct values from the perspective of `UNIQUE`: https://www.sqlite.org/lang_createtable.html#unique_constraints

zcash_client_backend/src/data_api.rs
nuttycom: With respect to proposals, `propose_shielding` (which is currently the only way to spend transparent UTXOs) takes an explicit set of `TransparentAddress` values, and only UTXOs for these addresses will be shielded. 

I think that it's a programming error for a wallet to request shielding for a transparent address for which the spending key is not available.

str4d: Conclusion from our pairing:
- We do currently have the property that every account creatable and modifiable by `WalletWrite` is structurally guaranteed to have all of its funds spendable with a single USK (`WalletWrite::import_account_ufvk` means we aren't guaranteed that every account is derived from a single seed, but we still have the guarantee that a single USK is constructable by the caller).
- This means that every account has the property that either all of the funds it can see are spendable, or none of them are.
- We do intend to preserve this property, even with the addition of this new method for mutating accounts.
- Therefore, we need copious documentation everywhere that a downstream developer creating accounts will see, that emphasises this property:
  - The root crate documentation.
  - The `WalletWrite` trait's documentation.
  - This method's documentation.
  - The `Account` trait's documentation.
  - Probably `WalletRead::get_account` as well.
- Then this method needs to also document that the caller MUST ensure that if this account is already considered "spendable", that the spending key for this transparent pubkey can be provided to _every_ transaction signing operation for spends from this account (and that we as developers of `zcash_client_backend` or any of our downstream impls of it will not do anything to help you if you mess up and create an account that is only partially spendable, causing the account's funds to be unspendable).

zcash_client_backend/CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_client_backend::data_api`:
+  - `AddressSource`
+  - `AddressInfo::source` replaces `AddressInfo::diversifier_index`
+  - `wallet::SpendingKeys`
```
str4d: Why are these changes part of `Unreleased`, and the rest planned for 0.20.0?

zcash_client_backend/src/data_api.rs
```diff
@@ -635,8 +651,8 @@ impl AddressInfo {
     }
 
     /// Returns the diversifier index the address was derived at.
-    pub fn diversifier_index(&self) -> DiversifierIndex {
-        self.diversifier_index
+    pub fn source(&self) -> AddressSource {
```
str4d, line 680: Documentation for this method was not updated when the method was altered.

zcash_client_backend/src/data_api.rs
```diff
@@ -599,10 +605,20 @@ impl<A: Copy> Account for (A, UnifiedIncomingViewingKey) {
...
 }
 
+/// Source information about an address in the wallet.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum AddressSource {
+    /// The address was produced by HD derivation via a known path, with the given diversifier
+    /// index.
+    Derived { diversifier_index: DiversifierIndex },
+    /// The address was imported into the wallet, and no derivation information is available.
+    Imported,
```
str4d: This feels like an odd point to split at. I'm comparing this to `AccountSource` which has the same categories, and I don't think they make sense here:
- `Imported` as documented here can' be used for shielded addresses (unless the intention is for this to be used for some kind of future imported address that is not used for input detection at all, but just for identification?)
- Conversely, we can know the diversifier index for a shielded address, without it being produced by HD derivation (a FROST address would be an example: it may have some HD-derived components but it is not a derived address).

I think maybe the split we want here is `Diversified` (implying we know the corresponding IVK) vs `Standalone` (we don't)? Or if we actually want to indicate the address source (as pruned from `AccountSource`), that should be a separate axis from whether or not the address is diversified?

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -974,7 +974,11 @@ where
             &network,
             &prover,
             &prover,
-            usk,
+            &SpendingKeys::new(
+                usk,
+                #[cfg(feature = "transparent-inputs")]
+                &HashMap::new(),
```
str4d: I really do not like this construction API (not least because of the forced feature flag on one of the arguments). If all users are going to need to use `SpendingKeys`, there should be a constructor that directly converts a `USK` into a `SpendingKeys`.

zcash_client_backend/src/data_api.rs
```diff
@@ -1636,6 +1652,7 @@ pub trait WalletRead {
         &self,
         _account: Self::AccountId,
         _include_change: bool,
+        _include_standalone: bool,
```
str4d, line 1694: Needs documentation (as does the `include_change` parameter I notice, but arguably those counted as "non-ephemeral... known to have been derived under this account" whereas the standalone addresses do not).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -748,6 +748,28 @@ struct StepResult<AccountId> {
     utxos_spent: Vec<OutPoint>,
 }
 
+pub struct SpendingKeys<'a> {
```
str4d: Needs documentation (both inherent, and tying it to the definition of an account per https://github.com/zcash/librustzcash/pull/1768/files#r2334105673

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -766,7 +788,7 @@ pub fn create_proposed_transactions<DbT, ParamsT, InputsErrT, FeeRuleT, ChangeEr
     params: &ParamsT,
     spend_prover: &impl SpendProver,
     output_prover: &impl OutputProver,
-    usk: &UnifiedSpendingKey,
+    spending_keys: &SpendingKeys,
```
str4d, line 807: It is very weird to take `SpendingKeys` as a reference when it is just a wrapper around immutable references. If `SpendingKeys` is not going to have ownership over the in-memory key material, then we should just `#[derive(Clone, Copy)]` (which is safe as it only copies the pointers) and then the caller can either create it inside the method call (and pass ownership), or construct it separately and then pass into multiple calls (which will implicitly copy).
```suggestion
    spending_keys: SpendingKeys,
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1512,7 +1537,7 @@ fn create_proposed_transaction<DbT, ParamsT, InputsErrT, FeeRuleT, ChangeErrT, N
...
```
str4d, line 1557: Ditto:
```suggestion
    spending_keys: SpendingKeys,
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -2433,7 +2473,7 @@ pub fn shield_transparent_funds<DbT, ParamsT, InputsT, ChangeT>(
     input_selector: &InputsT,
     change_strategy: &ChangeT,
     shielding_threshold: Zatoshis,
-    usk: &UnifiedSpendingKey,
+    spending_keys: &SpendingKeys,
```
str4d, line 2494: Ditto:
```suggestion
    spending_keys: SpendingKeys,
```

zcash_client_backend/src/wallet.rs
```diff
@@ -628,27 +628,40 @@ impl OvkPolicy {
 /// This is implicitly scoped to an account.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg(feature = "transparent-inputs")]
-pub struct TransparentAddressMetadata {
-    scope: TransparentKeyScope,
-    address_index: NonHardenedChildIndex,
+pub enum TransparentAddressMetadata {
+    Derived {
```
str4d, line 633: Needs documentation (probably moving the top-level documentation here).

zcash_client_backend/src/wallet.rs
```diff
@@ -628,27 +628,40 @@ impl OvkPolicy {
 /// This is implicitly scoped to an account.
```
str4d: This documentation is now incorrect.

zcash_client_backend/src/wallet.rs
```diff
@@ -628,27 +628,40 @@ impl OvkPolicy {
...
 #[cfg(feature = "transparent-inputs")]
-pub struct TransparentAddressMetadata {
-    scope: TransparentKeyScope,
-    address_index: NonHardenedChildIndex,
+pub enum TransparentAddressMetadata {
+    Derived {
+        scope: TransparentKeyScope,
+        address_index: NonHardenedChildIndex,
+    },
+    Standalone(secp256k1::PublicKey),
```
str4d, line 641: Needs documentation.

zcash_client_sqlite/src/lib.rs
```diff
@@ -940,18 +940,20 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> WalletRea
...
     ) -> Result<HashMap<TransparentAddress, Option<TransparentAddressMetadata>>, Self::Error> {
-        let key_scopes: &[KeyScope] = if include_change {
-            &[KeyScope::EXTERNAL, KeyScope::INTERNAL]
-        } else {
-            &[KeyScope::EXTERNAL]
-        };
+        let key_scopes = Some(KeyScope::EXTERNAL)
+            .iter()
+            .chain(include_change.then_some(KeyScope::INTERNAL).iter())
+            .chain(include_standalone.then_some(KeyScope::Foreign).iter())
```
str4d: These `.iter`s should not be necessary, as `.chain()` takes `IntoIter`:
```suggestion
            .chain(include_change.then_some(KeyScope::INTERNAL))
            .chain(include_standalone.then_some(KeyScope::Foreign))
```
If this results in a compilation error, that implies the initial `.iter()` before the first `.chain()` should be an `.into_iter()`.

The other way I work around those kinds of errors is to use `std::iter::empty()` (usually with a `use std::iter`) at the start, and then add everything in with `.chain()` so no explicit `.iter()` or `.into_iter()` is necessary.

str4d (changes requested):
Partial review of 8ae3c47fd90969098f4ae2c96a5b887097447ea8, flushing my comments.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: It ends up that to make a USK-only constructor, it needs to take ownership anyway because you can't reference an empty hashmap created with `HashMap::new`.

zcash_client_backend/src/data_api.rs
```diff
@@ -599,50 +621,72 @@ impl<A: Copy> Account for (A, UnifiedIncomingViewingKey) {
...
+        transparent_key_scope: Option<TransparentKeyScope>,
+    },
+    /// No derivation information is available; this is common for imported addresses.
+    #[cfg(feature = "transparent-key-import")]
+    Standalone,
+}
+
+impl AddressSource {
+    #[cfg(feature = "transparent-inputs")]
+    pub fn transparent_key_scope(&self) -> Option<&TransparentKeyScope> {
```
str4d, line 643: Needs documentation.

zcash_client_sqlite/src/wallet/init/migrations/support_zcashd_wallet_import.rs
str4d, line 1: I believe this migration has not yet been released, but please confirm this @nuttycom.

Separately, we should probably start maintaining an internal constant for "leaf migrations as of this commit", which we can then just copy to a new public constant as part of the release process for this crate.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -339,7 +388,15 @@ pub(crate) fn select_addrs_to_reserve<P: consensus::Parameters>(
...
                         a,
-                        TransparentAddressMetadata::new(key_scope.into(), i),
+                        match <Option<TransparentKeyScope>>::from(key_scope) {
+                            Some(scope) => TransparentAddressMetadata::Derived {
+                                scope,
+                                address_index: i,
+                            },
+                            None => {
+                                todo!("Use the standalone pubkey associated with the address")
+                            }
```
str4d, line 398: We can't derive addresses within the gap limit for standalone addresses, so shouldn't it be an error to call `select_addrs_to_reserve` with `KeyScope::Foreign`? Or alternatively, require the caller to provide `TransparentKeyScope`?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -382,7 +439,8 @@ pub(crate) fn reserve_next_n_addresses<P: consensus::Parameters>(
 
     if addresses_to_reserve.len() < n {
         return Err(SqliteClientError::ReachedGapLimit(
-            key_scope.into(),
+            <Option<TransparentKeyScope>>::from(key_scope)
+                .expect("reservation relies on key derivation"),
```
str4d, line 443: Likewise, we could require the caller to provide `TransparentKeyScope`?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -680,6 +680,41 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+        INSERT INTO addresses (
+          account_id, key_scope, address, cached_transparent_receiver_address,
+          receiver_flags, imported_transparent_receiver_pubkey
+        )
+        SELECT
+          id, :key_scope, :address, :address,
+          :receiver_flags, :imported_transparent_receiver_pubkey
+          FROM accounts
+          WHERE accounts.uuid = :account_uuid
+        ON CONFLICT (imported_transparent_receiver_pubkey) DO NOTHING
```
str4d, line 730: Hmm, this behaviour is not documented, and feels weird. I don't like the idea that the caller can think they added a transparent address to an account when it was actually silently ignored because it already existed in a different account. I think this should be an error.

zcash_client_backend/CHANGELOG.md
```diff
@@ -118,6 +132,31 @@ workspace.
...
+- `zcash_client_backend::wallet::TransparentAddressMetadata::scope` now
+  returns `Option<TransparentKeyScope>` instead of `TransparentKeyScope`.
+- `zcash_client_backend::wallet::TransparentAddressMetadata::address_index`
+  now returns `Option<NonHardenedChildIndex>` instead of `NonHardenedChildIndex`.
+
+### Removed
+- `zcash_client_backend::data_api::AddressInfo::diversifier_index`
 
 ### Removed
 - `zcash_client_backend::data_api::SpendableNotes` (renamed to `ReceivedNotes`)
```
```suggestion
### Removed
- `zcash_client_backend::data_api`:
  - `AddressInfo::diversifier_index` (use `AddressInfo::source` instead)
  - `SpendableNotes` (renamed to `ReceivedNotes`)
```

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Yeah, this should be an error.

zcash_client_sqlite/src/wallet.rs
nuttycom: mmm... yeah, okay. It's annoying to make it an error because that requires an entirely separate query (or parsing the conflict error string) but I guess that's not too awful, given that this isn't going to be called in a hot loop.

zcash_client_sqlite/src/wallet/init/migrations/support_zcashd_wallet_import.rs
nuttycom: That's correct; we haven't made a `zcash_client_sqlite` release since 0.17.3, and this migration is entirely new since then.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -314,6 +314,12 @@ pub const V_0_17_3: &[Uuid] = &[
     fix_v_transactions_expired_unmined::MIGRATION_ID,
 ];
 
+/// Leaf migrations as of the current repository state.
+pub const CURRENT_LEAF_MIGRATIONS: &[Uuid] = &[
```
str4d, line 318: Hmm, I guess it's fine for this to be in a release, as it would always be equal to the release state (assuming we remember to create it).

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -314,6 +314,12 @@ pub const V_0_17_3: &[Uuid] = &[
...
+    fix_v_transactions_expired_unmined::MIGRATION_ID,
+    support_zcashd_wallet_import::MIGRATION_ID,
```
str4d, line 321: Bug: this is missing one.
```suggestion
    tx_retrieval_queue_expiry::MIGRATION_ID,
    fix_v_transactions_expired_unmined::MIGRATION_ID,
    support_zcashd_wallet_import::MIGRATION_ID,
```

zcash_client_sqlite/src/error.rs
```diff
@@ -160,6 +163,10 @@ pub enum SqliteClientError {
         expected: BlockHeight,
         actual: BlockHeight,
     },
+
+    /// An
```
str4d: Unfinished documentation.

PR #1780, opened 2025-04-25
fix comment

AloeareV:


zcash_client_backend/proto/service.proto
```diff
@@ -174,7 +174,9 @@ service CompactTxStreamer {
     // Submit the given transaction to the Zcash network
     rpc SendTransaction(RawTransaction) returns (SendResponse) {}
 
-    // Return the txids corresponding to the given t-address within the given block range
+    // Return RawTransactions corresponding to the given t-address within the given block range
+    //
+    // Note: This function is misnamed, it returns complete rawtransactions, not txids
```
nuttycom, line 179:
```suggestion
    // Return RawTransactions that match the given transparent address filter.
    //
    // Note: This function is misnamed, it returns complete `RawTransaction` values, not TxIds.
```

nuttycom (approved):
doc-only ACK with suggestion.

nuttycom (approved):
doc-only ACK b4e9a152c9f5b75bc1612577565559bfb156e83e

zcash_client_backend/proto/service.proto
```diff
@@ -174,7 +174,9 @@ service CompactTxStreamer {
     // Submit the given transaction to the Zcash network
     rpc SendTransaction(RawTransaction) returns (SendResponse) {}
 
-    // Return the txids corresponding to the given t-address within the given block range
+    // Return RawTransactions that match the given transparent address filter.
+    //
+    // Note: This function is misnamed, it returns complete `RawTransaction` values, not TxIds.
```
str4d, line 179: Changes to `.proto` files must always be paired in their commits with the equivalent changes to the generated source code (CI checks for this).

str4d (changes requested):
Reviewed b4e9a152c9f5b75bc1612577565559bfb156e83e.

Instead of making a targeted change, what should be done instead is that the entire `service.proto` file should be copied over as-is from the `lightwalletd` repo (which is its canonical source). I believe this comment is present there, but if it isn't then this PR should be made there.

str4d (approved):
utACK df4d7fbbac1c626adc5e0ac743acbf2d68671630. Please ensure this change is also made to the source file in the `lightwalletd` repo.

PR #1781, opened 2025-04-25
Add transparent data to the `CompactBlock` format.

nuttycom:
This also modifies the BlockRange` type to allow callers to specify a set of pool types for which data should be returned. In the case that no pool types are specified, a server should return compact blocks containing only data relevant to the shielded (Sapling and Orchard) pools.

This PR now depends on zcash/lightwallet-protocol#1

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -42,7 +42,6 @@ workspace.
 - `zcash_client_backend::wallet::WalletTransparentOutput::with_known_input_size`
 
 ### Changed
-- Updated to `lightwallet_protocol v0.4.1`
 - `zcash_client_backend::data_api::error::Error::MemoForbidden` has been replaced
   by `Error::Payment(zip321::PaymentError)`, which can represent both memo-to-transparent
   and zero-valued-transparent-output errors.
@@ -58,7 +57,7 @@ workspace.
   `proposed_version: Option<TxVersion>` parameter under the `unstable` feature
   flag.
 - Migrated to `orchard 0.12`, `sapling-crypto 0.6`, `zip321 0.7`
-- Migrated to `lightwallet-protocol v0.4.0`. This results in the following API
+- Migrated to `lightwallet-protocol v0.4.1`. This results in the following API
   changes:
   - in `zcash_client_backend::proto::compact_formats`:
     - `CompactTx::hash` has been renamed to `CompactTx::txid`


zcash_client_backend/proto/service.proto
```diff
@@ -8,18 +8,31 @@ option go_package = "lightwalletd/walletrpc";
...
+    ORCHARD = 2;
+}
+
 // A BlockID message contains identifiers to select a block: a height or a
 // hash. Specification by hash is not implemented, but may be in the future.
 message BlockID {
-     uint64 height = 1;
-     bytes hash = 2;
+    uint64 height = 1;
+    bytes hash = 2;
```
nuttycom, line 22: There's something deeply annoying here: because the `BlockID` type is used directly as the argument to a GRPC request, and `BlockRange` specifies two `BlockID`s, we can't add the set of pool types here without the meaning of `BlockRange` becoming confusing. 

GRPC interfaces should be defined in terms of explicit `Request` types, so that additive changes of this sort are not invasive. Of course, `BlockID` should have been defined as a `oneof` in the first place, but that's a separate complaint. Really we should overhaul this whole interface, and create a `V2` API endpoint service type that does things properly.

zcash_client_backend/proto/service.proto
```diff
@@ -8,18 +8,31 @@ option go_package = "lightwalletd/walletrpc";
 option swift_prefix = "";
 import "compact_formats.proto";
 
+// An identifier for a Zcash value pool.
+enum PoolType {
+    TRANSPARENT = 0;
+    SAPLING = 1;
+    ORCHARD = 2;
```
daira: These are not the internal values used in librustzcash. Check whether these discriminants are ever exposed by language bindings, and if they are, consider making them consistent.

zcash_client_backend/proto/service.proto
```diff
@@ -8,18 +8,31 @@ option go_package = "lightwalletd/walletrpc";
...
 // Both BlockIDs must be heights; specification by hash is not yet supported.
+//
+// A server may optionally prune `CompactBlocks` returned to include only data
+// relevant to the requested pool types. If no pool types are specified, the
+// server should default to the legacy behavior of returning only data relevant
+// to the shielded (Sapling and Orchard) pools.
 message BlockRange {
     BlockID start = 1;
     BlockID end = 2;
+    repeated PoolType poolTypes = 3;
```
daira: `BlockRange` is also used in `TransparentAddressBlockFilter`, where it does not make sense to restrict the pool types. We should document that `poolTypes` MUST NOT be present when it is used in `TransparentAddressBlockFilter` (which is the argument type to `rpc GetTaddressTxids`).

zcash_client_backend/proto/service.proto
pacu: This might be more cumbersome than it seems

The [Field Presence](https://protobuf.dev/programming-guides/field_presence/) is quite deceiving at some point when it comes to values that might or might not be present in a message. Also the absence of the value can be filled with a default value that for enums is the value 0. see https://protobuf.dev/programming-guides/proto3/#default


Also when it comes to enums Swift and Kotlin treatment of enums differ. https://protobuf.dev/programming-guides/enum/ 

IDK if this can be avoided or not.

zcash_client_backend/proto/service.proto
```diff
@@ -8,18 +8,31 @@ option go_package = "lightwalletd/walletrpc";
 option swift_prefix = "";
 import "compact_formats.proto";
 
+// An identifier for a Zcash value pool.
+enum PoolType {
+    TRANSPARENT = 0;
```
pacu: According to best practices: https://protobuf.dev/best-practices/dos-donts/#unspecified-enum enums should include an unspecified value. This probably is because zero is the default value you never get a default value by accident.

zcash_client_backend/proto/service.proto
pacu: > Note that this is not generally safe for numeric types, including bools and enums.

https://protobuf.dev/programming-guides/proto3/#conditionally-safe-changes

zcash_client_backend/proto/service.proto
daira: Thanks, I hate it. (Yes we should do this. See https://protobuf.dev/programming-guides/enum/ for why we don't want a closed enum, therefore we have to define the useless "unspecified" value.)

zcash_client_backend/proto/service.proto
daira: For `service.proto` as of this PR, `PoolType` is only used as `repeated PoolType poolTypes = 3;` in `message BlockRange`, so the default for the repeated field will be that no pool types are specified. However, to avoid a footgun in case `PoolType` is used as a singular (not repeated) field, I agree that 0 should be added as an unspecified value.

zcash_client_backend/proto/service.proto
daira: That is not applicable because it is talking about whether "singular" fields are compatible with "repeated" fields. Here we're not changing an existing singular field.

zcash_client_backend/proto/service.proto
daira:
```suggestion
    // do not use `POOLTYPE_UNSPECIFIED`
    POOLTYPE_UNSPECIFIED = 0;
    TRANSPARENT = 1;
    SAPLING = 2;
    ORCHARD = 3;
```

zcash_client_backend/proto/service.proto
nuttycom: Fixed in https://github.com/zcash/lightwallet-protocol/pull/1

zcash_client_backend/proto/service.proto
nuttycom: Fixed in https://github.com/zcash/lightwallet-protocol/pull/1

nuttycom: We only use proper types in librustzcash, and the encodings for each purpose (for example, for the sqlite bindings) are isolated from each other.

zcash_client_backend/proto/service.proto
nuttycom: We should just document that the argument is ignored.

zcash_client_backend/lightwallet-protocol/walletrpc/compact_formats.proto
```diff
@@ -55,6 +55,32 @@ message CompactTx {
...
     repeated CompactSaplingOutput outputs = 5;
     repeated CompactOrchardAction actions = 6;
+    repeated CompactTxIn vin = 7;
+    repeated TxOut vout = 8;
+}
+
+// The unique identifier for a transparent transaction output.
+message OutPoint {
+    // The txid of the transaction that generated this output.
+    bytes txid = 1;
```
pacu: I believe we have a limitation when it comes to representing coinbase OutPoints if we want to account for them in the compact block model

this is the full block data for a regular Tx
```JSON
  "vin": [
    {
      "vout": 0,
      "valueSat": 97191200,
      "value": 0.971912,
      "txid": "85ada8b5633ff186f845217bb9ce30712c060b35a0705d30121388b714c8200d",
      "sequence": 4294967295,
      "scriptSig": {
        "hex": "483045022100aeafe0c5a65fc815ba709a19cba81989db7444a324b5ad9cab9f41cdfaee7e3c022009288390289222180b2668a27dc78b4f9a7000409b77005c80200ef828404465012102fbb01f68144bb27ee4dd5c9272f4ec7cd326ba4afa059268949d1c274dabc7c2",
        "asm": "3045022100aeafe0c5a65fc815ba709a19cba81989db7444a324b5ad9cab9f41cdfaee7e3c022009288390289222180b2668a27dc78b4f9a7000409b77005c80200ef828404465[ALL] 02fbb01f68144bb27ee4dd5c9272f4ec7cd326ba4afa059268949d1c274dabc7c2"
      },
      "address": "t1MgWgsdSb4s32JECaN7rPzStsj6RQTPRhU"
    }
```

This is the vin for a coinbase transaction
```JSON
  "vin": [
    {
      "sequence": 4294967295,
      "coinbase": "03bd2b2f00"
    }
  ],
```

does it make sense to actually represent this? Is this of any relevance for a light-client?

nuttycom: For a coinbase transaction, the `vout` array will have the outputs generated by the transaction, and `vin` will be empty. What does the light client need that `vin` data for?

nuttycom: If it wants the input metadata for coinbase, it can just download the full transaction, same as if it wants to get the memo.

zcash_client_backend/CHANGELOG.md
```diff
@@ -40,8 +40,17 @@ workspace.
...
+  - `OutPoint`
+  - `CompactTxIn`
+  - `TxOut`
+  - `CompactTx::{vin, vout}`
+- `zcash_client_backend::proto::service`:
+  - `PoolType`
+  - `BlockRange::pool_types`
 
 ### Changed
+- Updated to `lightwallet_protocol v0.4.1`
```
str4d: This should instead update the existing changelog entry.
```suggestion
```

zcash_client_backend/CHANGELOG.md
```suggestion
- Migrated to `lightwallet-protocol v0.4.1`. This results in the following API
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -40,8 +40,17 @@ workspace.
...
   - `zcash_client_backend::wallet::TransparentAddressMetadata::redeem_script()`
 - `zcash_client_backend::wallet::WalletTransparentOutput::with_known_input_size`
+- `zcash_client_backend::proto::compact_formats`:
+  - `OutPoint`
+  - `CompactTxIn`
+  - `TxOut`
+  - `CompactTx::{vin, vout}`
+- `zcash_client_backend::proto::service`:
+  - `PoolType`
+  - `BlockRange::pool_types`
```
str4d: Most of these are already in the changelog above here. Merge these into there instead.

PR #1786, opened 2025-05-01
Add OP_RETURN and external signature support to builders

zlyzol:
This PR adds necessary methods to the transaction builders to support workflows where signing happens externally (e.g., TSS,) and where OP_RETURN memos are needed.

**Key Additions:**

1.  **`add_transparent_op_return`:** Allows adding `OP_RETURN` outputs via the `Builder`.
2.  **`apply_external_signatures`:** Enables authorizing a transparent bundle using signatures generated elsewhere.

**Implementation Details:**

*   The `OpCode` enum in `zcash_transparent` is made public to facilitate `OP_RETURN` script construction.
*   An `InvalidOpReturn` error variant is added.
*   The `apply_external_signatures` method constructs standard P2PKH `scriptSig`s using the provided DER signatures and public keys from the input context.

These changes were developed as part of integrating Zcash into MAYAChain.

### DIFF
--- a/zcash_transparent/src/address.rs
+++ b/zcash_transparent/src/address.rs
@@ -14,7 +14,9 @@ use zcash_encoding::Vector;
 /// Most of the opcodes are unused by this crate, but we define them so that the alternate
 /// `Debug` impl for [`Script`] renders correctly for unexpected scripts.
 #[derive(Debug)]
-enum OpCode {
+
+// Maya Protocol integration needs OpCode to be public for add_op_return_output implementation
+pub enum OpCode {
     // push value
     Op0 = 0x00, // False
     PushData1 = 0x4c,
--- a/zcash_transparent/src/builder.rs
+++ b/zcash_transparent/src/builder.rs
@@ -404,4 +404,55 @@ impl Bundle<Unauthorized> {
             authorization: Authorized,
         })
     }
+    
+    pub fn apply_external_signatures(
+        self,
+        #[cfg(feature = "transparent-inputs")] signatures: Vec<secp256k1::ecdsa::Signature>,
+    ) -> Bundle<Authorized> {
+        #[cfg(feature = "transparent-inputs")]
+        let script_sigs = { 
+            // Check that the number of signatures matches the number of inputs
+            assert_eq!(
+                self.authorization.inputs.len(),
+                signatures.len(),
+                "Number of signatures must match number of inputs"
+            );
+
+            // Compute the scriptSigs
+            self
+            .authorization
+            .inputs
+            .iter()
+            .zip(signatures)
+            .map(|(info, signature)| { // For each (input info, signature) pair...
+                // Serialize the provided signature to DER format
+                let mut sig_bytes: Vec<u8> = signature.serialize_der().to_vec();
+                // Append the SIGHASH_ALL byte (0x01)
+                sig_bytes.push(SIGHASH_ALL);
+
+                // Construct the P2PKH scriptSig: <DER sig + SIGHASH byte> <compressed pubkey>
+                // Assumes P2PKH inputs. Needs info.pubkey which comes from TransparentInputInfo.
+                Script::default() << &sig_bytes[..] << &info.pubkey.serialize()[..]
+            })
+        };
+
+        #[cfg(not(feature = "transparent-inputs"))]
+        let script_sigs = std::iter::empty::<Script>();
+
+        // Construct the new authorized Bundle
+        Bundle {
+            vin: self
+                .vin // Take the original vin (which has empty script_sig fields)
+                .iter()
+                .zip(script_sigs) // Pair the old TxIn with the newly computed scriptSig
+                .map(|(txin, sig)| TxIn {
+                    prevout: txin.prevout.clone(),
+                    script_sig: sig, // Assign the computed scriptSig
+                    sequence: txin.sequence,
+                })
+                .collect(),
+            vout: self.vout, // Keep the original vout
+            authorization: Authorized,
+        }
+    }
 }
--- a/zcash_transparent/src/builder.rs
+++ b/zcash_transparent/src/builder.rs
@@ -13,6 +13,8 @@ use crate::{
     sighash::{SignableInput, TransparentAuthorizingContext},
 };
 
+use crate::address::OpCode;
+
 #[cfg(feature = "transparent-inputs")]
 use {
     crate::{
@@ -28,6 +30,7 @@ pub enum Error {
     InvalidAmount,
     /// A bundle could not be built because a required signing keys was missing.
     MissingSigningKey,
+    InvalidOpReturn,
 }
 
 impl fmt::Display for Error {
@@ -36,6 +39,7 @@ impl fmt::Display for Error {
             Error::InvalidAddress => write!(f, "Invalid address"),
             Error::InvalidAmount => write!(f, "Invalid amount"),
             Error::MissingSigningKey => write!(f, "Missing signing key"),
+            Error::InvalidOpReturn => write!(f, "Invalid op_return"),
         }
     }
 }
@@ -275,6 +279,28 @@ impl TransparentBuilder {
 
             Some(pczt::Bundle { inputs, outputs })
         }
+    }
+
+    /// Adds a zero-value OP_RETURN output containing the given data.
+    pub fn add_op_return_output(&mut self, data: &[u8]) -> Result<(), Error> {
+        // Check 80 bytes limit.
+        const MAX_OP_RETURN_RELAY_BYTES: usize = 80;
+        if data.len() > MAX_OP_RETURN_RELAY_BYTES {
+            // Use the appropriate error variant from the builder's Error enum
+            // InvalidOpReturn might need to be added if it doesn't exist.
+            return Err(Error::InvalidOpReturn);
+        }
+
+        // Construct the OP_RETURN script using the Shl (<<) operators
+        let script = Script::default() // Start with an empty script
+                     << OpCode::Return  // Append the OP_RETURN opcode
+                     << data;           // Append the data push opcode(s) and the data itself
+
+        self.vout.push(TxOut {
+             value: Zatoshis::ZERO,
+             script_pubkey: script,
+        });
+        Ok(())
     }
 }
 
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -521,6 +521,16 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_, 
         self.transparent_builder.add_output(to, value)
     }
 
+    /// Adds a transparent OP_RETURN output with the given data payload.
+    pub fn add_transparent_op_return<FE>(
+        &mut self, 
+        data: &[u8]
+    ) -> Result<(), Error<FE>> {
+        self.transparent_builder
+           .add_op_return_output(data)
+           .map_err(Error::TransparentBuild)
+   }
+
     /// Returns the sum of the transparent, Sapling, Orchard, and TZE value balances.
     fn value_balance(&self) -> Result<ZatBalance, BalanceError> {
         let value_balances = [


zcash_transparent/src/builder.rs
```diff
@@ -276,6 +280,28 @@ impl TransparentBuilder {
             Some(pczt::Bundle { inputs, outputs })
         }
     }
+
+    /// Adds a zero-value OP_RETURN output containing the given data.
+    pub fn add_op_return_output(&mut self, data: &[u8]) -> Result<(), Error> {
+        // Check 80 bytes limit.
+        const MAX_OP_RETURN_RELAY_BYTES: usize = 80;
```
sellout, line 287: I know this is just ported from the C++, but I was curious why 80 was chosen. I asked the core team, and @str4d found https://github.com/bitcoin/bitcoin/pull/2738#discussion_r6530132, in case anyone else is curious.

zcash_transparent/src/address.rs
```diff
@@ -14,7 +14,9 @@ use zcash_encoding::Vector;
 /// Most of the opcodes are unused by this crate, but we define them so that the alternate
 /// `Debug` impl for [`Script`] renders correctly for unexpected scripts.
 #[derive(Debug)]
-enum OpCode {
+
+// Maya Protocol integration needs OpCode to be public for add_op_return_output implementation
+pub enum OpCode {
```
sellout, line 19:
```suggestion
pub(crate) enum OpCode {
```

I think the opcodes are likely to end up fully `pub` eventually, but probably not until the switch to the Rust implementation of zcash_script.

zcash_transparent/src/builder.rs
```diff
@@ -276,6 +280,28 @@ impl TransparentBuilder {
...
+        if data.len() > MAX_OP_RETURN_RELAY_BYTES {
+            // Use the appropriate error variant from the builder's Error enum
+            // InvalidOpReturn might need to be added if it doesn't exist.
```
sellout, line 290:
```suggestion
```
The comment has been addressed, and doesn’t add anything to the implementation.

zcash_transparent/src/builder.rs
```diff
@@ -276,6 +280,28 @@ impl TransparentBuilder {
...
+            return Err(Error::InvalidOpReturn);
+        }
+
+        // Construct the OP_RETURN script using the Shl (<<) operators
+        let script = Script::default() // Start with an empty script
+                     << OpCode::Return  // Append the OP_RETURN opcode
+                     << data;           // Append the data push opcode(s) and the data itself
```
sellout, line 297:
```suggestion
        let script = Script::default() << OpCode::Return << data;
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +404,55 @@ impl Bundle<Unauthorized> {
...
         })
     }
+    
+    pub fn apply_external_signatures(
+        self,
+        #[cfg(feature = "transparent-inputs")] signatures: Vec<secp256k1::ecdsa::Signature>,
+    ) -> Bundle<Authorized> {
+        #[cfg(feature = "transparent-inputs")]
+        let script_sigs = { 
+            // Check that the number of signatures matches the number of inputs
```
sellout, line 414:
```suggestion
```
The assert message says the same thing as the comment.

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +404,55 @@ impl Bundle<Unauthorized> {
...
+            self
+            .authorization
+            .inputs
+            .iter()
+            .zip(signatures)
+            .map(|(info, signature)| { // For each (input info, signature) pair...
+                // Serialize the provided signature to DER format
+                let mut sig_bytes: Vec<u8> = signature.serialize_der().to_vec();
+                // Append the SIGHASH_ALL byte (0x01)
+                sig_bytes.push(SIGHASH_ALL);
```
sellout, line 431:
```suggestion
            .map(|(info, signature)| {
                let mut sig_bytes: Vec<u8> = signature.serialize_der().to_vec();
                sig_bytes.push(SIGHASH_ALL);
```

sellout (approved):
utACK

Some non-blocking suggestions, but it looks good.

zcash_transparent/src/address.rs
pacu: this also lacks a CHANGELOG entry if they went `pub`

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +404,55 @@ impl Bundle<Unauthorized> {
             authorization: Authorized,
         })
     }
+
+    pub fn apply_external_signatures(
+        self,
```
pacu, line 409: this lacks a CHANGELOG entry as well.

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +404,55 @@ impl Bundle<Unauthorized> {
             authorization: Authorized,
         })
     }
+
+    pub fn apply_external_signatures(
```
pacu, line 408: is this function called anywhere in this PR? are there any follow-up changes?

zcash_transparent/src/builder.rs
zlyzol: I think the 80 bytes here was chosen to align with the standard relay policies for OP_RETURN on Bitcoin and, by extension, for Zcash's transparent layer,

zcash_transparent/src/builder.rs
zlyzol: No, this `apply_external_signatures` function isn't called directly in this PR's. It's used by the Mayanode Bifrost service, which provides the external signatures.

zcash_transparent/src/builder.rs
zlyzol: accepted

zcash_transparent/src/builder.rs
zlyzol: added

zcash_transparent/src/address.rs
```diff
@@ -14,7 +14,8 @@ use zcash_encoding::Vector;
 /// Most of the opcodes are unused by this crate, but we define them so that the alternate
 /// `Debug` impl for [`Script`] renders correctly for unexpected scripts.
 #[derive(Debug)]
-enum OpCode {
+
```
str4d, line 17:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +30,7 @@ pub enum Error {
     InvalidAmount,
     /// A bundle could not be built because a required signing keys was missing.
     MissingSigningKey,
+    InvalidOpReturn,
```
str4d, line 33: - Document this new variant.
- Document this change to the enum in the changelog (adding variants to a public enum is a breaking change if it doesn't have a `nonexhaustive` annotation).

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +39,7 @@ impl fmt::Display for Error {
             Error::InvalidAddress => write!(f, "Invalid address"),
             Error::InvalidAmount => write!(f, "Invalid amount"),
             Error::MissingSigningKey => write!(f, "Missing signing key"),
+            Error::InvalidOpReturn => write!(f, "Invalid op_return"),
```
str4d, line 42:
```suggestion
            Error::InvalidOpReturn => write!(f, "Invalid OP_RETURN"),
```

zcash_transparent/src/builder.rs
```diff
@@ -276,6 +280,23 @@ impl TransparentBuilder {
             Some(pczt::Bundle { inputs, outputs })
         }
     }
+
+    /// Adds a zero-value "null data" (OP_RETURN) output containing the given data.
+    pub fn add_null_data_output(&mut self, data: &[u8]) -> Result<(), Error> {
+        // Check 80 bytes limit.
+        const MAX_OP_RETURN_RELAY_BYTES: usize = 80;
+        if data.len() > MAX_OP_RETURN_RELAY_BYTES {
+            return Err(Error::InvalidOpReturn);
```
str4d, line 289: This is not actually an invalid `OP_RETURN`, as the 80-byte limit is a standard rule rather than a consensus rule. However, in practice it means that this is an unusable length.

Rename the error variant to `Error::NullDataTooLong` and document it as "the provided null data is longer than the maximum supported length" (and consider storing both the data length and the maximum length inside the error variant and including them in the `Display` impl).

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +399,54 @@ impl Bundle<Unauthorized> {
             authorization: Authorized,
         })
     }
+
+    pub fn apply_external_signatures(
```
str4d, line 403: Document this new method. (I know some of the existing methods are undocumented, but we fix that by ensuring that new methods do get documented as we add them.)

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +399,54 @@ impl Bundle<Unauthorized> {
...
+        self,
+        #[cfg(feature = "transparent-inputs")] signatures: Vec<secp256k1::ecdsa::Signature>,
```
str4d, line 405: I do not like making individual arguments to methods conditionally compiled in a public library API, as that leads to all sorts of annoying breakage. (We do this in a few public APIs, but as much as possible only in crates that are effectively "app crates" and thus less susceptible to feature-flag-at-a-distance changes; `zcash_transparent` is not such a crate).

Instead this entire method should be behind the feature flag, and should just append signatures, with separate `prepare` and `finalize` methods like [`orchard::builder`](https://github.com/zcash/orchard/blob/fcb14defd75c0dd79512289c17a1ac46b5001d3a/src/builder.rs#L970-L1098) has.

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +399,54 @@ impl Bundle<Unauthorized> {
...
+                self.authorization.inputs.len(),
+                signatures.len(),
+                "Number of signatures must match number of inputs"
+            );
+
+            // Compute the scriptSigs
+            self.authorization
+                .inputs
+                .iter()
+                .zip(signatures)
```
str4d, line 419: Bug: this has a silent undocumented requirement that the caller provide the signatures in the correct order. This method should instead try each provided signature against each input that currently lacks a signature, using signature verification to determine whether it is for that input. (This is exactly what [`orchard::builder`](https://github.com/zcash/orchard/blob/fcb14defd75c0dd79512289c17a1ac46b5001d3a/src/builder.rs#L1052-L1079) does.)

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +399,54 @@ impl Bundle<Unauthorized> {
...
+                .map(|(info, signature)| {
+                    // For each (input info, signature) pair...
+                    // Serialize the provided signature to DER format
+                    let mut sig_bytes: Vec<u8> = signature.serialize_der().to_vec();
+                    // Append the SIGHASH_ALL byte (0x01)
+                    sig_bytes.push(SIGHASH_ALL);
+
+                    // Construct the P2PKH scriptSig: <DER sig + SIGHASH byte> <compressed pubkey>
+                    // Assumes P2PKH inputs. Needs info.pubkey which comes from TransparentInputInfo.
+                    Script::default() << &sig_bytes[..] << &info.pubkey.serialize()[..]
```
str4d, line 429: Bug: this will silently construct a `Bundle<Authorized>` with invalid signatures if the caller makes a mistake. The fix to this bug is same as above: verify the provided signature in the context of this input before using it.

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +399,54 @@ impl Bundle<Unauthorized> {
...
+        self,
+        #[cfg(feature = "transparent-inputs")] signatures: Vec<secp256k1::ecdsa::Signature>,
+    ) -> Bundle<Authorized> {
+        #[cfg(feature = "transparent-inputs")]
+        let script_sigs = {
+            assert_eq!(
+                self.authorization.inputs.len(),
+                signatures.len(),
+                "Number of signatures must match number of inputs"
+            );
```
str4d, line 413: Undocumented panic. The fix to this is the same as above, and the same as what `orchard::builder` does: try each provided signature against the remaining unsigned inputs. If a signature is provided that does not match any unsigned input, then an error can be returned instead of a panic.

str4d (changes requested):
Reviewed d7c098de2cf4bdf5f7f67c101e360f3d212e667d.

zcash_transparent/src/builder.rs
zlyzol: All new variants are documented now

zcash_transparent/src/builder.rs
zlyzol: renamed

zcash_transparent/src/builder.rs
zlyzol: All new methods are documented now.

zcash_transparent/src/builder.rs
zlyzol: I've refactored this by introducing a `TransparentSignatureContext` to manage the state for applying external signatures.
This enables a staged approach with prepare, append, and finalize methods, similar to orchard::builder. This design keeps the existing Bundle structure untouched. The original method has been removed.

zcash_transparent/src/builder.rs
zlyzol: Solved by refactoring.

zcash_transparent/src/builder.rs
zlyzol: Solved by refactoring.

zcash_transparent/src/builder.rs
zlyzol: Solved.

zcash_transparent/src/builder.rs
zlyzol: I also considered making the Bundle itself a state machine with an intermediate Bundle<PartiallyAuthorizedTransparent> state (where TxIn would use Option<Script> for signatures), which would align closely with the orchard::Bundle pattern. I opted for TransparentSignatureContext to minimize changes to the existing Bundle and TxIn structures. However, I'm open to feedback if you feel the Bundle-integrated state machine approach would be a better fit here.

daira (approved):
utACK

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
+  - `InputCountMismatch`
+  - `SighashGeneration`
+  - `SignatureVerificationFailed { input_index: usize }`
+  - `NoMatchingInputForSignature { input_index: usize }`
+  - `InputAlreadySigned { input_index: usize }`
+  - `AmbiguousSignature`
+  - `NotAllInputsSigned`
+  - `UnusedExternalSignature { input_index: usize }`
+  - `Secp256k1Error(String)`
+  - `InternalBuilderError(String)`
```
str4d: Non-blocking: these need to be moved under `Changed` (because as I said in my previous review, they are breaking changes to a regular enum, not additions to a non-exhaustive enum).

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_transaction::builder::Builder::add_transparent_null_data_output`
```
str4d: This path is wrong:
```suggestion
- `zcash_primitives::transaction::builder::Builder::add_transparent_null_data_output`
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +475,136 @@ impl Bundle<Unauthorized> {
...
+        }
+
+        let num_sigs = signatures.len();
+        let mut sig_is_used = vec![false; num_sigs];
+
+        // Iterate through each input that is not yet signed from previous calls
+        for input_idx in 0..num_inputs {
+            if self.final_script_sigs[input_idx].is_some() {
+                continue;
+            }
```
str4d: Non-blocking: This logic ends up more complex than it needs to be, due to iterating over the inputs. Iterating over the signatures (like `orchard` does) would be simpler, and is probably what I will refactor it to do after this PR is merged.

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    InputCountMismatch, // Use this for prepare_transparent_signatures
+    /// A pre-calculated sighash could not be correctly formed or was invalid.
+    SighashGeneration,
+    /// An external signature failed cryptographic verification for a specific transparent input.
+    SignatureVerificationFailed {
+        input_index: usize,
+    },
+    /// A provided external signature did not cryptographically match any available unsigned transparent input.
+    NoMatchingInputForSignature {
+        input_index: usize,
```
str4d: This should not be named `input_index` because it does not index into the transaction inputs. AFAICT it is an index into the provided signatures list, so it needs to be renamed to:
```suggestion
        sig_index: usize,
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    },
+    /// An attempt was made to apply a signature to a transparent input that had already been signed.
+    InputAlreadySigned {
+        input_index: usize,
+    },
```
str4d: Now that the prior bugs have been fixed (so that signatures are not applied by-index), this error case is unused:
```suggestion
```

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
+- `zcash_transparent::builder::Bundle<Unauthorized>::prepare_transparent_signatures` to initialize the signing context.
+- `zcash_transparent::builder::TransparentSignatureContext::append_external_signatures` to add signatures to the context.
+- `zcash_transparent::builder::TransparentSignatureContext::finalize_signatures` to produce an authorized bundle from the context.
+- New error variants to `zcash_transparent::builder::Error`:
+  - `SignatureCountMismatch`
+  - `InputCountMismatch`
+  - `SighashGeneration`
+  - `SignatureVerificationFailed { input_index: usize }`
+  - `NoMatchingInputForSignature { input_index: usize }`
+  - `InputAlreadySigned { input_index: usize }`
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
...
             Error::MissingSigningKey => write!(f, "Missing signing key"),
+            Error::NullDataTooLong { actual, limit } => write!(f, "Provided null data is longer than the maximum supported length (actual: {}, limit: {}", actual, limit),
+            Error::SignatureCountMismatch => write!(f, "The number of provided external signatures does not match the number of transparent inputs"),
+            Error::InputCountMismatch => write!(f, "The number of inputs does not match the number of sighashes"),
+            Error::SighashGeneration => write!(f, "A pre-calculated sighash could not be correctly formed or was invalid"),
+            Error::SignatureVerificationFailed { input_index } => write!(f, "External signature verification failed for transparent input at index {}", input_index),
+            Error::NoMatchingInputForSignature { input_index } => write!(f, "A provided external signature at index {} did not cryptographically match any available unsigned transparent input", input_index),
+            Error::InputAlreadySigned { input_index } => write!(f, "Transparent input at index {} has already been signed", input_index),
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
     InvalidAmount,
     /// A bundle could not be built because a required signing keys was missing.
     MissingSigningKey,
+    /// Provided null data is longer than the maximum supported length
+    NullDataTooLong {
+        actual: usize,
+        limit: usize,
+    },
+    /// The number of provided external signatures does not match the number of transparent inputs.
+    SignatureCountMismatch,
```
str4d: This is also now unused:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
             Error::InvalidAddress => write!(f, "Invalid address"),
             Error::InvalidAmount => write!(f, "Invalid amount"),
             Error::MissingSigningKey => write!(f, "Missing signing key"),
+            Error::NullDataTooLong { actual, limit } => write!(f, "Provided null data is longer than the maximum supported length (actual: {}, limit: {}", actual, limit),
+            Error::SignatureCountMismatch => write!(f, "The number of provided external signatures does not match the number of transparent inputs"),
```
str4d:
```suggestion
```

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
 ## [Unreleased]
 
+### Added
+- `zcash_transparent::builder::TransparentBuilder::add_null_data_output`
+- `zcash_transparent::builder::TransparentSignatureContext` struct for staged application of external signatures.
+- `zcash_transparent::builder::Bundle<Unauthorized>::prepare_transparent_signatures` to initialize the signing context.
+- `zcash_transparent::builder::TransparentSignatureContext::append_external_signatures` to add signatures to the context.
+- `zcash_transparent::builder::TransparentSignatureContext::finalize_signatures` to produce an authorized bundle from the context.
+- New error variants to `zcash_transparent::builder::Error`:
+  - `SignatureCountMismatch`
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    NullDataTooLong {
+        actual: usize,
+        limit: usize,
+    },
+    /// The number of provided external signatures does not match the number of transparent inputs.
+    SignatureCountMismatch,
+    /// The number of inputs does not match number of sighashes.
+    InputCountMismatch, // Use this for prepare_transparent_signatures
+    /// A pre-calculated sighash could not be correctly formed or was invalid.
+    SighashGeneration,
```
str4d: I don't know if this error variant was ever used, but it isn't in the current PR state.
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
             Error::InvalidAddress => write!(f, "Invalid address"),
             Error::InvalidAmount => write!(f, "Invalid amount"),
             Error::MissingSigningKey => write!(f, "Missing signing key"),
+            Error::NullDataTooLong { actual, limit } => write!(f, "Provided null data is longer than the maximum supported length (actual: {}, limit: {}", actual, limit),
+            Error::SignatureCountMismatch => write!(f, "The number of provided external signatures does not match the number of transparent inputs"),
+            Error::InputCountMismatch => write!(f, "The number of inputs does not match the number of sighashes"),
+            Error::SighashGeneration => write!(f, "A pre-calculated sighash could not be correctly formed or was invalid"),
```
str4d:
```suggestion
```

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
+  - `InputCountMismatch`
+  - `SighashGeneration`
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    /// The number of provided external signatures does not match the number of transparent inputs.
+    SignatureCountMismatch,
+    /// The number of inputs does not match number of sighashes.
+    InputCountMismatch, // Use this for prepare_transparent_signatures
+    /// A pre-calculated sighash could not be correctly formed or was invalid.
+    SighashGeneration,
+    /// An external signature failed cryptographic verification for a specific transparent input.
+    SignatureVerificationFailed {
+        input_index: usize,
+    },
```
str4d: This is unused, and also doesn't make sense with the current PR.
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
             Error::InvalidAddress => write!(f, "Invalid address"),
             Error::InvalidAmount => write!(f, "Invalid amount"),
             Error::MissingSigningKey => write!(f, "Missing signing key"),
+            Error::NullDataTooLong { actual, limit } => write!(f, "Provided null data is longer than the maximum supported length (actual: {}, limit: {}", actual, limit),
+            Error::SignatureCountMismatch => write!(f, "The number of provided external signatures does not match the number of transparent inputs"),
+            Error::InputCountMismatch => write!(f, "The number of inputs does not match the number of sighashes"),
+            Error::SighashGeneration => write!(f, "A pre-calculated sighash could not be correctly formed or was invalid"),
+            Error::SignatureVerificationFailed { input_index } => write!(f, "External signature verification failed for transparent input at index {}", input_index),
```
str4d:
```suggestion
```

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
+  - `SignatureVerificationFailed { input_index: usize }`
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+        input_index: usize,
+    },
+    /// A single provided external signature was cryptographically valid for more than one distinct transparent input.
+    AmbiguousSignature,
+    /// Not all transparent inputs that require signing received a valid signature.
+    NotAllInputsSigned,
+    /// One or more provided external signatures were not used to sign any transparent input.
+    UnusedExternalSignature {
+        input_index: usize,
+    },
```
str4d: This error case is a duplicate of `NoMatchingInputForSignature` and is unused.
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
...
+            Error::SighashGeneration => write!(f, "A pre-calculated sighash could not be correctly formed or was invalid"),
+            Error::SignatureVerificationFailed { input_index } => write!(f, "External signature verification failed for transparent input at index {}", input_index),
+            Error::NoMatchingInputForSignature { input_index } => write!(f, "A provided external signature at index {} did not cryptographically match any available unsigned transparent input", input_index),
+            Error::InputAlreadySigned { input_index } => write!(f, "Transparent input at index {} has already been signed", input_index),
+            Error::AmbiguousSignature => write!(f, "A single provided external signature was cryptographically valid for more than one distinct transparent input"),
+            Error::NotAllInputsSigned => write!(f, "Not all transparent inputs that require signing received a valid signature"),
+            Error::UnusedExternalSignature { input_index } => write!(f, "External signature at index {} was not used to sign any transparent input", input_index),
```
str4d:
```suggestion
```

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
+  - `NoMatchingInputForSignature { input_index: usize }`
+  - `InputAlreadySigned { input_index: usize }`
+  - `AmbiguousSignature`
+  - `NotAllInputsSigned`
+  - `UnusedExternalSignature { input_index: usize }`
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    /// The number of inputs does not match number of sighashes.
+    InputCountMismatch, // Use this for prepare_transparent_signatures
+    /// A pre-calculated sighash could not be correctly formed or was invalid.
+    SighashGeneration,
+    /// An external signature failed cryptographic verification for a specific transparent input.
+    SignatureVerificationFailed {
+        input_index: usize,
+    },
+    /// A provided external signature did not cryptographically match any available unsigned transparent input.
+    NoMatchingInputForSignature {
```
str4d: This is used as an equivalent error to `orchard`'s `BuildError::InvalidExternalSignature`, and I prefer the latter's naming (because we cannot distinguish between an external signature being for a different input, and an external signature being intended for one of the inputs but invalid).

This error also is not restricted to available unsigned transparent inputs, because the logic (correctly) checks every provided signature against every transparent input (in order to correctly detect `AmbiguousSignature`).
```suggestion
    /// A provided external signature was not valid for any transparent input.
    InvalidExternalSignature {
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+        input_index: usize,
+    },
+    /// An attempt was made to apply a signature to a transparent input that had already been signed.
+    InputAlreadySigned {
+        input_index: usize,
+    },
+    /// A single provided external signature was cryptographically valid for more than one distinct transparent input.
+    AmbiguousSignature,
```
str4d: This is used like `orchard`'s `BuildError::DuplicateSignature`, and should be named and documented equivalently because it indicates a critical failure:
```suggestion
    /// A signature is valid for more than one input. This should never happen if `alpha`
    /// is sampled correctly, and indicates a critical failure in randomness generation.
    DuplicateSignature,
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    /// Not all transparent inputs that require signing received a valid signature.
+    NotAllInputsSigned,
```
str4d: This is equivalent to `orchard`'s `BuildError::MissingSignatures`, and should be named the same:
```suggestion
    /// A bundle could not be built because required signatures were missing.
    MissingSignatures,
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +475,136 @@ impl Bundle<Unauthorized> {
...
+
+        // Iterate through each input that is not yet signed from previous calls
+        for input_idx in 0..num_inputs {
+            if self.final_script_sigs[input_idx].is_some() {
+                continue;
+            }
+
+            let input_info = &self.authorization_inputs[input_idx];
+            let sighash_msg = secp256k1::Message::from_digest_slice(&self.sighashes[input_idx][..])
+                .map_err(|e| Error::Secp256k1Error(e.to_string()))?;
```
str4d: This can only error if the digest is not 32 bytes long. `secp256k1 0.29` (our current dependency) has [an infallible API](https://docs.rs/secp256k1/0.29.1/secp256k1/struct.Message.html#method.from_digest) that should be used instead:
```suggestion
            let sighash_msg = secp256k1::Message::from_digest(*self.sighashes[input_idx]);
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    /// A single provided external signature was cryptographically valid for more than one distinct transparent input.
+    AmbiguousSignature,
+    /// Not all transparent inputs that require signing received a valid signature.
+    NotAllInputsSigned,
+    /// One or more provided external signatures were not used to sign any transparent input.
+    UnusedExternalSignature {
+        input_index: usize,
+    },
+    /// An error occurred within the secp256k1 cryptographic library.
+    Secp256k1Error(String), // Stores String to satisfy PartialEq/Eq
```
str4d: After using the correct `secp256k1` API, this error is unused.
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
...
+            Error::SignatureVerificationFailed { input_index } => write!(f, "External signature verification failed for transparent input at index {}", input_index),
+            Error::NoMatchingInputForSignature { input_index } => write!(f, "A provided external signature at index {} did not cryptographically match any available unsigned transparent input", input_index),
+            Error::InputAlreadySigned { input_index } => write!(f, "Transparent input at index {} has already been signed", input_index),
+            Error::AmbiguousSignature => write!(f, "A single provided external signature was cryptographically valid for more than one distinct transparent input"),
+            Error::NotAllInputsSigned => write!(f, "Not all transparent inputs that require signing received a valid signature"),
+            Error::UnusedExternalSignature { input_index } => write!(f, "External signature at index {} was not used to sign any transparent input", input_index),
+            Error::Secp256k1Error(msg) => write!(f, "Secp256k1 cryptographic library error: {}", msg),
```
str4d:
```suggestion
```

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,28 @@ workspace.
...
+  - `Secp256k1Error(String)`
```
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -28,6 +32,41 @@ pub enum Error {
...
+    /// Not all transparent inputs that require signing received a valid signature.
+    NotAllInputsSigned,
+    /// One or more provided external signatures were not used to sign any transparent input.
+    UnusedExternalSignature {
+        input_index: usize,
+    },
+    /// An error occurred within the secp256k1 cryptographic library.
+    Secp256k1Error(String), // Stores String to satisfy PartialEq/Eq
+    /// Generic internal error during transparent builder operations.
+    InternalBuilderError(String),
```
str4d: This is another unused variant:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -36,6 +75,18 @@ impl fmt::Display for Error {
...
+            Error::SignatureVerificationFailed { input_index } => write!(f, "External signature verification failed for transparent input at index {}", input_index),
+            Error::NoMatchingInputForSignature { input_index } => write!(f, "A provided external signature at index {} did not cryptographically match any available unsigned transparent input", input_index),
+            Error::InputAlreadySigned { input_index } => write!(f, "Transparent input at index {} has already been signed", input_index),
+            Error::AmbiguousSignature => write!(f, "A single provided external signature was cryptographically valid for more than one distinct transparent input"),
+            Error::NotAllInputsSigned => write!(f, "Not all transparent inputs that require signing received a valid signature"),
+            Error::UnusedExternalSignature { input_index } => write!(f, "External signature at index {} was not used to sign any transparent input", input_index),
+            Error::Secp256k1Error(msg) => write!(f, "Secp256k1 cryptographic library error: {}", msg),
+            Error::InternalBuilderError(msg) => write!(f, "Internal transparent builder error: {}", msg),
```
str4d:
```suggestion
```

zcash_transparent/CHANGELOG.md
str4d:
```suggestion
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +475,136 @@ impl Bundle<Unauthorized> {
             authorization: Authorized,
         })
     }
+
+    /// Prepares the bundle for staged application of external signatures.
+    #[cfg(feature = "transparent-inputs")]
+    pub fn prepare_transparent_signatures<'a>(
+        self,
+        sighashes: &'a [[u8; 32]],
```
str4d: Bug: this has a silent undocumented requirement that the caller provide the sighashes in the correct order. This is effectively the same as the `signatures` bug in the previous iteration of this API.

Compare this to `Bundle<Unauthorized>::apply_signatures`, which takes a `Fn(SignableInput) -> [u8; 32]` closure that eliminates the index-matching bug, and makes it less likely the caller will provide the wrong sighash (because the caller is required to provide a sighash matching the provided `SignableInput`).

The only justifications I can think of for this API are:
- Maybe there's a desire for the caller to compute and then keep in memory the sighashes. Is that a requirement for your usage, and if so what about your usage requires it? Allocations are clearly not a problem given that `TransparentSignatureContext` contains a bunch of them, so it would be little different to compute a `sighashes: Vec<[u8; 32]>` by calling a `calculate_sighash` closure appropriately.
- Maybe you're trying to ensure you have access to `sighashes` for producing the signatures. But that can be instead achieved with a getter on `TransparentSignatureContext` for the calculated sighashes. (I am somewhat surprised that there isn't already a getter on it for other fields; are you instead doing external matching with the original builder inputs?)

zcash_transparent/CHANGELOG.md
zlyzol: Moved under `Changed`

zcash_transparent/src/builder.rs
zlyzol: Renamed.

zcash_transparent/src/builder.rs
zlyzol: I've refactored this to iterate over the signatures.

zcash_transparent/src/builder.rs
zlyzol: Removed.

zcash_transparent/CHANGELOG.md
zlyzol: This one is still used.

zcash_transparent/src/builder.rs
zlyzol: Removed.

zcash_transparent/src/builder.rs
zlyzol: Removed.

zcash_transparent/CHANGELOG.md
zlyzol: Removed.

zcash_transparent/src/builder.rs
zlyzol: There was no specific requirement on our side for the caller to pre-compute the sighashes. I've refactored `prepare_transparent_signatures` to take a `calculate_sighash` closure instead of a slice, mirroring the pattern in `apply_signatures`.

.github/workflows/zizmor.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb # v6.1.0
+        uses: astral-sh/setup-uv@445689ea25e0de0a23313031f5fe577c74ae45a1 # v6.3.0
```
daira: This is indeed the v6.3.0 tag: https://github.com/astral-sh/setup-uv/releases/tag/v6.3.0

zcash_transparent/src/builder.rs
daira: The reference to `alpha` in the comment from the `orchard` crate is specific to key randomization for RedDSA shielded spend auth signatures. A RedDSA spend auth signature being valid for more than one input would indeed indicate some kind of critical bug (probably but not necessarily in randomness generation), but ECDSA signatures are not randomized in the same way.

It certainly still is the case that you shouldn't expect an ECDSA signature to be valid for more than one input. Either the signature is with the same key and signed input, which indicates a double-spend [*], or an attack on ECDSA such as key substitution, which can't happen accidentally. (Note that RedDSA and EdDSA are not vulnerable to key substitution.) In any case, the reference to `alpha` is still incorrect.

```suggestion
    /// A signature is valid for more than one input. This should not happen: it
    /// indicates either a duplicate signed input and key (and therefore a double-spend),
    /// or an attack on ECDSA such as key substitution.
    DuplicateSignature,
```

[*] It should not be possible to have the same signed input for two inputs that are not double-spends, because the `outpoint` is included in the signed input. For v4 this is specified in [[ZIP 243, Specification](https://zips.z.cash/zip-0243#specification) item 13], and for v5 in [[ZIP 244, S.2g: txin_sig_digest]](https://zips.z.cash/zip-0244#s-2g-txin-sig-digest).

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +451,365 @@ impl Bundle<Unauthorized> {
...
+        let bundle = builder.build().unwrap();
+
+        // Prepare the signing context
+        let secp = Secp256k1::verification_only();
+        let calculate_sighash = |input: SignableInput| [input.index as u8; 32];
+        let sig_context = bundle
+            .prepare_transparent_signatures(calculate_sighash, &secp)
+            .unwrap();
+
+        // Create a signature from adifferent key
```
daira:
```suggestion
        // Create a signature from a different key
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +451,365 @@ impl Bundle<Unauthorized> {
...
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let num_inputs = self.authorization.inputs.len();
+        Ok(TransparentSignatureContext {
+            original_vin_unauthorized: self.vin,
+            original_vout: self.vout,
+            authorization_inputs: self.authorization.inputs,
+            sighashes, // The newly computed and owned Vec of sighashes
```
daira:
```suggestion
            sighashes,
```
Comment is too obvious:
* it was computed just above;
* it's not usually necessary to comment on ownership except in particularly tricky cases, which this is not.

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +451,365 @@ impl Bundle<Unauthorized> {
...
+}
+
+#[cfg(feature = "transparent-inputs")]
+impl<'a> TransparentSignatureContext<'a, secp256k1::VerifyOnly> {
+    /// Appends a new batch of external signatures to the transparent inputs.
+    ///
+    /// This method iterates through the provided signatures, applying each one to the
+    /// single input for which it is valid.
+    pub fn append_external_signatures(
+        self, // Takes ownership of self
```
daira:
```suggestion
        self,
```
This is also obvious by being spelled just `self`. Possibly needs a rustfmt after applying the change.

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +451,365 @@ impl Bundle<Unauthorized> {
...
+                current_self.append_external_signature(sig_idx, sig)
+            })
+    }
+
+    /// Applies a single external signature to the one input for which it is valid.
+    ///
+    /// An error will be returned if the signature is not valid for any available
+    /// unsigned inputs, or if it is valid for more than one input.
+    fn append_external_signature(
+        mut self, // Also takes ownership
```
daira:
```suggestion
        mut self,
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +451,365 @@ impl Bundle<Unauthorized> {
...
+        })
+    }
+}
+
+#[cfg(feature = "transparent-inputs")]
+impl<'a> TransparentSignatureContext<'a, secp256k1::VerifyOnly> {
+    /// Appends a new batch of external signatures to the transparent inputs.
+    ///
+    /// This method iterates through the provided signatures, applying each one to the
+    /// single input for which it is valid.
```
daira, line 501:
```suggestion
    /// single input for which it is valid.
    ///
    /// An error will be returned if any of the signatures are not valid for any
    /// available unsigned inputs, or valid for more than one input.
```

zcash_transparent/src/builder.rs
```diff
@@ -378,4 +451,365 @@ impl Bundle<Unauthorized> {
...
+        };
+
+        let txid = [0u8; 32]; // Dummy txid
+        let outpoint = OutPoint::new(txid, 0);
+
+        (sk, pk, txout, outpoint)
+    }
+
+    #[test]
+    fn happy_path_append_and_finalize_signatures() {
```
daira:
```suggestion
    fn append_and_finalize_signatures() {
```
Style nit: it's sufficient to mark tests about failures by having "fails" in the name; it's not necessary to mark tests about successes, and we don't do so elsewhere AFAIK.

daira (approved):
utACK if comments (including @str4d's) are addressed.

zcash_transparent/CHANGELOG.md
daira: No, @str4d is correct.

daira (approved):
utACK, but the last two commits need another review.

nuttycom (approved):
utACK a9a5d06175781315fc9fd0745b2cebbf0708184f

PR #1787, opened 2025-05-03
Add the ability to obtain the type of a `ZcashAddress`

nuttycom:


components/zcash_address/src/lib.rs
```diff
@@ -315,6 +315,28 @@ impl ZcashAddress {
...
+            AddressKind::Unified(_) => AddressType::Unified,
+            AddressKind::P2pkh(_) => AddressType::P2pkh,
+            AddressKind::P2sh(_) => AddressType::P2sh,
+            AddressKind::Tex(_) => AddressType::Tex,
+        }
+    }
+}
+
+/// The enumeration of address types that are representable as a [`ZcashAddress`].
+pub enum AddressType {
```
str4d, line 333: The rest of the `zcash_address` API is intentionally designed so that a) we can add new address kinds via point releases to get maximum ecosystem parser uptake, and b) downstream code automatically does the correct thing when we do so and treats new address kinds as unsupported rather than unknown.

If we're to have an explicit address type enum in the API, it must be non-exhaustive in order to not immediately break both of these design goals.
```suggestion
///
/// This enum is non-exhaustive because new address types may be added in point releases of
/// the crate. Your code must treat the remainder (`_`) case as a valid but
/// currently-unsupported Zcash address (for example in error messages).
#[non_exhaustive]
pub enum AddressType {
```

str4d (changes requested):
Reviewed

components/zcash_address/src/lib.rs
nuttycom: I disagree with this design goal. If a new address type is added, users of the library can either remain on the older version, or they can upgrade to handle the new address type.

I think that it's fundamentally a mistake for the introduction of a new address type to result in runtime errors for users. The handling of unsupported address types should be explicit; it should not flow through the runtime error pathway.

nuttycom: An alternative would be to make the AddressKind type public. Then, users of the library who want to opt in to explicit handling without relying on the error path can do so; right now there's no way to have the compiler tell you that you need to update to handle a new address type.

nuttycom:
Subsumed by #1790

PR #1791, opened 2025-05-07
zcash_address: Add `Converter` trait and `ZcashAddress::convert_with`

nuttycom:


components/zcash_protocol/src/consensus.rs
```diff
@@ -286,7 +286,7 @@ pub trait Parameters: Clone {
     /// Determines whether the specified network upgrade is active as of the
     /// provided block height on the network to which this Parameters value applies.
     fn is_nu_active(&self, nu: NetworkUpgrade, height: BlockHeight) -> bool {
-        self.activation_height(nu).map_or(false, |h| h <= height)
+        self.activation_height(nu).is_some_and(|h| h <= height)
```
str4d, line 289: Bug: this was added in Rust 1.70, but the MSRV for this particular crate is Rust 1.52 (because it's so simple, we've left it lower in the past in order to enable it to be used in the widest possible set of environments).

I expect your entire "Fix beta clippy complaints" commit needs to be removed. We certainly can't bump MSRV in 0.6.3 or 0.7.1; the earliest would be 0.8 if we want to bump it at all.

str4d (changes requested):
Reviewed b1d04cb02f4313ad75a594b399d217879200cf2c.

str4d (approved):
utACK 456e83e789af14fa006c52b62139444f26d3b705

PR #1800, opened 2025-05-07
Release zcash_keys 0.8.1

nuttycom:


zcash_keys/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+## [0.4.1, 0.5.1, 0.6.1, 0.7.1, 0.8.1] - 2025-05-07
+
+### Added
+- `zcash_keys::Address::to_transparent_address`
```
str4d, line 12: Ditto: 0.8.1 should be in the same group as 0.7.1 without this line.

(Also update the release dates when you make the post-release changelog cleanup.)

str4d (approved):
utACK dcae628639f1cd74a3dd34e2eedf43e290c8dd4d with follow-up

PR #1807, opened 2025-05-15
zcash_keys: Add parsing and serialization for `zcashd`-compatible transparent key encodings.

nuttycom:


components/zcash_protocol/CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ represent the transitive `semver` implications of changes within the enclosing
 workspace.
 
 ## [Unreleased]
+### Added
+  - `{mainnet, testnet, regtest}::B58_SECRET_KEY_PREFIX`
```
str4d, line 14:
```suggestion
- `zcash_protocol::constants::`
  - `{mainnet, testnet, regtest}::B58_SECRET_KEY_PREFIX`
```

components/zcash_protocol/src/constants/mainnet.rs
```diff
@@ -36,14 +36,21 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "zs";
 /// [sproutpaymentaddrencoding]: https://zips.z.cash/protocol/protocol.pdf#sproutpaymentaddrencoding
 pub const B58_SPROUT_ADDRESS_PREFIX: [u8; 2] = [0x16, 0x9a];
 
+/// The prefix for a Base58Check-encoded DER-encoded mainnet [`SecretKey`], as specified via the
+/// bitcoin-derived [`EncodeSecret`] format function.
+///
+/// [`SecretKey`]: https://docs.rs/secp256k1/latest/secp256k1/struct.SecretKey.html
+/// [`EncodeSecret`]: https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/key_io.cpp#L298
+pub const B58_SECRET_KEY_PREFIX: [u8; 1] = [0x80];
```
str4d, line 44: I checked this against https://github.com/zcash/zcash/blob/fe3e645ca9f1de4ff7feaaa1ddb763ae714c93c6/src/chainparams.cpp#L154-L155.

components/zcash_protocol/src/constants/mainnet.rs
```diff
@@ -36,14 +36,21 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "zs";
...
+
 /// The prefix for a Base58Check-encoded mainnet [`PublicKeyHash`].
 ///
-/// [`PublicKeyHash`]: https://docs.rs/zcash_primitives/latest/zcash_primitives/legacy/enum.TransparentAddress.html
+/// [`PublicKeyHash`]: https://docs.rs/zcash_transparent/latest/address/enum.TransparentAddress.html
```
str4d: Incorrect URL modification:
```suggestion
/// [`PublicKeyHash`]: https://docs.rs/zcash_transparent/latest/zcash_transparent/address/enum.TransparentAddress.html
```

components/zcash_protocol/src/constants/mainnet.rs
```diff
@@ -36,14 +36,21 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "zs";
...
 pub const B58_PUBKEY_ADDRESS_PREFIX: [u8; 2] = [0x1c, 0xb8];
 
 /// The prefix for a Base58Check-encoded mainnet [`ScriptHash`].
 ///
-/// [`ScriptHash`]: https://docs.rs/zcash_primitives/latest/zcash_primitives/legacy/enum.TransparentAddress.html
+/// [`ScriptHash`]: https://docs.rs/zcash_transparent/latest/address/enum.TransparentAddress.html
```
str4d: Ditto:
```suggestion
/// [`ScriptHash`]: https://docs.rs/zcash_transparent/latest/zcash_transparent/address/enum.TransparentAddress.html
```

components/zcash_protocol/src/constants/regtest.rs
```diff
@@ -41,6 +41,13 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "zregtestsapling";
 /// [sproutpaymentaddrencoding]: https://zips.z.cash/protocol/protocol.pdf#sproutpaymentaddrencoding
 pub const B58_SPROUT_ADDRESS_PREFIX: [u8; 2] = [0x16, 0xb6];
 
+/// The prefix for a Base58Check-encoded DER-encoded mainnet [`SecretKey`], as specified via the
+/// bitcoin-derived [`EncodeSecret`] format function.
+///
+/// [`SecretKey`]: https://docs.rs/secp256k1/latest/secp256k1/struct.SecretKey.html
+/// [`EncodeSecret`]: https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/key_io.cpp#L298
+pub const B58_SECRET_KEY_PREFIX: [u8; 1] = [0xef];
```
str4d, line 49: I checked this against https://github.com/zcash/zcash/blob/fe3e645ca9f1de4ff7feaaa1ddb763ae714c93c6/src/chainparams.cpp#L752.

components/zcash_protocol/src/constants/testnet.rs
```diff
@@ -36,6 +36,13 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "ztestsapling";
 /// [sproutpaymentaddrencoding]: https://zips.z.cash/protocol/protocol.pdf#sproutpaymentaddrencoding
 pub const B58_SPROUT_ADDRESS_PREFIX: [u8; 2] = [0x16, 0xb6];
 
+/// The prefix for a Base58Check-encoded DER-encoded mainnet [`SecretKey`], as specified via the
+/// bitcoin-derived [`EncodeSecret`] format function.
+///
+/// [`SecretKey`]: https://docs.rs/secp256k1/latest/secp256k1/struct.SecretKey.html
+/// [`EncodeSecret`]: https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/key_io.cpp#L298
+pub const B58_SECRET_KEY_PREFIX: [u8; 1] = [0xef];
```
str4d, line 44: I checked this against https://github.com/zcash/zcash/blob/fe3e645ca9f1de4ff7feaaa1ddb763ae714c93c6/src/chainparams.cpp#L477-L478.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+            Err(_) => {
+                secret_buf.zeroize();
+                Err(())
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
str4d, line 329: `zcashd` has several sets of test vectors that should be ported to here:
- `src/test/data/base58_keys_valid.json` (the ones with `isPrivKey: true` should be valid `Key`s, the rest should be invalid).
- `src/test/data/base58_keys_invalid.json`

We don't need to keep the JSON format; converting to embedded Rust consts like we do for other test vectors is fine.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+        let decoded = SecretVec::new(
+            bs58::decode(encoded.expose_secret())
+                .with_check(None)
+                .into_vec()?,
+        );
+        let prefix = network.b58_secret_key_prefix();
+        let decoded_len = decoded.expose_secret().len();
+        let compressed = decoded_len == (33 + prefix.len());
+        if (decoded_len == 32 + prefix.len()
+            || (compressed && decoded.expose_secret().last() == Some(&1)))
```
str4d: By moving the `compressed` expression outside the conditional for reuse, it is now technically invalid. In the C++ code this condition was a subset of the conditional expression and therefore fine to use as a representative inside the conditional branch (because the state space was sufficiently constrained by the conditional), but here it is outside the branch and would be incorrect if accidentally used elsewhere.
```suggestion
        let compressed = decoded_len == (33 + prefix.len()) && decoded.expose_secret().last() == Some(&1);
        if (decoded_len == 32 + prefix.len() || compressed)
```
(plus `cargo fmt`).

It might also be clearer to define an `uncompressed` variable, to make it clear that `uncompressed != !compressed` in this encoding (that can only be assumed inside the conditional branch).

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+            && decoded.expose_secret()[0..prefix.len()] == prefix
+        {
+            let key_end = decoded_len - if compressed { 1 } else { 0 };
+            let bytes = PrivateKeyBytes::try_from(&decoded.expose_secret()[prefix.len()..key_end])?;
+            Ok(Self {
+                secret: SecretKey::from_bytes(&bytes)?,
+                compressed,
+            })
+        } else {
+            Err(ParseError::LengthInvalid)
```
str4d: This is not the only error that reaches here:
- The compressed encoding might not end in a `1` byte.
- The prefix might be incorrect.

Either rename `LengthInvalid` to a generic `InvalidEncoding` variant, or add error variants that cover the other error cases (one more variant is probably fine).

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    /// Decodes a base58-encoded secret key.
```
str4d:
```suggestion
    /// Encodes a base58-encoded secret key.
```

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// This corresponds to <https://github.com/zcash/zcash/blob/1f1f7a385adc048154e7f25a3a0de76f3658ca09/src/key_io.cpp#L298>
+    pub fn encode_base58<N: NetworkConstants>(&self, network: &N) -> SecretString {
+        let input = SecretVec::new(
+            network
+                .b58_secret_key_prefix()
+                .iter()
+                .chain(self.secret.secret_bytes().iter())
+                .chain(if self.compressed { &[1] } else { &[0] })
```
str4d: Bug: The uncompressed encoding omits the trailing byte.
```suggestion
                .chain(if self.compressed { &[1] } else { &[] })
```
This PR is lacking a round-trip test of the Base58Check encoding, which would have caught this (as would the test vectors if encoding were checked). Please add the tests (and confirm they indeed catch this).

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+
+impl From<bip32::Error> for ParseError {
+    fn from(value: bip32::Error) -> Self {
+        ParseError::Bip32(value)
+    }
+}
+
+/// A secp256k1 secret key, along with a flag indicating whether a compressed encoding should be
+/// used when performing DER serialization.
+pub struct Key {
```
str4d, line 45: Implement `Zeroize` manually for this type calling `self.secret.non_secure_erase()` (for best-effort zeroization), and `#[derive(ZeroizeOnDrop)]`.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+        SecretString::new(
+            bs58::encode(input.expose_secret())
+                .with_check()
+                .into_string(),
+        )
+    }
+
+    /// Returns the wrapped secp256k1 secret key.
+    pub fn secret(&self) -> SecretKey {
+        self.secret
```
str4d: This forces a copy of the key onto the stack which is going to be very hard to clear. Instead return the secret by reference (and then it's up to the caller to decide whether to copy):
```suggestion
    /// Returns the wrapped secp256k1 secret key.
    pub fn secret(&self) -> &SecretKey {
        &self.secret
```
Or alternatively wrap it in `secrecy::Secret` to force the caller to think about it.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    /// Derives the secp256k1 public key corresponding to the secret key.
+    pub fn pubkey(&self) -> PublicKey {
+        let secp = secp256k1::Secp256k1::new();
+        self.pubkey_internal(&secp)
+    }
+
+    /// Derives the secp256k1 public key corresponding to the secret key,
+    /// using the provided secp context.
+    pub fn pubkey_internal<C: Signing>(&self, secp: &Secp256k1<C>) -> PublicKey {
```
str4d: Why is a public API named `*_internal`? Would be better as e.g. `pubkey_with_context` or `pubkey_with`.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    /// Generates the "openssh-inspired" DER encoding of the secret key used by zcashd.
+    pub fn der_encode(&self) -> SecretVec<u8> {
+        let secp = secp256k1::Secp256k1::new();
+        self.der_encode_internal(&secp)
+    }
+
+    /// Generates the "openssh-inspired" DER encoding of the secret key used by zcashd,
+    /// using the provided secp context for pubkey encoding.
+    pub fn der_encode_internal<C: Signing>(&self, secp: &Secp256k1<C>) -> SecretVec<u8> {
```
str4d: Ditto.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+                0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xFF, 0xFF, 0xFC, 0x2F, 0x30,
+                0x06, 0x04, 0x01, 0x00, 0x04, 0x01, 0x07, 0x04, 0x21, 0x02, 0x79, 0xBE, 0x66, 0x7E,
+                0xF9, 0xDC, 0xBB, 0xAC, 0x55, 0xA0, 0x62, 0x95, 0xCE, 0x87, 0x0B, 0x07, 0x02, 0x9B,
+                0xFC, 0xDB, 0x2D, 0xCE, 0x28, 0xD9, 0x59, 0xF2, 0x81, 0x5B, 0x16, 0xF8, 0x17, 0x98,
+                0x02, 0x21, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF, 0x48, 0xA0, 0x3B, 0xBF,
+                0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x41, 0x02, 0x01, 0x01, 0xA1, 0x24, 0x03, 0x22,
+                0x00,
+            ];
```
str4d, line 158: I did not check the contents of this array.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+                0x06, 0x04, 0x01, 0x00, 0x04, 0x01, 0x07, 0x04, 0x41, 0x04, 0x79, 0xBE, 0x66, 0x7E,
+                0xF9, 0xDC, 0xBB, 0xAC, 0x55, 0xA0, 0x62, 0x95, 0xCE, 0x87, 0x0B, 0x07, 0x02, 0x9B,
+                0xFC, 0xDB, 0x2D, 0xCE, 0x28, 0xD9, 0x59, 0xF2, 0x81, 0x5B, 0x16, 0xF8, 0x17, 0x98,
+                0x48, 0x3A, 0xDA, 0x77, 0x26, 0xA3, 0xC4, 0x65, 0x5D, 0xA4, 0xFB, 0xFC, 0x0E, 0x11,
+                0x08, 0xA8, 0xFD, 0x17, 0xB4, 0x48, 0xA6, 0x85, 0x54, 0x19, 0x9C, 0x47, 0xD0, 0x8F,
+                0xFB, 0x10, 0xD4, 0xB8, 0x02, 0x21, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+                0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF,
+                0x48, 0xA0, 0x3B, 0xBF, 0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x41, 0x02, 0x01, 0x01,
+                0xA1, 0x44, 0x03, 0x42, 0x00,
+            ];
```
str4d, line 193: I did not check the contents of this array.

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        //    /* sequence header */
+        //    if (end - seckey < 1 || *seckey != 0x30u) {
+        //        return 0;
+        //    }
+        //    seckey++;
+        if seckey.is_empty() || seckey[0] != 0x30 {
+            return Err(());
+        }
+        let seckey = &seckey[1..];
```
str4d: Non-blocking: Alternatives for this (and similarly below):
```suggestion
        if seckey.first() != Some(&0x30) {
            return Err(());
        }
        let seckey = &seckey[1..];
```
```suggestion
        let seckey = match seckey.split_first() {
            Some((&0x30, rest)) => Ok(rest),
            _ => Err(()),
        }?;
```

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+        if seckey.len() < len {
+            return Err(());
+        }
+
+        //    /* sequence element 0: version number (=1) */
+        //    if (end - seckey < 3 || seckey[0] != 0x02u || seckey[1] != 0x01u || seckey[2] != 0x01u) {
+        //        return 0;
+        //    }
+        //    seckey += 3;
+        if seckey.len() < 3 || seckey[0] != 0x02 || seckey[1] != 0x01 || seckey[2] != 0x01 {
```
str4d: This is much clearer while still being close to the C++ code:
```suggestion
        if seckey.len() < 3 || &seckey[..3] != &[0x02, 0x01, 0x01] {
```

Non-blocking: This could use `slice::split_at` (similar to the earlier alternative).

zcash_keys/src/keys/transparent.rs
```diff
@@ -0,0 +1,340 @@
...
+        let seckey = &seckey[2..];
+        if oslen > 32 || seckey.len() < oslen {
+            return Err(());
+        }
+        let mut secret_buf = [0u8; 32];
+        secret_buf[(32 - oslen)..].copy_from_slice(&seckey[..oslen]);
+        match SecretKey::from_bytes(&secret_buf) {
+            Ok(secret) => Ok(Self { secret, compressed }),
+            Err(_) => {
+                secret_buf.zeroize();
```
str4d: We want to zeroize this unconditionally. In the C++ code `out32` was how the secret was returned in the success case; here `secret_buf` is internal.

str4d (changes requested):
Reviewed e8807e5d2259cfced1dd96c60ca7927155d5065f.

components/zcash_protocol/src/constants/regtest.rs
```diff
@@ -41,6 +41,13 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "zregtestsapling";
 /// [sproutpaymentaddrencoding]: https://zips.z.cash/protocol/protocol.pdf#sproutpaymentaddrencoding
 pub const B58_SPROUT_ADDRESS_PREFIX: [u8; 2] = [0x16, 0xb6];
 
+/// The prefix for a Base58Check-encoded DER-encoded mainnet [`SecretKey`], as specified via the
```
daira:
```suggestion
/// The prefix for a Base58Check-encoded DER-encoded regtest [`SecretKey`], as specified via the
```

components/zcash_protocol/src/constants/testnet.rs
```diff
@@ -36,6 +36,13 @@ pub const HRP_SAPLING_PAYMENT_ADDRESS: &str = "ztestsapling";
 /// [sproutpaymentaddrencoding]: https://zips.z.cash/protocol/protocol.pdf#sproutpaymentaddrencoding
 pub const B58_SPROUT_ADDRESS_PREFIX: [u8; 2] = [0x16, 0xb6];
 
+/// The prefix for a Base58Check-encoded DER-encoded mainnet [`SecretKey`], as specified via the
```
daira:
```suggestion
/// The prefix for a Base58Check-encoded DER-encoded testnet [`SecretKey`], as specified via the
```

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/f3996c83e591467526dc94e6e2cfed795ccb3d31..9bfb2538f547be8865b15a13da0115b4a9278cd2) to fix rustfmt error.

zcash_keys/src/keys/transparent.rs
nuttycom: Fixed in 0600a4d5d7226550b01ef150f032f2725564eb75

zcash_keys/src/keys/transparent/test_vectors.rs
```diff
@@ -0,0 +1,419 @@
...
+    },
+    TestVector {
+        base58_encoding: "t3TCuHUxH3LGnsFYTUbSwHrRXxTaEA8hAaf",
+        raw_bytes_hex: "a9145ece0cadddc415b1980f001785947120acdb36fc87",
+        kind: VectorKind::Pubkey,
+        network: NetworkType::Main,
+    },
+];
+
+// From
```
str4d, line 367: Unfinished sentence.

str4d (approved):
utACK 3c8a611c441d5622f83ca46c1f01fe6b9e9142a9

PR #1809, opened 2025-05-17
zcash_history: allow to return `V::NodeData` from `Entry`

Metalcape:
As discussed on discord, I am working on implementing flyclient prover functionalities into zebrad. Because zebrad's implementation of history trees uses `Entry` as the type for nodes, it would be useful to have access to its inner `NodeData`, to avoid having to operate on the serialized bytes directly. This is needed especially when reading nodes from the database and sending them to clients.

### DIFF
--- a/zcash_history/CHANGELOG.md
+++ b/zcash_history/CHANGELOG.md
@@ -9,6 +9,9 @@ and this library adheres to Rust's notion of
 ### Changed
 - MSRV is now 1.81.0.
 
+### Added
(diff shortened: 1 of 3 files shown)


str4d (changes requested):
Reviewed 6f509946eb59468c21d17a399359ba9192758536.

str4d (approved):
utACK 8d732c3a347b458365d42fc247f46d7a87c90d8a

PR #1811, opened 2025-05-22
zcash_client_backend: Expose HTTP GET capability in public API

str4d:


zcash_client_backend/src/tor/http.rs
```diff
@@ -57,15 +58,47 @@ impl Client {
...
+            retry_limit,
+        )
+        .await
+    }
+
+    /// Makes an HTTP request over Tor.
+    ///
+    /// On error, retries will be attempted as follows:
+    /// - A successful request that resulted in a client error (HTTP 400-499) will cause a
+    ///   retry with an isolated client.
```
nuttycom: We should probably choose a subset of this range that will result in retries; for example, 403 and 404 almost certainly should not be retried.

nuttycom (approved):
utACK 5d68044df51cfcab43cbcf740188352ed50be8af

zcash_client_backend/src/tor/http.rs
str4d: IIRC 403 is the most common error code I see when a Tor exit node is blocked, and is precisely when I want this to trigger.

nuttycom: Okay; maybe 404 is worth respecting though?

str4d: I think if we're going to filter it to less than "all 4XX status codes", then we probably should make the set of status codes configurable by the user, which makes the API more complicated. The current behaviour was fine when we just exposed the limited "fetch exchange rate" API, as we knew the kinds of status codes to expect.

I'll try and come up with an alternative API that isn't horrible.

nuttycom (approved):
utACK 9629192e4c45b38345cc38d8f73d84cc8853c21c

PR #1813, opened 2025-05-30
zcash_client_sqlite: Do not store a transaction record if there is no wallet involvement.

nuttycom:
Prior to this change, `decrypt_and_store_transaction` would erroneously create a record in the `transactions` table, even if the transaction didn't involve the wallet at all.

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -1044,15 +1044,7 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
     params: &P,
     output: &WalletTransparentOutput,
 ) -> Result<(AccountRef, KeyScope, UtxoId), SqliteClientError> {
-    put_transparent_output(
-        conn,
-        params,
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1372,9 +1363,9 @@ pub(crate) fn put_transparent_output<P: consensus::Parameters>(
              mined_height = :mined_height
          RETURNING id_tx",
         named_params![
-           ":txid": &outpoint.hash().to_vec(),
+           ":txid": &output.outpoint().hash().to_vec(),
            ":block": block,
-           ":mined_height": output_height
+           ":mined_height": output.mined_height().map(u32::from)
```
str4d, line 1368: The existing code was correct:
```suggestion
           ":mined_height": output_height
```

str4d (approved):
utACK 174a0dfdb2f329d7d7805cdc3f9ae92cac37e717

str4d (approved):
Re-utACK

PR #1815, opened 2025-05-31
zcash_client_backend: Don't convert HTTP status errors into Rust errors

str4d:
In some contexts it is important to get both the HTTP status code and the headers/body for non-success response codes. We didn't need this for the exchange rate fetcher, and so we were dropping the headers and body on the floor. That was fine while the HTTP GET etc. APIs were internal, but now that we expose them we need to enable the caller to decide what to do.


nuttycom (approved):
utACK 8769467dc192a284bd68576978a51d29922f48f1

PR #1821, opened 2025-06-12
zcash_client_sqlite: Do not queue retrieval for coinbase inputs.

nuttycom:


components/zcash_protocol/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+  - `zcash_protocol::txid::TxId::is_coinbase`
```
str4d: Nit:
```suggestion
- `zcash_protocol::txid::TxId::is_null`
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3642,7 +3642,10 @@ pub(crate) fn queue_transparent_input_retrieval<AccountId>(
         // queue the transparent inputs for enhancement
         queue_tx_retrieval(
             conn,
-            b.vin.iter().map(|txin| *txin.prevout.txid()),
+            b.vin
+                .iter()
+                .map(|txin| *txin.prevout.txid())
+                .filter(|txid| !txid.is_coinbase()),
```
str4d: Document this as a fix to `WalletDb::transaction_data_requests` ("no longer generates requests for the null (all-zeroes) transaction ID").

components/zcash_protocol/src/txid.rs
```diff
@@ -48,18 +48,27 @@ impl From<TxId> for [u8; 32] {
...
 
+    /// Writes the 32-byte payload directly to the provided writer.
     pub fn write<W: Write>(&self, mut writer: W) -> io::Result<()> {
         writer.write_all(&self.0)?;
         Ok(())
     }
+
+    /// Returns true when the txid consists of all zeros; this only occurs for coinbase
+    /// transactions.
+    pub fn is_coinbase(&self) -> bool {
```
str4d: This alone is not a coinbase signifier, it is a "null" signifier (as in, it corresponds to `uint256::IsNull` in `zcashd`):
```suggestion
    pub fn is_null(&self) -> bool {
```

str4d (changes requested):
Reviewed b54bd02908cac39d653c0a3e24e09cac57310a5c

components/zcash_protocol/src/txid.rs
```diff
@@ -48,18 +48,27 @@ impl From<TxId> for [u8; 32] {
...
 
+    /// Writes the 32-byte payload directly to the provided writer.
     pub fn write<W: Write>(&self, mut writer: W) -> io::Result<()> {
         writer.write_all(&self.0)?;
         Ok(())
     }
+
+    /// Returns true when the txid consists of all zeros; this only occurs for coinbase
+    /// transactions.
+    pub fn is_null(&self) -> bool {
```
str4d, line 71: Non-blocking: it would be a good idea to use this in `zcash_transparent` (though in a separate PR as it would be a no-op that doesn't need a patch release):
https://github.com/zcash/librustzcash/blob/1f61df1627e4a125fee951da4086e94975ec9621/zcash_transparent/src/bundle.rs#L171-L177

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3642,7 +3642,10 @@ pub(crate) fn queue_transparent_input_retrieval<AccountId>(
         // queue the transparent inputs for enhancement
         queue_tx_retrieval(
             conn,
-            b.vin.iter().map(|txin| *txin.prevout.txid()),
+            b.vin
+                .iter()
+                .map(|txin| *txin.prevout.txid())
+                .filter(|txid| !txid.is_null()),
```
str4d, line 3648: An alternate fix here is to make the queueing conditional on `!b.is_coinbase()` (because AFAIR consensus rules enforce that only coinbase txs can have any inputs with null prevout txids), but the more general approach here is fine (because we inherently in Zcash make the assumption in full nodes that a transaction will never have a null txid).

str4d (approved):
utACK 46b35504a635eefe4568537750590ea82c13ac2a

zcash_client_sqlite/src/wallet.rs
nuttycom: Took this approach in #1822

PR #1822, opened 2025-06-12
Skip coinbase input retrieval

nuttycom:


nuttycom:
The clippy failures here require an API-breaking change to fix (and that fix is already in `main`)

str4d (approved):
utACK 61fda669cf6ceb813e1ddc68dd76f205aaf0bb9d

PR #1823, opened 2025-06-12
Add `zcash_protocol::txid::TxId::is_null`

nuttycom:


zcash_transparent/src/bundle.rs
```diff
@@ -173,7 +173,7 @@ impl OutPoint {
     fn is_null(&self) -> bool {
         // From `BaseOutPoint::IsNull()` in zcashd:
         //   return (hash.IsNull() && n == (uint32_t) -1);
-        self.hash.as_ref() == &[0; 32] && self.n == u32::MAX
+        self.hash.is_null() && self.n == u32::MAX
```
str4d, line 176: `zcash_transparent` needs to depend on at least `zcash_protocol 0.5.3`, but the workspace that it gets its dependency from allows `zcash_protocol 0.5.1`. Technically we don't need to bump the version for any crate other than `zcash_transparent`, so maybe we just temporarily alter its `Cargo.toml` directly instead of the workspace one (copying in the version and path dep and then bumping the version)? We'd presumably remember to put it back to a workspace dep when we release `zcash_protocol 0.6` as the path dependency in `zcash_transparent` would break.

str4d (approved):
utACK c2590a47569c6d303fc8c4b12e24bdb0e6268bb3 with a question about how we should handle intra-workspace inter-crate minimum patch dependencies.

zcash_transparent/src/bundle.rs
nuttycom: There's not really a downside to bumping the workspace to 0.5.3, is there?

str4d: Probably not, it's technically forcing an update for other people but only if/when we cut other releases from the same branch state, which we probably won't end up doing often now that we are trying to use crate-specifoc patch branches.

str4d (approved):
utACK dcc1424482038309482f2e85cf8c21f276370cea

Issue #1805, opened 2025-05-13
zcash_client_backend: Add a way to notify wallet that a `TransactionDataRequest` has been served


str4d:
We currently have `WalletWrite::set_transaction_status` for two of the three TDRs, which implicitly marks the request as served. However, for `TransactionDataRequest::TransactionsInvolvingAddress` it is valid for the request to be served by making no changes, and we have no way to tell the wallet we did so. This leads to the request being repeated every call, causing the wallet to hot-loop.

We need something like `set_request_processed(TransactionDataRequest, block hash/height)`, where the caller can notify the wallet that it served the request relative to the specified chain tip state. Then the wallet can internally determine whether the chain tip state has moved relative to when it made the request, and update its queued requests appropriately. Similarly, when `update_chain_tip` is called this should also trigger request queue re-evaluations.

PR #1824, opened 2025-06-12
zcash_client_backend: Add `WalletWrite::notify_address_checked`.

nuttycom:
Fixes zcash/librustzcash#1805


zcash_client_backend/src/data_api.rs
```diff
@@ -2644,6 +2723,15 @@ pub trait WalletWrite: WalletRead {
...
         _status: TransactionStatus,
     ) -> Result<(), Self::Error>;
+
+    /// Notifies the wallet backend that the given query for transactions involving a particular
+    /// address has completed evaluation.
+    #[cfg(feature = "transparent-inputs")]
+    fn notify_address_checked(
+        &mut self,
+        request: TransactionsInvolvingAddress,
+        as_of_height: BlockHeight,
```
str4d, line 2719: Passing just a height here implies that the request was evaluated relative to a main chain state, but nothing in the documentation of this method constrains _which_ main chain state. Unlike the scanning APIs (where we have block hash information and can therefore do continuity checks to detect inconsistencies), here we do not have enough information to tell whether this height is related in any way to the wallet's current chain state view. But the method documentation does not give me enough information to know whether or not this potential state inconsistency matters.

Does `as_of_height` need to be a height that the backend knows about? i.e. does the caller need to externally synchronize to ensure that `as_of_height` is either the same block as, or a parent of, whatever was passed to the wallet in its most recent `update_chain_tip` call (thus making the height on its own consistent with the internal wallet state)? If so, document this implementor requirement on the method, and check it in the `zcash_client_sqlite` implementation. If not, document this on the method.

Alternatively, pass in `as_of_chain_tip` as a `(BlockHash, BlockHeight)` tuple (or maybe a struct representing a chain tip), so that the backend wallet has enough information to detect (either in this method call, or later) any state inconsistency (and e.g. re-queue the address check query immediately).

zcash_client_backend/src/data_api.rs
```diff
@@ -828,35 +887,55 @@ pub enum TransactionDataRequest {
     /// request by detecting transactions involving the specified address within the provided block
     /// range; if using `lightwalletd` for access to chain data, this may be performed using the
     /// [`GetTaddressTxids`] RPC method. It should then call [`wallet::decrypt_and_store_transaction`]
-    /// for each transaction so detected.
+    /// for each transaction so detected. If no transactions are detected within the given range,
+    /// the caller should instead invode `WalletWrite::set_request_processed
```
str4d: Incomplete docs, referring to presumably an earlier intermediate state of this PR.

zcash_client_backend/src/data_api.rs
```diff
@@ -788,6 +788,65 @@ pub enum OutputStatusFilter {
...
+    /// correlations between addresses to untrusted parties; for example, a wallet application
+    /// that uses a private, trusted-for-privacy supplier of chain data can safely ignore this
+    /// field.
+    request_at: Option<SystemTime>,
+    /// The caller should respond to this request only with transactions that conform to the
+    /// specified transaction status filter.
+    tx_status_filter: TransactionStatusFilter,
+    /// The caller should respond to this request only with transactions containing outputs
+    /// that conform to the specified output status filter.
+    output_status_filter: OutputStatusFilter,
```
str4d: All of this documentation needs to be moved onto the getters for these fields. The prior enum variant fields were user-visible; here only the getters are.

Alternatively, consider using the `getset` crate to implement the getters (which would then use the documentation defined on the fields).

zcash_client_backend/src/data_api.rs
```diff
@@ -788,6 +788,65 @@ pub enum OutputStatusFilter {
     All,
 }
 
+/// Payload data for [`TransactionDataRequest::TransactionsInvolvingAddress`].
+///
+/// Values of this type are not constructed directly, but are instead constructed using
+/// [`TransactionDataRequest::transactions_involving_address`].
+#[cfg(feature = "transparent-inputs")]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TransactionsInvolvingAddress {
```
str4d, line 798: Document this addition in the changelog. (The changelog already references this type in the changes, but the fact that this is a newly-added type is not mentioned.)

zcash_client_backend/src/data_api.rs
```diff
@@ -828,35 +887,55 @@ pub enum TransactionDataRequest {
...
+    ///   addresses to untrusted parties; for example, a wallet application that uses a private,
+    ///   trusted-for-privacy supplier of chain data can safely ignore this field.
+    /// - `tx_status_filter`: The caller should respond to this request only with transactions that
+    ///   conform to the specified transaction status filter.
+    /// - `output_status_filter: The caller should respond to this request only with transactions
+    ///   containing outputs that conform to the specified output status filter.
+    ///
+    /// See [`TransactionDataRequest::TransactionsInvolvingAddress`] for more information.
+    #[cfg(feature = "transparent-inputs")]
+    pub fn transactions_involving_address(
```
str4d, line 903: Document this addition in the changelog.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1038,6 +1039,43 @@ pub(crate) fn mark_transparent_utxo_spent(
     Ok(affected_rows > 0)
 }
 
+/// Sets the max observed unspent height for all unspent transparent outputs received at the given
+/// address to at least the given height (calling this method will not cause the max observed
+/// unspent height to decrease).
```
str4d, line 1043: Re: my API question, it's a valid state transition for a chain height to go backwards in a reorg. Presumably when a truncation is performed on the zcs state, it lowers the `max_observed_unspent_height`s? Which would mean that if we don't change the API to make reorgs detectable here, we won't re-check the address until the reorg is detected elsewhere, which could lead to weird detection pauses. But maybe that's fine if it's documented given that `TransactionsInvolvingAddress` is currently only used for ephemeral addresses? It may be a problem if that changes in future and we forget to make changes here.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1038,6 +1039,43 @@ pub(crate) fn mark_transparent_utxo_spent(
...
+) -> Result<(), SqliteClientError> {
+    let addr_str = address.encode(params);
+    debug!(
+        "Setting max_observed_unspent_height to {} for address {}",
+        checked_at, addr_str
+    );
+
+    let mut stmt_update_observed_unspent = conn.prepare(
+        "UPDATE transparent_received_outputs AS tro
+         SET max_observed_unspent_height = CASE
```
str4d, line 1061: The current documentation for this column is:
```rust
/// - `max_observed_unspent_height`: The maximum block height at which this TXO was either
///   observed to be a member of the UTXO set at the start of the block, or observed
///   to be an output of a transaction mined in the block. This is intended to be used to
///   determine when the TXO is no longer a part of the UTXO set, in the case that the
///   transaction that spends it is not detected by the wallet.
```
I think that means we need to be setting it to `:checked_at + 1`, since we've checked it as of the end of `checked_at`? Or it means that the documentation for the column needs updating if this is a separate third way that a TXO is observed.

Re: my API question, the implication here is that this height should be connected to the wallet's view of the chain tip, but not necessarily in the chain state that has the wallet's chain tip? That is, I think it's fine for this to be a height that id after the wallet's view of the chain tip as long as `checked_at` is some direct child of the wallet's view of the chain tip? Although then if there's a reorg after this call in the chain region between the wallet's chain tip and `checked_at`, we would roll forward the height at which we select UTXOs for spending without re-evaluating `transparent_received_outputs`, and thus might not detect a spend in the re-orged chain and continue treating this output as spendable.

Do we have any tests for this kind of situation?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1126,34 +1164,49 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
         super::chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)?;
 
+    debug!(
+        "Generating transaction data requests as of chain tip height {}",
+        chain_tip_height
+    );
+
     // We cannot construct address-based transaction data requests for the case where we cannot
     // determine the height at which to begin, so we require that either the target height or mined
     // height be set.
```
str4d: This documentation needs updating.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1126,34 +1164,49 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
     let mut spend_requests_stmt = conn.prepare_cached(
         "SELECT
             ssq.address,
-            IFNULL(t.target_height, t.mined_height)
+            COALESCE(tro.max_observed_unspent_height, t.mined_height, t.target_height)
```
str4d: Why do we now prefer `t.mined_height` over `t.target_height`? Is this fixing a bug, or is this a semantic change? (The documentation above once updated might help explain.)

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1126,34 +1164,49 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
-         WHERE t.target_height IS NOT NULL
-         OR t.mined_height IS NOT NULL",
+         JOIN transparent_received_outputs tro ON tro.transaction_id = t.id_tx
+         LEFT OUTER JOIN transparent_received_output_spends tros
+            ON tros.transparent_received_output_id = tro.id
+         WHERE tros.transaction_id IS NULL
+         AND (t.target_height IS NOT NULL OR t.mined_height IS NOT NULL)
+         AND (
+             tro.max_observed_unspent_height IS NULL
+             OR tro.max_observed_unspent_height < :chain_tip_height
```
str4d, line 1195: I think this needs to be `<=` given the (current) definition of `tro.max_observed_unspent_height` as being unspent at the start of the specified block (corresponding to setting it to `checked_at + 1` in my other comment)? Otherwise we don't request a TDR unless we're at least 2 blocks outdated.

str4d (changes requested):
Reviewed 2c19c1521c1de78e6ae0dba541facb23cbc6c782.

zcash_client_backend/src/data_api.rs
nuttycom: The reason that `as_of_height` (really `as_of_chain_tip`, as you note) is sufficient is that wallet truncation also resets the `max_observed_unspent_height` field to `null` for all block heights that are invalidated in truncation. So the name change is useful; I don't think that a block hash parameter is necessary here.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: `max_observed_unspent_height` is already set correctly via truncation.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: I think that the safest thing here is to never set this height to a value that is greater than the wallet's view of the chain tip. I'll make that change.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: ~It now occurs to me that there was already a subtle bug here and it's unrelated to this change; the existing code incorrectly made the assumption that if a transaction was mined in a block, that its outputs were unspent as of the end of that block. However, that's not the case for transparent outputs, which can be spent with zero confirmations.~

nuttycom: I believe the correct behavior here is:

* Change the semantics of  `max_observed_unspent_height` (which is not read prior to this PR) to have the meaning "the TXO is known to be unspent as of the *end* of the given block"

and then use:

```
COALESCE(tro.max_observed_unspent_height + 1, t.mined_height, t.target_height)
```
The reason to prefer to start scanning for spends at the mined height rather than the target height is obvious; if we know that the outputs have been mined, they can't be spent prior to that height. Maybe we should ignore target height entirely? There's no reason to check for spentness until we know the output to have been mined.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/ba4605a75a5f391bccf91984cb08390ff2bd92ca..caae2b9cc8d68498d862d2cff2e6e150902089fa) to fix feature-flagging problems.

zcash_client_backend/Cargo.toml
```diff
@@ -134,6 +134,11 @@ webpki-roots = { workspace = true, optional = true }
 # Workaround for https://anonticket.torproject.org/user/projects/arti/issues/pending/4028/
 time-core.workspace = true
 
+# Dependencies used internally:
+# (Breaking upgrades to these are usually backwards-compatible, but check MSRVs.)
```
str4d, line 138: Nit: we already had this boilerplate further up (if the dependency were inserted in the category-alphabetic spot it would be right underneath the existing one).

zcash_client_backend/src/data_api.rs
```diff
@@ -828,35 +867,56 @@ pub enum TransactionDataRequest {
     /// request by detecting transactions involving the specified address within the provided block
     /// range; if using `lightwalletd` for access to chain data, this may be performed using the
     /// [`GetTaddressTxids`] RPC method. It should then call [`wallet::decrypt_and_store_transaction`]
-    /// for each transaction so detected.
+    /// for each transaction so detected. If no transactions are detected within the given range,
+    /// the caller should instead invoke [`WalletWrite::notify_address_checked`] with
+    /// `block_end_height` as the `as_of_height` argument.
```
str4d: Off-by-one: given that `as_of_height` will be the height of a real block (as documented on `notify_address_checked`), but `block_end_height` is documented as end-exclusive (and thus may be a block that does not yet exist), this should be:
```suggestion
    /// `block_end_height - 1` as the `as_of_height` argument.
```

(And after reviewing further down in the PR, I see that this is enforced by `notify_address_checked` so it is important that this documentation be correct.)

zcash_client_sqlite/src/error.rs
```diff
@@ -222,6 +233,10 @@ impl fmt::Display for SqliteClientError {
             #[cfg(feature = "transparent-inputs")]
             SqliteClientError::Scheduling(err) => {
                 write!(f, "The wallet was unable to schedule an event: {err}")
+            },
+            #[cfg(feature = "transparent-inputs")]
+            SqliteClientError::NotificationMismatch { expected, actual } => {
+                write!(f, "The client performed an address check over a block range that did not match the requested range; expected height: {expected}, actual height: {actual}")
```
str4d: The error will be more helpful to the user if it directs them to which parameter was set incorrectly (which also helps avoid ambiguity about whether these heights are real or end-exclusive):
```suggestion
                write!(f, "The client performed an address check over a block range that did not match the requested range; expected as_of_height: {expected}, actual as_of_height: {actual}")
```

str4d (approved):
utACK caae2b9cc8d68498d862d2cff2e6e150902089fa with documentation bugs.

PR #1830, opened 2025-06-17
zcash_client_sqlite: Use `table_constants` in more places & remove `per_protocol_names`.

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -115,7 +116,7 @@ use zcash_protocol::{
     value::{ZatBalance, Zatoshis},
     PoolType, ShieldedProtocol, TxId,
 };
-use zip32::{fingerprint::SeedFingerprint, DiversifierIndex};
+use zip32::{self, fingerprint::SeedFingerprint, DiversifierIndex};
```
str4d: Don't know why this was done. Remove this change from the PR.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -73,6 +73,7 @@ use std::{
     time::SystemTime,
 };
 
+use common::per_protocol_names;
```
str4d: Move this down to the `self` / `crate` block of imports, and change it to `use self::common::per_protocol_names`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2018,11 +2021,11 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
     #[cfg(feature = "orchard")]
     {
         let orchard_trace = tracing::info_span!("orchard_balances").entered();
-        count_notes(
+        with_pool_balances(
             tx,
             summary_height,
             &mut account_balances,
-            ORCHARD_TABLES_PREFIX,
```
str4d, line 2025: Is the import of this still necessary?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2036,11 +2039,11 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
     }
 
     let sapling_trace = tracing::info_span!("sapling_balances").entered();
-    count_notes(
+    with_pool_balances(
         tx,
         summary_height,
         &mut account_balances,
-        SAPLING_TABLES_PREFIX,
```
str4d, line 2043: Is the import of this still necessary?

zcash_client_sqlite/src/wallet.rs
nuttycom: It's an artifact of just how old these changes are (the branch is from 2022)

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -21,19 +21,47 @@ use crate::{
...
+#[cfg(feature = "orchard")]
+const ORCHARD_TABLE_CONSTANTS: TableConstants = TableConstants {
+    table_prefix: ORCHARD_TABLES_PREFIX,
+    output_index_col: "action_index",
+    output_count_col: "orchard_action_count",
+    note_reconstruction_cols: "rho, rseed",
+    shard_height: ORCHARD_SHARD_HEIGHT,
+};
+
+#[allow(dead_code)]
```
str4d, line 51: AFAICT `ErrUnsupportedPool` is not unused. Why is the annotation necessary?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2120,40 +2114,34 @@ pub(crate) fn get_received_memo(
...
                 ":txid": note_id.txid().as_ref(),
                 ":output_index": note_id.output_index()
             ],
-            |row| row.get(0),
+            |row| row.get::<_, Option<Vec<u8>>>(0),
         )
-        .optional()
-    };
+        .optional()?
+        .flatten();
```
str4d, line 2138: This looks like an undocumented bugfix: it converts an error if the received note's tx is unenhanced (because `null` can't be parsed into `Vec<u8>`) into `Ok(None)`. Update the method documentation to clarify this, and add a "Fixed" entry to the `zcash_client_sqlite` changelog (because this alters the implementation to match the documented behaviour of `WalletRead::get_memo`).

str4d (changes requested):
Reviewed e9042f92aa0407211421cd1db13f3549aa50ac60.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: The method it defines is unused when the `transparent-inputs` feature is not enabled. Rather than feature-flag the method of the trait, I thought it was better to just allow it to go unused.

zcash_client_sqlite/src/wallet.rs
nuttycom: It is not; the flatten was previously repeated on lines 2141 and 2143; this just factors it out.

str4d: Ah okay, so previously the `memo_bytes: Option<Vec<_>>` combined with the `.flatten()` was enough for Rust to figure out that the return type of `fetch_memo` was `Result<Option<Option<Vec<_>>>, _>`. I've confirmed this in current `main`.

daira (commented):
Post-hoc ACK

PR #1831, opened 2025-06-17
Fix beta clippy complaints.

nuttycom:

PR #1832, opened 2025-06-18
Interim merge of the long-lived `zallet_dev` branch to `main` 

nuttycom:


str4d (approved):
utACK 28ae6243b986fe093978e5d6bdfdc10b9f123e0d. I didn't closely review the second merge, but I did confirm that it only included https://github.com/zcash/librustzcash/pull/1831 and that the first merge had the expected diff.

PR #1841, opened 2025-06-23
zcash_client_sqlite: Ensure unsatisfiable tx status requests get deleted.

nuttycom:
`TransactionDataRequest::GetStatus` requests for txids that do not correspond to unexpired transactions in the transactions table will now be deleted from the status check queue when `set_transaction_status` is called with a status of either `TxidNotRecognized` or `NotInMainChain`.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -107,7 +107,7 @@ use zcash_keys::{
 use zcash_primitives::{
     block::BlockHash,
     merkle_tree::read_commitment_tree,
-    transaction::{Transaction, TransactionData},
+    transaction::{builder::DEFAULT_TX_EXPIRY_DELTA, Transaction, TransactionData},
 };
 use zcash_protocol::{
     consensus::{self, BlockHeight, BranchId, NetworkUpgrade, Parameters},
@@ -123,7 +123,7 @@ use crate::{
     util::Clock,
     wallet::commitment_tree::{get_max_checkpointed_height, SqliteShardStore},
     AccountRef, AccountUuid, AddressRef, SqlTransaction, TransferType, TxRef,
-    WalletCommitmentTrees, WalletDb, PRUNING_DEPTH, SAPLING_TABLES_PREFIX, VERIFY_LOOKAHEAD,
+    WalletCommitmentTrees, WalletDb, PRUNING_DEPTH, SAPLING_TABLES_PREFIX,
 };
 
 #[cfg(feature = "transparent-inputs")]
@@ -2758,16 +2758,16 @@ pub(crate) fn set_transaction_status(
     // is whether that transaction ends up being mined or expires.
     match status {
         TransactionStatus::TxidNotRecognized | TransactionStatus::NotInMainChain => {
-            // If the transaction is now expired, remove it from the retrieval queue.
+            // Remove the txid from the retrieval queue unless an unexpired transaction having that
+            // txid exists in the transactions table.
             if let Some(chain_tip) = chain_tip_height(conn)? {
                 conn.execute(
-                    "DELETE FROM tx_retrieval_queue WHERE txid IN (
-                        SELECT txid FROM transactions
-                        WHERE txid = :txid AND expiry_height < :chain_tip_minus_lookahead
-                    )",
+                    "DELETE FROM tx_retrieval_queue
+                     WHERE txid = :txid
+                     AND request_expiry <= :chain_tip",
                     named_params![
                         ":txid": txid.as_ref(),
-                        ":chain_tip_minus_lookahead": u32::from(chain_tip).saturating_sub(VERIFY_LOOKAHEAD)
+                        ":chain_tip": u32::from(chain_tip)
                     ],
                 )?;
             }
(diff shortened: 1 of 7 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -2761,13 +2761,19 @@ pub(crate) fn set_transaction_status(
...
-                    )",
+                    "DELETE FROM tx_retrieval_queue q
+                     LEFT OUTER JOIN transactions t
+                        ON t.txid = q.txid
+                        AND (
+                          t.expiry_height IS NULL
+                          OR t.expiry_height >= :chain_tip_minus_lookahead
+                        )
+                     WHERE q.txid = :txid
+                     AND t.id_tx IS NULL",
```
str4d: Tested this, and it fails:
```
INFO steady_state: zallet::components::sync: New chain tip: 3458302 0005c7fd81d914a8d46520486213a7f5f96beb8318516981c522c045a937ae3f
INFO steady_state: zallet::components::sync::steps: Fetching block 0005c7fd81d914a8d46520486213a7f5f96beb8318516981c522c045a937ae3f
INFO data_requests: zallet::components::sync: 11 transaction data requests to service
INFO data_requests: zallet::components::sync: Enhancing 0000000000000000000000000000000000000000000000000000000000000000
INFO poll_transparent: zallet::components::sync: Fetching mined UTXOs
INFO steady_state: zallet::components::sync: Scanning ChainTip(3458302..3458303)
INFO zallet::commands::start: Wallet data-requests sync task exited wallet_sync_result=Err(Error(Context { kind: Sync, backtrace: Some(   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: <unknown>
  15: <unknown>
), source: Some(Other(DbError(SqlInputError { error: Error { code: Unknown, extended_code: 1 }, msg: "near \"q\": syntax error", sql: "DELETE FROM tx_retrieval_queue q\n                     LEFT OUTER JOIN transactions t\n                        ON t.txid = q.txid\n                        AND (\n                          t.expiry_height IS NULL\n                          OR t.expiry_height >= :chain_tip_minus_lookahead\n                        )\n                     WHERE q.txid = :txid\n                     AND t.id_tx IS NULL", offset: 31 }))) }))
INFO zallet::commands::start: Exiting Zallet because an ongoing task exited; asking other tasks to stop
```

```
DbError(
  SqlInputError {
    error: Error {
      code: Unknown,
      extended_code: 1
    },
    msg: "near \"q\": syntax error",
    sql: "DELETE FROM tx_retrieval_queue q\n                     LEFT OUTER JOIN transactions t\n                        ON t.txid = q.txid\n                        AND (\n                          t.expiry_height IS NULL\n                          OR t.expiry_height >= :chain_tip_minus_lookahead\n                        )\n                     WHERE q.txid = :txid\n                     AND t.id_tx IS NULL",
    offset: 31
  }
)
```

str4d (changes requested):
Tested e8ff5845dee23f420b937acb49a9bf0b5f1f0fee

zcash_client_sqlite/src/wallet.rs
str4d: The syntax in https://www.sqlite.org/lang_delete.html suggests that aliasing in `DELETE` isn't allowed, and neither is joining.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2758,13 +2758,21 @@ pub(crate) fn set_transaction_status(
...
-                    )",
+                    "DELETE FROM tx_retrieval_queue
+                     WHERE txid = :txid
+                     AND NOT EXISTS (
+                        SELECT id_tx
+                        FROM transactions t
+                        WHERE t.txid = :txid
+                        AND (
+                          t.expiry_height = 0
+                          OR t.expiry_height >= :chain_tip_minus_lookahead
```
nuttycom: There's a state that is not adequately addressed by this version, which is when `expiry_height IS NULL`. It's unclear though what the right behavior should be in this case - we don't know the expiry height, so we don't have any reference for when we should accept the "not recognized / not in main chain" determination as final.

A transaction record with no expiry information can exist as a consequence of a call to `decrypt_and_store_transaction` constructing the empty transaction record for a transparent transaction input in https://github.com/zcash/librustzcash/blob/zcash_client_sqlite-0.16.x/zcash_client_sqlite/src/wallet.rs#L3377 (which invokes https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/transparent.rs#L1358), For such records we probably want to keep checking for "a while".

One possible solution would be to add an "expiry height" field to `tx_retrieval_queue` and then use that to make the determination as to when we can safely stop making the request.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -497,6 +497,7 @@ CREATE TABLE tx_retrieval_queue (
     txid BLOB NOT NULL UNIQUE,
     query_type INTEGER NOT NULL,
     dependent_transaction_id INTEGER,
+    request_expiry INTEGER,
```
str4d, line 503: Document this new column.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -175,6 +176,7 @@ pub(super) fn all_migrations<
         Box::new(ensure_default_transparent_address::Migration {
             params: params.clone(),
         }),
+        Box::new(tx_retrieval_queue_expiry::Migration),
```
str4d, line 179: Note to self: we need to remember to add a new `V_0_16_4` leaf migration tracker. We also should as part of the 0.16.4 release backport fd9a8ba1ef7b93a4b322ff283afa03930bdf648a which fixes the contents of the `V_0_16_0` and `V_0_16_2` constants. I don't think we need to backport the entirety of #1764, though the migration in d53c02de5268b260f71a0f8be09d9bb6219e33d2 is arguably fixing a bug in 0.16.0 so maybe we should just backport that entire PR.

zcash_client_sqlite/src/wallet/init/migrations/tx_retrieval_queue_expiry.rs
```diff
@@ -0,0 +1,74 @@
...
+        conn.execute_batch(
+            "ALTER TABLE tx_retrieval_queue ADD COLUMN request_expiry INTEGER;
+
+             UPDATE tx_retrieval_queue
+             SET request_expiry = t.expiry_height
+             FROM transactions t
+             WHERE t.txid = tx_retrieval_queue.txid;",
+        )?;
+
+        conn.execute(
```
str4d, line 53: Add a comment to this saying why we're adding an expiry for txids that aren't in `transactions`. In particular, note that it's not possible to have any entries in `tx_retrieval_queue` without also having a `chain_tip_height`. (Hmm, should that occurring be interpreted as database corruption? Or will that edge case be handled elsewhere?)

str4d (approved):
utACK 8a72561ffe535a1d653b140427573adcd0713b85

zcash_client_sqlite/src/wallet/init/migrations.rs
str4d: We forgot; this was added in 0.17.3 (#1927).

Issue #1284, opened 2024-03-18
`zcash_client_sqlite`: Enable migrations to be performed without any seed


str4d:
Some migrations require a seed phrase. Currently we have two:
- UFVK support: seed required to derive the Orchard FVK component for each (derived) account.
- Account ID changes: seed required to derive the seed fingerprint to store with each derived account.

As discussed in #1283, there are cases where a caller wants to query a wallet DB without necessarily having any of the necessary seeds. We should therefore have an opt-in way to migrate a database without any of the relevant seeds, which "downgrades" all accounts to "read-only" (i.e. from `Derived` to `Imported`).

- In a pairing with @nuttycom, we considered this transition in terms of allowing `hd_seed_fingerprint` to be null while `hd_account_index` was non-null. That would be sufficient if we were modifying the latest migration making the account ID changes (as part of Zashi 1.0), but in the general migration case we'd either need to have some other way to track this, or do something like delete the seed fingerprint from the database. For #1283 (and therefore Zashi 1.0) we are avoiding this complexity.

We would also want to pair this with some new method like `WalletWrite::link_seed(Self::AccountId, seed, zip32::AccountId)` that enables "upgrading" a UFVK account from `Imported` to `Derived` (enabing spends from them again).

- We cannot upgrade accounts imported from UIVKs, as we have not performed the necessary historic scanning, and can't do that scanning in-line within a `WalletWrite` call. The caller would need to delete the UIVK account and then separately add the derived account with its UFVK.

PR #1845, opened 2025-06-26
Preparatory work for Zallet wallet.dat import support.

nuttycom:


zcash_client_sqlite/src/lib.rs
```diff
@@ -382,6 +382,12 @@ impl Borrow<rusqlite::Connection> for SqlTransaction<'_> {
     }
 }
 
+impl<C, P, CL, R> WalletDb<C, P, CL, R> {
+    pub fn params(&self) -> &P {
```
str4d, line 387: Document this method.

Also I thought this already existed, but I was thinking of `DbConnection::params` in Zallet.

zcash_keys/Cargo.toml
```diff
@@ -34,6 +34,7 @@ rand_core.workspace = true
 bech32.workspace = true
 bs58.workspace = true
 core2.workspace = true
+regex.workspace = true
```
str4d: Are you using `regex` elsewhere at some point? It's added along with the `zcashd-compat` feature flag and only used behind it in that commit. If it isn't used elsewhere, make it an optional dependency.

zcash_keys/src/keys/zcashd.rs
```diff
@@ -0,0 +1,212 @@
...
+pub fn derive_mnemonic(legacy_seed: SecretVec<u8>) -> Option<Mnemonic> {
+    if legacy_seed.expose_secret().len() != 32 {
+        return None;
+    }
+
+    let mut offset = 0u8;
+    loop {
+        let mut entropy = legacy_seed.expose_secret().clone();
+        entropy[0] += offset;
+        match Mnemonic::<English>::from_entropy(entropy) {
```
str4d, line 24: I looked at the `zcashd` impl, and it does indeed seem like we only ever generated mnemonics using the English wordlist. So this should be fine.

zcash_keys/src/keys/zcashd.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl From<LegacyAddressIndex> for ChildIndex {
+    fn from(id: LegacyAddressIndex) -> Self {
+        // Account IDs are always hardened in derivation paths.
+        ChildIndex::hardened(id.0)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ZcashdHdDerivation {
```
str4d, line 75: Document this enum and its variants.

zcash_keys/src/keys/zcashd.rs
```diff
@@ -0,0 +1,212 @@
...
+/// Errors that can occur in parsing the string representation of an HD key path.
+#[derive(Debug)]
+pub enum PathParseError {
+    PathInvalid,
+    CoinTypeInvalid(String),
+    CoinTypeMismatch { expected: u32, actual: u32 },
+    AccountIndexInvalid(String),
+    AccountIdInvalid(u32),
+    AddressIndexMissing,
+    AddressIndexInvalid(String),
```
str4d, line 105: Document these error variants.

zcash_keys/src/keys/zcashd.rs
```diff
@@ -0,0 +1,212 @@
...
+        if account_index == Self::ZCASHD_LEGACY_ACCOUNT_INDEX {
+            if parts.len() != 4 {
+                return Err(PathParseError::AddressIndexMissing);
+            }
+            let address_index = match parts[3].parse::<u32>() {
+                Ok(v) if v < (1 << 31) => Ok(LegacyAddressIndex(v)),
+                _ => Err(PathParseError::AddressIndexInvalid(parts[3].to_string())),
+            }?;
+            Ok(ZcashdHdDerivation::Post470LegacySapling { address_index })
+        } else {
```
str4d, line 159: Check in this case for `parts.len() != 3`.

zcash_client_backend/src/data_api.rs
```diff
@@ -352,14 +355,22 @@ impl AccountBalance {
...
+    legacy_address_index: Option<zcashd::LegacyAddressIndex>,
 }
 
 impl Zip32Derivation {
     /// Constructs new derivation metadata from its constituent parts.
-    pub fn new(seed_fingerprint: SeedFingerprint, account_index: zip32::AccountId) -> Self {
+    pub fn new(
+        seed_fingerprint: SeedFingerprint,
+        account_index: zip32::AccountId,
+        #[cfg(feature = "zcashd-compat")] legacy_address_index: Option<zcashd::LegacyAddressIndex>,
```
str4d, line 367: We know that this field is only set when `account_index` is the legacy one, so why not instead have a separate constructor for this case behind the feature flag that doesn't have an `account_index` parameter?

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- A `zcashd-compat` feature flag has been added in service of being able to
+  import data from the zcashd `wallet.dat` format. For additional information
+  refer to the `zcash_client_backend-0.20.0` release notes.
+
+### Changed
+- Migrated to `zcash_client_backend-0.20.0`
```
```suggestion
- Migrated to `zcash_client_backend 0.20`.
```

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- A `zcashd-compat` feature flag has been added in service of being able to
+  import data from the zcashd `wallet.dat` format. For additional information
+  refer to the `zcash_client_backend-0.20.0` release notes.
```
```suggestion
  refer to the `zcash_client_backend 0.20.0` release notes.
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -18,7 +18,55 @@ use zcash_protocol::consensus::{NetworkUpgrade, Parameters};
 
 use crate::wallet::scanning::priority_code;
 
-/// Stores information about the accounts that the wallet is tracking.
+/// Stores information about the accounts that the wallet is tracking. An account corresponds to a
+/// logical "bucket of funds" that has its own balance within the wallets and for which spending
+/// operations should treat received value 4
```
str4d: Unfinished sentence. Also, split this paragraph after the first sentence.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -18,7 +18,55 @@ use zcash_protocol::consensus::{NetworkUpgrade, Parameters};
...
+/// - `name`: A human-readable reference for the account. This column is present merely as a
+///   convenience for front-ends and debugging; it has no stable semantics and values are not
+///   required to be unique.
+/// - `uuid`: A wallet-instance-specific identifier for the account. This identifier will remain
+///   stable for the lifetime of the wallet database, but is not expected or required to be
+///   stable across wallet restores and it should not be stored in external backup formats.
+/// - `account_kind`: 0 for accounts derived from a mnemonic seed, 1 for imported accounts
+///   for which derivation path information may not be available. This column may be removed in the
+///   future; the distinction between whether an account is derived or imported is better
+///   represented by the presence or absence of HD seed fingerprint and HD account index data.
```
str4d, line 40: I disagree that presence or absence of derivation information is a "better" measure of this distinction, because they encode two different axes of information. `account_kind` encodes whether the seed for the account has been sighted (and thus can be again if necessary), vs only being accessible indirectly (e.g. on a hardware wallet). This is different from `has_spend_key` (which would be `true` in both cases). The distinction has mattered in the past, and we need to think through issues like #1284.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -18,7 +18,55 @@ use zcash_protocol::consensus::{NetworkUpgrade, Parameters};
...
+///   if any.
+/// - `p2pkh_fvk_item_cache`: The serialized representation of the P2PKH item of the `ufvk`,
+///   if any.
+/// - `birthday_height`: The minimum block height among blocks that may potentially contain
+///   shielded funds belonging to the account.
+/// - `birthday_sapling_tree_size`: A cache of the size of the Sapling note commitment tree
+///   as of the start of the birthday block.
+/// - `birthday_orchard_tree_size`: A cache of the size of the Orchard note commitment tree
+///   as of the start of the birthday block.
+/// - `recover_until_height`: The block height at which wallet recovery was initiated.
```
str4d:
```suggestion
/// - `recover_until_height`: The boundary between recovery and regular scanning for this account.
///   Unscanned blocks up to and excluding this height are counted towards recovery progress. It
///   is initially set via the `AccountBirthday` parameter of the `WalletWrite::import_account_*`
///   methods (usually to the chain tip height at which account recovery was initiated), and may
///   in future be automatically updated by the backend if the wallet is offline for an extended
///   period (to keep the scan progress percentage accurate to what actually needs scanning).
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -18,7 +18,55 @@ use zcash_protocol::consensus::{NetworkUpgrade, Parameters};
...
+///   this wallet database has access to spending key(s) for the account.
+/// - `zcash_legacy_address_index`: This column is only potentially populated for wallets imported
+///   from a `zcashd` `wallet.dat` file, for "standalone" Sapling addresses (each of which
+///   corresponds to an independent account) derived after the introduction of mnemonic seed
+///   derivation in the `4.7.0` `zcashd` release. This column will only be non-null in
+///   the case that the `hd_account_index` column has the value `0x7FFFFFFF`, in accordance with
+///   how post-v4.7.0 Sapling addresses were produced by the `z_getnewaddress` RPC method.
+///   This relationship is not currently enforced by a CHECK constraint; such a constraint should
+///   be added the next time that the `accounts` table is deleted and re-created to support a
+///   SQLite-breaking change to the columns of the table.
```
str4d, line 77: Open an issue for this, so we are less likely to forget.

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -708,18 +739,30 @@ impl<P: consensus::Parameters, C: Clock, R: RngCore> RusqliteMigration for Migra
...
-                generate_gap_addresses(
-                    conn,
-                    &self.params,
-                    AccountRef(account_id),
-                    key_scope,
-                    &GapLimits::default(),
-                    UnifiedAddressRequest::unsafe_custom(Allow, Allow, Require),
-                    false,
-                )?;
+                let gap_limit = match key_scope {
```
str4d, line 747: Add a comment that the code below duplicates `generate_gap_addresses` (and say why it does so; I believe it is to avoid the `get_account_internal` call).

zcash_transparent/src/address.rs
```diff
@@ -407,6 +410,14 @@ impl TransparentAddress {
             }
         }
     }
+
+    /// Derives the P2PKH transparent address corresponding to the given pubkey.
+    #[cfg(feature = "transparent-inputs")]
+    pub fn from_pubkey(pubkey: &secp256k1::PublicKey) -> Self {
```
str4d, line 416: Nit: Does this method actually need to be public? The only place I could find `zcash_transparent::keys::pubkey_to_address` being used is in `zcash-devtool inspect`, and the code there is doing manual script parsing that will all get replaced by `zcash_script`. I guess it's not a particularly harmful method, but in most cases what we want is a way to go from some type containing a `PublicKey` to `TransparentAddress`, rather than directly from `PublicKey` (and the deprecation message on `pubkey_to_address` implied this).

zcash_keys/src/address.rs
```diff
@@ -234,6 +234,17 @@ impl UnifiedAddress {
         );
         result.collect()
     }
+
+    /// Returns the set of receivers in the unified address, excluding unknown receiver types.
+    pub fn receivers(&self) -> Vec<Receiver> {
```
str4d: I dislike that `UnifiedAddress::receivers` is not symmetric to `UnifiedAddress::receiver_types`. Either return unknown receivers from this method, or rename it to be explicit that it returns the subset of known receivers.

zcash_keys/src/address.rs
```diff
@@ -459,6 +470,21 @@ impl Address {
...
+    /// Returns the [`Receiver`]s of this address as a vector. In the case that the underlying
+    /// address is the [`Address::Unified`] variant, this is equivalent to
+    /// [`UnifiedAddress::receivers`] in that it does not return unknown receiver data.
+    pub fn receivers(&self) -> Vec<Receiver> {
+        match self {
+            #[cfg(feature = "sapling")]
+            Address::Sapling(addr) => vec![Receiver::Sapling(*addr)],
+            Address::Transparent(addr) => vec![Receiver::Transparent(*addr)],
+            Address::Unified(ua) => ua.receivers(),
+            Address::Tex(addr) => vec![Receiver::Transparent(TransparentAddress::PublicKeyHash(
```
str4d, line 491: Hmm, this is a potentially dangerous method as it type-erases the TEX nature of those addresses, potentially leading to violating MUST requirements of [ZIP 320](https://zips.z.cash/zip-0320#design-considerations-for-senders).

How is this method intended to be used? Is there any reasonable way to adjust the API such that the ZIP 320 requirements are more obvious to the caller? If not, the method needs to at a minimum document the ZIP 320 requirements.

(The same concern applies to `Address::to_transparent_address`, which is used in `get_transparent_receivers` inside `zcash_client_sqlite`. That method should also have the ZIP 320 requirements documented on it.)

str4d (changes requested):
Reviewed bfb3f14f6b0b0952087261bcd2bc14e68bcc1d00.

zcash_keys/src/keys/zcashd.rs
nuttycom: That would imply a bug in `regex`, but this is a fine defensive check since the type doesn't give us that additional assurance.

zcash_client_backend/src/data_api.rs
nuttycom: This makes for much awkward code in `zcash_client_sqlite`, because then you have to switch which constructor you're using depending on whether the feature flag is enabled or not. I tried it out and it ends up roughly doubling the amount of code for the same meaning.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: I'm not sure what you mean by "sighted"; for example, if a user restores their Zashi backing database to a new device but has lost their seed phrase, this cached "account kind" information may no longer be correct. As such, the wallet shouldn't depend on this metadata; the application should instead consult what capabilities it has access to. Having the mnemonic seed, combined with the derivation path information, is the capability that is required for spending. The `has_spend_key` metadata is only relevant to the backend insofar as we eventually implement the optimization that skips updating the note commitment tree unless the account is marked as being intended for spending from.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/1846

zcash_keys/src/address.rs
nuttycom: The intent of this method is that it be used to get relevant addresses to check (mostly the transparent bits) when the wallet is inspecting chain data to detect that a transaction pertains to the wallet. Since there is no dependency in either direction between `zcash_keys` and `zcash_primitives`, there isn't a good way to implement that check internal to one or the other of those crates. I'll update the documentation to reflect that it should not be used for operations related to transaction creation.

zcash_transparent/src/address.rs
nuttycom: It's needed later in `zcash_client_sqlite`.

zcash_client_sqlite/src/wallet/init/migrations/transparent_gap_limit_handling.rs
```diff
@@ -737,7 +737,9 @@ impl<P: consensus::Parameters, C: Clock, R: RngCore> RusqliteMigration for Migra
 
         // At this point, we have completed updating the infrastructure for gap limit handling,
         // so we can regenerate the gap limit worth of addresses for each account that we
-        // recorded.
+        // recorded. The body of this method duplicates thje logic in `generate_gap_addresses`
```
str4d, line 740:
```suggestion
        // recorded. The body of this method duplicates the logic in `generate_gap_addresses`
```

PR #1852, opened 2025-07-14
zcash_client_backend: Add Near Intents 1Click API as a cryptex source

str4d:


zcash_client_backend/src/tor/http/cryptex/near_intents.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        // Request a quote for selling 1 ZEC, which gives us a bid price (what the DEX
+        // will pay for ZEC).
+        let bid_request = request_quote(
+            "EXACT_INPUT".into(),
+            ZEC_ASSET.into(),
+            USDT_ASSET.into(),
+            DUMMY_ZCASH_MAINNET_ADDR.into(),
+            DUMMY_USDT_ADDR.into(),
+        );
```
str4d, line 127: When submitting this, I get a 400 Bad Request error with the following message (newlines inserted for readability):
```
dry should not be empty,
dry must be a boolean value,
swapType should not be empty,
swapType must be one of the following values: EXACT_INPUT, EXACT_OUTPUT,
slippageTolerance should not be empty,
slippageTolerance must not be greater than 10000,
slippageTolerance must not be less than 0,
slippageTolerance must be an integer number,
originAsset should not be empty,
originAsset must be a string,
depositType should not be empty,
depositType must be one of the following values: ORIGIN_CHAIN, INTENTS,
destinationAsset should not be empty,
destinationAsset must be a string,
amount should not be empty,
amount must be a number string,
refundTo should not be empty,
refundTo must be a string,
refundType should not be empty,
refundType must be one of the following values: ORIGIN_CHAIN, INTENTS,
recipient should not be empty,
recipient must be a string,
recipientType should not be empty,
recipientType must be one of the following values: DESTINATION_CHAIN, INTENTS,
deadline must be a valid ISO 8601 date string
```
As far as I can tell, all of these conditions are satisfied by the above. The only one that I can see might be ambiguous is `deadline must be a valid ISO 8601 date string`, but that would imply the Near Intents 1Click API is not allowing some kinds of valid ISO 8601 date strings. Alternatively, the API might be performing additional validation and is returning the wrong (or a generic) error message.

str4d: The actual JSON being POSTed for `bid` is (newlines inserted for readability):
```json
{
    "dry": true,
    "swapType": "EXACT_OUTPUT",
    "slippageTolerance": 50,
    "originAsset": "usdt.tether-token.near",
    "depositType": "ORIGIN_CHAIN",
    "destinationAsset": "zec.omft.near",
    "amount": "100000000",
    "refundTo": "T9yD14Nj9j7xAB4dbGeiX9h8unkKHxuWwb",
    "refundType": "ORIGIN_CHAIN",
    "recipient": "t1Hsc1LR8yKnbbe3twRp88p6vFfC5t7DLbs",
    "recipientType": "DESTINATION_CHAIN",
    "deadline": "2025-07-14T20:00:58.112864Z"
}
```

PR #1858, opened 2025-07-15
Post-release merge of zcash protocol-0.5.4 to `main`

nuttycom:


zcash_client_backend/src/tor/http/cryptex.rs
```diff
@@ -63,7 +63,6 @@ pub mod exchanges {
 /// An exchange that can be queried for ZEC data.
 #[trait_variant::make(Exchange: Send)]
 #[dynosaur::dynosaur(DynExchange = dyn Exchange)]
-#[dynosaur::dynosaur(DynLocalExchange = dyn LocalExchange)]
```
str4d, line 66: Why is this removed? This is a breaking change AFAICT, and shouldn't ne present even in the difference of the two branches. Is this a merge conflict resolution gone wrong?

zcash_history/src/tree.rs
```diff
@@ -27,7 +27,7 @@ pub struct Tree<V: Version> {
 
 impl<V: Version> Tree<V> {
     /// Resolve link originated from this tree
-    pub fn resolve_link(&self, link: EntryLink) -> Result<IndexedNode<V>, Error> {
+    pub fn resolve_link(&self, link: EntryLink) -> Result<IndexedNode<'_, V>, Error> {
```
str4d, line 30: Where does this change come from?

zcash_history/src/tree.rs
```diff
@@ -274,7 +274,7 @@ impl<V: Version> Tree<V> {
     }
 
     /// Reference to the root node.
-    pub fn root_node(&self) -> Result<IndexedNode<V>, Error> {
+    pub fn root_node(&self) -> Result<IndexedNode<'_, V>, Error> {
```
str4d, line 277: Ditto. I'm confused as to why this compiles.

zcash_client_backend/src/tor/http/cryptex.rs
nuttycom: This generated a dead code warning: 
```
Checking zcash_client_backend v0.19.0 (/home/nuttycom/work/librustzcash/work/zcash_client_backend)
warning: trait `ErasedLocalExchange` is never used
  --> zcash_client_backend/src/tor/http/cryptex.rs:66:1
   |
66 | #[dynosaur::dynosaur(DynLocalExchange = dyn LocalExchange)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

zcash_history/src/tree.rs
nuttycom: These warnings:
```
    Checking zcash_address v0.8.0 (/home/nuttycom/work/librustzcash/work/components/zcash_address)
warning: lifetime flowing from input to output with different syntax can be confusing
  --> zcash_history/src/tree.rs:30:25
   |
30 |     pub fn resolve_link(&self, link: EntryLink) -> Result<IndexedNode<V>, Error> {
   |                         ^^^^^                             -------------- the lifetime gets resolved as `'_`
   |                         |
   |                         this lifetime flows to the output
   |
   = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: one option is to remove the lifetime for references and use the anonymous lifetime for paths
   |
30 |     pub fn resolve_link(&self, link: EntryLink) -> Result<IndexedNode<'_, V>, Error> {
   |                                                                       +++

warning: lifetime flowing from input to output with different syntax can be confusing
   --> zcash_history/src/tree.rs:277:22
    |
277 |     pub fn root_node(&self) -> Result<IndexedNode<V>, Error> {
    |                      ^^^^^            -------------- the lifetime gets resolved as `'_`
    |                      |
    |                      this lifetime flows to the output
    |
help: one option is to remove the lifetime for references and use the anonymous lifetime for paths
    |
277 |     pub fn root_node(&self) -> Result<IndexedNode<'_, V>, Error> {
    |                                                   +++
```

zcash_client_backend/src/tor/http/cryptex.rs
str4d: I looked into this further, and it turns out that neither `DynExchange` nor `DynLocalExchange` was in the public API. This was an accident, but it's also why the dead code error came up, so for now it's fine to remove this (though it prevents putting an `impl LocalExchange` into a `Box`).

str4d (approved):
utACK a3cdb1beea9fe601e9be68683c521f0b7c6514bd

PR #1859, opened 2025-07-16
fix: proper error types for `UnifiedFullViewingKey::decode`

### LINKED ISSUE #598: Add proper error types for the result of `UnifiedFullViewingKey::decode`
This string matching is a bit ugly; can `UnifiedFullViewingKey::decode` return an enum for its error type instead?

_Originally posted by @nuttycom in https://github.com/zcash/zcash-android-wallet-sdk/pull/530#discussion_r902009347_

schell:
These changes refactor and add to the `DecodeError` enum to provide some extra variants for improving error handling in `zcash_keys`. Fixes #598.

### DIFF
--- a/zcash_keys/Cargo.toml
+++ b/zcash_keys/Cargo.toml
@@ -67,6 +67,9 @@ byteorder = { workspace = true, optional = true }
 # - Digests
 blake2b_simd = { workspace = true }
 
+# - Error handling
+thiserror.workspace = true
+
 [dev-dependencies]
 hex.workspace = true
 jubjub.workspace = true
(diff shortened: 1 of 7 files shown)


zcash_keys/Cargo.toml
```diff
@@ -67,6 +67,9 @@ byteorder = { workspace = true, optional = true }
 # - Digests
 blake2b_simd = { workspace = true }
 
+# - Error handling
+snafu.workspace = true
```
nuttycom, line 71: Could `snafu` be a dev dependency instead? If so, I'm amenable to using this instead of `thiserror`, but I would prefer that it not result in an added dependency that could ultimately affect the public API, since `snafu` includes types that should *not* be used in library crates.

schell: I think we talked about this at standup and we net out at "no, it can't just be a dev dependency, but that's ok"?

PR #1862, opened 2025-07-17
zcash_client_sqlite: Restore fees in wallet recovery.

### LINKED ISSUE #1706: Restore of the fees
### What?
Currently when I restore the wallet I see no fees at all. For that case we fallback to a string `typical fee < 0.001`. 

### Goal
If possible, please do the magic and restore the fees or at least as much of it for the transactions as possible. It will significantly improve the UI/UX when going through the history of the transactions but most importantly, it will fix the problem with a brand new feature - tax reporting from an exported csv file. The fee amount is part of it and for the reason mentioned above, we currently fill in no values.

nuttycom:
Fixes #1706

This also discovers and fixes a separate error in transparent balance when `min_confirmations == 0`.


zcash_client_backend/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- The signature of `zcash_client_backend::data_api::testing::TestState::create_proposed_transaction`
+  has been modified to allow transactions to be created from shielding proposals; this API
+  was previously overconstrained. This only affects users of the `test-dependencies` feature.
```
str4d: Is the rationale for this breaking change targeting a `zcash_client_backend 0.18.x` release (which I presume is the intent due to this PR targeting the `zcash_client_sqlite-0.16.x` maintenance branch) that it only affects test usage? If we're going to make semver-breaking changes in patch releases for some feature flags, we should document that policy on the relevant feature flags.

zcash_keys/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+- `zcash_keys::keys::{UnifiedFullViewingKey, UnifiedIncomingViewingKey}::default_transparent_address`
+  has been added under the `test-dependencies` feature.
```
str4d: Going forward, we will need to separate out these kinds of changes into their own maintenance branches. It's less of a concern right now given the messy history of `zcash_client_sqlite-0.16.x`, but for the new maintenance policy we've discussed, these kinds of extra APIs will need to flow through from the dependency's maintenance branch into the downstream ones (or at least have their changes performed in standalone commits that can be cherry-picked around, though that does not work for the patch releases of the dependencies themselves which must be on their relevant maintenance branch for correct tagging).

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3245,3 +3246,86 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, DSF>(
...
+
+    // Get some funds in the source account
+    let note_value = Zatoshis::const_from_u64(350000);
+    st.generate_next_block(&from, AddressType::DefaultExternal, note_value);
+    st.generate_next_block(&from, AddressType::DefaultExternal, note_value);
+    st.scan_cached_blocks(source_account.birthday().height(), 2);
+
+    // Create a sequence of three transactions: two transactions sending from the source account to
+    // a transparent address in the destination account, and then a shielding transaction in the
+    // destination account that pays appropriate fees.
```
str4d: This comment is incorrect in the commit it is added in (5e41581da74df4d51798c3e5ddcdf180782fd2b1); only the first two transactions are being made. The third (shielding) transaction is added to the test in 90527446eec76bf8c90e778cdf8891279a473aa5.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -931,47 +939,54 @@ pub(crate) fn add_transparent_account_balances(
...
+             JOIN accounts a ON a.id = u.account_id
+             JOIN transactions t ON t.id_tx = u.transaction_id
+             WHERE (
+                 -- the transaction that created the output is mined with not enough confirmations
+                (
+                    t.mined_height < :mempool_height
+                    AND :mempool_height - t.mined_height < :min_confirmations -- tx is mined but not confirmed
+                )
+                -- or the tx is unmined but definitely not expired
+                OR (t.expiry_height = 0 OR t.expiry_height >= :mempool_height)
```
str4d: Bug: This is not an unmined check. We can reach this conditional in two cases:
- `t.mined_height >= :mempool_height` (TXO is unmined - good!)
- `:mempool_height - t.mined_height >= :min_confirmations` (TXO is mined and confirmed - bad!)

I believe this is an existing bug, because `AND` has higher precedence than `OR`, so the previous query had the same SQLite operator precedence as this query does after parens insertion. However it is a transient bug because the mined TXO stops being counted once it would have expired. This bug and its transient-ness should be testable.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -898,8 +898,16 @@ pub(crate) fn add_transparent_account_balances(
...
+            -- tx is mined and has at least has at least min_confirmations
+            (
+                t.mined_height < :mempool_height -- tx is mined
+                AND :mempool_height - t.mined_height >= :min_confirmations
+            )
+            -- or the utxo can be spent with zero confirmations and is definitely unexpired
+            OR (
+                :min_confirmations = 0
+                AND (t.expiry_height = 0 OR t.expiry_height >= :mempool_height)
+            )
```
str4d, line 910: This by comparison is fine, because the first conditional always implies `:mempool_height - t.mined_height >= 1` (the `:min_confirmations` in the second half of the first conditional just tightens that bound), and the `:min_confirmations = 0` here thus covers TXOs that are never selected by the first conditional.

zcash_transparent/src/bundle.rs
```diff
@@ -230,6 +230,14 @@ pub struct TxOut {
 }
 
 impl TxOut {
+    // Constructs a new `TxOut` from its constituent parts.
+    pub fn new(value: NonNegativeAmount, script_pubkey: Script) -> Self {
```
str4d, line 234: Document this addition in the changelog. (Also it is currently unnecessary as the struct is fully public, but I agree with transitioning these types to constructors over time.)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3088,13 +3089,82 @@ pub(crate) fn put_block(
...
+            // The sum of the value balances represents the total value of inputs not consumed by
+            // outputs. Without transparent outputs to consume any leftover value, the total is
+            // available for and treated as fee.
+            Ok(Some(
+                value_balance
+                    .try_into()
+                    .map_err(|_| BalanceError::Underflow)?,
+            ))
+        }
+    }
```
str4d, line 3161: This should all be using `TransactionData::fee_paid` (via the `Transaction` deref, so `tx.fee_paid(_)` should work). That method takes a `FnMut` closure that should therefore be able to do whatever it needs to with `conn` internally to look up the input outpoints.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3256,7 +3256,7 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, DSF>(
 pub fn wallet_recovery_computes_fees<T: ShieldedPoolTester, DsF: DataStoreFactory>(
     ds_factory: DsF,
     cache: impl TestCache,
-    intervene: impl FnOnce(&mut DsF::DataStore, TxId) -> Result<(), DsF::DsError>,
+    intervene: &mut impl FnMut(&mut DsF::DataStore, TxId) -> Result<(), DsF::DsError>,
```
str4d: An `&mut` should be unnecessary here. If Rust is requiring `intervene` to be mutable, `mut impl FnMut(_)` is the way to do it.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2111,6 +2111,56 @@ pub(crate) fn get_received_memo(
         .transpose()
 }
 
+fn parse_tx<P: consensus::Parameters>(
```
str4d, line 2114: Nit: the diff for 1b78d1e2bfba0a3bd0cf51325a96583221885650 would be much cleaner if `parse_tx` were just below `get_transaction` instead of just above it.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3167,11 +3167,9 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
     d_tx: DecryptedTransaction<AccountUuid>,
     #[cfg(feature = "transparent-inputs")] gap_limits: &GapLimits,
 ) -> Result<(), SqliteClientError> {
-    // If the transaction is fully shielded, or all transparent inputs are available, set the
-    // fee value.
     let fee = determine_fee(conn, d_tx.tx())?;
```
str4d: Why did d40b3f31a2837cbc51c8e7766fe749f52daffdce remove this comment? AFAICT it is still accurate.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3548,6 +3546,68 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
+            ON ss.transaction_id = t.id_tx
+         LEFT OUTER JOIN sapling_received_notes srn
+            ON srn.tx = :transaction_id
+            AND srn.id = ss.sapling_received_note_id
+         -- find transactions that spend Orchard outputs of the decrypted tx
+         LEFT OUTER JOIN orchard_received_note_spends os
+            ON os.transaction_id = t.id_tx
+         LEFT OUTER JOIN orchard_received_notes orn
+            ON orn.tx = :transaction_id
+            AND orn.id = os.orchard_received_note_id
```
str4d: Do these output cases matter? Having the full transaction that receives a shielded note should have no bearing on the ability to determine the fee of the downstream tx spending that note.

If there is a motivation for doing this for all downstreams (rather than only looking at transparent downstreams), please document it.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3548,6 +3546,68 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
+         )",
+    )?;
+
+    let mut spending_txs_rows = spending_txs_stmt.query(named_params! {
+        ":transaction_id": tx_ref.0
+    })?;
+
+    while let Some(row) = spending_txs_rows.next()? {
+        let spending_tx_ref = row.get(0).map(TxRef)?;
+        let tx_bytes: Vec<u8> = dbg!(row.get(1))?;
```
str4d: Leftover debugging:
```suggestion
        let tx_bytes: Vec<u8> = row.get(1)?;
```

str4d (changes requested):
Reviewed d40b3f31a2837cbc51c8e7766fe749f52daffdce.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: I think you're right. I believe that the fix is:
```suggestion
                -- or the tx is unmined but definitely not expired
                OR (
                    t.mined_height IS NULL
                    AND (t.expiry_height = 0 OR t.expiry_height >= :mempool_height)
                )
```

zcash_client_backend/CHANGELOG.md
nuttycom: Yeah; I think that making the breaking change here is okay because it only affects test dependencies. The problem is that otherwise, I can't actually get tests that use `propose_shielding` to compile at all; apparently, prior to now there haven't been any tests that exercised that code.

zcash_client_backend/src/data_api/testing/pool.rs
nuttycom:
```suggestion
    intervene: mut impl FnMut(&mut DsF::DataStore, TxId) -> Result<(), DsF::DsError>,
```

zcash_client_backend/src/data_api/testing/pool.rs
nuttycom: Ah, yup, that's a commit-splitting error.

zcash_client_sqlite/src/wallet.rs
nuttycom: I know that it's old-fashioned, but my C habits of defining functions before they're used haven't completely gone away. :)

zcash_client_sqlite/src/wallet.rs
nuttycom: Hmm, good point; I guess this does only apply to transparent.

zcash_keys/CHANGELOG.md
nuttycom: Yeah, I guess it makes sense to pull these changes out into those branches, and then merge those branches to this one. Makes for hassles in having to revise/rebase around merges though.

zcash_transparent/src/bundle.rs
nuttycom: Starting that transition was my intent here. Since I need to pull this out into a maintenance branch anyway, I will also change all of the sites where `TxOut` is constructed to use the constructor instead; that's a good maintenance change.

zcash_client_sqlite/src/wallet.rs
nuttycom: This would best be served via a breaking change to the signature of `TransactionData::fee_paid` to allow for the case where the output cannot be obtained by the closure. In retrospect, it's bad that that API makes it seem like that function is total; one has to then plumb the fact that the output is missing into the error type.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/d40b3f31a2837cbc51c8e7766fe749f52daffdce..d5d0b01572d3fcd902b5e6a4306837b6cecc64b0) to address comments from code review. This includes a delta that arose from rebasing after integrating dependency crate changes into the appropriate maintenance branches.

zcash_client_sqlite/src/wallet.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/1869 to track this; I think it's better to make this change in `main` after https://github.com/zcash/librustzcash/pull/1870 is merged.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -898,8 +898,16 @@ pub(crate) fn add_transparent_account_balances(
          JOIN transactions t ON t.id_tx = u.transaction_id
          -- the transaction that created the output is mined and with enough confirmations
          WHERE (
-            t.mined_height < :mempool_height -- tx is mined
-            AND :mempool_height - t.mined_height >= :min_confirmations -- has at least min_confirmations
+            -- tx is mined and has at least has at least min_confirmations
```
schell, line 901: I know I'm late to the game, but there's just a little typo here.

PR #1864, opened 2025-07-18
feat: note selection policy - WIP

schell:
This is a very rough sketch of some confirmation heights/depths changes. 

I'm mostly posting this so we can get some questions out of the way since it's my Friday, and we won't have the usual stand-up together tomorrow, or on my Monday. 

I'll add my questions in a follow up comment as this description will change.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -2078,8 +2078,6 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         &st,
         account_id,
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        // TODO(schell): this height needs to be changed as we're now taking the target_height (chain tip + 1)
-        // instead of anchor height
         (received_tx_height + 10).into(),
         ConfirmationsPolicy::default(),
         &[],
@@ -2096,8 +2094,6 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         &st,
         account_id,
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        // TODO(schell): this height needs to be changed as we're now taking the target_height (chain tip + 1)
-        // instead of anchor height
         (received_tx_height + 10).into(),
         ConfirmationsPolicy::default(),
         &[],
@@ -2154,7 +2150,6 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, DSF: DataStoreFactory>(
         &st,
         account.id(),
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        // TODO(schell): revisit this height, we should be passing chain tip + 1
         (account.birthday().height() + 5).into(),
         ConfirmationsPolicy {
             // 5


nuttycom:
> ## Questions
> Some of these details may have changed since I write them down in my devlog and then update the code after but don't revisit these to make them current, so bear with me.
> 
> ### Type variable `T` suffixes
> In `zcash_client_backend/src/data_api/wallet.rs` we have this wrapper function, which I'm editing to include `ConfirmationPolicy`:
> 
> ```rust
> pub fn propose_transfer<DbT, ParamsT, InputsT, ChangeT, CommitmentTreeErrT>(
>     wallet_db: &mut DbT,
>     params: &ParamsT,
>     spend_from_account: <DbT as InputSource>::AccountId,
>     input_selector: &InputsT,
>     change_strategy: &ChangeT,
>     request: zip321::TransactionRequest,
>     min_confirmations: ConfirmationPolicy,
> ) -> Result<
>     Proposal<ChangeT::FeeRule, <DbT as InputSource>::NoteRef>,
>     ProposeTransferErrT<DbT, CommitmentTreeErrT, InputsT, ChangeT>,
> >
> where
>     DbT: WalletRead + InputSource<Error = <DbT as WalletRead>::Error>,
>     <DbT as InputSource>::NoteRef: Copy + Eq + Ord,
>     ParamsT: consensus::Parameters + Clone,
>     InputsT: InputSelector<InputSource = DbT>,
>     ChangeT: ChangeStrategy<MetaSource = DbT>,
> ```
> 
> Looking at the type sig, I see all the introduced type variables are suffixed with `T`, a bit like denoting a Haskell transformer monad... maybe?

Nope, this is just a suffix to indicate that it's a type parameter; we don't have monad transformers in Rust, and in many cases (elsewhere) we use type parameters that might otherwise collide with the names of concrete types, so I've kind of adopted this convention here; single-character type parameters like we use in Haskell are not in style for this codebase except in relatively trivial cases.

> ### Error handling is quite complicated
> The error handling amongst the various functions in question are quite complicated due to the associated types of each capability. I'm sure this is more me and my unfamiliarity with the different layers of the onion.
> 
> To pass down the `ConfirmationsPolicy` into deeper functions (deeper than the outer `propose_transfer`), without also passing two sets of heights, we would need to provide `WalletRead` capabilities to those functions. I'm not sure if that's acceptable or not, or what the implications of that are, so I'm going to take another approach where I do the height computation up at the top and then pass a struct down.

It should be sufficient to pass the `ConfirmationsPolicy` along with `target_height`; the other heights are then computed by subtraction from the target height (which is the minimum height at which the transaction could potentially be included in the chain, ignoring reorgs.)

> ### Depths vs Heights
> In the literature (https://zips.z.cash/zip-0315#anchor-selection) we seem to use "depth" and "height" interchangeably. The code has `BlockHeight`. Hrm, well maybe it's "depth" for the number of confirmations and "height" for blocks. Why use "depth"? Are there other dimensions to confirmations?

`depth` is a count of a number of blocks back from the chain tip. So depth 0 is the `chain_tip_height`, depth 1 is `chain_tip_height - 1`, and so forth. The number of confirmations is yet a third thing: 0 confirmations for a transaction in the mempool, 1 confirmation for a transaction at depth 0, 2 confirmations for a transaction at depth 1, and so forth.

> ### Height Arithmetic
> @nuttycom - we talked briefly about doing some anchor height arithmetic between trusted and untrusted heights, but I neglected to write it down and now have completely forgotten the details. Can you refresh my memory? 🙇

The height to use for querying for notes should be the target height minus the number of confirmations. The changes in  #1862 to the queries can be instructive here.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -270,7 +289,7 @@ pub fn propose_transfer<DbT, ParamsT, InputsT, ChangeT, CommitmentTreeErrT>(
     input_selector: &InputsT,
     change_strategy: &ChangeT,
     request: zip321::TransactionRequest,
-    min_confirmations: NonZeroU32,
+    min_confirmations: ConfirmationsPolicy,
```
schell: NOTE: Should probably change the name of this (at all sites).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -282,22 +301,23 @@ where
     InputsT: InputSelector<InputSource = DbT>,
     ChangeT: ChangeStrategy<MetaSource = DbT>,
 {
-    let (target_height, anchor_height) = wallet_db
-        .get_target_and_anchor_heights(min_confirmations)
-        .map_err(|e| Error::from(InputSelectorError::DataSource(e)))?
-        .ok_or_else(|| Error::from(InputSelectorError::SyncRequired))?;
+    let maybe_intial_heights = wallet_db
+        .get_target_and_anchor_heights(min_confirmations.untrusted)
```
schell: NOTE: this should be the **trusted** height because we want to be able to spend trusted notes with fewer confirmations than untrusted notes, and then we can use the untrusted depth for the selection later on.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -224,6 +229,7 @@ where
     //    well as a single note for which the sum was greater than or equal to the
     //    required value, bringing the sum of all selected notes across the threshold.
     let mut stmt_select_notes = conn.prepare_cached(
+        // TODO(schell): query should take anchor height of untrusted as well
```
schell: NOTE: determine if trusted or untrusted, then filter based on that

.helix/languages.toml
```diff
@@ -0,0 +1,5 @@
+[language-server.rust-analyzer.config.cargo]
+features = "all"
+
+[language-server.rust-analyzer.config.server.extraEnv]
+RUSTUP_TOOLCHAIN = "stable"
```
schell, line 5: This is very similar to the `.vscode/settings.json` file that includes extra features for the IDE's `rust-analyzer` instance.

zcash_client_backend/src/data_api.rs
```diff
@@ -1294,7 +1294,8 @@ pub trait InputSource {
         account: Self::AccountId,
         target_value: TargetValue,
         sources: &[ShieldedProtocol],
-        anchor_height: BlockHeight,
+        target_height: zcash_protocol::consensus::TargetHeight,
```
schell, line 1297: There's something about the `delegatable_trait` attribute macro that is hiding the import of `TargetHeight`. I'm not sure why `BlockHeight` isn't affected. Maybe this is just on my machine? Anyway - that's why I'm qualifying here.

nuttycom: What's happening here is likely to be related to feature flagging.

PR #1867, opened 2025-07-18
zcash_client_backend: Fix overly-constrained type signature of TestState::create_proposed_transaction

nuttycom:
The existing type signature of this method made it impossible to use it with a proposal created by `zcash_client_backend::data_api::wallet::propose_shielding`.

This is technically a semver-breaking change, but as it only affects test code under the `test-dependencies` feature flag, it's worth it to include this in a patch release.

Once it has an approving review, this PR will merge as part of #1862

### DIFF
--- a/zcash_client_backend/Cargo.toml
+++ b/zcash_client_backend/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "zcash_client_backend"
 description = "APIs for creating shielded Zcash light clients"
-version = "0.18.0"
+version = "0.18.1"
 authors = [
     "Jack Grigg <jack@z.cash>",
     "Kris Nuttycombe <kris@electriccoin.co>"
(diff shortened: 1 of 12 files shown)


zcash_client_backend/Cargo.toml
```diff
@@ -220,6 +220,9 @@ tor = [
 ]
 
 ## Exposes APIs that are useful for testing, such as `proptest` strategies.
+##
+## NOTE: Semver-breaking changes to the APIs exposed by this feature may be
+## present in minor release versions.
```
str4d:
```suggestion
## present in patch release versions.
```
or
```suggestion
## present in any release.
```
(I know you mean "non-semver-breaking", but for pre-1.0 crates minor release versions are the breaking ones, so the current wording is confusing.)

Make an equivalent change to the changelog.

str4d (changes requested):
utACK ef2e67c7ab0a419222fda8fe8b245162020cb86b once the docs are fixed.

PR #1870, opened 2025-07-18
zcash_primitives: Make `Transaction::fee_paid` internally resilient to missing inputs.

nuttycom:
Previously, the closure passed to `Transaction::fee_paid` that performs input value retrieval did not provide any way to directly report that a requested UTXO could not be found, and consequently this absence had to be handled via the error pathway. This change modifies both the type of this function, and the result type of `fee_paid` to reflect this common circumstance. In addition, this change modifies the overall API to ensure that negative input amounts and negative fees are not representable.

### DIFF
--- a/zcash_transparent/src/bundle.rs
+++ b/zcash_transparent/src/bundle.rs
@@ -130,7 +130,8 @@ impl<A: Authorization> Bundle<A> {
             .sum::<Option<ZatBalance>>()
             .ok_or(BalanceError::Overflow)?;
 
-        Ok(Some(ZatBalance::from(input_sum) - output_sum).ok_or(BalanceError::Underflow)?)
+        let balance = (ZatBalance::from(input_sum) - output_sum).ok_or(BalanceError::Underflow)?;
+        Ok(Some(balance))
     }
 }
 


daira (approved):
utACK

daira (dismissed):
re-utACK

zcash_transparent/src/bundle.rs
```diff
@@ -109,24 +109,28 @@ impl<A: Authorization> Bundle<A> {
...
             .vout
             .iter()
-            .map(|p| Amount::from(p.value))
-            .sum::<Option<Amount>>()
+            .map(|p| ZatBalance::from(p.value))
+            .sum::<Option<ZatBalance>>()
             .ok_or(BalanceError::Overflow)?;
 
-        (input_sum - output_sum).ok_or_else(|| BalanceError::Underflow.into())
+        Ok(Some(ZatBalance::from(input_sum) - output_sum).ok_or(BalanceError::Underflow)?)
```
str4d: This line is equivalent to:
```rust
let balance: Option<_> = ZatBalance::from(input_sum) - output_sum;
Ok(Some(balance).ok_or(BalanceError::Underflow)?)
```
which means the `Option` that is getting underflow-checked is the hard-coded `Some`, not the actual potentially-underflowing operation. I think this should instead be:
```suggestion
        Ok(Some((ZatBalance::from(input_sum) - output_sum).ok_or(BalanceError::Underflow)?))
```

nuttycom: Good catch; this was a parenthesization error. I think it's clearer to assign the balance on a separate line.

PR #1879, opened 2025-07-21
[ZIP-233] Network Sustainability Mechanism: Burning

nuttycom:
This builds atop #1567 to bring it up to date with the current state of `main` and address comments from code review.


nuttycom (commented):
utACK 304c71bf35c279be75bb275dfd536116f292366c - this is left as a comment because I opened this PR to subsume #1567

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1281,7 +1318,7 @@ mod tests {
         let mut witness1 = IncrementalWitness::from_tree(tree.clone()).unwrap();
 
         // Fail if there is insufficient input
-        // 0.0003 z-ZEC out, 0.0002 t-ZEC out, 0.0001 t-ZEC fee, 0.00059999 z-ZEC in
+        // 0.0003 z-ZEC out, 0.00015 t-ZEC out, 0.0001 t-ZEC fee, 0.00059999 z-ZEC in
```
daira: I believe this is incorrect (pre-existing error):
```suggestion
        // 0.0003 z-ZEC out, 0.00015 t-ZEC out, 0.00015 t-ZEC fee, 0.00059999 z-ZEC in
```
There are three logical actions in this tx: two for Sapling (max of two spends after padding, and one output), and one for the transparent output.

(Nit: these amounts are simulated TAZ, not simulated ZEC, but don't fix that in this PR.)

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1315,6 +1352,43 @@ mod tests {
             );
         }
 
+        // Fail if there is insufficient input
+        // 0.0003 z-ZEC out, 0.00005 t-ZEC out, 0.0001 burned, 0.0001 t-ZEC fee, 0.00059999 z-ZEC in
```
daira: Same issue as above.
```suggestion
        // 0.0003 z-ZEC out, 0.00005 t-ZEC out, 0.0001 burned, 0.00015 t-ZEC fee, 0.00059999 z-ZEC in
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1326,6 +1400,54 @@ mod tests {
...
+            assert_eq!(
+                res.transaction()
+                    .fee_paid(|_| Err(BalanceError::Overflow))
+                    .unwrap(),
+                ZatBalance::const_from_i64(15_000)
+            );
+        }
+
+        // Succeeds if there is sufficient input
+        // 0.0003 z-ZEC out, 0.00005 t-ZEC out, 0.0001 burned, 0.00015 t-ZEC fee, 0.0006 z-ZEC in
```
daira, line 1473: This is correct.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1326,6 +1400,54 @@ mod tests {
 
         // Succeeds if there is sufficient input
         // 0.0003 z-ZEC out, 0.00015 t-ZEC out, 0.00015 t-ZEC fee, 0.0006 z-ZEC in
```
daira, line 1426: This is correct.

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -216,6 +223,14 @@ pub fn write_v5_bundle<W: Write>(
     Ok(())
 }
```
daira, line 224: I suggest writing
```rust
// also used by `write_v6_bundle`
```
at the top of the body of `write_v5_bundle` (not in the doc comment), and similarly for other cases, since it's a potential maintenance hazard to have a function called `write_v5_bundle` be used also for v6.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -60,7 +66,7 @@ pub use zcash_protocol::TxId;
 /// transaction fields. Note that this is not dependent on epoch, only on transaction encoding.
 /// For example, if a particular epoch defines a new transaction version but also allows the
 /// previous version, then only the new version would be added to this enum.
```
daira:
```suggestion
/// previous version, then only the new version would be added to this enum.
///
/// The derivation of `PartialOrd` depends on the versions being in order with
/// `ZFuture` just after the tx version that it is based on.
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -730,22 +772,51 @@ impl Transaction {
...
+    fn read_v6<R: Read>(mut reader: R, version: TxVersion) -> io::Result<Self> {
+        let (consensus_branch_id, lock_time, expiry_height, _zip233_amount) =
+            Self::read_v6_header_fragment(&mut reader)?;
+
+        let transparent_bundle = Self::read_transparent(&mut reader)?;
+        let sapling_bundle = sapling_serialization::read_v5_bundle(&mut reader)?;
+        let orchard_bundle = orchard_serialization::read_v6_bundle(&mut reader)?;
+
+        #[cfg(zcash_unstable = "zfuture")]
+        let tze_bundle = Self::read_tze(&mut reader)?;
```
daira:
```suggestion
        let tze_bundle = if version.has_tze() {
            Self::read_tze(&mut reader)?
        } else {
            None
        }
```
(or ``version.has_tze().then(|| Self.read_tze(&mut reader)).transpose()?``, I think.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -761,7 +832,7 @@ impl Transaction {
         let consensus_branch_id = reader.read_u32_le().and_then(|value| {
```
daira, line 902:
```suggestion
        // also used by `read_v6_header_fragment`
        let consensus_branch_id = reader.read_u32_le().and_then(|value| {
```

components/zcash_protocol/src/value.rs
```diff
@@ -478,6 +478,14 @@ impl Div<NonZeroU64> for Zatoshis {
     }
 }
 
+impl Neg for Zatoshis {
+    type Output = ZatBalance;
+
+    fn neg(self) -> ZatBalance {
+        ZatBalance(-i64::try_from(self.0).expect("Zatoshis value fits into i64"))
+    }
```
arya2: Nitpick: The `try_from()` & `expect()` calls are unnecessary.

```suggestion
    fn neg(self) -> Self::Output {
        ZatBalance::from(self).neg()
    }
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -864,6 +879,8 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
             consensus_branch_id: unauthed_tx.consensus_branch_id,
             lock_time: unauthed_tx.lock_time,
             expiry_height: unauthed_tx.expiry_height,
+            #[cfg(all(zcash_unstable = "nu7", feature = "zip-233"))]
+            zip233_amount: self.zip233_amount,
```
arya2: Nitpick: Set field from `unauthed_tx` for consistency

```suggestion
            zip233_amount: unauthed_tx.zip233_amount,
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -603,6 +632,17 @@ impl Transaction {
         Transaction { txid, data }
     }
 
+    #[cfg(any(zcash_unstable = "zfuture", zcash_unstable = "nu7"))]
+    fn from_data_v6(data: TransactionData<Authorized>) -> Self {
```
arya2, line 693: @daira should `from_data_v5()` be renamed to `from_data_v5_and_v6()` instead?

zcash_primitives/src/transaction/mod.rs
```diff
@@ -621,9 +661,9 @@ impl Transaction {
             }
             TxVersion::V5 => Self::read_v5(reader.into_base_reader(), version),
             #[cfg(zcash_unstable = "nu7")]
-            TxVersion::V6 => todo!(),
+            TxVersion::V6 => Self::read_v6(&mut reader.into_base_reader(), version),
```
arya2: Nitpick: Borrow is unnecessary here.

```suggestion
            TxVersion::V6 => Self::read_v6(reader.into_base_reader(), version),
```

zcash_primitives/src/transaction/mod.rs
arya2: Alternative needs a `flatten()` call.

```rust
        let tze_bundle = version
            .has_tze()
            .then(|| Self.read_tze(&mut reader))
            .transpose()?
            .flatten();
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -777,13 +848,36 @@ impl Transaction {
         Ok((consensus_branch_id, lock_time, expiry_height))
     }
 
+    #[cfg(any(zcash_unstable = "zfuture", zcash_unstable = "nu7"))]
+    fn read_v6_header_fragment<R: Read>(
+        mut reader: R,
+    ) -> io::Result<(BranchId, u32, BlockHeight, Zatoshis)> {
```
arya2: Nitpick:

Could use a type alias to return a different type if the `zip-233` feature is selected, like:

```rust
#[cfg(feature = "zip-233")]
type READ_V6_HEADER_FRAGMENT_RETURN = (BranchId, u32, BlockHeight, Zatoshis);
#[cfg(not(feature = "zip-233"))]
type READ_V6_HEADER_FRAGMENT_RETURN= (BranchId, u32, BlockHeight);

    fn read_v6_header_fragment<R: Read>(
        mut reader: R,
    ) -> io::Result<READ_V6_HEADER_FRAGMENT_RETURN> {

```

Or read the zip_233_amount outside of `read_v6_header_fragment()`.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -777,13 +848,36 @@ impl Transaction {
...
         reader: R,
     ) -> io::Result<Option<sapling::Bundle<sapling::bundle::Authorized, ZatBalance>>> {
         sapling_serialization::read_v5_bundle(reader)
     }
 
+    #[cfg(all(zcash_unstable = "nu7", feature = "zip-233"))]
+    fn read_zip233_amount<R: Read>(mut reader: R) -> io::Result<Zatoshis> {
+        Zatoshis::from_u64(reader.read_u64_le()?)
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "zip233Amount out of range"))
+            .map(Into::into)
```
arya2:
```suggestion
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -804,7 +898,7 @@ impl Transaction {
             TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 => self.write_v4(writer),
             TxVersion::V5 => self.write_v5(writer),
             #[cfg(zcash_unstable = "nu7")]
-            TxVersion::V6 => todo!(),
+            TxVersion::V6 => self.write_v6(writer),
```
arya2, line 971: `TxVersion::ZFuture` arm should use `write_v6()` for consistency with `read()`.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -879,6 +973,25 @@ impl Transaction {
...
+            ));
+        }
+        self.write_v6_header(&mut writer)?;
+
+        self.write_transparent(&mut writer)?;
+        self.write_v5_sapling(&mut writer)?;
+        orchard_serialization::write_v6_bundle(self.orchard_bundle.as_ref(), &mut writer)?;
+
+        #[cfg(zcash_unstable = "zfuture")]
+        self.write_tze(&mut writer)?;
```
arya2: This can be removed from `write_v5()` if using `read_v6()`/`write_v6()` for ZFuture.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -887,6 +1000,18 @@ impl Transaction {
         Ok(())
     }
 
+    #[cfg(zcash_unstable = "nu7")]
+    pub fn write_v6_header<W: Write>(&self, mut writer: W) -> io::Result<()> {
+        self.version.write(&mut writer)?;
+        writer.write_u32_le(u32::from(self.consensus_branch_id))?;
+        writer.write_u32_le(self.lock_time)?;
+        writer.write_u32_le(u32::from(self.expiry_height))?;
```
arya2, line 1077: @daira should this re-use `write_v5_header()`?

```suggestion
        self.write_v5_header(&mut writer)?;
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1033,7 +1162,7 @@ pub mod testing {
         }
     }
 
-    #[cfg(not(zcash_unstable = "zfuture"))]
+    #[cfg(all(not(zcash_unstable = "nu7"), not(zcash_unstable = "zfuture")))]
```
arya2, line 1240: This looks like a duplicate of the next `prop_compose!`

```suggestion
    #[cfg(all(not(feature= "zip-233"), not(zcash_unstable = "zfuture")))]
```

zcash_primitives/src/transaction/tests.rs
```diff
@@ -36,6 +39,7 @@ fn tx_read_write() {
     assert_eq!(&data[..], &encoded[..]);
 }
 
+#[cfg(test)]
```
arya2, line 42: It would be nice to export `transaction::tests::data` from lib directly instead of adding everything in `transaction::tests` into compilations with the `test-dependencies` feature enabled and hiding most of it individually.

arya2 (commented):
This is looking good.

zcash_primitives/src/transaction/components/orchard.rs
nuttycom: Given the changes in ZIP 230, we will need an entirely separate `write_v6_bundle` implementation, so I will leave this as a TODO.

zcash_primitives/src/transaction/mod.rs
nuttycom: This should not have ordering; instead, the correct approach is to define `TxVersion::has_zip233`, and have the implementation of that method internally perform the necessary determination.

zcash_primitives/src/transaction/mod.rs
nuttycom: No, having separate methods with identical implementations is preferable here.

zcash_primitives/src/transaction/mod.rs
nuttycom: This is where a struct would be better.

zcash_primitives/src/transaction/mod.rs
nuttycom: This is short enough that the duplication is okay; also, explicit fees with change the implementation here further.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -813,7 +928,7 @@ impl Transaction {
             TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 => self.write_v4(writer),
             TxVersion::V5 => self.write_v5(writer),
             #[cfg(zcash_unstable = "nu7")]
-            TxVersion::V6 => todo!(),
+            TxVersion::V6 => self.write_v6(writer),
             #[cfg(zcash_unstable = "zfuture")]
             TxVersion::ZFuture => self.write_v5(writer),
```
daira:
```suggestion
            TxVersion::ZFuture => self.write_v6(writer),
```
Blocking.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1051,6 +1200,84 @@ pub mod testing {
...
+    }
+
+    #[cfg(all(zcash_unstable = "nu7", feature = "zip-233"))]
+    prop_compose! {
+        pub fn arb_txdata(consensus_branch_id: BranchId)(
+            version in arb_tx_version(consensus_branch_id)
+        )(
+            lock_time in any::<u32>(),
+            expiry_height in any::<u32>(),
+            zip233_amount in 0..MAX_MONEY,
```
daira:
```suggestion
            zip233_amount in 0..=MAX_MONEY,
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1061,18 +1288,20 @@ pub mod testing {
...
 
-    #[cfg(zcash_unstable = "zfuture")]
+    #[cfg(all(zcash_unstable = "zfuture", feature = "zip-233"))]
     prop_compose! {
         pub fn arb_txdata(consensus_branch_id: BranchId)(
             version in arb_tx_version(consensus_branch_id),
         )(
             lock_time in any::<u32>(),
             expiry_height in any::<u32>(),
+            zip233_amount in 0..MAX_MONEY,
```
daira:
```suggestion
            zip233_amount in 0..=MAX_MONEY,
```

daira (dismissed):
One blocking comment, otherwise utACK.

daira (dismissed):
re-utACK

aphelionz:
Whooo!!! 🥳

PR #1885, opened 2025-07-30
zcash_protocol: Stabilise `NetworkUpgrade::Nu6_1`

str4d:


nuttycom (dismissed):
re-utACK 56ddab94b66f701bfd0a80a35841613c87b94f61

components/zcash_protocol/src/consensus.rs
```diff
@@ -441,7 +441,7 @@ impl Parameters for TestNetwork {
             NetworkUpgrade::Canopy => Some(BlockHeight(1_028_500)),
             NetworkUpgrade::Nu5 => Some(BlockHeight(1_842_420)),
             NetworkUpgrade::Nu6 => Some(BlockHeight(2_976_000)),
-            NetworkUpgrade::Nu6_1 => None,
+            NetworkUpgrade::Nu6_1 => Some(BlockHeight(3_536_500)),
```
nuttycom, line 444: I have verified that this height is approximately two weeks from the release height.

Issue #982, opened 2023-09-21
zcash_client_backend: Use knowledge of inserted treestates to reduce time to spendability


nuttycom:
Goal: to reduce the time to spendability by adjusting how work at the chain tip is prioritized.

TODO: summarize in more detail what is now proposed taking into account the discussion below.

----

### Original description

In order to minimize the amount of scanning work required at the chain tip, we should implement the following:

- Add a priority that is higher than `ChainTip`, call it `AnchorRange` or something of the sort. The semantics of an `AnchorRange` scanning range are that when a wallet encounters such a range, it immediately downloads the treestate corresponding to the last block prior to the start of the range and inserts that frontier directly into the note commitment tree (using a to-be-defined `WalletCommitmentTrees` method).
- After updating scan ranges, the following must hold:
   - the `AnchorRange` scanning range is at most `PRUNING_DEPTH` blocks long and has its end equal to the `chain_tip + 1`
   - any previously existing `AnchorRange` range with a starting height < `stable_height` has its priority reduced to `Historic`
   - if the the shard containing `stable_height` has no unscanned ranges below `stable_height`, no `AnchorRange` must exist, because in this case it's unnecessary to download or insert any additional tree state.
   
It might be possible to repurpose the `ChainTip` priority to have these semantics, but I think it's better to be explicit about the circumstance where the tree state should be downloaded, and we should continue to support wallets that don't want to download tree states; such wallets will treat `AnchorRange` ranges in the same fashion as they currently treat `ChainTip` ranges.

str4d:
Do we even need to modify the scan ranges? This feels like it is trying to shoe-horn additional semantics or signals into `suggest_scan_ranges`, that complicate it over "scan these ranges in priority order".

The need to obtain a tree state at the anchor height is associated with a change to the chain tip height. We could therefore instead modify `update_chain_tip` to return metadata that includes "height at which the wallet should be given a tree state to make its desired anchor spendable".

nuttycom:
> The need to obtain a tree state at the anchor height is associated with a change to the chain tip height. We could therefore instead modify `update_chain_tip` to return metadata that includes "height at which the wallet should be given a tree state to make its desired anchor spendable". If the caller ignores that metadata, then they continue to work correctly because they scan `ChainTip`.

[...]

PR #1888, opened 2025-07-31
feat: zip 315 confirmations policy

schell:
These changes add `ConfirmationsPolicy` which specify a customizable number of confirmations for selecting spendable notes.

[ZIP 315](https://zips.z.cash/zip-0315)

* adds `ConfirmationsPolicy` for selecting spendable notes 
* adds `TargetHeight` as a wrapper for `BlockHeight` 
* fixes an expiry bug in note selection 
 
Related:
* https://github.com/zcash/librustzcash/pull/1864 (closed in favor of this PR)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,8 @@ members = [
     "zcash_transparent",
 ]
 
+resolver = "2"
+
 [workspace.package]
 edition = "2021"
 rust-version = "1.81"
--- a/components/zcash_protocol/src/consensus.rs
+++ b/components/zcash_protocol/src/consensus.rs
@@ -129,9 +129,8 @@ impl Sub<BlockHeight> for BlockHeight {
 /// A wrapper type around [`BlockHeight`] that represents the _next_ chain tip.
 ///
 /// Addition and subtraction are provided by proxying to [`BlockHeight`].
-// TODO(schell): determine if we actually need PartialEq, Eq, Hash
 #[repr(transparent)]
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct TargetHeight(BlockHeight);
 
 impl From<BlockHeight> for TargetHeight {
--- a/components/zcash_protocol/src/consensus.rs
+++ b/components/zcash_protocol/src/consensus.rs
@@ -126,6 +126,60 @@ impl Sub<BlockHeight> for BlockHeight {
     }
 }
 
+/// A wrapper type around [`BlockHeight`] that represents the _next_ chain tip.
+///
+/// Addition and subtraction are provided by proxying to [`BlockHeight`].
+// TODO(schell): determine if we actually need PartialEq, Eq, Hash
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct TargetHeight(BlockHeight);
+
+impl From<BlockHeight> for TargetHeight {
+    fn from(value: BlockHeight) -> Self {
+        TargetHeight(value)
+    }
+}
+
+impl From<TargetHeight> for BlockHeight {
+    fn from(value: TargetHeight) -> Self {
+        value.0
+    }
+}
+
+impl From<TargetHeight> for u32 {
+    fn from(value: TargetHeight) -> Self {
+        value.0 .0
+    }
+}
+
+impl From<u32> for TargetHeight {
+    fn from(value: u32) -> Self {
+        BlockHeight::from_u32(value).into()
+    }
+}
+
+impl<I> Add<I> for TargetHeight
+where
+    BlockHeight: Add<I>,
+{
+    type Output = <BlockHeight as Add<I>>::Output;
+
+    fn add(self, rhs: I) -> Self::Output {
+        self.0 + rhs
+    }
+}
+
+impl<I> Sub<I> for TargetHeight
+where
+    BlockHeight: Sub<I>,
+{
+    type Output = <BlockHeight as Sub<I>>::Output;
+
+    fn sub(self, rhs: I) -> Self::Output {
+        self.0 - rhs
+    }
+}
+
 /// Constants associated with a given Zcash network.
 pub trait NetworkConstants: Clone {
     /// The coin type for ZEC, as defined by [SLIP 44].
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -64,7 +64,6 @@ use std::{
     io,
     num::{NonZeroU32, TryFromIntError},
 };
-use wallet::ConfirmationsPolicy;
 
 use incrementalmerkletree::{frontier::Frontier, Retention};
 use shardtree::{error::ShardTreeError, store::ShardStore, ShardTree};
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -76,7 +76,7 @@ use zcash_keys::{
 };
 use zcash_primitives::{block::BlockHash, transaction::Transaction};
 use zcash_protocol::{
-    consensus::BlockHeight,
+    consensus::{BlockHeight, TargetHeight},
     memo::{Memo, MemoBytes},
     value::{BalanceError, Zatoshis},
     ShieldedProtocol, TxId,
@@ -1294,7 +1294,7 @@ pub trait InputSource {
         account: Self::AccountId,
         target_value: TargetValue,
         sources: &[ShieldedProtocol],
-        target_height: BlockHeight,
+        target_height: zcash_protocol::consensus::TargetHeight,
         confirmations_policy: wallet::ConfirmationsPolicy,
         exclude: &[Self::NoteRef],
     ) -> Result<SpendableNotes<Self::NoteRef>, Self::Error>;
@@ -1339,7 +1339,7 @@ pub trait InputSource {
     fn get_spendable_transparent_outputs(
         &self,
         _address: &TransparentAddress,
-        _target_height: BlockHeight,
+        _target_height: zcash_protocol::consensus::TargetHeight,
         _min_confirmations: u32,
     ) -> Result<Vec<WalletTransparentOutput>, Self::Error> {
         Ok(vec![])
@@ -1507,7 +1507,7 @@ pub trait WalletRead {
     fn get_target_and_anchor_heights(
         &self,
         min_confirmations: NonZeroU32,
-    ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error>;
+    ) -> Result<Option<(zcash_protocol::consensus::TargetHeight, BlockHeight)>, Self::Error>;
 
     /// Returns the block height in which the specified transaction was mined, or `Ok(None)` if the
     /// transaction is not in the main chain.
@@ -2098,7 +2098,7 @@ impl<'a, AccountId> DecryptedTransaction<'a, AccountId> {
 pub struct SentTransaction<'a, AccountId> {
     tx: &'a Transaction,
     created: time::OffsetDateTime,
-    target_height: BlockHeight,
+    target_height: TargetHeight,
     account: AccountId,
     outputs: &'a [SentTransactionOutput<AccountId>],
     fee_amount: Zatoshis,
@@ -2121,7 +2121,7 @@ impl<'a, AccountId> SentTransaction<'a, AccountId> {
     pub fn new(
         tx: &'a Transaction,
         created: time::OffsetDateTime,
-        target_height: BlockHeight,
+        target_height: TargetHeight,
         account: AccountId,
         outputs: &'a [SentTransactionOutput<AccountId>],
         fee_amount: Zatoshis,
@@ -2166,7 +2166,7 @@ impl<'a, AccountId> SentTransaction<'a, AccountId> {
     }
 
     /// Returns the block height that this transaction was created to target.
-    pub fn target_height(&self) -> BlockHeight {
+    pub fn target_height(&self) -> TargetHeight {
         self.target_height
     }
 }
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -64,6 +64,7 @@ use std::{
     io,
     num::{NonZeroU32, TryFromIntError},
 };
+use wallet::ConfirmationsPolicy;
 
 use incrementalmerkletree::{frontier::Frontier, Retention};
 use shardtree::{error::ShardTreeError, store::ShardStore, ShardTree};
@@ -1294,7 +1295,8 @@ pub trait InputSource {
         account: Self::AccountId,
         target_value: TargetValue,
         sources: &[ShieldedProtocol],
-        anchor_height: BlockHeight,
+        target_height: BlockHeight,
+        confirmations_policy: wallet::ConfirmationsPolicy,
         exclude: &[Self::NoteRef],
     ) -> Result<SpendableNotes<Self::NoteRef>, Self::Error>;
 
--- a/zcash_client_backend/src/data_api/testing.rs
+++ b/zcash_client_backend/src/data_api/testing.rs
@@ -922,8 +922,6 @@ where
             from_account.usk(),
             request,
             OvkPolicy::Sender,
-            // TODO(schell): Check this passes tests.
-            // This was previously `NonZeroU32::MIN`, which is 1.
             ConfirmationsPolicy::MIN,
         )
     }
--- a/zcash_client_backend/src/data_api/testing.rs
+++ b/zcash_client_backend/src/data_api/testing.rs
@@ -35,7 +35,7 @@ use zcash_primitives::{
 };
 use zcash_proofs::prover::LocalTxProver;
 use zcash_protocol::{
-    consensus::{self, BlockHeight, Network, NetworkUpgrade, Parameters as _},
+    consensus::{self, BlockHeight, Network, NetworkUpgrade, Parameters as _, TargetHeight},
     local_consensus::LocalNetwork,
     memo::{Memo, MemoBytes},
     value::{ZatBalance, Zatoshis},
@@ -2501,7 +2501,7 @@ impl InputSource for MockWalletDb {
         _account: Self::AccountId,
         _target_value: TargetValue,
         _sources: &[ShieldedProtocol],
-        _anchor_height: BlockHeight,
+        _anchor_height: TargetHeight,
         _confirmations_policy: ConfirmationsPolicy,
         _exclude: &[Self::NoteRef],
     ) -> Result<SpendableNotes<Self::NoteRef>, Self::Error> {
@@ -2621,7 +2621,7 @@ impl WalletRead for MockWalletDb {
     fn get_target_and_anchor_heights(
         &self,
         _min_confirmations: NonZeroU32,
-    ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error> {
+    ) -> Result<Option<(TargetHeight, BlockHeight)>, Self::Error> {
         Ok(None)
     }
 
--- a/zcash_client_backend/src/data_api/testing.rs
+++ b/zcash_client_backend/src/data_api/testing.rs
@@ -51,7 +51,7 @@ use super::{
     wallet::{
         create_proposed_transactions,
         input_selection::{GreedyInputSelector, InputSelector},
-        propose_standard_transfer_to_address, propose_transfer,
+        propose_standard_transfer_to_address, propose_transfer, ConfirmationsPolicy,
     },
     Account, AccountBalance, AccountBirthday, AccountMeta, AccountPurpose, AccountSource,
     AddressInfo, BlockMetadata, DecryptedTransaction, InputSource, NoteFilter, NullifierQuery,
@@ -922,7 +922,9 @@ where
             from_account.usk(),
             request,
             OvkPolicy::Sender,
-            NonZeroU32::MIN,
+            // TODO(schell): Check this passes tests.
+            // This was previously `NonZeroU32::MIN`, which is 1.
+            ConfirmationsPolicy::MIN,
         )
     }
 
@@ -935,7 +937,7 @@ where
         usk: &UnifiedSpendingKey,
         request: zip321::TransactionRequest,
         ovk_policy: OvkPolicy,
-        min_confirmations: NonZeroU32,
+        confirmations_policy: ConfirmationsPolicy,
     ) -> Result<NonEmpty<TxId>, super::wallet::TransferErrT<DbT, InputsT, ChangeT>>
     where
         InputsT: InputSelector<InputSource = DbT>,
@@ -957,7 +959,7 @@ where
             input_selector,
             change_strategy,
             request,
-            min_confirmations,
+            confirmations_policy,
         )?;
 
         create_proposed_transactions(
@@ -979,7 +981,7 @@ where
         input_selector: &InputsT,
         change_strategy: &ChangeT,
         request: zip321::TransactionRequest,
-        min_confirmations: NonZeroU32,
+        confirmations_policy: ConfirmationsPolicy,
     ) -> Result<
         Proposal<ChangeT::FeeRule, <DbT as InputSource>::NoteRef>,
         super::wallet::ProposeTransferErrT<DbT, Infallible, InputsT, ChangeT>,
@@ -996,7 +998,7 @@ where
             input_selector,
             change_strategy,
             request,
-            min_confirmations,
+            confirmations_policy,
         )
     }
 
@@ -1007,7 +1009,7 @@ where
         &mut self,
         spend_from_account: <DbT as InputSource>::AccountId,
         fee_rule: StandardFeeRule,
-        min_confirmations: NonZeroU32,
+        confirmations_policy: ConfirmationsPolicy,
         to: &Address,
         amount: Zatoshis,
         memo: Option<MemoBytes>,
@@ -1028,7 +1030,7 @@ where
             &network,
             fee_rule,
             spend_from_account,
-            min_confirmations,
+            confirmations_policy,
             to,
             amount,
             memo,
@@ -2500,6 +2502,7 @@ impl InputSource for MockWalletDb {
         _target_value: TargetValue,
         _sources: &[ShieldedProtocol],
         _anchor_height: BlockHeight,
+        _confirmations_policy: ConfirmationsPolicy,
         _exclude: &[Self::NoteRef],
     ) -> Result<SpendableNotes<Self::NoteRef>, Self::Error> {
         Ok(SpendableNotes::empty())
--- a/zcash_client_backend/src/data_api/testing/orchard.rs
+++ b/zcash_client_backend/src/data_api/testing/orchard.rs
@@ -15,7 +15,7 @@ use zcash_keys::{
 use zcash_note_encryption::try_output_recovery_with_ovk;
 use zcash_primitives::transaction::Transaction;
 use zcash_protocol::{
-    consensus::{self, BlockHeight},
+    consensus::{self, BlockHeight, TargetHeight},
     memo::MemoBytes,
     ShieldedProtocol,
 };
@@ -109,7 +109,7 @@ impl ShieldedPoolTester for OrchardPoolTester {
         st: &TestState<Cache, DbT, P>,
         account: <DbT as InputSource>::AccountId,
         target_value: TargetValue,
-        target_height: BlockHeight,
+        target_height: TargetHeight,
         min_confirmations: ConfirmationsPolicy,
         exclude: &[DbT::NoteRef],
     ) -> Result<Vec<ReceivedNote<DbT::NoteRef, Self::Note>>, <DbT as InputSource>::Error> {
--- a/zcash_client_backend/src/data_api/testing/orchard.rs
+++ b/zcash_client_backend/src/data_api/testing/orchard.rs
@@ -24,6 +24,7 @@ use crate::{
     data_api::{
         chain::{CommitmentTreeRoot, ScanSummary},
         testing::{pool::ShieldedPoolTester, TestState},
+        wallet::ConfirmationsPolicy,
         DecryptedTransaction, InputSource, TargetValue, WalletCommitmentTrees, WalletSummary,
         WalletTest,
     },
@@ -108,7 +109,8 @@ impl ShieldedPoolTester for OrchardPoolTester {
         st: &TestState<Cache, DbT, P>,
         account: <DbT as InputSource>::AccountId,
         target_value: TargetValue,
-        anchor_height: BlockHeight,
+        target_height: BlockHeight,
+        min_confirmations: ConfirmationsPolicy,
         exclude: &[DbT::NoteRef],
     ) -> Result<Vec<ReceivedNote<DbT::NoteRef, Self::Note>>, <DbT as InputSource>::Error> {
         st.wallet()
@@ -116,7 +118,8 @@ impl ShieldedPoolTester for OrchardPoolTester {
                 account,
                 target_value,
                 &[ShieldedProtocol::Orchard],
-                anchor_height,
+                target_height,
+                min_confirmations,
                 exclude,
             )
             .map(|n| n.take_orchard())
--- a/zcash_client_backend/src/data_api/testing/orchard.rs
+++ b/zcash_client_backend/src/data_api/testing/orchard.rs
@@ -110,7 +110,7 @@ impl ShieldedPoolTester for OrchardPoolTester {
         account: <DbT as InputSource>::AccountId,
         target_value: TargetValue,
         target_height: TargetHeight,
-        min_confirmations: ConfirmationsPolicy,
+        confirmation_policy: ConfirmationsPolicy,
         exclude: &[DbT::NoteRef],
     ) -> Result<Vec<ReceivedNote<DbT::NoteRef, Self::Note>>, <DbT as InputSource>::Error> {
         st.wallet()
@@ -119,7 +119,7 @@ impl ShieldedPoolTester for OrchardPoolTester {
                 target_value,
                 &[ShieldedProtocol::Orchard],
                 target_height,
-                min_confirmations,
+                confirmation_policy,
                 exclude,
             )
             .map(|n| n.take_orchard())
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -21,7 +21,7 @@ use zcash_primitives::{
     },
 };
 use zcash_protocol::{
-    consensus::{self, BlockHeight, NetworkUpgrade, Parameters},
+    consensus::{self, BlockHeight, NetworkUpgrade, Parameters, TargetHeight},
     local_consensus::LocalNetwork,
     memo::{Memo, MemoBytes},
     value::Zatoshis,
@@ -150,7 +150,7 @@ pub trait ShieldedPoolTester {
         st: &TestState<Cache, DbT, P>,
         account: <DbT as InputSource>::AccountId,
         target_value: TargetValue,
-        target_height: BlockHeight,
+        target_height: TargetHeight,
         confirmations_policy: ConfirmationsPolicy,
         exclude: &[DbT::NoteRef],
     ) -> Result<Vec<ReceivedNote<DbT::NoteRef, Self::Note>>, <DbT as InputSource>::Error>;
@@ -772,7 +772,7 @@ pub fn send_multi_step_proposed_transfer<T: ShieldedPoolTester, DSF>(
 
     let mut builder = Builder::new(
         *st.network(),
-        height + 1,
+        (height + 1).into(),
         BuildConfig::Standard {
             sapling_anchor: None,
             orchard_anchor: None,
@@ -1209,6 +1209,7 @@ pub fn spend_fails_on_unverified_notes<T: ShieldedPoolTester>(
     );
 }
 
+/// TODO(schell): write docs about what this test does
 pub fn spend_fails_on_locked_notes<T: ShieldedPoolTester>(
     ds_factory: impl DataStoreFactory,
     cache: impl TestCache,
@@ -1989,7 +1990,7 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
         // TODO(schell): this height needs to be changed as we're now taking the target_height (chain tip + 1)
         // instead of anchor height
-        received_tx_height + 10,
+        (received_tx_height + 10).into(),
         ConfirmationsPolicy::default(),
         &[],
     )
@@ -2005,7 +2006,9 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         &st,
         account_id,
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        received_tx_height + 10,
+        // TODO(schell): this height needs to be changed as we're now taking the target_height (chain tip + 1)
+        // instead of anchor height
+        (received_tx_height + 10).into(),
         ConfirmationsPolicy::default(),
         &[],
     )
@@ -2061,7 +2064,8 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, DSF: DataStoreFactory>(
         &st,
         account.id(),
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        account.birthday().height() + 5,
+        // TODO(schell): revisit this height, we should be passing chain tip + 1
+        (account.birthday().height() + 5).into(),
         ConfirmationsPolicy {
             // 5
             untrusted: NonZeroU32::MIN.saturating_add(4),
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -344,6 +344,98 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
     );
 }
 
+/// Tests that inputs from trusted sources (the same wallet) can be spent according to
+/// the `ConfirmationPolicy`.
+///
+/// The test:
+/// - Adds funds to the wallet in a single note.
+/// - Checks that the wallet balances are correct.
+/// - Constructs a request to spend part of that balance to an external address in the
+///   same pool.
+/// - Builds the transaction.
+pub fn can_spend_trusted_inputs_by_confirmations_policy<T: ShieldedPoolTester>(
+    dsf: impl DataStoreFactory,
+    cache: impl TestCache,
+) {
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(dsf)
+        .with_block_cache(cache)
+        .with_account_from_sapling_activation(BlockHash([0; 32]))
+        .build();
+
+    let account = st.test_account().cloned().unwrap();
+    let dfvk = T::test_account_fvk(&st);
+
+    // Add funds to the wallet in a single note
+    let value = Zatoshis::const_from_u64(60000);
+    let (h, _, _) = st.generate_next_block(&dfvk, AddressType::DefaultExternal, value);
+    st.scan_cached_blocks(h, 1);
+
+    // Spendable balance matches total balance
+    assert_eq!(st.get_total_balance(account.id()), value);
+    assert_eq!(st.get_spendable_balance(account.id(), 1), value);
+
+    assert_eq!(
+        st.wallet()
+            .block_max_scanned()
+            .unwrap()
+            .unwrap()
+            .block_height(),
+        h
+    );
+
+    let to_extsk = T::sk(&[0xf5; 32]);
+    let to: Address = T::sk_default_address(&to_extsk);
+    let payment_amount = Zatoshis::const_from_u64(10000);
+    let request = zip321::TransactionRequest::new(vec![Payment::without_memo(
+        to.to_zcash_address(st.network()),
+        payment_amount,
+    )])
+    .unwrap();
+
+    let fee_rule = StandardFeeRule::Zip317;
+
+    let change_memo = "Test change memo".parse::<Memo>().unwrap();
+    let change_strategy = standard::SingleOutputChangeStrategy::new(
+        fee_rule,
+        Some(change_memo.clone().into()),
+        T::SHIELDED_PROTOCOL,
+        DustOutputPolicy::default(),
+    );
+    let input_selector = GreedyInputSelector::new();
+
+    let proposal = st
+        .propose_transfer(
+            account.id(),
+            &input_selector,
+            &change_strategy,
+            request,
+            ConfirmationsPolicy::MIN,
+        )
+        .unwrap();
+
+    let create_proposed_result = st.create_proposed_transactions::<Infallible, _, Infallible, _>(
+        account.usk(),
+        OvkPolicy::Sender,
+        &proposal,
+    );
+    assert_matches!(&create_proposed_result, Ok(txids) if txids.len() == 1);
+
+    let sent_tx_id = create_proposed_result.unwrap()[0];
+
+    // Ensure that when the transaction is mined, we cannot spend the change, yet.
+    let (h, _) = st.generate_next_block_including(sent_tx_id);
+    st.scan_cached_blocks(h, 1);
+
+    let fee = Zatoshis::from_u64(1000).unwrap();
+    assert_eq!(
+        st.get_total_balance(account.id()),
+        (value - payment_amount - fee).unwrap()
+    );
+    // Spendable balance matches total balance
+    assert_eq!(st.get_spendable_balance(account.id(), 1), value);
+}
+
 pub fn send_with_multiple_change_outputs<T: ShieldedPoolTester>(
     dsf: impl DataStoreFactory,
     cache: impl TestCache,
@@ -1209,7 +1301,6 @@ pub fn spend_fails_on_unverified_notes<T: ShieldedPoolTester>(
     );
 }
 
-/// TODO(schell): write docs about what this test does
 pub fn spend_fails_on_locked_notes<T: ShieldedPoolTester>(
     ds_factory: impl DataStoreFactory,
     cache: impl TestCache,
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -2078,8 +2078,6 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         &st,
         account_id,
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        // TODO(schell): this height needs to be changed as we're now taking the target_height (chain tip + 1)
-        // instead of anchor height
         (received_tx_height + 10).into(),
         ConfirmationsPolicy::default(),
         &[],
@@ -2096,8 +2094,6 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         &st,
         account_id,
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        // TODO(schell): this height needs to be changed as we're now taking the target_height (chain tip + 1)
-        // instead of anchor height
         (received_tx_height + 10).into(),
         ConfirmationsPolicy::default(),
         &[],
@@ -2154,7 +2150,6 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, DSF: DataStoreFactory>(
         &st,
         account.id(),
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        // TODO(schell): revisit this height, we should be passing chain tip + 1
         (account.birthday().height() + 5).into(),
         ConfirmationsPolicy {
             // 5
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -344,96 +344,95 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
     );
 }
 
-/// Tests that inputs from trusted sources (the same wallet) can be spent according to
-/// the `ConfirmationPolicy`.
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct ConfirmationStep {
+    i: u32,
+    confirmation_requirement: u32,
+    number_of_confirmations: u32,
+    pending_balance: Zatoshis,
+    spendable_balance: Zatoshis,
+    total_balance: Zatoshis,
+}
+
+/// Tests that inputs from a source can be spent according to the default
+/// `ConfirmationPolicy`.
 ///
 /// The test:
-/// - Adds funds to the wallet in a single note.
-/// - Checks that the wallet balances are correct.
-/// - Constructs a request to spend part of that balance to an external address in the
-///   same pool.
-/// - Builds the transaction.
-pub fn can_spend_trusted_inputs_by_confirmations_policy<T: ShieldedPoolTester>(
+/// - Adds funds to the wallet in a single note from an certain source.
+/// - Checks that the wallet balances are correct after N confirmations, according to
+///   the policy.
+pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
     dsf: impl DataStoreFactory,
     cache: impl TestCache,
+    input_is_trusted: bool,
 ) {
     let mut st = TestBuilder::new()
         .with_data_store_factory(dsf)
         .with_block_cache(cache)
         .with_account_from_sapling_activation(BlockHash([0; 32]))
         .build();
-
     let account = st.test_account().cloned().unwrap();
     let dfvk = T::test_account_fvk(&st);
+    let starting_balance = Zatoshis::const_from_u64(60_000);
+    // Add funds to the wallet in a single note, owned by the internal spending key,
+    // this is the first confirmation
+    let address_type = if input_is_trusted {
+        AddressType::Internal
+    } else {
+        AddressType::DefaultExternal
+    };
 
-    // Add funds to the wallet in a single note
-    let value = Zatoshis::const_from_u64(60000);
-    let (h, _, _) = st.generate_next_block(&dfvk, AddressType::DefaultExternal, value);
+    let (h, _, _) = st.generate_next_block(&dfvk, address_type, starting_balance);
     st.scan_cached_blocks(h, 1);
 
-    // Spendable balance matches total balance
-    assert_eq!(st.get_total_balance(account.id()), value);
-    assert_eq!(st.get_spendable_balance(account.id(), 1), value);
+    // Spendable balance matches total balance at 1 confirmation.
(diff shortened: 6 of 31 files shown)


components/zcash_protocol/src/consensus.rs
```diff
@@ -126,6 +126,59 @@ impl Sub<BlockHeight> for BlockHeight {
     }
 }
 
+/// A wrapper type around [`BlockHeight`] that represents the _next_ chain tip.
+///
+/// Addition and subtraction are provided by proxying to [`BlockHeight`].
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TargetHeight(BlockHeight);
```
nuttycom: This addition needs to be noted in the CHANGELOG

zcash_client_backend/src/data_api.rs
```diff
@@ -1294,7 +1294,8 @@ pub trait InputSource {
         account: Self::AccountId,
         target_value: TargetValue,
         sources: &[ShieldedProtocol],
-        anchor_height: BlockHeight,
+        target_height: zcash_protocol::consensus::TargetHeight,
```
nuttycom: Note this change in the CHANGELOG, and do the same for other conversions to use `TargetHeight`.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -2499,7 +2499,8 @@ impl InputSource for MockWalletDb {
         _account: Self::AccountId,
         _target_value: TargetValue,
         _sources: &[ShieldedProtocol],
-        _anchor_height: BlockHeight,
+        _anchor_height: TargetHeight,
```
nuttycom: nit: this should now be named `_target_height`

zcash_client_backend/src/data_api.rs
```suggestion
        target_height: TargetHeight,
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1338,7 +1339,7 @@ pub trait InputSource {
     fn get_spendable_transparent_outputs(
         &self,
         _address: &TransparentAddress,
-        _target_height: BlockHeight,
+        _target_height: zcash_protocol::consensus::TargetHeight,
```
```suggestion
        _target_height: TargetHeight,
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1506,7 +1507,7 @@ pub trait WalletRead {
     fn get_target_and_anchor_heights(
         &self,
         min_confirmations: NonZeroU32,
-    ) -> Result<Option<(BlockHeight, BlockHeight)>, Self::Error>;
+    ) -> Result<Option<(zcash_protocol::consensus::TargetHeight, BlockHeight)>, Self::Error>;
```
```suggestion
    ) -> Result<Option<(TargetHeight, BlockHeight)>, Self::Error>;
```

zcash_client_backend/src/data_api/testing/orchard.rs
```diff
@@ -108,15 +109,17 @@ impl ShieldedPoolTester for OrchardPoolTester {
         st: &TestState<Cache, DbT, P>,
         account: <DbT as InputSource>::AccountId,
         target_value: TargetValue,
-        anchor_height: BlockHeight,
+        target_height: TargetHeight,
+        confirmation_policy: ConfirmationsPolicy,
```
daira:
```suggestion
        confirmations_policy: ConfirmationsPolicy,
```

zcash_client_backend/src/data_api/testing/orchard.rs
```diff
@@ -108,15 +109,17 @@ impl ShieldedPoolTester for OrchardPoolTester {
...
         exclude: &[DbT::NoteRef],
     ) -> Result<Vec<ReceivedNote<DbT::NoteRef, Self::Note>>, <DbT as InputSource>::Error> {
         st.wallet()
             .select_spendable_notes(
                 account,
                 target_value,
                 &[ShieldedProtocol::Orchard],
-                anchor_height,
+                target_height,
+                confirmation_policy,
```
daira:
```suggestion
                confirmations_policy,
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +344,97 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    i: u32,
+    confirmation_requirement: u32,
+    number_of_confirmations: u32,
+    pending_balance: Zatoshis,
+    spendable_balance: Zatoshis,
+    total_balance: Zatoshis,
+}
+
+/// Tests that inputs from a source can be spent according to the default
+/// `ConfirmationPolicy`.
```
```suggestion
/// `ConfirmationsPolicy`.
```

components/zcash_protocol/src/consensus.rs
```diff
@@ -126,6 +126,59 @@ impl Sub<BlockHeight> for BlockHeight {
...
+
+impl From<TargetHeight> for u32 {
+    fn from(value: TargetHeight) -> Self {
+        value.0 .0
+    }
+}
+
+impl From<u32> for TargetHeight {
+    fn from(value: u32) -> Self {
+        BlockHeight::from_u32(value).into()
```
daira:
```suggestion
        TargetHeight::from(BlockHeight::from_u32(value))
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -770,7 +863,7 @@ pub fn send_multi_step_proposed_transfer<T: ShieldedPoolTester, DSF>(
 
     let mut builder = Builder::new(
         *st.network(),
-        height + 1,
+        (height + 1).into(),
```
```suggestion
        TargetHeight::from(height + 1),
```
I prefer `T::from(...)` in general: it makes clearer why the conversion is needed, it avoids having to guess the type or rely on an IDE (often not available in code review), and it is less fragile if new conversions are added.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1985,7 +2078,8 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
         &st,
         account_id,
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        received_tx_height + 10,
+        (received_tx_height + 10).into(),
```
```suggestion
        TargetHeight::from(received_tx_height + 10),
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -2000,7 +2094,8 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
...
```
```suggestion
        TargetHeight::from(received_tx_height + 10),
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -2049,12 +2144,18 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, DSF: DataStoreFactory>(
...
     st.scan_cached_blocks(account.birthday().height() + 10, 1);
 
+    println!("account_birthday_height: {}", account.birthday().height());
     // Verify that our note is considered spendable
     let spendable = T::select_spendable_notes(
         &st,
         account.id(),
         TargetValue::AtLeast(Zatoshis::const_from_u64(300000)),
-        account.birthday().height() + 5,
+        (account.birthday().height() + 5).into(),
```
```suggestion
        TargetHeight::from(account.birthday().height() + 5),
```

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -63,7 +63,7 @@ fn check_balance<DSF>(
     );
     assert_eq!(
         st.wallet()
-            .get_spendable_transparent_outputs(taddr, mempool_height, min_confirmations)
+            .get_spendable_transparent_outputs(taddr, mempool_height.into(), min_confirmations)
```
daira: Similarly for these and in other files below.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -228,17 +235,19 @@ where
             "WITH eligible AS (
                  SELECT
                      {table_prefix}_received_notes.id AS id, txid, {output_index_col},
-                     diversifier, value, {note_reconstruction_cols}, commitment_tree_position,
+                     diversifier, value, {note_reconstruction_cols}, commitment_tree_position, 
                      SUM(value) OVER (ROWS UNBOUNDED PRECEDING) AS so_far,
-                     accounts.ufvk as ufvk, recipient_key_scope
+                     accounts.ufvk as ufvk, recipient_key_scope,
+                     transactions.block AS mined_height
```
nuttycom: Using `transactions.block` here is correct because we need to have scanned the block (we don't know about it just from having observed a transparent UTXO, for example.)

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -251,7 +260,7 @@ where
                    JOIN transactions stx ON stx.id_tx = transaction_id
                    WHERE stx.block IS NOT NULL -- the spending tx is mined
                    OR stx.expiry_height IS NULL -- the spending tx will not expire
-                   OR stx.expiry_height > :anchor_height -- the spending tx is unexpired
+                   OR stx.expiry_height >= :target_height -- the spending tx is unexpired
```
nuttycom: This is correct; the transaction is unexpired if its expiry is the target height, because it can still be mined at the target height.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -291,17 +300,38 @@ where
...
+            ":target_height": &u32::from(target_height),
             ":target_value": &u64::from(target_value),
             ":exclude": &excluded_ptr,
             ":wallet_birthday": u32::from(birthday_height)
         ],
-        |r| to_spendable_note(params, r),
+        |r| -> Result<_, SqliteClientError> {
+            let maybe_note = to_spendable_note(params, r)?;
+            let recipient_key_scope: i64 = r.get("recipient_key_scope")?;
+            let note_is_trusted = recipient_key_scope == 1;
```
nuttycom:
```suggestion
            // for now, we only treat wallet-internal outputs as trusted
            let note_is_trusted = recipient_key_scope == KeyScope::INTERNAL.encode();
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +344,97 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+    } else {
+        AddressType::DefaultExternal
+    };
+
+    let (h, _, _) = st.generate_next_block(&dfvk, address_type, starting_balance);
+    st.scan_cached_blocks(h, 1);
+
+    // Spendable balance matches total balance at 1 confirmation.
+    assert_eq!(st.get_total_balance(account.id()), starting_balance);
+    assert_eq!(st.get_spendable_balance(account.id(), 1), starting_balance);
```
nuttycom: blocking: I now am remembering that `WalletRead::get_spendable_balance` also needs to take the confirmations policy into account. Otherwise, the wallet app will not be able to use the same rules for selection of notes as it does for displaying usable balance, which will lead to bugs and/or user confusion.

nuttycom (changes requested):
One blocking issue that I belatedly realized: WalletRead::get_spendable_balance also needs to take the confirmations policy into account.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1883,7 +1880,8 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
     let recover_until_height = recover_until_height(tx)?;
 
     let fully_scanned_height = block_fully_scanned(tx, params)?.map(|m| m.block_height());
-    let summary_height = (chain_tip_height + 1).saturating_sub(std::cmp::max(min_confirmations, 1));
+    let summary_height =
+        (chain_tip_height + 1).saturating_sub(u32::from(confirmations_policy.trusted));
```
nuttycom: It's not sufficient merely to change the summary height; the balance is split between confirmed and unconfirmed value, and so the confirmations policy values need to be taken into account when adding up the value of the wallet's notes.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -793,8 +794,11 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
     params: &P,
     address: &TransparentAddress,
     target_height: consensus::TargetHeight,
-    min_confirmations: u32,
+    confirmations_policy: ConfirmationsPolicy,
 ) -> Result<Vec<WalletTransparentOutput>, SqliteClientError> {
+    // TODO(schell): not sure if this is correct
+    let min_confirmations = u32::from(confirmations_policy.untrusted)
+        .saturating_sub(u32::from(confirmations_policy.trusted));
```
nuttycom:
```suggestion
    // At present, wallets based on `zcash_client_sqlite` do not produce transparent
    // change, and so all transparent UTXOs received by the wallet should be treated
    // as untrusted.
    // 
    // The outputs of a shielding transaction will be treated by the wallet as trusted, 
    // so if we shield the transparent UTXOs with zero confirmations, we need to wait 
    // a total of `untrusted` blocks from the block in which the UTXOs were received 
    // for the outputs of the shielding transaction to be spendable. The wallet *must*
    // wait at for at least `trusted` confirmations of the shielding outputs in before
    // attempting to spend them (for indistinguishability and to avoid anchor 
    // invalidation) and so we allow shielding of UTXOs having `untrusted - trusted`
    // confirmations. 
    let min_confirmations = u32::from(confirmations_policy.untrusted)
        .saturating_sub(u32::from(confirmations_policy.trusted));
```

nuttycom (changes requested):
One additional blocking issue; also, tests are failing.

schell:
I think somehow we missed two commits from last week. I'm not sure if it got clobbered or what, but I've cherry picked them and then fixed a typo. There are two tests left to fix after my changes, which I'll be working on today.

nuttycom:
> I think somehow we missed two commits from last week. I'm not sure if it got clobbered or what, but I've cherry picked them and then fixed a typo. There are two tests left to fix after my changes, which I'll be working on today.

Hmm, sorry about that. I did a little looking at this on Friday, and something that I realized is that we probably need to separately represent "un-economic" value in account totals: https://github.com/zcash/librustzcash/compare/feat/zip-315-confirmations-policy...nuttycom:librustzcash:feat/zip-315-confirmations-policy; I'll rebase this atop your current work.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2001,12 +2002,14 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
                JOIN transactions t ON t.id_tx = transaction_id
                WHERE t.block IS NOT NULL -- the spending transaction is mined
                OR t.expiry_height IS NULL -- the spending tx will not expire
+               -- TODO(schell): a bug in select_spendable_notes was fixed by changing this to
+               -- `>= :summary_height`. Should that also happen here?
                OR t.expiry_height > :summary_height -- the spending tx is unexpired
```
nuttycom: There's a separate bug here; this should be `:target_height`, not `:summary_height`. The expiry height is the last block in which the transaction can be mined, and so if the expiry height is equal to the target height, then the spending tx is unexpired, so yes this should be `>=`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2034,14 +2037,27 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
                 },
             )?;
 
+            // TODO(schell): ask under what circumstances would `received_height` be None?
+            // It seems that from the query above, `block` will always be present
```
nuttycom: Not true; the `block IS NOT NULL` condition is just one branch of an `OR`

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2034,14 +2037,27 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
             let received_height = row.get::<_, Option<u32>>(4)?.map(BlockHeight::from);
+            let recipient_key_scope = row.get::<_, i64>(5)?;
+            // for now, we only treat wallet-internal outputs as trusted
+            let note_is_trusted = recipient_key_scope == KeyScope::INTERNAL.encode();
+            let required_confirmations = u32::from(confirmations_policy.untrusted);
+            // If the note is from a trusted source (this wallet), then we don't have to determine the
+            // number of confirmations, as we've already queried above based on the trusted summary height.
+            // So we only check if it's trusted, or if it meets the untrusted number of confirmations.
```
nuttycom: Since we should be querying based on the target height and then filtering into different categories, this isn't true.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -291,17 +301,39 @@ where
...
-        .collect::<Result<_, _>>()
+        .filter_map(|result_maybe_note| {
+            let result_note = result_maybe_note.transpose()?;
+            result_note
+                .map(|(note, is_trusted, mined_height)| {
+                    let number_of_confirmations = target_height - mined_height;
+                    let required_confirmations = u32::from(confirmations_policy.untrusted);
+                    // If the note is from a trusted source (this wallet), then we don't have to determine the
+                    // number of confirmations, as we've already queried above based on the trusted anchor height.
+                    // So we only check if it's trusted, or if it meets the untrusted number of confirmations.
```
nuttycom: Here, this is correct; we're not querying unconditionally for all notes.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -355,7 +382,13 @@ where
...
         .unwrap();
 
-    check_balance::<DSF>(&st, &account, taddr, 2, &not_confirmed_yet_value);
+    check_balance::<DSF>(
+        &st,
+        &account,
+        taddr,
+        ConfirmationsPolicy::new_symmetrical(2).unwrap(),
+        &not_confirmed_yet_value,
+    );
```
nuttycom, line 403: After https://github.com/zcash/librustzcash/compare/feat/zip-315-confirmations-policy...nuttycom:librustzcash:feat/zip-315-confirmations-policy, this is the one remaining test failure and I'm wondering whether the test is actually wrong; with transparent funds, we should always allow 0-conf shielding, so the `spendable_value` field of the transparent balance is really "shieldable value".

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -945,6 +949,9 @@ pub(crate) fn add_transparent_account_balances(
          )
          GROUP BY a.uuid",
     )?;
+    // TODO(schell): figure out if this in any way a correct treatment for min_confirmations
```
nuttycom: I've been thinking about this and I don't think it's quite right, we can revisit in the onboarding pairing.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/28de05e9bb0963e5bffca97bf26656c67c41a7b2..a6963a504f3cd5e7e084ef0ed987b6313df2bbab) for additional cleanup and to get the CHANGELOGs right. The one significant semantic change here is that I decided that the `TargetHeight` type didn't really belong in `zcash_protocol` and that it wasn't a good idea to require a complete semver-breaking release of `zcash_primitives` to support the rest of these changes.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1511,11 +1701,34 @@ pub fn change_note_spends_succeed<T: ShieldedPoolTester>(
...
-    assert_eq!(st.get_pending_shielded_balance(account_id, 10), value);
-    assert_eq!(st.get_spendable_balance(account_id, 10), Zatoshis::ZERO);
+    assert_eq!(
+        st.get_pending_shielded_balance(
+            account_id,
+            ConfirmationsPolicy::new_symmetrical_unchecked(
+                10,
+                #[cfg(feature = "transparent-inputs")]
+                false
+            )
```
nuttycom: Should these be using `ConfirmationsPolicy::default()` instead?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1230,12 +1404,19 @@ pub fn spend_fails_on_locked_notes<T: ShieldedPoolTester>(
...
 
     // Send some of the funds to another address, but don't mine the tx.
     let extsk2 = T::sk(&[0xf5; 32]);
     let to = T::sk_default_address(&extsk2);
-    let min_confirmations = NonZeroU32::new(1).unwrap();
+    let min_confirmations = ConfirmationsPolicy::new_symmetrical_unchecked(
+        1,
+        #[cfg(feature = "transparent-inputs")]
+        false,
+    );
```
nuttycom: This could be `ConfirmationsPolicy::MIN`

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -291,17 +302,39 @@ where
...
-            ":anchor_height": &u32::from(anchor_height),
+            ":anchor_height": &u32::from(trusted_anchor_height),
+            ":target_height": &u32::from(target_height),
             ":target_value": &u64::from(target_value),
             ":exclude": &excluded_ptr,
             ":wallet_birthday": u32::from(birthday_height)
         ],
-        |r| to_spendable_note(params, r),
+        |r| -> Result<_, SqliteClientError> {
+            let maybe_note = to_spendable_note(params, r)?;
```
nuttycom: `ReceivedNote` already has a `spending_key_scope` field, which likely should be populated by `to_spendable_note`. It might also be useful to add `mined_height` there as an optional property (optional because notes detected in the mempool can't have a mined height.)

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1793,7 +1794,11 @@ pub(crate) mod tests {
                 &input_selector,
                 &change_strategy,
                 request,
-                NonZeroU32::new(10).unwrap(),
+                ConfirmationsPolicy::new_symmetrical_unchecked(
```
nuttycom: Check whether this can use `ConfirmationsPolicy::default()`

nuttycom (dismissed):
utACK with a few minor suggestions.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -359,18 +361,114 @@ impl ConfirmationsPolicy {
...
+    /// provided values.
+    ///
+    /// The number of confirmations required for trusted notes must be less than or equal to the
+    /// number of confirmations required for untrusted notes; this returns `Err(())` if this
+    /// invariant is violated.
+    pub fn new(
+        trusted: NonZeroU32,
+        untrusted: NonZeroU32,
+        #[cfg(feature = "transparent-inputs")] allow_zero_conf_shielding: bool,
+    ) -> Result<Self, ()> {
```
schell, line 410: QQ - why return `Result<_, ()>` here instead of `Option`? Is it for `map_err` somewhere down the line?

nuttycom: This is very minor, but I tend to prefer `Result<_, ()>` to signal that there's no value due to an invariant being violated and `Option<_>` to signal that there's no value because no value should be expected given the inputs. It's a bit subtle, I guess.

nuttycom (approved):
re-utACK https://github.com/zcash/librustzcash/pull/1888/commits/d55c08a76414f1f98f313d8a1d63b4b0dc6f0f00

zcash_client_sqlite/src/testing/db.rs
```diff
@@ -12,11 +12,13 @@ use tempfile::NamedTempFile;
 use rusqlite::{self};
 use secrecy::SecretVec;
 use shardtree::{error::ShardTreeError, ShardTree};
+
 use zcash_client_backend::{
     data_api::{
         chain::{ChainState, CommitmentTreeRoot},
         scanning::ScanRange,
         testing::{DataStoreFactory, Reset, TestState},
+        wallet::{ConfirmationsPolicy, TargetHeight},
```
str4d, line 21: These imports are unused in this file's diff. I'm surprised there isn't a lint failure.

Cargo.toml
```diff
@@ -18,6 +18,8 @@ members = [
     "zcash_transparent",
 ]
 
+resolver = "2"
```
str4d, line 21: This change is not explained or documented anywhere AFAICT. What about this PR required it to be made?

zcash_client_backend/src/scan.rs
```diff
@@ -255,6 +255,7 @@ impl<Item: Task + DynamicUsage> Tasks<Item> for WithUsage {
 
 /// A task that will clean up its own heap usage from the overall running usage once it is
 /// complete.
+#[allow(dead_code)]
```
str4d, line 258: What in this PR caused `WithUsageTask` to become dead? AFAICT it is not dead in `main`.

zcash_client_backend/CHANGELOG.md
```diff
@@ -38,9 +42,53 @@ workspace.
     as its argument instead of its parts. This minimizes the API complexity
     that would otherwise arise due to the presence of the `zcashd-compat`
     feature flag.
+  - Arguments to `InputSource::select_spendable_notes` have changed; it now takes
```
str4d: "Arguments to FOO have changed" is verbose and duplicative of this being a bullet point in the "Changed" section, and it makes spotting what the change affects harder (because the line does not start with the type that is being changed).

zcash_client_backend/CHANGELOG.md
```diff
@@ -38,9 +42,53 @@ workspace.
...
+    It now returns a `TargetHeight` instead of a `BlockHeight` for the target
+    height element of its result tuple.
+  - Arguments to `WalletRead::get_wallet_summary` have changed; it now takes a
+    `ConfirmationsPolicy` instead of a `u32` `min_confirmations` argument.
+  - Arguments to `WalletRead::get_transparent_balances` have changed. The balance
+    is now retrieved in terms of a target height and confirmations policy.
+  - The semantics of `zcash_client_backend::data_api::Balance::total` and
+    `zcash_client_backend::data_api::AccountBalance::total` have changed;
+    these totals no longer include the value of uneconomic ("dust") notes
+    and transparent UTXOs.
```
```suggestion
  - The following methods now take `TargetHeight` and `ConfirmationsPolicy`
    arguments instead of an anchor height:
    - `InputSource::select_spendable_notes`
    - `wallet::input_selection::InputSelector::propose_transaction`
    The signatures of `wallet::propose_transfer` and
    `wallet::propose_standard_transfer_to_address` have also been modified to
    take these changes into account, as have the signatures of the
    `test-dependencies`-flagged methods
    `testing::TestState::{spend, propose_transfer, propose_standard_transfer}`.
  - The following methods now use `TargetHeight` instead of `BlockHeight` in
    their arguments and return values:
    - `InputSource::get_spendable_transparent_outputs`
    - `WalletRead::get_target_and_anchor_heights`
    - `wallet::input_selection::ShieldingSelector::propose_shielding`
    Related `test-dependencies`-flagged methods on `testing::TestState` have
    also been changed accordingly.
  - The following methods now take a `ConfirmationsPolicy` instead of a `u32`
    `min_confirmations` argument:
    - `InputSource::get_spendable_transparent_outputs`
    - `WalletRead::get_wallet_summary`
    - `wallet::propose_shielding`
    - `wallet::shield_transparent_funds`
    - `wallet::input_selection::ShieldingSelector::propose_shielding`
  - `WalletRead::get_transparent_balances` now retrieves the balance in terms of
    a target height and confirmations policy.
  - The semantics of `Balance::total` and `AccountBalance::total` have changed;
    these totals no longer include value that is uneconomic to spend ("dust").
```

zcash_client_backend/src/data_api.rs
```diff
@@ -223,6 +226,18 @@ impl Balance {
...
 
+    /// Returns the value in the account of notes that have value less than the marginal
+    /// fee, and consequently cannot be spent except as a grace input.
+    pub fn uneconomic_value(&self) -> Zatoshis {
+        self.uneconomic_value
+    }
+
+    /// Adds the specified value to the uneconomic value total, checking for overflow.
+    pub fn add_uneconomic_value(&mut self, value: Zatoshis) -> Result<(), BalanceError> {
+        self.uneconomic_value = (self.uneconomic_value + value).unwrap();
```
str4d: This `unwrap` is incorrect. It was fine in the existing methods because the overflow check was performed via `self.check_total_adding(value)?`, so the actual addition was duplicative. But here, this is the only overflow check.
```suggestion
        self.uneconomic_value = (self.uneconomic_value + value)?;
```

zcash_client_backend/CHANGELOG.md
str4d: This sentence is misleading. I originally interpreted it as excluding:
- uneconomic ("dust") notes
- transparent UTXOs

when it actually should be interpreted as excluding:
- uneconomic ("dust"):
  - shielded notes
  - transparent UTXOs

(The lack of "shielded" prefixing notes, but the presence of "transparent" prefixing UTXOs, is part of what confused my parser.)

Change this to either:
```suggestion
  - The semantics of `Balance::total` and `AccountBalance::total` have changed;
    these totals no longer include the value of uneconomic ("dust") notes and coins.
```
(as notes are inherently shielded and coins are inherently transparent in our standard lexicon), or:
```suggestion
  - The semantics of `Balance::total` and `AccountBalance::total` have changed;
    these totals no longer include value that is uneconomic to spend ("dust").
```

zcash_client_backend/src/data_api.rs
```diff
@@ -302,6 +312,14 @@ impl AccountBalance {
     }
 
     /// Returns the [`Balance`] of unshielded funds in the account.
+    ///
+    /// Note that because transparent UTXOs may be shielded with zero confirmations and this crate
+    /// does not provide capabilities to directly spend transparent UTXOs in non-shielding
+    /// transactions, the [`change_pending_confirmation`] and [`value_pending_spendability`] fields
+    /// of the returned [`Balance`] will always be zero.
```
str4d, line 319: Note to future-us: This comment may become incorrect once we start making the changes to support spend-from-transparent in Zallet (depending on how we introduce the changes).

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -176,8 +180,9 @@ pub trait InputSelector {
         &self,
         params: &ParamsT,
         wallet_db: &Self::InputSource,
-        target_height: BlockHeight,
+        target_height: TargetHeight,
         anchor_height: BlockHeight,
+        confirmations_policy: ConfirmationsPolicy,
```
str4d, line 185: Do we need all three of these in `InputSelector::propose_transaction`? Or is this a side-effect of doing the `anchor_height` calculation at a higher layer so we want to access it here?

zcash_client_backend/src/proposal.rs
```diff
@@ -183,7 +184,7 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
     /// * `steps`: A vector of steps that make up the proposal.
     pub fn multi_step(
         fee_rule: FeeRuleT,
-        min_target_height: BlockHeight,
+        min_target_height: TargetHeight,
```
str4d, line 187: Document this change in the changelog.

zcash_client_backend/src/proposal.rs
```diff
@@ -273,7 +274,7 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
         shielded_inputs: Option<ShieldedInputs<NoteRef>>,
         balance: TransactionBalance,
         fee_rule: FeeRuleT,
-        min_target_height: BlockHeight,
+        min_target_height: TargetHeight,
```
str4d, line 277: Document this change in the changelog.

zcash_client_backend/src/proposal.rs
```diff
@@ -301,7 +302,7 @@ impl<FeeRuleT, NoteRef> Proposal<FeeRuleT, NoteRef> {
     ///
     /// The chain must contain at least this many blocks in order for the proposal to
     /// be executed.
-    pub fn min_target_height(&self) -> BlockHeight {
+    pub fn min_target_height(&self) -> TargetHeight {
```
str4d, line 305: Document this change in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -393,6 +394,7 @@ impl<NoteRef, NoteT> ReceivedNote<NoteRef, NoteT> {
         note: NoteT,
         spending_key_scope: Scope,
         note_commitment_tree_position: Position,
+        mined_height: Option<BlockHeight>,
```
str4d, line 400: Document this change in the changelog.

zcash_client_backend/src/wallet.rs
```diff
@@ -422,6 +425,9 @@ impl<NoteRef, NoteT> ReceivedNote<NoteRef, NoteT> {
     pub fn note_commitment_tree_position(&self) -> Position {
         self.note_commitment_tree_position
     }
+    pub fn mined_height(&self) -> Option<BlockHeight> {
```
str4d, line 442: Document this addition in the changelog.

zcash_client_backend/src/data_api/wallet.rs
schell: All good! Thanks for the explanation :)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2499,37 +2502,46 @@ pub(crate) fn chain_tip_height(
...
+
+    #[cfg(not(feature = "orchard"))]
+    let orchard_anchor_height: Option<BlockHeight> = None;
+
+    Ok(sapling_anchor_height
+        .zip(orchard_anchor_height)
+        .map(|(s, o)| std::cmp::min(s, o))
+        .or(sapling_anchor_height)
+        .or(orchard_anchor_height))
+}
```
schell, line 2555: I actually had this change in one of my commits for a moment, so I appreciate this :)

schell (commented):
Still looks good to me 🧂 🧂 but of course I'm the noob, so grain of salt.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -632,4 +641,9 @@ pub(crate) mod tests {
     fn wallet_recovery_compute_fees() {
         testing::pool::wallet_recovery_computes_fees::<SaplingPoolTester>();
     }
+
+    #[test]
+    fn zip315_can_spend_inputs_by_confirmations_policy() {
+        testing::pool::can_spend_inputs_by_confirmations_policy::<SaplingPoolTester>();
+    }
```
str4d, line 653: This test (using `OrchardPoolTester`) was not added to `zcash_client_sqlite/src/wallet/orchard.rs`.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -255,6 +281,197 @@ pub type ExtractErrT<DbT, N> = Error<
...
+
+    fn sub(self, rhs: I) -> Self::Output {
+        self.0 - rhs
+    }
+}
+
+/// The minimum number of confirmations required for trusted and untrusted
+/// transactions.
+#[derive(Clone, Copy, Debug)]
+pub struct ConfirmationsPolicy {
```
str4d: This needs more documentation, in particular references to ZIP 315 (both to explain what "trusted" and "untrusted" mean, and as a source of the default values, which also need to be documented).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -255,6 +281,197 @@ pub type ExtractErrT<DbT, N> = Error<
...
+            // 10
+            untrusted: NonZeroU32::MIN.saturating_add(9),
+            #[cfg(feature = "transparent-inputs")]
+            allow_zero_conf_shielding: true,
+        }
+    }
+}
+
+impl ConfirmationsPolicy {
+    pub const MIN: Self = ConfirmationsPolicy {
```
str4d, line 390: Document this.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -255,6 +281,197 @@ pub type ExtractErrT<DbT, N> = Error<
...
+        if trusted > untrusted {
+            panic!("trusted must be <= untrusted")
+        }
+
+        Self {
+            trusted: NonZeroU32::new(trusted).expect("trusted must be nonzero"),
+            untrusted: NonZeroU32::new(untrusted).expect("untrusted must be nonzero"),
+            #[cfg(feature = "transparent-inputs")]
+            allow_zero_conf_shielding,
+        }
```
str4d:
```suggestion
        Self::new(
            NonZeroU32::new(trusted).expect("trusted must be nonzero"),
            NonZeroU32::new(untrusted).expect("untrusted must be nonzero"),
            #[cfg(feature = "transparent-inputs")]
            allow_zero_conf_shielding,
        ).expect("trusted must be <= untrusted")
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -255,6 +281,197 @@ pub type ExtractErrT<DbT, N> = Error<
...
+        #[cfg(feature = "transparent-inputs")] allow_zero_conf_shielding: bool,
+    ) -> Self {
+        let confirmations =
+            NonZeroU32::new(min_confirmations).expect("min_confirmations must be nonzero");
+        Self {
+            trusted: confirmations,
+            untrusted: confirmations,
+            #[cfg(feature = "transparent-inputs")]
+            allow_zero_conf_shielding,
+        }
```
str4d:
```suggestion
        Self::new_symmetrical(
            NonZeroU32::new(min_confirmations).expect("min_confirmations must be nonzero")
            #[cfg(feature = "transparent-inputs")]
            allow_zero_conf_shielding,
        )
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -342,6 +347,101 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
...
+
+    // Spendable balance matches total balance at 1 confirmation.
+    assert_eq!(st.get_total_balance(account.id()), starting_balance);
+    assert_eq!(
+        st.get_spendable_balance(account.id(), ConfirmationsPolicy::MIN),
+        starting_balance
+    );
+
+    // Generate N confirmations by mining blocks
+    let confirmations_policy = ConfirmationsPolicy::default();
```
str4d, line 399: The test currently only checks one side of the boundary. Prior to generating `N` confirmations, check that the spendable balance is zero with the default `ConfirmationsPolicy`.

I note that `steps` below collects the necessary information, but it is not checked in any way until `min_confirmations` is reached.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -265,16 +262,6 @@ where
...
-                 AND NOT EXISTS (
-                   SELECT 1 FROM v_{table_prefix}_shard_unscanned_ranges unscanned
-                   -- select all the unscanned ranges involving the shard containing this note
-                   WHERE rn.commitment_tree_position >= unscanned.start_position
-                   AND rn.commitment_tree_position < unscanned.end_position_exclusive
-                   -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
-                   AND unscanned.block_range_start <= :anchor_height
-                   -- exclude unscanned ranges that end below the wallet birthday
-                   AND unscanned.block_range_end > :wallet_birthday
-                 )
```
nuttycom, line 265: It occurs to me now that this change is not correct in general, in that it could have problems if the target height is a ways back from the chain tip (where there are unscanned blocks above the target height within the tip shard). However, because of the way we handle transaction expiry, the target height for note selection should *always* be the chain tip; this makes me think that some of the other top-level APIs that currently take (and have always taken) a target height are malformed, and that top-level target height arguments should be removed in favor of only working in terms of the chain tip.

The exception to this rule is wallet summary APIs; for these, it would ideally be the case that the wallet summary could be queried as of any given height; currently, the wallet summary is *only* available as of the chain tip.

My conclusion here is that this change is okay, but incomplete without a more comprehensive review of how target height is used.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -1,17 +1,5 @@
...
-            AddressType, DataStoreFactory, ShieldedProtocol, TestBuilder, TestCache, TestState,
-        },
-        wallet::{decrypt_and_store_transaction, input_selection::GreedyInputSelector},
-        Account as _, Balance, InputSource, WalletRead, WalletWrite,
-    },
-    fees::{standard, DustOutputPolicy, StandardFeeRule},
-    wallet::WalletTransparentOutput,
-};
 use assert_matches::assert_matches;
+use std::collections::BTreeMap;
```
str4d, line 1: We do not group `std`, `core` etc. stdlib imports with third party dependency imports. Revert this move.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1946,83 +1942,104 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
-        let any_spendable = is_any_spendable(tx, summary_height, table_prefix)?;
+        let trusted_height =
+            target_height.saturating_sub(u32::from(confirmations_policy.trusted()));
+        let untrusted_height =
+            target_height.saturating_sub(u32::from(confirmations_policy.untrusted()));
+
+        let any_spendable = is_any_spendable(tx, trusted_height, table_prefix)?;
+
+        // The trusted height will be used as the anchor height.
         let mut stmt_select_notes = tx.prepare_cached(&format!(
```
str4d: Looks to me like the comment is in the wrong place; it should be just above the `let any_spendable` line, as that's where `trusted_height` is first used as the anchor height.
```suggestion

        // The trusted height will be used as the anchor height.
        let any_spendable = is_any_spendable(tx, trusted_height, table_prefix)?;

        let mut stmt_select_notes = tx.prepare_cached(&format!(
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2034,23 +2051,40 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
+            let (
+                spendable_value,
+                change_pending_confirmation,
+                value_pending_spendability,
+                uneconomic_value,
+            ) = {
                 let zero = Zatoshis::ZERO;
-                if is_spendable {
-                    (value, zero, zero)
+                if value <= Zatoshis::const_from_u64(5000) {
```
```suggestion
                if value <= zip317::MARGINAL_FEE {
```
along with a `use zcash_primitives::transaction::fees::zip317;` above.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -223,58 +228,51 @@ where
...
-                   OR stx.expiry_height IS NULL -- the spending tx will not expire
-                   OR stx.expiry_height > :anchor_height -- the spending tx is unexpired
-                 )
-                 AND NOT EXISTS (
-                    SELECT 1 FROM v_{table_prefix}_shard_unscanned_ranges unscanned
-                    -- select all the unscanned ranges involving the shard containing this note
-                    WHERE {table_prefix}_received_notes.commitment_tree_position >= unscanned.start_position
-                    AND {table_prefix}_received_notes.commitment_tree_position < unscanned.end_position_exclusive
-                    -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
-                    AND unscanned.block_range_start <= :anchor_height
```
str4d, line 262: I agree that this is fine to remove, because if a note is mined inside a range that starts above the anchor height, then it would already be excluded by the existing `t.block <= :anchor_height` check.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -782,38 +785,36 @@ pub(crate) fn get_wallet_transparent_output(
 /// * the output is unspent as of the current chain tip.
 ///
 /// An output that is potentially spent by an unmined transaction in the mempool is excluded
-/// iff the spending transaction will not be expired at `target_height`.
+/// if the spending transaction will not be expired at `target_height`.
```
str4d: Is this "if and only if" no longer true? (`iff` is never a typo in Zcash repos.)

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -782,38 +785,36 @@ pub(crate) fn get_wallet_transparent_output(
 /// * the output is unspent as of the current chain tip.
```
str4d, line 786: The documentation above here is now incorrect, as it refers to the now-removed `min_confirmations` field. Rewrite it to reflect `confirmations_policy`.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -822,19 +823,24 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
...
-            -- as of the target height, even if they might actually be spendable due to expiry
-            -- of the spending transaction as of the chain tip
+            OR tx.expiry_height >= :target_height -- the spending tx is unexpired
          )",
     )?;
 
     let addr_str = address.encode(params);
+
+    // We treat all transparent UTXOs as untrusted; however, if zero-conf shielding
+    // is enabled, we set the minimum number of confirmations to zero.
```
str4d, line 834: Note to future-us: when we alter this crate to support spend-from-transparent (for Zallet), this will need to change to detect and support trusted transparent UTXOs.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -782,38 +785,36 @@ pub(crate) fn get_wallet_transparent_output(
...
-            -- OR (
-            --     :min_confirmations = 0
-            --     AND (
-            --         t.expiry_height = 0 -- tx will not expire
-            --         OR t.expiry_height >= :target_height
-            --     )
-            -- )
+            -- tx is mined and has at least min_confirmations
+            (
+                t.mined_height < :target_height -- tx is mined
```
str4d, line 811: Remind me what this evaluates to if `t.mined_height IS NULL`? Because this is a bit of a weird statement: a transaction is mined as long as `t.mined_height IS NOT NULL`, and this calculation only happens to work because by convention `target_height` always means `chain_height + 1`.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -856,25 +862,36 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
     params: &P,
     account_uuid: AccountUuid,
-    summary_height: BlockHeight,
-) -> Result<HashMap<TransparentAddress, Zatoshis>, SqliteClientError> {
-    let chain_tip_height = chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)?;
+    target_height: TargetHeight,
+    confirmations_policy: ConfirmationsPolicy,
+) -> Result<HashMap<TransparentAddress, Balance>, SqliteClientError> {
+    // We treat all transparent UTXOs as untrusted; however, if zero-conf shielding
+    // is enabled, we set the minimum number of confirmations to zero.
```
str4d, line 870: Ditto.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -884,70 +901,133 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
     ])?;
+
     while let Some(row) = rows.next()? {
         let taddr_str: String = row.get(0)?;
         let taddr = TransparentAddress::decode(params, &taddr_str)?;
         let value = Zatoshis::from_nonnegative_i64(row.get(1)?)?;
 
-        res.insert(taddr, value);
+        let entry = result.entry(taddr).or_insert(Balance::ZERO);
+        if value < Zatoshis::const_from_u64(5000) {
```
```suggestion
        if value < zip317::MARGINAL_FEE {
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -884,70 +901,133 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+            ":min_confirmations": min_confirmations
+        ])?;
+
+        while let Some(row) = rows.next()? {
+            let taddr_str: String = row.get(0)?;
+            let taddr = TransparentAddress::decode(params, &taddr_str)?;
+            let value = Zatoshis::from_nonnegative_i64(row.get(1)?)?;
+
+            let entry = result.entry(taddr).or_insert(Balance::ZERO);
+            if value < Zatoshis::const_from_u64(5000) {
```
```suggestion
            if value < zip317::MARGINAL_FEE {
```

str4d (changes requested):
Reviewed a0cd29f1f353cf2580e9b4ad2e3216c0807f9668.

zcash_client_sqlite/src/testing/db.rs
nuttycom: They're used by `#[derive(Delegate)]` code. The delegate derivation doesn't qualify the types of imported argument/result types, so they have to be imported manually; that's how it's always been for this file.

zcash_client_backend/src/scan.rs
nuttycom: This is a beta lint fix.

nuttycom: (with the most recent clippy beta)

Cargo.toml
nuttycom: @schell?

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: Yes, this is a side effect of the `anchor_height` being computed by `get_target_and_anchor_heights`. An alternative would be to move `get_target_and_anchor_heights` into the `InputSelector` trait from `WalletRead` and use it here; I had considered that, but as that's a somewhat invasive change I thought this was the best compromise for this PR.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: The subtlety here is that the replacement conditions:
```
            LEFT OUTER JOIN v_{table_prefix}_shards_scan_state scan_state
                ON rn.commitment_tree_position >= scan_state.start_position
                AND rn.commitment_tree_position < scan_state.end_position_exclusive
            WHERE
            ...
            AND scan_state.max_priority <= :scanned_priority            
            ...            
```
is batching scan state by _shard_, rather than by _scan range_. But my conclusion (in the comment just below this) is that we essentially don't support spending unless we have fully scanned the shard at the chain tip anyway. This query (and the others like it) will have to change in order to properly support multi-account functionality (related to https://github.com/zcash/librustzcash/issues/982#issuecomment-2024078941) but my rationale for making this change now is that I want all of these queries to behave as uniformly as possible. Ideally we could have a view that factors out all of the common logic, but the way that the target height is used in the `OR` clause makes this tricky; it's unfortunate that SQLite doesn't have stored procedures.

zcash_client_sqlite/src/wallet.rs
nuttycom: Actually, in order to be consistent with the other methods, we should call `get_target_and_anchor_heights` directly, and use the anchor height from there.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Any comparison to `NULL` except the `IS NULL` comparison, always returns `false`. 

The consistent use of `target_height` (and the newtype) is intended to make the convention you mention closer to true by construction. Given transaction expiry, it doesn't make sense to have target heights that are defined in any other way.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: I think `iff` is still correct; I can't think of another case where a mempool transaction that spends an output would be disregarded.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: This comment is correct?

Cargo.toml
schell: *tl;dr* It's not required. It fixes a cargo warning when building the project. In general using resolver 2 makes cargo behave a bit more like you would expect with regard to the way it determines features.

See:
* <https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html> to read about the default
* <https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2> to read about resolver 2

str4d: Given that this is a significant change to how the entire workspace builds, I would prefer in future that changes like it be introduced in separate PRs if they are not specifically required by a feature PR like this one.

zcash_client_sqlite/src/wallet/transparent.rs
str4d: The ditto was incorporating my earlier comment on an identical pair of lines: https://github.com/zcash/librustzcash/pull/1888#discussion_r2279387452

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -777,43 +781,41 @@ pub(crate) fn get_wallet_transparent_output(
 
 /// Returns the list of spendable transparent outputs received by this wallet at `address`
 /// such that, at height `target_height`:
-/// * the transaction that produced the output had or will have at least `min_confirmations`
-///   confirmations; and
+/// * the transaction that produced the output had or will have at least the number of
+///   confirmations required by the specified confirmations policy, and;
```
str4d: When using trailing punctuation in a list, the previous order was correct:
```suggestion
///   confirmations required by the specified confirmations policy; and
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -255,6 +281,223 @@ pub type ExtractErrT<DbT, N> = Error<
...
 
+/// A wrapper type around [`BlockHeight`] that represents the _next_ chain tip.
+///
+/// Addition and subtraction are provided by proxying to [`BlockHeight`].
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TargetHeight(BlockHeight);
+
+impl TargetHeight {
+    pub fn saturating_sub(self, value: u32) -> BlockHeight {
```
str4d, line 295: Document this method.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -255,6 +281,223 @@ pub type ExtractErrT<DbT, N> = Error<
...
+}
+
+/// The default confirmations policy according to [`ZIP 315`].
+///
+/// * Require 3 confirmations for "trusted" transaction outputs (outputs produced by the wallet)
+/// * Require 10 confirmations for "untrusted" outputs (those sent to the wallet by external/third
+///   parties)
+/// * Allow zero-conf shielding of transparent UTXOs irrespective of their origin, but treat the
+///   resulting shielding transaction's outputs as though the original transparent UTXOs had
+///   instead been received as untrusted shielded outputs.
```
str4d, line 368: Where is the second part of this policy implemented? I don't recall this from my review.

nuttycom: Added https://github.com/zcash/librustzcash/pull/1888/commits/d6978545e169b910ba4c4477875202f1ce40bfa9

Cargo.toml
schell: Will do, thanks 🙇

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1946,83 +1943,115 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
-            "SELECT a.uuid, n.value, n.is_change, scan_state.max_priority, t.block
-             FROM {table_prefix}_received_notes n
-             JOIN accounts a ON a.id = n.account_id
-             JOIN transactions t ON t.id_tx = n.tx
+            "SELECT accounts.uuid, rn.value, rn.is_change, rn.recipient_key_scope,
+                    scan_state.max_priority,
+                    t.block AS mined_height,
+                    MAX(tt.block) AS max_shielding_input_height
+             FROM {table_prefix}_received_notes rn
+             INNER JOIN accounts accounts ON accounts.id = rn.account_id
```
str4d: Why did you change this from the `accounts a` rename to instead rename the table to its own name?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2034,23 +2063,48 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
-                && received_height.iter().any(|h| h <= &summary_height)
-                && max_priority <= ScanPriority::Scanned;
+                && max_priority <= ScanPriority::Scanned
+                && match recipient_key_scope {
+                    Some(KeyScope::INTERNAL) => {
+                        // The note was has at least `trusted` confirmations.
+                        received_height.iter().any(|h| h <= &trusted_height) &&
+                        // And, if the note was the output of a shielding transaction, its
+                        // have at least `untrusted` confirmations
+                        max_shielding_input_height.iter().all(|h| h <= &untrusted_height)
```
```suggestion
                        // And, if the note was the output of a shielding transaction, its
                        // transparent inputs have at least `untrusted` confirmations.
                        max_shielding_input_height.iter().all(|h| h <= &untrusted_height)
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -223,58 +239,65 @@ where
...
-             SELECT id, txid, {output_index_col},
-                    diversifier, value, {note_reconstruction_cols}, commitment_tree_position,
-                    ufvk, recipient_key_scope
-             FROM (SELECT * from eligible WHERE so_far >= :target_value LIMIT 1)",
-        )
-    )?;
+             GROUP BY
+                rn.id, t.txid, rn.{output_index_col},
+                rn.diversifier, rn.value, {note_reconstruction_cols}, rn.commitment_tree_position,
+                ufvk, rn.recipient_key_scope, t.block
```
str4d, line 287: I can't find a reference for this, but I believe the reason this works (instead of double-counting value in `so_far`) is that the order of execution here is:
- `GROUP BY` to reorder the rows by this grouping, which excludes both `so_far` and `max_shielding_input_height`.
- `MAX` to compute the aggregate function producing `max_shielding_input_height`.
- `SUM OVER` to compute the window function over the now aggregated rows to get `so_far`.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -292,16 +315,43 @@ where
...
+                        .mined_height()
+                        .expect("mined height checked to be non-null");
+
+                    let is_spendable = match note.spending_key_scope() {
+                        Scope::Internal => {
+                            // The note was has at least `trusted` confirmations.
+                            received_height <= trusted_height &&
+                            // And, if the note was the output of a shielding transaction, its
+                            // have at least `untrusted` confirmations
+                            note.max_shielding_input_height().iter().all(|h| h <= &untrusted_height)
```
str4d:
```suggestion
                            // And, if the note was the output of a shielding transaction, its
                            // transparent inputs have at least `untrusted` confirmations.
                            note.max_shielding_input_height().iter().all(|h| h <= &untrusted_height)
```

zcash_client_sqlite/src/wallet.rs
nuttycom: Oh, weird, I meant to delete the alias for consistency with the other queries.

nuttycom:
```suggestion
             INNER JOIN accounts ON accounts.id = rn.account_id
```

zcash_client_sqlite/src/wallet/common.rs
nuttycom: ~Hmm, actually, I wonder if instead of `ROWS UNBOUNDED PRECEDING` it should be `GROUPS UNBOUNDED PRECEDING` https://sqlite.org/syntax/frame-spec.html~

Nope, that doesn't work.

PR #1893, opened 2025-08-06
add(zcash_primitives): Support for constructing coinbase transactions

upbqdn:
Extends the tx builder in `zcash_primitives` to support building coinbase transactions.


upbqdn:
### Notes for Reviewers

- I tried following https://github.com/zcash/librustzcash/blob/main/CONTRIBUTING.md.
- Each commit passes `cargo test --all-features`, and I tried producing a clean commit history.
- I didn't edit changelogs yet. I'm happy to do it after getting concept ACK for the PR.
- Similarly to changelogs, I didn't add any tests yet. I'd like pointers from the reviewers as to what they'd like to see tested.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -285,6 +285,21 @@ impl Authorization for Unauthorized {
...
+/// [`Authorization`] marker type for coinbase transactions without authorization data.
+pub struct Coinbase;
+
+#[cfg(feature = "circuits")]
+impl Authorization for Coinbase {
+    type TransparentAuth = ::transparent::builder::Coinbase;
+    type SaplingAuth =
+        sapling_builder::InProgress<sapling_builder::Proven, sapling_builder::Unsigned>;
+    type OrchardAuth =
+        orchard::builder::InProgress<orchard::builder::Unproven, orchard::builder::Unauthorized>;
```
nuttycom, line 297: The inconsistency between Sapling and Orchard here is odd. It would make more sense to me to require `orchard::Proven`, even if that's slightly more restrictive.

zcash_transparent/Cargo.toml
```diff
@@ -26,7 +26,10 @@ blake2b_simd.workspace = true
 sha2.workspace = true
 
 # - Encodings
+bs58.workspace = true
 core2.workspace = true
+hex.workspace = true
+serde.workspace = true
```
nuttycom: Can the `serde` dependency be feature-flagged? I would prefer not to include it in most cases.

zcash_primitives/src/transaction/mod.rs
upbqdn: I was following https://github.com/zcash/librustzcash/blob/8086b5583be0588b5d188bb548276f580aedae09/zcash_primitives/src/transaction/mod.rs#L280-L282

upbqdn: And `orchard` only seems to have `Unproven`: https://docs.rs/orchard/latest/orchard/?search=proven.

zcash_transparent/src/bundle.rs
```diff
@@ -228,6 +236,45 @@ impl TxIn<Authorized> {
...
+    /// Creates an input for a coinbase transaction. Does not support creating inputs for the
+    /// genesis block.
+    pub fn coinbase(
+        height: BlockHeight,
+        miner_data: &MinerData,
+        sequence: u32,
+    ) -> Result<Self, coinbase::Error> {
+        let h = u32::from(height);
+
+        let mut script_sig = match h {
```
nuttycom:
```suggestion
        // The following algorithm is defined by https://zips.z.cash/protocol/protocol.pdf#txnconsensus
        let mut script_sig = match h {
```

zcash_transparent/src/bundle.rs
```diff
@@ -228,6 +236,45 @@ impl TxIn<Authorized> {
...
+                vec![0x01, h as u8]
+            }
+            128..=32_767 => {
+                let b = h.to_le_bytes();
+                vec![0x02, b[0], b[1]]
+            }
+            32_768..=8_388_607 => {
+                vec![0x03, h as u8, (h >> 8) as u8, (h >> 16) as u8]
+            }
+            8_388_608..=u32::MAX => {
```
nuttycom: AFAICT his does not match the spec; the height is specified to be encoded as a _signed_ little-endian representation, with length in the range `1..=5`; note that in the protocol specification `..` ranges are end-inclusive:

> Otherwise, let heightBytes be the signed little-endian representation of height, using the
minimum nonzero number of bytes such that the most significant byte is < 0x80. The length of heightBytes MUST be in the range {1 .. 5}. Then the encoding is the length of heightBytes encoded as one byte, followed by heightBytes itself. This matches the encoding used by Bitcoin in the implementation of [BIP-34] (but the description here is to be considered normative).

It might be useful to use [zcash_script::script::serialize_num](https://github.com/ZcashFoundation/zcash_script/blob/master/src/script.rs#L407) here.

zcash_transparent/Cargo.toml
upbqdn: Done in `zcash_transparent: Derive Serialize for MinerData`.

zcash_transparent/src/bundle.rs
upbqdn: Thanks for catching this and suggesting [zcash_script::script::serialize_num](https://github.com/ZcashFoundation/zcash_script/blob/master/src/script.rs?rgh-link-date=2025-08-11T14%3A21%3A57Z#L407). I updated the serialization in `zcash_transparent: Construct coinbase inputs`. It's much cleaner now.

upbqdn:
- Rebased and addressed review remarks.
- Added corresponding changelog entries to each commit.

zcash_transparent/src/bundle.rs
upbqdn: I pulled `zcash_script` as a git dependency for now since `serialize_num` is private in the current release. The new dependency requires updating `cargo vet`, which I think will be easier to do only after https://github.com/ZcashFoundation/zcash_script/pull/255 gets released.

upbqdn: Should I do anything about `zcash_script` failing to compile for Wasm and ARM?

upbqdn: It fails because of the C++ compilation. I'll try skipping it so that at least pure Rust code is available.

upbqdn: Hm, but it looks like we'll need to make `zcash_script` `no_std` to be able to compile it for the ARM target, which I don't know how to do trivially.

upbqdn: So I fixed the compilation for the Wasm target in `zcash_script` locally by skipping the C++ parts, which was a rather trivial change. However, making the crate `no_std` will be a much heavier lift if there isn't a more suitable way, which I don't see now.

nuttycom: A couple of things: First, @str4d has already done the `no_std` work, so that just needs to make it into a PR.

Second, we should extract all the stuff that depends on C++ into a separate crate in some fashion. I think that we should create a new crate, `legacy_zcash_script`, and have that depend on the `zcash_script` crate which then would contain just the Rust implementation. @sellout what do you think?

upbqdn: @str4d, feel free to tag me as a reviewer.

sellout: > we should extract all the stuff that depends on C++ into a separate crate in some fashion. I think that we should create a new crate, `legacy_zcash_script`, and have that depend on the `zcash_script` crate which then would contain just the Rust implementation. @sellout what do you think?

Yes, definitely want to do this. I think code-wise it’s trivial[^1], but I’m concerned about the outstanding work getting all conflicty if we split it now.

[^1]: C++-related code is all in lib.rs and the `cxx` module. I think the `zcash_script` module would also move to `legacy_zcash_script` since (after ZcashFoundation/zcash_script#209) Rust-only usage has a nicer API.

@upbqdn In ZcashFoundation/zcash_script#253, `script::serialize_num` becomes `num::serialize` (and that becomes public in ZcashFoundation/zcash_script#254).

upbqdn: @sellout, the PRs are in, but the `num` mod is not `pub`: https://github.com/ZcashFoundation/zcash_script/blob/ddd65552524df5e6c6cde0b24ef3e417d42248f2/src/lib.rs#L15. Should I make it `pub`?

sellout: > @sellout, the PRs are in, but the `num` mod is not `pub`

Ah, the whims of the public API …

There are more (smaller) PRs still landing. One of them is specifically about getting the public API right. `num` should definitely be `pub`, so I’ve made the change on that branch. (But separating out the C++ code won’t land until after 0.4.0 is cut).

upbqdn:
Rebased on `main`.

zcash_transparent/src/bundle.rs
sellout: @upbqdn Ok, it’s pub now for sure.

upbqdn: @sellout, this PR now depends on `script::Code` being `PartialEq`: https://github.com/ZcashFoundation/zcash_script/compare/master...make-code-partialeq. Should I do a patch release?

Cargo.toml
```diff
@@ -224,3 +224,7 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(zcash_unstable, values("zfuture", "nu7"))',
   'cfg(live_network_tests)',
 ] }
+
+
+[patch.crates-io]
+zcash_script = { git = "https://github.com/ZcashFoundation/zcash_script.git", rev = "d6dbbf4e69ea37a87132a652570d963662e88619" }
```
upbqdn: This patch references https://github.com/ZcashFoundation/zcash_script/compare/master...make-code-partialeq.

zcash_transparent/src/bundle.rs
sellout: I’ll loom at the code shortly, but should this PR instead be parsing the `Code` and then looking at opcodes?

upbqdn: I use it here: https://github.com/zcash/librustzcash/blob/bd6853863a46a5328f7069da2e36025983c172ad/zcash_transparent/src/bundle.rs#L291 and here: https://github.com/zcash/librustzcash/blob/bd6853863a46a5328f7069da2e36025983c172ad/zcash_transparent/src/builder.rs#L188-L194 since `Script` now wraps `Code` instead of `Vec<u8>`.

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +266,34 @@ impl TxIn<Authorized> {
...
+    ) -> Result<Self, coinbase::Error> {
+        let height = i64::from(height);
+
+        // Serialize the height. The serialization is specified in
+        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>.
+        let mut script_sig = match height {
+            0 => Err(coinbase::Error::GenesisInputNotSupported)?,
+            1..=16 => vec![0x50 + height.to_le_bytes()[0]],
+            _ => zcash_script::num::serialize(height),
+        };
```
sellout:
```suggestion
        let mut script_sig = match height {
            0 => Err(coinbase::Error::GenesisInputNotSupported)?,
            _ => script::Component(vec![zcash_script::pattern::push_num(height)]),
        }
```
`push_num` and a few of the other functions from `pattern` should probably move to the `op` and `pv` modules.

zcash_transparent/src/builder.rs
```diff
@@ -178,6 +185,37 @@ pub struct TransparentSignatureContext<'a, V: secp256k1::Verification = secp256k
     final_script_sigs: Vec<Option<script::Sig>>,
 }
 
+/// [`Authorization`] marker type for coinbase transactions without authorization data.
+#[derive(Debug, Clone)]
+pub struct Coinbase;
+
+impl Authorization for Coinbase {
+    type ScriptSig = script::Code;
```
sellout:
```suggestion
    type ScriptSig = script::Sig;
```

zcash_transparent/src/bundle.rs
upbqdn: Here: https://github.com/zcash/librustzcash/blob/bd6853863a46a5328f7069da2e36025983c172ad/zcash_transparent/src/address.rs#L19-L21

upbqdn: I need `Script` here: https://github.com/zcash/librustzcash/blob/bd6853863a46a5328f7069da2e36025983c172ad/zcash_transparent/src/builder.rs#L206-L212 since `Authorized` uses it: https://github.com/zcash/librustzcash/blob/bd6853863a46a5328f7069da2e36025983c172ad/zcash_transparent/src/bundle.rs#L49-L54

zcash_transparent/src/builder.rs
```diff
@@ -178,6 +185,37 @@ pub struct TransparentSignatureContext<'a, V: secp256k1::Verification = secp256k
...
+        vec![]
+    }
+}
+
+impl MapAuth<Coinbase, Authorized> for Coinbase {
+    fn map_script_sig(
+        &self,
+        s: <Coinbase as Authorization>::ScriptSig,
+    ) -> <Authorized as Authorization>::ScriptSig {
+        Script(s)
```
sellout:
```suggestion
        Script::from(s)
```
(Edited this comment because I didn’t realize the `address::Script` type had changed from the last time I looked.)

zcash_transparent/src/builder.rs
```diff
@@ -4,13 +4,17 @@ use alloc::collections::BTreeMap;
 use alloc::vec::Vec;
 use core::fmt;
 
-use zcash_protocol::value::{BalanceError, ZatBalance, Zatoshis};
+use zcash_protocol::{
+    consensus::BlockHeight,
+    value::{BalanceError, ZatBalance, Zatoshis},
+};
 
 use zcash_script::{op, script};
```
sellout:
```suggestion
use zcash_script::{op, script::{self, Evaluable}};
```
I think you’ll need this trait in scope for the `to_bytes` call I added below.

upbqdn:
> Some suggestions for avoiding script::Code for Coinbase.

Thanks. I've committed them, and I'll resolve the arising compilation errors a bit later.

> I didn’t notice where the `PartialEq` impl was needed [...]

Here:

https://github.com/zcash/librustzcash/blob/b8a76f28cca225f494dcf5a9e49ffc601d35db43/zcash_transparent/src/bundle.rs#L21-L23

zcash_primitives/src/transaction/builder.rs
```diff
@@ -386,11 +406,26 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, ()> {
...
+        //
+        // > [NU5 onward] The `nExpiryHeight` field of a coinbase transaction MUST be equal to its
+        // > block height.
+        //
+        // ## Notes
+        //
+        // We set the expiry height for coinbase txs to the block height regardless of the network
+        // upgrade.
+        let expiry_height = if build_config.is_coinbase() {
+            target_height
```
bhaumikmaan, line 421: Not sure if it needs to be explicitly handled, how will the Zip212 enforcement be handled for Coinbase Transactions across different consensus, specifically for Sapling Transactions (also considering the grace period)

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// # Consensus
+///
+/// > The length of `heightBytes` MUST be in the range {1 .. 5}. Then the encoding is the length of
+/// > `heightBytes` encoded as one byte, followed by heightBytes itself.
+///
+/// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+pub const MAX_COINBASE_HEIGHT_LEN: usize = 6;
+
+/// The maximum lenght of the miner data in the coinbase script sig field.
```
bhaumikmaan: nit: typo
```suggestion
/// The maximum length of the miner data in the coinbase script sig field.
```

zcash_primitives/src/transaction/builder.rs
upbqdn: ZIP-212 is handled upstream.

zcash_primitives/CHANGELOG.md
```diff
@@ -41,8 +41,21 @@ workspace.
...
+      and `sequence: u32`.
+    - `BuildConfig::is_coinbase`
+    - `Builder::add_transparent_null_data_output`
+    - `Builder::add_orchard_output`'s `value` parameter is now `Zatoshis`
+      instead of `u64`.
+    - `Builder` now supports constructing coinbase transactions.
+  - `Coinbase` marker type
+  - `impl Authorization for Coinbase`
+  - `tests::data` has been exposed under the `test-dependencies` feature to
+    provide access to test vectors.
```
nuttycom: This section needs to be moved to the `Unreleased` part of the CHANGELOG, but with the part that was actually released left in place.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -202,14 +215,17 @@ impl Progress {
...
 pub enum BuildConfig {
     Standard {
         sapling_anchor: Option<sapling::Anchor>,
         orchard_anchor: Option<orchard::Anchor>,
     },
-    Coinbase,
+    Coinbase {
+        miner_data: MinerData,
+        sequence: u32,
+    },
```
nuttycom, line 229: Is this also required for shielded coinbase?

zcash_primitives/src/transaction/builder.rs
```diff
@@ -461,7 +496,7 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
         &mut self,
         ovk: Option<orchard::keys::OutgoingViewingKey>,
         recipient: orchard::Address,
-        value: u64,
+        value: Zatoshis,
```
arya2, line 501: Update: It would be nice to leave this as a `u64` so that this method can be re-used for adding Orchard outputs of custom assets.

<details>
<summary>
Outdated:
</summary>

It may be nice to ~add a `BalanceError` variant to `builder::Error` and~ allow any type that implements TryInto<Zatoshis> here with an unreachable `From` impl for `Infallible` errors when passing in `Zatoshis`.

```rust
impl<FE> From<Infallible> for Error<FE> {
    fn from(_: Infallible) -> Self {
        unreachable!()
    }
}
```


Update: I see there's already a `Balance` variant on `builder::Error`.

```suggestion
        value: impl TryInto<Zatoshis, Error: Into<Error<FE>>>,
```

</details>

zcash_transparent/src/bundle.rs
```diff
@@ -187,9 +187,7 @@ impl OutPoint {
     /// Returns `true` if this `OutPoint` is "null" in the Bitcoin sense: it has txid set to
     /// all-zeroes and output index set to `u32::MAX`.
     fn is_null(&self) -> bool {
-        // From `BaseOutPoint::IsNull()` in zcashd:
-        //   return (hash.IsNull() && n == (uint32_t) -1);
-        self.hash.is_null() && self.n == u32::MAX
```
arya2, line 185: We may want to remove the `is_null()` method on `TxId` if it's no longer used.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -104,6 +105,8 @@ pub enum Error<FE> {
     /// The builder was constructed with a target height before NU5 activation, but an Orchard
     /// spend or output was added.
     OrchardBuilderNotAvailable,
+    /// And error occurred in constructing a coinbase transaction.
```
arya2:
```suggestion
    /// An error occurred in constructing a coinbase transaction.
```

zcash_primitives/src/transaction/builder.rs
arya2: Yep, these are required for the transparent input, and the transparent input is at least implicitly required as coinbase transactions MUST NOT have JoinSplit descriptions or Spend descriptions, and for v5 transactions, the enableSpendsOrchard flag MUST be set to 0 (stated in the middle of page 124 in the latest protocol spec).

zcash_primitives/src/transaction/builder.rs
```diff
@@ -706,22 +751,162 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
+            Some((bundle, meta)) => (Some(bundle), meta),
+            None => (None, orchard::builder::BundleMetadata::empty()),
+        };
+
+        let unauthed_tx: TransactionData<Coinbase> = TransactionData {
+            version: TxVersion::suggested_for_branch(consensus_branch_id),
+            consensus_branch_id,
+            lock_time: 0,
+            expiry_height: self.expiry_height,
+            transparent_bundle: Some(transparent_bundle.clone()),
```
arya2:
```suggestion
            #[cfg(all(
                any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
                feature = "zip-233"
            ))]
            zip233_amount: self.zip233_amount,
            transparent_bundle: Some(transparent_bundle.clone()),
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -706,22 +751,162 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
+            })
+            .transpose()
+            .map_err(Error::OrchardBuild)?;
+
+        let authed_tx: TransactionData<Authorized> = TransactionData {
+            version: unauthed_tx.version,
+            consensus_branch_id: unauthed_tx.consensus_branch_id,
+            lock_time: unauthed_tx.lock_time,
+            expiry_height: unauthed_tx.expiry_height,
+            transparent_bundle: Some(transparent_bundle),
```
arya2:
```suggestion
            #[cfg(all(
                any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
                feature = "zip-233"
            ))]
            zip233_amount: self.zip233_amount,
            transparent_bundle: Some(transparent_bundle),
```

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +266,29 @@ impl TxIn<Authorized> {
...
+        miner_data: &MinerData,
+        sequence: u32,
+    ) -> Result<Self, coinbase::Error> {
+        let height = i64::from(height);
+
+        // Serialize the height. The serialization is specified in
+        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>.
+        let mut script_sig = match height {
+            0 => Err(coinbase::Error::GenesisInputNotSupported)?,
+            _ => zcash_script::pattern::push_num(height).to_bytes(),
```
arya2: What is the last step in `zcash_script::num::serialize()` doing?

```rust

result.last().map_or(true, |last| last & 0x80 != 0)

```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+/// # Consensus
+///
+/// > The length of `heightBytes` MUST be in the range {1 .. 5}. Then the encoding is the length of
+/// > `heightBytes` encoded as one byte, followed by heightBytes itself.
+///
+/// <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+pub const MAX_COINBASE_HEIGHT_LEN: usize = 6;
+
+/// The maximum length of the miner data in the coinbase script sig field.
+pub const MAX_MINER_DATA_LEN: usize = MAX_COINBASE_SCRIPT_LEN - MAX_COINBASE_HEIGHT_LEN;
```
arya2: This seems fine for constructing new coinbase transactions, but its use in Zebra's `parse_coinbase_height()` function seems wrong if coinbase transactions with heights that are encoded to fewer bytes are supposed to allow more miner data.

zcash_transparent/src/bundle.rs
arya2: This is great code re-use.

No changes needed here, just curious: Does anyone know why zcashd uses a sequence of `u32::MAX` for v4 transactions?

upbqdn: It does so for all coinbase tx versions, no? 

Doing that for all inputs in a tx essentially disables time-lock checks for that tx: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/primitives/transaction.h#L333-L336 and https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/main.cpp#L776-L778. If the output of `IsFinalTx` is `false`, the tx won't be mined: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/miner.cpp#L555-L556.

Zebra currently sets `sequence` to zero in coinbase txs. It can do so because it disables the time-lock checks by setting the lock time to zero, so `IsFinalTx` returns `true` here: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/main.cpp#L772-L773.
`zcashd` does the same as lock times don't apply to coinbase txs.

zcash_transparent/src/coinbase.rs
upbqdn: I think we'll make a bunch of things easier if we keep the limit static.

The miner data is currently set in Zebra's config file. If a miner sets the data so that it would exceed the limit at some future height, Zebra would have to handle that somehow.

zcash_primitives/src/transaction/builder.rs
upbqdn: Could that not be changed to `Zatoshis`, too?

Regarding the `TryInto`, I'm more cautious about using it after reading the fifth (the last) rule of thumb here: https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice.

zcash_primitives/src/transaction/builder.rs
upbqdn: Transparent inputs are also required explicitly, since the protocol spec defines coinbase txs as those with a single transparent input containing a null prevout.

The sequence field must be present in transparent inputs. The value is not relevant as long as the `lock_time` field in the tx is zero. `zcashd` sets the sequence field to `u32::MAX` and `zebrad` sets it to zero. Other than being legacy clutter in the protocol, there's no use for the field as far as I know.

The `MinerData` part is not required. `zcashd` doesn't support it, `zebrad` does. I think it's useful, I used it to label the block I mined: https://mainnet.zcashexplorer.app/transactions/472c432c8367e1cbc271d27649f11622240c39f2c33f1ea403ce7b5e94b42205 (the explorer doesn't render UTF-8 properly). For shielded outputs, miners can use the memo fields, but more code would be required to display a msg like that in the explorer.

I also used the `MinerData` field in this PR to comply with this consensus rule

> A coinbase transaction script MUST have length in {2 .. 100} bytes.

Vid. `impl Default for MinerData` in this PR. Essentially, `zcashd` always sets `MinerData` to a single zero byte. This PR does that only if the block height is serialized as a single byte and the user didn't set the field. We could also adjust the consensus rule and set the lower bound to 1 after zcashd deprecation.

> Is this also required for shielded coinbase?

Regardless of any shielded outputs in the tx, we could not expose any of those fields in the API and set the sequence implicitly to a static value, but I find the current design fine since it gives the caller the option to use what the protocol allows.

zcash_primitives/CHANGELOG.md
upbqdn: Done.

zcash_transparent/src/bundle.rs
upbqdn: Done in [c60b0b0](https://github.com/zcash/librustzcash/pull/1893/commits/c60b0b0a019264d9b6053a05bee1536771d7fca5).

zcash_primitives/CHANGELOG.md
str4d: This was not done.

zcash_transparent/src/bundle.rs
str4d: The behaviour is inherited from Bitcoin Core, which defined the maximum sequence value of `-1` (which is equal to `u32::MAX` in two's complement, and the maximum value you can increment `sequence` to starting from 0 before it wraps back around) as disabling the `nLockTime` feature.

zcash_transparent/src/bundle.rs
```diff
@@ -177,12 +180,17 @@ impl OutPoint {
         writer.write_all(&self.n.to_le_bytes())
     }
 
+    /// Returns the null [`OutPoint`] in the Bitcoin sense:
+    /// - [`TxId`] is set to all-zeroes, and
+    /// - output index is set to [`u32::MAX`].
+    pub fn null() -> Self {
```
str4d: This should be a `const fn` (or an associated constant).

zcash_primitives/CHANGELOG.md
upbqdn: Thanks for catching that. It should be now.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -236,12 +256,16 @@ impl BuildConfig {
...
                 .map(|a| (orchard::builder::BundleType::DEFAULT, *a)),
-            BuildConfig::Coinbase => Some((
+            BuildConfig::Coinbase { .. } => Some((
                 orchard::builder::BundleType::Coinbase,
                 orchard::Anchor::empty_tree(),
             )),
         }
     }
+
+    pub fn is_coinbase(&self) -> bool {
```
str4d, line 264: Document this method.

zcash_transparent/src/builder.rs
```diff
@@ -412,6 +453,20 @@ impl TransparentBuilder {
...
     }
 
+    /// Builds a coinbase bundle.
+    pub fn build_coinbase(
+        self,
+        height: BlockHeight,
+        miner_data: &MinerData,
+        sequence: u32,
+    ) -> Result<Bundle<Coinbase>, coinbase::Error> {
+        Ok(Bundle {
```
str4d, line 471: This code currently discards any inputs that the caller (incorrectly) asked to be spent. It instead needs to check that `self.inputs.is_empty()` before returning the bundle, and otherwise return an error. Also, document this restriction in the method docs.

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +266,29 @@ impl TxIn<Authorized> {
...
+        let height = i64::from(height);
+
+        // Serialize the height. The serialization is specified in
+        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>.
+        let mut script_sig = match height {
+            0 => Err(coinbase::Error::GenesisInputNotSupported)?,
+            _ => zcash_script::pattern::push_num(height).to_bytes(),
+        };
+
+        script_sig.extend(miner_data.as_ref());
```
str4d: It's been long enough since I messed with coinbase inputs that I've forgotten: is this correct for how miner data is appended? Should it not be in a `PushData`? Document the answer in this code so it's clear to the next reviewer.

zcash_primitives/CHANGELOG.md
```diff
@@ -20,6 +34,7 @@ workspace.
     - `zcash_primitives::memo`
     - `zcash_primitives::zip32`
     - `zcash_primitives::legacy`
+- `zcash_primitives::transaction::builder::BuildConfig` is not `Copy` anymore.
```
nuttycom: What is the rationale for this change?

components/zcash_protocol/src/txid.rs
```diff
@@ -65,10 +65,4 @@ impl TxId {
         writer.write_all(&self.0)?;
         Ok(())
     }
-
-    /// Returns true when the txid consists of all zeros; this only occurs for coinbase
-    /// transactions.
-    pub fn is_null(&self) -> bool {
```
str4d, line 71: Do not remove this from the public API. `TxId` nullability is an annoying thing inherited from Bitcoin, but one we need to preserve due to consensus (i.e. that an all-zeros txid cannot exist outside of coinbase). The fact that this is not used internally to the librustzcash workspace does not mean it is unused. This is also an unnecessary breaking change to `zcash_protocol` that contributes to this PR not being additive.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -285,6 +285,21 @@ impl Authorization for Unauthorized {
     type TzeAuth = tze::builder::Unauthorized;
 }
 
+/// [`Authorization`] marker type for coinbase transactions without authorization data.
+pub struct Coinbase;
```
str4d: This type does not need to be public AFAICT, because we do not actually want to expose this intermediate state.
```suggestion
struct Coinbase;
```

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,20 @@ workspace.
...
+    - `Error::Coinbase` enum variant
+    - `impl<FE> From<coinbase::Error> for Error<FE>`
+    - `BuildConfig::Coinbase` enum variant now contains `miner_data: MinerData`
+      and `sequence: u32`.
+    - `BuildConfig::is_coinbase`
+    - `Builder::add_orchard_output`'s `value` parameter is now `Zatoshis`
+      instead of `u64`.
+    - `Builder` now supports constructing coinbase transactions.
+  - `Coinbase` marker type
+  - `impl Authorization for Coinbase`
```
str4d:
```suggestion
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+            Error::GenesisInputNotSupported => write!(
+                f,
+                "creating a coinbase input for the genesis block is not supported",
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
```
str4d: The `zcash_transparent` changes are additive, so they should be based on `zcash_transparent 0.6.x` in order to be releasable without a breaking release.

That being said, the next breaking release of `zcash_transparent` has MSRV 1.85.1, for which we don't need the `std` feature flag and can `impl core::error::Error` instead.

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,32 @@ workspace.
...
+    - `MIN_COINBASE_SCRIPT_LEN`
+    - `MAX_COINBASE_HEIGHT_LEN`
+    - `MAX_MINER_DATA_LEN`
+    - `MinerData` struct
+    - `impl Default for MinerData`
+    - `impl TryFrom<&[u8]> for MinerData`
+    - `impl AsRef<[u8]> for MinerData`
+    - `Error` enum
+    - `impl fmt::Display for Error`
+    - `impl std::error::Error for Error`
```
str4d: This is an entirely new module, so we don't need to list everything it introduces.
```suggestion
  - `coinbase` module, containing helpers for constructing coinbase transactions.
```

zcash_transparent/Cargo.toml
```diff
@@ -30,7 +30,10 @@ blake2b_simd.workspace = true
 sha2.workspace = true
 
 # - Encodings
+bs58.workspace = true
 core2.workspace = true
+hex.workspace = true
```
str4d: AFAICT these were incorrectly moved to "dependencies exposed in a public API". They were previously categorised as "dependencies used internally", and I don't see any changes to `zcash_transparent` that expose their types. Revert this move.

zcash_transparent/CHANGELOG.md
```diff
@@ -10,6 +10,32 @@ workspace.
...
+  - Optional dependency on `serde` under the implicit `serde` feature.
```
str4d: We do not use implicit feature flags. Define a `serde` feature flag that enables `dep:serde` and is documented similarly to `transparent-inputs` (so it shows up in rustdoc renders).

zcash_primitives/src/transaction/builder.rs
nuttycom: > Could that not be changed to `Zatoshis`, too?

The output of a custom asset is explicitly NOT in zatoshis; it's denominated in the custom asset value type. In addition, `Zatoshis` is bounded to the valid range of ZEC values `0..(21000000 * 100000000)`. However, the signature of this method will need to change to take an `AssetValue` once ZSA handling is added, so this change is fine for now (and will actually help us avoid errors in the process of making the later change.)

str4d (changes requested):
Flushing partial review

upbqdn:
> We plan to rework this PR in the next few days [...]

Should I address the current review remarks then?

zcash_primitives/CHANGELOG.md
daira: Is this breaking change needed?

zcash_primitives/src/transaction/builder.rs
daira: As you say, the sequence number doesn't matter when `lock_time` is 0, and so there is no reason to complicate the API by including it here. librustzcash doesn't need to support varying what it sends just because the legacy protocol allows it; librustzcash is opinionated in many other ways. I would just do what Zebra does, unless there's some reason not to that I'm missing.

`miner_data`, on the other hand, is important for miners to voluntarily identify themselves. (At some point we might want to have that identification be authenticatable, but I doubt we want to remove it.)

zcash_primitives/src/transaction/builder.rs
daira: Properly supporting ZSAs for wallet functionality will require major changes and we shouldn't try to anticipate how that will be done. But I definitely hope that it will not be done with raw `u64`s.

I don't think the cost of monomorphization is very significant; OTOH, `impl TryInto` seems entirely unnecessary here. It's trying to avoid an explicit `.try_into()` at the call site but we *want* that to be explicit.

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,20 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_primitives::transaction`:
+  - `builder`:
+    - `Error::Coinbase` enum variant
+    - `impl<FE> From<coinbase::Error> for Error<FE>`
+    - `BuildConfig::Coinbase` enum variant now contains `miner_data: MinerData`
+      and `sequence: u32`.
```
daira: This is not an addition, it's a breaking change, because enum variants are public. Also: if `sequence` is removed, don't forget to also remove the reference to it here (but the `miner_data` change remains breaking).

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,20 @@ workspace.
...
+    - `BuildConfig::is_coinbase`
+    - `Builder::add_orchard_output`'s `value` parameter is now `Zatoshis`
+      instead of `u64`.
```
daira: This is not an addition, it's a breaking change.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -692,8 +731,18 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
 
     /// Builds a transaction from the configured spends and outputs.
     ///
-    /// Upon success, returns a tuple containing the final transaction, and the
-    /// [`SaplingMetadata`] generated during the build process.
+    /// Upon success, returns a [tuple] containing:
```
daira:
```suggestion
    /// Upon success, returns a [`BuildResult`] containing:
```
(qualify if needed).

zcash_primitives/src/transaction/builder.rs
```diff
@@ -692,8 +731,18 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
+    ///
+    /// - the [final transaction],
+    /// - the [Sapling metadata], and
+    /// - the [Orchard metadata]
+    ///
+    /// generated during the build process.
+    ///
+    /// [Sapling metadata]: ::sapling::builder::SaplingMetadata
+    /// [Orchard metadata]: ::orchard::builder::BundleMetadata
+    /// [tuple]: BuildResult
```
daira:
```suggestion
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -706,22 +755,172 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
+
+        let (miner_data, seq) = match self.build_config {
+            BuildConfig::Coinbase {
+                miner_data,
+                sequence,
+            } => (miner_data, sequence),
+            _ => panic!(
+                "the build_coinbase method should not be called without a coinbase build config"
+            ),
+        };
```
daira: Why not do the match in `build` and then call this method with `miner_data` (and `sequence` if it still exists)? Then `is_coinbase` would also not be needed.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -706,22 +755,172 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
+            sapling_meta,
+            orchard_meta,
+        })
     }
 
     /// Builds a transaction from the configured spends and outputs.
     ///
-    /// Upon success, returns a tuple containing the final transaction, and the
-    /// [`SaplingMetadata`] generated during the build process.
+    /// Upon success, returns a [tuple] containing:
```
daira:
```suggestion
    /// Upon success, returns a [`BuildResult`] containing:
```
(qualify if needed).

zcash_primitives/src/transaction/builder.rs
```diff
@@ -706,22 +755,172 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
+    ///
+    /// - the [final transaction],
+    /// - the [Sapling metadata], and
+    /// - the [Orchard metadata]
+    ///
+    /// generated during the build process.
+    ///
+    /// [Sapling metadata]: ::sapling::builder::SaplingMetadata
+    /// [Orchard metadata]: ::orchard::builder::BundleMetadata
+    /// [tuple]: BuildResult
```
daira:
```suggestion
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -738,21 +937,26 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
-            transparent_signing_set,
-            sapling_extsks,
-            orchard_saks,
-            rng,
-            spend_prover,
-            output_prover,
-            fee,
-        )
+        if self.build_config.is_coinbase() {
+            self.build_coinbase(rng, spend_prover, output_prover)
```
daira: Same comment as above.

zcash_transparent/src/coinbase.rs
daira: I think this is fine.

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+/// - The data cannot be empty.
+/// - The data must be less than [`MAX_MINER_DATA_LEN`] bytes.
+#[derive(Clone, Eq, PartialEq, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+pub struct MinerData(Vec<u8>);
+
+impl Default for MinerData {
+    fn default() -> Self {
+        // `zcashd` includes an empty byte after the coinbase height [1]. We do that only if
+        // `miner_data` would be empty to comply with the following consensus rule:
```
daira: This is stated confusingly. Suggestion for if the `Default` impl is not removed.
```suggestion
        // `zcashd` uses a single zero byte as `miner_data`, in order to comply in all cases
        // with the following consensus rule:
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// # Invariants
+///
+/// - The data cannot be empty.
+/// - The data must be less than [`MAX_MINER_DATA_LEN`] bytes.
+#[derive(Clone, Eq, PartialEq, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+pub struct MinerData(Vec<u8>);
+
+impl Default for MinerData {
```
daira: Consider just removing this impl, to provide a soft encouragement to set the `miner_data` appropriately. What zcashd did isn't really relevant any more.

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// The maximum length of the miner data in the coinbase script sig field.
+pub const MAX_MINER_DATA_LEN: usize = MAX_COINBASE_SCRIPT_LEN - MAX_COINBASE_HEIGHT_LEN;
+
+/// Arbitrary data inserted by miners into a coinbase transaction.
+///
+/// # Invariants
+///
+/// - The data cannot be empty.
+/// - The data must be less than [`MAX_MINER_DATA_LEN`] bytes.
```
daira: The rationale for the data not being empty belongs here, not in the non-doc comment inside `default()`.
```suggestion
/// - The data cannot be empty.
/// - The data must be less than [`MAX_MINER_DATA_LEN`] bytes.
///
/// ## Rationale for not allowing the data to be empty
///
/// Coinbase heights < 17 are serialized as a single byte, and if there is no miner
/// data, the script of a coinbase tx with such a height would consist only of this
/// single byte for those blocks, violating the consensus rule. Also, this acts as a
/// soft encouragement to think about what the miner data should be.
```
An alternative would be to just not support creating miner data for block heights < 17.

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+        Self(vec![0])
+    }
+}
+
+impl TryFrom<&[u8]> for MinerData {
+    type Error = Error;
+
+    fn try_from(data: &[u8]) -> Result<Self, Self::Error> {
+        match data.len() {
+            0 => Err(Error::EmptyMinerData),
```
daira: Remove this error if the alternative of only supporting creating coinbase blocks at height >= 17 is taken.

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// Errors related to coinbase transactions.
+#[derive(Clone, Debug)]
+pub enum Error {
+    /// The miner data exceeds [`MAX_MINER_DATA_LEN`].
+    OversizedMinerData(usize),
+    /// The miner data is empty.
+    EmptyMinerData,
+    /// Creating a coinbase input for the genesis block is not supported.
+    GenesisInputNotSupported,
```
daira:
```suggestion
    /// Creating a coinbase input for blocks at height < 17 is not supported.
    BlockTooEarly,
```

daira (dismissed):
utACK with comments, which are not strictly blocking so I'll check Approve.

zcash_transparent/src/coinbase.rs
upbqdn: > An alternative would be to just not support creating miner data for block heights < 17.

The ability to create blocks at heights < 17 is useful for regtest, so I think we should keep it.

zcash_primitives/src/transaction/builder.rs
upbqdn: I removed the `sequence` field from the API and set it to `u32::MAX` implicitly.

zcash_primitives/CHANGELOG.md
upbqdn: It's because the `BuildConfig::Coinbase` variant now contains `MinerData`, which isn't `Copy` because it wraps `Vec<u8>`.

zcash_primitives/src/transaction/builder.rs
upbqdn: I originally wrote it as you described, but I couldn't find a way to satisfy the borrow checker since `miner_data` isn't `Copy`.

If I do the match in `build` and borrow `miner_data` from `self` there, then I can't call `build_coinbase` on `self` anymore because `build_coinbase` needs to consume `self`.

upbqdn: That's also why I added `is_coinbase`.

zcash_transparent/src/bundle.rs
upbqdn: I'm looking at how `zcashd` does it.

If I understand the code correctly, it doesn't use miner data at all in the `getblocktemplate` RPC: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/miner.cpp#L308.

It does use optional miner data in its internal miner, though: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/miner.cpp#L802; and serializes it as a script. I'll do that, then.

zcash_transparent/src/coinbase.rs
upbqdn: Changing the consensus rule from

_"A coinbase transaction script MUST have length in {2 .. 100} bytes."_

to

_"A coinbase transaction script MUST have length in {1 .. 100} bytes."_

would eliminate this edge case. Would this be feasible?

zcash_transparent/src/coinbase.rs
upbqdn: I did it this way to abstract the minimal-script-length edge case away from users.

zcashd doesn't use miner data at all in the `getblocktemplate` RPC: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/miner.cpp#L308; and always puts a zero after the height instead. It does use optional miner data only in its internal miner: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/miner.cpp#L802.

Another solution would be to put the logic that deals with the use case here: https://github.com/zcash/librustzcash/blob/7378d3b7778f2645576c8029e20bc831a6ec2e1a/zcash_transparent/src/bundle.rs#L273 This would more closely resemble zcashd's approach.

upbqdn: OK, I'm going to discard all of this. The miner data should be a script represented as `PushValue`, and an empty `PushValue` is serialized as `OP_0`: https://github.com/zcashfoundation/zcash_script/blob/39294a8c3bc0c9b0cb8c8bad8ad29dca511f8e71/zcash_script/src/opcode/mod.rs#L99. I didn't know.

zcash_transparent/src/coinbase.rs
upbqdn: Discarding, vid. https://github.com/zcash/librustzcash/pull/1893#discussion_r2576505506.

zcash_primitives/src/transaction/builder.rs
upbqdn: Vid. https://github.com/zcash/librustzcash/pull/1893#discussion_r2574178371.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/0e6b0bc4b2dcf525b1bd7fca47acbcd45cedcce1..87cfafc56048bd5eb137ba6a133df4e671bb3837) to rebase atop #2075

nuttycom (dismissed):
utACK 0dccfa1b2a0782477f95959995947472602ac7e8

zcash_primitives/CHANGELOG.md
str4d: Both `MinerData` (at the time) and `PushValue` (now) can include allocations via `Vec` / `BoundedVec`.

nuttycom (dismissed):
re-utACK cb0e2ea41434b0a9db3986c69d0fb850784609de

str4d (approved):
utACK 4b122638cdcdbaf98f7221434e2b39c1dc8d760c

nuttycom (approved):
re-utACK bd3282c3dda45ea3cbb0b81aad812329d191cdbc

zcash_primitives/CHANGELOG.md
upbqdn: I didn't see this question earlier, sorry for not responding.

PR #1894, opened 2025-08-06
zcash_keys: Minor API tweaks.

nuttycom:
The surface area affected by these tweaks is all in currently-unreleased APIs, so no CHANGELOG entries are necessary.


daira (commented):
Post-hoc ACK

PR #1895, opened 2025-08-06
Merge `zallet_dev` to `main`

nuttycom:


components/zcash_protocol/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ represent the transitive `semver` implications of changes within the enclosing
 workspace.
 
 ## [Unreleased]
+### Added
+- `zcash_protocol::constants::`
+  - `{mainnet, testnet, regtest}::B58_SECRET_KEY_PREFIX`
+
+### Changed
```
nuttycom, line 17: It's pretty inconvenient that this change requires a major version bump to `zcash_protocol`, which then would require the whole release chain. It would be nice to be able to release this as a minor version bump, and for ~all consumers of this library it would appear to be strictly additive, since I don't believe that we're aware of third-party implementations of the `NetworkConstants` trait. 

Something we might want to do in a future release is actually to seal `NetworkConstants` such that third-party implementations are prohibited; then, we could confidently release changes here without breaking SemVer.

daira: Why don't we seal it now? (And take the breaking change now for doing so.)

daira (approved):
utACK

components/zcash_protocol/CHANGELOG.md
nuttycom: I don't want to hold up the Zebra release, and I feel like while a patch release of `zcash_protocol` with the existing change is going to be invisible to essentially all users, sealing the trait is a bigger change which might not be so invisible. If we did it this way, we'd definitely have to take the approach of releasing 0.6.1 and then yanking 0.6.0.

PR #1899, opened 2025-08-08
Add zcash/lightwallet-protocol as a subtree & replace proto files with symlinks to canonical files.

nuttycom:


pacu:
I wonder how this change will impact this PR https://github.com/zcash/librustzcash/pull/1781/files 👀

zcash_client_backend/lightwallet-protocol/walletrpc/compact_formats.proto
```diff
@@ -0,0 +1,84 @@
...
+message CompactTx {
+    // Index and hash will allow the receiver to call out to chain
+    // explorers or other data structures to retrieve more information
+    // about this transaction.
+    uint64 index = 1;   // the index within the full block
+    bytes hash = 2;     // the ID (hash) of this transaction, same as in block explorers
+
+    // The transaction fee: present if server can provide. In the case of a
+    // stateless server and a transaction with transparent inputs, this will be
+    // unset because the calculation requires reference to prior transactions.
```
daira, line 50: Aside (non-blocking since this should be addressed in `lightwallet-protocol`): There is no such thing as "unset" in proto3. It will be zero if it is not explicitly set, which means that a transaction with zero fee &mdash;which is allowed by consensus&mdash; can't be distinguished from one that the server couldn't calculate. (Transactions with zero fee aren't usually relayed, except for coinbase transactions, so maybe this doesn't matter so much in practice.)

zcash_client_backend/lightwallet-protocol/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Electric Coin Company
```
daira, line 3:
```suggestion
Copyright (c) 2019-2025 Electric Coin Company
```
Should be addressed in `lightwalletd-protocol`.

zcash_client_backend/proto/compact_formats.proto
daira, line 1: I assume these have the correct git properties for symlinks even though the GitHub UI doesn't show that.

zcash_client_backend/src/proto/compact_formats.rs
daira, line 1: Should these generated files also be symlinks into generated files in `lightwallet-protocol`? That would make the subtree update process easier, and perhaps more importantly it would ensure consistency between repos that depend on `lightwalletd-protocol` in terms of the version of `protoc` used to generate the Rust bindings. Non-blocking; I can see the arguments for either approach.

zcash_client_backend/src/proto/service.rs
```diff
@@ -151,6 +168,8 @@ pub struct Balance {
     #[prost(int64, tag = "1")]
     pub value_zat: i64,
 }
+/// The a shortened transaction ID is the prefix in big-endian (hex) format
```
daira, line 171: Typo (should be addressed in `lightwalletd-protocol`).

zcash_client_backend/src/proto/service.rs
```diff
@@ -639,6 +688,9 @@ pub mod compact_tx_streamer_client {
         /// more bandwidth-efficient; if two or more transactions in the mempool
         /// match a shortened txid, they are all sent (none is excluded). Transactions
         /// in the exclude list that don't exist in the mempool are ignored.
+        ///
+        /// The a shortened transaction ID is the prefix in big-endian (hex) format
```
daira, line 692: Typo (should be addressed in `lightwalletd-protocol`).

zcash_client_backend/src/proto/compact_formats.rs
nuttycom: No; it's important that each project that imports the protobufs do their own build of the parsing layer.

zcash_client_backend/src/proto/compact_formats.rs
```diff
@@ -79,7 +79,7 @@ pub struct CompactTx {
 /// protocol specification.
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct CompactSaplingSpend {
-    /// Nullifier (see the Zcash protocol specification)
+    /// nullifier (see the Zcash protocol specification)
```
str4d, line 82: These revert part of #1102, which was only partially extracted into the new repository. The history of changes to `compact_formats.proto` (which until now was canonically specified in this repository) should be extracted to the new repo (in the same way that the history of lightwalletd was), which will then undo this diff.

str4d: Done in https://github.com/zcash/lightwallet-protocol/pull/3.

PR #1906, opened 2025-08-13
Use the right prefix for decoding transparent secret keys

daira:


zcash_keys/src/keys/transparent.rs
```diff
@@ -66,7 +66,7 @@ impl Key {
                 .with_check(None)
                 .into_vec()?,
         );
-        let prefix = network.b58_script_address_prefix();
+        let prefix = network.b58_secret_key_prefix();
```
nuttycom, line 69: This was a bad merge resolution in 62089d297e9e1a9753d7f827b7d556113b9229dd

daira: This was [released in zcash_keys 0.10.0](https://github.com/zcash/librustzcash/blob/zcash_keys-0.10.0/zcash_keys/src/keys/transparent.rs#L69) ~~and is a regression~~. Do we need to release a 0.10.1 and yank 0.10.0?

daira: Hotfix done in #1910.

PR #1910, opened 2025-08-14
zcash_keys: Hotfix release for use of incorrect secret key prefix in Base58 encoding

nuttycom:


zcash_keys/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ workspace.
 
 ## [Unreleased]
 
+## [0.10.1] - 2025-08-06
+
+### Fixed
+- Use of an incorrect prefix for transparent secret key DER encoding
```
str4d: It's not the DER encoding that has a bug, it's the Base58 encoding.
```suggestion
- Use of an incorrect prefix for transparent secret key Base58 encoding
```

PR #1917, opened 2025-08-21
feat: EIP-681 transaction request parser

schell:
These changes add a new `eip681` crate, similar to the `zip321` crate, which provides a parser for Ethereum transaction requests according to the [EIP-681](https://eips.ethereum.org/EIPS/eip-681) spec.

### DIFF
--- a/components/eip681/src/error.rs
+++ b/components/eip681/src/error.rs
@@ -116,7 +116,7 @@ pub enum ValidationError {
     /// that denotes a single value.
     #[snafu(display(
         "More than one parameter by the key '{key}', saw: [{}]",
-        values.iter().map(|(k, v)| format!("{k}={}", v.to_string())).collect::<Vec<_>>().join(", ")
+        values.iter().map(|(k, v)| format!("{k}={v}")).collect::<Vec<_>>().join(", ")
     ))]
     MultipleParameterValues {
         key: &'static str,


components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1437 @@
...
+
+/// Parse at least `min` digits.
+pub fn parse_min(i: &str, min: usize, is_hex: bool) -> nom::IResult<&str, Vec<u8>, ParseError<'_>> {
+    let radix = if is_hex { 16 } else { 10 };
+    let (i, chars) = nom::bytes::complete::take_while(|c: char| c.is_digit(radix))(i)?;
+    let data = chars
+        .chars()
+        .map(|c| {
+            c.to_digit(radix)
+                .unwrap_or_else(|| unreachable!("we already checked that this char was a digit"))
```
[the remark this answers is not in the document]
schell: `expect` also panics.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1437 @@
...
+/// 4. only `"-" / "+"` is present
+/// 5. ...others
+///
+/// For this reason, in this library, parsing is separate from validation.
+///
+/// Other implementations use regular expressions instead of parsing, and only
+/// support very specific values.  
+///
+// TODO(schell):
+//   - tell someone about the bugs in the ABNF syntax
```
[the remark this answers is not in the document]
schell: It's definitely actionable.

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = [
+members = [ "components/eip681",
```
str4d:
```suggestion
members = [
    "components/eip681",
```

components/eip681/README.md
```diff
@@ -0,0 +1,22 @@
+# eip681
+
+This library contains Rust parsing functions and data types for working
+with [Ethereum EIP 681 Payment Request
+URIs](https://eips.ethereum.org/EIPS/eip-681).
```
daira:
```suggestion
with [Ethereum EIP 681 Payment Request URIs](https://eips.ethereum.org/EIPS/eip-681).
```
Non-blocking.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+/// *DIGIT
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct Digits {
+    places: Vec<u8>,
+}
+
+impl core::fmt::Display for Digits {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for n in self.places.iter() {
```
daira, line 22:
```suggestion
        for n in self.places {
```
(This is not quite equivalent in general; it calls `.into_iter()` which takes ownership. But in this case it should make no difference.)

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+        let denominator = 10u64.pow(self.places.len() as u32);
+        // Skip the prefixed zeros to build an integer numerator
+        let rest = self
+            .places
+            .iter()
+            .skip_while(|n| **n == 0)
+            .collect::<Vec<_>>();
+        for (mag, digit) in rest.into_iter().rev().enumerate() {
+            numerator += *digit as u64 * 10u64.pow(mag as u32);
+        }
```
daira: Why not process `self.places` without reversing it, and multiply by 10 on each iteration? In that case you don't need to collect it into a `Vec` and don't need to explicitly skip zeroes.
(untested)
```suggestion
        let denominator = 10u64.pow(self.places.len() as u32);
        let mut numerator = 0;
        for digit in self.places {
            numerator = numerator * 10 + (*digit as u64);
        }
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+
+        let mut digits = Digits { places: vec![] };
+        for log in (0..=v.ilog10()).rev() {
+            let top = 10u64.pow(log);
+            let digit = v / top;
+            digits.places.push(digit as u8);
+            v -= digit * top;
+        }
+        digits
```
daira: This is simpler I think (untested):
```suggestion
        let mut places: Vec<u8> = vec![];
        while v > 0 {
            places.push((v % 10) as u8);
            v /= 10;
        }
        places.reverse();
        Digits { places }
```

components/eip681/src/parse.rs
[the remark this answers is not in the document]
daira: Copilot made a correct suggestion (although I prefer your panic message) with the wrong explanation. Using `.expect` is completely equivalent here, modulo a trivial difference in the prefix of the panic message, and preferred just because it's more concise.
```suggestion
                .expect("we already checked that this char was a digit")
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+        DigitsMinimumSnafu {
+            min,
+            digits: data,
+            input: i
+        }
+    );
+    Ok((i, data))
+}
+
+/// Zero or more consecutive hexidecimal digits.
```
daira:
```suggestion
/// Zero or more consecutive hexadecimal digits.
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for n in self.places.iter() {
+            f.write_fmt(format_args!("{n}"))?;
+        }
+        Ok(())
+    }
+}
+
+impl Digits {
+    pub const MAX_PLACES: usize = u64::MAX.ilog10() as usize;
```
daira: I think there's an off-by-one here.
```suggestion
    pub const MAX_PLACES: usize = (u64::MAX.ilog10() as usize) + 1;
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+/// *HEXDIG
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+pub struct HexDigits {
+    places: Vec<u8>,
+}
+
+impl core::fmt::Display for HexDigits {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for n in self.places.iter() {
```
daira, line 115:
```suggestion
        for n in self.places {
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+/// exponent MUST be greater or equal to the number of decimals after the point.
+///
+/// ## Note
+/// This ABNF notation doesn't seem correct, as it allows for quite a few
+/// cases that don't make sense - for example:
+///
+/// 1. `*DIGIT` is missing and `[ "." 1*DIGIT ]` is present while `[ ( "e" / "E" ) [ 1*DIGIT] ]`
+///    is not
+/// 2. `[ ( "e" / "E" ) [ 1*DIGIT] ]` is missing the `[ 1*DIGIT ]` (eg, just "e")
+/// 3. The decimal is present while the exponent is missing
```
daira: That's fairly clearly intended to be an error: the exponent isn't greater or equal to the number of decimals in that case. Also, isn't this a superset of case 1? It's an error regardless of whether `*DIGIT` is empty.

Case 2 seems to be just a bug: it makes no sense to write `[ 1*DIGIT ]` rather than `*DIGIT`. I suspect it was intended to just be `1*DIGIT`.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+/// 4. only `"-" / "+"` is present
```
daira: Since this isn't clearly prohibited, it should probably be interpreted as 0.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1437 @@
...
+
+    /// Parse an `EnsName`.
+    pub fn parse(i: &str) -> nom::IResult<&str, Self, ParseError<'_>> {
+        fn continue_parsing(c: char) -> bool {
+            !c.is_whitespace() && !EnsName::DELIMITERS.contains(&c)
+        }
+
+        let (i, name) = nom::bytes::complete::take_till(|c| !continue_parsing(c))(i)?;
+        snafu::ensure!(!name.is_empty(), EnsMissingSnafu);
+        snafu::ensure!(name.contains('.'), EnsMissingSnafu);
```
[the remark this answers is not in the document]
daira: Copilot has a point. What counts as `EnsDomainSnafu`? I don't see it used anywhere. If a missing domain can't be distinguished from an invalid domain, just delete `EnsDomainSnafu`. Also check whether any other errors are unused.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+    pub fn is_recoverable(&self) -> bool {
+        true
+    }
+}
+
+#[derive(Debug, Snafu, PartialEq)]
+pub enum ValidationError {
+    #[snafu(display("Exponent is too small, expected at least {expected}, saw {seen}"))]
+    SmallExponent { expected: usize, seen: u64 },
+
```
daira:
```suggestion

    #[snafu(display("Exponent is too large, expected at most {expected}, saw {seen}"))]
    LargeExponent { expected: usize, seen: u64 },
        
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+            .map(|is_positive| if is_positive { 1 } else { -1 })
+            .unwrap_or(1i128);
+        let integer = self.integer();
+        let (decimal_numerator, decimal_denominator) = self.decimal();
+        let exp = self
+            .exponent
+            .as_ref()
+            .and_then(|(_, maybe_exp)| maybe_exp.as_ref().map(|digits| digits.as_u64() as u32))
+            .unwrap_or_default();
+        let multiplier = 10i128.saturating_pow(exp);
```
daira:
```suggestion
        let multiplier = 10i128.checked_pow(exp).ok_or_else(|| LargeExponentSnafu {
            expected: u128::MAX.ilog10() as usize,
            seen: exp as u64,
        })?;
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+        snafu::ensure!(
+            modulo == 0 || decimal_numerator == 0,
+            SmallExponentSnafu {
+                expected: decimal_denominator.ilog10() as usize,
+                seen: exp
+            }
+        );
+        let multiplied_integer = (integer as i128).saturating_mul(multiplier);
+        let decimal_multiplier = multiplier / decimal_denominator as i128;
+        let multiplied_decimal = (decimal_numerator as i128).saturating_mul(decimal_multiplier);
```
daira: I recommend using `checked_mul` here as well.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1437 @@
...
+        //
+        // In practice this should be fine because values either end with `&` (in the case of
+        // more parameters after it) or with the end of the input.
+
+        let mut results = [number.parse(i), ethereum_address.parse(i), string.parse(i)];
+        results.sort_by_key(|a| {
+            // Sort by the amount of input left over, smallest first
+            a.as_ref().ok().map(|(i, _)| i.len()).unwrap_or(usize::MAX)
+        });
+        // UNWRAP: safe because this was an array of 3
```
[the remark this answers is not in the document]
daira: Here Copilot just doesn't understand that "this was an array of 3" is more useful than the low-level thing it suggested.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+        results.into_iter().next().unwrap()
```
daira:
```suggestion
        results.into_iter().next().expect("safe because this was an array of 3")
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1449 @@
...
+        // string used as key.
+        match &key {
+            Key::Value | Key::Gas | Key::GasLimit | Key::GasPrice => {
+                snafu::ensure!(
+                    matches!(value, Value::Number(_)),
+                    InvalidParameterValueSnafu
+                );
+            }
+            Key::Type(_ty) => {
+                // For the moment, we're not going to check the type here.
```
daira: Should this be a TODO comment, or is it intended because the type is checked elsewhere?

components/eip681/src/parse.rs
schell: Good catch!

components/eip681/src/parse.rs
schell: That's a nice simplification.

components/eip681/src/parse.rs
schell: 🐑

components/eip681/src/parse.rs
schell: This actually causes a compilation error, as it's trying to move out of `places` as it does implicitly call `.into_iter()`, but `places` is borrowed and so cannot be moved, and `Vec` doesn't implement `Copy`.

components/eip681/src/parse.rs
schell: Yes, this is the case. I've added a sanity test to ensure this.

components/eip681/src/parse.rs
schell: I think a TODO is probably appropriate. The problem is that parsing [arbitrary solidity types](https://docs.soliditylang.org/en/develop/abi-spec.html) is a large amount of work in itself, and it's not clear that we need this at the moment.

schell: Ah, right, the TODO is already in place in the documentation for `EthereumAbiTypeName`. I updated both places with more documentation and comments.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1452 @@
...
+            }
+        );
+        let multiplied_integer = (integer as i128)
+            .checked_mul(multiplier)
+            .with_context(|| SaturationSnafu { seen: self.clone() })?;
+        let decimal_multiplier = multiplier / decimal_denominator as i128;
+        let multiplied_decimal = (decimal_numerator as i128)
+            .checked_mul(decimal_multiplier)
+            .with_context(|| SaturationSnafu { seen: self.clone() })?;
+        let value = multiplied_integer + multiplied_decimal;
```
sam0x17: Should just use checked math here, that way we don't need to restrict using `i128::MAX`, also this restriction doesn't really do much to catch scenarios where we have wrapped around -- could be `i128::MAX - 3` for example and this wouldn't catch that. Should not use wrap around math at all here

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1452 @@
...
+        let (i, name) = nom::bytes::complete::take_till(|c| !continue_parsing(c))(i)?;
+        snafu::ensure!(!name.is_empty(), EnsMissingSnafu);
+        snafu::ensure!(name.contains('.'), EnsDomainSnafu);
+
+        // Now we have our name, normalize
+        let normalized_name = ens_normalize_rs::normalize(name).context(EnsNormalizationSnafu)?;
+
+        Ok((i, EnsName(normalized_name)))
+    }
+}
```
sam0x17, line 390: Few problems here:
1. need to conform to the EIP spec w.r.t how we parse whitespace, and this should be consistent across all parsing we do in this file, not specific to `EnsName`, should at least have a helper we call a bunch of times that does this properly
2. We use `ens_normalize_rs` to normalize the `EnsName`, probably as a cheap way to get parsing errors out of the `EnsName`, but this action is destructive -- ideally if we must use `ens_normalize_rs` to validate an `EnsName`, we call normalize, bubble up its errors, but throw out the result

components/eip681/src/parse.rs
schell: Thanks @sam0x17, I've switched to `checked_*` now. 🙇

components/eip681/src/parse.rs
schell: Thanks for your help!

Can you point me to the EIP spec that dictates how to deal with whitespace here? EIP-681 doesn't seem to mention anything, and I can't find mention of whitespace anywhere down the rabbit hole until https://unicode.org/reports/tr46/, which doesn't go into any detail about what the whitespace _is_. 

I think moving this into a helper at this point is not necessary since this is the only callsite that deals with any whitespace.

Validating with `ens_normalize_rs` is indeed a plot to save developer time. Though now that I'm re-reading the [EIP-137](https://eips.ethereum.org/EIPS/eip-137#name-syntax) section on ENS names it looks like the name we're parsing here should already be normalized, so it is indeed safe to throw out, if it matches the input. Thanks 👍 .

sam0x17: I think @daira noticed the exact place in the spec where it mentioned the whitespace stuff

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1524 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for n in self.places.iter() {
+            f.write_fmt(format_args!("{n}"))?;
+        }
+        Ok(())
+    }
+}
+
+impl Digits {
+    pub const MAX_PLACES: usize = u64::MAX.ilog10() as usize;
```
nuttycom, line 31: It looks like construction of `Digits` should be checked and bounded to this length, to make the casts in `as_u64`, `as_decimal_ratio`, etc, safe. It didn't look to me like `MAX_PLACES` was checked in parsing.

nuttycom (changes requested):

components/eip681/src/parse.rs
schell: No it currently isn't, so this would be a good addition, thanks.

schell: Though this is a limitation of the chosen data types and not suggested by the spec, though I think it makes more sense to add this as a validation error instead of a parse error.

schell: Hrm, though in `as_us64` we now use checked arithmetic to cover the possibility for overflow, which does account for this. So I think a better change would be to use checked arithmetic in `as_decimal_ratio` and have that return `Result`. Then I'll make MAX_PLACES private since it's only used for generating arbitrary proptest values.

schell: Instead of making this private I cfg'd it for test-only.

components/eip681/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "eip681"
+description = "Parsing functions and data types for Ethereum EIP 681 Payment Request URIs"
+version = "0.1.0"
```
str4d: We don't have crates in our namespace with un-patchable versions. We always keep the in-repo code as based on some published version. Given that 0.1.0 will be the first version we want to publish, we should not use it yet:
```suggestion
version = "0.0.0"
```
(In the past I'd have also said to publish an empty 0.0.0, but crates.io prohibits that practice now, so we should just get to 0.1.0 readiness quickly).

supply-chain/audits.toml
```diff
@@ -1820,6 +1867,18 @@ who = "Daira-Emma Hopwood <daira@jacaranda.org>"
 criteria = "safe-to-deploy"
 delta = "1.6.0 -> 1.7.0"
 
+[[trusted.async-trait]]
```
str4d: In our `cargo vet` ecosystem, we only use trust to convey that we trust our own reviews of our own crates. Replace all of these with exemptions.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
+//! Types and functions used for parsing.
+
+use std::borrow::Cow;
+
+use nom::{AsChar, Parser};
+use snafu::prelude::*;
+
+#[derive(Debug, Snafu)]
+pub enum ParseError<'a> {
```
str4d: Style: I'd prefer this to go in a separate `src/parse/error.rs` file (like we tend to do in our other crates) so it doesn't get in the way of reading the parser impl.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+                .checked_mul(10u64.checked_pow(mag as u32).context(OverflowSnafu)?)
+                .context(OverflowSnafu)?;
+            total = total.checked_add(increment).context(OverflowSnafu)?;
+        }
+        Ok(total)
+    }
+
+    /// Returns the digits as a ratio, where prefixed zeros
+    /// are treated as a denominator.
+    pub fn as_decimal_ratio(&self) -> Result<(u64, u64), ValidationError> {
```
str4d: Does this need to be public?
```suggestion
    fn as_decimal_ratio(&self) -> Result<(u64, u64), ValidationError> {
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+    ///
+    /// ## Errors
+    /// Errors if internal arithmetic operations overflow.
+    pub fn integer(&self) -> Result<u64, ValidationError> {
+        self.integer.as_u64()
+    }
+
+    /// Returns the value of the decimal portion of the number as a ratio
+    /// of `u64`s.
+    pub fn decimal(&self) -> Result<(u64, u64), ValidationError> {
```
str4d: Does this need to be public?
```suggestion
    fn decimal(&self) -> Result<(u64, u64), ValidationError> {
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+        Ok(())
+    }
+}
+
+impl HexDigits {
+    /// Parse at least `min` digits.
+    pub fn parse_min(i: &str, min: usize) -> nom::IResult<&str, Self, ParseError<'_>> {
+        let (i, places) = parse_min(i, min, true)?;
+        Ok((i, HexDigits { places }))
+    }
```
str4d, line 127: Parsers like this should be written in the nom combinatorial style, separating grammar-level arguments from the input string:
```suggestion
    pub fn parse_min(min: usize) -> impl Fn(&str) -> nom::IResult<&str, Self, ParseError<'_>> {
        move |i| map(parse_min(min, true), |places| HexDigits { places } ))
    }
```

I also think some of the parsers could be a bit clearer making use of nom's tuple parsers (in particular to avoid some of the `i`-shadowing style that is also common to nom parsers).

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+        ))(i)
+    }
+}
+
+/// A key-value pair.
+#[derive(Debug, PartialEq)]
+pub struct Parameter {
+    key: Key,
+    value: Value,
+}
```
str4d: Instead of having the key-level type safety be on `Key`, I would instead have it be on `Parameter`, because that also allows the value-level type safety.
```suggestion
/// A key-value pair.
///
/// TODO This needs to explain the two classes of parameters.
#[derive(Debug, PartialEq)]
pub enum Parameter {
    /// A token amount.
    ///
    /// If [`EthereumTransactionRequest.function_name`] is `None`, this is denominated in
    /// atomic units of the native token of the blockchain indicated by
    /// [`EthereumTransactionRequest.schema_prefix`] (i.e. Ethereum wei).
    ///
    /// If [`EthereumTransactionRequest.function_name`] is `Some(_)`, this cannot be
    /// interpreted without understanding the function.
    Value(Number),
    /// TODO docs etc below.
    Gas(Number),
    GasLimit(Number),
    GasPrice(Number),
    AbiType {
        name: EthereumAbiTypeName,
        value: Value,
    },
}
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+        match self {
+            Key::Value => f.write_str("value"),
+            Key::Gas => f.write_str("gas"),
+            Key::GasLimit => f.write_str("gasLimit"),
+            Key::GasPrice => f.write_str("gasPrice"),
+            Key::Type(ty) => ty.fmt(f),
+        }
+    }
+}
+
```
str4d: Then `Key` goes away entirely:
```suggestion
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+                Value::parse(i).map_err(|_| ParseError::InvalidParameterValue {
+                    ty: type_name.to_string(),
+                })?
+            }
+        };
+        Ok((i, Parameter { key, value }))
+    }
+}
+
+/// A collection of [`Parameter`].
```
str4d, line 688: Add a bunch more documentation here to explain how these parameters should be used. In particular:
- For function calls, non-[`Parameter::AbiType`] parameters must be filtered out. (This is obviously correct regardless of how the questions below are answered.)
- What happens if e.g. [`Parameter::Value`] is present twice? Which takes precedence?
- What are the semantics for mixed parameters? Is [`Parameter::Value`] allowed alongside [`Parameter::AbiType`]? How are gas params handled?

components/eip681/src/parse.rs
nuttycom:
```suggestion
        for n in &self.places {
```
would work instead but this is trivial, nonblocking.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1533 @@
...
+            .map(|is_positive| if is_positive { 1 } else { -1 })
+            .unwrap_or(1i128);
+        let integer = self.integer()?;
+        let (decimal_numerator, decimal_denominator) = self.decimal()?;
+        let exp = self
+            .exponent
+            .as_ref()
+            .and_then(|(_, maybe_exp)| maybe_exp.as_ref().map(|digits| digits.as_u64()))
+            .transpose()?
+            .unwrap_or_default() as u32;
```
nuttycom: This cast should be replaced with `try_into().map_err(|_| Overflow)` or the like. Also, I dislike having to think through `default`, so I would write this as `.map_or(Ok(0), u32::try_from).map_err(|_| LargeExponentSnafu { .. })?`.

components/eip681/src/parse.rs
schell: We still need `Key` for querying parameters, and it is also helpful for parsing - we can parse `Key` first, then match on that for the value, so I've kept it around.

components/eip681/src/parse.rs
schell: I think the answer to these questions lies downstream of the parser. This library aims to get the values out and then it's up to the user to determine if they're "good" and how to use them.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+
+    /// Convert this [`Number`] into an i128, if possible.
+    ///
+    /// ## Errors
+    /// Errors if internal arithmetic operations overflow.
+    pub fn as_i128(&self) -> Result<i128, ValidationError> {
+        let signum = self
+            .signum
+            .map(|is_positive| if is_positive { 1 } else { -1 })
+            .unwrap_or(1i128);
```
nuttycom:
```suggestion
            .map_or(1i128, |is_positive| if is_positive { 1 } else { -1 });
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+        let integer = self.integer()?;
+        let (decimal_numerator, decimal_denominator) = self.decimal()?;
+        let exp = self
+            .exponent
+            .as_ref()
+            .and_then(|(_, maybe_exp)| maybe_exp.as_ref().map(|digits| digits.as_u64()))
+            .transpose()?
+            .unwrap_or_default()
```
nuttycom: I don't like having to reason about this `unwrap_or_default`. In general, I think that `Default` shouldn't exist, and I prefer that we never use it.
```suggestion
            .unwrap_or(0)
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+    /// The amount to be paid, in the atomic unit of the native token of the blockchain.
+    ///
+    /// In most cases this will denote wei on the ether blockchain, but it depends on
+    /// context outside the scope of this library.
+    Value(Number),
+    /// Gas number value.
+    Gas(Number),
+    /// Synonym for "gas".
+    GasLimit(Number),
+    /// Synonym for "gas".
```
str4d: Bug: `gasPrice` is not (specified as) synonymous with `gas`.
```suggestion
    /// Suggested user-editable value for the gas price of the transaction.
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+pub enum Parameter {
+    /// The amount to be paid, in the atomic unit of the native token of the blockchain.
+    ///
+    /// In most cases this will denote wei on the ether blockchain, but it depends on
+    /// context outside the scope of this library.
+    Value(Number),
+    /// Gas number value.
+    Gas(Number),
+    /// Synonym for "gas".
+    GasLimit(Number),
```
str4d:
```suggestion
    /// Synonym for [`Self::GasLimit`].
    Gas(Number),
    /// Suggested user-editable value for the gas limit of the transaction.
    GasLimit(Number),
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+
+    /// Return the _first_ value of the parameter with the given `key`, if any.
+    pub fn get_value(&self, key: &Key) -> Option<Value> {
+        for param in self.0.iter() {
+            if &param.key() == key {
+                return Some(param.value());
+            }
+        }
+        None
+    }
```
str4d: Per conversation in pairing, we cannot make an assertion about which of multiple is the correct one without outreach to the Ethereum community. In the absence of that, we should only return `Some(_)` if there is precisely one parameter matching the key.

Additionally, I think we need to separate queries for non-ABI parameters from queries for ABI parameters. The latter we don't actually want to query individual parameters; we want the full sequence in order.

So I think what we want is:
```suggestion
    /// Returns the `value` parameter, if there is precisely one.
    ///
    /// Returns `Ok(None)` if it is not present, and `Err(_)` if it is ambiguous.
    pub fn value(&self) -> Result<Option<Number>, Error> {
        todo!()
    }

    /// Returns the `gasLimit` or `gas` parameter, if there is precisely one.
    ///
    /// Returns `Ok(None)` if neither is present, and `Err(_)` if it is ambiguous.
    pub fn gas_limit(&self) -> Result<Option<Number>, Error> {
        todo!()
    }

    /// Returns the `gasPrice` parameter, if there is precisely one.
    ///
    /// Returns `Ok(None)` if it is not present, and `Err(_)` if it is ambiguous.
    pub fn gas_price(&self) -> Result<Option<Number>, Error> {
        todo!()
    }

    /// Returns the sequence of ABI parameters.
    pub fn abi_parameters(&self) -> impl Iterator<Item = (EthereumAbiTypeName, Value)> {
        todo!()
    }
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+            c.is_alphanum() || chars.contains(&c)
+        }
+        let (i, s) = nom::bytes::complete::take_while1(is_type_char)(i)?;
+        Ok((i, EthereumAbiTypeName { name: s.to_owned() }))
+    }
+}
+
+/// The key of a parameter.
+#[derive(Clone, Debug, PartialEq)]
+pub enum Key {
```
str4d: Then we can make this private:
```suggestion
enum Key {
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+            Key::Gas => f.write_str("gas"),
+            Key::GasLimit => f.write_str("gasLimit"),
+            Key::GasPrice => f.write_str("gasPrice"),
+            Key::Type(ty) => ty.fmt(f),
+        }
+    }
+}
+
+impl Key {
+    pub fn parse(i: &str) -> nom::IResult<&str, Self, ParseError<'_>> {
```
str4d:
```suggestion
    fn parse(i: &str) -> nom::IResult<&str, Self, ParseError<'_>> {
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+    /// ```
+    pub fn parse(i: &str) -> nom::IResult<&str, Self, ParseError<'_>> {
+        let mut number = Number::parse.map(Value::Number);
+        let mut ethereum_address = EthereumAddress::parse.map(Value::Address);
+        let mut string = UrlEncodedUnicodeString::parse.map(Value::String);
+
+        // Here there are some interesting corner cases:
+        //
+        // 1. In `number`'s spec, absolutely everything is optional, so it _never_ fails.
+        // 2. `ethereum_address` may be prefixed with `0` in the hex-address case, so both number
```
str4d:
```suggestion
        // 2. `ethereum_address` may be prefixed with `0` (in the hex-address case), so both number
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+        if let Some(head) = maybe_head {
+            let parse_next_param =
+                nom::sequence::preceded(nom::bytes::complete::tag("&"), Parameter::parse);
+            let (i, tail) = nom::multi::many0(parse_next_param)(i)?;
+            let mut parameters = vec![head];
+            parameters.extend(tail);
+            Ok((i, Parameters(parameters)))
+        } else {
+            Ok((i, Parameters(vec![])))
+        }
```
str4d: As I mentioned in our pairing, given the horrors that live inside the `ethereum_address` parser (and in particular its interaction with the `value` parser), I'm inclined to parse this top-down instead of left-to-right. We should first parse `i` as `nom::multi::separated_list0(tag("&"), nom::bytes::complete::is_not("&"))`, and then parse the results of the item parsers with `Parameter::parse` (thus preventing the `value` parser from going beyond the end of a single parameter). I think that looks like this:
```suggestion
        use nom::{bytes::complete::{is_not, tag}, combinator::map_parser, multi::separated_list0};

        separated_list0(tag("&"), map_parser(is_not("&"), Parameter::parse)).parse(i)
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1492 @@
...
+        Ok((
+            i,
+            Self {
+                schema_prefix,
+                target_address,
+                chain_id,
+                function_name,
+                parameters,
+            },
+        ))
```
str4d, line 939: Likewise, I'd prefer to gate the `EthereumAddress::parse` inside an outer parser:

```suggestion
        use nom::{bytes::complete::{is_not, tag}, combinator::{map, map_parser, opt}, sequence::preceded};

        map(
            (
                SchemaPrefix::parse,
                map_parser(
                    is_not("@/?"),
                    EthereumAddress::parse,
                ),
                opt(preceded(tag("@"), |i| Digits::parse_min(i, 1))),
                opt(preceded(tag("/"), UrlEncodedUnicodeString::parse)),
                opt(preceded(tag("?"), Parameters::parse)),
            ),
            |
                schema_prefix,
                target_address,
                chain_id,
                function_name,
                parameters,
            | Self {
                schema_prefix,
                target_address,
                chain_id,
                function_name,
                parameters,
            },
        ).parse(i)
```

components/eip681/src/error.rs
```diff
@@ -0,0 +1,133 @@
+//! Any and all errors encountered during parsing and validation.
+
+use snafu::prelude::*;
+
+use crate::{Number, Value};
+
+/// Errors discovered during parsing.
+#[derive(Debug, Snafu)]
```
str4d, line 8: We've decided to use non-exhaustive enums for errors going forward:
```suggestion
#[derive(Debug, Snafu)]
#[non_exhaustive]
```

components/eip681/src/error.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl ParseError<'_> {
+    pub fn is_recoverable(&self) -> bool {
+        true
+    }
+}
+
+/// Errors discovered after parsing, usually when attempting to convert parsed
+/// types into numeric types.
+#[derive(Debug, Snafu, PartialEq)]
```
str4d, line 93:
```suggestion
#[derive(Debug, Snafu, PartialEq)]
#[non_exhaustive]
```

components/eip681/src/error.rs
```diff
@@ -0,0 +1,133 @@
...
+            nom::Err::Error(value)
+        } else {
+            nom::Err::Failure(value)
+        }
+    }
+}
+
+impl ParseError<'_> {
+    pub fn is_recoverable(&self) -> bool {
+        true
```
str4d: I'm confused as to why this method exists in the public API. If we think that some future `ParseError` will be unrecoverable, then we can just add this method at the time we add that variant.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+        for (mag, digit) in self.places.iter().rev().enumerate() {
+            let increment = (*digit as u64)
+                .checked_mul(
+                    10u64
+                        .checked_pow(u32::try_from(mag).context(IntegerSnafu)?)
+                        .context(OverflowSnafu)?,
+                )
+                .context(OverflowSnafu)?;
+            total = total.checked_add(increment).context(OverflowSnafu)?;
+        }
```
str4d: This is simpler with the same approach you are already using in `Digits::as_decimal_ratio`:
```suggestion
        for digit in &self.places {
            total = total
                .checked_mul(10)
                .context(OverflowSnafu)?
                .checked_add(*digit as u64)
                .context(OverflowSnafu)?;
        }
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+            .context(OverflowSnafu)?;
+        let mut numerator = 0u64;
+
+        for digit in self.places.iter() {
+            numerator = numerator
+                .checked_mul(10)
+                .context(OverflowSnafu)?
+                .checked_add(*digit as u64)
+                .context(OverflowSnafu)?;
+        }
```
str4d: And then it's clear these are identical:
```suggestion
        let numerator = self.as_u64();
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+                        .context(OverflowSnafu)?,
+                )
+                .context(OverflowSnafu)?;
+            total = total.checked_add(increment).context(OverflowSnafu)?;
+        }
+        Ok(total)
+    }
+
+    /// Returns the digits as a ratio, where prefixed zeros
+    /// are treated as a denominator.
```
str4d: This comment is confusing to me, because there is no period character in `Digits`, so `self` is not a ratio.

I *think* what you are saying is that this method returns the ratio representing the decimal number `0.<digits>`, i.e. re-interpreting `self` as the fractional component of some other decimal number.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+                .checked_mul(10)
+                .context(OverflowSnafu)?
+                .checked_add(*digit as u64)
+                .context(OverflowSnafu)?;
+        }
+
+        Ok((numerator, denominator))
+    }
+
+    pub fn from_u64(mut v: u64) -> Self {
```
str4d: Document this method.

Style nit: I prefer constructors to be at the start of `impl` blocks (near the definition of the struct) so it's easier to check that constructors uphold any necessary invariants. I then tend to put conversions straight afterwards for symmetry, so in this case putting `from_u64` above `as_u64` would satisfy these both.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+                exponent,
+            },
+        ))
+    }
+
+    /// Returns the value of the integer portion of the number.
+    ///
+    /// ## Errors
+    /// Errors if internal arithmetic operations overflow.
+    pub fn integer(&self) -> Result<u64, ValidationError> {
```
str4d: This API can be confused with `Number::as_i128` if you don't carefully read the documentation. I also don't see any case where this would be useful externally, so we should just keep it private:
```suggestion
    fn integer(&self) -> Result<u64, ValidationError> {
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+            .checked_rem(decimal_denominator as i128)
+            .context(OverflowSnafu)?;
+        snafu::ensure!(
+            modulo == 0 || decimal_numerator == 0,
+            SmallExponentSnafu {
+                expected: decimal_denominator.ilog10() as usize,
+                seen: exp
+            }
+        );
+        let multiplied_integer = (integer as i128)
```
str4d, line 299: To help clarify what is going on:
```suggestion
        // signum * (
        //     (integer * multiplier) +
        //     (decimal_numerator * (multiplier / decimal_denominator))
        // )
        let multiplied_integer = (integer as i128)
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+            .checked_mul(multiplier)
+            .with_context(|| SaturationSnafu { seen: self.clone() })?;
+        let decimal_multiplier = multiplier
+            .checked_div(decimal_denominator as i128)
+            .context(OverflowSnafu)?;
```
str4d: Expanding this out:
```
let decimal_multiplier = 10^exp / 10^(self.decimal.places.len());
```

This would be easier to understand, simpler to prove avoids computational edge cases (by avoiding the sole use of `div`), and closer to the spec, if implemented as:
```
let decimal_multiplier = 10^(exp.checked_sub(self.decimal.places.len()));
```

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+                _ => None,
+            })
+            .collect::<Vec<_>>();
+        snafu::ensure!(
+            values.len() <= 1,
+            MultipleParameterValuesSnafu {
+                key: "gasLimit",
+                values: values.into_iter().map(Value::Number).collect::<Vec<_>>(),
+            }
+        );
```
str4d, line 793: Non-blocking nit: this is over-constraining the parser. If both `gas` and `gasLimit` are set, it is still unambiguous as long as they are both set to the same value.

If `values` were collected into a `HashSet` instead of a `Vec` (and similarly for `value()` and `gas_price()`, that would I think be sufficient (and also allow `gas=X&gas=X` etc, which should be fine as there is no sensible array interpretation of these non-ABI fields).

supply-chain/audits.toml
```diff
@@ -18,6 +18,12 @@ who = "Daira-Emma Hopwood <daira@jacaranda.org>"
 criteria = "safe-to-deploy"
 delta = "1.0.82 -> 1.0.83"
 
+[[audits.anyhow]]
+who = "Schell Carl Scivally <efsubenovex@gmail.com>"
+criteria = "safe-to-deploy"
+delta = "1.0.97 -> 1.0.100"
+notes = "Contains mostly configuration for no_std support and conditionally generating backtraces."
```
str4d: This update includes changes to the build script, changes to existing `unsafe` blocks, and new `unsafe` blocks. In our `cargo-vet` audits we always declare when these kinds of changes occur (because they are directly relevant to the audit criteria), and what they are doing (as evidence that we fully understand the changes, as required by the audit criteria).

In this case, I would likely have skipped auditing the crate because understanding it to the necessary degree would have taken me too long, which I know because I've just gone and done so. My audit of this crate is:
```suggestion
who = "Jack Grigg <jack@electriccoin.co>"
criteria = "safe-to-deploy"
delta = "1.0.97 -> 1.0.100"
notes = """
Changes to the build script are to enable reusing the probe file within the crate, enable
more features on newer Rust versions, and catch more errors.

Most of the changes to `unsafe` code are refactors to surrounding safe code, and not
actually altering the `unsafe` blocks. The new instances of `unsafe` are to allow errors
to be converted to `Box<dyn std::error:Error>` in a way that supports downcasting (at the
cost of backtraces), via a new vtable method. The `unsafe` blocks in the new code appear
to match how boxing and downcasting logic is done in other existing vtable methods.
"""
```

supply-chain/audits.toml
```diff
@@ -173,16 +179,34 @@ who = "Jack Grigg <jack@electriccoin.co>"
 criteria = "safe-to-deploy"
 delta = "0.20.9 -> 0.20.10"
 
+[[audits.darling]]
+who = "Schell Carl Scivally <efsubenovex@gmail.com>"
+criteria = "safe-to-deploy"
+delta = "0.20.10 -> 0.21.3"
+notes = "Mostly added tests and documentation. The bulk of the changes were made to `darling_core`."
```
str4d, line 195: By contrast, I would likely not have written any notes for this audit, because I do not audit test code or documentation (as neither affect our codebase), and dependency updates are visible in lockfile changes so they can't go unnoticed by others. But I don't mind notes being left in these cases.

supply-chain/audits.toml
```diff
@@ -785,6 +809,17 @@ who = "Jack Grigg <jack@electriccoin.co>"
 criteria = "safe-to-deploy"
 delta = "0.4.8 -> 0.6.0"
 
+[[audits.schemars]]
+who = "Schell Carl Scivally <efsubenovex@gmail.com>"
+criteria = "safe-to-deploy"
+delta = "0.8.12 -> 0.9.0"
+
```
str4d, line 830: There is a lot of codegen in this crate, and the `safe-to-deploy` criteria requires auditing not just for usage of `unsafe`, but also powerful imports, and confirming that generated code won't contain problematic usage of either. Given that this update has 3644 additions and 2834 removals, I am very reluctant to accept an audit with no notes.

In particular, in a brief examination myself I noted it lowers the restriction from `forbid(unsafe_code)` to `deny(unsafe_code)`, which enables it to be overridden anywhere within the crate via `allow(unsafe_code)`. That change at a minimum should be noted, even if this update does not make use of it.

Either provide notes about your audit, or remove it and replace with an exemption:
```suggestion
```

supply-chain/audits.toml
```diff
@@ -785,6 +809,17 @@ who = "Jack Grigg <jack@electriccoin.co>"
...
+[[audits.schemars]]
+who = "Schell Carl Scivally <efsubenovex@gmail.com>"
+criteria = "safe-to-deploy"
+delta = "0.9.0 -> 1.1.0"
+notes = "These changes improve security by forbidding unsafe code."
```
str4d: As I noted above, this is actually reverting a change from a previous update that weakened the `unsafe_code` guarantee.

What were the other changes to generated code in this update?

str4d (changes requested):
Reviewed fd77aa70b04a3dd48d6cc10b6566843bf0bf4914.

components/eip681/src/parse.rs
str4d: This has not been resolved.

components/eip681/src/parse.rs
schell: You're right, though it would have to be a `BTreeSet` as we have to preserve ordering.

components/eip681/src/error.rs
```diff
@@ -0,0 +1,133 @@
...
+#[snafu(visibility(pub(crate)))]
+pub enum ValidationError {
+    #[snafu(display("Exponent is too small, expected at least {expected}, saw {seen}"))]
+    SmallExponent { expected: usize, seen: u64 },
+
+    #[snafu(display("Exponent is too large, expected at most {expected}, saw {seen}"))]
+    LargeExponent { expected: usize, seen: u64 },
+
+    #[snafu(display("Number {seen} is too big and has saturated"))]
+    Saturation { seen: Number },
```
daira: Using saturation to detect overflow is an implementation detail (and usually It is preferable to use `checked` arithmetic operations instead). It should be sufficient to just have the `Overflow` variant.

components/eip681/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+//! target_address   = ethereum_address
+//! chain_id         = 1*DIGIT
+//! function_name    = STRING
+//! ethereum_address = ( "0x" 40*HEXDIG ) / ENS_NAME
+//! parameters       = parameter *( "&" parameter )
+//! parameter        = key "=" value
+//! key              = "value" / "gas" / "gasLimit" / "gasPrice" / TYPE
+//! value            = number / ethereum_address / STRING
+//! number           = [ "-" / "+" ] *DIGIT [ "." 1*DIGIT ] [ ( "e" / "E" ) [ 1*DIGIT ] ]
+//! ```
```
daira, line 19: It would be useful to include the definitions of `DIGIT`, `STRING`, `ENS_NAME`, and `TYPE`. For any of these that can't reasonably be defined precisely in ABNF, it's sufficient here to give a covering grammar and an informally stated restriction on it.

components/eip681/src/parse.rs
```diff
@@ -0,0 +1,1565 @@
...
+
+use nom::{AsChar, Parser};
+use snafu::{OptionExt, ResultExt};
+
+use crate::error::*;
+
+/// Zero or more consecutive digits.
+///
+/// ```abnf
+/// *DIGIT
```
daira, line 13: Please repeat the definition of `DIGIT` here (and similarly for other productions in this file that use terminals), since it isn't at all obvious.

components/eip681/src/parse.rs
daira:
```suggestion
    /// Returns the ratio corresponding to the decimal number `0.<digits>`,
    /// i.e. the denominator will be `10^len(digits)`.
```

components/eip681/src/parse.rs
schell: My comment was made out of context of the rest of the function. `HashSet` is fine 👍.

components/eip681/src/lib.rs
schell: `DIGIT` is a core rule in ABNF so I think it's fine to omit that, right? We can assume that people understand the basics of ABNF, otherwise the spec itself wouldn't make sense.

schell: I wrote some extra documentation to provide links to the core rules.

components/eip681/src/parse.rs
```diff
@@ -731,68 +731,72 @@ impl Parameters {
...
     /// ## Errors
-    /// Errors if there are more than one parameter with the key "gasPrice".
+    /// Errors if there are more than one parameter with the key "gasPrice"
+    /// or "gas", and the values of those parameters are not equal.
     pub fn gas_price(&self) -> Result<Option<Number>, ValidationError> {
-        let mut values = self
+        let values = self
             .iter()
             .filter_map(|p| match p {
+                Parameter::Gas(n) => Some(n.clone()),
```
str4d: Bug: 97b7b82e45f15551afa3921a3ba4a1a179661dbc moved `gas` from `gas_limit()` to `gas_price()`, but `gas` is synonymous with `gas_limit`, not `gas_price`. Fix the commit to not make this change (both in code and docs).

components/eip681/src/parse.rs
```diff
@@ -731,68 +731,72 @@ impl Parameters {
...
+        let values = self
             .iter()
             .filter_map(|p| match p {
                 Parameter::Value(n) => Some(n.clone()),
                 _ => None,
             })
-            .collect::<Vec<_>>();
+            .collect::<std::collections::HashSet<_>>();
         snafu::ensure!(
             values.len() <= 1,
```
str4d, line 756: This check now relies on the structural equality from `Eq` (which is necessary for `HashSet` to work), not semantic equality (equivalent encodings of the same numeric value). I think we want to use the latter for determining whether two `value`s are the same, but if it's too complex to implement then I'm okay with this being instead a documented limitation of the method.

components/eip681/src/parse.rs
```diff
@@ -731,68 +731,72 @@ impl Parameters {
...
             .iter()
             .filter_map(|p| match p {
-                Parameter::Gas(n) => Some(n.clone()),
                 Parameter::GasLimit(n) => Some(n.clone()),
                 _ => None,
             })
-            .collect::<Vec<_>>();
+            .collect::<std::collections::HashSet<_>>();
         snafu::ensure!(
             values.len() <= 1,
```
str4d, line 785: Ditto.

schell: All three of these parameter query functions now collect the parameter key+values by semantic equality in a `BTreeMap` to preserve deterministic order.

str4d: Why does ordering matter here?
- Ordering for re-encoding purposes is handled by `Parameters` wrapping a `Vec<_>`.
- We only return something from these three methods if there is at most a single entry in the local `values` variable, which implies order is irrelevant.

components/eip681/src/parse.rs
schell: I'm not sure why github isn't marking this outdated, but this is now resolved.

components/eip681/src/parse.rs
schell: But we also return errors that include _all_ the aliased keys and values. It's possible downstream users write tests that expect errors to be consistent between runs.

str4d: I'd argue that those tests would be incorrectly written, but if that's the reason for the determinism then I guess it's fine.

daira (commented):
Post-hoc ACK changes since my last review.

PR #1923, opened 2025-08-22
Add `zcash_client_memory` crate.

nuttycom:
This builds upon and subsumes https://github.com/zcash/librustzcash/pull/1634

The objective of this PR is to get the contents of #1634 merged so that that PR will stop bitrotting; it does so at the cost of most of the `zcash_client_memory` tests having to be ignored because the merge from `main` breaks them; they can be fixed in follow-up PRs.


Cargo.lock
```diff
@@ -6418,6 +6464,50 @@ dependencies = [
...
+ "serde_json",
+ "shardtree",
+ "static_assertions",
+ "subtle",
+ "thiserror 1.0.63",
+ "time",
+ "tokio",
+ "tracing",
+ "wasm_sync",
+ "which 6.0.3",
```
daira: This should be updated to avoid introducing the duplicate `which`.

zcash_client_memory/Cargo.toml
```diff
@@ -0,0 +1,101 @@
...
+static_assertions.workspace = true
+
+# - Shielded protocols
+orchard = { workspace = true, optional = true }
+sapling.workspace = true
+
+# - Note commitment trees
+incrementalmerkletree.workspace = true
+shardtree = { workspace = true, features = ["legacy-api"] }
+thiserror = "1.0.61"
```
daira, line 58: Why such a specific version?
```suggestion
thiserror = "1"
```

Cargo.toml
```diff
@@ -10,6 +10,7 @@ members = [
     "zcash",
     "zcash_client_backend",
     "zcash_client_sqlite",
+    "zcash_client_memory",
```
str4d, line 12: Nit: this isn't sorted like the rest of the array is.

zcash_client_backend/Cargo.toml
```diff
@@ -262,6 +264,11 @@ unstable-serialization = ["dep:byteorder"]
 ## Exposes the [`data_api::scanning::spanning_tree`] module.
 unstable-spanning-tree = []
 
+## feature to allow building for wasm using wasm-bindgen
+wasm-bindgen = [
+    "time/wasm-bindgen",
+]
+
```
str4d: Unaddressed from my previous review (https://github.com/zcash/librustzcash/pull/1634#discussion_r1953808479):

Is this necessary or only a convenience? AFAICT downstream WASM crates can themselves add a dependency on time with the necessary feature flag enabled, without us needing to re-export the feature flag. I'm reluctant to re-export uncommon feature flags from crates we do not control.
```suggestion
```

zcash_client_backend/src/data_api/chain.rs
```diff
@@ -499,7 +499,7 @@ impl ScanSummary {
 }
 
 /// The final note commitment tree state for each shielded pool, as of a particular block height.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ChainState {
```
str4d, line 503: Unaddressed from my previous review (https://github.com/zcash/librustzcash/pull/1634#discussion_r1953825361):

This is one of those kinds of structs where I'm not entirely convinced that full struct comparison is what we want, and instead `block_hash` equivalence should be the appropriate proxy. Technically all other fields in the struct are committed to by `block_hash`; if `a.block_hash == b.block_hash` but the rest of their fields do not match, that means (except with cryptographically negligible probability) that either `ChainState::empty` or `ChainState::new` was called with invalid data.

But the real question here is, for what purpose is `ChainState` equality needed? I'm guessing the motivation here is a `HashMap` or something, and there we care about plain struct equality, for which the derive is fine.

In [the example from the stdlib docs](https://doc.rust-lang.org/stable/std/cmp/trait.PartialEq.html#how-can-i-implement-partialeq), `PartialEq` is used in a way that means "I do not care whether the other fields match or not", whereas for our environment we _do_ care whether the other fields match. We don't want two chain states with the same block hash but different final tree states to be treated interchangeably (one of them is invalid!) but we also don't want to end up with the same block hash occurring in multiple places (i.e. keeping around invalid state alongside potentially-valid state).

All this is to say, this is why I tend to avoid adding `PartialEq` to structs unless we know we want to use them in that way. So my question here is, how are you using this `ChainState`?

zcash_client_backend/src/data_api.rs
```diff
@@ -2351,6 +2373,11 @@ impl AccountBirthday {
         self.recover_until
     }
 
+    /// Returns the [`ChainState`] corresponding to the last block prior to the wallet's birthday
+    pub fn prior_chain_state(&self) -> &ChainState {
```
str4d, line 2383: Unaddressed from my previous review (https://github.com/zcash/librustzcash/pull/1634#discussion_r1953829633):

Why does this need to be exposed?

zcash_keys/src/address.rs
```diff
@@ -176,7 +176,7 @@ impl UnifiedAddress {
         &self.unknown
     }
 
-    fn to_address(&self, net: NetworkType) -> ZcashAddress {
+    pub fn to_zcash_address(&self, net: NetworkType) -> ZcashAddress {
```
str4d, line 180: Unaddressed from my previous review (https://github.com/zcash/librustzcash/pull/1634#discussion_r1953834241):

- Needs documentation now that this is in the public API.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -16,6 +8,13 @@ use rand::{CryptoRng, Rng, RngCore, SeedableRng};
 use rand_chacha::ChaChaRng;
 use secrecy::{ExposeSecret, Secret, SecretVec};
 use shardtree::{error::ShardTreeError, store::memory::MemoryShardStore, ShardTree};
+use std::{
+    collections::{BTreeMap, HashMap},
+    convert::Infallible,
+    fmt,
+    hash::Hash,
+    num::NonZeroU32,
+};
```
str4d: We do not combine stdlib imports with third-party dependency imports. Undo this move.

zcash_client_memory/src/proto/memory_wallet.proto
str4d, line 1: `.proto` files should not be checked in under `zcash_client_memory/src/`. They should instead be under `zcash_client_memory/proto/`  (like we do in `zcash_client_backend`).

Required but non-blocking (can be done in a follow-up PR instead).

zcash_client_backend/src/data_api.rs
```diff
@@ -245,6 +245,25 @@ impl Balance {
...
+                .ok_or(BalanceError::Overflow)?,
+            value_pending_spendability: (self.value_pending_spendability
+                + rhs.value_pending_spendability)
+                .ok_or(BalanceError::Overflow)?,
+            uneconomic_value: (self.uneconomic_value + rhs.uneconomic_value)
+                .ok_or(BalanceError::Overflow)?,
+        })
+    }
+}
+
```
str4d: This is new since my last review. @nuttycom where does this get used, and why wasn't it needed before?

Also, bug: there is no verification that the total post-addition will not overflow. The existing constructors / mutators perform sufficient validation to ensure that `Balance::total` won't panic, but this does not.

zcash_client_backend/src/data_api.rs
```diff
@@ -2260,7 +2282,7 @@ impl<AccountId> SentTransactionOutput<AccountId> {
 
 /// A data structure used to set the birthday height for an account, and ensure that the initial
 /// note commitment tree state is recorded at that height.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct AccountBirthday {
```
str4d, line 2292: I'm guessing that the reason `ChainState` gains these derives is solely because of this. Where is this comparison needed, and how do my previous review comments affect this derive impl (or should we be using a manual impl here)?

zcash_client_backend/src/data_api.rs
```diff
@@ -1806,6 +1824,10 @@ pub trait WalletTest: InputSource + WalletRead {
         &self,
         protocol: ShieldedProtocol,
     ) -> Result<Vec<ReceivedNote<Self::NoteRef, Note>>, <Self as InputSource>::Error>;
+
+    /// Optionally perform final checks at the conclusion of each test
+    /// Allows wallet backend developers to perform any necessary consistency checks or cleanup
+    fn finally(&self) {}
```
```suggestion
    /// Performs final checks at the conclusion of each test.
    ///
    /// This method allows wallet backend developers to perform any necessary consistency
    /// checks or cleanup. By default it does nothing.
    fn finally(&self) {}
```

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1342,10 +1347,42 @@ pub fn single_output_change_strategy<DbT: InputSource>(
 fn check_proposal_serialization_roundtrip<DbT: InputSource>(
     wallet_data: &DbT,
     proposal: &Proposal<StandardFeeRule, DbT::NoteRef>,
-) {
+) where
+    DbT::NoteRef: std::fmt::Debug,
+{
     let proposal_proto = crate::proto::proposal::Proposal::from_standard_proposal(proposal);
-    let deserialized_proposal = proposal_proto.try_into_standard_proposal(wallet_data);
-    assert_matches!(deserialized_proposal, Ok(r) if &r == proposal);
```
str4d: @nuttycom why did this need to change to only verify the first step in the proposal, rather than ensuring that the entire proposal round-trips correctly? This seems like a regression in test coverage.

zcash_client_backend/src/wallet.rs
```diff
@@ -67,7 +67,7 @@ impl NoteId {
 /// * for wallet-internal outputs, the internal account ID and metadata about the note.
 /// * if the `transparent-inputs` feature is enabled, for ephemeral transparent outputs, the
 ///   internal account ID and metadata about the outpoint;
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Recipient<AccountId> {
```
str4d, line 71: How is equality used for this type? Is the automatic derive correct for the desired usage, or should we only be checking equality on the recipient part of `Recipient`, and not the outpoint / note metadata? It could be rather unintuitive if a "recipient" type results in the same recipient being present twice in some map; whatever is decided here needs to be documented.

zcash_transparent/src/builder.rs
```diff
@@ -459,7 +459,7 @@ impl Bundle<Unauthorized> {
         self,
         calculate_sighash: F,
         secp_ctx: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
-    ) -> Result<TransparentSignatureContext<secp256k1::VerifyOnly>, Error>
+    ) -> Result<TransparentSignatureContext<'_, secp256k1::VerifyOnly>, Error>
```
str4d: @nuttycom why is this changing here? Is this an unrelated beta clippy lint that you're fixing?

zcash_client_memory/build.rs
```diff
@@ -0,0 +1,41 @@
+use std::env;
+use std::fs;
+use std::io;
+use std::path::PathBuf;
+
+fn main() -> io::Result<()> {
+    // - We check for the existence of protoc in the same way as prost_build, so that
+    //   people building from source do not need to have protoc installed. If they make
+    //   changes to the proto files, the discrepancy will be caught by CI.
+    if env::var_os("PROTOC")
```
str4d: This crate should be updated to do the same thing we do in `zcash_client_backend` (not include the `.proto` files in releases so that downstreams don't regenerate the bindings even if protoc is present).

Required but non-blocking (can be done in a follow-up PR instead).

zcash_client_memory/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+mod block_source;
```
str4d, line 1: Crate is missing documentation.

zcash_client_memory/src/types/mod.rs
str4d, line 1: Nit: we rarely use `foo/mod.rs` files, instead placing them at `foo.rs`.

Non-blocking, can be fixed in a follow-up PR.

zcash_client_memory/Cargo.toml
```diff
@@ -0,0 +1,100 @@
+[package]
+name = "zcash_client_memory"
+version = "0.1.0"
```
str4d: This is unreleased (and thus the changes in this PR are not a diff on top of an already-existing 0.1.0 release), so per our repository conventions this should be:
```suggestion
version = "0.0.0"
```

zcash_client_memory/Cargo.toml
```diff
@@ -0,0 +1,100 @@
...
+zcash_keys = { workspace = true, features = ["sapling"] }
+zcash_primitives.workspace = true
+zcash_protocol.workspace = true
+transparent.workspace = true
+zip32.workspace = true
+
+tokio = { workspace = true, features = ["sync"] }
+
+# Dependencies exposed in a public API:
+# (Breaking upgrades to these require a breaking upgrade to this crate.)
```
str4d, line 28: I have not verified that any of these dependency categorisations are accurate for this crate. Doing so is required but non-blocking (as if this is left unverified for too long after the PR merges, it will become a maintenance hazard).

zcash_client_memory/src/types/block.rs
```diff
@@ -0,0 +1,154 @@
...
+use zcash_protocol::memo::MemoBytes;
+
+/// Internal wallet representation of a Block.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct MemoryWalletBlock {
+    pub(crate) height: BlockHeight,
+    pub(crate) hash: BlockHash,
+    pub(crate) block_time: u32,
+    // Just the transactions that involve an account in this wallet
+    pub(crate) _transactions: HashSet<TxId>,
```
str4d, line 17: This is a problematic representation, because it loses transaction index information (i.e. _where_ in the block each transaction is). I guess that is currently not relevant for the in-memory wallet, but I did see transaction indices used in other places, so I presume that means the information gets stored elsewhere?

zcash_client_memory/src/proto/memory_wallet.proto
```diff
@@ -0,0 +1,186 @@
...
+  bytes block_hash = 2;
+  bytes final_sapling_tree = 3;
+  bytes final_orchard_tree = 4;
+}
+
+message WalletBlock {
+  uint32 height = 1; // the height of this block
+  bytes hash = 2; // the ID (hash) of this block, same as in block explorers
+  uint32 block_time = 3; // Unix epoch time when the block was mined
+  repeated bytes transactions = 4; // the txids of transactions in this block
```
str4d: This is the protobuf field corresponding to my previous comment. There is only implicit mention that these might only be wallet-relevant transactions, and there is no documentation that this is a set (and thus the protobuf array order must be strictly ignored).

zcash_client_memory/src/types/block.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub(crate) _memos: HashMap<NoteId, MemoBytes>,
```
str4d, line 18: This is an odd representation given that `_transactions` doesn't keep the transaction index information, but I guess it's fine.

zcash_transparent/src/builder.rs
nuttycom: Yes, this is a beta lint that I've had to fix in every branch I've worked on lately.

zcash_client_backend/src/data_api/chain.rs
str4d: If the `PartialEq, Eq` is only required for test comparisions, then instead do:
```rust
#[derive(Debug, Clone)]
#[cfg_attr(test, derive(PartialEq, Eq)]
pub struct ChainState {
```
and then I will accept it without question (as it is then not part of the public API, and a direct structural equality is fine).

zcash_client_backend/src/data_api.rs
str4d: Ditto my other comment; I would accept
```rust
#[derive(Clone, Debug)]
#[cfg_attr(test, derive(PartialEq, Eq)]
pub struct AccountBirthday {
```
if this is only used for test comparisons.

zcash_client_memory/src/types/notes/received.rs
```diff
@@ -0,0 +1,398 @@
...
+                    note.note_id,
+                    note.txid(),
+                    note.output_index.try_into().unwrap(), // this overflow can never happen or else the chain is broken
+                    inner.clone(),
+                    note.recipient_key_scope
+                        .ok_or(Error::Missing("recipient key scope".into()))?,
+                    note.commitment_tree_position
+                        .ok_or(Error::Missing("commitment tree position".into()))?,
+                    None,
+                    None,
```
nuttycom, line 313: Bug, this breaks proposal round-tripping.

zcash_client_memory/src/types/notes/received.rs
```diff
@@ -0,0 +1,398 @@
...
+                    note.note_id,
+                    note.txid(),
+                    note.output_index.try_into().unwrap(), // this overflow can never happen or else the chain is broken
+                    *inner,
+                    note.recipient_key_scope
+                        .ok_or(Error::Missing("recipient key scope".into()))?,
+                    note.commitment_tree_position
+                        .ok_or(Error::Missing("commitment tree position".into()))?,
+                    None,
+                    None,
```
nuttycom, line 336: Bug, this breaks proposal round-tripping.

zcash_client_memory/src/types/notes/received.rs
```diff
@@ -0,0 +1,398 @@
...
+    fn from(value: ReceivedNote) -> Self {
+        zcash_client_backend::wallet::ReceivedNote::from_parts(
+            value.note_id,
+            value.txid,
+            value.output_index.try_into().unwrap(),
+            value.note,
+            value.recipient_key_scope.unwrap(),
+            value.commitment_tree_position.unwrap(),
+            None,
+            None,
```
nuttycom, line 190: Bug, this breaks proposal round-tripping.

zcash_client_backend/src/data_api.rs
nuttycom: So that it can be persisted in the in-memory wallet backend. It's a little odd in any case to pass a value of this type in the constructor and then not be able to get it back.

zcash_client_backend/src/data_api/chain.rs
nuttycom: From my perspective, structural equality is well-defined for this type, and there is no good reason *not* to include it in our API.

zcash_client_backend/src/data_api/testing.rs
nuttycom: The commit that Pacu added that made this change is being removed (as it didn't actually fix the broken tests.)

zcash_client_backend/src/data_api.rs
nuttycom: I also think that this is a reasonable type on which to define equaility.

zcash_client_backend/src/data_api/testing.rs
nuttycom: This was a temporary change that I made in order to make it clear where the failure was occurring; the debug printing was eliding the parts of the objects that were actually not matching. This commit is being removed.

zcash_client_backend/src/wallet.rs
nuttycom: This type should be named `OutputRecipient`. I will fix this in a follow-up PR.

.github/workflows/ci.yml
```diff
@@ -209,7 +209,6 @@ jobs:
           --workspace
           ${{ steps.prepare.outputs.feature-flags }}
           --features expensive-tests
-          -- --ignored
```
daira, line 212: There are some serialization roundtrip tests in `zcash_primitives/src/transaction/tests.rs` that are marked both `#[ignore]` and `#[cfg(feature = "expensive-tests")]`. They pass and run relatively quickly, so I don't know what the intent was there. Perhaps both `#[ignore]` and `#[cfg(feature = "expensive-tests")]` attributes should be removed from them.

nuttycom: The commit that introduces this change removes the `ignore` from those tests. But I agree, I don't know why these are considered expensive.

zcash_client_memory/Cargo.toml
```diff
@@ -0,0 +1,100 @@
...
+name = "zcash_client_memory"
+version = "0.0.0"
+repository.workspace = true
+# readme = "README.md"
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+categories.workspace = true
+build = "build.rs"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
str4d: The change to `.proto` file handling was incomplete.
```suggestion

# Exclude proto files so crates.io consumers don't need protoc.
exclude = ["*.proto"]
```

zcash_client_memory/build.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::path::PathBuf;
+
+const MEMORY_WALLET_PROTO: &str = "proto/memory_wallet.proto";
+
+fn main() -> io::Result<()> {
+    // - We don't include the proto files in releases so that downstreams do not need to
+    //  regenerate the bindings even if protoc is present.
```
str4d, line 11: This is the comment that was added without making the change to actually exclude the proto files in releases.

zcash_client_memory/build.rs
```diff
@@ -0,0 +1,47 @@
...
+            "proto/notes.proto",
+            "proto/primitives.proto",
+            "proto/shardtree.proto",
+            "proto/transparent.proto",
+        ],
+        &["proto/"],
+    )?;
+
+    // Copy the generated types into the source tree so changes can be committed.
+    fs::copy(out.join("memwallet.rs"), "src/proto/generated.rs")?;
```
str4d: The change to `.proto` file handling was incomplete.
```suggestion
    fs::copy(out.join("memwallet.rs"), "src/proto/memwallet.rs")?;
```

zcash_client_memory/src/proto.rs
```diff
@@ -0,0 +1,2 @@
+pub mod generated;
+pub use generated as memwallet;
```
str4d: The change to `.proto` file handling was incomplete.
```suggestion
#[rustfmt::skip]
#[allow(unknown_lints)]
#[allow(clippy::derive_partial_eq_without_eq)]
pub mod memwallet;
```

zcash_client_memory/src/proto/generated.rs
str4d, line 1: The change to `.proto` file handling was incomplete.
- Delete this file and check in the generated `memwallet.rs` instead.
- Edit `.gitattributes` in the root to add the line
  ```
  zcash_client_memory/src/proto/memwallet.rs linguist-generated=true
  ```

zcash_client_backend/src/data_api/chain.rs
str4d: I don't disagree with including structural equality in our API. I disagree with your assertion that `PartialEq, Eq` _is_ structural equality.

zcash_client_memory/Cargo.toml
nuttycom: Thanks, I wasn't aware that this required an explicit exclude; I thought that it just had to be outside the source tree.

str4d (approved):
utACK e8123c6dd17501b0d48880d3fc2cdc15e9020675

PR #1925, opened 2025-08-28
zcash_client_backend: Improve ephemeral address metadata types in test code.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -1862,7 +1862,8 @@ pub trait WalletTest: InputSource + WalletRead {
 pub struct OutputOfSentTx {
     value: Zatoshis,
     external_recipient: Option<Address>,
-    ephemeral_address: Option<(Address, u32)>,
+    #[cfg(feature = "transparent-inputs")]
+    ephemeral_address: Option<(Address, NonHardenedChildIndex)>,
```
daira, line 1866: Change log entries required.

daira (approved):
utACK but needs change log entries.

PR #1929, opened 2025-08-30
zcash_client_backend: Remove misleading default trait method implementations.

nuttycom:
In the past, when adding trait methods that were guarded by feature flags, we added default implementations for those trait methods that returned empty results. However, if these default implementations end up inadvertently used in scenarios where those methods end up being called, they may result in misleading results being silently returned to the user. This change replaces those default implementations with `unimplemented!()` macro calls that will panic in the event that such a default implementation is inadvertently used.


nuttycom:
> Why not remove the default implementations instead? Wouldn’t that result in a compile time error to implement those methods, rather than a run-time panic? I would think removing the default impls and using `unimplemented!` are both breaking changes..
> 
> If you want to avoid a breaking change, is there any way to deprecate a default impl? I imagine putting `#[deprecated]` on the method declaration would treat the method itself as deprecated, not the implementation.

The reason to have default impls for feature-flagged trait methods is to avoid the "spooky action at a distance" nature of feature flags; if some import path enables the feature flag that enables the feature flag, it can break compilation for a downstream user in a way that they can't correct even if they never use any of the flagged functionality. 

The correct solution here is actually to, in the future, split up these traits such that everything that's currently on feature-flagged methods is instead on a separate trait that is feature-flagged in its entirety. But this is a larger change than I want to take on at the moment.

PR #1930, opened 2025-08-30
zcash_client_backend: Add `InputSource::select_unspent_notes`

nuttycom:
This utility function makes use of the same machinery as does note selection in order to return information about all the valid unspent notes in the wallet, irrespective of current spendability (which may depend on wallet scanning state, chosen anchor heights, and other wallet-local state.)


nuttycom:
This PR is in draft pending live testing in `zallet`.

zcash_client_backend/src/data_api.rs
```diff
@@ -841,14 +841,14 @@ impl<NoteRef> NoteRetention<NoteRef> for SimpleNoteRetention {
 
 /// Spendable shielded outputs controlled by the wallet.
```
str4d: Update the documentation to convey the new intended semantics of `ReceivedNotes`.

zcash_client_backend/src/data_api.rs
```diff
@@ -1364,7 +1364,17 @@ pub trait InputSource {
...
+    ) -> Result<ReceivedNotes<Self::NoteRef>, Self::Error>;
+
+    /// Returns the list of notes belonging to the wallet that are unspent as of the specified
+    /// target height.
+    fn select_unspent_notes(
+        &self,
+        _account: Self::AccountId,
+        _sources: &[ShieldedProtocol],
+        _target_height: TargetHeight,
+        _exclude: &[Self::NoteRef],
```
str4d: Nit: I dislike having the underscores in the trait argument names (because IDEs then proliferate it to trait impls), and there's no need for it (Rust doesn't require that a trait and its impl use the same argument names, so a default impl somewhere can underscore-prefix unused ones as required within its own context).
```suggestion
        account: Self::AccountId,
        sources: &[ShieldedProtocol],
        target_height: TargetHeight,
        exclude: &[Self::NoteRef],
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -396,9 +396,6 @@ impl Default for ConfirmationsPolicy {
 impl ConfirmationsPolicy {
     /// A policy to use the minimum number of confirmations possible: 1 confirmation for shielded
     /// notes irrespective of origin, and 0 confirmations for transparent UTXOs.
```
str4d: This documentation is incorrect for the now-public `ConfirmationsPolicy::MIN`. The configuration only expresses that transparent UTXOs can be *shielded* with 0 confirmations, not spent arbitrarily (which for the current structure of `ConfirmationsPolicy` would still be 1 confirmation).

zcash_client_backend/src/wallet.rs
```diff
@@ -377,8 +377,8 @@ impl Note {
     }
 }
 
-/// Information about a note that is tracked by the wallet that is available for spending,
-/// with sufficient information for use in note selection.
+/// A note that was receivd by the wallet, along with contextual information about the output that
```
str4d:
```suggestion
/// A note that was received by the wallet, along with contextual information about the output that
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -289,7 +319,7 @@ where
            SELECT {table_prefix}_received_note_id
```
str4d, line 317: The documentation just above this statement (specifically the text after the semicolon) is incorrect after this PR.

str4d (changes requested):
utACK c531bd0fe1a7e5ed05b436b6a6cc37106b064e35 after documentation fixes.

zcash_client_backend/src/data_api.rs
nuttycom: This was an artifact; originally I had added a default implementation for this method.

PR #1936, opened 2025-09-05
zcash_client_sqlite: Fix logging in `store_decrypted_tx`

str4d:
After zcash/librustzcash#1813 we don't store every transaction that this method is called with, but still logged that we were.


nuttycom (approved):
log-only ACK

PR #1941, opened 2025-09-09
Release zcash_primitives version 0.24.1

nuttycom:


daira (dismissed):
utACK (modulo @str4d's approval since apparently it's missing something).

zcash_primitives/CHANGELOG.md
```diff
@@ -15,6 +15,13 @@ workspace.
 - `zcash_primitives::transaction::tests::data` has been exposed under the
   `test-dependencies` feature to provide access to test vectors.
 
+## [0.24.1] - 2025-09-09
+
+### Changed
+- This release makes no public API changes, but provides pre-release support
```
str4d: This is incorrect. The exposing of  `zcash_primitives::transaction::tests::data` is released in this point release:

str4d (changes requested):
Reviewed daaa593235c92c6588c4acf1ade58c8cc6bd50b4.

zcash_primitives/CHANGELOG.md
nuttycom: Fixed.

str4d (dismissed):
utACK d011d1a29f54fbbdd07a5fa365ce3b03bcc57814

supply-chain/audits.toml
```diff
@@ -899,6 +899,12 @@ criteria = "safe-to-deploy"
 delta = "0.15.1 -> 0.16.0"
 notes = "The primary change here is the switch from the `hdwallet` dependency to using `bip32`."
 
+[[audits.zcash_primitives]]
```
str4d, line 902: Nit: we generally do not add audits of our own Zcash code, because it would be "self-auditing" mixed in with our third-party auditing. Instead we mark that we trust our own crates (which is a correct position to convey to external consumers of our audits).

str4d (approved):
utACK e9a2fca699880a4ade918ee4022a14d4cb41caa0 with a nit

supply-chain/audits.toml
nuttycom: Okay; we can fix this up in a later commit.

PR #1952, opened 2025-09-15
zcash_client_sqlite: Add address & diversifier index information to `v_tx_outputs`.

### LINKED ISSUE #1951: zcash_client_sqlite: Expose the address & diversifier index at which each output is received in `v_tx_outputs`

nuttycom:
Fixes #1951

### DIFF
--- a/zcash_client_sqlite/src/wallet/db.rs
+++ b/zcash_client_sqlite/src/wallet/db.rs
@@ -1018,9 +1018,36 @@ GROUP BY notes.account_id, notes.txid";
 ///   both outputs received as a consequence of wallet-internal transfers
 ///   and as change.
 ///
-/// The `to_address` column will only contain an address when the recipient is
-/// external. In all other cases, the recipient account id indicates the account
-/// that controls the output.
+/// # Columns
+/// - `txid`: The id of the transaction in which the output was sent or received.
+/// - `output_pool`: The value pool for the transaction; valid values for this are:
+///   - 0: Transparent
+///   - 2: Sapling
+///   - 3: Orchard
+/// - `output_index`: The index of the output within the transaction bundle associated with
+///   the `output_pool` value; that is, within `vout` for transparent, the vector of
+///   Sapling `OutputDescription` values, or the vector of Orchard actions.
+/// - `from_account_uuid`: The UUID of the wallet account that created the output, if the wallet
+///   spent notes in creating the transaction. Note that if multiple accounts in the wallet
+///   contributed funds in creating the associated transaction, redundant rows will exist in the
+///   output of this view, one for each such account.
+/// - `to_account_uuid`: The UUID of the wallet account that received the output, if any; for
+///   outgoing transaction outputs this will be `NULL`.
+/// - `address`: The address to which the output was sent; for received outputs, this is the
+///   address at which the output was received, or `NULL` for wallet-internal outputs.
+/// - `diversifier_index_be`: The big-endian representation of the diversifier index (or, for
+///   transparent addresses, the BIP 44 change-level index of the derivation path) of the receiving
+///   address. This will be `NULL` for outgoing transaction outputs.
+/// - `value`: The value of the output, in zatoshis.
+/// - `is_change`: `0` for outgoing outputs and outputs received at external-facing addresses, `1`
+///   for outputs received at wallet-internal addresses. This represents a best-effort judgement
+///   for whether or not the output should be considered change, and may not be correct for
+///   cross-account internal transactions, shielding transactions, or outputs explicitly sent from
+///   the wallet to itself. The determination of what counts as change is somewhat subjective and
+///   the value of this column should be used with caution.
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1030,13 +1030,15 @@ WITH unioned AS (
            ro.output_index              AS output_index,
            from_account.uuid            AS from_account_uuid,
            to_account.uuid              AS to_account_uuid,
-           NULL                         AS to_address,
+           a.address                    AS to_address,
```
str4d, line 1060: As noted in pairing, documentation of this needs updating because this column is no longer guaranteed to be null for non-external (and instead downstreams need to use `to_account_uuid IS NULL` to identify external.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1030,13 +1030,15 @@ WITH unioned AS (
...
+           a.diversifier_index_be       AS diversifier_index_be,
            ro.value                     AS value,
            ro.is_change                 AS is_change,
            ro.memo                      AS memo
     FROM v_received_outputs ro
     JOIN transactions
         ON transactions.id_tx = ro.transaction_id
+    JOIN addresses a ON a.id = ro.address_id
```
str4d: Hmm, I think this needs to be `LEFT JOIN`, otherwise wallet-interal shielded notes will be omitted.

str4d (changes requested):
Reviewed 9d97d00f97fe35fd277aa18bf19ff2bfcfd95b17

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Ahh, right, it's the account ID that's non-null for received outputs, not the address ID.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/9d97d00..a13f8257e5ee97231514e6a244c6c91fde004377) to address comments from code review.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/9d97d00..13fb221ea0d36f05cae9c73a25eda6bea839b643) to address comments from code review.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1018,9 +1018,40 @@ GROUP BY notes.account_id, notes.txid";
...
+///   cross-account internal transactions, shielding transactions, or outputs explicitly sent from
+///   the wallet to itself. The determination of what counts as change is somewhat subjective and
+///   the value of this column should be used with caution.
+/// - `memo`: The binary content of the memo associated with the output, if the output is a
+///   shielded output and the memo was received by the wallet, sent by the wallet or was able to be
+///   decrypted with the wallet's outgoing viewing key.
+///
+/// Wallet-internal outputs will always have the `to_account_uuid` column set to a non-null value.
+/// The `to_address` column will be set to the wallet-internal address in this case, and to the
+/// address to which the output was sent if the recipient is external to the wallet.
```
str4d: The first half of this sentence is incorrect (for the same reason we needed a `LEFT JOIN addresses`).

str4d (changes requested):
Reviewed 13fb221ea0d36f05cae9c73a25eda6bea839b643

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Hmmm, this is a phrasing thing - it's an address belonging to the wallet, but I agree, not specifically an address associated with the wallet's internal IVK.

nuttycom: I think I'll just delete this extra paragraph.

str4d (approved):
utACK e02889e2f8e9cec8c6290b78ec203a55ad6fcd0a

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1018,9 +1018,36 @@ GROUP BY notes.account_id, notes.txid";
...
+///   spent notes in creating the transaction. Note that if multiple accounts in the wallet
+///   contributed funds in creating the associated transaction, redundant rows will exist in the
+///   output of this view, one for each such account.
+/// - `to_account_uuid`: The UUID of the wallet account that received the output, if any; for
+///   outgoing transaction outputs this will be `NULL`.
+/// - `address`: The address to which the output was sent; for received outputs, this is the
+///   address at which the output was received, or `NULL` for wallet-internal outputs.
+/// - `diversifier_index_be`: The big-endian representation of the diversifier index (or, for
+///   transparent addresses, the BIP 44 change-level index of the derivation path) of the receiving
+///   address. This will be `NULL` for outgoing transaction outputs.
```
daira, line 1040: This doc for when `diversifier_index_be` is NULL doesn't correspond to the description in the change log.

Issue #1948, opened 2025-09-14
zcash_client_sqlite: Fix bug in `zcashd` support migration


str4d:
While testing https://github.com/zcash/wallet/pull/152 I encountered the following error:

```
UNIQUE constraint failed: accounts.hd_seed_fingerprint, accounts.hd_account_index
```

I think the bug is that the unique index `INDEX_HD_ACCOUNT ` needs to be over `(hd_seed_fingerprint, hd_account_index, zcashd_legacy_address_index)` (i.e. allowing one account with zcashd_legacy_address_index IS NULL, and then one for each non-null legacy address index).

PR #1953, opened 2025-09-16
zcash_client_sqlite: Fix `hd_account` uniqueness constraint

str4d:
We can't have `zcashd_legacy_address_index` be a nullable column, as that would permit multiple accounts with the same values for `(hd_seed_fingerprint, hd_account_index)` because every `NULL` is treated as unique. Instead we use -1 as a sentinel.

Closes zcash/librustzcash#1948.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -146,7 +146,10 @@ use {
 use zcash_client_backend::data_api::ORCHARD_SHARD_HEIGHT;
 
 #[cfg(feature = "zcashd-compat")]
-use zcash_keys::keys::zcashd;
+use {
+    crate::wallet::encoding::{decode_legacy_account_index, encode_legacy_account_index},
+    zcash_keys::keys::zcashd,
+};
 
 pub mod commitment_tree;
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -1360,15 +1356,15 @@ pub(crate) fn get_derived_account<P: consensus::Parameters>(
          WHERE hd_seed_fingerprint = :hd_seed_fingerprint
          AND hd_account_index = :hd_account_index
          AND (
-             :zcashd_legacy_address_index IS NULL
+             :zcashd_legacy_address_index = -1
```
nuttycom, line 1359: nit: we use `encode/decode`, but this being hardcoded here means that the encoding and decoding are sharing knowledge about the implementation of these queries. Maybe use a constant instead?

nuttycom (approved):
utACK 92004fa89ee3bd7ab107295becfcbb44dae24b35 with nonblocking nit

nuttycom (approved):
re-utACK 09e0301dad78bd988bea81cb4c0133075bdc2051

PR #1958, opened 2025-09-17
zcash_client_backend: Add `WalletWrite::mark_tx_trusted`

nuttycom:
This adds the capability to mark a transaction as explicitly "trusted", meaning that its outputs will be treated in the same fashion as wallet-internal outputs for purposes of spendability.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -2840,6 +2840,11 @@ pub trait WalletWrite: WalletRead {
         received_tx: DecryptedTransaction<Self::AccountId>,
     ) -> Result<(), Self::Error>;
 
+    /// Marks the given transaction as being trusted, such that its outputs will be available for
+    /// spending with [`ConfirmationsPolicy::trusted`] confirmations even if the output is not
+    /// wallet-internal.
+    fn mark_tx_trusted(&mut self, txid: TxId) -> Result<(), Self::Error>;
+
     /// Saves information about transactions constructed by the wallet to the persistent
     /// wallet store.
     ///
--- a/zcash_client_memory/src/wallet_write.rs
+++ b/zcash_client_memory/src/wallet_write.rs
@@ -935,6 +935,10 @@ impl<P: consensus::Parameters> WalletWrite for MemoryWalletDb<P> {
         Ok(())
     }
 
+    fn mark_tx_trusted(&mut self, _txid: TxId) -> Result<(), Self::Error> {
+        todo!()
+    }
+
     /// Truncates the database to the given height.
     ///
     /// If the requested height is greater than or equal to the height of the last scanned
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -3939,6 +3939,23 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
     Ok(())
 }
 
+pub(crate) fn mark_tx_trusted(
+    conn: &rusqlite::Transaction,
+    txid: TxId,
+) -> Result<(), SqliteClientError> {
+    conn.execute(
+        "UPDATE transactions
+         SET trust_status = :trust_status
+         WHERE txid = :txid",
+        named_params! {
+           ":txid": &txid.as_ref()[..],
+           ":trust_status": true
+        },
+    )?;
+
+    Ok(())
+}
+
 /// Inserts information about a mined transaction that was observed to
 /// contain a note related to this wallet into the database.
 pub(crate) fn put_tx_meta(
--- a/zcash_client_sqlite/src/wallet/common.rs
+++ b/zcash_client_sqlite/src/wallet/common.rs
@@ -288,7 +288,9 @@ where
              accounts.ufvk as ufvk, rn.recipient_key_scope,
              t.block AS mined_height,
              scan_state.max_priority,
-             MAX(tt.mined_height) AS max_shielding_input_height
+             IFNULL(t.trust_status, 0) AS trust_status,
+             MAX(tt.mined_height) AS max_shielding_input_height,
+             MIN(IFNULL(tt.trust_status, 0)) AS min_shielding_input_trust
          FROM {table_prefix}_received_notes rn
          INNER JOIN accounts ON accounts.id = rn.account_id
          INNER JOIN transactions t ON t.id_tx = rn.tx
@@ -350,6 +352,8 @@ where
         |row| -> Result<_, SqliteClientError> {
             let result_note = to_received_note(params, row)?;
             let max_priority_raw = row.get::<_, Option<i64>>("max_priority")?;
+            let tx_trust_status = row.get::<_, bool>("trust_status")?;
+            let tx_shielding_inputs_trusted = row.get::<_, bool>("min_shielding_input_trust")?;
             let shard_scan_priority = max_priority_raw
                 .map(|code| {
                     parse_priority_code(code).ok_or_else(|| {
@@ -360,7 +364,12 @@ where
                 })
                 .transpose()?;
 
-            Ok((result_note, shard_scan_priority))
+            Ok((
+                result_note,
+                shard_scan_priority,
+                tx_trust_status,
+                tx_shielding_inputs_trusted,
+            ))
         },
     )?;
 
@@ -370,7 +379,7 @@ where
 
     row_results
         .map(|t| match t? {
-            (Some(note), max_shard_priority) => {
+            (Some(note), max_shard_priority, trusted, tx_shielding_inputs_trusted) => {
                 let shard_scanned = max_shard_priority
                     .iter()
                     .any(|p| *p <= ScanPriority::Scanned);
@@ -385,13 +394,20 @@ where
                     (Some(received_height), Scope::Internal) => {
                         // The note has the required number of confirmations for a trusted note.
                         received_height <= trusted_height &&
-                        // And, if the note was the output of a shielding transaction, its inputs
-                        // have at least `untrusted` confirmations
-                        note.max_shielding_input_height().iter().all(|h| h <= &untrusted_height)
+                        // if the note was the output of a shielding transaction
+                        note.max_shielding_input_height().iter().all(|h| {
+                            // its inputs have at least `untrusted` confirmations
+                            h <= &untrusted_height ||
+                            // or its inputs are trusted and have at least `trusted` confirmations
+                            (h <= &trusted_height && tx_shielding_inputs_trusted)
+                        })
                     }
                     (Some(received_height), Scope::External) => {
                         // The note has the required number of confirmations for an untrusted note.
-                        received_height <= untrusted_height
+                        received_height <= untrusted_height ||
+                        // or it is the output of an explicitly trusted tx and has at least
+                        // `trusted` confirmations
+                        (received_height <= trusted_height && trusted)
                     }
                 };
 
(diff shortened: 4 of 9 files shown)


str4d:
I extracted the bugfixes into #1963 for inclusion in this week's crate releases. This PR is now just the `WalletWrite::mark_tx_trusted` addition, which is a breaking change but could go in with next week's crate releases (as the NU 6.1 mainnet details won't be ready for this week).

nuttycom:
Rebased.

zcash_client_memory/src/wallet_write.rs
```diff
@@ -935,6 +935,10 @@ impl<P: consensus::Parameters> WalletWrite for MemoryWalletDb<P> {
         Ok(())
     }
 
+    fn mark_tx_trusted(&mut self, _txid: TxId) -> Result<(), Self::Error> {
+        todo!()
```
schell, line 938: Is it ok to leave this as a `todo!()`?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3938,6 +3938,23 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
+         SET trust_status = :trust_status
+         WHERE txid = :txid",
+        named_params! {
+           ":txid": &txid.as_ref()[..],
+           ":trust_status": true
+        },
+    )?;
+
+    Ok(())
+}
```
schell, line 3991: Nice and easy 👍 .

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_trust_marker.rs
```diff
@@ -0,0 +1,123 @@
+//! Adds support for marking transactions as explicitly trusted for the purpose of satisfying the
+//! ZIP 315 confirmations policy
+use std::collections::HashSet;
+
+use schemerz_rusqlite::RusqliteMigration;
+use uuid::Uuid;
+
+use crate::wallet::init::{migrations::fix_v_transactions_expired_unmined, WalletMigrationError};
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x4e68277f_6269_467e_9437_f3853cc4a41f);
```
schell, line 11: Where does this `Uuid` come from?

schell (changes requested):
Looks good apart from the `todo!()`.

schell:
Rebased :)

zcash_client_memory/src/wallet_write.rs
nuttycom: Yeah, this is in `zcash_client_memory` which ChainSafe has a grant to work on support for.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_trust_marker.rs
nuttycom: `uuidgen` on my desktop; it just needs to be a distinct identifier from the other migration IDs.

zcash_client_backend/src/data_api.rs
```diff
@@ -2848,6 +2848,11 @@ pub trait WalletWrite: WalletRead {
         received_tx: DecryptedTransaction<Self::AccountId>,
     ) -> Result<(), Self::Error>;
 
+    /// Marks the given transaction as being trusted, such that its outputs will be available for
+    /// spending with [`ConfirmationsPolicy::trusted`] confirmations even if the output is not
+    /// wallet-internal.
+    fn mark_tx_trusted(&mut self, txid: TxId) -> Result<(), Self::Error>;
```
str4d: Design question: do we want to allow unmarking txs that were previously marked? I'm thinking of the "fat-finger" use case, where a user accidentally presses a button they didn't mean to, and wants to undo the change. This could be solved-ish in UIs by delaying application of the button (like how "Undo" buttons are implemented on emails), but that's a bit weird here as it's entirely local.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_trust_marker.rs
```diff
@@ -0,0 +1,123 @@
...
+
+const DEPENDENCIES: &[Uuid] = &[fix_v_transactions_expired_unmined::MIGRATION_ID];
```
str4d: Bug: the diagram (and the update to `CURRENT_LEAF_MIGRATIONS`) shows this migration as descending from `v_tx_outputs_return_addrs`, but it actually descends from that migration's parent. Decide which of the two is incorrect and fix it.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -285,6 +285,7 @@ CREATE TABLE "transactions" (
     target_height INTEGER,
     min_observed_height INTEGER NOT NULL,
     confirmed_unmined_at_height INTEGER,
+    trust_status INTEGER,
```
str4d, line 291: Document this new column.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1074,7 +1075,10 @@ WITH unioned AS (
            a.diversifier_index_be       AS diversifier_index_be,
            ro.value                     AS value,
            ro.is_change                 AS is_change,
-           ro.memo                      AS memo
+           ro.memo                      AS memo,
+           IFNULL(t.trust_status, 0)    AS is_trusted,
```
str4d: Bug: `t` is undefined. Presumably it is meant to be:
```suggestion
           IFNULL(transactions.trust_status, 0)    AS is_trusted,
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1084,6 +1088,12 @@ WITH unioned AS (
     -- join on the accounts table to obtain account UUIDs
     LEFT JOIN accounts from_account ON from_account.id = sent_notes.from_account_id
     LEFT JOIN accounts to_account ON to_account.id = ro.account_id
+    -- join to transparent_received_output_spends to gather data about transaction inputs
+    LEFT OUTER JOIN transparent_received_output_spends ros
+       ON ros.transaction_id = t.id_tx
```
str4d: Bug: `t` is undefined. Presumably it is meant to be:
```suggestion
       ON ros.transaction_id = transactions.id_tx
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1095,7 +1105,10 @@ WITH unioned AS (
            NULL                         AS diversifier_index_be,
            sent_notes.value             AS value,
            0                            AS is_change,
-           sent_notes.memo              AS memo
+           sent_notes.memo              AS memo,
+           1                            AS is_trusted,
+           NULL                         AS t_input_mined_height,
+           1                            AS t_input_trust_status
```
str4d: Bug: `t_input_trust_status` is only fixed to true for wallet-created transactions that solely spend wallet-created outputs. If the transaction spends inputs that were received externally, then `is_trusted` is fixed to true but `t_input_trust_status` would be false (unless you have a more subtle semantic meaning for `t_input_trust_status` than what I am interpreting it as, which I don't know because it is undocumented). This kind of transaction cannot occur in Zashi (as externally-received transparent inputs are always shielded first), but it *will* occur in Zallet (which this library also backs).

`t_input_mined_height` should not be fixed to NULL unless it is documented that this is the case for wallet-created outputs. But if the `t_input_trust_status` bug is fixed with joins, then we can trivially look up the input mined height same as for the other side of the union.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1113,7 +1126,9 @@ SELECT
     max(to_address) AS to_address,
     max(value) AS value,
     max(is_change) AS is_change,
-    max(memo) AS memo
+    max(memo) AS memo,
+    max(t_input_mined_height) AS max_shielding_input_height,
+    min(t_input_trust_status) AS min_shielding_input_trust
```
str4d: Document the two new columns.

zcash_client_sqlite/src/wallet/db.rs
str4d: Possible bug: the new `is_trusted` column of `unioned` is never used or exposed in `v_tx_outputs`.

zcash_client_sqlite/src/wallet/common.rs
str4d, line 1: I have not reviewed the changes to this file yet.

str4d (changes requested):
Reviewed 3fd16772a7cb401df539cb96dae92c8fd25d2283 except for one file.

zcash_client_sqlite/src/wallet/common.rs
str4d: I have now reviewed the changes to this file as of the specified commit.

zcash_client_backend/src/data_api.rs
nuttycom: Yeah, that's probably valuable.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: For wallet-created outputs sent to external recipients, it doesn't make sense for `t_input_trust_status` to ever be false because we already have a proof that the input was trusted, either via confirmation depth or explicit trust; otherwise, we would not have been able to spend it to create the output in question.  

This does depend somewhat on the invariant that we never spend transparent coins when sending to external recipients, except in the case of TEX address sends. But the TEX interstitial outputs are always trusted.

nuttycom: All that being said, it's been long enough now since I worked on this PR that I no longer remember why it was that it alters `v_tx_outputs` in the first place.

nuttycom: Given that I can't figure it out, I'm changing this to modify `v_transactions` to include the `trust_status` and leaving it at that.

zcash_client_memory/src/wallet_write.rs
schell: Could we put a note about that in here? Either in comments or in the todo itself: `todo!("to be implemented in grant work by ChainSafe")`.

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_trust_marker.rs
```diff
@@ -0,0 +1,183 @@
...
+            LEFT JOIN sent_note_counts
+                 ON sent_note_counts.account_id = notes.account_id
+                 AND sent_note_counts.txid = notes.txid
+            GROUP BY notes.account_id, notes.txid"
+        )?;
+        Ok(())
+    }
+
+    fn down(&self, _transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        Ok(())
```
schell: I'm guessing we don't ever run migrations backward?

Also, just for my education, how did you generate the SQL above? Did it come from another part of the codebase, by hand or a tool?

nuttycom: The SQL is written by hand. But it is largely copied from the previous version of the source code for the view.

nuttycom: With respect to the down migration, this should actually return `Err`, not sure how I missed that.

schell: > this should actually return Err

Oh! Glad I asked!

str4d (approved):
utACK 135d72172f1e0f414f3a8d656dbb38c8e578821e

zcash_client_sqlite/src/wallet/init/migrations/add_transaction_trust_marker.rs
```diff
@@ -0,0 +1,184 @@
...
+
+use schemerz_rusqlite::RusqliteMigration;
+use uuid::Uuid;
+
+use super::tx_observation_height;
+use crate::wallet::init::WalletMigrationError;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x4e68277f_6269_467e_9437_f3853cc4a41f);
+
+const DEPENDENCIES: &[Uuid] = &[tx_observation_height::MIGRATION_ID];
```
str4d: Bug: This migration needs to also depend on `fix_v_transactions_expired_unmined` which alters `v_transactions`.
```suggestion
const DEPENDENCIES: &[Uuid] = &[
    tx_observation_height::MIGRATION_ID,
    fix_v_transactions_expired_unmined::MIGRATION_ID,
];
```

nuttycom (commented):
utACK ac6e040d008801f6005e75d10705a04a0b48c36b

PR #1960, opened 2025-09-18
zcash_client_backend: Fix missing feature flag.

nuttycom:
Fixes a build failure that occurred when both the `pczt` and `transparent-inputs` features were enabled without `transparent-key-import` being enabled.


daira (commented):
Post-hoc ACK

PR #1835, opened 2025-06-19
pczt: Add a test exercising the P2SH multisig spend workflow

str4d:
Closes zcash/librustzcash#1833.

PR #1962, opened 2025-09-24
zcash_transparent: Implement ZIP 48

str4d:


str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/39a2df323bc6d5ff268b625cf5de886d8720334a..e6d132187fcba316db824c89462713d935b54fc6) to rebase on #1835 after it was finished up.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/e6d132187fcba316db824c89462713d935b54fc6..6f693ef398ecf671ed54e64c66538d920337b779) to fix rebase error.

str4d:
The no-std / WASM builds are failing due to AFAICT the `secp256k1/alloc` feature flag being enabled. Might just mean we need to update our CI, because I've confirmed that this commit builds fine inside the Keystone firmware.

zcash_transparent/src/zip48.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// Derives the signing key for this account's scoped address at the given index.
+    pub fn derive_signing_key(
+        &self,
+        scope: zip32::Scope,
+        address_index: NonHardenedChildIndex,
+    ) -> SecretKey {
+        *self
+            .key
+            .derive_child(TransparentKeyScope::from(scope).into())
+            .expect("chance of failure is around 2^-127")
```
nuttycom, line 89: Elsewhere I think we propagate this error; we should check and be consistent.

zcash_transparent/src/zip48.rs
```diff
@@ -0,0 +1,450 @@
...
+impl FullViewingKey {
+    /// Constructs a full viewing key for a standard ZIP 48 account.
+    ///
+    /// This uses the standard wallet descriptor template. For example:
+    /// - `threshold == 2`
+    /// - `key_info.len() == 3`
+    /// - Wallet descriptor template: `"sh(sortedmulti(2,@0/**,@1/**,@2/**))"`
+    pub fn standard(
+        threshold: u8,
+        key_info: Vec<AccountPubKey>,
```
nuttycom, line 199:
```suggestion
        key_info: &[AccountPubKey],
```

zcash_transparent/src/zip48.rs
```diff
@@ -0,0 +1,450 @@
...
+                    .iter()
+                    .map(|key| key.origin.derivation())
+                    .collect::<BTreeSet<_>>();
+                if derivations.len() != 1 {
+                    return Err(FullViewingKeyError::IncompatiblePubKeys);
+                }
+            }
+
+            // TODO: Decide whether `key_info` should be sorted (or checked to be sorted)
+            // to ensure a canonical multipath descriptor.
```
nuttycom, line 224: Do we ever expect this type to be constructed via parsing? If so, I think that we'll need to preserve the order of `key_info` so that the parsing will round-trip.

nuttycom (approved):
utACK 6f693ef398ecf671ed54e64c66538d920337b779, looks great.

zcash_transparent/src/zip48.rs
```diff
@@ -0,0 +1,450 @@
...
+                // information. We check this by collecting into a set, and then checking
+                // it contains one entry. We don't need to verify the derivation's
+                // structure because `AccountPubKey` enforces it by construction.
+                let derivations = key_info
+                    .iter()
+                    .map(|key| key.origin.derivation())
+                    .collect::<BTreeSet<_>>();
+                if derivations.len() != 1 {
+                    return Err(FullViewingKeyError::IncompatiblePubKeys);
+                }
```
str4d, line 220: Notes:
- This requires BIP 380 key origin information to be provided by all participants. It's optional in BIP 380 but I think fine for us to require in ZIP 48.
- This currently requires the `account` value to be identical for all keys in a ZIP 48 account. BIP 48 is very underspecified but every example of its use that I can find uses the same account number for each participant, and it's easier to lift this restriction in future than to impose it later, so for now I'm imposing it.

I plan to open a PR with these updates to the ZIP 48 draft.

str4d: A downside to requiring `account` to match would be that impls in e.g. hardware wallets could not just use the next available account, and would need the participants to agree ahead of time on which `account` to use. This is only a problem if a user wants to use the same seed with multiple different sets of participants; if they always use a given seed with the same participants then there's no coordination problem (as each new set of participants has its own set of seeds, from which the account numbering would restart at zero).

zcash_transparent/src/zip48.rs
str4d: We take ownership of `key_info` here. If we took a slice, we'd then need to clone it.

zcash_transparent/src/zip48.rs
str4d: Right, that's why I raised this as a question. For now we can avoid it, and kick parsing to the layer above (in my case, `zcash-devtool`).

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/3c3157c3f3a68b23f91c14881cb0a4fa6a8a9ccd..bb71ab63b12ba8b9c6a7da17daf2a223997dd150) to bump release date to Monday, and get `cargo vet` passing in CI (I will of course be pushing a post-release commit to also update it again before merging this).

PR #1963, opened 2025-09-24
zcash_client_sqlite: Bugfixes

str4d:
Extracted from #1958.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations.rs
@@ -116,9 +116,10 @@ pub(super) fn all_migrations<
     //                              \      ensure_default_transparent_address    /
     //                               \                     |                    /
     //                                `---- fix_transparent_received_outputs --'
-    //                                        /            |            \
-    //             support_zcashd_wallet_import            |            fix_v_transactions_expired_unmined
-    //                                         v_tx_outputs_return_addrs
+    //                                        /                         \
+    //                     support_zcashd_wallet_import          fix_v_transactions_expired_unmined
+    //                                                                           |
+    //                                                                v_tx_outputs_return_addrs
     let rng = Rc::new(Mutex::new(rng));
     vec![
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -116,9 +116,10 @@ pub(super) fn all_migrations<
     //                              \      ensure_default_transparent_address    /
     //                               \                     |                    /
     //                                `---- fix_transparent_received_outputs --'
-    //                                        /            |            \
-    //             support_zcashd_wallet_import            |            fix_v_transactions_expired_unmined
-    //                                         v_tx_outputs_return_addrs
+    //                                        /                         \
+    //                     support_zcashd_wallet_import          fix_v_transactions_expired_unmined
+    //                                                                           |
+    //                                                                v_tx_outputs_return_addrs
```
str4d, line 122: In the commit message, @nuttycom says that `fix_v_transactions_expired_unmined` modifies `v_tx_outputs`. AFAICT it does not, and these were fine to be in parallel. However, there's no harm in this graph reorganisation, because `v_tx_outputs_return_addrs` has not yet been released (whereas `fix_v_transactions_expired_unmined` has).

str4d (commented):
utACK 5193c4b4645af0ebaa11c97ec2db4d3cff9c2724

schell (approved):
LGTM 🧂

zcash_client_sqlite/src/wallet/init/migrations.rs
nuttycom: I definitely was getting `verify_schema` errors due to  `v_tx_outputs` migrations being run out-of-order.

nuttycom: https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/init/migrations/fix_v_transactions_expired_unmined.rs#L167-L218

PR #1967, opened 2025-09-25
zcash_client_sqlite: Fix incorrect `GROUP BY` in wallet balance query.

nuttycom:


zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -148,9 +148,7 @@ where
                WHERE stx.block IS NOT NULL -- the spending tx is mined
                OR stx.expiry_height IS NULL -- the spending tx will not expire
              )
-             GROUP BY rn.id, t.txid, rn.{output_index_col},
-                      rn.diversifier, rn.value, {note_reconstruction_cols}, rn.commitment_tree_position,
-                      accounts.ufvk, rn.recipient_key_scope, t.mined_height"
+             GROUP BY rn.id"
```
str4d, line 151: I added checks in the test that would catch a bug here if one existed, but it didn't because we were already grouping by `rn.id` as it's a returned column.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -321,11 +319,7 @@ where
            OR stx.expiry_height IS NULL -- the spending tx will not expire
            OR stx.expiry_height >= :target_height -- the spending tx is unexpired
          )
-         GROUP BY
-            rn.id, t.txid, rn.{output_index_col},
-            rn.diversifier, rn.value, {note_reconstruction_cols}, rn.commitment_tree_position,
-            ufvk, rn.recipient_key_scope, 
-            t.block, scan_state.max_priority"
+         GROUP BY rn.id"
```
str4d, line 322: I added checks in the test that would catch a bug here if one existed, but it didn't because we were already grouping by `rn.id` as it's a returned column.

zcash_client_memory/src/testing/pool/orchard.rs
```diff
@@ -157,3 +157,9 @@ fn scan_cached_blocks_finds_change_notes() {
 fn scan_cached_blocks_detects_spends_out_of_order() {
     testing::pool::scan_cached_blocks_detects_spends_out_of_order::<OrchardPoolTester>()
 }
+
+#[test]
+#[ignore] //FIXME
```
nuttycom, line 162: Why are these currently ignored?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -145,6 +145,8 @@ pub trait ShieldedPoolTester {
 
     fn next_subtree_index<A: Hash + Eq>(s: &WalletSummary<A>) -> u64;
 
+    fn note_value(note: &Self::Note) -> Zatoshis;
```
nuttycom, line 148: Do we document changes to this trait in the CHANGELOG? It's test-only, but it is technically part of the public API.

nuttycom (commented):
utACK 705e695b53d18a946cf94dfc6949487010f1c2e4 (just the test changes as the other changes are mine). If we're happy with this PR, we should make the hotfix release on this branch. Since we've yanked the 0.18.0 release, we can mark it as yanked in the changelog and record all of the semver-breaking changes as changes relative to the 0.18.1 release.

zcash_client_backend/src/data_api/testing/pool.rs
schell: Better safe than sorry? 🤷

str4d (approved):
utACK 6a3725126e389a5502b6c5a9ca7285f0c7ad934b

str4d (approved):
utACK a1ba00d17b4a185ce3543bb60bb0bdad256fe5e2

str4d (approved):
Re-utACK 85d528e9e60826b329c365a4f92c355200fda52c

PR #1969, opened 2025-09-27
zcash_client_backend: Add `WalletRead` method for retrieval of recently-exposed ephemeral addresses.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -1698,6 +1698,31 @@ pub trait WalletRead {
...
+    ///
+    /// # Parameters
+    /// - `account`: The identifier for the account from which transparent receivers should be
+    ///   returned.
+    /// - `exposure_depth`: Implementations of this method should return only addresses exposed at
+    ///   heights greater than or equal to `chain_tip_height - exposure_depth`
+    /// - `exclude_used`: When set to `true`, do not return addresses that are known have already
+    ///   received funds in a transaction.
+    #[cfg(feature = "transparent-inputs")]
+    fn get_ephemeral_transparent_receivers(
```
str4d, line 1805: How does this differ in intention, usage, or returned data from `get_known_ephemeral_addresses`? The documentation for this method does not reference the other, and the documentation for the other method is not updated to account for the presence of this method. Both methods need sufficient documentation for callers to understand how to use them both correctly, and why there are two.

zcash_client_backend/src/data_api.rs
```diff
@@ -2942,6 +2942,17 @@ pub trait WalletWrite: WalletRead {
         _status: TransactionStatus,
     ) -> Result<(), Self::Error>;
 
+    /// Schedules a UTXO check for the given address at a random time that has an expected value of
+    /// `offset_seconds` from the current system time.
+    #[cfg(feature = "transparent-inputs")]
+    fn schedule_next_check(
```
str4d, line 3032: Is this a method that was needed for the TEX address returned-value-or-failure case usage? If not, how was the TEX address case planned to be handled, and why is this new method necessary?

zcash_client_backend/src/data_api.rs
nuttycom: The primary difference is that `get_known_ephemeral_addresses` allows the user to query by child index, whereas this method queries by exposure depth. Naming is a bit of a hassle; I don't think that either of these have good names. I guess one possibility would be to combine them and use a more sophisticated query/request type, but what I wanted here was basically an additive change from the perspective of an API consumer.

zcash_client_backend/src/data_api.rs
nuttycom: No; the TEX use case is handled via the normal `TransactionDataRequest` path, and those checks are performed without API user intervention as part of the normal handling of `scan_cached_blocks` and `decrypt_and_store_transaction`. This is explicitly for use in scenarios like https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/244 where you need to schedule a check manually.

zcash_client_backend/src/data_api.rs
```diff
@@ -1698,6 +1698,31 @@ pub trait WalletRead {
...
+    ///   heights greater than or equal to `chain_tip_height - exposure_depth`
+    /// - `exclude_used`: When set to `true`, do not return addresses that are known have already
+    ///   received funds in a transaction.
+    #[cfg(feature = "transparent-inputs")]
+    fn get_ephemeral_transparent_receivers(
+        &self,
+        _account: Self::AccountId,
+        _exposure_depth: u32,
+        _exclude_used: bool,
+    ) -> Result<HashMap<TransparentAddress, Option<TransparentAddressMetadata>>, Self::Error> {
```
nuttycom: This should also return the number of unexposed addresses remaining in the gap limit. Wallets need this in order to be able to anticipate when they may need to send a transaction to reset the gap limit.

zcash_client_backend/src/data_api.rs
nuttycom: I somewhat want to just get rid of `get_known_ephemeral_addresses` entirely. We don't use it in the SDK, we don't use it in the devtool, it was added speculatively when Daira first implemented ephemeral address handling and it's currently only used for testing. I'm tempted to move it to `WalletTest`.

zcash_client_backend/src/data_api.rs
```diff
@@ -1694,10 +1694,35 @@ pub trait WalletRead {
         _account: Self::AccountId,
         _include_change: bool,
         _include_standalone: bool,
-    ) -> Result<HashMap<TransparentAddress, Option<TransparentAddressMetadata>>, Self::Error> {
+    ) -> Result<HashMap<TransparentAddress, TransparentAddressMetadata>, Self::Error> {
         unimplemented!("WalletRead::get_transparent_receivers must be overridden for wallets to use the `transparent-inputs` feature")
```
schell, line 1781: What is the rational behind implementing these default trait fns but leaving them with an `unimplemented!` body? Do we really want folks to be able to `impl WalletRead` using the `transparent-inputs` feature, but _not_ require them to implement these functions, bringing a lurking `panic` into play? We could enforce this check at compile time by not implementing a default for these, so I'm sure I'm lacking some education here.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -385,21 +471,26 @@ pub(crate) fn select_addrs_to_reserve<P: consensus::Parameters>(
...
+                                        scope,
+                                        i,
+                                        Exposure::Unexposed,
+                                        None,
+                                    ))
+                                },
+                            )?,
+                        ))
+                    })
+                    .transpose()
```
schell, line 495: I think this could benefit from some refactoring. At the very least `<Option<TransparentKeyScope>>::from(key_scope)` could be done at the function's top level scope, and I think that might flatten a lot of this out.

zcash_client_backend/src/data_api.rs
nuttycom: The important situation that this accounts for is if downstream user uses an implementation of `zcash_client_backend` that doesn't implements the `transparent-inputs` feature-flagged methods, but activates that feature inadvertently via some other path (like some other dependency which enables that feature by default.)

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1365,15 +1368,6 @@ where
...
              to: TransparentAddress|
              -> Result<(), CreateErrT<DbT, InputsErrT, FeeRuleT, ChangeErrT, N>> {
-                // Always reject sending to one of our known ephemeral addresses.
-                #[cfg(feature = "transparent-inputs")]
-                if wallet_db
-                    .find_account_for_ephemeral_address(&to)
-                    .map_err(Error::DataSource)?
-                    .is_some()
-                {
-                    return Err(Error::PaysEphemeralTransparentAddress(to.encode(params)));
```
daira, line 1375: If this is no longer an error, remove the error variant.

zcash_client_backend/src/data_api.rs
```diff
@@ -1686,18 +1686,43 @@ pub trait WalletRead {
...
+    /// under this account within `exposure_depth` blocks of the chain tip. Wallets may scan the
+    /// chain for UTXOs sent to these receivers, but should do so in a fashion that does not reveal
+    /// that they are controlled by the same wallet.
+    ///
+    /// # Parameters
+    /// - `account`: The identifier for the account from which transparent receivers should be
+    ///   returned.
+    /// - `exposure_depth`: Implementations of this method should return only addresses exposed at
+    ///   heights greater than or equal to `chain_tip_height - exposure_depth`
+    /// - `exclude_used`: When set to `true`, do not return addresses that are known have already
```
```suggestion
    /// - `exclude_used`: When set to `true`, do not return addresses that are known to have already
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1921,6 +1857,83 @@ pub trait WalletTest: InputSource + WalletRead {
...
+    ///     and second transactions, meaning that this wallet does not need to observe
+    ///     the second transaction to determine when it would have expired.
+    ///
+    /// - A TEX address recipient decided to return funds that the wallet had sent to
+    ///   them.
+    ///
+    /// In all cases, the wallet should re-shield the unspent outputs, in a separate
+    /// transaction per ephemeral address, before re-spending the funds.
+    #[cfg(feature = "transparent-inputs")]
+    fn get_known_ephemeral_addresses(
```
daira, line 1987: Is this not redundant now with `get_ephemeral_transparent_receivers`?

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +627,139 @@ impl OvkPolicy {
...
+pub enum GapMetadata {
+    /// The address, or an address at a greater child index, has received transparent funds and
+    /// will be discovered by wallet recovery by exploration over the space of
+    /// [`NonHardenedChildIndex`]es using the provided gap limit.
+    GapRecoverable { gap_limit: u32 },
+    /// The address exists within an address gap of the given limit size, and will be discovered by
+    /// wallet recovery by exploration using the provided gap limit. In the view of the wallet, no
+    /// addresses at the given position or greater (up to the gap limit) have received funds. The
+    /// number of addresses remaining within the gap limit before no additional addresses can be
+    /// allocated is given by `gap_limit - (gap_position + 1)`.
```
daira, line 651: Is this sufficient information, or do we also to know the absolute starting index of the gap?

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +624,83 @@ impl OvkPolicy {
     }
 }
 
+/// Metadata describing whether and when a transparent address was exposed by the wallet.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg(feature = "transparent-inputs")]
+pub enum ExposedAt {
+    Height(BlockHeight),
```
str4d: Document this variant to match the `exposed_at_height` column from which it is drawn:
```suggestion
    /// The address was first exposed to the wider ecosystem at this height, to the best
    /// of our knowledge.
    ///
    /// - For user-generated addresses, this is the chain tip height at the time that the
    ///   address was generated by an explicit request by the user or reserved for use in
    ///   a ZIP 320 transaction. These heights are not recoverable from chain.
    /// - In the case of an address with its first use discovered in a transaction
    ///   obtained by scanning the chain, this will be set to the mined height of that
    ///   transaction. In recover from seed cases, this is what user-generated addresses
    ///   will be assigned.
    Height(BlockHeight),
```

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +624,83 @@ impl OvkPolicy {
...
+    Unexposed,
```
str4d: Document precisely what this variant is intended to convey.

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +624,83 @@ impl OvkPolicy {
...
+    Unknown,
```
str4d, line 691: Document this variant (wording will depend on how `Unexposed` is documented).

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -145,6 +147,12 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
             })?
             .to_transparent_address();
 
+        let exposed_at = row
+            .get::<_, Option<u32>>("exposed_at_height")?
+            .map_or(ExposedAt::Unexposed, |h| {
```
str4d: Re: the documentation question, it's not clear to me whether `exposed_at_height IS NULL` should always map to `ExposedAt::Unexposed`. The presence of a row in the `addresses` table does not IMO mean that `ExposedAt::Unknown` doesn't apply. In particular, if the address is in the middle of a gap (that is smaller than the gap limit), then it's almost certainly the case that this address was exposed at some point, but never used on chain by whomever it was exposed to (so we have no information about when), in which case it should be `ExposedAt::Unknown`. Documenting these two variants would help me to know what you intend them to convey to users.

... except we *might* have some information about when: if an address derived earlier than it is known to be exposed, then arguably that is also a lower bound on when this address could have been exposed (if we assume that wallets always derive transparent addresses in-order). But I'd be fine with that being a variant that is added later on (or maybe `Unknown` is altered to have a lower bound, which we can always set to either the genesis block or the wallet birthday depending on which conveys the intended information when we have nothing else (the latter could be an issue for "re-birthday"ed wallets).

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -385,21 +394,25 @@ pub(crate) fn select_addrs_to_reserve<P: consensus::Parameters>(
...
+                        Ok::<_, SqliteClientError>((
+                            address_id,
+                            a,
+                            <Option<TransparentKeyScope>>::from(key_scope).map_or(
+                                Err(SqliteClientError::UnknownZip32Derivation),
+                                |scope| {
+                                    Ok(TransparentAddressMetadata::derived(
+                                        scope,
+                                        i,
+                                        ExposedAt::Unexposed,
```
str4d: Similarly here: by convention we can assume that addresses beyond the gap limit at the time we first fully sync have never been exposed, and that would hold as we continue to derive addresses even when they are subsequently within the gap limit. But what should we convey about addresses that are within the gap limit at the time we first fully sync during a recover-from-seed? I suspect `ExposedAt::Unknown` is what we should be using for them.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1477,22 +1490,28 @@ pub(crate) fn get_transparent_address_metadata<P: consensus::Parameters>(
...
+            "SELECT diversifier_index_be, key_scope, imported_transparent_receiver_pubkey, exposed_at_height
              FROM addresses
              JOIN accounts ON addresses.account_id = accounts.id
              WHERE accounts.uuid = :account_uuid
              AND cached_transparent_receiver_address = :address",
             named_params![":account_uuid": account_uuid.0, ":address": &address_str],
             |row| {
                 let scope_code = row.get("key_scope")?;
+                let exposed_at = row.get::<_, Option<u32>>("exposed_at_height")?.map_or(
+                    ExposedAt::Unexposed,
```
str4d: This should be updated to reflect whatever is decided above.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -125,9 +180,14 @@ pub(crate) fn schedule_ephemeral_address_checks<C: Clock, R: RngCore>(
...
     if let Some((_, max_check_time)) = rows.last().as_ref() {
+        // Updating the next check time should not result in an already-scheduled check being
+        // further deferred.
         let mut set_check_time = conn.prepare(
             "UPDATE addresses
-             SET transparent_receiver_next_check_time = :next_check
+             SET transparent_receiver_next_check_time = MIN(
+                :next_check,
+                MAX(:next_check, IFNULL(transparent_receiver_next_check_time, :next_check))
+             )
```
daira: Would it be clearer to write this as `IFNULL(:next_check, transparent_receiver_next_check_time)`?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1336,6 +1424,40 @@ pub(crate) fn next_check_time<R: RngCore, D: DerefMut<Target = R>>(
...
+    let addr_str = address.encode(params);
+    let start_time = clock.now();
+    let next_check = next_check_time(&mut rng, start_time, offset_seconds)?;
+    let scheduled_next_check = conn
+        .query_row(
+            "UPDATE addresses
+             SET transparent_receiver_next_check_time = MIN(
+                :next_check,
+                MAX(:next_check, IFNULL(transparent_receiver_next_check_time, :next_check))
+             )
```
daira: Same potential simplification.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1474,28 +1596,76 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
                     let address_index = address_index_from_diversifier_index_be(row.get("diversifier_index_be")?)?;
+                    let exposed_at_height = row.get::<_, Option<u32>>("exposed_at_height")?.map(BlockHeight::from);
+
                     match <Option<TransparentKeyScope>>::from(key_scope).zip(address_index) {
                         Some((scope, address_index)) => {
-                            Ok(TransparentAddressMetadata::Derived {
+                            let exposure = exposed_at_height.map_or(
+                                Ok::<_, SqliteClientError>(Exposure::Unexposed),
+                                |at_height| {
+                                    let gap_metadata =  match gap_limits.limit_for(key_scope) {
```
daira:
```suggestion
                                    let gap_metadata = match gap_limits.limit_for(key_scope) {
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1698,6 +1698,31 @@ pub trait WalletRead {
...
+    /// The set contains all ephemeral transparent receivers that are known to have been derived
+    /// under this account within `exposure_depth` blocks of the chain tip. Wallets may scan the
+    /// chain for UTXOs sent to these receivers, but should do so in a fashion that does not reveal
+    /// that they are controlled by the same wallet.
+    ///
+    /// # Parameters
+    /// - `account`: The identifier for the account from which transparent receivers should be
+    ///   returned.
+    /// - `exposure_depth`: Implementations of this method should return only addresses exposed at
+    ///   heights greater than or equal to `chain_tip_height - exposure_depth`
```
str4d: This is annoyingly different from the "depth in main chain" of a transaction, which is governed by the relation `mined_height = chain_tip_height - depth + 1`. I'd prefer that a depth of 1 continue to mean "the block at the chain tip", which could be specified here as:
```suggestion
    /// - `exposure_depth`: Implementations of this method should return only addresses exposed at
    ///   heights greater than `chain_tip_height - exposure_depth`.
```

Alternatively, don't call this a "depth" so there's no chance of confusion.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -112,9 +112,19 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
...
     scopes: &[KeyScope],
+    exposure_depth: Option<u32>,
+    exclude_used: bool,
 ) -> Result<HashMap<TransparentAddress, Option<TransparentAddressMetadata>>, SqliteClientError> {
     let mut ret: HashMap<TransparentAddress, Option<TransparentAddressMetadata>> = HashMap::new();
 
+    let min_exposure_height = exposure_depth
+        .map(|d| {
+            Ok::<_, SqliteClientError>(
+                chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)? - d,
```
str4d: If the alternative of "unify meaning of depth in chain" is taken:
```suggestion
                chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)? - d + 1,
```

zcash_client_backend/src/data_api.rs
```diff
@@ -1686,18 +1686,43 @@ pub trait WalletRead {
...
         unimplemented!("WalletRead::get_transparent_receivers must be overridden for wallets to use the `transparent-inputs` feature")
     }
 
+    /// Returns the set of previously-exposed ephemeral transparent receivers generated by the given
+    /// account controlled by this wallet.
+    ///
+    /// The set contains all ephemeral transparent receivers that are known to have been derived
+    /// under this account within `exposure_depth` blocks of the chain tip. Wallets may scan the
+    /// chain for UTXOs sent to these receivers, but should do so in a fashion that does not reveal
+    /// that they are controlled by the same wallet.
```
str4d: Reference `TransparentAddressMetadata::next_check_time` here.

zcash_client_backend/src/wallet.rs
```diff
@@ -684,6 +700,12 @@ impl TransparentAddressMetadata {
         self.exposed_at
     }
 
+    /// Returns the timestamp of the next time that the light wallet server should be queried for
+    /// UTXOs associated with this address.
```
str4d:
```suggestion
    /// UTXOs associated with this address, or `None` if the wallet backend is not placing any
    /// restriction on when this address can be queried.
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -432,6 +439,7 @@ pub(crate) fn select_addrs_to_reserve<P: consensus::Parameters>(
                                         scope,
                                         i,
                                         ExposedAt::Unexposed,
+                                        None,
```
str4d: For clarity, the intention is that `transparent_receiver_next_check_time` is always `NULL` for any address returned from this API?

zcash_client_backend/src/data_api.rs
```diff
@@ -2942,6 +2942,17 @@ pub trait WalletWrite: WalletRead {
         _status: TransactionStatus,
     ) -> Result<(), Self::Error>;
 
+    /// Schedules a UTXO check for the given address at a random time that has an expected value of
+    /// `offset_seconds` from the current system time.
```
str4d, line 3027:
```suggestion
    /// `offset_seconds` from the current system time.
    ///
    /// Returns the time at which the address is scheduled for checking, or `None` if the
    /// address is not being tracked by the wallet.
```
(assuming I understand the code correctly).

zcash_client_sqlite/src/lib.rs
```diff
@@ -1218,8 +1222,8 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> WalletTes
     }
 }
 
-impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R> WalletWrite
-    for WalletDb<C, P, CL, R>
+impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R: RngCore>
+    WalletWrite for WalletDb<C, P, CL, R>
```
str4d, line 1276: Document this breaking change to `impl WalletWrite for WalletDb` in the changelog.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -150,9 +150,14 @@ pub(crate) fn schedule_ephemeral_address_checks<C: Clock, R: RngCore>(
...
```
str4d: - I understand the purpose of the `MIN(:next_check, _)`: if the proposed next-check time for this address is sooner than the already-scheduled time, then use it.
- I don't understand the purpose if the `MAX(:next_check, _)`: that seems to suggest that if the proposed next-check time for this address is later than the already-scheduled time, use it - but that would result in further deferring the already-scheduled check, violating the comment.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1379,6 +1380,40 @@ pub(crate) fn next_check_time<R: RngCore, D: DerefMut<Target = R>>(
     Ok(from_event + Duration::new(event_delay, 0))
 }
 
+pub(crate) fn schedule_next_check<P: consensus::Parameters, C: Clock, R: RngCore>(
```
str4d, line 1481: It would be helpful to have a test of this, pinning the intended behaviour.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1379,6 +1380,40 @@ pub(crate) fn next_check_time<R: RngCore, D: DerefMut<Target = R>>(
...
+    let addr_str = address.encode(params);
+    let start_time = clock.now();
+    let next_check = next_check_time(&mut rng, start_time, offset_seconds)?;
+    let scheduled_next_check = conn
+        .query_row(
+            "UPDATE addresses
+             SET transparent_receiver_next_check_time = MIN(
+                :next_check,
+                MAX(:next_check, IFNULL(transparent_receiver_next_check_time, :next_check))
+             )
```
str4d: Same question: I understand the purpose of the `MIN` but not the `MAX`. If both are present then @daira's simplification applies, but I don't think both are intended to be present.

zcash_client_backend/CHANGELOG.md
```diff
@@ -38,6 +43,16 @@ workspace.
...
 
+### Removed
+- `zcash_client_backend::data_api::WalletRead::get_known_ephemeral_addresses`
+  has been removed. It was previously only used in tests and is not well-designed,
+  for wallet use; a replacement for use in tests has been added to the `WalletTest`
+  trait.
+- `zcash_client_backend::data_api::WalletTest::find_account_for_ephemeral_address`
+  has been removed. It was previously only used in tests and is not well-designed,
+  for wallet use; a replacement for use in tests has been added to the `WalletTest`
+  trait.
```
```suggestion
- `zcash_client_backend::data_api`:
  - `WalletRead::{get_known_ephemeral_addresses, find_account_for_ephemeral_address}`
    have been removed. They were previously only used in tests and are not
    well-designed for wallet use; replacements for use in tests have been added
    to the `WalletTest` trait.
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1369,15 +1368,6 @@ where
...
-                }
```
str4d, line 1376: I was confused as to why this code was removed in the "Remove test-only WalletRead methods" commit (61523ff792578eef7e762bb5dee6c3d665d0f13c) until I read the rest of the commit message. The commit message should be reworked so the top-line message indicates this restriction is being removed, because it is a significant change to allowed wallet behaviour (or the commit should be split in two, with this removal happening first).

The change should also be documented in the changelog, as now the only defense against misusing send-to-ephemeral behaviour is the caller being very careful to follow the documentation. The documentation for `propose_shielding` and `shield_transparent_funds` should be similarly updated to clearly specify how to correctly use them with ephemeral addresses.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -2059,29 +2059,6 @@ pub fn send_multi_step_proposed_transfer<T: ShieldedPoolTester, DSF>(
...
-        .unwrap();
-
-    let create_proposed_result = st.create_proposed_transactions::<Infallible, _, Infallible, _>(
-        account.usk(),
-        OvkPolicy::Sender,
-        &proposal,
-    );
-    assert_matches!(
-        &create_proposed_result,
-        Err(Error::PaysEphemeralTransparentAddress(address_str)) if address_str == &ephemeral0.encode(st.network()));
```
str4d, line 2051: If the restriction removal is split out into a separate commit, the test change should go with it.

zcash_client_memory/src/wallet_read.rs
```diff
@@ -655,18 +655,15 @@ impl<P: consensus::Parameters> WalletRead for MemoryWalletDb<P> {
...
                         )
-                        .map(|i| {
-                            TransparentAddressMetadata::derived(
-                                Scope::External.into(),
-                                i,
-                                ExposedAt::Unknown,
-                                None,
-                            )
-                        });
+                        .expect("non-hardened address corresponds to a valid diversifier index"),
```
str4d: The claimed condition in the `.expect()` is backwards. Indeed this claim holds, but the claim we require is that `diversifier_index` will fit into a `NonHardenedChildIndex`.

What we are actually relying on here is the database invariant that if `ua.transparent().is_some()`, then the transparent address was derived using `diversifier_index`, and thus the latter fits into a `NonHardenedChildIndex`. This happens to hold because in `zcash_client_memory`, `account.addresses()` only returns addresses derived by the wallet itself.

zcash_client_backend/CHANGELOG.md
```diff
@@ -12,12 +12,48 @@ workspace.
 
 ## [0.21.0] - PLANNED
 
+### Added
+- `zcash_client_backend::wallet::{Exposure, GapMetadata}`
+- `zcash_client_backend::wallet::TransparentAddressSource`
+- `zcash_client_backend::wallet::TransparentAddressMetadata::`
+    - `derived`
+    - `standalone`
+    - `exposed_at`
```
```suggestion
    - `exposure`
```

zcash_client_backend/src/wallet.rs
```diff
@@ -627,11 +627,45 @@ impl OvkPolicy {
...
+    Exposed {
+        /// The height at which the address was exposed.
+        at_height: BlockHeight,
+        /// As of the time of the query that produced this exposure metadata, the number of
+        /// potentially-exposed addresses at indexes greater than the [`NonHardenedChildIndex`]
+        /// corresponding to the address before the gap limit is exhausted for the address's key
+        /// scope. `None` if the start of the transparent address gap is at an index greater than
+        /// this address's child index; that is, if this address or an address having a greater
+        /// child index within the associated key scope has already received funds.
+        gap_metadata: GapMetadata,
```
str4d: The documentation doesn't match the type. Did you start down one path and document it, then change tack to `GapMetadata` without rewriting the documentation?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -174,10 +194,40 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
...
                     at_height: BlockHeight::from(h),
+                    gap_metadata: gap_limit_starts
+                        .get(&key_scope)
+                        .zip(address_index_opt)
+                        .map_or(
+                            GapMetadata::DerivationUnknown,
+                            |((gap_limit, start), idx)| {
+                                if *start <= idx {
+                                    GapMetadata::InGap {
+                                        gap_position: idx.index() - start.index(),
```
```suggestion
                                if let Some(gap_position) = idx.index().checked_sub(start.index()) {
                                    GapMetadata::InGap {
                                        gap_position,
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1584,8 +1627,39 @@ pub(crate) fn get_transparent_address_metadata<P: consensus::Parameters>(
...
+                                |at_height| {
+                                    let gap_metadata =  match gap_limits.limit_for(key_scope) {
+                                        None => GapMetadata::DerivationUnknown,
+                                        Some(gap_limit) => {
+                                            find_gap_start(conn, account_id, scope, gap_limit)?.map_or(
+                                                GapMetadata::GapRecoverable { gap_limit },
+                                                |gap_start| {
+                                                    if address_index >= gap_start {
+                                                        GapMetadata::InGap {
+                                                            gap_position: address_index.index() - gap_start.index(),
```
```suggestion
                                                    if let Some(gap_position) = address_index.index().checked_sub(gap_start.index()) {
                                                        GapMetadata::InGap {
                                                            gap_position,
```

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -83,7 +97,22 @@ pub(crate) fn get_known_ephemeral_addresses<P: consensus::Parameters>(
...
                     Exposure::Unexposed,
-                    |h| Exposure::Exposed { at_height: BlockHeight::from(h) }
+                    |h| Exposure::Exposed {
+                        at_height: BlockHeight::from(h),
+                        gap_metadata: gap_start.map_or(
+                            GapMetadata::DerivationUnknown,
+                            |gap_start| {
+                                if address_index >= gap_start {
+                                    GapMetadata::InGap {
+                                        gap_position: address_index.index() - gap_start.index(),
```
str4d:
```suggestion
                                if let Some(gap_position) = address_index.index().checked_sub(gap_start.index()) {
                                    GapMetadata::InGap {
                                        gap_position,
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -474,20 +474,20 @@ pub(crate) fn select_addrs_to_reserve<P: consensus::Parameters>(
                 transparent_child_index
                     .zip(address)
                     .map(|(i, a)| {
+                        // If we have a transparent child index, we must know the transparent key
+                        // scope & be able to derive for it.
+                        let key_scope = <Option<TransparentKeyScope>>::from(key_scope)
+                            .ok_or(SqliteClientError::UnknownZip32Derivation)?;
```
str4d: I am very confused by this: the input `key_scope` is already of type `TransparentKeyScope`, so this conversion should be a no-op.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: This ended up being a bit of an adventure: 74455091b63a1a6b197bd735a21bf7c23cba4997

zcash_client_backend/src/data_api.rs
nuttycom: This is now a member of `WalletTest`. It is used in a number of test cases, and it is not redundant given that it allows more unrestricted access to addresses.

zcash_client_backend/src/wallet.rs
nuttycom: I think that the absolute index should be irrelevant to the user.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
nuttycom: `:next_check` will never be null.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
nuttycom: I just realized that this needs to take the current time into account to make sense. If a preexisting check time has passed, relative to the system clock, then the check should be rescheduled forward; otherwise, it should take the minimum.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: At some point in the past, this was `KeyScope` and this was residue that wasn't fixed whenever that change was made (see removed line 394, that this was adapted from.)

zcash_client_backend/src/data_api/wallet.rs
nuttycom: We still have a check that prevents ephemeral address reuse when sending *to* an ephemeral address, so I don't think that we're entirely without defenses here. I don't have a clear view of what would need to be changed in the documentation of `propose_shielding` or `shield_transparent_funds`; in both cases, they're just shielding from a transparent address, and an ephemeral address is not special with respect to avoiding linking addresses.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Yes, we do not create a next-check time when exposing an address for the first time; all check scheduling is done independent of generation/reservation.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: The reason that I created a distinction between `Unexposed` and `Unknown` was to use `Unknown` for imported transparent addresses, and `Unexposed` for derived addresses. It is not intended that `Unexposed` be considered definitive, because we can know with certainty that an address has not been exposed by some other wallet using a different gap limit. We we could potentially rename `Unknown` to `Imported` or something of the sort and its semantics would be unchanged.

str4d: In that case, `Unexposed` needs to be renamed, because its current name is misleading.

zcash_client_backend/src/data_api.rs
```diff
@@ -1734,23 +1764,10 @@ pub trait WalletRead {
     #[cfg(feature = "transparent-inputs")]
```
daira, line 1846: The equivalent code in the doc comment is no longer valid; it should be changed to use `get_ephemeral_transparent_receivers` or removed.

zcash_client_backend/CHANGELOG.md
```diff
@@ -15,16 +15,49 @@ workspace.
...
 - `zcash_client_backend::tor::http::cryptex::exchanges::GateIo`
+- `zcash_client_backend::data_api`:
+  - `WalletRead::{get_known_ephemeral_addresses, find_account_for_ephemeral_address}`
+    have been removed. They were previously only used in tests and are not
+    well-designed for wallet use; replacements for use in tests have been added
+    to the `WalletTest` trait.
+- `zcash_client_backend::data_api::error::Error::PaysEphemeralTransparentAddress`
+  has been removed; the wallet no longer restricts the caller from constructing
+  payments to explicit ephemeral transparent addresses, in order to allow
+  for gap limit management operations to be performed by the wallet application.
```
daira, line 79:
```suggestion
  for gap limit management operations to be performed by the wallet application.
  If it is desired to maintain the previous policy of excluding explicit payments
  to ephemeral addresses, call `get_transparent_address_metadata` on the address
  and check whether its `scope` is `TransparentKeyScope::EPHEMERAL` if it is `Derived`.
```

zcash_client_backend/src/data_api.rs
daira: I think those tests are mainly testing `get_known_ephemeral_addresses`, aren't they?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -1891,12 +1874,8 @@ pub fn send_multi_step_proposed_transfer<T: ShieldedPoolTester, DSF>(
 
         let total_sent = (expected_step0_fee + expected_step1_fee + transfer_amount).unwrap();
 
-        // Generate a ZIP 320 proposal, sending to the wallet's default transparent address
+        // Generate a ZIP 320 proposal, sending to the another wallet's default transparent address
```
```suggestion
        // Generate a ZIP 320 proposal, sending to another wallet's default transparent address
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -2424,12 +2376,8 @@ pub fn spend_all_funds_multi_step_proposed_transfer<T: ShieldedPoolTester, DSF>(
 
     let total_sent = (expected_step0_fee + expected_step1_fee + transfer_amount).unwrap();
 
-    // Generate a ZIP 320 proposal, sending to the wallet's default transparent address
-    // expressed as a TEX address.
-    let tex_addr = match default_addr {
-        TransparentAddress::PublicKeyHash(data) => Address::Tex(data),
-        _ => unreachable!(),
-    };
+    // Generate a ZIP 320 proposal, sending to the an external TEX address.
```
```suggestion
    // Generate a ZIP 320 proposal, sending to an external TEX address.
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -2579,13 +2527,8 @@ pub fn proposal_fails_if_not_all_ephemeral_outputs_consumed<T: ShieldedPoolTeste
...
     add_funds(&mut st, value);
 
-    // Generate a ZIP 320 proposal, sending to the wallet's default transparent address
-    // expressed as a TEX address.
-    let tex_addr = match account.usk().default_transparent_address().0 {
-        TransparentAddress::PublicKeyHash(data) => Address::Tex(data),
-        _ => unreachable!(),
-    };
-
+    // Generate a ZIP 320 proposal, sending to the an external TEX address.
```
```suggestion
    // Generate a ZIP 320 proposal, sending to an external TEX address.
```

zcash_client_backend/src/wallet.rs
daira: How will the policy of sending a dummy output to extend the gap limit be implemented in that case? Don't we need the child index in order to derive the address to send to?

daira: Oh, I guess there is sufficient information available from `get_ephemeral_transparent_receivers` and the `TransparentAddressSource` of each address' metadata. But it's a bit inconvenient to have to get it that way when we could just include it here.

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +627,153 @@ impl OvkPolicy {
...
+        ///   will be assigned.
+        at_height: BlockHeight,
+        /// Transparent address gap metadata, as of the time the query that produced this exposure
+        /// metadata was executed.
+        gap_metadata: GapMetadata,
+    },
+    /// The address is not known to have been exposed to an external caller by the wallet. This is
+    /// the state that an address is in when it has been generated by the advancement of the
+    /// transparent address gap; this judgement may be incorrect for restored wallets.
+    Unexposed,
```
daira: Maybe rename to `UnexposedByThisWallet`?

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +627,153 @@ impl OvkPolicy {
...
+#[cfg(feature = "transparent-inputs")]
+pub struct TransparentAddressMetadata {
+    source: TransparentAddressSource,
+    exposure: Exposure,
+    next_check_time: Option<SystemTime>,
+}
+
+#[cfg(feature = "transparent-inputs")]
+impl TransparentAddressMetadata {
+    /// Constructs a new [`TransparentAddressMetadata`] value from its constitutent parts.
```
daira:
```suggestion
    /// Constructs a new [`TransparentAddressMetadata`] value from its constituent parts.
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -111,30 +116,76 @@ fn address_index_from_diversifier_index_be(
...
-         AND key_scope IN rarray(:scopes_ptr)",
+         AND key_scope IN rarray(:scopes_ptr)
+         AND (
+             :min_exposure_height IS NULL
+             OR exposed_at_height >= :min_exposure_height
+         )
+         AND (
+             NOT(:exclude_used)
+             -- if we're only retrieving unused addresses, do not return those for which we have
+             -- observed an output.
```
daira, line 177: Does this match an intuitive idea of "used"? Specifically, addresses that we've used in transactions we've sent are intuitively "used", are those included in `transparent_received_outputs`?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1311,6 +1462,42 @@ pub(crate) fn next_check_time<R: RngCore, D: DerefMut<Target = R>>(
...
+    offset_seconds: u32,
+) -> Result<Option<SystemTime>, SqliteClientError> {
+    let addr_str = address.encode(params);
+    let now = clock.now();
+    let next_check = next_check_time(&mut rng, now, offset_seconds)?;
+    let scheduled_next_check = conn
+        .query_row(
+            "UPDATE addresses
+             SET transparent_receiver_next_check_time = CASE
+                WHEN transparent_receiver_next_check_time < :current_time THEN :next_check
```
daira, line 1496: Isn't there a race condition here? The current time being after `transparent_receiver_next_check_time` isn't equivalent to us actually having processed the event that is supposed to be at `transparent_receiver_next_check_time`.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1448,28 +1635,76 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+                    .map_err(|e| rusqlite::Error::ToSqlConversionFailure(Box::new(e)))?;
+
                 Ok(KeyScope::decode(scope_code).and_then(|key_scope| {
                     let address_index = address_index_from_diversifier_index_be(row.get("diversifier_index_be")?)?;
+                    let exposed_at_height = row.get::<_, Option<u32>>("exposed_at_height")?.map(BlockHeight::from);
+
                     match <Option<TransparentKeyScope>>::from(key_scope).zip(address_index) {
                         Some((scope, address_index)) => {
-                            Ok(TransparentAddressMetadata::Derived {
+                            let exposure = exposed_at_height.map_or(
```
daira, line 1689: My eyes are glazing over at this point and I know I'm not giving this the thoroughness of review it needs.

daira (dismissed):
There is a comment about code I haven't reviewed sufficiently, but utACK for the rest.

zcash_transparent/src/keys.rs
```diff
@@ -49,6 +51,17 @@ impl TransparentKeyScope {
     pub const EPHEMERAL: Self = TransparentKeyScope(2);
 }
 
+impl fmt::Debug for TransparentKeyScope {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            Self::EXTERNAL => f.write_str("EXTERNAL"),
+            Self::INTERNAL => f.write_str("INTERNAL"),
+            Self::EPHEMERAL => f.write_str("EPHEMERAL"),
+            TransparentKeyScope(other) => f.write_str(&format!("Custom({other})")),
```
str4d: This makes the standalone debug impl more confusing as you would just see the string "EXTERNAL" without any type context. We should strike a middle ground:
```suggestion
            Self::EXTERNAL => f.write_str("TransparentKeyScope::EXTERNAL"),
            Self::INTERNAL => f.write_str("TransparentKeyScope::INTERNAL"),
            Self::EPHEMERAL => f.write_str("TransparentKeyScope::EPHEMERAL"),
            TransparentKeyScope(other) => f.write_str(&format!("TransparentKeyScope({other})")),
```

zcash_transparent/src/keys.rs
```diff
@@ -67,9 +80,15 @@ impl From<TransparentKeyScope> for ChildNumber {
 /// A child index for a derived transparent address.
 ///
 /// Only NON-hardened derivation is supported.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct NonHardenedChildIndex(u32);
 
+impl core::fmt::Debug for NonHardenedChildIndex {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&format!("{}", self.0))
```
str4d: Similarly, I disagree with this change because the standalone debug string now has no type context.

zcash_transparent/src/address.rs
```diff
@@ -115,12 +115,27 @@ impl From<&script::Sig> for Script {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::PublicKeyHash(arg0) => f
+                .debug_tuple("PublicKeyHash")
+                .field(&hex::encode(arg0))
+                .finish(),
+            Self::ScriptHash(arg0) => f
+                .debug_tuple("ScriptHash")
+                .field(&hex::encode(arg0))
+                .finish(),
```
str4d, line 134: These are at least meaningful-enough names that dropping the `TransparentAddress::` prefix isn't too bad, but I weakly disagree with doing so.

OTOH hex-encoding the argument is a very good change for reducing noise.

zcash_client_backend/CHANGELOG.md
```diff
@@ -49,14 +49,15 @@ workspace.
...
-  trait.
+- `zcash_client_backend::data_api`:
+  - `WalletRead::{get_known_ephemeral_addresses, find_account_for_ephemeral_address}`
+    have been removed. They were previously only used in tests and are not
+    well-designed for wallet use; replacements for use in tests have been added
+    to the `WalletTest` trait.
+- `zcash_client_backend::data_api::error::Error::PaysEphemeralTransparentAddress`
+  has been removed; the wallet no longer restricts the caller from constructing
+  payments to explicit ephemeral transparent addresses, in order to allow
+  for gap limit management operations to be performed by the wallet application.
```
str4d, line 79: Very minor nit: this was included in the "Apply suggestions" commit, before the variant was removed.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2898,11 +2899,44 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
             Recipient::External {
-                recipient_address: _addr,
+                recipient_address: _zaddr,
                 output_pool: _pool,
-                ..
             } => {
-                // Nothing to do for external recipients.
+                // In the case that a transaction sends to a transparent address belonging to the
+                // wallet (such as is the case for gap limit management transactions) then we need
+                // to add the received transparent output to our wallet. For shielded outputs,
```
str4d: Unfinished sentence.

zcash_client_backend/src/wallet.rs
```diff
@@ -624,10 +627,153 @@ impl OvkPolicy {
...
+        /// Transparent address gap metadata, as of the time the query that produced this exposure
+        /// metadata was executed.
+        gap_metadata: GapMetadata,
+    },
+    /// The address is not known to have been exposed to an external caller by the wallet. This is
+    /// the state that an address is in when it has been generated by the advancement of the
+    /// transparent address gap; this judgement may be incorrect for restored wallets.
+    Unexposed,
+    /// It is not known whether the address has been exposed.
+    Unknown,
```
str4d: Per the conversation in our pairing today:
```suggestion
    /// The address is not known to have been exposed to an external caller by the wallet.
    ///
    /// The wallet makes its determination based on observed chain data and inference from
    /// standard wallet address generation patterns. In particular, this is the state that
    /// an address is in when it has been generated by the advancement of the transparent
    /// address gap. This judgement may be incorrect for restored wallets.
    Unknown,
    /// It is not possible for the wallet to determine whether the address has been exposed,
    /// given the information the wallet has access to.
    CannotKnow,
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1903,6 +1968,15 @@ pub(crate) fn put_transparent_output<P: consensus::Parameters>(
...
         .query_row(sql_args, |row| row.get::<_, i64>(0).map(UtxoId))?;
 
+    #[cfg(feature = "transparent-inputs")]
+    update_gap_limits(
+        conn,
+        params,
+        gap_limits,
+        *output.outpoint().txid(),
+        output_height.map_or(observation_height, BlockHeight::from),
+    )?;
```
str4d: Possible bug: I think this needs to go below the conditional `mark_transparent_utxo_spent()` below, otherwise the maps that `update_gap_limits` relies on are not fully current (because spends might still be cached in `transparent_spend_map`) and we won't correctly update gap limits for the involved downstream addresses.

zcash_client_backend/src/wallet.rs
nuttycom: The metadata is returned with/for a specific address; you don't have to re-derive the address, you can just use it.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: An output that is sent to an address in our wallet is also an output that is received by our wallet, so yes, it ends up in `transparent_received_outputs`.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: `transparent_receiver_next_check_time` is intended to serve as a _lower bound_ on the times at which the check can be performed. It's best interpreted as "do not check before this time." The reason for this interpretation is that it's not possible for mobile wallets to schedule checks at exact times; a wallet application may be not be able to perform any operations at the scheduled time (it could be offline, the application could be closed, etc.). 

All of this is a hacky workaround that attempts to de-correlate the ephemeral transparent address checks from other operations of the wallet in the absence of transparent output information being included in compact blocks. In practical use, this will be called to establish the once-a-day checks for ephemeral addresses for returned TEX funds, and immediately after a check has been performed (with back-off) after providing an ephemeral address to receive swap outputs. See https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/244/files#diff-f64796ccc388f201986ea7eaaf07bbefbd8ac1a56bb995fb505da80879872a39R3514 for the call site that uses this operation.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2898,11 +2899,44 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
...
+                #[cfg(feature = "transparent-inputs")]
+                if _pool == &PoolType::Transparent {
+                    let address = Address::try_from_zcash_address(params, _zaddr.clone())
+                        .expect("recipient is an understood Zcash address.");
+                    if let Some(taddr) = address.to_transparent_address() {
+                        if transparent::find_account_uuid_for_transparent_address(
```
str4d, line 2916: It's a bit weird handling this under `Recipient::External`, but I guess it's fine for now.

str4d (changes requested):
Reviewed 13b5a174ede06e642d1f0a042ed88f2cb208d1c7

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Gap limits are only advanced as a consequence of transactions being _mined_, so I don't think that the order of the operation with respect to marking the UTXO spent matters? I'll double check though.

zcash_transparent/src/keys.rs
nuttycom:
```suggestion
        f.write_str(&format!("NonHardenedChildIndex({})", self.0))
```

zcash_client_backend/src/data_api/wallet.rs
nuttycom: In pairing, @str4d and I discussed that we should make it a runtime error to attempt to shield from more than on address if any of the addresses being shielded from is an ephemeral address.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: It's a minor enough change that I'll just do it.

zcash_client_sqlite/src/wallet.rs
nuttycom: Yeah; the thing is, from the perspective of the ZIP 321 request, this _is_ an external recipient. It just happens to belong to the wallet, so we check that here.

zcash_client_backend/src/data_api/wallet.rs
daira: I'm confused, did I misunderstand the changelog message or this code change? I thought that this was precisely about changing the API to allow sending *to* an address that happens to be ephemeral. For reference, the changelog message is:

> - `zcash_client_backend::data_api::error::Error::PaysEphemeralTransparentAddress`
>   has been removed; the wallet no longer restricts the caller from constructing
>   payments to explicit ephemeral transparent addresses, in order to allow
>   for gap limit management operations to be performed by the wallet application.

I know that we thought about this carefully for the original code and intentionally disallowed it.

See this comment for a possible resolution: https://github.com/zcash/librustzcash/pull/1969/files#r2441454057

daira (dismissed):
utACK 81491e86eab9cfbab9d56cd900b8a92c9759975b

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1039,20 +1039,37 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+
+                ephemeral_addrs.extend(utxos.iter().filter_map(|utxo| {
+                    (utxo.recipient_key_scope() == Some(TransparentKeyScope::EPHEMERAL))
+                        .then_some(utxo.recipient_address())
+                }));
+                input_addrs.extend(utxos.iter().map(|utxo| utxo.recipient_address()));
+                inputs.extend(utxos.into_iter().map(|utxo| utxo.into_wallet_output()));
+
+                // Funds may be spent from at most one ephemeral address at a time. If there are no
+                // ephemeral addresses, we allow shielding from multiple transparent addresses.
```
daira:
```suggestion
                // Funds may be spent from at most one ephemeral address at a time. If no recipients
                // are ephemeral addresses, we allow shielding from multiple transparent addresses.
```

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/81491e86eab9cfbab9d56cd900b8a92c9759975b..615fc8977b62f71c0e3185b37b4282b4343a1209)  to address comments from code review then [force_pushed](https://github.com/zcash/librustzcash/compare/615fc8977b62f71c0e3185b37b4282b4343a1209..8bd3802b7e9f2a03ded3fbfa11e0f7f09a17dabd) to rebase on `main`.

daira:
Why couldn't `recipient_key_scope()` just be added to `WalletTransparentOutput`, rather than adding the `WalletUtxo` type?

zcash_client_backend/src/data_api.rs
```diff
@@ -984,6 +984,83 @@ impl<NoteRef> ReceivedNotes<NoteRef> {
...
 }
 
+/// An unspent transparent output belonging to the wallet, along with derivation key metadata (if
+/// available).
+#[derive(Clone, Debug)]
+#[cfg(feature = "transparent-inputs")]
+pub struct WalletUtxo {
+    wallet_output: WalletTransparentOutput,
+    recipient_key_scope: Option<TransparentKeyScope>,
+}
```
daira, line 994: Why couldn't `recipient_key_scope()` just be added to `WalletTransparentOutput`, rather than adding the `WalletUtxo` type?

zcash_client_backend/src/data_api.rs
daira: This comment is still valid after recent pushes.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1039,20 +1039,37 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+                if !ephemeral_addrs.is_empty() && input_addrs.len() > 1 {
```
daira, line 1070: This doesn't take into account transparent recipient addresses belonging to this wallet. That is, arguably the policy should be that the union of source and recipient addresses should not include multiple transparent addresses belonging to this wallet, if at least one of them is ephemeral. Note that ZIP 320 satisfies this, as long as we are paying to an external TEX address.

zcash_client_backend/src/wallet.rs
daira, line 177:
```suggestion
    recipient_address: TransparentAddress,
    recipient_key_scope: Option<TransparentKeyScope>,
```

daira, line 188:
```suggestion
        mined_height: Option<BlockHeight>,
        recipient_key_scope: Option<TransparentKeyScope>,
```

daira, line 196:
```suggestion
                recipient_address,
                recipient_key_scope,
```

zcash_client_backend/src/wallet.rs
```diff
@@ -189,22 +197,27 @@ impl WalletTransparentOutput {
...
 
+    /// Returns the height at which the UTXO was mined, if any.
     pub fn mined_height(&self) -> Option<BlockHeight> {
         self.mined_height
     }
 
+    /// Returns the wallet address that received the UTXO.
     pub fn recipient_address(&self) -> &TransparentAddress {
         &self.recipient_address
     }
```
daira, line 219:
```suggestion

    /// Returns the transparent key scope at which this address was derived, if known.
    pub fn recipient_key_scope(&self) -> Option<TransparentKeyScope> {
        self.recipient_key_scope
    }

```

zcash_client_backend/src/wallet.rs
```diff
@@ -174,6 +178,10 @@ pub struct WalletTransparentOutput {
 }
 
 impl WalletTransparentOutput {
+    /// Constructs a new [`WalletTransparentOutput`] from its constituent parts.
+    ///
+    /// Returns None if the recipient address for the provided [`TxOut`] cannot be determined
+    /// based on the set of output script patterns understood by this wallet.
```
daira:
```suggestion
    /// based on the set of output script patterns understood by this wallet.
    ///
    /// `recipient_key_scope` MUST be provided for any transparent address derived by the wallet;
    /// this metadata is used by `propose_shielding` to ensure that shielding transactions do not
    /// inadvertently link ephemeral addresses to other wallet activity on-chain.
```

zcash_client_backend/src/data_api.rs
daira: I checked all the places where we call `WalletTransparentOutput::from_parts` outside test code. It seems like we should know the scope for our own addresses in all of those places (there are six of them).

zcash_client_backend/src/wallet.rs
daira: What I mean is that if you have the metadata from a given address in the gap but you want the one at a *different* position relative to the gap, you have to find the `address_index` from that metadata (if known), add the desired offset from `gap_position`, and then derive the address at the desired position. I guess that isn't so bad now that I write it out.

daira (commented):
Re-reviewed 8bd3802b7e9f2a03ded3fbfa11e0f7f09a17dabd.

zcash_client_backend/src/data_api.rs
nuttycom: This is not the case; we use `WalletTransparentOutput` [here](https://github.com/zcash/librustzcash/blob/main/zcash_client_backend/src/sync.rs#L510) to represent the results of the `GetAddressUtxosStream` GRPC call. We do not have the key scope for the address at this point. In addition, having a wrapper type here gives us more flexibility to add other metadata (such as might be required for working with P2SH multisig) in the future.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: This code is explicitly in the `propose_shielding` method; there is no possibility of there being a transparent recipient here.

zcash_client_backend/src/wallet.rs
nuttycom: See https://github.com/zcash/librustzcash/pull/1969#discussion_r2445074521

nuttycom: See https://github.com/zcash/librustzcash/pull/1969#discussion_r2445074521

nuttycom: See https://github.com/zcash/librustzcash/pull/1969#discussion_r2445074521

zcash_client_backend/src/wallet.rs
nuttycom: See https://github.com/zcash/librustzcash/pull/1969#discussion_r2445074521

zcash_client_backend/src/wallet.rs
nuttycom: See https://github.com/zcash/librustzcash/pull/1969#discussion_r2445074521

zcash_client_backend/src/proposal.rs
```diff
@@ -61,6 +61,9 @@ pub enum ProposalError {
     /// change output when needed for sending to a TEX address.
     #[cfg(feature = "transparent-inputs")]
     EphemeralOutputsInvalid,
+    /// The request to generate a proposal that would link activity on
```
str4d:
```suggestion
    /// The requested proposal would link activity on an ephemeral address to other wallet activity.
```

zcash_client_backend/src/wallet.rs
```diff
@@ -174,6 +178,10 @@ pub struct WalletTransparentOutput {
 }
 
 impl WalletTransparentOutput {
+    /// Constructs a new [`WalletTransparentOutput`] from its constituent parts.
+    ///
+    /// Returns None if the recipient address for the provided [`TxOut`] cannot be determined
```
str4d:
```suggestion
    /// Returns `None` if the recipient address for the provided [`TxOut`] cannot be determined
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1039,20 +1039,37 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+                BTreeSet::<TransparentAddress>::new(),
+            ),
+            |(mut inputs, mut ephemeral_addrs, mut input_addrs), taddr| {
+                use transparent::keys::TransparentKeyScope;
+
+                let utxos = wallet_db
+                    .get_spendable_transparent_outputs(taddr, target_height, confirmations_policy)
+                    .map_err(InputSelectorError::DataSource)?;
+
+                ephemeral_addrs.extend(utxos.iter().filter_map(|utxo| {
```
str4d, line 1061:
```suggestion
                // `InputSource::get_spendable_transparent_outputs` is required to return
                // outputs received by `taddr`, so these `.extend()` calls are guaranteed
                // to add at most a single new address to each set. But it's more
                // convenient this way as we can reuse `utxo.recipient_key_scope()`
                // instead of needing to query the wallet twice for each address to
                // determine their scopes.
                ephemeral_addrs.extend(utxos.iter().filter_map(|utxo| {
```

zcash_client_backend/src/data_api.rs
```diff
@@ -984,6 +984,83 @@ impl<NoteRef> ReceivedNotes<NoteRef> {
...
+
+    /// Returns the value of the UTXO
+    pub fn value(&self) -> Zatoshis {
+        self.wallet_output().value()
+    }
+
+    /// Returns the transparent key scope at which this address was derived, if known. This
+    /// metadata MUST be returned for any transparent address derived by the wallet; this metadata
+    /// is used by `propose_shielding` to ensure that shielding transactions do not inadvertently
+    /// link ephemeral addresses to other wallet activity on-chain.
```
```suggestion
    /// Returns the transparent key scope at which this address was derived, if known.
    ///
    /// This metadata MUST be returned for any transparent address derived by the wallet;
    /// this metadata is used by `propose_shielding` to ensure that shielding transactions
    /// do not inadvertently link ephemeral addresses to other wallet activity on-chain.
```

zcash_client_backend/src/data_api.rs
str4d: Fixed example code:
```rust
    ///     if let Some(result) = self.get_transparent_receivers(account, true)?.get(address) {
    ///         Some(result.clone())
    ///     } else {
    ///         self.get_ephemeral_transparent_receivers(account, u32::MAX, false)?
    ///             .get(address)
    ///             .cloned()
    ///     },
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -113,30 +117,76 @@ fn address_index_from_diversifier_index_be(
...
-    let mut ret: HashMap<TransparentAddress, Option<TransparentAddressMetadata>> = HashMap::new();
+    exposure_depth: Option<u32>,
+    exclude_used: bool,
+) -> Result<HashMap<TransparentAddress, TransparentAddressMetadata>, SqliteClientError> {
+    let mut ret: HashMap<TransparentAddress, TransparentAddressMetadata> = HashMap::new();
+
+    let min_exposure_height = exposure_depth
+        .map(|d| {
+            Ok::<_, SqliteClientError>(
+                mempool_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)? - d,
```
str4d: While updating the example code, I noticed this bug:
```suggestion
                mempool_height(conn)?
                    .ok_or(SqliteClientError::ChainHeightUnknown)?
                    .saturating_sub(d),
```

str4d (changes requested):
Reviewed fa91bfe3bcdc9665ddb36a2efcd4cab9581fc277. One bug, otherwise utACK (though I did get a bit lost in the rebases; hopefully I saw all the changes from my last review).

nuttycom (commented):
ACK 1c967e6cfe9cdd8891aac9f7e1996628203f7608

PR #1970, opened 2025-09-28
zcash_client_sqlite: Enable zero-conf UTXO selection for shielding.

nuttycom:


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -804,9 +804,8 @@ pub(crate) fn get_wallet_transparent_output(
              OR (
                  (
                     t.mined_height IS NOT NULL -- tx is mined
-                    -- TODO: uncomment the following two lines in order to enable zero-conf spends
-                    -- OR t.expiry_height = 0 -- tx will not expire
-                    -- OR t.expiry_height >= :mempool_height -- tx has not yet expired
+                    OR t.expiry_height = 0 -- tx will not expire
+                    OR t.expiry_height >= :mempool_height -- tx has not yet expired
```
schell, line 808: So why was this commented out @nuttycom? I saw this during my confirmations work, but didn't point it out as it was an historical TODO.

schell (approved):
UNTODO: 👍

zcash_client_sqlite/src/wallet/transparent.rs
str4d: This was initially added commented-out in #1402 in what was called `wallet::transparent::get_unspent_transparent_output` (renamed in #1462).

That same PR also added a similar commented-out line in `get_spendable_transparent_outputs`, but that instance was removed by @nuttycom in 8f9efaf621c47d0841a2f07ef4051462ad720e7b as part of the confirmations policy PR #1888.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/944d69b2e9ef92efba1f134192bdc2d4d2ee3bb9..308cc99edd71ba66c3dcf23b56314e002f6e3e49) to fix `cargo vet` metadata.

zcash_client_sqlite/src/wallet/transparent.rs
str4d, line 808: I checked that this matches the logic in `wallet::transparent::get_spendable_transparent_outputs`.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -804,9 +804,8 @@ pub(crate) fn get_wallet_transparent_output(
              OR (
                  (
                     t.mined_height IS NOT NULL -- tx is mined
```
str4d, line 806: I note that this uses the previous approach for checking "is not mined", whereas you migrated to a different approach in `get_spendable_transparent_outputs`: https://github.com/zcash/librustzcash/pull/1888#discussion_r2279436790

We should ensure that both methods agree on what is unspendable. In `get_wallet_transparent_output` we specifically mean consensus-unspendable, while `get_spendable_transparent_outputs` also takes a confirmation policy, so I think what we want here is that when `allow_unspendable = false`, this method should only return a UTXO if `get_spendable_transparent_outputs` would have returned it for `confirmations_policy.allow_zero_conf_shielding() = true`. Given that the latter means `min_confirmations = 0`, I believe the logic in `get_spendable_transparent_outputs` simplifies down to the same as here, as long as `t.mined_height IS NOT NULL` is guaranteed to filter the same as `t.mined_height < :target_height`.

Separate nit: we should be consistent on using either `target_height` or `mempool_height` across queries.

nuttycom: I think for consistency it would be fine to change this to `t.mined_height < :target_height` (and use `target_height` consistently) but I believe that there's no semantic difference here. Since this is a hotfix, I don't want to make any API changes here but I think that for an upcoming semver-breaking release it would be good to deprecate `get_unspent_transparent_output` and add a `get_spendable_transparent_output` method that follows `get_spendable_transparent_outputs` more closely.  In addition, I would like to move all of the `transparent-inputs`-flagged methods out into a separate trait; it's going to be time to do a broader `data_api` cleanup some time soon, so I think any such changes can be deferred until then.

nuttycom: Opened https://github.com/zcash/librustzcash/issues/1971

daira (commented):
Post-hoc ACK

PR #1972, opened 2025-09-29
[WIP] split zip321

schell:


components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,924 @@
...
+
+    Ok(payment)
+}
+
+/// A single payment being requested.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Payment {
+    // TODO(schell): changed
+    /// The address to which the payment should be sent.
+    recipient_address: String,
```
schell, line 332: The type here changed from `ZcashAddress` to `String`. 

The `zip321::Payment` type wraps this one and adds `ZcashAddress` on top, validating it.

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,924 @@
...
+    // TODO(schell): changed, make separate types for whole and fractional
+    /// The amount of the payment that is being requested.
+    amount_coins: u64,
+    amount_zatoshis: u64,
```
schell: Similarly these fields came from `amount: Zatoshis`.

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,924 @@
...
+    // TODO(schell): changed
+    /// A memo that, if included, must be provided with the payment.
+    /// If a memo is present and [`recipient_address`] is not a shielded
+    /// address, the wallet should report an error.
+    ///
+    /// [`recipient_address`]: #structfield.recipient_address
+    memo: Option<Memo>,
```
schell, line 337: Same deal here, coming from `MemoBytes`.

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,924 @@
...
+    ) -> Self {
+        Self {
+            recipient_address,
+            amount_coins: amount.0,
+            amount_zatoshis: amount.1,
+            memo,
+            label,
+            message,
+            other_params,
+        }
```
schell, line 378: Notice here that there's no check to ensure the recipient address can receive a memo. This is by design since we don't have enough information to do this. The check still happens in `zip321::Payment::new`.

components/zip321/src/lib.rs
```diff
@@ -8,108 +8,68 @@ use core::fmt::Debug;
...
     /// The payment at the wrapped index attempted to include a memo when sending to a
     /// transparent recipient address, which is not supported by the protocol.
     TransparentMemo(usize),
-    /// The payment at the wrapped index did not include a recipient address.
-    RecipientMissing(usize),
     /// The ZIP 321 URI was malformed and failed to parse.
-    ParseError(String),
+    ParseError(zip321_parse::Error),
+    /// Any other stringly typed error
+    Other(String),
```
schell, line 32: I split `ParseError` up into two variants - most have moved to `zip321_parser` but a few validation messages remain.

components/zip321/src/lib.rs
```diff
@@ -186,29 +199,22 @@ impl Payment {
...
-    pub(crate) fn normalize(&mut self) {
-        self.other_params.sort();
-    }
+/// Combine ZEC and a fractional remainder of zatoshis.
+fn combine_zatoshis((zec, zats): (u64, u64)) -> Result<Zatoshis, Zip321Error> {
+    zec.checked_mul(COIN)
+        .and_then(|coin_zats| coin_zats.checked_add(zats))
+        .ok_or(BalanceError::Overflow)
+        .and_then(Zatoshis::from_u64)
+        .map_err(|_| Zip321Error::Other(format!("Not a valid zat amount: {zec}.{zats}")))
```
schell, line 217: I lifted this logic out of the parser, which now keeps ZEC and the fractional zats separately. `zip321::Payment` still contains `Zatoshis` and uses this logic for validation.

components/zip321/src/lib.rs
```diff
@@ -123,17 +83,62 @@ pub struct Payment {
...
-    /// of the transaction request.
-    label: Option<String>,
-    /// A human-readable message to be displayed to the user describing the
-    /// purpose of this payment.
-    message: Option<String>,
-    /// A list of other arbitrary key/value pairs associated with this payment.
-    other_params: Vec<(String, String)>,
+
+    inner: zip321_parse::Payment,
+}
```
schell, line 88: `zip321::Payment` is essentially a pointer to a parsed `zip321_parse::Payment`, and derefs as such. It contains the validated domain specific types for address, amount and memo on top.

components/zip321/src/lib.rs
```diff
@@ -219,563 +225,70 @@ impl Payment {
 /// payment value in the request.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TransactionRequest {
-    payments: BTreeMap<usize, Payment>,
+    inner: zip321_parse::TransactionRequest<Payment>,
 }
```
schell, line 229: Much like `zip321::Payment`, `zip321::TransactionRequest` is a smart pointer to a parse `zip321_parse::TransactionRequest<zip321::Payment>`.

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+            _ => {
+                let query_params = self
+                    .payments
+                    .iter()
+                    .flat_map(|(i, payment)| {
+                        let idx = if *i == 0 { None } else { Some(*i) };
+                        let primary_address = payment.as_ref().recipient_address.clone();
+                        std::iter::empty()
+                            // TODO(schell): mention changed
+                            .chain(Some(render::addr_param_encoded(&primary_address, idx)))
```
schell, line 602: This changed when it was moved to _not_ encode the address, as a `zip321_parse::Payment`'s address is already encoded (it doesn't get decoded and validated until the "last mile").

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,903 @@
+use core::fmt::Debug;
+use std::collections::BTreeMap;
+
+use base64::{prelude::BASE64_URL_SAFE_NO_PAD, Engine};
```
daira, line 4: At the time of writing of https://eprint.iacr.org/2022/361.pdf, the `base64` crate did not check canonicity. We do want to check canonicity, because base64 is [defined to require it at least for encoders](https://datatracker.ietf.org/doc/html/rfc4648#section-3.5), even though the RFC uses unfortuitously loose wording for decoders. (It shows signs of being a compromise between compatibility with an earlier Postel's Law spec, and later spec writers who understood the security relevance of canonicity.)

`base64::prelude::BASE64_URL_SAFE_NO_PAD` is defined as `GeneralPurpose::new(&alphabet::URL_SAFE, NO_PAD)`. `NO_PAD` is defined like this:

```rust
pub const NO_PAD: GeneralPurposeConfig = GeneralPurposeConfig::new()
    .with_encode_padding(false)
    .with_decode_padding_mode(DecodePaddingMode::RequireNone);
```

`GeneralPurposeConfig::new()` [sets `decode_allow_trailing_bits` to `false`](https://docs.rs/base64/0.22.1/src/base64/engine/general_purpose/mod.rs.html#252-264), and so the decoder checks canonicity. The problem is that this is only a property of the current implementation and is not actually documented. It should be documented; that's a bug. Filed https://github.com/marshallpierce/rust-base64/issues/289 .

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,903 @@
...
+};
+use percent_encoding::percent_decode;
+use snafu::prelude::*;
+
+/// Parsing errors.
+#[derive(Debug, Clone, PartialEq, Eq, Snafu)]
+pub enum Error {
+    /// The ZIP 321 request included more payments than can be created within a single Zcash
+    /// transaction. The wrapped value is the number of payments in the request.
+    #[snafu(display("Cannot create a Zcash transation container {count} payments"))]
```
daira, line 21:
```suggestion
    #[snafu(display("Cannot create a Zcash transaction making {count} payments"))]
```

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,903 @@
...
+    TooManyPayments { count: usize },
+
+    /// A memo field in the ZIP 321 URI was not properly base-64 encoded
```
daira, line 24:
```suggestion
    /// A memo field in the ZIP 321 URI was not properly base-64 encoded.
```

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,903 @@
...
+    #[snafu(display("Memo value was not correctly base64-encoded: {source:?}"))]
+    InvalidBase64 { source: base64::DecodeError },
+
+    /// A memo byte array was too long.
+    #[snafu(display("Memo length {count} is larger than maximum of 512"))]
```
daira, line 29:
```suggestion
    #[snafu(display("Memo length {count} bytes is larger than maximum of 512 bytes"))]
```

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,903 @@
...
+    /// The payment at the wrapped index did not include a recipient address.
+    #[snafu(display("Payment {idx} is missing its recipient address"))]
+    RecipientMissing { idx: usize },
+}
+
+type Result<T, E = Error> = core::result::Result<T, E>;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct Memo([u8; 512]);
```
daira, line 51: Can we not use `zcash_protocol::memo::MemoBytes` directly instead? I know it isn't `repr(transparent)`, but it has `as_array` which should be enough, shouldn't it?

components/zip321-parse/src/lib.rs
```diff
@@ -0,0 +1,903 @@
...
+
+impl Memo {
+    /// Create a new `Memo`.
+    pub fn new(bytes: [u8; 512]) -> Self {
+        Self(bytes)
+    }
+
+    /// Converts a [`MemoBytes`] value to a ZIP 321 compatible base64-encoded string.
```
daira, line 59: This appears not to involve `MemoBytes` (as it currently stands), was this left over from a previous state or copied from somewhere else?

In any case you should be able to do something like:
```rust
pub fn memo_byte_to_base64(&mb) -> String {
    BASE64_URL_SAFE_NO_PAD.encode(mb.as_bytes().as_slice())
}
```
(not to be confused with `mb.as_slice()` which is a slice over only the bytes before the first null byte).

In general there's no need to create a new wrapper type to define functions that only depend on the public API of an existing type.

components/zip321-parse/src/lib.rs
nuttycom: `zip321-parse` is being created explicitly to drop the `zcash_protocol` dependency.

nuttycom: That being said, this should be wrapping `Vec<u8>` instead of `[u8; 512]`, because with ZIP 231 coming in NU7 we don't want fixed memo lengths (though construction of `Memo` should be bounded on the ZIP 231maximum length, or maybe this parsing needs to be dependent on the target version of the transaction?)

PR #1976, opened 2025-09-30
zcash_client_sqlite: Ensure that we find out whether purely-transparent mempool transactions received by the wallet are ever mined.

nuttycom:


schell (approved):
Seems good to me 🧂

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3054,6 +3063,66 @@ pub(crate) fn set_transaction_status(
...
+                    named_params! {
+                        ":new_expiry": tx_retrieval_expiry(chain_tip, tx_expiry),
+                        ":txid": &txid[..],
+                        ":status_type": TxQueryType::Status.code(),
+                    },
+                )?;
+            } else {
+                conn.execute(
+                    "INSERT INTO tx_retrieval_queue (txid, query_type, request_expiry)
+                     VALUES (:txid, :status_type, :new_expiry)",
```
nuttycom: Bug: there might already be an enhancement request in the queue, in which case this will cause a conflict.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -10,6 +10,16 @@ workspace.
 
 ## [Unreleased]
 
+## [0.18.3] - 2025-09-30
+
+### Changed
+- The `zcash_client_sqlite` implementation of `WalletWrite::update_chain_tip`
+  now ensures that a transaction status request is queued for any transactions
+  for which we do not have mined-height information and which are known to be
+  unexpired.
```
daira, line 19: This is confusing to me. How do we know they are unexpired if we don't have mined-height information? Do you mean "and which are not known to be expired"?

nuttycom: We can judge whether a transaction is expired purely based on the expiry height and the current chain tip height.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3010,12 +3010,21 @@ pub(crate) fn set_transaction_status(
     match status {
         TransactionStatus::TxidNotRecognized | TransactionStatus::NotInMainChain => {
             // Remove the txid from the retrieval queue unless an unexpired transaction having that
-            // txid exists in the transactions table.
+            // txid exists in the transactions table, with a `VERIFY_LOOKAHEAD` buffer.
```
daira, line 3013: It's not clear what this means.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/95d556dadeb595b25ca03675008ca8fd4bf3f0bc..fdf65ee6620ae5bb014286da9e89ad6578738d7) to fix Clippy complaints.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/95d556dadeb595b25ca03675008ca8fd4bf3f0bc..1aa4e6cd6e2443ab8176e67a5d4d5016317926e1) to fix Clippy and rustfmt complaints.

zcash_client_sqlite/CHANGELOG.md
daira: That tells us whether it would be able to be mined now. It doesn't tell us whether it is expired; if it was mined prior to its expiry height but the current chain tip is past the expiry height, then it's not expired, it's mined.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3010,12 +3010,21 @@ pub(crate) fn set_transaction_status(
...
                      WHERE txid = :txid
-                     AND request_expiry <= :chain_tip",
+                     AND request_expiry <= :chain_tip
+                     AND txid NOT IN (
+                        SELECT tx.txid
+                        FROM transactions tx
+                        WHERE tx.mined_height IS NULL
+                        AND (
+                            tx.expiry_height == 0 -- tx will never expire
+                            OR tx.expiry_height > :chain_tip -- tx is unexpired
```
daira, line 3025:
```suggestion
                            OR tx.expiry_height > :chain_tip -- tx could still be mined
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3054,6 +3063,67 @@ pub(crate) fn set_transaction_status(
...
+            "SELECT
+                t.txid,
+                t.expiry_height,
+                r.query_type IS NOT NULL AS status_request_exists
+             FROM transactions t
+             LEFT OUTER JOIN tx_retrieval_queue r ON r.txid = t.txid
+             WHERE t.mined_height IS NULL
+             AND (
+                t.expiry_height IS NULL -- expiry is unknown
+                OR t.expiry_height = 0 -- tx will note expire
```
daira, line 3080:
```suggestion
                OR t.expiry_height = 0 -- tx will never expire
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3054,6 +3063,67 @@ pub(crate) fn set_transaction_status(
...
+                OR t.expiry_height > :chain_tip -- tx is unexpired
```
daira, line 3081:
```suggestion
                OR t.expiry_height > :chain_tip -- tx could still be mined
```

PR #1978, opened 2025-10-02
Post-release merge of `zcash_client_sqlite-0.18.3`

nuttycom:


str4d (approved):
utACK ddf80d0e9d7bf6dc3d1d777c41d7c06d8d3bf348

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3054,6 +3063,67 @@ pub(crate) fn set_transaction_status(
...
+        while let Some(row) = rows.next()? {
+            let txid = row.get::<_, Vec<u8>>("txid")?;
+            let tx_expiry = row.get::<_, Option<u32>>("expiry_height")?;
+
+            if row.get::<_, bool>("status_request_exists")? {
+                conn.execute(
+                    "UPDATE tx_retrieval_queue
+                     SET request_expiry = :new_expiry
+                     WHERE txid = :txid
+                     AND query_type = :status_type",
```
daira, line 3102: `AND query_type = :status_type` appears to be redundant because we have `r.query_type IS NULL OR r.query_type = :status_type` from the SELECT above, so `query_type = :status_type` whenever `status_request_exists` is true (since the whole function is in a transaction), and `query_type IS NULL` otherwise.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3054,6 +3063,67 @@ pub(crate) fn set_transaction_status(
...
+                        ":new_expiry": tx_retrieval_expiry(chain_tip, tx_expiry),
+                        ":txid": &txid[..],
+                        ":status_type": TxQueryType::Status.code(),
+                    },
+                )?;
+            } else {
+                conn.execute(
+                    "INSERT INTO tx_retrieval_queue (txid, query_type, request_expiry)
+                     VALUES (:txid, :status_type, :new_expiry)
+                     ON CONFLICT (txid) DO NOTHING",
```
daira, line 3113: This could have been `ON CONFLICT (txid) DO UPDATE SET request_expiry = :status_type`. And then `status_request_exists` is not needed and the `INSERT ... ON CONFLICT ...` can be unconditional.

(`ON CONFLICT (txid) DO UPDATE SET request_expiry = IIF(query_type = :status_type, :new_expiry, request_expiry)` would more closely match the existing logic, but as noted above the `AND query_type = :status_type` condition is unnecessary.)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3054,6 +3063,67 @@ pub(crate) fn set_transaction_status(
...
+                    named_params! {
+                        ":new_expiry": tx_retrieval_expiry(chain_tip, tx_expiry),
+                        ":txid": &txid[..],
+                        ":status_type": TxQueryType::Status.code(),
+                    },
+                )?;
+            }
```
daira, line 3120:
```suggestion
            conn.execute(
                "INSERT INTO tx_retrieval_queue (txid, query_type, request_expiry)
                 VALUES (:txid, :status_type, :new_expiry)
                 ON CONFLICT (txid) DO UPDATE SET request_expiry = :status_type",
                named_params! {
                    ":new_expiry": tx_retrieval_expiry(chain_tip, tx_expiry),
                    ":txid": &txid[..],
                    ":status_type": TxQueryType::Status.code(),
                },
            )?;
```

nuttycom: No, this doesn't work; this would overwrite existing enhancement requests with status requests.

daira: It wouldn't because of the `r.query_type IS NULL OR r.query_type = :status_type` in the SELECT, unless I'm missing something. Please check the reasoning in my other comments.

daira (commented):
Post-hoc ACK

PR #1982, opened 2025-10-08
Remove deprecations from zcash_primitives

schell:
This removes deprecated modules from `zcash_primitives` and fixes all callsites.

### DIFF
--- a/zcash_client_memory/src/types/account.rs
+++ b/zcash_client_memory/src/types/account.rs
@@ -4,6 +4,11 @@ use std::{
 };
 
 use subtle::ConditionallySelectable;
+use transparent::address::TransparentAddress;
+#[cfg(feature = "transparent-inputs")]
+use transparent::keys::{
+    AccountPubKey, EphemeralIvk, IncomingViewingKey, NonHardenedChildIndex, TransparentKeyScope,
+};
 use zcash_address::ZcashAddress;
 #[cfg(feature = "transparent-inputs")]
 use zcash_client_backend::wallet::TransparentAddressMetadata;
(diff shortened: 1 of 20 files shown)


zcash_client_memory/src/types/account.rs
```diff
@@ -4,6 +4,11 @@ use std::{
 };
 
 use subtle::ConditionallySelectable;
+use transparent::address::TransparentAddress;
```
daira, line 7: The only use of `TransparentAddress` in this file that is not guarded by "transparent-inputs" is the `EphemeralAddress` structure, which is `pub(crate)`. Consider guarding that and its impl by "transparent-inputs" as well.

zcash_client_memory/src/wallet_write.rs
```diff
@@ -1196,7 +1195,7 @@ Instead derive the ufvk in the calling code and import it using `import_account_
         use zcash_keys::keys::AddressGenerationError;
         // TODO: We need to implement first_unsafe_index to make sure we dont violate gap invarient
```
daira, line 1196:
```suggestion
        // TODO: We need to implement `first_unsafe_index` to make sure we don't violate the gap invariant.
```
Please file an issue for this if there isn't one, and reference it here.

zcash_client_memory/src/types/account.rs
nuttycom: I'm surprised that this doesn't cause dead code issues in CI, if that's the case. We need to understand why `EphemeralAddress` isn't already flagged.

zcash_client_memory/src/wallet_write.rs
schell: As another step - I could log every TODO in this repo as an issue, with a `todo` label. It would only take a moment as I've written a tool to do it. We can keep it in sync automatically if we care to go that far.

zcash_client_memory/src/types/account.rs
schell: I think it's because it's used in the `Account` struct, which is `pub`.

zcash_client_memory/src/wallet_write.rs
schell: Look like there's 53 distinct todos in 65 locations.

schell:
@nuttycom @daira I'm not sure we can hide that `TransparentAddress` import, as it's used in `EphemeralAddress`, which is used in `Account`, which is `pub`. The entire module is `pub(crate)` but then `Account` is used inside `pub` functions, like `WalletWrite::import_account_ufvk` for `MemoryWalletDb`.

zcash_client_memory/src/types/account.rs
nuttycom: In any case, changing anything associated with this is not appropriate for this PR. Also, this is the `zcash_client_memory` crate, which we're not currently putting any energy into.

PR #1985, opened 2025-10-08
zcash_client_sqlite: Exclude funds received on ephemeral addresses from transparent balance.

nuttycom:


zcash_client_sqlite/CHANGELOG.md
```diff
@@ -10,13 +10,24 @@ workspace.
 
 ## [Unreleased]
 
+## [0.18.4] - 2025-10-08
+
+### Fixed
+- This modifies balance calculation to explicitly ignore balance held in
+  ephemeral addresses. This will be altered in a future PR; at present,
```
str4d: Nit:
```suggestion
  ephemeral addresses. This will be altered in a future release; at present,
```

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -10,13 +10,24 @@ workspace.
...
+  the only use of ephemeral addresses is as interstitial addresses in TEX
+  address transfers, and so it is safe to ignore these funds (which would only
+  appear in the case of a partial TEX transfer failure or funds being returned
+  to a TEX address, which would not be detected by normal scanning but which
+  could be detected by the new mempool detection logic implemented by Zashi.
```
str4d, line 22: Wording here feels odd but I can't think of better wording, so it's fine for now.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -10,13 +10,24 @@ workspace.
...
 ## [Unreleased]
 
+## [0.18.4] - 2025-10-08
+
+### Fixed
+- This modifies balance calculation to explicitly ignore balance held in
+  ephemeral addresses. This will be altered in a future PR; at present,
+  the only use of ephemeral addresses is as interstitial addresses in TEX
+  address transfers, and so it is safe to ignore these funds (which would only
+  appear in the case of a partial TEX transfer failure or funds being returned
```
daira:
```suggestion
  address transfers, and so it is safe to ignore these funds. Funds would only
  be missed in the case of a partial TEX transfer failure or funds being returned
```

zcash_client_sqlite/CHANGELOG.md
daira: It probably feels odd partly because of the unbalanced `(`.

PR #1987, opened 2025-10-10
 zcash_client_sqlite: Correctly handle transaction expiry heights in balance & unspent note queries.

nuttycom:
Prior to this change, transactions with unknown expiry heights were always treated as unexpired. This could cause funds to be indefinitely "locked" if a spending transaction was mined, detected via chain scanning, then reorg'ed away and never mined; in this circumstance, the wallet would see the notes and/or UTXOs spent by that transaction as being permanently unavailable to the wallet.

This change adds a non-nullable `min_observed_height` column to the `transactions` table that tracks the block height at which a transaction is first observed. This change needs to be coupled to a wallet feature that continues to attempt to retrieve such orphaned transactions for the period that the transaction would cause a note to be locked by the orphan that spends it.


zcash_client_memory/src/input_source.rs
```diff
@@ -48,10 +49,6 @@ impl<P: consensus::Parameters> InputSource for MemoryWalletDb<P> {
         >,
         Self::Error,
     > {
-        let target_height = TargetHeight::from(
-            self.chain_height()?
-                .ok_or(crate::error::Error::ChainHeightUnknown)?,
-        );
```
str4d, line 54: There was an off-by-one bug here: this was using the chain height as the mempool height, not `chain_height + 1`.

zcash_client_backend/CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ workspace.
 - Migrated to `zcash_protocol 0.7`, `zcash_address 0.10`, `zip321 0.6`,
   `zcash_transparent 0.6`, `zcash_primitives 0.26`, `zcash_proofs 0.26`
 - `zcash_client_backend::data_api`:
+  - `InputSource::get_spendable_note` now takes an additional `target_height`
+    argument; spendability isn't a well-defined property in absence of target
+    height information.
```
str4d: Technically it is well-defined: it's relative to the input source's view of the chain tip. But given that all of the other `InputSource` methods take a target height, I agree it makes more sense for this method to also take an external target height.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1168,6 +1176,10 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> WalletTes
         &self,
         protocol: ShieldedProtocol,
     ) -> Result<Vec<ReceivedNote<Self::NoteRef, Note>>, <Self as InputSource>::Error> {
+        let (target_height, _) = self
+            .get_target_and_anchor_heights(NonZeroU32::MIN)?
+            .expect("chain height is known");
```
str4d: Are we guaranteed to know the chain height here? If not, this should be returning `SqliteClientError::ChainHeightUnknown`.

Also, we should probably be consistent in using either `self.get_target_and_anchor_heights(_)?.0` or `wallet::chain_tip_height(_)? + 1`.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1088,11 +1088,12 @@ pub(crate) fn put_shard_roots<
 pub(crate) fn check_witnesses(
     conn: &rusqlite::Transaction<'_>,
 ) -> Result<Vec<Range<BlockHeight>>, SqliteClientError> {
-    let chain_tip_height =
-        super::chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)?;
+    let target_height = TargetHeight::from(
+        super::chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)? + 1,
```
str4d: I've now seen at least two `+ 1`s, that really should motivate a new `target_height(conn)` method, or using `self.get_target_and_anchor_heights(_)` as I noted above.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -406,7 +412,7 @@ pub(crate) fn get_orchard_nullifiers(
                FROM orchard_received_note_spends spends
                JOIN transactions stx ON stx.id_tx = spends.transaction_id
                WHERE stx.block IS NOT NULL  -- the spending tx is mined
```
str4d: Doc bug: the last commit claims to update all the places where this could change, but here is one that didn't (while the equivalent spot in `wallet/sapling.rs` was updated. If we're going to change this, I agree it should be done consistently across the codebase.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -278,6 +278,7 @@ CREATE TABLE "transactions" (
     raw BLOB,
     fee INTEGER,
     target_height INTEGER,
+    min_observed_height INTEGER NOT NULL,
```
str4d, line 286: New column needs to be documented.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -805,27 +825,25 @@ pub(crate) fn get_wallet_transparent_output(
...
+                        AND t.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
+                    )
                  )
                  -- and the output is unspent
-                 AND u.id NOT IN (
-                    SELECT txo_spends.transparent_received_output_id
-                    FROM transparent_received_output_spends txo_spends
-                    JOIN transactions tx ON tx.id_tx = txo_spends.transaction_id
-                    WHERE tx.mined_height IS NOT NULL  -- the spending tx is mined
-                    OR tx.expiry_height = 0 -- the spending tx will not expire
```
str4d, line 816: Mmk, so some of this logic _was_ correct, and yet almost identical to the incorrect clauses. Was the incorrect clause copied to a new place in a later PR, and then that PR got review comments identifying the bug that were fixed in the PR's changes but not where the clause was copied from (presumably because the reviewer didn't know it had been copied, and thus didn't ask for it to be fixed elsewhere)?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -778,6 +778,26 @@ fn to_unspent_transparent_output(row: &Row) -> Result<WalletTransparentOutput, S
...
+        r#"
+        SELECT txo_spends.transparent_received_output_id
+        FROM transparent_received_output_spends txo_spends
+        JOIN transactions stx ON stx.id_tx = txo_spends.transaction_id
+        WHERE stx.mined_height IS NOT NULL -- the spending transaction is mined
+        OR stx.expiry_height = 0 -- the spending tx will not expire
+        OR stx.expiry_height >= :target_height -- the spending tx is unexpired
+        -- if we don't know an actual mined height or expiry height, we will treat the
+        -- spending transaction as unexpired _only_ if we just observed it in the last
+        -- <expiry_delta> blocks
```
```suggestion
        -- <expiry_delta> blocks, guessing that the wallet creating the transaction
        -- used the same default expiry delta as us. If our guess is wrong (and the
        -- wallet used a larger expiry delta or disabled expiry), then the UTXO will
        -- be treated as spendable when it shouldn't be for as long as it takes this
        -- wallet to either observe the spending transaction being mined, or enhance
        -- the spending transaction to learn its expiry height.
```
or if my extraction suggestion is accepted:
```suggestion
/// `expiry_delta` blocks, guessing that the wallet creating the transaction
/// used the same default expiry delta as us. If our guess is wrong (and the
/// wallet used a larger expiry delta or disabled expiry), then the transaction
/// will be treated as unexpired when it shouldn't be for as long as it takes
/// this wallet to either observe the transaction being mined, or enhance it
/// to learn its expiry height.
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -805,27 +825,25 @@ pub(crate) fn get_wallet_transparent_output(
                  (
                     t.mined_height IS NOT NULL -- tx is mined
                     OR t.expiry_height = 0 -- tx will not expire
-                    OR t.expiry_height >= :mempool_height -- tx has not yet expired
+                    OR t.expiry_height >= :target_height -- tx has not yet expired
+                    OR (
+                        t.expiry_height IS NULL -- the expiry height is unknown
+                        AND t.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
+                    )
```
str4d: This is identical to the condition within `spent_utxos_clause`. Maybe we should consider extracting that as well? With something like:
```rust
/// Generates an SQL condition that a transaction is unexpired.
///
/// TODO: Insert comment from `spent_utxos_clause` here, adapted as appropriate.
///
/// # Usage requirements
/// - `tx` must be set to the SQL variable name for the transaction in the parent.
/// - The parent must provide `:target_height` as a named argument.
/// - The parent is responsible for enclosing this condition in parentheses as appropriate.
fn tx_unexpired_condition(tx: &str) -> String {
    format!(
        r#"
        {tx}.mined_height IS NOT NULL -- the transaction is mined
        OR {tx}.expiry_height = 0 -- the tx will not expire
        OR {tx}.expiry_height >= :target_height -- the tx is unexpired
        OR (
            {tx}.expiry_height IS NULL -- the expiry height is unknown
            AND {tx}.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
        )
        "#
    )
}
```

pczt/tests/end_to_end.rs
```diff
@@ -228,7 +228,7 @@ fn transparent_p2sh_multisig_to_orchard() {
     ];
     let redeem_script = script::Component(check_multisig(2, &pks, false).unwrap());
     let p2sh_addr = TransparentAddress::ScriptHash(
-        *ripemd::Ripemd160::digest(sha2::Sha256::digest(&redeem_script.to_bytes())).as_ref(),
+        *ripemd::Ripemd160::digest(sha2::Sha256::digest(redeem_script.to_bytes())).as_ref(),
```
nuttycom: This was an unrelated beta Clippy lint that was automatically fixed.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -875,15 +893,9 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
             )
          )
          -- and the output is unspent
-         AND u.id NOT IN (
-            SELECT txo_spends.transparent_received_output_id
-            FROM transparent_received_output_spends txo_spends
-            JOIN transactions tx ON tx.id_tx = txo_spends.transaction_id
-            WHERE tx.mined_height < :target_height -- the spending transaction is mined
```
str4d, line 882: To confirm, this PR correctly indicates your intent to standardise on `tx.mined_height IS NOT NULL`, not `tx.mined_height < :target_height` (the latter of which I thought you started using more recently IIRC)?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1266,7 +1254,9 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
     params: &P,
     output: &WalletTransparentOutput,
 ) -> Result<(AccountRef, KeyScope, UtxoId), SqliteClientError> {
-    put_transparent_output(conn, params, output, true)
+    let observation_height =
+        chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)?;
+    put_transparent_output(conn, params, output, observation_height, true)
```
str4d: Should the observation height be the chain tip, or the target height for this chain? I suspect the latter:
- If `output.mined_height().is_none()` then the correct `min_observed_height` (based on my understanding of that column from usage, given that it is undocumented) is the mempool height (i.e. the target height), as that's the first block it can be mined in.
- If `output.mined_height().is_some()` then the only way that the chain tip (or lower) is the correct `min_observed_height` is if we haven't yet observed the transparent transactions from our chain tip block. I guess this could happen in a race condition (UTXOs fetched, reorg occurs that includes the UTXO's tx mined, wallet is informed of the reorg before it is informed of the UTXOs). But given that the purpose of `min_observed_height` (again inferred due to no documentation) is to enable us to guess when a tx is expired (so we don't select unspendable TXOs and brick transaction creation), it seems better for reliability to use the target height here until we actually observe evidence of the UTXO's tx being mined in the chain tip block.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2079,20 +2079,18 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
              WHERE (
-                t.block IS NOT NULL -- the receiving tx is mined
-                OR t.expiry_height IS NULL -- the receiving tx will not expire
+                t.mined_height IS NOT NULL -- the receiving tx is mined
+                OR t.expiry_height = 0 -- the receiving tx will not expire
                 OR t.expiry_height >= :target_height -- the receiving tx is unexpired
+                OR (
+                    t.expiry_height IS NULL -- the expiry height is unknown
+                    AND t.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
+                )
```
str4d: Ditto, this could also use my proposed `tx_unexpired_condition` helper.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3644,7 +3644,12 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
     // fee value.
     let fee = determine_fee(conn, d_tx.tx())?;
 
-    let tx_ref = put_tx_data(conn, d_tx.tx(), fee, None, None)?;
+    let observation_height = d_tx.mined_height().map_or_else(
+        || chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown),
```
str4d: Same question here: should this be the target height? (I think same analysis applies.)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4017,12 +4027,13 @@ pub(crate) fn put_tx_meta(
...
-        "INSERT INTO transactions (txid, block, mined_height, tx_index)
-        VALUES (:txid, :block, :block, :tx_index)
+        "INSERT INTO transactions (txid, block, mined_height, tx_index, min_observed_height)
+        VALUES (:txid, :block, :block, :tx_index, :block)
         ON CONFLICT (txid) DO UPDATE
         SET block = :block,
+            tx_index = :tx_index,
             mined_height = :block,
-            tx_index = :tx_index
+            min_observed_height = MIN(min_observed_height, :block)
```
str4d: Here by comparison, we have observed the transaction mined in a block, so this is a correct update.

I note that if we were to actually rely on `min_observed_height` later for expiry calculations (i.e. somewhere we forget to check `mined_height` first), it could produce misleading results if `:block` were larger here. (I'm specifically thinking of a transaction that is created at height X with expiry height X + E, then mined at height X - D which would give it a "larger" expiry delta than what we use). I think all I want is that the documentation for `min_observed_height` should note that it is a rough guess for the target height that the transaction was created for, and should only be used in queries when `mined_height` and `expiry_height` are both `NULL` (either of which is a better basis for a guess).

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -74,6 +74,27 @@ pub(crate) fn table_constants<E: ErrUnsupportedPool>(
     }
 }
 
+// A helper funtion that generates a SQL expression
+pub(crate) fn spent_notes_clause(table_prefix: &str) -> String {
```
str4d: I see why this is separate from `spent_utxos_clause`, and I'm not going to suggest you try and merge them, because my suggestion of a `tx_unexpired_condition` helper would remove most of the duplication.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -307,19 +325,17 @@ where
          AND nf IS NOT NULL
          AND (
             t.block IS NOT NULL -- the receiving tx is mined
-            OR t.expiry_height IS NULL -- the receiving tx will not expire
+            OR t.expiry_height = 0 -- the receiving tx will not expire
             OR t.expiry_height >= :target_height -- the receiving tx is unexpired
+            OR (
+                t.expiry_height IS NULL -- the expiry height is unknown
+                AND t.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
+            )
```
str4d: Ditto: this could use `tx_unexpired_condition`.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -787,19 +786,24 @@ pub(crate) fn spendable_notes_meta(
...
+                        WHERE stx.mined_height IS NOT NULL -- the spending transaction is mined
+                        OR stx.expiry_height = 0 -- the spending tx will not expire
+                        OR stx.expiry_height >= :target_height -- the spending tx is unexpired
+                        -- if we don't know an actual mined height or expiry height, we will treat the
+                        -- spending transaction as unexpired _only_ if we just observed it in the last
+                        -- <expiry_delta> blocks
+                        OR (
+                            stx.expiry_height IS NULL -- the expiry height is unknown
+                            AND stx.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
+                        )
```
str4d: Ditto: this could use `tx_unexpired_condition`.

zcash_client_backend/CHANGELOG.md
nuttycom: The input source trait doesn't, in the abstract, have a view of the chain state, so it can't know what the "chain tip" is. It's essentially treated just as a bag of inputs with selection criteria.

zcash_client_sqlite/src/wallet/commitment_tree.rs
nuttycom: Given that `self.get_target_and_anchor_height` relies on a number of confirmations, I think that what I would name this is `mempool_height(conn)`.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: I don't think this was a consequence of review; I think that this was written from scratch at a different time than the erroneous `tx.expiry_height IS NULL` clause. Note the use of `:mempool_height` instead of `:target_height` - all the `:target_height` changes were made at the same time.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Hm, no, good catch; this is a bug. We should standardize on `tx.mined_height < :target_height`.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: The intent of `min_observed_height` is that it be used more like a clock than a chain reference. It is the chain tip height at whatever time the wallet first observed the transaction; if the mined height is later discovered to be earlier, we set the observation height to that block height because that is the earliest height at which the transaction *could have been* observed.

zcash_client_sqlite/src/lib.rs
nuttycom: This is a bug. This should use `SqliteClientError::ChainHeightUnknown` - at the time I wrote this, I was making changes both to the `zcash_client_backend` code and to `zcash_client_sqlite` code and I think I thought that I didn't have access to that error type here (due to the abstract type that is used in the function signature.)

zcash_client_sqlite/src/wallet/transparent.rs
str4d: > It is the chain tip height at whatever time the wallet first observed the transaction

That doesn't match how you are initializing the column in the migration; there you use the mempool height as the fallback.

I'll re-consider the analysis once you have documented `min_observed_height` so I know what you were actually intending.

nuttycom: Now I'm considering whether it _should_ be the minimum observed height.

The important case is:
* The transaction discovered by mempool scanning - only the min_observed_height is set.
* The transaction is reorged away. It's now effectively observed at mempool height.

So I think you're right, it should just consistently be set to mempool height when there's no better information.

zcash_client_memory/src/input_source.rs
daira:
```suggestion
    /// that belongs to the given transaction, provided that it is spendable at the
    /// given target height.
```

zcash_client_sqlite/src/wallet/init/migrations/tx_observation_height.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, conn: &Transaction) -> Result<(), WalletMigrationError> {
+        // We use the target height as the minimum observed height for transactions
+        // for which we don't already have any other height information.
+        //
+        // chain_tip_height should only be `None` for a brand new wallet, in which case there
+        // should be no entries in the `transactions` table and the Sapling activation height end
```
daira:
```suggestion
        // should be no entries in the `transactions` table and the Sapling activation height will end
```

zcash_client_sqlite/src/wallet/init/migrations/tx_observation_height.rs
```diff
@@ -0,0 +1,116 @@
...
+                tx_index,
+                expiry_height,
+                raw,
+                fee,
+                target_height,
+                MIN(
+                    IFNULL(target_height, {fallback_height}),
+                    IFNULL(mined_height, {fallback_height}),
+                    {fallback_height}
+                )
```
daira: This seems wrong. If both `target_height` and `mined_height` are null, we want to use `{fallback_height}`. If exactly one of them is non-null, we want to use that one (I think), and if both are non-null we want to use the minimum of them. That can be expressed as `IFNULL(MIN(IFNULL(target_height, mined_height), IFNULL(mined_height, target_height)), {fallback_height})`.

Explanation: `MIN` as a scalar function (as opposed to an aggregate function) will return `NULL` if any of its arguments are null. Therefore we need to make sure that both arguments are non-null when either is non-null in order to get the right semantics. The above expression is ugly but it does that correctly (tested with https://sqlime.org/ ).

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -74,6 +74,27 @@ pub(crate) fn table_constants<E: ErrUnsupportedPool>(
     }
 }
 
+// A helper funtion that generates a SQL expression
```
daira:
```suggestion
// A helper funtion that generates a SQL expression.
// Maintenance hazard: this is similar to `spent_utxos_clause` in `wallet\transparent.rs`.
// Consider changing both when changing either.
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -778,6 +778,26 @@ fn to_unspent_transparent_output(row: &Row) -> Result<WalletTransparentOutput, S
     })
 }
```
daira, line 782:
```suggestion

// A helper funtion that generates a SQL expression.
// Maintenance hazard: this is similar to `spent_notes_clause` in `wallet\common.rs`.
// Consider changing both when changing either.
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -74,6 +74,27 @@ pub(crate) fn table_constants<E: ErrUnsupportedPool>(
...
+    // spending transaction as unexpired _only_ if we just observed it in the last
+    // DEFAULT_TX_EXPIRY_DELTA blocks
+    format!(
+        r#"
+        SELECT {table_prefix}_received_note_id
+        FROM {table_prefix}_received_note_spends rns
+        JOIN transactions stx ON stx.id_tx = rns.transaction_id
+        WHERE stx.mined_height IS NOT NULL -- the spending transaction is mined
+        OR stx.expiry_height = 0 -- the spending tx will not expire
+        OR stx.expiry_height >= :target_height -- the spending tx is unexpired
```
daira: `NULL` will never be `>= :target_height` so this correctly uses only the condition on line 92 when `stx.expiry_height IS NULL`.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -613,40 +624,34 @@ pub(crate) fn select_unspent_note_meta(
...
+         AND recipient_key_scope IS NOT NULL
+         AND nf IS NOT NULL
+         AND commitment_tree_position IS NOT NULL
+         AND rn.id NOT IN ({})
+         AND NOT EXISTS (
+            SELECT 1 FROM v_{table_prefix}_shard_unscanned_ranges unscanned
+            -- select all the unscanned ranges involving the shard containing this note
+            WHERE rn.commitment_tree_position >= unscanned.start_position
+            AND rn.commitment_tree_position < unscanned.end_position_exclusive
+            -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
```
daira, line 715:
```suggestion
            -- exclude unscanned ranges that start above the target height (they don't affect spendability)
```

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -395,6 +397,10 @@ pub(crate) fn get_orchard_nullifiers(
     // Get the nullifiers for the notes we are tracking
     let mut stmt_fetch_nullifiers = match query {
         NullifierQuery::Unspent => conn.prepare(
+            // This may over-select nullifiers and return those that have been spent in un-mined
+            // transactions that have not yet expired, or for which the expiry height is unknown.
+            // This is fine because these nullifiers are primarily used to detect the spends of our
+            // own notes in scanning; if we select a few too many nullifiers, it's not a big deal.
```
daira: This belongs in the function's doc comment I think.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -278,6 +280,10 @@ pub(crate) fn get_sapling_nullifiers(
     // Get the nullifiers for the notes we are tracking
     let mut stmt_fetch_nullifiers = match query {
         NullifierQuery::Unspent => conn.prepare(
+            // This may over-select nullifiers and return those that have been spent in un-mined
+            // transactions that have not yet expired, or for which the expiry height is unknown.
+            // This is fine because these nullifiers are primarily used to detect the spends of our
+            // own notes in scanning; if we select a few too many nullifiers, it's not a big deal.
```
daira: Similarly this belongs in the function's doc comment.

zcash_client_sqlite/src/wallet/sapling.rs
```diff
@@ -288,8 +294,8 @@ pub(crate) fn get_sapling_nullifiers(
                SELECT spends.sapling_received_note_id
                FROM sapling_received_note_spends spends
                JOIN transactions stx ON stx.id_tx = spends.transaction_id
-               WHERE stx.block IS NOT NULL  -- the spending tx is mined
-               OR stx.expiry_height IS NULL -- the spending tx will not expire
+               WHERE stx.mined_height IS NOT NULL  -- the spending tx is mined
+               OR stx.expiry_height = 0 -- the spending tx will not expire
```
daira:
```suggestion
               OR stx.expiry_height = 0  -- the spending tx will not expire
```

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -406,7 +412,7 @@ pub(crate) fn get_orchard_nullifiers(
                FROM orchard_received_note_spends spends
                JOIN transactions stx ON stx.id_tx = spends.transaction_id
                WHERE stx.block IS NOT NULL  -- the spending tx is mined
-               OR stx.expiry_height IS NULL -- the spending tx will not expire
+               OR stx.expiry_height = 0 -- the spending tx will not expire
```
daira:
```suggestion
               OR stx.expiry_height = 0  -- the spending tx will not expire
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -778,6 +778,26 @@ fn to_unspent_transparent_output(row: &Row) -> Result<WalletTransparentOutput, S
...
 
+pub(crate) fn spent_utxos_clause() -> String {
+    format!(
+        r#"
+        SELECT txo_spends.transparent_received_output_id
+        FROM transparent_received_output_spends txo_spends
+        JOIN transactions stx ON stx.id_tx = txo_spends.transaction_id
+        WHERE stx.mined_height IS NOT NULL -- the spending transaction is mined
```
daira: Nit: two spaces before an SQL comment.

zcash_client_sqlite/src/wallet/transparent.rs
daira:
```rust
/// `DEFAULT_TX_EXPIRY_DELTA` blocks, guessing that the wallet creating the transaction 
/// used the same expiry delta as our default. [...]
```
(What is relevant is the expiry delta the other wallet used for this transaction as compared to our default; not the other wallet's default.)

zcash_client_sqlite/src/wallet/transparent.rs
daira: > The intent of `min_observed_height` is that it be used more like a clock than a chain reference. It is the chain tip height at whatever time the wallet first observed the transaction; if the mined height is later discovered to be earlier, we set the observation height to that block height because that is the earliest height at which the transaction could have been observed.

If we intend `mined_height IS NULL OR min_observed_height <= mined_height` to always hold, we should add a `CONSTRAINT` for that. Suggested in https://github.com/zcash/librustzcash/pull/1987/files#r2422854005 .

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -278,6 +278,7 @@ CREATE TABLE "transactions" (
...
     FOREIGN KEY (block) REFERENCES blocks(height),
     CONSTRAINT height_consistency CHECK (block IS NULL OR mined_height = block)
```
daira:
```suggestion
    CONSTRAINT height_consistency CHECK (block IS NULL OR mined_height = block),
    CONSTRAINT observed_consistent_with_mined_height CHECK (mined_height IS NULL OR min_observed_height <= mined_height)
```
Please check that all updates to `mined_height` maintain this constraint.

zcash_client_sqlite/src/wallet/orchard.rs
nuttycom: I don't see what you're talking about in wallet/sapling.rs; the code there is the same. But I agree that this is a good candidate for a refactoring into `common.rs` - all we need is the function from `&[u8] -> Result<Nullifier, ...>` to be provided and the rest of the code can just use the technique of table name substituion.

zcash_client_memory/src/input_source.rs
nuttycom: The doc comments in this `impl` should actually be deleted; they're not the source of truth (the `InputSource` trait is.)

zcash_client_sqlite/src/wallet/init/migrations/tx_observation_height.rs
nuttycom: Ehhh... I can't see a circumstance where, in the context of this migration, the mined height or the target height will be greater than the fallback height (the mempool height.)

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: This is true. We have no idea what this is, so we use the default.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: Why? We don't use that anywhere that I know of.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: Hmm, this actually *does* need to be the anchor height.

zcash_client_sqlite/src/lib.rs
```diff
@@ -685,13 +693,16 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> InputSour
         selector: &NoteFilter,
         exclude: &[Self::NoteRef],
     ) -> Result<AccountMeta, Self::Error> {
-        let chain_tip_height = wallet::chain_tip_height(self.conn.borrow())?
-            .ok_or(SqliteClientError::ChainHeightUnknown)?;
+        let target_height = TargetHeight::from(
```
daira: This can use `mempool_height`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3644,7 +3641,16 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
     // fee value.
     let fee = determine_fee(conn, d_tx.tx())?;
 
-    let tx_ref = put_tx_data(conn, d_tx.tx(), fee, None, None)?;
+    let observation_height = d_tx.mined_height().map_or_else(
```
daira: Prefer `observed_height` for consistency.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -74,6 +74,48 @@ pub(crate) fn table_constants<E: ErrUnsupportedPool>(
     }
 }
 
+/// Generates an SQL condition that a transaction is unexpired.
+///
+/// TODO: Insert comment from `spent_utxos_clause` here, adapted as appropriate.
```
daira: Address the TODO.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3077,7 +3074,7 @@ pub(crate) fn refresh_status_requests(
              WHERE t.mined_height IS NULL
              AND (
                 t.expiry_height IS NULL -- expiry is unknown
-                OR t.expiry_height = 0 -- tx will note expire
+                OR t.expiry_height = 0 -- tx will never expire
                 OR t.expiry_height > :chain_tip -- tx is unexpired
```
daira: Should this be using `tx_unexpired_condition`?

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -74,6 +74,48 @@ pub(crate) fn table_constants<E: ErrUnsupportedPool>(
...
+        "#
+    )
+}
+
+// Generates a SQL expression that returns the identifiers of all spent notes in the wallet.
+pub(crate) fn spent_notes_clause(table_prefix: &str) -> String {
+    // If we don't know an actual mined height or expiry height, we will treat the spending
+    // transaction as unexpired _only_ if we just observed it in the last DEFAULT_TX_EXPIRY_DELTA
+    // blocks, guessing that the wallet creating the transaction used the same default expiry delta
+    // as us. If our guess is wrong (and the wallet used a larger expiry delta or disabled expiry),
```
daira: "used the same expiry delta as our default"

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -96,6 +138,58 @@ fn unscanned_tip_exists(
...
+                 JOIN accounts a ON a.id = rn.account_id
+                 JOIN transactions tx ON tx.id_tx = rn.tx
+                 WHERE rn.nf IS NOT NULL
+                 AND tx.block IS NOT NULL
+                 AND rn.id NOT IN (
+                   SELECT rns.{table_prefix}_received_note_id
+                   FROM {table_prefix}_received_note_spends rns
+                   JOIN transactions stx ON stx.id_tx = rns.transaction_id
+                   WHERE stx.mined_height IS NOT NULL  -- the spending tx is mined
+                   OR stx.expiry_height = 0 -- the spending tx will not expire
```
daira, line 180: Should this be using `tx_unexpired_condition`?

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -787,19 +852,24 @@ pub(crate) fn spendable_notes_meta(
...
+                        WHERE stx.mined_height IS NOT NULL -- the spending transaction is mined
+                        OR stx.expiry_height = 0 -- the spending tx will not expire
+                        OR stx.expiry_height >= :target_height -- the spending tx is unexpired
+                        -- if we don't know an actual mined height or expiry height, we will treat the
+                        -- spending transaction as unexpired _only_ if we just observed it in the last
+                        -- <expiry_delta> blocks
+                        OR (
+                            stx.expiry_height IS NULL -- the expiry height is unknown
+                            AND stx.min_observed_height + {DEFAULT_TX_EXPIRY_DELTA} >= :target_height
+                        )
```
daira: Should this be using `tx_unexpired_condition`?

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -613,40 +690,34 @@ pub(crate) fn select_unspent_note_meta(
...
+         AND rn.id NOT IN ({})
+         AND NOT EXISTS (
+            SELECT 1 FROM v_{table_prefix}_shard_unscanned_ranges unscanned
+            -- select all the unscanned ranges involving the shard containing this note
+            WHERE rn.commitment_tree_position >= unscanned.start_position
+            AND rn.commitment_tree_position < unscanned.end_position_exclusive
+            -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
+            AND unscanned.block_range_start < :anchor_height
+            -- exclude unscanned ranges that end below the wallet birthday
+            AND unscanned.block_range_end > :wallet_birthday
```
daira, line 718: It's bothering me that we have so much duplication in our queries, and although having helpers to construct common elements as strings is better than nothing, I'm not sure that it is really the right fix. Should some of these be views?

zcash_client_sqlite/src/wallet/init/migrations/tx_observation_height.rs
daira: `fallback_height` can be the Sapling activation height. But I think you're right.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -802,30 +815,20 @@ pub(crate) fn get_wallet_transparent_output(
...
+        spent_utxos_clause()
+    ))?;
 
     let result: Result<Option<WalletTransparentOutput>, SqliteClientError> = stmt_select_utxo
         .query_and_then(
             named_params![
                 ":txid": outpoint.hash(),
                 ":output_index": outpoint.n(),
-                ":mempool_height": chain_tip_height.map(|h| u32::from(h) + 1),
+                ":target_height": chain_tip_height.map(|h| u32::from(h) + 1),
```
daira: This can use the `mempool_height` function I think.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1266,7 +1239,9 @@ pub(crate) fn put_received_transparent_utxo<P: consensus::Parameters>(
     params: &P,
     output: &WalletTransparentOutput,
 ) -> Result<(AccountRef, KeyScope, UtxoId), SqliteClientError> {
-    put_transparent_output(conn, params, output, true)
+    let observation_height =
```
daira: Use `observed_height` for consistency.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: Unfortunately you can't parameterize views. In another database, I would be using stored procedures for these conditions but SQLite doesn't have them.

zcash_client_sqlite/src/wallet.rs
nuttycom: No, in this specific case we do not want to consider the observation height. If the wallet observes a transaction via scanning (and thus doesn't have an expiry height), then observes a reorg that un-mines the transaction, and then goes offline for more than `DEFAULT_TX_EXPIRY_DELTA`, we definitely want the wallet to check on the status of that transaction. I will add a comment to this effect.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: No, see the method documentation.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: Yes, this should.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -578,9 +593,8 @@ pub(super) const INDEX_SENT_NOTES_TX: &str = r#"CREATE INDEX sent_notes_tx ON "s
 ///   about transparent inputs to a transaction, this is a reference to that transaction record.
 ///   NULL for transactions where the request for enhancement data is based on discovery due
 ///   to blockchain scanning.
-/// - `request_expiry`: The block height at which this transaction data request will be considered
-///   expired. This is used to ensure that permanently-unsatisfiable transaction data requests
-///   do not stay in the queue forever.
+/// - `request_expiry`: UNUSED; this column should be removed the next time that this table
+///   should be rewritten (TODO).
```
daira, line 597: Why not remove it now in the `tx_observation_height` migration?

daira (approved):
utACK

zcash_client_sqlite/src/wallet/db.rs
nuttycom: I think that this table is going to have to be rewritten to support account deletion anyway (it has a foreign key to `transactions` that will have to have `ON DELETE CASCADE` added), so we don't need to rewrite it twice.

zcash_client_sqlite/src/wallet/transparent.rs
str4d: In a prior comment two spaces were used, not as a blanket format, but so the comments on two adjacent lines would be aligned. It's not necessary but just convenient to whoever first wrote it I guess.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -96,6 +143,60 @@ fn unscanned_tip_exists(
...
+/// Retrieves the set of nullifiers for "potentially spendable" notes that the wallet is tracking.
+///
+/// This may over-select nullifiers and return those that have been spent in un-mined transactions
+/// that have not yet expired, or for which the expiry height is unknown. This is fine because
+/// these nullifiers are primarily used to detect the spends of our own notes in scanning; if we
+/// select a few too many nullifiers, it's not a big deal.
+///
+/// "Potentially spendable" means:
+/// - The transaction in which the note was created has been observed as mined.
+/// - No transaction in which the note's nullifier appears has been observed as mined.
```
str4d: Nit: this makes more sense the other way round:
```suggestion
/// "Potentially spendable" means:
/// - The transaction in which the note was created has been observed as mined.
/// - No transaction in which the note's nullifier appears has been observed as mined.
///
/// This may over-select nullifiers and return those that have been spent in un-mined transactions
/// that have not yet expired, or for which the expiry height is unknown. This is fine because
/// these nullifiers are primarily used to detect the spends of our own notes in scanning; if we
/// select a few too many nullifiers, it's not a big deal.
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -96,6 +143,60 @@ fn unscanned_tip_exists(
...
+    let mut stmt_fetch_nullifiers = match query {
+        NullifierQuery::Unspent => conn.prepare(&format!(
+            // See the method documentation for why this does not use `spent_notes_clause`.
+            // We prefer to be more restrictive in determining whether a note is spent here.
+            "SELECT a.uuid, rn.nf
+                 FROM {table_prefix}_received_notes rn
+                 JOIN accounts a ON a.id = rn.account_id
+                 JOIN transactions tx ON tx.id_tx = rn.tx
+                 WHERE rn.nf IS NOT NULL
+                 AND tx.block IS NOT NULL
```
str4d:
```suggestion
                 AND tx.mined_height IS NOT NULL
```
like the second half of the query does, unless there's a reason for waiting on the corresponding block to be scanned? I think there's currently no way for these to be set independently for shielded transactions, but could there be in future (say, for a sent tx that we then determine is mined via some polling mechanism rather than block scanning)? I guess it depends on what we mean by "observed as mined" in the doc comment.

If we do keep this as-is, add an SQL comment noting this is intentional so we don't mass-change it in future.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -613,40 +697,34 @@ pub(crate) fn select_unspent_note_meta(
...
+         AND nf IS NOT NULL
+         AND commitment_tree_position IS NOT NULL
+         AND rn.id NOT IN ({})
+         AND NOT EXISTS (
+            SELECT 1 FROM v_{table_prefix}_shard_unscanned_ranges unscanned
+            -- select all the unscanned ranges involving the shard containing this note
+            WHERE rn.commitment_tree_position >= unscanned.start_position
+            AND rn.commitment_tree_position < unscanned.end_position_exclusive
+            -- exclude unscanned ranges that start above the anchor height (they don't affect spendability)
+            AND unscanned.block_range_start < :anchor_height
```
str4d: Bug: You changed this from anchor to target height, which meant changing this bound (as you can't scan the target height block). Then you changed back to anchor height, but didn't revert the bounds check.
```suggestion
            AND unscanned.block_range_start <= :anchor_height
```

zcash_client_sqlite/src/lib.rs
```diff
@@ -483,7 +486,13 @@ impl<C: BorrowMut<Connection>, P, CL, R> WalletDb<C, P, CL, R> {
     ///
     /// This method is intended for repairing wallets that broke due to bugs in `shardtree`.
     pub fn check_witnesses(&mut self) -> Result<Vec<Range<BlockHeight>>, SqliteClientError> {
-        self.transactionally(|wdb| wallet::commitment_tree::check_witnesses(wdb.conn.0))
+        self.transactionally(|wdb| {
+            if let Some(anchor_height) = chain_tip_height(wdb.conn.0)? {
+                wallet::commitment_tree::check_witnesses(wdb.conn.0, anchor_height)
+            } else {
+                Ok(vec![])
```
str4d, line 493: I agree with replacing the error with an empty set: if there's no blocks scanned, then there's no witnesses to check.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3240,7 +3170,7 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
     // transaction entries may be created as a consequence of receiving transparent TXOs.
     conn.execute(
         "UPDATE transactions
-         SET block = NULL, mined_height = NULL, tx_index = NULL
+         SET block = NULL, mined_height = NULL, tx_index = NULL, confirmed_unmined_at_height = NULL
```
str4d, line 3173: I think this is a no-op, as `WHERE mined_height > :height` can only be true if `mined_height IS NOT NULL` (and thus by constraint `confirmed_unmined_at_height IS NULL` already). But it doesn't hurt to be present.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4188,120 +4140,106 @@ pub(crate) fn queue_tx_retrieval(
...
     conn: &rusqlite::Connection,
 ) -> Result<Vec<TransactionDataRequest>, SqliteClientError> {
-    let mut tx_retrieval_stmt =
-        conn.prepare_cached("SELECT txid, query_type FROM tx_retrieval_queue")?;
+    // We will return both explicitly constructed status requests, and a status request for each
+    // transaction that is known to the wallet for which we don't have the mined height,
+    // and for which we have no positive confirmation that the transaction expired unmined.
+    //
+    // For transactions with a known expiry height of 0, we will continue to query indefinitely.
+    // Such transactions should be rebroadcast by the wallet until they are mined.
```
```suggestion
    // Such transactions should be rebroadcast by the wallet until they are either mined or
    // conflict with another mined transaction.
```
which implies we also need to ensure that we stop querying once we observe the transaction is conflicted (i.e. the junction tables tell us that it spends notes or coins that are also spent in some other tx that is mined), and that we _start_ querying again if that conflicting tx itself gets unmined in a reorg.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4188,120 +4140,106 @@ pub(crate) fn queue_tx_retrieval(
...
+         WHERE mined_height IS NULL
+         AND (
+            -- we have no confirmation of expiry
+            confirmed_unmined_at_height IS NULL
+            -- a nonzero expiry height is known, and we have confirmation that the transaction was
+            -- not mined as of a height greater than that expiry height
+            OR (
+                expiry_height > 0
+                AND confirmed_unmined_at_height > expiry_height
+            )
```
str4d, line 4202: Bug: the conditional here is backwards, and is selecting only transactions that are guaranteed to be expired unmined.

`confirmed_unmined_at_height` is documented as:
> the maximum block height at which the wallet has observed positive proof that the transaction has not been mined in a block.

which I interpret to mean "we've confirmed this tx is not present in any block up to and including this height".

That means the earliest the tx could be mined is `confirmed_unmined_at_height + 1`, and so to be unexpired it must hold that `confirmed_unmined_at_height + 1 <= expiry_height`, or:
```suggestion
            -- a nonzero expiry height is known, and we have confirmation that the transaction was
            -- not unmined as of a height greater than or equal to that expiry height
            OR (
                expiry_height > 0
                AND confirmed_unmined_at_height < expiry_height
            )
```

It would be useful to have a test that catches this bug.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4188,120 +4140,106 @@ pub(crate) fn queue_tx_retrieval(
...
+            -- the expiry height is unknown and the default expiry height for it is in the stable
+            -- block range according to the PRUNING_DEPTH
+            OR (
+                expiry_height IS NULL
+                AND confirmed_unmined_at_height > min_observed_height + :certainty_depth
+            )
```
str4d: Ditto bug: this selects transactions that are likely to be expired unmined based on the heuristic, instead of the opposite of that (transactions that are not likely to be expired unmined).
```suggestion
            -- the expiry height is unknown and the default expiry height for it is not yet in the
            -- stable block range according to the PRUNING_DEPTH
            OR (
                expiry_height IS NULL
                AND confirmed_unmined_at_height < min_observed_height + :certainty_depth
            )
```

(I just flipped the condition instead of using `<=`, for symmetry with the one aboce, and because off-by-nes don't really matter for this particular condition.)

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -330,6 +334,7 @@ pub const CURRENT_LEAF_MIGRATIONS: &[Uuid] = &[
     tx_retrieval_queue_expiry::MIGRATION_ID,
     support_zcashd_wallet_import::MIGRATION_ID,
     v_tx_outputs_return_addrs::MIGRATION_ID,
+    tx_observation_height::MIGRATION_ID,
```
str4d, line 346: Duplicate `CURRENT_LEAF_MIGRATIONS` as the frozen `V_0_18_5` state in the 0.18.5 release commit (and add it to `PUBLIC_MIGRATION_STATES` along with any others that were previously missed like `V_0_18_0`).

zcash_client_sqlite/CHANGELOG.md
str4d, line 6: It occurs to me that we should document here explicitly that the database format is not covered by SemVer (we already disclaim this in the code documentation IIRC).

str4d (changes requested):
Reviewed 0bccc10ecf37e44aeca8eff17ecfa01331dde4e6.

zcash_client_sqlite/src/wallet.rs
nuttycom: No, this is incorrect. We need to to have gotten a `TxidNotRecognized` or `NotInMainChain` result at a height greater than the expiry height to be certain that the transaction has expired without having been mined.

nuttycom: Oh, wait. You're correct. Bleh.

str4d (approved):
utACK ae60e142b23050a1fa3f324402d078e02a61a532 with nits:

- I'm a tad uncomfortable about an "Apply suggestions" commit being after the release commit (0bccc10ecf37e44aeca8eff17ecfa01331dde4e6), but not enough to block.
- The release commit doesn't actually contain the release bumps; it just has a bugfix, and the actual release bumps are in its parent f8afca18802b4f1092e8751661ed80e9cf9ecbf4.

PR #1992, opened 2025-10-13
feat: Use TransactionData::fee_paid in wallet::determine_fee

### LINKED ISSUE #1869: zcash_client_sqlite: Use `TransactionData::fee_paid` for the implementation of `wallet::determine_fee`
This should all be using `TransactionData::fee_paid` (via the `Transaction` deref, so `tx.fee_paid(_)` should work). That method takes a `FnMut` closure that should therefore be able to do whatever it needs to with `conn` internally to look up the input outpoints.

_Originally posted by @str4d in https://github.com/zcash/librustzcash/pull/1862#discussion_r2216476947_

varun-doshi:
Fixes #1869

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -113,7 +113,7 @@ use zcash_primitives::{
 use zcash_protocol::{
     consensus::{self, BlockHeight, BranchId, NetworkUpgrade, Parameters},
     memo::{Memo, MemoBytes},
-    value::{BalanceError, ZatBalance, Zatoshis},
+    value::{ZatBalance, Zatoshis},
     PoolType, ShieldedProtocol, TxId,
 };
 use zip32::{fingerprint::SeedFingerprint, DiversifierIndex};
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3539,67 +3539,24 @@ fn detect_wallet_transparent_outputs<P: consensus::Parameters>(
...
-                    .map_err(|_| BalanceError::Underflow)?,
-            ))
-        }
+        });
+
+    match fee_result {
+        Ok(Some(fee)) => Ok(Some(fee)),
+        Ok(None) => Ok(None), // cannot compute fee because some inputs are missing
+        Err(e) => Err(SqliteClientError::from(e)),
     }
```
str4d: This is unnecessary; you are already requiring that `fee_result` have type `Result<_, SqliteClientError>`. You can just return that result:
```suggestion
        })
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3539,67 +3539,24 @@ fn detect_wallet_transparent_outputs<P: consensus::Parameters>(
...
-                .map(|o| o.value())
-                .sum::<Option<Zatoshis>>()
-                .ok_or(BalanceError::Overflow)?;
-            let value_balance = (value_balance - t_out_total).ok_or(BalanceError::Underflow)?;
-
-            if t_bundle.vin.is_empty() {
-                Zatoshis::try_from(value_balance)
-                    .map(Some)
-                    .map_err(SqliteClientError::from)
+    let fee_result: Result<std::option::Option<Zatoshis>, SqliteClientError> =
```
```suggestion
```
(and a corresponding `cargo fmt` to fix the indentations).

Removing this could lead to an error where the closure doesn't know what error type to return. If that occurs, the simplest way to fix that is to leave this in and replace the `match fee_result {..}` with `fee_result`, but a cleaner version would be to change the `Ok(None)` in the closure to `Ok::<_, SqliteClientError>(None)` (to provide the error type hint to the compiler).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3539,67 +3539,24 @@ fn detect_wallet_transparent_outputs<P: consensus::Parameters>(
...
+        tx.fee_paid(|outpoint| {
+            // This closure can do DB lookups to fetch the value of each transparent input.
+            if let Some(out) = get_wallet_transparent_output(conn, outpoint, true)? {
```
str4d: Bug: `get_wallet_transparent_output` is behind a feature flag. You need to provide an alternative when the feature flag is disabled:
```suggestion
        tx.fee_paid(|outpoint| {
            #[cfg(not(feature = "transparent-inputs"))]
            {
                // Transparent inputs aren't supported, so this closure should never be
                // called during transaction construction. But in case it is, handle it
                // correctly.
                Ok(None)
            }
            
            // This closure can do DB lookups to fetch the value of each transparent input.
            #[cfg(feature = "transparent-inputs")]
            if let Some(out) = get_wallet_transparent_output(conn, outpoint, true)? {
```

str4d (changes requested):

PR #1996, opened 2025-10-15
zcash_client_backend: Add target height to `InputSource::get_spendable_note`

nuttycom:
It doesn't make sense to consider a note "spendable" absent the context of the target height at which it is intended to be spent.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1440,10 +1440,11 @@ pub trait InputSource {
         exclude: &[Self::NoteRef],
     ) -> Result<AccountMeta, Self::Error>;
 
-    /// Fetches the transparent output corresponding to the provided `outpoint`.
+    /// Fetches the transparent output corresponding to the provided `outpoint` if it is considered
+    /// spendable as of the provided `target_height`.
     ///
-    /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or is not
-    /// spendable as of the chain tip height.
+    /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or would not be
+    /// spendable in a transaction mined in the block at the target height.
     #[cfg(feature = "transparent-inputs")]
     fn get_unspent_transparent_output(
         &self,
@@ -1457,7 +1458,8 @@ pub trait InputSource {
     /// such that, at height `target_height`:
     /// * the transaction that produced the output had or will have at least `min_confirmations`
     ///   confirmations; and
-    /// * the output is unspent as of the current chain tip.
+    /// * the output can potentially be spent in a transaction mined in a block at the given
+    ///   `target_height`.
     ///
     /// An output that is potentially spent by an unmined transaction in the mempool is excluded
     /// iff the spending transaction will not be expired at `target_height`.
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -1391,12 +1391,13 @@ pub trait InputSource {
...
     ///
     /// Returns `Ok(None)` if the note is not known to belong to the wallet or if the note
-    /// is not spendable.
+    /// is not spendable as of the given height.
     fn get_spendable_note(
         &self,
         txid: &TxId,
         protocol: ShieldedProtocol,
         index: u32,
+        target_height: TargetHeight,
```
str4d, line 1400: If we're going to make this change with the given rationale (that an `InputSource` can't know what the "chain tip" is: https://github.com/zcash/librustzcash/pull/1987#discussion_r2422223178), we should be consistent.
- Update the `InputSource` trait documentation to make clear that implementations of it must not assume a particular chain tip height.
- Either add a `target_height` field to `InputSource::get_account_metadata` (which must be used to decide whether notes are spent and thus excluded), or documentation how it should handle the above requirement that it not assume a particular chain height, or move it to `WalletRead`.
- Either add a `target_height` field to `InputSource::get_unspent_transparent_output`, or (given that it is almost a pure coin getter) move it to `WalletRead`.

zcash_client_backend/src/fees.rs
```diff
@@ -510,6 +510,7 @@ pub trait ChangeStrategy {
         &self,
         meta_source: &Self::MetaSource,
         account: <Self::MetaSource as InputSource>::AccountId,
+        target_height: TargetHeight,
```
str4d, line 513: Document this change in the changelog.

zcash_client_backend/src/data_api.rs
```diff
@@ -1446,6 +1448,7 @@ pub trait InputSource {
     fn get_unspent_transparent_output(
         &self,
         _outpoint: &OutPoint,
+        _target_height: TargetHeight,
```
str4d, line 1452: Update the method documentation to reflect this change.

zcash_client_backend/src/data_api.rs
str4d: This method already takes a `target_height` but its documentation references the chain tip. Fix the documentation to correctly constrain what the method should be doing without reference to any specific chain tip.

zcash_client_backend/src/data_api.rs
```diff
@@ -1904,13 +1907,18 @@ pub trait WalletTest: InputSource + WalletRead {
...
     ///
-    /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or is not
-    /// spendable as of the chain tip height.
+    /// # Parameters
+    /// - `outpoint`: The identifier for the output to be retrieved.
+    /// - `spendable_as_of`: The target height of a transaction under construction that will spend the
+    ///   returned output. If this is `None`, no spendability checks are performed.
+    ///
+    /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or if `spendable_as_of`
+    /// is set and the output is not spendable as of the given target height.
```
str4d: "spendable as of the given target height" makes it sound like this is interpreting the target height like a chain tip (i.e. spendable as of a hypothetical chain tip block at this height), when I think what is actually intended is "spendable in a block at the given target height". Documentation should be clarified, and it might be a good idea to rename `spendable_as_of`.

str4d (approved):
utACK f066597704a7b353e040101082e812edfaf54e32

PR #1997, opened 2025-10-16
zcash_client_sqlite: delete irrelevant enhancement requests

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3565,6 +3565,7 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
         && wallet_transparent_outputs.is_empty()
         && !d_tx.has_decrypted_outputs()
     {
+        delete_retrieval_queue_entries(conn, d_tx.tx().txid())?;
```
daira, line 3568: I suspect that we are not guaranteed to reach https://github.com/zcash/librustzcash/pull/1997/files#diff-a1724f7547ecabc613a07520eaf7917d9084d4277219d186698b0c8e352495a4R3946 where the next call to `delete_retrieval_queue_entries` is, if there is an `?` early return between here and there.

str4d: There's an early return on the very next line, so we absolutely do not reach the other one.

PR #1999, opened 2025-10-17
zcash_client_sqlite: Set `min_observed_height` correctly in `set_transaction_status`

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3030,6 +3030,7 @@ pub(crate) fn set_transaction_status(
             conn.execute(
                 "UPDATE transactions
                  SET mined_height = :height,
+                     min_observed_height = MIN(IFNULL(mined_height, :height), :height),
```
str4d: Bug: this ignores the current value of `min_observed_height` and can cause it to increase. We need the minimum of all three values.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1035,14 +1035,14 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> WalletRea
     }
 
     fn transaction_data_requests(&self) -> Result<Vec<TransactionDataRequest>, Self::Error> {
-        if let Some(chain_tip_height) = wallet::chain_tip_height(self.conn.borrow())? {
+        if let Some(_chain_tip_height) = wallet::chain_tip_height(self.conn.borrow())? {
```
daira, line 1038: BTW, I mildly prefer `#[allow(unused_variables)]` (because I think of that as expressing "~~clippy~~ rustc is wrong here"). Actually cases like this are arguably an outright ~~clippy~~ rustc bug, so I filed

daira: Filed https://github.com/rust-lang/rust/issues/147836

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3030,6 +3030,11 @@ pub(crate) fn set_transaction_status(
             conn.execute(
                 "UPDATE transactions
                  SET mined_height = :height,
+                     min_observed_height = MIN(
+                        min_observed_height,
+                        IFNULL(mined_height, :height),
+                        :height
+                     ),
```
daira, line 3037: `MIN(NULL, ...)` returns NULL, which I think is not what you want here if `min_observed_height` was previously NULL.

str4d: Per `zcash_client_sqlite/src/wallet/db.rs`:
https://github.com/zcash/librustzcash/blob/32f8ad963c1d2a62e03236c2daa0bcefa2daa1c5/zcash_client_sqlite/src/wallet/db.rs#L286

PR #2005, opened 2025-10-22
zcash_client_sqlite: Exclude dust-valued outputs from input selection & correctly represent them in balance calculation.

nuttycom:


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -866,20 +896,10 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
...
-            )
-            -- or the utxo can be spent with zero confirmations and is definitely unexpired
-            OR (
-                :min_confirmations = 0
-                AND (t.expiry_height = 0 OR t.expiry_height >= :target_height)
-            )
-         )
-         -- and the output is unspent
-         AND u.id NOT IN ({})",
+         AND u.value_zat > :min_value
```
nuttycom: This is the critical change to avoid dust in note selection.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1081,7 +1091,13 @@ pub(crate) fn add_transparent_account_balances(
         account_balances
             .entry(account)
             .or_insert(AccountBalance::ZERO)
-            .with_unshielded_balance_mut(|bal| bal.add_spendable_value(value))?;
+            .with_unshielded_balance_mut(|bal| {
```
nuttycom, line 1090: This is one of the two critical changes to exclude dust from balance calculation.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1127,7 +1143,13 @@ pub(crate) fn add_transparent_account_balances(
             account_balances
                 .entry(account)
                 .or_insert(AccountBalance::ZERO)
-                .with_unshielded_balance_mut(|bal| bal.add_pending_spendable_value(value))?;
+                .with_unshielded_balance_mut(|bal| {
+                    if value >= zip317::MARGINAL_FEE {
```
nuttycom, line 1143: This is the other.

zcash_client_sqlite/src/wallet/transparent.rs
str4d: Possible bug: this uses `> MARGINAL_FEE`, but the other two instances use `>= MARGINAL_FEE`. This means the spendable balance may appear larger than what can actually be spent, leaving a user's wallet permanently showing a "Shield funds" button. Given that these notes can always be spent, and sometimes spent without being entirely consumed as fee (using an already-paid-for spend):
```suggestion
         AND u.value_zat >= :min_value
```

str4d (changes requested):
Reviewed e45a4db04c376ed2505dd14df3e9c9b93a88632b.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -896,7 +916,8 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
     let mut rows = stmt_utxos.query(named_params![
         ":address": addr_str,
         ":target_height": u32::from(target_height),
-        ":min_confirmations": min_confirmations
+        ":min_confirmations": min_confirmations,
+        ":min_value": u64::from(zip317::MARGINAL_FEE)
     ])?;
 
     let mut utxos = Vec::<WalletTransparentOutput>::new();
```
daira, line 923: Lines 1009-1020 below appear to be another instance where `tx_unexpired_condition` could be used.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1046,29 +1067,16 @@ pub(crate) fn add_transparent_account_balances(
          JOIN accounts a ON a.id = u.account_id
          JOIN transactions t ON t.id_tx = u.transaction_id
          JOIN addresses ON addresses.id = u.address_id
-         WHERE addresses.key_scope != :key_scope_ephemeral
```
nuttycom, line 1049: Hmmm... removal of the `key_scope` check may either need the changes from https://github.com/zcash/librustzcash/pull/1969/commits/a0a84c32c7cf3929bb2994cfad69f062be79d4c8 in order to avoid problems until the next release, or else that removal needs to be reverted and a corresponding key scope check should be introduced for UTXO selection.

str4d: This removal would be a reversion of #1985, and would on its own undo whatever bugfix that was doing.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: No; in this case, we're explicitly looking for transactions that are in a "pending" state with respect to the minimum number of confirmations, and if you look carefully it's not the same set of conditions at all.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -936,29 +957,21 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
-                AND (t.expiry_height = 0 OR t.expiry_height >= :target_height)
-            )
-         )
-         -- and the output is unspent
-         AND u.id NOT IN ({})",
+         AND a.key_scope != :ephemeral_key_scope
+         AND u.value_zat > 0
+         AND ({}) -- the transaction is mined or unexpired with minconf 0
+         AND u.id NOT IN ({}) -- and the output is unspent",
+        tx_unexpired_condition("t"),
```
nuttycom, line 966: Bug: this is supposed to be `tx_unexpired_condition_minconf_0`

daira (commented):
Post-hoc ACK modulo the bug @nuttycom pointed out.

PR #2007, opened 2025-10-23
More CI fixes

str4d:


schell (dismissed):
LGTM. BTW, I've also had good luck with https://github.com/moonrepo/setup-rust for the initial toolchain.

.github/workflows/ci.yml
```diff
@@ -405,26 +405,30 @@ jobs:
   clippy:
     name: Clippy (MSRV)
     runs-on: ubuntu-latest
+    permissions:
+      checks: write
```
nuttycom, line 409: Why are write permissions needed by this action?

nuttycom (dismissed):
utACK with question.

.github/workflows/ci.yml
schell: Hrm - maybe this is a clippy action that writes lints inline inside the PR itself?

str4d: Indeed that's what it does (hence specifically write permissions for GitHub checks).

.github/workflows/ci.yml
```diff
@@ -495,14 +497,21 @@
           persist-credentials: false
       - id: prepare
         uses: ./.github/actions/prepare
+      - uses: dtolnay/rust-toolchain@nightly
```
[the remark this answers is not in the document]
schell: Does this specify the nightly toolchain or the nightly version of this action?

nuttycom (approved):
utACK 06950e3ad51b6c84d11d1682762c752e0981ac40

.github/workflows/ci.yml
```diff
@@ -405,26 +405,30 @@ jobs:
...
-        uses: actions-rs/clippy-check@v1
+        uses: auguwu/clippy-action@94a9ff2f6920180b89e5c03d121d0af04a9d3e03 # 1.4.0
         with:
-          name: Clippy (MSRV)
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: >
-            ${{ steps.prepare.outputs.feature-flags }}
+          working-directory: ${{ inputs.target }}
+          check-args: >
+            --all-features
```
str4d, line 422: This uses `--all-features` now because `auguwu/clippy-action` doesn't support spaces inside arguments (which is how `steps.prepare.outputs.feature-flags` is built). As a side-effect, we will now detect clippy errors under feature flags that we don't normally enable (namely unstable feature flags), which might be fragile at times but is better than not detecting clippy errors under regular feature flags.

.github/workflows/ci.yml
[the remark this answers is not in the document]
str4d: The nightly toolchain. As I said in the dismissal for this, we *want* this to be unpinned so that we are using latest nightly as of the CI run.

PR #2008, opened 2025-10-23
zcash_client_backend: Return key scope from WalletRead::get_transparent_balances

nuttycom:
Wallets need key scope information to inform their behavior whenever they are interacting with light wallet servers or shielding funds. Returning the key scope facilitates this interaction.

### DIFF
--- a/zcash_client_memory/src/types/memory_wallet/mod.rs
+++ b/zcash_client_memory/src/types/memory_wallet/mod.rs
@@ -1038,6 +1038,7 @@ impl<P: consensus::Parameters> MemoryWalletDb<P> {
     }
 
     #[cfg(feature = "transparent-inputs")]
+    #[allow(unreachable_code, unused_variables)] //FIXME: need address key scope detection
     pub(crate) fn put_transparent_output(
         &mut self,
         output: &WalletTransparentOutput,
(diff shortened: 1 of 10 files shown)


zcash_client_sqlite/src/testing/db.rs
```diff
@@ -44,7 +44,11 @@ use crate::{
 #[cfg(feature = "transparent-inputs")]
 use {
     crate::TransparentAddressMetadata,
-    ::transparent::{address::TransparentAddress, bundle::OutPoint, keys::NonHardenedChildIndex},
+    ::transparent::{
+        address::TransparentAddress,
+        bundle::OutPoint,
+        keys::{NonHardenedChildIndex, TransparentKeyScope},
```
str4d, line 50: Why does this import change, when nothing else in the file does to use it?

nuttycom: Because of the way that the delegate macro works.

zcash_client_memory/src/types/memory_wallet/mod.rs
```diff
@@ -1103,6 +1105,7 @@ impl<P: consensus::Parameters> MemoryWalletDb<P> {
                     txid,
                     *receiving_account,
                     *address,
+                    todo!("look up the key scope for the address"),
```
daira, line 1108: Is there an issue for this TODO?

nuttycom: This is in `zcash_client_memory` again. We do not create issues for all the changes that need to be made here (there are dozens.)

PR #2009, opened 2025-10-23
zcash_client_sqlite: Exclude only wallet-internal ephemeral outputs from balance computation.

nuttycom:
We expect all outputs to ephemeral addresses to fall into one of two categories:
- A wallet-internal deshielding output created as part of sending to a TEX address.
- An output sent to us after we provided a one-time use ephemeral address to some third party, such as for a DEX swap or onramp transaction.

As such, we modify transparent balance calculation to ignore ephemeral outputs that fall into the first category, because we do not expect these to become part of the wallet's spendable balance; only in the case of a partial TEX transaction failure (where the first transaction in the pair is mined, but the second fails to broadcast and/or is never mined) will we exclude outputs that need to be re-shielded to the wallet.

We need to make this change because we cannot yet reliably detect purely-transparent transactions involving ephemeral addresses without introducing privacy leaks, except in limited cases. In wallet restore scenarios, we want to make sure that transparent functionality in the wallet doesn't end up bricked by the wallet incorrectly treating previously-spent UTXOs as spendable.

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -952,7 +952,7 @@ pub(crate) fn spent_utxos_clause() -> String {
 /// - `tx` must be set to the alias for the `transactions` table in the enclosing scope.
 /// - `accounts` must be set to the alias for the `accounts` table in the enclosing scope.
 /// - The parent is responsible for enclosing this condition in parentheses as appropriate.
-pub(crate) fn tx_account_internal_clause(tx: &str, accounts: &str) -> String {
+pub(crate) fn tx_spends_account_outputs_condition(tx: &str, accounts: &str) -> String {
     format!(
         r#"
         {tx}.id_tx IN (
@@ -1006,7 +1006,7 @@ pub(crate) fn get_wallet_transparent_output(
          )",
         tx_unexpired_condition("t"),
         spent_utxos_clause(),
-        tx_account_internal_clause("t", "accounts")
+        tx_spends_account_outputs_condition("t", "accounts")
     ))?;
 
     let result: Result<Option<WalletUtxo>, SqliteClientError> = stmt_select_utxo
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -945,6 +945,25 @@ pub(crate) fn spent_utxos_clause() -> String {
...
 }
 
+/// Generates a SQL condition that checks whether a transaction spends outputs controlled by an
+/// account.
+///
+/// # Usage requirements
+/// - `tx` must be set to the alias for the `transactions` table in the enclosing scope.
+/// - `accounts` must be set to the alias for the `accounts` table in the enclosing scope.
+/// - The parent is responsible for enclosing this condition in parentheses as appropriate.
+pub(crate) fn tx_account_internal_clause(tx: &str, accounts: &str) -> String {
```
str4d: We've used `*_clause` to indicate full `SELECT` etc statements, and `*_condition` to specifically indicate conditions for use in a `WHERE`:
```suggestion
pub(crate) fn tx_spends_account_outputs_condition(tx: &str, accounts: &str) -> String {
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -976,10 +996,17 @@ pub(crate) fn get_wallet_transparent_output(
...
+                 AND NOT (
+                    addresses.key_scope = :ephemeral_key_scope
+                    AND {} -- the transaction that created the txo spends account-internal outputs
+                 )
              )
          )",
         tx_unexpired_condition("t"),
-        spent_utxos_clause()
+        spent_utxos_clause(),
+        tx_account_internal_clause("t", "accounts")
```
```suggestion
        tx_spends_account_outputs_condition("t", "accounts"),
```
and similarly below. (Also added a trailing comma because they are allowed here, and it would avoid an extra line change in future additions to the format macro.)

zcash_client_sqlite/src/wallet/transparent.rs
str4d: (Also there's nothing about this condition that specialises it to checking for account-internal txs only, so I suggest a rename.)

str4d (changes requested):
utACK 390742a5c264f0a394b04269e03f596b91c8e986 after rename.

PR #2011, opened 2025-10-24
chore: Testing Scenario DSL

schell:
This PR introduces a wrapper over `TestBuilder` and `TestState` that provides a scenario-driven API with which to create tests that are easily grokked. 

So far the test API covers the "add funds" portion of a test scenario.

Existing tests were refactored in terms of this DSL where appropriate.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -207,7 +207,7 @@ pub fn send_single_step_proposed_transfer<T: ShieldedPoolTester>(
     let mut st = TestDsl::with_sapling_birthday_account(dsf, cache).build::<T>();
 
     // Add funds to the wallet in a single note
-    let (h, _, _) = st.add_a_single_note(Zatoshis::const_from_u64(60000));
+    let (h, _, _) = st.add_a_single_note_checking_balance(Zatoshis::const_from_u64(60000));
 
     let to_extsk = T::sk(&[0xf5; 32]);
     let to: Address = T::sk_default_address(&to_extsk);
@@ -385,8 +385,9 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
     };
     let min_confirmations = u32::from(min_confirmations);
 
-    let (_, r, _) = st
-        .add_a_single_note(TestNoteConfig::from(starting_balance).with_address_type(address_type));
+    let (_, r, _) = st.add_a_single_note_checking_balance(
+        TestNoteConfig::from(starting_balance).with_address_type(address_type),
+    );
     let txid = r.txids()[0];
 
     // Mark the external input as explicitly trusted, if so requested
@@ -459,7 +460,7 @@ pub fn spend_max_spendable_single_step_proposed_transfer<T: ShieldedPoolTester>(
     // Add funds to the wallet in two notes over 5 blocks
     let value = Zatoshis::const_from_u64(60000);
     let h = st
-        .add_notes([Some(value), None, None, None, Some(value)])
+        .add_notes_checking_balance([Some(value), None, None, None, Some(value)])
         .block_height()
         .unwrap();
 
@@ -611,7 +612,7 @@ pub fn spend_everything_single_step_proposed_transfer<T: ShieldedPoolTester>(
     let account = st.test_account().cloned().unwrap();
 
     // Add funds to the wallet in a single note
-    let (h, _, _) = st.add_a_single_note(Zatoshis::const_from_u64(60000));
+    let (h, _, _) = st.add_a_single_note_checking_balance(Zatoshis::const_from_u64(60000));
 
     let to_extsk = T::sk(&[0xf5; 32]);
     let to: Address = T::sk_default_address(&to_extsk);
@@ -743,7 +744,7 @@ pub fn fails_to_send_max_spendable_to_transparent_with_memo<T: ShieldedPoolTeste
     let mut st = TestDsl::with_sapling_birthday_account(dsf, cache).build::<T>();
 
     // Add funds to the wallet in a single note
-    st.add_a_single_note(Zatoshis::const_from_u64(60000));
+    st.add_a_single_note_checking_balance(Zatoshis::const_from_u64(60000));
 
     let account = st.test_account().cloned().unwrap();
     let (default_addr, _) = account.usk().default_transparent_address();
@@ -785,7 +786,7 @@ pub fn spend_everything_proposal_fails_when_unconfirmed_funds_present<T: Shielde
     cache: impl TestCache,
 ) {
     let mut st = TestDsl::with_sapling_birthday_account(dsf, cache).build::<T>();
-    st.add_notes([
+    st.add_notes_checking_balance([
         Some(Zatoshis::const_from_u64(60000)),
         None,
         None,
@@ -840,8 +841,7 @@ pub fn spend_everything_proposal_fails_when_unconfirmed_funds_present<T: Shielde
 /// - Add more funds
 /// - Attempts to construct a request to spend the whole balance to an external address in the
 ///   same pool.
-/// - succeds at doing so
-// TODO(schell): as far as I can tell this function is unused in our test suite
+/// - succeeds at doing so
 pub fn send_max_spendable_proposal_succeeds_when_unconfirmed_funds_present<
     T: ShieldedPoolTester,
 >(
@@ -850,7 +850,7 @@ pub fn send_max_spendable_proposal_succeeds_when_unconfirmed_funds_present<
 ) {
     let mut st = TestDsl::with_sapling_birthday_account(dsf, cache).build::<T>();
     let h = st
-        .add_notes([
+        .add_notes_checking_balance([
             Some(Zatoshis::const_from_u64(60000)),
             None,
             None,
@@ -1010,7 +1010,7 @@ pub fn spend_everything_multi_step_single_note_proposed_transfer<T: ShieldedPool
     let value = Zatoshis::const_from_u64(100000);
 
     // Add funds to the wallet.
-    st.add_a_single_note(value);
+    st.add_a_single_note_checking_balance(value);
     let initial_balance = value;
     assert_eq!(
         st.get_spendable_balance(account_id, ConfirmationsPolicy::MIN),
@@ -1167,7 +1167,7 @@ pub fn spend_everything_multi_step_many_notes_proposed_transfer<T: ShieldedPoolT
 
     // Add funds to the wallet.
     for _ in 0..number_of_notes {
-        st.add_a_single_note(note_value);
+        st.add_a_single_note_checking_balance(note_value);
     }
 
     let initial_balance = value;
@@ -1321,8 +1321,8 @@ pub fn spend_everything_multi_step_with_marginal_notes_proposed_transfer<
         (note_value * number_of_notes).expect("sum of notes should not fail.");
 
     for _ in 0..number_of_notes {
-        st.add_a_single_note(note_value);
-        st.add_a_single_note(zip317::MARGINAL_FEE);
+        st.add_a_single_note_checking_balance(note_value);
+        st.add_a_single_note_checking_balance(zip317::MARGINAL_FEE);
     }
 
     let account = st.test_account().cloned().unwrap();
@@ -1463,7 +1463,7 @@ pub fn send_with_multiple_change_outputs<T: ShieldedPoolTester>(
 
     // Add funds to the wallet in a single note
     let value = Zatoshis::const_from_u64(650_0000);
-    let (h, _, _) = st.add_a_single_note(value);
+    let (h, _, _) = st.add_a_single_note_checking_balance(value);
 
     let to_extsk = T::sk(&[0xf5; 32]);
     let to: Address = T::sk_default_address(&to_extsk);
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1475,7 +1477,7 @@ impl Default for TestBuilder<(), ()> {
 
 impl<A> TestBuilder<(), A> {
     /// Adds a block cache to the test environment.
-    pub fn with_block_cache<C: TestCache>(self, cache: C) -> TestBuilder<C, A> {
+    pub fn with_block_cache<C>(self, cache: C) -> TestBuilder<C, A> {
```
schell, line 1470: This may seem controversial since it appears to be reducing the safety of the resulting `TestBuilder` (in that with this change we can't guarantee the `C` impls `TestCache` on the result), but `with_block_cache` doesn't use any of those methods internally, and `build` _also_ doesn't have this constraint. On top of that I had a genuine need to relax this.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -2121,9 +2110,10 @@ impl TestFvk for ::orchard::keys::FullViewingKey {
 /// Configures how a [`TestFvk`] receives a particular output.
 ///
 /// Used with [`TestFvk::add_output`] and [`TestFvk::add_logical_action`].
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Default, Debug)]
 pub enum AddressType {
     /// The output will be sent to the default address of the full viewing key.
+    #[default]
```
daira: Is this needed? I think it is clearer to explicitly say `AddressType::DefaultExternal`.

schell: Good catch! Indeed it's not needed.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4878,14 +4477,15 @@ pub fn scan_cached_blocks_finds_change_notes<T: ShieldedPoolTester, DSF>(
...
 
-    // Scan the cache
-    st.scan_cached_blocks(received_height, 1);
+    // // Scan the cache
+    // st.scan_cached_blocks(received_height, 1);
 
-    // Account balance should reflect the received note
-    assert_eq!(st.get_total_balance(account.id()), value);
+    // // Account balance should reflect the received note
+    // assert_eq!(st.get_total_balance(account.id()), value);
```
daira: Why is this commented out? Are we losing any useful test coverage here?

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+        .into()
+    }
+}
+
+/// A proxy for `FakeCompactOutput` that allows test code to omit the `fvk` and
+/// `address_type` fields, which can be derived from the `TestState` in most cases.
+pub struct TestNoteConfig<T: ShieldedPoolTester> {
+    pub value: Zatoshis,
+    /// Diversifiable full viewing key of
```
daira: Incomplete sentence.

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,377 @@
...
+            _phantom: PhantomData,
+        }
+        .into()
+    }
+}
+
+/// A proxy for `FakeCompactOutput` that allows test code to omit the `fvk` and
+/// `address_type` fields, which can be derived from the `TestState` in most cases.
+pub struct TestNoteConfig<T: ShieldedPoolTester> {
+    pub value: Zatoshis,
```
daira, line 97: `pub` fields should have doc comments.

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,377 @@
...
+            address_type.unwrap_or(AddressType::DefaultExternal),
+            *value,
+        )
+    }
+
+    /// Add funds from multiple notes in one or more blocks, or generate empty blocks.
+    ///
+    /// This step also verifies that the test account contains the expected
+    /// funds as part of the _total_. Keep in mind that these funds may not yet
+    /// be _spendable_ due to the number confirmations required.
```
daira:
```suggestion
    /// be _spendable_ due to the number of confirmations required.
```

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,377 @@
...
+            .unwrap_or(Zatoshis::ZERO);
+        let mut expected_total = starting_balance;
+        let mut summary = AddFundsSummary::default();
+        for notes in blocks.into_iter() {
+            // Collect the notes while also counting their value.
+            let (fake_outputs, note_configs): (Vec<_>, Vec<_>) = notes
+                .into_iter()
+                .map(|into_note_config| {
+                    let note_config = into_note_config.into();
+                    if note_config.value > zip317::MARGINAL_FEE {
```
daira, line 328: This is correct, but led me to file #2041.

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,377 @@
...
+                        // Don't include dust in the expected total, as the balance
+                        // won't include it.
```
daira:
```suggestion
                        // Don't include uneconomic notes in the expected total, as the
                        // balance won't include them.
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3519,38 +3190,24 @@ pub fn external_address_change_spends_detected_in_restore_from_seed<T: ShieldedP
...
 
     // Add funds to the wallet
-    let (h1, _, _) = st.generate_next_block(
-        &dfvk,
-        AddressType::Internal,
-        Zatoshis::const_from_u64(50000),
-    );
+    st.add_notes([Some(Zatoshis::const_from_u64(50000))]);
 
     // Add 10 dust notes to the wallet
```
```suggestion
    // Add 10 uneconomic notes to the wallet
```

daira (dismissed):
utACK with minor suggestions.

zcash_client_backend/src/data_api/testing/pool.rs
schell: Good catch, but no, these assertions happen implicitly inside `add_single_note` now.

nuttycom (changes requested):
Blocking on opportunity to review - I will try to get to this today.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -954,42 +840,25 @@ pub fn spend_everything_proposal_fails_when_unconfirmed_funds_present<T: Shielde
 /// - Attempts to construct a request to spend the whole balance to an external address in the
 ///   same pool.
 /// - succeds at doing so
+// TODO(schell): as far as I can tell this function is unused in our test suite
```
daira: That's probably a bug. It should be called in the same pattern as the other functions in this module. This is quite error-prone (it's easy to forget to call a test) and I'd like to find a better way of doing it.

daira (approved):
utACK with a non-blocking comment. If that comment is not resolved then please file an issue about it. This does not override @nuttycom's request to wait for his review.

zcash_client_backend/src/data_api/testing.rs
nuttycom: If the method doesn't depend on the constraint, this is fine (and, indeed, preferable).

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///
+    /// Shorthand for the following:
+    /// ```rust,ignore
+    /// let dsl: TestDsl<TestBuilder<_, _>> = TestBuilder::new()
+    ///     .with_data_store_factory(dsf)
+    ///     .with_block_cache(tc)
+    ///     .with_account_from_sapling_activation(BlockHash([0; 32]))
+    ///     .into();
+    /// ```
+    pub fn with_standard_sapling_account(dsf: Dsf, tc: Cache) -> Self {
```
nuttycom: This naming is slightly off; it's not a sapling account, it's an account with its birthday at the Sapling activation height. Maybe `with_account_from_sapling_activation`?

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+impl<T: ShieldedPoolTester, C: TestCache> AddFundsSummary<T, C> {
+    /// Return the first block height.
+    pub fn first_block_height(&self) -> BlockHeight {
+        if let Some(step) = self.steps.first() {
+            step.results.block_height
+        } else {
+            BlockHeight::from_u32(0)
```
nuttycom: This result isn't documented, and I'm puzzled trying to figure out what this is used for. If the list of steps is empty, why don't we just return `None`?

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+
+    /// Return the latest block height after generating the blocks
+    /// that added funds.
+    pub fn block_height(&self) -> BlockHeight {
+        if let Some(step) = self.steps.last() {
+            step.results.block_height
+        } else {
+            BlockHeight::from_u32(0)
```
nuttycom: I would have both of these return `None`. Also, do we not have `BlockHeight::GENESIS` or `BlockHeight::ZERO`? We should.

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///
+    /// Returns a summary of steps.
+    ///
+    /// ## Note
+    /// Keep in mind:
+    /// * Each block coalesces these notes into a single transaction.
+    /// * Funds are added to the default test account.
+    pub fn add_notes(
+        &mut self,
+        blocks: impl IntoIterator<Item = impl IntoIterator<Item = impl Into<TestNoteConfig<T>>>>,
```
nuttycom, line 310: This argument needs better documentation; it's kind of obvious what it should do from the type but not totally so.

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,379 @@
...
+    ) -> AddFundsSummary<T, Cache> {
+        let mut from_height = None;
+        let mut current_height = BlockHeight::from_u32(0);
```
nuttycom: This should probably be initialized to 
```
self.wallet()
                .block_max_scanned()
                .unwrap()
                .unwrap()
                .block_height(),
```
which then will avoid a confusing assertion if no blocks were provided (which is abnormal, but still a valid call to the API.)

zcash_client_backend/src/data_api/testing/pool/dsl.rs
```diff
@@ -0,0 +1,379 @@
...
+            // Collect the notes while also counting their value.
+            let (fake_outputs, note_configs): (Vec<_>, Vec<_>) = notes
+                .into_iter()
+                .map(|into_note_config| {
+                    let note_config = into_note_config.into();
+                    if note_config.value > zip317::MARGINAL_FEE {
+                        // Don't include uneconomic (dust) notes in the expected
+                        // total, as the balance won't include them.
+                        expected_total = (expected_total + note_config.value).unwrap();
+                    }
```
nuttycom, line 332: This is a weird place to hardcode this behavior. Maybe the dust threshold should at least be part of the builder, but when we add the ability to consume dust notes in grace inputs then even that would likely break.

Maybe this is okay for now, but it seems like something that will be annoying to track down from test failures if we change how balance calculation is done.

This would be fine if this reflected the documented semantics of balance calculation, but I'm not sure that the balance calculation APIs currently document such a contract.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -411,17 +385,9 @@ pub fn zip_315_confirmations_test_steps<T: ShieldedPoolTester>(
     };
     let min_confirmations = u32::from(min_confirmations);
 
-    // Generate a block to create the inputs that we will spend.
-    let (h, r, _) = st.generate_next_block(&dfvk, address_type, starting_balance);
+    let (_, r, _) = st
+        .add_a_single_note(TestNoteConfig::from(starting_balance).with_address_type(address_type));
```
nuttycom: Maybe this would be better named `add_a_single_note_checking_balance`

zcash_client_backend/src/data_api/testing/pool.rs
nuttycom: If we don't have any uses of this, we should add one and see whether it passes.

nuttycom (dismissed):
utACK with naming nits. Also, I would prefer that the `DSF -> Dsf` capitalization change be reverted because it makes the diff here noisier than it needs to be.

zcash_client_backend/src/data_api/testing/pool.rs
schell: I added a test for this function on `main` and it does not pass. I'm guessing that's why it's currently unused?

PR #2012, opened 2025-10-24
zcash_client_backend: support account deletion

### LINKED ISSUE #1633: zcash_client_backend: support account deletion
It should be possible to entirely delete data associated with a single account.

nuttycom:
This has been rebased after merging #1958.

Closes #1633.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -2881,7 +2881,8 @@ pub trait WalletWrite: WalletRead {
     ///
     /// WARNING: This is a destructive operation and may result in the permanent loss of
     /// potentially important information that is not recoverable from chain data, including:
-    /// * Data about transactions sent by the account for which a random OVK was used;
+    /// * Data about transactions sent by the account for which [`OvkPolicy::Discard`] (or
+    ///   [`OvkPolicy::Custom`] with random OVKs) was used;
     /// * Data related to transactions that the account attempted to send that expired or were
     ///   otherwise invalidated without having been mined in the main chain;
     /// * Data related to transactions that were observed in the mempool as having inputs or
@@ -2889,6 +2890,9 @@ pub trait WalletWrite: WalletRead {
     /// * Data related to transactions that were received by the wallet in a mined block, where
     ///   that block was later un-mined in a chain reorg and the transaction was either invalidated
     ///   or was never re-mined.
+    ///
+    /// [`OvkPolicy::Discard`]: crate::wallet::OvkPolicy::Discard
+    /// [`OvkPolicy::Custom`]: crate::wallet::OvkPolicy::Custom
     fn delete_account(&mut self, account: Self::AccountId) -> Result<(), Self::Error>;
 
     /// Imports the given pubkey into the account without key derivation information, and adds the
--- a/zcash_client_sqlite/src/wallet/init.rs
+++ b/zcash_client_sqlite/src/wallet/init.rs
@@ -772,6 +772,9 @@ mod tests {
         let re = Regex::new(r"\s+").unwrap();
         let re_paren = Regex::new(r"([\(\)])").unwrap();
 
+        // Surround each opening or closing parenthesis character with whitespace, and then
+        // replace each occurrence of any amount of whitespace (including newlines) with a single
+        // space.
         let normalize = |s: &str| -> String {
             re.replace_all(&re_paren.replace_all(s, " $1 "), " ")
                 .trim()


zcash_client_backend/src/data_api.rs
```diff
@@ -2877,6 +2876,10 @@ pub trait WalletWrite: WalletRead {
         key_source: Option<&str>,
     ) -> Result<Self::Account, Self::Error>;
 
+    /// Deletes the specified account, and all transactions that exclusively involve it, from the
+    /// wallet database.
+    fn delete_account(&mut self, account: Self::AccountId) -> Result<(), Self::Error>;
```
str4d, line 2896: Document this trait change in the changelog.

zcash_client_sqlite/src/wallet/init.rs
```diff
@@ -805,8 +806,10 @@ mod tests {
         assert_eq!(rows.len(), expected_tables.len());
         for (actual, expected) in rows.iter().zip(expected_tables.iter()) {
             assert_eq!(
-                re.replace_all(actual, " "),
-                re.replace_all(expected, " ").trim(),
+                re.replace_all(&re_paren.replace_all(actual, " $1 "), " ")
+                    .trim(),
```
str4d: This is now complex enough that we should have a `normalize` helper closure just underneath the regex definitions to deduplicate it, with documentation of what it is doing.

str4d:
This is the foreign key graph in the database prior to this PR (bold lines indicate `ON DELETE CASCADE`):
```mermaid
flowchart LR
    addresses ==>|account_id| accounts
    transactions -->|block_height| blocks
    sapling_received_notes ==>|address_id| addresses
    sapling_received_notes ==>|transaction_id| transactions
    sapling_received_notes ==>|account_id| accounts
    sapling_received_note_spends ==>|sapling_received_note_id| sapling_received_notes
    sapling_received_note_spends ==>|transaction_id| transactions
    orchard_received_notes ==>|address_id| addresses
    orchard_received_notes ==>|transaction_id| transactions
    orchard_received_notes ==>|account_id| accounts
    orchard_received_note_spends ==>|orchard_received_note_id| orchard_received_notes
    orchard_received_note_spends ==>|transaction_id| transactions
    transparent_received_outputs ==> |address_id| addresses
    transparent_received_outputs ==>|transaction_id| transactions
    transparent_received_outputs ==>|account_id| accounts
    transparent_received_output_spends ==>|transparent_received_output_id| transparent_received_outputs
    transparent_received_output_spends ==>|transaction_id| transactions
    transparent_spend_map ==>|spending_transaction_id| transactions
    sent_notes ==>|transaction_id| transactions
    sent_notes ==>|from_account_id| accounts
    sent_notes ==>|to_account_id| accounts
    tx_retrieval_queue ==>|dependent_transaction_id| transactions
    transparent_spend_search_queue ==>|transaction_id| transactions
    sapling_tree_checkpoint_marks_removed ==>|checkpoint_id| sapling_tree_checkpoints
    orchard_tree_checkpoint_marks_removed ==>|checkpoint_id| orchard_tree_checkpoints
    nullifier_map ==>|block_height, tx_index| tx_locator_map
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -169,7 +169,7 @@ pub(crate) fn get_nullifiers<N, F: Fn(&[u8]) -> Result<N, SqliteClientError>>(
             "SELECT a.uuid, rn.nf
                  FROM {table_prefix}_received_notes rn
                  JOIN accounts a ON a.id = rn.account_id
-                 JOIN transactions tx ON tx.id_tx = rn.tx
+                 JOIN transactions tx ON tx.id_tx = rn.transaction_id
```
str4d, line 172: Have you confirmed that none of the modified queries are used in any prior migration? If any are, we need to first archive identical versions of the queries within those migrations, before modifying the queries for the renamed columns.

CI / testing is unlikely to spot this, because it will be the data migrations that break.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -564,27 +599,34 @@ CREATE TABLE transparent_spend_map (
...
+        REFERENCES transactions(id_tx) ON DELETE CASCADE,
     output_pool INTEGER NOT NULL,
     output_index INTEGER NOT NULL,
-    from_account_id INTEGER NOT NULL,
+    from_account_id INTEGER NOT NULL
+        REFERENCES accounts(id) ON DELETE CASCADE,
     to_address TEXT,
-    to_account_id INTEGER,
+    to_account_id INTEGER
+        REFERENCES accounts(id) ON DELETE CASCADE,
```
str4d: Bug: This cannot be `ON DELETE CASCADE`, as that results in data inconsistency for cross-account transactions. It similarly can't be `ON DELETE SET NULL` because that violates the `ck_send_note_recipient` constraint.

Instead, `to_account_id` edges should be explicitly converted into `to_address` edges during deletion (matching the fact that the deleted account is now external to the wallet), and:
```suggestion
    to_account_id INTEGER
        REFERENCES accounts(id),
```

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -564,27 +599,34 @@ CREATE TABLE transparent_spend_map (
...
 CREATE TABLE "sent_notes" (
     id INTEGER PRIMARY KEY,
-    tx INTEGER NOT NULL,
+    transaction_id INTEGER NOT NULL
+        REFERENCES transactions(id_tx) ON DELETE CASCADE,
     output_pool INTEGER NOT NULL,
     output_index INTEGER NOT NULL,
-    from_account_id INTEGER NOT NULL,
+    from_account_id INTEGER NOT NULL
+        REFERENCES accounts(id) ON DELETE CASCADE,
```
str4d, line 611: By contrast, this is fine to be `ON DELETE CASCADE`, as the column is non-null reflecting the fact that we never add entries to the `sent_notes` table if we don't know the source (they would instead only appear in the relevant `*_received_*` table).

I think that's going to be a more general rule:
- If a foreign key column is non-null, it must be `ON DELETE CASCADE` because we don't have anything else we can sensibly set it to upon parent deletion.
- If a foreign key column is nullable, it *might* be `ON DELETE CASCADE` depending on why it is nullable (e.g. `sapling_received_notes.address_id` which is nullable to represent notes sent to the internal IVK for which we do not expose addresses, but for which there is no other address it could be set to if a non-internal address *did* connect it to is deleted), but we need to analyse them on a case-by-case basis.

zcash_client_backend/src/data_api.rs
```diff
@@ -2877,6 +2876,10 @@ pub trait WalletWrite: WalletRead {
         key_source: Option<&str>,
     ) -> Result<Self::Account, Self::Error>;
 
+    /// Deletes the specified account, and all transactions that exclusively involve it, from the
+    /// wallet database.
```
str4d, line 2880: The documentation should loudly state that this is a destructive operation, and list what information is discarded as a result of it. In particular, it should point out what information may be permanently lost due to being unrecoverable from chain (e.g. `sent_notes` data that was sent using a `None` OVK, or transactions that expired unmined; we should make an exhaustive list to the best of our knowledge).

We may want to also either add a `WalletRead` method, or alter the return type of this method, to return either a backup of all of an account's data, or a backup of all non-chain-recoverable account data. Wallets could then choose to expose (a summary of) the unrecoverable information to users on the deletion screen, or offer to export it.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -597,15 +639,16 @@ pub(super) const INDEX_SENT_NOTES_TX: &str = r#"CREATE INDEX sent_notes_tx ON "s
...
 pub(super) const TABLE_TX_RETRIEVAL_QUEUE: &str = r#"
-CREATE TABLE tx_retrieval_queue (
+CREATE TABLE "tx_retrieval_queue" (
     txid BLOB NOT NULL UNIQUE,
     query_type INTEGER NOT NULL,
-    dependent_transaction_id INTEGER,
-    request_expiry INTEGER,
-    FOREIGN KEY (dependent_transaction_id) REFERENCES transactions(id_tx)
+    dependent_transaction_id INTEGER
+        REFERENCES transactions(id_tx) ON DELETE CASCADE
```
str4d, line 648: If this retrieval request originated from a transparent input of some transaction, and that downstream transaction is deleted, do we still want to scan this upstream transaction for information? I'm imagining a race condition where two accounts are added to the wallet, and then before syncing is completed, one of them is removed. Could there be edge cases where not traversing this edge means we don't observe data?
- If this upstream transaction involved shielded then we should be fine because we detect that through CompactBlock scanning (either via a spent nullifier or an IVK or OVK detection).
- For transparent information I'm not so sure. This table traverses the transaction graph backwards, and with out-of-order scanning, gap limits don't work correctly (because we aren't looking for the later transparent addresses until we've found the earlier ones). That's not a problem right now because we do direct queries, but if we shift to CompactBlock transparent scanning then we will need to cache transparent outputs that we can't tell aren't ours until we've eliminated any gaps prior to their mined height (same as we need to cache the nullifier map). And I think that cache is either part of this table, or part of `transparent_spend_search_queue`, or similar to them.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -617,12 +660,16 @@ CREATE TABLE tx_retrieval_queue (
 /// spent soon after it is received in a purely transparent transaction, which the wallet
 /// currently has no means of detecting otherwise.
 pub(super) const TABLE_TRANSPARENT_SPEND_SEARCH_QUEUE: &str = r#"
-CREATE TABLE transparent_spend_search_queue (
+CREATE TABLE "transparent_spend_search_queue" (
     address TEXT NOT NULL,
-    transaction_id INTEGER NOT NULL,
+    transaction_id INTEGER NOT NULL
+        REFERENCES transactions(id_tx) ON DELETE CASCADE,
```
str4d, line 667: Similar line of questioning applies here as to `tx_retrieval_queue`. It might be that one or both of these tables becomes obsoleted by a more general map table at some point in the future, and that it's fine to keep them as-is for now, but I do want to nail down whether or not we should care about losing this discovery path in a race condition.

zcash_client_sqlite/src/wallet/db.rs
str4d, line 244: Add a new paragraph documenting the conditions under which transactions are deleted from this table. I believe it to specifically be when no account would remain in the wallet that the transaction can be counted as "belonging to".

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -360,25 +365,29 @@ CREATE INDEX "sapling_received_notes_tx" ON "sapling_received_notes" (
...
-        ON DELETE CASCADE,
-    FOREIGN KEY (transaction_id)
-        -- We do not delete transactions, so this does not cascade
-        REFERENCES transactions(id_tx),
+pub(super) const TABLE_SAPLING_RECEIVED_NOTE_SPENDS: &str = r#"
+CREATE TABLE "sapling_received_note_spends" (
+    sapling_received_note_id INTEGER NOT NULL
+        REFERENCES sapling_received_notes(id) ON DELETE CASCADE,
+    transaction_id INTEGER NOT NULL
+        REFERENCES transactions(id_tx) ON DELETE CASCADE,
```
str4d, line 377: If this edge is triggering a cascade, it implies that a transaction is being deleted because (assuming my belief on the conditions of deletion is correct) no account will exist in the wallet that this transaction belongs to. This further implies that the account spending a note in this transaction is also being deleted, and thus we would delete this note-spend via the `sapling_received_note_id` edge cascading if the account were deleted prior to the transaction instead of after.

The corollary is that as long as any account exists that funds a transaction, it must not be deleted, or else this cascade edge would trigger and cause data corruption. Given that I agree we need to delete transactions prior to deleting accounts (in order to correctly clean up all information related to the account), the only way to avoid this cascade would be to make this edge nullable. But that is the wrong data model, so I'm content with this just being a documented restriction on how rows are deleted from the `transactions` table per my prior comment.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -360,25 +365,29 @@ CREATE INDEX "sapling_received_notes_tx" ON "sapling_received_notes" (
...
-    FOREIGN KEY (sapling_received_note_id)
-        REFERENCES sapling_received_notes(id)
-        ON DELETE CASCADE,
-    FOREIGN KEY (transaction_id)
-        -- We do not delete transactions, so this does not cascade
-        REFERENCES transactions(id_tx),
+pub(super) const TABLE_SAPLING_RECEIVED_NOTE_SPENDS: &str = r#"
+CREATE TABLE "sapling_received_note_spends" (
+    sapling_received_note_id INTEGER NOT NULL
+        REFERENCES sapling_received_notes(id) ON DELETE CASCADE,
```
str4d, line 375: By contrast, it's okay for this edge to cascade at any time, because notes in `*_received_notes` are always associated with a single account (due to us ensuring that IVKs are always non-overlapping with any existing account at account creation), and so when an account is deleted we always want to remove all spent-in data. Transactions that were pointed to by removed entries but left after their deletion are presumed to be cross-account transactions, assuming the initial tx deletion pass was correctly filtering.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -683,6 +683,55 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     Ok(account)
 }
 
+pub(crate) fn delete_account(
+    conn: &rusqlite::Transaction,
+    account_uuid: AccountUuid,
+) -> Result<(), SqliteClientError> {
+    conn.execute(
```
str4d, line 729:
```suggestion
    // Delete all transaction information that is solely linked to this account. This
    // effectively reverts the wallet state for this account to the time before its
    // viewing keys and transparent addresses had been used to scan for information.
    conn.execute(
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -683,6 +683,55 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+            EXCEPT
+            SELECT transaction_id FROM non_account_transactions
+        )
+        "#,
+        named_params![
+            ":account_uuid": account_uuid.0,
+        ],
+    )?;
+
+    conn.execute(
```
str4d, line 766:
```suggestion
    // At this point, the only information remaining about the account is its entry in the
    // accounts table and its addresses; delete them.
    conn.execute(
```

Also, between these two statements, per my earlier bug comment, we'd need to relink any `sent_notes` entries that reference this account via `to_account_id` (which as a side-effect means the comment I suggest here is accurate).

zcash_client_sqlite/src/wallet/db.rs
str4d: There should also be a test for this (add two accounts to a wallet, send funds from the first to the second, then delete the second account).

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3226,6 +3233,109 @@ pub fn change_note_spends_succeed<T: ShieldedPoolTester>(
...
+
+    assert_eq!(st.get_total_balance(account2), amount_sent);
+    assert_eq!(st.get_total_balance(account1), amount_left);
+
+    // txid0 should exist; we haven't enhanced it so we'll have the mined height, but not the raw
+    // transaction data.
+    assert_matches!(st.wallet_mut().get_tx_height(txid0), Ok(Some(_)));
+
+    // delete account 1
+    assert_matches!(st.wallet_mut().delete_account(account1), Ok(_));
```
str4d, line 3319: Re: my bug comment, this tests deleting the source account in a cross-account transfer, but we need a test of deleting the destination account as well.

str4d (changes requested):
Reviewed 1ba912344779223767a0620fc05a542d8ac5a4a8.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: In this case, we may have to set the `to_address` column to some "unknown" sentinel value; if we set a random OVK, we will not be able to obtain the recipient address by trial-decryption and if it was a wallet-internal shielded address we will not have ever generated that address string, so we have nowhere to obtain it from.

In current use of the wallet APIs, I believe that `to_account_id` will always equal `from_account_id` in the cases where it is set, so the cascade is fine. In that case, another option would be to add a check constraint that is `((to_account_id IS NULL) OR (to_account_id = from_account_id))`.

zcash_client_backend/src/data_api.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/2023 for the latter; it's not something that we're going to block this release on. In general, we will eventually want to provide wallet backup functionality, and this will then be subsumed by that; we don't have time to do it now.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: On further consideration, I think that we have not been adequately handling the case where the wallet does not intend to retain information about a transaction recipient, as indicated by the use of a null OVK. Instead, I think that we should remove the `ck_send_note_recipient` check, and use `ON DELETE SET NULL`. We should still however scan the transaction using _only_ the sending account's OVK and update the `to_address` column if we're able to recover the address using that.

nuttycom: When constructing a cross-account transaction, we should explicitly use `OvkPolicy::None` when the recipient address is an internal address of the receiving account.

nuttycom: Opened https://github.com/zcash/librustzcash/issues/2024

zcash_client_sqlite/src/wallet/common.rs
nuttycom: I have now reviewed every use of the database connection in every migration, and confirmed that none of the uses of that database connection in calls to "live" `zcash_client_sqlite` methods touch any of the affected columns.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: I agree that this isn't a problem for right now, and that in general for `CompactBlock`-based transparent scanning that we will have to account for race conditions in gap limit discovery. This is something where out-of-order scanning will cause us problems, because transparent outputs sent to addresses beyond the original gap will not be detected. What we will probably need to do is, specifically for wallet recovery, add the functionality for background UTXO discovery that we had originally planned for ephemeral addresses; I don't see another way at the moment, but since it's in a recovery context I think that delays are probably tolerable.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: This is now much less relevant given that we no longer rely exclusively on this queue for txid discovery.

zcash_client_sqlite/src/wallet/init.rs
str4d: The helper closure was introduced, but it is missing documentation (which I want because the `re_parens` regex is very difficult to parse visually, and it's not clear why this is needed and that it's not just stripping out all the parens and potentially introducing bugs in change detection).

zcash_client_backend/src/data_api.rs
```diff
@@ -2878,6 +2878,17 @@ pub trait WalletWrite: WalletRead {
 
     /// Deletes the specified account, and all transactions that exclusively involve it, from the
     /// wallet database.
+    ///
+    /// WARNING: This is a destructive operation and may result in the permanent loss of
+    /// potentially important information that is not recoverable from chain data, including:
+    /// * Data about transactions sent by the account for which a random OVK was used;
```
```suggestion
    /// * Data about transactions sent by the account for which [`OvkPolicy::Discard`] (or
    ///   [`OvkPolicy::Custom`] with random OVKs) was used;
```

zcash_client_backend/src/data_api.rs
```diff
@@ -2878,6 +2878,17 @@ pub trait WalletWrite: WalletRead {
...
+    /// * Data related to transactions that the account attempted to send that expired or were
+    ///   otherwise invalidated without having been mined in the main chain;
+    /// * Data related to transactions that were observed in the mempool as having inputs or
+    ///   outputs that involved the account, but that were never mined in the main chain;
+    /// * Data related to transactions that were received by the wallet in a mined block, where
+    ///   that block was later un-mined in a chain reorg and the transaction was either invalidated
+    ///   or was never re-mined.
```
str4d, line 2892:
```suggestion
    ///   or was never re-mined.
    ///
    /// [`OvkPolicy::Discard`]: crate::wallet::OvkPolicy::Discard
    /// [`OvkPolicy::Custom`]: crate::wallet::OvkPolicy::Custom
```

str4d (approved):
utACK 9852ee86cb3d79a49dbac955402ded1b8a589e69 with documentation requests (below and in https://github.com/zcash/librustzcash/pull/2012#discussion_r2482737960).

PR #2013, opened 2025-10-25
Update `zcash_client_*` dependencies

str4d:
These are dependencies which don't require updates to our low-level crates, and (in the case of `arti-client`) to versions that don't require an MSRV bump.


nuttycom (dismissed):
utACK 5ad1620d0732491d77ab1d28334494e85c3ca5b6

PR #2018, opened 2025-10-27
zcash_client_sqlite: Return correct gap metadata from `reserve_next_n_addresses`

nuttycom:


zcash_client_backend/src/wallet.rs
```diff
@@ -761,6 +761,22 @@ impl TransparentAddressMetadata {
         self.exposure
     }
 
+    /// Returns a copy of this metadata, with its exposure metadata updated
+    pub fn with_exposure_at(
```
str4d, line 765: Document this addition in the changelog.

str4d (dismissed):
utACK 288bdbeec9a3e10f700b5fb84deff90e1b806702 with changelog fix.

zcash_client_backend/src/wallet.rs
nuttycom: I had misremembered `TransparentAddressMetadata` as being a new type in this update, but no, it just changed shape.

PR #2022, opened 2025-10-29
CI: Fix usage of `./.github/actions/prepare`

str4d:
The default value for an input only applies if the input is completely unset. If we are setting it from a matrix strategy field, we need to manually fill in the default if the matrix strategy doesn't set the field.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -5,6 +5,14 @@ inputs:
     description: 'Extra feature flags to enable'
     required: false
     default: ''
+  transparent-pool:
+    description: 'Enable all transparent pool-specific feature flags'
+    required: false
+    default: false
+  orchard-pool:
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/data_api/testing/pool.rs
nuttycom, line 1: This is also fixed, slightly differently, in https://github.com/zcash/librustzcash/pull/2012/files#diff-1e9367e5fa59a041939850dc9e410a4b3f563ce75b10e2c9ae8be45b67dc19ed

nuttycom (approved):
utACK 8485c92d4d447d718221b3506485316f65a0dacc; I will rebase #2012 after this merges.

nuttycom (approved):
re-utACK 2500475262d4865008ab766984c94e47578367bb

PR #2029, opened 2025-10-31
zcash_client_sqlite: Extend ephemeral transparent address gap limit to 10 addresses.

nuttycom:
Builds atop #2028

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -326,7 +326,7 @@ impl GapLimits {
 ///
 /// - external addresses: 10
 /// - transparent internal (change) addresses: 5
-/// - ephemeral addresses: 5
+/// - ephemeral addresses: 10
 ///
 /// These limits are chosen with the following rationale:
 /// - At present, many wallets query light wallet servers with a set of addresses, because querying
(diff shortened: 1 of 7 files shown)


zcash_client_sqlite/src/lib.rs
```diff
@@ -351,7 +351,7 @@ impl Default for GapLimits {
         Self {
             external: 10,
             internal: 5,
-            ephemeral: 5,
+            ephemeral: 10,
```
str4d, line 353: Bug: the documentation for the defaults was not updated, either to declare the new default or to document the rationale for this new limit.

zcash_client_sqlite/src/lib.rs
```diff
@@ -3023,7 +3023,7 @@ mod tests {
                             // This is an imprecise check; the objective of the randomized time
                             // selection is that all ephemeral address checks be performed within a
                             // day, and that their check times be distinct.
-                            let result = t_delta < day && !check_times.contains(&t);
+                            let result = t_delta < 2 * day && !check_times.contains(&t);
```
str4d, line 3030: Bug: The comment no longer matches the code, and it's not clear to me that this code change *should* be correct. The alternative would be to adjust scheduling times to account for the doubled gap limit while still targeting one day for checking all ephemeral addresses. And I feel like that is the more correct approach, because in the limit of many ephemeral addresses being used, this change to the gap limit becomes an increasingly smaller contribution to the overall set size, and thus should have increasingly less effect (but this test change suggests otherwise).

str4d (changes requested):
Reviewed 570d19eb32ac894a740aab91a09af9c4b9ca2a50.

zcash_client_sqlite/src/lib.rs
str4d: This change should also be documented in the changelog.

zcash_client_sqlite/src/lib.rs
nuttycom: Not a bug. What happened here is that, with a larger sample, we (because randomness) ended up with a delta that's larger than a day.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/570d19eb32ac894a740aab91a09af9c4b9ca2a50..cfe3b444b4f77e89cc2af8fb9c73e4057a8d16e6) to address comments from code review.

zcash_client_sqlite/src/lib.rs
str4d: Mmk. That means there's still a bug, but it's in this test instead of the scheduling code. If the scheduler is not bounding the maximum delta then this test was incorrectly failing on those bounds being exceeded, even before this PR (and we just happened to not yet run into it). This change is misleading because the check no longer matches the comment explaining it.

The test should instead be collecting up all of the check times (and continuing to ensure they are unique), and then bounding the maximum percentage of deltas that may be greater than 1 day to the expected percentage from the scheduler distribution.

nuttycom: The alternative is to just re-roll anything that comes up with a delta that's more than a day, and return this test to its original state.

str4d: Rejection sapling would indeed also be acceptable. It's in theory an unbounded loop, but in practice you'll only go around it once.

nuttycom: Hmm, I just remembered, that isn't how this works, because we're simulating a Poisson process: https://github.com/zcash/librustzcash/blob/752db0fb7fd81ba4b92d8b91bd28fe9c5b2317e6/zcash_client_sqlite/src/wallet/transparent/ephemeral.rs#L207

We divide the day by the number of addresses to be checked, and then each check is scheduled to have an expected value of 1/n days after the previous check. So I'm not sure how to bound this; we don't want to just cut the tail off the distribution.

str4d (approved):
utACK d860a332a5627d5719e722f5a2aa578687640b49

zcash_client_sqlite/src/lib.rs
```diff
@@ -3022,8 +3021,13 @@ mod tests {
...
                             // This is an imprecise check; the objective of the randomized time
                             // selection is that all ephemeral address checks be performed within a
-                            // day, and that their check times be distinct.
-                            let result = t_delta < day && !check_times.contains(&t);
+                            // day, and that their check times be distinct. We are bounding the
+                            // overall delta to two days to limit the probability of test failure,
+                            // since due to randomization of the check intervals it's possible that
+                            // scheduled checks might exceed one day; if we find an example where
+                            // checks exceed two days then we're in some long tail of the
+                            // distribution.
```
daira, line 3029: Nit: an exponential distribution is not a "[long-tailed](https://en.wikipedia.org/wiki/Long_tail) distribution", either in the [strict statistical sense](https://en.wikipedia.org/wiki/Heavy-tailed_distribution#Definition_of_long-tailed_distribution), or in the sense of power-law-like distributions sometimes used in business. ([By definition](https://en.wikipedia.org/wiki/Heavy-tailed_distribution) an exponential distribution is not heavy-tailed, and therefore not long-tailed.) There is a much fuzzier colloquial use of "long tail" that this fits, but I'd argue that it is an incorrect use.

```suggestion
                            // checks exceed two days then we are in an exponentially unlikely tail
                            // of the distribution.
```

daira (commented):
Post-hoc ACK (with wording nit in a comment, which would not have been blocking).

PR #2031, opened 2025-11-03
zcash_client_sqlite: Relax exclusions on wallet-internal ephemeral UTXOs

nuttycom:
In zcash/librustzcash#1985, a clause was added to UTXO queries to ensure that ephemeral outputs produced by wallet-internal transactions are excluded from no such outputs are ever regarded as spendable. However, this is too strict; it inhibits recovery of funds that end up "stuck" in an ephemeral address due to the partial failure of a TEX send. This relaxes that check such that if a TXO has actually been observed to be unspent after the second transaction in a TEX chain would have expired, we consider that unspentness information to be authoritative.

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -995,21 +995,52 @@ pub(crate) fn tx_unexpired_condition_minconf_0(tx: &str) -> String {
     )
 }
 
-/// Generates a SQL condition that checks whether a transaction spends outputs controlled by an
-/// account.
+/// Generates a SQL condition that checks that if a TXO was received at an ephemeral
+/// address, it either had no inputs belonging to the wallet, or has been confirmed
+/// to be unspent after the transaction that created it would have expired.
+///
+/// TODO: This fragment is very unwieldy; it really doesn't work well as a fragment. It would be
+/// much better if it could be expressed as a view somehow; the problem is that views can't be
+/// parameterized, and there are numerous interacting checks that need to be made against the
+/// target height in the context where this fragment is being used.
 ///
 /// # Usage requirements
-/// - `tx` must be set to the alias for the `transactions` table in the enclosing scope.
-/// - `accounts` must be set to the alias for the `accounts` table in the enclosing scope.
+/// - `transparent_received_outputs` must be set to the alias to the `transparent_received_outputs`
+///   table in the enclosing scope.
+/// - `addresses` must be set to alias for the `addresses` table in the enclosing scope such that
+///   `addresses.id = transparent_received_outputs.address_id`.
+/// - `tx` must be set to the alias for the `transactions` table in the enclosing scope such that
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -995,21 +995,52 @@ pub(crate) fn tx_unexpired_condition_minconf_0(tx: &str) -> String {
...
             FROM v_received_output_spends
             WHERE v_received_output_spends.account_id = {accounts}.id
         )
+        -- or the transaction that generated the TXO would be considered expired as of the TXOs
+        -- max_observed_unspent_height; this operates under the assumption that the second
+        -- transaction in a TEX chain has the same expiry as the transaction that generated the
+        -- ephemeral output, and the output having been observed to be unspent above this height
+        -- indicates that the subsequent spend failed and the spending transaction will have
+        -- expired.
+        OR {transparent_received_outputs}.max_observed_unspent_height > {tx}.expiry_height
```
str4d, line 1043: Bug: this should be
```suggestion
        OR {transparent_received_outputs}.max_observed_unspent_height >= {tx}.expiry_height
```
because the expiry height is the last valid height the tx can be mined at, so if the tx is observed unspent as of the end of the block at that height (which is what `max_observed_unspent_height` records), then the tx cannot be mined.

Add a test that catches this bug by exercising the boundary.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1250,21 +1263,15 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
              )
-             -- and the output is unspent
-             AND u.id NOT IN ({})
-             -- excluding received txos where the address is an ephemeral address and the
-             -- transaction that produced the TXO spent outputs belonging to the same account.
-             AND NOT (
-                addresses.key_scope = :ephemeral_key_scope
-                AND {} -- the transaction that created the txo spends account-internal outputs
-             )",
+
```
str4d: Nit:
```suggestion
```

str4d (changes requested):
Reviewed 3110d0a693ea8e04794198c1fa3acdc86f9db98c

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: This isn't a bug; using `>` instead of `>=` is the more conservative choice: we only allow the TXO to be selected once it has been observed to be unspent at a height _greater_ than the expiry height.

In fact, I think your suggested change here would be a bug. Note that the sense of this whole clause has changed, I rewrote it via De Morgan's law in order to be able to include this third clause via an `OR`.

str4d: My brain was misparsing `unspent` as `unmined`, which led to my mis-analysis.

str4d (dismissed):
utACK 3110d0a693ea8e04794198c1fa3acdc86f9db98c with prior nit.

PR #2033, opened 2025-11-04
Clarify the use of OVKs and FVKs in compact block creation utilities for testing.

nuttycom:


zcash_client_backend/CHANGELOG.md
```diff
@@ -73,6 +73,12 @@ workspace.
     consequence of this change, the signature of
     `TransparentAddressMetadata::new` has changed; use
     `TransparentAddressMetadata::derived` instead.
+- `zcash_client_backend::data_api::testing`:
+  - The `TestFvk` trait has changed in several ways. It has a ne `OutgoingViewingKey`
```
str4d, line 77:
```suggestion
  - The `TestFvk` trait has changed in several ways. It has a new `OutgoingViewingKey`
```

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -2312,7 +2315,7 @@ fn fake_compact_block_spending<P: consensus::Parameters, Fvk: TestFvk>(
     height: BlockHeight,
     prev_hash: BlockHash,
     (nf, in_value): (Fvk::Nullifier, Zatoshis),
-    fvk: &Fvk,
+    recipient_fvk: &Fvk,
```
str4d: The FVK happens to receive change, but it would be a better clarification to mark this as a sender:
```suggestion
    sender_fvk: &Fvk,
```
and similarly below.

EDIT: I see that this exact rename is done in the third commit. I'd mildly prefer that the rename be moved to the first commit, given that it is ostensibly the "clarify naming" commit.

str4d (approved):
utACK d60ad5ad001b6bec23bc7ac2329035989fbc78dd with nits.

zcash_client_backend/CHANGELOG.md
nuttycom: I'll fix this in a subsequent PR.

PR #2035, opened 2025-11-06
refactor(zcash_encoding): expose a reverse-hex decoder

### LINKED ISSUE #1934: zcash_encoding: Add a parser and serializer for the Bitcoin "big-endian hex" format
Bitcoin Core's `uint256` C++ type is essentially a `[u8; 32]`. However, when printed as a string, it is serialized in the following manner:
- Reverse the bytes.
- Encode the reversed bytes as a hex string.

The reason is that while `uint256` was used for both block hashes and transaction IDs and thus contained an opaque hash output, when specifically used for block hashes it was also interpreted as a little-endian integer for proof-of-work difficulty comparisons. The byte-reversal meant that the hex string for block hashes would start with leading zeros corresponding to the amount of work.

Unfortunately, `zcashd` made liberal use of `uint256` as a `[u8; 32]` to store everything from hash outputs to field elements in Sprout, and printed these out in JSON-RPC using `uint256::GetHex`. This led to the precedent that 32-byte hex values in JSON-RPC outputs were always reversed, but non-32-byte hex values were not. We maintained this for Sapling under duress, and then decided to just not propagate the problem with Orchard.

However, it means that there are many legacy JSON-RPC fields that need a way to parse this format. Currently I tend to do this by misusing `TxId`, but that just leads to confusion.

We should instead extract the parser and serializer from `TxId` into a helper in `zcash_encoding`, that we can then use everywhere it is needed.

distractedm1nd:
Intended to resolve #1934.

Draft, because I have a question that blocks: Is it fine to just expose a single function? The issue says to extract the parser and serializer. Is this because there is a reason to actively reverse the hex? Anywhere for compatibility?

### DIFF
--- a/components/zcash_encoding/src/lib.rs
+++ b/components/zcash_encoding/src/lib.rs
@@ -316,6 +316,9 @@ impl ReverseHex {
 
     /// Decodes a reversed-hex string into a byte array by decoding the hex and
     /// reversing the order of the bytes.
+    ///
+    /// Returns `None` if the hex decoding fails or if the output is the wrong
+    /// size.
     pub fn decode(hex_str: &str) -> Option<[u8; 32]> {
         let mut bytes = [0; 32];
         let res = hex::decode_to_slice(hex_str, &mut bytes);
--- a/components/zcash_encoding/CHANGELOG.md
+++ b/components/zcash_encoding/CHANGELOG.md
@@ -6,6 +6,8 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `zcash_encoding::ReverseHex::{encode, decode}`
 
 ## [0.3.0] - 2025-02-21
 ### Changed


components/zcash_encoding/src/lib.rs
```diff
@@ -300,6 +300,21 @@ impl Optional {
...
 
+/// This handles the legacy format from Bitcoin Core's `uint256::GetHex`, which reverses
+/// bytes before hex encoding. `zcashd` applied this to all 32-byte JSON-RPC values
+/// (transaction IDs, block hashes, Sprout field elements, etc.), though non-32-byte
+/// hex values are not reversed.
+pub struct ReverseHex;
+
+impl ReverseHex {
+    /// Decodes a reversed-hex string into a byte array.
+    pub fn decode(bytes: &[u8; 32]) -> alloc::string::String {
```
nuttycom: This implements the encoding function, not the decoding function (which is String -> Option<[u8; 32]>)
```suggestion
    /// Encodes a byte array by reversing the order of the bytes and then hex encoding.
    pub fn decode(bytes: &[u8; 32]) -> alloc::string::String {
```

nuttycom (changes requested):

components/zcash_encoding/src/lib.rs
distractedm1nd: [cf39300](https://github.com/zcash/librustzcash/pull/2035/commits/cf39300ffa47aca9f2dec1f55727d17a5301f2fe)

components/zcash_encoding/src/lib.rs
```diff
@@ -300,10 +300,48 @@ impl Optional {
...
+        let mut bytes = [0; 32];
+        let res = hex::decode_to_slice(hex_str, &mut bytes);
+        if res.is_err() {
+            return None;
+        }
+
+        bytes.reverse();
+        Some(bytes)
+    }
+}
```
nuttycom, line 332: Returning `Option` here instead of `Result` conflates multiple different possible types of errors, so it's potentially not as useful as it could be in reporting what went wrong to the caller, but so long as the `None` cases are documented I think this is fine.

In addition, please document the addition of these functions in the `CHANGELOG.md` file for `zcash_encoding.`

nuttycom (changes requested):
One minor documentation improvement & CHANGELOG additions required, otherwise utACK on the code changes.

components/zcash_encoding/src/lib.rs
distractedm1nd: Yeah, I mentioned the ergonomics of the Option being a little weird in the zallet PR, but I also think it's fine. The alternative would be introducing custom errors to this crate IIUC, which also doesn't feel right when this is the only case.

Updated changelog + rustdoc, hope formatting is okay 

[b664dbd](https://github.com/zcash/librustzcash/pull/2035/commits/b664dbd5693d7fe12dbc8f51d40ed2c42416d571)

nuttycom (approved):
utACK b664dbd5693d7fe12dbc8f51d40ed2c42416d571

PR #2036, opened 2025-11-07
zcash_client_backend: Add low-level data API

### LINKED ISSUE #2114: zcash_client_backend: Extract logic of `put_blocks` and `decrypt_and_store_transaction` from `zcash_client_sqlite`
Over the past few years, a lot of functionality has accumulated in zcash_client_sqlite that is not directly coupled to storing data in the SQL database. This PR migrates important business logic into zcash_client_backend, with two objectives: first, to make the logic reusable across backend implementations, and second, to expose additional functionality in zcash_client_backend that is necessary for the purpose of implementing handling for transparent transaction information in scan_cached_blocks.

nuttycom:
Over the past few years, a lot of functionality has accumulated in `zcash_client_sqlite` that is not directly coupled to storing data in the SQL database. This PR migrates important business logic into `zcash_client_backend`, with two objectives: first, to make the logic reusable across backend implementations, and second, to expose additional functionality in `zcash_client_backend` that is necessary for the purpose of implementing handling for transparent transaction information in `scan_cached_blocks`.

Best reviewed commit-by-commit.

Fixes #2114.
Fixes COR-360.

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -21,7 +21,7 @@ workspace.
 - `impl zcash_client_backend::data_api::ll::LowLevelWalletRead for WalletDb`
 - `impl zcash_client_backend::data_api::ll::LowLevelWalletWrite for WalletDb`
 - `impl Hash for zcash_client_sqlite::{ReceivedNoteId, UtxoId, TxRef}`
-- `impl {PartialOrd, Ord} for zcash_client_sqlite::{UtxoId, PartialOrd, Ord}`
+- `impl {PartialOrd, Ord} for zcash_client_sqlite::UtxoId`
 
 ### Changed
 - Migrated to `orchard 0.12`, `sapling-crypto 0.6`.


zcash_client_backend/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_client_backend::wallet::transparent`
```
```suggestion
- `zcash_client_backend::wallet::transparent` module, behind the
  `transparent-inputs` feature flag.
```

zcash_client_backend/src/wallet/transparent.rs
str4d, line 1: Document this module.

zcash_primitives/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_primitives::transaction::Transaction::data`
```
str4d: There is no change to the `zcash_primitives` crate in 34c22898e13e07820dcda4f9ffc63e5144f900cf where this changelog entry is introduced. Move it to the correct commit.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -700,10 +700,16 @@ impl Transaction {
         Transaction { txid, data }
     }
 
+    /// Consumes this [`Transaction`] and returns the wrapped [`TransactionData`] value.
     pub fn into_data(self) -> TransactionData<Authorized> {
         self.data
     }
 
+    /// Accessor for the wrapped [`TransactionData`] value.
+    pub fn data(&self) -> &TransactionData<Authorized> {
```
str4d: What motivates this accessor? We already have a `Deref` impl which makes everything about `TransactionData` directly accessible from `Transaction`, without going through a getter like this.

`Deref` impls are a widely-used feature of Rust; the most common examples of them are on [`String`](https://doc.rust-lang.org/stable/std/string/struct.String.html#deref-methods-str) (sharing methods with `&str`) and [`Vec`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#deref-methods-%5BT%5D) (sharing methods with slices).

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+            .into_iter()
+            .flat_map(|bundle| bundle.actions().iter().map(|action| action.nullifier()))
+    }
+
+    fn fee_paid<E, F>(&self, get_prevout: F) -> Result<Option<Zatoshis>, E>
+    where
+        E: From<BalanceError>,
+        F: FnMut(&OutPoint) -> Result<Option<Zatoshis>, E>,
+    {
+        self.data().fee_paid(get_prevout)
```
str4d: Per the rustdocs, [`Transaction::fee_paid`](https://docs.rs/zcash_primitives/0.26.1/zcash_primitives/transaction/struct.Transaction.html#method.fee_paid) already exists via the `Deref` (like you are already using for `self.orchard_bundle()` etc. above):
```suggestion
        self.fee_paid(get_prevout)
```

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
```diff
@@ -741,7 +737,7 @@ mod tests {
                     // Create subtrees from the note commitments in parallel.
                     const CHUNK_SIZE: usize = 1024;
                     let subtrees = sapling_commitments
-                        .par_chunks_mut(CHUNK_SIZE)
+                        .chunks_mut(CHUNK_SIZE)
```
str4d, line 740: This is a performance regression in the migration, and I see no reason for removing this. Undo this change.

zcash_client_sqlite/src/lib.rs
```diff
@@ -140,18 +133,6 @@ use crate::wallet::encoding::KeyScope;
...
-trait ParallelSliceMut<T> {
-    fn par_chunks_mut(&mut self, chunk_size: usize) -> std::slice::ChunksMut<'_, T>;
-}
-#[cfg(not(feature = "multicore"))]
-impl<T> ParallelSliceMut<T> for [T] {
-    fn par_chunks_mut(&mut self, chunk_size: usize) -> std::slice::ChunksMut<'_, T> {
-        self.chunks_mut(chunk_size)
-    }
-}
-
```
str4d, line 151: Undo this deletion (to enable undoing the migration change).

zcash_client_backend/src/data_api.rs
```diff
@@ -2332,19 +2332,19 @@ impl<A> ScannedBlock<A> {
 ///
 /// The purpose of this struct is to permit atomic updates of the
 /// wallet database when transactions are successfully decrypted.
-pub struct DecryptedTransaction<'a, AccountId> {
+pub struct DecryptedTransaction<'a, Tx, AccountId> {
```
str4d, line 2346: Document this change in the changelog.

It is also not clear to me *why* this change was made, as it doesn't appear to be used anywhere in this PR.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// Returns an iterator over the nullifiers of Sapling notes spent in this transaction.
+    fn sapling_spent_note_nullifiers(&self) -> impl Iterator<Item = &::sapling::Nullifier>;
+
+    /// Returns an iterator over the nullifiers of Orchard notes spent in this transaction.
+    #[cfg(feature = "orchard")]
+    fn orchard_spent_note_nullifiers(&self) -> impl Iterator<Item = &::orchard::note::Nullifier>;
+
+    /// Returns the fee paid by this transaction, given a function that can retrieve the value of
+    /// prior transparent outputs spent in the transaction. Returns `None` if insufficient
+    /// information is available for computing the fee.
```
str4d:
```suggestion
    /// Returns the fee paid by this transaction, given a function that can retrieve the value of
    /// prior transparent outputs spent in the transaction.
    ///
    /// Returns `None` if insufficient information is available for computing the fee.
```

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+        funding_accounts.extend(self.detect_accounts_orchard(tx.orchard_spent_note_nullifiers())?);
+
+        Ok(funding_accounts)
+    }
+
+    /// Returns the most likely wallet address that corresponds to the protocol-level receiver of a
+    /// note or UTXO.
+    ///
+    /// If the wallet database has stored a wallet address that contains the given receiver, then
+    /// that address is returned; otherwise, a new address containing that receiver is generated.
```
str4d:
```suggestion
    /// that address is returned; otherwise, the most likely address containing that receiver is
    /// generated.
```
(by definition it cannot be a new address, because the receiver has been observed outside the wallet). Also it might be worth clarifying that the generated address is not stored (just to avoid misunderstandings about the word "generated"); this is implied by the method taking `&self` and being on the `LowLevelWalletRead` trait, but not guaranteed.

Also, document what a return value of `Ok(None)` means (I presume it is either "this receiver doesn't belong to this account" for shielded receivers following standard diversifier index derivation, or "we cannot determine whether or not this receiver belongs to this account" for all other receivers).

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+    fn select_receiving_address(
+        &self,
+        account: Self::AccountId,
+        receiver: &Receiver,
+    ) -> Result<Option<ZcashAddress>, Self::Error>;
+
+    /// Detects and returns the identifier for the account that generated the given address, if the
+    /// address belongs to an account in the wallet.
```
str4d: "generated" is the wrong word here, given that standalone transparent addresses are now supported in backends. Maybe "the account to which the given address belongs, if any"?

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+        &self,
+        spends: impl Iterator<Item = &'a orchard::note::Nullifier>,
+    ) -> Result<HashSet<Self::AccountId>, Self::Error>;
+
+    /// Get information about a transparent output controlled by the wallet.
+    ///
+    /// # Parameters
+    /// - `outpoint`: The identifier for the output to be retrieved.
+    /// - `spendable_as_of`: The target height of a transaction under construction that will spend the
+    ///   returned output. If this is `None`, no spendability checks are performed.
```
str4d: This is a confusing parameter name to me. The backend wallet will have its own view of the chain tip, and thus of whether the outpoint is spendable or not; the target of the `spendable` adjective in the name is unclear. If you don't like `target_height` for this, then maybe `check_if_spendable_as_of` to clarify it?

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+        outpoint: &OutPoint,
+        spendable_as_of: Option<TargetHeight>,
+    ) -> Result<Option<WalletUtxo>, Self::Error>;
+
+    /// Returns the vector of transactions in the wallet that spend the outputs of the referenced
+    /// transaction.
+    ///
+    /// This is used as part of [`wallet::store_decrypted_tx`] to allow downstream transactions'
+    /// fee amounts to be updated once the value of all their inputs are known.
+    fn get_spending_transactions(
```
str4d: Document whether or not this API should return conflicted transactions that attempt to spend common outputs (particularly if none of them are mined or expired, which could occur in certain combinations of reorgs). I suspect we always need to return conflicts if none are mined, but maybe if one is mined we can exclude the rest? What is the most consistent response that trait implementers can follow?

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+        &self,
+        tx_ref: Self::TxRef,
+    ) -> Result<Vec<(Self::TxRef, Transaction)>, Self::Error>;
```
str4d, line 243: Are we always guaranteed to know the `Transaction`? I don't think so, e.g. if we observe via compact blocks a receive followed by a spend (either in a recover-from-seed or parallel wallet) before doing any enhancing?

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,554 @@
...
+}
+
+/// A capability trait that provides low-level wallet write operations. These operations are used
+/// to provide standard implementations for certain [`WalletWrite`] trait methods.
+///
+/// [`WalletWrite`]: super::WalletWrite
+pub trait LowLevelWalletWrite: LowLevelWalletRead {
```
str4d, line 265: I got to here in the fourth commit (bbde5ca438f66837efce83dee086f1a1db727fbd).

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
nuttycom: `par_chunks_mut` is unavailable as the wrapper & helper trait that enabled it was moved out of this crate.

zcash_client_backend/src/data_api/ll.rs
nuttycom: I started out with this approach, but I consistently get this error and haven't been able to figure out a way around it: 
```
warning: function cannot return without recursing
  --> zcash_client_backend/src/data_api/ll.rs:90:5
   |
90 | /     fn fee_paid<E, F>(&self, get_prevout: F) -> Result<Option<Zatoshis>, E>
91 | |     where
92 | |         E: From<BalanceError>,
93 | |         F: FnMut(&OutPoint) -> Result<Option<Zatoshis>, E>,
   | |___________________________________________________________^ cannot return without recursing
94 |       {
95 |           (*self).fee_paid(get_prevout)
   |           ----------------------------- recursive call site
   |
   = help: a `loop` may express intention better if this is on purpose
   = note: `#[warn(unconditional_recursion)]` on by default
```

nuttycom: Fixed via `TransactionData::fee_paid(self, ...)`

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
nuttycom: Also note that this is in test code; I don't think it's worth reintroducing the wrapper just for a test?

zcash_client_sqlite/src/lib.rs
nuttycom: https://github.com/zcash/librustzcash/pull/2036#discussion_r2599827088

zcash_client_backend/src/data_api.rs
nuttycom: This is preparatory work for using `DecryptedTransaction` with `CompactBlock` scanning; at present, it's only used by `decrypt_and_store` which gets full transaction data, but the sync process will use it in a follow-up PR.

zcash_client_backend/src/data_api/ll.rs
nuttycom: I guess `target_height` is fine, we use that often enough.

zcash_client_backend/src/data_api/ll.rs
nuttycom: Hm. This is a bit challenging, because all that we need is to return something that satisfies the `TxMeta` constraint, but that's not a concrete type. Is this an instance where it's actually appropriate to have the return type be `Box<dyn TxMeta>` or something of the sort?

zcash_client_backend/src/data_api/ll.rs
str4d: Ah yep, that is caused by the introduction of the `TxMeta` trait. You may want to consider renaming `TxMeta::fee_paid` to avoid this ambiguity, but the approach you've taken is the way to resolve the ambiguity when it occurs.

zcash_client_sqlite/src/wallet/init/migrations/receiving_key_scopes.rs
str4d: Hmm, I thought I saw this in the migration code. If it's only in tests then I care much less.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,562 @@
...
+        funding_accounts.extend(self.detect_accounts_orchard(tx.orchard_spent_note_nullifiers())?);
+
+        Ok(funding_accounts)
+    }
+
+    /// Returns the most likely wallet address that corresponds to the protocol-level receiver of a
+    /// note or UTXO.
+    ///
+    /// If the wallet database has stored a wallet address that contains the given receiver, then
+    /// that address is returned; otherwise, a the most likely address containing that receiver
```
str4d:
```suggestion
    /// that address is returned; otherwise, the most likely address containing that receiver
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4065,6 +4065,46 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
+         -- find transactions that spend transparent outputs of the decrypted tx
+         LEFT OUTER JOIN transparent_received_output_spends ts
+            ON ts.transaction_id = t.id_tx
+         LEFT OUTER JOIN transparent_received_outputs tro
+            ON tro.transaction_id = :transaction_id
+            AND tro.id = ts.transparent_received_output_id
+         WHERE t.fee IS NULL
+         AND t.raw IS NOT NULL
+         AND ts.transaction_id IS NOT NULL",
+    )?;
```
str4d, line 3469: Bug: this only selects transactions that spend the **transparent** outputs of the referenced transaction. But `LowLevelWalletRead::get_spending_transactions` is documented as selecting transactions that spend the **outputs** of the referenced transaction, which means all outputs. One of the following is true:
- This code is wrong.
- The trait method name is wrong, and should be a transparent-specific name.
- The trait method is missing a pool selector argument.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4065,6 +4065,46 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
     Ok(())
 }
 
+pub(crate) fn get_spending_transactions<P: consensus::Parameters>(
+    conn: &rusqlite::Connection,
+    params: &P,
+    tx_ref: TxRef,
+) -> Result<Vec<(TxRef, Transaction)>, SqliteClientError> {
+    // For each transaction that spends a transparent output of this transaction and does not
+    // already have a known fee value, set the fee if possible.
```
str4d: This SQL select query has nothing to do with fee-setting. Is this copy-pasta from somewhere else in `zcash_client_sqlite`?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4065,6 +4065,46 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
+    let mut spending_txs_stmt = conn.prepare(
+        "SELECT DISTINCT t.id_tx, t.raw, t.mined_height, t.expiry_height
+         FROM transactions t
+         -- find transactions that spend transparent outputs of the decrypted tx
+         LEFT OUTER JOIN transparent_received_output_spends ts
+            ON ts.transaction_id = t.id_tx
+         LEFT OUTER JOIN transparent_received_outputs tro
+            ON tro.transaction_id = :transaction_id
+            AND tro.id = ts.transparent_received_output_id
+         WHERE t.fee IS NULL
```
str4d, line 3466: Related to the above, is this fee filter supposed to be here?

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,562 @@
...
+    /// properties of that output.
+    ///
+    /// - If `recipient` is a Unified address, `output_index` is an index into the outputs of the
+    ///   transaction within the bundle associated with the recipient's output pool.
+    /// - If `recipient` is a Sapling address, `output_index` is an index into the Sapling outputs of
+    ///   the transaction.
+    /// - If `recipient` is a transparent address, `output_index` is an index into the transparent
+    ///   outputs of the transaction.
+    /// - If `recipient` is an internal account, `output_index` is an index into the outputs of
+    ///   the transaction in the transaction bundle corresponding to the recipient pool.
```
str4d: This documentation is misleading. It doesn't matter whether `recipient` is a UA or not; the `Recipient` enum is split along different boundaries (namely `External`, `EphemeralTransparent`, and `InternalAccount`); all of these contain explicit pool information that we therefore don't need to duplicate in the method arguments.

Below is what I originally wrote before going and checking the `Recipient` enum type (which invalidated my original line of reasoning which was based on this documentation):

> This is not precise enough. I presume you intended either "the preferred output pool", or "the output pool into which the recipient received an output at `output_index` within this transaction".
>
> But what happens if a transaction contains multiple outputs that are decryptable with different receivers within the same UA (such as in a multi-pool transaction created by ezcash)?

Instead, this should say that `output_index` is an index into the outputs of the bundle indicated by `recipient`. Maybe add a helper method to `Recipient` that returns `PoolType`?

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,562 @@
...
+        tx_ref: Self::TxRef,
+        output_index: usize,
+        recipient: &Recipient<Self::AccountId>,
+        value: Zatoshis,
+        memo: Option<&MemoBytes>,
+    ) -> Result<(), Self::Error>;
+
+    /// Updates the wallet's view of a transaction to indicate the miner's fee paid by the
+    /// transaction.
+    fn update_tx_fee(&mut self, tx_ref: Self::TxRef, fee: Zatoshis) -> Result<(), Self::Error>;
```
str4d, line 446: I think this should be documented as (and the `zcs` impl updated so it is) returning an error if the fee is already set and the new value doesn't match the existing value, because if that occurs it indicates either database corruption, or a fee calculation bug.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,562 @@
...
+
+    /// Adds a transparent output observed by the wallet to the data store, or updates any existing
+    /// record for that output.
+    ///
+    /// # Parameters
+    /// - `output`: The output data.
+    /// - `observation_height`: The chain tip height at the time that the update is made. Note that
+    ///   this is likely to differ from the height at which the output was mined (if any.)
```
str4d:
```suggestion
    ///   this is likely to differ from the height at which the output was mined (if any).
```

str4d: AFAICT this will end up directly used to set `min_observed_height` in the wallet DB. That column is documented as "the mempool height at the time that the wallet observed the transaction, or the mined height of the transaction, whichever is less" (which makes sense, as the mined height *is* the last mempool height the tx was in the mempool at). In the current `zcs` impl it is correctly pulled from mined-height and mempool heights everywhere *except* `put_received_transparent_utxo` which uses the chain tip height.

Either fix this documentation to refer to "mempool height", or fix the `zcs` impl to calculate the mempool height from this provided height. The pre-existing bug in `put_received_transparent_utxo` should also be fixed.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,562 @@
...
+    #[cfg(feature = "transparent-inputs")]
+    fn generate_transparent_gap_addresses(
+        &mut self,
+        account_id: Self::AccountId,
+        key_scope: TransparentKeyScope,
+        request: UnifiedAddressRequest,
+    ) -> Result<(), Self::Error>;
+
+    /// Adds a [`TransactionDataRequest::TransactionsInvolvingAddress`] request to the transaction
+    /// data request queue. When the transparetn output of `tx_ref` at output index `output_index`
```
str4d:
```suggestion
    /// data request queue. When the transparent output of `tx_ref` at output index `output_index`
```

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,562 @@
...
+    /// [`TransactionDataRequest::TransactionsInvolvingAddress`]: super::TransactionDataRequest
+    #[cfg(feature = "transparent-inputs")]
+    fn queue_transparent_spend_detection(
+        &mut self,
+        receiving_address: TransparentAddress,
+        tx_ref: Self::TxRef,
+        output_index: u32,
+    ) -> Result<(), Self::Error>;
+
+    /// Adds [`TransactionDataRequest::Enhancement`] requests  for transactions that generated the
```
str4d:
```suggestion
    /// Adds [`TransactionDataRequest::Enhancement`] requests for transactions that generated the
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3480,21 +3448,21 @@ pub(crate) fn put_block(
...
+    from_account_uuid: AccountId,
     output_index: usize,
-    recipient: Recipient<AccountUuid>,
+    recipient: Recipient<AccountId>,
     value: Zatoshis,
 }
 
 #[derive(Debug)]
-struct WalletTransparentOutputs {
+pub struct WalletTransparentOutputs<AccountId> {
```
str4d: Document this struct (in its new home in `zcb`).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3610,38 +3582,46 @@ fn detect_wallet_transparent_outputs<P: consensus::Parameters>(
...
             // If we can’t find it, fee computation can't complete accurately
-            Ok::<_, SqliteClientError>(None)
+            Ok(None)
         }
     })
 }
 
-pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
-    conn: &rusqlite::Transaction,
+pub fn store_decrypted_tx<DbT, P>(
```
str4d: Document this function (in its new home in `zcb`).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3667,35 +3647,20 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
-
-    let observed_height = d_tx.mined_height().map_or_else(
-        || {
-            mempool_height(conn)?
-                .ok_or(SqliteClientError::ChainHeightUnknown)
-                .map(BlockHeight::from)
-        },
-        Ok,
-    )?;
+    let observed_height = d_tx.mined_height().unwrap_or(chain_tip_height + 1);
```
str4d: Indeed we are still passing the mempool height to `LowLevelWalletWrite::put_transparent_output` (below), confirming my earlier comment.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3985,63 +3897,19 @@ pub(crate) fn store_decrypted_tx<P: consensus::Parameters>(
...
-                 SET fee = :fee
-                 WHERE id_tx = :transaction_id",
-                named_params! {
-                    ":transaction_id": spending_tx_ref.0,
-                    ":fee": u64::from(fee)
-                },
-            )?;
+    for (spending_tx_ref, spending_tx) in wallet_db.get_spending_transactions(tx_ref)? {
+        if let Some(fee) = determine_fee(wallet_db, &spending_tx)? {
+            wallet_db.update_tx_fee(spending_tx_ref, fee)?;
```
str4d: Aha, indeed my earlier comments found copy-pasta; that explains the `fee` filter in the select query. I think we should be doing that filtering here, and making `get_spending_transactions` general (either specifically for transparent, or more generally and then filtering for transparent here).

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -40,6 +40,8 @@ use {
 #[cfg(feature = "orchard")]
 use crate::wallet::WalletOrchardOutput;
 
+pub mod wallet;
```
str4d, line 46: Document this addition in the changelog.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -256,6 +303,28 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
         spent_in: Option<Self::TxRef>,
     ) -> Result<(), Self::Error>;
 
+    /// Updates the backing store to indicate that the Sapling output having the given nullifier is
+    /// spent in the transaction referenced by `spent_in_tx`.
+    fn mark_sapling_note_spent(
```
str4d, line 333: - Document what the returned `bool` means.
- Document what happens if the Sapling output is already spent, specifically that the record should always be created (resulting in a conflict if neither transaction is mined), except in the case where both transactions are mined (this would be an invalid state, meaning either the caller is making a mistake or there is database corruption).

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -267,6 +336,32 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
         spent_in: Option<Self::TxRef>,
     ) -> Result<(), Self::Error>;
 
+    /// Updates the backing store to indicate that the Orchard output having the given nullifier is
+    /// spent in the transaction referenced by `spent_in_tx`.
+    #[cfg(feature = "orchard")]
+    fn mark_orchard_note_spent(
```
str4d, line 382: Same comments as for `mark_sapling_note_spent`.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -256,6 +303,28 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
...
+
+    /// Causes the given Sapling output nullifiers to be tracked by the wallet.
+    ///
+    /// When scanning the chain out-of-order, it is necessary to store the any nullifiers observed
+    /// after a gap in the scanned blocks until the blocks in that gap have been fully scanned, in
+    /// order to be able to immediately detect that a received output has already been spent. The
+    /// data store should track a mapping from each nullifier to the block, transaction ID and
+    /// input index where the nullifier was revealed, so that the state of a note receive within
+    /// that "gap" can be properly updated when it is discovered.
+    fn track_block_sapling_nullifiers(
```
str4d: This documentation is misleading. It, combined with the type signature of `nfs`, implies that the `u16` is the input index, but that doesn't make sense because an input can't have multiple nullifiers. The documentation should instead document the format of `nfs`, namely that it matches `ScannedBundles::nullifier_map` and the `u16` is the index of the transaction within the block, and the input indices are obtained from the enumeration of the `Vec<Nullifier>` (i.e. the caller must provide *all* nullifiers from the transaction, not omitting any that might have already been detected as spent).

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -267,6 +336,32 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
...
+    /// Causes the given Orchard output nullifiers to be tracked by the wallet.
+    ///
+    /// When scanning the chain out-of-order, it is necessary to store the any nullifiers observed
+    /// after a gap in the scanned blocks until the blocks in that gap have been fully scanned, in
+    /// order to be able to immediately detect that a received output has already been spent. The
+    /// data store should track a mapping from each nullifier to the block, transaction ID and
+    /// input index where the nullifier was revealed, so that the state of a note receive within
+    /// that "gap" can be properly updated when it is discovered.
+    #[cfg(feature = "orchard")]
+    fn track_block_orchard_nullifiers(
```
str4d: Same comment as for `track_block_sapling_nullifiers`.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -267,6 +336,32 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
...
+        &mut self,
+        block_height: BlockHeight,
+        nfs: &[(TxId, u16, Vec<::orchard::note::Nullifier>)],
+    ) -> Result<(), Self::Error>;
+
+    fn prune_tracked_nullifiers(&mut self, pruning_depth: u32) -> Result<(), Self::Error>;
```
str4d, line 422: Document this trait method.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -372,18 +480,20 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
...
-
     /// Deletes all [`TransactionDataRequest::Enhancement`] requests for the given transaction ID
     /// from the transaction data request queue.
     ///
     /// [`TransactionDataRequest::Enhancement`]: super::TransactionDataRequest
     fn delete_retrieval_queue_entries(&mut self, txid: TxId) -> Result<(), Self::Error>;
+
+    /// Updates the state of the wallet backend to indicate that the given range of blocks has been
+    /// fully scanned, identifying the position in the note commitment tree of any notes belonging
+    /// to the wallet that were discoveredd in the process of scanning.
```
str4d:
```suggestion
    /// to the wallet that were discovered in the process of scanning.
```

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -149,6 +149,26 @@ where
...
+    #[cfg(feature = "orchard")]
+    for action in d_tx
+        .tx()
+        .orchard_bundle()
+        .iter()
+        .flat_map(|b| b.actions().iter())
+    {
+        wallet_db.mark_orchard_note_spent(action.nullifier(), tx_ref)?;
+    }
+
```
str4d: Where were these calls happening before in `store_decrypted_tx` when it was over in `zcs`? Is this an undocumented bugfix in f8450d30f1a19b5a9fd719e078ad8f079a8bfccc? If so, document it in the changelog.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -233,6 +253,16 @@ where
...
+    // Mark Sapling notes as spent when we observe their nullifiers.
+    for spend in d_tx
+        .tx()
+        .sapling_bundle()
+        .iter()
+        .flat_map(|b| b.shielded_spends().iter())
+    {
+        wallet_db.mark_sapling_note_spent(spend.nullifier(), tx_ref)?;
+    }
+
```
str4d, line 751: Wait... this exact code was added just earlier in this function. Is this a rebase bug? Is the earlier addition also a rebase bug?

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -316,6 +346,17 @@ where
...
+    #[cfg(feature = "orchard")]
+    for action in d_tx
+        .tx()
+        .orchard_bundle()
+        .iter()
+        .flat_map(|b| b.actions().iter())
+    {
+        wallet_db.mark_orchard_note_spent(action.nullifier(), tx_ref)?;
+    }
+
```
str4d, line 845: Same question: is this a rebase bug?

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -90,6 +111,449 @@ where
     })
 }
 
+pub enum PutBlocksError<SE, TE> {
```
str4d, line 120: - Document this error enum and its variants.
- Make it non-exhaustive.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -90,6 +111,449 @@ where
...
+    ShardTree(ShardTreeError<TE>),
+}
+
+impl<SE, TE> From<ShardTreeError<TE>> for PutBlocksError<SE, TE> {
+    fn from(value: ShardTreeError<TE>) -> Self {
+        PutBlocksError::ShardTree(value)
+    }
+}
+
+pub fn put_blocks<DbT, SE, TE>(
```
str4d, line 147: Document this method.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -373,8 +838,8 @@ where
         let (account_id, _) =
             wallet_db.put_transparent_output(received_t_output, observed_height, false)?;
 
-        if let Some(key_scope) = key_scope {
-            gap_update_set.insert((account_id, *key_scope));
+        if let Some(t_key_scope) = key_scope {
+            gap_update_set.insert((account_id, *t_key_scope));
```
str4d: Unclear whether this diff in ab895a8df30e468b7e07d9ff99ef45b0180d4aa7 is fixing the naming from an earlier commit, or undoing fixed naming from an earlier commit.

zcash_client_sqlite/src/lib.rs
```diff
@@ -252,7 +233,7 @@ impl ConditionallySelectable for AccountRef {
 }
 
 /// An opaque type for received note identifiers.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct ReceivedNoteId(pub(crate) ShieldedProtocol, pub(crate) i64);
```
str4d, line 237: Document this added `impl` in the changelog.

zcash_client_sqlite/src/lib.rs
```diff
@@ -264,15 +245,15 @@ impl fmt::Display for ReceivedNoteId {
 }
 
 /// A newtype wrapper for sqlite primary key values for the utxos table.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct UtxoId(pub(crate) i64);
```
str4d, line 249: Document these added `impl`s in the changelog.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -90,6 +111,449 @@ where
...
+    };
+
+    let mut sapling_commitments = vec![];
+    #[cfg(feature = "orchard")]
+    let mut orchard_commitments = vec![];
+    let mut last_scanned_height = None;
+    let mut note_positions = vec![];
+
+    #[cfg(feature = "transparent-inputs")]
+    let mut tx_refs = HashSet::new();
```
str4d, line 190: The move commit ab895a8df30e468b7e07d9ff99ef45b0180d4aa7 alters this from `BTreeSet` to `HashSet`; it isn't clear to me why.

zcash_client_sqlite/src/wallet.rs
nuttycom: The method name is wrong, so, not a bug, but potentially misleading.

zcash_client_backend/src/data_api/ll/wallet.rs
nuttycom: Essentially, for the generalized API we need to choose a constraint, either Ord or Hash. BTreeSet is slightly more performant for small sets, but for a set that is of a size of several blocks worth of transactions, I think that HashSet might be a better choice.

zcash_client_backend/src/data_api/ll/wallet.rs
nuttycom: Making this non-exhaustive would mean adding a case for the unhandled case to `SqliteClientError`, since there's a function (which probably should just be a `From` impl) that maps the cases of PutBlocksError into `SqliteClientError`. It feels weird to add `SqliteClientError::UnrecognizedPutBlocksError`. I guess instead of having a conversion, `SqliteClientError` could just wrap `PutBlocksError`, but then you get the weird stack `SqliteClientError::PutBlocks(PutBlocksError::Storage(SqliteClientError))` which is just gross. I'm going to leave this as an exhaustive type for now.

zcash_client_backend/src/data_api/ll.rs
nuttycom: This doesn't need specific documentation, because the whole `ll` module is added in this PR.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/405d2132b2c6cf6d3e0bd528db4a37cbefb487f5..640445fb9cf22a7ef1a4ae06d2c1d482cc042345) to address comments from code review.

nuttycom:
Then [force-pushed](https://github.com/zcash/librustzcash/compare/640445fb9cf22a7ef1a4ae06d2c1d482cc042345..6c32fbd73e53092b54b06bfdbc9593adae4beef3) to rebase and resolve conflicts with `main`.

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,682 @@
...
+        output: &T,
+        tx_ref: Self::TxRef,
+        target_or_mined_height: Option<BlockHeight>,
+        spent_in: Option<Self::TxRef>,
+    ) -> Result<(), Self::Error>;
+
+    /// Updates the backing store to indicate that the Sapling output having the given nullifier is
+    /// spent in the transaction referenced by `spent_in_tx`. This may result in multiple distinct
+    /// transactions being recorded as having spent the note; only one of these transactions will
+    /// end up having been mined (by consensus.)
```
str4d:
```suggestion
    /// end up having been mined (by consensus).
```

str4d: I'm okay with this being the documentation resolution to https://github.com/zcash/librustzcash/pull/2036#discussion_r2636673607 as long as you are on the same page that if the caller sets `spent_in_tx` to a mined transaction when `nf` is already marked as spent by another mined transaction, that this method would return an error (because this is the caller making a mistake, not something permitted by consensus, and allowing it to be saved would result in database corruption).

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,682 @@
...
+        output: &T,
+        tx_ref: Self::TxRef,
+        target_or_mined_height: Option<BlockHeight>,
+        spent_in: Option<Self::TxRef>,
+    ) -> Result<(), Self::Error>;
+
+    /// Updates the backing store to indicate that the Orchard output having the given nullifier is
+    /// spent in the transaction referenced by `spent_in_tx`. This may result in multiple distinct
+    /// transactions being recorded as having spent the note; only one of these transactions will
+    /// end up having been mined (by consensus.)
```
str4d:
```suggestion
    /// end up having been mined (by consensus).
```

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,682 @@
...
+        spent_in_tx: Self::TxRef,
+    ) -> Result<bool, Self::Error>;
+
+    /// Causes the given Sapling output nullifiers to be tracked by the wallet.
+    ///
+    /// When scanning the chain out-of-order, it is necessary to store the any nullifiers observed
+    /// after a gap in the scanned blocks until the blocks in that gap have been fully scanned, in
+    /// order to be able to immediately detect that a received output has already been spent. The
+    /// data store should track a mapping from each nullifier to the block, transaction index and
+    /// ID of the transaction where the spend occurred.
```
str4d: This documentation change (in [this diff](https://github.com/zcash/librustzcash/compare/17f8f87a41cbd1451a48845744805b32237ad3de..640445fb9cf22a7ef1a4ae06d2c1d482cc042345)) is incorrect. This dosn't refer to the `nfs` argument structure, but to the structure that the low-level data store needs to maintain *derived from* `nfs`. Either rewrite this to be documenting the `nfs` argument and make no reference to what the data store should do with it, or revert this change and separately document `nfs`.
```suggestion
    /// data store should track a mapping from each nullifier to the block, transaction ID and
    /// input index where the nullifier was revealed, so that the state of a note received within
    /// that "gap" can be properly updated when it is discovered.
```

(Note that my [original review comment](https://github.com/zcash/librustzcash/pull/2036#discussion_r2636678445) about the documentation being confusing was resolved by the introduction of a `TxIndex` newtype, so there's no `u16` that could be misunderstood as "the input index where the nullifier was revealed"; documenting `nfs` separately from what the data store should store would still be useful.)

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,682 @@
...
+        tx_ref: Self::TxRef,
+    ) -> Result<bool, Self::Error>;
+
+    /// Causes the given Orchard output nullifiers to be tracked by the wallet.
+    ///
+    /// When scanning the chain out-of-order, it is necessary to store the any nullifiers observed
+    /// after a gap in the scanned blocks until the blocks in that gap have been fully scanned, in
+    /// order to be able to immediately detect that a received output has already been spent. The
+    /// data store should track a mapping from each nullifier to the block, transaction index and
+    /// ID of the transaction where the spend occurred.
```
str4d: Ditto:
```suggestion
    /// data store should track a mapping from each nullifier to the block, transaction ID and
    /// input index where the nullifier was revealed, so that the state of a note receive within
    /// that "gap" can be properly updated when it is discovered.
```

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,682 @@
...
+    #[cfg(feature = "orchard")]
+    fn track_block_orchard_nullifiers(
+        &mut self,
+        block_height: BlockHeight,
+        nfs: &[(TxIndex, TxId, Vec<::orchard::note::Nullifier>)],
+    ) -> Result<(), Self::Error>;
+
+    /// Removes nullifiers tracked by the wallet for transactions mined in blocks at heights less
+    /// than `pruning_depth` blocks below the height to which the wallet has been fully scanned.
```
str4d: This is ambiguous. Is "less than" referring to `pruning_depth`, or `wallet.fully_scanned_height - pruning_depth`? These are non-overlapping block ranges.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -16,10 +16,21 @@ workspace.
   - `tx_mined_height`
   - `tx_trust_status`
   - `recipient_key_scope`
+- `zcash_client_sqlite::TxRef`
+- `impl<'a> Borrow<rusqlite::Transaction<'a>> for zcash_client_sqlite::SqlTransaction<'a>`
+- `impl zcash_client_backend::data_api::ll::LowLevelWalletRead for WalletDb`
+- `impl zcash_client_backend::data_api::ll::LowLevelWalletWrite for WalletDb`
+- `impl Hash for zcash_client_sqlite::{ReceivedNoteId, UtxoId, TxRef}`
+- `impl {PartialOrd, Ord} for zcash_client_sqlite::{UtxoId, PartialOrd, Ord}`
```
```suggestion
- `impl {PartialOrd, Ord} for zcash_client_sqlite::UtxoId`
```

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -0,0 +1,1022 @@
...
+                },
+                &note_positions,
+            )
+            .map_err(PutBlocksError::Storage)?;
+    }
+
+    Ok(())
+}
+
+pub fn store_decrypted_tx<DbT, P>(
```
str4d, line 597: Document this function.

zcash_client_backend/src/data_api/ll.rs
str4d: In a pairing we re-determined that the reason `put_received_transparent_utxo` uses a chain tip height is not because we know the UTXO is unmined and should be using the mempool height, but because we know the UTXO is mined but not where, so we use the chain tip as the best guess.

zcash_client_sqlite/src/testing.rs
```diff
@@ -43,13 +43,15 @@ impl BlockCache {
...
-pub struct BlockCacheInsertionResult {
+pub(crate) struct BlockCacheInsertionResult {
     txids: Vec<TxId>,
+    #[allow(dead_code)]
     note_commitments: NoteCommitments,
 }
 
 impl BlockCacheInsertionResult {
-    pub fn note_commitments(&self) -> &NoteCommitments {
+    #[allow(dead_code)]
```
nullcopy, line 53: The commit message mentions a TODO to remove these pragmas later. Is there a git issue to track this?

zcash_client_backend/src/wallet/transparent.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl GapLimits {
+    /// Constructs a new `GapLimits` value from its constituent parts.
+    ///
+    /// The gap limits recommended for use with this crate are supplied by the [`Default`]
+    /// implementation for this type.
+    ///
+    /// This constructor is only available under the `unstable` feature, as it is not recommended
+    /// for general use.
```
nullcopy, line 35: (nit) You mention this constructor is not recommended for general use. You might mention that `GapLimits::default()` is the recommendation instead

zcash_client_backend/src/data_api/ll.rs
```diff
@@ -0,0 +1,683 @@
...
+    ///   returned output. If this is `None`, no spendability checks are performed.
+    #[cfg(feature = "transparent-inputs")]
+    fn get_wallet_transparent_output(
+        &self,
+        outpoint: &OutPoint,
+        target_height: Option<TargetHeight>,
+    ) -> Result<Option<WalletUtxo>, Self::Error>;
+
+    /// Returns the vector of transactions in the wallet that spend the transparent outputs of the
+    /// referenced transaction, but for which the amount of fee paid is unknown. This should
```
nullcopy, line 234: > but for which the amount of fee paid is unknown

This constraint is non-obvious from the function name. I'm curious the rationale here.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -149,6 +149,26 @@ where
...
 
+    // Mark notes as spent and remove them from the scanning cache when we observe their
+    // nullifiers.
+    for spend in d_tx
+        .tx()
+        .sapling_bundle()
+        .iter()
+        .flat_map(|b| b.shielded_spends().iter())
+    {
+        wallet_db.mark_sapling_note_spent(spend.nullifier(), tx_ref)?;
```
nullcopy: Is there a reason you don't also mark transparent outputs as spent?

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -0,0 +1,1022 @@
...
+) -> Result<(), <DbT as LowLevelWalletRead>::Error>
+where
+    DbT: LowLevelWalletWrite,
+    DbT::AccountId: core::fmt::Debug,
+    DbT::Error: From<BalanceError>,
+    P: consensus::Parameters,
+{
+    let funding_accounts = wallet_db.get_funding_accounts(d_tx.tx())?;
+
+    // TODO(#1305): Correctly track accounts that fund each transaction output.
```
schell, line 611: It looks like the TODO here (https://github.com/zcash/librustzcash/issues/1305) is at least partially addressed by these changes?

zcash_client_backend/src/wallet/transparent.rs
```diff
@@ -0,0 +1,88 @@
+//! Types and operations specific to handling of transparent addresses and transparent outputs
+//! received by the wallet.
+use transparent::keys::TransparentKeyScope;
+
+/// A data structure that can be used to configure custom gap limits for use in transparent address
+/// rotation.
```
schell: It would be nice to have a quick explainer of what gap limits _are_, to close the circular documentation here. Something like:

```suggestion
/// Configuration for gap limits used in transparent address rotation.
///
/// A "gap limit" is the maximum number of consecutive unused addresses that a wallet
/// will scan before assuming no more addresses have been used. This is necessary for
/// HD (hierarchical deterministic) wallets, which can derive an unlimited number of
/// addresses from a single seed. During wallet recovery or synchronization, the wallet
/// scans addresses sequentially and stops when it encounters this many consecutive
/// addresses with no transaction history.
///
/// This struct allows configuring separate gap limits for different address types:
/// - `external`: Addresses shared with others for receiving payments (default: 10)
/// - `internal`: Change addresses used internally by the wallet (default: 5)
/// - `ephemeral`: Single-use addresses for specific transaction types, e.g., TEX addresses (default: 10)
///
/// Higher gap limits increase the chance of finding all funds during wallet recovery,
/// but also increase sync time and computational cost.
```

schell: Though I see this is all just being moved, so up to you! (Also I'm not sure if my suggestion is even accurate.)

zcash_client_backend/src/data_api/ll/wallet.rs
schell: I'm guessing we don't care about deterministic ordering here?

zcash_client_backend/src/data_api/ll.rs
nuttycom: I've updated the function name as @str4d previously suggested; at some point in the history of this PR that change seemed to have gotten lost, but even with the update it doesn't mention the fee in the function name. I think that's okay though, this is quite a low-level API, and so long as the documentation clearly specifies the restriction and the rationale, that's probably sufficient; it's hard to fit all the documentation into the function name itself, but suggestions would be gratefully accepted.

zcash_client_backend/src/data_api/ll.rs
nuttycom: In ZCS this requires an extra roundtrip to the database, do you think that's a cost worth paying?

zcash_client_backend/src/data_api/ll/wallet.rs
nuttycom: @schell correct; the ordering would be arbitrary anyway as the keys are the internal database identifiers.

nuttycom:
Then [force-pushed](https://github.com/zcash/librustzcash/compare/39ffc349cc4151056be46c9bdf3f1bc99a574ce2..01e91190f5fba1c4b6e5d08a1d0b9dfbd2286f09) to rebase on `main`.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/1ca44387f41df95ba0fc9255900dfa54faf77341..a60e3ab35b8072366f24c8f5c8e8e0e770880f37) to address comments from code review.

zcash_client_sqlite/src/wallet.rs
nuttycom: This did not need to remain part of the public API; it was made public for the transient state prior to everything being moved to `zcash_client_backend.`

zcash_client_backend/src/data_api/ll/wallet.rs
nuttycom: Fixing the naming from the original introduction of this code.

zcash_client_backend/src/wallet/transparent.rs
nuttycom: That's literally said in the previous sentence - the recommended limits are supplied by the `Default` impl.

zcash_client_backend/src/data_api/ll/wallet.rs
nuttycom: `ScannedBlock` is shielded-only, so we don't (yet) observe any transparent outputs here.

zcash_client_backend/src/data_api/ll.rs
nuttycom: This is another one of those situations where it's annoying to detect the error, because it requires an entirely separate query; it can't be done as part of the upsert. But I guess perhaps it's worth it since these methods are now exposed externally.

nuttycom: Fixed in 3c5f82ebc1a4142a6a935f2232bab9fe9bb78b70

zcash_client_backend/src/data_api/ll.rs
nuttycom: Fixed in c43975e7e8c620118dba55bd8481987b91666540

nuttycom: I do have a question about the nullifier map code, because it looks like it's including the txid in the nullifier map entries, but an internal comment suggests that the txid was originally intended to be omitted for space efficiency.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -0,0 +1,1048 @@
...
+    }
+
+    // Mark Sapling notes as spent when we observe their nullifiers.
+    for spend in d_tx
+        .tx()
+        .sapling_bundle()
+        .iter()
+        .flat_map(|b| b.shielded_spends().iter())
+    {
+        wallet_db.mark_sapling_note_spent(spend.nullifier(), tx_ref)?;
```
str4d, line 749: Nit: we mark Sapling and Orchard notes as spent after calling `put_sent_output` for the Sapling and Orchard outputs respectively, but we mark transparent UTXOs as spent before calling `put_sent_output` for the transparent outputs (and `put_transparent_output`).

Non-blocking, it's just a consistency thing I noticed while re-reviewing (previously these were in the same order, but one of the recent force-pushes moved them).

Issue #1912, opened 2025-08-18
Proptest failure in `zip321` parsing


nuttycom:
Proptest kicked this up this morning:
```
thread 'tests::prop_zip321_roundtrip_uri' panicked at components/zip321/src/lib.rs:1163:65:
called `Result::unwrap()` on an `Err` value: ParseError("Error parsing query parameters: Parsing Error: Error { input: \"&req-H-kz7iN=0.%2F%3C%3D%F0%96%AB%B0%F0%9F%AA%B2%5C($%22%F0%9E%BA%92b%C2%A5\", code: Eof }")
...
proptest: Aborting shrinking after the PROPTEST_MAX_SHRINK_ITERS environment variable or ProptestConfig.max_shrink_iters iterations (set 1024 to a large(r) value to shrink more; current configuration: 1024 iterations)
proptest: Saving this and future failures in /home/runner/work/librustzcash/librustzcash/components/zip321/proptest-regressions/lib.txt
proptest: If this test was run on a CI system, you may wish to add the following line to your copy of the file. (You may need to create it.)
cc b040fce0ef8c3a97cea73b1d467f4a1b2bc3e3e017e41d96b091c125d2b2b57e
thread 'tests::prop_zip321_roundtrip_uri' panicked at components/zip321/src/lib.rs:1115:5:
Test failed: called `Result::unwrap()` on an `Err` value: ParseError("Error parsing query parameters: Parsing Error: Error { input: \"&req-H-kz7iN=0.%2F%3C%3D%F0%96%AB%B0%F0%9F%AA%B2%5C($%22%F0%9E%BA%92b%C2%A5\", code: Eof }").
[...]

PR #2038, opened 2025-11-09
Prevent usage of unrecognized required params in test payment generation

replicationlag:
This addresses #1912.

The `VALID_PARAMNAME` regex does not  preclude generation of parameter names of the form `req-*`. As such, it's possible that the generated URI will contain an unrecognized required parameter, causing the test case to fail on rare occasion. This was the case for the linked issue, where `&req-H-kz7iN=` was present in the URI.

This change prevents this issue by explicitly excluding parameters of this form.

I also added a test example for this specific validation.


nuttycom (approved):
utACK a6144b71bc6a758a9004db567d36f103d5c25055

PR #2043, opened 2025-11-14
fix(memo): return payment error when memo to t-address

bradcush:
Recipients using a transparent address can't receive memos but the current `Option` doesn't express that to the caller. We explicitly return a `Result` now with a `TransparentMemo(0)` error when that type of failure happens. This issue was originally mentioned in https://github.com/zcash/zcash-devtool/issues/95 as part of the `zcash-devtool` project.

### DIFF
--- a/components/zip321/src/lib.rs
+++ b/components/zip321/src/lib.rs
@@ -136,7 +136,7 @@ pub struct Payment {
 impl Payment {
     /// Constructs a new [`Payment`] from its constituent parts.
     ///
-    /// Returns `None` if the payment requests that a memo be sent to a recipient that cannot
+    /// Returns `Err` if the payment requests that a memo be sent to a recipient that cannot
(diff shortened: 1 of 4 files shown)


components/zip321/src/lib.rs
```diff
@@ -156,7 +156,7 @@ impl Payment {
                 other_params,
             })
         } else {
-            None
+            Err(Zip321Error::TransparentMemo(0))
```
nuttycom, line 159: The wrapped value here is supposed to contain the parameter index, so hardcoding to 0 isn't the correct solution. It would be okay to pass in the index to the constructor here.

nuttycom: Hmm, but then the index could be assigned incorrectly. I'm now seeing why `None` was used instead, and how it was documented.

PR #2047, opened 2025-11-20
pczt: Support appending external signatures to inputs

str4d:


zcash_transparent/src/pczt/signer.rs
```diff
@@ -81,11 +81,101 @@ impl super::Input {
...
+    ///
+    /// It is the caller's responsibility to perform any semantic validity checks on the
+    /// PCZT (for example, comfirming that the change amounts are correct) before calling
+    /// this method.
+    ///
+    /// Returns an error if the signature does not match any pubkey involved with spend
+    /// control of the input's spent coin. The supported script formats are:
+    /// - P2PKH
+    ///   - The `hash160_preimages` field must contan a mapping from the `pubkeyhash` to
+    ///     the pubkey.
```
nuttycom: Can this doc comment provide a reference link to how that mapping is structured?

zcash_transparent/src/pczt/signer.rs
```diff
@@ -81,11 +81,101 @@ impl super::Input {
...
+            }
+        }
+
+        Err(SignerError::InvalidExternalSignature)
+    }
+}
+
+fn to_pubkey(data: &[u8]) -> Result<[u8; 33], SignerError> {
+    data.try_into().map_err(|_| SignerError::UnsupportedPubkey)
 }
```
nuttycom: nit: This could be an inner function.

nuttycom (dismissed):
utACK a87b7d391309fd383e65f024a22e18d79d86684f with nits

zcash_transparent/src/pczt/signer.rs
str4d: Added a link to the field itself which defines that mapping (namely `key = RIPEMD160(SHA256(value))`).

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/5798789d04eb70f110e2609a155cf36470490736..370bf256abe3f0037e6bc04cb8a4db176bde0f0a) to fix dependency patch error.

pczt/src/roles/signer/mod.rs
```diff
@@ -170,6 +208,30 @@ impl Signer {
...
+    /// It is the caller's responsibility to perform any semantic validity checks on the
+    /// PCZT (for example, comfirming that the change amounts are correct) before calling
+    /// this method.
+    pub fn apply_sapling_signature(
+        &mut self,
+        index: usize,
+        signature: secp256k1::ecdsa::Signature,
+    ) -> Result<(), Error> {
+        self.generate_or_apply_sapling_signature(index, |spend, shielded_sighash| {
+            spend.append_signature(index, shielded_sighash, signature)
```
nuttycom: I'm confused, this is showing up as an error when I try to build this locally. Is this supposed to be `apply_signature`?

str4d: Yeah, copy-paste error locally. Fixed.

str4d:
@gstohl I generally prefer to fix my own PRs in-situ; in this case a dependency bug hid the actual compilation failure prior to me heading off for the weekend, so I didn't get back to it until today. As for the sighash refactor, the approach I've taken encapsulates the `SignableInput` logic into a single place, rather than spreading it across the `pczt` and `zcash_transparent` crates.

pczt/src/roles/signer/mod.rs
```diff
@@ -310,6 +413,18 @@ impl Authorization for EffectsOnly {
     type TzeAuth = core::convert::Infallible;
 }
 
+/// Helper to produce the correct sighash for a PCZT.
+fn sighash(
+    tx_data: &TransactionData<EffectsOnly>,
+    signable_input: &SignableInput,
+    txid_parts: &TxDigests<Blake2bHash>,
+) -> [u8; 32] {
+    v5_signature_hash(tx_data, signable_input, txid_parts)
```
daira, line 422: Shouldn't this be using the correct hash for the transaction version?

str4d: The PCZT v1 format only supports v5 tx creation.

zcash_transparent/src/pczt/signer.rs
```diff
@@ -8,6 +8,26 @@ use crate::{
...
 impl super::Input {
+    /// Helper to prepare a [`SignableInput`] for this input.
+    ///
+    /// This can be used to calculate the sighash for this input within its transaction,
+    /// to produce a signature externally suitable for passing to [`Self::append_signature`].
+    pub fn with_signable_input<T, F>(&self, index: usize, f: F) -> T
+    where
+        F: FnOnce(SignableInput) -> T,
+    {
+        // For P2PKH, `script_code` is always the same as `script_pubkey`.
```
daira, line 19: Is this guaranteed to be a P2PKH input (rather than P2SH)?

zcash_transparent/src/pczt/signer.rs
```diff
@@ -81,11 +101,103 @@ impl super::Input {
...
+    /// PCZT (for example, comfirming that the change amounts are correct) before calling
+    /// this method.
+    ///
+    /// Returns an error if the signature does not match any pubkey involved with spend
+    /// control of the input's spent coin. The supported script formats are:
+    /// - P2PKH
+    ///   - The [`Input::hash160_preimages`] field must contan a mapping from the `pubkeyhash` to
+    ///     the pubkey.
+    /// - P2MS
+    /// - P2PK
```
daira, line 117: Do we want to support P2PK? I thought it was only used for early mining rewards?

pczt/src/roles/signer/mod.rs
daira: This will need to be updated for v6 then, even if it's behind a feature flag.

zcash_transparent/src/pczt/signer.rs
str4d: If any of those early moving rewards are unspent, someone may still want to spend them. I haven't checked if this is the case, but it was easy to support.

pczt/src/roles/signer/mod.rs
str4d: It cannot be updated any earlier than when we start defining PCZT v2. Currently we have feature flagged code in the builder that returns an error if you try to build a v6 tx via the PCZT route.

zcash_transparent/src/pczt/signer.rs
str4d: No, it is P2SH if `redeem_script` is `Some(_)`. The comment is explaining the `script_pubkey` fallback.

PR #2053, opened 2025-11-25
zcash_client_sqlite: Fix table aliasing bug in unspent notes metadata queries (hotfix 0.18.10)

nuttycom:


daira (commented):
Post-hoc ACK

PR #2058, opened 2025-11-26
zcash_keys: Add `UnifiedFullViewingKey::select_ovk`.

nuttycom:


zcash_client_backend/src/data_api/error.rs
```diff
@@ -43,6 +43,7 @@ pub enum Error<DataSourceError, CommitmentTreeError, SelectionError, FeeError, C
     /// * spend a prior shielded output;
     /// * pay to an output pool for which the corresponding feature is not enabled;
     /// * pay to a TEX address if the "transparent-inputs" feature is not enabled.
+    /// * a proposal step has no inputs
```
str4d, line 46: How would this be a structurally valid proposal? Even coinbase transactions have a single transparent input (spending the null prevout).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1141,6 +1141,25 @@ where
...
+        if !sapling_inputs.is_empty() {
+            sources.push(PoolType::SAPLING);
+        }
+        #[cfg(feature = "orchard")]
+        if !orchard_inputs.is_empty() {
+            sources.push(PoolType::ORCHARD);
+        }
+        #[cfg(feature = "transparent-inputs")]
+        if !(proposal_step.transparent_inputs().is_empty()
+            && proposal_step.prior_step_inputs().is_empty())
```
str4d, line 1153: `proposal_step.prior_step_inputs()` can refer to a shielded output of a prior step. We don't currently support it and detect/reject it at the top of this function, so we should either collect these input sources at that time for use here, or we should document here that we are assuming the checks further up (so that if/when they are changed, we don't cause this to become a bug).

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1259,44 +1278,28 @@ where
...
-    };
-
-    #[cfg(feature = "orchard")]
-    let orchard_internal_ovk = || {
-        #[cfg(feature = "transparent-inputs")]
-        if proposal_step.is_shielding() {
-            return ufvk
-                .transparent()
-                .map(|k| orchard::keys::OutgoingViewingKey::from(k.internal_ovk().as_bytes()));
+    let external_ovk = match ovk_policy {
```
str4d: Document this policy change in the changelog, as well as updating the `OvkPolicy` documentation.

zcash_client_backend/CHANGELOG.md
```diff
@@ -10,6 +10,24 @@ workspace.
...
+- `zcash_client_backend::wallet::OvkPolicy` has been substantially modified
+  to reflect the view that a single outgoing viewing key should be uniformly
+  applied to encrypt all external transaction outputs, irrespective of which
+  pool those outputs are produced into, and that by default wallet-internal
+  outputs should not be decryptable using an OVK (and will therefore only
+  be recoverable using the wallet's internal IVK).
+  - The semantics of `OvkPolicy::Sender` have changed. This policy now
+    specifies that wallet-internal change outputs should be encrypted
+    using a random key, rendering them unrecoverable to entities that
+    do not posess the internal IVK for the wallet.
```
str4d: "wallet-internal change outputs should be encrypted using a random key" is misleading. They are both always encrypted with a random key (because ephemera-static DH), and not encrypted with a random key from the recipient's perspective (otherwise the IVK wouldn't work).

Reword this to clarify that it is solely the outgoing ciphertexts (which do not encrypt the change outputs themselves) that should be made unrecoverable.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: A shielded prior step output cannot be used as a step input, because we cannot compute the witness that would be required to spend it.

nuttycom: Attempting to spend a prior step input already results in an error on https://github.com/zcash/librustzcash/pull/2058/files/80b1d26d6b813029481a714838e725dd3c5f1440..d1700ce2abfb64cc97b254d811a8ede29b765827#diff-dc0076ef8eaf064e73a80f1a6210a9b41da40eb49b24cefdf00b2797a255abafL1033-L1038

zcash_client_backend/src/data_api/error.rs
nuttycom: Isn't a transaction that pays no fee and has only zero-valued outputs still valid? You'd need a miner to explicitly include it, but I don't believe there's any consensus rule that prohibits it.

str4d: Per consensus, there must be at least one input to every transaction: https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/main.cpp#L1492-L1520

However, that input *may* be a dummy shielded spend, and I don't recall whether a `Proposal` has already decided what dummy spends will be present, or if it only includes real spends.

zcash_client_backend/src/data_api/wallet.rs
str4d: That error checking is precisely what I was referring to, and there's a TODO above it for potentially supporting shielded prior step outputs in future. If that TODO is implemented, this code becomes a bug.

zcash_client_backend/src/data_api/wallet.rs
str4d: This was not done in the changelog. The changelog update in the third commit covers the policy changes to `OvkPolicy::Sender` from that commit, but not the policy change to `OvkPolicy::Sender` from the second commit (that the per-pool OVKs are now not used simultaneously).

zcash_client_backend/src/data_api/wallet.rs
nuttycom: That TODO is impossible to implement.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: The CHANGELOG entry for `OvkPolicy` *does* reflect this change. The change is not specific to the `Sender` policy, so the paragraph above covers it.

zcash_client_backend/src/data_api/error.rs
str4d: We confirmed that the Proposal does not include dummies (they are added by the builder later on), so it's fine to consider this "unsupported".

str4d (approved):
utACK 4e634914ccf86f955a63eb29105286cf19b3f923

zcash_client_backend/src/data_api/wallet.rs
daira: Please modify the TODO to be a comment saying why that cannot be supported (and close the associated issue if there is one).

Issue #1946, opened 2025-09-11
zcash_keys: Migrate transparent gap limit address generation rules from `zcash_client_sqlite` to `zcash_keys`


nuttycom:
Correct gap limit handling is tricky, and is good to standardize for the wallet ecosystem. The gap limit handling logic in `zcash_client_sqlite` can be abstracted to take advantage of a capability-oriented approach to the wallet database interaction, and then that logic can be extracted into the `zcash_keys` crate so that others can make use of it. This will also be advantageous for the maintenance of the `zcash_client_memory` crate, which has not yet implemented general gap limit handling logic.

ANISH-SR:
Can I take this up?

nuttycom:
Completed in #2065 and follow-ups.

PR #2065, opened 2025-12-07
refactor(gap): move gap limits logic into zcash keys

bradcush:
This PR moves shared gap limit logic from `zcash_client_sqlite` into `zcash_keys` as represented in https://github.com/zcash/librustzcash/issues/1946. Moving gap limit logic there should make it easier and less error prone when supporting other backends like `zcash_client_memory`. Since a lot of related work has been done in abstracting `LowLevelWalletl{Read,Write}` out of `zcash_client_sqlite`, this work currently builds on top of https://github.com/zcash/librustzcash/pull/2036. The goal will be to wait until that PR is merged and then rebase on main.

## A few details on the approach taken

- Existing gap limit stuff is moved from `zcash_client_sqlite` to `zcash_keys`
- Introduces a new `GapLimits` trait decoupled from `LowLevelWallet{Read,Write}`
- Shared logic with gap limit specific implementation details created in `zcash_keys`
- Exposed new account related DB stuff in `LowLevelWallet{Read,Write}` that was needed
- Gap limits related stuff is hidden behind `transparent-inputs` feature
  - This required intermediate traits for some conditional bounds
- Kept duplicated logic in `zcash_client_sqlite` as it still calls `generate_gap_addresses`
  - This feels acceptable for now but could be removed/changed in another PR
- Interfaces weren't changed much regarding DB access, things were mostly moved

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -8,17 +8,15 @@ use std::{
 use assert_matches::assert_matches;
 use incrementalmerkletree::{Level, Position, frontier::Frontier};
 use rand::{Rng, RngCore};
-use rand_core::OsRng;
 use secrecy::Secret;
 use shardtree::error::ShardTreeError;
 
-use transparent::{address::TransparentAddress, builder::TransparentSigningSet};
+use transparent::address::TransparentAddress;
 use zcash_keys::{address::Address, keys::UnifiedSpendingKey};
 use zcash_primitives::{
     block::BlockHash,
     transaction::{
         Transaction,
-        builder::{BuildConfig, Builder},
         fees::zip317::{FeeRule as Zip317FeeRule, MARGINAL_FEE, MINIMUM_FEE},
     },
 };
@@ -29,7 +27,6 @@ use zcash_protocol::{
     memo::{Memo, MemoBytes},
     value::Zatoshis,
 };
-use zcash_script::opcode::PushValue;
 use zip32::Scope;
 use zip321::{Payment, TransactionRequest};
 
@@ -84,7 +81,11 @@ use zcash_protocol::PoolType;
 #[cfg(feature = "pczt")]
 use {
     pczt::roles::{prover::Prover, signer::Signer},
+    rand_core::OsRng,
+    transparent::builder::TransparentSigningSet,
+    zcash_primitives::transaction::builder::{BuildConfig, Builder},
     zcash_proofs::prover::LocalTxProver,
+    zcash_script::opcode::PushValue,
 };
 
 pub mod dsl;


zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -112,13 +118,73 @@ where
...
+        wallet_db
+            .generate_address_range_internal(
+                account_ref,
+                &account.uivk(),
+                account.ufvk(),
+                key_scope,
+                request,
+                gap_start..gap_start.saturating_add(gap_limit),
+                false,
+            )
```
nuttycom: One of the primary things I would like to see is for all the "business logic" parts of `generate_address_range_internal` to be moved up into `zcash_keys`, which AFAICT this PR does not yet do. The implementation here, which involves a few IO operations against the backing data store that should be abstracted over, is what needs to be "pulled up" into `zcash_keys` and made reusable. What's currently being moved up is kind of trivial.

bradcush: That makes sense, I've just moved the business logic from `generate_address_range_internal` into `zcash_keys`, separating address list generation from storage, so there's a new abstraction to `store_address_range`. List generation lives inside `zcash_keys` as well and can use it directly but since we have some internal calls in `zcash_client_sqlite` between gap limit functions I've also imported that and called it directly there. Otherwise, I think a lot more would need to change, so I felt this duplication was okay for now. Many functions dealing with gap limits in `zcash_client_sqlite` still exist with the same API to accommodate that, but they should be pretty thin now that the business logic is moved, we really just store stuff. I figure we'll eventually be able to remove more over time which this sets up. I also kept `GapMetadata` where it was since that was moved in your PR recently. Happy to move that if we agree it should also live in `zcash_keys`. I've rebased on the latest changes in your branch.

nuttycom: Thanks for doing this work, I will re-review it tomorrow!

nuttycom: A quick note, I had to rebase my PR because of conflicts with `main`, so you may want to also rebase this one.

bradcush: Sounds good, thanks, I've just rebased on your PR.

bradcush:
> This was automatically closed because the upstream PR merged. Will reopen.

Thanks for re-targeting `main` and rebasing the PR as well.

zcash_keys/src/keys/transparent/wallet.rs
```diff
@@ -0,0 +1,45 @@
...
+    ) -> Result<Option<NonHardenedChildIndex>, Self::Error>;
+
+    fn store_address_range(
+        &self,
+        account_id: Self::AccountRef,
+        key_scope: TransparentKeyScope,
+        list: Vec<(Address, TransparentAddress, NonHardenedChildIndex)>,
+    ) -> Result<(), Self::Error>;
+
+    fn update_gap_limits(
```
nuttycom: This method is not used in `zcash_keys`, so it should not be a member of this trait. I will move it to `LowLevelWalletWrite`.

zcash_keys/src/keys/transparent/wallet.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// non-hardened transparent child index.
+    fn find_gap_start(
+        &self,
+        account_ref: Self::AccountRef,
+        key_scope: TransparentKeyScope,
+        gap_limit: u32,
+    ) -> Result<Option<NonHardenedChildIndex>, Self::Error>;
+
+    fn store_address_range(
+        &self,
```
nuttycom: As a mutative capability, this needs to take `&mut self` rather than `&self`.

nuttycom (approved):
utACK 66d1f76833c34dba68f6ea0cd3dadcdabe74fe36

PR #2066, opened 2025-12-08
fix: zcash_client_sqlite take coinbase maturity into account

### LINKED ISSUE #821: Verify whether shielding functionality takes into account coinbase maturity.
If a miner mines directly to the transparent address of an autoshielding mobile wallet, we need to ensure that the coins in question have satisfied the coinbase maturity constraint before we attempt to shield them.

schell:
These changes add a query to exclude immature coinbase outputs from note selection.
 
Fixes #821.

Requirement for:
* https://github.com/zcash/wallet/issues/89

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -4968,15 +4968,15 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
 }
 
 #[cfg(feature = "pczt")]
-fn build_coinbase_tx(
+fn build_transparent_coinbase_tx(
     network: &LocalNetwork,
-    target_height: BlockHeight,
+    target_height: TargetHeight,
     value: Zatoshis,
     recipient: TransparentAddress,
     miner_data: Option<PushValue>,
 ) -> zcash_primitives::transaction::builder::BuildResult {
     let build_config = BuildConfig::Coinbase { miner_data };
-    let mut builder = Builder::new(*network, target_height, build_config);
+    let mut builder = Builder::new(*network, BlockHeight::from(target_height), build_config);
 
     // Add transparent output to recipient
     builder.add_transparent_output(&recipient, value).unwrap();
@@ -5016,8 +5016,13 @@ pub fn immature_coinbase_outputs_are_excluded_from_note_selection<T: ShieldedPoo
     let coinbase_height = st.sapling_activation_height();
 
     // Construct the coinbase transaction and mine the block
-    let coinbase_build_result =
-        build_coinbase_tx(st.network(), coinbase_height, coinbase_value, t_addr, None);
+    let coinbase_build_result = build_transparent_coinbase_tx(
+        st.network(),
+        TargetHeight::from(coinbase_height),
+        coinbase_value,
+        t_addr,
+        None,
+    );
     let coinbase_tx = coinbase_build_result.transaction();
     let (h, _) = st.generate_next_block_from_tx(0, coinbase_tx);
     st.scan_cached_blocks(h, 1);
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -1067,9 +1067,6 @@ pub(crate) fn excluding_immature_coinbase_outputs(tx: &str) -> String {
         r#"
         NOT (
             -- the output is a coinbase output
-            -- -- TODO(schell): After https://github.com/zcash/librustzcash/pull/2036 merges we may be able
-            -- -- to simplify this query to:
-            -- -- {tx}.tx_index == 0
             IFNULL({tx}.tx_index, 1) == 0
             -- the coinbase output is immature (< 100 confirmations)
             AND :target_height - {tx}.mined_height < {COINBASE_MATURITY_BLOCKS}


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1045,6 +1045,27 @@ pub(crate) fn excluding_wallet_internal_ephemeral_outputs(
...
+///   used as a strictly constricting clause on the set of outputs.
+pub(crate) fn excluding_immature_coinbase_outputs(tx: &str) -> String {
+    format!(
+        r#"
+        -- the output is _not_ a coinbase output
+        {tx}.tx_index != 0
+        -- or tx is mined and has at least 100 confirmations
+        OR (
+            {tx}.mined_height < :target_height -- tx is mined
+            AND :target_height - {tx}.mined_height >= 100
```
nuttycom: Since we're already in a `format!` block, maybe introduce a constant?
```suggestion
            AND :target_height - {tx}.mined_height >= {COINBASE_MATURITY_BLOCKS}
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1045,6 +1045,27 @@ pub(crate) fn excluding_wallet_internal_ephemeral_outputs(
...
+///   outputs that have already otherwise been verified to be spendable, i.e. it must be
+///   used as a strictly constricting clause on the set of outputs.
+pub(crate) fn excluding_immature_coinbase_outputs(tx: &str) -> String {
+    format!(
+        r#"
+        -- the output is _not_ a coinbase output
+        {tx}.tx_index != 0
+        -- or tx is mined and has at least 100 confirmations
+        OR (
+            {tx}.mined_height < :target_height -- tx is mined
```
nuttycom: This condition seems like it shouldn't be necessary, because the `:target_height - {tx}.mined_height` on the next line will always return `NULL` and thus the test against `{COINBASE_MATURITY_BLOCKS}` will fail.
```suggestion
```

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1045,6 +1045,27 @@ pub(crate) fn excluding_wallet_internal_ephemeral_outputs(
...
+            AND :target_height - {tx}.mined_height >= 100
+        )
```
nuttycom:
```suggestion
        OR :target_height - {tx}.mined_height >= {COINBASE_MATURITY_BLOCKS}
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4939,3 +4939,81 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
...
+        AddressType::DefaultExternal,
+        value,
+    )];
+    let total_value = value;
+
+    // `st.generate_next_block` with multiple outputs.
+    let pre_activation_block = CachedBlock::none(st.sapling_activation_height() - 1);
+    let prior_cached_block = st.latest_cached_block().unwrap_or(&pre_activation_block);
+    let h = prior_cached_block.height() + 1;
+    st.generate_block_at(
```
nuttycom: This won't correctly add a transparent coinbase output, because we don't (yet) support transparent inputs and outputs in compact blocks. One possible solution is to update to `lightwallet-protocol v0.4.0` and then update the test harness accordingly, but that really depends on `scan_cached_blocks` correctly handling coinbase outputs, which I'm currently working on. Alternately, we need to get https://github.com/zcash/librustzcash/pull/1893 merged so that we can explicitly create a coinbase transaction and add it to a block using `fake_compact_block_from_tx`.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4954,3 +4957,114 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
...
 }
+
+fn build_coinbase_tx(
+    network: &LocalNetwork,
+    value: Zatoshis,
+    recipient: TransparentAddress,
+    miner_data: Option<PushValue>,
+) -> zcash_primitives::transaction::builder::BuildResult {
+    let build_config = BuildConfig::Coinbase { miner_data };
+    let mut builder = Builder::new(network.clone(), BlockHeight::from_u32(1), build_config);
```
nuttycom: For fidelity, this shouldn't hardcode the target height, but should take it as an argument.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4954,3 +4957,114 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
...
+) {
+    let mut st = TestDsl::with_sapling_birthday_account(dsf, cache).build::<T>();
+
+    // Get the default transparent address
+    let (t_addr, _) = st.get_account().usk().default_transparent_address();
+
+    let coinbase_value = Zatoshis::const_from_u64(50000);
+
+    // Construct the coinbase transaction and mine the block
+    let coinbase_build_result = build_coinbase_tx(st.network(), coinbase_value, t_addr, None);
```
nuttycom: Given that the Sapling birthday height is far above block height 1, this transaction should actually be expired when `generate_next_block_from_tx` is called, but as we don't have that consensus check in place it's incorrectly succeeding. We should probably do that check in `generate_next_block_from_tx` because there are cases around transaction expiry that we should be testing in some places.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4954,3 +4957,114 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
...
+    let coinbase_tx = coinbase_build_result.transaction();
+    let (h, _) = st.generate_next_block_from_tx(0, coinbase_tx);
+    st.scan_cached_blocks(h, 1);
+
+    // Store the UTXO in the wallet (not sure about this step)
```
nuttycom: This is okay; another way to do this would be to call `decrypt_and_store_transaction`.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4954,3 +4957,114 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
...
+        .unwrap();
+    assert!(
+        !spendable_utxos.is_empty(),
+        "Coinbase output should be spendable at blockheight {latest_height} \
+        with {confirmations} confirmations since the coinbase tx was mined (at {h})\n \
+        target_height {target_height:?} - coinbase_tx.mined_height {h} = {}",
+        u32::from(target_height) - u32::from(h)
+    );
+
+    // TODO(schell): Verify we can propose shielding the coinbase utxo
```
nuttycom: This is probably where transaction expiry will bite you.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1045,6 +1045,34 @@ pub(crate) fn excluding_wallet_internal_ephemeral_outputs(
...
+/// - The parent is responsible for enclosing this condition in parentheses as appropriate.
+/// - The parent is responsible for ensuring that this condition will only be checked for
+///   outputs that have already otherwise been verified to be spendable, i.e. it must be
+///   used as a strictly constricting clause on the set of outputs.
+pub(crate) fn excluding_immature_coinbase_outputs(tx: &str) -> String {
+    // FIXME: If a coinbase transaction is discovered via the get_compact_utxos RPC call
+    // we won't have sufficient info to identify it as coinbase, so it may not be excluded
+    // unless decrypt_and_store_transaction has been called on the transaction that produced it.
+    //
+    // To fix this we'll need to add the `tx_index` field to the GetAddressUtxosReply proto type.
```
nullcopy, line 1063: Is there an issue tracking this?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1045,6 +1045,34 @@ pub(crate) fn excluding_wallet_internal_ephemeral_outputs(
...
+    format!(
+        r#"
+        NOT (
+            -- the output is a coinbase output
+            -- -- TODO(schell): After https://github.com/zcash/librustzcash/pull/2036 merges we may be able
```
nullcopy: So is this PR blocked on #2036 ?

zcash_client_memory/src/input_source.rs
```diff
@@ -155,6 +155,8 @@ impl<P: consensus::Parameters> InputSource for MemoryWalletDb<P> {
         target_height: TargetHeight,
         confirmations_policy: ConfirmationsPolicy,
     ) -> Result<Vec<WalletUtxo>, Self::Error> {
+        // TODO: take into consideration coinbase maturity
+        // See <https://github.com/zcash/librustzcash/issues/821>
```
nullcopy, line 159: Did you mean to leave this TODO here? This PR resoves #821, right?

zcash_client_sqlite/src/wallet/transparent.rs
schell: Good call. I've created one at https://github.com/zcash/lightwallet-protocol/issues/17.

zcash_client_sqlite/src/wallet/transparent.rs
schell: It's not blocked, as the `IFNULL(...)` implementation works around this, but feels a bit hacky.

zcash_client_memory/src/input_source.rs
schell: This is left for folks implementing `zcash_client_memory`, just as a reminder.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4954,3 +4957,113 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
         assert_eq!(T::note_value(note.note()), value);
     }
 }
+
+#[cfg(feature = "pczt")]
+fn build_coinbase_tx(
```
nuttycom: Naming nit: this is specifically for transparent coinbase.
```suggestion
fn build_transparent_coinbase_tx(
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4954,3 +4957,113 @@ pub fn receive_two_notes_with_same_value<T: ShieldedPoolTester>(
...
+    network: &LocalNetwork,
+    target_height: BlockHeight,
```
nuttycom: nonblocking nit:
```suggestion
    target_height: TargetHeight,
```

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: #2036 actually doesn't change the situation here, I don't believe; you're thinking of a yet-to-be-opened PR that actually implements transparent output detection as part of normal block scanning. Even with that, though, wallets will have the option to use the direct UTXO retrieval strategy, so I think this will have to remain `IFNULL` and this TODO can be removed.

nuttycom (dismissed):
utACK ddddf660e897d42333561ddefe7827fde81ea6fa

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1067,9 +1067,6 @@ pub(crate) fn excluding_immature_coinbase_outputs(tx: &str) -> String {
         r#"
         NOT (
             -- the output is a coinbase output
-            -- -- TODO(schell): After https://github.com/zcash/librustzcash/pull/2036 merges we may be able
-            -- -- to simplify this query to:
-            -- -- {tx}.tx_index == 0
             IFNULL({tx}.tx_index, 1) == 0
```
nuttycom, line 1070: Oh, actually, `{tx}.tx_index == 0` will always return `false` if `{tx}.tx_index IS NULL` anyway, so this `IFNULL` is superfluous.

nuttycom: Any comparison `NULL == foo` returns `false`.

schell: Previously, when using `{tx}.tx_index == 0`, tests would fail because many of the test transactions were erroneously being filtered by this condition. For example, if `tx_index` is set to `NULL`, `{tx}.tx_index == 0` evaluates to `NULL`, **not false**. I checked this by removing the filter and adding `(t.tx_index == 0) AS tx_index_eq_zero` to `get_spendable_transparent_outputs`, but you can also verify in sqlite itself:

```sql
sqlite> SELECT (NULL == 0) WHERE 1;

sqlite> SELECT (0 == 0) WHERE 1;
1
sqlite> SELECT (1 == 0) WHERE 1;
0
```

The failing tests were:

```
failures:
    wallet::orchard::tests::shield_transparent
    wallet::sapling::tests::shield_transparent
    wallet::transparent::tests::put_received_transparent_utxo
    wallet::transparent::tests::transparent_balance_across_shielding
    wallet::transparent::tests::transparent_balance_spendability
```

Using `IFNULL` here was the crux of getting the new coinbase maturity test to pass, while also keeping the previous tests passing, since they don't insert a `tx_index`. 

*edit*
Otherwise what happens is the whole `excluding_immature_coinbase_outputs` essentially simplifies down to `NOT NULL` preventing the transaction from being selected, even though it is _not_ a coinbase transaction...

PR #2070, opened 2025-12-08
Update lightwallet protocol protobuf descriptors to zcash/lightwallet-protocol@v0.4.0

nuttycom:
This is the minimal API change required; it does not adopt any new functionality.

### DIFF
--- a/zcash_client_backend/src/data_api/testing.rs
+++ b/zcash_client_backend/src/data_api/testing.rs
@@ -2199,7 +2199,7 @@ fn fake_compact_tx<R: RngCore + CryptoRng>(rng: &mut R) -> CompactTx {
     let mut ctx = CompactTx::default();
     let mut txid = vec![0; 32];
     rng.fill_bytes(&mut txid);
-    ctx.hash = txid;
+    ctx.txid = txid;
 
     ctx
 }
@@ -2294,7 +2294,7 @@ fn fake_compact_block_from_tx(
     // Create a fake CompactTx containing the transaction.
     let mut ctx = CompactTx {
         index: tx_index as u64,
-        hash: tx.txid().as_ref().to_vec(),
+        txid: tx.txid().as_ref().to_vec(),
         ..Default::default()
     };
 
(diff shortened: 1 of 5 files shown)


zcash_client_backend/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ workspace.
 
 ### Changed
 - Migrated to `orchard 0.12`, `sapling-crypto 0.6`.
+- Migrated to `lightwallet-protocol v0.4.0`
```
str4d: We need to mention the changes that result from this. At a minimum we should be pointing out that `zcash_client_backend::proto` will have a bunch of changes; ideally we'd list them all as it is part of our public Rust API (and for this update is reasonably short).

zcash_client_backend/src/proto/service.rs
```diff
@@ -1,4 +1,30 @@
...
+/// thus it is not "compact".
+#[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
+pub struct TxOut {
+    /// The value of the output, in Zatoshis.
+    #[prost(uint64, tag = "1")]
+    pub value: u64,
+    /// The script pubkey that must be satisfied in order to spend this output.
+    #[prost(bytes = "vec", tag = "2")]
+    pub script_pub_key: ::prost::alloc::vec::Vec<u8>,
+}
```
str4d: These should not have been generated here, which reminded me of the change this PR is missing:
https://github.com/zcash/librustzcash/blob/146245b4b1cc40b05f032168600a90c12a16cb55/zcash_client_backend/build.rs#L51-L74

These extern paths need to be updated with the new types in `compact_formats.proto` that are referenced in `service.proto`, prior to regenerating the files.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/6e37885731463638839abef97c91933a216b3e0c..4ee193db62d706c2e6ff3eb7db2a1602cc650f1c) to fix incorrect generation of `compact_formats` types in `service` module.

str4d (approved):
utACK 4ee193db62d706c2e6ff3eb7db2a1602cc650f1c

PR #2071, opened 2025-12-10
zip321: make `zip321::Payment::amount` an optional field.

nuttycom:


components/zcash_address/src/lib.rs
```diff
@@ -317,6 +317,23 @@ impl ZcashAddress {
         }
     }
 
+    /// Returns whether the only recognized receivers of this address are transparent protocol
+    /// receivers. A unified addresse with unknown receivers may be considered transparent-only if
```
str4d:
```suggestion
    /// receivers.
    ///
    /// A unified address with unknown receivers may be considered transparent-only if
```

components/zip321/src/lib.rs
```diff
@@ -802,7 +845,7 @@ pub mod testing {
     prop_compose! {
         pub fn arb_zip321_payment(network: NetworkType)(
             recipient_address in arb_address(network),
-            amount in arb_zatoshis(),
+            amount in option::of(arb_zatoshis()),
```
str4d: If `arb_zip321_payment()` is meant to return an arbitrary valid ZIP 321 payment URI, then both `amount` and `memo` need to have their values gated on `recipient_address`. If not, we should document the fact that it can return invalid states.

str4d (approved):
utACK 4150c765ae9dfc807f7c087c021739644e43a8ef with nits

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/4150c765ae9dfc807f7c087c021739644e43a8ef..a078a78881f13d633ef36dddf52453f8a89c052a) to address comments from review.

str4d (approved):
utACK a078a78881f13d633ef36dddf52453f8a89c052a

components/zip321/CHANGELOG.md
```diff
@@ -8,10 +8,28 @@ indicated by the `PLANNED` status in order to make it possible to correctly
...
+  Previously, the amount field of the payment would be set to zero if no amount
+  parameter was present; however, this behavior is not specified by ZIP 321.
+  The semantics of this field have now been altered such that if no amount is
+  specified by a payment request, this should be interpreted to mean that the
+  sender of the transaction should specify an amount for the payment.
+- `zip321::Payment::new` now takes its `amount` parameter as `Option<Zatoshis>`
+  instead of `Zatoshis`. It now returns `None` if the payment requests that a
+  zero-valued output OR a memo be sent to a transparent-only address.
+- `zip321::Zip321Error` has added variant `Zip321Error::ZeroValuedTransparentOutput`.
+  Zero-valued transparent outputs are rejected by the Zcash consensus rules,
```
daira, line 25: There is no such consensus rule. Zcash is in this respect like Bitcoin, which also allows zero-valued UTXOs: https://bitcoin.stackexchange.com/questions/104937/transaction-outputs-with-value-0

There is a standard rule in both Zcash and Bitcoin that transactions with "dust" outputs (for Zcash, call this "transparent dust" to avoid ambiguity), should not be mined. So the restriction on constructing zero-valued transparent outputs is perfectly reasonable, but it should be described accurately as a consequence of a standard rule, not a consensus rule.

```suggestion
  Zero-valued transparent outputs are rejected by the Zcash standard rules,
```

components/zip321/src/lib.rs
```diff
@@ -39,6 +40,9 @@ pub enum Zip321Error {
     /// The payment at the wrapped index attempted to include a memo when sending to a
     /// transparent recipient address, which is not supported by the protocol.
     TransparentMemo(usize),
+    /// The payment at the wrapped index requested a zero-valued output to a transparent recipient
+    /// address, which is disallowed by the Zcash consensus rules.
```
daira, line 44:
```suggestion
    /// address, which is disallowed by the Zcash standard rules.
```

components/zip321/src/lib.rs
```diff
@@ -75,6 +79,10 @@ impl Display for Zip321Error {
                 f,
                 "Payment {idx} is invalid: cannot send a memo to a transparent recipient address"
             ),
+            Zip321Error::ZeroValuedTransparentOutput(idx) => write!(
+                f,
+                "Payment {idx} is invalid: zero-valued transparent outputs are disallowed by consensus"
```
daira, line 84:
```suggestion
                "Payment {idx} is invalid: zero-valued transparent outputs are disallowed"
```

components/zip321/src/lib.rs
```diff
@@ -801,8 +845,14 @@ pub mod testing {
 
     prop_compose! {
         pub fn arb_zip321_payment(network: NetworkType)(
-            recipient_address in arb_address(network),
-            amount in arb_zatoshis(),
+            (recipient_address, amount) in arb_address(network).prop_flat_map(|addr| {
+                arb_zatoshis().prop_filter_map(
+                    "zero-valued outputs to transparent addresses are disallowed by consensus",
```
daira, line 850:
```suggestion
                    "zero-valued outputs to transparent addresses are disallowed",
```

components/zcash_address/src/lib.rs
```diff
@@ -317,6 +317,25 @@ impl ZcashAddress {
         }
     }
 
+    /// Returns whether the only recognized receivers of this address are transparent protocol
+    /// receivers.
+    ///
+    /// A unified address with unknown receivers may be considered transparent-only if it does not
```
daira, line 323:
```suggestion
    /// A unified address with unknown receivers is considered transparent-only iff it does not also include
```

components/zcash_address/src/lib.rs
```diff
@@ -317,6 +317,25 @@ impl ZcashAddress {
...
+    /// also contain a Sapling or Orchard receiver, i.e. if ordinary transaction construction
```
daira, line 324:
```suggestion
    /// a supported shielded (Sapling or Orchard) receiver, i.e. if ordinary transaction construction
```
Note that it is possible for a payment to a UA to result in a transparent output even if it does include receivers for shielded protocols, but all of those protocols are configured off. (This currently applies just to Sapling since there is no config flag for Orchard.) In that case l believe the code after this PR will behave as before, i.e. creating a transaction with a zero-valued transparent output that probably will not be mined.

components/zcash_address/src/lib.rs
```diff
@@ -317,6 +317,25 @@ impl ZcashAddress {
...
+    /// methods would produce a transaction sending to a transparent address.
+    pub fn is_transparent_only(&self) -> bool {
+        use AddressKind::*;
+        match &self.kind {
+            Sprout(_) | Sapling(_) => false,
```
daira, line 329: We don't support sending to Sprout, but it's okay that this returns false for Sprout addresses because a different error will occur.

daira (commented):
Post-hoc reviewed with substantive documentation suggestions. I will file a follow-up PR.

PR #2073, opened 2025-12-10
zcash_protocol: Construct null outpoints

nuttycom:


nuttycom (commented):
utACK; these are some of @upbqdn's changes broken out from #1893 to allow for a `zcash_protocol` patch release.

PR #2074, opened 2025-12-10
zcash_transparent: Add the ability to construct the transparent bundle for a coinbase transaction

nuttycom:


zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,108 @@
...
+            .map(MinerData)
+            .ok_or(Error::OversizedMinerData)
+    }
+}
+
+impl AsRef<[u8]> for MinerData {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
```
str4d, line 52: These implementations are not symmetric: the bytes you get out of a `MinerData` are script-encoded, but the bytes you put in are not. I think this is likely to lead to errors.

This is made even more confusing by the `serde` impl, which as it stands will serialize the script-encoded bytes, and then if someone wants to take those serialized bytes and parse them, the only avenue is the `TryFrom` but that will double-encode them.

Instead, I think `MinerData` should store the raw miner data, applying any length constraints as needed, and the conversion to script form should happen where `MinerData` is used in `TxIn::coinbase`. Alternatively, the `TryFrom` impl should be replaced by an explicit constructor that clarifies it is doing the encoding.

cc @upbqdn, I need to understand how Zebra is trying to use these APIs.

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +267,72 @@ impl TxIn<Authorized> {
...
+        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+        //
+        // ## Note
+        //
+        // The height is encoded in the `push_num` fn above.
+        //
+        // [BIP-34]: <https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki>
+        if script_sig.is_empty() {
+            Err(coinbase::Error::EmptyHeight)?;
+        } else if script_sig.len() > coinbase::MAX_COINBASE_HEIGHT_LEN {
```
str4d: This error cannot occur. `push_num` always returns an opcode, and every opcode serializes to at least one byte. We can therefore check this with an assertion, as a failure here would be a programming error.
```suggestion
        assert!(!script_sig.is_empty());
        if script_sig.len() > coinbase::MAX_COINBASE_HEIGHT_LEN {
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Clone, Debug)]
+pub enum Error {
+    /// Encoded coinbase script exceeds [`MAX_COINBASE_SCRIPT_LEN`].
+    OversizedScript,
+    /// Encoded coinbase script is shorter than [`MIN_COINBASE_SCRIPT_LEN`].
+    UndersizedScript,
+    /// Encoded block height exceeds [`MAX_COINBASE_HEIGHT_LEN`].
+    OversizedHeight,
+    /// Encoded block height is empty.
+    EmptyHeight,
```
str4d:
```suggestion
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,108 @@
...
+            Error::UndersizedScript => write!(
+                f,
+                "encoded coinbase script is shorter than the minimum of {} bytes",
+                MIN_COINBASE_SCRIPT_LEN,
+            ),
+            Error::OversizedHeight => write!(
+                f,
+                "encoded block height exceeds the limit of {MAX_COINBASE_HEIGHT_LEN}",
+            ),
+            Error::EmptyHeight => write!(f, "encoded block height is empty"),
```
str4d:
```suggestion
```

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +267,72 @@ impl TxIn<Authorized> {
...
+            Err(coinbase::Error::OversizedHeight)?;
+        }
+
+        // We can embed `miner_data` directly into `script_sig` since it is already encoded as a
+        // push value.
+        script_sig.extend(miner_data.as_ref());
```
str4d, line 305: This will always serialize an empty data at the end if `miner_data` is empty. I note that `zcashd` does do this, but as an explicit choice vs this implicit approach. I have two comments:
- We need to confirm that the consensus rules only apply to the first `PushData` in a coinbase `scriptSig`. If the `PushData([])` (i.e. `OP_0`) in `zcashd` is actually required by consensus, then it would be incorrect to ever set `miner_data` here which is unintended.
- Rather than immediately serializing to bytes (and assuming that my earlier `MinerData` question is resolve in favour of storing raw bytes), we should instead have a `PushData` for the height, and then an optional `PushData` for the `MinerData` if non-empty. Then `script_sig` can have type `zcash_script::script::Sig` (i.e. `Component<PushValue>`, and we can use `script_sig.byte_len()` to check length limits.

zcash_transparent/src/builder.rs
```diff
@@ -239,6 +249,37 @@ pub struct TransparentSignatureContext<'a, V: secp256k1::Verification = secp256k
     final_script_sigs: Vec<Option<script::Sig>>,
 }
 
+/// [`Authorization`] marker type for coinbase transactions without authorization data.
+#[derive(Debug, Clone)]
+pub struct Coinbase;
+
+impl Authorization for Coinbase {
+    type ScriptSig = Vec<u8>;
```
str4d, line 257: Then this could be:
```suggestion
    type ScriptSig = script::Sig;
```
i.e. constrained to be a `Component<PushData>` instead of completely arbitrary bytes.

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +267,72 @@ impl TxIn<Authorized> {
...
+        //
+        // ## Note
+        //
+        // These bounds are met implicitly, but we check explicitly to document compliance with the
+        // consensus rule.
+        if script_sig.len() < coinbase::MIN_COINBASE_SCRIPT_LEN {
+            Err(coinbase::Error::UndersizedScript)?;
+        } else if script_sig.len() > coinbase::MAX_COINBASE_SCRIPT_LEN {
+            Err(coinbase::Error::OversizedScript)?;
+        }
```
str4d, line 321: To avoid calculating the length more than once from a `script::Sig`:
```suggestion
        match script_sig.byte_len() {
            0..coinbase::MIN_COINBASE_SCRIPT_LEN => Err(coinbase::Error::UndersizedScript)?,
            coinbase::MIN_COINBASE_SCRIPT_LEN..=coinbase::MAX_COINBASE_SCRIPT_LEN => (),
            (coinbase::MAX_COINBASE_SCRIPT_LEN + 1).. =>Err(coinbase::Error::OversizedScript)?,
        }
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,108 @@
+//! Types, constants, and functions related to coinbase transactions.
+
+use alloc::{fmt, vec::Vec};
+use zcash_script::{opcode::Evaluable, pv::push_value};
+
+/// The maximum length of the coinbase transaction script.
+pub const MAX_COINBASE_SCRIPT_LEN: usize = 100;
+
+/// The minimum length of the coinbase transaction script.
+pub const MIN_COINBASE_SCRIPT_LEN: usize = 2;
```
str4d, line 10: I confirmed that these match the consensus rules implemented in `zcashd`:
https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/main.cpp#L1754-L1755

zcash_transparent/src/bundle.rs
str4d: For the first point, I checked the consensus rules in `zcashd`:
https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/main.cpp#L5169-L5181

It specifically checks the encoded `scriptSig` bytes against `CScript() << nHeight`, i.e. it only checks the first `PushData`. So I think it is safe to interpret the `OP_0` pushed by `zcashd` as "empty miner data", but it would be a good idea to check both the commit history and Bitcoin Core (as I suspect that is where it comes from).

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// The maximum length of the encoded height in the coinbase transaction script.
+pub const MAX_COINBASE_HEIGHT_LEN: usize = 6;
```
str4d, line 13: I believe the origin of this number is:
- The maximum block height in Zcash is `u32::MAX`.
- The script number encoding of this is the little-endian `u32` encoding followed by an `0x00` byte.
- The `PushData` encoding will use `OP_5` because the byte length (5) fits in a single byte.

str4d (changes requested):
Reviewed 599485782aa99783da91d54cc2e0a554e0964716.

zcash_transparent/src/coinbase.rs
upbqdn: I'll document why the value is 6.

zcash_transparent/src/bundle.rs
```diff
@@ -258,6 +267,71 @@ impl TxIn<Authorized> {
...
+        // > description here is to be considered normative).
+        //
+        // <https://zips.z.cash/protocol/protocol.pdf#txnconsensus>
+        //
+        // ## Note
+        //
+        // The height is encoded in the `push_num` fn above.
+        //
+        // [BIP-34]: <https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki>
+        assert!(script_sig.is_empty());
```
upbqdn:
```suggestion
        assert!(!script_sig.is_empty());
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Encoded coinbase script is shorter than [`MIN_COINBASE_SCRIPT_LEN`].
+    UndersizedScript,
+    /// Encoded block height exceeds [`MAX_COINBASE_HEIGHT_LEN`].
+    OversizedHeight,
+    /// Encoded miner data exceeds [`MinerData::MAX_LEN`].
+    OversizedMinerData,
+    /// Creating a coinbase input for the genesis block is not supported.
+    GenesisInputNotSupported,
+    /// Coinbase transactions can contain only predefined inputs.
+    ExcessiveInputs,
```
daira:
```suggestion
    /// Coinbase transactions must contain only a dummy input.
    UnexpectedInputs,
```

zcash_transparent/src/builder.rs
```diff
@@ -392,6 +433,26 @@ impl TransparentBuilder {
...
+    ///
+    /// Returns an error if any inputs were added to the builder.
+    pub fn build_coinbase(
+        self,
+        height: BlockHeight,
+        miner_data: &MinerData,
+    ) -> Result<Bundle<Coinbase>, coinbase::Error> {
+        #[cfg(feature = "transparent-inputs")]
+        if !self.inputs.is_empty() {
+            return Err(coinbase::Error::ExcessiveInputs);
```
daira:
```suggestion
            return Err(coinbase::Error::UnexpectedInputs);
```

zcash_transparent/src/bundle.rs
daira: I concur that zcashd only checks that the script starts with the expected PushData. This is consistent with the [protocol spec](https://zips.z.cash/protocol/protocol-dark.pdf#txnconsensus), so I think this is fine as-is.

zcash_transparent/src/coinbase.rs
daira: This is not correct. Per the spec:
> For height in the range {1 .. 16}, the encoding is a single byte of value 0x50 + height. Otherwise, let heightBytes be the signed little-endian representation of height, using the minimum nonzero number of bytes such that the most significant byte is < 0x80. The length of heightBytes MUST be in the range {1 .. 5}. Then the encoding is the length of heightBytes encoded as one byte, followed by heightBytes itself.

So the encoding has length `2..=6`. But the height encoding is not a `u32`. In any case, the last two sentences quoted from the spec are sufficient here.

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,105 @@
...
+                "encoded coinbase script exceeds the limit of {MAX_COINBASE_SCRIPT_LEN} bytes",
+            ),
+            Error::UndersizedScript => write!(
+                f,
+                "encoded coinbase script is shorter than the minimum of {} bytes",
+                MIN_COINBASE_SCRIPT_LEN,
+            ),
+            Error::OversizedHeight => write!(
+                f,
+                "encoded block height exceeds the limit of {MAX_COINBASE_HEIGHT_LEN}",
```
daira:
```suggestion
                "encoded block height exceeds the limit of {MAX_COINBASE_HEIGHT_LEN} bytes",
```

zcash_transparent/src/coinbase.rs
```diff
@@ -0,0 +1,105 @@
...
+            ),
+            Error::OversizedMinerData => write!(
+                f,
+                "encoded miner data exceeds the limit of {}",
```
daira:
```suggestion
                "encoded miner data exceeds the limit of {} bytes",
```

daira (approved):
utACK with @schell. No blocking comments.

zcash_transparent/src/builder.rs
upbqdn: I was using `script::Code` originally: https://github.com/zcashfoundation/zcash_script/blob/d6dbbf4e69ea37a87132a652570d963662e88619/zcash_script/src/script/mod.rs#L315, but that wasn't `PartialEq` at that time. I had another PR for that, but ended up using `Vec` in the end.

zcash_transparent/src/coinbase.rs
upbqdn: The original idea was to have a type with a safe constructor. The way I used `MinerData` in Zebra was that I deserialized it from Zebra's config TOML file at startup, so the user would get an early error if anything was wrong with it. And then I'd pass it along all the way to `TxIn` here in `zcash_transparent`.

I don't see much sense in storing raw data in `MinerData`, so I'd drop it entirely instead. I'll open a parallel PR to this one with that.

upbqdn: Opened #2078.

nuttycom: @upbqdn I think that having a safe constructor makes sense; you can still have length-bounded raw data, and then you know internal to `TxIn::coinbase` that the length has already been checked. @str4d's complaint here is just that the encoding is being done in the wrong place.

Personally, I would *both* have `MinerData` store raw unencoded bytes, *and* I would use an explicit constructor instead of `TryFrom`.

nuttycom: Also, I think that not being able to provide miner data (as in #2078) is a worse option.

upbqdn: The problem is that I'd have to do the encoding multiple times, then: when constructing (and deserializing), and when using it here. Would that be fine? If I wanted to do the bound checking without encoding, I would end up almost reimplementing the encoding again.

zcash_transparent/src/bundle.rs
upbqdn: I rewrote this in #2078, although the logic is similar there.

I wanted to use this code to generate blocks from height one on Regtest. @str4d, note that the first 15 heights are encoded as a single byte, so we need some padding to satisfy this consensus rule for those heights:

> A coinbase transaction script MUST have length in {2 .. 100} bytes.

I'm using `miner_data` for that; `zcashd` does it with the extra `OP_0`.

zcash_transparent/src/builder.rs
upbqdn: Done in #2078.

zcash_transparent/src/bundle.rs
upbqdn: #2078 also uses only a single call for the length.

zcash_transparent/src/coinbase.rs
upbqdn: If #2078 doesn't work, then I can use a safe type for miner data again, but couple encoding with bound-checking. I can't think of another solution.

upbqdn: > and then you know internal to `TxIn::coinbase` that the length has already been checked.

Yup, this would be fine. I would only need to run through `push_value` again.

PR #2087, opened 2025-12-15
zcash_transparent: Add `zip48::FullViewingKey::derive_matching_account_priv_key`

str4d:


zcash_transparent/src/zip48.rs
```diff
@@ -229,6 +257,18 @@ impl FullViewingKey {
         }
     }
 
+    /// Returns the ZIP 48 coin type and account ID for this full viewing key.
+    fn coin_type_and_account(&self) -> (u32, AccountId) {
```
nuttycom, line 261: nonblocking: do we not have a `CoinType` type anywhere? If not, we should add one at some point.

zcash_transparent/src/zip48.rs
```diff
@@ -420,6 +482,20 @@ mod tests {
                 assert_eq!(&key.key_info_expression(&params), expected);
             }
 
+            for (i, seed) in (0..tv.key_information_vector.len()).zip(&seeds) {
+                if i < tv.key_information_vector.len() {
+                    assert!(matches!(
+                        fvk.derive_matching_account_priv_key(seed),
+                        Ok(Some(_)),
+                    ));
+                } else {
```
nuttycom, line 486: blocking: How do we ever get to this `else` clause? We zipped `0..tv.key_information_vector.len()` with `&seeds` but `zip` halts at the end of the shorter collection, so AFAICT we can never get an index that would trigger this `else`.

zcash_transparent/src/zip48.rs
nuttycom: nonblocking: Given the documentation added in this PR on line 262, we should probably change this to `NonEmpty`.

nuttycom (changes requested):
Reviewed 6e50c0d5cef81fd1129153e9fad97434a42b55cb

zcash_transparent/src/zip48.rs
str4d: The test vectors can use at most 16 seeds, but the number of participants in each test vector wallet is random (see https://github.com/zcash/zcash-test-vectors/pull/111 which still hasn't been reviewed).

What the test does is verify that the seeds expected to be involved are involved, and the seeds expected to not be involved are not involved.

str4d: Oh wait, I intended to change the bound on the iterator but forgot.

str4d: Also, the maximum number of participants is 15 (which the test vectors are bounded on in the TV PR), but this module's test vectors initialize `seeds` with 16 seeds, so for this test there is always guaranteed to be one seed that is never involved in any FVK.

nuttycom (approved):
utACK 7eb630e79cffb795359ff53fa73d14457d97bc26

str4d (commented):
utACK 572a91a0ff207d821016758f116c1923535ae62e (changes after my commits)

PR #2088, opened 2025-12-16
zcash_primitives: Fix transitivity of `zcash_primitives/std` & improve Builder type bound ergonomics.

nuttycom:


str4d (approved):
Re-utACK c3425f9c3c7f6deb20720bb78b18f35fbbed8edd

PR #2091, opened 2025-12-18
Fix a mis-merge in the `zcash_transparent` CHANGELOG

nuttycom:
The replacement of the [Unreleased] tag with the [0.6.3] line on the maintenance branch allowed a no-conflict merge to occur that resulted in CHANGELOG entries being in the wrong order on `main`. I have modified the `Unreleased` line for `zcash_transparent` and `zcash_primitives` to use `-PENDING` release versions to help prevent this sort of error in the future. In general, we should probably not use [Unreleased] for this reason.


str4d:
Post-hoc ACK

PR #2093, opened 2025-12-18
`zcash_keys`: Fix missing `std` feature dependencies.

nuttycom:
`zcash_keys/std` now implies `orchard?/std`, `sapling?/std`, `transparent/std`, and `zcash_address/std`. In addition, downstream crates of `zcash_keys` within this workspace now must explicitly enable `zcash_keys/std` if its features are required.

### DIFF
--- a/zcash_keys/CHANGELOG.md
+++ b/zcash_keys/CHANGELOG.md
@@ -16,8 +16,8 @@ workspace.
 ### Changed
 - MSRV is now 1.85.1.
 - Migrated to `orchard 0.12`, `sapling-crypto 0.6`.
-- `zcash_keys/std` now implies `orchard?/std`, `sapling?/std`,
-  `transparent/std`, and `zcash_address/std`.
+- The `std` feature flag now enables the equivalent flag on the dependencies
+  `orchard`, `sapling-crypto`, `zcash_transparent`, and `zcash_address`.
 
 ## [0.12.0] - 2025-10-13
 


Cargo.toml
```diff
@@ -39,7 +39,7 @@ equihash = { version = "0.2", path = "components/equihash", default-features = f
 zcash_address = { version = "0.10", path = "components/zcash_address", default-features = false }
 zcash_client_backend = { version = "0.21", path = "zcash_client_backend" }
 zcash_encoding = { version = "0.3", path = "components/zcash_encoding", default-features = false }
-zcash_keys = { version = "0.12", path = "zcash_keys" }
+zcash_keys = { version = "0.12", path = "zcash_keys", default-features = false  }
```
nuttycom, line 42: This change is perhaps not strictly required because none of the crates downstream of `zcash_keys` in this workspace have `no_std` builds available, but I think it's useful to do anyway because it's better for `std` to be visibly and explicitly enabled.

zcash_keys/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ workspace.
 ### Changed
 - MSRV is now 1.85.1.
 - Migrated to `orchard 0.12`, `sapling-crypto 0.6`.
+- `zcash_keys/std` now implies `orchard?/std`, `sapling?/std`,
+  `transparent/std`, and `zcash_address/std`.
```
str4d:
```suggestion
- The `std` feature flag now enables the equivalent flag on the dependencies
  `orchard`, `sapling-crypto`, `zcash_transparent`, and `zcash_address`.
```

PR #2094, opened 2025-12-19
Make `ChangeStrategy` values usable outside of the context of `zcash_client_backend`

nuttycom:
This exposes types and trait implementations from `zcash_transparent` and `zcash_client_backend` to make it easier to use the `ChangeStrategy` type in contexts outside of `zcash_client_backend` proposal construction.

### DIFF
--- a/zcash_transparent/src/builder.rs
+++ b/zcash_transparent/src/builder.rs
@@ -155,8 +155,8 @@ fn construct_script_sig(
     ])
 }
 
-/// Information required to spend a transparent coin; the secp256k public key when spending a p2pkh
-/// output, or the redeem script when spending a p2sh output.
+/// Information required to spend a transparent coin; the secp256k1 public key when spending a P2PKH
+/// output, or the redeem script when spending a P2SH output.
 #[cfg(feature = "transparent-inputs")]
 #[derive(Debug, Clone)]
 pub enum SpendInfo {


zcash_transparent/src/builder.rs
```diff
@@ -155,32 +155,57 @@ fn construct_script_sig(
...
-    kind: InputKind,
     utxo: OutPoint,
     coin: TxOut,
+    spend_info: SpendInfo,
 }
 
 #[cfg(feature = "transparent-inputs")]
 impl TransparentInputInfo {
+    /// Constructs a [`TransparentInputInfo`] value from its constituent parts.
+    pub fn new(utxo: OutPoint, coin: TxOut, spend_info: SpendInfo) -> Self {
```
str4d: I dislike exposing a constructor for `TransparentInputInfo` that doesn't enforce the correctness constraints that `Builder::add_input` and `Builder::add_p2sh_input` do. Instead, we should move the correctness checks from those methods into here, and then have those existing methods be thin wrappers around this one.

We could also change `Builder::add_input` to take `TransparentInputInfo` as its argument, and rename the current `Builder::add_input` to `Builder::add_p2pkh_input` (so it's explicitly a convenience function). That would then remove the need for the `TransparentInputInfo::into_parts` method.

zcash_transparent/src/builder.rs
```diff
@@ -155,32 +155,97 @@ fn construct_script_sig(
     ])
 }
 
+/// Information required to spend a transparent coin; the secp256k public key when spending a p2pkh
+/// output, or the redeem script when spending a p2sh output.
```
str4d:
```suggestion
/// Information required to spend a transparent coin; the secp256k1 public key when spending a P2PKH
/// output, or the redeem script when spending a P2SH output.
```
Or (because it's just restating the type signatures which are visible because enum variants are fully public):
```suggestion
/// Information required to spend a transparent coin.
```

str4d (dismissed):
utACK 77c422f1bd56400c9647f089c87f3776e16fd212 with nit

PR #2099, opened 2026-01-10
Fix db schema migration break due to `confirmed_unmined_at_height` column

### LINKED ISSUE #2096: zcash_client_sqlite migration fails with missing `confirmed_unmined_at_height` column
When migrating an old database with the latest librustzcash, I get this error:

> An error occurred while applying migration 9fa43ce0-a387-45d1-be03-57a3edc76d01 (Truncates away bad note commitment tree state for users whose wallets were broken by bad reorg handling.) up: no such column: confirmed_unmined_at_height.

The failing migration is `fix_broken_commitment_trees`.
The migration that adds the missed column is `tx_observation_height`.

Looking at [the migration dependency graph](https://github.com/zcash/librustzcash/blob/ade8b67e220522f34237281db3565402669d181b/zcash_client_sqlite/src/wallet/init/migrations.rs#L117-L127), we see that in fact `tx_observation_height` comes well _after_ `fix_broken_commitment_trees`. Thus, the former migration takes a dependency on the later migration.

The `fix_broken_commitment_trees` migration does not in fact directly mention the `confirmed_unmined_at_height` column. But it calls out to non-migration code (e.g. `wallet::truncate_to_height` in this case) which apparently assumes that the db has the latest schema, which of course it doesn't. 

IMO migrations really should not call into any non-migration code because of such risks. Any non-migration dependency could be copied into the migration script to 'freeze' the SQL that it executes.

AArnott:
The `fix_broken_commitment_trees` migration fails on databases that haven't yet applied `tx_observation_height`, which adds the `confirmed_unmined_at_height` column. The migration calls `wallet::truncate_to_height()`, which unconditionally references this column, causing "no such column" errors.

## Changes

- Modified `truncate_to_height()` to check for column existence via `PRAGMA table_info(transactions)` before updating `confirmed_unmined_at_height`
- Conditional SQL execution based on schema state, following the pattern established in `ufvk_support.rs` migration
- Added comprehensive tests to verify both code paths work correctly

```rust
let has_confirmed_unmined = {
    let mut stmt = conn.prepare("PRAGMA table_info(transactions)")?;
    let mut rows = stmt.query_map([], |row| {
        let col_name: String = row.get(1)?;
        Ok(col_name)
    })?;
    rows.any(|name| name.as_ref().map(|n| n == "confirmed_unmined_at_height").unwrap_or(false))
};

if has_confirmed_unmined {
    // Include confirmed_unmined_at_height in UPDATE
} else {
    // Omit column from UPDATE
}
```

## Testing

- ✅ `migrate_without_confirmed_unmined_at_height_column` - Tests the upgrade path from an older database schema (before `tx_observation_height`) to verify the migration works when the column doesn't exist
- ✅ `migrate_with_confirmed_unmined_at_height_column` - Tests the case when the column exists to ensure both branches of the conditional logic are covered
- ✅ All 47 migration tests in the suite pass

This ensures migrations work correctly regardless of database state while maintaining existing behavior when the column exists.

Fixes #2096

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
@@ -65,170 +65,8 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
             .optional()?
             .flatten();
 
-        if let Some(max_height) = max_block_height {
-            // This migration inlines the logic from `wallet::truncate_to_height` that was present
-            // at the time of the migration, before the `confirmed_unmined_at_height` column was
-            // added to the transactions table. This ensures the migration works on databases
-            // that have not yet applied the `tx_observation_height` migration.
-
-            // Determine a checkpoint to which we can rewind, if any.
-            #[cfg(not(feature = "orchard"))]
-            let truncation_height_query = r#"
-                SELECT MAX(height) FROM blocks
-                JOIN sapling_tree_checkpoints ON checkpoint_id = blocks.height
-                WHERE blocks.height <= :block_height
-            "#;
-
-            #[cfg(feature = "orchard")]
-            let truncation_height_query = r#"
-                SELECT MAX(height) FROM blocks
-                JOIN sapling_tree_checkpoints sc ON sc.checkpoint_id = blocks.height
-                JOIN orchard_tree_checkpoints oc ON oc.checkpoint_id = blocks.height
-                WHERE blocks.height <= :block_height
-            "#;
-
-            let truncation_height = transaction
-                .query_row(
-                    truncation_height_query,
-                    named_params! {":block_height": u32::from(max_height)},
-                    |row| row.get::<_, Option<u32>>(0),
-                )
-                .optional()?
-                .flatten()
-                .map_or_else(
-                    || {
-                        // If we don't have a checkpoint at a height less than or equal to the requested
-                        // truncation height, query for the minimum height to which it's possible for us to
-                        // truncate so that we can report it to the caller.
-                        #[cfg(not(feature = "orchard"))]
-                        let min_checkpoint_height_query =
-                            "SELECT MIN(checkpoint_id) FROM sapling_tree_checkpoints";
-                        #[cfg(feature = "orchard")]
-                        let min_checkpoint_height_query = "SELECT MIN(sc.checkpoint_id)
-                             FROM sapling_tree_checkpoints sc
-                             JOIN orchard_tree_checkpoints oc
-                             ON oc.checkpoint_id = sc.checkpoint_id";
-
-                        let min_truncation_height = transaction
-                            .query_row(min_checkpoint_height_query, [], |row| {
-                                row.get::<_, Option<u32>>(0)
-                            })
-                            .optional()?
-                            .flatten()
-                            .map(BlockHeight::from);
-
-                        Err(WalletMigrationError::from(
-                            SqliteClientError::RequestedRewindInvalid {
-                                safe_rewind_height: min_truncation_height,
-                                requested_height: max_height,
-                            },
-                        ))
-                    },
-                    |h| Ok(BlockHeight::from(h)),
-                )?;
-
-            let last_scanned_height =
-                transaction.query_row("SELECT MAX(height) FROM blocks", [], |row| {
-                    let h = row.get::<_, Option<u32>>(0)?;
-
-                    Ok(h.map_or_else(
-                        || {
-                            self.params
-                                .activation_height(NetworkUpgrade::Sapling)
-                                .expect("Sapling activation height must be available.")
-                                - 1
-                        },
-                        BlockHeight::from,
-                    ))
-                })?;
-
-            // Delete from the scanning queue any range with a start height greater than the
-            // truncation height, and then truncate any remaining range by setting the end
-            // equal to the truncation height + 1. This sets our view of the chain tip back
-            // to the retained height.
-            transaction.execute(
-                "DELETE FROM scan_queue
-                WHERE block_range_start >= :new_end_height",
-                named_params![":new_end_height": u32::from(truncation_height + 1)],
-            )?;
-            transaction.execute(
-                "UPDATE scan_queue
-                SET block_range_end = :new_end_height
-                WHERE block_range_end > :new_end_height",
-                named_params![":new_end_height": u32::from(truncation_height + 1)],
-            )?;
-
-            // Mark transparent utxos as un-mined. Since the TXO is now not mined, it would ideally be
-            // considered to have been returned to the mempool; it _might_ be spendable in this state, but
-            // we must also set its max_observed_unspent_height field to NULL because the transaction may
-            // be rendered entirely invalid by a reorg that alters anchor(s) used in constructing shielded
-            // spends in the transaction.
-            transaction.execute(
-                "UPDATE transparent_received_outputs
-                 SET max_observed_unspent_height = CASE
-                    WHEN tx.mined_height <= :height THEN :height
-                    ELSE NULL
-                 END
-                 FROM transactions tx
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3306,12 +3306,35 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    // Check if the confirmed_unmined_at_height column exists (added by tx_observation_height migration)
+    let has_confirmed_unmined = {
+        let mut stmt = conn.prepare("PRAGMA table_info(transactions)")?;
+        let mut rows = stmt.query_map([], |row| {
+            let col_name: String = row.get(1)?;
+            Ok(col_name)
+        })?;
+        rows.any(|name| {
+            name.as_ref()
+                .map(|n| n == "confirmed_unmined_at_height")
```
str4d: This column was added in the `tx_observation_height` migration, which was released in `zcash_client_sqlite 0.18.5`. Instead of being merged directly to `main` (which is at `zcash_client_sqlite 0.19.0`), the fix should be released as an 0.18.6 bugfix release.

nuttycom: This also isn't the correct fix; instead, the prior implementation of truncate_to_height needs to be copied into the migration where it is used. That's our standard practice for this kind of breakage; otherwise, we would start polluting lots of methods with structural checks for the sake of old migrations.

AArnott: Thanks for the feedback. I'll fix it.

AArnott: Fix updated and rebased onto zcash_client_sqlite-0.18.10.

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -61,14 +65,170 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
-                &self.params,
-                #[cfg(feature = "transparent-inputs")]
-                &GapLimits::default(),
-                h,
+        if let Some(max_height) = max_block_height {
+            // This migration inlines the logic from `wallet::truncate_to_height` that was present
+            // at the time of the migration, before the `confirmed_unmined_at_height` column was
+            // added to the transactions table. This ensures the migration works on databases
+            // that have not yet applied the `tx_observation_height` migration.
+
```
str4d: We do not normally inline the code like this (which is harder to maintain as it makes it more difficult to distinguish migration code from frozen general logic). Instead we copy in the function wholesale and then change the migration to reference the copied function. See 52549a80d4e9da83cd7bdacc2c7104a6b004852a for an example.

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -81,10 +241,200 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
 mod tests {
-    use crate::wallet::init::migrations::tests::test_migrate;
+    use rusqlite::params;
+    use tempfile::NamedTempFile;
+    use zcash_protocol::consensus::Network;
+
+    use crate::{
+        testing::db::{test_clock, test_rng},
+        wallet::init::{
+            migrations::{support_legacy_sqlite, tx_observation_height},
```
str4d: Undo this import change:
```suggestion
            migrations::{support_legacy_sqlite, tx_observation_height, tests::test_migrate},
```

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -81,10 +241,200 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+        },
+        WalletDb,
+    };
+
+    use super::MIGRATION_ID;
 
     #[test]
     fn migrate() {
-        test_migrate(&[super::MIGRATION_ID]);
+        crate::wallet::init::migrations::tests::test_migrate(&[MIGRATION_ID]);
```
str4d:
```suggestion
        test_migrate(&[MIGRATION_ID]);
```

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -79,12 +77,371 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+/// This is a copy of [`crate::wallet::truncate_to_height`] as of the expected database
+/// state corresponding to this migration. It is duplicated here as later updates to the
+/// database schema require incompatible changes to `truncate_to_height` (specifically,
+/// the addition of the `confirmed_unmined_at_height` column in the `tx_observation_height`
+/// migration).
+fn truncate_to_height<P: consensus::Parameters>(
+    conn: &rusqlite::Transaction,
+    params: &P,
+    max_height: BlockHeight,
+) -> Result<BlockHeight, WalletMigrationError> {
```
str4d, line 89: Nit: this could have remained `Result<BlockHeight, SqliteClientError>` to avoid changes to the frozen code (the `?` in the migration would automatically handle the conversion to `WalletMigrationError`).

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -79,12 +77,371 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+                    .optional()?
+                    .flatten()
+                    .map(BlockHeight::from);
+
+                Err(WalletMigrationError::from(
+                    SqliteClientError::RequestedRewindInvalid {
+                        safe_rewind_height: min_truncation_height,
+                        requested_height: max_height,
+                    },
+                ))
```
str4d, line 141: That would revert this to:
```suggestion
                Err(SqliteClientError::RequestedRewindInvalid {
                    safe_rewind_height: min_truncation_height,
                    requested_height: max_height,
                })
```

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -79,12 +77,371 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
+    // affected block records from the database.
+    if truncation_height < last_scanned_height {
+        // Truncate the note commitment trees
+        let mut wdb = WalletDb {
+            conn: SqlTransaction(conn),
+            params: params.clone(),
+            clock: (),
+            rng: (),
+            #[cfg(feature = "transparent-inputs")]
+            gap_limits: GapLimits::default(),
```
str4d, line 212: Nit: we would also normally leave this as a passed-in parameter, again to minimize changes to the frozen code.

zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -79,12 +77,371 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
     }
 }
 
+/// This is a copy of [`crate::wallet::truncate_to_height`] as of the expected database
+/// state corresponding to this migration. It is duplicated here as later updates to the
+/// database schema require incompatible changes to `truncate_to_height` (specifically,
+/// the addition of the `confirmed_unmined_at_height` column in the `tx_observation_height`
+/// migration).
+fn truncate_to_height<P: consensus::Parameters>(
```
str4d, line 85: Other than as noted in my prior comments, I verified that this copy matches the version in 58913805e1b991bc4ee7a040f5d7ce2994f9f16a immediately prior to the incompatible change.

Issue #2101, opened 2026-01-19
bug: Github pages deployment fails due to cargo-doc errors


nullcopy:
Expected docs to be found at https://zcash.github.io/librustzcash, but it looks like the build job isn't functioning or has an error.

nullcopy:
Looks like a build error with `cargo doc`:
```bash
$ cargo doc --open
warning: unresolved link to `Input::sign`
  --> zcash_transparent/src/pczt.rs:63:60
   |
63 |     /// This is used by Signers to apply signatures with [`Input::sign`].
   |                                                            ^^^^^^^^^^^ the struct `Input` has no field or associated item named `sign`
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: `zcash_transparent` (lib doc) generated 1 warning
   Compiling zcash_client_backend v0.21.0 (/home/user/Projects/zcash/librustzcash/zcash_client_backend)
warning: unresolved link to `recipient_address`
   --> components/zip321/src/lib.rs:152:17
    |
152 |     ///   and [`recipient_address`] is not a shielded address, the wallet should report an error.
    |                 ^^^^^^^^^^^^^^^^^ no item named `recipient_address` in scope
    |
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

   Compiling zcash_client_memory v0.0.0 (/home/user/Projects/zcash/librustzcash/zcash_client_memory)
warning: `zip321` (lib doc) generated 1 warning
 Documenting zcash_client_backend v0.21.0 (/home/user/Projects/zcash/librustzcash/zcash_client_backend)
error: unresolved link to `WalletWrite::import_standalone_transparent_pubkey`
    --> zcash_client_backend/src/data_api.rs:2765:37
     |
2765 | /// transparent keys imported via [`WalletWrite::import_standalone_transparent_pubkey`] break this
     |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `WalletWrite` has no associated item named `import_standalone_transparent_pubkey`
     |
note: the lint level is defined here
    --> zcash_client_backend/src/lib.rs:60:9
     |
60   | #![deny(rustdoc::broken_intra_doc_links)]
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #2103, opened 2026-01-20
fix: Fix github-pages build

nullcopy:
cargo-doc was failing due to doc references to modules behind feature gates. This commit enables all features when building docs for the gitbook.

Resolves https://github.com/zcash/librustzcash/issues/2101


nuttycom:
Looks like this might need to be rebased after the merge of the `cargo vet` fix.

PR #2117, opened 2026-01-27
eip681: Simplify and fix bugs

str4d:


components/eip681/src/parse.rs
```diff
@@ -826,29 +816,28 @@ impl Parameter {
 
     pub fn parse(i: &str) -> nom::IResult<&str, Self, ParseError<'_>> {
         // Parse the key blob
-        let (i, key_blob) = nom::bytes::complete::take_till1(|c| c == '=')(i)?;
-        let (i, _) = nom::bytes::complete::tag("=")(i)?;
+        let (i, key_blob) = terminated(take_till1(|c| c == '='), tag("="))(i)?;
```
nuttycom, line 954: Isn't the `take_till1` here superfluous? This looks to me like it could just be `terminated(anychar, tag("="))`.

components/eip681/src/parse.rs
```diff
@@ -902,17 +891,12 @@ impl Parameters {
...
-        let mut params = vec![];
-        for blob in blobs.into_iter() {
-            let (j, param) = Parameter::parse(blob)?;
-            snafu::ensure!(j.is_empty(), UnexpectedLeftoverInputSnafu { input: j });
-            params.push(param);
-        }
-        Ok((i, Parameters(params)))
+        separated_list0(
+            tag("&"),
+            map_parser(take_till1(|c| c == '&'), all_consuming(Parameter::parse)),
```
nuttycom, line 1031: Similar question here about `take_till1` - what is it needed for?

components/eip681/src/parse.rs
str4d: [`terminated()` is greedy](https://docs.rs/nom/7.1.3/nom/sequence/fn.terminated.html):
> Gets an object from the first parser, then matches an object from the second parser and discards it.

So we need the first parser to stop just before the `=`.

str4d:
[Force-pushed](https://github.com/zcash/librustzcash/compare/486caedb3dcb746aec26e54de12c0727c208862b..0767e5e833d8de744f7b94ebc68d0f14d40591a2) to rebase on `main`. The first commit (removing commented-out unused code) was duplicated in other PRs; everything else was still relevant.

components/eip681/src/parse.rs
str4d: [`separated_list0` is greedy](https://docs.rs/nom/7.1.3/nom/multi/fn.separated_list0.html):
> Alternates between two parsers to produce a list of elements.

So we need two mutually-exclusive parsers, which I achieve here by alternating between `not "&"` and `"&"`, and then re-parsing the fomer as a `Parameter`.

components/eip681/src/parse.rs
str4d: Note that `take_till(|c| c == '=')` is identical to `is_not("=")`, but we can't use the latter here because we want to ensure that we match on at least one character (i.e the key blob is not empty), because the empty key is not valid.

PR #2120, opened 2026-01-28
fix: eip681 ethereum address is case-sensitive

### LINKED ISSUE #2118: EIP-681 parser hexdigits needs to preserve casing to adhere to ERC-55 checksumming
EVM addresses parsed with the EIP-681 parser crate don't roundtrip because they drop hexadecimal character casing. Check-summed EVM addresses need to preserve case so the checksum matches the ERC-55 spec. 

Ideally this would also validate the parsed addresses with ERC-55.

PR:
* #2120

schell:
These changes make `HexDigits` case sensitive, include an ERC-55 function and change the name of `EthereumAddress` to `EthereumAddressOrEnsName`.

Fixes #2118
Fixes COR-361

### DIFF
--- a/components/eip681/src/parse.rs
+++ b/components/eip681/src/parse.rs
@@ -1809,7 +1809,7 @@ mod test {
             // All lower
             "d1220a0cf47c7b9be7a2e6ba89f429762e7b9adb",
             // All upper
-            "d1220a0cf47c7b9be7a2e6ba89f429762e7b9adb",
+            "D1220A0CF47C7B9BE7A2E6BA89F429762E7B9ADB",
         ];
 
         for addy in borked_cases {


components/eip681/src/parse.rs
```diff
@@ -122,9 +217,86 @@ impl core::fmt::Display for HexDigits {
...
+        snafu::ensure!(
+            self.places.len() == 40,
+            IncorrectEthAddressLenSnafu {
+                len: self.places.len()
+            }
+        );
+
+        // If all the digits are lowercase or all are uppercase, we can validate anything besides
+        // the length.
+        if self.is_all_lowercase() {
```
nullcopy: I'm curious if it is correct to skip checksum validation here in these all upper/lower-case scenarios. Although not likely, isn't it possible to have an ERC55 checksum which yields all upper or all lower characters? I think we should still attempt to validate the checksum, and only IFF the checksum fails AND the address is all upper/lower do we warn that the address was not validated.

components/eip681/src/parse.rs
```diff
@@ -1568,8 +1752,9 @@ mod test {
     fn test_vectors_eip_681() {
         let input = "ethereum:0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359?value=2.014e18";
         let (_, seen) = EthereumTransactionRequest::parse(input).unwrap();
+        // Ensure the address is case-sensitive
```
nullcopy, line 1763: IMO it would be good to include the ERC55 test vectors here: https://eips.ethereum.org/EIPS/eip-55

components/eip681/src/parse.rs
daira: ERC-55 checksum validation is necessarily probabilistic because it could be that there are insufficient `a`-`f` nibbles in the address. The chance of all of the `a`-`f` nibbles being uppercase or all of them being lowercase is just another similar low-probability failure case (which all compressing checksums have to some extent).

[EIP-55](https://eips.ethereum.org/EIPS/eip-55) is a woeful underspecification. It would never pass muster as a Zcash specification: the very first review comment would be "what are the conformance requirements on decoders?" There actually aren't any; what is supposed to pass for a specification is just pseudocode for an encoder, not a decoder.

daira: I calculated the chance of all of the `a`-`f` nibbles being uppercase or all of them being lowercase.
```python
#!/bin/env python

from scipy.stats import binom

def calc():
    """
    Calculate the probability that the letters `a`-`f` in an Ethereum address
    are either all-uppercase or all-lowercase.
    """

    # The number of letters has a binomial distribution with these parameters:
    n = 40
    p = 6.0/16

    # The probability of k letters being all uppercase or all lowercase is (1/2)^{k-1}.
    def same_case(k): return 0.5**(k-1)

    # binom.expect: <https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binom.html#scipy-stats-binom>
    return binom.expect(same_case, args=(n, p))

if __name__ == "__main__":
    print(calc())
```
```shell
$ ./eip681.py 
0.0004942706034105575
```

components/eip681/src/parse.rs
```diff
@@ -122,9 +217,86 @@ impl core::fmt::Display for HexDigits {
...
+    /// Returns whether all alphabetic digits are lowercase.
+    pub fn is_all_lowercase(&self) -> bool {
+        for digit in self.places.iter() {
+            match digit {
+                CaseSensitiveHexDigit::A { is_upper: true } => return false,
+                CaseSensitiveHexDigit::B { is_upper: true } => return false,
+                CaseSensitiveHexDigit::C { is_upper: true } => return false,
+                CaseSensitiveHexDigit::D { is_upper: true } => return false,
+                CaseSensitiveHexDigit::E { is_upper: true } => return false,
+                CaseSensitiveHexDigit::F { is_upper: true } => return false,
```
daira: This could just match on the characters, and would be much simpler:
```suggestion
                'A'..='F' => return false,
```

components/eip681/src/parse.rs
```diff
@@ -93,27 +94,121 @@ pub fn parse_min(i: &str, min: usize, is_hex: bool) -> nom::IResult<&str, Vec<u8
...
+            'B' => CaseSensitiveHexDigit::B { is_upper: true },
+            'C' => CaseSensitiveHexDigit::C { is_upper: true },
+            'D' => CaseSensitiveHexDigit::D { is_upper: true },
+            'E' => CaseSensitiveHexDigit::E { is_upper: true },
+            'F' => CaseSensitiveHexDigit::F { is_upper: true },
+            character => NotAHexDigitSnafu { character }.fail()?,
+        })
+    }
+}
+
```
daira, line 193: I'm not sure what the benefit is of representing a parsed as this enum rather than as the original character in `[0-9a-fA-F]`? I don't think you really need this enum at all.

```suggestion
```

components/eip681/src/parse.rs
```diff
@@ -93,27 +94,121 @@ pub fn parse_min(i: &str, min: usize, is_hex: bool) -> nom::IResult<&str, Vec<u8
...
+/// Zero or more consecutive and case-sensitive hexadecimal digits.
 ///
 /// ```abnf
 /// *HEXDIG
 /// ```
 #[derive(Clone, Debug, PartialEq)]
 pub struct HexDigits {
-    places: Vec<u8>,
+    places: Vec<CaseSensitiveHexDigit>,
```
daira, line 201:
```suggestion
    places: Vec<char>,
```

components/eip681/src/parse.rs
schell: I cargo-culted this part of the logic from @str4d's PR here https://github.com/zcash/zcash-devtool/pull/98/changes#diff-0493ab6feeddc01f1f98d0199b3ffe7ccf23d567baf56711b458bf58f9193ec3R49-R75.

@daira, I'm not sure if your comments are in support of validating the all upper/lower addresses or not, but @nullcopy's comment seems fair, and it's not a difficult change.

components/eip681/src/parse.rs
schell: Restricting the enum to only valid values makes the invalid case of anything _not_ `a..=f, A..=F, 0..=9` unrepresentable, at the cost of being verbose. Maybe the rigor is unnecessary, though if the innards of this type ever turned public, there's no guarantee any constructed value is valid.

schell: There might be a memory concern as well, not a big one, but `char` is 4x bigger than `u8`.

schell: Ok, so the internal representation I've set up takes 2 bytes, so it's only half as big as a `char`, but it's still an improvement.

nullcopy (approved):
utACK

components/eip681/src/parse.rs
schell: You've convinced me that it should only take 1 byte! And having a variant for each letter (uppercase and lowercase) does that as well as makes checking for uppers and lowers easier by using ranges.

PR #2122, opened 2026-01-29
zcash_client_backend: Enable connections to Tor hidden services

emersonian:
Zec.rocks maintains a few .onion lightwalletd instances, but they are not currently possible to connect to using librustzcash's built-in Tor support because Arti is not built with the onion-service-client feature enabled.

[zcash-devtool now supports SOCKS connections](https://github.com/zcash/zcash-devtool/pull/143), so it is possible to connect to .onion addresses with that functionality using an external Tor process.

This PR will make it possible to connect to .onion addresses using the built-in Arti functionality as well.

The primary advantage of using Tor hidden services is that server operators never get the opportunity to log IP addresses. The second advantage is that it greatly simplifies self-hosting your own lightwalletd, as Tor traverses NAT amazingly, and no SSL certificate or domain name registration is required.


Cargo.toml
```diff
@@ -170,7 +170,7 @@ incrementalmerkletree-testing = "0.3"
 # Tor
 # - `arti-client` depends on `rusqlite`, and a version mismatch there causes a compilation
 #   failure due to incompatible `libsqlite3-sys` versions.
-arti-client = { version = "0.35", default-features = false, features = ["compression", "rustls", "tokio"] }
+arti-client = { version = "0.35", default-features = false, features = ["compression", "onion-service-client", "rustls", "tokio"] }
```
str4d, line 173: This change is completely unnecessary here, and instead forces us to deal with extra transitive dependencies during development.

Instead, applications using `zcash_client_backend` should themselves add to their own `Cargo.toml` file:
```toml
arti-client = { version = "0.35", default-features = false, features = ["onion-service-client"] }
```
(and bump the version whenever `zcash_client_backend` bumps its dependency).

emersonian: With that rationale why include Arti at all in librustzcash? Supporting .onions is a very core part of using Tor, why ever exit the Tor network.

str4d: I think you are confused. We *already* support .onion addresses; there is nothing that needs to change in `zcash_client_backend` for that.

This statement from your OP is simply incorrect:
> they are not currently possible to connect to using librustzcash's built-in Tor support because Arti is not built with the onion-service-client feature enabled.

The choice to build `arti-client` (or any other library) with or without a given feature is *always* the responsibility of the application (or whatever part of the stack is building the final Rust binaries, like the Rust side of an FFI). Rust libraries should only enable feature flags that they actually require.

> why include Arti at all in librustzcash?

Because I wanted to provide a nice Zcash-oriented API on top of `arti-client`, to make it easier to use. That API does not care about, or require knowledge of, whether the lightwallet server is accessible via .onion or via an exit node. But it is also entirely optional: anyone can instead use `arti-client` directly to do the same thing, or use a separate Tor client's SOCKS proxy, or whatever they prefer. The API here is purely for convenience.

PR #2123, opened 2026-01-29
CI fixes for `maint/zcash_client_sqlite-0.18.x`

str4d:


zcash_client_sqlite/src/wallet/init/migrations/fix_broken_commitment_trees.rs
```diff
@@ -180,10 +191,7 @@ fn truncate_to_height<P: consensus::Parameters>(
     // spends in the transaction.
     conn.execute(
         "UPDATE transparent_received_outputs
-         SET max_observed_unspent_height = CASE
-            WHEN tx.mined_height <= :height THEN :height
-            ELSE NULL
-         END
+         SET max_observed_unspent_height = CASE WHEN tx.mined_height <= :height THEN :height ELSE NULL END
```
nuttycom, line 194: Why is this reformatting part of this PR?

str4d: Because I could not do it in #2099 due to that PR being opened from a GitHub org. The `rustfmt` change was also due to that PR but also caused CI to fail.

PR #2130, opened 2026-02-02
CI: Add `cargo-mutants` workflow

str4d:
Closes COR-695.

### DIFF
--- a/.github/workflows/mutants.yml
+++ b/.github/workflows/mutants.yml
@@ -1,0 +1,109 @@
+name: Mutation testing
+
+env:
+  CARGO_TERM_COLOR: always
+
+on:
+  # Run weekly on Sundays at 2 AM UTC
+  schedule:
+    - cron: "0 2 * * 0"
+  # Allow manual triggering
+  workflow_dispatch:
+    inputs:
+      package:
+        description: "Package to test (select 'all' for all packages)"
+        required: false
+        default: "all"
+        type: choice
+        options:
+          - all
+          - eip681
+          - equihash
+          - f4jumble
+          - pczt
+          - zcash_address
+          - zcash_client_backend
+          - zcash_client_sqlite
+          - zcash_encoding
+          - zcash_extensions
+          - zcash_history
+          - zcash_keys
+          - zcash_primitives
+          - zcash_proofs
+          - zcash_protocol
+          - zcash_transparent
+          - zip321
+
+permissions:
+  contents: read
+
+# Only run one mutation test at a time
+concurrency:
+  group: mutation-testing
+  cancel-in-progress: false
+
+jobs:
+  mutants:
+    name: Mutation testing of ${{ matrix.package }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - eip681
+          - equihash
+          - f4jumble
+          - pczt
+          - zcash_address
+          - zcash_client_backend
+          - zcash_client_sqlite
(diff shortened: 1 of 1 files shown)


.github/workflows/mutants.yml
```diff
@@ -0,0 +1,109 @@
...
+          - zcash_keys
+          - zcash_primitives
+          - zcash_proofs
+          - zcash_protocol
+          - zcash_transparent
+          - zip321
+        # Only filter if a specific package was requested via workflow_dispatch
+        # For push/pull_request events, inputs.package is undefined, so default to 'all'
+        exclude:
+          - package: ${{ (github.event.inputs.package || 'all') != 'all' && (github.event.inputs.package || 'all') != 'eip681' && 'eip681' || 'NONE' }}
```
nuttycom: I'm confused by the precedence of `&&` and `||` here with respect to the `|| 'NONE'` at the end. I can't make this typecheck in my head; we have (bool && bool && String || String). What does it mean?

str4d: I took this logic from https://github.com/RustCrypto/KEMs/pull/216 and AIUI the type resolves to `String`. But I can't actually find anything in the GitHub docs that specifies the output of an expression, so we could rewrite this using [the `case` function](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case):
```suggestion
          - package: ${{ case((github.event.inputs.package || 'all') != 'all' && (github.event.inputs.package || 'all') != 'eip681', 'eip681', 'NONE') }}
```
although that still relies on the same `String || String -> String` expression evaluation logic, just not in a nested fashion.

nuttycom: So, what I don't understand is, if `'eip681'` is truthy and `&&` has higher precedence than `||`, what does `(True && True && 'eip681')` resolve to? does that return `'eip681'`? If that's the case, your `case` modification` doesn't return the same value; it would always return the first truthy branch, which is `'eip681'`. So I guess that as-written this is okay; if anything, it would just be helpful to parenthesize it: 
```suggestion
          - package: ${{ ((github.event.inputs.package || 'all') != 'all' && (github.event.inputs.package || 'all') != 'eip681' && 'eip681') || 'NONE' }}
```

nuttycom (approved):
utACK with minor suggestion for duck-typing clarity.

.github/workflows/mutants.yml
str4d: > `&&` has higher precedence than `||`,

I think it's that they have the same precedence and thus are evaluated left-to-right, with the first "truthy" value being returned (as opposed to `true` value which the boolean sub-expressions evaluate to). It would be nice if GitHub documented this better.

str4d: Aha, it's fake ternary: https://github.com/actions/runner/issues/409

The correct semantic parsing of this is:
```yaml
(github.event.inputs.package || 'all') != 'all' && (github.event.inputs.package || 'all') != 'eip681'
&& 'eip681'
|| 'NONE'
```

And from that same thread, it turns out the `case` function was added four days ago to enable more-than-ternary conditionals.

str4d: It's "fake ternary" because if the `<yes>` side of the ternary happens to be the empty string, then it is considered `false` and thus the expression always evaluates to `<no>` regardless of `<condition>`. This isn't an issue for us with hard-coded output strings, but the new `case` function would also avoid this.

daira (approved):
utACK

PR #2134, opened 2026-02-05
feat: COR-696 EIP-681 high level API

schell:
These changes add a high-level API for the `eip681` parser that also does validation.

Depends on 
* #2133 (rebase after that merges)


schell:
It looks like `proptest` caught something... 👷

schell:
But of course that's a lot of work, and arguably out of scope when we really only care about `uint256` values for the purpose of transaction requests.

PR #2140, opened 2026-02-06
zcash_client_backend: Refactor low-level wrappers

str4d:


zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -1026,3 +1006,59 @@ where
...
+        <DbT as LowLevelWalletRead>::AccountId,
+        TransparentKeyScope,
+    ),
+) -> Result<(), <DbT as LowLevelWalletRead>::Error>
+where
+    DbT: LowLevelWalletWrite,
+{
+    #[cfg(feature = "transparent-inputs")]
+    for (received_t_output, key_scope) in &outputs.received {
+        let (account_id, _) = put_received_output(wallet_db, received_t_output)?;
```
nuttycom, line 1034: I'm not sure why this is handled via a closure; what do you anticipate invoking here other than `wallet_db.put_transparent_output`?

nuttycom (approved):
utACK with question

zcash_client_backend/src/data_api/ll/wallet.rs
str4d: I've done this with a closure for closer API symmetry with `put_shielded_outputs`; in my ideal end state the per-pool logic is unified inside `store_decrypted_tx`, which requires a bit more massaging (as currently some of the logic in `put_shielded_outputs` maps to logic in `detect_wallet_transparent_outputs`, rather than `put_transparent_outputs` where it should end up).

str4d: To put this another way, I'm using "function with closure arguments" as a stepping stone to figuring out what a trait interface might need to look like, without forcing everything to fit into traits right now. And it might be that the correct thing to do is alter the `LowLevelWalletWrite` trait to unify these `put_*` methods instead of having per-pool APIs, but I don't want to try that before getting the refactors in place.

PR #2144, opened 2026-02-11
Implement tracking of ZIP 48 transparent multisig accounts

### LINKED ISSUE #2197: zcash_client_backend: ZIP 48 account + coin tracking
In service to https://github.com/zcash/librustzcash/issues/1361, we need to support importing ZIP 48 wallets (deriving addresses and tracking balances)

nullcopy:
This PR implements support for balance tracking of ZIP 48 transparent multisig accounts. This is a first step towards general P2SH account tracking.

I wanted to get feedback on this, before deciding how to approach general P2SH account support.

resolves https://github.com/zcash/librustzcash/issues/2197
resolves COR-1068

### DIFF
--- a/zcash_client_backend/Cargo.toml
+++ b/zcash_client_backend/Cargo.toml
@@ -187,11 +187,24 @@ transparent-inputs = [
     "zcash_primitives/transparent-inputs",
 ]
 
-# Enables operations that permit arbitrary transparent pubkeys to be associated
-# with an account in the wallet. Applications that rely on this feature must
-# maintain spending keys corresponding to all imported pubkeys for any account
-# from which it permits spending.
+## Enables operations that permit arbitrary transparent pubkeys to be associated
+## with an account in the wallet.
+##
+## Applications that enable this feature **MUST** also enable the
+## `transparent-inputs` feature of the crate they are using that implements the
+## `zcash_client_backend` traits.
+##
+## Applications that rely on this feature must maintain spending keys
+## corresponding to all imported pubkeys for any account from which it permits
+## spending.
 transparent-key-import = ["transparent-inputs"]
+
+## Enables tracking of ZIP 48 transparent multisig wallets.
+##
+## Applications that enable this feature **MUST** also enable the
+## `transparent-inputs` feature of the crate they are using that implements the
+## `zcash_client_backend` traits.
+zip-48 = ["transparent-inputs"]
 
 ## Enables receiving and spending Orchard funds.
 orchard = ["dep:orchard", "dep:pasta_curves", "zcash_keys/orchard"]
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -489,6 +489,10 @@ pub enum AccountSource {
         purpose: AccountPurpose,
         key_source: Option<String>,
     },
+
+    /// A ZIP 48 transparent multisig account.
+    #[cfg(feature = "zip-48")]
+    Zip48,
 }
 
 impl AccountSource {
@@ -500,6 +504,8 @@ impl AccountSource {
                 purpose: AccountPurpose::Spending { derivation },
                 ..
             } => derivation.as_ref(),
+            #[cfg(feature = "zip-48")]
+            AccountSource::Zip48 => None,
             _ => None,
         }
     }
@@ -509,6 +515,8 @@ impl AccountSource {
         match self {
             AccountSource::Derived { key_source, .. } => key_source.as_ref().map(|s| s.as_str()),
             AccountSource::Imported { key_source, .. } => key_source.as_ref().map(|s| s.as_str()),
+            #[cfg(feature = "zip-48")]
+            AccountSource::Zip48 => None,
         }
     }
 }
@@ -549,6 +557,8 @@ pub trait Account {
                 derivation: Some(derivation.clone()),
             },
             AccountSource::Imported { purpose, .. } => purpose.clone(),
+            #[cfg(feature = "zip-48")]
+            AccountSource::Zip48 => AccountPurpose::ViewOnly,
         }
     }
 
@@ -2744,6 +2754,10 @@ impl AccountBirthday {
 /// - [`WalletWrite::create_account`]
 /// - [`WalletWrite::import_account_hd`]
 /// - [`WalletWrite::import_account_ufvk`]
+#[cfg_attr(
+    feature = "zip-48",
+    doc = "/// - [`WalletWrite::import_account_zip48_multisig`]"
+)]
 ///
 /// All of these methods take an [`AccountBirthday`]. The birthday height is defined as
 /// the minimum block height that will be scanned for funds belonging to the wallet. If
@@ -2949,6 +2963,29 @@ pub trait WalletWrite: WalletRead {
         purpose: AccountPurpose,
         key_source: Option<&str>,
     ) -> Result<Self::Account, Self::Error>;
+
+    /// Imports a ZIP 48 transparent multisig wallet for tracking.
+    ///
+    /// This creates a new account of kind `account_kind = 2` (ZIP 48 multisig) with the
+    /// provided full viewing key. The account can track received funds and derive addresses.
+    ///
+    /// # Parameters
+    /// - `name`: A human-readable name for the account.
+    /// - `fvk`: The ZIP 48 full viewing key containing threshold and participant public keys.
+    /// - `birthday`: The account birthday, used for determining scan start height.
+    ///
+    /// Returns details about the imported account.
+    #[cfg(feature = "zip-48")]
+    fn import_account_zip48_multisig(
+        &mut self,
+        _name: &str,
+        _fvk: &transparent::zip48::FullViewingKey,
+        _birthday: &AccountBirthday,
+    ) -> Result<Self::Account, Self::Error> {
+        unimplemented!(
+            "WalletWrite::import_account_zip48_multisig must be overridden for wallets to use the `zip-48` feature"
+        )
+    }
 
     /// Deletes the specified account, and all transactions that exclusively involve it, from the
     /// wallet database.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -3069,6 +3069,56 @@ pub trait WalletWrite: WalletRead {
         request: UnifiedAddressRequest,
     ) -> Result<Option<UnifiedAddress>, Self::Error>;
 
+    /// Generates, persists, and marks as exposed the next available address for a ZIP 48
+    /// multisig account.
+    ///
+    /// The address and its redeem script are derived from the account's full viewing key
+    /// at the next available index for the specified scope. The redeem script is stored
+    /// alongside the address for later use in transaction construction.
+    ///
+    /// # Parameters
+    /// - `account`: The identifier for the ZIP 48 multisig account.
+    /// - `scope`: The derivation scope (`External` for receiving, `Internal` for change).
+    ///
+    /// Returns `Ok(None)` if the account identifier does not correspond to a known
+    /// ZIP 48 multisig account.
+    #[cfg(feature = "zip-48")]
+    fn get_next_zip48_multisig_address(
+        &mut self,
+        _account: Self::AccountId,
+        _scope: zip32::Scope,
+    ) -> Result<Option<(TransparentAddress, TransparentAddressMetadata)>, Self::Error> {
+        unimplemented!(
+            "WalletWrite::get_next_zip48_multisig_address must be overridden for wallets to use the `zip-48` feature"
+        )
+    }
+
+    /// Generates, persists, and marks as exposed an address at a specific index for a ZIP 48
+    /// multisig account.
+    ///
+    /// This method allows generating addresses at specific indices, which may be useful for
+    /// wallet recovery or coordinating with external systems.
+    ///
+    /// # Parameters
+    /// - `account`: The identifier for the ZIP 48 multisig account.
+    /// - `scope`: The derivation scope (`External` for receiving, `Internal` for change).
+    /// - `address_index`: The specific index at which to derive the address.
+    ///
+    /// # Warning
+    /// If the chosen index is outside the wallet's gap limit, funds sent to this address
+    /// may not be discovered on recovery from the viewing key alone.
+    #[cfg(feature = "zip-48")]
+    fn get_zip48_multisig_address_for_index(
+        &mut self,
+        _account: Self::AccountId,
+        _scope: zip32::Scope,
+        _address_index: NonHardenedChildIndex,
+    ) -> Result<Option<(TransparentAddress, TransparentAddressMetadata)>, Self::Error> {
+        unimplemented!(
+            "WalletWrite::get_zip48_multisig_address_for_index must be overridden for wallets to use the `zip-48` feature"
+        )
+    }
+
     /// Updates the wallet's view of the blockchain.
     ///
     /// This method is used to provide the wallet with information about the state of the
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -1338,6 +1338,10 @@ where
                         .expect("spending key derivation should not fail"),
                     #[cfg(feature = "transparent-key-import")]
                     TransparentAddressSource::Standalone(pubkey) => *pubkey,
+                    #[cfg(feature = "zip-48")]
+                    TransparentAddressSource::Zip48 { .. } => {
+                        return Err(Error::ProposalNotSupported);
+                    }
                 };
 
                 utxos_spent.push(outpoint.clone());
@@ -1706,6 +1710,10 @@ where
                 .standalone_transparent_keys
                 .get(&_address)
                 .ok_or(Error::AddressNotRecognized(_address))?,
+            #[cfg(feature = "zip-48")]
+            TransparentAddressSource::Zip48 { .. } => {
+                return Err(Error::ProposalNotSupported);
+            }
         });
     }
     let sapling_extsks = &[
@@ -2078,6 +2086,8 @@ where
                                 } => Some((index, *scope, *address_index)),
                                 #[cfg(feature = "transparent-key-import")]
                                 TransparentAddressSource::Standalone(_) => None,
+                                #[cfg(feature = "zip-48")]
+                                TransparentAddressSource::Zip48 { .. } => None,
                             })
                     })
                     .collect::<Vec<_>>();
--- a/zcash_client_backend/src/wallet.rs
+++ b/zcash_client_backend/src/wallet.rs
@@ -17,6 +17,8 @@ use zcash_protocol::{
     consensus::{BlockHeight, TxIndex},
     value::{BalanceError, Zatoshis},
 };
+#[cfg(feature = "zip-48")]
+use zcash_script::script;
 use zip32::Scope;
 
 use crate::fees::sapling as sapling_fees;
@@ -778,6 +780,30 @@ impl TransparentAddressMetadata {
         }
     }
 
+    /// Returns a [`TransparentAddressMetadata`] for a ZIP 48 transparent multisig address derived
+    /// from a [`::transparent::zip48::FullViewingKey`].
+    ///
+    /// The `scope`, `address_index`, and `redeem_script` should be the values used when
+    /// calling [`::transparent::zip48::FullViewingKey::derive_address`].
+    #[cfg(feature = "zip-48")]
+    pub fn zip48_multisig(
+        scope: TransparentKeyScope,
+        address_index: NonHardenedChildIndex,
+        redeem_script: script::Redeem,
+        exposure: Exposure,
+        next_check_time: Option<SystemTime>,
+    ) -> Self {
+        Self {
+            source: TransparentAddressSource::Zip48 {
+                scope,
+                address_index,
+                redeem_script,
+            },
+            exposure,
+            next_check_time,
+        }
+    }
+
     /// Returns the source metadata for the address.
     pub fn source(&self) -> &TransparentAddressSource {
         &self.source
@@ -825,6 +851,13 @@ impl TransparentAddressMetadata {
     pub fn address_index(&self) -> Option<NonHardenedChildIndex> {
         self.source.address_index()
     }
+
+    /// Returns the redeem script for the address, if this is a P2SH address.
+    /// Returns `None` for non-P2SH addresses.
+    #[cfg(feature = "zip-48")]
+    pub fn redeem_script(&self) -> Option<&script::Redeem> {
+        self.source.redeem_script()
+    }
 }
 
 /// Source information for a transparent address.
@@ -842,6 +875,14 @@ pub enum TransparentAddressSource {
     /// This variant provides the public key directly.
     #[cfg(feature = "transparent-key-import")]
     Standalone(secp256k1::PublicKey),
+    /// ZIP 48 transparent multisig derivation information for the address below account
+    /// pubkey level, i.e. the `change` and `index` elements of the path.
+    #[cfg(feature = "zip-48")]
+    Zip48 {
+        scope: TransparentKeyScope,
+        address_index: NonHardenedChildIndex,
+        redeem_script: script::Redeem,
+    },
 }
 
 #[cfg(feature = "transparent-inputs")]
@@ -853,6 +894,8 @@ impl TransparentAddressSource {
             TransparentAddressSource::Derived { scope, .. } => Some(*scope),
             #[cfg(feature = "transparent-key-import")]
             TransparentAddressSource::Standalone(_) => None,
+            #[cfg(feature = "zip-48")]
+            TransparentAddressSource::Zip48 { scope, .. } => Some(*scope),
         }
     }
 
@@ -863,6 +906,21 @@ impl TransparentAddressSource {
             TransparentAddressSource::Derived { address_index, .. } => Some(*address_index),
             #[cfg(feature = "transparent-key-import")]
             TransparentAddressSource::Standalone(_) => None,
+            #[cfg(feature = "zip-48")]
+            TransparentAddressSource::Zip48 { address_index, .. } => Some(*address_index),
+        }
+    }
+
+    /// Returns the redeem script for the address, if this is a P2SH address.
+    /// Returns `None` for non-P2SH addresses.
+    #[cfg(feature = "zip-48")]
+    pub fn redeem_script(&self) -> Option<&script::Redeem> {
+        match self {
+            TransparentAddressSource::Derived { .. } => None,
+            #[cfg(feature = "transparent-key-import")]
+            TransparentAddressSource::Standalone(_) => None,
+            #[cfg(feature = "zip-48")]
+            TransparentAddressSource::Zip48 { redeem_script, .. } => Some(redeem_script),
         }
     }
 }
--- a/zcash_client_memory/Cargo.toml
+++ b/zcash_client_memory/Cargo.toml
@@ -125,6 +125,9 @@ zcashd-compat = ["zcash_keys/zcashd-compat", "zcash_client_backend/zcashd-compat
 ## Enables the prospective ZIP 233 (Network Sustainability Mechanism) feature.
 zip-233 = ["zcash_primitives/zip-233"]
 
+## Enables support for ZIP 48 transparent P2SH multisig accounts.
+zip-48 = ["transparent-inputs", "zcash_client_backend/zip-48"]
+
 [build-dependencies]
 prost-build.workspace = true
 which.workspace = true
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -1299,6 +1299,26 @@ impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R:
         })
     }
 
+    #[cfg(feature = "zip-48")]
+    fn import_account_zip48_multisig(
+        &mut self,
+        name: &str,
+        fvk: &transparent::zip48::FullViewingKey,
+        birthday: &AccountBirthday,
+    ) -> Result<Self::Account, Self::Error> {
+        self.transactionally(|wdb| {
+            wallet::add_account(
+                wdb.conn.0,
+                &wdb.params,
+                name,
+                &AccountSource::Zip48,
+                wallet::ViewingKey::Zip48Full(Box::new(fvk.clone())),
+                birthday,
+                &wdb.gap_limits,
+            )
+        })
+    }
+
     fn delete_account(&mut self, account_uuid: Self::AccountId) -> Result<(), Self::Error> {
         self.transactionally(|wdb| wallet::delete_account(wdb.conn.0, account_uuid))
     }
@@ -1340,6 +1360,11 @@ impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R:
     ) -> Result<Option<UnifiedAddress>, Self::Error> {
         if let Some(account) = self.get_account(account)? {
             use zcash_keys::keys::AddressGenerationError::*;
+
+            #[cfg(feature = "zip-48")]
+            if matches!(account.source(), AccountSource::Zip48) {
+                return Err(SqliteClientError::Zip48UnsupportedOperation);
+            }
 
             match account.uivk().address(diversifier_index, request) {
                 Ok(address) => {
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -1391,6 +1391,41 @@ impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R:
         }
     }
 
+    #[cfg(feature = "zip-48")]
+    fn get_next_zip48_multisig_address(
+        &mut self,
+        account: Self::AccountId,
+        scope: zip32::Scope,
+    ) -> Result<Option<(TransparentAddress, TransparentAddressMetadata)>, Self::Error> {
+        self.transactionally(|wdb| {
+            wallet::get_next_available_zip48_multisig_address(
+                wdb.conn.0,
+                &wdb.params,
+                &wdb.gap_limits,
+                account,
+                scope,
+            )
+        })
+    }
+
+    #[cfg(feature = "zip-48")]
+    fn get_zip48_multisig_address_for_index(
+        &mut self,
+        account: Self::AccountId,
+        scope: zip32::Scope,
+        address_index: NonHardenedChildIndex,
+    ) -> Result<Option<(TransparentAddress, TransparentAddressMetadata)>, Self::Error> {
+        self.transactionally(|wdb| {
+            wallet::get_zip48_multisig_address_for_index(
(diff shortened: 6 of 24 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -468,6 +539,20 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     let zcashd_legacy_address_index: i64 = LEGACY_ADDRESS_INDEX_NULL;
 
     let ufvk_encoded = viewing_key.ufvk().map(|ufvk| ufvk.encode(params));
+
+    // For ZIP 48 multisig accounts, we store a placeholder UIVK and the serialized FVK
```
nullcopy: placeholder UIVK necessary, because `uivk` column is `TEXT NOT NULL`. I don't like that 'fake' key material gets stored here for ZIP-48 accounts. I would prefer to nullify this column, or overload with `zip48::FullViewingKey`

nullcopy:
Force pushed to rebase and fix test-build errors

zcash_client_backend/src/data_api.rs
```diff
@@ -489,6 +494,13 @@ pub enum AccountSource {
         purpose: AccountPurpose,
         key_source: Option<String>,
     },
+
+    /// A ZIP 48 transparent multisig account.
+    #[cfg(feature = "zip48-multisig")]
+    Zip48Multisig {
+        purpose: AccountPurpose,
+        key_source: Option<String>,
```
nuttycom: Hmmm... the AccountPurpose has the ZIP 32 derivation information associated with it, but that's not quite the right information to attach; we would need the BIP 48 derivation path for this wallet's contribution, but there's an added complication: this is an "account" not in the ZIP 32 account sense, but in a "viewing authority with potentially attached signing capability" sense. Suppose that the overall wallet is associated with two of three signers - say a zodl spending key and a Keystone device that is associated with the wallet, such that the derivation path from the zodl spending key uses a different BIP 44 account-level identifier than the Keystone device. In that case, what would we want here?

nullcopy:
Force pushed to fix test build errors

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -95,17 +99,31 @@ CREATE TABLE "accounts" (
...
         account_kind = 1
         AND (hd_seed_fingerprint IS NULL) = (hd_account_index IS NULL)
+        AND zip48_fvk IS NULL
+      )
+      OR
+      (
+        account_kind = 2
+        AND zip48_fvk IS NOT NULL
+        AND hd_seed_fingerprint IS NULL
+        AND hd_account_index IS NULL
```
nuttycom, line 125: This makes me think that the hd seed fingerprint and hd account index should potentially no longer be part of the account table; instead, maybe they should be stored in something like an `account_derivation` table that foreign keys to this one? This relates to my comment about `AccountPurpose`.

zcash_client_sqlite/src/lib.rs
```diff
@@ -130,12 +131,23 @@ use {
...
+    not(feature = "transparent-inputs"),
+))]
+use zcash_keys::address::Address;
+
 #[cfg(any(test, feature = "test-dependencies", feature = "transparent-inputs"))]
 use crate::wallet::encoding::KeyScope;
 
+#[cfg(all(
+    any(test, feature = "test-dependencies"),
+    not(feature = "transparent-inputs"),
```
nuttycom: I don't understand this feature gate. Everything changed in this file was added under the `zip48-multisig` feature, but doesn't that imply `transparent-inputs`? So what is this `not` doing?

zcash_client_sqlite/src/lib.rs
```diff
@@ -1367,6 +1409,41 @@ impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R:
         }
     }
 
+    #[cfg(feature = "zip48-multisig")]
+    fn get_next_zip48_multisig_address(
```
nuttycom, line 1395: It feels like there should be a relationship between this and getting ephemeral addresses; in both cases, we just want to get the transparent address by gap limit generation and return it. This also raises a question about what we'll want to do with FROST integration; if a user is sending to an ephemeral address from a FROST multisig spending account to a TEX address, we'll have problems. Probably best to just not support that use case and make users manually send to a transparent multisig, if they really want to do that.

zcash_client_sqlite/src/wallet.rs
nuttycom, line 251: Note that ZIP 316 does provide handling for P2SH addresses and UFVK items. And UFVKs are not subject to the same must-contain-shielded restrictions that UAs are; you just can't generate a UA from a transparent-only UFVK.

nuttycom (commented):
Flushing a few comments from review.

zcash_client_sqlite/src/wallet.rs
nuttycom: Oh, wait, I was wrong:
> There is no defined way to represent a Viewing Key for a Transparent P2SH Address in a UFVK or UIVK (because P2SH Addresses cannot, in general, be diversified in an unlinkable way). The Typecode 0x01 MUST NOT be included in a UFVK or UIVK by Producers, and MUST be treated as unrecognised by Consumers.

nuttycom: That "in general" is doing the heavy lifting here, and I think it's actually wrong; even if P2SH keys can't be diversified in general, we should permit the specific cases where they can (since you need extra metadata to diversify them anyway.)

str4d: The "in general" was added in https://github.com/zcash/zips/pull/1074/changes/62c8eb6f908496a302871d51edbd004deca677a5 as part of the ZIP 48 PR. As noted in its commit message:
> P2MS-in-P2SH is a counter-example, but also evidence that each P2SH address needs to be considered individually.

The MUST NOTs and MUSTs were present in ZIP 316 Revision 0 before ZIP 48 was written, and thus cannot be modified by ZIP 48 itself (but instead would need to be handled by a ZIP 316 revision).

zcash_client_backend/Cargo.toml
```diff
@@ -193,6 +193,9 @@ transparent-inputs = [
 # from which it permits spending.
 transparent-key-import = ["transparent-inputs"]
 
+## Enables tracking of ZIP 48 transparent multisig wallets
```
str4d:
```suggestion
## Enables tracking of ZIP 48 transparent multisig wallets.
##
## Applications that enable this feature **MUST** also enable the
## `transparent-inputs` feature of the crate they are using that implements the
## `zcash_client_backend` traits.
```

zcash_client_backend/Cargo.toml
```diff
@@ -193,6 +193,9 @@ transparent-inputs = [
 # from which it permits spending.
```
str4d: Unrelated issue I noticed while reviewing this PR: these comments won't show up because they only have a single octothope. Also, this should have the same `MUST` as my other suggestion.
```suggestion
## Enables operations that permit arbitrary transparent pubkeys to be associated
## with an account in the wallet.
##
## Applications that enable this feature **MUST** also enable the
## `transparent-inputs` feature of the crate they are using that implements the
## `zcash_client_backend` traits.
##
## Applications that rely on this feature must maintain spending keys
## corresponding to all imported pubkeys for any account from which it permits
## spending.
```

zcash_client_backend/src/data_api.rs
nullcopy: Good point about the derivation paths.

> Suppose that the overall wallet is associated with two of three signers

I'm not sure I follow. From the perspective of viewing and coin tracking, every multisig is associated with all N signers, not merely two of them. When it comes to spending, I envisioned interaction of multiple "accounts", where each account represents one of the participants in the multisig.

So, in your example, one account for the ZODL hotkey and and one for the Keystone device. Both accounts would be `AccountSource::Zip48Multisig`, and each would have to coordinate to build the PCZT to build a spend transaction.

Is that different from how you envisioned?

zcash_client_sqlite/src/lib.rs
nullcopy: Ooops, this shouldn't be here. This is an artifact from when I was trying to implement a standalone `p2sh-inputs` feature gate.

nullcopy: removed: 5847fc54f9

zcash_client_sqlite/src/lib.rs
nullcopy: I feel like there was a reason I couldn't do that my first time around, but the reason is escaping me. 

Looking at this again, I agree with you. I'll see about moving these into the `WalletRead` trait and merge with (or at least follow the same pattern as) the handling of ephemeral transparent receivers

zcash_client_backend/src/data_api.rs
nuttycom: Hmm, yeah, I guess that makes sense, to have the account correspond to the signer as opposed to having it correspond with the multisig as a whole.

zcash_client_sqlite/src/wallet.rs
str4d: In general I'm fine with making `NOT NULL` columns start being nullable, as long as the database constraints still enforce normalization. Previously, every entry in the table required a `uivk`; now every entry in the table would need either a `uivk` or the ZIP 48 incoming viewing key equivalent. So if there were an additional column, then both columns could be nullable but with a constraint that exactly one of them must be set.

Note also that if we do add a ZIP 48 column, we should really add two: one for the FVK, and one for the IVK (as it's the latter that pairs with UIVK and should be "exactly one is set").

zcash_client_backend/src/data_api.rs
str4d: I disagree with having accounts correspond to signers. In the context of `zcash_client_backend`, we *do not track signing keys*. Everything is tracked at a viewing level, and it is up to the caller to store and track the necessary signing data. For ZIP 48, every participating signer in the multisig has the same wallet policy and thus the same full viewing key.

Moreover, the wallet backend assumes that an output can only be received by a single account:
https://github.com/zcash/librustzcash/blob/1c15b44ab7f356afac8fd11aa402bf5fba1e9431/zcash_client_sqlite/src/wallet/db.rs#L497-L504

`account_id` is documented as:
https://github.com/zcash/librustzcash/blob/1c15b44ab7f356afac8fd11aa402bf5fba1e9431/zcash_client_sqlite/src/wallet/db.rs#L485

which I think is still consistent with accounts being tracked at the viewing level rather than the signer level, because for ZIP 48 an individual signer does not control spend authority - only a threshold of signers does.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -219,6 +220,7 @@ pub(super) fn all_migrations<
         Box::new(add_transaction_trust_marker::Migration),
         Box::new(account_delete_cascade::Migration),
         Box::new(v_tx_outputs_key_scopes::Migration),
+        Box::new(zip48_multisig::Migration),
```
str4d, line 225: Remember to update the ASCII graph above this when adding migrations.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -66,6 +67,9 @@ use crate::wallet::scanning::priority_code;
 ///   period (to keep the scan progress percentage accurate to what actually needs scanning).
 /// - `has_spend_key`: A boolean flag (0 or 1) indicating whether the application that embeds
 ///   this wallet database has access to spending key(s) for the account.
+/// - `zip48_fvk`: For ZIP 48 transparent multisig accounts (account_kind = 2), the serialized
+///   ZIP 48 full viewing key. This contains the threshold and all information needed for address
+///   derivation. Must be null for non-multisig accounts.
```
str4d: Change this to "the full viewing key serialized as a BIP 389 multipath descriptor template" (the "ZIP 48" is duplicative), and add a note that arbitrary BIP 389 multipath descriptor templates are not supported (only the subset usable by ZIP 48).

I'm currently undecided on whether to use the BIP 389 encoding here, or a BIP 388 wallet policy. The latter is composed of a wallet descriptor template and a key information vector; each of these is a stricter encoding than BIP 389, but there is no specification for encoding a "wallet policy" as a single thing.BIP 389 by contrast gives you a single descriptor string, albeit with parser scope we need to document as "invalid for our usage".

zcash_client_sqlite/src/wallet.rs
```diff
@@ -218,15 +222,60 @@ fn parse_account_source(
...
+        )),
         (_, _) => Err(SqliteClientError::CorruptedData(
             "Unrecognized account_kind".to_string(),
         )),
     }
 }
 
+/// Parses a ZIP 48 full viewing key from its serialized BIP 389 multipath descriptor format.
+#[cfg(feature = "zip48-multisig")]
+fn parse_zip48_fvk<P: zcash_protocol::consensus::Parameters>(
```
str4d: This should eventually move to a constructor method `transparent::zip48::FullViewingKey::parse_multipath_descriptor` (just below `FullViewingKey::standard`), and be documented as only parsing the subset of BIP 389 that can be created by other `FullViewingKey` constructors.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -241,6 +290,13 @@ pub(crate) enum ViewingKey {
     /// Accounts that have this kind of viewing key cannot be used in wallet contexts,
     /// because they are unable to maintain an accurate balance.
     Incoming(Box<UnifiedIncomingViewingKey>),
+
+    /// A ZIP 48 transparent multisig full viewing key.
+    ///
+    /// These accounts only support transparent P2SH multisig addresses and do not have
+    /// unified viewing keys.
+    #[cfg(feature = "zip48-multisig")]
+    Zip48Multisig(Box<::transparent::zip48::FullViewingKey>),
```
str4d: I think this is the wrong generalisation layer, because this is still a "full viewing key", and it is also possible to define an "incoming viewing key" for ZIP 48 multisig accounts (namely, with key expressions that end in `/0/*` instead of `/<0;1>/*`).

Because this is crate-internal enum I don't think the nested format I'd prefer in a public API is needed here. Instead, rename this to:
```suggestion
    Zip48Full(Box<::transparent::zip48::FullViewingKey>),
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -408,6 +479,23 @@ pub(crate) fn add_account<P: consensus::Parameters>(
             return Err(SqliteClientError::AccountCollision(existing_account.id()));
         }
     }
+
+    // Check whether a ZIP 48 multisig account with the same FVK already exists.
+    #[cfg(feature = "zip48-multisig")]
+    if let Some(fvk_bytes) = viewing_key.zip48_fvk_bytes(params) {
+        if let Some(existing_uuid) = conn
+            .query_row(
+                "SELECT uuid FROM accounts WHERE zip48_fvk = :zip48_fvk",
```
str4d, line 453: String equality is insufficient to detect collisions here. Because we are using `sortedmulti()` for deterministic derivation, it means that the order of the key expressions within the BIP 388 key information vector or BIP 389 multipath descriptor template has no effect on the generated addresses.

Instead, we will need to do one of the following:
- Enforce canonical ordering inside the encodings. This corresponds to implementing the TODO inside `FullViewingKey::standard`.
- Parse each ZIP 48 FVK from the database, and then check equality on the `transparent::zip48::FullViewingKey`s (after implementing equality on that type appropriately).

zcash_client_backend/src/wallet.rs
```diff
@@ -820,6 +853,14 @@ pub enum TransparentAddressSource {
     /// This variant provides the public key directly.
     #[cfg(feature = "transparent-key-import")]
     Standalone(secp256k1::PublicKey),
+    /// ZIP 48 transparent multisig derivation information for the address below account
+    /// pubkey level, i.e. the `change` and `index` elements of the path.
+    #[cfg(feature = "zip48-multisig")]
+    Zip48Multisig {
```
str4d: I was initially unsure that this was the correct place to present the redeem script. But then I remembered that `(scope, address_index)` is insufficient to derive unless you also know which BIP to derive with, and `TransparentAddressSource::Derived` is specifically for BIP 44 derivation. So yes, I agree that this should be a new variant.

I'm not convinced that the suffix is necessary:
```suggestion
    Zip48 {
```
but I'm only weakly against it. If we dropped the suffix, then I'd also be weakly in favour of renaming the `Derived` variant to `Bip44` (but not in this PR).

zcash_client_memory/Cargo.toml
```diff
@@ -125,6 +125,9 @@ zcashd-compat = ["zcash_keys/zcashd-compat", "zcash_client_backend/zcashd-compat
 ## Enables the prospective ZIP 233 (Network Sustainability Mechanism) feature.
 zip-233 = ["zcash_primitives/zip-233"]
 
+## Enables support for ZIP 48 transparent P2SH multisig accounts.
+zip48-multisig = ["transparent-inputs", "zcash_client_backend/zip48-multisig"]
```
str4d: We should try and be consistent with feature flags. Just above this we have `zip-233`, and in the `zcash_note_encryption` crate we have `pre-zip-212`.

I'd be inclined for this to be either `zip-48` or `transparent-multisig`, instead of `zip48-multisig`. If the ZIP number is in the feature flag then a user seeing it enabled somewhere needs to go look up the ZIP to determine what it means, regardless of whether the `-multisig` suffix is present or not. `transparent-multisig` by contrast tells the user everything they need to know (because we only support one way to do transparent multisig).

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -167,6 +185,9 @@ pub(super) const INDEX_HD_ACCOUNT: &str = r#"CREATE UNIQUE INDEX hd_account ON a
 //    obtained via derivation from an HD seed associated with the account. In cases that
 //    `cached_transparent_receiver_address` is non-null, either this column or
 //    `transparent_child_index` must also be non-null.
+/// - `redeem_script`: For P2SH addresses, the redeem script required to spend funds at this address.
+///   The redeem script encodes the spending conditions and is required when constructing
+///   transactions. This is null for non-P2SH addresses.
```
str4d, line 193: Hmm, do we *need* to store this here?

- For ZIP 48 we can re-derive it via `(account_id, key_scope, transparent_child_index)`, so this should only be `cached_redeem_script` in that context.
- For arbitrary non-derived P2SH addresses, the redeem script feels closer to a "viewing key" than "address".

I guess for standalone transparent addresses we put the pubkey in this table, and that is also essentially a redeem script. But as I recall, we only allow standalone transparent addresses to be added as side-cars to existing accounts, not be their own accounts. ZIP 48 by contrast is going to be its own account, so it feels more correct to have this be viewing-level material, and since it can be re-derived at any time, it therefore makes more sense to me to just derive it on-the-fly whenever we need it.

The main question then is what to do about arbitrary P2SH scripts, insofar as we support them from `zcashd`. If we want to support completely arbitrary scripts, then I think each of them should be its own account with the redeem script being the "viewing key". But if we only support scripts that `zcashd` supported, then we should look at each of them individually and determine whether the redeem script is the "viewing key", or whether there is some other thing (like a ZIP 48 FVK) that serves that purpose.

str4d (changes requested):
Flushing my current comments as of 3fd60d3458b8d77506f831a7e0689e8bc567399c. I haven't closely reviewed the PR, just focused on the areas I think need input for overall design.

zcash_client_sqlite/src/lib.rs
nullcopy: Ok, so the difference (unless I'm misunderstanding) is that we do not track address exposure for ephemeral addresses. IMO for a ZIP48 account, we still want to mark an address as exposed so that the caller can request new, previously unexposed addresses on each call. So, IMO this is analogous to `get_next_available_address()`, but specifically for P2SH account types (today ZIP 48 only)

zcash_client_backend/src/wallet.rs
nullcopy: I like the idea of dropping the suffix and renaming Derived->Bip44 (in another PR). As I think about generic P2SH support, I imagine more variants being added here, and naming each variant by ZIP/BIP number seems logical

zcash_client_sqlite/src/wallet.rs
nullcopy: > Note also that if we do add a ZIP 48 column, we should really add two: one for the FVK, and one for the IVK

AFAICT ZIP 48 only has FVK. Am I just missing the IVK type somewhere?

zcash_client_memory/Cargo.toml
nullcopy: Makes sense. I've renamed to `zip-48` everywhere

zcash_client_backend/src/data_api.rs
nullcopy: Ok, that's good context I was missing. In that case, yeah, a single account will need to be able to coordinate signatures from all involved signers.

For now, I've created a `Zip48Derivation` type, which I believe is sufficient for the view-only case. I have some ideas how to tackle spending with multiple signers, but that will be another PR

zcash_client_sqlite/src/wallet/db.rs
nullcopy: I stored them, in an attempt to follow the same pattern as standalone p2pkh addresses, but I am not married to it if you'd prefer I remove it.

> what to do about arbitrary P2SH scripts [...]. If we want to support completely arbitrary scripts, then I think each of them should be its own account with the redeem script being the "viewing key"

I'm weakly of the opinion that each script type we support should be a new account type, rather than supporting arbitrary P2SH accounts. IMO its a lot easier to gain confidence around a finite set of known account types, but I'm sure there's been a lot of discussion around this I'm not privy to.

zcash_client_sqlite/src/wallet.rs
nullcopy: I opted to implement key sorting when constructing a standard `zip48::FullViewingKey`: https://github.com/nullcopy/librustzcash/commit/bdab8172edb53d25ff56031f69b81e92c64dc4d8

zcash_client_sqlite/src/wallet.rs
nullcopy: > it is also possible to define an "incoming viewing key" for ZIP 48 multisig accounts

Is that implemented anywhere?

nullcopy: In chat you clarified that IVK has not been implemented. I don't see any need to implement it yet today, though, so the rename should be sufficient for now

nullcopy: Oops, I had renamed this as you suggested, but just looked at the PR and realized it got re-renamed... fixed in dc51ecd1c5

zcash_client_sqlite/src/wallet.rs
nullcopy: This was answered elsewhere: IVK type does not exist, but _could_ be created.

I made the UIVK columns nullable and ensured normalization. I haven't added a column for the IVK yet, but I'm questioning if we really need it? If the difference between IVK and FVK really is just whether or not we derive the BIP32 change address tree, it feels like we could reasonably derive the IVK or FVK from the other... what is the benefit to storing both?

zcash_client_backend/src/data_api.rs
nullcopy: Looking at this again, I don't like the new `Zip48Derivation` type I added. The data it holds today is really just a placeholder since this PR only implements view-only capability. I've gone ahead and removed it for now. It will come back in my next PR to add generic key structures

nullcopy:
Force pushed to rebase and drop the `Zip48Derivation` type (deferring that to a future PR)

zcash_transparent/src/zip48.rs
```diff
@@ -258,7 +258,7 @@ impl FullViewingKey {
     }
 
     /// Returns the ZIP 48 coin type and account ID for this full viewing key.
-    fn coin_type_and_account(&self) -> (u32, AccountId) {
+    pub fn coin_type_and_account(&self) -> (u32, AccountId) {
```
nuttycom, line 316: If we're going to make this public, we should probably add a newtype wrapper for `CoinType`.

zcash_client_backend/src/wallet.rs
```diff
@@ -863,6 +906,21 @@ impl TransparentAddressSource {
...
             TransparentAddressSource::Standalone(_) => None,
+            #[cfg(feature = "zip-48")]
+            TransparentAddressSource::Zip48 { address_index, .. } => Some(*address_index),
+        }
+    }
+
+    /// Returns the redeem script for the address, if this is a P2SH address.
+    /// Returns `None` for non-P2SH addresses.
+    #[cfg(feature = "zip-48")]
+    pub fn redeem_script(&self) -> Option<&script::Redeem> {
```
nuttycom, line 917: This addition should be noted in the CHANGELOG

zcash_client_backend/src/wallet.rs
```diff
@@ -842,6 +875,14 @@ pub enum TransparentAddressSource {
     /// This variant provides the public key directly.
     #[cfg(feature = "transparent-key-import")]
     Standalone(secp256k1::PublicKey),
+    /// ZIP 48 transparent multisig derivation information for the address below account
+    /// pubkey level, i.e. the `change` and `index` elements of the path.
+    #[cfg(feature = "zip-48")]
```
nuttycom, line 880: This addition should be noted in the CHANGELOG

zcash_client_backend/src/wallet.rs
```diff
@@ -825,6 +851,13 @@ impl TransparentAddressMetadata {
     pub fn address_index(&self) -> Option<NonHardenedChildIndex> {
         self.source.address_index()
     }
+
+    /// Returns the redeem script for the address, if this is a P2SH address.
+    /// Returns `None` for non-P2SH addresses.
+    #[cfg(feature = "zip-48")]
+    pub fn redeem_script(&self) -> Option<&script::Redeem> {
```
nuttycom, line 858: CHANGELOG

zcash_client_backend/src/wallet.rs
```diff
@@ -778,6 +780,30 @@ impl TransparentAddressMetadata {
         }
     }
 
+    /// Returns a [`TransparentAddressMetadata`] for a ZIP 48 transparent multisig address derived
+    /// from a [`::transparent::zip48::FullViewingKey`].
+    ///
+    /// The `scope`, `address_index`, and `redeem_script` should be the values used when
+    /// calling [`::transparent::zip48::FullViewingKey::derive_address`].
+    #[cfg(feature = "zip-48")]
+    pub fn zip48_multisig(
```
nuttycom, line 789: CHANGELOG

zcash_client_backend/src/data_api.rs
```diff
@@ -549,6 +557,8 @@ pub trait Account {
                 derivation: Some(derivation.clone()),
             },
             AccountSource::Imported { purpose, .. } => purpose.clone(),
+            #[cfg(feature = "zip-48")]
+            AccountSource::Zip48 => AccountPurpose::ViewOnly,
```
nuttycom, line 561: Document the rationale behind the Zip48 purpose being declared as ViewOnly

zcash_client_backend/src/data_api.rs
```diff
@@ -489,6 +489,10 @@ pub enum AccountSource {
         purpose: AccountPurpose,
         key_source: Option<String>,
     },
+
+    /// A ZIP 48 transparent multisig account.
+    #[cfg(feature = "zip-48")]
+    Zip48,
```
nuttycom, line 495: CHANGELOG

zcash_client_backend/src/data_api/wallet.rs
nuttycom, line 1081: The changes to this function further convince me that we should move away from the `Proposal` type and just use PCZTs everywhere instead.

zcash_client_sqlite/src/wallet/init/migrations/zip48_multisig.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    fn description(&self) -> &'static str {
+        "Adds support for tracking ZIP 48 transparent P2SH multisig wallets."
+    }
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
```
nuttycom, line 40: Should this migration also add a table that records the relationship between the wallet's HD seed(s) and its ZIP 48 accounts? I'm wondering whether we shouldn't pull the `hd_seed_fingerprint` and `hd_account_index` out of the `accounts` table (since we have to reconstruct it). 

Something I've never been happy with is how we define the spending / viewing split; this shouldn't be an "account kind" column, it should be a property of "do we have derivation metadata". So I would like to take this opportunity to clean this up, get rid of account kinds, and instead have separate table(s) that store this information for an account.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -133,6 +134,8 @@ pub(super) fn all_migrations<
     //                       `------------------- account_delete_cascade ---------------------------------'
     //                                                      |
     //                                           v_tx_outputs_key_scopes
+    //                                                      |
+    //                                               zip48_multisig
```
nuttycom, line 138: There isn't actually a dependency between these two migrations, right? The graph should reflect the actual dependencies.

zcash_client_sqlite/src/wallet/transparent/zip48_multisig.rs
```diff
@@ -0,0 +1,377 @@
...
+    range_to_store: Range<NonHardenedChildIndex>,
+) -> Result<(), SqliteClientError> {
+    let scope = match key_scope {
+        TransparentKeyScope::EXTERNAL => zip32::Scope::External,
+        TransparentKeyScope::INTERNAL => zip32::Scope::Internal,
+        _ => return Err(SqliteClientError::Zip48UnsupportedOperation),
+    };
+
+    for address_index in NonHardenedChildRange::from(range_to_store) {
+        let (address, redeem_script) = fvk.derive_address(scope, address_index);
```
nuttycom, line 49: ~It's probably an error that `fvk.derive_address` takes a `zip32::Scope` as its argument; instead, it should take a `TransparentKeyScope`. That would make the error on line 45 unnecessary.~

@str4d noted that this restriction comes from BIP 388. We need to document this in the documentation of the `scope` parameter of `derive_address`.

zcash_client_sqlite/src/wallet/transparent/zip48_multisig.rs
```diff
@@ -0,0 +1,377 @@
...
+/// (used for gap-filling). When `Some(height)`, the address is marked as exposed at that height.
+/// On conflict, `COALESCE(existing, new)` preserves an already-set exposure height.
+pub(crate) fn insert_zip48_multisig_address<P: consensus::Parameters>(
+    conn: &rusqlite::Transaction<'_>,
+    params: &P,
+    account_id: AccountRef,
+    scope: TransparentKeyScope,
+    address_index: NonHardenedChildIndex,
+    address: &TransparentAddress,
+    redeem_script: &script::Redeem,
```
nuttycom: I'm not sure that we should have the redeem script here? We need the key info vector, but that's an account-level concern, not an address-level concern, and with that we can always derive the redeem script.

In zcashd, `addmultisigaddress` takes a list of keys and the number of signatures required, and creates and stores the redeem script, so there's some precedent here, but there too, I think that we should probably treat this as account-level information. But then again, while that's a P2SH multisig account, it's not (identifiably) a ZIP 48 multisig account; ZIP 48 is specifically about defining that derivation. For zcashd migration, we'll need non-ZIP-48 multisig accounts in addition to ZIP 48 accounts; for that purpose, I would like to keep the **address** level "clean".

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Aha, I agree with @str4d that I don't think we need the redeem script here if we can produce it. And as I mentioned above, I think that `zip48::FullViewingKey::derive_address` should take a `TransparentKeyScope` instead of a `zip32::Scope`; then, we have fewer special cases.

zcash_client_sqlite/src/wallet.rs
nuttycom: Let's discuss how to revise ZIP 316 in ZIP Sync or R&D.

zcash_client_sqlite/src/wallet/transparent/zip48_multisig.rs
```diff
@@ -135,6 +150,81 @@ pub(crate) fn insert_zip48_multisig_address<P: consensus::Parameters>(
...
+) -> Result<NonHardenedChildIndex, SqliteClientError> {
+    let key_scope = KeyScope::try_from(scope)?;
+
+    let max_index: Option<u32> = conn
+        .query_row(
+            "SELECT MAX(transparent_child_index)
+                 FROM addresses
+                 WHERE account_id = :account_id
+                 AND key_scope = :key_scope
+                 AND redeem_script IS NOT NULL",
```
nuttycom, line 169: I think we should take the "diversifiability" of an address from information at the account level; this query seems fragile to me in the presence of adding other types of redeem scripts.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -4603,6 +4607,147 @@ pub mod testing {
...
+    )))
+}
+
+/// Derives and persists an address at a specific index for a ZIP 48 transparent multisig account.
+#[cfg(feature = "zip-48")]
+pub(crate) fn get_zip48_multisig_address_for_index<P: consensus::Parameters>(
+    conn: &rusqlite::Transaction<'_>,
+    params: &P,
+    account_uuid: AccountUuid,
+    scope: zip32::Scope,
```
nuttycom, line 4699: ~This should also be `TransparentKeyScope`.~

The choice to use `zip32::Scope` instead of `TransparentKeyScope` is due to BIP 388 only supporting 0 and 1 for change-level indices.

zcash_client_backend/src/data_api/wallet.rs
nullcopy: That seems sensible. Is that something to take on in this PR?

zcash_client_sqlite/src/wallet/init/migrations.rs
nullcopy: Good catch. Yeah, `account_delete_cascade` is actually the correct dependency. Ive updated the specified dependencies and this graph.

zcash_client_sqlite/src/wallet.rs
nullcopy: Closing this as there's no action for this PR unless ZIP 316 changes, and its unclear if/how/when that will happen.

zcash_client_sqlite/src/wallet/db.rs
nullcopy: I agree. Since ZIP 48 is a new account type, this information should be stored as derivation info under that account.

Standalone P2SH addresses will still need to store a script with the addresses, but that's for [another PR](https://github.com/zcash/librustzcash/pull/2194)

zcash_client_sqlite/src/wallet/transparent/zip48_multisig.rs
nullcopy: Agreed. Removing this in favor of storing key info with the account.

PR #2146, opened 2026-02-13
Fix feature-flagging errors induced by recent gap limits reorg.

nuttycom:


zcash_keys/Cargo.toml
```diff
@@ -105,7 +105,7 @@ transparent-inputs = [
 
 ## Enables use of zcashd-derived secp256k1 secret key encodings.
 transparent-key-encoding = [
-    "transparent-inputs",
+    "dep:bip32",
```
nullcopy, line 108: Just want to confirm this was intentional, as it doesn't _immediately appear_ related to the other feature-flagging errors

zcash_keys/src/keys.rs
```diff
@@ -1686,14 +1686,18 @@ mod tests {
...
         alloc::string::ToString,
         alloc::vec::Vec,
-        zcash_address::test_vectors,
-        zip32::DiversifierIndex,
     };
 
+    #[cfg(all(
+        feature = "transparent-inputs",
+        any(feature = "orchard", feature = "sapling")
+    ))]
```
nullcopy, line 1698: Idea: could we move feature-specific logic into modules, rather than one large file that holds the logic for several different permutations of feature sets? I've noticed similar complexity (and more) in my own work which is adding yet another feature flag, and I think a lot of this could be avoided if we can align feature gates along module boundaries.

zcash_keys/Cargo.toml
nuttycom: Yes; in fact, this is a core issue with the feature flags - the dependency on `transparent-inputs` was causing a transitive dependency problem that showed up in CI, where the zcash_keys crate had `transparent-key-encoding` enabled, which transitively enabled `transparent-inputs`; however, in that test, `zcash_client_sqlite` didn't have `transparent-inputs` enabled and this resulted in errors when zcash_client_sqlite was trying to do address generation with address components that it couldn't handle.

zcash_keys/src/keys.rs
nuttycom: In some cases we're able to do this; in this particular situation, the feature flag is enforcing the ZIP 316 rule that unified addresses MUST contain a shielded receiver. We relaxed that rule in the (undeployed) Revision 1 to unified addresses, but this library still only supports R0 because https://github.com/zcash/librustzcash/pull/1135 got delayed by review, and then bitrotted so badly that it would probably need to be rewritten. I still think we should get R1 UAs deployed because I really want address expiry, but we probably need to go back to the drawing board on other facets of that specification; there's been a strong community desire to remove transparent components from UAs.

nullcopy: > ZIP 316 rule that unified addresses MUST contain a shielded receiver. We relaxed that rule in the (undeployed) Revision 1 to unified addresses

:eyes: Oh really?? This would have a significant positive benefit to https://github.com/zcash/librustzcash/pull/2144

str4d (commented):
Post-hoc utACK 1488770f6191cca05d0786d9767424779b264c80

Issue #1545, opened 2024-09-20
Finish documentation for pool tests extracted into `zcash_client_backend`


str4d:
Follow-up work after #1543.

PR #2147, opened 2026-02-13
Add `WalletWrite::truncate_to_chain_state` for precise rewind.

### LINKED ISSUE #2121: zcash_client_sqlite: Enable truncation to a specific block height, not an approximate block height.
At present, truncation essentially "rounds" to note commitment tree state tracked by the wallet; outside of the last 100 blocks, this means that truncation may rewind significantly farther than is intended. Instead, we should make it possible to truncate to a specific height by providing the chain state needed to insert the frontier prior to the truncated block into the note commitment tree, allowing the note commitment tree to then be updated going forward from the truncated state.

nuttycom:
`truncate_to_height` fails when the target height's checkpoint has been pruned from the note commitment tree (i.e. it is beyond PRUNING_DEPTH blocks in the past). `truncate_to_chain_state` solves this by accepting a `ChainState` whose frontiers are used to reconstruct the checkpoint:

- If a usable checkpoint already exists, delegates directly to `truncate_to_height`.
- Otherwise, truncates to the earliest available checkpoint first, inserts the provided frontier as a new checkpoint at the target height, then truncates to that checkpoint.

Also adds `CachedBlock::chain_state()` accessor to the test infrastructure and a test covering the pruned-checkpoint scenario.

Depends on incrementalmerkletree rev 7b6263ee (insert_frontier).

Fixes #2121
Depends on zcash/incrementalmerkletree#139

Builds upon #2146 and should retarget to `main` when that merges.

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -2057,7 +2057,7 @@ end = "2026-10-26"
 criteria = "safe-to-deploy"
 user-id = 169181 # Kris Nuttycombe (nuttycom)
 start = "2022-12-15"
-end = "2025-04-22"
+end = "2027-02-23"
 
 [[trusted.sinsemilla]]
 criteria = "safe-to-deploy"
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -114,8 +114,8 @@ user-login = "str4d"
 user-name = "Jack Grigg"
 
 [[publisher.shardtree]]
-version = "0.6.1"
-when = "2025-01-31"
+version = "0.6.2"
+when = "2026-02-21"
 user-id = 169181
 user-login = "nuttycom"
 user-name = "Kris Nuttycombe"


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3330,6 +3331,113 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    //
+    // We find the minimum checkpoint height and truncate to it.
+    #[cfg(not(feature = "orchard"))]
+    let min_checkpoint_height_query =
+        "SELECT MIN(checkpoint_id) FROM sapling_tree_checkpoints";
+    #[cfg(feature = "orchard")]
+    let min_checkpoint_height_query = "SELECT MIN(sc.checkpoint_id)
+         FROM sapling_tree_checkpoints sc
+         JOIN orchard_tree_checkpoints oc
+         ON oc.checkpoint_id = sc.checkpoint_id";
```
nuttycom: This doesn't work, we need to truncate to the maximum of the two the two minimum checkpoint ids if `orchard` is enabled.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3330,6 +3331,113 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    let min_checkpoint_height: BlockHeight = wdb
+        .conn
+        .0
+        .query_row(min_checkpoint_height_query, [], |row| {
+            row.get::<_, Option<u32>>(0)
+        })
+        .optional()?
+        .flatten()
+        .map(BlockHeight::from)
+        .ok_or(SqliteClientError::RequestedRewindInvalid {
```
nuttycom: Since we're able to insert a frontier, there should be *no* condition under which a truncation to a chain state is invalid. If there are no checkpoints, then we simply don't have to do the next `truncate_to_height`.

Cargo.toml
```diff
@@ -228,3 +228,8 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(zcash_unstable, values("zfuture", "nu7"))',
   'cfg(live_network_tests)',
 ] }
+
+[patch.crates-io]
+incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree.git", rev = "7b6263ee2af31b2db0131e28928b83a4d52f4e83" }
```
nullcopy: For my own understanding: would we merge patches like this to main? Or will we wait for updated releases for each of those crates?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    }
+
+    // Step 1: Truncate to the earliest available checkpoint. This removes all later
+    // checkpoints and tree data, making room for the frontier insertion.
+    //
+    // We find the minimum checkpoint height for each tree and take the maximum of those
+    // minimums, since that is the earliest height at which all trees have a checkpoint.
+    // The orchard table exists unconditionally but is empty when the feature is not active;
+    // MAX ignores NULLs, so this correctly returns only the sapling minimum in that case.
+    let min_checkpoint_height_query = "SELECT MAX(m) FROM (
```
nullcopy: I see this query is identical to: https://github.com/zcash/librustzcash/blob/3a5cb42a24ab34eb73e66ce3c0fd6697272d62e9/zcash_client_sqlite/src/wallet.rs#L3201

Can/should this be defined somewhere shared, instead of duplicating code?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+            #[cfg(feature = "transparent-inputs")]
+            &wdb.gap_limits,
+            min_checkpoint_height,
+        )?;
+    }
+
+    // Step 2: Insert the frontier from the chain state, creating a checkpoint at
+    // the target height. Because we just truncated down to a single checkpoint,
+    // there is room for this new one.
+    wdb.with_sapling_tree_mut(|tree| {
```
nullcopy, line 3426: > Because we just truncated down to a single checkpoint there is room for this new one.

We may not have truncated at all, if `min_checkpoint_height` is `None`. Does this statement hold true?

zcash_client_sqlite/src/wallet.rs
nuttycom: Yeah, that's probably worth factoring out at this point.

zcash_client_sqlite/src/wallet.rs
nuttycom: The reason for the initial truncation is that we need to ensure that there are fewer than 100 checkpoints; otherwise, when we insert the new (older) checkpoint it'll be immediately pruned. If there *are* no checkpoints, then we definitely have fewer than 100 and can just insert one.

Cargo.toml
nuttycom: We're fine with patches so long as they're on `zcash` org forks. Once these merge to `main` though we will not rebase those PRs, because we want the hashes to eventually end up in history. We would not *release* code that depends on these crates without crate releases of the dependencies.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3173,25 +3173,18 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
-    let truncation_height_query = r#"
-        SELECT MAX(height) FROM blocks
-        JOIN sapling_tree_checkpoints sc ON sc.checkpoint_id = blocks.height
-        JOIN orchard_tree_checkpoints oc ON oc.checkpoint_id = blocks.height
-        WHERE blocks.height <= :block_height
-    "#;
-
+    // Determine a checkpoint to which we can rewind, if any. We need a height
+    // that has a checkpoint in every tree that contains data. The orchard table
+    // exists unconditionally but is empty when the orchard feature is not active.
```
str4d: It can also be empty if no syncing has been done yet, or only syncing operations that don't trigger updates to this table. I can't immediately think of a pathway that would lead to this situation, but it might be worth noting in the comment here that this is a situation we believe should resolve itself (i.e. data gets entered into the table) prior to any reasonable point at which a caller would want to trigger truncation (usually when the syncing APIs detect a reorg and tell the caller to truncate, which requires having scanned at least one block, but I guess the caller could get unlucky and scan reorged blocks that don't contain any Orchard transactions?)

zcash_client_backend/src/data_api.rs
```diff
@@ -3096,6 +3096,15 @@ pub trait WalletWrite: WalletRead {
...
     fn truncate_to_height(&mut self, max_height: BlockHeight) -> Result<BlockHeight, Self::Error>;
 
+    /// Truncates the wallet database to the specified chain state.
+    ///
+    /// In contrast to [`truncate_to_height`], this method allows the caller to truncate the wallet
+    /// database to a precise height by providing additional chain state information needed for
+    /// note commitment tree maintenance after the truncation.
+    ///
+    /// [`truncate_to_height`]: WalletWrite::truncate_to_height
+    fn truncate_to_chain_state(&mut self, chain_state: ChainState) -> Result<(), Self::Error>;
```
str4d, line 3106: Document this change to the `WalletWrite` trait in the changelog. (I concur that this makes the PR a breaking change, thus making `main` an appropriate target for it.)

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -337,6 +337,11 @@ impl CachedBlock {
         }
     }
 
+    /// Returns the chain state as of the end of this block.
+    pub fn chain_state(&self) -> &ChainState {
```
str4d, line 341: I don't recall if we've been documenting additions to the testing APIs, but if we have then document this addition in the changelog.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4256,6 +4256,118 @@ where
     );
 }
 
+pub fn truncate_to_chain_state<T: ShieldedPoolTester, Dsf>(ds_factory: Dsf, cache: impl TestCache)
```
str4d, line 4259: - It would be nice to document this test, given that it's in the public API.
- I don't recall if we've been documenting additions to the testing APIs, but if we have then document this addition in the changelog.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    // Try the simple case first: if a checkpoint exists at or below the target height,
+    // truncate_to_height will succeed directly.
+    match truncate_to_height(
+        wdb.conn.0,
+        &wdb.params,
+        #[cfg(feature = "transparent-inputs")]
+        &wdb.gap_limits,
+        target_height,
+    ) {
+        Ok(_) => return Ok(()),
```
str4d: Possible bug: per the API contract for `WalletWrite::truncate_to_height`, it is allowed to truncate to a lower height than `target_height`. IDK if the zcs `truncate_to_height` implementation has this property, but we might want to gate its call on a check that it will produce an exact truncation (and then check that the returned truncation height matches `target_height`, or maybe assert it as I think this would be a programming error rather than data corruption).

Claude: when reusing existing `zcash_client_sqlite` functions that implement parts of the `WalletRead` or `WalletWrite` traits, please always ensure you satisfy their preconditions (which are documented on those traits in `zcash_client_backend`, and not usually documented on the `zcash_client_sqlite` functions themselves).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+        Err(SqliteClientError::RequestedRewindInvalid { .. }) => {
+            // The target height is below the oldest checkpoint. We need to use the
+            // frontier-based approach.
+        }
+        Err(e) => return Err(e),
+    }
+
+    // Step 1: Truncate to the earliest available checkpoint. This removes all later
```
str4d: These step numberings make no sense, given that Step 3 is a truncation but there are other truncations before and in between the steps that have no numberings. Explicit step numberings in comments are also a maintenance hazard. Remove all of them.

Claude: Please stop adding step numbers to your code comments, unless they correspond to explicit step numbers in a published specification that you are following (and thus are guaranteed to be stable).

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    // checkpoints and tree data, making room for the frontier insertion.
```
str4d: What does "making room" mean? The frontier insertion is necessarily happening before the earliest available checkpoint, so it's guaranteed to overwrite at least some data in the tree regardless of what has been truncated. What problem occurs if it is overwriting *more* data?

EDIT: Reading further down, I see that what is meant here is "room in the checkpoints list" (i.e. so that this earlier checkpoint doesn't immediately get pruned due to being over the 100-checkpoint limit). This comment is therefore a bit misleading: it isn't the insertion of the frontier itself that needs room, but the the `Retention::Checkpoint` associated with that frontier. Suggested reword:
```suggestion
    // checkpoints and tree data, making room for the new checkpoint created by the
    // frontier insertion.
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    // no checkpoints at all, we can skip this step and proceed directly to inserting the
+    // frontier.
+    if let Some(min_checkpoint_height) = min_checkpoint_height {
+        truncate_to_height(
+            wdb.conn.0,
+            &wdb.params,
+            #[cfg(feature = "transparent-inputs")]
+            &wdb.gap_limits,
+            min_checkpoint_height,
+        )?;
```
str4d: Similar possible bug: what guarantee do we have here that `truncate_to_height` does not truncate to below `target_height`? At this point in the function (and assuming the earlier checks I suggested are added), I think that it doing so here would be a programming error; add an assertion that the returned height is `> target_height`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3321,6 +3322,114 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+    })?;
+
+    // Step 3: Now truncate to the target height checkpoint we just created.
+    truncate_to_height(
+        wdb.conn.0,
+        &wdb.params,
+        #[cfg(feature = "transparent-inputs")]
+        &wdb.gap_limits,
+        target_height,
+    )?;
```
str4d: Similarly, add an assertion that the returned height `== target_height`, to catch programming errors in future refactors of `truncate_to_height`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3322,6 +3309,29 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
     Ok(truncation_height)
 }
 
+// Find the minimum checkpoint height for each tree and take the maximum of those minima.
+//
+// This returns the earliest height at which all trees have a checkpoint. The orchard table exists
+// unconditionally but is empty when the `orchard` feature is not active; MAX ignores NULLs, so
+// this correctly returns only the sapling minimum in that case.
```
str4d: Bug: These are code comments, not function docs. Replace `//` with `///`.

Claude: Please use `///` comments when documenting Rust functions, even for private ones, so that the documentation correctly shows up in IDEs at callsites of the function. Only use `//` when adding private comments about a function (e.g. an internal TODO that is only relevant when looking at the function implementation). Also, please follow [Rust documentation conventions](https://doc.rust-lang.org/1.0.0/style/style/comments.html#sentence-structure) by using third person singular present indicative form for summary lines.

str4d (changes requested):
Reviewed 6d82262e0a8aedaaba73ba7b1eea034d3183ec1c

zcash_client_backend/src/data_api/testing/pool.rs
nuttycom: We generally have not documented the tests in this file, either with API documentation or in the CHANGELOG. If we want to start documenting them, then I can dispatch an agent to do the grunt work on all of them for a different PR.

zcash_client_backend/src/data_api/testing.rs
nuttycom: Same answer as above, we have generally not tracked `test-dependencies` changes.

zcash_client_sqlite/src/wallet.rs
nuttycom: You know that I respond to your comments directly, right? You don't need to provide LLM instructions.

nuttycom: Also, this is my error, not Claude's; it just did what I told it to.

nuttycom: I had forgotten that we would truncate further back than was specified.

nuttycom: Realistically, I don't think that this is an issue that could ever, in any circumstances, actually cause a wallet genuine problems, but I'll fix it anyway.

zcash_client_sqlite/src/wallet.rs
nuttycom: Claude does not learn from one session to the next, so giving instructions to future instances needs to be done in the context of an AGENTS.md file or similar.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/6d82262..11c7ed67cbb2ef1ec48c6273ea091584bb13fa47)  to rfix conflicts with `main`, address comments from code review and update to released `shardtree` version.

zcash_client_sqlite/src/wallet.rs
str4d: Claude can determine that your local branch is pushed to this repo and thus figure out this PR exists, and inspect its comments. It could easily be doing this inside the session in which you are addressing PR comments, and I have no way to confirm that it is not doing this. So if I'm reviewing a PR where basically every commit has a Claude co-authorship, how can I assume that you are the only one reading my review comments? I know you can infer what I intend from short review comments, but I also know that an AI reading those same comments is not going to extract the necessary instructions from them, and so if I'm going to end up remote-instructing an AI agent, I may as well do so explicitly and precisely.

nuttycom:
You should not assume your comments on my PRs are being read by the agent;
I don’t give blanket web access permissions, and often don’t maintain
long-running agent sessions; I close the agent once I switch branches
locally.

str4d (approved):
utACK 11c7ed67cbb2ef1ec48c6273ea091584bb13fa47

Cargo.toml
```diff
@@ -55,7 +55,7 @@ bellman = { version = "0.14", default-features = false, features = ["groth16"] }
 ff = { version = "0.13", default-features = false }
 group = "0.13"
 incrementalmerkletree = { version = "0.8.2", default-features = false }
-shardtree = "0.6.1"
+shardtree = "0.6.2"
```
str4d, line 58: `cargo-vet` was not updated for this dependency bump.

schell (approved):
🧂 utACK https://github.com/zcash/librustzcash/commit/11c7ed67cbb2ef1ec48c6273ea091584bb13fa47

zcash_client_backend/src/data_api/testing/pool.rs
nuttycom: Note we already have an issue that is related to this: https://github.com/zcash/librustzcash/issues/1545

PR #2152, opened 2026-02-19
Update `zcash_primitives` and `zcash_client_backend` to permit explicit selection of a transaction version.

nuttycom:
This is a prerequisite for https://github.com/zcash/zcash-devtool/pull/152, which has merged; as such this PR MUST NOT be rebased.

### DIFF
--- a/zcash_primitives/src/transaction/mod.rs
+++ b/zcash_primitives/src/transaction/mod.rs
@@ -256,9 +256,9 @@ impl TxVersion {
                 Sprout | Overwinter => false,
                 Sapling | Blossom | Heartwood | Canopy | Nu5 | Nu6 | Nu6_1 => true,
                 #[cfg(zcash_unstable = "nu7")]
-                Nu7 => false,
+                Nu7 => false, // ZIP 2003
                 #[cfg(zcash_unstable = "zfuture")]
-                ZFuture => false,
+                ZFuture => false, // ZIP 2003
             },
             TxVersion::V5 => match consensus_branch_id {
                 Sprout | Overwinter | Sapling | Blossom | Heartwood | Canopy => false,
@@ -272,7 +272,7 @@ impl TxVersion {
             TxVersion::V6 => match consensus_branch_id {
                 Sprout | Overwinter | Sapling | Blossom | Heartwood | Canopy | Nu5 | Nu6
                 | Nu6_1 => false,
-                Nu7 => true,
+                Nu7 => true, // ZIP 230 or ZIP 248, whichever is chosen for activation
             },
             #[cfg(zcash_unstable = "zfuture")]
             TxVersion::ZFuture => match consensus_branch_id {


zcash_primitives/src/transaction/builder.rs
```diff
@@ -379,6 +389,45 @@ impl<P, U> Builder<'_, P, U> {
             .as_ref()
             .map_or_else(|| &[][..], |b| b.outputs())
     }
+
+    /// Checks that the given version supports all features required by the inputs and
+    /// outputs already added to the builder.
+    fn check_version_compatibility<FE>(&self, version: TxVersion) -> Result<(), Error<FE>> {
```
str4d, line 402: Bug: this needs to confirm that the proposed version is compatible with the `consensus_branch_id` (selected by the user with `target_height`). It also means the `ProposedVersionIncompatible` error variant is insufficient as-is to represent the possible error states.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -379,6 +389,45 @@ impl<P, U> Builder<'_, P, U> {
...
+        if !version.has_sapling()
+            && (!self.sapling_inputs().is_empty() || !self.sapling_outputs().is_empty())
```
str4d: This would read easier as:
```suggestion
        if !(version.has_sapling()
            || (self.sapling_inputs().is_empty() && self.sapling_outputs().is_empty()))
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -379,6 +389,45 @@ impl<P, U> Builder<'_, P, U> {
...
+            return Err(Error::ProposedVersionIncompatible(
+                version,
+                PoolType::SAPLING,
+            ));
+        }
+        if !version.has_orchard()
+            && self
+                .orchard_builder
+                .as_ref()
+                .is_some_and(|b| !b.spends().is_empty() || !b.outputs().is_empty())
```
str4d: Ditto:
```suggestion
        if !(version.has_orchard()
            || self
                .orchard_builder
                .as_ref()
                .is_none_or(|b| b.spends().is_empty() && b.outputs().is_empty())
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -398,6 +403,10 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
         Self::InputSource: InputSource,
         ChangeT: ChangeStrategy<MetaSource = DbT>,
     {
+        // proposed_version is only used in orchard-gated code paths
+        #[cfg(all(feature = "unstable", not(feature = "orchard")))]
+        let _ = proposed_version;
+
```
str4d: Why? If we're adding this field now, we should also hook it into the Sapling code paths so that in future when a Sapling-free transaction version is defined, it works correctly. You've already hooked up the Sapling path inside `Builder`.
```suggestion
```

str4d (changes requested):
Reviewed 0d6e7fdc32321c60fa39b13a3017296a23e205ee

zcash_primitives/src/transaction/builder.rs
nuttycom: I realize that these are equivalent, but I don't actually find this easier to reason through. The way it's currently written, it's clearly saying "there are nonempty sapling inputs and outputs but sapling is not supported." The other way around, it's saying "not (sapling is supported or sapling inputs and outputs are empty)" and I find it harder to puzzle through that to figure out why it's an error.

zcash_primitives/src/transaction/builder.rs
nuttycom: Same answer as above, I think this is harder to reason about.

zcash_primitives/src/transaction/builder.rs
str4d: Interesting. I find the other way round easier to figure out why it's an error, because I see it saying "to be valid, either the transaction format supports Sapling, or we are not using Sapling. If not valid, error".

zcash_primitives/src/transaction/builder.rs
```diff
@@ -393,21 +393,26 @@ impl<P, U> Builder<'_, P, U> {
     /// Checks that the given version supports all features required by the inputs and
     /// outputs already added to the builder.
     fn check_version_compatibility<FE>(&self, version: TxVersion) -> Result<(), Error<FE>> {
-        if !version.has_sapling()
+        let sapling_available = version.has_sapling() && self.consensus_branch_id.has_sapling();
```
str4d, line 411: Gah sorry, my previous review comment was unclear. I didn't mean for you to alter `consensus::BranchId` to expose whether a shielded protocol was enabled in that epoch, but to check whether the requested transaction version is mineable during that epoch. The latter encompasses the former, as a shielded protocol is functionally disabled if no transaction interacting with it can be mined under the consensus rules.

That's also why I said the new error variant was incomplete: this tx version check is pool-independent. If a particular transaction version cannot be mined within a given epoch, that means it can't be used for *any* shielded protocol the tx version supports.

components/zcash_protocol/src/consensus.rs
```diff
@@ -863,6 +863,46 @@ impl BranchId {
...
+        use BranchId::*;
+        match self {
+            Sprout | Overwinter | Sapling | Blossom | Heartwood | Canopy => false,
+            Nu5 | Nu6 | Nu6_1 => true,
+            #[cfg(zcash_unstable = "nu7")]
+            BranchId::Nu7 => true,
+            #[cfg(zcash_unstable = "zfuture")]
+            BranchId::ZFuture => true,
+        }
+    }
```
str4d, line 905: Per my other comment, none of these new APIs are necessary. They might be of independent interest, but they don't actually help with the checks needed in this PR.

zcash_primitives/src/transaction/builder.rs
nuttycom: What about ZIP 248, though? Once it is live, the transaction version will not dictate what protocols are available for a given consensus branch id. I don't think we should couple these right now; the approach I have taken is effectively to take the intersection and ensure that both the transaction version and the consensus branch support the protocol. I will add a check to ensure that the transaction version and the consensus branch ID are also compatible.

components/zcash_protocol/src/consensus.rs
nuttycom: They do, because after ZIP 248 the transaction version will not fully determine the valid set of protocols for use on the branch.

zcash_primitives/src/transaction/builder.rs
nuttycom: The difference to me is essentially which branch of the "if" you're on. If there was something to do in the "else" branch (using the test that is currently written) then I would absolutely switch, put that other thing in the "if true" branch and the error in the "else". I dislike "if not ... then ..." though, that's more negation than my brain is happy with. This is also why I *detest* the `guard` keyword in Swift.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -242,6 +242,46 @@ impl TxVersion {
...
+        // Note: we intentionally use `match` expressions instead of the `matches!`
+        // macro below because we want exhaustivity.
+        match self {
+            TxVersion::Sprout(_) => consensus_branch_id == Sprout,
+            TxVersion::V3 => consensus_branch_id == Overwinter,
+            TxVersion::V4 => match consensus_branch_id {
+                Sprout | Overwinter => false,
+                Sapling | Blossom | Heartwood | Canopy | Nu5 | Nu6 | Nu6_1 => true,
+                #[cfg(zcash_unstable = "nu7")]
+                Nu7 => false,
```
str4d: Nit: this is implementing ZIP 2003. It would be nice for this (and tbh all other changes-in-validity) to have a comment with a reference to the specification enacting the change, not least so that when we get around to stabilising NU7, we remember that we need to confirm here whether or not ZIP 2003 is part of NU7.

nullcopy (commented):
post-hoc utACK https://github.com/zcash/librustzcash/pull/2152/changes/80ad4cbbad4b09aa2e44ea4dede57306b441f990

PR #2162, opened 2026-02-24
zcash_client_backend: Add note locking functionality.

### LINKED ISSUE #2161: zcash_client_backend: Support note locking
We need to add a "note locking" feature to `zcash_client_backend` and `zcash_client_sqlite`. The essential changes that we need are: add a `lock_output` method to the `WalletWrite` trait that takes a (pool, txid, output_index)  tuple as its argument, and that adds a `locked` flag to that note's record in the backing data store; add a database migration to update the storage appropriately. Note selection methods and queries should be modified to take an  `include_locked` parameter; when the value of that parameter is false, locked notes should not be returned from note selection queries and should not be included in balance. Methods that create transaction proposals should also  take an added `lock_notes` parameter (that defaults to `false`); when `true`, the notes selected for spending should be locked before the proposal is returned.

When a proposal to spend locked notes (or a PCZT that spends locked notes) is finally built into an actual transaction, the notes should be unlocked - we will allow the preexisting transaction expiry-based locking functionality to  function there. Note locking has effect only between the time of creating a proposal or pczt, and the time that the proposal is acted upon to construct a transaction.

nuttycom:


zcash_client_backend/src/data_api.rs
```diff
@@ -1657,6 +1657,12 @@ pub trait WalletRead {
         confirmations_policy: ConfirmationsPolicy,
     ) -> Result<Option<WalletSummary<Self::AccountId>>, Self::Error>;
 
+    /// Returns the set of currently locked outputs.
+    ///
+    /// Locked outputs are excluded from note selection and balance computations
+    /// (unless `include_locked` is set to `true`).
+    fn get_locked_outputs(&self) -> Result<Vec<OutputRef>, Self::Error>;
```
nuttycom: I'm not confident that this is a useful API, except perhaps for testing? If it's present only to simplify testing, we should potentially move it to the test framework.

zcash_client_memory/src/testing/mod.rs
```diff
@@ -110,6 +110,10 @@ impl<P> WalletTest for MemoryWalletDb<P>
 where
     P: zcash_protocol::consensus::Parameters + Clone + Debug + PartialEq,
 {
+    fn get_locked_outputs(&self) -> Result<Vec<OutputRef>, Error> {
+        Ok(vec![])
```
nuttycom:
```suggestion
        todo!()
```

zcash_client_backend/CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ workspace.
   - `Note::receiver`
   - `impl From<sapling_crypto::Note> for Note`
   - `impl From<orchard::Note> for Note`
+- `zcash_client_backend::data_api::ll`
```
dannywillems: I don't see this change as part of this commit (`6bf3123a0330b14ed31274eb3ad82180bb16eab1`). Is it expected?

PR #2165, opened 2026-02-25
feat: eip681 TransactionRequest from parts

schell:
These changes add constructors for `TransactionRequest` from parts. This will be useful for FFI, when we want to construct a request address from pieces previously parsed.

### DIFF
--- a/components/eip681/src/parse.rs
+++ b/components/eip681/src/parse.rs
@@ -3,13 +3,13 @@
 use std::{borrow::Cow, collections::BTreeMap};
 
 use nom::{
+    AsChar, Parser,
     branch::alt,
     bytes::complete::{is_not, tag, take_till, take_till1, take_while, take_while1},
     character::complete::{char, digit0},
     combinator::{map_parser, opt, success, value},
     multi::separated_list0,
     sequence::{preceded, separated_pair, terminated, tuple},
-    AsChar, Parser,
 };
 use primitive_types::U256;
 use sha3::{Digest, Keccak256};
@@ -1048,7 +1048,7 @@ impl Parameters {
         let mut values = BTreeMap::default();
         for p in self.iter() {
             let (number, n) = match p {
-                Parameter::Value(n) => (n.as_i128()?, n),
+                Parameter::Value(n) => (n.as_uint256()?, n),
                 _ => continue,
             };
             values.insert(("value", number), n.clone());
@@ -1075,8 +1075,8 @@ impl Parameters {
         let mut values = BTreeMap::default();
         for p in self.iter() {
             let (k, number, n) = match p {
-                Parameter::Gas(n) => ("gas", n.as_i128()?, n),
-                Parameter::GasLimit(n) => ("gasLimit", n.as_i128()?, n),
+                Parameter::Gas(n) => ("gas", n.as_uint256()?, n),
+                Parameter::GasLimit(n) => ("gasLimit", n.as_uint256()?, n),
                 _ => continue,
             };
             // Also return the actual key for error reporting
@@ -1105,7 +1105,7 @@ impl Parameters {
         let mut values = BTreeMap::default();
         for p in self.iter() {
             let (number, n) = match p {
-                Parameter::GasPrice(n) => (n.as_i128()?, n),
+                Parameter::GasPrice(n) => (n.as_uint256()?, n),
                 _ => continue,
             };
             values.insert(("gasPrice", number), n.clone());
--- a/components/eip681/src/request.rs
+++ b/components/eip681/src/request.rs
@@ -52,6 +52,52 @@ pub enum TransactionRequest {
 }
 
 impl TransactionRequest {
+    /// Construct a `TransactionRequest` from `NativeRequest` parts, if possible.
+    pub fn from_native_request_parts(
+        schema_prefix: &str,
+        has_pay: bool,
+        chain_id: Option<u64>,
+        recipient: &str,
+        value: Option<U256>,
+        gas_limit: Option<U256>,
+        gas_price: Option<U256>,
+    ) -> Result<Self, Error> {
+        let chain_id = chain_id.map(|id| format!("@{id}")).unwrap_or_default();
+        let value = value.map(|v| format!("value={v}"));
+        let gas_limit = gas_limit.map(|v| format!("gasLimit={v}"));
+        let gas_price = gas_price.map(|v| format!("gasPrice={v}"));
+        let params: String = [value, gas_limit, gas_price]
+            .into_iter()
+            .flatten()
+            .collect::<Vec<_>>()
+            .join("&");
+        let pay = if has_pay { "pay-" } else { "" };
+        let query = if params.is_empty() {
+            String::new()
+        } else {
+            format!("?{params}")
+        };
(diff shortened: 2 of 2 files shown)


components/eip681/src/request.rs
```diff
@@ -52,6 +52,43 @@ pub enum TransactionRequest {
 }
 
 impl TransactionRequest {
+    /// Construct a `TransactionRequest` from `NativeRequest` parts, if possible.
+    pub fn from_native_request_parts(
+        schema_prefix: &str,
```
schell, line 57: This forgets the possibility of the "pay-" portion of a schema prefix. I'm still thinking of a good way to represent that. I might just have to pull in all of `SchemaPrefix` here.

components/eip681/src/request.rs
```diff
@@ -52,6 +52,47 @@ pub enum TransactionRequest {
...
+    ) -> Result<Self, Error> {
+        let chain_id = chain_id.map(|id| format!("@{id}")).unwrap_or_default();
+        let value = value.map(|v| format!("value={v}")).unwrap_or_default();
+        let gas_limit = gas_limit
+            .map(|v| format!("gasLimit={v}"))
+            .unwrap_or_default();
+        let gas_price = gas_price
+            .map(|v| format!("gasPrice={v}"))
+            .unwrap_or_default();
+        let params = [value, gas_limit, gas_price].join("&");
```
nuttycom: bug: This will result in a sequence of `&` characters if gas limit, gas price, or values are None in the inputs because they get mapped to the empty string, and then concatenated. The `unwrap_or_default`s here are incorrect.

nuttycom (changes requested):
One bug, plus Clippy failures.

components/eip681/src/request.rs
```diff
@@ -343,6 +389,16 @@ impl TryFrom<&RawTransactionRequest> for Erc20Request {
 }
 
 impl Erc20Request {
+    /// Retuns the schema prefix.
```
nullcopy:
```suggestion
    /// Returns the schema prefix.
```

components/eip681/src/request.rs
```diff
@@ -52,6 +52,47 @@ pub enum TransactionRequest {
...
+        let gas_price = gas_price
+            .map(|v| format!("gasPrice={v}"))
+            .unwrap_or_default();
+        let params = [value, gas_limit, gas_price].join("&");
+        let pay = if has_pay { "pay-" } else { "" };
+        let req = format!("{schema_prefix}:{pay}{recipient}{chain_id}?{params}",);
+        Self::parse(&req)
+    }
+
+    // Construct a `TransactionRequest` from `Erc20Request` parts, if possible.
```
nullcopy:
```suggestion
    /// Construct a `TransactionRequest` from `Erc20Request` parts, if possible.
```

components/eip681/src/request.rs
```diff
@@ -452,4 +508,122 @@ mod test {
...
+        assert_eq!(
+            erc20.token_contract_address(),
+            erc20_b.token_contract_address()
+        );
+        assert_eq!(erc20.recipient_address(), erc20_b.recipient_address());
+        assert_eq!(erc20.value_atomic(), erc20_b.value_atomic());
+    }
+
+    #[test]
+    fn from_erc20_request_with_pay_parts_roundtrip() {
```
nullcopy: FYI, none of these tests cover cases where any of the optional parameters are `None`. Might be nice to have at least one test covering edge cases relating to optional arguments

components/eip681/src/request.rs
schell: Ah, yeah, good catch.

components/eip681/src/request.rs
schell: I think I'll convert these into one proptest. Thanks :)

nullcopy (approved):
utACK df6b7c498a727802cae929793f5804063635f2f5

components/eip681/src/parse.rs
```diff
@@ -1151,7 +1149,8 @@ pub struct SchemaPrefix {
 
 impl core::fmt::Display for SchemaPrefix {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str("ethereum:")?;
+        f.write_str(&self.prefix)?;
```
str4d, line 1154: This is a fix for a bug in #2134 that was not caught in review.

In general, please separate bugfixes from feature additions (same PR is fine if the bugfix is needed for the PR, as long as they are in separate commits), so it is clear that it was an intentional bugfix (and so the bugfix can be cherry-picked in the event that the prior PR had already been released, in case the feature being added herein were a breaking change that couldn't also be cherry-picked).

components/eip681/src/request.rs
```diff
@@ -459,4 +513,125 @@ mod test {
...
+        "0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed",
+        "0xdbF03B407c01E7cD3CBea99509d93f8DDDC8C6FB",
+    ];
+
+    fn arb_erc55_address() -> impl Strategy<Value = &'static str> {
+        prop::sample::select(ERC55_ADDRESSES)
+    }
+
+    fn arb_schema_prefix() -> impl Strategy<Value = &'static str> {
+        prop::sample::select(&["ethereum", "zcash", "chaos_emerald"][..])
```
str4d: Do not use `zcash` here. EIP 681 is incompatible with ZIP 321, and we should not imply otherwise here.
```suggestion
        prop::sample::select(&["ethereum", "chaos_emerald"][..])
```

components/eip681/src/request.rs
```diff
@@ -459,4 +513,125 @@ mod test {
...
+    }
+
+    fn arb_u256() -> impl Strategy<Value = U256> {
+        any::<u64>().prop_map(U256::from)
```
str4d: This doesn't cover the full `U256` range. The most useful way to fix this (i.e. the way that maps closest to how `proptest` will do its reductions) would be to sample two `u128`s, construct two `U256`s, and then return `(upper << 128) | lower`.

str4d (changes requested):
Reviewed df6b7c498a727802cae929793f5804063635f2f5. utACK after test fixes.

components/eip681/src/parse.rs
schell: I've separated the commits now.

components/eip681/src/parse.rs
```diff
@@ -1048,7 +1048,7 @@ impl Parameters {
         let mut values = BTreeMap::default();
         for p in self.iter() {
             let (number, n) = match p {
-                Parameter::Value(n) => (n.as_i128()?, n),
+                Parameter::Value(n) => (n.as_uint256()?, n),
```
str4d, line 1051: Are these more bugfixes identified after my suggested proptest change? These fixes should also be separated out from the feature commit (the commit message for which mentions the change but does not indicate that it is a bugfix).

In this case, given that the test that found the bug requires the new feature, I'd be fine with adding the feature and test, and then applying the unrelated fix in a follow-up commit (this is functionally equivalent to red-green bugfixing which we also do when we figure out an obscure bug: add a test exercising the bug in a commit that can be verified to fail, then in the next commit fix it).

PR #2174, opened 2026-02-26
Release pczt 0.4.1

nuttycom:


pczt/Cargo.toml
```diff
@@ -90,6 +90,9 @@ sapling = [
 ## Enables functionality that requires Zcash transparent protocol types.
 transparent = ["dep:secp256k1", "dep:transparent", "dep:zcash_script"]
 
+## Enables transparent input signing support.
+transparent-inputs = ["transparent", "transparent/transparent-inputs"]
+
```
str4d: This new feature flag doesn't make sense. All that `transparent/transparent-inputs` does is enable `dep:secp256k1`, which we already do in the `transparent` feature flag. We should instead enable `transparent/transparent-inputs` in `transparent`.
```suggestion
transparent = [
    "dep:secp256k1",
    "dep:transparent",
    "dep:zcash_script",
    "transparent/transparent-inputs",
]

```

pczt/src/roles/signer/mod.rs
```diff
@@ -326,6 +334,7 @@ pub enum Error {
     SaplingVerify(sapling::pczt::VerifyError),
     TransparentExtract(transparent::pczt::TxExtractorError),
     TransparentParse(transparent::pczt::ParseError),
+    #[cfg(feature = "transparent-inputs")]
```
str4d: I disagree that this is a backwards-compatible change. With my other suggestion, this also becomes unnecessary, as the entire `signer` module requires `transparent`.
```suggestion
```
(and similarly above).

str4d (changes requested):
Reviewed 00160591b1f63df6df4cb7185bd2d93a0c35474f.

pczt/src/roles/signer/mod.rs
nuttycom: Ah, I forgot to cut that out of the PR that does more rework to this error type.

pczt/src/roles/signer/mod.rs
```diff
@@ -47,9 +47,12 @@ pub struct Signer {
     sapling: sapling::pczt::Bundle,
     orchard: orchard::pczt::Bundle,
     /// Cached across multiple signatures.
+    #[allow(dead_code)] // only used outside constructor when transparent-inputs is enabled
     tx_data: TransactionData<EffectsOnly>,
+    #[allow(dead_code)] // only used outside constructor when transparent-inputs is enabled
     txid_parts: TxDigests<Blake2bHash>,
     shielded_sighash: [u8; 32],
+    #[allow(dead_code)] // only used outside constructor when transparent-inputs is enabled
```
str4d: These shouldn't be necessary either now? Or are the comments just out-of-date?

str4d (approved):
utACK 93bac2f520630db4ca3afcbccc8a0c234f918728 with question.

pczt/src/roles/signer/mod.rs
nuttycom: Ah, yeah, that's correct.

PR #2179, opened 2026-02-27
Generalize Pczt -> TransactionData extraction & eliminate io-finalizer dependency on signer role.

### LINKED ISSUE #2180: pczt: Eliminate dependency of `io_finalizer` on the `signer` role.
There is also a substantial opportunity here for eliminating redundant code for PCZT bundle parsing and transaction data extraction.

nuttycom:
The individual commits here may be interesting, but in general I think it's better to review the overall diff.

Depends on https://github.com/zcash/orchard/pull/477 and https://github.com/zcash/sapling-crypto/pull/174
Closes #2180.
Closes COR-1050.


pczt/src/lib.rs
```diff
@@ -88,24 +107,163 @@ impl Pczt {
...
+}
+
+/// Helper to produce the correct sighash for a PCZT.
+#[cfg(any(feature = "io-finalizer", feature = "signer"))]
+pub(crate) fn sighash(
+    tx_data: &TransactionData<EffectsOnly>,
+    signable_input: &SignableInput,
+    txid_parts: &TxDigests<Blake2bHash>,
+) -> [u8; 32] {
+    v5_signature_hash(tx_data, signable_input, txid_parts)
```
str4d, line 251: The AI split apart the version check from the sighash logic where previously they were always side-by-side, and dropped the TODO that previously annotated them collectively without resolving it.

Re-introduce the TODO:
```suggestion
    // TODO: Pick sighash based on tx version.
    v5_signature_hash(tx_data, signable_input, txid_parts)
```

and document this helper method to ensure that the caller is aware of the precondition on its use (namely, that it cannot be used on anything except the output of `Pczt::extract_tx_data`, because that's where the version check now happens).

pczt/src/roles/signer/mod.rs
```diff
@@ -357,95 +338,20 @@ impl Signer {
...
-        ))]
-        Zatoshis::ZERO,
-        transparent_bundle,
-        None,
-        sapling_bundle,
-        orchard_bundle,
-    ))
-}
-
-pub struct EffectsOnly;
```
str4d, line 406: Document this removal in the changelog, with a "(use `pczt::EffectsOnly` instead)" suffix.

pczt/src/lib.rs
```diff
@@ -88,24 +107,163 @@ impl Pczt {
...
+pub(crate) struct ParsedPczt<A: Authorization> {
+    pub(crate) global: Global,
+    pub(crate) transparent: ::transparent::pczt::Bundle,
+    pub(crate) sapling: ::sapling::pczt::Bundle,
+    pub(crate) orchard: ::orchard::pczt::Bundle,
+    pub(crate) tx_data: TransactionData<A>,
+}
+
+#[cfg(any(feature = "io-finalizer", feature = "signer"))]
+pub struct EffectsOnly;
```
str4d, line 228: Document this addition in the changelog.

pczt/src/lib.rs
```diff
@@ -88,24 +107,163 @@ impl Pczt {
...
+    v5_signature_hash(tx_data, signable_input, txid_parts)
+        .as_ref()
+        .try_into()
+        .expect("correct length")
+}
+
+/// Errors that can occur while parsing PCZT bundles and extracting transaction data.
+#[cfg(any(feature = "io-finalizer", feature = "signer", feature = "tx-extractor"))]
+#[derive(Debug)]
+pub enum ExtractError {
```
str4d, line 261: Didn't we decide to make new error enums non-exhaustive?

str4d (changes requested):
Reviewed caac5ae6d2a7faa13f1ba77839f14fa7bc29bdae.

In addition to my review comments, `rustfmt` failures are present.

Cargo.toml
```diff
@@ -228,3 +228,7 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(zcash_unstable, values("zfuture", "nu7"))',
   'cfg(live_network_tests)',
 ] }
+
+[patch.crates-io]
+sapling = { package = "sapling-crypto", git = "https://github.com/zcash/sapling-crypto.git", rev = "4f95c2286dbe90f05e6f44d634bc2924b992fab4" }
+orchard = { package = "orchard", git = "https://github.com/zcash/orchard.git", rev = "6b12c77260aa7fac0d804983fc31b71b584d48e0" }
```
str4d, line 234: These are breaking changes, so all of the changelogs will need to be bumped to say "Migrated to `orchard 0.13`, `sapling-crypto 0.7`" (as that will be the next breaking release of these crated).

Normally I'd do this in the PR I add the patch, but it could also be done in the commit that replaces these patches with the eventual published crates.

PR #2182, opened 2026-03-02
Update `zcash_script` dep from 0.4.2 to 0.4.3

upbqdn:


Cargo.toml
```diff
@@ -78,7 +78,7 @@ crypto-common = { version = "=0.2.0-rc.1" } # later RCs require edition2024
 ripemd = { version = "0.1", default-features = false }
 secp256k1 = { version = "0.29", default-features = false, features = ["alloc"] }
 transparent = { package = "zcash_transparent", version = "0.6.2", path = "zcash_transparent", default-features = false }
-zcash_script = { version = "0.4.2", default-features = false }
+zcash_script = { version = "0.4.3", default-features = false }
```
str4d, line 81: Why is changing the minimum required version necessary here? This won't do anything until breaking releases are made. Isn't it sufficient for downstreams (and here as well) to just run `cargo update -p zcash_script`?

PR #2183, opened 2026-03-03
Prune dependencies and maintain README graph

str4d:


.github/helpers/check-dep-graph.py
```diff
@@ -0,0 +1,109 @@
...
+            continue
+
+        while len(crate_stack) > depth:
+            crate_stack.pop()
+        if crate_stack[-1] in CRATES_IN_GRAPH and crate in CRATES_IN_GRAPH:
+            cargo_edges.add((crate_stack[-1], crate))
+        crate_stack.append(crate)
+
+    # Check for missing edges.
+    missing_edges = cargo_edges.difference(readme_edges_set)
```
str4d, line 94: Currently this lint requires direct edges for all dependencies, and ignores the existence of transitive dependencies. We have not been consistent in the README graph on this; for some crates, we show direct dependencies even if there are also transitive dependency paths.

If we don't want to show all edges, I can add a filter here to allow missing direct edges if a transitive edge exists. This would I think allow the current mixed behaviour to continue.

nuttycom: I think I would rather see the *direct* dependencies; when looking at the graph, the transitive dependencies are obvious.

nuttycom (approved):
utACK 415b835a4eadb97e788fdb0b345b762b02f4f4bf

.github/helpers/check-dep-graph.py
str4d: I think you misunderstood me. The linter is only saying to add direct dependencies where the `Cargo.toml` files express direct dependencies. That is still more edges than are currently present in the graph, because currently we present them as transitive dependencies, not the direct dependencies they are.

I'll push a commit to satisfy the linter so you can see what it looks like.

str4d: I pushed two commits:
- 0c5296f3be3511d5ca64c000e2170629ed69e633 adds every direct edge, like the linter and @nuttycom were asking form
- 322067aefd48ad94f1e2d0f7277e43f1a15c1a5f then comments out every direct edge where the effective relationship is also maintained by other transitive edges (and adjusts the linter to allow this while still having it track all direct edges).

I think the second commit's graph is clearer, but I'm happy to remove it and do the first commit's graph instead.

PR #2185, opened 2026-03-03
Migrate to `Swatinem/rust-cache` action.

nuttycom:
This eliminates some manual cache configuration and ensures correct use of `restore-keys`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,15 +65,10 @@ jobs:
           all-pools: ${{ matrix.all-pools || false }}
           all-features: ${{ matrix.all-features || false }}
           extra-features: ${{ matrix.extra_flags || '' }}
-      - uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
+      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -65,15 +65,9 @@ jobs:
...
+      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-msrv-${{ hashFiles('**/Cargo.lock') }}
+          shared-key: msrv-${{ matrix.state }}
```
str4d: Bugs:
- Using `shared-key` without including `runner.os` means that the cache will be shared across targets (Linux, macOS, Windows).
- This will generate and use a separate cache for every single state we test, which will cause way more cache overflows than we want (we have 8 states in this job alone, which will cause anywhere between 6 and 12 GiB of cache).

In terms of what we want our caches to split across:
- We definitely want a new cache per `runner.os`.
- We *maybe* want a separate cache for MSRV and latest builds (i.e. per Rust version)? That can be achieved with `key: msrv`.

nuttycom: This is incorrect; the OS is included in the key - `shared-key` is always just used as a prefix: https://github.com/Swatinem/rust-cache/blob/master/src/config.ts#L92-L97

str4d (approved):
utACK 7dd68dfcc392ce61f0cbc0f0257a21c35a898765

PR #2190, opened 2026-03-04
Fix documentation in unified address and fvk modules

Cosmos-Harry:
- Add spec-referenced documentation to the Receiver enum in address.rs
- Clarify the purpose of the Unknown variant for forward compatibility
- Fix incorrect method reference in Ufvk::items_as_parsed doc comment
  (`Ufvk::receivers` → `Ufvk::items`)

No functional changes.


nuttycom (approved):
doc-only ACK

PR #2191, opened 2026-03-04
zip321: Return Result<Self, PaymentError> from Payment::new

### LINKED ISSUE #1794: `zip321::Payment::new` should return a `Result` instead of an `Option`
Currently the caller has to make sense of the `None` result returned by `new()` while there's a very specific meaning to it which is attempting to send a shielded memo to a transparent recipient. 

Example:

```Rust
   let payment = zip321::Payment::new(recipient, amount, memo, None, None, vec![])
            .ok_or_else(|| {
                InputSelectorError::Proposal(ProposalError::SendsMemoToTransparentRecipient)
            })?;

```

underlying API:

```Rust
    /// Constructs a new [`Payment`] from its constituent parts.
    ///
    /// Returns `None` if the payment requests that a memo be sent to a recipient that cannot
    /// receive a memo.
    pub fn new(
        recipient_address: ZcashAddress,
        amount: Zatoshis,
        memo: Option<MemoBytes>,
        label: Option<String>,
        message: Option<String>,
        other_params: Vec<(String, String)>,
    ) -> Option<Self> {
```

Desired API:

```Rust
    /// Constructs a new [`Payment`] from its constituent parts.
    ///
    /// Returns `None` if the payment requests that a memo be sent to a recipient that cannot
    /// receive a memo.
    pub fn new(
        recipient_address: ZcashAddress,
        amount: Zatoshis,
        memo: Option<MemoBytes>,
        label: Option<String>,
        message: Option<String>,
        other_params: Vec<(String, String)>,
    ) -> Result<Self, Error> {
```

nuttycom:
Changes `Payment::new` to return `Result<Self, PaymentError>` instead of `Option<Self>`, giving callers a meaningful error when payment construction fails. The new `PaymentError` enum distinguishes between sending a memo to a transparent recipient and creating a zero-valued transparent output.

Closes #1794.

### DIFF
--- a/components/zip321/src/lib.rs
+++ b/components/zip321/src/lib.rs
@@ -23,6 +23,44 @@ use zcash_protocol::{
     value::Zatoshis,
 };
 
+/// Errors that may be produced in constructing a [`Payment`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PaymentError {
+    /// A memo was provided for a recipient address that cannot receive a memo.
+    TransparentMemo,
+    /// A zero-valued output was requested for a transparent recipient address, which is
+    /// disallowed by the Zcash consensus rules.
+    ZeroValuedTransparentOutput,
+}
+
+impl Display for PaymentError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PaymentError::TransparentMemo => {
+                write!(f, "Cannot send a memo to a transparent recipient address")
+            }
+            PaymentError::ZeroValuedTransparentOutput => write!(
+                f,
+                "Zero-valued transparent outputs are disallowed by consensus"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for PaymentError {}
+
+impl PaymentError {
+    /// Adds a payment index to this error to produce a [`Zip321Error`].
+    pub fn with_index(self, index: usize) -> Zip321Error {
+        match self {
+            PaymentError::TransparentMemo => Zip321Error::TransparentMemo(index),
+            PaymentError::ZeroValuedTransparentOutput => {
+                Zip321Error::ZeroValuedTransparentOutput(index)
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -716,7 +716,7 @@ where
             None,
             vec![],
         )
-        .ok_or(Error::MemoForbidden)?,
+        .map_err(|_| Error::MemoForbidden)?,
```
nuttycom: This mapping is now not always correct; the Error type should be updated to represent both error conditions (if they're not already representable).

components/zip321/src/lib.rs
```diff
@@ -166,13 +204,13 @@ impl Payment {
...
         message: Option<String>,
         other_params: Vec<(String, String)>,
-    ) -> Option<Self> {
-        if (recipient_address.is_transparent_only() && amount == Some(Zatoshis::ZERO))
-            || (memo.is_some() && !recipient_address.can_receive_memo())
-        {
-            None
+    ) -> Result<Self, PaymentError> {
+        if memo.is_some() && !recipient_address.can_receive_memo() {
+            Err(PaymentError::TransparentMemo)
```
nullcopy, line 209: (nit) this feels somewhat brittle. If future behavior of `can_receive_memo()` returns false for any other reason than the receiver being transparent, this error would be misleading.

components/zip321/src/lib.rs
```diff
@@ -23,6 +23,44 @@ use zcash_protocol::{
...
 
+/// Errors that may be produced in constructing a [`Payment`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PaymentError {
+    /// A memo was provided for a recipient address that cannot receive a memo.
+    TransparentMemo,
+    /// A zero-valued output was requested for a transparent recipient address, which is
+    /// disallowed by the Zcash consensus rules.
+    ZeroValuedTransparentOutput,
+}
```
schell, line 34: `snafu`!

Issue #1892, opened 2025-08-05
Add support for import of standalone P2SH addresses & key material.


nuttycom:
* P2SH address
* Redeem script
* Imported pubkeys corresponding to signatures that need to be applied.

PR #2194, opened 2026-03-05
Add support for standalone P2SH address import

nullcopy:
resolves https://github.com/zcash/librustzcash/issues/1892
resolves https://github.com/zcash/librustzcash/issues/2196
resolves COR-197
resolves COR-1062

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -3017,6 +3017,11 @@ pub trait WalletWrite: WalletRead {
     ///
     /// [`create_proposed_transactions`]: crate::data_api::wallet::create_proposed_transactions
     /// [`propose_shielding`]: crate::data_api::wallet::propose_shielding
+    ///
+    /// # Spending limitations
+    ///
+    /// P2PKH-in-P2SH scripts are unsupported by PCZT at this time, so the only way to spend
+    /// from such an address is to use the [`create_proposed_transactions`] signing path.
     #[cfg(feature = "transparent-key-import")]
     fn import_standalone_transparent_script(
         &mut self,
--- a/zcash_client_backend/src/data_api/testing/transparent.rs
+++ b/zcash_client_backend/src/data_api/testing/transparent.rs
@@ -811,7 +811,7 @@ where
 
     // Build SpendingKeys with the standalone key for the P2PKH address.
     let mut standalone_keys = HashMap::new();
-    standalone_keys.insert(taddr, secret_key);
+    standalone_keys.insert(taddr, vec![secret_key]);
     let spending_keys = SpendingKeys::new(
         account.usk().clone(),
         #[cfg(feature = "transparent-key-import")]
@@ -853,6 +853,26 @@ where
         &taddr,
         ConfirmationsPolicy::MIN,
         &Balance::ZERO,
+    );
+
+    // The shielded balance should now include the value minus the fee.
+    let fee = st
+        .get_tx_from_history(*txids.first())
+        .unwrap()
+        .unwrap()
+        .fee_paid()
+        .expect("fee should be known for wallet-created transactions");
+    let summary = st
+        .wallet()
+        .get_wallet_summary(ConfirmationsPolicy::MIN)
+        .unwrap()
+        .unwrap();
+    let account_balance = summary.account_balances().get(&account_id).unwrap();
+    assert_eq!(
+        account_balance
+            .sapling_balance()
+            .change_pending_confirmation(),
+        (value - fee).unwrap(),
     );
 }
 
@@ -1061,4 +1081,119 @@ where
         .unwrap();
     assert_eq!(utxos.len(), 1);
     assert_eq!(utxos[0].value(), value);
+}
+
+/// Tests spending from a standalone P2SH (multisig) address by shielding its balance.
+#[cfg(feature = "transparent-key-import")]
+pub fn spend_from_standalone_p2sh<DSF>(dsf: DSF, cache: impl TestCache)
+where
+    DSF: DataStoreFactory,
+{
+    use crate::data_api::wallet::{self, SpendingKeys};
+    use std::collections::HashMap;
+
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(dsf)
+        .with_block_cache(cache)
+        .with_account_from_sapling_activation(BlockHash([0; 32]))
+        .build();
+
+    let account = st.test_account().cloned().unwrap();
+    let account_id = account.id();
+    // Build the redeem script and get the signing key.
+    let (redeem_script, secret_key) = build_test_redeem_script();
+
+    // Import the P2SH address.
+    st.wallet_mut()
+        .import_standalone_transparent_script(account_id, redeem_script.clone())
+        .unwrap();
+
+    // Derive the P2SH address.
+    let script_pubkey = sh(&redeem_script);
+    let taddr = TransparentAddress::from_script_pubkey(&script_pubkey).expect("valid P2SH address");
+
+    // Initialize chain data with blocks (needed for shielding transaction creation).
+    let not_our_key = ExtendedSpendingKey::master(&[]).to_diversifiable_full_viewing_key();
+    let not_our_value = Zatoshis::const_from_u64(10000);
+    let (start_height, _, _) =
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    for _ in 1..10 {
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    }
+    st.scan_cached_blocks(start_height, 10);
+
+    // Create a fake UTXO at the P2SH address.
+    let value = Zatoshis::from_u64(100000).unwrap();
+    let height = st.wallet().chain_height().unwrap().unwrap();
+    let txout = TxOut::new(value, taddr.script().into());
+    let utxo = WalletTransparentOutput::from_parts(OutPoint::fake(), txout, Some(height)).unwrap();
+    st.wallet_mut()
+        .put_received_transparent_utxo(&utxo)
+        .unwrap();
+
+    // Build SpendingKeys with the standalone key for the P2SH address.
+    let mut standalone_keys = HashMap::new();
+    standalone_keys.insert(taddr, vec![secret_key]);
+    let spending_keys = SpendingKeys::new(
+        account.usk().clone(),
+        #[cfg(feature = "transparent-key-import")]
+        standalone_keys,
+    );
+
+    // Shield the P2SH UTXO.
+    let input_selector = GreedyInputSelector::new();
+    let change_strategy = standard::SingleOutputChangeStrategy::new(
+        StandardFeeRule::Zip317,
+        None,
+        ShieldedProtocol::Sapling,
+        DustOutputPolicy::default(),
+    );
+
+    let prover = ::zcash_proofs::prover::LocalTxProver::bundled();
+    let network = *st.network();
+    let txids = wallet::shield_transparent_funds(
+        st.wallet_mut(),
+        &network,
+        &prover,
+        &prover,
+        &input_selector,
+        &change_strategy,
+        value,
+        &spending_keys,
+        &[taddr],
+        account_id,
+        ConfirmationsPolicy::MIN,
+    )
+    .unwrap();
+
+    assert!(!txids.is_empty());
+
+    // The wallet should have zero transparent balance after shielding.
+    check_balance::<DSF>(
+        &st,
+        &account,
+        &taddr,
+        ConfirmationsPolicy::MIN,
+        &Balance::ZERO,
+    );
+
+    // The shielded balance should now include the value minus the fee.
+    let fee = st
+        .get_tx_from_history(*txids.first())
+        .unwrap()
+        .unwrap()
+        .fee_paid()
+        .expect("fee should be known for wallet-created transactions");
+    let summary = st
+        .wallet()
+        .get_wallet_summary(ConfirmationsPolicy::MIN)
+        .unwrap()
+        .unwrap();
+    let account_balance = summary.account_balances().get(&account_id).unwrap();
+    assert_eq!(
+        account_balance
+            .sapling_balance()
+            .change_pending_confirmation(),
+        (value - fee).unwrap(),
+    );
 }
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -96,6 +96,9 @@ use {
     std::collections::HashMap,
 };
 
+#[cfg(feature = "transparent-key-import")]
+use zcash_script::script::{self as zs_script, Evaluable};
+
 #[cfg(feature = "pczt")]
 use {
     crate::data_api::error::PcztError,
@@ -853,7 +856,7 @@ struct StepResult<AccountId> {
 pub struct SpendingKeys {
     usk: UnifiedSpendingKey,
     #[cfg(feature = "transparent-key-import")]
-    standalone_transparent_keys: HashMap<TransparentAddress, secp256k1::SecretKey>,
+    standalone_transparent_keys: HashMap<TransparentAddress, Vec<secp256k1::SecretKey>>,
 }
 
 impl SpendingKeys {
@@ -862,7 +865,7 @@ impl SpendingKeys {
         usk: UnifiedSpendingKey,
         #[cfg(feature = "transparent-key-import")] standalone_transparent_keys: HashMap<
             TransparentAddress,
-            secp256k1::SecretKey,
+            Vec<secp256k1::SecretKey>,
         >,
     ) -> Self {
         Self {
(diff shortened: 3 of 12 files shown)


zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -183,19 +183,27 @@ CREATE TABLE "addresses" (
...
+                AND redeem_script IS NULL
             )
             OR (
                 cached_transparent_receiver_address IS NOT NULL
-                AND (transparent_child_index IS NULL) == (imported_transparent_receiver_pubkey IS NOT NULL)
+                AND (
+                    (transparent_child_index IS NULL) == (
+                        (imported_transparent_receiver_pubkey IS NOT NULL)
+                        OR (redeem_script IS NOT NULL)
+                    )
```
schell, line 212: I know there's no set precedent here yet, but I could use some comments here about this boolean logic, as it's hard for my mind to flip these things back and forth and keep track of what it means.

That said - do we care about catching the case where _both_ are not null? `imported_transparent_receiver_pubkey IS NOT NULL AND redeem_script IS NOT NULL`?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1203,14 +1261,12 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
+        // `TransparentKeyScope`. We default to External as a placeholder.
         let key_scope = KeyScope::decode(key_scope_code)?
             .as_transparent()
-            .ok_or_else(|| {
-                SqliteClientError::CorruptedData(format!(
-                    "Invalid key scope code for transparent received output: {}",
-                    key_scope_code
-                ))
-            })?;
+            .unwrap_or(TransparentKeyScope::from(zip32::Scope::External));
```
schell: Maybe a comment about why this is not an error would be helpful. Like - how is it that downstream consumers don't care that there's no derivation scope?

schell (commented):
Just had some comments. Not really my wheelhouse ATM.

zcash_client_sqlite/src/wallet/db.rs
nullcopy: > That said - do we care about catching the case where both are not null?

Yes, essentially, this check is supposed to ensure that if an address is imported, then it may not have derivation information, and vice versa.

There is actually a bug here. Because the redeem script may be non-null for derived keys  (e.g. upcoming ZIP 48 pr), then presence of a redeem_script is not enough to ensure the address is imported. We also need to check the key_scope to ensure that, which I have fixed in the next push.

> but I could use some comments here about this boolean logic

Good catch. I've updated the documentation above with a short description

zcash_client_sqlite/src/wallet/transparent.rs
nullcopy: Good point. Looking at it again, I agree the error was helpful. Changing it back

nullcopy: Ah no, I remember now. The error was a problem, because it was throwing whenever an address had foreign key scope. This was a bug, causing standalone addresses to be ignored. This wasn't noticed until I added tests around standalone addresses.

The better fix is to modify the function (and all its callers) to allow an optional keyscope, so that standalone addresses with no `TransparentKeyScope` won't be excluded. I've fixed this in d4fefd9993e5d146592f3ac8f24622c17311c1b9

nullcopy:
Force pushed to address comments from @schell, rebase on main, and fix failing tests

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -183,19 +187,27 @@ CREATE TABLE "addresses" (
     receiver_flags INTEGER NOT NULL,
     transparent_receiver_next_check_time INTEGER,
     imported_transparent_receiver_pubkey BLOB,
+    redeem_script BLOB,
```
nullcopy: Since this is not needed by ZIP 48, its only purpose is for imported accounts.

TODO: make this imported_transparent_receiver_script with similar normalization check

zcash_transparent/src/builder.rs
```diff
@@ -1046,4 +1136,113 @@ mod tests {
...
+        // Build a 2-of-2 multisig redeem script:
+        // OP_2 PUSHBYTES_33 <pubkey1> PUSHBYTES_33 <pubkey2> OP_2 OP_CHECKMULTISIG
+        let mut script_bytes: Vec<u8> = Vec::new();
+        script_bytes.push(0x52); // OP_2
+        script_bytes.push(0x21); // PUSHBYTES_33
+        script_bytes.extend_from_slice(&pk1.serialize());
+        script_bytes.push(0x21); // PUSHBYTES_33
+        script_bytes.extend_from_slice(&pk2.serialize());
+        script_bytes.push(0x52); // OP_2
+        script_bytes.push(0xae); // OP_CHECKMULTISIG
```
nuttycom: nit: could one of the builders in https://docs.rs/zcash_script/latest/zcash_script/pattern/index.html be used here instead of manual construction?

zcash_client_backend/src/wallet.rs
```diff
@@ -763,16 +796,31 @@ impl TransparentAddressMetadata {
         }
     }
 
-    /// Returns a [`TransparentAddressMetadata`] with [`TransparentAddressSource::Standalone`] source
-    /// information and the specified exposure height.
+    /// Returns a [`TransparentAddressMetadata`] with [`TransparentAddressSource::StandalonePubkey`]
+    /// source information for a P2PKH address and the specified exposure height.
     #[cfg(feature = "transparent-key-import")]
-    pub fn standalone(
+    pub fn standalone_p2pkh(
```
nullcopy, line 802: TODO from R&D meeting: lets not add support for p2sh-p2pkh. P2PKH use is already covered with native P2PKH, and we don't need to take on additional script types which don't add value to justify the maintenance burden

zcash_transparent/src/builder.rs
```diff
@@ -252,56 +252,67 @@ impl TransparentInputInfo {
...
             SpendInfo::P2pkh { .. } => {
-                // P2PKH `script_sig` format is:
-                // - PushData(signature || sigtype)
-                // - PushData(pubkey)
+                let fake_sig = pv::push_value(&[0; 72 + 1]).expect("short enough");
                 let fake_pubkey = pv::push_value(&[0; secp256k1::constants::PUBLIC_KEY_SIZE])
                     .expect("short enough");
-                Some(script::Component(vec![fake_sig, fake_pubkey]).byte_len())
+                let script_len = script::Component(vec![fake_sig, fake_pubkey]).byte_len();
+                let prevout_len = 32 + 4;
```
nuttycom: nit: extract this as a constant so that they can be documented.

zcash_transparent/src/builder.rs
```diff
@@ -252,56 +252,67 @@ impl TransparentInputInfo {
...
+                let script_sig_len = CompactSize::serialized_size(script_len) + script_len;
+                let sequence_len = 4;
```
nuttycom: nit: extract as a constant

zcash_transparent/src/builder.rs
```diff
@@ -252,56 +252,67 @@ impl TransparentInputInfo {
...
     }
 }
 
+/// Computes the serialized input size for a P2SH input given the redeem script.
+///
+/// Returns `None` if the redeem script is not a recognized P2SH script kind (currently
+/// only P2MS-in-P2SH is supported).
+#[cfg(feature = "transparent-inputs")]
+pub fn p2sh_input_serialized_len(redeem_script: &script::FromChain) -> Option<usize> {
+    let fake_sig = pv::push_value(&[0; 72 + 1]).expect("short enough");
```
nuttycom: nit: extract the `72 + 1` as a constant and document it.

zcash_transparent/src/builder.rs
```diff
@@ -252,56 +252,67 @@ impl TransparentInputInfo {
...
+    let fake_pubkey =
+        pv::push_value(&[0; secp256k1::constants::PUBLIC_KEY_SIZE]).expect("short enough");
+
+    let script_len = redeem_script
+        .refine()
+        .ok()
+        .as_ref()
+        .and_then(solver::standard)
+        .and_then(|spend_info| match spend_info {
+            solver::ScriptKind::PubKeyHash { .. } => Some(
```
nuttycom:
```suggestion
            // p2pkh-in-p2sh; this is distinct from the ordinary p2pkh handling
            solver::ScriptKind::PubKeyHash { .. } => Some(
```

zcash_transparent/src/builder.rs
```diff
@@ -649,18 +656,99 @@ impl Bundle<Unauthorized> {
...
-                        ]))
+                        Ok(construct_script_sig(&sig, pubkey))
+                    }
+                    SpendInfo::P2sh { redeem_script } => {
+                        let refined = redeem_script
+                            .refine()
+                            .map_err(|_| Error::UnsupportedScript)?;
+                        let script_code = Script::from(redeem_script);
+                        match solver::standard(&refined) {
+                            Some(solver::ScriptKind::PubKeyHash { hash }) => {
```
nuttycom:
```suggestion
                            // p2pkh-in-p2sh
                            Some(solver::ScriptKind::PubKeyHash { hash }) => {
```

zcash_transparent/src/builder.rs
```diff
@@ -649,18 +656,99 @@ impl Bundle<Unauthorized> {
...
+                                let sighash = calculate_sighash(SignableInput {
+                                    hash_type: SighashType::ALL,
```
nuttycom: Is SIGHASH_ALL always correct here? Or do we need to document that only SIGHASH_ALL is supported? And the same question for the multisig signing below.

zcash_transparent/src/builder.rs
nuttycom: Or
```suggestion
                            #[cfg(feature = any(test, feature = "test-dependencies"))]
                            Some(solver::ScriptKind::PubKeyHash { hash }) => {
```

zcash_transparent/src/builder.rs
nuttycom: Alternately, should this branch of the match just be supported under the test-dependencies feature? That might be the right choice to start with.
```suggestion
            #[cfg(any(test, feature = "test-dependencies"))]        
            solver::ScriptKind::PubKeyHash { .. } => Some(
```

zcash_transparent/src/pczt/spend_finalizer.rs
```diff
@@ -51,6 +51,33 @@ impl super::Bundle {
                             .map_err(|_| SpendFinalizerError::UnsupportedRedeemScript)?;
 
                         match solver::standard(&redeem_script) {
+                            Some(solver::ScriptKind::PubKeyHash { hash }) => {
```
nuttycom:
```suggestion
                            #[cfg(any(test, feature = "test-dependencies"))]
                            Some(solver::ScriptKind::PubKeyHash { hash }) => {
```

zcash_client_backend/src/data_api.rs
```diff
@@ -132,6 +132,10 @@ pub mod wallet;
 #[cfg(any(test, feature = "test-dependencies"))]
 pub mod testing;
 
+/// A mapping from transparent addresses to their key scope and balance.
+#[cfg(feature = "transparent-inputs")]
+pub type TransparentBalances = HashMap<TransparentAddress, (Option<TransparentKeyScope>, Balance)>;
```
nuttycom: Instead of an Option here, let's use a bespoke enum that has `Standalone, Derived { scope: TransparentKeyScope }` as its variants. That will be more descriptive and more adaptable in the future.

zcash_client_backend/src/wallet.rs
nuttycom, line 802:
```suggestion
    #[cfg(any(test, feature = "test-dependencies"))]
    pub fn standalone_p2pkh(
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -827,6 +827,104 @@ pub(crate) fn import_standalone_transparent_pubkey<P: consensus::Parameters>(
         ],
     )?;
 
+    if rows_affected == 0 {
+        return Err(SqliteClientError::AccountUnknown);
+    }
```
nuttycom, line 832: This looks like a bug in the context of the `ON CONFLICT DO NOTHING` - are you certain that this works correctly?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -827,6 +827,104 @@ pub(crate) fn import_standalone_transparent_pubkey<P: consensus::Parameters>(
...
+            "Redeem script exceeds maximum P2SH size of {MAX_P2SH_REDEEM_SCRIPT_SIZE} bytes (got {} bytes)",
+            rs_bytes.len()
+        )));
+    }
+
+    // Do not import script types which do not have a supported spend flow.
+    if !matches!(
+        zcash_script::solver::standard(&redeem_script),
+        Some(zcash_script::solver::ScriptKind::MultiSig { .. })
+            | Some(zcash_script::solver::ScriptKind::PubKeyHash { .. })
```
nuttycom: Might need to replace this with a match expression if PubKeyHash ends up feature-flagged off.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -827,6 +827,104 @@ pub(crate) fn import_standalone_transparent_pubkey<P: consensus::Parameters>(
...
+            ":account_uuid": account_uuid.0,
+            ":key_scope": KeyScope::Foreign.encode(),
+            ":address": addr_str,
+            ":receiver_flags": ReceiverFlags::P2SH.bits(),
+            ":imported_transparent_receiver_script": &rs_bytes[..]
+        ],
+    )?;
+
+    if rows_affected == 0 {
+        return Err(SqliteClientError::AccountUnknown);
```
nuttycom, line 923: Same question here about the ON CONFLICT DO NOTHING.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1203,14 +1276,7 @@ pub(crate) fn get_transparent_balances<P: consensus::Parameters>(
...
         let key_scope_code: i64 = row.get("key_scope")?;
-        let key_scope = KeyScope::decode(key_scope_code)?
-            .as_transparent()
-            .ok_or_else(|| {
-                SqliteClientError::CorruptedData(format!(
-                    "Invalid key scope code for transparent received output: {}",
-                    key_scope_code
-                ))
-            })?;
+        let key_scope = KeyScope::decode(key_scope_code)?.as_transparent();
```
nuttycom: This looks like a bug. `as_transparent` returns an Option, but a decoding failure here isn't the same as the key scope being unknown. `key_scope_code` should be Option<i64>, right? And then you'd do key_scope_code.map(...).transpose()? and then map the result into the bespoke enum previously discussed.

zcash_client_sqlite/src/wallet.rs
nullcopy: The `PubKeyHash` variant is actually going away (per our discussion to only support native P2PKH), so this becomes a non-issue, however I've gone ahead and refactored this to a match expression since a similar situation is likely to arise if we add other script types in the future.

zcash_client_backend/src/wallet.rs
nullcopy: This is going away, per the decision not to bring in wrapped P2PKH script support

zcash_transparent/src/builder.rs
nullcopy: This goes away with the removal of wrapped P2PKH

zcash_transparent/src/builder.rs
nullcopy: This goes away with the removal of wrapped P2PKH

zcash_transparent/src/pczt/spend_finalizer.rs
nullcopy: This is going away, per the decision not to bring in wrapped P2PKH script support

zcash_client_sqlite/src/wallet/transparent.rs
nullcopy: Good catch! I don't follow why `key_scope_code` should be an option; `row.get(...)?` gets a result and unwraps it to an `i64` which seems sensible IMO

I like your suggestion of adding the bespoke enum, though, and I went ahead and added `KeyScope::as_key_origin()` for the mapping, since this is reused in a few places.

zcash_transparent/src/builder.rs
nullcopy: I used `SighashType::ALL` because that's what was already being used for the P2PKH case, and I just assumed that was a design choice for the transaction builder (I see the PCZT flow allows configurable sighash types). I'm adding a note about this in the doc comment.

zcash_client_sqlite/src/wallet.rs
nullcopy: Oh good catch! I'd added `ON CONFLICT DO NOTHING` to deal with duplicate imports. It wasn't actually necessary, as duplicate import is already handled on ~799, so I've removed it now

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: I was confused, I had forgotten that `key_scope` is a non-null column, so no `Option` needed.

nuttycom (dismissed):
utACK aa99a79b840e6f216f3bb8eef5c72e9f2dc7612d

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -534,3 +540,620 @@ where
...
+    assert!(!txids.is_empty());
+
+    // The wallet should have zero transparent balance after shielding.
+    check_balance::<DSF>(
+        &st,
+        &account,
+        &taddr,
+        ConfirmationsPolicy::MIN,
+        &Balance::ZERO,
+    );
```
daira, line 856: Check that it has the expected shielded balance.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -534,3 +540,620 @@ where
...
```
daira, line 1178: Check that it has the expected shielded balance.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -2077,7 +2103,9 @@ where
                                     address_index,
                                 } => Some((index, *scope, *address_index)),
                                 #[cfg(feature = "transparent-key-import")]
-                                TransparentAddressSource::Standalone(_) => None,
+                                TransparentAddressSource::StandalonePubkey(_) => None,
+                                #[cfg(feature = "transparent-key-import")]
+                                TransparentAddressSource::StandaloneScript(_) => None,
```
daira: Style nit: this could be an `|` pattern, avoiding the need for multiple `#[cfg(...)]` directives.

zcash_client_backend/src/wallet.rs
daira: Edit: I made an incorrect argument here about the value of P2PKH-in-P2SH being that it doesn't reveal the pubkey until it is spent from.  However, P2PKH already has that advantage. It's P2PK that doesn't, and that is no longer used.

zcash_client_sqlite/src/wallet/init/migrations/standalone_p2sh.rs
```diff
@@ -0,0 +1,303 @@
...
+                exposed_at_height INTEGER,
+                receiver_flags INTEGER NOT NULL,
+                transparent_receiver_next_check_time INTEGER,
+                imported_transparent_receiver_pubkey BLOB,
+                imported_transparent_receiver_script BLOB,
+                UNIQUE (account_id, key_scope, diversifier_index_be),
+                UNIQUE (imported_transparent_receiver_pubkey),
+                UNIQUE (imported_transparent_receiver_script),
+                CONSTRAINT ck_addr_transparent_index_consistency CHECK (
+                    (transparent_child_index IS NULL OR diversifier_index_be < x'0000000F00000000000000')
```
daira, line 63: I think this was preexisting, but to satisfy my curiosity, where does this constant ($15 \times 2^{56}$) come from?

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -183,19 +188,29 @@ CREATE TABLE "addresses" (
...
+                AND imported_transparent_receiver_script IS NULL
             )
             OR (
                 cached_transparent_receiver_address IS NOT NULL
-                AND (transparent_child_index IS NULL) == (imported_transparent_receiver_pubkey IS NOT NULL)
+                AND (
+                    (transparent_child_index IS NULL) == (
+                        key_scope = -1 AND (
+                            imported_transparent_receiver_pubkey IS NOT NULL
+                            OR imported_transparent_receiver_script IS NOT NULL
```
daira: Exactly one of `imported_transparent_receiver_pubkey` and `imported_transparent_receiver_script` should be non-NULL in this case, right? 

That can be expressed concisely as `(imported_transparent_receiver_pubkey IS NULL) != (imported_transparent_receiver_script IS NULL)`.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -243,51 +244,107 @@ pub(crate) fn get_transparent_receivers<P: consensus::Parameters>(
...
 
+            #[cfg(not(feature = "transparent-key-import"))]
+            let metadata = p2pkh_metadata()?;
+
+            #[cfg(not(feature = "transparent-key-import"))]
+            {
+                if matches!(key_scope, KeyScope::Foreign) {
+                    // Foreign-scoped addresses (standalone imports) require
+                    // transparent-key-import. Skip gracefully for DB compatibility.
+                    continue;
```
daira, line 349: Should this log a warning?

zcash_client_sqlite/src/wallet/init/migrations/standalone_p2sh.rs
nullcopy: Yep, this is a preexisting constraint. It was added in 8c0abf9a, so I just kept it as is.

I couldn't find good documentation for the constraint. At first I thought that was just a sentinel from ZIP 32 (perhaps big endian representation of max derivation index or something), but looking at the spec now, I'm unable to connect the dots. @nuttycom I'd be curious to know where this constraint comes from as well.

zcash_client_sqlite/src/wallet/db.rs
nullcopy: Actually, yours is not only more concise, it is more correct. Mine would incorrectly allow the case where each is `NOT NULL`. Thanks for this!

daira (approved):
utACK

PR #2199, opened 2026-03-07
Implement ZIP 316, Revision 2

### LINKED ISSUE #1153: zcash_address: Support Zip 316, Revision 2 (MUST-understand metadata)

nuttycom:


components/zcash_address/src/kind/unified/address.rs
```diff
@@ -219,34 +246,44 @@ pub mod testing {
     }
 
     /// Generates an arbitrary Unified address containing receivers corresponding to the provided
-    /// set of typecodes. The receivers of this address are likely to not represent valid protocol
-    /// receivers, and should only be used for testing parsing and/or encoding functions that do
-    /// not concern themselves with the validity of the underlying receivers.
+    /// set of typecodes.
     pub fn arb_unified_address_for_typecodes(
```
nuttycom, line 244: This should be modified to generate a `Vec<Uitem>` instead of a `Vec<Receiver>`, given how it is documented. Also, it should take a Set of `DataTypecode` rather than a `Vec<Typecode>` in its inputs, since duplication is not allowed.

components/zcash_address/src/kind/unified/fvk.rs
nuttycom, line 287: This should not be deleted.

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -227,9 +213,21 @@ mod tests {
...
+                match (a, b) {
+                    (Uitem::Data(a), Uitem::Data(b)) => Ivk::encoding_order(a, b),
+                    _ => core::cmp::Ordering::Equal,
+                }
+            });
+            Uivk {
+                revision: Revision::R0,
+                items,
+            }
         }
```
nuttycom, line 233: This should not be deleted.

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -268,14 +265,14 @@ mod tests {
         // Short padding (padded to 15 bytes instead of 16)
         let truncated_padding = [
             0x96, 0x73, 0x6a, 0x56, 0xbc, 0x44, 0x38, 0xe2, 0x47, 0x41, 0x1c, 0x70, 0xe4, 0x6,
-            0x87, 0xbe, 0xb6, 0x90, 0xbd, 0xab, 0x1b, 0xd8, 0x27, 0x10, 0x0, 0x21, 0x30, 0x2, 0x77,
+            0x87, 0xbe, 0xb6, 0x90, 0xbd, 0xab, 0x1b, 0xd8, 0x27, 0x10, 0x0, 0x21, 0x30, 0x2, 0x79,
             0x87, 0x0, 0x25, 0x96, 0x94, 0x8f, 0x1e, 0x39, 0xd2, 0xd8, 0x65, 0xb4, 0x3c, 0x72,
```
nuttycom, line 272: Restore.

components/zcash_address/src/kind/unified.rs
```diff
@@ -321,7 +555,43 @@ pub(crate) mod private {
...
+                                let meta = parse_must_understand_metadata(tc_val, &data)?;
+                                result.push(Uitem::Metadata(meta));
+                            }
+                        }
+                    } else {
+                        // SHOULD-understand (unknown) metadata: valid in both R0 and R2.
+                        result.push(Uitem::Metadata(MetadataItem::Unknown {
+                            typecode: tc_val,
+                            data,
+                        }));
```
nuttycom, line 624: Unknown metadata is only permitted in `tu` addresses for Revision 2; `zu` addresses prohibit it.

components/zcash_address/src/kind/unified.rs
```diff
@@ -321,7 +555,43 @@ pub(crate) mod private {
...
+                    }
+                } else {
+                    // Data typecode (0x00..=0xBF or 0xFD+).
+                    match Typecode::try_from(tc_val)? {
```
nuttycom, line 628: Why not use typecode parsing instead of the if/else above? Perform the metadata typecode checks witthin the metadata typecode match arm.

nuttycom (commented):
Flushing comments

zcash_keys/src/address.rs
```diff
@@ -180,62 +211,203 @@ impl UnifiedAddress {
...
-        );
-
-        #[cfg(feature = "sapling")]
-        let items = items.chain(
-            self.sapling
-                .as_ref()
-                .map(|pa| pa.to_bytes())
-                .map(unified::Receiver::Sapling),
-        );
+    /// Returns the expiry height metadata for this address, if present.
```
nuttycom, line 214: [P1] Enforce address expiry during payment construction

The PR parses and exposes expiry metadata, but payment selection never consults `addr.expiry_height()` or `addr.expiry_time()`. Consequently, the wallet can send to an address it knows is expired, and it does not constrain the transaction's `nExpiryHeight`. ZIP 316 Revision 2 requires wallets not to send to known-expired addresses and requires a nonzero, compatible `nExpiryHeight` for expiry-height metadata. These constraints must cover every recipient in a multi-recipient transaction.

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -59,31 +57,30 @@ impl fmt::Debug for Fvk {
...
         let data = data.to_vec();
-        match typecode.try_into()? {
-            Typecode::P2pkh => data.try_into().map(Fvk::P2pkh),
-            Typecode::P2sh => Err(data),
-            Typecode::Sapling => data.try_into().map(Fvk::Sapling),
-            Typecode::Orchard => data.try_into().map(Fvk::Orchard),
-            Typecode::Unknown(_) => Ok(Fvk::Unknown { typecode, data }),
+        match typecode {
+            DataTypecode::P2pkh => data.try_into().map(Fvk::P2pkh),
+            DataTypecode::P2sh => Err(data),
```
nuttycom: [P1] Accept Revision 2 P2SH viewing-key items

Typecode `0x01` is unconditionally rejected for UFVKs, and `Ivk` likewise has no P2SH representation. Revision 2 explicitly defines P2SH FVK/UIVK items using BIP 388 wallet policies, so valid `uvf` and `uvi` encodings cannot be consumed. The conformance test currently conceals this by skipping every vector containing a P2SH item in `zcash_keys/src/keys.rs`.

components/zcash_address/src/kind/unified.rs
```diff
@@ -85,25 +210,29 @@ impl TryFrom<u32> for Typecode {
...
-            0x00 => Ok(Typecode::P2pkh),
-            0x01 => Ok(Typecode::P2sh),
-            0x02 => Ok(Typecode::Sapling),
-            0x03 => Ok(Typecode::Orchard),
-            0x04..=0x02000000 => Ok(Typecode::Unknown(typecode)),
+            0x00 => Ok(Typecode::Data(DataTypecode::P2pkh)),
+            0x01 => Ok(Typecode::Data(DataTypecode::P2sh)),
+            0x02 => Ok(Typecode::Data(DataTypecode::Sapling)),
+            0x03 => Ok(Typecode::Data(DataTypecode::Orchard)),
+            0x04..=0xBF => Ok(Typecode::Data(DataTypecode::Unknown(typecode))),
```
nuttycom, line 221: [P2] Reject typecodes prohibited by the Revision 2 registry

The parser treats `0x04..=0xBF`, `0xC0..=0xDF`, and most values from `0xFD` upward as valid unknown items. The Revision 2 registry marks these ranges unassigned or prohibited; only `0xFFFA..=0xFFFF` may be used experimentally. In particular, `parse_items` explicitly accepts non-MUST-understand metadata in both revisions even though the registry permits none of `0xC0..=0xDF` in R0, `zu`, or `tu`. This allows invalid containers to decode and be re-emitted.

zcash_keys/src/address.rs
```diff
@@ -180,62 +211,203 @@ impl UnifiedAddress {
...
+                        .as_ref()
+                        .map(|pa| Uitem::Data(unified::Receiver::Sapling(pa.to_bytes())))
+                        .into_iter()
+                }
+                #[cfg(not(feature = "sapling"))]
+                {
+                    core::iter::empty()
+                }
+            })
+            .chain(self.transparent.as_ref().map(|taddr| match taddr {
```
nuttycom: [P2] Make the default `UnifiedAddress` encoding match its shielded-only contract

`to_zcash_address` includes `self.transparent`, so `encode()` produces `tu`, contrary to its documentation that it strips the transparent receiver and produces `zu`. The separate explicit `encode_transparent_including` path therefore provides no privacy distinction and additionally refuses valid metadata-free `tu` addresses. Either omit the transparent receiver from the default path as documented or revise the API coherently; the current behavior defeats the intended explicit choice between Revision 2 address forms.

nuttycom (commented):
Review against ZIP 316 Revision 2.

Co-Authored-By: OpenAI Codex <noreply@openai.com>

components/zcash_address/src/kind/unified/address.rs
nuttycom: The strategy now takes a `BTreeSet<DataTypecode>` and generates a `Vec<Uitem<Receiver>>`; `DataTypecode` derives `Ord`, documented as coinciding with typecode-value order given that `Unknown` never holds a known code's value. The spurious `Metadata(_)` arm is gone, and the original doc caveat about protocol-invalid receivers is restored.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified/fvk.rs
nuttycom: Restored, updated to the current constructor syntax.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified/ivk.rs
nuttycom: Restored. The hand-rolled comparators here and in `Encoding::try_from_items` are now consolidated into a single `Uitem::encoding_order`, and the `items_as_parsed` doc comment is back.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified/ivk.rs
nuttycom: Restored to `0x77`; the byte change was gratuitous — the test passes with the published vector.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified.rs
nuttycom: I believe this is superseded by the current draft: zips@81803ee7 added "If a new Metadata Item type should not be allowed in `zu` addresses, it MUST be assigned in the MUST-understand Metadata range", which makes future SHOULD-understand (0xC0–0xDF) assignments `zu`-safe by construction. Accepting unknown SHOULD-understand metadata in `zu` — while rejecting un-understood MUST-understand items — therefore matches the current spec, so I've made no change here.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified.rs
nuttycom: Restructured as suggested: `parse_items` now classifies via `Typecode::try_from`, with metadata handling (the per-revision MUST-understand rules and length validation) in a `parse_metadata_item` helper reached from the `Typecode::Metadata` arm. The `unreachable!` is gone.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

zcash_keys/src/address.rs
nuttycom: Implemented. Wallet operations now take an explicit `clock: &impl Clock` (the `Clock` trait moved from `zcash_client_sqlite::util` to a new `zcash_client_backend::util`, with the former re-exporting it), and every payment recipient is checked during transaction construction: sending to an address that is expired by time (per the clock) fails, and an address expiry height requires the transaction to carry a nonzero expiry height no greater than it — deliberately not clamped, per the spec's anti-leak rationale — failing with the new `Error::RecipientAddressExpiry`. The PCZT path now applies its expiry override through the transaction builder, so both construction paths share a single enforcement point. New pool fixtures cover the height-conflict case, time-expiry on the second recipient of a multi-recipient request, and the success path.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified/fvk.rs
nuttycom: Implemented. `Fvk::P2sh` and `Ivk::P2sh` carry the policy payload, structurally validated on parse: container framing, US-ASCII template, one-to-one correspondence between `@N` placeholders and key entries, and the `/**` vs `/*` notation required by the container kind. Full BIP 388 grammar validation is left to consumers that interpret the policy, consistent with the crate validating other items by length only. `zcash_keys` stores the item as `P2shPolicy`, round-trips it, and derives the UIVK form when deriving a UIVK from a UFVK — including re-sorting the derived key entries lexicographically per the deterministic-order rule, which the reference vectors caught my first draft omitting. The conformance tests no longer skip P2SH or metadata vectors, and additionally assert the stored item bytes and the derived-UIVK item against the reference data.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified.rs
nuttycom: I've left the parser's accept-and-retain behaviour in place, because I believe the registry's prohibited/unassigned entries bind producers rather than consumers in the current draft: 77e236db originally said "and Consumers MUST reject addresses that contain such Items", but 81803ee7 deliberately replaced that clause with "Producers MUST NOT … generate UA/UVKs containing Items with typecodes that are Unassigned". Three surviving pieces of spec text depend on consumers not rejecting: the ignore rule ("With the exception of MUST-understand Metadata Items, Consumers MUST ignore constituent Items with Typecodes they do not recognise"), the item-ordering rationale (round-trip retention of unrecognised items), and the note that a MUST-understand entry can later change to "only allowed in `tu` when understood" — which only works if deployed consumers do not hard-reject the range. A consumer that rejected today's unassigned ranges would break the moment a 2xxx ZIP assigns one of them. If the intent is producer-side registry enforcement, the open question is that `try_from_items` currently serves both minting and round-trip reconstruction; splitting those roles seems like a follow-up design decision.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

zcash_keys/src/address.rs
nuttycom: Fixed: `to_zcash_address`/`encode` now strip the transparent receiver and produce `zu`, falling back to `tu` only for an address with no shielded receiver (its only valid Revision 2 form), and `to_transparent_including_zcash_address` drops the metadata gate, returning `None` iff there is no transparent receiver. Since wallet storage needs a lossless encoding — persisted addresses must round-trip their transparent receiver — this also adds `to_receiver_preserving_zcash_address`/`encode_receiver_preserving` on `UnifiedAddress` and `Address`, which `zcash_client_sqlite` now uses for address persistence and lookup, keeping the privacy-preserving `zu` default for share/display paths.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -59,37 +69,43 @@ impl fmt::Debug for Fvk {
     }
 }
 
-impl TryFrom<(u32, &[u8])> for Fvk {
-    type Error = ParseError;
-
-    fn try_from((typecode, data): (u32, &[u8])) -> Result<Self, Self::Error> {
+impl SealedItem for Fvk {
+    fn parse(typecode: DataTypecode, data: &[u8]) -> Result<Self, ParseError> {
+        if typecode == DataTypecode::P2sh {
```
schell, line 74: **GLM-5.3 says:**
> [P1] R0 UFVKs must treat P2SH items as unrecognised, but this parses them for all revisions. The spec (P2SH Viewing Key Items) says: "For Revision 0: Typecode 0x01 MUST NOT be included in a UFVK or UIVK by Producers, and MUST be treated as unrecognised by Consumers." `SealedItem::parse` receives no `Revision`, so an R0 UFVK containing 0x01 is surfaced as `Fvk::P2sh`, inviting BIP 388 interpretation, instead of `Fvk::Unknown { typecode: 1, .. }` as the general "Consumers MUST ignore unrecognised typecodes" rule requires. Same issue in the `Ivk::parse` impl (ivk.rs). Fix: thread `Revision` into `SealedItem::parse` and, for R0, fall through to the `Unknown` arm. (P2SH receivers in R0 addresses are unaffected; that path is fine.)

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1679,6 +1685,46 @@ struct BuildState<P, AccountId> {
...
+/// payment must not be made to an address that is known to have expired, and when the
+/// recipient defines an expiry height, the transaction must carry a nonzero expiry
+/// height no greater than it. All recipients of a multi-recipient transaction are
+/// subject to these constraints.
+fn check_recipient_expiry(
+    ua: &UnifiedAddress,
+    tx_expiry_height: Option<BlockHeight>,
+    now: SystemTime,
+) -> Result<(), AddressExpiryError> {
+    if let Some(address_expiry) = ua.expiry_height() {
```
schell, line 1703: **GLM-5.3 says:**
> [P1] This check enforces the nExpiryHeight-conflict rule and expiry-time, but never compares the address expiry height to the current chain height. The spec says: "A Unified Address containing metadata Typecode 0xE0 MUST be considered expired when the height of the Zcash chain is greater than this value" and "A wallet MUST NOT send to an address that it knows to have expired." A caller-supplied small `expiry_height` (or a low `min_target_height + DEFAULT_TX_EXPIRY_DELTA`) can pass both branches here for an address whose expiry height is already in the past. `min_target_height` is available at the call site; pass it in and reject when `min_target_height > address_expiry`.
>
> Related [P2]: `AddressExpiryError::Expired` only carries `expiry_time: u64`, so once the height check exists there is no clean place to report the expired height. The error enum will need a height-aware variant (it is `#[non_exhaustive]`, so that is compatible).
>
> Also worth noting: the new `create_to_address_respects_recipient_expiry` fixture masks the missing height check, because `with_expiry(Some(birthday), None)` produces a height low enough that the default transaction expiry trips the conflict branch before any chain-height comparison could run.

components/zcash_address/src/kind/unified.rs
```diff
@@ -326,38 +711,252 @@ pub(crate) mod private {
...
+                    if has_transparent {
+                        return Err(ParseError::TransparentReceiverInR2Address);
+                    }
+                    // zu: must have at least one shielded receiver
+                    let has_shielded = result.items_as_parsed().iter().any(|item| {
+                        matches!(item, Uitem::Data(d) if matches!(d.typecode(),
+                            super::DataTypecode::Sapling
+                            | super::DataTypecode::Orchard))
+                    });
+                    if !has_shielded {
```
schell: **GLM-5.3 says:**
> [P2] This `zu` restriction over-rejects experimental-only addresses. The typecode registry reserves 0xFFFA..=0xFFFF for experiments, permitted "at the discretion of wallets that have opted into the corresponding experiment", and the zu restriction only forbids transparent items (0x00, 0x01). An opt-in zu address carrying only an experimental receiver violates no consumer rule, but is rejected here with a misleading `OnlyTransparent`. Consider gating on "at least one non-transparent data item" to match the literal zu constraint, and surfacing a dedicated error for the no-shielded-receiver case.

components/zcash_address/src/kind/unified.rs
```diff
@@ -326,38 +711,252 @@ pub(crate) mod private {
...
 
-            if only_transparent {
-                Err(ParseError::OnlyTransparent)
-            } else {
-                // All checks pass!
-                Ok(Self::from_inner(items))
+            match revision {
+                Revision::R0 => {
+                    // R0: Must contain at least one shielded item (for both UAs and UVKs).
+                    if !has_data_item || only_transparent {
```
schell: **GLM-5.3 says:**
> [P3] An R0 container consisting solely of metadata items is correctly rejected, but reports `OnlyTransparent`, which is semantically wrong (there are no transparent items at all). R2 uses the more accurate `NoDataItems`; consider unifying on that for the `!has_data_item` sub-case in both revisions.

components/zcash_address/src/kind/unified.rs
```diff
@@ -326,38 +711,252 @@ pub(crate) mod private {
...
+                return Err(P2shItemError::PlaceholderCount {
+                    placeholders: placeholders + 1,
+                    keys: n_keys,
+                });
+            }
+            seen[index] = true;
+            placeholders += 1;
+
+            let suffix = &template[digits_end..];
+            let multipath_ok = match kind {
```
schell, line 899: **GLM-5.3 says:**
> [P3] The multipath check is prefix-only, so a placeholder like `@0/**bar` (UFVK) or `@0/*foo` (UIVK) passes structural validation even though those are not valid BIP 388 multipath tokens. Deferring full grammar validation to consumers is fine and documented, but it would help to say explicitly in the `validate_p2sh_item` doc comment that only the required prefix is checked, not that the suffix is a well-formed BIP 388 path component.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -9824,3 +9834,138 @@ pub fn self_migration_keeps_spending_orchard<Dsf: DataStoreFactory>(
         "the second crossing must still be canonical"
     );
 }
+
+/// Tests that payment construction enforces the ZIP 316 Revision 2 address expiration
+/// rules: a payment to an address whose expiry height conflicts with the transaction's
+/// expiry height is refused, a payment to an address whose expiry time has passed is
+/// refused (for any recipient of a multi-recipient transaction), and a payment to an
+/// address with unexpired metadata succeeds.
+pub fn create_to_address_respects_recipient_expiry<T: ShieldedPoolTester>(
```
schell, line 9841: **GLM-5.3 says:**
> [P3] The expiry fixtures cover height-conflict, time-expiry on the second recipient, and the success path, but not the boundary cases: `expiry_time == now` (spec: expired only when the current time is *after* the expiry time) and `nExpiryHeight == address_expiry` (spec: forbidden only when it *would be greater than*). The success fixture also carries both 0xE0 and 0xE1, so a time-only (no 0xE0) address is untested in the success path. Adding equality-boundary and metadata-only variants would lock in the off-by-one semantics.

zcash_keys/src/keys.rs
```diff
@@ -2126,125 +2521,82 @@ mod tests {
...
-            sapling,
-            #[cfg(feature = "orchard")]
-            orchard,
-        );
+        use super::test_vectors::unified_viewing_keys_r2::TEST_VECTORS;
+        use zcash_protocol::consensus::NetworkType;
+
+        for tv in TEST_VECTORS {
+            let decoded =
+                UnifiedIncomingViewingKey::parse(&unified::Uivk::decode(tv.unified_ivk).unwrap().2)
```
schell, line 2659: **GLM-5.3 says:**
> [P3] The UFVK round-trip test asserts `expiry_height()`/`expiry_time()` against the vectors, but this UIVK round-trip loop does not, even though the spec requires expiry metadata to be "retained unmodified in the derived UIVK". Also, the vectors file carries `derived_ua` and per-receiver fields (`p2pkh_addr`, `p2sh_addr`, `sapling_raw_addr`, `orchard_raw_addr`, `diversifier_index`) that are never asserted anywhere, so UA derivation from a decoded UIVK (including the P2SH receiver) has no conformance coverage against the R2 vectors.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1391,7 +1391,7 @@ pub(crate) fn find_account_for_address<P: consensus::Parameters>(
     params: &P,
     address: &Address,
 ) -> Result<Option<AccountUuid>, FindAccountForAddressError<SqliteClientError>> {
-    let addr_str = address.encode(params);
+    let addr_str = address.encode_receiver_preserving(params);
```
schell, line 1394: **GLM-5.3 says:**
> [P3] Reads and writes here switched to `encode_receiver_preserving`, but existing databases store pre-migration R0 `u...` strings in the `addresses` table and no migration re-encodes them. Decoding still accepts R0 so lookups should work, but the table will hold a mix of R0 and R2 encodings after upgrade. Worth a regression test that `find_account_for_address` matches a pre-migration stored address, to lock in that both encodings of the same UA hit the same row.

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -227,9 +249,67 @@ mod tests {
...
         ) -> Uivk {
-            let mut items: Vec<_> = transparent.into_iter().chain(shielded).collect();
-            items.sort_unstable_by(Ivk::encoding_order);
-            Uivk(items)
+            let mut items: Vec<Uitem<Ivk>> = transparent
+                .into_iter()
+                .chain(shielded)
+                .map(Uitem::Data)
+                .collect();
+            items.sort_unstable_by(Uitem::encoding_order);
```
schell, line 257: nit: I don't think there's a functional reason to use `sort_unstable_by` here?

schell (dismissed):
I'm still looking at this, but flushing comments. Mostly just nits (agent caught a few bigger ones).

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/c6ad8c53b603aa08bce9d15d8fc7405b2822bd31..f4f1125044ba1aaeba7205d0a20f206fa9b5f71d) to address comments from review.

components/zcash_address/src/kind/unified/ivk.rs
nuttycom: It's in the `arb_unified_ivk` proptest strategy, and `Uitem::encoding_order` is a total order on distinct typecodes (duplicates are rejected downstream), so stable versus unstable isn't observable here. Happy to switch to `sort_by` if you'd rather not rely on that argument.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/987b2d9efe2ae4c02618545133a815ea64ab99f5..e2709badcf6f5a7d3f8704ab28350464e13ae229) to rebase on `main`.

zcash_transparent/src/keys.rs
```diff
@@ -521,6 +521,21 @@ pub trait IncomingViewingKey: private::SealedChangeLevelKey + core::marker::Size
...
+    /// This is the key material behind [`derive_address`](Self::derive_address), which a
+    /// caller needs when the address is determined by a script over several keys rather
+    /// than by a single key hash.
+    fn derive_pubkey(
+        &self,
+        address_index: NonHardenedChildIndex,
+    ) -> Result<PublicKey, bip32::Error> {
+        Ok(*self
+            .extended_pubkey()
+            .derive_child(address_index.into())?
```
nullcopy, line 536: This derivation is not at a standard address path. This derives the pubkey at the BIP 44 change level, but I'm pretty sure you want to derive the pubkey at the address level.

zcash_keys/src/keys.rs
```diff
@@ -845,18 +914,218 @@ impl AsRef<[u8; 32]> for OutgoingViewingKey {
 pub struct UnifiedFullViewingKey {
     #[cfg(feature = "transparent-inputs")]
     transparent: Option<::transparent::keys::AccountPubKey>,
+    p2sh: Option<P2shPolicy>,
```
nullcopy: This name is confusing, because of the `transparent` member which precedes it.

What if the members were `transparent_p2pkh` and `transparent_p2sh`?

nullcopy: Also, doesn't the `p2sh` member need to be gated by `transparent-inputs`?

zcash_keys/src/keys.rs
```diff
@@ -734,22 +764,56 @@ impl ReceiverRequirements {
     /// Constructs a new unified address request that requires an Orchard receiver, and no others.
     pub const ORCHARD: ReceiverRequirements = { Self::unsafe_new(Require, Omit, Omit) };
 
+    /// Constructs a new unified address request that requires a transparent P2PKH receiver, and
```
nullcopy, line 763: I thought ZIP 316 R2 allowed a P2SH-only address, but this says P2PKH is required

components/zcash_address/src/kind/unified.rs
```diff
@@ -114,14 +247,143 @@ impl TryFrom<Typecode> for usize {
...
+/// The kinds of unified viewing key container in which a P2SH viewing key item can
+/// occur, determining the multipath notation its descriptor template must use.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum P2shItemKind {
+    /// An item in a Unified Full Viewing Key, whose descriptor template must use the
+    /// `/**` multipath notation for each key placeholder.
+    FullViewing,
+    /// An item in a Unified Incoming Viewing Key, whose descriptor template must use the
+    /// `/*` notation for each key placeholder.
+    IncomingViewing,
```
nullcopy: No corresponding `OutgoingViewing` for P2SH?

components/zcash_address/src/kind/unified.rs
```diff
@@ -114,14 +247,143 @@ impl TryFrom<Typecode> for usize {
...
+                // items with equal typecodes are of the same kind.
+                _ => cmp::Ordering::Equal,
+            })
+    }
+}
+
+/// The kinds of unified viewing key container in which a P2SH viewing key item can
+/// occur, determining the multipath notation its descriptor template must use.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum P2shItemKind {
```
nullcopy, line 335: I'm surprised not to find a analogous `P2pkhItemKind` type. I'm just flagging the asymmetry, but I haven't thought hard about why that may be the case

components/zcash_protocol/src/address.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// See [ZIP 316](https://zips.z.cash/zip-0316) for details.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Revision {
+    /// Revision 0 of the unified encoding format.
+    ///
+    /// This is the original format, which requires at least one shielded receiver
+    /// in both Unified Addresses and Unified Viewing Keys.
+    R0,
+    /// Revision 2 of the unified encoding format.
```
nullcopy, line 20: Might be nice to add a note why Rev1 is skipped

nullcopy (changes requested):
Reviewed as of e2709badcf6f5a7d3f8704ab28350464e13ae229

zcash_transparent/src/keys.rs
nuttycom: You'll note that this trait inherits from `private::SealedChangeLevelKey`, so this key *is* derived at the correct path. Compare to `derive_address` above.

nuttycom: I will note that `derive_address` could now usefully be updated to call `derive_pubkey` instead of reimplementing the same logic, for consistency.

components/zcash_address/src/kind/unified.rs
nuttycom: There's no equivalent descriptor template for P2pkh items.

zcash_keys/src/keys.rs
nuttycom: Construction of P2SH addresses doesn't go through the paths involving `ReceiverRequirements`. Consider it this way: the counterpart to a P2SH address is a FROST multisig address, not ordinary ZIP 32 derivation (like the Orchard and Sapling components use). The `ReceiverRequirements` type is exclusively for ZIP 32; we should maybe make that more clear, but I don't think that it's worth renaming the type.

zcash_keys/src/keys.rs
nuttycom: This is modeled incorrectly in any case; the P2shPolicy and the ZIP 32 `AccountPubKey` should be mutually exclusive anyway.

components/zcash_address/src/kind/unified.rs
```diff
@@ -21,62 +21,191 @@ pub(crate) mod ivk;
...
+/// crate's parsers, as an `Unknown` typecode never holds a known code's value.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum DataTypecode {
+    /// A transparent P2PKH address, FVK, or IVK encoding.
     P2pkh,
     /// A transparent P2SH address.
-    ///
-    /// This typecode cannot occur in a [`Ufvk`] or [`Uivk`].
     P2sh,
-    /// A Sapling raw address, FVK, or IVK encoding as specified in [ZIP 316](https://zips.z.cash/zip-0316).
```
nuttycom, line 39: What is the rationale for the removal of the specification link?

components/zcash_address/src/kind/unified.rs
```diff
@@ -21,62 +21,191 @@ pub(crate) mod ivk;
...
-pub enum Typecode {
-    /// A transparent P2PKH address, FVK, or IVK encoding as specified in [ZIP 316](https://zips.z.cash/zip-0316).
+/// The derived ordering coincides with ascending typecode value order (the canonical
+/// encoding order of the corresponding items) for all values constructible by this
+/// crate's parsers, as an `Unknown` typecode never holds a known code's value.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum DataTypecode {
+    /// A transparent P2PKH address, FVK, or IVK encoding.
     P2pkh,
     /// A transparent P2SH address.
```
nuttycom: Doesn't this need to also mention redeem scripts or descriptor templates, for the FVK/IVK types?

components/zcash_address/src/kind/unified.rs
```diff
@@ -205,40 +515,105 @@ pub(crate) mod private {
         }
     }
 
+    /// Write a metadata item's raw encoding.
+    fn write_metadata_raw_encoding<W: Write>(item: &MetadataItem, mut writer: W) {
+        let tc_val: u32 = item.typecode().into();
+        let data = item.data();
+        CompactSize::write(&mut writer, tc_val.try_into().unwrap()).unwrap();
+        CompactSize::write(&mut writer, data.len()).unwrap();
```
nuttycom: Write paths have io::Error; these should map to proper IO errors, not `unwrap`s, or `expect`s if the failure cases of these write methods are provably unreachable.

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -62,6 +64,14 @@ workspace.
...
 - `wallet::init::init_wallet_db` and `wallet::init::WalletMigrator::init_or_migrate`
   no longer fail on wallets containing accounts imported by UIVK.
+- Upgrading a wallet database whose `support_zcashd_wallet_import` migration
+  ran before 2025-09-16 no longer fails with `NOT NULL constraint failed:
+  accounts_new.zcashd_legacy_address_index`. In such a database every account
+  that existed at that time has a NULL `accounts.zcashd_legacy_address_index`;
+  the migration that rebuilds the `accounts` table now maps those to the
+  sentinel value that column has carried since, and gives the database the
+  `hd_account` uniqueness index over `(hd_seed_fingerprint, hd_account_index,
+  zcashd_legacy_address_index)`.
```
nuttycom: This chunk need not have been moved.

zcash_keys/src/keys.rs
```diff
@@ -844,19 +914,244 @@ impl AsRef<[u8; 32]> for OutgoingViewingKey {
...
+///
+/// The payload is structurally validated when the containing unified viewing key is
+/// parsed. A policy whose descriptor template this implementation can evaluate yields a
+/// P2SH receiver during address generation.
+///
+/// [ZIP 316]: https://zips.z.cash/zip-0316
+/// [BIP 388]: https://github.com/bitcoin/bips/blob/master/bip-0388.mediawiki
+#[cfg(feature = "transparent-inputs")]
+#[derive(Clone, PartialEq, Eq)]
+pub struct P2shPolicy(Vec<u8>);
```
nuttycom: Why is this preserved as a `Vec<u8>` instead of parsed out into a structure containing the descriptor template and key information vector as well-typed values? This doesn't appear to provide any mechanism for external construction, limiting its utility.

zcash_keys/src/keys.rs
```diff
@@ -844,19 +914,244 @@ impl AsRef<[u8; 32]> for OutgoingViewingKey {
...
+/// advanced to its non-hardened child at that index, the descriptor template is
+/// instantiated with the derived keys, and the receiver is the HASH160 of the resulting
+/// output script.
+///
+/// Only the `sh(sortedmulti(..))` descriptor template is supported; any other template
+/// yields [`AddressGenerationError::UnsupportedP2shPolicy`].
+///
+/// [ZIP 316]: https://zips.z.cash/zip-0316
+#[cfg(feature = "transparent-inputs")]
+fn derive_p2sh_receiver(
```
nuttycom: This is not implemented in a fashion that lends itself readily to future extension to support other descriptor templates.

nuttycom (commented):
The last comment, that P2shPolicy needs to wrap a structured value with safe construction primitives, is blocking.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/573465932961c9d96a95fe8e94171380edd824e7..15d18ef96d210cb5821e2808b53d3a8d699428d5) to address comments from self-review; self-utACK.

nullcopy (approved):
utACK 15d18ef96d210cb5821e2808b53d3a8d699428d5

PR #2204, opened 2026-03-10
zcash_client_sqlite: Fix more regtest breakage

### LINKED ISSUE #2202: zcash_client_sqlite: Fix more regtest breakage
https://github.com/zcash/librustzcash/blob/707959e0d4b5eb203f86a658378035bdde7202d4/zcash_client_sqlite/src/wallet.rs#L1976-L1978
https://github.com/zcash/librustzcash/blob/707959e0d4b5eb203f86a658378035bdde7202d4/zcash_client_sqlite/src/wallet.rs#L2001-L2003

str4d:
Closes #2202.
Closes COR-1076.

### DIFF
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -1,10 +1,6 @@
 
 # cargo-vet imports lock
 
-[[unpublished.zcash_client_sqlite]]
-version = "0.19.5"
-audited_as = "0.19.4"
-
 [[publisher.bumpalo]]
 version = "3.16.0"
 when = "2024-04-08"
@@ -452,11 +448,11 @@ user-login = "str4d"
 user-name = "Jack Grigg"
 
 [[publisher.zcash_client_sqlite]]
-version = "0.19.4"
-when = "2026-02-27"
-user-id = 169181
-user-login = "nuttycom"
-user-name = "Kris Nuttycombe"
+version = "0.19.5"
+when = "2026-03-10"
+user-id = 6289
+user-login = "str4d"
+user-name = "Jack Grigg"
 
 [[publisher.zcash_encoding]]
 version = "0.3.0"


zcash_client_sqlite/src/wallet/init/migrations/v_sapling_shard_unscanned_ranges.rs
```diff
@@ -64,11 +64,12 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
-                self.params
-                    .activation_height(NetworkUpgrade::Sapling)
-                    .unwrap()
-            ),
+            // Sapling might not be active in regtest mode.
+            self.params
+                .activation_height(NetworkUpgrade::Sapling)
+                .map(|h| u32::from(h).to_string())
+                .as_deref()
+                .unwrap_or("NULL"),
```
nuttycom, line 72: This is a bit weird, in that the substitution is occurring inside a `CREATE VIEW`, meaning that the null value ends up hardcoded within the view, and therefore in any case where `prev_shard.subtree_end_height IS NULL` the view will return no rows. I haven't been able to deeply analyze the implications of this, but it seems like it could create some very weird boundary conditions if Sapling is activated after the test begins. I expect there may be bugs with respect to scanning the initial shard that result.

str4d: In general this shouldn't be a problem, because most tests were updated in `zcashd` to assume a baseline of Sapling a few years ago (and currently `zebrad` forces everything up to Canopy to be activated in regtest at height 1, but that could change). I don't actually recall whether or not there are tests that are solely pre-Sapling, or that activate Sapling after starting.

In theory we could have just left the Sapling assertions alone, and only fixed the NU5 ones. But I prefer us having consistent regtest behaviour across the codebase (and also some of these Sapling assertions would absolutely have broken Zallet trying to sync from genesis rather than from Sapling activation - https://github.com/zcash/wallet/issues/266), so IMO better to just change them all and then work around any specific regtest issues it results in.

nuttycom: Maybe this pitfall should be documented in `wallet/db.rs` on the view documentation so that if there is weirdness in testing in the future, the source of that weirdness is not entirely opaque?

str4d: And the other factor: we do not support changing networks across databases, so this issue won't affect mainnet or testnet databases (created with params that have always had Sapling activation), and regtest databases being repurposed for mainnet and testnet is unsupported and likely breaks in a bunch of other ways.

nuttycom (approved):
utACK 1fd7966b33b96e3486309e303568f71f2d9746b4

Issue #1944, opened 2025-09-10
zcash_client_backend: Add `WalletRead::find_account_for_address` method


str4d:
This would enable more efficient versions of querying `WalletRead::get_account_ids` folowed by `WalletRead::list_addresses` and address comparisions:
https://github.com/zcash/wallet/blob/2eb2d4b027195f664b4a0fe5ea3877402e00e37a/zallet/src/components/json_rpc/payments.rs#L427-L451

We could also have that as the default method, allowing impls like `zcash_client_sqlite` to override it for efficiency.

In the case of UAs, we should probably also handle the case where we detect different parts of the UA being relevant to different accounts, and report this as an error (as it would violate the intended properties of `Account`s).

PR #2205, opened 2026-03-10
zcash_client: add address-to-account lookup with UA conflict detection

fndecroix:
• Add WalletRead::find_account_for_address to resolve the owning account for a wallet address, including Unified Address receiver-aware matching. 
• Introduce FindAccountForAddressError with Backend and UnifiedAddressConflict variants. 
• Add address_receiver_matches_ua and a default trait implementation in zcash_client_backend.
• Override in zcash_client_sqlite::WalletDb with a single-query implementation (instead of scanning all accounts/addresses). 
• Add backend and sqlite tests for transparent, Sapling, Orchard, unknown-address, and cross-account UA conflict cases. 
• Update zcash_client_backend and zcash_client_sqlite changelogs.

Co-authored-by: @LorenzoRD2003
Fixes: Issue [#1944](https://github.com/zcash/librustzcash/issues/1944)

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -679,6 +679,7 @@ impl AddressSource {
 }
 
 /// Information about an address in the wallet.
+#[derive(Clone)]
 pub struct AddressInfo {
     address: Address,
     source: AddressSource,
@@ -1603,6 +1604,59 @@ pub trait InputSource {
     }
 }
 
+/// Errors that may occur when resolving the account controlling an address.
+#[derive(Debug)]
+pub enum FindAccountForAddressError<E> {
+    /// Error returned by the underlying wallet backend.
+    Backend(E),
+
+    /// A Unified Address whose receivers map to different accounts.
+    UnifiedAddressConflict,
+}
+
+impl<E> From<E> for FindAccountForAddressError<E> {
+    fn from(err: E) -> Self {
+        Self::Backend(err)
+    }
+}
+
+/// Returns `true` if `address` contains any receiver that matches the given `ua`'s
+/// receivers.
+pub fn address_receiver_matches_ua(address: &Address, ua: &UnifiedAddress) -> bool {
+    match address {
+        Address::Transparent(t) => ua.transparent().map(|nt| nt == t).unwrap_or(false),
+        Address::Sapling(s) => ua.sapling().map(|ns| ns == s).unwrap_or(false),
+        Address::Unified(ua2) => {
+            let matches_transparent = ua
+                .transparent()
+                .zip(ua2.transparent())
+                .map(|(a, b)| a == b)
+                .unwrap_or(false);
+            let matches_sapling = ua
+                .sapling()
+                .zip(ua2.sapling())
+                .map(|(a, b)| a == b)
+                .unwrap_or(false);
+            let matches_orchard = {
+                #[cfg(feature = "orchard")]
+                {
+                    ua.orchard()
+                        .zip(ua2.orchard())
+                        .map(|(a, b)| a == b)
+                        .unwrap_or(false)
+                }
+                #[cfg(not(feature = "orchard"))]
+                {
+                    false
+                }
+            };
+            matches_transparent || matches_sapling || matches_orchard
+        }
+        // TEX addresses can never appear as a receiver inside a UnifiedAddress
+        _ => false,
+    }
+}
+
 /// Read-only operations required for light wallet functions.
 ///
 /// This trait defines the read-only portion of the storage interface atop which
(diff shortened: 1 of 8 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -3294,3 +3406,401 @@ pub trait WalletCommitmentTrees {
...
+        UnifiedAddress::from_receivers(
+            Some(orchard).flatten(),
+            Some(sapling).flatten(),
+            transparent,
+        )
+        .expect("test UA must be valid")
+        .into()
+    }
+
+    impl WalletRead for TestWalletDb {
```
nuttycom: Why is an entirely separate `WalletRead` impl being defined here? Can you improve https://github.com/zcash/librustzcash/blob/main/zcash_client_backend/src/data_api/testing.rs#L2628 instead?

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ workspace.
 ## [Unreleased]
 
 ### Added
+- `WalletDb` now overrides `WalletRead::find_account_for_address` with a single SQL query,
+  replacing the `O(accounts × addresses)` scan that the default trait implementation would
+  require.
```
nuttycom: This isn't really the kind of thing we usually add changelog entries for; the change here is that this crate has been migrated to use `zcash_client_backend 0.22`.

zcash_client_backend/src/data_api.rs
```diff
@@ -1571,6 +1572,59 @@ pub trait InputSource {
...
+
+impl<E> From<E> for FindAccountForAddressError<E> {
+    fn from(err: E) -> Self {
+        Self::Backend(err)
+    }
+}
+
+/// Returns `true` if `address` contains any receiver that matches the given `ua`'s
+/// receivers.
+pub fn address_receiver_matches_ua(address: &Address, ua: &UnifiedAddress) -> bool {
```
nuttycom: This method doesn't belong in `zcash_client_backend`, it belongs in `zcash_keys`, and there are already some methods in that crate that should be used instead, for example https://github.com/zcash/librustzcash/blob/2e7455de13c24f468a11bec2e3de48ed7b6a655d/zcash_keys/src/address.rs#L487-L497

nuttycom: Also consider https://github.com/zcash/librustzcash/blob/2e7455de13c24f468a11bec2e3de48ed7b6a655d/zcash_keys/src/address.rs#L292

zcash_client_backend/src/data_api.rs
fndecroix: Hi @nuttycom ! Great, we had missed that implementation. I just removed TestWalletDb and extended MockWalletDb instead. Let me know if that's ok!

zcash_client_backend/src/data_api.rs
fndecroix: Hi @nuttycom . Thanks for pointing those functions! I changed `address_receiver_matches_ua` to use them. 

I'm not sure about the network parameter I need to convert the Address into a ZcashAddress, for that I introduced a a prameter to the `find_account_for_address` method. Is that reasonable? Should every structure that implements the WalletRead trait have access to its network?

zcash_client_sqlite/CHANGELOG.md
fndecroix: Hi @nuttycom I didn't understand this. What should we add to this file?

nuttycom: The CHANGELOG isn't intended to document internal implementation details; it should document changes to the public API.

fndecroix: Done!

zcash_client_backend/src/data_api.rs
nuttycom: Ordinarily, implementations of the `WalletRead` trait will bind internally to a specific network; you can see this with the existing `WalletRead` implementations. The high-level idea is that we don't want the same data store being used to store a mismash of testnet, mainnet, and regtest data.

nuttycom:
@zodl-review

nuttycom:
Hmm, some spurious stuff in that automated review. I've strikethrough'ed the stuff that's wrong.

y4ssi:
@zodl-review retrigger (testing dual-model: Claude Opus 4.7 1M + DeepSeek-R1, with enriched context: full files + commits + prior comments)

nuttycom:
Closing as subsumed by #2312, which squashes this branch's commits into a single commit and then does further cleanup.

y4ssi:
@zodl-review retrigger (dual-model: Claude Opus 4.7 1M + Llama 4 Maverick, both via Bedrock, full context)

Issue #2137, opened 2026-02-05
zcash_client_backend: Refactor `put_blocks` and `store_decrypted_tx` to share logic


str4d:
Currently we have two similar-but-separate pathways for processing transaction data:
- Compact tx data as part of compact blocks.
  - Only handles shielded, triggers transparent requests
- Full tx data standalone as part of mempool scanning or tx enhancement.
  - Handles shielded and transparent

We have several additional flows we want to support:
- Full tx data as part of full blocks
- Transparent data within compact blocks
- A better sync engine that can efficiently batch-process block data (for Zallet)

Towards this, we should refactor the existing logic to tease out the functional components that we will need for the full block engine.

str4d:
The current low-level DAA usage (plus a few helper methods) by `store_decrypted_tx`:
```mermaid
flowchart TD
    store_decrypted_tx --> get_funding_accounts
    get_funding_accounts --> detect_wallet_transparent_outputs
    detect_wallet_transparent_outputs --> delete_retrieval_queue_entries
    detect_wallet_transparent_outputs --> determine_fee
    determine_fee --> put_tx_data
    put_tx_data --> set_transaction_status

    set_transaction_status --> select_receiving_address-sapling-out
    select_receiving_address-sapling-out --> put_sent_output-sapling-out
    put_sent_output-sapling-out --> mark_sapling_note_spent

    set_transaction_status --> put_received_sapling_note-int
    put_received_sapling_note-int --> put_sent_output-sapling-int
    put_sent_output-sapling-int --> mark_sapling_note_spent

    set_transaction_status --> put_received_sapling_note-in
    put_received_sapling_note-in --> select_receiving_address-sapling-in
    select_receiving_address-sapling-in --> put_sent_output-sapling-in
    put_sent_output-sapling-in --> mark_sapling_note_spent

    mark_sapling_note_spent --> select_receiving_address-orchard-out
    select_receiving_address-orchard-out --> put_sent_output-orchard-out
    put_sent_output-orchard-out --> mark_orchard_note_spent

    mark_sapling_note_spent --> put_received_orchard_note-int
    put_received_orchard_note-int --> put_sent_output-orchard-int
    put_sent_output-orchard-int --> mark_orchard_note_spent

    mark_sapling_note_spent --> put_received_orchard_note-in
    put_received_orchard_note-in --> select_receiving_address-orchard-in
    select_receiving_address-orchard-in --> put_sent_output-orchard-in
[...]

PR #2219, opened 2026-03-18
zcash_client_backend: Further regularize `ll` tx handling

str4d:


zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -866,45 +832,16 @@ where
...
+    put_outputs(
         wallet_db,
         Some(params),
         tx_ref,
         funding_account,
+        observed_height,
+        d_tx.mined_height(),
+        &wallet_transparent_outputs,
         d_tx.sapling_outputs(),
-        |_, _| Ok(None),
```
str4d: This part of the second commit needs consideration. The rest of the PR should preserve existing behaviour, but here I replace the `spent_in = None` behaviour of `store_decrypted_tx` with the `detect_note_spent_in` behaviour of `put_blocks`. I think this matches the intention of what @nuttycom and I discussed in February (https://github.com/zcash/librustzcash/issues/2137#issuecomment-3856989137) but I want someone else to think about the implications.

str4d: Never mind, a common `put_outputs()` function would have required annoying type shenanigans, so I removed it. We might still want to unify these at some point, but it's not forced by this PR.

PR #2223, opened 2026-03-19
fix: redact sensitive material in `Debug` implementations

fndecroix:
## Summary

Replace derived `Debug` implementations with redacted manual implementations for types containing sensitive key material.

This prevents accidental disclosure of secrets via logs, panics, or error formatting.

---

## Rationale

Several types in `librustzcash` contain sensitive material (e.g. spending keys, private keys, viewing keys) while deriving `Debug`.

In Rust, `Debug` prints all fields recursively, which makes these types unsafe to log. Downstream consumers may inadvertently expose secrets via:

- logging (`log`, `tracing`)
- panic messages (`unwrap`, `expect`)
- error formatting (`anyhow`, etc.)

Redacting at the type level avoids relying on downstream discipline.

---

## Downstream impact

A review of `zingolib` shows that:

- High-level wrappers such as `LightWallet` and `UnifiedKeyStore` derive `Debug`
- These types contain:
  - `Mnemonic` (seed phrase, printed verbatim in its `Debug`)
  - `UnifiedSpendingKey`
  - `UnifiedFullViewingKey`

Although no current callsite was found that directly logs these types, they are effectively **one `Debug` call away from exposing sensitive material**.

A minimal instrumentation confirms that formatting these structures with `{:?}` reveals seed phrases and key material.

---

## Changes

Manual `Debug` implementations are introduced for key types, and transparent / viewing key components.

All implementations replace sensitive internal fields with `"..."`.

---

## Notes

- Non-breaking change (affects only `Debug` output)
- No changes to serialization or protocol logic
- Includes unit tests verifying redacted output

---

## Precedents

A similar issue was noted in the Zcash ecosystem audit of `ed25519-zebra` [NCC Group, 2023](https://www.nccgroup.com/media/bw3nq0vm/_ncc_group_zcashfoundation_e005955_report_2023-06-27_v10.pdf), where secret material could be exposed via `Debug`.

### DIFF
--- a/zcash_keys/src/keys.rs
+++ b/zcash_keys/src/keys.rs
@@ -839,7 +839,7 @@ impl core::fmt::Debug for UnifiedFullViewingKey {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         let mut d = f.debug_struct("UnifiedFullViewingKey");
         #[cfg(feature = "transparent-inputs")]
-        d.field("transparent", &self.transparent);
+        d.field("transparent", &self.transparent.as_ref().map(|_| "..."));
         #[cfg(feature = "sapling")]
         d.field("sapling", &self.sapling.as_ref().map(|_| "..."));
         #[cfg(feature = "orchard")]
@@ -2244,6 +2244,8 @@ mod tests {
         .unwrap();
 
         let debug_str = format!("{:?}", ufvk);
+        #[cfg(feature = "transparent-inputs")]
+        assert!(debug_str.contains("transparent: Some(\"...\")"));
         #[cfg(feature = "sapling")]
         assert!(debug_str.contains("sapling: Some(\"...\")"));
         #[cfg(feature = "orchard")]
--- a/zcash_keys/CHANGELOG.md
+++ b/zcash_keys/CHANGELOG.md
@@ -32,9 +32,9 @@ workspace.
   secret key material.
 - `Debug` output for `zcash_keys::keys::{UnifiedSpendingKey,
   OutgoingViewingKey, UnifiedFullViewingKey, UnifiedIncomingViewingKey}` now
-  redacts sensitive key material. `UnifiedFullViewingKey` and
-  `UnifiedIncomingViewingKey` no longer delegate shielded components to
-  external `Debug` implementations or emit raw unknown-item bytes.
+  redacts sensitive key material. `UnifiedFullViewingKey` and `UnifiedIncomingViewingKey`
+  no longer delegates shielded components to external `Debug` implementations
+  or emits raw unknown-item bytes.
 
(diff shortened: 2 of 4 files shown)


zcash_keys/src/keys.rs
```diff
@@ -816,6 +835,27 @@ pub struct UnifiedFullViewingKey {
...
 
+impl core::fmt::Debug for UnifiedFullViewingKey {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let mut d = f.debug_struct("UnifiedFullViewingKey");
+        #[cfg(feature = "transparent-inputs")]
+        d.field("transparent", &self.transparent);
+        #[cfg(feature = "sapling")]
+        d.field("sapling", &self.sapling.as_ref().map(|_| "..."));
+        #[cfg(feature = "orchard")]
+        d.field("orchard", &self.orchard.as_ref().map(|_| "..."));
```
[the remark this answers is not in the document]
LorenzoRD2003: You're right. Even if the transparent component is account-level material rather than spend-authorizing secret data, it is still part of viewing/key-management state and should not be emitted from `Debug` by default. For consistency with the rest of `UnifiedFullViewingKey` and to avoid leaking account/viewing metadata through logs, I’ll redact that field as well.

This means I'll also include modifications in the commit to add the redacted `Debug` for `AccountPubKey`, which is not in this message but otherwise we would be inconsistent. Thank you.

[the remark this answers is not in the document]
LorenzoRD2003: Done!

components/zcash_address/src/kind/unified/fvk.rs
```diff
@@ -41,6 +44,21 @@ pub enum Fvk {
     },
 }
 
+impl fmt::Debug for Fvk {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
daira, line 48: This almost looks like it shouldn't have a `Debug` impl. I know that would be a breaking change, though. It's fine like this.

components/zcash_address/src/kind/unified/ivk.rs
```diff
@@ -46,6 +49,21 @@ pub enum Ivk {
     },
 }
 
+impl fmt::Debug for Ivk {
```
daira, line 52: Same comment as for `Fvk` — and similarly for the other classes changed by this PR.

components/zcash_address/src/kind/unified/fvk.rs
nuttycom: I think it's good to have these explicitly redacting `Debug` impls in place, because that prevents a non-redacting `Debug` from being added without thinking through the consequences.

PR #2226, opened 2026-03-23
Add `TransparentOutputFilter` to `propose_shielding` and UTXO selection

schell:
Adds a `TransparentOutputFilter` enum to `zcash_client_backend::data_api` with variants `All` (default) and `CoinbaseOnly`, allowing callers to restrict which transparent outputs are eligible for selection.

This enables wallet implementations like Zallet's `z_shieldcoinbase` RPC to correctly shield only coinbase UTXOs, matching the behavior of the `zcashd` wallet which uses `TransparentCoinbasePolicy::Require` for this purpose.

The filter parameter is threaded through:
- `InputSource::get_spendable_transparent_outputs`
- `ShieldingSelector::propose_shielding`
- The public `propose_shielding` function
- `GreedyInputSelector::propose_shielding`

The `shield_transparent_funds` convenience function always passes `TransparentOutputFilter::All` internally.

In `zcash_client_sqlite`, the `CoinbaseOnly` filter adds a SQL clause `IFNULL(t.tx_index, 1) == 0` to restrict results to coinbase transactions, using the same `tx_index` heuristic as `excluding_immature_coinbase_outputs`.

In `zcash_client_memory`, the filter checks `tx_index().unwrap_or(1) == 0`, conservatively treating unknown `tx_index` as non-coinbase.

### DIFF
--- a/zcash_primitives/src/block.rs
+++ b/zcash_primitives/src/block.rs
@@ -1,10 +1,15 @@
 //! Structs and methods for handling Zcash block headers.
 
+// MSRV lints do not recognize the `clippy::io_other_error` lint
+#![allow(unknown_lints)]
+// beta lints attempt to rewrite `core2::io::Error::new` calls to `std::io::Error::other`
+#![allow(clippy::io_other_error)]
+
 use alloc::string::ToString;
 use alloc::vec::Vec;
 use core::fmt;
 use core::ops::Deref;
-use core2::io::{self, Read, Write};
+use core2::io::{self, ErrorKind, Read, Write};
 use transparent::bundle::OutPoint;
 use zcash_script::{Opcode, opcode::PossiblyBad};
 
@@ -190,7 +195,10 @@ impl Block {
         // individually read the coinbase transaction before the rest.
         let n_tx: usize = CompactSize::read_t(&mut reader)?;
         if n_tx == 0 {
-            return Err(io::Error::other("block has no coinbase tx"));
+            return Err(core2::io::Error::new(
+                ErrorKind::Other,
+                "block has no coinbase tx",
+            ));
         }
 
         // Parse with an arbitrary consensus branch ID initially; we will fix it later.
@@ -201,11 +209,16 @@ impl Block {
             let coinbase_transparent = coinbase
                 .transparent_bundle()
                 .filter(|b| b.is_coinbase())
-                .ok_or_else(|| io::Error::other("first tx of block is not coinbase"))?;
+                .ok_or_else(|| {
+                    core2::io::Error::new(ErrorKind::Other, "first tx of block is not coinbase")
+                })?;
 
             // Verify some simple structural consensus rules on the coinbase transaction.
             if coinbase.sprout_bundle().is_some() {
-                return Err(io::Error::other("coinbase tx has sprout data"));
+                return Err(core2::io::Error::new(
+                    ErrorKind::Other,
+                    "coinbase tx has sprout data",
+                ));
             }
 
             let coinbase_txin = coinbase_transparent.vin.first().expect("checked above");
@@ -214,16 +227,25 @@ impl Block {
                 Some(Ok(PossiblyBad::Good(Opcode::PushValue(height)))) => height
                     .to_num()
                     .map_err(|_| {
-                        io::Error::other("coinbase input specifies an invalid block height")
+                        core2::io::Error::new(
+                            ErrorKind::Other,
+                            "coinbase input specifies an invalid block height",
+                        )
                     })
                     .and_then(|h| {
                         BlockHeight::try_from(h).map_err(|_| {
-                            io::Error::other("coinbase input specifies an invalid block height")
+                            core2::io::Error::new(
+                                ErrorKind::Other,
+                                "coinbase input specifies an invalid block height",
+                            )
                         })
                     }),
                 // TODO: Permit missing height in genesis blocks by matching their block
                 // hash against the provided params.
-                _ => Err(io::Error::other("coinbase scriptSig missing height")),
+                _ => Err(core2::io::Error::new(
+                    ErrorKind::Other,
+                    "coinbase scriptSig missing height",
+                )),
             }?
         };
 
@@ -238,7 +260,8 @@ impl Block {
             // All later tx versions directly commit to the consensus branch ID.
             _ => {
                 if coinbase.consensus_branch_id() != consensus_branch_id {
-                    return Err(io::Error::other(
+                    return Err(core2::io::Error::new(
+                        ErrorKind::Other,
                         "coinbase tx's claimed height doesn't match its consensus branch ID for given network parameters",
                     ));
                 }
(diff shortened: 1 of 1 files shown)


zcash_client_memory/src/input_source.rs
```diff
@@ -166,6 +167,15 @@ impl<P: consensus::Parameters> InputSource for MemoryWalletDb<P> {
...
                     .unwrap_or(false)
             })
+            .filter(|(_, _, tx)| match output_filter {
+                TransparentOutputFilter::All => true,
+                // Coinbase transactions have tx_index == 0. If tx_index is unknown
+                // (None), conservatively treat it as **non-coinbase**, matching the SQL
+                // implementation's IFNULL(tx_index, 1) == 0 behavior.
+                TransparentOutputFilter::CoinbaseOnly => {
+                    tx.map_or(false, |tx| tx.tx_index().unwrap_or(1) == 0)
+                }
```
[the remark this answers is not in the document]
schell: Resolved by updating the PR description

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1096,6 +1096,7 @@ where
     /// [`propose_shielding`]: crate::data_api::wallet::propose_shielding
     #[cfg(feature = "transparent-inputs")]
     #[allow(clippy::type_complexity)]
+    #[allow(clippy::too_many_arguments)]
```
nullcopy, line 1103: This isn't really specific to your PR, but I wonder if we should relax the clippy checks? I see overrides like this all over the place, and feel it would be better to just disable the handful of lints we're always overriding (namely `type_complexity` and `too_many_arguments`)

zcash_client_backend/src/data_api.rs
```diff
@@ -1460,6 +1460,22 @@ impl NoteFilter {
...
+#[cfg(feature = "transparent-inputs")]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum TransparentOutputFilter {
+    /// Select all spendable transparent outputs.
+    #[default]
+    All,
+    /// Select only coinbase transparent outputs.
+    ///
+    /// Coinbase transactions are identified by having `tx_index == 0` within
+    /// their containing block. Outputs for which the transaction index is
```
nullcopy, line 1484: Curious to learn more about the case where a tx index may be unknown. I didn't realize that was possible

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -2639,6 +2644,7 @@ where
         from_addrs,
         to_account,
         confirmations_policy,
+        TransparentOutputFilter::All,
```
nullcopy, line 2661: I think this is missing from the CHANGELOG

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1167,7 +1167,13 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
     address: &TransparentAddress,
     target_height: TargetHeight,
     confirmations_policy: ConfirmationsPolicy,
+    output_filter: TransparentOutputFilter,
 ) -> Result<Vec<WalletUtxo>, SqliteClientError> {
+    let coinbase_only = match output_filter {
```
nullcopy, line 1172: Rather than manually encoding here, this we should probably add encode/parse methods for `TransparentOutputFilter`, similar to what exists for other db metadata items, like [PoolType](https://github.com/zcash/librustzcash/blob/90c76221f0e8d32e754e6dbea88b90d1b725a04e/zcash_client_sqlite/src/wallet/encoding.rs#L34), [AccountSource](https://github.com/zcash/librustzcash/blob/90c76221f0e8d32e754e6dbea88b90d1b725a04e/zcash_client_sqlite/src/wallet/encoding.rs#L56), etc

zcash_client_backend/src/data_api/testing.rs
schell: I mostly agree.

zcash_client_backend/src/data_api/wallet.rs
schell: The change is mentioned here https://github.com/zcash/librustzcash/pull/2226/changes#diff-1cc140d33c5912f6e98cc8bd07ee744a81aa7a70ee4becabe24617a5c9608f49R52 for `propose_shielding` :)

zcash_client_backend/src/data_api.rs
schell: It happens in our tests as they don't usually make a distinction between coinbase or not. AFAICT it also happens (at least sometimes) when getting transactions from `lightwalletd`, see https://github.com/zcash/lightwallet-protocol/issues/17.

zcash_client_backend/src/data_api/testing.rs
schell: I mean - I totally agree. But I do like it when I feel like I can simply create a parameters type, or a builder to satisfy clippy. But in this codebase that would be a big change, and I'm not sure the juice is worth the squeeze.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1167,7 +1167,9 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
     address: &TransparentAddress,
     target_height: TargetHeight,
     confirmations_policy: ConfirmationsPolicy,
+    output_filter: TransparentOutputFilter,
 ) -> Result<Vec<WalletUtxo>, SqliteClientError> {
+    let coinbase_only = output_filter.into_u32();
```
nuttycom: Hmm, I don't really care for the `repr u32` being used in this way, to map an enum into a sqlite boolean. I was wondering what that `repr` was for; I don't think it's appropriate to use it in this way. Explicitly match on the output filter here.

nuttycom (changes requested):
Looks good except for the mingling of `repr u32` and sqlite boolean semantics.

nuttycom (dismissed):
utACK 7ad85a202b437d6e6de7259c35dcf2c294722e23

zcash_client_backend/src/data_api.rs
```diff
@@ -1463,35 +1463,17 @@ impl NoteFilter {
...
-
-#[cfg(feature = "transparent-inputs")]
-impl TransparentOutputFilter {
-    /// Convert to u32 encoding.
-    pub fn into_u32(self) -> u32 {
-        self as u32
-    }
-
-    /// Attempt to parse from `u32` encoding.
-    pub fn parse_from_u32(encoding: u32) -> Option<Self> {
```
nullcopy: Curious, why was this reverted?

nullcopy: Its not worth blocking, but it was originally added to address https://github.com/zcash/librustzcash/pull/2226#discussion_r3002061358

nuttycom: This isn't persistent data; this is effectively a little DSL for query construction, and it never needs a stable encoding. In fact, the one place it *is* encoded as an i32, it would probably better be encoded as a boolean (which `rusqlite` handles just fine).

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: No; this is basically just a query DSL that's intended to alter a boolean query parameter. It doesn't need a stable encoding.

nullcopy (dismissed):
utACK e6a6a33a5c67188da6151cba741358fc6e3d5e1b

PR #2236, opened 2026-03-25
feat(zcash_client_sqlite): enable #![deny(missing_docs)]

### LINKED ISSUE #491: Enable `#![deny(missing_docs)]` on `zcash_client_sqlite`
Once we've ensured that the full API is documented, we can enable this to ensure we don't merge undocumented APIs in future.

nuttycom:
## Summary
- Enable `#![deny(missing_docs)]` on `zcash_client_sqlite` to prevent undocumented public APIs from being merged
- Add doc comments for all public items that were missing documentation across 8 files

## Test plan
- [x] All 90 tests pass (1 ignored), all 3 doc-tests pass
- [x] `cargo check -p zcash_client_sqlite` succeeds with `#![deny(missing_docs)]`
- [ ] CI passes

Closes #491

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -41,6 +41,8 @@ workspace.
 ### Removed
 - `zcash_client_sqlite::GapLimits` use `zcash_keys::keys::transparent::GapLimits` instead.
 - `zcash_client_sqlite::UtxoId` contents are now private.
+- The inadvertently-exposed `zcash_client_sqlite::chain::migrations::blockmeta::init` 
+  module has been removed from the public API.
 
 ### Fixed
 - `get_transparent_balances` no longer fails for standalone transparent addresses
--- a/zcash_client_sqlite/src/chain/migrations/blockmeta.rs
+++ b/zcash_client_sqlite/src/chain/migrations/blockmeta.rs
@@ -7,20 +7,18 @@ pub fn all_migrations() -> Vec<Box<dyn RusqliteMigration<Error = rusqlite::Error
     vec![Box::new(init::Migration {})]
 }
 
-/// The initial migration that creates the `compactblocks_meta` table.
-pub mod init {
+mod init {
     use rusqlite::{self};
     use schemerz::{self, migration};
     use schemerz_rusqlite::RusqliteMigration;
     use uuid::Uuid;
 
-    /// The migration that creates the `compactblocks_meta` table.
-    pub struct Migration;
+    pub(super) struct Migration;
 
     /// The migration that added the `compactblocks_meta` table.
     ///
     /// 68525b40-36e5-46aa-a765-720f8389b99d
-    pub const MIGRATION_ID: Uuid = Uuid::from_fields(
+    pub(super) const MIGRATION_ID: Uuid = Uuid::from_fields(
         0x68525b40,
         0x36e5,
         0x46aa,


zcash_client_sqlite/src/chain/migrations.rs
```diff
@@ -1 +1,2 @@
+/// Migrations for the `compactblocks_meta` table used by [`FsBlockDb`](crate::FsBlockDb).
```
str4d: We do not document modules this way; in this case it is particularly misleading because it appears at the top of the file, so it *looks* like documentation for `zcash_client_sqlite::chain::migrations` but is actually for `zcash_client_sqlite::chain::migrations::blockmeta`.

Move this to be a `//!` comment in `zcash_client_sqlite/src/chain/migrations/blockmeta.rs`

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -94,6 +100,7 @@ impl error::Error for Error {
     }
 }
 
+/// An implementation of [`ShardStore`] backed by a SQLite database.
```
str4d:
```suggestion
/// An implementation of [`ShardStore`] backed by an SQLite database.
```

zcash_client_sqlite/src/chain.rs
```diff
@@ -19,6 +19,7 @@ use {
 };
 
 pub mod init;
+/// Database migrations for the chain cache database.
```
str4d: Ditto: move this to be a `//!` comment in `zcash_client_sqlite/src/chain/migrations.rs`.

zcash_client_sqlite/src/util.rs
```diff
@@ -31,6 +31,7 @@ impl<C: Clock> Clock for &C {
 }
 
 #[cfg(any(test, feature = "test-dependencies"))]
+/// Test utilities for clock simulation.
 pub mod testing {
```
str4d: We put doc comments before conditional compilation, so that the conditions are as close to the conditioned code as feasible (particularly if there is a lot of documentation).
```suggestion
/// Test utilities for clock simulation.
#[cfg(any(test, feature = "test-dependencies"))]
pub mod testing {
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -154,6 +154,8 @@ use {
     zcash_keys::keys::zcashd,
 };
 
+/// SQLite-backed implementation of [`ShardStore`](shardtree::store::ShardStore)
+/// for note commitment trees.
 pub mod commitment_tree;
```
str4d: Ditto: move this to a `//!` comment in `zcash_client_sqlite/src/wallet/commitment_tree.rs`

zcash_client_sqlite/src/error.rs
```diff
@@ -163,11 +166,16 @@ pub enum SqliteClientError {
     ///
     /// [`TransactionsInvolvingAddress`]: zcash_client_backend::data_api::TransactionsInvolvingAddress
     #[cfg(feature = "transparent-inputs")]
+    /// The notification endpoint returned a range of block heights that did not match the
+    /// expected range.
```
str4d: Umm, what? This comment was added to an enum variant that was already documented, and the AI did so in a way that splits the doc comment around a feature flag.
```suggestion
```

zcash_client_sqlite/src/chain.rs
schell: Yeah, I didn't even know you could document modules this way.

PR #2238, opened 2026-03-25
feat(zcash_client_sqlite): cache IVKs and detect collisions when adding accounts

### LINKED ISSUE #1490: Cache incoming viewing keys and detect IVK collisions when adding accounts

nuttycom:
## Summary
- Add database migration introducing `orchard_ivk_item_cache`, `sapling_ivk_item_cache`, and `p2pkh_ivk_item_cache` columns to the `accounts` table with UNIQUE constraints
- Change account collision detection from FVK-based to IVK-based — the previous check skipped IVK-only accounts entirely since it only ran when a UFVK was present
- Add `get_account_for_uivk` query function parallel to existing `get_account_for_ufvk`

## Test plan
- [x] All 91 unit tests and 3 doc-tests pass
- [x] Compiles cleanly with default features
- [ ] CI passes with `--all-features`

Closes #1490

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -437,28 +437,13 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     birthday: &AccountBirthday,
     #[cfg(feature = "transparent-inputs")] gap_limits: &GapLimits,
 ) -> Result<Account, SqliteClientError> {
-    // Check whether any IVK component collides with an existing account. If two accounts
-    // share an FVK they necessarily share an IVK, and there is negligible chance that two
-    // different FVKs produce colliding IVKs.
+    // Check whether any IVK component collides with an existing account.
     let uivk = viewing_key.uivk();
     if let Some(existing_account) = get_account_for_uivk(conn, params, &uivk)? {
-        let existing_uivk = existing_account.uivk();
         match (&viewing_key, existing_account.ufvk()) {
-            (ViewingKey::Full(new_ufvk), existing_ufvk) => {
-                // FVK import over an existing account (either IVK-only or FVK).
-                // This is permitted as a capability upgrade if every existing IVK item
-                // has a corresponding item in the new FVK, AND the new FVK is strictly
-                // additive (has at least one new item). If the item sets are identical,
-                // it's a duplicate import and is an error.
-                if !new_ufvk.subsumes_uivk(&existing_uivk) {
-                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
-                }
-                // Check that the new FVK actually adds capability. If the existing
-                // account already has a UFVK that subsumes the new one as well,
-                // the item sets are identical and this is a duplicate import.
-                if existing_ufvk.is_some_and(|efvk| efvk.subsumes_ufvk(new_ufvk)) {
-                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
-                }
+            (ViewingKey::Full(new_ufvk), _) => {
+                // FVK import over an existing account. The upgrade function
+                // validates that the new FVK strictly adds capability.
                 return upgrade_account_ufvk(conn, params, &existing_account, new_ufvk);
             }
             (ViewingKey::Incoming(_), Some(_)) => {
@@ -467,15 +452,8 @@ pub(crate) fn add_account<P: consensus::Parameters>(
                 return Err(SqliteClientError::AccountCollision(existing_account.id()));
             }
             (ViewingKey::Incoming(_), None) => {
-                // IVK-over-IVK: permitted if the new UIVK strictly adds items.
-                if !uivk.subsumes(&existing_uivk) {
-                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
-                }
-                // If the existing UIVK also subsumes the new one, the item sets
-                // are identical — this is a duplicate import.
-                if existing_uivk.subsumes(&uivk) {
-                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
-                }
+                // IVK-over-IVK: the upgrade function validates that the new
+                // UIVK strictly adds capability.
                 return upgrade_account_uivk(conn, params, &existing_account, &uivk);
             }
         }
@@ -1593,14 +1571,30 @@ pub(crate) fn get_account_for_uivk<P: consensus::Parameters>(
 /// Upgrades an existing account to store a full viewing key, updating the UIVK
 /// and IVK cache columns to reflect any newly-added items.
 ///
-/// The caller must have already verified that the existing account's IVK items
-/// are a subset of those derivable from `ufvk`.
+/// Returns [`SqliteClientError::AccountCollision`] if the new UFVK does not
+/// strictly add capability over the existing account's key material.
 fn upgrade_account_ufvk<P: consensus::Parameters>(
     conn: &rusqlite::Connection,
     params: &P,
     existing_account: &Account,
     ufvk: &UnifiedFullViewingKey,
 ) -> Result<Account, SqliteClientError> {
+    let existing_uivk = existing_account.uivk();
+
+    // The new FVK must subsume the existing account's IVK items.
+    if !ufvk.subsumes_uivk(&existing_uivk) {
+        return Err(SqliteClientError::AccountCollision(existing_account.id()));
+    }
+
+    // If the existing account already has a UFVK that subsumes the new one,
+    // this is a duplicate import (no new capability).
+    if existing_account
+        .ufvk()
+        .is_some_and(|efvk| efvk.subsumes_ufvk(ufvk))
+    {
+        return Err(SqliteClientError::AccountCollision(existing_account.id()));
+    }
+
     let account_id = existing_account.internal_id();
     let ufvk_encoded = ufvk.encode(params);
     let uivk = ufvk.to_unified_incoming_viewing_key();
@@ -1641,14 +1635,22 @@ fn upgrade_account_ufvk<P: consensus::Parameters>(
 /// Upgrades an existing IVK-only account with a UIVK that adds new items,
 /// updating the UIVK encoding and IVK cache columns.
 ///
-/// The caller must have already verified that the existing account's IVK items
-/// are a strict subset of those in `uivk`.
+/// Returns [`SqliteClientError::AccountCollision`] if the new UIVK does not
+/// strictly add capability over the existing UIVK.
 fn upgrade_account_uivk<P: consensus::Parameters>(
     conn: &rusqlite::Connection,
     params: &P,
     existing_account: &Account,
     uivk: &UnifiedIncomingViewingKey,
 ) -> Result<Account, SqliteClientError> {
+    let existing_uivk = existing_account.uivk();
+
+    // The new UIVK must strictly add items: it must subsume the existing,
+    // but not be identical to it.
+    if !uivk.subsumes(&existing_uivk) || *uivk == existing_uivk {
+        return Err(SqliteClientError::AccountCollision(existing_account.id()));
+    }
+
     let account_id = existing_account.internal_id();
     let uivk_encoded = uivk.encode(params);
 
--- a/zcash_keys/src/keys.rs
+++ b/zcash_keys/src/keys.rs
@@ -1293,6 +1293,14 @@ pub struct UnifiedIncomingViewingKey {
     unknown: Vec<(u32, Vec<u8>)>,
 }
 
+impl PartialEq for UnifiedIncomingViewingKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.subsumes(other) && other.subsumes(self)
+    }
+}
+
+impl Eq for UnifiedIncomingViewingKey {}
+
 impl core::fmt::Debug for UnifiedIncomingViewingKey {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         let mut d = f.debug_struct("UnifiedIncomingViewingKey");
--- a/zcash_keys/CHANGELOG.md
+++ b/zcash_keys/CHANGELOG.md
@@ -13,6 +13,7 @@ workspace.
 - `zcash_keys::keys::UnifiedFullViewingKey::subsumes_ufvk`
 - `zcash_keys::keys::UnifiedFullViewingKey::subsumes_uivk`
 - `zcash_keys::keys::UnifiedIncomingViewingKey::subsumes`
+- `impl {PartialEq, Eq} for zcash_keys::keys::UnifiedIncomingViewingKey`
 - `zcash_keys::keys::OutgoingViewingKey`
 - `zcash_keys::keys::UnifiedFullViewingKey::select_ovk`
 - `zcash_keys::keys::transparent::gap_limits` module (behind the


zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,136 @@
...
+        "Adds IVK item cache columns and UNIQUE indices for FVK/IVK collision detection."
+    }
+}
+
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        // Add the new IVK cache columns to the accounts table.
+        transaction.execute_batch(
```
nuttycom: Instead of simply adding columns, after populating these columns, we should remove the fvk_item_cache columns; there's no need for them as the IVK components provide a better match: not only will comparing on ivk items catch redundant IVK imports, but they'll catch FVK imports of previously  imported IVKs and vice versa. These cases should be handled gracefully: importing an FVK over an existing IVK is a capability upgrade (all existing IVK items *MUST* match an overwriting FVK item, or fail) and importing an IVK over an existing FVK should return an error indicating that the key corresponds to an already-managed account.

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,200 @@
...
+                uuid BLOB NOT NULL,
+                account_kind INTEGER NOT NULL DEFAULT 0,
+                key_source TEXT,
+                hd_seed_fingerprint BLOB,
+                hd_account_index INTEGER,
+                ufvk TEXT,
+                uivk TEXT NOT NULL,
+                orchard_ivk_item_cache BLOB,
+                sapling_ivk_item_cache BLOB,
+                p2pkh_ivk_item_cache BLOB,
```
nuttycom, line 119: Since we'll have p2sh IVK items after ZIP 316 Revision 2, we should go ahead and prepare for that now by adding a `p2pkh_ivk_item_cache` and a constraint that at least one of this and `p2pkh_ivk_item_cache` must be null - `NOT (p2pkh_item_cache IS NOT NULL AND p2sh_item_cache IS NOT NULL)`

zcash_client_sqlite/src/wallet.rs
```diff
@@ -403,13 +403,35 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+    if let Some(existing_account) = get_account_for_uivk(conn, params, &uivk)? {
+        match (&viewing_key, existing_account.ufvk()) {
+            (ViewingKey::Full(new_ufvk), None) => {
+                // FVK-over-IVK: the new key is a capability upgrade. Verify that all
+                // IVK items derived from the new UFVK match the existing account's IVK
+                // items, then upgrade the account to store the UFVK.
+                let existing_uivk = existing_account.uivk();
+                if uivk.encode(params) != existing_uivk.encode(params) {
+                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
+                }
```
nuttycom: This is not quite correct: as the FVK represents a capability upgrade, we need to check the inverse, which is that each item in the existing IVK corresponds to an item in the FVK. The FVK can have additional items (newly added capabilities). Also, FVK-over-FVK import should be permitted in the case that the new FVK represents a strictly additive change to the set of items.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -429,22 +451,16 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
-    let sapling_item = viewing_key
-        .ufvk()
-        .and_then(|ufvk| ufvk.sapling().map(|k| k.to_bytes()));
+    let sapling_ivk_item: Option<Vec<u8>> = uivk.sapling().as_ref().map(|k| k.to_bytes().to_vec());
 
     #[cfg(feature = "transparent-inputs")]
-    let transparent_item = viewing_key
-        .ufvk()
-        .and_then(|ufvk| ufvk.transparent().map(|k| k.serialize()));
+    let transparent_ivk_item: Option<Vec<u8>> = uivk.transparent().as_ref().map(|k| k.serialize());
```
nuttycom: Let's rename this to `p2pkh_ivk_item`; in the future we'll have one or the other of `p2pkh_ivk_item` and `p2sh_ivk_item`.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1489,49 +1511,93 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
...
+fn upgrade_account_ufvk<P: consensus::Parameters>(
+    conn: &rusqlite::Connection,
+    params: &P,
+    existing_account: &Account,
+    ufvk: &UnifiedFullViewingKey,
+) -> Result<Account, SqliteClientError> {
+    let account_id = existing_account.internal_id();
+    let ufvk_encoded = ufvk.encode(params);
+    conn.execute(
+        "UPDATE accounts SET ufvk = :ufvk WHERE id = :id",
```
nuttycom: This isn't sufficient; it may be necessary to cache newly-added IVK items associated with the key.

zcash_keys/src/keys.rs
```diff
@@ -1110,6 +1110,55 @@ impl UnifiedFullViewingKey {
...
 
+    /// Returns `true` if this UFVK subsumes the given UIVK: every IVK item present in
+    /// `other` has a matching item derivable from `self`. The UFVK may have additional
+    /// items representing capabilities not present in the UIVK.
+    pub fn subsumes_uivk(&self, other: &UnifiedIncomingViewingKey) -> bool {
+        let self_uivk = self.to_unified_incoming_viewing_key();
+
+        #[cfg(feature = "orchard")]
+        match (other.orchard(), self_uivk.orchard()) {
+            (Some(e), Some(n)) => {
```
nuttycom, line 1127: Orchard IncomingViewingKey elements may be compared using `==`; it's only Sapling IVKs that don't define `Eq`.

zcash_keys/src/keys.rs
```diff
@@ -1110,6 +1110,55 @@ impl UnifiedFullViewingKey {
...
+                }
+            }
+            (Some(_), None) => return false,
+            _ => {}
+        }
+
+        #[cfg(feature = "transparent-inputs")]
+        match (other.transparent(), self_uivk.transparent()) {
+            (Some(e), Some(n)) => {
+                if e.serialize() != n.serialize() {
```
nuttycom: We might as well derive `PartialEq` and `Eq` for `ExternalIvk` (make sure to note in the CHANGELOG)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -403,13 +403,41 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
+                }
+                return upgrade_account_ufvk(conn, params, &existing_account, new_ufvk);
+            }
+            (ViewingKey::Incoming(_), Some(_)) => {
+                // IVK-over-FVK: the existing account already has full viewing
+                // capability. Importing a lower-capability key is not permitted.
+                return Err(SqliteClientError::AccountCollision(existing_account.id()));
+            }
+            (ViewingKey::Incoming(_), None) => {
```
nuttycom, line 454: Ivk-over-IVK should also be permitted if the new IVK subsumes the old and is additive. Add subsumption checks in `zcash_keys` as needed.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1489,49 +1519,174 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
...
+    #[cfg(not(feature = "transparent-inputs"))]
+    let p2pkh_ivk_item: Option<Vec<u8>> = None;
+
+    conn.execute(
+        "UPDATE accounts
+         SET uivk = :uivk,
+             orchard_ivk_item_cache = :orchard_ivk,
+             sapling_ivk_item_cache = :sapling_ivk,
+             p2pkh_ivk_item_cache = :p2pkh_ivk
+         WHERE id = :id",
```
nuttycom: Add `AND ufvk IS NULL` as a guardrail here, and check that the number of rows affected is 1 to ensure that we can't end up with a UIVK that's inconsistent with respect to the UFVK.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -46,12 +46,14 @@ use crate::wallet::scanning::priority_code;
...
-///   if any.
-/// - `p2pkh_fvk_item_cache`: The serialized representation of the P2PKH item of the `ufvk`,
-///   if any.
+/// - `orchard_ivk_item_cache`: The serialized representation of the Orchard IVK item derived
+///   from the account's viewing key, if any. Used for collision detection.
+/// - `sapling_ivk_item_cache`: The serialized representation of the Sapling IVK item derived
+///   from the account's viewing key, if any. Used for collision detection.
+/// - `p2pkh_ivk_item_cache`: The serialized representation of the transparent P2PKH IVK item
+///   derived from the account's viewing key, if any. Used for collision detection.
+/// - `p2sh_ivk_item_cache`: Reserved for future ZIP 316 Revision 2 P2SH IVK items.
```
nullcopy: (nit) this comment wont age well, assuming ZIP 316 Revision 2 merges soon

zcash_client_sqlite/src/wallet/init/migrations.rs
nullcopy, line 1: Also need to add this migration to the `CURRENT_LEAF_MIGRATIONS` below, right?

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,211 @@
...
+        Err(WalletMigrationError::CannotRevert(MIGRATION_ID))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::wallet::init::migrations::tests::test_migrate;
+
+    #[test]
+    fn migrate() {
```
nullcopy, line 221: I would love to see a test demonstrating failure if the db already contained conflicts which weren't detected before

zcash_keys/src/keys.rs
```diff
@@ -1110,6 +1110,20 @@ impl UnifiedFullViewingKey {
...
     }
 
+    /// Returns `true` if this UFVK subsumes the given UIVK: every IVK item present in
+    /// `other` has a matching item derivable from `self`. The UFVK may have additional
+    /// items representing capabilities not present in the UIVK.
+    pub fn subsumes_uivk(&self, other: &UnifiedIncomingViewingKey) -> bool {
+        self.to_unified_incoming_viewing_key().subsumes(other)
+    }
+
+    /// Returns `true` if this UFVK subsumes the given UFVK: every IVK item derivable
```
nullcopy: Is it true to say one UFVK subsumes the other UFVK, just because its IVK subsumes the other's IVK?

zcash_keys/src/keys.rs
```diff
@@ -1457,6 +1471,46 @@ impl UnifiedIncomingViewingKey {
...
+                }
+            }
+            (Some(_), None) => return false,
+            _ => {}
+        }
+
+        #[cfg(feature = "sapling")]
+        match (other.sapling(), &self.sapling) {
+            (Some(e), Some(n)) => {
+                if e.to_bytes() != n.to_bytes() {
```
nullcopy: Is `PartialEq` not implemented for sapling types? Would prefer to see that implemented and used here, unless there's a reason not to

zcash_keys/src/keys.rs
nuttycom: Hmm, so, the metadata items are a question.

zcash_client_sqlite/src/lib.rs
```diff
@@ -2708,14 +2708,16 @@ mod tests {
     ) where
         DbT::Account: core::fmt::Debug,
     {
+        // Re-importing the same UFVK is a duplicate (no new capability added), so
+        // it should produce an AccountCollision error.
```
nullcopy, line 2867: I know this preexists your PR, but would it be better for these tests to be in 'zcash_client_backend'? The requirement not to allow duplicate account insertions feels like a reasonable API constraint

nuttycom: For now we'll keep them here, because `zcash_client_backend` doesn't expose methods for uivk import; when we add that, we'll move the tests over.

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,299 @@
...
+                transparent_ivk_item,
+            ));
+        }
+        // Drop stmt/rows before modifying the table.
+        drop(rows);
+        drop(stmt);
+
+        // Rebuild the accounts table: drop FVK cache columns, add IVK cache columns.
+        // Use PRAGMA legacy_alter_table to preserve views and foreign key references.
+        transaction.execute_batch(
```
nullcopy, line 106: Curious what we would do in the case of a migration failure.

e.g. in the (probably unlikely) case that the database contains one account by FVK which subsumes the IVK of another account.

zcash_keys/src/keys.rs
nullcopy: @nuttycom I see you implemented `PartialEq` and `Eq`, but I think you forgot to remove the bytewise comparison here

zcash_client_sqlite/src/wallet.rs
```diff
@@ -403,13 +433,49 @@ pub(crate) fn add_account<P: consensus::Parameters>(
     birthday: &AccountBirthday,
     #[cfg(feature = "transparent-inputs")] gap_limits: &GapLimits,
 ) -> Result<Account, SqliteClientError> {
-    if let Some(ufvk) = viewing_key.ufvk() {
-        // Check whether any component of this UFVK collides with an existing imported or derived FVK.
-        if let Some(existing_account) = get_account_for_ufvk(conn, params, ufvk)? {
-            return Err(SqliteClientError::AccountCollision(existing_account.id()));
+    // Check whether any IVK component collides with an existing account. If two accounts
+    // share an FVK they necessarily share an IVK, and there is negligible chance that two
+    // different FVKs produce colliding IVKs.
```
nullcopy: > and there is negligible chance that two different FVKs produce colliding IVKs.

I can imagine some scenarios which might make this probability non-negligible, e.g. if we later add capabilities to FVKs which do not change the IVK, but which are not understood today. Can't this simply be handled by enforcing that the new FVK subsumes the existing?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -403,13 +433,49 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+                // it's a duplicate import and is an error.
+                if !new_ufvk.subsumes_uivk(&existing_uivk) {
+                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
+                }
+                // Check that the new FVK actually adds capability. If the existing
+                // account already has a UFVK that subsumes the new one as well,
+                // the item sets are identical and this is a duplicate import.
+                if existing_ufvk.is_some_and(|efvk| efvk.subsumes_ufvk(new_ufvk)) {
+                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
+                }
```
nullcopy: Per my previous comment, I think we should enforce that the new FVK subsumes the old, before allowing this upgrade

zcash_client_sqlite/src/wallet.rs
```diff
@@ -403,13 +433,49 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+                if !uivk.subsumes(&existing_uivk) {
+                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
+                }
+                // If the existing UIVK also subsumes the new one, the item sets
+                // are identical — this is a duplicate import.
+                if existing_uivk.subsumes(&uivk) {
+                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
+                }
+                return upgrade_account_uivk(conn, params, &existing_account, &uivk);
+            }
```
nullcopy:
```suggestion
                // IVK-over-IVK: permitted if the new UIVK strictly adds items.
                if !uivk.subsumes(&existing_uivk) || uivk == existing_uivk {
                    return Err(SqliteClientError::AccountCollision(existing_account.id()));
                }
                return upgrade_account_uivk(conn, params, &existing_account, &uivk);
            }
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1490,48 +1540,147 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
...
         Ok(accounts.into_iter().next())
     }
 }
 
+/// Upgrades an existing account to store a full viewing key, updating the UIVK
+/// and IVK cache columns to reflect any newly-added items.
+///
+/// The caller must have already verified that the existing account's IVK items
+/// are a subset of those derivable from `ufvk`.
+fn upgrade_account_ufvk<P: consensus::Parameters>(
```
nullcopy, line 1576: IMO this function should return an error if an illegal upgrade is attempted. e.g. if the new FVK does not add capabilities over the existing, etc.

You have these checks in `add_account()`, but this way those constraints will still be enforced if these upgrades get called elsewhere in the future. Also, it allows you to simplify the checks in `add_account()`

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1490,48 +1540,147 @@ pub(crate) fn get_account_for_ufvk<P: consensus::Parameters>(
...
+        Ok(parse_account_row(row, params))
+    })?
+}
+
+/// Upgrades an existing IVK-only account with a UIVK that adds new items,
+/// updating the UIVK encoding and IVK cache columns.
+///
+/// The caller must have already verified that the existing account's IVK items
+/// are a strict subset of those in `uivk`.
+fn upgrade_account_uivk<P: consensus::Parameters>(
```
nullcopy, line 1640: Same comment as above about returning error for illegal upgrades

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nuttycom: We *should* be secure against that given that we've never opened up IVK import before. But the answer regardless is that in the case of migration failure, the transaction is rolled back and the wallet fails to start.

zcash_keys/src/keys.rs
nuttycom: Oops, actually, I totally failed and missed implementing these for the incoming viewing key types: https://github.com/zcash/sapling-crypto/pull/180

zcash_client_sqlite/src/lib.rs
```diff
@@ -3034,6 +3035,195 @@ mod tests {
         );
     }
 
+    #[test]
+    pub(crate) fn ivk_only_account_upgrade_paths() {
```
nuttycom, line 3039: These tests should be moved to `zcash_client_backend` once we support IVK-only account creation in the API there.

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,299 @@
...
+            )?;
+        }
+
+        // Re-create indices (the old ones were dropped with the table).
+        transaction.execute_batch(
+            "CREATE UNIQUE INDEX accounts_uuid ON accounts (uuid);
+             CREATE UNIQUE INDEX accounts_ufvk ON accounts (ufvk);
+             CREATE UNIQUE INDEX accounts_uivk ON accounts (uivk);
+             CREATE UNIQUE INDEX hd_account ON accounts (hd_seed_fingerprint, hd_account_index, zcashd_legacy_address_index);
+             CREATE UNIQUE INDEX accounts_orchard_ivk ON accounts (orchard_ivk_item_cache);
```
daira, line 189: Because the UNIQUE indices for the ivk cache columns are created after populating them, any duplicates would cause the index creation to fail rather than being reported when a duplicate is added. Since it *should* be infeasible to find two fvks that derive the same ivk, this shouldn't make any difference. However, it could be argued that you would get better error reporting by doing it in the other order. Any difference in efficiency will be insignificant for the expected number of accounts.

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,299 @@
...
+            PRAGMA legacy_alter_table = OFF;",
+        )?;
+
+        // Populate the IVK cache columns.
+        for (account_id, orchard_ivk, sapling_ivk, p2pkh_ivk) in &ivk_updates {
+            transaction.execute(
+                "UPDATE accounts
+                 SET orchard_ivk_item_cache = :orchard_ivk,
+                     sapling_ivk_item_cache = :sapling_ivk,
+                     p2pkh_ivk_item_cache = :p2pkh_ivk
```
nullcopy, line 172: Need P2SH here too now, right?

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,299 @@
...
+                 WHERE id = :account_id",
+                named_params![
+                    ":orchard_ivk": orchard_ivk,
+                    ":sapling_ivk": sapling_ivk,
+                    ":p2pkh_ivk": p2pkh_ivk,
```
nullcopy, line 177: Missing P2SH

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,299 @@
...
+                    AND ufvk IS NOT NULL
+                  )
+                  OR
+                  (
+                    account_kind = 1
+                    AND (hd_seed_fingerprint IS NULL) = (hd_account_index IS NULL)
+                  )
+                ),
+                CHECK (
+                  NOT (p2pkh_ivk_item_cache IS NOT NULL AND p2sh_ivk_item_cache IS NOT NULL)
```
nullcopy, line 141: (Not blocking) I wonder if it would be better to have one `transparent_ivk` column, instead of separate p2pkh & p2sh columns

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
```diff
@@ -0,0 +1,299 @@
...
+            let transparent_ivk_item: Option<Vec<u8>> =
+                uivk.transparent().as_ref().map(|k| k.serialize());
+            #[cfg(not(feature = "transparent-inputs"))]
+            let transparent_ivk_item: Option<Vec<u8>> = None;
+
+            ivk_updates.push((
+                account_id,
+                orchard_ivk_item,
+                sapling_ivk_item,
+                transparent_ivk_item,
```
nullcopy, line 97: Here you have a unified transparent IVK type, but below you have distinct P2PKH and P2SH IVK types. I think some wires are crossed.

I like the idea of just having a single column, so perhaps the unified type here makes more sense, but I'm not picky

zcash_client_sqlite/src/wallet.rs
```diff
@@ -321,6 +321,36 @@ impl ViewingKey {
     }
 }
 
+/// Serialized IVK items extracted from a [`UnifiedIncomingViewingKey`] for storage
+/// in the `accounts` table cache columns.
+struct IvkItemCache {
+    orchard: Option<Vec<u8>>,
+    sapling: Option<Vec<u8>>,
+    p2pkh: Option<Vec<u8>>,
```
nullcopy, line 329: Need p2sh here now, too, right?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -501,10 +529,10 @@ pub(crate) fn add_account<P: consensus::Parameters>(
                 ":zcashd_legacy_address_index": zcashd_legacy_address_index,
                 ":key_source": key_source,
                 ":ufvk": ufvk_encoded,
-                ":uivk": viewing_key.uivk().encode(params),
-                ":orchard_fvk_item_cache": orchard_item,
-                ":sapling_fvk_item_cache": sapling_item,
-                ":p2pkh_fvk_item_cache": transparent_item,
+                ":uivk": uivk.encode(params),
+                ":orchard_ivk_item_cache": ivk_cache.orchard,
+                ":sapling_ivk_item_cache": ivk_cache.sapling,
```
nullcopy, line 534: Missing p2sh

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nuttycom: Not until we actually have P2SH IVK items, I think?

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nuttycom: I considered this; the problem is that they're fundamentally different encodings. I think that I prefer to have dedicated parsing for each encoding, instead of using a parse failure to indicate that the next decoder should be attempted.

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nuttycom: Here the transparent ivk item is actually the p2pkh IVK; we don't have representable P2SH IVKs yet.

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nuttycom: Same thing, we have nothing to write here.

zcash_client_sqlite/src/wallet.rs
nuttycom: Just in case R2 IVKs are rejected, we shouldn't make that change just yet, I think; that addition should be made in #2199

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nullcopy: Good point! The addition of the P2SH item column threw me off, as I assumed they would come together

zcash_client_sqlite/src/wallet/init/migrations/ivk_item_cache.rs
nullcopy: I think my confusion stemmed from how P2SH and P2PKH IVKs are accounted for separately elsewhere (update loop, db schema, etc) , but the naming here implies there will be just one combined transparent IVK type

PR #2246, opened 2026-03-27
zcash_client_backend: Add full block scanning logic

### LINKED ISSUE #2137: zcash_client_backend: Refactor `put_blocks` and `store_decrypted_tx` to share logic

str4d:


nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e533601658c1e8641de159b2ea5ef5f70b42091c..fc3c566a426c83cf5422ab35686c03fb154c556b) with cleanup, test fixing, and Claude-implemented completion. Noting this force push specifically as the previous pushes were all just rebase operations.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/e533601658c1e8641de159b2ea5ef5f70b42091c..be7b669) complete changes since original no-semantic-change rebase on `main`.

nuttycom (dismissed):
utACK be7b66907940caac49883dd651e4a56ac7c9773f; requires another review because I was responsible for the changes in https://github.com/zcash/librustzcash/compare/e533601658c1e8641de159b2ea5ef5f70b42091c..be7b669

nuttycom:
@zodl-review

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+                    .map(|(id, key)| (id.clone(), key.prepare())),
+            ),
+            #[cfg(not(feature = "orchard"))]
+            orchard: PhantomData,
+        }
+    }
+
+    /// Adds the given transaction's shielded outputs to the various batch runners.
+    ///
+    /// `block_tag` is the hash of the block that triggered this txid being added to the
```
dannywillems: Nit: `block_tag` is not a parameter.

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+    /// batch, or `None` to indicate that no block triggered it (i.e. it was a mempool
+    /// change).
+    ///
+    /// `height` is the height of the block that triggered this txid being added to the
+    /// batch, or the mempool height (chain tip height + 1) if `block_tag` is `None`.
```
dannywillems: `block_tag` does not exist.

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+        self.sapling.flush();
+        #[cfg(feature = "orchard")]
+        self.orchard.flush();
+    }
+}
+
+/// A pending batch decryption result for a single transaction.
+pub(crate) struct PendingBatch<IvkTag> {
+    tx: Transaction,
+    sapling_batch: Option<BatchReceiver<IvkTag, SaplingDomain, [u8; 512]>>,
```
dannywillems: Nit: `512` could be coming from a `const` definition.

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+        match self {
+            ScanBlockError::Scan(e) => Some(e),
+            ScanBlockError::AddressLookup(e) => Some(e),
+        }
+    }
+}
+
+/// Scans a block with a set of [`ScanningKeys`].
+///
+/// Returns a vector of [`WalletTx`]s decryptable by any of the given keys. If an output is
```
dannywillems: The documentation and the code for the returned type mismatch.

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+    for (tx_index, batch) in vtx.into_iter().enumerate() {
+        let BatchResult {
+            tx,
+            sapling: sapling_decrypted,
+            #[cfg(feature = "orchard")]
+                orchard: orchard_decrypted,
+        } = batch;
+        let txid = tx.txid();
+        let tx_index =
+            TxIndex::try_from(tx_index).expect("Cannot fit more than 2^16 transactions in a block");
```
dannywillems, line 424: `Cannot fit more than 2^16 transactions in a block`: is it a consensus rule?

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+
+        let (sapling_spends, sapling_unlinked_nullifiers) = tx
```
dannywillems, line 426: Nit: below, when the feature `orchard` is activated, the equivalent of `sapling_unlinked_nullifiers` for `orchard`, namely `orchard_unlinked_nullifiers`, is added in a nested scope. For consistency, the same pattern could be followed.

zcash_client_backend/src/sync/decryptor.rs
```diff
@@ -0,0 +1,418 @@
...
+
+use crate::scanning::{
+    ScanningKeys,
+    full::{BatchResult, BatchRunners, DEFAULT_BATCH_SIZE_THRESHOLD},
+};
+
+/// Constructs a new batch decryption engine.
+pub fn new() -> Builder {
+    Builder {
+        queue_size: 1000,
```
nullcopy: (nit) can we use a named constant here, instead of the magic number?

zcash_client_backend/src/sync/decryptor.rs
```diff
@@ -0,0 +1,418 @@
...
+
+    /// Sets the number of outputs at which the batch runner will immediately flush a
+    /// batch, for both the Sapling and Orchard runners.
+    ///
+    /// To configure the pools independently, use [`Self::sapling_batch_size_threshold`]
+    /// and [`Self::orchard_batch_size_threshold`].
+    ///
+    /// Default is 200.
+    pub fn batch_size_threshold(mut self, batch_size_threshold: usize) -> Self {
+        self.sapling_batch_size_threshold = batch_size_threshold;
```
nullcopy: (nit) I'd prefer to call the independent setters from within this combined setter

zcash_client_backend/src/sync/decryptor.rs
```diff
@@ -0,0 +1,418 @@
...
+    /// since the engine started (or since the last reload) are applied to subsequently
+    /// queued blocks and transactions.
+    ///
+    /// The reload is ordered within the queue: every block or transaction queued before
+    /// this call is decrypted with the previous keys, and everything queued afterwards
+    /// with the reloaded keys.
+    ///
+    /// Returns a receiver that completes once the reload has been applied, or `None` if
+    /// the batch decryptor has shut down.
+    pub async fn reload_keys(&self) -> Option<oneshot::Receiver<()>> {
```
nullcopy, line 270: What is the scenario where we reload keys without caring which block range get's scanned with the new keys? I'm wondering if key reload should be coupled with a scan restart

zcash_client_backend/src/wallet.rs
```diff
@@ -166,6 +169,12 @@ impl<AccountId> WalletTx<AccountId> {
         self.block_index
     }
 
+    /// Returns a record for each transparent coin received or produced by the wallet in
+    /// the transaction.
```
nullcopy:
```suggestion
    /// Returns a record for each transparent coin received or produced by the wallet
```

zcash_client_backend/src/scan.rs
```diff
@@ -534,16 +555,34 @@ where
...
+    ///
+    /// Returns a handle for receiving the results of the batch. To have the batch runner
+    /// manage this for you, use [`Self::add_outputs`] instead of this method.
+    ///
+    /// If after adding the given outputs, the accumulated batch size is at least the size
+    /// threshold that was set via `Self::new`, `Self::flush` is called. Subsequent calls
+    /// to `Self::add_outputs` will be accumulated into a new batch.
+    pub(crate) fn process_outputs(
+        &mut self,
+        domain: impl Fn(&Output) -> D,
```
nullcopy, line 578: Could use a comment explaining the expected behavior of this fn

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+    orchard: PhantomData<TO>,
+}
+
+impl<IvkTag, TS, TO> BatchRunners<IvkTag, TS, TO>
+where
+    IvkTag: Clone + Send + 'static,
+    TS: SaplingTasks<IvkTag>,
+    TO: OrchardTasks<IvkTag>,
+{
+    pub(crate) fn for_keys<AccountId>(
```
nullcopy, line 104: Needs documentation

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+            warn!(
+                "More than one wallet account detected as funding transaction {:?}, selecting {:?}",
+                txid,
+                funding_account
+                    .expect("funding_account is Some when spent_from_accounts is nonempty")
+            )
+        }
+
+        // Transparent spend detection for full blocks is not yet implemented; only
+        // received transparent outputs are scanned here.
```
nullcopy, line 478: Does this need a TODO/Issue?

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,721 @@
...
+        #[cfg(feature = "orchard")]
+        ScannedBundles::new(
+            pos_tracker.orchard_final_tree_size,
+            orchard_note_commitments,
+            orchard_nullifier_map,
+        ),
+    ))
+}
+
+impl PositionTracker {
```
nullcopy, line 655: These methods could use some unit tests

nullcopy (changes requested):
Reviewed as of 4913e6ab6a533fc45feb542716531438a20597e0

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,834 @@
...
+        assert_eq!(self.sapling_tree_position, self.sapling_final_tree_size);
+        #[cfg(feature = "orchard")]
+        assert_eq!(self.orchard_tree_position, self.orchard_final_tree_size);
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dannywillems, line 755: All of these tests are independent from the shielded protocol. Therefore, we could generalize it using proptest. I suggest [a patch](https://github.com/zcash/librustzcash/pull/2401) in zcash_protocol adding an `arb_protocol` strategy that can be used here.

nullcopy:
[Force-pushed](https://github.com/zcash/librustzcash/compare/afc93dac85ff304ab8a6f278772b34fe7bc88618..ef3b4880b1475bacef12aeace34c7780d39513e3) to rebase on main

zcash_client_backend/src/scanning/full.rs
nullcopy: Not directly. The bound comes from TxIndex being a u16 newtype in zcash_protocol. It can't actually fire in practice because the 2 MB consensus block-size limit caps the number of transactions per block well below 2¹⁶

nullcopy:
Pushed https://github.com/zcash/librustzcash/pull/2246/commits/b4d01ca40da359a1169cbbd6c17623466376062d to address review feedback

zcash_client_backend/src/scanning.rs
```diff
@@ -656,22 +680,25 @@ impl PositionTracker {
...
-    spends: &[Spend],
+fn find_spent<'a, I, AccountId, Spend, Nf, WS>(
+    spends: I,
     nullifiers: &[(AccountId, Nf)],
     extract_nf: impl Fn(&Spend) -> Nf,
     construct_wallet_spend: impl Fn(usize, Nf, AccountId) -> WS,
-) -> (Vec<WS>, Vec<Nf>) {
+) -> (Vec<WS>, Vec<Nf>)
+where
+    I: IntoIterator<Item = &'a Spend>,
```
dannywillems, line 690: I like that we are moving the type constraints in the `where` clause. I saw that it is not consistent everywhere; we might want to unify this.

zcash_client_backend/src/scan.rs
```diff
@@ -166,6 +152,45 @@ impl<IvkTag, D: Domain, M> DynamicUsage for BatchReceiver<IvkTag, D, M> {
     }
 }
 
+impl<IvkTag, D: Domain, M> BatchReceiver<IvkTag, D, M> {
+    /// Blocks until the results of the batch are ready.
+    pub(crate) fn into_results(self) -> HashMap<usize, DecryptedOutput<IvkTag, D, M>> {
+        // This iterator will end once the channel becomes empty and disconnected. We
```
dannywillems, line 158: Nit: Is this actually useful information? This seems to be explaining the inner working of `flume`, if the claims are correct.

zcash_client_backend/src/scan.rs
```diff
@@ -134,30 +134,16 @@ impl<IvkTag, D: Domain, M> DynamicUsage for OutputReplier<IvkTag, D, M> {
...
-        let item_size = std::mem::size_of::<OutputItem<IvkTag, D, M>>();
-        const ATOMIC_USIZE_SIZE: usize = std::mem::size_of::<AtomicUsize>();
-        let block_size = PTR_SIZE + ITEMS_PER_BLOCK * (item_size + ATOMIC_USIZE_SIZE);
-
-        num_blocks * block_size
+        // We count the memory usage of items still buffered in the channel on the
+        // receiver side. This is a loose lower bound rather than an exact figure:
+        // `flume` stores queued items in an internal buffer that may be over-allocated
+        // relative to its length and carries per-item bookkeeping, so the true heap use
+        // can exceed this. It is only used as a soft memory-pressure heuristic.
```
dannywillems, line 145: I confirm that the value is not the actual _heap_ memory allocated, and the claim seems correct by looking at the code of `flume`, which uses a `VecDeque` behind an `Arc`. `self.0.len(0)` returns the number of elements in the channel. The elements that have been dequeued are still in the memory, and will only be overwritten after a certain amount of time.

It does not seem to be critical at this point. I added this comment for future reference, as I went quickly through the code of `flume` and `VecDeque`.

zcash_client_backend/src/scanning.rs
```diff
@@ -510,6 +511,13 @@ pub enum ScanError {
         protocol: ShieldedProtocol,
         at_height: BlockHeight,
     },
+
+    /// The size of the note commitment tree for the given protocol would exceed the
+    /// `u32` range as a result of applying the outputs in the block being scanned.
+    TreeSizeOverflow {
```
dannywillems, line 517: Nit: this only exists because `tx_outputs` in `full.rs` is a value of type `usize`, because `tree_sizes_around` expects an iterator on `usize` values, coming from a call of `.len()` on a structure, which returns a `usize`.
This would normally never happen.

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,1007 @@
...
+/// - `activation_height` is the height at which `protocol` activated, if set.
+/// - `prior_tree_size` is the tree size as of the end of the previous block, if known.
+/// - `output_counts` yields the number of `protocol` outputs in each transaction of the
+///   block, in order.
+///
+/// Returns [`ScanError::TreeSizeUnknown`] if the starting size cannot be determined (the
+/// block is at or above the protocol's activation height but no prior size is known), or
+/// [`ScanError::TreeSizeOverflow`] if applying the block's outputs would take the tree
+/// size beyond the `u32` range.
+fn tree_sizes_around(
```
dannywillems, line 595: Nit: `protocol` could be moved at the type level as it doesn't contribute to the computation in this case; it is solely used for the error (which could also leverage the type system).

zcash_client_backend/src/scanning/full.rs
```diff
@@ -0,0 +1,1007 @@
...
+
+    fn increment_over_tx(&mut self, tx: &Transaction) {
+        self.increment(
+            sapling_output_count(tx),
+            #[cfg(feature = "orchard")]
+            orchard_action_count(tx),
+        );
+    }
+
+    fn check_end_of_block_consistency(&self) -> Result<(), ScanError> {
```
dannywillems, line 742: The return type is incorrect; there is no value of type `ScanError` returned if the check fails. The routine panicks.

dannywillems (approved):
tACK bb7ce3836788e104038c25e3be8a5c1104571d2f.

Left a few comments that I don't consider blocking, but would be nice to give another pass one day. Focusing on unblocking work.

nullcopy (approved):
I agree with every one of your points @dannywillems, but I'm going to merge this as is, since they're all minor/stylistic and time is of the essence right now. I'm making a note to revisit your suggestions

zcash_client_backend/src/scanning/full.rs
nuttycom: This should have been a blocking comment, IMO.

nullcopy: Agreed. We were too hasty here. Here's a fix: https://github.com/zcash/librustzcash/pull/2420

PR #2260, opened 2026-04-01
zcash_client_backend: Yet more refactors to `scanning` module

str4d:


zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -850,34 +864,37 @@ where
...
                         WalletTransparentOutput::from_parts(
-                            OutPoint::new(
-                                d_tx.tx().txid().into(),
-                                u32::try_from(output_index).unwrap(),
-                            ),
+                            OutPoint::new(tx.txid().into(), u32::try_from(output_index).unwrap()),
                             txout.clone(),
-                            d_tx.mined_height(),
+                            mined_height,
+                            // TODO: Confirm this is correct for `TransferType::WalletInternal`.
```
nuttycom: This will probably misclassify send-to-self transactions as incoming instead of wallet-internal, but that's okay; I don't like the existing `TransferType` machinery and will look at getting rid of it. In the mobile wallets send-to-self is essentially always just testing, because you can't directly spend transparent funds. In zallet, the situation will be more complex.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3171,6 +3174,10 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
                         outpoint.clone(),
                         TxOut::new(output.value(), ephemeral_address.script().into()),
                         None,
+                        // TODO: Is this correct?
+                        TransferType::Outgoing,
```
nuttycom: It'll do until I can eradicate `TransferType`. I'll begin that as soon as it will be nondisruptive to everything else in-flight.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -822,114 +789,113 @@ where
...
-                            OutPoint::new(
-                                d_tx.tx().txid().into(),
-                                u32::try_from(output_index).unwrap(),
-                            ),
+                            OutPoint::new(tx.txid().into(), u32::try_from(output_index).unwrap()),
                             txout.clone(),
-                            d_tx.mined_height(),
+                            mined_height,
+                            TransferType::Outgoing,
+                            account_id,
```
nuttycom: This use of `account_id` is completely wrong; the account ID is intended to be the id of the account to which the output belongs, but this is using the _funding_ account - the account that spent funds in this transaction - as that account. This is a significant downside of unifying `TransparentSentOutput` and `WalletTransparentOutput`.

nuttycom (commented):
NACK 5c04b120fd2044e9401041d8c2cb2ab34fb04c32

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -1086,36 +1046,73 @@ fn put_transparent_outputs<DbT>(
...
-        wallet_db.queue_transparent_spend_detection(
-            *received_t_output.recipient_address(),
-            tx_ref,
-            received_t_output.outpoint().n(),
-        )?;
-    }
+        // Send side: record the output as sent for each wallet account that
+        // funded the transaction. If the recipient is also a wallet account,
+        // the send is recorded as an internal transfer.
+        for &from_account in output.funding_accounts() {
```
nuttycom: Using a loop here will probably cause bugs, because this ends up duplicating `sent_notes` entries.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -948,7 +948,53 @@ pub(crate) fn utxo_query_height(
     }
 }
 
-fn to_unspent_transparent_output(row: &Row) -> Result<WalletTransparentOutput, SqliteClientError> {
+/// Returns a wallet account whose received outputs/notes were spent in the
+/// transaction with the given internal id, if any. If multiple wallet accounts
+/// funded the transaction, an arbitrary (deterministic) one is returned;
+/// `zcash_client_backend` does not currently support representing multiple
+/// funding accounts on a single output.
+fn list_funding_accounts(
```
nuttycom, line 956: This should be renamed to `find_funding_account` and should deterministically choose the account that contributed the most value; in the case of ties, break in favor of the oldest spend.

zcash_client_backend/CHANGELOG.md
```diff
@@ -74,6 +75,10 @@ workspace.
   `zcash_transparent 0.7`, `zcash_primitives 0.27`, `zcash_proofs 0.27`,
   `zcash_keys 0.13`, `pczt 0.6`.
 - `zcash_client_backend::data_api`:
+  - Changes to the `InputSource` trait:
+    - The result types of `InputSource::get_unspent_transparent_output` and
+      `InputSource::get_unspent_transparent_outputs` have each changed; these
+      have reverted to returning `WalletTransparentOutput`.
```
nullcopy: These CHANGELOG entries are being inserted into a historic release, but I assume they should be added to the PLANNED release at the top

zcash_client_backend/CHANGELOG.md
nullcopy, line 1: Claude had a lot of good feedback about CHANGELOG consistency:

```
  1. CHANGELOG and commit messages contradict the actual API (must fix before merge)

  The CHANGELOG (zcash_client_backend/CHANGELOG.md) and the bodies of commits f675015fab, d2c8d50648, and d4473639b1 consistently describe the new API as:

  - field funding_accounts: HashSet<AccountId>
  - from_parts(... funding_accounts: HashSet<AccountId>)
  - accessor funding_accounts()
  - helper list_funding_accounts(creating_tx_id) -> HashSet<AccountId>

  The actual code has:

  - field funding_account: Option<AccountId> (zcash_client_backend/src/wallet.rs:189)
  - from_parts(... funding_account: Option<AccountId>) (zcash_client_backend/src/wallet.rs:206)
  - accessor funding_account() (zcash_client_backend/src/wallet.rs:327)
  - per-backend find_funding_account(creating_tx_id) -> Option<AccountUuid> (zcash_client_sqlite/src/wallet/transparent.rs:960, zcash_client_memory/src/types/memory_wallet/mod.rs:299) that
   picks one account by a value-weighted heuristic

  The set-vs-option-vs-heuristic discrepancy reads like a mid-PR design pivot whose docs never got updated. Fix:
  - CHANGELOG funding_accounts → funding_account (lines 17, 43, 48, 61).
  - Commit message bodies (rewrite if you reflow before merge).
  - The docstring on funding_account() at wallet.rs:317-326 already correctly describes the singular-with-heuristic design, so it's just the marketing layer that's stale.
```

zcash_client_memory/src/types/transparent.rs
```diff
@@ -124,11 +124,18 @@ impl ReceivedTransparentOutput {
...
+    ) -> Option<WalletTransparentOutput<AccountId>> {
         WalletTransparentOutput::from_parts(
             outpoint.clone(),
             self.txout.clone(),
             mined_height,
+            match self.key_scope {
+                TransparentKeyScope::INTERNAL | TransparentKeyScope::EPHEMERAL => {
+                    zcash_client_backend::TransferType::WalletInternal
+                }
+                _ => zcash_client_backend::TransferType::Incoming,
```
nullcopy: (nit) IMO better to be exhaustive here. I don't know why we would, but if a new `TransparentKeyScope` was added, we'd want to explicitly choose how it gets classified here, rather than defaulting to `TransferType::Incoming`

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -960,13 +962,22 @@ fn to_unspent_transparent_output(row: &Row) -> Result<WalletTransparentOutput, S
...
     WalletTransparentOutput::from_parts(
         outpoint,
         TxOut::new(value, script_pubkey),
         height.map(BlockHeight::from),
+        match key_scope {
+            // TODO: Is this right?
+            Some(TransparentKeyScope::INTERNAL | TransparentKeyScope::EPHEMERAL) => {
+                zcash_client_backend::TransferType::WalletInternal
+            }
+            _ => zcash_client_backend::TransferType::Incoming,
```
nullcopy: Same comment as [above](https://github.com/zcash/librustzcash/pull/2260/changes/276779a3d901d7798d2040d7dd6d7d3fe95b9147#r3237385353)

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -1086,36 +1046,73 @@ fn put_transparent_outputs<DbT>(
...
-            on_received(account_id, *t_key_scope);
+            if let Some(t_key_scope) = output.recipient_key_scope() {
+                on_received(account_id, t_key_scope);
+            }
+
+            // When we receive transparent funds (particularly as ephemeral outputs
+            // in transaction pairs sending to a ZIP 320 address) it becomes
+            // possible that the spend of these outputs is not then later detected
+            // if the transaction that spends them is purely transparent. This is
+            // especially a problem in wallet recovery.
```
nullcopy: For my own edification, I'd like to understand this scenario better

zcash_client_backend/src/wallet.rs
```diff
@@ -86,6 +84,14 @@ pub enum Recipient<AccountId> {
         ephemeral_address: TransparentAddress,
         outpoint: OutPoint,
     },
+    /// A transparent output sent to a non-ephemeral transparent address belonging to
+    /// a wallet account. Used to record the send side of a transparent output that
+    /// the wallet both funded and received.
```
nullcopy, line 94: (nit) would this be a good time to document the other variants like this?

zcash_client_backend/src/wallet.rs
```diff
@@ -86,6 +84,14 @@ pub enum Recipient<AccountId> {
...
+    #[cfg(feature = "transparent-inputs")]
+    InternalTransparent {
+        receiving_account: AccountId,
+        recipient_address: TransparentAddress,
+    },
     InternalAccount {
```
nullcopy, line 111: Now `InternalAccount` feels misnamed. Isn't `InternalTransparent` also destined to an internal account?

zcash_client_backend/src/wallet.rs
```diff
@@ -192,18 +201,41 @@ impl WalletTransparentOutput {
...
+                funding_account,
                 known_input_size: None,
             })
     }
 
+    /// Strips the `AccountId` from this output struct.
+    ///
+    /// Used by the ZIP 321 Proposal type.
+    #[cfg(feature = "transparent-inputs")]
+    pub(crate) fn without_account_id(self) -> WalletTransparentOutput<()> {
```
nullcopy: This method feels unusual to me. If the only purpose is to redact the funding account before a ZIP321 proposal gets shared, wouldn't it be more appropriate to perform that redaction at the point where the proposal get's wire encoded for sharing?

It feels like an opportunity to have two versions of ostensibly the same data in the same type, but which disagree about the funding account.

zcash_client_backend/src/wallet.rs
```diff
@@ -228,18 +265,76 @@ impl WalletTransparentOutput {
...
+    pub fn transfer_type(&self) -> TransferType
+    where
+        AccountId: PartialEq,
+    {
+        match (&self.recipient_account, self.recipient_key_scope) {
+            (None, _) => TransferType::Outgoing,
+            (Some(_), Some(TransparentKeyScope::INTERNAL | TransparentKeyScope::EPHEMERAL)) => {
+                TransferType::AccountInternal
+            }
+            (Some(r), _) if self.funding_account.as_ref() == Some(r) => {
```
nullcopy: I think it does, but I just want to confirm this will catch standalone addresses in an account

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -948,7 +948,66 @@ pub(crate) fn utxo_query_height(
...
+/// Returns the wallet account that most likely funded the transaction with the given
+/// internal id, if any. If multiple wallet accounts contributed inputs to the
+/// transaction, the account that contributed the most value is selected; ties are
+/// broken in favor of the account whose oldest contributed input has the lowest
+/// mined height (with unmined inputs sorting last).
+///
+/// `zcash_client_backend` does not currently support representing multiple funding
+/// accounts on a single output; this heuristic provides a deterministic
+/// single-account choice when more than one wallet account contributed funds.
+fn find_funding_account(
```
nullcopy: I would prefer if this helper returned a list of funding accounts (and amounts) and let the caller decide if they only require the largest account.

Or at minimum rename the function to `find_largest_funding_account()`, but the former seems more useful IMO

nullcopy (commented):
Overall looks good. Left a few suggestions to consider.

Reviewed as of d4473639b14da1072b49cfa98799849b43b081ce

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/d4473639b14da1072b49cfa98799849b43b081ce..0773eed93509861d99b6b76228391272909f062d) to correct CHANGELOGs.

zcash_client_backend/src/wallet.rs
nuttycom: Essentially the problem is that for internal shielded outputs, the recipient address doesn't matter but for internal transparent outputs, the address matters (and we may not actually have the account identifier at that point in the data flow.)

zcash_client_backend/src/wallet.rs
nuttycom: It should probably be named `redact_account_data`, but IIRC this is an internal utility method for a case related to serizalization. I'll double-check whether we really need this.

zcash_client_backend/src/wallet.rs
nullcopy: Should `InternalAccount` be renamed to `InternalShielded`?

nullcopy (approved):
utACK c585cac66b720b8a46b978872d05bd3a8c5587fa

PR #2271, opened 2026-04-07
Implement ZIP 248

### LINKED ISSUE #2365: zcash_primitives: Add support for ZIP 248 transaction parsing.
---

arya2:
Implements [ZIP 248], the extensible transaction format, in `zcash_primitives`.

Resolves COR-1253.

## What this is

ZIP 248 replaces the fixed transaction layout with a typecode-length-value
encoding of protocol bundles plus a value pool delta map, so that a wallet can
parse, identify and value-account for a transaction containing bundle types it
does not implement. This PR adds that format behind the existing
`zcash_unstable = "nu7"` / `"zfuture"` gates (`zcash_v7` cfg alias): parsing and
serialization, the txid and authorizing data commitment digest trees, the v0
signature digest, and the transaction-local consensus rules.

## Updated to the current draft (2026-09-02)

The implementation has been brought in line with the ZIP as it now stands in
`zcash/zips`:

- **Deployed as transaction version 7.** The draft proposes the extensible
  format as V7, leaving V6 to ZIP 229 / ZIP 230. Everything is named
  accordingly (`TxVersion::V7`, `zcash_v7`, `sighash_v7.rs`, …), and
  `zcash_protocol::constants` gains `V7_TX_VERSION` plus a placeholder
  `V7_VERSION_GROUP_ID` (the ZIP has not assigned one).
- **Registry renumbered:** transparent 0, coinbase 1, Sapling 2, Orchard 3,
  Ironwood 4, fee 5, ZIP 233 NSM 6. The reserved slot is gone, and key rotation
  and lockbox disbursement moved to the unallocated table. Parsing enforces the
  registry's ❌ columns.
- **Coinbase bundle.** The coinbase is now a bundle carrying `blockHeight`,
  `blockSubsidy`, `lockboxValue` and `coinbaseData`, with its own
  `ZTxIdCoinbasHash` digest and no authorizing data. Its presence is what makes
  a transaction a coinbase transaction, so `check_v7_consensus_rules` no longer
  takes an `is_coinbase` flag. Because the bundle contributes
  `blockSubsidy - lockboxValue` to the transparent transaction value pool, the
  delta balance rule is now universal — every transaction sums to zero in every
  asset, coinbase included.
- **Ironwood bundle.** A second Orchard protocol bundle, acting on the
  *Ironwood pool*. It shares the Orchard encoding and reuses the same read and
  write paths; only the digest personalizations differ (`ZTxIdIronwd_Hash`,
  `ZTxIdIrnAct{C,M,N}Hash`, `ZTxAuthIrnwdHash`). The Orchard effect and auth
  digests are generalized over a pool's personalizations, and the `f_orchard`
  closure of `map_bundles` / `map_authorization` now applies to both pools.
- **Anchors are authorizing data.** `anchorSapling` and `anchorOrchard` moved
  out of the effecting data and are now written first in the bundle's
  authorizing data, with the re-personalized digests the draft requires:
  `ZTxIdSaplingH_v7`, `ZTxAuthSapliH_v7`, `ZTxIdOrchardH_v7`,
  `ZTxAuthOrchaH_v7`.
- **Consensus rules** follow the reorganized bundle-local / cross-bundle split,
  including the new coinbase rules (no transparent inputs, no Sapling spends,
  `enableSpends = 0` in every Orchard protocol bundle, ZEC-only asset class,
  and the value pool delta equal to `blockSubsidy - lockboxValue`).

## Known deviations and limitations

- **Unknown-bundle digests.** The draft defines no digest for a bundle a client
  does not understand; it expects the client to be *given* the 32-byte
  `bundle_effects_digest`. To keep txid computation infallible, this
  implementation falls back to a flat hash of the raw `vBundleData` under an
  implementation-defined personalization. That digest is not what a validator
  which understands the bundle computes, so a txid derived over an unknown
  bundle is provisional. This is documented at
  `zip248::opaque_effects_personalization`.
- **`enableCrossAddress`.** The draft leaves this flag unrestricted in an
  Ironwood bundle, but `orchard::bundle::Flags` has no representation for the
  bit, so a bundle that sets it cannot be parsed yet. The Orchard-bundle rule
  ("MUST be 0") is therefore enforced structurally by the `orchard` crate.
- Rules needing block context — the per-block fee sum, exactly one coinbase
  transaction first in the block, and `blockHeight` / `blockSubsidy` /
  `lockboxValue` matching the block and the issuance schedule — are left to the
  consumer and called out in the rustdoc.

## Testing

`cargo test -p zcash_primitives --all-features` passes with no cfg, with
`--cfg zcash_unstable="nu7"` and with `--cfg zcash_unstable="zfuture"`;
`cargo clippy --all-features --all-targets` and `cargo fmt --check` are clean.

New tests cover the coinbase bundle (field constraints, wire roundtrip,
whole-transaction roundtrip, the value pool delta relation), an Orchard
protocol bundle carried as the Ironwood bundle (roundtrip, and that the same
bundle digests differently in the two pools), the renumbered registry's parsing
rejections, and that the v0 signature digest of a transaction with no
transparent inputs is its txid. The serialization roundtrip property now also
compares the txid and the authorizing data commitment, which is what catches a
field being moved between effecting and authorizing data.

## Still to do

This branch is based on a commit ~50 behind `main` and has not been rebased;
current `main` renamed `core2` to `corez` and moved `orchard::Bundle::from_parts`
to `try_from_parts`, which the rebase will need to account for.

## AI disclosure

Claude Code was used for the update to the current draft: the code changes,
tests and this description. The contributor is the responsible author.

[ZIP 248]: https://zips.z.cash/zip-0248

https://claude.ai/code/session_012NvtrjynTDvJAh5tywmyb7

### DIFF
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -346,7 +346,7 @@ pub struct Builder<'a, P, U> {
     target_height: BlockHeight,
     expiry_height: BlockHeight,
     #[cfg(all(
-        any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
+        zcash_v6,
         feature = "zip-233"
     ))]
     zip233_amount: Zatoshis,
@@ -507,7 +507,7 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, ()> {
             target_height,
             expiry_height,
             #[cfg(all(
-                any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
+                zcash_v6,
                 feature = "zip-233"
             ))]
             zip233_amount: Zatoshis::ZERO,
@@ -541,7 +541,7 @@ impl<'a, P: consensus::Parameters> Builder<'a, P, ()> {
             target_height: self.target_height,
             expiry_height: self.expiry_height,
             #[cfg(all(
-                any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
+                zcash_v6,
                 feature = "zip-233"
             ))]
             zip233_amount: self.zip233_amount,
@@ -694,7 +694,7 @@ impl<P: consensus::Parameters, U> Builder<'_, P, U> {
                 },
             )?,
             #[cfg(all(
-                any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
+                zcash_v6,
                 feature = "zip-233"
             ))]
             -ZatBalance::from(self.zip233_amount),
@@ -803,7 +803,7 @@ impl<P: consensus::Parameters, U> Builder<'_, P, U> {
     }
 
     #[cfg(all(
-        any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
+        zcash_v6,
         feature = "zip-233"
     ))]
     pub fn set_zip233_amount(&mut self, zip233_amount: Zatoshis) {
@@ -1097,7 +1097,7 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_, 
             vp_deltas.set_fee(fee);
         }
         #[cfg(all(
-            any(zcash_unstable = "nu7", zcash_unstable = "zfuture"),
+            zcash_v6,
             feature = "zip-233"
         ))]
         vp_deltas.set_zip233(self.zip233_amount);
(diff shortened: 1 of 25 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -1161,23 +1176,17 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<'_,
...
-            ))]
-            zip233_amount: unauthed_tx.zip233_amount,
+        let authorized_tx = TransactionData::from_parts(
+            unauthed_tx.version(),
+            unauthed_tx.consensus_branch_id(),
+            unauthed_tx.lock_time(),
+            unauthed_tx.expiry_height(),
+            unauthed_tx.value_pool_deltas().clone(),
             transparent_bundle,
-            sprout_bundle: unauthed_tx.sprout_bundle,
```
nuttycom, line 1175: This removal assumes that v4 transactions will no longer be supported on the network. If we wanted the in-memory representation of the transaction to change prior to the NU7 upgrade, then this would need to be reverted.

PR #2274, opened 2026-04-09
WIP: for transactions in the "stable" block range, do not remove their mined-height information in truncation.

nuttycom:
Also, never truncate the note commitment tree to greater than the stable depth.

This work-in-progress will be modified to also set a flag on the wallet's notes once the shard fragment after the note has been fully scanned and the end of that scan range is within the stable block range; this flag will then supersede the scan-queue based determination for spendability for notes received within the stable block range. For notes received within the unstable range, we still need to consult the scan queue, but since that's only 100 blocks we can dispense with that quickly.

These changes will mean that notes that have been previously determined to be spendable (in terms of our ability to construct witnesses) will remain spendable so long as the 100 blocks at the chain tip have been scanned. 

Note that we should also make sure that spentness determinations based on discovered nullifiers also be considered final for spending transactions within the stable range as part of this effort.


nullcopy:
Superceded by https://github.com/zcash/librustzcash/pull/2289

PR #2275, opened 2026-04-09
Impls accessor methods on `JsDescription` and Clone on `TransactionData`

arya2:


nuttycom (changes requested):
Code LGTM; please update the CHANGELOGs.

arya2:
@nuttycom CHANGELOG has been updated in https://github.com/zcash/librustzcash/pull/2275/commits/e24afd4820591fe6d00420ac57a6eaf28bd05eeb.

nuttycom:
@arya2 looks like there are likely feature-flagging related CI failures.

arya2:
The same tests passed in the previous commit, the only changes since then have been to the changelog. 

I need to add a way to mutate transactions to make them invalid to keep some tests working in Zebra which will probably require a few more updates here.

nuttycom:
> I need to add a way to mutate transactions to make them invalid to keep some tests working in Zebra which will probably require a few more updates here.

@arya2 I don't care for this - the whole point is to make as much code as we can correct by construction, and using mutation to construct states that could never otherwise occur at the in-memory transaction representation layer seems like the wrong way to go about it to me. Is it instead possible to create serialized test vectors that represent the invalid states you're trying to reach? Such negative examples would be a valuable addition there, but here it just gives anyone a feature flag that they can use to violate invariants. What sort of tests are these needed for?

nuttycom:
@arya2 at very least, I would request that these methods be prefaced by something like `test_only_unsafe_*`

arya2:
These tests were removed from zebra-chain because they don't make sense there anymore. They might not make sense here either, as these checks happen in the block/tx verifiers rather than in the codec, but there should ideally be similar test coverage here now:

<details>

| Test | Reason |
|------|--------|
| `v5_with_duplicate_orchard_action` | Duplicating an orchard action requires modifying the orchard bundle proofs and action count, which is not feasible with byte-level manipulation |
| `coinbase_outputs_are_decryptable_for_fake_v5_blocks` | Requires constructing custom orchard actions with specific note encryption test vector fields, not feasible with byte-level manipulation |
| `shielded_outputs_are_not_decryptable_for_fake_v5_blocks` | Requires constructing custom orchard actions with specific note encryption test vector fields, not feasible with byte-level manipulation |
| `mempool_skip_accepts_block_with_garbage_orchard_proofs` | Requires constructing and modifying orchard bundles with garbage proofs, not feasible with byte-level manipulation |


</details>

arya2:
> Is it instead possible to create serialized test vectors that represent the invalid states you're trying to reach? Such negative examples would be a valuable addition there, but here it just gives anyone a feature flag that they can use to violate invariants. What sort of tests are these needed for?

It is for most of them, a few of them didn't make sense, and the ones that are no longer possible probably belong in this workspace instead.

I was hoping for this sort of feedback, thank you.

nuttycom:
> > Is it instead possible to create serialized test vectors that represent the invalid states you're trying to reach? Such negative examples would be a valuable addition there, but here it just gives anyone a feature flag that they can use to violate invariants. What sort of tests are these needed for?
> 
> It is for most of them, a few of them didn't make sense, and the ones that are no longer possible probably belong in this workspace instead.
> 
> I was hoping for this sort of feedback, thank you.

If there are tests that make sense to be added here instead, that seems ideal.

arya2:
AI was wrong about two of the tests, they can be restored in Zebra with the current zcash_primitives API on this branch.

zcash_primitives/src/transaction/tests.rs
```diff
@@ -459,3 +459,209 @@ fn zip_0233() {
...
+        nonempty::NonEmpty::from_vec(actions_vec).expect("vec has at least two elements");
+    let dup_bundle = orchard::bundle::Bundle::from_parts(
+        dup_actions,
+        *bundle.flags(),
+        *bundle.value_balance(),
+        *bundle.anchor(),
+        bundle.authorization().clone(),
+    );
+
+    assert!(orchard_bundle_has_duplicate_nullifiers(&dup_bundle));
```
nuttycom: I don't understand the use of this test; it's just checking that the new test code above can correctly identify Orchard bundles that contain duplicate nullifiers? This seems tautological.

zcash_primitives/src/transaction/tests.rs
```diff
@@ -459,3 +459,209 @@ fn zip_0233() {
...
+        keys::OutgoingViewingKey,
+        note::{ExtractedNoteCommitment, Nullifier, TransmittedNoteCiphertext},
+        note_encryption::OrchardDomain,
+        primitives::redpallas::{self, SpendAuth},
+        value::ValueCommitment,
+    };
+    use zcash_note_encryption::try_output_recovery_with_ovk;
+
+    // Test vector 1 from the Orchard note encryption test vectors.
+    // Source: https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_note_encryption.py
```
nuttycom: For these test vectors we prefer to copy them unmodified from the test vectors into their own module; see for example https://github.com/zcash/librustzcash/pull/2199/changes/0f6864b3b15638ca64f605e529f78d137c48ccee#diff-a970457f8a351680457fdb55858451dc94508be1cbddde8adaf20cb773934e0c - we add a `test_vectors` module in the appropriate location, and then just copy in the test vectors. Also, as a note, the repository has moved so the canonical test vector source is https://github.com/zcash/zcash-test-vectors.

zcash_primitives/src/transaction/tests.rs
```diff
@@ -459,3 +459,209 @@ fn zip_0233() {
...
+/// checked that a coinbase transaction's Orchard outputs are decryptable with the correct
+/// OVK. For coinbase transactions, ZIP-212 mandates the all-zeros OVK so that any
+/// observer can recover the output. Here we use the OVK from the Orchard note encryption
+/// test vectors (which uses a non-zero OVK) to verify the mechanism is correct; the
+/// exact same code path applies when the OVK is all-zeros.
+///
+/// This also verifies that a *wrong* OVK fails, which corresponds to Zebra's
+/// `shielded_outputs_are_not_decryptable_for_fake_v5_blocks`.
+#[test]
+fn orchard_output_recovery_with_known_ovk() {
```
nuttycom: Correct me if I'm wrong, but it looks like this test could be in the `orchard` crate instead? I don't see that it depends on anything from `zcash_primitives`.

nuttycom (changes requested):
I have some questions about the tests being added here. Also as a nit, the commit message of acef4765ddc86337297749dcfcde1d53a5ac7074 doesn't meet our contribution guidelines as expressed in CONTRIBUTING.md - remember that we don't squash commits, so we require an informative commit message for each commit.

arya2:
@nuttycom the second test should go in the `orchard` crate and the first test can be restored in Zebra.

zcash_primitives/src/transaction/tests.rs
arya2: It was useful in bringing my attention to the API it was using, but otherwise pointless.

nuttycom (approved):
utACK e24afd4820591fe6d00420ac57a6eaf28bd05eeb

PR #2278, opened 2026-04-10
harden transparent sighash against out-of-range input index

Zk-nd3r:
move the transparent-input index check into `zcash_transparent::sighash::SignableInput::from_parts`, so invalid indices are rejected at construction time rather than at sighash use sites.

`SignableInput::from_parts` now takes `&Bundle<_>` and returns `Result<_, InvalidInputIndex>`. this makes the invariant explicit in the type layer and lets `zcash_primitives` trust constructed transparent signable inputs.

updated:
- `zcash_transparent/src/sighash.rs`
- `zcash_transparent/CHANGELOG.md`
- `zcash_primitives/src/transaction/tests.rs`
- `pczt/tests/end_to_end.rs`

### DIFF
--- a/pczt/tests/end_to_end.rs
+++ b/pczt/tests/end_to_end.rs
@@ -152,28 +152,36 @@ fn transparent_to_orchard() {
         Some(
             v5_signature_hash(
                 &tx_effects,
-                &SignableInput::Transparent(::transparent::sighash::SignableInput::from_parts(
-                    match (hash_type.signed_outputs(), hash_type.anyone_can_pay()) {
-                        (zcash_script::signature::SignedOutputs::All, false) => SighashType::ALL,
-                        (zcash_script::signature::SignedOutputs::All, true) => {
-                            SighashType::ALL_ANYONECANPAY
-                        }
-                        (zcash_script::signature::SignedOutputs::Single, false) => {
-                            SighashType::SINGLE
-                        }
-                        (zcash_script::signature::SignedOutputs::Single, true) => {
-                            SighashType::SINGLE_ANYONECANPAY
-                        }
-                        (zcash_script::signature::SignedOutputs::None, false) => SighashType::NONE,
-                        (zcash_script::signature::SignedOutputs::None, true) => {
-                            SighashType::NONE_ANYONECANPAY
-                        }
-                    },
-                    0,
-                    &Script(script_code.clone()),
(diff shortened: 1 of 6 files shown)


nuttycom (dismissed):
utACK 0ef69f880c094b1b43c627bf53f07473dcc5dfe6. If you want to make the breaking API change now, I would accept it as we're about to make a major-version release of `zcash_primitives` and it makes sense to include this change.

Zk-nd3r:
Appreciate the utACK and the release-window offer.

Caught that the transparent input is wrapped: `SignableInput::Transparent(transparent::sighash::SignableInput<'a>)`. Clean typed-index hardening wants `zcash_transparent`, not `zcash_primitives`. Two paths:

(A) `v4_signature_hash` / `v5_signature_hash` return `Result<_, SighashError>` with `InvalidInputIndex`. Stays in `zcash_primitives`. Moves the error earlier but doesn't eliminate the class: callers can still construct a bad `SignableInput`.

(B) Validating constructor on `zcash_transparent::sighash::SignableInput` taking `&TransparentBundle<_>`, returning `Result`. Defense-at-type-layer, sighash trusts its input unconditionally. Is `zcash_transparent` in the same major-version train?

Your call on scope. I'll push on top of `0ef69f8`.

nuttycom:
> Appreciate the utACK and the release-window offer.
> 
> Caught that the transparent input is wrapped: `SignableInput::Transparent(transparent::sighash::SignableInput<'a>)`. Clean typed-index hardening wants `zcash_transparent`, not `zcash_primitives`. Two paths:
> 
> (A) `v4_signature_hash` / `v5_signature_hash` return `Result<_, SighashError>` with `InvalidInputIndex`. Stays in `zcash_primitives`. Moves the error earlier but doesn't eliminate the class: callers can still construct a bad `SignableInput`.
> 
> (B) Validating constructor on `zcash_transparent::sighash::SignableInput` taking `&TransparentBundle<_>`, returning `Result`. Defense-at-type-layer, sighash trusts its input unconditionally. Is `zcash_transparent` in the same major-version train?
> 
> Your call on scope. I'll push on top of `0ef69f8`.

Let's go with the `zcash_transparent` change, that sounds like the correct resolution here.

nuttycom (approved):
utACK 1ecb60ea582057d593920b126f0e781fe2e9aac9

PR #2279, opened 2026-04-10
feat(zcash_client_sqlite): Retain stable witness data during truncation

### LINKED ISSUE #2276: feat: Stabilized notes should be spendable after truncation

**Proposed solution:**

nullcopy:
Add a `witness_stabilized` flag to received notes that durably records when a note's shard has been fully scanned. Spendability queries use this flag instead of transient scan-queue state, so notes remain spendable when scan ranges are rewritten (e.g. importing a new account with an old birthday) or when the wallet is explicitly truncated.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -2413,6 +2413,7 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
         let mut stmt_select_notes = tx.prepare_cached(&format!(
             "SELECT accounts.uuid, rn.id, rn.value, rn.is_change, rn.recipient_key_scope,
                     scan_state.max_priority,
+                    rn.witness_stabilized,
                     t.mined_height,
                     IFNULL(t.trust_status, 0) AS trust_status,
                     MAX(tt.mined_height) AS max_shielding_input_height,
@@ -2483,20 +2484,31 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
 
             let tx_shielding_inputs_trusted = row.get::<_, bool>("min_shielding_input_trust")?;
 
-            // A note is spendable if we have enough chain tip information to construct witnesses,
-            // the shard that its witness resides in is sufficiently scanned that we can construct
-            // the witness for the note, and the note has enough confirmations to be spent.
-            let is_spendable = any_spendable
-                && max_priority <= ScanPriority::Scanned
-                && confirmations_policy.confirmations_until_spendable(
-                    target_height,
-                    PoolType::Shielded(protocol),
-                    recipient_key_scope.and_then(|k| zip32::Scope::try_from(k).ok()),
-                    received_height,
-                    tx_trusted,
-                    max_shielding_input_height,
-                    tx_shielding_inputs_trusted,
-                ) == 0;
+            let witness_stabilized = row.get::<_, bool>("witness_stabilized")?;
+
+            // A stabilized note whose containing transaction is still mined is
+            // unconditionally spendable: at stabilization it was already confirmed beyond
+            // `PRUNING_DEPTH` (which exceeds any reasonable confirmation policy), and its
+            // witness data is durable in the shard tree. This remains true even when
+            // truncation has rolled the wallet's scanned chain tip back below the note's
+            // `mined_height`, because `truncate_to_height_internal` retains `mined_height`
+            // for fully-stabilized transactions. If the transaction has been un-mined —
+            // e.g. because some sibling note in the same tx was not stabilized — then the
+            // note is not currently spendable even if its own flag is still set.
+            // Non-stabilized notes require their shard to be sufficiently scanned and the
+            // usual confirmations policy to be met.
+            let is_spendable = (witness_stabilized && received_height.is_some())
+                || (any_spendable
+                    && max_priority <= ScanPriority::Scanned
+                    && confirmations_policy.confirmations_until_spendable(
+                        target_height,
+                        PoolType::Shielded(protocol),
+                        recipient_key_scope.and_then(|k| zip32::Scope::try_from(k).ok()),
+                        received_height,
+                        tx_trusted,
+                        max_shielding_input_height,
+                        tx_shielding_inputs_trusted,
+                    ) == 0);
 
             let is_pending_change =
                 is_change && received_height.iter().all(|h| h > &trusted_height);
@@ -3513,6 +3525,31 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
     )
 }
 
+/// Returns `true` if the given pool has any received notes marked as `witness_stabilized`
+/// whose containing transaction is currently mined at a height strictly greater than
+/// `truncation_height`. Tree operations that would otherwise delete witness data above
+/// that height (namely `truncate_to_checkpoint` and `insert_frontier` at a position
+/// below `mined_height`) must be skipped for such pools.
+fn pool_has_stabilized_notes_above(
+    conn: &rusqlite::Transaction<'_>,
+    table_prefix: &'static str,
+    truncation_height: BlockHeight,
+) -> Result<bool, SqliteClientError> {
+    conn.query_row(
+        &format!(
+            "SELECT EXISTS (
+                 SELECT 1 FROM {table_prefix}_received_notes rn
+                 INNER JOIN transactions t ON t.id_tx = rn.transaction_id
+                 WHERE rn.witness_stabilized = 1
+                   AND t.mined_height > :height
+             )"
+        ),
+        named_params![":height": u32::from(truncation_height)],
+        |row| row.get::<_, bool>(0),
+    )
+    .map_err(SqliteClientError::from)
+}
+
 pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
     conn: &rusqlite::Transaction,
     params: &P,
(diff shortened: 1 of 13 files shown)


zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,114 @@
...
+// `zcash_client_backend::data_api::ORCHARD_SHARD_HEIGHT` is only defined under that feature
+// flag. We define a local mirror here; it must equal `ORCHARD_SHARD_HEIGHT`, which is a
+// protocol-defined constant derived from `orchard::NOTE_COMMITMENT_TREE_DEPTH / 2`.
+const ORCHARD_SHARD_HEIGHT: u8 = 16;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x688f7dd4_a8d4_45ba_94a3_dfc520fefed5);
+
+const DEPENDENCIES: &[Uuid] = &[
+    v_tx_outputs_key_scopes::MIGRATION_ID,
+    ivk_item_cache::MIGRATION_ID,
```
nullcopy: Why does this depend on both of these? In fact, I think this migration depends on neither of these. It only depends on whatever changes led up to the current note tables, right?

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::wallet::init::migrations::tests::test_migrate;
+
+    #[test]
+    fn migrate() {
+        test_migrate(&[super::MIGRATION_ID]);
```
nullcopy: Need more comprehensive migration tests

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +370,57 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
 
     replace_queue_entries::<SqliteClientError>(conn, &query_range, replacement, false)?;
 
+    // Mark notes whose containing shard is complete and confirmed beyond the pruning depth.
+    let last_scanned_height = range.end - 1;
+    mark_stabilized_notes(conn, last_scanned_height)?;
```
nullcopy, line 385: Does this have to wait for completion? I feel like the marking notes as stabilized can be done JIT (or at least somewhat synchronously with scanning) as shard end heights are finalized

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,114 @@
...
+
+             ALTER TABLE orchard_received_notes
+               ADD COLUMN witness_stabilized INTEGER NOT NULL DEFAULT 0;",
+        )?;
+
+        // Backfill: set witness_stabilized for existing notes whose containing shard is
+        // complete and confirmed beyond the pruning depth. When scan_queue is empty the
+        // MAX subquery yields NULL, causing the <= comparison to evaluate to false, so
+        // no rows are updated.
+        transaction.execute(
```
nullcopy: Should these backfill transactions use `mark_stabilized_notes()`? Or perhaps a helper can abstract away the commonalities between `mark_stabilized_notes()` and the sqlite transactions here?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3533,21 +3570,39 @@ pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
         ))
     })?;
 
-    // Delete from the scanning queue any range with a start height greater than the
-    // truncation height, and then truncate any remaining range by setting the end
-    // equal to the truncation height + 1. This sets our view of the chain tip back
-    // to the retained height.
-    conn.execute(
```
nullcopy: I think these scanning queue truncations were correct as is

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
nullcopy: the migration test is underpowered for this migration. test_migrate() just runs the migration against an empty database and checks it doesn't error — so it validates only the ALTER TABLE statements, not the
  backfill logic.

  The backfill has three non-trivial conditions it won't exercise:
  - shard.subtree_end_height IS NOT NULL
  - shard.subtree_end_height <= (MAX(block_range_end) - 1 - PRUNING_DEPTH) — with an empty scan_queue the MAX is NULL and the comparison silently returns false
  - (commitment_tree_position >> SHARD_HEIGHT) = shard.shard_index

  A table-driven test with a few seeded scenarios would be worth adding:
  - Note in a shard with subtree_end_height well below stable_height → should become stabilized.
  - Note in a shard whose subtree_end_height is above stable_height → should remain unstable.
  - Note in a shard with no subtree_end_height → should remain unstable.
  - Note where commitment_tree_position IS NULL → should remain unstable.
  - Wallet with an empty scan_queue → backfill is a no-op.

  The current integration tests (stabilized_notes_survive_truncation, etc.) exercise mark_stabilized_notes post-migration, but they don't catch a bug in the one-shot backfill that runs only during the migration
  itself. If an existing wallet with stabilized shards in the old schema migrated incorrectly, it'd be invisible to the current tests.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +370,69 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+}
+
+/// Marks notes as stabilized when their containing shard is complete and the shard's end
+/// height is at or below `stable_height`. Once stabilized, a note's witness data is
+/// considered durable and will be preserved across truncation.
+///
+/// `stable_height` is the height beyond which chain reorganizations are not expected to
+/// occur; typically the max scanned height minus [`PRUNING_DEPTH`]. Callers are
+/// responsible for computing it.
+pub(crate) fn mark_stabilized_notes(
```
nullcopy, line 393: Should this move to db.rs?

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,386 @@
...
+    fn migrate() {
+        test_migrate(&[MIGRATION_ID]);
+    }
+
+    /// End-to-end exercise of the backfill:
+    /// seeds the pre-migration schema with a variety of notes whose stabilization
+    /// outcomes are fully determined, runs the migration, and asserts the expected
+    /// `witness_stabilized` value for each.
+    #[test]
+    fn migrate_backfills_stabilized_notes() {
```
nullcopy, line 110: I like these migration tests

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -133,10 +134,10 @@ pub(super) fn all_migrations<
     //                     \                       \         v_received_output_spends_account      /        /
     //                      \                       \               /                             /        /
     //                       `------------------- account_delete_cascade ---------------------------------'
-    //                                              /               \
-    //                               v_tx_outputs_key_scopes    standalone_p2sh
-    //                                                                  |
-    //                                                           ivk_item_cache
+    //                                          /       |        \
+    //                    v_tx_outputs_key_scopes  standalone_p2sh  witness_stabilized_notes
```
nullcopy, line 138: This is correct

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1092,6 +1092,34 @@ pub(crate) fn put_shard_roots<
...
+    // `put_shard_roots` is typically invoked when subtree roots are downloaded from
+    // the light wallet server — independently of scan completion — stabilize here
+    // rather than waiting for the next `scan_complete` call. The equivalent call in
+    // `scan_complete` handles the complementary case where `stable_height` advances
+    // without any shard roots being finalized.
+    //
+    // `stable_height` is derived from the scan queue using the same rule as the
+    // migration backfill. If the scan queue is empty (i.e., no scanning has occurred
+    // yet), the helper is not called and the UPDATE is skipped.
+    let stable_height: Option<u32> = conn
```
nullcopy: This is a good addition

nullcopy (commented):
Commented on 7a47b3831e5ec0e4c6a768e786c8ac608cd502d8

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3533,21 +3570,39 @@ pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
...
-        named_params![":new_end_height": u32::from(truncation_height + 1)],
-    )?;
+    // If there are stabilized notes whose `mined_height` is above the truncation height,
+    // we perform a "soft" truncation that preserves the state needed to spend them
+    // without re-scanning: the scan queue is not trimmed (so `chain_tip_height` stays at
+    // the pre-truncation value), the note commitment trees are not truncated (so
+    // stabilized witness data remains queryable), and the `blocks` rows above the
+    // truncation height are retained. Non-stabilized transactions and un-mined blocks
+    // are handled by the tx-level un-mine SQL below regardless.
+    let sapling_has_stabilized_above =
```
nullcopy, line 3580: I'm concerned that this soft/hard truncation approach modifies the scan queue in a way that will break re-scan. e.g. if ZODL truncates the state, I'm worried the 'soft truncate' path will prevent the necessary blocks from being re-scanned.

I don't think this regression would be caught by integration tests, so first we should add a test to affirm this.

PR #2281, opened 2026-04-11
pczt: Add public getter for spend_auth_sig

p0mvn:
Based on `maint/pczt-0.5.x` (branched from `pczt-0.5.0` series). This is a
SemVer-compatible addition intended for a `0.5.x` point release.

Expose the Orchard spend authorization signature via `getset`, enabling downstream consumers to read back the signature after signing 

## How We Use It

Token Holder Voting.

We use spend_auth_sig() as a read-only accessor to extract the 64-byte spend authorization signature from a Keystone-signed PCZT, so it can be forwarded to the cosmos chain as part of the delegation transaction.

### DIFF
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -13,6 +13,7 @@ workspace.
 ### Added
 - `pczt::ExtractError`
 - `pczt::EffectsOnly`
+- `pczt::orchard::Spend::spend_auth_sig` getter (via `getset`).
 - `pczt::roles::signer`:
   - `Signer::sighash`
   - `Signer::append_transparent_signature`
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK 2f6769b7710f4359dd569ba462c44df2e03d0e97

nuttycom:
CI is broken here because of the core2 yank. Given that we have to release the whole stack anyway, I'm going to rebase and retarget this to `main`.

nuttycom:
@p0mvn oh, I just realized I can't push to the valargroup branches, so you'll have to do the rebase.

p0mvn:
@nuttycom rebase completed. Also, granted access to the valargroup fork.

ValarDragon:
Thank you!!

nuttycom (approved):
re-utACK

PR #2282, opened 2026-04-11
pczt: Add public getter for shielded sighash

p0mvn:
Based on `maint/pczt-0.5.x` (branched from `pczt-0.5.0` series). This is a
SemVer-compatible addition intended for a `0.5.x` point release.

Backports the `Signer::shielded_sighash` accessor from main (added in zcash/librustzcash#2047) to the 0.5.x series, enabling downstream consumers to obtain the sighash for producing signatures externally (e.g. via a hardware wallet).

## How We Use It

Token Holder Voting.

We use `shielded_sighash()` to extract the ZIP-244 transaction digest from the finalized PCZT so that both signing paths (software key derivation and Keystone hardware) produce a spend-auth signature over the same canonical hash.

### DIFF
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -14,10 +14,11 @@ workspace.
 - `pczt::ExtractError`
 - `pczt::EffectsOnly`
 - `pczt::roles::signer`:
-  - `Signer::sighash`
   - `Signer::append_transparent_signature`
-  - `Signer::apply_sapling_signature`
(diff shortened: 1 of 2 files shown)


nuttycom (dismissed):
utACK

nuttycom (dismissed):
utACK

nuttycom (approved):

nuttycom:
Duplicate of https://github.com/zcash/librustzcash/pull/2047

PR #2283, opened 2026-04-11
zcash_client_sqlite: Add historical height witness generation

p0mvn:
## Summary

Add `WalletDb::generate_orchard_witnesses_at_historical_height`, which produces Orchard Merkle witnesses anchored at an earlier tree state.

This is needed for applications such as token-holder voting, where the wallet tree has advanced past the height at which witnesses are required.

The implementation copies shard and cap data into an ephemeral in-memory database, inserts the caller-provided frontier as a checkpoint, and generates a witness for each requested note position. The wallet DB is strictly read-only.

See analysis of the implementation choice [here](https://valargroup.gitbook.io/shielded-vote-docs/appendices/membership-proofs-at-historical-heights)

### Branching note

This PR targets `maint/zcash_client_sqlite-0.19.x` as a
SemVer-compatible addition (new public method, no breaking changes).

### How We Use It

Token Holder Voting.

Generate witnesses at a snapshot height to delegate voting authority.

### Changes

- **`zcash_client_sqlite/src/wallet/commitment_tree.rs`** — new
  `pub(crate) fn generate_orchard_witnesses_at_historical_height` plus
  private helpers `create_orchard_tree_tables` / `copy_orchard_tree_data`
  that clone the shard-tree schema and data into an ephemeral in-memory DB.
  DDL reuses the canonical `TABLE_ORCHARD_TREE_*` constants from `db.rs`.
- **`zcash_client_sqlite/src/lib.rs`** — public
  `WalletDb::generate_orchard_witnesses_at_historical_height` method
  that delegates to the above.
- **`zcash_client_sqlite/CHANGELOG.md`** — added entry under
  `[Unreleased]`.
- Unit test (`witnesses_at_historical_height`) verifying the generated
  witness root matches the expected frontier root.


zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1171,6 +1173,175 @@ pub(crate) fn check_witnesses(
...
+        incrementalmerkletree::MerklePath<
+            orchard::tree::MerkleHashOrchard,
+            { orchard::NOTE_COMMITMENT_TREE_DEPTH as u8 },
+        >,
+    >,
+    SqliteClientError,
+> {
+    let frontier_position = frontier_at_height.position();
+
+    let mem_conn = rusqlite::Connection::open_in_memory().map_err(SqliteClientError::DbError)?;
```
nuttycom: An alternative to using an in-memory rusqlite would be to use the fully in-memory `ShardStore` impl: https://github.com/zcash/incrementalmerkletree/blob/main/shardtree/src/store/memory.rs - that might be better for your purposes?

p0mvn: @nuttycom thanks for the rebase and the recommendation.

I have applied the recommendation here:
[b3fabf3](https://github.com/zcash/librustzcash/pull/2283/commits/b3fabf316ceeb32a542d4e86710894e4286c36c6)

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1171,6 +1176,120 @@ pub(crate) fn check_witnesses(
...
+            marking: Marking::None,
+        },
+    )
+    .map_err(|e| {
+        SqliteClientError::CorruptedData(format!("failed to insert frontier nodes: {}", e))
+    })?;
+
+    tree.store_mut()
+        .add_checkpoint(height, Checkpoint::at_position(frontier_position))
+        .expect("add_checkpoint is infallible");
```
nuttycom: I'm confused by this explicit checkpoint addition; the frontier insertion should have added the checkpoint?

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1171,6 +1176,120 @@ pub(crate) fn check_witnesses(
...
+    // Insert frontier + checkpoint
+    tree.insert_frontier_nodes(
+        frontier_at_height,
+        Retention::Checkpoint {
+            id: height,
+            marking: Marking::None,
+        },
+    )
+    .map_err(|e| {
+        SqliteClientError::CorruptedData(format!("failed to insert frontier nodes: {}", e))
```
nuttycom: This now seems like the wrong error; indeed, is `SqliteClientError` the correct error type for this function at all?

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1171,6 +1176,120 @@ pub(crate) fn check_witnesses(
...
+    for &pos in note_positions {
+        let merkle_path = tree
+            .witness_at_checkpoint_id(pos, &height)
+            .map_err(|e| {
+                SqliteClientError::CorruptedData(format!(
+                    "failed to generate witness for position {}: {} \
+                     (wallet may need to sync through historical height)",
+                    u64::from(pos),
+                    e
+                ))
```
nuttycom: This also argues for a bespoke error type; the caller can't automatically respond to this error, but it sort of needs to be able to do so.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1171,6 +1176,120 @@ pub(crate) fn check_witnesses(
...
+            })?
+            .ok_or_else(|| {
+                SqliteClientError::CorruptedData(format!(
+                    "no witness available for position {} \
+                     (wallet missing shard data — sync through historical height)",
+                    u64::from(pos)
+                ))
```
nuttycom: Same issue here.

nuttycom (commented):
A couple of the error types now seem incorrect after switching to use the in-memory shard store.

zcash_client_sqlite/src/wallet/commitment_tree.rs
p0mvn: Right, `Retention::Checkpoint { id: height, .. }` already registers the checkpoint inside `insert_frontier_nodes`. Initially, I thought it would be helpful for redundancy and explicitness, but I can see that it adds confusion.

So, I removed the explicit `add_checkpoint` in [7043dd1](https://github.com/zcash/librustzcash/pull/2283/commits/7043dd1732e5d2af958fe77593546167c9048e11)

zcash_client_sqlite/src/wallet/commitment_tree.rs
p0mvn: All other methods in the crate returned `SqliteClientError` error type, so this was chosen for consistency with the pattern.

I have improved the errors so that the caller can respond to them. However, I think that it would be best to keep `SqliteClientError` for consistency with the crate patterns.

Concretely, [c68d769](https://github.com/zcash/librustzcash/pull/2283/commits/c68d7692fa1330ed765b905e6bfa4b1da49aeb85) adds two orchard-gated variants that callers can match on:

- `SqliteClientError::HistoricalFrontierInvalid(InsertionError)` - the caller supplied a bad frontier.
- `SqliteClientError::HistoricalWitnessUnavailable { position, height }` - the wallet isn't synced through height, the checkpoint was pruned, or position isn't in the wallet.
Shard-read failures continue to surface as `SqliteClientError::CommitmentTree`.

If we want this changed, I think it would make sense to file a tech debt issue that would refactor the errors in the crate as a whole.

Please let me know if there is an alternative recommendation.

zcash_client_sqlite/src/wallet/commitment_tree.rs
p0mvn: Fixed in [c68d769](https://github.com/zcash/librustzcash/pull/2283/commits/c68d7692fa1330ed765b905e6bfa4b1da49aeb85): this now returns `SqliteClientError::HistoricalWitnessUnavailable { position, height }`, which the caller can match on programmatically. That lets the caller trigger further syncing or drop position from the requested set, instead of string-matching on a formatted message.

See the thread on [L1259](https://github.com/zcash/librustzcash/pull/2283#discussion_r3139557453) for the rationale for keeping these as `SqliteClientError` variants rather than a bespoke error type.

zcash_client_sqlite/src/wallet/commitment_tree.rs
p0mvn: Same fix: [c68d769](https://github.com/zcash/librustzcash/pull/2283/commits/c68d7692fa1330ed765b905e6bfa4b1da49aeb85). `witness_at_checkpoint_id` returns `Result<Option<MerklePath>, ShardTreeError>`, so "no witness" has two representations: `Err(ShardTreeError::Query(_))` (handled on L1293) and `Ok(None)` (handled on L1302). 

Both now produce `SqliteClientError::HistoricalWitnessUnavailable { position, height }`, since they're semantically the same "no witness at this height for this position" situation from the caller's perspective.

p0mvn:
@nuttycom Thanks for the careful review. I pushed two follow-up commits addressing all four inline comments:

- [7043dd1](https://github.com/zcash/librustzcash/pull/2283/commits/7043dd1732e5d2af958fe77593546167c9048e11) removes the redundant `add_checkpoint` (L1264).
- [c68d769](https://github.com/zcash/librustzcash/pull/2283/commits/c68d7692fa1330ed765b905e6bfa4b1da49aeb85) replaces the misleading `CorruptedData` stringification with two new orchard-gated `SqliteClientError` variants so callers can respond programmatically (L1259, L1277, L1284):
   * `HistoricalFrontierInvalid(InsertionError)` - caller-supplied frontier inconsistent with the wallet's shards.
   * `HistoricalWitnessUnavailable { position, height }` - no witness available (wallet likely not synced through height, checkpoint pruned, or position not in wallet).


I kept the errors as `SqliteClientError` variants rather than introducing a bespoke error type, for consistency with the rest of the crate. Please see [my reply on L1259](https://github.com/zcash/librustzcash/pull/2283#discussion_r3139557453) for the reasoning.

Happy to split them out into a dedicated error type if you'd prefer. This seemed the more conservative of the two options.

Ready for another look when you have a moment.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1171,6 +1178,138 @@ pub(crate) fn check_witnesses(
...
+        );
+
+    // Insert the frontier. `Retention::Checkpoint` causes `ShardTree` to
+    // register a checkpoint at `height` internally.
+    //
+    // `MemoryShardStore::Error` is `Infallible`, so the only variants of
+    // `ShardTreeError` we can observe are `Insert` (a caller-supplied
+    // frontier inconsistent with the loaded shards) and `Query` (an
+    // inconsistency in the loaded shards themselves).
+    tree.insert_frontier_nodes(
```
nuttycom, line 1289: One more issue here: you may want to truncate the existing chain-tip shard by at least one checkpoint; otherwise, if the added frontier corresponds to a prunable state of the tree, you could end up with your newly added checkpoint immediately pruned if it causes the number of checkpoints in the tree to exceed MAX_CHECKPOINTS. It would be good to add a test to make sure, but I'm pretty sure this is the case.

p0mvn: On it

p0mvn: Since the in-memory store starts with zero checkpoints and we add exactly one, the newly added checkpoint cannot be pruned on the current path. Even if the frontier sits at a "prunable" position relative to the loaded shard data. The `CHECKPOINT` retention flag stored inside the shard leaf is independent of `MemoryShardStore::checkpoint_count()`.

The existing test `witnesses_at_historical_height` already exercises a "historical" scenario (wallet advanced past historical height, frontier extracted at an earlier position), and it passes.

I am still working on making a convincing test, but I am fairly confident the current flow is safe.

p0mvn: Dug into this. The current code is safe.

`MemoryShardStore::empty()` starts with zero checkpoints, `put_shard/put_cap` don't touch the checkpoint map, and `max_checkpoints = 1`. So when `insert_frontier_nodes` calls `add_checkpoint` then `prune_excess_checkpoints`, the prune check is `1 > 1 == false` and the historical checkpoint survives.

Added in [5e08af9802](https://github.com/zcash/librustzcash/pull/2283/commits/5e08af9802):

- Regression test:
   * Appends a marked note + 250 checkpoints (with `WALLET_MAX_CHECKPOINTS = 100`) to force the wallet pruner to evict the historical checkpoint from SQLite, asserts min_checkpoint_id > historical_height as a precondition, then verifies witness reconstruction matches the captured frontier root.
- "Pruning-safety invariant" comment on the production code, naming the three changes that would break it:
      * pre-loading wallet checkpoints
      * an extra add_checkpoint call
      * `max_checkpoints < 1`.

nuttycom (dismissed):
utACK 5e08af9802b5eeb9c7c5efdd6d7104187b74a90f

nuttycom:
@p0mvn the merge of your other PR caused conflicts here, can you please resolve them?

p0mvn:
Yes, on it. Thanks for the reviews

p0mvn:
@nuttycom conflicts have been resolved. Thank you again

nuttycom (approved):
re-utACK 2a1ee686bab46e9016195c766d628ef2b5afc1d7; once CI passes I'll push this directly to avoid an extraneous merge commit.

p0mvn:
> re-utACK [2a1ee68](https://github.com/zcash/librustzcash/commit/2a1ee686bab46e9016195c766d628ef2b5afc1d7); once CI passes I'll push this directly to avoid an extraneous merge commit.

Ah, my bad on the merge commit instead of rebase.

Tracking the rules now, but I am going to avoid resetting/rebasing myself to keep CI running.

PR #2284, opened 2026-04-11
zcash_client_sqlite: Add WalletDb::get_unspent_orchard_notes_at_historical_height

p0mvn:
## Summary

- Add `WalletDb::get_unspent_orchard_notes_at_historical_height`, which returns all Orchard notes that existed and were unspent at a given block height for a specific account.

See the full integration note: https://hackmd.io/HstXV5RfSdOSPdEauotrXw

### Branching note

This PR targets maint/zcash_client_sqlite-0.19.x as a
SemVer-compatible addition (new public method, no breaking changes).

### How We Use It

Token Holder Voting.

Query the set of notes a wallet held at a governance snapshot height, then
pair with witnesses from `generate_orchard_witnesses_at_historical_height`
to construct voting proofs. See https://github.com/zcash/librustzcash/pull/2283

### Changes

- **`zcash_client_sqlite/src/wallet/orchard.rs`** — new `pub(crate) fn
  get_unspent_orchard_notes_at_historical_height` query that selects received notes
  mined at or before the given height, excluding any spent at or before
  that height.
- **`zcash_client_sqlite/src/lib.rs`** — public
  `WalletDb::get_unspent_orchard_notes_at_historical_height` method that delegates
  to the above.
- **`zcash_client_sqlite/CHANGELOG.md`** — added entry under `[Unreleased]`.
- Unit test (`get_unspent_orchard_notes_at_historical_height_boundary_heights`)
  verifying correct note visibility across receive, spend, and subsequent
  receive heights.

### DIFF
--- a/zcash_client_sqlite/src/wallet/orchard.rs
+++ b/zcash_client_sqlite/src/wallet/orchard.rs
@@ -173,6 +173,15 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
 /// Unlike `select_spendable_notes` (which applies confirmation, dust, and
 /// expiry filters for transaction construction), this returns every note
 /// that existed and was unspent at the given height.
+///
+/// Height filtering uses `transactions.mined_height`, not `transactions.block`.
+/// A transaction is considered to have occurred at its mined height as soon
+/// as the wallet learns of that height (for example, from transparent UTXO
+/// retrieval), even if the containing compact block has not been fully
+/// scanned. In practice the two columns are equivalent for the notes this
+/// query can return, because `nf IS NOT NULL` and
+/// `commitment_tree_position IS NOT NULL` already require a scan of the
+/// block that contains the receiving transaction.
 pub(crate) fn get_unspent_orchard_notes_at_historical_height<P: consensus::Parameters>(
     conn: &Connection,
     params: &P,
@@ -184,14 +193,13 @@ pub(crate) fn get_unspent_orchard_notes_at_historical_height<P: consensus::Param
              rn.id AS id, t.txid, rn.action_index,
              rn.diversifier, rn.value, rn.rho, rn.rseed, rn.commitment_tree_position,
              accounts.ufvk AS ufvk, rn.recipient_key_scope,
-             t.block AS mined_height,
+             t.mined_height,
              NULL AS max_shielding_input_height
          FROM orchard_received_notes rn
          INNER JOIN accounts ON accounts.id = rn.account_id
          INNER JOIN transactions t ON t.id_tx = rn.transaction_id
          WHERE accounts.uuid = :account_uuid
-           AND t.block IS NOT NULL
-           AND t.block <= :height
+           AND t.mined_height <= :height
            AND rn.nf IS NOT NULL
            AND rn.commitment_tree_position IS NOT NULL
            AND rn.recipient_key_scope IN (0, 1)
@@ -200,8 +208,7 @@ pub(crate) fn get_unspent_orchard_notes_at_historical_height<P: consensus::Param
                SELECT rns.orchard_received_note_id
                FROM orchard_received_note_spends rns
                JOIN transactions t_spend ON t_spend.id_tx = rns.transaction_id
-               WHERE t_spend.block IS NOT NULL
-                 AND t_spend.block <= :height
+               WHERE t_spend.mined_height <= :height
            )
          ORDER BY rn.commitment_tree_position",
     )?;
--- a/zcash_client_sqlite/src/wallet/orchard.rs
+++ b/zcash_client_sqlite/src/wallet/orchard.rs
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
     )
 }
 
+/// Return all Orchard notes that were received at or before `height`
+/// and unspent as of `height`, for the given account.
+///
+/// Unlike `select_spendable_notes` (which applies confirmation, dust, and
+/// expiry filters for transaction construction), this returns every note
+/// that existed and was unspent at the given height.
+pub(crate) fn get_unspent_orchard_notes_at_historical_height<P: consensus::Parameters>(
+    conn: &Connection,
+    params: &P,
+    account: AccountUuid,
+    height: BlockHeight,
+) -> Result<Vec<ReceivedNote<ReceivedNoteId, Note>>, SqliteClientError> {
+    let mut stmt = conn.prepare_cached(
+        "SELECT
+             rn.id AS id, t.txid, rn.action_index,
+             rn.diversifier, rn.value, rn.rho, rn.rseed, rn.commitment_tree_position,
+             accounts.ufvk AS ufvk, rn.recipient_key_scope,
+             t.block AS mined_height,
+             NULL AS max_shielding_input_height
+         FROM orchard_received_notes rn
+         INNER JOIN accounts ON accounts.id = rn.account_id
+         INNER JOIN transactions t ON t.id_tx = rn.transaction_id
+         WHERE accounts.uuid = :account_uuid
+           AND t.block IS NOT NULL
+           AND t.block <= :height
+           AND rn.nf IS NOT NULL
+           AND rn.commitment_tree_position IS NOT NULL
+           AND rn.recipient_key_scope IN (0, 1)
+           AND accounts.ufvk IS NOT NULL
+           AND rn.id NOT IN (
+               SELECT rns.orchard_received_note_id
+               FROM orchard_received_note_spends rns
+               JOIN transactions t_spend ON t_spend.id_tx = rns.transaction_id
+               WHERE t_spend.block IS NOT NULL
+                 AND t_spend.block <= :height
+           )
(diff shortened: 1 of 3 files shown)


zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
...
+           AND rn.nf IS NOT NULL
+           AND rn.commitment_tree_position IS NOT NULL
+           AND rn.recipient_key_scope IN (0, 1)
+           AND accounts.ufvk IS NOT NULL
+           AND rn.id NOT IN (
+               SELECT rns.orchard_received_note_id
+               FROM orchard_received_note_spends rns
+               JOIN transactions t_spend ON t_spend.id_tx = rns.transaction_id
+               WHERE t_spend.block IS NOT NULL
+                 AND t_spend.block <= :height
```
nuttycom:
```suggestion
               WHERE t_spend.block <= :height
```
Comparisons to `NULL` always return `false`, so the` NOT NULL` check is unnecessary.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
...
+    params: &P,
+    account: AccountUuid,
+    height: BlockHeight,
+) -> Result<Vec<ReceivedNote<ReceivedNoteId, Note>>, SqliteClientError> {
+    let mut stmt = conn.prepare_cached(
+        "SELECT
+             rn.id AS id, t.txid, rn.action_index,
+             rn.diversifier, rn.value, rn.rho, rn.rseed, rn.commitment_tree_position,
+             accounts.ufvk AS ufvk, rn.recipient_key_scope,
+             t.block AS mined_height,
```
nuttycom:
```suggestion
             t.mined_height
```
There's a subtle distinction here: `t.block` requires that the associated compact block has been retrieved and scanned, whereas `t.mined_height` can be set for transactions that have been discovered by other paths (such as transactions downloaded as part of transparent UTXO graph retrieval). I'm not sure which you want here. If `t.block` is set, then `t.mined_height` will always hold the same value, but `t.block` may be NULL when `t.mined_height` is non-null. You should document your choice here, whichever way you go.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
...
+             NULL AS max_shielding_input_height
+         FROM orchard_received_notes rn
+         INNER JOIN accounts ON accounts.id = rn.account_id
+         INNER JOIN transactions t ON t.id_tx = rn.transaction_id
+         WHERE accounts.uuid = :account_uuid
+           AND t.block IS NOT NULL
```
nuttycom:
```suggestion
```

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
...
+           AND t.block <= :height
```
nuttycom:
```suggestion
           AND t.mined_height <= :height
```

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
...
+           AND rn.nf IS NOT NULL
+           AND rn.commitment_tree_position IS NOT NULL
+           AND rn.recipient_key_scope IN (0, 1)
```
nuttycom: Do not use hardcoded constants; either use a `format!(...)` directive to encode the key scope encodings (which is what we usually do in situations like this) or pass the key scopes in as a parameter.

nuttycom (changes requested):
The suggestions in this review don't necessarily require changes, but I'd like to get responses to the questions they raise before ACK'ing.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -167,6 +167,56 @@ pub(crate) fn select_spendable_orchard_notes<P: consensus::Parameters>(
     )
 }
 
+/// Return all Orchard notes that were received at or before `height`
+/// and unspent as of `height`, for the given account.
+///
+/// Unlike `select_spendable_notes` (which applies confirmation, dust, and
+/// expiry filters for transaction construction), this returns every note
+/// that existed and was unspent at the given height.
+pub(crate) fn get_unspent_orchard_notes_at_historical_height<P: consensus::Parameters>(
```
nuttycom, line 193: Another question about this method: right now, it doesn't do any checking to determine whether it's possible to construct a witness for the notes being returned. Is that a concern for this method, or is that addressed elsewhere?

zcash_client_sqlite/src/wallet/orchard.rs
p0mvn: Agreed. Dropped both IS NOT NULL guards (receive-side and spend-side) in [aa4cb3833f](https://github.com/zcash/librustzcash/pull/2284/commits/aa4cb3833f)

zcash_client_sqlite/src/wallet/orchard.rs
p0mvn: Thanks for spelling out the distinction. Went with t.mined_height in [aa4cb3833f](https://github.com/zcash/librustzcash/pull/2284/commits/aa4cb3833f) for both the receive and spend sides, and documented the choice in the rustdoc:

> Height filtering uses transactions.mined_height, not transactions.block. A transaction is considered to have occurred at its mined height as soon as the wallet learns of that height (for example, from transparent UTXO retrieval), even if the containing compact block has not been fully scanned. In practice the two columns are equivalent for the notes this query can return, because nf IS NOT NULL and commitment_tree_position IS NOT NULL already require a scan of the block that contains the receiving transaction.

The AS mined_height alias is dropped along with the swap, since the column is now selected under its native name.

zcash_client_sqlite/src/wallet/orchard.rs
p0mvn: Done in [aa4cb3833f](https://github.com/zcash/librustzcash/pull/2284/commits/aa4cb3833f). Same rationale as comment #1.

zcash_client_sqlite/src/wallet/orchard.rs
p0mvn: Applied in [aa4cb3833f](https://github.com/zcash/librustzcash/pull/2284/commits/aa4cb3833f), with the matching change on the spend-side subquery and a rustdoc paragraph (see the response on the `t.block AS mined_height` thread) documenting why `mined_height` is the right column to filter on.

zcash_client_sqlite/src/wallet/orchard.rs
p0mvn: Addressed in [aee7cb39d3](https://github.com/zcash/librustzcash/pull/2284/commits/aee7cb39d3). Switched to the `format!(...)` pattern, matching the existing usages in wallet/transparent.rs and wallet/init/migrations/transparent_gap_limit_handling.rs:

```rust
let external_scope = KeyScope::EXTERNAL.encode();
let internal_scope = KeyScope::INTERNAL.encode();

let mut stmt = conn.prepare_cached(&format!(
    "...
       AND rn.recipient_key_scope IN ({external_scope}, {internal_scope})
     ..."
))?;
```

This way the SQL filter stays in sync with `KeyScope` if its on-disk encoding ever changes.

zcash_client_sqlite/src/wallet/orchard.rs
p0mvn: You're right that this method does no witness-availability check. That's intentional. Witness construction is a separate caller concern, and I documented that in 268b7260ef:

As an example, a companion `WalletDb::generate_orchard_witnesses_at_historical_height` (in #2283) returns an actionable error for any position the wallet cannot witness at height (for example, because the wallet has not synced through height, the checkpoint was pruned, or the position does not belong to the wallet).

The motivation for keeping them separate:

- `get_unspent_orchard_notes_at_historical_height` is independently useful for audit / balance-snapshot use cases that don't need witnesses.
- By keeping the function's narrow focus, future callers may benefit from reuse.
- In terms of our use case, a pre-check here would duplicate the shard reads that the witness generator already performs.

If you'd prefer a witness pre-check baked into this method still, I'm happy to add it. Please let me know which direction you'd like.

p0mvn:
@nuttycom thanks for the review. All suggestions have been addressed in three follow-up commits.

Each inline thread has a reply with the specific commit that addresses it.

`cargo fmt, cargo clippy --all-features --all-targets -- -D warnings`, and the existing `get_unspent_orchard_notes_at_historical_height_boundary_heights` test all pass locally.

Ready for another look when you have a moment.

nuttycom (approved):
utACK 268b7260ef3564c1a71a6bab838173e3a3f3af07

PR #2289, opened 2026-04-14
feat: Mark notes as `witness_stabilized` and enable using them even after truncation

### LINKED ISSUE #2276: feat: Stabilized notes should be spendable after truncation

nullcopy:
This PR modifies truncation to respect the `PRUNING_DEPTH`, adds a `witness_stabilized` flag to notes, and modifies note selection criteria to use that flag in place of chain data, to allow using notes even during truncation.

Truncation will still reset the scan queue to the requested height, but other chain data will only be rewound as far back as the pruning depth. This ensures that any witness data necessary to spend a transaction will be available up to `tip - PRUNING_DEPTH`. Notes after the pruning depth will trivially become spendable as the wallet need only scan the most recent blocks to recover the remaining witness data up to the tip.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -4258,21 +4258,28 @@ where
     );
 }
 
-pub fn truncate_to_chain_state<T: ShieldedPoolTester, Dsf>(ds_factory: Dsf, cache: impl TestCache)
-where
+pub fn truncate_to_chain_state_deep<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
     Dsf: DataStoreFactory,
     <Dsf as DataStoreFactory>::AccountId: std::fmt::Debug,
 {
-    // Test plan:
-    // 1. Set up test environment with account
+    // Deep-truncation test plan:
+    // 1. Set up test environment with a birthday-aligned account
     // 2. Generate and scan initial blocks to populate the note commitment tree
-    // 3. Capture the chain state at a specific height
-    // 4. Generate and scan blocks beyond PRUNING_DEPTH to ensure early checkpoints are pruned
-    // 5. Verify that normal truncate_to_height fails due to missing checkpoints
+    // 3. Capture the chain state at a specific height (the future truncation target)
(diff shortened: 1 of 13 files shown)


zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,376 @@
...
+
+use schemerz_rusqlite::RusqliteMigration;
+use uuid::Uuid;
+use zcash_protocol::consensus::BlockHeight;
+
+use super::account_delete_cascade;
+use crate::wallet::init::WalletMigrationError;
+use crate::wallet::scanning::mark_stabilized_notes;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x688f7dd4_a8d4_45ba_94a3_dfc520fefed5);
```
nullcopy: Use UUID `0x0x688f7dd4_a8d4_45ba_94a3_dfc520fefed5` instead

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4776,6 +4776,159 @@ pub fn truncate_to_chain_state_above_scanned<T: ShieldedPoolTester, Dsf>(
...
+        &mut st,
+        0,
+        &[CommitmentTreeRoot::from_parts(
+            birthday,
+            T::empty_tree_leaf(),
+        )],
+    )
+    .unwrap();
+
+    // Scan `PRUNING_DEPTH` more blocks so the note is considered stabilized.
```
nullcopy: This comment is slightly innaccurate. Should be "more than" or "at least" `PRUNING_DEPTH` blocks, since the code actually does 10 blocks more

nullcopy (commented):
Comments on 771de250ccbfcca7bc65157d25017047e22b808b

zcash_client_sqlite/src/wallet/init/migrations.rs
nullcopy: Need to add our new leaf migration

nullcopy (commented):
Comments on c408572bbfbd54e5ba9d3b38564a75b1aa94d544

nullcopy (commented):
Comments on ffd22d3a5bf82d7ab4745234c456d1550ac50ccc

PR #2294, opened 2026-04-15
feat: Mark notes as stabilized and enable using them even after truncation

### LINKED ISSUE #2276: feat: Stabilized notes should be spendable after truncation

nullcopy:
This PR modifies truncation to respect the `PRUNING_DEPTH`, adds a `witness_stabilized` flag to notes, and modifies note selection criteria to use that flag in place of chain data, to allow using notes even during truncation.

Truncation will still reset the scan queue to the requested height, but other chain data will only be rewound as far back as the pruning depth. This ensures that any witness data necessary to spend a transaction will be available up to `tip - PRUNING_DEPTH`. Notes after the pruning depth will trivially become spendable as the wallet need only scan the most recent blocks to recover the remaining witness data up to the tip.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -3190,6 +3190,28 @@ pub trait WalletWrite: WalletRead {
     /// [`truncate_to_height`]: WalletWrite::truncate_to_height
     fn truncate_to_chain_state(&mut self, chain_state: ChainState) -> Result<(), Self::Error>;
 
+    /// Rewinds the wallet to at most the given block height, preserving any wallet data which has
+    /// been confirmed beyond the pruning depth.
+    ///
+    /// In contrast to [`truncate_to_height`], which unconditionally deletes wallet state above
+    /// `max_height` (transaction & note data is retained, but committment trees, blocks, etc are
+    /// removed to the truncation height), this rewinds the scan queue to `max_height` but only
+    /// rewinds blocks, note commitment trees, transactions, transparent UTXO observations, and
+    /// nullifier-map entries as far back as the pruning floor (`chain_tip - (PRUNING_DEPTH - 1)`).
+    /// Data at or below that height is preserved. Because `PRUNING_DEPTH` is a property of chain
+    /// depth, the floor is derived from the wallet's view of the chain tip rather than from
+    /// `MAX(blocks.height)`.
+    ///
+    /// The floor is clamped to an actual shard-tree checkpoint at or above the pruning floor so
+    /// that the underlying truncation has a real checkpoint to truncate to under non-contiguous
+    /// scan orders.
+    ///
+    /// Returns the actual scan-queue rewind height (`max_height` clamped to the max scanned height
+    /// when `max_height` is above it).
+    ///
+    /// [`truncate_to_height`]: WalletWrite::truncate_to_height
+    fn rewind_to_height(&mut self, max_height: BlockHeight) -> Result<BlockHeight, Self::Error>;
+
     /// Reserves the next `n` available ephemeral addresses for the given account.
     /// This cannot be undone, so as far as possible, errors associated with transaction
     /// construction should have been reported before calling this method.
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -156,8 +156,16 @@ pub const SAPLING_SHARD_HEIGHT: u8 = sapling::NOTE_COMMITMENT_TREE_DEPTH / 2;
 ///
 /// This conforms to the structure of subtree data returned by
 /// `lightwalletd` when using the `GetSubtreeRoots` GRPC call.
+///
+/// Exposed unconditionally so downstream consumers (e.g. SQL code that targets the
+/// Orchard note tables regardless of whether the `orchard` feature is enabled) can
+/// rely on a single source of truth. The `const_assert` below pins this value
+/// against `orchard::NOTE_COMMITMENT_TREE_DEPTH / 2` whenever the feature is
+/// enabled, so any drift in the Orchard crate traps at compile time.
+pub const ORCHARD_SHARD_HEIGHT: u8 = 16;
+
 #[cfg(feature = "orchard")]
-pub const ORCHARD_SHARD_HEIGHT: u8 = { orchard::NOTE_COMMITMENT_TREE_DEPTH as u8 } / 2;
+const _: () = assert!(ORCHARD_SHARD_HEIGHT == orchard::NOTE_COMMITMENT_TREE_DEPTH as u8 / 2);
 
 /// An enumeration of constraints that can be applied when querying for nullifiers for notes
 /// belonging to the wallet.
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -4258,21 +4258,28 @@ where
     );
 }
 
-pub fn truncate_to_chain_state<T: ShieldedPoolTester, Dsf>(ds_factory: Dsf, cache: impl TestCache)
-where
+pub fn truncate_to_chain_state_deep<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
     Dsf: DataStoreFactory,
     <Dsf as DataStoreFactory>::AccountId: std::fmt::Debug,
 {
-    // Test plan:
-    // 1. Set up test environment with account
+    // Deep-truncation test plan:
+    // 1. Set up test environment with a birthday-aligned account
     // 2. Generate and scan initial blocks to populate the note commitment tree
-    // 3. Capture the chain state at a specific height
-    // 4. Generate and scan blocks beyond PRUNING_DEPTH to ensure early checkpoints are pruned
-    // 5. Verify that normal truncate_to_height fails due to missing checkpoints
+    // 3. Capture the chain state at a specific height (the future truncation target)
+    // 4. Generate and scan strictly more than PRUNING_DEPTH extra blocks so that the
+    //    captured checkpoint is pruned AND the target lies below `tip - PRUNING_DEPTH`
+    //    (the "deep" branch of `truncate_to_height_internal`)
+    // 5. Verify that normal truncate_to_height fails due to the missing checkpoint
     // 6. Test that truncate_to_chain_state succeeds using the captured chain state
-    // 7. Verify wallet state after truncation
+    // 7. Verify wallet state after truncation: the scan_queue is rewound all the way
+    //    to the target, but blocks, transactions, tx_locator_map entries, and note
+    //    commitment trees are only rewound to `tip - (PRUNING_DEPTH - 1)` (the oldest
+    //    retained checkpoint)
 
-    use crate::data_api::ll::wallet::PRUNING_DEPTH;
+    use crate::data_api::{ll::wallet::PRUNING_DEPTH, scanning::ScanPriority};
 
     let mut st = TestDsl::with_sapling_birthday_account(ds_factory, cache).build::<T>();
 
@@ -4283,7 +4290,7 @@ where
 
     // Step 2: Generate and scan initial blocks to populate the note commitment tree.
     // We use an "other" fvk so that notes won't be tracked by the wallet (keeping the
-    // test focused on tree state rather than wallet balances).
+    // test focused on tree/block state rather than wallet balances).
     let seed = [1u8; 32];
     let other_sk = T::sk(&seed);
     let other_fvk = T::sk_to_fvk(&other_sk);
@@ -4309,9 +4316,154 @@ where
         .clone();
     assert_eq!(captured_chain_state.block_height(), capture_height);
 
-    // Step 4: Generate and scan blocks well beyond PRUNING_DEPTH so that the checkpoint
-    // at capture_height is pruned from the note commitment tree.
-    let extra_blocks = PRUNING_DEPTH + 10;
+    // Step 4: Generate and scan strictly more than PRUNING_DEPTH extra blocks so that
+    // the checkpoint at capture_height is pruned from the note commitment tree AND
+    // capture_height is below `tip - PRUNING_DEPTH`, which exercises the deep branch
+    // of `truncate_to_height_internal`.
+    let extra_blocks = PRUNING_DEPTH + 1;
+    for _ in 0..extra_blocks {
+        st.generate_next_block(
+            &other_fvk,
+            AddressType::DefaultExternal,
+            Zatoshis::const_from_u64(5000),
+        );
+    }
+    st.scan_cached_blocks(capture_height + 1, extra_blocks as usize);
+
+    let pre_truncation_tip = st
+        .wallet()
+        .chain_height()
+        .unwrap()
+        .expect("chain tip should be set");
+    assert!(
+        pre_truncation_tip > capture_height + PRUNING_DEPTH,
+        "tip should be strictly beyond pruning depth from capture height"
+    );
+
+    // Step 5: Verify that truncate_to_height fails at capture_height because the
+    // checkpoint has been pruned.
+    let truncation_result = st.wallet_mut().truncate_to_height(capture_height);
+    assert!(
+        truncation_result.is_err(),
+        "truncate_to_height should fail when checkpoint has been pruned"
+    );
+
+    // Step 6: truncate_to_chain_state should succeed. For a deep target it skips frontier
+    // insertion entirely and relies on the existing tree checkpoints above the target.
+    st.wallet_mut()
+        .truncate_to_chain_state(captured_chain_state.clone())
+        .expect("truncate_to_chain_state should succeed");
+
+    // Step 7: Verify wallet state after truncation.
+    // 7a. The chain tip (derived from scan_queue) should now report the truncation target.
+    let new_tip = st
+        .wallet()
+        .chain_height()
+        .unwrap()
+        .expect("chain tip should match truncation height");
+    assert_eq!(new_tip, capture_height);
+
+    // 7b. No `Scanned` range in the scan queue may extend above the truncation target.
+    let ranges = st
+        .wallet()
+        .suggest_scan_ranges()
+        .expect("suggest_scan_ranges should succeed");
+    let future_scanned = ranges
+        .iter()
+        .find(|r| r.priority() == ScanPriority::Scanned && r.block_range().end > capture_height);
+    assert!(
+        future_scanned.is_none(),
+        "no `Scanned` range should extend above the truncation target, found: {future_scanned:?}"
+    );
+
+    // 7c. A deep truncation rewinds the scan queue all the way to the target, but blocks,
+    // transactions, tx_locator_map entries, and note commitment trees are only rewound to the
+    // oldest retained checkpoint at `tip - (PRUNING_DEPTH - 1)`. Data at that boundary is kept
+    // (so stabilized notes remain spendable); data above it is removed.
+    let prune_boundary = pre_truncation_tip - (PRUNING_DEPTH - 1);
+    let wallet = st.wallet();
+    assert!(
+        wallet.get_block_hash(prune_boundary).unwrap().is_some(),
+        "block entry at (tip - (PRUNING_DEPTH - 1)) should be preserved by a deep truncation"
+    );
+    assert!(
+        wallet.get_block_hash(prune_boundary + 1).unwrap().is_none(),
+        "block entries above (tip - (PRUNING_DEPTH - 1)) must be removed by a deep truncation"
+    );
+    assert!(
+        wallet.get_block_hash(pre_truncation_tip).unwrap().is_none(),
+        "block entries up to the pre-truncation tip must be removed by a deep truncation"
+    );
+    assert_eq!(
+        wallet
+            .block_max_scanned()
+            .unwrap()
+            .map(|m| m.block_height()),
+        Some(prune_boundary),
+        "block_max_scanned should equal (tip - (PRUNING_DEPTH - 1)) after a deep truncation"
+    );
+}
+
+pub fn truncate_to_chain_state_shallow<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
+    Dsf: DataStoreFactory,
+    <Dsf as DataStoreFactory>::AccountId: std::fmt::Debug,
+{
+    // Shallow-truncation test plan:
+    // 1. Set up test environment with a birthday-aligned account
+    // 2. Generate and scan initial blocks to populate the note commitment tree
+    // 3. Capture the chain state at a specific height (the future truncation target)
+    // 4. Generate and scan `PRUNING_DEPTH - 1` extra blocks so that the target sits at the
+    //    shallow boundary (`target == tip - (PRUNING_DEPTH - 1)`, the oldest retained
+    //    checkpoint)
+    // 5. Test that truncate_to_chain_state succeeds using the captured chain state
+    // 6. Verify wallet state after truncation: scan_queue, blocks, tx_locator_map and
+    //    note commitment trees are all rewound to the truncation target. Data at the
+    //    target is preserved; anything above is removed.
+
+    use crate::data_api::{ll::wallet::PRUNING_DEPTH, scanning::ScanPriority};
+
+    let mut st = TestDsl::with_sapling_birthday_account(ds_factory, cache).build::<T>();
+
+    let sapling_activation = st
+        .network()
+        .activation_height(consensus::NetworkUpgrade::Sapling)
+        .unwrap();
+
+    // Step 2: Generate and scan initial blocks to populate the note commitment tree.
+    // We use an "other" fvk so that notes won't be tracked by the wallet (keeping the
+    // test focused on tree/block state rather than wallet balances).
+    let seed = [1u8; 32];
+    let other_sk = T::sk(&seed);
+    let other_fvk = T::sk_to_fvk(&other_sk);
+
+    let initial_block_count = 8u32;
+    for _ in 0..initial_block_count {
+        st.generate_next_block(
+            &other_fvk,
+            AddressType::DefaultExternal,
+            Zatoshis::const_from_u64(10000),
+        );
+    }
+    let scan_start = sapling_activation;
+    st.scan_cached_blocks(scan_start, initial_block_count as usize);
+
+    // Step 3: Capture the chain state at the current tip. The CachedBlock tracks the
+    // exact frontier that corresponds to the end of each generated block.
+    let capture_height = sapling_activation + initial_block_count - 1;
+    let captured_chain_state = st
+        .latest_cached_block()
+        .expect("should have cached blocks")
+        .chain_state()
+        .clone();
+    assert_eq!(captured_chain_state.block_height(), capture_height);
+
+    // Step 4: Scan `PRUNING_DEPTH - 1` extra blocks so that the truncation target sits at the
+    // shallow boundary (`target == tip - (PRUNING_DEPTH - 1)`, which is exactly the oldest
+    // retained checkpoint given the tree's `max_checkpoints = PRUNING_DEPTH`).
+    let extra_blocks = PRUNING_DEPTH - 1;
     for _ in 0..extra_blocks {
         st.generate_next_block(
             &other_fvk,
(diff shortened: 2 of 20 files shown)


zcash_client_sqlite/CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ workspace.
 - `impl<'conn, P, CL, R> WalletWrite for WalletDb<SqlTransaction<'conn>, P, CL, R>` to
   enable calling `WalletWrite` methods inside `WalletDb::transactionally` (amortizing the
   database transaction overhead).
+- `zcash_client_sqlite::wallet::commitment_tree::min_checkpoint_id_at_or_above`
```
nuttycom, line 33: Does this need to be a public function?

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -48,6 +52,17 @@ workspace.
   `zcash_client_sqlite::error::StandaloneImportConflict`
 - P2SH UTXOs returned by `get_spendable_transparent_outputs` now include a
   precomputed input size for accurate ZIP 317 fee estimation.
+- Truncation no longer unconditionally rewinds all wallet data to the requested
```
nuttycom, line 55: I've been thinking about this more. It occurs to me that there are potentially two separate use cases, with different purposes: there's "rewind", which is what we want to do ordinarily, and there's "truncate" which behaves as `truncate_to_chain_state` does before this PR. Truncation *should* never be necessary, but suppose that we somehow end up with note commitment tree corruption? Maybe we should separate these use cases out into two separate methods, or one `rewind_to_chain_state` method with two modes, `Rewind` and `Truncate`?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3634,19 +3684,41 @@ pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
...
 
-    // Only truncate trees when the maximum scanned height is greater than the target height. When
-    // the target height is at or above the max scanned height, we skip frontier insertion (it is
-    // unnecessary at the max scanned height, and could introduce a subtree root discontinuity
-    // above it; the frontier will be added naturally during scanning). We will however still need
-    // to truncate the scan queue so that ranges above the target are removed.
-    let truncate_trees = block_max_scanned(wdb.conn.0, &wdb.params)?
-        .is_some_and(|meta| meta.block_height() > target_height);
+    // Only truncate trees when the target is strictly below the maximum scanned height AND
+    // inside the prune window `[max_scanned - (PRUNING_DEPTH - 1), max_scanned]`. Targets at or
```
nuttycom, line 3710:
```suggestion
   // inside the pruning window `(max_scanned - (PRUNING_DEPTH - 1))..=max_scanned`. Targets at or
```
nit: make inclusion/exclusion explicit

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3634,19 +3684,41 @@ pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
 ///   oldest checkpoint to ensure that the a checkpoint added at the provided frontier position does
 ///   not get immediately pruned, then inserts the provided frontier as a new checkpoint at the
 ///   target height, and finally truncates to that new checkpoint.
+///
+/// # Spending stabilized notes after a deep truncation
+///
+/// A deep truncation rewinds the scan queue to `target_height` but preserves blocks, trees,
+/// transactions, and other wallet data above the lowest shard-tree checkpoint inside the
```
nuttycom, line 3691: The "above the lowest shard-tree checkpoint" language here is a little confusing; I know what is meant here, but the use of "checkpoints" in the shard tree is an implementation detail. I guess that this is fine since this is pub(crate), but it took me a moment to understand what was meant here.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4339,34 +4348,303 @@ where
...
+        .wallet()
+        .suggest_scan_ranges()
+        .expect("suggest_scan_ranges should succeed");
+    let future_scanned = ranges
+        .iter()
+        .find(|r| r.priority() == ScanPriority::Scanned && r.block_range().end > capture_height);
+    assert!(
+        future_scanned.is_none(),
+        "no `Scanned` range should extend above the truncation target, found: {future_scanned:?}"
+    );
```
nuttycom, line 4377: I would expect that the maximum extent of the scan queue would be just up to new_tip; that would be a stronger check than this one. Why would we expect scan ranges that extend past the chain tip?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4339,34 +4348,303 @@ where
...
+
+    // 7c. A deep truncation rewinds the scan queue all the way to the target, but blocks,
+    // transactions, tx_locator_map entries, and note commitment trees are only rewound to the
+    // oldest retained checkpoint at `tip - (PRUNING_DEPTH - 1)`. Data at that boundary is kept
+    // (so stabilized notes remain spendable); data above it is removed.
+    let prune_boundary = pre_truncation_tip - (PRUNING_DEPTH - 1);
+    let wallet = st.wallet();
+    assert!(
+        wallet.get_block_hash(prune_boundary).unwrap().is_some(),
```
nuttycom, line 4386: Instead of checking `is_some`, it should be checked against the expected value.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4339,34 +4348,303 @@ where
...
+    assert_eq!(new_tip, capture_height);
+
+    // 6b. No `Scanned` range in the scan queue may extend above the truncation target.
+    let ranges = st
+        .wallet()
+        .suggest_scan_ranges()
+        .expect("suggest_scan_ranges should succeed");
+    let future_scanned = ranges
+        .iter()
+        .find(|r| r.priority() == ScanPriority::Scanned && r.block_range().end > capture_height);
```
nuttycom, line 4510: Same comment here as above; this should should be able to ignore priority. If there are ranges with the end above the capture height, I want to understand what they are.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4339,34 +4348,303 @@ where
...
+    );
+}
+
+/// Regression test: after the scan scheduler processes a `ChainTip` range before a lower
+/// `Historic` range, `MAX(height) FROM blocks` points into one scanned region while
+/// `last_scanned - (PRUNING_DEPTH - 1)` lands inside the unscanned gap between the two
+/// regions. `truncate_to_chain_state` must still succeed: the previous implementation
+/// returned `CorruptedData` because it expected a checkpoint at exactly the PD floor; the
+/// current implementation clamps forward to the lowest checkpoint that still lies inside
+/// the prune window.
```
nuttycom, line 4544: The comment about the "previous implementation" here doesn't make a lot of sense - this will certainly degrade over time. Which previous implementation is this talking about?

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,402 @@
...
+        // stabilization path share a single source of truth. We restrict the subquery to
+        // ranges with priority `Scanned`, because `scan_queue` also contains unscanned
+        // ranges (historic / found-note / chain-tip / verify) whose `block_range_end`
+        // would otherwise pollute the max. If the wallet has no scanned ranges yet, the
+        // subquery returns `NULL` and no rows are backfilled, which is the correct
+        // behavior for a freshly-created wallet.
+        let last_scanned_height: Option<u32> = transaction.query_row(
+            &format!(
+                "SELECT MAX(block_range_end) - 1 FROM scan_queue WHERE priority = {}",
+                priority_code(&ScanPriority::Scanned),
```
nuttycom, line 72: I'm pretty sure this has a bug: this is the maximum end of a scanned range, but if there are unscanned ranges below this height, they'll be ignored in determining whether those notes are stabilized.

If you use `zcash_client_sqlite::wallet::block_fully_scanned` I think that will give you what you want instead?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +369,57 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+    mark_stabilized_notes(conn, last_scanned_height)?;
+
+    Ok(())
+}
+
+/// Marks notes as stabilized when their containing shard is complete and the shard's end
+/// height lies at or below the truncation rewind boundary
+/// (`last_scanned_height - (PRUNING_DEPTH - 1)`). Once stabilized, a note's witness data is
+/// considered durable and will be preserved across truncation.
+pub(crate) fn mark_stabilized_notes(
```
nuttycom, line 383: I think that I must be missing something here; how does this determine that the shard has been fully scanned?

nuttycom (changes requested):
Overall this is looking good, but I'm confused about some things with respect to how stabilization is being determined.

In the migration, the safest way to determine witness completeness would simply be to attempt to construct the witness for each note in the wallet at the oldest anchor available. If the witness is constructable and the shard's maximum extent is confirmed beyond the pruning depth, then (and only then) can you mark it stable.

zcash_client_sqlite/src/wallet/scanning.rs
nullcopy: The intent was to only prune shards whose `subtree_end_height` is not NULL and has been confirmed to the stable depth, because I was (wrongly) under the impression that `subtree_end_height` would only be populated once a shard had been filled. This needs to be fixed

zcash_client_sqlite/CHANGELOG.md
nullcopy: Good point. No. I'll remove this and make it private

zcash_client_sqlite/CHANGELOG.md
nullcopy: Yes, I believe we should leave truncate as is and create a new `rewind_to_chain_state` method. I'll refactor it this

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Something to think through that might work: a query that verifies that there are no gaps in the `blocks` table for the extent of the shard could give us a good proxy for "fully scanned". I'm not 100% confident that this is the right idea, but it's a direction to explore.

zcash_client_backend/src/data_api/testing/pool.rs
nullcopy: You're right. I added this test case in a past iteration (because of a bug I was chasing), but its completely useless here. I'll remove 7b entirely

zcash_client_backend/src/data_api/testing/pool.rs
nullcopy: Good idea. Will do 👍

zcash_client_sqlite/src/wallet.rs
nullcopy: Agree this could be confusing. I'm removing this to revert `truncate_to_chain_state()` to its previous behavior, but I'll take a stab at phrasing this better for the new `rewind_to_chain_state()` method

zcash_client_sqlite/src/wallet/scanning.rs
nullcopy: That would be covered by `block_fully_scanned()`, right? `block_fully_scanned()` would be a more efficient query, and the only case where it would be weaker than your proposal, would be in the scenario where ANY unscanned range exists BEFORE the shard in question. But, that should be a transient scenario that I don't think is practically meaningful here?

nullcopy:
[Force-pushed](https://github.com/zcash/librustzcash/compare/a22ffd8bf70fe864c45a35a3101cee5e5b3903e9..f4155576786db6592e4c49e62fb1a849f1156c4b) to address feedback from @nuttycom

Given the significance of the refactor, IMO it's easier to review the commits again, one by one

nullcopy:
Force-pushed to rebase on main and use `corez` replacement for yanked `core2`

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3744,6 +3744,98 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
...
+        crate::ORCHARD_TABLES_PREFIX,
+        truncation_target,
+    )
+    .map_err(ShardTreeError::Storage)?;
+    #[cfg(not(feature = "orchard"))]
+    let orchard_window_floor: Option<BlockHeight> = None;
+
+    // Combine the per-pool floors by taking the shallower (larger height).
+    let truncation_height = match (sapling_window_floor, orchard_window_floor) {
+        (Some(s), Some(o)) => s.max(o),
```
nuttycom: These should always be equal; we should probably return `SqliteClientError::CorruptedData` if they aren't.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3744,6 +3744,98 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
...
+        params,
+        #[cfg(feature = "transparent-inputs")]
+        gap_limits,
+        truncation_height,
+    )?;
+
+    // When the caller asked to rewind below the truncation_height floor, trim the scan queue the
+    // rest of the way down to `max_height` so the sync loop will re-scan the blocks between
+    // `max_height` and `data_rewind_height`.
+    if max_height < truncation_height {
```
nuttycom: It would be useful to factor the body of this out into a function, nonblocking.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +369,96 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+    conn: &rusqlite::Transaction<'_>,
+) -> Result<(), SqliteClientError> {
+    let Some(birthday_height) = wallet_birthday(conn)? else {
+        return Ok(());
+    };
+
+    // The fully-scanned height is the last height in the first `Scanned` range of
+    // `scan_queue`, provided that range starts at or before the wallet birthday.
+    // `block_range_end` is exclusive.
+    let fully_scanned_height = match conn
```
nuttycom: This reimplements block_fully_scanned, but also I think has a bug because a scanned range can exist at heights above unscanned ranges (this is common).

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +369,96 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+                Ok((start, end))
+            },
+        )
+        .optional()?
+    {
+        Some((start, end)) if start <= birthday_height => end - 1,
+        _ => return Ok(()),
+    };
+
+    let Some(max_scanned_height): Option<BlockHeight> = conn
```
nuttycom: block_max_scanned should be used here.

zcash_client_sqlite/src/wallet/common.rs
nuttycom:
```suggestion
/// - The `to_received_note` function is expected to return `Ok(None)` in the case that spending
```

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -457,25 +460,33 @@ where
...
-                    .iter()
-                    .any(|p| *p <= ScanPriority::Scanned);
+            (
+                Some(note),
+                witness_stabilized,
+                max_shard_priority,
+                tx_trusted,
+                tx_shielding_inputs_trusted,
+            ) => {
+                let shard_scanned = witness_stabilized
```
nuttycom: `shard_scanned` is no longer a good name for this variable; I would call it `shard_witness_available` or something of the sort - either the witness is stabilized, or the witness is not yet stable but the shard has been fully scanned.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -523,9 +534,13 @@ where
...
-    if unscanned_tip_exists(conn, anchor_height, table_prefix)? {
-        return Ok(vec![]);
-    }
+
+    // When the anchor's shard has unscanned ranges at or below `anchor_height`, the tree
+    // frontier at the anchor is not reliably reconstructible, so un-stabilized notes can't
+    // be safely selected. Stabilized notes are unaffected: their witness data is durable
+    // in the shard tree regardless of what `scan_queue` says. We pass `tip_unscanned` as a
+    // parameter and let the SQL mix the two cases, rather than early-exiting.
+    let tip_unscanned = unscanned_tip_exists(conn, anchor_height, table_prefix)?;
```
nuttycom: See https://github.com/zcash/librustzcash/issues/982#issuecomment-4282764862 - I think that the "Stabilized notes are unaffected" piece of this comment is pointing to the wrong thing. For stabilized notes, we can compute a witness so long as we have downloaded the tree state corresponding to the frontier block, which we should always have done anyway as part of chain tip scanning. The only thing that the "tip_unscanned" determination might be useful for is figuring out whether we can spend a note that is *within* the chain-tip shard.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -571,26 +587,39 @@ where
...
+             -- unscanned ranges (`:tip_unscanned = 1`) or its containing shard is not
+             -- yet fully scanned. Non-stabilized notes require a fully-scanned containing
+             -- shard AND a fully-scanned anchor shard. The `scan_state.max_priority IS
+             -- NULL` case (which arises when `rn.commitment_tree_position IS NULL`) is
+             -- excluded by the priority check.
+             AND (
+                 rn.witness_stabilized = 1
+                 OR (
+                     :tip_unscanned = 0
+                     AND scan_state.max_priority <= :scanned_priority
```
nuttycom:
```suggestion
                     :tip_unscanned = 0 -- the tip shard has no unscanned ranges
                     AND scan_state.max_priority <= :scanned_priority -- the note shard is fully scanned or ignored
```

y4ssi:
@zodl-review

y4ssi:
@zodl-review

nullcopy:
~~Force-pushed a moderate refactor: Instead of marking notes as stabilized according to shard scan state and confirmation, I now mark shards as stabilized according to their scan state and confirmation. Additionally, I fixed the stabilization criteria to ensure no unscanned ranges exist in the extent of a given shard, before it can be considered stabilized.~~

~~The modifications to note selection still use note stabilization in their selection criteria, but note stabilization is now read from the stabilization status of the note's containing shard~~

~~I recommend re-reviewing commit by commit, but for those who are interested, the diff is [here](https://github.com/zcash/librustzcash/compare/e5c32e060d5318391224d6816ea58c336c7af147..e0e66e1bf6ddfaa37023cabeaae4ebb931bbcacd)~~

[edit] I abandoned this refactor, as shard-stabilization is not sensible. Keep scrolling :arrow_down:

y4ssi:
@zodl-review

nullcopy:
Force-pushed to address PR feedback:
* checks for any unscanned range intersecting the shard's extent, before marking notes as stabilized
* clarified several comments
* added a check to ensure orchard and sapling pool always arrive at the same checkpoint
* gate orchard updates so that `ORCHARD_SHARD_HEIGHT` doesn't need to be unconditionally exported anymore
* refactored end-to-end test to build fully populated shards, instead of just simulating shard completion
* add test to ensure stabilization of notes discovered during rescan
* use `block_max_scanned()` in place of bespoke SQL to query max scanned height

[full diff since last review](https://github.com/zcash/librustzcash/compare/e5c32e060d5318391224d6816ea58c336c7af147..5abbdf828213e4a2fb22f834bf10136eb6e2fcf9)

zcash_client_backend/src/data_api.rs
```diff
@@ -3190,6 +3190,26 @@ pub trait WalletWrite: WalletRead {
     /// [`truncate_to_height`]: WalletWrite::truncate_to_height
     fn truncate_to_chain_state(&mut self, chain_state: ChainState) -> Result<(), Self::Error>;
 
+    /// Rewinds the wallet to at most the given block height, preserving any wallet data which has
+    /// been confirmed beyond the pruning depth.
+    ///
+    /// In contrast to [`truncate_to_height`], which unconditionally deletes all wallet state above
```
nuttycom: It's not quite right to say that `truncate_to_height` unconditionally deletes wallet state, because that could be misinterpreted to indicate that transaction data is also deleted, which is not the case.

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,644 @@
...
+        transaction.execute_batch(
+            "ALTER TABLE sapling_received_notes
+               ADD COLUMN witness_stabilized INTEGER NOT NULL DEFAULT 0;
+
+             ALTER TABLE orchard_received_notes
+               ADD COLUMN witness_stabilized INTEGER NOT NULL DEFAULT 0;",
+        )?;
+
+        // Backfill: Identify any notes which have stable witness data, and mark them as such.
+        mark_stabilized_notes(transaction, &self.params)?;
```
nuttycom, line 58: This still looks like the wrong API to me, if this method is intended to be called at all outside of this migration. We can't do a global search for notes that have been stabilized in any sort of performant fashion; if we were able to do so, we wouldn't need the stabilized flag to begin with. We have an opportunity at this migration to specifically mark notes stabilized based on the state of the scan queue, and we should do so, but after this migration we'll never be able to do that again.

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -0,0 +1,644 @@
...
+impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), WalletMigrationError> {
+        transaction.execute_batch(
+            "ALTER TABLE sapling_received_notes
+               ADD COLUMN witness_stabilized INTEGER NOT NULL DEFAULT 0;
+
+             ALTER TABLE orchard_received_notes
+               ADD COLUMN witness_stabilized INTEGER NOT NULL DEFAULT 0;",
```
nuttycom: These columns should be indexed if they're important to querying.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +370,66 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+    params: &P,
+) -> Result<(), SqliteClientError> {
+    let Some(max_scanned_height) = block_max_scanned(conn, params)?.map(|m| m.block_height())
+    else {
+        return Ok(());
+    };
+
+    let pruning_floor: u32 = u32::from(max_scanned_height).saturating_sub(PRUNING_DEPTH - 1);
+
+    conn.execute(
```
nuttycom: We should document that we can never stabilize the witness for a note within the chain-tip shard, because the anchor height is always within the unstable range.

nuttycom (commented):
Flushing comments; some of these may be obsolete given the changes I haven't had time to re-review yet.

zcash_client_sqlite/src/wallet/scanning.rs
nullcopy: This has been removed

zcash_client_sqlite/src/wallet/common.rs
nullcopy: I've fixed the comment for today, and also responded on the issue (See https://github.com/zcash/librustzcash/issues/982#issuecomment-4299270935) with improvements to the stabilization approach in this PR, which IMO can come as a future enhancement.

Do you agree?

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
nullcopy: IIRC, our conclusion from pairing was that the updates are efficient, because the query filters out any note which has previously been stabilized (as long as we index the stabilization column).

I plan to enable indexing on that column, which I believe alleviates your concern.

I could also update this method to restrict its query to the block range corresponding to the affected shard, and add a loop over all note-containing shards here in the migration. I think that would technically be less efficient, though... Consider the steady-state behavior while scanning:
1) First, the wallet is likely to have zero unstabilized note to consider, so the existing query will efficiently short circuit on `WHERE witness_stabilized = 0`.
2) Second, if unstabilized notes exist, they USUALLY (except during re-scan or catching up after being offline for several shards) will be in a recently completed shard. Filtering for a block range will be a heavier query than the existing `WHERE witness_stabilized = 0`.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -457,25 +460,33 @@ where
...
+                Some(note),
+                witness_stabilized,
+                max_shard_priority,
+                tx_trusted,
+                tx_shielding_inputs_trusted,
+            ) => {
+                let shard_scanned = witness_stabilized
+                    || max_shard_priority
+                        .iter()
+                        .any(|p| *p <= ScanPriority::Scanned);
```
nuttycom: BTW, we can now use 
```suggestion
                    || max_shard_priority.is_some_and(|p| *p <= ScanPriority::Scanned);
```

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ workspace.
 - `impl<'conn, P, CL, R> WalletWrite for WalletDb<SqlTransaction<'conn>, P, CL, R>` to
   enable calling `WalletWrite` methods inside `WalletDb::transactionally` (amortizing the
   database transaction overhead).
+- `WalletWrite::rewind_to_height` respects the pruning window: it rewinds the
+  scan queue to the requested target but only rewinds blocks, note commitment
+  trees, transactions, transparent UTXO observations, and nullifier-map entries
+  as far back as the lowest shard-tree checkpoint inside the pruning window.
```
nuttycom: The addition of `rewind_to_height` to `WalletWrite` needs to be documented in the `Changed` section; it's an API-breaking change to a public trait, not an addition.

nuttycom: That being said, it actually shouldn't be in the `zcash_client_sqlite` changelog at all; this will be covered under a single "Migrated to `zcash_client_backend 0.22.0` entry."

zcash_client_sqlite/CHANGELOG.md
```diff
@@ -48,6 +52,17 @@ workspace.
...
+  `orchard_received_notes` tables. The column is set to 1 at the end of each
+  scan batch (and once as a backfill by the `witness_stabilized_notes`
+  migration) for notes whose containing shard is fully Scanned and whose
+  `subtree_end_height` has received at least `PRUNING_DEPTH` confirmations.
+- Note-selection (`select_unspent_notes`,
+  `select_spendable_notes_matching_value`) and balance reporting
+  (`get_wallet_summary`) now consult the `witness_stabilized` flag in addition to
+  the existing scan state. A stabilized note bypasses the shard-scanned,
+  unscanned-tip, and confirmations checks, so stabilized notes remain spendable
+  after a deep `rewind_to_height` rewinds `scan_queue` below them.
```
nuttycom: This is an internal implementation detail, I'm not sure it belongs in the CHANGELOG.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3744,6 +3753,98 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
...
+/// The floor is clamped to an actual shard-tree checkpoint at or above the pruning floor (via
+/// [`commitment_tree::min_checkpoint_id_at_or_above`]) so that [`truncate_to_height_internal`] has
+/// a real checkpoint to truncate to under non-contiguous scan orders.
+///
+/// Returns the actual scan-queue rewind height (`max_height` clamped to the max scanned height when
+/// `max_height` is above it).
+pub(crate) fn rewind_to_height<P: consensus::Parameters>(
+    conn: &rusqlite::Transaction,
+    params: &P,
+    #[cfg(feature = "transparent-inputs")] gap_limits: &GapLimits,
```
nuttycom, line 3767: nonblocking: this API requiring gap limits is a wart; the reason that truncate_to_height_internal requires it is that we're reconstructing a `WalletDb` instance in order to reuse methods from the `ShardTree` API. It would be good to disentangle this. I may make follow-up PR that does this refactoring; it's not worth addressing in this PR.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3744,6 +3753,98 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
     Ok(())
 }
 
+/// Rewinds the wallet to at most the given block height, preserving any wallet data which has been
+/// confirmed beyond the pruning depth.
+///
+/// In contrast to [`truncate_to_height`], which unconditionally deletes all wallet state above
```
nuttycom: nit: this duplicates the subtly-incorrect "unconditionally deletes all wallet state" language that I objected to in `data_api.rs

zcash_client_sqlite/CHANGELOG.md
nullcopy: Removed

zcash_client_sqlite/CHANGELOG.md
nullcopy: Agreed. Removed.

zcash_client_sqlite/src/wallet.rs
nullcopy: Fixed

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
nuttycom: Agreed, with the indexing, this is fine.

nullcopy:
[Force-pushed](https://github.com/zcash/librustzcash/compare/5abbdf828213e4a2fb22f834bf10136eb6e2fcf9..ee69897e625ce39f18b4a1cf1d209910067899f9) to address review feedback

nullcopy:
[Force-pushed](https://github.com/zcash/librustzcash/compare/ee69897e625ce39f18b4a1cf1d209910067899f9..7fe3757ce7708ec6224c16ad8a9f38f8065ab7b4) to fix off-by-one error in new `trim_scan_queue_to()` causing truncation failures

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -370,6 +370,58 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+        Ok(())
+    }
+
+    if let Some(max_scanned_height) = block_max_scanned(conn, params)?.map(|m| m.block_height()) {
+        let pruning_floor: u32 = u32::from(max_scanned_height).saturating_sub(PRUNING_DEPTH - 1);
+
+        // Mark stabilized notes in each pool
+        mark_pool(conn, "sapling", SAPLING_SHARD_HEIGHT, pruning_floor)?;
+        #[cfg(feature = "orchard")]
+        mark_pool(conn, "orchard", ORCHARD_SHARD_HEIGHT, pruning_floor)?;
```
nuttycom, line 422: nonblocking: we have constants for "sapling" and "orchard"

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4593,6 +4603,875 @@ pub fn truncate_to_chain_state_above_scanned<T: ShieldedPoolTester, Dsf>(
...
+    let a_positions: [u32; 3] = [
+        2 * SHARD_POSITIONS,                       // first slot of shard 2
+        2 * SHARD_POSITIONS + SHARD_POSITIONS / 2, // middle slot of shard 2
+        3 * SHARD_POSITIONS - 1,                   // last slot of shard 2
+    ];
+    let b_positions: [u32; 3] = [
+        2 * SHARD_POSITIONS + 1,                       // one after A's first
+        2 * SHARD_POSITIONS + SHARD_POSITIONS / 2 + 1, // one after A's middle
+        3 * SHARD_POSITIONS - 2,                       // one before A's last
+    ];
```
nuttycom, line 5351: This is a weird test setup, for the reason that the scenario described is impossible; you can't have a block that fills a shard.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4593,6 +4603,875 @@ pub fn truncate_to_chain_state_above_scanned<T: ShieldedPoolTester, Dsf>(
...
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, filler_value);
+    }
+
+    // Step 1d: scan the note block, declare shard 2 complete, scan the
+    // filler blocks. The batch ordering mirrors
+    // `stabilized_note_spendable_after_deep_rewind`: shard 2's root can
+    // only be computed after its leaves are in the wallet's tree, and
+    // `put_subtree_roots` must run before the next scan batch so
+    // `mark_stabilized_notes` sees shard 2's `subtree_end_height`.
+    st.scan_cached_blocks(note_height, 1);
```
nuttycom, line 5400: I'm surprised that the performance of this is acceptable for a test. Computers are fast, I guess.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4593,6 +4603,875 @@ pub fn truncate_to_chain_state_above_scanned<T: ShieldedPoolTester, Dsf>(
...
+    // Step 3: import account B, sharing A's birthday so `add_account`
+    // rewrites the post-birthday `Scanned` range to `Historic` (forcing
+    // a re-scan).
+    let b_birthday = account_a.birthday().clone();
+    let seed = Secret::new(TEST_SEED.to_vec());
+    let (account_b, _usk_b) = st
+        .wallet_mut()
+        .import_account_hd("account B", &seed, zip32_index_b, &b_birthday, None)
+        .expect("account B import should succeed");
+
```
nuttycom, line 5429: This reminds me of follow-up that is needed: account insertion should now be calling `rewind_to_height`.

nuttycom (approved):
utACK 7fe3757ce7708ec6224c16ad8a9f38f8065ab7b4. Some of the test setup looks like it doesn't reflect realities of how the chain functions, but I am not able to say that that impairs the test and the implementation all looks reasonable to me.

PR #2303, opened 2026-04-17
zcash_client_backend: Add scanning::ScanningKey::new

### LINKED ISSUE #2297: Support for scanning with only an IVK
A `ScanningKey` apparently makes the nullifier key optional:

https://github.com/zcash/librustzcash/blob/f646a1119d5e0214d02fb1575e662c6b0ec57c8d/zcash_client_backend/src/scanning.rs#L128-L133

That's great, because it looks like scanning with only an IVK is a supported thing.

Except that the struct has private fields, and the only function I can find that creates these structs requires an FVK:

https://github.com/zcash/librustzcash/blob/f646a1119d5e0214d02fb1575e662c6b0ec57c8d/zcash_client_backend/src/scanning.rs#L267-L274

It seems like adding an IVK->scanning key function might be relatively simple and perhaps the last step in support IVK-based scanning.

nuttycom:
Fixes #2297

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -33,6 +33,7 @@ workspace.
   - `LightdInfo` has added fields `upgrade_name`, `upgrade_height`, and `lighwallet_protocol_version`
   - `GetMempoolTxRequest` (previously named `Exclude`) has added field `pool_types`
 - `zcash_client_backend::fees::MetaSource`
+- `zcash_client_backend::scanning::ScanningKey::new`:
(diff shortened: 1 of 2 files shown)


nullcopy (approved):
utACK ef07d9c722fb3cdc12393c57ad16ae6b2763a9f8

nullcopy (approved):
utACK fd4076196cae04ad4bd09e9487a3df985fe951a1

ValarDragon (commented):
Lgtm, simple new api here

PR #2305, opened 2026-04-17
feat: WalletWrite::mark_transparent_addresses_exposed

### LINKED ISSUE #2304: feat(zcash_client_backend): WalletWrite method to report address exposure
It is possible for a wallet to know of an address's exposure for reasons outside of the purview of librustzcash. In those cases, the wallet needs a mechanism to inform the backend that an address has been exposed.

nullcopy:
Allows wallet callers to inform the backend that a transparent address has been exposed externally, for reasons outside the purview of librustzcash. The SQLite implementation updates `exposed_at_height` to the lesser of the existing value and the provided height, and returns `AddressNotRecognized` when the address is not tracked by the wallet.

Resolves: https://github.com/zcash/librustzcash/issues/2304
Resolves: [COR-1200](https://linear.app/zodl/issue/COR-1200/featzcash-client-backend-walletwrite-method-to-report-address-exposure)

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -46,6 +46,12 @@ workspace.
     most likely has not synced through that height).
   Shard-read failures continue to surface via the existing
   `SqliteClientError::CommitmentTree` variant.
+- `impl zcash_client_backend::data_api::WalletWrite::mark_transparent_addresses_exposed()`
+  (behind the `transparent-inputs` feature flag). Updates the `exposed_at_height`
+  column of each provided transparent address to the lesser of the existing
+  value and the provided height. The operation is atomic: if any address is not
+  tracked by the wallet, no updates are applied and the call returns
+  `SqliteClientError::AddressNotRecognized` identifying the first such address.
 
 ### Changed
 - Migrated to `sapling-crypto 0.7`, `orchard 0.13`, `zcash_encoding 0.4`, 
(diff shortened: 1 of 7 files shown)


zcash_client_sqlite/CHANGELOG.md
```diff
@@ -46,6 +46,12 @@ workspace.
     most likely has not synced through that height).
   Shard-read failures continue to surface via the existing
   `SqliteClientError::CommitmentTree` variant.
+- `impl zcash_client_backend::data_api::WalletWrite::mark_transparent_addresses_exposed()`
+  (behind the `transparent-inputs` feature flag). Updates the `exposed_at_height`
+  column of each provided transparent address to the lesser of the existing
+  value and the provided height. The operation is atomic: if any address is not
+  tracked by the wallet, no updates are applied and the call returns
```
nuttycom, line 53: Nonblocking nit; we don't document things like this in the CHANGELOG because it's covered by the "updated to `zcash_client_backend x.y` entry"; also, the information in the changelog is intended to assist users in migrating from one version to another, it should not include API documentation except insofar as that documentation is necessary to contextualize a change for migration purposes..

nuttycom (approved):
utACK 3fa5661fa3a39fcd4c21290697e4018f5f063fb2

zcash_client_sqlite/CHANGELOG.md
nullcopy: Ok, good to know. I'd been keeping the `zcash_client_backend` CHANGELOG entries very high level, but I allowed more context in the `zcash_client_sqlite` CHANGELOG. I'll keep that in mind going forward

PR #2310, opened 2026-04-22
fix: COR-1209 eip681 arb non numeric url encoded string

schell:
@nullcopy found a `proptest` failure in the `eip681` crate where roundtrips were broken because certain strings without any digits were being parsed as `Value::Number`. 

These changes fix that by changing the way `Value` is parsed. Previously the `Value::Number` variant parsing followed the ambiguous EIP-681 ABNF syntax and treated strings like "E", "e", "+" and "-" as the number `0`. This is (at least to me) obviously wrong. It's my opinion these should be treated as strings, not numbers.

## breakdown
b3c18d5e7b — cherry-picked upstream 6108185a (attribution to nullcopy and Co-Authored-By: Claude Opus 4.7 (1M context) preserved). Adds the CI-discovered proptest seed and the now-superseded generator tweak.
c0c7fc941f — the real fix, co-authored. Three files changed, +88/-7:
- components/eip681/src/parse.rs — restores the generator regex to [^\d], adds Number::has_mantissa_digits(), and filters digit-less Number candidates out of Value::parse so they fall through to STRING. Adds two unit tests (parse_value_lone_exponent_marker_is_string, parse_value_number_with_mantissa_digits).
- components/eip681/src/error.rs — adds ParseError::NumberMissingDigits variant (non-exhaustive enum, so no SemVer break).
- components/eip681/CHANGELOG.md — Added + Fixed entries under [Unreleased].

### DIFF
--- a/components/eip681/CHANGELOG.md
+++ b/components/eip681/CHANGELOG.md
@@ -11,6 +11,18 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `error::ParseError::NumberMissingDigits` (non-exhaustive): emitted internally
+  by `Value::parse` when a `Number` candidate has no mantissa digits.
+
+### Fixed
+- `Value::parse` no longer classifies inputs that lack any integer digits and
+  any decimal part (e.g. the lone alphabetic inputs `"e"` and `"E"`) as
+  `Value::Number`. Such inputs now correctly parse as `Value::String`, matching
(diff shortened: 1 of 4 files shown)


nullcopy (approved):
utACK c0c7fc941f396a8be7924c75c83a700647f88a31

nuttycom (approved):
utACK c0c7fc941f396a8be7924c75c83a700647f88a31

PR #2312, opened 2026-04-23
zcash_client: add address-to-account lookup with UA conflict detection

### LINKED ISSUE #1944: zcash_client_backend: Add `WalletRead::find_account_for_address` method

nuttycom:
• Add WalletRead::find_account_for_address to resolve the owning account for a wallet address, including Unified Address receiver-aware matching. 
• Introduce FindAccountForAddressError with Backend and UnifiedAddressConflict variants. • Add address_receiver_matches_ua and a default trait implementation in zcash_client_backend. 
• Override in zcash_client_sqlite::WalletDb with a single-query implementation (instead of scanning all accounts/addresses). 
• Add backend and sqlite tests for transparent, Sapling, Orchard, unknown-address, and cross-account UA conflict cases. • Update zcash_client_backend and zcash_client_sqlite changelogs.

Fixes #1944
Subsumes #2205


nuttycom (commented):
self-utACK 7575b70ad096fd89cd8f8bc78d0c3983819cd9d4

nuttycom:
@zodl-review

zcash_client_backend/src/data_api.rs
```diff
@@ -1673,6 +1715,36 @@ pub trait WalletRead {
...
+    /// - `Ok(None)` if no receiver of the address is recognized as belonging to any account.
+    /// - `Err(FindAccountForAddressError::Backend(_))` if the lookup fails due to a
+    ///   backend error.
+    /// - `Err(FindAccountForAddressError::UnifiedAddressConflict)` if the provided address
+    ///   is a Unified Address whose receiver components map to different accounts.
+    fn find_account_for_address<P: consensus::Parameters>(
+        &self,
+        params: &P,
+        address: &zcash_keys::address::Address,
+    ) -> Result<Option<Self::AccountId>, FindAccountForAddressError<Self::Error>>;
```
nullcopy: Did you mean to wire up the default here?

nuttycom:
> ## 🤖 Claude Review
> ## Suggestions
> * **Blocking:** Give `WalletRead::find_account_for_address` a default impl that delegates to `defaults::find_account_for_address`, to keep this additive rather than trait-breaking.

No. It is important that trait implementations explicitly opt-in to the provided inefficient approach, rather than silently inheriting it; the next release is API-breaking with respect to this trait anyway.

> * Consider unifying the non-UA path in `defaults` with the UA path via `decrypt_diversifier` on synthesized single-receiver UAs, so a Sapling address derivable from (but not previously exposed by) an account is still resolved.

:+1: 

> * Non-blocking: document whether Sapling-before-Orchard in `decrypt_diversifier` is contractual or incidental.

A better approach here is to return a set of diversifier indices instead of an option; that then leaves it to the caller to decide what to do if we have a frankenstein UA, instead of silently ignoring the case.

> * Non-blocking: consider additional test coverage (empty wallet; UA receiver matched against stored UA via the sqlite `receiver_flags` path specifically).
> 
> ## Areas of disagreement
> * **`#[non_exhaustive]` on `FindAccountForAddressError`:** Reviewer A asserts the enum is **not** `#[non_exhaustive]` and flags this as blocking. Reviewer B asserts it **is** marked `#[non_exhaustive]`. Verifying against `zcash_client_backend/src/data_api.rs`, the definition reads `#[derive(Debug)]` immediately followed by `pub enum FindAccountForAddressError<E> {` — i.e. there is **no** `#[non_exhaustive]` attribute. Reviewer A is correct: adding a variant later (e.g. for a new pool type) would be a breaking change. Recommend adding `#[non_exhaustive]`.

:+1:

zcash_client_backend/src/data_api/defaults.rs
```diff
@@ -0,0 +1,93 @@
...
+                    None => found_acc_id = Some(acc_id),
+                    Some(prev) if prev == acc_id => {}
+                    Some(_) => return Err(FindAccountForAddressError::UnifiedAddressConflict),
+                }
+            }
+        }
+        Ok(found_acc_id)
+    } else {
+        // Non-UA addresses have a single receiver, so no cross-account conflict is
+        // possible; return the first match.
```
nullcopy: Aside from my other question about the mapping between UAs and accounts, this non-UA raises a question about standalone imported addresses and accounts.

Is it documented somewhere that imported address are not allowed to be imported into multiple accounts? Currently we do not allow it in the sqlite implementation, but I don't see anything in the trait documentation instructing implementers to disallow that

zcash_client_backend/src/data_api/defaults.rs
```diff
@@ -0,0 +1,93 @@
...
+        let zcash_address = address.to_zcash_address(params);
+        for acc_id in wallet.get_account_ids()? {
+            for addr_info in wallet.list_addresses(acc_id)? {
```
nullcopy: This only lists exposed addresses (at least for the sqlite implementation). It might be worth documenting this limitation as it applies to the behavior of this account lookup

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -2587,8 +2590,39 @@ impl MockWalletDb {
...
+        let mut wallet = Self::new(network);
+        for (account_id, addresses) in entries {
+            wallet.account_ids.push(account_id);
+            wallet.addresses_by_account.insert(account_id, addresses);
+        }
+        wallet.account_ids.sort_unstable();
+        wallet
+    }
+
+    #[cfg(all(test, feature = "orchard"))]
```
nullcopy: Why is this test limited to orchard? I would expect functionality for non-orchard builds as well

zcash_client_backend/src/data_api.rs
```diff
@@ -1603,6 +1606,45 @@ pub trait InputSource {
     }
 }
 
+/// Errors that may occur when resolving the account controlling an address.
+#[derive(Debug)]
+pub enum FindAccountForAddressError<E> {
```
nullcopy: Would this be better located in `data_api/error.rs`?

zcash_client_backend/src/data_api.rs
```diff
@@ -1673,6 +1715,36 @@ pub trait WalletRead {
...
+    /// [`defaults::find_account_for_address`], which implements the semantics described below
+    /// via a linear scan over [`Self::get_account_ids`] and [`Self::list_addresses`].
+    ///
+    /// # Unified Addresses
+    ///
+    /// For Unified Addresses, each receiver component (transparent, Sapling, Orchard) is
+    /// matched individually against the wallet's tracked addresses. All matched receivers
+    /// must resolve to the **same** account; if any two receivers map to different accounts,
+    /// this is considered an inconsistent wallet state and `Err` is returned rather than
+    /// arbitrarily selecting one account.
```
nullcopy, line 1690: > All matched receivers must resolve to the **same** account

Is this a documented limitation of our implementation of UAs? I don't see it documented anywhere

I ask, because it feels weird to call this out as a specific limitation of this method, when its really an invariant we're prescribing across the wallet API.

zcash_client_memory/src/wallet_read.rs
```diff
@@ -734,6 +734,14 @@ impl<P: consensus::Parameters> WalletRead for MemoryWalletDb<P> {
         todo!()
     }
 
+    fn find_account_for_address<Q: consensus::Parameters>(
+        &self,
+        params: &Q,
+        address: &zcash_keys::address::Address,
+    ) -> Result<Option<Self::AccountId>, FindAccountForAddressError<Self::Error>> {
+        defaults::find_account_for_address(self, params, address)
```
nullcopy, line 743: This wouldn't be necessary if the default was provided in the trait definition. Curious if this is intentional

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1152,6 +1153,169 @@ pub(crate) fn list_addresses<P: consensus::Parameters>(
...
+/// address-kind-specific fallback:
+///
+/// - For Unified Addresses, each account's [`UnifiedIncomingViewingKey`] is asked whether it
+///   derived one of the UA's receivers, via
+///   [`UnifiedIncomingViewingKey::decrypt_diversifier`]. This finds every UA that any wallet
+///   account could have produced, whether or not it was previously exposed.
+/// - For Sapling addresses, stored UAs whose `receiver_flags` indicate a matching receiver
+///   are scanned and [`address_receiver_matches_ua`] confirms the actual overlap.
+/// - For transparent addresses, no fallback is needed: the exact-match query already covers
+///   both standalone transparent rows and transparent receivers cached on stored UAs.
```
nullcopy, line 1179: Again, might be worth mentioning that transparent addresses will only be matched if they'd previously been exposed by this wallet.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -1152,6 +1153,169 @@ pub(crate) fn list_addresses<P: consensus::Parameters>(
...
+        .prepare_cached(
+            "SELECT accounts.uuid, addresses.address
+             FROM addresses
+             JOIN accounts ON accounts.id = addresses.account_id
+             WHERE (receiver_flags & :shielded_flag) != 0",
+        )
+        .map_err(|e| E::Backend(e.into()))?;
+
+    let mut rows = stmt
+        .query(named_params![":shielded_flag": shielded_flag.bits()])
```
nullcopy, line 1274: (nit) s/shielded_flag/shielded_flags/

nullcopy (commented):
Flushing comments

zcash_client_backend/src/data_api/testing.rs
nuttycom: I think it's only *used* in tests that require the `orchard` feature, and so this is just to prevent a dead code warning when that feature is not enabled.

zcash_client_backend/src/data_api.rs
nuttycom: Such UAs violate ZIP 316, but we can't actually prevent them from existing, so I think the catch here is correct.

nuttycom: Someone can deliberately construct a UA that results in this situation, where it's indeterminate which account it should be considered belonging to. The wallet hasn't done anything wrong; it's the caller that has created the incoherent situation. I guess a very weird setup could import an orchard-only UFVK, and then import the Sapling-only UFVK from the same seed, but we also can't prevent that; a caller that's doing these kinds of shenanigans deserves whatever they get. :)

zcash_client_backend/src/data_api.rs
nuttycom: No, I intentionally want wiring up that default to be an intentional choice by an implementer; I wrote the same response to the AI review.

zcash_client_memory/src/wallet_read.rs
nuttycom: See my other responses.

zcash_client_sqlite/src/wallet/encoding.rs
```diff
@@ -231,6 +231,9 @@ bitflags! {
         const SAPLING = 0b00000100;
         /// The associated address can receive Orchard outputs.
         const ORCHARD = 0b00001000;
+
+        /// The union of all shielded receiver flags.
+        const SHIELDED = Self::SAPLING.bits() | Self::ORCHARD.bits();
```
nullcopy: You added this flag, but I don't see it used anywhere. Is that intentional?

zcash_client_backend/src/data_api.rs
```diff
@@ -3348,3 +3443,288 @@ pub trait WalletCommitmentTrees {
...
+            &[seed_tag; 32],
+            zip32::AccountId::ZERO,
+        )
+        .expect("valid seed")
+        .to_unified_full_viewing_key()
+    }
+
+    #[cfg(feature = "orchard")]
+    #[test]
+    fn find_account_for_unified_address_returns_account_when_receivers_map_to_same_account() {
```
nullcopy, line 3554: Why are these tests orchard only?

zcash_client_backend/src/data_api/testing.rs
nullcopy: I guess that begs the question why those tests are orchard only. I left a comment on one of the calling tests

zcash_client_sqlite/src/lib.rs
```diff
@@ -3387,4 +3407,367 @@ mod tests {
...
+    ) -> (UnifiedAddress, DiversifierIndex) {
+        state
+            .wallet_mut()
+            .get_next_available_address(account_id, UnifiedAddressRequest::AllAvailableKeys)
+            .unwrap()
+            .expect("address generation for account 1 must succeed")
+    }
+
+    #[cfg(all(feature = "orchard", feature = "transparent-inputs"))]
+    fn remove_account_from_db(
```
nullcopy: This method doesn't actually remove the account entry, it only removes the address records pertaining to the specified account ID. Not sure if this function is misnamed, or if bug

zcash_client_sqlite/src/lib.rs
```diff
@@ -3387,4 +3407,367 @@ mod tests {
...
+        state: &mut TestState<(), TestDb, LocalNetwork>,
+        acc1_id: AccountUuid,
```
nullcopy: (nit) Why `acc1_id`? This makes me expect other account ids, like `acc2_id`, `acc3_id`, etc

zcash_client_sqlite/src/wallet/encoding.rs
nuttycom: Ah, that's an appendix; it was added for a previous version that did more querying in the SQLite database that was then removed.

nullcopy:
utACK 9f101fc731b05c45682edb37aa1b4cc3110d4138

daira (commented):
Post-hoc ACK

PR #2317, opened 2026-04-25
zcash_transparent pczt signer: add expected_amount param to Input::sign for compiler-enforced caller responsibility

Zk-nd3r:
## summary

zcash_transparent::pczt::signer::Input::sign at zcash_transparent/src/pczt/signer.rs:42 trusts the updater-provided self.value blindly and signs whatever sighash the caller produces. caller-responsibility for amount validation is documented in the docstring at signer.rs:33-35 and as a literal TODO at pczt/src/roles/signer/mod.rs:158 ("Check consistency of the input being signed").

zip-244 binds input amounts into the sighash. if signer signs claimed-value and validator computes actual-value the signatures differ and the tx is rejected at consensus. self-correcting denial of broadcast not fund loss. but the failure mode is silent at the signer side and the caller has no compile-time signal that they are responsible for cross-checking value before calling.

this pr adds `expected_amount: Zatoshis` as a required parameter to `Input::sign`. the function returns `SignerError::ValueMismatch` if `self.value` differs from `expected_amount`. caller responsibility becomes compiler-enforced.

## changes

- `zcash_transparent/src/pczt/signer.rs` Input::sign signature gets `expected_amount: Zatoshis` param plus early `ValueMismatch` check.
- `zcash_transparent/src/pczt/signer.rs` SignerError gets `ValueMismatch { expected, updater_provided }` variant.
- `zcash_transparent/src/pczt/signer.rs` test added covering the mismatch path.
- `pczt/src/roles/signer/mod.rs` caller updated to pass expected_amount from already-trusted context. TODO at line 158 marked done.

## breaking change

yes. all callers of Input::sign must now pass expected_amount. internal callers in this PR are updated. external embedders see a one-line addition.

## context

found while filing the panic-on-untrusted-input GHSA family in librustzcash this week (GHSA-m8hw-x85c-j3qq, GHSA-7mqm-gq3v-pmcp, GHSA-f3vg-grqw-2q2g, GHSA-c66h-h8v6-c8pg, GHSA-9m3x-v6f2-mh47, GHSA-6r2c-r2qx-m358). this is the hardening companion. not a security advisory itself, self-correcting via consensus rejection. opens the contract for the caller earlier.

### DIFF
--- a/pczt/src/roles/signer/mod.rs
+++ b/pczt/src/roles/signer/mod.rs
@@ -116,6 +116,11 @@ impl Signer {
     /// that consensus will reject. Caller derives `expected_amount` from already-trusted
     /// context (the spend record being signed) before invoking the signer. The check is
     /// compiler-enforced via [`transparent::pczt::SignerError::ValueMismatch`].
+    ///
+    /// This check does not replace the caller's broader responsibility to perform
+    /// semantic validity checks on the PCZT before signing. The caller must still
+    /// verify recipients, outputs, change amounts, fees, and any policy-specific
+    /// expectations against trusted context.
     pub fn sign_transparent(
         &mut self,
         index: usize,
@@ -174,8 +179,8 @@ impl Signer {
 
         // Consistency of input value is now checked at the input.sign() boundary via the
         // `expected_amount: Zatoshis` precondition; see `transparent::pczt::Input::sign`
-        // and `SignerError::ValueMismatch`. The closure `f` is responsible for threading
-        // the caller-supplied expected amount through to that check.
+        // and `SignerError::ValueMismatch`. Other semantic checks remain the caller's
+        // responsibility before invoking the signer.
 
         // Generate or apply the signature.
         f(input, &self.tx_data, &self.txid_parts, &self.secp).map_err(Error::TransparentSign)?;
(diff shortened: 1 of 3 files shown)


pczt/src/roles/signer/mod.rs
```diff
@@ -110,20 +111,24 @@ impl Signer {
 
     /// Signs the transparent spend at the given index with the given spending key.
     ///
-    /// It is the caller's responsibility to perform any semantic validity checks on the
-    /// PCZT (for example, comfirming that the change amounts are correct) before calling
-    /// this method.
```
nuttycom, line 127: I agree that the expected amount is a reasonable thing to always enforce is checked, but it's not the only potential validity check; I think that perhaps this is why no specific checks were originally included here. We should probably retain some variant of the removed documentation; for example, it's still important for the caller to check that outputs and change amounts are what they expect before signing.

nuttycom (commented):
Why single out the transparent input amount for checking? Everything else that goes into the sighash also needs to be validated by the caller.

pczt/src/roles/signer/mod.rs
Zk-nd3r: Agreed.  The amount check is useful, but it should not imply that it is the only semantic validation the caller needs to perform before signing.  I think the caller-responsibility warning should stay, with explicit mention that outputs, change amounts, recipients, fees, and policy-specific expectations still need to be validated before calling into the signer.

Zk-nd3r: Docs updated in the current diff. The amount check is framed as one enforced precondition, and the broader caller-responsibility warning is kept for recipients, outputs, change, fees, and policy checks.

Zk-nd3r:
Thanks for the review. Closing this old approach rather than rebasing it: the added argument does not prove the second amount came from trusted context, and the current signer architecture has moved substantially while retaining the broader caller-validation contract. If this check is revisited, it should be designed against the current signer/verifier API rather than as a one-field breaking change.

PR #2319, opened 2026-04-27
feat: initial pr for upstream snap zcash libs

### LINKED ISSUE #2323: Update `zcash_client_memory` to reflect `zcash_client_backend` changes up to version 0.22
`zcash_client_memory` has bit-rotted due to persistence issues and needs to be brought up to date.

sqhell:
librustzcash changes to support ChainSafe zcash snap wallet project

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1360,7 +1360,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.0",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1614,7 +1614,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4010,7 +4010,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/zcash_client_memory/CHANGELOG.md
+++ b/zcash_client_memory/CHANGELOG.md
@@ -1,0 +1,68 @@
+# Changelog
+All notable changes to this library will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this library adheres to Rust's notion of
+[Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+- `ReceivedNote::from_decrypted_sapling_output` and
+  `ReceivedNote::from_decrypted_orchard_output` for constructing received notes
+  from decrypted incoming transaction outputs.
+- `ReceivedNote::note()` and `ReceivedNote::memo()` public accessor methods.
+- `SentNote::from_account_id()`, `SentNote::to()`, `SentNote::value()`, and
+  `SentNote::memo()` public accessor methods.
+- `TransactionEntry::expiry_height()`, `TransactionEntry::status()`, and
+  `TransactionEntry::mined_height()` are now public.
+- `MemoryWalletDb::received_notes()`, `MemoryWalletDb::sent_notes()`,
+  `MemoryWalletDb::tx_table()`, and `MemoryWalletDb::get_block_time()` public
+  accessor methods for transaction history queries.
+- `ReceivedNoteTable::from_notes()` constructor that rebuilds indexes from a
+  `Vec<ReceivedNote>`.
+- `ReceivedNoteTable` now maintains O(1) lookup indexes by nullifier and note ID.
+- Unit tests for `ReceivedNoteTable` indexed lookup operations.
+- `WalletRead::list_addresses()` implementation.
+- `WalletRead::get_last_generated_address_matching()` implementation.
+- `WalletRead::utxo_query_height()` implementation (behind `transparent-inputs`
+  feature flag).
+- `WalletRead::get_transparent_address_metadata()` implementation (behind
+  `transparent-inputs` feature flag).
+- `TransparentKeyScope` serialization field in protobuf for
+  `ReceivedTransparentOutput`, with backwards-compatible deserialization that
+  defaults to `EXTERNAL` for old data.
+- `Account::find_key_scope_for_transparent_address()` for looking up the key
+  scope of a transparent address across ephemeral, unified, and legacy addresses.
+- `WalletWrite::store_decrypted_tx()` now stores incoming Sapling and Orchard
+  notes instead of panicking.
+- `WalletWrite::notify_address_checked()` is now a no-op instead of panicking.
+- `ReceivedNote`, `ReceivedNoteTable`, `SentNote`, `SentNoteTable`,
+  `TransactionEntry`, and `TransactionTable` are now re-exported as public types.
+
+### Fixed
+- `get_wallet_summary()` no longer incorrectly filters mined transactions by
+  expiry height. Only expired *unmined* transactions are now excluded from
+  balance calculations.
+- `get_transparent_receivers()` no longer panics when `include_change` or
+  `include_standalone` are true; it now returns external addresses regardless
+  of these flags.
+- `TransparentKeyScope` deserialization no longer panics with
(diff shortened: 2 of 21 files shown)


zcash_client_memory/src/types/memory_wallet/serialization.rs
```diff
@@ -353,7 +353,7 @@ impl<P: Parameters> From<&MemoryWalletDb<P>> for proto::MemoryWallet {
                 .map(|r| proto::ScanQueueRecord::from(*r))
                 .collect(),
 
-            sapling_tree: tree_to_protobuf(&wallet.sapling_tree).unwrap(),
+            sapling_tree: tree_to_protobuf(&wallet.sapling_tree).ok().flatten(),
```
czarcas7ic, line 356: Can you explain why this was changed from unwrap? My understanding is that `sapling_tree`, `orchard_tree`, and `birthday` should panic in absence, but could be missing some context here.

zcash_client_memory/src/types/memory_wallet/serialization.rs
```diff
@@ -366,7 +366,7 @@ impl<P: Parameters> From<&MemoryWalletDb<P>> for proto::MemoryWallet {
                 .collect(),
 
             #[cfg(feature = "orchard")]
-            orchard_tree: tree_to_protobuf(&wallet.orchard_tree).unwrap(),
+            orchard_tree: tree_to_protobuf(&wallet.orchard_tree).ok().flatten(),
```
czarcas7ic, line 369: Same comment as https://github.com/zcash/librustzcash/pull/2319/changes#r3183423484

zcash_client_memory/src/types/account.rs
```diff
@@ -627,7 +664,7 @@ mod serialization {
                 }
                 .clone(),
                 viewing_key: acc.viewing_key.encode(&EncodingParams),
-                birthday: Some(acc.birthday().clone().try_into().unwrap()),
+                birthday: acc.birthday().clone().try_into().ok(),
```
czarcas7ic, line 667: Same comment as https://github.com/zcash/librustzcash/pull/2319/changes#r3183423484

zcash_client_memory/src/wallet_read.rs
```diff
@@ -745,15 +803,24 @@ impl<P: consensus::Parameters> WalletRead for MemoryWalletDb<P> {
...
-        _account: Self::AccountId,
+        account: Self::AccountId,
         _address_filter: UnifiedAddressRequest,
     ) -> Result<Option<UnifiedAddress>, Self::Error> {
-        todo!()
+        // Return the last (highest diversifier index) generated address for this account
+        Ok(self
+            .accounts
+            .get(account)
+            .and_then(|acc| acc.addresses().values().last().cloned()))
```
czarcas7ic, line 813: I checked the trait contract here: https://github.com/zcash/librustzcash/blob/525d61a2363786906ca2107cadace29ae32a9484/zcash_client_backend/src/data_api.rs#L1726-L1736

It says this must return the most recent address that conforms to the requested `address_filter`, or `Ok(None)` if none does. We still ignore `address_filter` here, so it can return the latest address even when it does not match the requested receiver requirements (for example, a shielded-only request could get back a UA with a transparent receiver.

zcash_client_memory/src/wallet_read.rs
```diff
@@ -651,12 +664,10 @@ impl<P: consensus::Parameters> WalletRead for MemoryWalletDb<P> {
...
-        if _include_change {
-            unimplemented!("include_change is not yet supported");
-        }
-        if _include_standalone {
-            unimplemented!("include_standalone is not yet supported");
-        }
+        // NOTE: For the memory wallet, we only track external (receiving) addresses.
+        // Change addresses (Internal scope) are not separately tracked - they're derived
+        // on-demand during transaction creation. Standalone addresses are not supported.
+        // We simply return the external addresses regardless of the include_* flags.
```
czarcas7ic, line 670: I checked the trait contract here: https://github.com/zcash/librustzcash/blob/525d61a2363786906ca2107cadace29ae32a9484/zcash_client_backend/src/data_api.rs#L1862-L1886

Wouldn't it be better to return an explicit unsupported error when these flags request receivers this implementation cannot return? `include_change=true` should include transparent change receivers, and `include_standalone=true` should matter when `transparent-key-import` is enabled, but this implementation always returns only external UA transparent receivers. If the memory wallet intentionally cannot support those sets yet, I think the API should make that visible rather than letting callers believe they got the full scan set.

zcash_client_memory/src/types/notes/received.rs
```diff
@@ -222,38 +324,57 @@ impl ReceivedNoteTable {
...
+        if let Some(&idx) = self.note_id_index.get(&note.note_id) {
+            // Upsert: update existing note
+            let existing = &mut self.notes[idx];
+            // If nullifier changed, update the nullifier index
+            let old_nf = existing.nf;
+            existing.nf = note.nf.or(existing.nf);
+            existing.is_change = note.is_change || existing.is_change;
+            existing.commitment_tree_position = note
+                .commitment_tree_position
+                .or(existing.commitment_tree_position);
```
czarcas7ic, line 337: Now that this PR stores incoming decrypted transaction outputs, should this upsert also merge `memo`? A note may already have been inserted by compact-block scanning with `Memo::Empty`. If full transaction decryption later inserts the same note with the real memo, this branch updates the nullifier/change/position fields but leaves the empty memo in place.

zcash_client_memory/src/types/memory_wallet/mod.rs
```diff
@@ -1101,11 +1093,19 @@ impl<P: consensus::Parameters> MemoryWalletDb<P> {
                 entry.get_mut().txout = output.txout().clone();
             }
             Entry::Vacant(entry) => {
+                // Look up the key scope for this address by checking ephemeral,
+                // unified, and legacy transparent addresses in the account
+                let key_scope = self
+                    .accounts
+                    .get(*receiving_account)
+                    .and_then(|account| account.find_key_scope_for_transparent_address(address))
+                    .unwrap_or(::transparent::keys::TransparentKeyScope::EXTERNAL);
```
czarcas7ic, line 1102: Should this return an error instead of defaulting to `EXTERNAL`? At this point the output has already been classified as belonging to `receiving_account`, so failing to find its transparent key scope seems like inconsistent wallet state. Treating unknown scope as external could hide that inconsistency and persist incorrect metadata.

zcash_client_memory/CHANGELOG.md
```diff
@@ -0,0 +1,68 @@
...
+- `SentNote::from_account_id()`, `SentNote::to()`, `SentNote::value()`, and
+  `SentNote::memo()` public accessor methods.
+- `TransactionEntry::expiry_height()`, `TransactionEntry::status()`, and
+  `TransactionEntry::mined_height()` are now public.
+- `MemoryWalletDb::received_notes()`, `MemoryWalletDb::sent_notes()`,
+  `MemoryWalletDb::tx_table()`, and `MemoryWalletDb::get_block_time()` public
+  accessor methods for transaction history queries.
+- `ReceivedNoteTable::from_notes()` constructor that rebuilds indexes from a
+  `Vec<ReceivedNote>`.
+- `ReceivedNoteTable` now maintains O(1) lookup indexes by nullifier and note ID.
```
czarcas7ic, line 24: nit: The indexes are `BTreeMap`s (`received.rs` lines 57 and 59), so these lookups are log rather than O(1).

Cargo.lock
```diff
@@ -1360,7 +1360,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.0",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
```
schell, line 1363: These downgrades were not intentional, correct?

zcash_client_memory/src/wallet_read.rs
```diff
@@ -722,6 +733,37 @@ impl<P: consensus::Parameters> WalletRead for MemoryWalletDb<P> {
...
+            .get_account(account)?
+            .ok_or(Error::AccountUnknown(account))?;
+
+        // Search through account addresses for matching transparent receiver
+        for (diversifier_index, ua) in acc.addresses().iter() {
+            if let Some(transparent_addr) = ua.transparent() {
+                if transparent_addr == address {
+                    let metadata = TransparentAddressMetadata::derived(
+                        Scope::External.into(),
+                        NonHardenedChildIndex::from_index((*diversifier_index).try_into().unwrap())
```
schell, line 752: Stray `.unwrap()` here. At the very least we need an `.expect(...)` with a good description of why it will never happen.

PR #2326, opened 2026-04-28
zcash_client_backend: Deprecate `WalletWrite::truncate_to_height`

nuttycom:
Height-based truncation is tricky to use due to the need to find a height that the note commitment tree can be pruned to. With the combination of `rewind_to_height` and `truncate_to_chain_state` we should no longer need it.


nuttycom:
@zodl-review

nuttycom:
This isn't the correct thing to do; we actually need truncation to height for reorgs in scanning.

PR #2331, opened 2026-05-01
ci: skip rust-cache save on Windows all-features

S1nus:
The Swatinem/rust-cache post-step seems to fail on large diffs for Windows, even though all tests pass


nuttycom (approved):
utACK

PR #2332, opened 2026-05-01
Add top-level warning: The only way to check Zcash consensus validity is to use a Zcash consensus node

daira:


nuttycom (approved):
doc-only ACK

PR #2333, opened 2026-05-02
Spendability redesign: height-based rule with anchor-priority window

### LINKED ISSUE #982: zcash_client_backend: Use knowledge of inserted treestates to reduce time to spendability

### LINKED ISSUE #2329: rewindToHeight affects spendability
We implemented new rewindToHeight method in a way that when calling in while adding a Keystone device, the spendability of Zodl wallet is not affected except last 100 blocks. However, when I test the new method on my Zodl wallet and add a KS device, ~30% of my balance becomes unspendable and flips to pending state. I can send funds from those 70%, that works. But those 30% shouldn't be affected according to the design and technical architecture of the rewindToHeight method.

nuttycom:
Replaces the wallet's spendability rule with a single height-based predicate shared by balance reporting and note selection, pins the anchor to the tree state at the confirmations-policy depth, confines note commitment tree checkpoints to the pruning window, and keeps the scan queue contiguous. Together these close the gap in which `get_wallet_summary` reported value that `propose_transfer` then refused (`ScanRequired` / `InsufficientFunds`), and the converse gap in which a wallet that lagged the chain tip spent against, and published, an anchor as old as its offline period.

## Spendability rule

A shielded note is spendable when all of the following hold. `get_wallet_summary` and note selection (`select_spendable_notes` and the value-targeted selection behind `propose_transfer`) evaluate the same predicate.

- **Anchor floor.** The note's `witness_anchor_stable` height is set and lies at or below the chosen anchor. It is the height through which the note's witness data is settled, every block from the note's own block through it having been scanned: for a note in a completed shard, the shard's end height; for a note in the open shard, the pruning floor as of the last scan that reached the tip, never below the note's own height. Floors advance as scanning progresses and are cleared above any truncation height.
- **Pruning window scanned.** No `scan_queue` range above `Scanned` overlaps the chain-tip pruning window at or below the policy anchor (`target - trusted confirmations`). A not-yet-scanned tip extension strictly above the policy anchor does not suspend spendability.
- **Durable witness region.** Either the note's shard is complete and every block after the note's own block through the shard's end has been scanned, judged from the scan queue rather than from the subtree-root table alone (a server-reported root says nothing about which leaves the wallet holds), or the floor reaches the bottom of the window with no unscanned range in between. Blocks before the note within its shard are irrelevant: the frontier inserted when the note's block was scanned supplies the left side of its witness.
- **Anchor available.** The pool's tree has a checkpoint at the anchor height.
- **Confirmations.** The note meets the confirmations policy.

`ScanPriority::Anchor`, slotted between `ChainTip` and `Verify`, is stamped on the pruning window by `rewind_to_chain_state`, `truncate_to_height`, and `truncate_to_chain_state`, so re-establishing a usable anchor is scheduled ahead of ordinary forward sync.

## Anchor selection

`get_target_and_anchor_heights` returns the highest checkpoint at or below `target - min_confirmations` only when every block between that checkpoint and that depth has been scanned, so the anchor's tree state is exactly the state at the policy depth; for a wallet synced to the tip that checkpoint is the policy depth itself. Otherwise there is no anchor: balance reports nothing spendable, and proposals (transfers and shielding alike) fail with `ScanRequired` until the wallet has scanned to the policy depth. Previously the most recent checkpoint was returned unconditionally, so a wallet reopened after an offline period revealed the length of that period on chain. Shielding keeps the anchor requirement because Orchard and Ironwood actions carry dummy spends.

## Checkpoints

`put_blocks` uses `Checkpoint` retention only within `PRUNING_DEPTH` blocks of the chain tip, and creates a checkpoint at every scanned height in that window in every tree, whether or not the block carries a shielded output. Scanning older blocks creates no checkpoints; their commitments keep only the marking that protects a wallet note from pruning. Heights the ZIP 318 anchor-retention policy retains are checkpointed regardless. `ll::wallet::put_blocks` takes the wallet's chain tip as an argument for this. A stretch of more than `PRUNING_DEPTH` blocks without a shielded output no longer leaves notes in the open tip shard unspendable.

## Scan queue contiguity

`put_blocks`, `store_decrypted_tx`, `set_transaction_status`, and `put_received_transparent_utxo` extend the scan queue when they learn of a block above the known chain tip, so the queue is contiguous from the wallet birthday to the tip and both the spendability checks and `suggest_scan_ranges` see every unscanned height.

## Reorg detection

Two hardenings found while auditing the reorg guards on the rewind and re-scan paths this PR touches, independent of the other changes:

- `scan_cached_blocks` compares each block against the metadata the wallet recorded at that height before scanning it, and reports a mismatch as `ScanError::BlockHashMismatch`, a continuity error. Previously the mismatch surfaced only as the storage layer's `BlockConflict`, which a sync loop had no reason to treat as a signal to rewind.
- `WalletWrite::put_blocks` now returns `PutBlocksError<Self::Error>`, separating a chain-continuity failure detected while persisting a batch from a failure to persist it. A shardtree `InsertionError::Conflict` raised while updating a pool's tree (the inserted frontier or commitments disagree with a leaf or computable node the wallet holds) is classified as `ScanError::CommitmentTreeConflict` at the first block of the batch, and `scan_cached_blocks` reports it as `Error::Scan`. Previously it surfaced as `PutBlocksCommitmentTree`, which the sync loop treated as unrecoverable. Other tree errors are unchanged.

A sync loop that rewinds on `ScanError::is_continuity_error` therefore recovers from both a reorg below its scanned range and a tree conflict the same way it recovers from a parent-hash mismatch at the start of a batch. Recovery is still bounded by the checkpoint window; a reorg deeper than that is tracked in #3035.

## Schema

`*_received_notes.witness_stabilized` (boolean) is replaced by `witness_anchor_stable` (block height). The migration is automatic.

## Test support (`test-dependencies`)

`shard_stub::fake_advance_to` and the `ShieldedPoolTester` helpers listed in the `zcash_client_backend` changelog build note-commitment-tree fixtures with real notes placed near shard boundaries and the shard interior filled by random subtree stubs, replacing the 65,537-output mega-block fixtures.

## Verification

- `cargo test -p zcash_client_sqlite --lib --features orchard,transparent-inputs`: 537 passed, 0 failed, 2 ignored.
- `cargo clippy --all-targets -- -D warnings` on `zcash_client_backend` and `zcash_client_sqlite` with `--all-features`; on `zcash_client_sqlite` with no default features, `orchard` alone, and `transparent-inputs` alone; and on `zcash_client_backend` with no default features.
- `cargo doc --document-private-items` and `cargo fmt --check` are clean.
- Every commit builds its test targets (`git rebase --exec 'cargo check -p zcash_client_sqlite --tests --features orchard'`).

## Follow-ups

- #3035: a reorg deeper than the checkpoint window cannot be rewound and must be surfaced as a distinct condition requiring intervention.
- zcash/incrementalmerkletree#202: three shardtree insertion paths accept data that disagrees with stored node hashes without raising `InsertionError::Conflict`, so `ScanError::CommitmentTreeConflict` catches a frontier or commitment that meets an existing leaf or computable node, but not every divergent gap fill.
- `zcash_client_memory` maps `ScanPriority::Anchor` to `proto::ScanPriority::ChainTip` as a placeholder (lossy on round-trip).
- A wallet reopened after a long offline period must scan to the policy anchor before any shielded balance is spendable. Scheduling the pruning window ahead of the `ChainTip` backlog on a plain tip advance, or inserting the server's tree state at the policy anchor, would shorten that wait for notes in completed shards.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>


zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -5201,6 +5201,403 @@ where
...
+/// [`stabilized_note_spendable_after_deep_rewind`]), perform the deep rewind
+/// *without* a subsequent `update_chain_tip`, and check that
+///   - if `get_spendable_balance > 0`, then both `propose_transfer` and
+///     `create_proposed_transactions` succeed for at least the marginal fee, and
+///   - otherwise, `propose_transfer` fails with `ScanRequired` or
+///     `InsufficientFunds`.
+///
+/// This test is expected to fail until balance reporting is taught to consider
+/// whether an anchor's tree root can actually be constructed.
+pub fn rewind_balance_implies_spendable<T, Dsf>(ds_factory: Dsf, cache: impl TestCache)
```
nullcopy: `stabilized_note_spendable_after_deep_rewind` was intended to check that the post-rewind balance was fully usable. The fact that it doesn't test the difference between `get_spendable_balance` and what `propose_transfer` tries to use is an oversight. Could these conditions just be tested in "Step 6" of `stabilized_note_spendable_after_deep_rewind`, instead of adding a whole new test for this?

nullcopy: Same question for `rewind_with_chain_tip_balance_implies_spendable` below

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2557,7 +2557,7 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
         // If the shard containing the anchor height contains any unscanned ranges that start
         // below or including that height, none of our shielded balance is currently spendable.
         #[tracing::instrument(skip_all)]
-        fn is_any_spendable(
+        fn tip_shard_scanned(
```
nullcopy: s/tip_shard_scanned/shard_scanned/`

Technically this doesn't check the tip shard. It checks whichever shard the anchor happens to be in, which may or may not be the tip (especially likely for transactions constructed around the time of a shard boundary)

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2660,16 +2670,17 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
+            // reasonable confirmation policy, and its witness data cannot be removed by
+            // truncation.
             //
             // Non-stabilized notes require more checks: we must have enough chain tip information
             // to construct witnesses, the shard that the note resides in must be sufficiently
             // scanned that we can construct the witness for the note, and the note has enough
             // confirmations to be spent.
-            let is_spendable = witness_stabilized
-                || (any_spendable
+            let is_spendable = (witness_stabilized && anchor_available)
```
nullcopy: Should we also confirm the anchor height is greater than the note's mined height?

zcash_client_sqlite/src/wallet/init/migrations/witness_anchor_stable.rs
```diff
@@ -0,0 +1,339 @@
+//! Replaces the boolean `witness_stabilized` flag on received-note tables with a
+//! `witness_anchor_stable` column whose value is the containing shard's
+//! `subtree_end_height` at the time of stabilization. The new column carries the same
```
nullcopy: In our discussion yesterday, you mentioned marking the anchor stable even before the shard has been completed. Is that still the case? This comment might need updating.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -54,6 +54,7 @@ mod v_tx_outputs_key_scopes;
 mod v_tx_outputs_return_addrs;
 mod v_tx_outputs_use_legacy_false;
 mod wallet_summaries;
+mod witness_anchor_stable;
```
nullcopy, line 64: Need to update the migration dependency graph below

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -393,8 +395,18 @@ pub(crate) fn mark_stabilized_notes<P: consensus::Parameters>(
     ) -> Result<(), SqliteClientError> {
         let sql = format!(
             "UPDATE {pool}_received_notes
-             SET witness_stabilized = 1
-             WHERE witness_stabilized = 0
+             SET witness_anchor_stable = (
+                 SELECT shard.subtree_end_height
+                 FROM {pool}_tree_shards shard
+                 WHERE shard.subtree_end_height IS NOT NULL
```
nullcopy: I thought you were going to stabilize anchors even if the shard had not completed yet (i.e. in the tip shard)

zcash_client_memory/src/types/scanning.rs
```diff
@@ -176,6 +176,7 @@ mod serialization {
                     ScanPriority::OpenAdjacent => proto::ScanPriority::OpenAdjacent as i32,
                     ScanPriority::FoundNote => proto::ScanPriority::FoundNote as i32,
                     ScanPriority::ChainTip => proto::ScanPriority::ChainTip as i32,
+                    ScanPriority::Anchor => proto::ScanPriority::ChainTip as i32, //FIXME
```
nullcopy: Don't forget this FIXME

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -53,11 +54,60 @@ pub(crate) fn parse_priority_code(code: i64) -> Option<ScanPriority> {
...
+        None => return Ok(()),
+    };
+
+    let wallet_birthday = match wallet_birthday(conn)? {
+        Some(h) => h,
+        None => return Ok(()),
+    };
+
+    let chain_tip_u32 = u32::from(chain_tip);
+    let pruning_floor = chain_tip_u32.saturating_sub(PRUNING_DEPTH);
```
nullcopy: Pruning floor is defined elsewhere (e.g. in `rewind_to_height`) to be `tip - (PRUNING_DEPTH - 1)`. We should agree on one, and probably put this calculation in a helper to ensure we stay consistent

nullcopy: I believe our logic still lines up, because I essentially did the opposite of your `pruning_floor + 1` logic below.. but this is ripe for an off-by-one error

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +419,160 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
-/// A note within the chain-tip shard cannot be stabilized by this function, because the
-/// chain-tip shard does not yet have a `subtree_end_height` value.
-pub(crate) fn mark_stabilized_notes<P: consensus::Parameters>(
+/// The stored value represents a block height at which witness data for a note is **locally
+/// stable**. This is to say that within the shard containing the note, enough note commitment tree
+/// data is available that it is possible to construct this note's witness up to shoard boundaries.
+/// Once written it is monotonically non-decreasing: the value advances over a row's lifetime under
+/// two cases: first, when an active-shard floor is promoted to a completed-shard floor, and
+/// secondly, in the case of notes in the chain-tip shard, when a contiguously scanned region
+/// between the note's position and an anchor increases. In the chain tip shard only, chain reorgs
```
nullcopy: > secondly, in the case of notes in the chain-tip shard, when a contiguously scanned region between the note's position and an anchor increases.

Should mention the contiguously scanned region must include the note. The way this reads, it sounds like any contiguously scanned region after the note would count, even if the note itself is in a gap

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +419,160 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+/// and explicit rewind operations may cause this value to decrease.
+///
+/// - **First-time stabilize** (NULL → floor): for any note whose containing shard is
```
nullcopy: Use of "floor" in this whole paragraph confuses me. I think you mean "stable height"?

e.g. Its weird to refer to the shard's end height as the floor on L442

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +419,160 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+            named_params![
+                ":pruning_floor": pruning_floor,
+                ":shard_height": shard_height,
+            ],
+        )?;
+        Ok(())
+    }
+
+    if let Some(chain_tip) = chain_tip_height(conn)? {
+        let pruning_floor: u32 = u32::from(chain_tip).saturating_sub(PRUNING_DEPTH - 1);
```
nullcopy: another case of `pruning_floor` math being potentially off-by-one

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +419,160 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
-                   AND shard.subtree_end_height <= :pruning_floor
-                   AND ({pool}_received_notes.commitment_tree_position >> :shard_height)
-                       = shard.shard_index
-                   AND shard.shard_index NOT IN (
-                       SELECT shard_index FROM v_{pool}_shard_unscanned_ranges
-                   )
+             SET witness_anchor_stable = COALESCE(
+                 (SELECT shard.subtree_end_height
+                  FROM {pool}_tree_shards shard
+                  WHERE shard.subtree_end_height IS NOT NULL
```
nullcopy: Need to ensure `witness_anchor_stable IS NULL`, otherwise this doesn't count as a "first time" stabilization, and such notes should be handled by the promotion arm instead

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +419,160 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
 
-        // Mark stabilized notes in each pool
-        mark_pool(conn, "sapling", SAPLING_SHARD_HEIGHT, pruning_floor)?;
+        first_time_stabilize(
+            conn,
+            SAPLING_TABLES_PREFIX,
+            SAPLING_SHARD_HEIGHT,
+            pruning_floor,
+        )?;
+        promote_on_completion(
```
nullcopy, line 696: splitting this into two update transactions, doubles the database traffic in a potentially hot path. I don't think its too bad, because the bulk of the column index does a lot of the heavy lifting, but I feel like this could be done in one query if we care to optimize it.

zcash_client_backend/src/data_api/testing/shard_stub.rs
```diff
@@ -169,9 +166,11 @@ where
     if !shard_stubs.is_empty() {
         let entries: Vec<(u64, T::MerkleTreeHash)> = shard_stubs.into_iter().collect();
         let mut start_index = entries[0].0;
-        let mut roots: Vec<CommitmentTreeRoot<T::MerkleTreeHash>> = vec![
-            CommitmentTreeRoot::from_parts(advance_height, entries[0].1.clone()),
-        ];
+        let mut roots: Vec<CommitmentTreeRoot<T::MerkleTreeHash>> =
+            vec![CommitmentTreeRoot::from_parts(
```
nullcopy, line 170: (nit) these changes to the fake-advance fixture should probably be squashed into the commit which added that fixture

zcash_client_sqlite/src/wallet/scanning.rs
nullcopy: nvm, I see you do exactly that in a later commit 😅

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -58,6 +58,50 @@ pub(crate) fn parse_priority_code(code: i64) -> Option<ScanPriority> {
     }
 }
 
+/// Returns the inclusive lower bound of the chain-tip pruning window: the height below
+/// which the wallet's surviving tree checkpoints are no longer guaranteed to be available.
+/// Concretely, this is `chain_tip - PRUNING_DEPTH`. Heights strictly greater than this
+/// value (up through `chain_tip`) are within the pruning window.
+pub(crate) fn pruning_floor(chain_tip: BlockHeight) -> BlockHeight {
```
nullcopy, line 68: There are several other places in the codebase that should use this helper. Right now we calculate the pruning floor differently: in some places its `tip - PRUNING_DEPTH`, in others its `tip - (PRUNING_DEPTH - 1)`

nullcopy (commented):
Flushing comments

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2658,28 +2651,38 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
...
+                received_height,
+                tx_trusted,
+                max_shielding_input_height,
+                tx_shielding_inputs_trusted,
+            ) == 0;
+            let is_spendable = stored_at_or_below_chosen
+                && tip_window_scanned
+                && anchor_available
+                && confirmations_met;
+            let _ = max_priority; // no longer participates in the spendability rule
```
p0mvn: Since `max_priority` no longer participates in the spendability rule, can we remove it from this query entirely, along with the scan-state join and parsing block? 

Keeping it here makes it look like per-note shard priority is still semantically relevant.

p0mvn: I see good coverage for the broader balance/spend agreement and for the unscanned anchor-window negative case. What I think could be added is a regression test proving that per-note shard `max_priority` is no longer part of the rule. 

If we remove `max_priority` from the query, a targeted test for that boundary would make the cleanup easier to review.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -585,30 +564,28 @@ where
...
-                     AND scan_state.max_priority <= :scanned_priority -- the note shard is fully scanned or ignored
-                 )
-             )
+             -- The stored floor must lie at or below the chosen anchor, and the
+             -- chain-tip pruning window must have no `Anchor` priority overlap so the
+             -- wallet has fully scanned through any disturbance left by truncate,
+             -- rewind, or account-import.
+             AND rn.witness_anchor_stable IS NOT NULL
+             AND rn.witness_anchor_stable <= :anchor_height
+             AND :tip_window_scanned = 1
```
p0mvn: Would it make sense to consolidate the spendability predicate into one shared helper and use it from both balance reporting and note selection? 

Right now, `wallet.rs` and `common.rs` encode the same rule differently, with part of the spend path in SQL and part in Rust.

Since this PR is specifically about keeping reported balance and selected notes in agreement, having one canonical predicate would make future changes much safer.

This could totally be done in a separate PR and is non-blocking. Just thought I'd mention, as it came to mind.

p0mvn: As a side note, I noticed that balance reporting requires `anchor_available` ([ref](https://github.com/zcash/librustzcash/blob/b1100a47a1e9bed041e8881f7f57a5d4a4543655/zcash_client_sqlite/src/wallet.rs#L2570)), but note selection does not appear to check it before returning notes.

Consolidation could reduce the risk of divergence while reducing the need to review these details

zcash_client_sqlite/src/wallet.rs
nuttycom: Removed and replaced with scanning::tip_window_fully_scanned

zcash_client_backend/src/data_api/testing/pool.rs
nuttycom: These tests have been reworked and simplified.

zcash_client_sqlite/src/wallet.rs
nuttycom: Done.

zcash_client_memory/src/types/scanning.rs
nuttycom: We're not maintaining `zcash_client_memory` where they have persisted data; this fixme needs to be left in for ChainSafe to take care of.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +427,171 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+/// Records `witness_anchor_stable` on received notes whose witness is durably constructable given
+/// the ability to construct an anchor and scanning invariants.
+///
+/// The stored value is the note's **anchor floor**: the lowest anchor height at which the wallet
+/// has the data it needs to construct this note's witness. Once written it is monotonically
+/// non-decreasing: the anchor floor advances over a row's lifetime under two cases: first, when
+/// the active-shard floor is promoted to a completed-shard floor, and secondly, for notes in the
+/// chain-tip shard, when the contiguously scanned region containing the note grows toward an
+/// anchor. In the chain-tip shard only, chain reorgs and explicit rewind operations may cause
+/// the stored value to decrease.
```
nuttycom: In the cases that the height is a shard-end height, it's not really correct to refer to it as an anchor height (though it would be possible to use and refer to that anchor).

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +427,171 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
+        // the first-time stabilize step in the same call).
+        let sql = format!(
+            "UPDATE {pool}_received_notes AS rn
+             SET witness_anchor_stable = shard.subtree_end_height
+             FROM {pool}_tree_shards AS shard
+             WHERE rn.witness_anchor_stable IS NOT NULL
+               AND rn.commitment_tree_position IS NOT NULL
+               AND rn.witness_anchor_stable < shard.subtree_end_height
+               AND shard.subtree_end_height IS NOT NULL
+               AND shard.subtree_end_height <= :pruning_floor
```
nuttycom, line 616:
```suggestion
             WHERE rn.witness_anchor_stable < shard.subtree_end_height
               AND rn.commitment_tree_position IS NOT NULL
               AND shard.subtree_end_height <= :pruning_floor
```

Comparison to NULL is always falsy.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: This shouldn't be too hot, right? It's once per scanned range. But in any case, I looked into trying to consolidate these queries and it gets a bit hairy, not worth trying to optimize IMO.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -66,6 +66,42 @@ pub(crate) fn pruning_floor(chain_tip: BlockHeight) -> BlockHeight {
...
+) -> Result<bool, SqliteClientError> {
+    let pruning_floor_u32 = u32::from(pruning_floor(chain_tip));
+    let chain_end = u32::from(chain_tip) + 1;
+    let anchor_code = priority_code(&ScanPriority::Anchor);
+    conn.query_row(
+        "SELECT NOT EXISTS(
+             SELECT 1 FROM scan_queue
+             WHERE block_range_start < :chain_end
+               AND block_range_end > :pruning_floor
+               AND priority = :anchor_priority
```
nuttycom: This should be an inequality comparing against the `Scanned` priority, not an exact match with the anchor priority. We're trying to confirm that there's not a scanned range, not that there's not an anchor range.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -66,6 +66,42 @@ pub(crate) fn pruning_floor(chain_tip: BlockHeight) -> BlockHeight {
...
+/// wallet stamps on the pruning window after operations that disturb the anchor
+/// (rewind, truncate, account import), and as long as any such range remains
+/// unprocessed, the cap state needed to witness stabilized notes can't be reliably
+/// reconstructed.
+///
+/// Non-`Anchor` higher priorities (`Verify`, `ChainTip`, `Historic`, etc.) represent
+/// ordinary forward-sync work and don't gate this check — a wallet with a partial-sync
+/// gap below the anchor can still spend stabilized notes whose shard is itself
+/// scan-clean.
+pub(crate) fn tip_window_fully_scanned(
```
nuttycom: Naming nit: `prunable_window_fully_scanned`; we want to avoid confusion with the chain tip shard, which may be a superset or subset of the prunable window.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: https://github.com/zcash/librustzcash/pull/2333/commits/b90a6e93df6ebd961ecf23b4fd9230ebdff51a41

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Now handled by the `COALESCE` in `first_time_stabilize` (scanning.rs:534-552): if the containing shard has no `subtree_end_height` (or it sits above the pruning floor), the column falls through to the transaction's block height — the active/tip-shard interpretation. So tip-shard anchors are stabilized.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Rephrased — the current doc reads "when the contiguously scanned region **containing the note** grows toward an anchor" (scanning.rs:478).

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Renamed throughout: "anchor floor" / "stability height" → **"anchor-stable height"** (matching the column name `witness_anchor_stable`). The shard's `subtree_end_height` is still called by its name, and the docstring now explicitly notes that even though it is the *highest* height in the shard, it is the *lower* bound on anchors that can witness the note.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Addressed by renaming the stored value to "anchor-stable height" throughout (matches the column name). The shard's `subtree_end_height` is referred to by its name; the doc now clarifies that it serves as the anchor-stable height for completed shards because it is the *lower bound* on anchors that can witness the note (even though it's the highest height in the shard).

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: `WHERE witness_anchor_stable IS NULL` is present in the current first-time arm (scanning.rs:549), so notes that already have a stored value won't match this UPDATE — they're handled by the promotion arm.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Applied — dropped the redundant `IS NOT NULL` checks since the inequality comparisons (`<` and `<=`) already make NULL fall out of the WHERE.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Renamed to `prunable_window_fully_scanned`. Callers and docstring updated.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Fixed — the predicate is now `priority > :scanned_priority` (anything with pending work in the window blocks spendability, not just `Anchor`). Docstring updated to match.

zcash_client_sqlite/src/wallet/common.rs
nuttycom: Done in b90a6e93df. The predicate now lives in a single helper, `common::is_note_spendable_at_anchor` (common.rs:319), and is called from balance reporting (wallet.rs:2586) and from both note-selection paths (common.rs:509 and common.rs:722). `anchor_available` is part of the helper's signature, so the divergence you noted between balance and selection is also resolved — both call sites now pass it.

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: The stability of notes in the tip shard is essentially a continuously maintained invariant that the interval between the note and the chain tip has been scanned at least once, and there are no unscanned blocks in the "prunable" range.

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/d1c026c87a8246d1d3f2101eb234e9857214f06b..4ce2ab5d3a4aa11be914d6ba0b259e9b64c80a18) to fix an off-by-one error

zcash_client_backend/CHANGELOG.md
```diff
@@ -12,6 +12,24 @@ workspace.
 
 ### Added
 - `zcash_client_backend::data_api::error::RewindError`
+- `zcash_client_backend::data_api::scanning::ScanPriority::Anchor`, slotted
+  between `ChainTip` and `Verify`. Backends stamp `scan_queue` ranges with
+  this priority after operations that disturb the wallet's anchor; the
+  spendability rule treats any `Anchor`-priority range overlapping the
+  chain-tip pruning window as blocking.
+- The following utilities for fake note commitment tree construction have been
```
nullcopy, line 317: Do we typically document test utilities in the CHANGELOG?

zcash_client_backend/src/data_api/testing/shard_stub.rs
```diff
@@ -0,0 +1,353 @@
...
+
+/// The fixed Merkle tree depth used by both Sapling and Orchard note
+/// commitment trees.
+pub const NOTE_COMMITMENT_TREE_DEPTH: u8 = 32;
+
+/// Decomposes the half-open range `[start, end)` of leaf positions into the
+/// minimal set of aligned subtree addresses that exactly cover the range,
+/// capping each subtree at level `max_level`.
+///
+/// Each returned [`Address`] identifies a complete subtree at some
```
nullcopy: s/complete subtree/subtree/

Not sure what 'complete' means in this context

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -55,7 +61,44 @@ impl<P: consensus::Parameters> RusqliteMigration for Migration<P> {
...
         // Backfill: Identify any notes which have stable witness data, and mark them as such.
-        mark_stabilized_notes(transaction, &self.params)?;
+        // The SQL is inlined here so this migration remains stable even if
+        // `crate::wallet::scanning::mark_stabilized_notes` later evolves to write a
+        // different column.
+        if let Some(max_scanned_height) =
+            block_max_scanned(transaction, &self.params)?.map(|m| m.block_height())
+        {
+            let pruning_floor: u32 =
+                u32::from(max_scanned_height).saturating_sub(PRUNING_DEPTH - 1);
```
nullcopy, line 80: Should use `pruning_floor()` helper here, right? Or is this also intentionally inlined?

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -587,7 +631,9 @@ mod tests {
 
         let read_stabilized = |conn: &rusqlite::Connection, table: &str, pk_col: &str| -> i64 {
             conn.query_row(
-                &format!("SELECT witness_stabilized FROM {table} WHERE {pk_col} = 0"),
+                &format!(
+                    "SELECT witness_anchor_stable IS NOT NULL FROM {table} WHERE {pk_col} = 0"
```
nullcopy: Was this change intentional? `witness_anchor_stable` doesn't exist yet, as of this migration.

zcash_client_sqlite/src/wallet/scanning.rs
nullcopy: This should use the pruning floor helper

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -53,11 +54,60 @@ pub(crate) fn parse_priority_code(code: i64) -> Option<ScanPriority> {
...
+    };
+
+    let chain_tip_u32 = u32::from(chain_tip);
+    let pruning_floor = chain_tip_u32.saturating_sub(PRUNING_DEPTH);
+    let birthday_u32 = u32::from(wallet_birthday);
+    // The window's inclusive lower bound is the larger of `pruning_floor + 1` and
+    // `wallet_birthday`; we never stamp Anchor priority on pre-birthday blocks (which are
+    // `Ignored` and must stay that way).
+    let window_start = std::cmp::max(pruning_floor + 1, birthday_u32);
+    let window_end_exclusive = chain_tip_u32 + 1;
```
nullcopy: doesn't +1 make it _inclusive_ rather than exclusive?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +427,171 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
-/// fully scanned and whose shard end height has received at least `PRUNING_DEPTH`
-/// confirmations. The stored value is the shard's `subtree_end_height` — the anchor at which
-/// the leaf-level path from the note to the shard root has been finalized.
+/// Records `witness_anchor_stable` on received notes whose witness is durably constructable given
+/// the ability to construct an anchor and scanning invariants.
+///
+/// The stored value is the note's **anchor floor**: the lowest anchor height at which the wallet
+/// has the data it needs to construct this note's witness. Once written it is monotonically
+/// non-decreasing: the anchor floor advances over a row's lifetime under two cases: first, when
+/// the active-shard floor is promoted to a completed-shard floor, and secondly, for notes in the
```
nullcopy: Again, not sure what is meant by 'active-shard floor' or 'completed-shard floor' here. Does this just mean the point at which the tip shard has been filled and a new shard begins?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -421,67 +427,171 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
...
-        let pruning_floor: u32 = u32::from(max_scanned_height).saturating_sub(PRUNING_DEPTH - 1);
+    fn promote_on_completion(
+        conn: &rusqlite::Transaction<'_>,
+        pool: &str,
+        shard_height: u8,
+        pruning_floor: u32,
+    ) -> Result<(), SqliteClientError> {
+        // For a note that already has a stored floor, advance it to the shard's
+        // `subtree_end_height` if the shard has now reached the completed-and-
+        // stabilized state and the stored floor sits below that height. The join to
```
nullcopy: I'm curious, did you come up with a good rationale for promoting the stabilization height? I don't see anything wrong with it, but I'm not convinced its necessary either.

nullcopy: As an interesting data point, all tests still pass with the promotion arm disabled. Not sure if that's evidence of this being unnecessary or a gap in test coverage, though 🤷

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -66,6 +66,42 @@ pub(crate) fn pruning_floor(chain_tip: BlockHeight) -> BlockHeight {
     BlockHeight::from(u32::from(chain_tip).saturating_sub(PRUNING_DEPTH))
 }
 
+/// Returns true when no `scan_queue` range in the chain-tip pruning window has
+/// [`ScanPriority::Anchor`] priority. This is the wallet-level precondition for treating
```
nullcopy: > Returns true when no `scan_queue` range in the chain-tip pruning window has [`ScanPriority::Anchor`] priority

or higher, right? Completed ranges within the pruning window will obviously have lower than anchor priority

nullcopy: nvm, I see this gets removed in a later commit

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -295,6 +269,66 @@ impl NoteRequest {
...
+    conn: &Connection,
+    anchor_height: BlockHeight,
+    protocol: ShieldedProtocol,
+) -> Result<bool, SqliteClientError> {
+    let table_prefix = match protocol {
+        ShieldedProtocol::Sapling => SAPLING_TABLES_PREFIX,
+        #[cfg(feature = "orchard")]
+        ShieldedProtocol::Orchard => ORCHARD_TABLES_PREFIX,
+        #[cfg(not(feature = "orchard"))]
+        ShieldedProtocol::Orchard => return Ok(true),
```
nullcopy: I see why you did it this way, but IMO it's misleading saying the frontier is available in the orchard pool, even when the orchard pool isn't enabled.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -295,6 +269,66 @@ impl NoteRequest {
...
+                SqliteClientError::CommitmentTree(shardtree::error::ShardTreeError::Storage(e))
+            })?
+            .is_some(),
+    )
+}
+
+/// Encodes the wallet's spendability rule for a single note, given the four orthogonal
+/// predicates that compose it. All four must hold for the note to be spendable.
+///
+/// 1. **Stored floor at or below the chosen anchor.** `witness_anchor_stable` (the note's
```
nullcopy, line 329: What is a 'stored floor'? Does this just mean the note was stored at or below the chosen anchor?

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -589,15 +618,13 @@ where
...
-             -- A stabilized note's witness is durable across rewinds, so it bypasses
-             -- the scan-state gating
-             AND (
-                 rn.witness_anchor_stable IS NOT NULL
-                 OR (
-                     :tip_unscanned = 0 -- the tip shard has no unscanned ranges
-                     AND scan_state.max_priority <= :scanned_priority -- the note shard is fully scanned or ignored
-                 )
-             )
+             -- The stored floor must lie at or below the chosen anchor, and the
```
nullcopy: s/stored floor/stored anchor floor/

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -295,6 +269,66 @@ impl NoteRequest {
...
+/// Encodes the wallet's spendability rule for a single note, given the four orthogonal
+/// predicates that compose it. All four must hold for the note to be spendable.
+///
+/// 1. **Stored floor at or below the chosen anchor.** `witness_anchor_stable` (the note's
+///    *anchor floor* — the lowest anchor height for which the wallet has the data needed
+///    to construct this note's witness) must be set, and must lie at or below the chosen
+///    anchor height.
+/// 2. **Tip-window fully scanned.** The chain-tip pruning window must contain no
+///    `scan_queue` range with [`ScanPriority::Anchor`] priority. This is a wallet-state-wide
+///    precondition that the caller computes once per call.
```
nullcopy: Criteria 2 says no `ScanPriority::Anchor` range may exist in the pruning window (this language is repeated elsewhere throughout the PR), but we always use `prunable_window_fully_scanned()`, which actually requires no scan range with scan priority greater than `ScanPriority::Scanned` in the pruning window (more strict than documented).

I think the code is correct and perhaps the documentation just needs rephrasing?

nullcopy (changes requested):
Reviewed as of 4ce2ab5d3a4aa11be914d6ba0b259e9b64c80a18

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: Instead of using the chain tip as an inclusive end height, we add one and obtain an exclusive end height.

zcash_client_backend/CHANGELOG.md
nuttycom: The way I think about this is that additions that are available for others to use are worth mentioning; these are technically public APIs under the `test-dependencies` feature; it's mostly that I don't think we have to be rigorous about it. This CHANGELOG entry could be more concise.

nullcopy: That makes sense. I'm on board

zcash_client_sqlite/src/wallet/scanning.rs
nullcopy: Ah, I see what you mean now. I interpreted `window_end_exclusive` to mean "end of the window range excluding the chain tip", but I think you mean it as "the end position to be used with an exclusive range, such that the chain tip is still included"

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -333,6 +311,133 @@ impl NoteRequest {
     }
 }
 
+/// Returns whether the wallet's shardtree has a frontier at the given anchor height for
+/// the specified pool — i.e. whether a tree-state checkpoint exists from which a witness
+/// against this anchor can be reconstructed.
+///
+/// After a deep rewind that drops the chain tip below the lowest surviving tree
+/// checkpoint, no checkpoint exists at the anchor height and the wallet treats the
+/// anchor as unavailable for spendability purposes.
```
nuttycom, line 320: Is there a corner case here where if more than 100 blocks go by without any blocks including shielded components, all shielded notes become unspendable? We should always be able to create a checkpoint at an inserted frontier that's within the pruning window, to avoid this.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -333,6 +311,133 @@ impl NoteRequest {
...
+            .is_some(),
+    )
+}
+
+/// Encodes the wallet's spendability rule for a single note, given the orthogonal predicates
+/// that compose it. All must hold for the note to be spendable.
+///
+/// 1. **Stored floor at or below the chosen anchor.** `witness_anchor_stable` (the note's
+///    *anchor floor* — the lowest anchor height for which the wallet has the data needed to
+///    construct this note's witness) must be set, and must lie at or below the chosen anchor.
```
nuttycom: Describing `witness_anchor_stable` as the lower anchor height for which the wallet has the data needed to construct the witness makes it seem like the degenerate case of just treating the block height in which the note was mined would be valid? That misrepresents the usual case, which is that the height of the end of a fully-scanned shard containing the note's commitment is the stable height. Likewise, in the chain tip shard case, the stable height should be the maximum height for which all succeeding blocks are scanned.

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -333,6 +311,133 @@ impl NoteRequest {
...
+/// 3. **Witness region below the window is durable.** Either the note's shard is scan-clean —
+///    every block in its range is scanned, so every leaf after the note within the shard is
+///    present and the path to the shard root can be built regardless of gaps elsewhere below
+///    the window — or its stored floor reaches the bottom of the window: no `scan_queue` range
+///    above `Scanned` separates `witness_anchor_stable` from the window. `pruning_region_gap_top`
+///    (the top of the highest such gap, computed once per call) encodes that boundary; a note in
+///    a shard that is not scan-clean is durable iff its floor lies at or above it. Cleanliness
+///    is judged from the scan queue, never from the subtree-root table: a root reported by the
+///    server says nothing about which of the shard's leaves the wallet holds. Use
+///    [`is_shard_scan_clean`] on the shard's `v_*_shards_scan_state.max_priority`.
```
nuttycom, line 373: This is also not exactly the condition that I was expecting. The important condition is that the witness be constructable to the root of the shard. This is achieved right now when the range between the block the note was received in and the end of the shard is fully scanned; the left side of the witness is then constructable from the prior frontier that was inserted at the time of scanning, and the right side of the witness during the scanning of the remainder of the shard.

zcash_client_sqlite/src/wallet/common.rs
nuttycom, line 607: Instead of the start position, can we compare to the position of the note itself? That then would ignore the scanned-ness of previous positions in the shard, which are irrelevatn.

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -650,11 +655,19 @@ where
 ///   boundary block containing no shielded outputs in any pool would otherwise leave a permanent
 ///   hole in the retained grid, and the anchor there could never be proved against. `None`
 ///   disables anchor retention.
+/// - `chain_tip`: The wallet's view of the chain tip, if known. `Checkpoint` retention is used
+///   only within `PRUNING_DEPTH` blocks of the later of this height and the last block in
```
nuttycom:
```suggestion
///   only within `PRUNING_DEPTH` blocks of the greater of this height and the last block in
```

PR #2338, opened 2026-05-06
Fix incorrect progress computation after rewinding to an earlier birthday.

nuttycom:
This PR is best reviewed hiding whitespace.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -5373,4 +5373,333 @@ mod tests {
             first = initial_height,
         );
     }
+
+    #[test]
+    fn rewound_birthday_slow_path_recovery_stays_below_one() {
+        use std::num::NonZeroU8;
+
+        use incrementalmerkletree::frontier::Frontier;
+        use zcash_client_backend::data_api::{
+            chain::{ChainState, CommitmentTreeRoot},
+            testing::{
+                InitialChainState, pool::ShieldedPoolTester, sapling::SaplingPoolTester,
+            },
+        };
+        use zcash_protocol::{
+            ShieldedProtocol,
+            consensus::{NetworkUpgrade, Parameters},
+        };
+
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -2176,216 +2173,186 @@ fn subtree_scan_progress<P: consensus::Parameters>(
         FROM blocks
         WHERE height = :start_height",
     ))?;
-    let mut stmt_end_tree_size_at = conn.prepare_cached(&format!(
```
nullcopy, line 2179: I'm curious why this got inlined? It seemed like a helpful abstraction IMO

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2144,9 +2142,8 @@ fn subtree_scan_progress<P: consensus::Parameters>(
     params: &P,
     shielded_protocol: ShieldedProtocol,
     pool_activation_height: BlockHeight,
-    birthday_height: BlockHeight,
+    min_birthday_height: BlockHeight,
```
nullcopy, line 2145: (nit) would be nice to update the name in the trait signature too

zcash_client_sqlite/src/wallet.rs
nullcopy, line 2139: I wonder if this `allow` is still necessary

nullcopy (approved):
utACK c56e1493b44d9483caa78fddb79dd3e74950b53c

Minor comments, but none are blocking

zcash_client_sqlite/src/wallet.rs
nuttycom: It was only used once, and it's easier to read inline.

zcash_client_sqlite/src/wallet.rs
nuttycom: Good point, I'll fix that in a follow-up.

PR #2341, opened 2026-05-06
zcash_client_backend: Replace `rewind_to_height` with `rewind_to_chain_state`

nuttycom:
The `rewind_to_height` function did not provide sufficient information to update wallet account birthday metadata, which has the potential to break a lot of invariants in the wallet that assume correct birthday height information. This PR repairs that problem, and additionally reuses the `rewind_to_chain_state` logic to effect rewinds that occur on account addition; previously, this used a strategy of resetting the scan queue without altering note commitment tree data, but this introduced a potential failure due to possible block and note commitment tree conflicts under reorg scenarios. Using `rewind_to_chain_state` consistently avoids this potential issue.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -4688,18 +4688,20 @@ pub fn rewind_to_chain_state_deep<T: ShieldedPoolTester, Dsf>(
         )
         .expect("rewind_to_chain_state should succeed for a deep target");
 
-    // The chain tip (derived from scan_queue) should now report the rewind target.
+    // The chain tip (derived from scan_queue) should still report the pre-rewind tip:
+    // `rewind_to_chain_state` overwrites the scan-queue range above the rewind target
+    // with a `Historic` rescan range that extends up to the pre-rewind tip.
     let new_tip = st
         .wallet()
         .chain_height()
         .unwrap()
-        .expect("chain tip should match rewind target");
-    assert_eq!(new_tip, rewind_target);
+        .expect("chain tip should still be set after rewind");
+    assert_eq!(new_tip, pre_rewind_tip);
 
-    // A deep rewind keeps the scan queue rewound to the target, while blocks,
-    // transactions, tx_locator_map entries, and note commitment trees are only rewound
-    // to the oldest retained checkpoint at `tip - (PRUNING_DEPTH - 1)`. Data at that
-    // boundary is kept (so stabilized notes remain spendable); data above it is removed.
+    // A deep rewind preserves block, transaction, tx_locator_map, and note commitment tree
+    // data only as far back as the oldest retained checkpoint at `tip - (PRUNING_DEPTH - 1)`.
+    // Data at that boundary is kept (so stabilized notes remain spendable); data above it is
+    // removed.
     let wallet = st.wallet();
     assert_eq!(
         wallet.get_block_hash(prune_boundary).unwrap(),
@@ -4802,12 +4804,15 @@ pub fn rewind_to_chain_state_shallow<T: ShieldedPoolTester, Dsf>(
         )
         .expect("rewind_to_chain_state should succeed for a shallow target");
 
+    // The chain tip (derived from scan_queue) should still report the pre-rewind tip:
+    // `rewind_to_chain_state` overwrites the scan-queue range above the rewind target with
+    // a `Historic` rescan range that extends up to the pre-rewind tip.
     let new_tip = st
         .wallet()
         .chain_height()
         .unwrap()
-        .expect("chain tip should match rewind target");
-    assert_eq!(new_tip, rewind_target);
+        .expect("chain tip should still be set after rewind");
+    assert_eq!(new_tip, tip);
 
     // A shallow rewind truncates blocks, tx_locator_map, and note commitment trees
     // directly to the rewind target: data at the target is preserved (with the same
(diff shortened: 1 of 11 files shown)


zcash_client_sqlite/src/lib.rs
```diff
@@ -1332,8 +1332,29 @@ impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R:
         self.transactionally(|wdb| wdb.truncate_to_chain_state(chain_state))
     }
 
-    fn rewind_to_height(&mut self, max_height: BlockHeight) -> Result<BlockHeight, Self::Error> {
-        self.transactionally(|wdb| wdb.rewind_to_height(max_height))
```
nullcopy, line 1336: If you implement `From<rusqlite::Error>` for `RewindError`, then I believe you could still use the `transactionally()` helper (which is a nice abstraction IMO).

```rust
  impl From<rusqlite::Error> for RewindError<AccountUuid, SqliteClientError> {
      fn from(e: rusqlite::Error) -> Self {
          RewindError::DataSource(e.into())
      }
  }
```

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3878,92 +3877,181 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
...
-            params
-                .activation_height(NetworkUpgrade::Sapling)
-                // Fall back to the genesis block in regtest mode.
-                .map_or(BlockHeight::from_u32(0), |h| h - 1)
-        });
+    chain_state: &ChainState,
+    reset_account_birthdays: HashSet<AccountUuid>,
+) -> Result<(), RewindError<AccountUuid, SqliteClientError>> {
+    // Find account birthdays. We need to check that we either we have an account with a birthday
+    // less than or equal to the new birthday height, or that there is an intersection between the
```
nullcopy: Comment fix: Not just an intersection. `reset_account_birthdays` should be a pure subset of the accounts in the wallet.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3878,92 +3877,181 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
...
+            "Account UUIDs provided for birthday reset do not exist in the wallet database."
+                .to_string(),
+        )));
+    }
 
-    #[cfg(feature = "orchard")]
+    let target_height = chain_state.block_height();
+    let new_birthday = target_height + 1;
+    let birthday_reset_required =
+        account_birthdays.values().all(|b| b > &new_birthday) && reset_account_birthdays.is_empty();
```
nullcopy, line 3951: I don't understand the right side `&& reset_account_birthdays.is_empty()` here. What if `reset_account_birthdays` is not empty, but the offending account is not one of the provided accounts?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3878,92 +3877,181 @@ pub(crate) fn truncate_to_chain_state<P: consensus::Parameters, CL, R>(
...
-            truncation_target,
-        )
-        .map_err(ShardTreeError::Storage)?;
-        if orchard_window_floor != sapling_window_floor {
-            return Err(SqliteClientError::CorruptedData(
-                "Sapling and Orchard should have the same checkpoints".into(),
-            ));
+        if target_height < max_scanned_height {
+            // Compute the floor height of the pruning window.
+            let pruning_floor = max_scanned_height.saturating_sub(PRUNING_DEPTH - 1);
```
nullcopy, line 3970: Didn't one of your other PRs add a helper function to compute the pruning floor recently?

nullcopy: Oh, that may be in 2333... might be worth pulling into this PR?

zcash_client_backend/src/data_api/testing/pool.rs
nullcopy, line 1: Related to my comment about skipping reset checks when `reset_account_birthdays` is not empty: I would expect tests here to exercise various account birthday checks. Some that come to mind:
* rewind should error if `reset_account_birthdays` is empty, but an account fails the birthday test
* rewind should error if `reset_account_birthdays` is **not** empty, but an account **not in the reset list** fails the birthday test
* rewind should error if `reset_account_birthdays` is contains an `account_uuid` not in our wallet

zcash_client_sqlite/src/wallet.rs
```diff
@@ -572,61 +572,36 @@ pub(crate) fn add_account<P: consensus::Parameters>(
...
+        birthday.prior_chain_state(),
+        std::iter::once(account_uuid).collect(),
+    ) {
+        Ok(()) => {}
+        Err(RewindError::DataSource(e)) => return Err(e),
+        Err(RewindError::RewindBeyondBirthdays(_)) => {
+            // Cannot occur: `reset_account_birthdays` is non-empty (it contains the new
+            // account), so `rewind_to_chain_state`'s contract specifies that this variant is
+            // not returned.
+            unreachable!(
```
nullcopy, line 601: This comment is correct, in that the birthday check is bypassed when `reset_account_birthdays` is not empty, but I'd argue that is a bug. In practice, I would expect this _could_ be reachable if there already exists an account with a higher birthday than this new account.

nullcopy (changes requested):
I have some questions, as of 3cace58d5d2ca6da3cb274ce783d6bf986a29f73

zcash_client_sqlite/src/wallet.rs
nuttycom: We've already cleared the `all(|b| b > &new_birthday)` gate, so every account in `reset_account_birthdays` birthday height is known to be greater than `new_birthday`, if it's nonempty. If any account exists that has `b <= new_birthday`, then we don't *need* to reset any birthday height; the account birthdays are all correct and we're just rescanning because we added an account with a birthday height greater than or equal to the rewind height. So it's only when both conditions are satisfied that we require a birthday reset to be specified.

zcash_client_sqlite/src/wallet.rs
nuttycom: Probably not worth it here.

zcash_client_sqlite/src/wallet.rs
nuttycom: see https://github.com/zcash/librustzcash/pull/2341#discussion_r3205124335

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5730,4 +5736,101 @@ mod tests {
...
+            account_birthday(st.wallet().conn(), account_id),
+            Ok(b) if b == target_height + 1
+        );
+    }
+
+    /// `rewind_to_chain_state` must reject `reset_account_birthdays` containing an
+    /// `AccountUuid` that does not correspond to an account in the wallet, surfacing the
+    /// error via `RewindError::DataSource(CorruptedData)`.
+    #[test]
+    fn rewind_to_chain_state_with_unknown_uuid_in_reset_returns_data_source_error() {
```
nullcopy: Not worth blocking, but one other case that would be good to test, is: `reset_account_birthdays` contains a valid UUID, but its not the account which had a conflict.

zcash_client_sqlite/src/wallet.rs
nullcopy: Ahhh, I see what I was missing. I was mistakenly thinking that _all_ accounts had to be below the new birthday, but I'm remembering now that we only require _at least one_ account to be below the new birthday.

nullcopy (approved):
utACK 777200b02c2af153b8cf6bd571bfbcb760d3d314

PR #2343, opened 2026-05-07
Fix computation of sync and recovery progress to account for rewinds that leave blocks table entries in place.

nuttycom:


zcash_client_sqlite/src/wallet.rs
```diff
@@ -2153,35 +2153,58 @@ fn subtree_scan_progress<P: consensus::Parameters>(
...
 
+    // Each query against the `blocks` table that contributes to scan-progress accounting
+    // must exclude heights that fall within a `scan_queue` range whose priority indicates
+    // the range is pending re-scan. Without this filter, blocks whose tree state was
+    // recorded by a previous scan but whose enclosing range was subsequently re-queued
+    // (e.g., by `rewind_to_chain_state`) would be counted as scanned, over-reporting
+    // progress against a tree-size denominator that no longer reflects the wallet's
+    // actual scanned state.
+    let scanned_priority = priority_code(&ScanPriority::Scanned);
+    let unscanned_filter = "AND NOT EXISTS (
```
nullcopy, line 2164: (nit) you call this `unscanned_filter`, but allow the caller to specify whichever scan priority they want to filter by. I don't think it's practically a problem, but a possible opportunity for confusion in the future

nullcopy (approved):
utACK 4059803bed8508a7d8cd143019716f74d1affd1a

PR #2345, opened 2026-05-08
Add V5_Qr transaction selector for Orchard change

ValarDragon:
## Summary

- update the Orchard dependency to the ZIP 2005 QR note support branch
- add Builder::add_versioned_orchard_output so callers can select an Orchard note version per output
- add TxVersion::V5_Qr as a builder-level selector for ZIP 2005 QR Orchard change notes
- keep TxVersion::None / suggested_for_branch selecting V5, not V5_Qr
- route external Orchard outputs through the existing add_orchard_output path, while internal change can use add_versioned_orchard_output with NoteVersion::V3_Qr

## API consumer guidance

This API is behind the unstable feature. Consumers that want QR Orchard change notes should build with unstable enabled and set the proposed transaction version to TxVersion::V5_Qr for non-Keystone transactions.

For Keystone created transactions, callers should continue using TxVersion::V5. Keystone users must stay on plain V5 until keystone team publishes firmware that accepts the Orchard dependency bump / QR note support needed for V5_Qr change notes.

## Notes

V5_Qr is serialized, hashed, signed, and consensus-treated as V5. It is a construction-time selector used by wallet transaction creation. For now it affects Orchard change notes only; the enum comment leaves room for a future upgrade to support QR sends as well.

## Tests

- cargo fmt --check
- cargo test -p zcash_primitives suggested_version_for_v5_branches_is_not_qr
- cargo check -p zcash_primitives -p zcash_client_backend --features zcash_client_backend/unstable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2966,9 +2966,8 @@ checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
 
 [[package]]
 name = "orchard"
-version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a36733033adae631acacf46f542f027d508d9025d14797d9b35059494c99eb3"
+version = "0.13.1"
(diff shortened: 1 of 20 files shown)


daira:
Should this be closed? We are not doing QR (for the Orchard protocol) outside the Ironwood pool.

ValarDragon:
Yes, thank you!

PR #2347, opened 2026-05-08
Fix shielding failure with >=150 P2PKH UTXOs

### LINKED ISSUE #2346: Shielding fails with `ChangeRequired(5000)` when spending >=150 P2PKH UTXOs
Reported via [MOB-1147](https://linear.app/zodl/issue/MOB-1147): users on Android Zashi/Zodl hit the following error when shielding a transparent balance composed of many small UTXOs (e.g., regular mining payouts):

```
TransactionNotCreatedException(rootCause=java.lang.RuntimeException: Error while creating transactions:
  An error occurred building the transaction: The transaction requires an additional change output of ZatBalance(5000) zatoshis)
```

## Root cause

Proposal-time and builder-time ZIP 317 fee computation disagree on the per-input size for P2PKH inputs:

* The proposal layer (`WalletTransparentOutput::serialized_size` in `zcash_client_backend`) reports `STANDARD_P2PKH = 150` bytes per P2PKH input — the ZIP 317 standard upper bound.
* The builder layer (`TransparentInputInfo::serialized_size` in `zcash_primitives`) delegates to `serialized_len()`, which returns the exact serialized size of `149` bytes for P2PKH (PREVOUT 36 + script_sig 109 + sequence 4).

ZIP 317 derives `logical_actions` from `ceildiv(t_in_total_size, 150)`. With N P2PKH inputs:

* Proposal: `ceildiv(N * 150, 150) = N`
* Builder:  `ceildiv(N * 149, 150) = N - floor(N/150)` (when N is not a multiple of 150)

The two computations diverge starting at `N = 150` and grow by one logical action per additional 150 inputs. The proposal earmarks fee for one extra logical action that the builder does not include, and the builder's `balance_after_fees > 0` check rejects the transaction with `Error::ChangeRequired(5000)` (one marginal fee).

## User impact

Affected users: anyone shielding a transparent balance with ≥150 P2PKH UTXOs. Real-world reports came from Android users with \~216 mining-payout UTXOs (2 reports via Help Scout [#427](<https://github.com/zcash/librustzcash/issues/427>) and zcash/librustzcash#430).

## Fix

Align `TransparentInputInfo::serialized_size` with the proposal layer and the ZIP 317 standard: return `STANDARD_P2PKH` (150) for P2PKH inputs and continue using `serialized_len()` for P2SH (where the size depends on the redeem script).

A regression test (`shielding_many_transparent_utxos`) shielding 160 P2PKH UTXOs reproduces the failure pre-fix and passes post-fix.

nuttycom:
## Summary

Fixes #2346.

The proposal layer computes ZIP 317 fees using `STANDARD_P2PKH = 150` bytes per P2PKH input (via `WalletTransparentOutput::serialized_size`), but the builder was computing them using `TransparentInputInfo::serialized_len() = 149` bytes per P2PKH input. The two computations diverge by one logical action at every multiple of 150 inputs because `ceildiv(t_in_total_size, 150)` differs for `150*N` vs `149*N`. The proposal earmarks fee for one extra logical action that the builder does not include in its own fee computation, so the builder rejects the transaction with `Error::ChangeRequired(5000)`.

The user reports cited in the linked issue had ~216 transparent UTXOs from regular mining payouts, reproducing the 5000-zatoshi gap exactly.

### Commits

1. **`test: gate zcash_transparent builder tests on transparent-inputs`** — prerequisite: the `tests` module references `secp256k1`, `add_p2pkh_input`, and `prepare_transparent_signatures`, all gated behind `transparent-inputs`. Without the feature, `cargo test -p zcash_transparent` failed to build on `main`.
2. **`test: add regression test for shielding many P2PKH UTXOs`** — adds `shielding_many_transparent_utxos` (160 UTXOs) which fails on the parent commit with the exact error users report.
3. **`fix: align TransparentInputInfo P2PKH size with ZIP 317 standard`** — `TransparentInputInfo::serialized_size()` now returns `STANDARD_P2PKH` for P2PKH (matching the proposal layer and ZIP 317). P2SH continues to use `serialized_len()`.

## Test plan
- [x] `cargo test -p zcash_client_sqlite --features=transparent-inputs shielding_many_transparent_utxos` (new regression test passes after fix; fails before)
- [x] `cargo test -p zcash_client_sqlite --features=transparent-inputs shield` (existing shielding tests still pass)
- [x] `cargo test -p zcash_client_backend --features=transparent-inputs --lib` (all green)
- [x] `cargo test -p zcash_transparent` and `cargo test -p zcash_transparent --features=transparent-inputs` (both green; previously the default-feature run wouldn't build)
- [x] Verified each commit individually builds and the regression test transitions from failing → passing at the fix commit.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/transparent.rs
+++ b/zcash_client_backend/src/data_api/testing/transparent.rs
@@ -333,6 +333,97 @@ where
     );
 }
 
+/// Regression test for [PRO-291]: shielding a transparent balance composed of many P2PKH
+/// UTXOs must not fail with `ChangeRequired` due to fee disagreement between the proposal
+/// and builder layers.
+///
+/// At 150 P2PKH inputs the proposal-time fee computation (which uses
+/// `STANDARD_P2PKH = 150` bytes per input) starts to diverge from the builder-time
+/// fee computation (which historically used the actual serialized size, 149 bytes per
+/// input) due to the `ceildiv(t_in_total_size, 150)` term in the ZIP 317 fee formula.
+/// The discrepancy grows by one logical action (5000 zats) for every additional 150
+/// inputs.
+///
+/// [PRO-291]: https://linear.app/zodl/issue/PRO-291
+pub fn shielding_many_transparent_utxos<DSF>(dsf: DSF, cache: impl TestCache)
+where
+    DSF: DataStoreFactory,
+{
+    // Choose enough UTXOs to cross the first ceildiv(_, 150) boundary.
+    const NUM_UTXOS: usize = 160;
+    // Per-UTXO value comfortably above the marginal fee so none are treated as dust.
+    const PER_UTXO: u64 = 100_000;
+
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(dsf)
+        .with_block_cache(cache)
+        .with_account_from_sapling_activation(BlockHash([0; 32]))
+        .build();
+
+    let account = st.test_account().cloned().unwrap();
+    let uaddr = st
+        .wallet()
+        .get_last_generated_address_matching(account.id(), UnifiedAddressRequest::AllAvailableKeys)
+        .unwrap()
+        .unwrap();
+    let taddr = uaddr.transparent().unwrap();
+
+    // Initialize the wallet with chain data that has no shielded notes for us.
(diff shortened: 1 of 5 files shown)


nullcopy (approved):
utACK 66b6f07d688302720fb44a82119b0d326aaa94f9

PR #2353, opened 2026-05-13
zcash_client_backend: propose_shielding_coinbase

schell:
## Summary

Adds a new dedicated `propose_shielding_coinbase` method (and corresponding free function) to `zcash_client_backend` for shielding coinbase transparent outputs to an arbitrary shielded recipient.

This **replaces the previous approach on this branch**, which modified `propose_shielding` to accept an optional `to_address` and `memo`. That approach was rejected during review (thanks @nullcopy @nuttycom) because it broadened the public API in a privacy-regressing way: it would have allowed callers to spend arbitrary transparent outputs (not just coinbase) to arbitrary shielded recipients, exposing the sender's prior transparent transaction graph to the recipient.

This PR has been force-pushed with a fresh implementation that takes the new approach instead.

## Design

The new `propose_shielding_coinbase` method preserves the privacy property of the prior `zcashd` `z_shieldcoinbase` behavior by:

- **Hard-coding the input filter to `TransparentOutputFilter::CoinbaseOnly`**, ensuring that only coinbase outputs (which have no prior transparent transaction graph) are eligible. The restriction is enforced at the API boundary; callers cannot opt in to selecting non-coinbase outputs through this method.
- **Requiring a shielded `to_address`** (Sapling, or a Unified Address with a shielded receiver). Transparent and TEX destinations are rejected with `ProposalError::ShieldingRequiresShieldedRecipient`. The destination is not constrained to a wallet-local account; miners may legitimately need to shield to custodial/external wallets.
- Accepting an optional **`memo`** (always permitted, since a payment is always present).
- Accepting an optional **`limit: Option<usize>`** capping the number of selected UTXOs to the highest-value `n` of those eligible, with a stable tiebreaker by outpoint. `Some(0)` selects no inputs and falls through to the standard insufficient-funds path. This folds in the previously-deferred `limit` work so that zallet's `z_shieldcoinbase` `limit` parameter can be honored.
- **Producing a proposal with no transparent change**, matching the consensus rule that coinbase-spending transactions cannot produce transparent change.

## New public API

- `wallet::propose_shielding_coinbase` (free function)
- `wallet::input_selection::ShieldingSelector::propose_shielding_coinbase` (trait method; breaking change for any downstream impl of `ShieldingSelector`, but the only impl in this repo is `GreedyInputSelector`)
- `ProposalError::ShieldingRequiresShieldedRecipient` (returned for transparent/TEX destinations)
- `ProposalError::AttemptedShieldingToTransparentAddress` (defensive internal-invariant guard; unreachable in current callers but routes the residual `PoolType::Transparent` case through a real error variant rather than `unreachable!`)

## Code reuse

`propose_shielding` was refactored to share three new private helpers with `propose_shielding_coinbase`:

- `gather_shielding_inputs`: the per-address UTXO fold with ephemeral-address linkability check.
- `resolve_shielded_destination`: address-to-pool resolution with shielded-receiver enforcement.
- `compute_shielding_balance_with_dust_retry`: the dust-input retry pattern around `compute_balance`, parameterized on an optional destination pool.

The legacy `is_shielding = true` shape produced by `propose_shielding` is preserved unchanged. `propose_shielding_coinbase` always sets `is_shielding = false` because it always carries an explicit ZIP-321 payment.

## Tests

Five new pool-generic tests in `zcash_client_backend/src/data_api/testing/pool.rs`, each wired up for both `SaplingPoolTester` and `OrchardPoolTester` in `zcash_client_sqlite` (10 sqlite test registrations total, gated on `pczt-tests + transparent-inputs`):

- `propose_shielding_coinbase_succeeds`: happy path with a shielded recipient; verifies the proposal contains exactly one transparent input, one ZIP-321 payment, no change, and that `payment_amount + fee == coinbase_value`.
- `propose_shielding_coinbase_transparent_recipient_rejected`: verifies `ProposalError::ShieldingRequiresShieldedRecipient` for a transparent destination.
- `propose_shielding_coinbase_with_memo_succeeds`: verifies the supplied memo is attached to the resulting payment.
- `propose_shielding_coinbase_with_limit_truncates_inputs`: verifies `limit = Some(2)` selects the two highest-value coinbase UTXOs out of three eligible.
- `propose_shielding_coinbase_with_zero_limit_insufficient_funds`: verifies `limit = Some(0)` returns an insufficient-funds error rather than producing a proposal.

All 258 sqlite tests pass (0 failed, 2 ignored).

## Caller impact

`shield_transparent_funds` and `TestState::propose_shielding` are unchanged. `propose_shielding`'s public signature is also unchanged from `main`.

Required by:
- https://github.com/zcash/wallet/pull/402

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -23,6 +23,16 @@ workspace.
 - `zcash_client_backend::TransferType::AccountInternal`: indicates an output
   whose recipient and funder are the same wallet account (e.g. change). This
   has the semantics previously carried by `TransferType::WalletInternal`.
+- `zcash_client_backend::data_api::wallet::propose_shielding_coinbase` and
+  `zcash_client_backend::data_api::wallet::input_selection::ShieldingSelector::propose_shielding_coinbase`,
+  which propose a transaction that shields one or more coinbase transparent
+  outputs to an arbitrary shielded recipient.
+- `zcash_client_backend::proposal::ProposalError::ShieldingRequiresShieldedRecipient`,
+  returned by `propose_shielding_coinbase` when the supplied `to_address` is
+  a transparent or TEX address.
+- `zcash_client_backend::data_api::wallet::ProposeShieldingCoinbaseErrT` type
+  alias, parallel to `ProposeShieldingErrT` but parameterized on a `FeeRule`
+  instead of a `ChangeStrategy`.
 
 ### Changed
 - `zcash_client_backend::data_api`:
@@ -55,6 +65,9 @@ workspace.
     - It now has an `AccountId` generic parameter,
     - `from_parts` now takes additional `recipient_account`,
       `recipient_key_scope`, and `funding_account` parameters.
+- `zcash_client_backend::data_api::wallet::input_selection::ShieldingSelector`
+  now requires implementors to provide `propose_shielding_coinbase` in
+  addition to `propose_shielding`.
 
 ### Removed
 - `zcash_client_backend::data_api::WalletUtxo` (use `WalletTransparentOutput` 
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -6404,3 +6404,298 @@ where
         "All proposal should contain both transparent inputs"
     );
 }
+
+/// Verifies that `propose_shielding_coinbase` with a shielded destination produces
+/// a proposal containing a single ZIP-321 payment to the supplied address for the
+/// full available value (input total minus fee), with no change.
+#[cfg(all(feature = "pczt", feature = "transparent-inputs"))]
+pub fn propose_shielding_coinbase_succeeds<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
+    Dsf: DataStoreFactory,
+    <<Dsf as DataStoreFactory>::DataStore as WalletWrite>::UtxoRef: std::fmt::Debug,
+{
+    let mut st = TestDsl::with_sapling_birthday_account(ds_factory, cache).build::<T>();
+    let (t_addr, _) = st.get_account().usk().default_transparent_address();
+    let coinbase_value = Zatoshis::const_from_u64(50000);
+    let coinbase_build_result = build_transparent_coinbase_tx(
+        st.network(),
+        TargetHeight::from(st.sapling_activation_height()),
+        coinbase_value,
+        t_addr,
+        None,
+    );
+    let coinbase_tx = coinbase_build_result.transaction();
+    let (h, _) = st.generate_next_block_from_tx(0, coinbase_tx);
+    st.scan_cached_blocks(h, 1);
+    let params = *st.network();
+    decrypt_and_store_transaction(&params, st.wallet_mut(), coinbase_tx, Some(h)).unwrap();
+    // Coinbase outputs require 100 confirmations.
+    st.add_empty_blocks(100);
+
+    // The destination is a shielded address controlled by a separate spending key
+    // (i.e. potentially in a different wallet).
+    let to_extsk = T::sk(&[0xab; 32]);
+    let to_address = T::sk_default_address(&to_extsk).to_zcash_address(st.network());
+
+    let proposal = st
+        .propose_shielding_coinbase(
+            &GreedyInputSelector::new(),
+            &StandardFeeRule::Zip317,
+            Zatoshis::ZERO,
+            &[t_addr],
+            to_address.clone(),
+            None,
+            None,
+        )
+        .expect("propose_shielding_coinbase with a shielded destination should succeed");
+
+    let step = proposal.steps().first();
+    assert_eq!(
+        step.transparent_inputs().len(),
+        1,
+        "Expected exactly one coinbase transparent input"
+    );
+    let payments = step.transaction_request().payments();
+    assert_eq!(
+        payments.len(),
+        1,
+        "Expected exactly one payment in proposal"
+    );
+    let (idx, payment) = payments.iter().next().unwrap();
+    assert_eq!(*idx, 0);
+    assert_eq!(payment.recipient_address(), &to_address);
+    assert_eq!(
+        step.balance().proposed_change().len(),
+        0,
+        "Coinbase shielding must produce no change"
+    );
+
+    let fee = step.balance().fee_required();
+    let payment_amount = payment.amount().expect("payment must have an amount");
+    assert_eq!(
+        (payment_amount + fee).unwrap(),
+        coinbase_value,
+        "payment_amount + fee must equal coinbase input value"
+    );
+}
+
+/// Verifies that `propose_shielding_coinbase` rejects a transparent destination
+/// with [`ProposalError::ShieldingRequiresShieldedRecipient`].
+#[cfg(all(feature = "pczt", feature = "transparent-inputs"))]
+pub fn propose_shielding_coinbase_transparent_recipient_rejected<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
+    Dsf: DataStoreFactory,
+    <<Dsf as DataStoreFactory>::DataStore as WalletWrite>::UtxoRef: std::fmt::Debug,
+{
+    let mut st = TestDsl::with_sapling_birthday_account(ds_factory, cache).build::<T>();
+    let (t_addr, _) = st.get_account().usk().default_transparent_address();
+    let coinbase_value = Zatoshis::const_from_u64(50000);
+    let coinbase_build_result = build_transparent_coinbase_tx(
+        st.network(),
+        TargetHeight::from(st.sapling_activation_height()),
+        coinbase_value,
+        t_addr,
+        None,
+    );
+    let coinbase_tx = coinbase_build_result.transaction();
+    let (h, _) = st.generate_next_block_from_tx(0, coinbase_tx);
+    st.scan_cached_blocks(h, 1);
+    let params = *st.network();
+    decrypt_and_store_transaction(&params, st.wallet_mut(), coinbase_tx, Some(h)).unwrap();
+    st.add_empty_blocks(100);
+
+    let bad_to_address = Address::Transparent(TransparentAddress::PublicKeyHash([7; 20]))
+        .to_zcash_address(st.network());
+
+    let result = st.propose_shielding_coinbase(
+        &GreedyInputSelector::new(),
+        &StandardFeeRule::Zip317,
+        Zatoshis::ZERO,
+        &[t_addr],
+        bad_to_address,
+        None,
+        None,
+    );
+
+    assert_matches!(
+        result,
+        Err(Error::Proposal(
+            ProposalError::ShieldingRequiresShieldedRecipient
+        ))
+    );
+}
+
+/// Verifies that `propose_shielding_coinbase` propagates the supplied `memo`
+/// into the resulting payment's memo field.
+#[cfg(all(feature = "pczt", feature = "transparent-inputs"))]
+pub fn propose_shielding_coinbase_with_memo_succeeds<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
+    Dsf: DataStoreFactory,
+    <<Dsf as DataStoreFactory>::DataStore as WalletWrite>::UtxoRef: std::fmt::Debug,
+{
+    let mut st = TestDsl::with_sapling_birthday_account(ds_factory, cache).build::<T>();
+    let (t_addr, _) = st.get_account().usk().default_transparent_address();
+    let coinbase_value = Zatoshis::const_from_u64(50000);
+    let coinbase_build_result = build_transparent_coinbase_tx(
+        st.network(),
+        TargetHeight::from(st.sapling_activation_height()),
+        coinbase_value,
+        t_addr,
+        None,
+    );
+    let coinbase_tx = coinbase_build_result.transaction();
+    let (h, _) = st.generate_next_block_from_tx(0, coinbase_tx);
+    st.scan_cached_blocks(h, 1);
+    let params = *st.network();
+    decrypt_and_store_transaction(&params, st.wallet_mut(), coinbase_tx, Some(h)).unwrap();
+    st.add_empty_blocks(100);
+
+    let to_extsk = T::sk(&[0xcd; 32]);
+    let to_address = T::sk_default_address(&to_extsk).to_zcash_address(st.network());
+
+    let memo_text = "shielding to external wallet";
+    let memo_bytes = MemoBytes::from(memo_text.parse::<Memo>().unwrap());
+
+    let proposal = st
+        .propose_shielding_coinbase(
+            &GreedyInputSelector::new(),
+            &StandardFeeRule::Zip317,
+            Zatoshis::ZERO,
+            &[t_addr],
+            to_address,
+            Some(memo_bytes.clone()),
+            None,
+        )
+        .expect("propose_shielding_coinbase with memo should succeed");
+
+    let payments = proposal.steps().first().transaction_request().payments();
+    let (_, payment) = payments.iter().next().unwrap();
+    assert_eq!(payment.memo(), Some(&memo_bytes));
+}
+
+/// Verifies that `propose_shielding_coinbase` with `limit = Some(n)` selects at
+/// most `n` UTXOs, preferring the highest-value coinbase outputs.
+#[cfg(all(feature = "pczt", feature = "transparent-inputs"))]
+pub fn propose_shielding_coinbase_with_limit_truncates_inputs<T: ShieldedPoolTester, Dsf>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) where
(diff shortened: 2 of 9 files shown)


zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6391,3 +6399,280 @@ where
...
+        u64::from(payment_amount) > 0,
+        "Payment amount must be positive (was {})",
+        u64::from(payment_amount)
+    );
+    assert!(
+        u64::from(payment_amount) < u64::from(coinbase_value),
+        "Payment amount ({}) must be strictly less than coinbase input ({}) after fee",
+        u64::from(payment_amount),
+        u64::from(coinbase_value)
+    );
```
nullcopy, line 6465: Would be better to strictly check that the amount == coinbase - fee, instead of bound checking with the prior two assertions

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6391,3 +6399,280 @@ where
...
+        "Payment memo should match the supplied memo"
+    );
+}
+
+/// Verifies that `propose_shielding` with `to_address = None` preserves the
+/// legacy behavior: the resulting proposal has an empty `TransactionRequest`,
+/// so the shielded value flows entirely to the account's internal address
+/// via change.
+#[cfg(all(feature = "pczt", feature = "transparent-inputs"))]
+pub fn propose_shielding_to_none_preserves_legacy_behavior<T: ShieldedPoolTester, Dsf>(
```
nullcopy: Isn't this test redundant with the legacy tests which now pass `None`? Not that redundant tests are necessarily a bad thing 😅

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -242,6 +242,19 @@ pub trait ShieldingSelector {
     ///
     /// The `output_filter` parameter controls which transparent outputs are eligible for
     /// inclusion in the proposal. See [`TransparentOutputFilter`] for details.
+    ///
+    /// The `to_address` parameter, when `Some(addr)`, specifies the address that will
+    /// receive the shielded value. The address must have a shielded receiver (Sapling
+    /// or Orchard); transparent and TEX addresses are rejected.
```
nullcopy: I think we additionally want to require the `to_address` to be an account owned address, to prevent someone from using this method to build a payment transaction which would link their transparent outputs in the payment.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1131,60 +1148,254 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+                    return Err(InputSelectorError::Proposal(
+                        ProposalError::ShieldingRequiresShieldedRecipient,
+                    ));
+                }
+            }
+        } else {
+            None
+        };
+
+        // A memo can only be attached to a shielded destination payment.
```
nullcopy: This should be infallible, since we require the destination to be a shielded address. Perhaps would be better to `memo.expect()` rather than `if memo.is_some()`? I'm not sure our policy on when to crash vs error, though

nullcopy (changes requested):
I'm realizing this breaks an important property of propose shielding, and want to make sure we actually want this..

Previously it was _intentionally_ impossible to spend a transparent output to 3rd party. This PR breaks that property, allowing users to construct transactions which directly spend from transparent outputs to any shielded receiver. This could be misused to construct transactions that link user's transparent activity to a 3rd party.

[edit] From discussion with @nuttycom, it would be better not to modify `propose_shielding`, and instead create a new method which:
* allows creating a transaction to any arbitrary _shielded_ receiver
* BUT restricts input selection to coinbase outputs

schell:
Ok @nullcopy - thanks for your input! Now `propose_shielding` is reverted to its pre-PR interface; the new approach adds a dedicated propose_shielding_coinbase method that hard-codes `TransparentOutputFilter::CoinbaseOnly` at the API boundary and accepts an arbitrary shielded `to_address`. This preserves the "no spending non-coinbase transparent outputs to third parties" property while still letting miners shield coinbase to custodial/external wallets.

nullcopy:
FYI, the commit message talks about changes made from the prior version of this commit, to the current amended version. Could you please revise the commit message to drop out the historical bits and just describe what is being added/changed?

zcash_client_backend/CHANGELOG.md
```diff
@@ -12,6 +12,24 @@ workspace.
...
+- `zcash_client_backend::data_api::wallet::propose_shielding_coinbase` and
+  `zcash_client_backend::data_api::wallet::input_selection::ShieldingSelector::propose_shielding_coinbase`,
+  which propose a transaction that shields one or more coinbase transparent
+  outputs to an arbitrary shielded recipient. Unlike `propose_shielding`, this
+  method restricts input selection to coinbase outputs only (a privacy-relevant
+  property: coinbase outputs have no prior transparent transaction graph), and
+  carries an explicit ZIP-321 payment to the supplied address rather than
+  delivering the value via change. It also accepts an optional `memo` and an
+  optional `limit` capping the number of selected UTXOs to the highest-value
+  `n` of those eligible.
```
nullcopy: (nit) IMO this could be shortened. IMO remove "Unlike `propose_shielding`, this..." to the end.

zcash_client_backend/src/proposal.rs
```diff
@@ -70,6 +70,16 @@ pub enum ProposalError {
...
+    /// A shielding proposal was constructed with a destination address that has no shielded
+    /// receiver. Shielding requires the destination to be able to receive shielded value.
+    #[cfg(feature = "transparent-inputs")]
+    ShieldingRequiresShieldedRecipient,
+    /// An internal invariant violation: a shielding proposal's destination resolved to a
+    /// transparent pool. This error is unreachable today; it exists so that the internal
+    /// helper that computes the shielding bundle's trial balance can fail-safe in case a
+    /// future refactor bypasses the destination-resolution step.
+    #[cfg(feature = "transparent-inputs")]
+    AttemptedShieldingToTransparentAddress,
```
nullcopy: Isn't this redundant with `ShieldingRequiresShieldedRecipient`?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -267,6 +267,63 @@ pub trait ShieldingSelector {
...
+    ///
+    /// - Only coinbase transparent outputs are eligible for inclusion in the proposal. This
+    ///   restriction is hard-coded; callers cannot opt in to selecting non-coinbase outputs
+    ///   via this method. Coinbase outputs are uniquely suited to being sent to arbitrary
+    ///   shielded recipients because they have no prior transparent transaction graph that
+    ///   could be exposed to the recipient.
+    /// - The `to_address` argument specifies the destination of the shielded value. It must
+    ///   be a shielded address (Sapling, or a Unified Address with a shielded receiver); the
+    ///   method returns [`ProposalError::ShieldingRequiresShieldedRecipient`] for transparent
+    ///   or TEX destinations.
```
nullcopy: The key difference about `to_address` is not so much about it being shielded, but that it may be an external address not belonging to any account we control

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -267,6 +267,63 @@ pub trait ShieldingSelector {
...
+        wallet_db: &Self::InputSource,
+        change_strategy: &ChangeT,
+        shielding_threshold: Zatoshis,
+        source_addrs: &[TransparentAddress],
+        to_account: <Self::InputSource as InputSource>::AccountId,
+        to_address: ZcashAddress,
+        memo: Option<MemoBytes>,
+        limit: Option<usize>,
+        target_height: TargetHeight,
+        confirmations_policy: ConfirmationsPolicy,
```
nullcopy: IMO `confirmations_policy` doesn't need to be provided, since coinbase outputs have a protocol mandated 100 block conf

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -267,6 +267,63 @@ pub trait ShieldingSelector {
...
+    #[allow(clippy::type_complexity)]
+    #[allow(clippy::too_many_arguments)]
+    fn propose_shielding_coinbase<ParamsT, ChangeT>(
+        &self,
+        params: &ParamsT,
+        wallet_db: &Self::InputSource,
+        change_strategy: &ChangeT,
+        shielding_threshold: Zatoshis,
+        source_addrs: &[TransparentAddress],
+        to_account: <Self::InputSource as InputSource>::AccountId,
```
nullcopy: IMO we should rename `to_account` to `change_account`, to emphasize that the caller cannot specify destination by account (that is already perfectly served by the existing `propose_shielding`).

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -267,6 +267,63 @@ pub trait ShieldingSelector {
...
+    /// most `n`, keeping the highest-value UTXOs (with a stable tiebreaker by outpoint).
+    /// `Some(0)` selects no inputs and will therefore return
+    /// [`InputSelectorError::InsufficientFunds`].
+    ///
+    /// If the total value of selected inputs (after dust filtering and any cap imposed by
+    /// `limit`) is less than `shielding_threshold`, this method returns
+    /// [`InputSelectorError::InsufficientFunds`].
+    #[allow(clippy::type_complexity)]
+    #[allow(clippy::too_many_arguments)]
+    fn propose_shielding_coinbase<ParamsT, ChangeT>(
```
nullcopy: IMO, "shielding" implies that the coins stay in your wallet, just in a shielded pool. This function specifically enables sending to externally controlled accounts, though.

Perhaps rename to `sweep_coinbase_outputs`?

zcash_client_backend/src/data_api/wallet/input_selection.rs
nullcopy: Actually, thinking about it some more, I think we should drop this arg and disallow change altogether. Allowing change creates a scenario where the recipient learns about more of the sender's value than what they are receiving.

e.g. if a miner uses this method to send some of his mining rewards to an exchange, the exchange will learn of the full value (amount + change), by observing every UTXO that funded the transaction.

e.g. if I deposit 6.5 ZEC to an exchange, I don't want them to know about the other 6 ZEC I keep for myself.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -267,6 +267,63 @@ pub trait ShieldingSelector {
...
+        &self,
+        params: &ParamsT,
+        wallet_db: &Self::InputSource,
+        change_strategy: &ChangeT,
```
nullcopy: See my [other comment](https://github.com/zcash/librustzcash/pull/2353/changes/206485ca2761ecdeb0c367afad0060cc50fb8feb#r3243547803) about disallowing change

nullcopy (changes requested):
IMO we need to disallow change for this method. See my comments below.

Reviewed as of 206485ca2761ecdeb0c367afad0060cc50fb8feb

zcash_client_backend/src/data_api/wallet/input_selection.rs
schell: What you're saying makes sense, but I think it may be overfit to someone working in `librustzcash` =P. 

IMO "shielding"  (including `z_shieldcoinbase` itself) means "move transparent value into a shielded pool", and it doesn't _necessarily_ carry a same-wallet implication. The fact that the recipient may be external is the very feature this method adds over `propose_shielding`. It doesn't make the operation stop being shielding. Dropping the `propose_*` prefix also breaks the naming convention of this module (`propose_transfer`, `propose_send_max`, `propose_shielding`).

If it is important enough, though, I'd suggest `propose_sweep_coinbase_to_shielded` as a longer alternative? But I think `propose_shielding_coinbase` + a clear rustdoc bullet (already present) is more readable.

nullcopy: Yeah, perhaps its just my own personal bias :sweat_smile:

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1187,4 +1191,429 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+/// destination pool. The caller is responsible for computing the real payment
+/// amount from `input_total - balance.fee_required()` afterwards.
+///
+/// When `destination_pool` is `None`, no shielded output is supplied; the
+/// strategy will direct all available value into change (matching the
+/// `propose_shielding` "all-change" shape).
+#[cfg(feature = "transparent-inputs")]
+#[allow(clippy::type_complexity)]
+#[allow(clippy::too_many_arguments)]
+fn compute_shielding_balance_with_dust_retry<DbT, ChangeT, ParamsT>(
```
nullcopy, line 1456: Good feedback from Claude:
```
  compute_shielding_balance_with_dust_retry and compute_shielding_balance (input_selection.rs:1466-1618) are written to take an Option<PoolType> destination and dispatch into
  Sapling/Orchard/None bundle views. But propose_shielding_coinbase (1196-1310) never calls them — it computes fee and balance ad hoc via fee_rule.fee_required and a manual input_total -
  fee. The only remaining caller is propose_shielding, which passes None. That makes the following all dead:

  - compute_shielding_balance branches for Some(PoolType::SAPLING), Some(PoolType::ORCHARD), and Some(_) (input_selection.rs:1568-1604).
  - ShieldingBalanceError::TransparentDestination (1532) and the whole ShieldingBalanceError wrapper — without the Some(...) branches there's no error besides Change, so the wrapper isn't
  pulling its weight.
  - destination_pool: Option<PoolType> plumbed through both helpers.
  - The SaplingPayment(Zatoshis::ZERO) / OrchardPayment(Zatoshis::ZERO) zero-value scaffolding (1563-1565).

  Either commit to the ad-hoc shape (drop the destination_pool plumbing and ShieldingBalanceError, restore the helper to its pre-PR None-only shape) or have propose_shielding_coinbase
  actually go through compute_shielding_balance_with_dust_retry. The former is the smaller change and matches what the code currently does at runtime. The latter would re-introduce
  dust-filtering symmetry with the legacy path (see below).
```

IMO, if it doesn't end up being useful to your feature (`propose_shielding_coinbase`) then let's remove it and not change `propose_shielding` for the sake of minimizing surface area

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1187,4 +1191,429 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+
+/// Internal error type returned by [`compute_shielding_balance`] so that the
+/// caller can distinguish between change-strategy errors and the (unreachable
+/// in practice) case where `destination_pool` is `Some(PoolType::Transparent)`.
+#[cfg(feature = "transparent-inputs")]
+enum ShieldingBalanceError<ChangeErrT> {
+    /// The change strategy reported an error while computing the trial balance.
+    Change(ChangeError<ChangeErrT, Infallible>),
+    /// `destination_pool` was `Some(PoolType::Transparent)`. See
+    /// [`ProposalError::AttemptedShieldingToTransparentAddress`].
```
nullcopy: This error type doesn't exist anymore

nullcopy (changes requested):
Reviewed as of d18a2783f980f30f2bb6a2844daedbfa8c962d33

nullcopy (approved):
utACK 773e07ab1246b3256b486135257644c561852e30

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1187,4 +1190,343 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+                0,
+                sapling_output_count,
+                orchard_action_count,
+            )
+            // The `InputSelectorError::Change` variant is the only existing
+            // carrier capable of holding an arbitrary fee-rule error
+            // (`ChangeError::StrategyError` wraps `FeeRuleT::Error` in the
+            // generic position). We reuse it here rather than introduce a new
+            // top-level variant.
+            .map_err(|e| InputSelectorError::Change(ChangeError::StrategyError(e)))?;
```
nullcopy, line 1279: (nit) I wouldn't be against adding a new top-level error

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1187,4 +1190,343 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+        // rather than a more general "shield to address" path; only coinbase
+        // outputs are eligible because they have no prior transparent transaction
+        // graph that could be exposed to the shielded recipient.
+        let mut transparent_inputs = gather_shielding_inputs::<DbT, FeeRuleT::Error>(
+            wallet_db,
+            source_addrs,
+            target_height,
+            // It doesn't matter here if we pass a 100 confirmations or 1 confirmations policy,
+            // as coinbase txs require 100, which will be enforced by note selection.
+            ConfirmationsPolicy::MIN,
```
nuttycom, line 1225: nit: even though note selection will enforce this via `CoinbaseOnly`, why not just add a 100-block confirmation policy as defense-in-depth?

nuttycom (commented):
post-hoc utACK

PR #2368, opened 2026-05-20
zcash_client_memory: convert From<proto::Note> for Note to TryFrom

camofu:
## What

`zcash_client_memory::MemoryWalletDb::decode_new<B: Buf>(buf: B, ...) -> Result<Self>` is a `pub` API that returns a `Result`, but the `From<proto::Note> for Note` impl reached from inside it panics on malformed input rather than returning `Err` — defeating the `Result` contract.

The impl had 8 panic sites: wrong-length sapling/orchard recipient, missing/unrecognized `rseed_type` discriminator, wrong-length `rseed` payload, missing/invalid orchard `rho`, invalid curve points (sapling `Fr`, orchard `Address`/`Rho`/`RandomSeed`), invalid `orchard::Note::from_parts` combination, and a `panic!("invalid protocol")` arm when the `orchard` feature is disabled.

This PR converts the impl to `TryFrom<proto::Note> for Note` and propagates typed `Err` through the two call sites in `received.rs` and `sent.rs`. No new `Error` variants are needed — existing `ByteVecToArrayConversion`, `ProtoMissingField`, `CorruptedData`, and `OrchardNotEnabled` cover everything.

## Test

Three regression tests in `mod tests` assert the expected `Err` variant for the three Sapling panic sites (wrong-length recipient, unrecognized `rseed_type`, wrong-length rseed payload). The existing `test_note_roundtrip` still passes.

```sh
cargo test --lib -p zcash_client_memory --features 'orchard,transparent-inputs' \
    types::notes
```

All 5 tests pass (3 new + 1 existing roundtrip + 1 sibling). Full lib test suite (41 tests) and `--no-default-features` build also pass.

## Scope

Intentionally narrow: this PR fixes one impl. The same `unwrap`/`panic!`-instead-of-`Err` pattern recurs in `account.rs`, `block.rs`, `notes/received.rs`, `notes/sent.rs`, `memory_wallet/mod.rs`, `wallet_write.rs`, and `wallet_read.rs` — happy to follow up with the same shape once this lands and the error-plumbing precedent is agreed.

### DIFF
--- a/zcash_client_memory/src/types/notes/mod.rs
+++ b/zcash_client_memory/src/types/notes/mod.rs
@@ -87,48 +87,71 @@ mod serialization {
         }
     }
 
-    impl From<proto::Note> for Note {
-        fn from(note: proto::Note) -> Self {
+    impl TryFrom<proto::Note> for Note {
+        type Error = Error;
+
+        fn try_from(note: proto::Note) -> Result<Self, Self::Error> {
             match note.protocol() {
                 proto::ShieldedProtocol::Sapling => {
-                    let recipient =
-                        sapling::PaymentAddress::from_bytes(&note.recipient.try_into().unwrap())
-                            .unwrap();
+                    let recipient_bytes: [u8; 43] = note.recipient.try_into()?;
+                    let recipient = sapling::PaymentAddress::from_bytes(&recipient_bytes)
+                        .ok_or_else(|| {
+                            Error::CorruptedData("invalid sapling recipient".to_string())
+                        })?;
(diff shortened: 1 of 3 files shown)


camofu (commented):
Replaces a `From` impl full of `unwrap()` / `panic!` on proto-deserialized Notes with a `TryFrom` that returns `Error::ByteVecToArrayConversion` on bad lengths and `Error::CorruptedData` on invalid points/seeds. 

The `not(feature = "orchard")` arm now returns `Error::OrchardNotEnabled` instead of panicking. Added three negative tests on the Sapling branch.

PR #2372, opened 2026-05-22
zcash_primitives: Reject v4 tx with empty Sapling bundle and non-zero…

camofu:
… valueBalanceSapling

`read_v4_components` read `valueBalanceSapling` unconditionally when `tx_has_sapling`, and `read_v4` then dropped it whenever both Sapling vectors were empty (the empty-bundle branch synthesizes `sapling_bundle = None` via `binding_sig.and_then(...)`). The net effect was that a v4 encoding with `nShieldedSpend == 0 && nShieldedOutput == 0 && valueBalanceSapling != 0` was silently accepted, even though consensus rule §7.1.2 requires `valueBalanceSapling == 0` in that case. Zebra (`zebra-chain/src/transaction/serialize.rs:962-963`) and post-v6.12.2 zcashd both reject this encoding; `zcash_primitives` now agrees.

Add a regression test that constructs the wire bytes directly and asserts `Transaction::read` returns an error, plus a sanity test for the canonical `vb == 0` encoding's round-trip.

### DIFF
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -32,6 +32,13 @@ workspace.
   Orchard and Ironwood bundles. Pass `orchard::builder::BundleType::DEFAULT` to keep
   the previous (padded) behavior.
 
+### Fixed
+- `zcash_primitives::transaction::Transaction::read` now rejects v4
+  transactions whose Sapling section contains zero spends and zero outputs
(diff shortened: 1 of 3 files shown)


nuttycom (dismissed):
utACK 28f1fbeb2e5b52b438f5f70f71b78c0be3491c5a

PR #2373, opened 2026-05-22
fix(zcash_transparent): reject oversized PCZT partial_signatures at parse time

Zk-nd3r:
Input::parse accepts partial_signatures values of arbitrary length, and the P2PKH branch of spend_finalizer.rs pushes them into a script via pv::push_value(...).expect("short enough"). push_value returns None for any element over the 520-byte script push-value limit, so a partial_signatures entry longer than 520 bytes aborts the finalizer. The P2SH/P2MS branch already handles this with .ok_or(InvalidSignature)?; P2PKH was inconsistent.

- parse.rs: reject oversized partial_signatures entries at parse time, with a new ParseError::InvalidPartialSignature. This closes it at the boundary for every role, not just P2PKH.
- spend_finalizer.rs: convert the P2PKH signature push to .ok_or(SpendFinalizerError::InvalidSignature)?, matching the P2MS branch. The pubkey is a fixed 33-byte key that always fits, so it keeps an expect with an accurate message.
- Add a regression test for parse-time rejection of a 521-byte partial_signatures value.

Validation:
- cargo fmt --all
- cargo build -p zcash_transparent on Linux
- cargo test -p zcash_transparent on Linux, 9 passed

### DIFF
--- a/zcash_transparent/CHANGELOG.md
+++ b/zcash_transparent/CHANGELOG.md
@@ -10,6 +10,15 @@ workspace.
 
 ## [Unreleased]
 
+### Changed
+- `zcash_transparent::pczt::ParseError` has a new `InvalidPartialSignature`
+  variant. The `ParseError` enum is now marked `#[non_exhaustive]`.
+
+### Fixed
+- `zcash_transparent::pczt` spend finalization no longer panics on a
+  `partial_signatures` entry larger than the maximum script push-value size.
+  Such entries are now rejected at parse time.
+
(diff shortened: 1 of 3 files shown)


nuttycom (dismissed):
utACK 5e56aa4b6dd31cfa8e3a1c031ff73fd4c4045095

nuttycom (approved):
re-utACK 3a3d5d9aab1083336366b6219af8db000d7a7026

PR #2375, opened 2026-05-23
fix: align propose_shielding_coinbase fee with builder bundle minimums

### LINKED ISSUE #2374: `propose_shielding_coinbase` underestimates fee by one marginal unit
**Summary**

`propose_shielding_coinbase` (added in #2353) hardcodes the shielded output/action counts it passes to `FeeRule::fee_required`, but those counts don't reflect what the builder will actually produce. Both `sapling::builder::BundleType::DEFAULT` and `orchard::builder::BundleType::DEFAULT` pad up to a minimum of 2 (`MIN_SHIELDED_OUTPUTS` / `MIN_ACTIONS`), so the proposal underestimates the fee by exactly one ZIP‑317 marginal unit (5000 zat). The proposal succeeds; `create_proposed_transactions` then fails at build time with `Insufficient funds for transaction construction; need an additional ZatBalance(5000) zatoshis`.

**Location**

`zcash_client_backend/src/data_api/wallet/input_selection.rs:1249`:

```rust
let (sapling_output_count, orchard_action_count) = match destination_pool {
    PoolType::SAPLING => (1usize, 0usize),
    PoolType::ORCHARD  => (0usize, 1usize),
    ...
};

let fee = fee_rule.fee_required(
    params, BlockHeight::from(target_height),
    transparent_inputs.iter().map(transparent_fees::InputView::serialized_size),
    std::iter::empty::<usize>(),
    0,
    sapling_output_count,
    orchard_action_count,
)?;
```

The other proposal paths (e.g. `single_change_output_balance` in `fees/common.rs`) ask each bundle type for its real action count via `BundleType::DEFAULT.num_outputs(...)` / `.num_actions(...)`. This path doesn't, and so misses the padding.

**Reproduction**

End‑to‑end against a zaino‑backed regtest wallet: any `z_shieldcoinbase` call to a destination with a Sapling‑only or Orchard‑only receiver. The proposal returns a valid `Proposal`; `create_proposed_transactions` aborts with the "need additional 5000 zatoshis" error.

dianaravenborne:
Fixes #2374.
Fixes [COR-1261](https://linear.app/zodl/issue/COR-1261/propose-shielding-coinbase-underestimates-fee-by-one-marginal-unit).

`propose_shielding_coinbase` (added in #2353) hardcoded the shielded output / action counts it passes to `FeeRule::fee_required`:

```rust
let (sapling_output_count, orchard_action_count) = match destination_pool {
    PoolType::SAPLING => (1usize, 0usize),
    PoolType::ORCHARD  => (0usize, 1usize),
    ...
};
```

But both `sapling::builder::BundleType::DEFAULT` and `orchard::builder::BundleType::DEFAULT` pad up to a minimum of 2 outputs / actions when any output is added (`MIN_SHIELDED_OUTPUTS` and `MIN_ACTIONS` respectively). The proposal therefore underestimates the fee by exactly one ZIP-317 marginal unit (5000 zat). The proposal succeeds, then \`create_proposed_transactions\` fails at build time with:

> Insufficient funds for transaction construction; need an additional ZatBalance(5000) zatoshis

This makes \`z_shieldcoinbase\` (in zallet) unable to actually build a shielding tx end-to-end in either pool destination.

This fix queries each bundle type for the real count instead of hardcoding it, matching the pattern used by \`single_change_output_balance\` in \`fees/common.rs\`. The \`.expect()\` calls are safe because \`BundleType::DEFAULT\` is \`Transactional\` with both flags enabled, which permits any \`(num_spends, num_outputs)\` combination.

## Test plan

Verified end-to-end against a zaino-backed regtest wallet running \`z_shieldcoinbase\` to an Orchard receiver: before this fix the build step fails with the 5000-zat shortfall; after the fix the tx builds and confirms. Full integration test suite in zcash/integration-tests covering this and other code paths is green against this commit.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -1245,11 +1245,29 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
         // Compute the fee directly from the fee rule. This method produces no
         // change in any pool, so there is no change-strategy or wallet-metadata
         // computation to perform. The proposal will carry exactly one shielded
-        // output (in `destination_pool`) and zero transparent outputs.
+        // payment output (in `destination_pool`) and zero transparent outputs.
+        //
+        // Both Sapling and Orchard transactional bundles pad up to a minimum
+        // of 2 outputs/actions when any output is added (see `MIN_ACTIONS` /
+        // `MIN_SHIELDED_OUTPUTS` in the respective crates). The proposal fee
+        // must reflect the *padded* counts that the builder will actually
+        // produce, otherwise the subsequent `create_proposed_transactions`
+        // call fails with `InsufficientFunds(need additional <marginal_fee>)`.
+        // We query each bundle type for the count it will materialize.
         let (sapling_output_count, orchard_action_count) = match destination_pool {
-            PoolType::SAPLING => (1usize, 0usize),
+            PoolType::SAPLING => {
+                let count = ::sapling::builder::BundleType::DEFAULT
+                    .num_outputs(0, 1)
+                    .expect("sapling DEFAULT bundle type permits any (spends, outputs) count");
+                (count, 0usize)
+            }
             #[cfg(feature = "orchard")]
-            PoolType::ORCHARD => (0usize, 1usize),
+            PoolType::ORCHARD => {
+                let count = orchard::builder::BundleType::DEFAULT
+                    .num_actions(0, 1)
+                    .expect("orchard DEFAULT bundle type permits any (spends, outputs) count");
+                (0usize, count)
+            }
             // Unreachable: `resolve_shielded_destination` rejects transparent
             // destinations earlier with `ShieldingRequiresShieldedRecipient`.
             _ => {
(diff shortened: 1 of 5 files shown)


zcash_client_backend/src/data_api/wallet/input_selection.rs
nullcopy, line 1: Please add a regression test for this

dianaravenborne:
Done — added `propose_and_build_shielding_coinbase_succeeds` as a regression test (parameterized over `T: ShieldedPoolTester` so it runs once for Sapling, once for Orchard). The test runs proposal **and** `create_proposed_transactions` end-to-end; the build step is where the original bug surfaces.

Verified on the pre-fix code (reverted `input_selection.rs` to its pre-fix state with the new test in place): both `wallet::{sapling,orchard}::tests::propose_and_build_shielding_coinbase_succeeds` panic with `Insufficient funds for transaction construction; need an additional ZatBalance(5000) zatoshis`. With the fix restored, both pass. Squashed into the existing commit and force-pushed (`a72196fe86`).

nullcopy:
superseded by https://github.com/zcash/librustzcash/pull/2376

PR #2376, opened 2026-05-23
fix: align propose_shielding_coinbase fee with builder bundle minimums

### LINKED ISSUE #2374: `propose_shielding_coinbase` underestimates fee by one marginal unit

nullcopy:
Fixes #2374.
Fixes [COR-1261](https://linear.app/zodl/issue/COR-1261/propose-shielding-coinbase-underestimates-fee-by-one-marginal-unit).

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -23,7 +23,7 @@ use zcash_primitives::{
 };
 use zcash_protocol::{
     ShieldedProtocol,
-    consensus::{self, BlockHeight, NetworkUpgrade, Parameters},
+    consensus::{self, BlockHeight, COINBASE_MATURITY_BLOCKS, NetworkUpgrade, Parameters},
     local_consensus::LocalNetwork,
     memo::{Memo, MemoBytes},
     value::Zatoshis,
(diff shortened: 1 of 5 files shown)


zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6699,3 +6699,78 @@ pub fn propose_shielding_coinbase_with_zero_limit_insufficient_funds<T: Shielded
...
+            &StandardFeeRule::Zip317,
+            Zatoshis::ZERO,
+            &[t_addr],
+            to_address,
+            None,
+            None,
+        )
+        .expect("propose_shielding_coinbase should succeed");
+
+    // The actual regression: prior to #2375 this would fail at build time
```
daira: Was it a regression? "Regression test" means that it is testing for possible future regressions, not that the thing being fixed was a regression. Also the same comment as above about the PR number.
```suggestion
    // Prior to #2376 this would fail at build time
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6699,3 +6699,78 @@ pub fn propose_shielding_coinbase_with_zero_limit_insufficient_funds<T: Shielded
     // `available: 0, required: fee`.
     assert_matches!(result, Err(Error::InsufficientFunds { .. }));
 }
+
+/// Regression test for the propose-fee/build-fee mismatch fixed in #2375.
```
```suggestion
/// Regression test for the propose-fee/build-fee mismatch fixed in #2376.
```
#2375 was closed in favour of this PR.

daira (dismissed):
utACK with blocking comments.

nullcopy:
[Force-pushed](https://github.com/zcash/librustzcash/compare/e41d4aa9a84cef5d89aac894974c81e1469db2dc..39655aa21c19ab7b79c395fb10f022d72dfcc0f6) to address review feedback from @daira

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6699,3 +6699,77 @@ pub fn propose_shielding_coinbase_with_zero_limit_insufficient_funds<T: Shielded
...
+///
+/// Both `sapling::builder::BundleType::DEFAULT` and
+/// `orchard::builder::BundleType::DEFAULT` pad up to a minimum of 2
+/// outputs/actions (`MIN_SHIELDED_OUTPUTS` / `MIN_ACTIONS`). Before the fix,
+/// `propose_shielding_coinbase` hardcoded `(1, 0)` / `(0, 1)` when asking the
+/// fee rule what fee to charge, so the proposal underestimated the fee by
+/// exactly one ZIP-317 marginal unit (5000 zat). The proposal succeeded, but
+/// `create_proposed_transactions` then failed at build time with
+/// `Insufficient funds for transaction construction; need an additional
+/// ZatBalance(5000) zatoshis`.
```
```suggestion
/// `Insufficient funds for transaction construction; need an additional ZatBalance(5000) zatoshis`.
```
I think ` across lines doesn't work when generating the docs.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6699,3 +6699,77 @@ pub fn propose_shielding_coinbase_with_zero_limit_insufficient_funds<T: Shielded
...
+        t_addr,
+        None,
+    );
+    let coinbase_tx = coinbase_build_result.transaction();
+    let (h, _) = st.generate_next_block_from_tx(0, coinbase_tx);
+    st.scan_cached_blocks(h, 1);
+    let params = *st.network();
+    decrypt_and_store_transaction(&params, st.wallet_mut(), coinbase_tx, Some(h)).unwrap();
+    // Coinbase outputs require 100 confirmations.
+    st.add_empty_blocks(100);
```
```rust
./components/zcash_protocol/src/consensus.rs:686:pub const COINBASE_MATURITY_BLOCKS: u32 = 100;
```

daira (dismissed):
utACK with non-blocking suggestions.

zcash_client_backend/src/data_api/testing/pool.rs
nullcopy: I've fixed it in my additions, but there are several other places we could go improve this. I chose not to take on that cleanup task in this PR, but I can if you'd like

nullcopy:
[Force-pushed](https://github.com/zcash/librustzcash/compare/39655aa21c19ab7b79c395fb10f022d72dfcc0f6..454426f6062adc00e528beb0be37f01f3d3cf1de) to address feedback from @daira

nuttycom (approved):
utACK

PR #2383, opened 2026-05-26
Fix stale entries after truncation

### LINKED ISSUE #2382: zcash_client_sqlite: `truncate_to_height` leaves stale entries in `tx_retrieval_queue` and `transparent_spend_search_queue`
After a wallet truncate, the wallet's transaction retrieval queues still contain `txid` references to transactions that lived in the now-discarded portion of the chain. On the next sync tick, the `data_requests` task issues `getrawtransaction` for those txids, even though we should be unaware of those after truncation.

This rears its head in integration tests where I use truncation to simulate a reorg.. transactions before truncation are not present in the chain after reorg, and I expect Zalet NOT to go asking for them.

nullcopy:
Closes https://github.com/zcash/librustzcash/issues/2382

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -3737,6 +3737,27 @@ pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
         named_params![":height": u32::from(truncation_height)],
     )?;
 
+    // Remove entries from the transaction retrieval and transparent spend search queues that
+    // were queued solely as a consequence of having seen a transaction on-chain at a height
+    // that is about to be discarded. Without this, `transaction_data_requests` will keep
+    // returning requests for txids that lived only in the now-discarded portion of the chain.
+    // This must be done before un-mining the transactions below, since the criterion is based
+    // on the soon-to-be-cleared `mined_height` column.
+    conn.execute(
+        "DELETE FROM tx_retrieval_queue
+         WHERE dependent_transaction_id IN (
+            SELECT id_tx FROM transactions WHERE mined_height > :height
+         )",
+        named_params![":height": u32::from(truncation_height)],
+    )?;
+    conn.execute(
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet.rs
```diff
@@ -3737,6 +3737,27 @@ pub(crate) fn truncate_to_height_internal<P: consensus::Parameters>(
...
 
+    // Remove entries from the transaction retrieval and transparent spend search queues that
+    // were queued solely as a consequence of having seen a transaction on-chain at a height
+    // that is about to be discarded. Without this, `transaction_data_requests` will keep
+    // returning requests for txids that lived only in the now-discarded portion of the chain.
+    // This must be done before un-mining the transactions below, since the criterion is based
+    // on the soon-to-be-cleared `mined_height` column.
+    conn.execute(
+        "DELETE FROM tx_retrieval_queue
+         WHERE dependent_transaction_id IN (
```
czarcas7ic, line 3748: I think this only clears queue rows tied to a dependent transaction. Scan discovered txids are queued with `dependent_tx_ref = None` here:

https://github.com/zcash/librustzcash/blob/86d9764f3f1aacbfcdc87e682dc338e51cc9b7db/zcash_client_backend/src/data_api/ll/wallet.rs#L276-L278

That stores `dependent_transaction_id = NULL` here:

https://github.com/zcash/librustzcash/blob/86d9764f3f1aacbfcdc87e682dc338e51cc9b7db/zcash_client_sqlite/src/wallet.rs#L4557

Should this also remove queued txids whose own transaction row has `mined_height > :height`? If the goal is no `getrawtransaction` requests for discarded branch txids, I think the synthesized `GetStatus` path after truncation clears `mined_height` may need consideration too.

nullcopy:
Closing this, as the underlying issue was actually a misdiagnosis.

PR #2386, opened 2026-05-28
zcash_proofs: scope README to Sprout

ValarDragon:
The Sapling circuits have lived in `sapling-crypto` for some time; zcash_proofs now only contains the Sprout circuit (plus helpers for loading the Sapling proving parameters from disk). Update the README to reflect that.


dannywillems (approved):
LGTM.

PR #2392, opened 2026-05-29
zcash_transparent: reject zero threshold in zip48 FullViewingKey::standard

camofu:
## Summary

`zcash_transparent::zip48::FullViewingKey::standard(threshold, key_info)` validates that `key_info` is non-empty and that `threshold <= key_info.len()`, but does **not** reject `threshold == 0`.

Because `key_info` is a `NonEmpty`, `key_info.len()` is always `>= 1`, so the existing one-sided check `usize::from(threshold) > key_info.len()` can never be true for `threshold == 0` — a zero threshold flows straight through to `Ok(..)`. The resulting standard P2SH multisig redeem script is:

```
OP_0 <pk1> <pk2> ... <pkN> OP_<N> OP_CHECKMULTISIG
```

`OP_CHECKMULTISIG` with `required = 0` succeeds after consuming the customary off-by-one dummy element, so the resulting P2SH address is **anyone-can-spend**: funds sent to it can be swept by the first observer.

## Impact

This is **not a live vulnerability**. No wallet/RPC/PCZT path currently feeds a user-controlled `threshold` into `FullViewingKey::standard`; the only non-test caller (`pczt/tests/end_to_end.rs`) hardcodes `threshold = 2`.

It is a **defense-in-depth** fix for the recently-added ZIP 48 multisig public API. ZIP 48 multisig is intended for wallet integrators, and any wallet that surfaces `threshold` to the user (or reads it from a config/RPC payload) without its own `>= 1` validation would create a fund-loss footgun. Closing the gap in the library now prevents that before integrators ship against the API.

## Fix

Tighten the threshold check from one-sided to a closed range, and widen the `InvalidThreshold` doc comment accordingly:

```rust
} else if threshold == 0 || usize::from(threshold) > key_info.len() {
    Err(FullViewingKeyError::InvalidThreshold)
```

A regression test (`standard_rejects_zero_threshold`) asserts `Err(InvalidThreshold)` for a `threshold == 0` call with three valid account keys.

## Note on `zcash_script`

The same one-sided check exists one layer down in `zcash_script::descriptor::multi_inner` (`if k > n`, not `k == 0`). `zcash_script` is a separate crate (not part of this workspace), so that layer needs a separate PR against its own repository. Fixing both layers is appropriate, since a direct caller of `zcash_script::descriptor::{multi, sortedmulti}` with `k = 0` would produce the same anyone-can-spend script.


zcash_transparent/src/zip48.rs
```diff
@@ -239,7 +239,7 @@ impl FullViewingKey {
         let key_info = NonEmpty::from_vec(key_info).ok_or(FullViewingKeyError::NoPubKeys)?;
         if key_info.len() > 15 {
             Err(FullViewingKeyError::TooManyPubKeys)
-        } else if usize::from(threshold) > key_info.len() {
+        } else if threshold == 0 || usize::from(threshold) > key_info.len() {
```
nuttycom: Instead of this, we should use the https://doc.rust-lang.org/std/num/type.NonZeroU8.html type for `threshold` if we want to enforce this invariant. Then it is apparent in the type, and inviolable by construction.

nuttycom (changes requested):

zcash_transparent/src/zip48.rs
camofu: Done!

camofu (commented):
`FullViewingKey::standard` now takes threshold: `NonZeroU8` instead of `u8`.

nuttycom (approved):
utACK

PR #2393, opened 2026-05-29
zcash_client_sqlite: use regex version from workspace

dannywillems:
It seems that most of the dependencies are shared by the workspace, except a few. This is a small patch to get more uniformity.


dannywillems:
Closing in favor of https://github.com/zcash/librustzcash/pull/2397

PR #2397, opened 2026-05-30
zcash_client_sqlite: use regex version from workspace

dannywillems:
It seems that most of the dependencies are shared by the workspace, except a few. This is a small patch to get more uniformity.


nullcopy (approved):
utACK ccf933622798a420f58fc3f14ede330ddfaab453

PR #2410, opened 2026-06-06
Use From instead of widening as-casts (clippy::cast_lossless)

dannywillems:
## Summary

Replaces widening integer `expr as T` casts with the value-identical
`T::from(expr)` across the workspace. These are lossless conversions (`u8`/
`u16`/`u32` -> `u64`, `u32` -> `i64`, `u16` -> `u32`, `bool` -> `i64`, etc.),
so there is no behaviour change. This resolves every `clippy::cast_lossless`
warning across the workspace under `--all-features`.

## Why

`as` casts silently become lossy if the operand type later widens; the `From`
form makes the conversion fail to compile instead, which is the safer default
for arithmetic-heavy code.

## Scope

Pure refactor, no public API or behaviour change. 9 files, +33/-25:

- `components/eip681`, `components/equihash`, `components/zcash_address`,
  `components/zcash_encoding`, `components/zcash_protocol`
- `zcash_client_backend` (testing utilities), `zcash_client_memory`,
  `zcash_client_sqlite`

## Verification

- `cargo clippy --workspace --lib --all-features -- -W clippy::cast_lossless`
  reports 0 warnings.
- Workspace compiles; existing tests for the affected crates pass.

### DIFF
--- a/components/eip681/src/parse.rs
+++ b/components/eip681/src/parse.rs
@@ -82,7 +82,7 @@ impl Digits {
             total = total
                 .checked_mul(10)
                 .context(OverflowSnafu)?
-                .checked_add(*digit as u64)
+                .checked_add(u64::from(*digit))
                 .context(OverflowSnafu)?;
         }
         Ok(total)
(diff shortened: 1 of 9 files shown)


nuttycom (approved):
utACK

PR #2415, opened 2026-06-10
zcash_client_backend: use core instead of std

dannywillems:


nullcopy (approved):
utACK 99786ae75fdf77dbf5b0721c9bd7af643b450a40

PR #2419, opened 2026-06-16
zcash_client_backend: Support spending transparent UTXOs in propose_transaction

fndecroix:
Solves [issue 2350](https://github.com/zcash/librustzcash/issues/2350)

`propose_transaction` allows usage of transparent funds when shielded funds are not enough.

There is an optimization that should be considered in the future. Now that we are allowing transparent inputs, we should try using them by default when transfering to a TEX address instead of building an extra transaction. I didn't include it in this PR to limit the scope of this task. I could add an issue as a follow up for this.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -202,6 +202,7 @@ pub trait InputSelector {
         account: <Self::InputSource as InputSource>::AccountId,
         transaction_request: TransactionRequest,
         change_strategy: &ChangeT,
+        #[cfg(feature = "transparent-inputs")] transparent_source_addrs: &[TransparentAddress],
         #[cfg(feature = "unstable")] proposed_version: Option<TxVersion>,
     ) -> Result<
         Proposal<<ChangeT as ChangeStrategy>::FeeRule, <Self::InputSource as InputSource>::NoteRef>,
@@ -461,6 +462,7 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
         account: <DbT as InputSource>::AccountId,
         transaction_request: TransactionRequest,
         change_strategy: &ChangeT,
+        #[cfg(feature = "transparent-inputs")] transparent_source_addrs: &[TransparentAddress],
         #[cfg(feature = "unstable")] proposed_version: Option<TxVersion>,
     ) -> Result<
         Proposal<<ChangeT as ChangeStrategy>::FeeRule, DbT::NoteRef>,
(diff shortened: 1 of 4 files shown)


nuttycom:
I haven't yet reviewed this PR, but:

> Now that we are allowing transparent inputs, we should try using them by default when transfering to a TEX address

This is exactly the wrong framing; this directly eliminates a user's privacy by making their transfers linkable. If you spend funds from the shielded pool in the source of a TEX transfer, then you retain your privacy; we MUST NOT provide a default that allows pure t->t transfers. t->t transfers are necessary for support of the legacy zcashd wallet behavior, but we must NEVER use them except in the context of users who require those flows and explicitly acknowledge that they give up all potential privacy.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -202,6 +202,7 @@ pub trait InputSelector {
         account: <Self::InputSource as InputSource>::AccountId,
         transaction_request: TransactionRequest,
         change_strategy: &ChangeT,
+        #[cfg(feature = "transparent-inputs")] transparent_source_addrs: &[TransparentAddress],
```
nuttycom, line 205: Instead of a bare slice of addresses, what we need here is an enumeration that can support the `ANY_TADDR` zcashd semantics, which explicitly permits linking transparent addresses by including funds from multiple addresses when spending. For large wallets (for example, exchange wallets containing tens of thousands of transparent addresses) we should not be performing address selection ahead of time for the `ANY_TADDR` case, but instead be performing address selection *in tandem* with UTXO selection to meet the desired spend amount.

schell:
I'm closing this in favor of #2422. @fndecroix, I rolled some of your changes from this branch into that PR 🙇

fndecroix:
> I haven't yet reviewed this PR, but:
> 
> > Now that we are allowing transparent inputs, we should try using them by default when transfering to a TEX address
> 
> This is exactly the wrong framing; this directly eliminates a user's privacy by making their transfers linkable. If you spend funds from the shielded pool in the source of a TEX transfer, then you retain your privacy; we MUST NOT provide a default that allows pure t->t transfers. t->t transfers are necessary for support of the legacy zcashd wallet behavior, but we must NEVER use them except in the context of users who require those flows and explicitly acknowledge that they give up all potential privacy.

Thanks for the explanation @nuttycom ! I'm not sure if I get it though. If a user needs to transfer to a TEX address and it has enough transparent inputs to make the transaction using 100% transparent inputs, we should still try to use shielded funds by default?

If that's the case, the PR is working as expected. In this PR we are using shielded inputs by default and always following the tr0/tr1 scheme, so there shouldn't be any issue. My previous comment was only suggesting a possible future optimization to avoid the extra transaction in the case we are already using 100% transparent inputs, but it is not implemented and there's no need to do it if that's not the intended behavior.

PR #2421, opened 2026-06-17
Update zcash_history for Ironwood (adding NodeDataV3)

czarcas7ic:
Adds V3 chain history metadata for the Ironwood shielded pool, including `NodeDataV3`, the `V3` history tree version, and updated serialized size limits.

The new metadata tracks Ironwood start and end roots plus transaction counts through serialization, node combination, append, and truncate paths, with tests covering byte fixtures, maximum encoded sizes, and hash commitment changes.

Validation:
- cargo fmt --all -- --check
- cargo test -p zcash_history --all-features

solves COR-1385


zcash_history/CHANGELOG.md
```diff
@@ -8,9 +8,16 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 ### Changed
 - MSRV is now 1.85.1.
+- `zcash_history::MAX_NODE_DATA_SIZE` is now 317 bytes and the derived
+  `zcash_history::MAX_ENTRY_SIZE` is now 326 bytes, to account for Ironwood
+  history node metadata.
 
 ### Added
 - `zcash_history::Entry::data`
+- `zcash_history::NodeDataV2`, exposing the V2 history tree node metadata type.
```
ValarDragon, line 17: V2 was already exposed I believe?

czarcas7ic: `zcash_history::V2` (the `version::V2` marker enum) was already exported. This is about `node_data::V2`, the metadata struct holding the Orchard roots and tx count, which is a separate type.

It lives in the private mod `node_data`, and on `feat/ironwood` the only re-exports from there were `MAX_NODE_DATA_SIZE` and `NodeData`:

https://github.com/zcash/librustzcash/blob/a5532b85961eee157468e6fcb63d1b5c7131ac9f/zcash_history/src/lib.rs#L15-L17

This PR adds `V2` as `NodeDataV2` to that `pub use`.

ValarDragon (dismissed):

nuttycom (dismissed):
utACK 451228ac30ad31c3c226d79f4a6756d491cba968; added tests verifying against https://github.com/zcash/zcash-test-vectors/pull/131

nuttycom (approved):
re-utACK 8090078c2d7b0b7800c24b797f4639af564dd2e7 after test cleanup

zcash_history/src/node_data.rs
```diff
@@ -17,8 +17,11 @@ pub const MAX_NODE_DATA_SIZE: usize = 32 + // subtree commitment
...
-    9; // Orchard tx count (compact uint)
-// = total of 244
+    9 + // Orchard tx count (compact uint)
+    32 + // start Ironwood tree root
+    32 + // end Ironwood tree root
+    9; // Ironwood tx count (compact uint)
+// = total of 317
 
 /// V1 node metadata.
 #[repr(C)]
```
daira, line 27: This `#[repr(C)]` isn't necessary. (Claude confirmed that it has never been necessary since `zcash_history` was created from `zcash_mmr`, and in fact 2020-era `librustzcash_mmr_append` didn't need it either.)

```suggestion
```

daira (commented):
Post-hoc ACK

daira:
Post-hoc review of this change together with the zcash-test-vectors reference generator + vectors (zcash/zcash-test-vectors#131), checked against ZIP 258's "Changes to ZIP 221".

**TL;DR:** the spec (ZIP 258) ↔ this PR ↔ the zcash/zcash-test-vectors#131 generator are mutually consistent, and the `zip_0221_v{1,2,3}.rs` vectors committed here are **byte-for-byte identical** to #131's generated output. `src/test_vectors.rs` is a genuine cross-validation (real `Tree` vs the reference vectors: leaf + root serialization, `hashChainHistoryRoot`, peaks), which also pins down the peak-bagging order.

One non-blocking note: that cross-repo vector copy is **manual and un-CI'd**, and these were taken from #131 while it is still open — so they should be re-synced from the final #131 when it lands (ideally a regenerate-and-diff CI check, so drift can't hide silently).

Full review (with the consistency table and the couple of minor notes): https://github.com/zcash/zcash-test-vectors/pull/131#issuecomment-4857569305

Posted with the assistance of Claude Opus 4.8.

mcakr616-del (commented):
Güzel

PR #2422, opened 2026-06-17
feat: propose transaction can spend transparent utxos

### LINKED ISSUE #2350: zcash_client_backend: Support spending transparent UTXOs in `propose_transaction`
At present, `propose_transaction` only allows spending of shielded TXOs, but for general t->t support we need to relax this restriction.

---

schell:
These changes:
* Add a `TransparentSpendPolicy` enumeration used to specify the intent of `propose_transaction`.
* Add `CoinbaseFilter::NonCoinbaseOnly` to remove coinbase UTXOs from `propose_transaction`, while still allowing coinbase to be shielded with `propose_shielding`.
* Add `InputSource::select_spendable_transparent_outputs`, a value-bounded transparent input gather (ordered by descending value, stopping once the requested value is covered) so wallets with large numbers of transparent UTXOs don't have to materialize their whole UTXO set for a small transfer.
* The gather step accounts for its own ZIP 317 marginal fee cost as it accumulates (via `StandardFeeRule::Zip317`), so it rarely needs a second round trip to correct an under-estimated fee headroom.
* Refactors the transparent SQL query into something re-usable now that we have more than one callsite using it.

Fixes #2350 
Supersedes https://github.com/zcash/librustzcash/pull/2419.


zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,111 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+#[cfg(feature = "transparent-inputs")]
+impl TransparentSpendPolicy {
+    /// Creates a policy that only spends from shielded UTXOs.
+    pub fn shielded_only() -> Self {
+        Self::ShieldedOnly
+    }
+
+    /// Creates a policy that spends from arbitrary transparent receivers
+    /// belonging to the account.
+    pub fn any_account_taddr() -> Self {
```
nullcopy: (nit) s/any_account_taddr()/from_any_account_transparent_address()/

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,111 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+        Self::AnyAccountTaddr
+    }
+
+    /// Creates a policy that only spends from the specified transparent addresses,
+    /// potentially leaking them. (`ANY_TADDR`)
+    pub fn from_transparent_addresses(taddrs: NonEmpty<TransparentAddress>) -> Self {
```
nullcopy: (nit) s/from_transparent_addresses()/from_specific_transparent_addresses()/

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,111 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+}
+
+#[cfg(feature = "transparent-inputs")]
+/// Specifies the wallet's intent to spend transparent UTXOs in a transfer.
+///
+/// Spending transparent funds links the chosen transparent addresses on-chain,
+/// reducing privacy; callers must opt in explicitly. Corresponds to the legacy
+/// `AllowTransparentAddressLinking` privacy policy / `ANY_TADDR`.
+#[derive(Default)]
+pub enum TransparentSpendPolicy {
```
nullcopy, line 481: I feel like `TransparentOutputFilter` exists for this purpose. Did you consider adding new variants to that enum, instead of creating this new enum?

nullcopy: Also feels like a hybrid with zallet's `PrivacyPolicy`. If for some reason this is not suitable as a `TransparentOutputFilter`, then I'd wonder if we should move `PrivacyPolicy` from zallet into `zcash_client_backend` and use it here instead.

Either way, I'm not convinced a new enum is the way to go, but maybe I'm missing something

nullcopy (commented):
Flushing comments

zcash_client_backend/src/data_api/wallet/input_selection.rs
schell: > I feel like TransparentOutputFilter exists for this purpose. Did you consider adding new variants to that enum, instead of creating this new enum?

I didn't, and it's an interesting proposal - but ultimately I think they're different things: 

```rust
pub enum TransparentOutputFilter {
    /// Select all spendable transparent outputs.
    #[default]
    All,
    /// Select only coinbase transparent outputs.
    ///
    /// Coinbase transactions are identified by having `tx_index == 0` within
    /// their containing block. Outputs for which the transaction index is
    /// unknown are conservatively treated as non-coinbase and will be excluded
    /// when this filter is active.
    CoinbaseOnly,
}

pub enum TransparentSpendPolicy {
    /// Do not spend any transparent UTXOs (default; fully-shielded behavior).
    #[default]
    ShieldedOnly,
    /// Spend from arbitrary transparent receivers belonging to the account, as
    /// needed to satisfy the request. The proposer chooses the addresses,
    /// potentially linking them. (`ANY_TADDR`)
    AnyAccountTaddr,
    /// Spend only from the specified transparent addresses, intentionally
    /// linking them. (Wrap a non-empty set in an immutable type to enforce
    /// safe construction, per AGENTS.md.)
    FromAddresses(NonEmptyBTreeSet<TransparentAddress>),
}
```
If we combine them we have something like: 

```rust
pub enum TransparentOutputFilter {
    /// Select only coinbase transparent outputs.
    ///
    /// Coinbase transactions are identified by having `tx_index == 0` within
    /// their containing block. Outputs for which the transaction index is
    /// unknown are conservatively treated as non-coinbase and will be excluded
    /// when this filter is active.
    CoinbaseOnly,
    /// Only select shielded UTXOs
    #[default]
    ShieldedOnly,
    /// Spend from arbitrary transparent receivers belonging to the account, as
    /// needed to satisfy the request. The proposer chooses the addresses,
    /// potentially linking them. (`ANY_TADDR`)
    ///
    /// Subsumes `TransparentSpendPolicy::All`
    AnyAccountTaddr,
    /// Spend only from the specified transparent addresses, intentionally
    /// linking them. (Wrap a non-empty set in an immutable type to enforce
    /// safe construction, per AGENTS.md.)
    FromAddresses(NonEmptyBTreeSet<TransparentAddress>),
}
``` 

Which is conceptually appealing, but the callsites where `TransparentOutputFilter` is used don't lend themselves to adding `TransparentSpendPolicy::ShieldedOnly` (like `propose_shielding`, `gather_shielding_inputs` and `get_spendable_transparent_outputs`).

We could get rid of the `ShieldedOnly` variant, have `TransparentSpendPolicy` take a `CoinbaseOnly` variant, and then have `propose_transaction` take `Option<&TransparentSpendPolicy>` - that may work. It means adding the ability for `propose_transaction` to spend coinbase, though, instead of requiring the coinbase be shielded prior to spending. This would mean there are effectively two paths to spending coinbase.

If that's acceptable then this could be a nice simplification, but it may also be more work than it's worth, considering `TransparentOutputFilter` is really `coinbase_only: bool`. Maybe `TransparentOutputFilter` should be renamed to `CoinbaseOutputFilter`?

schell: > Also feels like a hybrid with zallet's PrivacyPolicy. If for some reason this is not suitable as a TransparentOutputFilter, then I'd wonder if we should move PrivacyPolicy from zallet into zcash_client_backend and use it here instead.

I think like your first idea, there's some good overlap between the two enums:

```rust
pub(super) enum PrivacyPolicy {
    /// Only allow fully-shielded transactions (involving a single shielded value pool).
    FullPrivacy,

    /// Allow funds to cross between shielded value pools, revealing the amount that
    /// crosses pools.
    AllowRevealedAmounts,

    /// Allow transparent recipients.
    ///
    /// This also implies revealing information described under
    /// [`PrivacyPolicy::AllowRevealedAmounts`].
    AllowRevealedRecipients,

    /// Allow transparent funds to be spent, revealing the sending addresses and amounts.
    ///
    /// This implies revealing information described under
    /// [`PrivacyPolicy::AllowRevealedAmounts`].
    AllowRevealedSenders,

    /// Allow transaction to both spend transparent funds and have transparent recipients.
    ///
    /// This implies revealing information described under
    /// [`PrivacyPolicy::AllowRevealedSenders`] and
    /// [`PrivacyPolicy::AllowRevealedRecipients`].
    AllowFullyTransparent,

    /// Allow selecting transparent coins from the full account, rather than just the
    /// funds sent to the transparent receiver in the provided Unified Address.
    ///
    /// This implies revealing information described under
    /// [`PrivacyPolicy::AllowRevealedSenders`].
    AllowLinkingAccountAddresses,

    /// Allow the transaction to reveal any information necessary to create it.
    ///
    /// This implies revealing information described under
    /// [`PrivacyPolicy::AllowFullyTransparent`] and
    /// [`PrivacyPolicy::AllowLinkingAccountAddresses`].
    NoPrivacy,
}
```

...but `PrivacyPolicy` is too granular to deal with inside `propose_transaction` without a really big overhaul (we'd also have to update `PrivacyPolicy` to include specific transparent addresses to achieve the goal of this PR, and that would have consequences for `zallet`'s RPC shape). 

I think the best thing to do here is to rename `TransparentOutputFilter` to better suit its use case as a coinbase thing. I know that's _neither_ of your suggestions, but maybe that's enough? 🤪

schell: For posterity - we talked it over on slack and decided that was enough. I'll make a note to float the idea of a more thorough rework of note selection.

nuttycom: The updated enumeration here looks reasonable to me.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -202,6 +202,7 @@ pub trait InputSelector {
         account: <Self::InputSource as InputSource>::AccountId,
         transaction_request: TransactionRequest,
         change_strategy: &ChangeT,
+        #[cfg(feature = "transparent-inputs")] spend_policy: &TransparentSpendPolicy,
```
nuttycom, line 212: With the ability to specify transparent addresses and/or "all account addresses", we *do* need to apply coinbase maturity rules here. The intersection with the `CoinbaseFilter` for `propose_shielding` is worth thinking about more - what special rules are needed for input selection in `propose_transaction` in the context of having coinbase outputs in the wallet?

schell: There aren't any special rules needed, per say, as the transparent notes are selected with the call graph `propose_transaction` -> `gather_account_transparent_inputs` -> `get_spendable_transparent_outputs`, which queries notes using `excluding_immature_coinbase_outputs`. This is the same path that `z_shieldcoinbase` takes, so it's re-using the coinbase maturity rules there.

But now I've convinced myself that even this implementation does indeed spend coinbase without shielding first. If you pass either `TransparentSpendPolicy::AnyAccountTaddr` _or_ specific taddrs - coinbase UTXOs that pass the maturity test will be selected.

I also notice that we've got some redundant coinbase maturity testing in our SQL...

schell: I think at some point we need to refactor the note selection to be more general - essentially take a record of parameters to match notes against.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1271,7 +1274,13 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
          AND u.id NOT IN ({}) -- and the output is unspent
          AND ({}) -- exclude likely-spent wallet-internal ephemeral outputs
          AND ({}) -- exclude immature coinbase outputs
-         AND (:coinbase_only == 0 OR IFNULL(t.tx_index, 1) == 0) -- coinbase filter: unknown tx_index defaults to 1 (non-coinbase) to avoid false positives",
+         AND (
+             :coinbase_filter == 0
+             OR (:coinbase_filter == 1 AND IFNULL(t.tx_index, 1) == 0)
+             OR (:coinbase_filter == 2 AND IFNULL(t.tx_index, 1) != 0)
+         ) -- coinbase filter: 0 = all, 1 = coinbase-only, 2 = non-coinbase-only;
+           -- unknown tx_index defaults to 1 (non-coinbase) to avoid false positives,
```
dannywillems, line 1354: This is an interesting condition. It seems to be reusing existing code, therefore I won't dig more into this during my review, but keeping a note for later. I would have expected the field `tx_index` to always be non-null when we want to spend funds.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,114 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+            tail: Default::default(),
+        })
+    }
+}
+
+#[cfg(feature = "transparent-inputs")]
+/// Gathers the account's spendable transparent UTXOs from each of the provided source
+/// addresses, for use as inputs to a general (non-shielding) transfer.
+///
+/// Unlike [`gather_shielding_inputs`], the source addresses here are the account's own
```
dannywillems: Nit: maybe this comment is unnecessary as the routine name is explicit?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,114 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+    pub fn from_one_transparent_address(taddr: TransparentAddress) -> Self {
+        Self::FromAddresses(NonEmptyBTreeSet {
+            head: taddr,
+            tail: Default::default(),
+        })
+    }
+}
+
+#[cfg(feature = "transparent-inputs")]
+/// Gathers the account's spendable transparent UTXOs from each of the provided source
```
dannywillems: The routine has an explicit filter, `CoinbaseFilter::NonCoinbaseOnly`, which has the following important hypothesis:
> We only want non-coinbase transparent UTXOs. This forces users to shield their coinbase first by using `propose_shielding` with the appropriate coinbase filter.

Neither the routine name nor the documentation mentions this hypothesis. As a caller, I would expect to be warned of this - as I should not be reading the routine body (except if I'm investigating a bug). Should we reflect in either the name or the documentation?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,114 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+///
+/// [`WalletRead::get_transparent_receivers`]: crate::data_api::WalletRead::get_transparent_receivers
+#[allow(clippy::type_complexity)]
+fn gather_account_transparent_inputs<DbT, ChangeErrT>(
+    wallet_db: &DbT,
+    source_addrs: &[TransparentAddress],
+    target_height: TargetHeight,
+    confirmations_policy: ConfirmationsPolicy,
+) -> Result<
+    Vec<WalletTransparentOutput<()>>,
```
dannywillems: Is it expected to allow transparent addresses from multiple account ID (i.e. "erasing" the type parameter of `WalletTransparentOutput` by replacing with a `unit` type ? Or do we want the source addresses to be all from the same account?

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -581,6 +695,36 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                }
+                TransparentSpendPolicy::AnyAccountTaddr => {
+                    // return _all_ the known, non-ephemeral transparent receivers
+                    let t_recvs = wallet_db
+                        .get_transparent_receivers(account, true, true)
+                        .map_err(InputSelectorError::DataSource)?;
+                    t_recvs.into_keys().collect()
+                }
+                TransparentSpendPolicy::FromAddresses(set) => {
+                    let mut t_recvs = vec![set.head];
```
dannywillems: Nit: we could have a `Into` implementation for this, if we think that the new structure `NonEmptyBTreeSet` will stay around.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -755,6 +758,8 @@ where
         &change_strategy,
         request,
         confirmations_policy,
+        #[cfg(feature = "transparent-inputs")]
+        &Default::default(),
```
dannywillems: Nit: personal taste - I like knowing what is exactly the kind of value passed to the method, without going through the method documentation. `TransparentSpendPolicy::default()` looks better in that case.

dannywillems: Note that default values are defined in variables above, even if it could be inline. Following the same strategy makes the code more consistent.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -999,6 +999,8 @@ where
             change_strategy,
             request,
             confirmations_policy,
+            #[cfg(feature = "transparent-inputs")]
+            &Default::default(),
```
dannywillems: Same as https://github.com/zcash/librustzcash/pull/2422/changes#r3466046468

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1042,6 +1044,47 @@ where
...
```
dannywillems: Same as https://github.com/zcash/librustzcash/pull/2422/changes#r3466046468

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1035,11 +1188,13 @@ struct EphemeralStepConfig {
     tr1_payment_pools: BTreeMap<usize, PoolType>,
 }
 
+#[allow(clippy::too_many_arguments)]
 fn build_proposal<FeeRuleT: FeeRule + Clone, NoteRef>(
     fee_rule: &FeeRuleT,
     tr0_balance: TransactionBalance,
     target_height: TargetHeight,
     shielded_inputs: Option<ShieldedInputs<NoteRef>>,
+    transparent_inputs: Vec<WalletTransparentOutput<()>>,
```
dannywillems, line 1452: Related to https://github.com/zcash/librustzcash/pull/2422/changes#r3466012268, it seems that no account ID is propagated. Therefore it is a feature out of the scope of this patch.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6629,23 +6630,42 @@ where
...
+    let non_coinbase_utxos = st
+        .wallet()
+        .get_spendable_transparent_outputs(
+            &t_addr,
+            target_height,
+            ConfirmationsPolicy::default(),
+            CoinbaseFilter::NonCoinbaseOnly,
+        )
+        .unwrap();
+    assert_eq!(
```
dannywillems, line 6677: Nit: reading the code above confirms the behavior is the expected one by checking the length equality, see L6590. We check the value above, but maybe we could check the combinaison txid + tx index too?
The current version follows the previous assertion; it is coherent with the existing code. That's why I mention this as a nit, and it could be out of the scope of this patch.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6666,6 +6686,26 @@ where
...
+            &single_output_change_strategy(StandardFeeRule::Zip317, None, T::SHIELDED_PROTOCOL),
+            Zatoshis::from_u64(10000).unwrap(),
+            &[t_addr],
+            account,
+            ConfirmationsPolicy::default(),
+            CoinbaseFilter::NonCoinbaseOnly,
+        )
+        .unwrap();
+    let non_coinbase_inputs = proposal_non_coinbase.steps().first().transparent_inputs();
+    assert_eq!(
```
dannywillems, line 6719: Same comment as https://github.com/zcash/librustzcash/pull/2422/changes#r3466124281

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -1582,3 +1583,275 @@ where
...
+    .unwrap()
+}
+
+/// Regression test enforcing the privacy invariant: with the default
+/// [`TransparentSpendPolicy::ShieldedOnly`] policy, a transfer must NOT silently spend
+/// the account's transparent UTXOs as a fallback. An account holding only transparent
+/// funds must fail with [`InsufficientFunds`] rather than producing a t->t proposal.
+///
+/// [`InsufficientFunds`]: crate::data_api::error::Error::InsufficientFunds
+pub fn propose_t2t_shielded_only_is_insufficient<DSF>(dsf: DSF, cache: impl TestCache)
```
dannywillems, line 1882: This only tests with a non-coinbase transaction. Should we have a proptest with the three cases? (only coinbase tx, only non coinbase tx, mix of coinbase tx + non coinbase tx).

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -1582,3 +1583,275 @@ where
...
+
+/// Sets up a wallet whose account holds no shielded notes and a single spendable
+/// transparent UTXO at its default external transparent receiver, returning the test
+/// state, the account, the funding outpoint, and the UTXO value.
+///
+/// The chain is seeded with blocks containing notes that do *not* belong to the wallet,
+/// so that target/anchor heights resolve while the account remains shielded-empty. This
+/// isolates the transparent-UTXO selection path.
+#[allow(clippy::type_complexity)]
+fn setup_transparent_only_account<DSF>(
```
dannywillems, line 1808: What about using a proptest to have randomized tests?

Additionnally, see https://github.com/zcash/librustzcash/pull/2422/changes#r3466145113

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -1582,3 +1583,275 @@ where
...
+        result,
+        Err(crate::data_api::error::Error::InsufficientFunds { .. }),
+        "shielded-only policy must not spend transparent UTXOs as a fallback",
+    );
+}
+
+/// With [`TransparentSpendPolicy::AnyAccountTaddr`] (the legacy `ANY_TADDR` behavior), a
+/// transfer may spend the account's transparent UTXOs. Verifies that the funding UTXO is
+/// selected as a transparent input and that the proposal balance is consistent.
+pub fn propose_t2t_any_account_taddr<DSF>(dsf: DSF, cache: impl TestCache)
```
dannywillems, line 1919: As in [this comment](https://github.com/zcash/librustzcash/pull/2422/changes#r3466145113), I notice we only handle the case of non-coinbase tx. Should we support more cases? Maybe using `proptest` could help.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -1582,3 +1583,275 @@ where
...
+    // input total minus the explicit payment.
+    assert_eq!(
+        step.balance().total(),
+        (utxo_value - transfer_amount).unwrap(),
+    );
+    assert!(step.balance().fee_required() > Zatoshis::ZERO);
+    assert!(!step.balance().proposed_change().is_empty());
+}
+
+/// With [`TransparentSpendPolicy::FromAddresses`], only the explicitly named transparent
```
dannywillems, line 2213: Same comment as above.
Additionnaly, I would test more cases like a wallet with multiple accounts. `proptest` could be used to cover a wide range of scenarii.

dannywillems (dismissed):
utACK 4b270bf2711ce8ce1a2c6957e58230ef53170adc, with suggestions to improve test coverage and documentation.

dannywillems:
Also, approving as it actually fixes the ticket https://github.com/zcash/librustzcash/issues/2350. I didn't check specific cases like conditionals on the block height, etc.

zcash_client_backend/src/data_api.rs
```diff
@@ -1396,17 +1396,19 @@ impl NoteFilter {
 /// Controls which transparent outputs are eligible for selection.
 #[cfg(feature = "transparent-inputs")]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
-pub enum TransparentOutputFilter {
+pub enum CoinbaseFilter {
```
nuttycom, line 1429: I'm a bit confused by this change. There's no rule that says you can't spend non-coinbase outputs alongside coinbase outputs; the rule is that if any coinbase outputs are selected, there may not be any transparent outputs or change. So `NonCoinbaseOnly` seems like the wrong sense here. This change also isn't well-motivated by the CHANGELOG or the documentation of this type. Can you explain more thoroughly?

nuttycom (dismissed):

zcash_client_sqlite/src/wallet/transparent.rs
schell: Yeah, it's unfortunate, but whether `tx_index` is populated depends on the path that the transaction took to get into the database.

zcash_client_backend/src/data_api/wallet/input_selection.rs
schell: This function is taken wholesale from PR that this one supersedes, #2419. I'm not opposed to redundant documentation, though, so long as it stays in sync with the actual code.

zcash_client_backend/src/data_api/wallet/input_selection.rs
schell: As far as I can tell, the `AccountId` type parameter on `WalletTransparentOutput` exists for account _data_ erasure (for ensuring privacy at the type level), not for enforcing single-account sourcing. So your question is orthogonal to the type itself.

Which accounts are sourced from in this function depend on the addresses given in `source_addrs: &[TransparentAddress],` which in the call graph come from a match on `TransparentSpendPolicy` in `propose_transaction`. If the caller of `propose_transaction` uses a policy with specific transparent addresses that come from more than one account then I guess (I haven't looked) it would select notes from more than one account. The other policy variants select no transparent notes, or select notes from the account that's in scope, respectively.

zcash_client_backend/src/data_api/wallet/input_selection.rs
schell: Totally. I think as soon as we have another module that wants to use a `NonEmptyBTreeSet` (or even a `NonEmpty<F, T>`, I guess) we could add some more scaffolding around it, but for now we just need a quick way for folks to ensure there's one or more addresses.

schell: I miss applicative, here, lol.

zcash_client_backend/src/data_api/testing/pool.rs
schell: Yeah, we can strengthen the checks here for both `NonCoinbaseOnly` and `CoinbaseOnly` variants.

zcash_client_backend/src/data_api/testing/transparent.rs
schell: I think `proptest`s would be nice, but given the time constraints it's out of scope.

zcash_client_backend/src/data_api/testing/transparent.rs
schell: Same as above - `proptest` _should_ be added, but I don't think this PR is the place.

zcash_client_backend/src/data_api/testing/transparent.rs
schell: Agreed, but same.

zcash_client_backend/src/data_api.rs
schell: `CoinbaseFilter` isn't an encoding of the coinbase-spend consensus rule. It's a UTXO-selection filter parameter to `get_spendable_transparent_outputs`. It answers "which UTXOs should the selector even consider?". 

You're right that the rule is transaction-level. Spending coinbase is fine as long as the tx is fully shielded with no transparent change. We're not modeling that rule in the filter. For general transfers, which can produce transparent change, we conservatively keep coinbase UTXOs out of the candidate set, which is precisely the "filter coinbase out of `propose_transaction`" behavior you asked for in standup, and `NonCoinbaseOnly` is just the selection that does it. 

The problem here, and the reason I didn't simply filter after calling `gather_acount_transparent_inputs` is that there's no existing mechanism to filter coinbase out at that layer. We'd have to surface coinbase-ness on `WalletTransparentOutput`, which is a larger public API change. Pushing the filter into the SQL query with a `CoinbaseFilter` variant was the smaller, more localized change.

Yeah, the documentation can be improved, which I'll do.

nuttycom: Okay, fair enough.

dannywillems: But why did the name of the enum changed? The enum is not only about coinbases.

zcash_client_backend/src/data_api.rs
```diff
@@ -1394,19 +1394,38 @@ impl NoteFilter {
 }
 
 /// Controls which transparent outputs are eligible for selection.
+///
+/// This is purely an input-selection control: it determines which transparent
+/// UTXOs a selector is permitted to consider. It does not model or enforce any
```
dannywillems: Why is this added?
```
/// It does not model or enforce any
/// consensus rule about coinbase spends, and callers should not treat it as
/// such.
```
Is it a necessary information?

zcash_client_backend/src/data_api.rs
```diff
@@ -1394,19 +1394,38 @@ impl NoteFilter {
...
     ///
     /// Coinbase transactions are identified by having `tx_index == 0` within
     /// their containing block. Outputs for which the transaction index is
     /// unknown are conservatively treated as non-coinbase and will be excluded
     /// when this filter is active.
     CoinbaseOnly,
+    /// Select only non-coinbase transparent outputs.
+    ///
+    /// Used for general (non-shielding) transfers, which may produce transparent
+    /// change. The consensus rules require that any transaction *spending* a
```
dannywillems: Can the documentation be summarized? It seems to be too verbose.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -414,6 +415,118 @@ impl orchard_fees::OutputView for OrchardPayment {
...
+    DbT: InputSource,
+{
+    let mut inputs = vec![];
+    for addr in source_addrs {
+        let utxos = wallet_db
+            .get_spendable_transparent_outputs(
+                addr,
+                target_height,
+                confirmations_policy,
+                // We exclude coinbase UTXOs from general-transfer input
```
dannywillems: Can this comment be summarized with the first two lines and a half? The rest seems unnecessary to me.

zcash_client_backend/src/data_api.rs
schell: The rename to `CoinbaseFilter` came out of the earlier thread with @nullcopy where we concluded the type is effectively a coinbase selector, as opposed to a broader spend policy like `TransparentSpendPolicy`/zallet's `PrivacyPolicy`. All three current variants are about the coinbase/non-coinbase distinction, so `CoinbaseFilter` describes the axis it filters on.

dannywillems (dismissed):
utACK debbd24f13d968dc93bbfca71dfde8b15acafe00

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: We don't have any plans to support spending from multiple accounts anywhere in `zcash_client_backend`.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: > I miss applicative, here, lol.

This is me most days.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -581,6 +695,36 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                TransparentSpendPolicy::ShieldedOnly => {
+                    // No transparent addresses required
+                    vec![]
+                }
+                TransparentSpendPolicy::AnyAccountTaddr => {
+                    // return _all_ the known, non-ephemeral transparent receivers
+                    let t_recvs = wallet_db
+                        .get_transparent_receivers(account, true, true)
+                        .map_err(InputSelectorError::DataSource)?;
+                    t_recvs.into_keys().collect()
```
nuttycom: This is going to fall over instantaneously for large wallets; you should expect wallets with >500k receivers.

nuttycom: We need to start thinking seriously about how operations like this behave for exchange wallet users of Zallet; if a wallet has half a million or a million transparent receivers, this approach is going to be totally intractable. We need some way to bound the set - ideally by total value of returned outputs, so that the selection query on the backend can properly "size" the inputs to the request - we don't want a strategy that returns $25M worth of ZEC in 0.25-zec outputs, for example.

nuttycom (changes requested):
Sorry for catching this late in the process, but I hadn't had time for close review before now. We need to improve the efficiency of the wallet overall, but for transparent handling in particular we need some approach to "right-sizing" output selection.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -482,7 +545,11 @@ impl<DbT> Default for GreedyInputSelector<DbT> {
     }
 }
 
-impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
+impl<DbT> InputSelector for GreedyInputSelector<DbT>
+where
+    DbT: WalletRead
```
nuttycom: This is a substantial change to the constraints here; we previously intentionally only required `InputSource` because Zingo uses this code, and implements `InputSource` but not `WalletRead`. Any functions that we need from `WalletRead` to make this work should probably be moved to `InputSource` instead.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: This is still an issue at this point.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -687,42 +683,64 @@ where
...
+            } else {
+                // Compute the initial bound for the value-bounded transparent gather: the
+                // total payment amount. We don't (yet) include a fee headroom in the
+                // initial bound because the SQLite gather's `estimated_additional_fees`
+                // parameter is currently informational; the loop re-gathers on
+                // `InsufficientFunds` if the bound was too low.
+                //
+                // A `TransactionRequest` may have payments with no fixed amount (e.g. for
+                // `propose_send_max`); in that case we fall back to `TargetValue::AllFunds`
+                // so the gather returns everything.
```
nuttycom: We're still going to be bounded by a maximum number of inputs that can fit in a block, and I think we should probably limit any given gather to 10% of that (which is what https://github.com/zcash/librustzcash/pull/2489 does)

zcash_client_sqlite/src/wallet/init/migrations/add_transparent_value_index.rs
```diff
@@ -0,0 +1,115 @@
...
+use super::utxos_to_txos;
+use crate::wallet::init::WalletMigrationError;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x8b1a3c2d_4e5f_6a7b_8c9d_0e1f2a3b4c5d);
+
+// `utxos_to_txos` is the migration that first creates the `transparent_received_outputs`
+// table (with `value_zat`), so depending on it is sufficient to ensure the table exists
+// before the index is created. The `fix_transparent_received_outputs` migration rebuilds
+// the table later, but the column name and meaning are preserved, so the index remains
+// valid across the rebuild.
```
nuttycom: I would take the more conservative route of depending on `fix_transparent_received_outputs` here; if it drops and re-creates the table, I'm not confident that the index would actually get re-created.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1393,6 +1393,80 @@ pub(crate) fn get_spendable_transparent_outputs_for_addresses<P: consensus::Para
...
+///
+/// The query is a single SQL statement that orders eligible UTXOs by descending value (using
+/// the `idx_transparent_received_outputs_value_zat` index) and lets the Rust side accumulate
+/// values until the bound is met. This bounds the work done in SQLite to the prefix of the
+/// table that can possibly satisfy the request, rather than materializing the wallet's full
+/// UTXO set as the unbounded gather did.
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn select_spendable_transparent_outputs<P: consensus::Parameters>(
+    conn: &rusqlite::Connection,
+    _params: &P,
```
nuttycom: If this parameter is unused why is it passed?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1393,6 +1393,80 @@ pub(crate) fn get_spendable_transparent_outputs_for_addresses<P: consensus::Para
...
+            let base = u64::from(z);
+            let headroom = estimated_additional_fees.map(u64::from).unwrap_or(0);
+            base.checked_add(headroom)
+        }
+        TargetValue::AllFunds(_) => None,
+    };
+
+    let coinbase_filter = coinbase_filter_encoding(output_filter);
+
+    let mut stmt_utxos = conn.prepare_cached(&spendable_transparent_outputs_query(
```
nuttycom, line 1540: Should `spendable_transparent_outputs_query` take a limit, so that the maximum number of inputs we select is bounded? Same thing as #2489 again.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -2140,21 +2140,25 @@ where
             ConfirmationsPolicy::MIN,
             CoinbaseFilter::AllTransparentOutputs,
             TargetValue::AtLeast(target),
-            None,
+            &StandardFeeRule::Zip317,
```
nuttycom, line 2390: 👍 StandardFeeRule is the right choice.

zcash_client_backend/src/data_api/wallet/input_selection.rs
schell: Right, we should probably fall back on that. 👍

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -423,6 +431,67 @@ const MAX_BLOCK_BYTES: usize = 2_000_000;
...
+    pub tail: BTreeSet<T>,
+}
+
+#[cfg(feature = "transparent-inputs")]
+/// Specifies the wallet's intent to spend transparent UTXOs in a transfer.
+///
+/// Spending transparent funds links the chosen transparent addresses on-chain,
+/// reducing privacy; callers must opt in explicitly. Corresponds to the legacy
+/// `AllowTransparentAddressLinking` privacy policy / `ANY_TADDR`.
+#[derive(Default)]
```
nuttycom, line 480: nit
```suggestion
/// Specifies the wallet's intent to spend transparent UTXOs in a transfer.
///
/// Spending transparent funds links the chosen transparent addresses on-chain,
/// reducing privacy; callers must opt in explicitly. Corresponds to the legacy
/// `AllowTransparentAddressLinking` privacy policy / `ANY_TADDR`.
#[derive(Default)]
#[cfg(feature = "transparent-inputs")]
```

nuttycom (dismissed):
utACK 5b67ceaf322b585f83243446759cfbc0fb9e51d5

On reflection, the issue with bounding the number of inputs and outputs to a transaction isn't limited to the transparent pool; we need to solve this for shielded transfers as well. So I'm okay following up with a PR that takes a more holistic approach, and I'm also okay with adding to this PR (since this is what's introducing new methods for bounding selected value, it might make sense to set the overall upper bounds for transparent here)

nuttycom (dismissed):
re-utACK e399e8cf1546bbc84c20dfaabbf8d784820aa0c3

nuttycom (approved):
re-utACK 9a80c1c64c5d0bac5f281bcba0841ea72b3b79b1

nuttycom:
# Review — transparent UTXO spending in `propose_transaction`

Verified by: full diff read against merge base `6788e5b0`, `cargo check -p zcash_client_backend` (default features), and `cargo test -p zcash_client_sqlite --all-features` on the new tests (all pass).

The overall design is good — the default (`ShieldedOnly`) preserves existing behavior and skips the gather entirely, the ZIP 317-aware accumulation in SQL-order is well thought out (O(1) incremental fee computation, with a documented validity argument), the re-gather fallback with the `transparent_gather_grew` progress flag terminates correctly, and re-parenting the index migration onto `fix_transparent_received_outputs` (because that migration drops/recreates the table, destroying indices) is a genuinely important catch. Test coverage is strong, including a privacy regression test for the default policy. But there is one significant bug.

## Bug: `TransparentSpendPolicy::FromAddresses` is not enforced

`spend_policy` is consulted in exactly two places in `GreedyInputSelector::propose_transaction`, both as `matches!(spend_policy, TransparentSpendPolicy::ShieldedOnly)`. The `FromAddresses` variant is never destructured anywhere in the workspace — the gather is account-wide (`select_spendable_transparent_outputs(account, …)` has no address parameter), and there is no post-gather filtering. So `FromAddresses` behaves identically to `AnyAccountTaddr`: the proposal can spend from — and link — addresses the caller explicitly excluded. That inverts the variant's privacy contract ("Spend only from the specified transparent addresses").

The `propose_t2t_from_addresses` test passes coincidentally. Both UTXOs have value 100,000, so the `ORDER BY u.value_zat DESC, u.output_index` sort ties (both `output_index` 0) and falls back to rowid/insertion order; the named address's UTXO was inserted first, and the value-bounded gather stops after one UTXO because 100,000 minus fee already covers the 40,000 payment. Fund the *unnamed* address with 150,000 instead and the test will fail. Worth strengthening the test that way regardless of how the fix lands.

For the fix: post-gather filtering is not sufficient, because excluded UTXOs would consume the value bound and under-fill the set. The address predicate needs to be pushed into the gather — e.g. an `Option<&BTreeSet<TransparentAddress>>` (or similar) parameter on `select_spendable_transparent_outputs`, wired to an address predicate in `spendable_transparent_outputs_query` the same way `get_spendable_transparent_outputs_for_addresses` already does with `rarray`.

## Other findings, by severity

**New warnings in the default-features build.** `amount_at_transparent_gather` (`input_selection.rs`) is declared outside the `cfg(transparent-inputs)` block but only read inside cfg-gated code, producing `unused_variables` + `unused_mut` when the feature is off — this will fail any `-D warnings` CI pass. Restructuring the gather block to return `(inputs, amount)` as a tuple would keep both bindings inside the cfg. (The third warning in that build, the unused `transparent_fees` import, is pre-existing at the merge base.)

**Duplicate changelog entry.** The rebase re-added the `spend-index` feature-flag bullet, which now appears twice in the same 0.24.0 "Added" section of `zcash_client_backend/CHANGELOG.md` — it already existed on `main`.

**Design question — transparent-first vs. "as needed".** Under `AnyAccountTaddr`/`FromAddresses`, the gather runs up-front, bounded by the full payment total — so transparent funds are spent *preferentially*, with shielded notes only topping up the remainder. The doc on `AnyAccountTaddr` says "as needed to satisfy the request," which reads as fallback semantics (shielded first, transparent only for the shortfall). One behavior or the other should be chosen deliberately: transparent-first has an upside (change goes to a shielded pool, so it opportunistically consolidates/shields transparent funds the moment the user opts in), but it also reveals more than necessary when shielded funds alone would have sufficed. If transparent-first is the intent, the docs should say so explicitly; if fallback is the intent, the gather bound should be the post-shielded-selection shortfall.

**`NonEmptyBTreeSet` API.** Public `head`/`tail` fields mean the invariants a "set" implies aren't maintained — `head` can duplicate a `tail` element, and `head` need not sort before `tail`, so it's really "one element plus a set." It has no `iter`/`contains`/`len`, no `Debug`/`PartialEq` derives, and (per the bug above) nothing consumes it yet. Since the constructor already takes `nonempty::NonEmpty`, consider whether the custom type pays its way — `NonEmpty<TransparentAddress>` alone, or private fields with a constructor and accessors, would both be safer. A general-purpose collection type also sits oddly in `input_selection`; if it stays, a more foundational home may be warranted.

**`TransparentSpendPolicy` derives only `Default`.** As a public API type that callers construct, store, and log, it should at least derive `Debug`, `Clone`, and `PartialEq`/`Eq` (all derivable given its contents).

**Doc nits.**
- `from_specific_transparent_addresses` is labeled "(`ANY_TADDR`)" — that legacy label belongs to `AnyAccountTaddr` only. Also "potentially leaking them" vs. the variant doc's "intentionally linking them" — the latter phrasing is the accurate one.
- `propose_transfer`'s doc comment wasn't updated to describe the new `spend_policy` parameter.

**Edge cases worth a look (low likelihood, non-blocking).**
- The payment-total overflow arm reports `InsufficientFunds { available: Zatoshis::ZERO, required: MAX_MONEY }` — `available: ZERO` is fabricated. Fine as a defensive branch, but if ZIP 321 request construction already caps totals, this may be unreachable and worth a comment either way.
- Transparent `DustInputs` handling: after pruning dust from `transparent_inputs`, the iteration falls through to the loop-progress check, which returns `InsufficientFunds` unless shielded notes grew — so a transparent-only account never gets a second balance computation against the pruned (possibly sufficient) set. Hard to reach under ZIP 317 because the gather filters `value_zat > MARGINAL_FEE`, but reachable with other change strategies or large P2SH inputs. Relatedly, a subsequent re-gather rebuilds the set from scratch and can re-introduce previously pruned dust.

## Positives worth keeping

- The privacy regression test (`propose_t2t_shielded_only_is_insufficient`) directly pins the "no silent transparent fallback" invariant.
- The re-gather test is thoughtfully constructed — it independently reproduces the initial gather to prove the success could only have come via the fallback path.
- The `max_inputs` cap (added in the latest revision) closes the unbounded-transaction-size hole for dust-heavy wallets and is tested at the cap boundary.
- Fee-aware gather documentation is exemplary: the `StandardFeeRule` fixing rationale, the linearity argument, and the heuristic-bound caveat are all stated where an implementor needs them.
- Migration hygiene: UUID fixed, dependency re-parented with a clear comment about the table rebuild, and `verify_schema` updated.
- CHANGELOG entries follow the project convention — Added entries as pointers, Changed entries explaining adaptation.

**Bottom line:** the architecture and test discipline are solid, but the `FromAddresses` non-enforcement is a must-fix privacy bug before merge (with a test that doesn't rely on insertion-order ties), and the default-features warnings plus the duplicated changelog bullet should be cleaned up alongside it.

---
*This review was performed with the assistance of Claude Code (Claude Fable 5).*

nuttycom (changes requested):
Blocking on a number of Fable-discovered issues.

nuttycom (approved):
utACK [delta since my last review](https://github.com/zcash/librustzcash/pull/2422/changes/9a80c1c64c5d0bac5f281bcba0841ea72b3b79b1..3424baf798b164772dc765dec2dd5602020fd52e)

PR #2423, opened 2026-06-17
zcash_primitives: Add feature flagging to enable `sapling` and `orchard` features to be independently enabled.

### LINKED ISSUE #1162: `zcash_primitives`: Add `orchard` and `sapling feature flags.
This has been requested by the Brave folks in order to allow them to avoid a dependency on the `sapling-crypto` crate, and we will add a similar flag for `orchard` for symmetry.

nuttycom:
This is a simpler alternative to #1388

Closes #1162, which asked for these flags so that a consumer can depend on `zcash_primitives` without pulling in `sapling-crypto`. Today both `orchard` and `sapling` are unconditional `workspace = true` dependencies, and the existing `circuits` / `multicore` / `std` features only forward to their features rather than making the crates optional.

### DIFF
--- a/zcash_primitives/src/transaction/components/orchard_raw.rs
+++ b/zcash_primitives/src/transaction/components/orchard_raw.rs
@@ -1,0 +1,262 @@
+//! Parsing and serialization of the Orchard components of a transaction when the `orchard`
+//! feature is disabled.
+//!
+//! When the `orchard` feature is not enabled, the `orchard` crate (and its `pasta_curves`
+//! dependency) is not compiled, so a transaction's Orchard bundle cannot be converted into
+//! the typed [`orchard::Bundle`] domain representation. These functions parse the exact same
+//! binary wire format as the typed path, retaining the parsed bytes in [`RawOrchardBundle`]
+//! so that the transaction can be re-serialized byte-identically and its consensus txid /
+//! authorizing-data digests computed (see [`crate::transaction::txid`]).
+//!
+//! Parsing here is *lenient*: it validates field lengths, the Orchard flags bits, and (under
+//! [`ProofSizeEnforcement::Strict`]) the canonical proof size, but it does NOT validate
+//! Pallas point membership of `cv_net`/`nullifier`/`rk`/`cmx`/`anchor` (that requires
+//! `pasta_curves`, which is only compiled with the `orchard` feature). Point-membership
+//! validation is therefore deferred to the point at which the bytes would be converted into
+//! a typed bundle, which requires the `orchard` feature. This is sufficient to re-serialize
+//! byte-identically and to compute the correct consensus txid.
+
+use alloc::vec::Vec;
+use corez::io::{self, Read, Write};
+
+use zcash_encoding::{Array, CompactSize, Vector};
+use zcash_protocol::value::ZatBalance;
+
+use crate::transaction::Transaction;
+
+const FLAG_SPENDS_ENABLED: u8 = 0b0000_0001;
+const FLAG_OUTPUTS_ENABLED: u8 = 0b0000_0010;
+const FLAGS_EXPECTED_UNSET: u8 = !(FLAG_SPENDS_ENABLED | FLAG_OUTPUTS_ENABLED);
+
+const ENC_CIPHERTEXT_SIZE: usize = 580;
+const OUT_CIPHERTEXT_SIZE: usize = 80;
+
+// The canonical byte length of an Orchard proof is a fixed base plus a fixed per-action
+// contribution; these constants mirror `orchard::Proof::expected_proof_size`. A proof of any
+// other length is non-canonical (e.g. padded; GHSA-2x4w-pxqw-58v9).
+const PROOF_BASE_SIZE: usize = 2720;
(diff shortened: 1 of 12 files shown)


zcash_primitives/src/transaction/mod.rs
```diff
@@ -1317,7 +1407,14 @@ pub enum DigestError {
...
 
-#[cfg(any(test, feature = "test-dependencies"))]
+// The proptest strategies in this module build typed Sapling and Orchard bundles, and so
+// require both protocol features. Feature-off builds exercise transaction parsing via the
+// static ZIP 244 / ZIP 143 / ZIP 243 test vectors instead.
+#[cfg(all(
+    any(test, feature = "test-dependencies"),
+    feature = "sapling",
+    feature = "orchard"
+))]
```
nuttycom, line 1183: Not all of the generators here depend on the shielded protocols; use more selective flagging internal to the module. (follow-up, not for this PR).

zcash_primitives/src/transaction/txid.rs
```diff
@@ -166,6 +175,28 @@ pub(crate) fn hash_sapling_spends<A: sapling::bundle::Authorization>(
...
+            nh.write_all(&s_spend.cv).unwrap();
+            nh.write_all(&s_spend.anchor).unwrap();
+            nh.write_all(&s_spend.rk).unwrap();
+        }
+
+        h.write_all(ch.finalize().as_bytes()).unwrap();
+        h.write_all(nh.finalize().as_bytes()).unwrap();
+    }
+    h.finalize()
+}
```
ValarDragon: It'd be nice for:
- performance concerns upstream ( https://github.com/valargroup/zebra/pull/131/changes#diff-db57c4ae23524be17d3e41ce4c68574dcf972c6d713ea295bca8c89d9ed8310fR107-R142 )
- Code re-use

to be able to always call this, and have the typed version call into this.

I'd like for Zebra to be able to call this as well in the relevant spot!

ValarDragon: Maybe we can always compile with the raw types, but the PCZT only does one. Then we can expose both sighashes here. (And they can internally re-use code via passing in every arg)

zcash_primitives/src/transaction/components/orchard_raw.rs
```diff
@@ -0,0 +1,338 @@
...
+#[derive(Clone, Debug)]
+pub struct RawOrchardAction {
+    pub(crate) cv_net: [u8; 32],
+    pub(crate) nullifier: [u8; 32],
+    pub(crate) rk: [u8; 32],
+    pub(crate) cmx: [u8; 32],
+    pub(crate) epk_bytes: [u8; 32],
+    pub(crate) enc_ciphertext: [u8; ENC_CIPHERTEXT_SIZE],
+    pub(crate) out_ciphertext: [u8; OUT_CIPHERTEXT_SIZE],
+    pub(crate) spend_auth_sig: [u8; 64],
```
nuttycom, line 64: The `spend_auth_sig` field here is populated from `vSpendAuthSigsOrchard`, which must be separately from parsing the action; see `with_auth` below.

zcash_primitives/src/transaction/components/orchard_raw.rs
```diff
@@ -0,0 +1,338 @@
...
+fn read_array<const N: usize, R: Read>(mut reader: R) -> io::Result<[u8; N]> {
+    let mut bytes = [0u8; N];
+    reader.read_exact(&mut bytes)?;
+    Ok(bytes)
+}
+
+fn read_flags<R: Read>(mut reader: R) -> io::Result<u8> {
+    let mut byte = [0u8; 1];
+    reader.read_exact(&mut byte)?;
+    if byte[0] & FLAGS_EXPECTED_UNSET != 0 {
```
nuttycom, line 142: This check will become contextual as of NU6.3

zcash_primitives/src/transaction/components/sapling_raw.rs
```diff
@@ -0,0 +1,409 @@
...
+}
+
+/// Reads a value commitment, enforcing canonical encoding and rejecting small-order points
+/// (matching `sapling::value::ValueCommitment::from_bytes_not_small_order`).
+fn read_cv<R: Read>(mut reader: R) -> io::Result<[u8; 32]> {
+    let mut bytes = [0u8; 32];
+    reader.read_exact(&mut bytes)?;
+    // Matches `sapling::value::ValueCommitment::from_bytes_not_small_order`, which decodes the
+    // canonical Jubjub point encoding and rejects small-order points.
+    let affine = jubjub::AffinePoint::from_bytes(bytes);
```
nuttycom: If we're disabling the `sapling-crypto` feature, shouldn't this also remove the `jubjub` dependency from the feature stack, making it not possible to perform these checks here?

nullcopy (dismissed):
utACK ecda7f49396ad985209fa3aaefc59c6f42685782 modulo conflicts

nuttycom:
Back to the drawing board. I thought this was pretty close, but I didn't realize that the `circuits` feature hadn't been split; this requires more changes to the transaction builder, which was the original place where Str4d's #1388 ran into trouble.

PR #2426, opened 2026-06-18
zcash_client_sqlite: Fix InvalidParameterName in delete_account

### LINKED ISSUE #2425: zcash_client_sqlite: delete_account fails if cross-account transactions exist
Originally reported by @noop-sk:
```
Bug Report: delete_account fails with Invalid parameter name: :address when sent_notes reference the deleted account

  Affects: zcash_client_sqlite 0.20.2 and 0.21.0
  Severity: High — renders Keystone hardware wallet disconnect permanently broken for affected users
  Reproducible: State-dependent (see trigger condition below)

  ---
  Summary

  WalletDb::delete_account panics with rusqlite::Error::InvalidParameterName(":address") when the account being deleted is referenced by any sent_notes row. The SQL statement uses the parameter name
  :to_address, but the binding code passes :address. This mismatch is silent when the loop body never executes (no matching rows), which is why the bug went unnoticed — it only manifests once a user has
  made a cross-account internal transfer to the Keystone account.

  ---
  Location

  zcash_client_sqlite/src/wallet.rs, function delete_account:

  let mut update_sent_note = conn.prepare(
      r#"
      UPDATE sent_notes
      SET to_address = :to_address, to_account_id = NULL   // ← expects :to_address
      WHERE id = :sent_note_id
      "#,
  )?;

  // ...

      if let Some(address) = row.get::<_, Option<String>>("to_address")? {
          update_sent_note.execute(named_params![
              ":sent_note_id": row.get::<_, i64>("sent_note_id")?,
              ":address": address   // ← BUG: should be ":to_address"
          ])?;
      }
  }

  ---
  Fix

  One-line change:

  // Before
  ":address": address

  // After
  ":to_address": address

  ---
  Trigger Condition

  The bug is only reachable when the following query returns at least one row:

  SELECT sn.id AS sent_note_id, ...
  FROM sent_notes sn
  JOIN v_received_outputs ro ON ro.sent_note_id = sn.id
  JOIN addresses ON addresses.id = ro.address_id
  JOIN accounts ta ON ta.id = [sn.to](http://sn.to/)_account_id
  WHERE ta.uuid = :account_uuid   -- the account being deleted

  In practice this means: a transaction was sent to an address belonging to the account being deleted, and that output was received and scanned. In a multi-account wallet (e.g. Zashi account + Keystone
  hardware wallet), this state is reached when the user sends ZEC from account A to an address of account B, or via auto-shielding of transparent Keystone funds.

  ---
  Observed Crash (Android)

  java.lang.RuntimeException: Invalid parameter name: :address
      at cash.z.ecc.android.sdk.internal.jni.RustBackend.deleteAccount(Native Method)
      ...
  Caused by: java.lang.RuntimeException: Invalid parameter name: :address
      at cash.z.ecc.android.sdk.internal.jni.RustBackend.deleteAccount(Native Method)

  Full JNI call chain:
  Synchronizer.deleteAccount(AccountUuid) → TypesafeBackendImpl.deleteAccount → RustBackend.deleteAccount (JNI) → wallet::delete_account → rusqlite InvalidParameterName(":address")

  App version: 3.6.0 (SDK 2.6.2-SNAPSHOT, zcash_client_sqlite 0.21.0, mainnet)

  ---
  Why It Appeared "Suddenly"

  The defect has been present since at least zcash_client_sqlite 0.20.2. delete_account succeeded for all users whose Keystone account had never appeared as to_account_id in sent_notes. The first
  cross-account transfer (internal send, auto-shield of Keystone transparent balance, or any multi-account operation) creates the triggering database state, after which every subsequent disconnect attempt
  fails permanently.

  ---
  Impact

  Users cannot disconnect their Keystone hardware wallet. The operation fails at the native layer and the error surfaces to the UI. The wallet database is left in a consistent state (the account is not
  deleted), so there is no data loss — but the account cannot be removed.
```

nullcopy:
Closes https://github.com/zcash/librustzcash/issues/2425
Closes [COR-1364](https://linear.app/zodl/issue/COR-1364/zcash-client-sqlite-delete-account-fails-if-cross-account-transactions)

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -3295,6 +3295,110 @@ where
     );
 }
 
+/// Regression test for a bug in which [`WalletWrite::delete_account`] failed with a
+/// `rusqlite::Error::InvalidParameterName(":address")` panic when the account being
+/// deleted was referenced by a `sent_notes` row via its `to_account_id` column.
+///
+/// The triggering state is reached when a transaction is sent from one account in the
+/// wallet to an address belonging to a second account in the same wallet, and the
+/// transaction is then decrypted via [`decrypt_and_store_transaction`] so that the
+/// cross-account transfer is recorded with a non-null `to_account_id` and a received
+/// output that has an associated address. Deleting the recipient account then exercises
+/// the `sent_notes` update path inside `delete_account`.
+///
+/// [`WalletWrite::delete_account`]: crate::data_api::WalletWrite::delete_account
+pub fn account_deletion_with_internal_transfer<T: ShieldedPoolTester, DSF>(
+    ds_factory: DSF,
+    cache: impl TestCache,
+) where
+    DSF: DataStoreFactory,
+    <DSF as DataStoreFactory>::DataStore: Reset,
+{
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(ds_factory)
+        .with_block_cache(cache)
+        .build();
+
+    // Add two accounts to the wallet, derived from the same seed.
+    let seed = Secret::new([0u8; 32].to_vec());
+    let birthday = AccountBirthday::from_sapling_activation(st.network(), BlockHash([0; 32]));
+    let (account1, usk1) = st
+        .wallet_mut()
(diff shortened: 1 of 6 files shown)


nuttycom (approved):
utACK

PR #2427, opened 2026-06-19
Move PCZT v1 into versioned module

ValarDragon:
## Summary

- Move the existing PCZT v1 implementation into an inline `pczt::v1` module.
- Preserve existing top-level `pczt::Pczt`, `ParseError`, `ExtractError`, and `EffectsOnly` paths via re-exports.
- Add a planned changelog entry for the new public module.

## Validation

- `cargo fmt --all`

## API surface

- Adds public module `pczt::v1`.
- Adds public paths `pczt::v1::Pczt`, `pczt::v1::ParseError`, and feature-gated `pczt::v1::ExtractError` / `pczt::v1::EffectsOnly`.
- Changes `Pczt` fields `global`, `transparent`, `sapling`, and `orchard` from private-at-crate-root to `pub(crate)` inside `v1` so existing role modules can construct and destructure the type.

### DIFF
--- a/pczt/src/lib.rs
+++ b/pczt/src/lib.rs
@@ -139,7 +139,7 @@ impl Pczt {
         let version = u32::from_le_bytes(bytes[4..8].try_into().unwrap());
         match version {
             PCZT_VERSION_1 => postcard::from_bytes::<v1::Pczt>(&bytes[8..])
-                .map(Into::into)
+                .map(Pczt::from)
                 .map_err(ParseError::Invalid),
             _ => Err(ParseError::UnknownVersion(version)),
         }
--- a/pczt/src/lib.rs
+++ b/pczt/src/lib.rs
@@ -24,7 +24,6 @@ extern crate alloc;
 use alloc::vec::Vec;
 
 use getset::Getters;
-use serde::{Deserialize, Serialize};
 
 #[cfg(all(
     any(feature = "io-finalizer", feature = "signer", feature = "tx-extractor"),
@@ -55,15 +54,20 @@ pub mod orchard;
 pub mod sapling;
 pub mod transparent;
 
-const MAGIC_BYTES: &[u8] = b"PCZT";
-const PCZT_VERSION_1: u32 = 1;
+pub(crate) const MAGIC_BYTES: &[u8] = b"PCZT";
+pub(crate) const PCZT_VERSION_1: u32 = 1;
+
+/// Parses a PCZT from its encoding.
+pub fn parse(bytes: &[u8]) -> Result<Pczt, ParseError> {
+    Pczt::parse(bytes)
+}
 
 /// A partially-created Zcash transaction.
-#[derive(Clone, Debug, Serialize, Deserialize, Getters)]
+#[derive(Clone, Debug, Getters)]
 pub struct Pczt {
     /// Global fields that are relevant to the transaction as a whole.
     #[getset(get = "pub")]
-    global: common::Global,
+    pub(crate) global: common::Global,
 
     //
     // Protocol-specific fields.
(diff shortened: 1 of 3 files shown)


pczt/src/lib.rs
```diff
@@ -90,21 +136,28 @@ impl Pczt {
...
-
-        // This is a v1 PCZT.
-        postcard::from_bytes(&bytes[8..]).map_err(ParseError::Invalid)
     }
 
     /// Serializes this PCZT.
     pub fn serialize(&self) -> Vec<u8> {
+        self.serialize_legacy_v1()
+            .expect("v1 can encode the current logical PCZT")
+    }
```
ValarDragon: I expect this API to transform to:
- If ironwood -> v2
- else v1

or we just delete the function.

pczt/src/lib.rs
```diff
@@ -74,13 +78,55 @@ pub struct Pczt {
...
+            orchard: pczt.orchard.clone(),
+        }
+    }
+}
+
+/// Errors that can occur while serializing a PCZT using the legacy v1
+/// encoding.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum LegacyV1SerializeError {}
```
ValarDragon: Bikeshed of if this should be LegacyV1 or just V1

nuttycom: Just V1

pczt/src/lib.rs
nuttycom: (1) remove "legacy"
(2) why is the "expect" needed at all here? I don't see `serialize_legacy_v1` being fallible?

pczt/src/lib.rs
```diff
@@ -90,21 +136,28 @@ impl Pczt {
...
+            .expect("v1 can encode the current logical PCZT")
+    }
+
+    /// Serializes this PCZT using the legacy v1 encoding.
+    pub fn serialize_legacy_v1(&self) -> Result<Vec<u8>, LegacyV1SerializeError> {
         let mut bytes = vec![];
         bytes.extend_from_slice(MAGIC_BYTES);
         bytes.extend_from_slice(&PCZT_VERSION_1.to_le_bytes());
-        postcard::to_extend(self, bytes).expect("can serialize into memory")
+        Ok(postcard::to_extend(&v1::Pczt::from(self), bytes).expect("can serialize into memory"))
```
nuttycom: Looks like we shouldn't need the `LegacyV1SerializeError` at all?)

nuttycom: Oh that error type will have some actual constructors in the future?

pczt/src/lib.rs
nuttycom: Oh I think I get it - that error will get some constructors in the future. Let's just make a SerializeError type that we use for both v1 and v2 serialization errors, that seems simpler? Make it non-exhaustive.

pczt/src/lib.rs
```diff
@@ -74,13 +78,55 @@ pub struct Pczt {
...
+            global: pczt.global,
+            transparent: pczt.transparent,
+            sapling: pczt.sapling,
+            orchard: pczt.orchard,
+        }
+    }
 }
 
+impl From<&Pczt> for v1::Pczt {
+    fn from(pczt: &Pczt) -> Self {
```
nuttycom: This approach seems good, low-cost and gives us a semantic PCZT type that's decoupled from the serialization.

nuttycom (dismissed):
utACK modulo naming/error nits

pczt/src/lib.rs
czarcas7ic: Changed to V1 https://github.com/zcash/librustzcash/blob/8fd1f60fbd9d5dd87835fb62d9eb85b51d6dc0bb/pczt/src/lib.rs#L124-L155

pczt/src/lib.rs
czarcas7ic: https://github.com/zcash/librustzcash/pull/2427/commits/4a86392902c803461d0a7e112f04a980aab18b96

Legacy removed in earlier push, and made this a SerializeError type.

pczt/src/lib.rs
nuttycom: This is okay for now, it did just occur to me though that once `v2` is introduced this will need to become a `TryFrom`, because `Pczt -> v1::Pczt` will be fallible.

pczt/src/lib.rs
```diff
@@ -90,21 +135,28 @@ impl Pczt {
         if &bytes[..4] != MAGIC_BYTES {
             return Err(ParseError::NotPczt);
         }
+
         let version = u32::from_le_bytes(bytes[4..8].try_into().unwrap());
-        if version != PCZT_VERSION_1 {
-            return Err(ParseError::UnknownVersion(version));
+        match version {
+            PCZT_VERSION_1 => postcard::from_bytes::<v1::Pczt>(&bytes[8..])
+                .map(Into::into)
```
nuttycom:
```suggestion
                .map(Pczt::from)
```
nit: `into` is a frequent source of typechecker inference failures; it's better to always use `From`.

pczt/src/lib.rs
nuttycom: The `expect` here will be removed, and the signature changed to return `Result` in the next PR?

nuttycom (approved):
utACK https://github.com/zcash/librustzcash/pull/2427/commits/6808acdea035ee3879bb04834288851aa95aab43

Current beta lint failures will be fixed in a subsequent commit by having a constructable error type.

PR #2429, opened 2026-06-19
Add Ironwood V6 transaction ID support

### LINKED ISSUE #2451: zcash_primitives: Implement TxVersion::V6 parsing & serialization
---

### LINKED ISSUE #2452: zcash_primitives: Implement NU6.3 txid, auth data &  signature hashing.
---

czarcas7ic:
Summary:
- Adds the V6 transaction ID and signature hash plumbing needed for Ironwood transaction support.
- Carries the related transaction version, PCZT, and ZIP 233 test coverage onto the Ironwood split-1 foundation branch.
- Keeps the ZIP 233 transaction test ordered with the base branch to avoid unnecessary diff noise.

Testing:
- `RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo test -p zcash_primitives --all-features transaction::tests::zip_0233`

Solves COR-1385


Cargo.toml
```diff
@@ -68,7 +68,7 @@ redjubjub = { version = "0.8", default-features = false }
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
 
 # - Orchard
-orchard = { version = "0.14", default-features = false }
+orchard = { git = "https://github.com/valargroup/qr_orchard", rev = "6f1a5eaddcf0e016101c0c62a477d48f31e7a83f", default-features = false }
```
czarcas7ic: This will change once https://github.com/zcash/orchard/pull/504 is merged as well as https://github.com/valargroup/qr_orchard/pull/1

pczt/src/orchard.rs
```diff
@@ -18,6 +16,27 @@ use crate::{
...
+
+#[cfg(feature = "orchard")]
+pub(crate) fn legacy_bundle_format() -> orchard::bundle::BundleFormat {
+    orchard::bundle::BundleFormat::PreNu6_3
+}
+
+#[cfg(feature = "orchard")]
+pub(crate) fn legacy_note_version() -> orchard::note::NoteVersion {
+    legacy_bundle_protocol().default_note_version()
+}
```
ValarDragon: So this part should get cleaned up in PCZT v2, going to commit a fix to remove this one. This PR's diff should just be on the zcash_primitives side

ValarDragon:
Adam's force pushing to fix conflicts + make it all a clean commit

zcash_client_backend/src/fees/orchard.rs
```diff
@@ -3,32 +3,40 @@
...
+use orchard::{BundleProtocol, builder::BundleType};
 use zcash_protocol::value::Zatoshis;
 
-/// A trait that provides a minimized view of Orchard bundle configuration
+pub(crate) fn transactional_action_count(
+    protocol: BundleProtocol,
+    num_spends: usize,
+    num_outputs: usize,
+) -> Result<usize, &'static str> {
+    BundleType::DEFAULT.num_actions(num_spends, num_outputs, protocol)
```
nuttycom, line 14: As we discussed yesterday, does this actually just depend on the number of spends and the number of outputs? I thought that we could potentially need a more complex view of the number of actions required based upon which are "change" actions. Or is that intended to come in later with the actual NU6.3 support when the bundle flags switch?

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -90,11 +90,23 @@ pub fn read_v5_bundle<R: Read>(
     }
 }
 
-#[cfg(any(zcash_unstable = "zfuture", zcash_unstable = "nu7"))]
+/// Reads an [`orchard::Bundle`] from a v5 transaction format.
+pub fn read_v5_bundle<R: Read>(
+    reader: R,
+    proof_size_enforcement: ProofSizeEnforcement,
+) -> io::Result<Option<orchard::Bundle<Authorized, ZatBalance>>> {
+    read_bundle(reader, proof_size_enforcement, BundleFormat::PreNu6_3)
```
nuttycom, line 108: Doesn't `read_v5_bundle` need to also take the consensus branch ID to make the determination of the format to read? Because NU6.3 doesn't disable v5 transactions.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -282,6 +289,37 @@ impl BuildConfig {
...
+        .checked_add(builder.changes().len())
+        .ok_or("num_outputs + num_changes overflowed")?;
+
+    let bundle_type = if is_coinbase {
+        orchard::builder::BundleType::Coinbase
+    } else {
+        orchard::builder::BundleType::DEFAULT
+    };
+
+    bundle_type.num_actions(num_spends, num_outputs, builder.protocol())
```
nuttycom, line 292: related/same question here.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1304,14 +1456,31 @@ pub trait TransactionDigest<A: Authorization> {
 
     fn digest_sapling(
         &self,
+        version: TxVersion,
         sapling_bundle: Option<&sapling::Bundle<A::SaplingAuth, ZatBalance>>,
     ) -> Self::SaplingDigest;
 
     fn digest_orchard(
         &self,
+        version: TxVersion,
```
nuttycom, line 1187: This works, but it's somewhat overloading the sighash digest version with the transaction version. In the ZIP 248 future, the digest algorithm will be independent of the transaction version, so while this is correct for now, in the future the digest algorithm identifier is probably what we'll want.

zcash_primitives/src/transaction/tests.rs
nuttycom, line 1: I have not yet thoroughly reviewed this file.

zcash_primitives/src/transaction/txid.rs
```diff
@@ -64,9 +66,66 @@ const ZCASH_SAPLING_OUTPUTS_NONCOMPACT_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxId
...
+fn orchard_commitment_domain(version: TxVersion) -> BundleCommitmentDomain {
+    #[cfg(zcash_unstable = "nu6.3")]
+    if matches!(version, TxVersion::V6) {
+        return BundleCommitmentDomain::ORCHARD_V6;
+    }
+
+    #[cfg(all(zcash_unstable = "nu7", not(zcash_unstable = "nu6.3")))]
+    if matches!(version, TxVersion::V6) {
+        return BundleCommitmentDomain::ORCHARD_V5_NU6_3;
+    }
```
nuttycom: The `cfg` directive here is nonsensical; you can't have two different values for the same `cfg` flag, so "and not" is tautological. Also why do we regress to ORCHARD_V5_NU6_3 under NU7?

zcash_primitives/src/transaction/txid.rs
```diff
@@ -64,9 +66,66 @@ const ZCASH_SAPLING_OUTPUTS_NONCOMPACT_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxId
...
+
+    #[cfg(zcash_unstable = "zfuture")]
+    if matches!(version, TxVersion::ZFuture) {
+        return BundleCommitmentDomain::ORCHARD_V5_NU6_3;
```
nuttycom: Same question about ORCHARD_V5_NU6_3 here.

zcash_primitives/src/transaction/txid.rs
nuttycom, line 1: I need to make an additional pass over this file in tandem with the spec.

nuttycom (changes requested):
A couple of bundle version errors are blocking; I will continue review here tomorrow.

zcash_client_backend/src/fees/orchard.rs
ValarDragon: Ah sorry forgot to update on that. With the cross-address restriction, as currently done in Sean's Orchard PR, each spend is making its own action per my understanding, with a dummy output.

This in particular is needed to prevent a privacy leakage with a quantum adversary. https://github.com/zcash/orchard/pull/504/changes/657b375c41734bc8779b394529216d952a272eb0

I hadn't realized the quantum leakage yday

zcash_primitives/src/transaction/components/orchard.rs
ValarDragon: reading the bundle should be equivalent, because the flags and serialization are identical pre/post upgrade for Orchard

Since its in the public API, not sure its worth breaking an additional break.

The read_v6_bundle is being re-used across Ironwood and Orchard though, hence needing the parameter passed into the caller.

We should definitely add code comment explaining this though

zcash_primitives/src/transaction/txid.rs
```diff
@@ -139,21 +198,30 @@ pub(crate) fn hash_tze_outputs(tze_outputs: &[TzeOut]) -> Blake2bHash {
...
+        let mut nh = hasher(sapling_spends_noncompact_personalization(version));
         for s_spend in shielded_spends {
             // we build the hash of nullifiers separately for compact blocks.
             ch.write_all(s_spend.nullifier().as_ref()).unwrap();
 
             nh.write_all(&s_spend.cv().to_bytes()).unwrap();
+            #[cfg(zcash_unstable = "nu6.3")]
+            if !matches!(version, TxVersion::V6) {
+                nh.write_all(&s_spend.anchor().to_repr()).unwrap();
+            }
```
nuttycom: nit: I prefer an explicit `match version { ... }` here without a default, because then we'll get a compilation error in the future when the `TxVersion` enum adds a new version. As an extra benefit, you then don't need the `not` pragma, I think, since the `zcash_unstable` annotation will guard `TxVersion::V6` and will just match nu6.3 (and maybe should also match nu7/zfuture)?

zcash_primitives/src/transaction/txid.rs
```diff
@@ -436,11 +529,86 @@ pub(crate) fn to_hash(
...
+    } else {
+        to_hash(
+            txversion,
+            consensus_branch_id,
+            digests.header_digest,
+            hash_transparent_txid_data(digests.transparent_digests.as_ref()),
+            digests.sapling_digest,
+            digests.orchard_digest,
+            #[cfg(zcash_unstable = "zfuture")]
+            digests.tze_digests.as_ref(),
```
nuttycom: This can now be removed; tze and zfuture support have been removed from upstream.

zcash_primitives/src/transaction/txid.rs
```diff
@@ -523,17 +694,43 @@ impl TransactionDigest<Authorized> for BlockTxCommitmentDigester {
...
+                b.authorizing_commitment_for_domain(orchard_commitment_domain(version))
+                    .0
+            },
+        )
+    }
+
+    #[cfg(zcash_unstable = "nu6.3")]
+    fn digest_ironwood(
+        &self,
+        _version: TxVersion,
```
nuttycom: This argument can be removed.

zcash_primitives/src/transaction/txid.rs
czarcas7ic: Addressed here https://github.com/zcash/librustzcash/pull/2429/commits/c663639194ff61df1b1f4fa7c792e00c8f69d14c

zcash_primitives/src/transaction/txid.rs
czarcas7ic: Zfuture removed https://github.com/zcash/librustzcash/pull/2429/commits/c663639194ff61df1b1f4fa7c792e00c8f69d14c

zcash_primitives/src/transaction/txid.rs
czarcas7ic: Addressed here https://github.com/zcash/librustzcash/pull/2429/commits/c663639194ff61df1b1f4fa7c792e00c8f69d14c

zcash_primitives/src/transaction/txid.rs
czarcas7ic: Removed from rebase

zcash_primitives/src/transaction/txid.rs
czarcas7ic: Removed here https://github.com/zcash/librustzcash/pull/2429/commits/2125bd3ea1b48f055962dc0e5a9b4e3a2a90d9ee

zcash_primitives/src/transaction/components/orchard.rs
czarcas7ic: Please let me know if this comment is sufficient, leaving unresolved until ack https://github.com/zcash/librustzcash/pull/2429/commits/1a5f2bca0ff4e652768716b82ee7cc04a4ce35e9

czarcas7ic:
The diff blew up from regenerating the test vectors here https://github.com/zcash/librustzcash/pull/2429/changes/815d44d7b2b50a99552885720063b98d03675f89

zcash_primitives/src/transaction/mod.rs
```diff
@@ -541,10 +615,12 @@ impl<A: Authorization> TransactionData<A> {
             Option<sapling::Bundle<A::SaplingAuth, ZatBalance>>,
         )
             -> Result<Option<sapling::Bundle<B::SaplingAuth, ZatBalance>>, E>,
-        f_orchard: impl FnOnce(
+        mut f_orchard: impl FnMut(
```
nuttycom, line 618: It's unclear why we need this change; it's not wrong, it just means that this isn't really a "map" operation because it can perform arbitrary side effects.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1168,6 +1292,11 @@ pub mod testing {
             transparent_bundle in transparent::arb_bundle(),
             sapling_bundle in sapling::arb_bundle_for_version(version),
             orchard_bundle in orchard::arb_bundle_for_version(version),
+            ironwood_bundle in if version.has_ironwood() {
+                orchard::arb_bundle_for_version(version).boxed()
+            } else {
+                Just(None).boxed()
+            },
```
nuttycom, line 1299: nonblocking: This looks wrong, particularly given that the whole generator is gated off for NU6.3. Does `orchard::arb_bundle_for_version` produce Ironwood bundles?

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1206,6 +1341,38 @@ pub mod testing {
...
+        pub fn arb_txdata(consensus_branch_id: BranchId)(
+            version in arb_tx_version(consensus_branch_id)
+        )(
+            lock_time in any::<u32>(),
+            expiry_height in any::<u32>(),
+            transparent_bundle in transparent::arb_bundle(),
+            sapling_bundle in sapling::arb_bundle_for_version(version),
+            orchard_bundle in orchard::arb_bundle_for_version(version),
+            ironwood_bundle in if version.has_ironwood() {
+                orchard::arb_bundle_for_version(version).boxed()
```
nuttycom, line 1360: Same question; is there no ironwood bundle generator?

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1194,6 +1324,11 @@ pub mod testing {
             transparent_bundle in transparent::arb_bundle(),
             sapling_bundle in sapling::arb_bundle_for_version(version),
             orchard_bundle in orchard::arb_bundle_for_version(version),
+            ironwood_bundle in if version.has_ironwood() {
+                orchard::arb_bundle_for_version(version).boxed()
```
nuttycom, line 1328: Same question; is there no ironwood bundle generator?

nuttycom (approved):
utACK 88082359cdcf8e994e7f60d2763389d196e6002f with nonblocking questions

zcash_primitives/src/transaction/mod.rs
czarcas7ic: Ironwood reuses the Orchard bundle type, so this closure has to run on both orchard_bundle and ironwood_bundle. Since a FnOnce can't be called twice, so it has to be FnMut. Please lmk if you have another suggestion for this.

zcash_primitives/src/transaction/mod.rs
czarcas7ic: An Ironwood bundle is an orchard::Bundle with the same V6 wire format and circuit. For V6 both has_orchard() and has_ironwood() are true, so generating one bundle per field matches a real V6 tx (which can carry either, both, or neither).

zcash_primitives/src/transaction/mod.rs
czarcas7ic: Replied above.

zcash_primitives/src/transaction/mod.rs
czarcas7ic: Replied above.

zcash_primitives/src/transaction/mod.rs
nuttycom: The alternative there would be to pass a `Fn` instead of a `FnOnce` - I guess though that `FnOnce` allows mutation in the calling scope, so maybe this would break things. `FnMut` is fine.

PR #2432, opened 2026-06-19
Adapt to upstream Orchard API

czarcas7ic:
Updates the Orchard dependency to upstream commit e81bccc6ab0e3235f01f10a75337c6f588e850f9 and adjusts the existing v5 Orchard and PCZT call sites for the new API shape.

Design notes:

- This PR is intentionally limited to updating `zcash_primitives` for the upstream Orchard API.
- Where the new API requires a bundle protocol, the existing behavior is preserved by inferring `PreNU6_3`.
- Where the new API requires `note_version`, the existing behavior is preserved by inferring `v2`.
- Follow up integration work will thread the protocol and note version from the correct sources. This first step keeps the old behavior while making `zcash_primitives` compile against the new Orchard interface.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ redjubjub = { version = "0.8", default-features = false }
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
 
 # - Orchard
-orchard = { git = "https://github.com/zcash/orchard", rev = "b2af0a11abe00f59c51258d349c2105fe7a16215", default-features = false }
+orchard = { version = "0.14", default-features = false }
 pasta_curves = "0.5"
 
 # - Transparent
@@ -186,6 +186,9 @@ zip32 = { version = "0.2", default-features = false }
 
 # Workaround for https://anonticket.torproject.org/user/projects/arti/issues/pending/4028/
 time-core = "=0.1.2"
+
+[patch.crates-io]
+orchard = { git = "https://github.com/zcash/orchard", rev = "b2af0a11abe00f59c51258d349c2105fe7a16215" }
 
 [workspace.metadata.release]
 consolidate-commits = false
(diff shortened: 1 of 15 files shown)


zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6097,7 +6097,9 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, Dsf>(
 
     // Create proofs.
     let sapling_prover = LocalTxProver::bundled();
-    let orchard_pk = ::orchard::circuit::ProvingKey::build();
+    let orchard_pk = ::orchard::circuit::ProvingKey::build(
+        ::orchard::circuit::OrchardCircuitVersion::FixedPostNu6_2,
+    );
```
ValarDragon, line 6102: This is the correct version for this test to match today's behavior

ValarDragon (dismissed):
Matched that every usage bump matches today

Cargo.toml
```diff
@@ -68,7 +68,7 @@ redjubjub = { version = "0.8", default-features = false }
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
 
 # - Orchard
-orchard = { version = "0.14", default-features = false }
+orchard = { git = "https://github.com/zcash/orchard", rev = "e81bccc6ab0e3235f01f10a75337c6f588e850f9", default-features = false }
```
nuttycom: This should be a `patch.crates-io` entry; it should not directly modify the Orchard dependency in this way.

nuttycom (changes requested):
Use a [patch.crates-io] directive for the orchard dependency, otherwise utACK 2de405fc521d787320cb25780501bfea4d97c2ea

Cargo.toml
czarcas7ic: https://github.com/zcash/librustzcash/pull/2432/commits/85c66404d241dff14109b9669b38931b3af98a5c

nuttycom (approved):
utACK 85c66404d241dff14109b9669b38931b3af98a5c

PR #2433, opened 2026-06-19
Add new errors to keep track of block height/range in CommitmentTree

dannywillems:


nuttycom (approved):
utACK 056960177626ef78fb58aaf9e7abb95d5d1f1682

PR #2436, opened 2026-06-20
Remove zcash_client_memory from the workspace

nuttycom:
The in-memory wallet backend now lives in its own repository at https://github.com/zcash/zcash_client_memory. Drop the crate from the workspace along with every reference to it: the workspace member, the README dependency graph, the dep-graph linter's crate list, the generated-file attribute, and the crate summary in AGENTS.md.


pacu (approved):
utACK

PR #2447, opened 2026-06-23
zcash_client_backend: recognize OP_RETURN outputs as data, not unsupported

nuttycom:
`detect_wallet_transparent_outputs` logged a `warn!` for every transparent output whose script kind could not be converted to an address, including `OP_RETURN` (nulldata) outputs. nulldata outputs are provably-unspendable data carriers with no recipient address and are never wallet outputs, so this flagged them as "unsupported script kind" once per output during scanning -- very noisy on chains with frequent OP_RETURN usage.

Recognize `ScriptKind::NullData` explicitly and skip it silently, while still logging other unrecognized script kinds and the un-parseable case.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -51,6 +51,10 @@ workspace.
     Tokio-based batch decryption engine for full blocks and transactions.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


nullcopy (approved):
utACK 8ed7506fdc05274b026ecf57272d7cc82a60c1bf

PR #2463, opened 2026-06-25
Add spend-index feature for per-outpoint transparent spend queries

nuttycom:
Adds a `spend-index` cargo feature (in zcash_client_backend and zcash_client_sqlite) for wallets whose chain-data source can resolve the spend of an individual transparent outpoint directly (e.g. a full node with a spent-outpoint index), as an alternative to address-based `TransactionsInvolvingAddress` requests.

- zcash_client_backend: gated `TransactionDataRequest::SpendsOfTransparentOutput(OutPoint)` request variant and `WalletWrite::notify_outpoint_unspent_checked`. The servicing contract documents that spentness should be taken from an authoritative source and the spend index used only to resolve the spender, with "spent but spender not yet resolvable" treated as retryable (see ZcashFoundation/zebra#10806).
- zcash_client_sqlite: when `spend-index` is enabled, `transaction_data_requests` emits the per-outpoint variant for spend detection instead of `TransactionsInvolvingAddress` (retained for ephemeral-address discovery and when the feature is disabled); implements `notify_outpoint_unspent_checked` by advancing the per-outpoint `max_observed_unspent_height` watermark.
- zcash_transparent: derive `Hash` for `OutPoint` (required for the `TransactionDataRequest` `Hash` derive; also a natural map/set key).

A unit test for the new emission/notify path is left as a TODO (noted in transparent.rs): populating the spend-search queue requires the low-level/block-scan store path.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1106,10 +1106,14 @@ pub enum TransactionDataRequest {
     /// transparent address is requested.
     ///
     /// Fully transparent transactions, and transactions that do not contain either shielded inputs
-    /// or shielded outputs belonging to the wallet, may not be discovered by the process of chain
-    /// scanning; as a consequence, the wallet must actively query to find transactions that spend
-    /// such funds. Ideally we'd be able to query by [`OutPoint`] but this is not currently
-    /// functionality that is supported by the light wallet server.
+    /// or shielded outputs belonging to the wallet, may not be discovered by the process of
+    /// out-of-order chain scanning due to race conditions related to advancing the transparent
+    /// address gap limit; as a consequence, the wallet must actively query to find transactions
+    /// that spend such funds. Ideally we'd be able to query by [`OutPoint`] but this is not
+    /// currently functionality that is supported by the light wallet server; for full-node wallets
+    /// or other arrangements that allow privacy-preserving retrieval of individually identifying
+    /// information such as backends that support private information retrieval (PIR) consider
+    /// enabling the `spend-index` feature.
     ///
     /// The caller evaluating this request on behalf of the wallet backend should respond to this
     /// request by detecting transactions involving the specified address within the provided block
@@ -1120,6 +1124,29 @@ pub enum TransactionDataRequest {
     /// `block_end_height - 1` as the `as_of_height` argument.
     #[cfg(feature = "transparent-inputs")]
     TransactionsInvolvingAddress(TransactionsInvolvingAddress),
+    /// The main-chain transaction that spends of a specific transparent output, or confirmation
+    /// that the output remains unspent, is requested.
+    ///
+    /// When the `spend-index` feature is enabled, `GetSpendingTx` requests will be emitted by the
+    /// backend instead of [`TransactionDataRequest::TransactionsInvolvingAddress`], in
+    /// circumstances when spentness determination is needed. This feature should only be enabled
+    /// for wallets whose chain-data source can resolve the spend of an individual outpoint
+    /// directly — for example a full node that maintains a spent-outpoint index. It avoids needing
+    /// to retrieve potentially large amounts of transaction history (in the cas of a
+    /// heavily-reused address) just to discover whether one output was spent.
+    ///
+    /// The caller evaluating this request on behalf of the wallet backend should determine whether
+    /// `outpoint` has been spent on the main chain. Spentness should be taken from an
+    /// authoritative source (e.g. the node's UTXO set); the spend index is used only to identify
+    /// the spending transaction. If the output is spent, the caller should provide the spending
+    /// transaction to [`wallet::decrypt_and_store_transaction`]. If the output is confirmed
+    /// unspent as of some height, the caller should invoke
+    /// [`WalletWrite::notify_output_verified_unspent`] with that height. If the output is known to
+    /// be spent but the spending transaction cannot yet be resolved (e.g. an index is still being
+    /// built — see ZcashFoundation/zebra#10806), the caller should do nothing, so that the request
+    /// is re-issued and retried later.
+    #[cfg(feature = "spend-index")]
+    GetSpendingTx(OutPoint),
 }
 
 impl TransactionDataRequest {
@@ -3304,6 +3331,25 @@ pub trait WalletWrite: WalletRead {
     ) -> Result<(), Self::Error> {
         unimplemented!(
             "WalletWrite::notify_address_checked must be overridden for wallets to use the `transparent-inputs` feature"
+        )
+    }
+
+    /// Notifies the wallet backend that a specific transparent output was confirmed unspent as of
+    /// the given height, in response to a [`TransactionDataRequest::GetSpendingTx`]
+    /// request.
+    ///
+    /// # Arguments
+    /// - `outpoint`: the transparent outpoint whose spend status was checked.
+    /// - `as_of_height`: the maximum height among blocks that were inspected, and through which
+    ///   the outpoint is confirmed to remain unspent.
+    #[cfg(feature = "spend-index")]
+    fn notify_output_verified_unspent(
+        &mut self,
+        _outpoint: OutPoint,
+        _as_of_height: BlockHeight,
+    ) -> Result<(), Self::Error> {
+        unimplemented!(
+            "WalletWrite::notify_output_verified_unspent must be overridden for wallets to use the `spend-index` feature"
         )
     }
 }
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -40,7 +40,9 @@ use zcash_keys::{
         transparent::gap_limits::{GapLimits, generate_address_list},
     },
 };
-use zcash_primitives::transaction::{builder::DEFAULT_TX_EXPIRY_DELTA, fees::zip317};
+#[cfg(not(feature = "spend-index"))]
+use zcash_primitives::transaction::builder::DEFAULT_TX_EXPIRY_DELTA;
+use zcash_primitives::transaction::fees::zip317;
 use zcash_protocol::{
     TxId,
     consensus::{self, BlockHeight, COINBASE_MATURITY_BLOCKS},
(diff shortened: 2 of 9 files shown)


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1648,6 +1650,47 @@ pub(crate) fn update_observed_unspent_heights<P: consensus::Parameters>(
...
+            ELSE max_observed_unspent_height
+         END
+         WHERE tro.output_index = :output_index
+         AND tro.transaction_id IN (
+             SELECT id_tx FROM transactions WHERE txid = :txid
+         )
+         AND tro.id NOT IN (
+             SELECT transparent_received_output_id
+             FROM transparent_received_output_spends
+         )",
```
nuttycom: Could this be implemented more efficiently using the `UPDATE...FROM` sqlite3 extension?

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1820,58 +1863,101 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+            "SELECT t.txid, ssq.output_index
+             FROM transparent_spend_search_queue ssq
+             JOIN transactions t ON t.id_tx = ssq.transaction_id
+             JOIN transparent_received_outputs tro
+                ON tro.transaction_id = ssq.transaction_id AND tro.output_index = ssq.output_index
+             JOIN addresses ON addresses.id = tro.address_id
+             LEFT OUTER JOIN transparent_received_output_spends tros
+                ON tros.transparent_received_output_id = tro.id
+             WHERE tros.transaction_id IS NULL
+             AND addresses.key_scope != :ephemeral_key_scope
```
nuttycom: Since we're assuming privacy-preserving retrieval of spend information, I don't think we need to exclude ephemeral address-associated keys here.

zcash_client_backend/src/data_api.rs
```diff
@@ -1106,10 +1106,14 @@ pub enum TransactionDataRequest {
...
     ///
     /// Fully transparent transactions, and transactions that do not contain either shielded inputs
-    /// or shielded outputs belonging to the wallet, may not be discovered by the process of chain
-    /// scanning; as a consequence, the wallet must actively query to find transactions that spend
-    /// such funds. Ideally we'd be able to query by [`OutPoint`] but this is not currently
-    /// functionality that is supported by the light wallet server.
+    /// or shielded outputs belonging to the wallet, may not be discovered by the process of
+    /// out-of-order chain scanning due to race conditions related to advancing the transparent
+    /// address gap limit; as a consequence, the wallet must actively query to find transactions
+    /// that spend such funds. Ideally we'd be able to query by [`OutPoint`] but this is not
```
nullcopy, line 1112: This note about what the lightwallet server currently supports feels unnecessary here. This would be a useful comment at the 'spend-index' feature: e.g. "only enable this feature if the lightwallet server supports it"

zcash_client_backend/src/data_api.rs
```diff
@@ -1120,6 +1124,29 @@ pub enum TransactionDataRequest {
...
+    /// authoritative source (e.g. the node's UTXO set); the spend index is used only to identify
+    /// the spending transaction. If the output is spent, the caller should provide the spending
+    /// transaction to [`wallet::decrypt_and_store_transaction`]. If the output is confirmed
+    /// unspent as of some height, the caller should invoke
+    /// [`WalletWrite::notify_output_verified_unspent`] with that height. If the output is known to
+    /// be spent but the spending transaction cannot yet be resolved (e.g. an index is still being
+    /// built — see ZcashFoundation/zebra#10806), the caller should do nothing, so that the request
+    /// is re-issued and retried later.
+    #[cfg(feature = "spend-index")]
+    GetSpendingTx(OutPoint),
```
nullcopy, line 1149: Might be nice to make this a batch query

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1648,6 +1650,47 @@ pub(crate) fn update_observed_unspent_heights<P: consensus::Parameters>(
...
+/// Sets the max observed unspent height for the unspent transparent output identified by the given
+/// outpoint to at least the given height (will not cause the height to decrease). Used to record
+/// the result of a [`TransactionDataRequest::GetSpendingTx`] check that found the
+/// output unspent.
+///
+/// [`TransactionDataRequest::GetSpendingTx`]: zcash_client_backend::data_api::TransactionDataRequest::GetSpendingTx
+#[cfg(feature = "spend-index")]
+pub(crate) fn update_observed_unspent_height_for_outpoint(
+    conn: &rusqlite::Transaction,
+    outpoint: &OutPoint,
```
nullcopy, line 1662: Same comment as above, might be nice to support batch queries

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1648,6 +1650,47 @@ pub(crate) fn update_observed_unspent_heights<P: consensus::Parameters>(
...
+    checked_at: BlockHeight,
+) -> Result<(), SqliteClientError> {
+    let chain_tip_height = chain_tip_height(conn)?.ok_or(SqliteClientError::ChainHeightUnknown)?;
+    let checked_at = std::cmp::min(checked_at, chain_tip_height);
```
nullcopy, line 1666: Why do we care if the checked height is above the chain tip? This feels like it could easily trip if we race a transaction that was confirmed in the latest block, before we may have updated our tip.

This will force us to retry after the tip updates, which doesn't seem necessary, since the lightwallet server is trusted and providing the tip updates as well

zcash_client_backend/src/data_api.rs
```diff
@@ -1120,6 +1124,29 @@ pub enum TransactionDataRequest {
...
     #[cfg(feature = "transparent-inputs")]
     TransactionsInvolvingAddress(TransactionsInvolvingAddress),
+    /// The main-chain transaction that spends of a specific transparent output, or confirmation
+    /// that the output remains unspent, is requested.
+    ///
+    /// When the `spend-index` feature is enabled, `GetSpendingTx` requests will be emitted by the
+    /// backend instead of [`TransactionDataRequest::TransactionsInvolvingAddress`], in
+    /// circumstances when spentness determination is needed. This feature should only be enabled
+    /// for wallets whose chain-data source can resolve the spend of an individual outpoint
+    /// directly — for example a full node that maintains a spent-outpoint index. It avoids needing
```
dannywillems, line 1134: > for example a full node that maintains a spent-outpoint index

I guess we expect that in Zaino? I see this comment [here](https://github.com/zcash/zallet/issues/507)

zcash_client_sqlite/src/wallet/transparent.rs
dannywillems, line 1666: Comment: this handles reorg.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1648,6 +1650,47 @@ pub(crate) fn update_observed_unspent_heights<P: consensus::Parameters>(
...
+
+    let mut stmt = conn.prepare(
+        "UPDATE transparent_received_outputs AS tro
+         SET max_observed_unspent_height = CASE
```
nullcopy, line 1670: Instead of casing the check-height, would it be better to filter `WHERE max_observed_unspent_height IS NULL OR max_observed_unspent_height < :checked_at`? The same criteria you use below to decide which outpoints to query

zcash_client_sqlite/src/wallet/transparent.rs
dannywillems: I guess it is expected.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1820,58 +1863,101 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+    // Create transaction data requests that can find spends of our received UTXOs.
+    //
+    // With the `spend-index` feature, the chain-data source can resolve the spend of an
+    // individual outpoint directly, so we request spends per-outpoint. Otherwise we fall back to
+    // address-based requests (which, so long as address-based transaction data requests are
+    // required at all, are served by address-based lookups rather than by querying the spends of
+    // the associated outpoints directly).
+    #[cfg(feature = "spend-index")]
+    let spend_search_requests = {
+        // Per-outpoint spend resolution is privacy-preserving (it does not correlate the
```
nullcopy, line 1875: > Per-outpoint spend resolution is privacy-preserving (it does not correlate the wallet's addresses to an untrusted server), so unlike the address-based path below there is no need to exclude ephemeral-address outpoints here.

This statement is only true if the requests are not temporally correlated and there is sufficient cover traffic at to the lightwallet server. I would not personally be comfortable with having my ephemeral outpoints queried in the same process.

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -2428,6 +2516,39 @@ mod tests {
...
+        // Exercises the `spend-index` `UPDATE ... FROM` in
+        // `update_observed_unspent_height_for_outpoint` (the outpoint matches no rows).
+        let tx = st.wallet().db().conn.unchecked_transaction().unwrap();
+        super::update_observed_unspent_height_for_outpoint(
+            &tx,
+            &OutPoint::new([1u8; 32], 0),
+            chain_tip,
+        )
+        .unwrap();
+        tx.commit().unwrap();
```
nullcopy, line 2549: Shouldn't you check the spend height was updated?

zcash_client_backend/src/data_api.rs
nuttycom: It's available in Zebra now with the `-indexer` feature enabled. There are some problems with it, but it's available.

zcash_client_sqlite/src/wallet/transparent.rs
nullcopy: Also need to test the negative case; attempting to reduce the spend height should be noop

nullcopy (changes requested):
Blocking 4cdba554f5545050da784fc9e1eb073ae958d819 to discuss privacy implications

Other comments are nice-to-have (IMO), but not blocking

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: I think it's better to have to re-check than it is to reference a block height that we don't know about anywhere; I don't know what invariants referencing such a height would break and I don't want to have to think about it. Remember that the "mempool height" or "target height" is used in a number of places, and we don't want to be in a situation where target heights could be confused for actual block heights.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: The whole gate for the spend-index feature is "I can retrieve spentness information in a privacy-preserving manner".

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1820,58 +1863,101 @@ pub(crate) fn transaction_data_requests<P: consensus::Parameters>(
...
+    // required at all, are served by address-based lookups rather than by querying the spends of
+    // the associated outpoints directly).
+    #[cfg(feature = "spend-index")]
+    let spend_search_requests = {
+        // Per-outpoint spend resolution is privacy-preserving (it does not correlate the
+        // wallet's addresses to an untrusted server), so unlike the address-based path below
+        // there is no need to exclude ephemeral-address outpoints here.
+        let mut spend_requests_stmt = conn.prepare_cached(
+            "SELECT t.txid, ssq.output_index
+             FROM transparent_spend_search_queue ssq
```
dannywillems, line 1880: Note/comment for me: this table is fed by `put_transaction_outputs`, with `queue_transaction_spend_detection`.

dannywillems (approved):
The SQL requests LGTM. I went through the code to check the different tables and how they were populated. They're behind `transparent-inputs`, which is activated by `spend-index`.

zcash_client_sqlite/src/wallet/transparent.rs
nullcopy: Ah, that was not clear to me. In that case, I have no further objections

nullcopy (approved):
@nuttycom and I discussed the requested changes, and agreed to defer/skip each
utACK 4cdba554f5545050da784fc9e1eb073ae958d819

PR #2480, opened 2026-06-26
Bump orchard to 30c4ea2 (feat/ironwood)

czarcas7ic:
Bumps the `orchard` git pin from `b2af0a11` to `30c4ea2` (current `zcash/orchard` `feat/ironwood` tip), realigning `feat/ironwood` off its divergent side-branch pin onto canonical orchard.

This is an API + light semantics migration:

- `BundleProtocol` → `BundlePoolRestrictions` (`OrchardPreNu6_3` → `OrchardNu6_2Only`, `OrchardPostNu6_3` → `OrchardNu6_3Onward`, `IronwoodPostNu6_3` → `IronwoodNu6_3Onward`).
- Commitment-domain API → fallible `(pool_restrictions, tx_version)` in `txid.rs`. Present bundles derive the pool restriction from their own flags, empty bundles from the tx version; txid/auth digests are byte-preserved.
- `BundleFormat` removed; (de)serialization takes a `BundlePoolRestrictions`. `read_v6_bundle`/`write_v6_bundle` take the pool restriction as an argument so the caller selects the Orchard or Ironwood v6 pool (the Ironwood pool permits the cross-address flag, the Orchard v6 pool does not; the old single `BundleFormat::Nu6_3` covered both), and reject pre-v6 restrictions.
- `Builder` now carries the selected pool restriction.
- Adapt to the new orchard pczt parse signatures (the `note_version` argument moved) and to the note-encryption `Domain`, which is now generic over a `DomainPolicy` (pass `OrchardDomain` / an explicit pool restriction).

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -28,8 +28,9 @@ pub(crate) enum NoteVersion {
 /// using an Orchard-equivalent flag format.
 ///
 /// PCZT v1 serializes Orchard flags with bits 2..=7 reserved and only supports
-/// V2 Orchard note plaintexts. The selected `BundlePoolRestrictions` value names
-/// that flag format; it is not an NU6.2 network-height assumption.
+/// Orchard note plaintexts (which always use lead byte 0x02). The selected
+/// `BundlePoolRestrictions` value names that flag format; it is not an NU6.2
+/// network-height assumption.
 #[cfg(feature = "orchard")]
 pub(crate) const ANY_ORCHARD_POOL_RESTRICTIONS: BundlePoolRestrictions =
     BundlePoolRestrictions::OrchardNu6_2Only;
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -9,7 +9,7 @@ use core::cmp::Ordering;
 use ff::PrimeField;
 use getset::Getters;
 #[cfg(feature = "orchard")]
-use orchard::bundle::BundleFormat;
+use orchard::bundle::BundlePoolRestrictions;
 #[cfg(feature = "orchard")]
 pub(crate) use orchard::note::NoteVersion;
 
@@ -635,7 +635,6 @@ impl Bundle {
                     action.spend.value,
                     action.spend.rho,
                     action.spend.rseed,
-                    note_version,
                     action.spend.fvk,
                     action.spend.witness,
                     action.spend.alpha,
@@ -650,6 +649,7 @@ impl Bundle {
                         })
                         .transpose()?,
                     action.spend.dummy_sk,
+                    note_version,
                     action.spend.proprietary,
                 )?;
 
@@ -662,7 +662,6 @@ impl Bundle {
                     action.output.recipient,
                     action.output.value,
                     action.output.rseed,
-                    note_version,
                     action.output.ock,
                     action
                         .output
@@ -675,6 +674,7 @@ impl Bundle {
                         })
                         .transpose()?,
                     action.output.user_address,
+                    note_version,
                     action.output.proprietary,
                 )?;
 
@@ -685,7 +685,7 @@ impl Bundle {
         orchard::pczt::Bundle::parse(
             actions,
             self.flags,
-            BundleFormat::PreNu6_3,
+            BundlePoolRestrictions::OrchardNu6_2Only,
             self.value_sum,
             self.anchor,
             self.zkproof,
@@ -798,7 +798,7 @@ impl Bundle {
             actions,
             flags: bundle
                 .flags()
-                .to_byte(BundleFormat::PreNu6_3)
+                .to_byte(BundlePoolRestrictions::OrchardNu6_2Only)
                 .expect("Orchard flags must be representable in the v5 transaction format"),
             value_sum,
             anchor: bundle.anchor().to_bytes(),
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -24,6 +24,16 @@ pub(crate) enum NoteVersion {
     V2,
 }
 
+/// Orchard pool restriction for PCZT operations whose result only depends on
+/// using an Orchard-equivalent flag format.
+///
+/// PCZT v1 serializes Orchard flags with bits 2..=7 reserved and only supports
+/// V2 Orchard note plaintexts. The selected `BundlePoolRestrictions` value names
+/// that flag format; it is not an NU6.2 network-height assumption.
+#[cfg(feature = "orchard")]
+pub(crate) const ANY_ORCHARD_POOL_RESTRICTIONS: BundlePoolRestrictions =
+    BundlePoolRestrictions::OrchardNu6_2Only;
+
 /// PCZT fields that are specific to producing the transaction's Orchard bundle (if any).
 #[derive(Clone, Debug, Getters)]
 pub struct Bundle {
@@ -685,7 +695,7 @@ impl Bundle {
         orchard::pczt::Bundle::parse(
             actions,
             self.flags,
-            BundlePoolRestrictions::OrchardNu6_2Only,
+            ANY_ORCHARD_POOL_RESTRICTIONS,
             self.value_sum,
             self.anchor,
             self.zkproof,
@@ -798,7 +808,7 @@ impl Bundle {
             actions,
             flags: bundle
                 .flags()
-                .to_byte(BundlePoolRestrictions::OrchardNu6_2Only)
+                .to_byte(ANY_ORCHARD_POOL_RESTRICTIONS)
                 .expect("Orchard flags must be representable in the v5 transaction format"),
             value_sum,
             anchor: bundle.anchor().to_bytes(),
--- a/pczt/src/roles/creator/mod.rs
+++ b/pczt/src/roles/creator/mod.rs
@@ -63,7 +63,7 @@ impl Creator {
     #[cfg(feature = "orchard")]
     pub fn with_orchard_flags(mut self, orchard_flags: orchard::bundle::Flags) -> Self {
         self.orchard_flags = orchard_flags
-            .to_byte(orchard::bundle::BundlePoolRestrictions::OrchardNu6_2Only)
+            .to_byte(crate::orchard::ANY_ORCHARD_POOL_RESTRICTIONS)
             .expect("Orchard flags must be representable in the v5 transaction format");
         self
     }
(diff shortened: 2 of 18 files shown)


pczt/src/roles/creator/mod.rs
```diff
@@ -63,7 +63,7 @@ impl Creator {
     #[cfg(feature = "orchard")]
     pub fn with_orchard_flags(mut self, orchard_flags: orchard::bundle::Flags) -> Self {
         self.orchard_flags = orchard_flags
-            .to_byte(orchard::bundle::BundleFormat::PreNu6_3)
+            .to_byte(orchard::bundle::BundlePoolRestrictions::OrchardNu6_2Only)
```
ValarDragon: This is Orchard flags, so every bundle pool restriction is actually equivalent here

Hence works

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -2291,8 +2291,17 @@ fn compact_orchard_action<R: RngCore + CryptoRng>(
...
+                ShieldedOutput::<::orchard::note_encryption::OrchardDomain, 52>::ephemeral_key(
+                    &compact_action,
+                )
+                .0
+                .to_vec(),
+            ciphertext:
+                ShieldedOutput::<::orchard::note_encryption::OrchardDomain, 52>::enc_ciphertext(
+                    &compact_action,
+                )[..52]
+                    .to_vec(),
```
ValarDragon, line 2304: I don't get why these changed

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1869,7 +1869,11 @@ where
                     .orchard_bundle()
                     .and_then(|bundle| {
                         bundle
-                            .decrypt_output_with_key(output_index, &orchard_internal_ivk)
+                            .decrypt_output_with_key(
+                                ::orchard::bundle::BundlePoolRestrictions::OrchardNu6_3Onward,
```
ValarDragon: All Orchard choices are equivalent here. I wonder if we should just add a constant for "OrchardFlagFormatResctrictions" where we set it to `orchard::bundle::BundlePoolRestrictions::OrchardNu6_3Onward`, since otherwise these are really confusing

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -50,14 +50,14 @@ impl MapAuth<Authorized, Authorized> for () {
 fn read_bundle<R: Read>(
```
ValarDragon, line 57: imo we should just delete the read_v5_bundle and read_v6_bundle fns

zcash_client_backend/src/data_api/wallet.rs
czarcas7ic: Ah this is actually a great recommendation, is much more obvious what is happening

https://github.com/zcash/librustzcash/pull/2480/commits/cfbc3fe2365e4d919131ab6dfea808415954c318

zcash_client_backend/src/data_api/testing.rs
czarcas7ic: The bump made `CompactAction` implement `ShieldedOutput` for more than one note domain, so the bare `ephemeral_key()`/`enc_ciphertext()` calls are now ambiguous and need to call `ShieldedOutput::<OrchardDomain, 52>` to pin the Orchard domain.

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -99,20 +99,47 @@ fn read_bundle<R: Read>(
...
+/// argument so the caller selects the Orchard or Ironwood v6 pool.)
 pub fn read_v5_bundle<R: Read>(
     reader: R,
     proof_size_enforcement: ProofSizeEnforcement,
 ) -> io::Result<Option<orchard::Bundle<Authorized, ZatBalance>>> {
-    read_bundle(reader, proof_size_enforcement, BundleFormat::PreNu6_3)
+    read_bundle(
+        reader,
+        proof_size_enforcement,
+        BundlePoolRestrictions::OrchardNu6_2Only,
```
nuttycom: This is more confusing now, given that this function will still be used for reading v5 transactions post-nu6.3. I'm not really content with the way that the pool restrictions are enumerated  in the `orchard` crate, which is intended to be independent and reusable (though its primary user is the Zcash protocol). For parsing in particular, this will need to be able to read both pre-nu6.3 and post-nu6.3 transaction data; we still have to be able to parse and validate the entire history of the chain.

pczt/tests/end_to_end.rs
```diff
@@ -589,7 +589,7 @@ fn orchard_to_orchard() {
     let value = orchard::value::NoteValue::from_raw(1_000_000);
     let note = {
         let mut orchard_builder = orchard::builder::Builder::new(
-            orchard::BundleProtocol::OrchardPreNu6_3,
+            orchard::bundle::BundlePoolRestrictions::OrchardNu6_2Only,
```
nuttycom, line 592:
```suggestion
            crate::ANY_ORCHARD_POOL_RESTRICTIONS,
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -870,7 +874,11 @@ where
             0
         };
         orchard_fees::transactional_action_count(
-            ::orchard::BundleProtocol::OrchardPreNu6_3,
+            // Preserve the legacy Orchard action-count estimate here. If this
+            // path targets restrictions that disable cross-address transfers,
+            // thread the target-height-selected `BundlePoolRestrictions`
+            // through this call.
```
nuttycom:
```suggestion
            // Preserve the legacy Orchard action-count estimate here. 
            // FIXME: When this path needs to respect restrictions that disable 
            // cross-address transfers, thread the target-height-selected 
            // `BundlePoolRestrictions` through this call.
```

We really need to avoid the foot-gun that's left in the code here. I'm okay with this being a FIXME but it's hazardous.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -710,7 +710,11 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                 ),
                 #[cfg(feature = "orchard")]
                 &(
-                    ::orchard::BundleProtocol::OrchardPreNu6_3,
+                    // Preserve the legacy Orchard action-count estimate here.
+                    // If this path targets restrictions that disable
```
nuttycom:
```suggestion
                    // FIXME: When this path targets restrictions that disable
```

zcash_primitives/src/transaction/tests.rs
```diff
@@ -412,6 +412,31 @@ fn v5_tx_data_with_sapling_bundle(
     )
 }
 
+/// Clears the cross-address flag on an Orchard bundle (preserving spends/outputs)
+/// so it is representable in a v6 Orchard slot (`OrchardNu6_3Onward`, which
+/// forbids cross-address transfers; cross-address is Ironwood-only).
+#[cfg(all(test, zcash_unstable = "nu6.3"))]
+fn disable_cross_address(
```
nuttycom: This is an exact duplicate of the code in https://github.com/zcash/librustzcash/pull/2480/changes#diff-c32a3c560d227172d7faab723c4f3512afc145d942e86dee08eb0435dfd7918fR431-R448

zcash_primitives/src/transaction/txid.rs
```diff
@@ -344,7 +376,9 @@ impl<A: Authorization> TransactionDigest<A> for TxIdDigester {
         orchard_bundle: Option<&orchard::Bundle<A::OrchardAuth, ZatBalance>>,
     ) -> Self::OrchardDigest {
         orchard_bundle.map(|b| {
-            b.commitment_for_domain(orchard_commitment_domain(version))
+            let (_, tx_version) = orchard_commitment_domain(version);
+            let flag_format = orchard_flag_format_for_flags(b.flags());
```
nuttycom, line 380: The commitment method here is pulling what's effectively a flag format from itself, and then passing it back to itself. That says to me that the pool restrictions shouldn't be passed into the commitment method at all.

nuttycom: The documentation for `commitment` says:
```
    /// `pool_restrictions` selects the flag-byte encoding; `tx_version` selects the commitment
    /// personalizations and whether the bundle anchor bytes are included here or in the
    /// authorizing digest. In a v5 transaction the anchor bytes are included here; in a v6
    /// transaction they are included by [`Bundle::authorizing_commitment`] instead.
```

So we pull the flag bytes from the bundle, figure out what restrictions they can correspond to using the extremely confusing method above, and then pass them back in to the commitment computation.

pczt/src/orchard.rs
```diff
@@ -24,6 +24,16 @@ pub(crate) enum NoteVersion {
     V2,
 }
 
+/// Orchard pool restriction for PCZT operations whose result only depends on
+/// using an Orchard-equivalent flag format.
+///
+/// PCZT v1 serializes Orchard flags with bits 2..=7 reserved and only supports
+/// V2 Orchard note plaintexts. The selected `BundlePoolRestrictions` value names
+/// that flag format; it is not an NU6.2 network-height assumption.
```
daira:
```suggestion
/// Orchard note plaintexts (which always use lead byte 0x02). The selected
/// `BundlePoolRestrictions` value names that flag format; it is not an NU6.2
/// network-height assumption.
```

pczt/src/orchard.rs
```diff
@@ -24,6 +24,16 @@ pub(crate) enum NoteVersion {
...
+#[cfg(feature = "orchard")]
+pub(crate) const ANY_ORCHARD_POOL_RESTRICTIONS: BundlePoolRestrictions =
+    BundlePoolRestrictions::OrchardNu6_2Only;
```
daira, line 36: The point of `BundlePoolRestrictions` is that it's supposed to be the information needed to interpret/produce the bundle format and to construct actions that meet the restrictions. This constant seems like it's trying to shoehorn a `BundlePoolRestrictions` value into contexts where that information is not actually known. It would be better to understand what those contexts are and have `orchard` provide an API that is explicitly designed for them.

(Note that using `BundleFormat::PreNu6_3` already had this problem. The PCZT v1 format is supposed to continue to be usable for v5 transactions after NU6.3, as far as I know.)

zcash_primitives/src/transaction/tests.rs
czarcas7ic: Re-used existing code here https://github.com/zcash/librustzcash/pull/2480/commits/ebae880371135f1684281c4153c4462fceccd877

nuttycom:
Subsumed by #2493

PR #2483, opened 2026-06-27
pczt: Make a v1 serialization type for Orchard PCZT bundle

ValarDragon:
## Summary

- Add a logical Orchard note plaintext version field to PCZT bundles.
- Split Orchard v1 serialization into crate-local DTOs that preserve the existing v1 wire shape.
- Default parsed v1 Orchard spends and outputs to `NoteVersion::V2`, and reject non-v1-compatible Orchard note versions during v1 encoding.
- Update the planned `pczt` changelog for the API and serialization behavior changes.

## Impact

The logical PCZT Orchard model can now carry the bundle `NoteVersion`, while the v1 encoding remains compatible with the existing V2-only field layout. Encoding a logical PCZT with an Orchard note version that cannot be represented by v1 now returns `EncodingError::UnsupportedOrchardNoteVersion`.

## Validation

- `cargo fmt --all`
- `cargo check -p pczt --features orchard`


pczt/src/orchard.rs
```diff
@@ -263,6 +266,246 @@ pub struct Output {
...
+            Self {
+                cv_net: action.cv_net,
+                spend: super::Spend::from(action.spend),
+                output: super::Output::from(action.output),
+                rcv: action.rcv,
+            }
+        }
+    }
+
+    impl TryFrom<super::Spend> for Spend {
```
nuttycom: This can just be `From`

pczt/src/orchard.rs
```diff
@@ -263,6 +266,246 @@ pub struct Output {
...
+                witness: spend.witness,
+                alpha: spend.alpha,
+                zip32_derivation: spend.zip32_derivation,
+                dummy_sk: spend.dummy_sk,
+                proprietary: spend.proprietary,
+            }
+        }
+    }
+
+    impl TryFrom<super::Output> for Output {
```
nuttycom: this can just be From

pczt/src/orchard.rs
```diff
@@ -263,6 +266,246 @@ pub struct Output {
...
+                value_sum: bundle.value_sum,
+                anchor: bundle.anchor,
+                note_version: NoteVersion::V2,
+                zkproof: bundle.zkproof,
+                bsk: bundle.bsk,
+            }
+        }
+    }
+
+    impl TryFrom<super::Action> for Action {
```
nuttycom: this can just be From

nuttycom (approved):
utACK b883c3df8309b7183eaa24a7a9bda235c4683e61

PR #2484, opened 2026-06-27
zcash_primitives: Charge ZIP 317 marginal fees for Ironwood actions

### LINKED ISSUE #2461: zcash_primitives: Update ZIP 317 support to handle Ironwood actions.
---

nuttycom:
## Summary

Updates the ZIP 317 fee computation in `zcash_primitives` to account for Ironwood actions, mirroring the existing Orchard handling. Ironwood is the shielded pool carried by the NU6.3 V6 transaction format.

- Adds an `ironwood_action_count: usize` argument to `FeeRule::fee_required`, folded into the ZIP 317 logical action count so that each Ironwood action is charged one marginal fee — the same treatment as an Orchard action.
- As with Orchard, the minimum-of-two-actions floor for a non-empty bundle is applied at the action-counting layer (the bundle type's `num_actions`), not in the fee rule itself; the fee rule simply adds the already-padded count.
- The transaction builder and `zcash_client_backend` callers thread through a count of zero, since neither constructs Ironwood bundles yet.

The fee-rule change is unconditional (fee computation is not feature-gated); the surrounding Ironwood/V6 support lives behind the `zcash_unstable = "nu6.3"` cfg.

## Follow-up

When Ironwood bundle construction is added to the builder, it will need an `ironwood_action_count` helper (paralleling the existing `orchard_action_count`) that applies the same `MIN_ACTIONS` padding, so that non-empty Ironwood bundles count as at least two actions.

## Testing

- New unit tests in `zip317.rs` covering Ironwood actions being charged the marginal fee, and combining additively with Orchard actions.
- `cargo check --workspace --all-features --tests` passes under `--cfg zcash_unstable="nu6.3"`.

Closes #2461

### DIFF
--- a/zcash_primitives/CHANGELOG.md
+++ b/zcash_primitives/CHANGELOG.md
@@ -42,6 +42,12 @@ workspace.
   `zcash_primitives::transaction::components::orchard::{read_v5_bundle,
   read_v6_bundle, read_flags}` now take a `BundleVersion` argument, while
   `write_v6_bundle` derives the version from the bundle.
+- `zcash_primitives::transaction::fees::FeeRule::fee_required` now takes an
+  additional `ironwood_action_count: usize` argument following
+  `orchard_action_count`. Implementors and callers must thread through the
+  number of Ironwood actions; pass `0` for transactions without an Ironwood
+  bundle. Under ZIP 317 each Ironwood action is charged one marginal fee, the
+  same as an Orchard action.
 
 ### Removed
 - All support for Transparent Zcash Extensions (TZEs), which was only ever
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -89,6 +89,11 @@ workspace.
   addition to `propose_shielding`.
 - `zcash_client_backend::wallet::WalletTx::new` now takes a `transparent_outputs`
   argument.
+- `zcash_client_backend::fees::StandardFeeRule` tracks the new
+  `ironwood_action_count: usize` argument added to
+  `zcash_primitives::transaction::fees::FeeRule::fee_required`. Code that calls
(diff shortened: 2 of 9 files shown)


zcash_primitives/CHANGELOG.md
```diff
@@ -33,6 +33,12 @@ workspace.
   implementations can distinguish Orchard commitments by transaction format.
 - `TransactionDigest::digest_sapling` now receives `TxVersion`, so digest
   implementations can distinguish Sapling commitments by transaction format.
+- `zcash_primitives::transaction::fees::FeeRule::fee_required` now takes an
+  additional `ironwood_action_count: usize` argument following
+  `orchard_action_count`. Implementors and callers must thread through the
+  number of Ironwood actions; pass `0` for transactions without an Ironwood
+  bundle. Under ZIP 317 each Ironwood action is charged one marginal fee, the
```
czarcas7ic, line 49: nit: I don't think ZIP 317 mentions Ironwood by name. So to be accurate, I would say 

```
Under the ZIP 317 action model, each Ironwood action is charged one marginal fee, mirroring the existing Orchard treatment.
```

ValarDragon (dismissed):

ValarDragon:
Tracking as a comment here, we are likely merging #2485 first, which means we need to update the fee call-site in the code for ironwood. (In #2485 we make orchard_actions argument be set to orchard_plus_ironwood_actions)

zcash_primitives/CHANGELOG.md
nuttycom: This is in https://github.com/zcash/zips/pull/1314

nuttycom (commented):
self-utACK; this'll need additional changes in zcash_client_backend as part of adding Ironwood wallet support.

ValarDragon (approved):

PR #2485, opened 2026-06-27
zcash_primitives: Add Ironwood wiring to the transaction builder

### LINKED ISSUE #2453: zcash_primitives: Add Ironwood support to the transaction builder
---

czarcas7ic:
Adds builder support for constructing Ironwood bundles in v6 transactions, in parallel to the existing Orchard path.

Stacked on `adam/bump-orchard-30c4ea2` (the Ironwood builder uses `BundlePoolRestrictions::IronwoodNu6_3Onward` and the pool-parameterized orchard builder from there); retarget to `feat/ironwood` once that merges.

- Five Ironwood `Error` variants plus `IronwoodBuilderNotAvailable`.
- `BuildConfig::ironwood_anchor` and `ironwood_builder()` (pins `IronwoodNu6_3Onward`).
- `Builder`: `add_ironwood_spend` (enforces `NoteVersion::V3`), `add_ironwood_output`, `add_ironwood_change_output`, and the `build()`/`build_for_pczt()` wiring (construct, prove, sign, fee/action count, metadata). The Ironwood builder is created exactly when the branch's default transaction version is V6 (NU6.3 onward).

The `ironwood_anchor` field is cfg-gated on `nu6.3`/`nu7`, so the existing `BuildConfig::Standard` constructors in `pczt`, a `zcash_client_sqlite` migration, and the `zcash_client_backend` build path gain `ironwood_anchor: None` (wallet-side Ironwood building is a later split).

### DIFF
--- a/zcash_client_backend/src/data_api/wallet.rs
+++ b/zcash_client_backend/src/data_api/wallet.rs
@@ -1330,6 +1330,8 @@ where
         BuildConfig::Standard {
             sapling_anchor,
             orchard_anchor,
+            #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+            ironwood_anchor: None,
         },
     );
 
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -1260,12 +1260,32 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
             .transpose()
             .map_err(Error::SaplingBuild)?;
 
+        // Orchard and Ironwood actions share the post-NU6.3 circuit, so a single
+        // proving key serves both proofs when a V6 transaction carries both
+        // bundles. Build it once, from whichever bundle is present; their circuit
+        // versions agree whenever both are present.
+        let orchard_proving_key = {
+            let circuit_version = unauthed_tx
+                .orchard_bundle
+                .as_ref()
+                .map(|b| b.circuit_version());
+            #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+            let circuit_version = circuit_version.or_else(|| {
+                unauthed_tx
+                    .ironwood_bundle
+                    .as_ref()
+                    .map(|b| b.circuit_version())
+            });
+            circuit_version.map(orchard::circuit::ProvingKey::build)
+        };
+
         let orchard_bundle = unauthed_tx
             .orchard_bundle
             .map(|b| {
-                let circuit_version = b.circuit_version();
                 b.create_proof(
-                    &orchard::circuit::ProvingKey::build(circuit_version),
+                    orchard_proving_key
+                        .as_ref()
+                        .expect("proving key is built when an Orchard bundle is present"),
                     &mut rng,
                 )
                 .and_then(|b| {
@@ -1279,9 +1299,10 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
         let ironwood_bundle = unauthed_tx
             .ironwood_bundle
             .map(|b| {
-                let circuit_version = b.circuit_version();
                 b.create_proof(
-                    &orchard::circuit::ProvingKey::build(circuit_version),
+                    orchard_proving_key
+                        .as_ref()
+                        .expect("proving key is built when an Ironwood bundle is present"),
                     &mut rng,
                 )
                 .and_then(|b| {
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -87,16 +87,32 @@ pub enum Error<FE> {
     SaplingBuild(sapling::builder::Error),
     /// An error occurred in constructing the Orchard parts of a transaction.
     OrchardBuild(orchard::builder::BuildError),
+    /// An error occurred in constructing the Ironwood parts of a transaction.
+    #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+    IronwoodBuild(orchard::builder::BuildError),
     /// An error occurred in adding an Orchard Spend to a transaction.
     OrchardSpend(orchard::builder::SpendError),
     /// An error occurred in adding an Orchard Output to a transaction.
     OrchardRecipient(orchard::builder::OutputError),
+    /// An error occurred in adding an Ironwood Spend to a transaction.
+    #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+    IronwoodSpend(orchard::builder::SpendError),
+    /// An Ironwood spend note used an unsupported note plaintext version.
+    #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+    IronwoodSpendUnsupportedNoteVersion(orchard::NoteVersion),
+    /// An error occurred in adding an Ironwood Output to a transaction.
+    #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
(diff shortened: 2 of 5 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -608,6 +724,85 @@ impl<P: consensus::Parameters, U> Builder<P, U> {
...
+                memo.into_bytes(),
+            )
+            .map_err(Error::IronwoodRecipient)
+    }
+
+    /// Adds a wallet-controlled Ironwood change output to the transaction.
+    ///
+    /// This uses [`orchard::note::NoteVersion::V3`].
+    #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+    pub fn add_ironwood_change_output<FE>(
```
czarcas7ic: Need to implement the orchard change output as well in a later PR so we allow for orchard change as well.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1330,6 +1330,8 @@ where
         BuildConfig::Standard {
             sapling_anchor,
             orchard_anchor,
+            #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+            ironwood_anchor: None,
```
ValarDragon, line 1334: ok this is a TODO for us to solve in a subsequent commit

zcash_primitives/src/transaction/builder.rs
```diff
@@ -752,17 +989,7 @@ impl<P: consensus::Parameters, U> Builder<P, U> {
...
-                    .as_ref()
-                    .map_or(Ok(0), |builder| {
-                        orchard_action_count(
-                            builder,
-                            self.build_config.is_coinbase(),
-                            self.orchard_pool_restrictions
-                                .expect("orchard builder present implies pool restrictions"),
-                        )
-                    })
-                    .map_err(FeeError::Bundle)?,
```
ValarDragon, line 780: Not sure we had to move this

ValarDragon: Ah this was moved to make it orchard+ironwood actions, because we don't have #2484 in

ValarDragon: Ok just need a note to self if we merge this first, to update in #2484 , but can we rename the variable to orchard_plus_ironwood_actions in the interrim?

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1015,6 +1257,22 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
...
 
+        #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+        let ironwood_bundle = unauthed_tx
+            .ironwood_bundle
+            .map(|b| {
+                let circuit_version = b.circuit_version();
+                b.create_proof(
+                    &orchard::circuit::ProvingKey::build(circuit_version),
+                    &mut rng,
+                )
```
ValarDragon: IF we want, we can just define this once for both proofs

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1015,6 +1257,22 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
...
+                .and_then(|b| {
+                    b.apply_signatures(&mut rng, *shielded_sig_commitment.as_ref(), orchard_saks)
```
ValarDragon, line 1311: This is `orchard_saks` because the signing key is the same between orchard/ironwood.

zcash_primitives/src/transaction/builder.rs
ValarDragon: In that PR we also need to update the Orchard transaction counts for cross-address disabled

ValarDragon (dismissed):
LGTM for Ironwood matching today's impl.

Two follow-ups noted, one for:
- Orchard cross address disabled updates. (Change notes, transaction counts)
- Second for fees post #2484

zcash_primitives/src/transaction/builder.rs
czarcas7ic: https://github.com/zcash/librustzcash/pull/2485/commits/7feae52fd342dfda6daa0a0a3352fa26f482fe3d

zcash_client_backend/src/data_api/wallet.rs
czarcas7ic: Opened a PR to thread this through here https://github.com/zcash/librustzcash/pull/2498

zcash_primitives/src/transaction/builder.rs
ValarDragon: (line ref moved, I meant proving key)

zcash_primitives/src/transaction/builder.rs
ValarDragon: Resolved

zcash_primitives/src/transaction/builder.rs
nuttycom: Why do we need `add_ironwood_change_output` at all at this point? I don't think we need to support someone disabling cross-address spends in ironwood at this layer.

czarcas7ic: Agreed, its not needed

https://github.com/zcash/librustzcash/pull/2485/commits/d8e15b5f044e3fb68c82ef600b7918baa978471d

zcash_primitives/src/transaction/builder.rs
czarcas7ic: Done here https://github.com/zcash/librustzcash/pull/2485/changes/300c5072ae7d35288928dd4a97c9dbaefead0f7d

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1018,12 +1260,32 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
...
+                .map(|b| b.circuit_version());
+            #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+            let circuit_version = circuit_version.or_else(|| {
+                unauthed_tx
+                    .ironwood_bundle
+                    .as_ref()
+                    .map(|b| b.circuit_version())
+            });
+            circuit_version.map(orchard::circuit::ProvingKey::build)
+        };
```
ValarDragon, line 1277: Can we just do "if orchard_bundle.is_some || ironwood_bundle.is_some { get circuit version from consensus branch ID"?

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1018,12 +1260,32 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
             .transpose()
             .map_err(Error::SaplingBuild)?;
 
+        // Orchard and Ironwood actions share the post-NU6.3 circuit, so a single
+        // proving key serves both proofs when a V6 transaction carries both
+        // bundles. Build it once, from whichever bundle is present; their circuit
+        // versions agree whenever both are present.
+        let orchard_proving_key = {
```
nuttycom, line 1267: nit: we don't even need to build this unless there's either an Orchard or Ironwood bundle.

nuttycom: Oh, I see, that's actually already an Option, nm

zcash_primitives/src/transaction/builder.rs
czarcas7ic: https://github.com/zcash/librustzcash/pull/2485/changes/38eefdcdb094a7203dc6948a832627ea44408cf8

nuttycom (approved):
utACK 38eefdcdb094a7203dc6948a832627ea44408cf8

Issue #2470, opened 2026-06-26
zcash_client_sqlite: Transparent wallet performance at scale (hundreds of thousands of addresses/UTXOs)


nuttycom:
## Context

Wallets that hold very large numbers of transparent addresses and UTXOs — on the order of hundreds of thousands (an exchange/custodial address-per-deposit pattern, or a `zcashd` wallet import; the default gap limits are external=10/internal=5/ephemeral=10, so reaching this scale implies one of those) — exercise the transparent code paths in `zcash_client_sqlite` (and the shielding input-selection paths in `zcash_client_backend`) far outside their original design point. At this scale, balance computation and the gather phase that mobilizes transparent funds become severe bottlenecks.

Note on architecture: the *pure* spend path (`propose_transfer`/`spend`) does **not** load transparent UTXOs — it funds payments from shielded notes only. Transparent funds are mobilized exclusively by **shielding** (`shield_transparent_funds` → `propose_shielding`), so "spending transparent funds" *is* the shielding path, and that is where the cost concentrates.

## Three structural problems

1. **Hot-path columns are unindexed.** `addresses.cached_transparent_receiver_address` and `transparent_received_outputs.address` (TEXT) are filtered by the address→account lookups and by `get_spendable_transparent_outputs`, but neither is indexed (only the integer `address_id` FK is). Every such lookup is a full scan of a multi-hundred-thousand-row table.
2. **The shielding gather is an uncapped N+1 over addresses.** `gather_shielding_inputs` calls `get_spendable_transparent_outputs` once per source address; combined with problem (1) each call is a full table scan, and the gather materializes the entire spendable set with no amount target or per-transaction input cap. A second per-input metadata lookup happens at transaction-build time.
3. **Balance/summary and gap-limit bookkeeping re-scan unconditionally.** `add_transparent_account_balances` (on every `get_wallet_summary`) full-scans the outputs table with a repeated `NOT IN (all spent outputs)` subquery; `find_gap_start` re-runs a window-function aggregation over a 3-way-UNION view redundantly, including once per UTXO on ingest.

## Measured evidence

Synthetic SQLite database built from the real schema and indexes, loaded with 500,000 addresses + 500,000 UTXOs (+50k spends), measured with `EXPLAIN QUERY PLAN` and `.timer` (warm cache, desktop — absolute numbers understate cold-cache/mobile, but the ratios are the robust signal):

| Operation | Before | After fix | Speedup |
|---|---|---|---|
[...]

PR #2486, opened 2026-06-28
zcash_client_sqlite: Add UNIQUE index on addresses.cached_transparent_receiver_address

### LINKED ISSUE #2311: zcash_client_sqlite: add UNIQUE constraint on `addresses.cached_transparent_receiver_address`
## Context

The `addresses` table in `zcash_client_sqlite` stores `cached_transparent_receiver_address` as a denormalised cache of the transparent receiver for each row that represents (or contains) a transparent address. The most recent schema revision (`standalone_p2sh.rs`) applies `UNIQUE` constraints to:

- `(account_id, key_scope, diversifier_index_be)`
- `imported_transparent_receiver_pubkey`
- `imported_transparent_receiver_script`

but **not** to `cached_transparent_receiver_address` itself.

## Problem

Code that looks up an account by transparent address — for example, the new SQLite override of `WalletRead::find_account_for_address` added for `find_account_for_unified_address` on the `wallet-1944-find-account-for-address` branch — has to either use `LIMIT 1` over a potentially non-deterministic ordering, or deduplicate results in Rust. Either way, the invariant that a given transparent address maps to at most one account is not enforced at the schema level, so callers cannot rely on it.

In practice a given transparent address should correspond to exactly one `(account, key_scope)` derivation (or exactly one standalone import); duplicate cached values would indicate a derivation collision or data-corruption bug that we would want the database to refuse rather than silently accept.

## Proposed fix

The next time the `addresses` table schema is rewritten (the table is already rewritten in several migrations: `full_account_ids.rs`, `account_delete_cascade.rs`, `transparent_gap_limit_handling.rs`, `support_zcashd_wallet_import.rs`, `standalone_p2sh.rs`), add:

```sql
UNIQUE (cached_transparent_receiver_address)
```

to the new `addresses` table definition, and verify that existing data satisfies the constraint as part of the migration.

## Notes

- `NULL` values do not violate SQLite's `UNIQUE` semantics, so rows without a transparent receiver are unaffected.
- The existing `ck_addr_transparent_index_consistency` CHECK already ensures that `cached_transparent_receiver_address` is populated iff the row represents an address with a transparent receiver, so the uniqueness constraint would cleanly cover "one account owns each transparent receiver."

nuttycom:
Part of the transparent-wallet-at-scale performance target #2470.

Account-by-transparent-address lookups (`find_account_for_transparent_address`, `get_transparent_address_metadata`, `put_transparent_output`, and related maintenance queries) filter on `addresses.cached_transparent_receiver_address`, which had no supporting index and so required a full scan of the `addresses` table. On a synthetic 500k-address database the lookup drops from a ~26 ms full scan to a ~0.03 ms index seek.

The index is declared `UNIQUE`, which additionally enforces the invariant that each transparent receiver belongs to at most one address record. The migration that adds it pre-checks existing data and fails with a clear `CorruptedData` error (rather than an opaque index-creation failure) if duplicates are present. `NULL` receivers are unaffected, as multiple `NULL`s do not violate SQLite's `UNIQUE` semantics.

Tests: a migration test, a uniqueness-enforcement test, and a test that a pre-existing duplicate produces a clean migration failure. The full `zcash_client_sqlite` suite passes under default and `transparent-inputs,orchard` features.

Closes #2311.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations.rs
@@ -12,6 +12,7 @@ mod add_account_birthdays;
 mod add_account_uuids;
 mod add_transaction_trust_marker;
 mod add_transaction_views;
+mod add_transparent_receiver_address_index;
 mod add_utxo_account;
 mod addresses_table;
 mod ensure_default_transparent_address;
@@ -136,8 +137,8 @@ pub(super) fn all_migrations<
     //                       `------------------- account_delete_cascade ---------------------------------'
     //                                        /               |              \
     //                       v_tx_outputs_key_scopes    standalone_p2sh    witness_stabilized_notes
-    //                                                        |
-    //                                                  ivk_item_cache
+    //                                                    /          \
+    //                                          ivk_item_cache    add_transparent_receiver_address_index
     //
     let rng = Rc::new(Mutex::new(rng));
     vec![
@@ -231,6 +232,7 @@ pub(super) fn all_migrations<
         Box::new(witness_stabilized_notes::Migration {
             params: params.clone(),
         }),
+        Box::new(add_transparent_receiver_address_index::Migration),
     ]
 }
 
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
@@ -1,0 +1,394 @@
+//! Adds a `UNIQUE` index on `addresses.cached_transparent_receiver_address`.
+//!
+//! The `addresses` table stores `cached_transparent_receiver_address` as a denormalized cache of
+//! the transparent receiver for each row that represents (or contains) a transparent address.
+//! Lookups by this value — for example, resolving the account that controls a transparent address
+//! during scanning and transaction construction — previously had no supporting index and so
+//! required a full table scan, which becomes a severe bottleneck for wallets that hold very large
+//! numbers of transparent addresses.
+//!
+//! In addition to making these lookups index-backed, the `UNIQUE` constraint enforces the
+//! invariant that a given transparent receiver is associated with at most one address record.
+//! Any pre-existing duplicates (for example, the same address both HD-derived and standalone
+//! imported) are resolved by the migration: a single canonical record is retained, the
+//! received-output foreign keys of the others are repointed to it, and the redundant records are
+//! deleted. A receiver duplicated across more than one account cannot be safely merged and aborts
+//! the migration. `NULL` values do not participate in SQLite's `UNIQUE` semantics, so rows without
+//! a transparent receiver are unaffected.
+
+use std::collections::HashSet;
+
+use rusqlite::named_params;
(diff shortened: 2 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -0,0 +1,243 @@
...
+                |row| row.get(0),
+            )
+            .optional()?;
+
+        if let Some(addr) = duplicate {
+            return Err(WalletMigrationError::CorruptedData(format!(
+                "Cannot add a uniqueness constraint on \
+                 addresses.cached_transparent_receiver_address: the transparent address {addr} is \
+                 associated with more than one address record."
+            )));
```
nuttycom: If we return an error here the user's wallet will be disabled unless we find a way to resolve the issue; we essentially need to resolve such errors by choosing which record should persist, ensuring all foreign keys reference that record, and then removing the offending record(s).

zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -0,0 +1,243 @@
...
+use super::{ivk_item_cache, v_tx_outputs_key_scopes, witness_stabilized_notes};
+use crate::wallet::init::WalletMigrationError;
+
+pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x3d4f12d6_3da9_4ace_ac65_a0dd0a7adc32);
+
+const DEPENDENCIES: &[Uuid] = &[
+    ivk_item_cache::MIGRATION_ID,
+    v_tx_outputs_key_scopes::MIGRATION_ID,
+    witness_stabilized_notes::MIGRATION_ID,
+];
```
nuttycom: Why do we depend on all three of these migrations?

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -136,8 +137,10 @@ pub(super) fn all_migrations<
     //                       `------------------- account_delete_cascade ---------------------------------'
     //                                        /               |              \
     //                       v_tx_outputs_key_scopes    standalone_p2sh    witness_stabilized_notes
-    //                                                        |
-    //                                                  ivk_item_cache
+    //                                    \                   |                  /
+    //                                     \           ivk_item_cache          /
+    //                                      \               |                 /
+    //                                       `-- add_transparent_receiver_address_index --'
```
nuttycom: We probably don't need to depend on all these migrations; really, we only need to depend on the topologically-latest migration that includes the addresses table.

nuttycom (commented):
self-utACK bfb050edeab769b2ca0f2b0c29c0c1ab01adc190

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/bfb050edeab769b2ca0f2b0c29c0c1ab01adc190..ef6214bc5e3323833b7e9b239cb80ee16dcdeefd) to fix Clippy complaints.

schell (approved):

PR #2487, opened 2026-06-28
zcash_client_sqlite: Make get_spendable_transparent_outputs use an indexed address lookup

### LINKED ISSUE #2471: zcash_client_sqlite: get_spendable_transparent_outputs filters on the unindexed transparent_received_outputs.address TEXT column
Part of the transparent-at-scale performance target #2470.

`get_spendable_transparent_outputs` (`zcash_client_sqlite/src/wallet/transparent.rs:1246`) drives its query with `WHERE u.address = :address`, where `transparent_received_outputs.address` is an **unindexed TEXT column**. The only index on that table relevant to addresses is on the integer `address_id` FK, not the `address` string. The denormalized `address` column was added specifically to make per-address UTXO lookup fast (`db.rs:526-529`), but no supporting index exists, so each call is a full table scan.

`EXPLAIN QUERY PLAN` on a 500k-UTXO synthetic DB: `SCAN u`, ~46 ms per single-address call (warm cache). With an index the plan becomes `SEARCH u USING INDEX … (address=?)` at ~0.035 ms — roughly 1300×. This is the per-call half of the shielding-gather bottleneck (the per-address loop is tracked separately).

### Suggested fix (either, prefer the first)
- **(a)** Resolve the address to its `address_id` once (via the `cached_transparent_receiver_address` index from #2311) and filter on `u.address_id = :address_id`, which uses the existing `idx_transparent_received_outputs_address`. No new index, and `address`/`address_id` are redundant anyway.
- **(b)** Add `CREATE INDEX … ON transparent_received_outputs(address)`.

Also switch the statement from `conn.prepare` to `conn.prepare_cached` (`transparent.rs:1259`) — it is currently recompiled on every iteration of the per-address gather loop.

---

nuttycom:
Part of the transparent-wallet-at-scale performance target #2470.

`get_spendable_transparent_outputs` filtered candidate UTXOs with `WHERE transparent_received_outputs.address = :address`, an unindexed TEXT column, forcing a full scan of the `transparent_received_outputs` table on every call. Because shielding invokes this once per source address, a wallet with many transparent addresses paid one full table scan per address.

This matches the target against the joined `addresses.cached_transparent_receiver_address` instead, so the planner locates the address row and reaches its outputs via the existing `idx_transparent_received_outputs_address` index on `address_id`, turning the per-call full scan into an index seek. The denormalized `address` column on `transparent_received_outputs` is redundant with the joined address, so the result set is unchanged. The lookup becomes a pure index seek once the companion index from #2311 is present, and is non-regressive without it. The statement is also now `prepare_cached`, since shielding re-runs it once per source address.

Independent of #2311 and #2471's sibling PRs (a pure query change, no migration). The existing shielding/transparent test suite passes.

Closes #2471.

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -1256,7 +1256,13 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
         TransparentOutputFilter::CoinbaseOnly => 1i32,
     };
 
-    let mut stmt_utxos = conn.prepare(&format!(
+    // This statement is re-run once per source address when shielding, so it is prepared via
+    // `prepare_cached` to avoid recompiling it on each call. The address is matched against
+    // `addresses.cached_transparent_receiver_address` (rather than the denormalized
+    // `transparent_received_outputs.address`) so that the lookup is served by an index on the
+    // `addresses` table and joined to the outputs via `idx_transparent_received_outputs_address`,
+    // instead of scanning the full `transparent_received_outputs` table.
+    let mut stmt_utxos = conn.prepare_cached(&format!(
         "SELECT t.txid, u.output_index, u.script,
                 u.value_zat, addresses.key_scope,
                 accounts.uuid AS account_uuid,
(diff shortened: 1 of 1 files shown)


nuttycom (commented):
sefl-utACK; this is a trivial change but a meaningful speedup.

daira (approved):
utACK

Issue #2472, opened 2026-06-26
zcash_client_backend: shielding input gather is an uncapped per-address N+1


nuttycom:
Part of the transparent-at-scale performance target #2470.

`gather_shielding_inputs` (`zcash_client_backend/src/data_api/wallet/input_selection.rs:1359`) does `source_addrs.iter().try_fold(...)`, calling `get_spendable_transparent_outputs` **once per source address** — one DB round-trip per address, issued even for addresses with zero spendable UTXOs (the common case under gap-limit scanning), and each call is currently a full table scan (see the index issue blocking #2470). The results are `extend`ed into a single `Vec` plus a `BTreeSet<TransparentAddress>`, with **no amount target and no per-transaction input cap**. The only cap (`limit` on the coinbase path, `input_selection.rs:1234`) is applied *after* full materialization and *after* an O(n log n) sort.

At hundreds of thousands of source addresses this is the dominant cost: ~500k full scans ≈ several hours of query time in the worst case, versus a single account-scoped scan (~0.16 s measured) that returns all unspent outputs grouped by address and skips empty addresses for free.

### Suggested direction
- Replace the per-address loop with a single account-scoped discovery query returning unspent outputs grouped by address; iterate the grouping in memory. Preserve the per-transaction single-address privacy property ("when shielding we always select UTXOs for only a single address at a time", `db.rs:526`) at *construction* time — discovery can be batched while each built shielding transaction still draws from a single address.
- Cap/target the gather during gathering (a single transaction cannot hold anywhere near this many inputs).

Relates to #1201 (proposing multiple transactions when shielding from multiple receivers).

---
*Filed with Claude Code assistance.*

PR #2488, opened 2026-06-28
zcash_client_backend: Batch transparent input gathering when shielding

nuttycom:
Part of the transparent-wallet-at-scale performance target #2470. This is the first of a stacked pair for #2472; a follow-up PR adds the per-transaction input cap.

`gather_shielding_inputs` called `InputSource::get_spendable_transparent_outputs` once per source address, issuing one database query per address — including for the many addresses that have no spendable outputs. For wallets that hold large numbers of transparent addresses (e.g. exchange/custodial wallets, or wallets recovered from a `zcashd` import), shielding therefore performed up to hundreds of thousands of queries.

This adds `InputSource::get_spendable_transparent_outputs_for_addresses`, a batched equivalent that returns the spendable outputs for a set of addresses. It has a default implementation that queries each address individually, so existing implementors are unaffected; the `zcash_client_sqlite` implementation overrides it with a single query that matches the receiving address against the set via `rarray`, ordered deterministically. `gather_shielding_inputs` now issues one query and applies the single-ephemeral linkability guard once over the full result, so the set of selected outputs (and hence the resulting proposal) is unchanged.

Includes a multi-address regression test (the prior shielding tests all used a single source address).

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -1328,6 +1328,107 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
     Ok(utxos)
 }
 
+/// Returns the list of spendable transparent outputs received by this wallet at any of the
+/// given `addresses`, under the same spendability conditions as
+/// [`get_spendable_transparent_outputs`].
+///
+/// This is the batched equivalent of [`get_spendable_transparent_outputs`]: it issues a single
+/// query over the entire set of provided addresses rather than one query per address, which avoids
+/// a per-address database round-trip (and, for each empty address, a wasted query) when shielding
+/// from a wallet that holds large numbers of transparent addresses. Each returned output
+/// identifies its receiving address, so a caller that needs to group results by address can do so
+/// from the returned values.
+///
+/// The query body mirrors that of [`get_spendable_transparent_outputs`], differing only in that
+/// the receiving address is matched against a set via `rarray` rather than a single value.
+pub(crate) fn get_spendable_transparent_outputs_for_addresses<P: consensus::Parameters>(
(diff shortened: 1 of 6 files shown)


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1322,6 +1322,107 @@ pub(crate) fn get_spendable_transparent_outputs<P: consensus::Parameters>(
...
+         AND u.id NOT IN ({}) -- and the output is unspent
+         AND ({}) -- exclude likely-spent wallet-internal ephemeral outputs
+         AND ({}) -- exclude immature coinbase outputs
+         AND (:coinbase_only == 0 OR IFNULL(t.tx_index, 1) == 0) -- coinbase filter: unknown tx_index defaults to 1 (non-coinbase) to avoid false positives
+         ORDER BY addresses.cached_transparent_receiver_address, u.output_index",
+        tx_unexpired_condition_minconf_0("t"),
+        spent_utxos_clause(),
+        excluding_wallet_internal_ephemeral_outputs("u", "addresses", "t", "accounts"),
+        excluding_immature_coinbase_outputs("t"),
+    ))?;
```
schell, line 1323: So this adds another big query to this file that will have to be kept in sync with the other big query. I think we should rewrite the above function `get_spendable_transparent_outputs_for_addresses` in terms of this one by calling it with a singleton address list. It would be a slightly bigger diff, but less surface area for bugs, which SQL tends to be rife with in my experience.

schell: Refactoring in this way would also make my changes in #2422 easier, as they would only apply to a single query.

schell (dismissed):
utACK 5041459007148186352a7e8d72b8175ed2457d47 👍

schell (approved):

PR #2489, opened 2026-06-28
zcash_client_backend: Cap shielding transparent inputs to a fraction of block space

### LINKED ISSUE #2472: zcash_client_backend: shielding input gather is an uncapped per-address N+1

nuttycom:
Part of the transparent-wallet-at-scale performance target #2470. Stacked on #2488 (the batching half of #2472); the base will become `main` once #2488 merges.

Shielding gathered every spendable transparent UTXO into a single proposed transaction, which for wallets with very large UTXO sets produces a transaction that cannot be built or relayed. This caps the transparent inputs of a single shielding transaction to a configurable fraction of a block's space (default 10%), selecting the highest-value inputs first; the remainder is left unspent, to be consolidated by a subsequent (manually-initiated) shielding transaction.

The fraction is configured via `GreedyInputSelector::with_shielding_block_space_percent`. The cap (`(2_000_000 × percent / 100) / 150` inputs) is applied in `gather_shielding_inputs`, so both the standard and coinbase shielding paths share it, and the coinbase path's existing `limit` now composes with it (the more restrictive of the two applies).

Tests: a unit test of the cap math and an integration test asserting that, with a 1%-of-block-space cap and 140 distinct-value UTXOs, exactly the 133 highest-value UTXOs are selected.

Closes #2472.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -414,12 +414,25 @@ impl orchard_fees::OutputView for OrchardPayment {
     }
 }
 
+/// The Zcash consensus maximum block size, in bytes.
+#[cfg(feature = "transparent-inputs")]
+const MAX_BLOCK_BYTES: usize = 2_000_000;
+
+/// The default maximum fraction of a block's space, as an integer percentage, that a single
+/// shielding transaction's transparent inputs may occupy.
+#[cfg(feature = "transparent-inputs")]
+const DEFAULT_SHIELDING_BLOCK_SPACE_PERCENT: u32 = 10;
+
 /// An [`InputSelector`] implementation that uses a greedy strategy to select between available
 /// notes.
 ///
 /// This implementation performs input selection using methods available via the
 /// [`InputSource`] interface.
 pub struct GreedyInputSelector<DbT> {
+    /// The maximum fraction of a block's space, as an integer percentage (0–100), that a single
+    /// shielding transaction's transparent inputs may occupy.
+    #[cfg(feature = "transparent-inputs")]
+    shielding_block_space_percent: u32,
     _ds_type: PhantomData<DbT>,
 }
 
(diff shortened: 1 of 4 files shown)


zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -435,9 +448,32 @@ impl<DbT> GreedyInputSelector<DbT> {
...
+        self
+    }
+}
+
+/// Returns the maximum number of transparent inputs that a single shielding transaction may
+/// select, given the configured fraction of a block's space (as an integer percentage) that its
+/// inputs may occupy.
+#[cfg(feature = "transparent-inputs")]
+fn shielding_max_inputs(block_space_percent: u32) -> usize {
+    (MAX_BLOCK_BYTES.saturating_mul(block_space_percent as usize) / 100) / P2PKH_STANDARD_INPUT_SIZE
```
ValarDragon, line 476: is there any risk of this API being used by a non-standard usage of bitcoin script that is much larger? If not, than 10% SGTM

nuttycom (commented):
self-utACK 7caed142bd1bacb9e7438209806a9b79a196f6c9

schell (approved):
Nothing jumps out at me, but I'm interested to hear an answer to @ValarDragon's question.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: Right now zcash_client_backend only supports p2pkh and a very limited subset of p2sh scripts, basically just standard multisigs, so this is fine I think.

PR #2490, opened 2026-06-28
Add PCZT v2 serialization

ValarDragon:
## Summary
- add PCZT version 2 serialization
- add Orchard v2 bundle serialization with bundle-level note version
- omit empty Transparent, Sapling, and Orchard bundles from v2 encodings

For Orchard v2 bundles, there isn't a "blank value struct" that covers all cases, because of "NoteVersion" and "Flags". So we detect whether to omit for serialization by running orchard::v2::is_empty, which counts as empty if everything but flags and note version are empty. Notably, we don't mark as empty if just Anchor is set even though we could in v6 txs due to it not being signed over. This is kept for v5 compatability.

## Stack
1. This PR
2. Add logical Ironwood PCZT bundle
3. Add PCZT Ironwood role support

## Testing
- cargo fmt --all
- cargo check -p pczt --all-features
- cargo test -p pczt --all-features empty_bundles_encode_as_none_and_decode_as_empty

### DIFF
--- a/pczt/src/lib.rs
+++ b/pczt/src/lib.rs
@@ -257,8 +257,9 @@ pub mod v2 {
         fn empty_bundles_encode_as_none_and_decode_as_empty() {
             // Zero anchors: the shielded bundles carry no anchor and no
             // spends/actions, so they are fully empty and omitted.
-            let pczt =
-                Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [0; 32], [0; 32]).build();
+            let pczt = Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [0; 32], [0; 32])
+                .unwrap()
+                .build();
 
             let encoded = Pczt::try_from(pczt).unwrap();
 
@@ -288,8 +289,9 @@ pub mod v2 {
             // A Sapling/Orchard bundle with a non-empty anchor differs from its
             // empty form, so it must not be omitted even with no spends/actions,
             // and the anchor must survive the v2 round-trip.
-            let pczt =
-                Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [1; 32], [2; 32]).build();
+            let pczt = Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [1; 32], [2; 32])
+                .unwrap()
+                .build();
 
             let encoded = Pczt::try_from(pczt).unwrap();
 
@@ -305,8 +307,9 @@ pub mod v2 {
 
         #[test]
         fn orchard_flags_and_note_version_do_not_prevent_omission() {
-            let mut pczt =
-                Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [0; 32], [0; 32]).build();
+            let mut pczt = Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [0; 32], [0; 32])
+                .unwrap()
+                .build();
             pczt.orchard.flags = 0;
             pczt.orchard.note_version = NoteVersion::V3;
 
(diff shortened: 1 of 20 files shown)


pczt/src/orchard.rs
```diff
@@ -475,6 +482,130 @@ pub mod v1 {
...
+                SerializedNoteVersion::V3 => Self::V3,
+            }
+        }
+    }
+
+    /// PCZT fields that are specific to producing the transaction's Orchard bundle.
+    #[derive(Clone, Debug, Serialize, Deserialize, Getters)]
+    pub struct Bundle {
+        #[getset(get = "pub")]
+        pub(crate) actions: Vec<v1::Action>,
```
ValarDragon: We use the v1 serialization defn of action here, because it is identical.

pczt/src/orchard.rs
```diff
@@ -19,13 +19,15 @@ use crate::{
 };
 
 #[cfg(not(feature = "orchard"))]
+#[allow(dead_code)]
```
ValarDragon: Erroneous leftover, I remove in #2491

ValarDragon: EDIT: Had to do another commit for @czarcas7ic 's comment, so fixing here !

pczt/src/orchard.rs
```diff
@@ -475,6 +482,130 @@ pub mod v1 {
...
+    };
+
+    /// Whether `bundle` can be omitted from the v2 encoding.
+    fn is_empty(bundle: &super::Bundle) -> bool {
+        bundle.actions == EMPTY_BUNDLE.actions
+            && bundle.value_sum == EMPTY_BUNDLE.value_sum
+            && bundle.anchor == EMPTY_BUNDLE.anchor
+            && bundle.zkproof == EMPTY_BUNDLE.zkproof
+            && bundle.bsk == EMPTY_BUNDLE.bsk
+    }
```
czarcas7ic, line 590: Lets add code comment explaining why we leave out `flags` and `note_version` from the is_empty check.

ValarDragon: Good call, done!

pczt/src/orchard.rs
```diff
@@ -475,6 +481,134 @@ pub mod v1 {
...
+
+    /// PCZT fields that are specific to producing the transaction's Orchard bundle.
+    #[derive(Clone, Debug, Serialize, Deserialize, Getters)]
+    pub struct Bundle {
+        #[getset(get = "pub")]
+        pub(crate) actions: Vec<v1::Action>,
+        #[getset(get = "pub")]
+        pub(crate) flags: u8,
+        #[getset(get = "pub")]
```
nuttycom: We don't need the `getset`s for the serialization types, right?

nuttycom: Also I thought these could be private rather than pub(crate)?

pczt/src/lib.rs
```diff
@@ -136,6 +137,145 @@ pub mod v1 {
...
+    /// omitting empty Transparent, Sapling, and Orchard bundles.
+    impl TryFrom<super::Pczt> for Pczt {
+        type Error = super::EncodingError;
+
+        fn try_from(pczt: super::Pczt) -> Result<Self, Self::Error> {
+            Ok(Self {
+                global: pczt.global,
+                transparent: (pczt.transparent != EMPTY_TRANSPARENT).then_some(pczt.transparent),
+                sapling: (pczt.sapling != EMPTY_SAPLING).then_some(pczt.sapling),
+                orchard: pczt.orchard.try_into()?,
```
nuttycom, line 216: Without going to look at the type of `pczt.orchard` it's not obvious why this is different from transparent and sapling.

nuttycom (approved):
utACK 236d47d4ab5245efd2d03374ed8dc1bad4be632d

PR #2491, opened 2026-06-28
Add logical Ironwood PCZT bundle

ValarDragon:
## Summary
- add an unstable logical Ironwood bundle to pczt::Pczt
- initialize empty Ironwood bundles in PCZT creation and decoding paths
- preserve Ironwood through existing PCZT role transitions without adding role-specific behavior yet

## Stack
1. #2490
2. This PR
3. Add PCZT Ironwood role support

## Testing
- cargo fmt --all
- cargo check -p pczt --all-features
- RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo check -p pczt --all-features
- RUSTFLAGS='--cfg zcash_unstable="nu6.3"' cargo check -p pczt --all-features
- RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo test -p pczt --all-features empty_bundles_encode_as_none_and_decode_as_empty

PR #2492, opened 2026-06-28
Add PCZT Ironwood role support

ValarDragon:
## Summary
- parse and serialize Ironwood PCZT bundles with the V6 Orchard bundle format
- integrate Ironwood into extraction, IO finalization, proving, low-level signing, redaction, and verification
- dispatch PCZT sighashes between V5 and V6
- make TransactionData::from_parts_v6 available under nu7 with ZIP 233 amount support

## Stack
1. #2490
2. #2491
3. This PR

## Testing
- cargo fmt --all
- cargo check -p pczt --all-features
- RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo check -p pczt --all-features
- RUSTFLAGS='--cfg zcash_unstable="nu6.3"' cargo check -p pczt --all-features
- RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo test -p pczt --all-features empty_bundles_encode_as_none_and_decode_as_empty

### DIFF
--- a/pczt/src/roles/io_finalizer/mod.rs
+++ b/pczt/src/roles/io_finalizer/mod.rs
@@ -29,10 +29,15 @@ impl IoFinalizer {
     pub fn finalize_io(self) -> Result<Pczt, Error> {
         let Self { pczt } = self;
 
+        let has_orchard_actions = !pczt.orchard.actions.is_empty();
+        #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
+        let has_ironwood_actions = !pczt.ironwood.actions.is_empty();
+        #[cfg(not(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7")))]
+        let has_ironwood_actions = false;
         let has_shielded_spends =
-            !(pczt.sapling.spends.is_empty() && pczt.orchard.actions.is_empty());
+            !(pczt.sapling.spends.is_empty() && !has_orchard_actions && !has_ironwood_actions);
         let has_shielded_outputs =
-            !(pczt.sapling.outputs.is_empty() && pczt.orchard.actions.is_empty());
+            !(pczt.sapling.outputs.is_empty() && !has_orchard_actions && !has_ironwood_actions);
 
         // We can't build a transaction that has no spends or outputs.
         // However, we don't attempt to reject an entirely dummy transaction.
@@ -49,7 +54,7 @@ impl IoFinalizer {
             mut sapling,
             mut orchard,
             #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
-            ironwood,
+            mut ironwood,
             tx_data,
         } = pczt.extract_tx_data(
             |t| {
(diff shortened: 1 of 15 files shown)


pczt/src/roles/io_finalizer/mod.rs
```diff
@@ -76,14 +85,19 @@ impl IoFinalizer {
...
 
         Ok(Pczt {
             global,
             transparent: crate::transparent::Bundle::serialize_from(transparent),
             sapling: crate::sapling::Bundle::serialize_from(sapling),
             orchard: crate::orchard::Bundle::serialize_from(orchard),
             #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
-            ironwood,
+            ironwood: empty_ironwood
+                .unwrap_or_else(|| crate::orchard::Bundle::serialize_from(ironwood)),
```
ValarDragon: Make this an update

`(!has_ironwood_actions).then(|| pczt.ironwood.clone()).else(crate::orchard::Bundle::serialize_from(ironwood));`

And add a comment for "We do this so the serialize from preserves note verison v3"

ValarDragon:
After rebase, logical type carries BundleVersion, use that to solve for orchard vs ironwood, instead of inferring off of note version

See if this also solves the io-finalizer oddity

ValarDragon:
Rebased, and both oddities solved!

pczt/Cargo.toml
```diff
@@ -98,7 +98,7 @@ transparent = [
 ]
 
 ## Enables building a PCZT from the output of `zcash_primitives`'s `Builder::build_for_pczt`.
-zcp-builder = ["dep:zcash_primitives"]
+zcp-builder = ["dep:zcash_primitives", "zcash_primitives/zip-233"]
```
nuttycom: These flag additions seem wrong.

nuttycom: All of these should be `?` suffixed so that `zip-233` isn't transitively enabled by the `zcp-builder` flag.

nuttycom (changes requested):
Fixing the zip-233 feature flagging issue is blocking.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -397,6 +397,36 @@ impl<A: Authorization> TransactionData<A> {
...
+        sprout_bundle: Option<sprout::Bundle>,
+        sapling_bundle: Option<sapling::Bundle<A::SaplingAuth, ZatBalance>>,
+        orchard_bundle: Option<orchard::Bundle<A::OrchardAuth, ZatBalance>>,
+    ) -> Self {
+        Self::from_parts(
+            version,
+            consensus_branch_id,
+            lock_time,
+            expiry_height,
+            #[cfg(all(zcash_unstable = "nu7", feature = "zip-233"))]
```
nuttycom: The feature isn't defined; do we even need `from_parts_default_zip233`? What's using it?

nuttycom: We can probably keep the feature, but shouldn't we then gate this whole function?

pczt/Cargo.toml
```diff
@@ -69,7 +69,6 @@ zip32.workspace = true
 [features]
 default = ["std"]
 std = ["document-features"]
-zip-233 = ["zcash_primitives/zip-233"]
```
nuttycom, line 72: If this is removed why is there still a bunch of new zip233 code?

nuttycom (approved):
utACK d012e1068d79a9a6df3400a7d28177d504bdd235

PR #2493, opened 2026-06-29
Update Orchard to the BundleVersion API

nuttycom:
Stacked on `adam/bump-orchard-30c4ea2`, and pairs with zcash/orchard#519.

Bumps the patched `orchard` dependency to the revision that replaces `BundlePoolRestrictions` with `BundleVersion` (each `Bundle` now carries its `(value pool, protocol version)`), and adapts the integration to the new API.

## Summary
- **Transaction (de)serialization** derives the `BundleVersion` for each Orchard/Ironwood slot from the consensus context and threads it into bundle construction. The v5 read maps the branch to the version (pre-NU6.2 → `orchard_insecure_v0`, NU6.2 → `orchard_v1`, NU6.3+ → `orchard_v2`), which also subsumes the former `ProofSizeEnforcement` choice.
- **txid / authorizing commitments** read the version from the bundle; `commitment`/`authorizing_commitment` drop their version argument, and serialization uses the new infallible `Bundle::flag_byte`. The empty-bundle commitment helpers now take a `ValuePool`.
- **PCZT Creator** takes `with_orchard_bundle_version(version, flags)`; the fee and decryption call sites drop their version arguments.

## Testing
Full workspace test suite passes under `--cfg zcash_unstable="nu6.3"` with `--all-features` (34/34 suites). Workspace also compiles clean under default and `nu7`. `cargo vet --locked` passes.

> [!NOTE]
> The `orchard` patch points at the `zcash/orchard` `chore/naming_nits` branch backing #519; it will need re-pointing as that PR merges/rebases into `feat/ironwood`.

---
Opened as a draft for review. Filed with Claude Code assistance.

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -9,7 +9,7 @@ use core::cmp::Ordering;
 use ff::PrimeField;
 use getset::Getters;
 #[cfg(feature = "orchard")]
-use orchard::bundle::BundlePoolRestrictions;
+use orchard::bundle::BundleVersion;
 #[cfg(feature = "orchard")]
 pub(crate) use orchard::note::NoteVersion;
 
@@ -23,16 +23,6 @@ use crate::{
 pub(crate) enum NoteVersion {
     V2,
 }
-
-/// Orchard pool restriction for PCZT operations whose result only depends on
-/// using an Orchard-equivalent flag format.
-///
-/// PCZT v1 serializes Orchard flags with bits 2..=7 reserved and only supports
-/// V2 Orchard note plaintexts. The selected `BundlePoolRestrictions` value names
-/// that flag format; it is not an NU6.2 network-height assumption.
-#[cfg(feature = "orchard")]
-pub(crate) const ANY_ORCHARD_POOL_RESTRICTIONS: BundlePoolRestrictions =
-    BundlePoolRestrictions::OrchardNu6_2Only;
 
 /// PCZT fields that are specific to producing the transaction's Orchard bundle (if any).
 #[derive(Clone, Debug, Getters)]
@@ -695,7 +685,8 @@ impl Bundle {
         orchard::pczt::Bundle::parse(
             actions,
             self.flags,
-            ANY_ORCHARD_POOL_RESTRICTIONS,
+            // PCZT v1 only carries pre-NU6.3 Orchard bundles (V2 note plaintexts, bit 2 reserved).
+            BundleVersion::orchard_v1(),
             self.value_sum,
             self.anchor,
             self.zkproof,
@@ -806,10 +797,7 @@ impl Bundle {
 
         Self {
             actions,
-            flags: bundle
-                .flags()
-                .to_byte(ANY_ORCHARD_POOL_RESTRICTIONS)
-                .expect("Orchard flags must be representable in the v5 transaction format"),
+            flags: bundle.flag_byte(),
             value_sum,
             anchor: bundle.anchor().to_bytes(),
             note_version,
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -686,7 +686,7 @@ impl Bundle {
             actions,
             self.flags,
             // PCZT v1 only carries pre-NU6.3 Orchard bundles (V2 note plaintexts, bit 2 reserved).
-            BundleVersion::orchard_v1(),
+            BundleVersion::orchard_v2(),
             self.value_sum,
             self.anchor,
             self.zkproof,
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -9,7 +9,7 @@ use core::cmp::Ordering;
 use ff::PrimeField;
 use getset::Getters;
 #[cfg(feature = "orchard")]
-use orchard::bundle::BundleFormat;
+use orchard::bundle::BundlePoolRestrictions;
 #[cfg(feature = "orchard")]
 pub(crate) use orchard::note::NoteVersion;
 
@@ -635,7 +635,6 @@ impl Bundle {
                     action.spend.value,
                     action.spend.rho,
                     action.spend.rseed,
-                    note_version,
                     action.spend.fvk,
                     action.spend.witness,
                     action.spend.alpha,
@@ -650,6 +649,7 @@ impl Bundle {
                         })
                         .transpose()?,
                     action.spend.dummy_sk,
+                    note_version,
                     action.spend.proprietary,
                 )?;
 
@@ -662,7 +662,6 @@ impl Bundle {
                     action.output.recipient,
                     action.output.value,
                     action.output.rseed,
-                    note_version,
                     action.output.ock,
                     action
                         .output
@@ -675,6 +674,7 @@ impl Bundle {
                         })
                         .transpose()?,
                     action.output.user_address,
+                    note_version,
                     action.output.proprietary,
                 )?;
 
@@ -685,7 +685,7 @@ impl Bundle {
         orchard::pczt::Bundle::parse(
             actions,
             self.flags,
-            BundleFormat::PreNu6_3,
+            BundlePoolRestrictions::OrchardNu6_2Only,
             self.value_sum,
             self.anchor,
             self.zkproof,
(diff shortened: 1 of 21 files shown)


nuttycom:
This is an alternative that builds atop #2480 and depends on zcash/orchard#519. The diff vs #2480 is https://github.com/zcash/librustzcash/compare/98e7383e9d5f2cc93fb5bbfaeee6befbf30a48ee..5740ea2

nuttycom:
[force-pushed](https://github.com/zcash/librustzcash/compare/33d9e7b..1fc64ed5ac918c266b8fc573fbaa85fc4abd1666) to address comments from code review.

zcash_primitives/src/transaction/txid.rs
```diff
@@ -80,19 +83,24 @@ fn sapling_auth_includes_anchor(version: TxVersion) -> bool {
...
+/// Selects the `(value pool, orchard tx version)` pair that reproduces the
+/// pre-bump `BundleCommitmentDomain` for a given transaction version. The value
+/// pool here is only used for empty-bundle commitments (which hash no flags);
+/// present bundles compute their commitments from the `BundleVersion` each
+/// bundle carries.
+fn orchard_commitment_domain(version: TxVersion) -> (ValuePool, OrchardTxVersion) {
     match version {
         TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 | TxVersion::V5 => {
-            BundleCommitmentDomain::ORCHARD_V5_PRE_NU6_3
+            (ValuePool::Orchard, OrchardTxVersion::V5)
```
nuttycom, line 94: This is really the critical change that the API updates enable. Later, when we extract the sighash computation from the Orchard crate, this is where we will get the benefits, but it's also clearer right now.

nuttycom (commented):
self-utACK

pczt/src/roles/creator/mod.rs
```diff
@@ -60,12 +62,27 @@ impl Creator {
...
-    pub fn with_orchard_flags(mut self, orchard_flags: orchard::bundle::Flags) -> Self {
-        self.orchard_flags = orchard_flags
-            .to_byte(orchard::bundle::BundleFormat::PreNu6_3)
-            .expect("Orchard flags must be representable in the v5 transaction format");
-        self
+    pub fn with_orchard_bundle_version(
+        mut self,
+        bundle_version: orchard::bundle::BundleVersion,
+        flags: orchard::bundle::Flags,
+    ) -> Result<Self, orchard::bundle::BundleError> {
```
ValarDragon: We probably want to keep the old name, right?

pczt/src/orchard.rs
```diff
@@ -685,7 +685,8 @@ impl Bundle {
         orchard::pczt::Bundle::parse(
             actions,
             self.flags,
-            BundleFormat::PreNu6_3,
+            // PCZT v1 only carries pre-NU6.3 Orchard bundles (V2 note plaintexts, bit 2 reserved).
+            BundleVersion::orchard_v1(),
```
ValarDragon: orchard_v1() confuses me 😅 , I'd think that is PreNu6_2

nuttycom: I'm mostly indifferent to the numbers here; if you want to index 1-based instead of 0-based that's fine with me.

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -98,21 +99,50 @@ fn read_bundle<R: Read>(
...
+    bundle_version: BundleVersion,
 ) -> io::Result<Option<orchard::Bundle<Authorized, ZatBalance>>> {
-    read_bundle(reader, proof_size_enforcement, BundleFormat::PreNu6_3)
+    read_bundle(reader, bundle_version)
 }
 
+/// Rejects bundle versions that are not valid for the v6 transaction format, which has exactly
+/// two Orchard-bundle slots: the Orchard slot ([`BundleVersion::orchard_v2`]) and the Ironwood
+/// slot ([`BundleVersion::ironwood_v2`]). A pre-NU6.3 version would (de)serialize the flag byte
+/// with the wrong cross-address (bit 2) semantics.
```
ValarDragon: A pre-NU6.3 version on Ironwood would!

In practice, it should be identical for NU6.3 on Orchard

pczt/src/roles/creator/mod.rs
```diff
@@ -29,6 +29,7 @@ pub struct Creator {
     expiry_height: u32,
     coin_type: u32,
     orchard_flags: u8,
+    orchard_note_version: crate::orchard::NoteVersion,
```
ValarDragon: I don't think this is right and we should remove it. In the call sites for Orchard bundle we always put v2, and in the call sites for Ironwood we just put v3

pczt/src/roles/creator/mod.rs
ValarDragon: We won't set note version in the final API here, but will address in #2491

pczt/src/roles/creator/mod.rs
```diff
@@ -60,12 +62,27 @@ impl Creator {
...
+    /// version).
     #[cfg(feature = "orchard")]
-    pub fn with_orchard_flags(mut self, orchard_flags: orchard::bundle::Flags) -> Self {
-        self.orchard_flags = orchard_flags
-            .to_byte(orchard::bundle::BundleFormat::PreNu6_3)
-            .expect("Orchard flags must be representable in the v5 transaction format");
-        self
+    pub fn with_orchard_bundle_version(
+        mut self,
+        bundle_version: orchard::bundle::BundleVersion,
```
ValarDragon: We realized we can remove this, because its implied by consensus branch ID

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -710,7 +710,11 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                 ),
                 #[cfg(feature = "orchard")]
                 &(
-                    ::orchard::BundleProtocol::OrchardPreNu6_3,
+                    // Preserve the legacy Orchard action-count estimate here.
+                    // If this path targets restrictions that disable
+                    // cross-address transfers, thread the target-height-selected
+                    // `BundleVersion` through this call.
+                    crate::ANY_ORCHARD_BUNDLE_VERSION,
```
ValarDragon, line 717: Right, because we will update in this the zcash_client_backend PR

pczt/src/roles/creator/mod.rs
nuttycom: Instead of `with_orchard_flags` taking the BundleVersion, we should determine the bundle version from the consensus branch ID (and make this fallible so that we can reject unknown consensus branch IDs or IDs below v5. Then, we can store the bundle version and `with_orchard_flags` can check against it.

zcash_primitives/src/transaction/txid.rs
```diff
@@ -344,7 +352,8 @@ impl<A: Authorization> TransactionDigest<A> for TxIdDigester {
         orchard_bundle: Option<&orchard::Bundle<A::OrchardAuth, ZatBalance>>,
     ) -> Self::OrchardDigest {
         orchard_bundle.map(|b| {
-            b.commitment_for_domain(orchard_commitment_domain(version))
+            let (_, tx_version) = orchard_commitment_domain(version);
```
ValarDragon, line 355: (not for this PR) but can we just remove the `orchard_commitment_domain` now?

zcash_primitives/src/transaction/txid.rs
```diff
@@ -356,7 +365,8 @@ impl<A: Authorization> TransactionDigest<A> for TxIdDigester {
         ironwood_bundle: Option<&orchard::Bundle<A::OrchardAuth, ZatBalance>>,
     ) -> Self::IronwoodDigest {
         ironwood_bundle.map(|b| {
-            b.commitment_for_domain(ironwood_v6_domain())
+            let (_, tx_version) = ironwood_v6_domain();
```
ValarDragon, line 368: (not for this PR) but can we just remove the `ironwood_v6_domain` now?

zcash_primitives/src/transaction/txid.rs
ValarDragon: We can;t for empty hashing

ValarDragon (dismissed):
LGTM! Great job on the cleaner API

Good to merge w/o re-ACK from me, on pczt/src/roles/creator update

ValarDragon (approved):
Latest commit LGTM! Nice

nuttycom (commented):
self-utACK

PR #2495, opened 2026-06-29
Add methods to differentiate sapling/orchard/transparent payments

### LINKED ISSUE #217: rpc: Design and implement `z_sendfromaccount`
`z_sendmany` has weird semantics inherited from several iterations of wallet design. In particular, you have to specify an address to send from, but for UAs we use those as representatives of their corresponding accounts (rather than only selecting funds that were sent to the specified addresses).

We should instead create a new RPC method that takes one (or more?) account UUID(s) as the source of funds.

Additionally, we should consider whether this RPC should be designed around proposals:
- `z_sendfromaccount` would return a proposal object, that includes information about the privacy implications of the transaction that the caller can review.
  - This does not necessarily have to be the same `Proposal` protobuf as the mobile SDKs use, because Zallet can internally cache proposed transactions. It could instead directly convey just the (economic and privacy) effects of the proposed transaction.
  - Zallet could prepare and create proofs for the proposed transaction asynchronously while waiting for spending approval.
- A second RPC would take a pointer (UUID?) to a proposal, and calling it would constitute acceptance of the transaction. Any required signatures would be applied at this time, and then the transaction would be broadcast.

This has similarities to / overlaps with the plan to add RPC methods for working with PCZTs (#99).

dannywillems:


components/zcash_address/src/lib.rs
```diff
@@ -348,6 +348,23 @@ impl ZcashAddress {
             _ => false,
         }
     }
+
+    /// Returns whether this address can be used in a Sapling transfer
+    pub fn is_sapling(&self) -> bool {
```
nuttycom, line 353:
```suggestion
    pub fn can_receive_sapling(&self) -> bool {
```

components/zcash_address/src/lib.rs
```diff
@@ -348,6 +348,23 @@ impl ZcashAddress {
...
+        match &self.kind {
+            AddressKind::Sapling(_) => true,
+            AddressKind::Unified(addr) => addr.has_receiver_of_type(PoolType::SAPLING),
+            _ => false,
+        }
+    }
+
+    /// Returns whether this address can be used in an Orchard transfer
+    pub fn is_orchard(&self) -> bool {
```
nuttycom, line 362:
```suggestion
    pub fn can_receive_orchard(&self) -> bool {
```

components/zip321/src/lib.rs
nuttycom, line 1: I can't fully understand the changes to this file, because it's not possible in ZIP 321 to resolve whether the Orchard or Sapling receiver of a unified address will be selected for use in transaction construction. Depending on the source of funds, either one could be used, or even both (to prevent payment amounts being revealed by crossing the turnstile.)

nuttycom (changes requested):

PR #2496, opened 2026-06-29
Document fields of Step and ReceivedNote

dannywillems:


dannywillems:
Closing. I've just understood, based on a quick overview of the other structures, that we don't document the private fields, and rely on the public accessors.

PR #2497, opened 2026-06-29
Extract lambdas from involves to reuse in zallet

dannywillems:
These methods will be reused in zallet in an upcoming patch.


nuttycom (approved):
utACK

PR #2499, opened 2026-06-30
Expose Orchard change outputs in tx builder

czarcas7ic:
This adds the `zcash_primitives` transaction-builder entry point for wallet-controlled Orchard change outputs by forwarding to Orchard's existing change-output builder API.

It also makes the combined Orchard and Ironwood action-count variable explicit, and strengthens the Orchard v3 action-count test coverage for the cross-address-disabled case where requested spends and wallet-controlled change outputs occupy separate actions.

### DIFF
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -746,6 +746,33 @@ impl<P: consensus::Parameters, U> Builder<P, U> {
             .map_err(Error::OrchardRecipient)
     }
 
+    /// Adds a wallet-controlled Orchard change output to the transaction.
+    ///
+    /// Returns [`Error::OrchardBuilderNotAvailable`] if this builder is not
+    /// configured with an Orchard bundle builder. Returns
+    /// [`Error::OrchardRecipient`] if the Orchard builder rejects the recipient
+    /// or cannot construct the output.
+    pub fn add_orchard_change_output<FE>(
+        &mut self,
+        fvk: orchard::keys::FullViewingKey,
+        ovk: Option<orchard::keys::OutgoingViewingKey>,
+        recipient: orchard::Address,
+        value: Zatoshis,
+        memo: MemoBytes,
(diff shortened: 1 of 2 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -991,7 +1018,10 @@ impl<P: consensus::Parameters, U> Builder<P, U> {
             .map_err(FeeError::Bundle)?;
 
         #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
-        let orchard_actions = orchard_actions + ironwood_actions;
+        let orchard_plus_ironwood_actions = orchard_actions + ironwood_actions;
+
+        #[cfg(not(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7")))]
+        let orchard_plus_ironwood_actions = orchard_actions;
```
czarcas7ic, line 1024: Do we really need / want to do this? Pre nu6.3 wont ironwood_actions just be zero anyway?

zcash_primitives/src/transaction/builder.rs
```diff
@@ -746,6 +746,33 @@ impl<P: consensus::Parameters, U> Builder<P, U> {
...
+        fvk: orchard::keys::FullViewingKey,
+        ovk: Option<orchard::keys::OutgoingViewingKey>,
+        recipient: orchard::Address,
+        value: Zatoshis,
+        memo: MemoBytes,
+    ) -> Result<(), Error<FE>> {
+        self.orchard_builder
+            .as_mut()
+            .ok_or(Error::OrchardBuilderNotAvailable)?
+            .add_change_output(
```
czarcas7ic, line 766: Uses the ct randomization orchard func

zcash_primitives/src/transaction/builder.rs
ValarDragon: So the API looks different when we target to main because of #2484

PR #2505, opened 2026-06-30
Add `Builder::with_expiry_height` and enforce coinbase expiry height

czarcas7ic:
Adds `Builder::with_expiry_height` to override the transaction expiry height — setting `0` disables expiry for non-coinbase transactions. Coinbase builders additionally reject an overridden expiry height that does not match the target block height (the new `Error::CoinbaseExpiryHeightMismatch`); the check runs in both `build_internal` and `build_for_pczt`, with unit tests covering the override and the coinbase rejection.

### DIFF
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -725,6 +725,11 @@ impl<P: consensus::Parameters, U> Builder<P, U> {
     /// For non-coinbase transactions, setting this to `BlockHeight::from(0)`
     /// disables transaction expiry. Coinbase builders reject overridden expiry
     /// heights that do not match the target block height.
+    ///
+    /// Disabling expiry by setting the height to `BlockHeight::from(0)` is not
+    /// recommended: non-expiring transactions are not yet well tested
+    /// end-to-end and may surface edge cases elsewhere in the stack. Callers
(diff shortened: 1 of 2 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -708,6 +720,29 @@ impl<P: consensus::Parameters> Builder<P, ()> {
 }
 
 impl<P: consensus::Parameters, U> Builder<P, U> {
+    /// Overrides the expiry height for the transaction under construction.
+    ///
+    /// For non-coinbase transactions, setting this to `BlockHeight::from(0)`
+    /// disables transaction expiry. Coinbase builders reject overridden expiry
+    /// heights that do not match the target block height.
```
ValarDragon, line 727: Lets just add a warning that we don;t recommend doing 0, because it not end-to-end tested well?

ValarDragon (dismissed):

nuttycom (dismissed):
utACK

zcash_primitives/src/transaction/builder.rs
czarcas7ic: Good idea https://github.com/zcash/librustzcash/pull/2505/commits/9b7780ada7903f37b6a891ebfc1ea6f1320811dd

nuttycom (approved):

Issue #184, opened 2019-12-10
Add an API to zcash_client_sqlite that takes a Transaction, decrypts it, and adds its information to the data_db


str4d:
Part of #156.

str4d:
The core API was implemented in the previous sprint. The remaining work is to expose the API in the Android and iOS SDKs, and then make any adjustments based on their feedback.

str4d:
Done:
- https://github.com/zcash/zcash-android-wallet-sdk/pull/119
- https://github.com/zcash/ZcashLightClientKit/pull/102

PR #2507, opened 2026-06-30
Explicit shardtree checkpoint retention (durable anchors)

nuttycom:
## Summary

Consumer-side support for explicitly retaining note commitment tree checkpoints as durable "anchors": a retained checkpoint is exempt from automatic pruning of excess checkpoints, so its root — and witnesses for marked notes at or before it — remain computable even after it has aged more than `max_checkpoints` behind the chain tip.

Builds on the `shardtree` retention API added in zcash/incrementalmerkletree#184.

> [!NOTE]
> `Cargo.toml` patches `shardtree`/`incrementalmerkletree`/`incrementalmerkletree-testing` to `zcash/incrementalmerkletree` main, which carries the retention API from #184. This is a normal git patch on a `zcash/` repo and doesn't block merge; it's dropped in favor of released crate versions as a pre-publish step (see follow-ups).

Rebased onto `main` (Ironwood is now merged there); anchor retention uses the NU6.3 (Ironwood) activation height as its retention floor.

## What's here

- **`zcash_client_sqlite`** — implements the new `ShardStore` retained-checkpoint methods (`add_retained_checkpoint` / `remove_retained_checkpoint` / `retained_checkpoints`) for `SqliteShardStore`, backed by new per-pool `{sapling,orchard}_tree_retained_checkpoints` tables added by a migration (`db.rs` canonical schema + `verify_schema` updated). `retained_checkpoints` reports an empty set when the table predates the retention migration, so the tree stays usable at intermediate migration states.
- **`zcash_client_backend`** — during `put_blocks`, checkpoints at heights on a fixed interval (every 288 blocks, ≈4 per day) at or above the NU6.3 activation height are retained as anchors via `ShardTree::ensure_retained`. The retention floor is threaded through `put_blocks`/`update_tree` as `Option<BlockHeight>`; the SQLite backend computes it as `params.activation_height(NetworkUpgrade::Nu6_3)`.
- **`WalletCommitmentTrees::remove_retained_checkpoints_below(max_height)`** — releases anchors that have aged below a threshold (both pools) so they can be pruned normally.

Anchor retention currently covers the **Sapling and Orchard** trees only; extending it to the **Ironwood** note commitment tree (which activates at NU6.3) is tracked in #2544.

## Dormancy

`NetworkUpgrade::Nu6_3` now builds unconditionally — the `zcash_unstable = "nu6.3"` cfg was removed upstream when Ironwood/NU6.3 became a default build — so retention is compiled into all builds and gated only by the activation height. It is therefore active on networks that have an assigned NU6.3 height (testnet, at 4,134,000) and dormant where it is still `None` (mainnet).

## Testing

- SQLite store round-trip (both pools), the migration `test_migrate`, and `verify_schema`.
- `WalletCommitmentTrees::remove_retained_checkpoints_below` against a real `WalletDb`.
- The `should_retain_anchor` interval/floor gating predicate.
- `cargo test -p zcash_client_sqlite -p zcash_client_backend --all-features` passes (Ironwood/NU6.3 builds by default).

The full scan→retain integration path is not yet covered by an automated test (would need a consensus parameter set with an assigned NU6.3 activation height wired into a scanning fixture).

## Follow-ups

- **Before publishing these crates:** drop the `[patch.crates-io]` entries in favor of released `shardtree`/`incrementalmerkletree`/`incrementalmerkletree-testing` versions.
- Add `cargo vet` policy entries for the patched crates if CI runs `cargo vet` on this branch (`[policy.<crate>] audit-as-crates-io = false`).


LukasKorba (approved):
LGTM

LukasKorba (approved):

PR #2509, opened 2026-06-30
Crate releases supporting NU6.3 testnet activation.

### LINKED ISSUE #2450: zcash_primitives: Support NU6.3
---

### LINKED ISSUE #2456: zcash_protocol: Set testnet NU6.3 activation height
---

### LINKED ISSUE #2457: zcash_protocol release v0.10.0-rc1
---

### LINKED ISSUE #2478: zcash_protocol: Set NU6.3 consensus branch ID
---

nuttycom:
This release supports the needs of consensus nodes, with the crate stack complete and ready-for-use on testnet up through `zcash_primitives`, `zcash_proofs`, `zcash_keys` and `zcash_history`.


components/zcash_protocol/CHANGELOG.md
```diff
@@ -8,17 +8,18 @@ indicated by the `PLANNED` status in order to make it possible to correctly
...
-- `zcash_protocol::consensus::{NetworkUpgrade::Nu6_3, BranchId::Nu6_3}` under
-  `zcash_unstable = "nu6.3"`. The NU6.3 consensus branch ID is provisional until
-  the Ironwood / NU6.3 value is chosen.
-- `zcash_protocol::local_consensus::LocalNetwork::nu6_3` under
-  `zcash_unstable = "nu6.3"`.
+- `zcash_protocol::constants::{V6_TX_VERSION, V6_VERSION_GROUP_ID}`. The v6
+  version group ID is provisional until the Ironwood / NU6.3 value is chosen.
+- `zcash_protocol::consensus::{NetworkUpgrade::Nu6_3, BranchId::Nu6_3}`. The
+  NU6.3 consensus branch ID is provisional until the Ironwood / NU6.3 value is
+  chosen.
```
nuttycom: These are no longer provisional.

ValarDragon (dismissed):
LGTM! Happy to get rid of the nu6.3 FF's

I didn't do a detailed pass on the NU7 FF's items, because it won't affect consensus code right now.

nuttycom (commented):
self-utACK 00dc7eeecc6cd7a0ce62f6ab0bea1d2044f01e70

zancas (commented):
UtAck

components/zcash_protocol/src/consensus.rs
```diff
@@ -729,9 +723,8 @@ pub enum BranchId {
     Nu6_2,
     /// The consensus rules to be deployed by [`NetworkUpgrade::Nu6_3`].
     ///
-    /// This variant is gated by `zcash_unstable = "nu6.3"` and currently uses a
-    /// placeholder branch ID until the Ironwood / NU6.3 consensus branch ID is chosen.
-    #[cfg(zcash_unstable = "nu6.3")]
+    /// This variant currently uses a placeholder branch ID until the Ironwood /
```
zancas: Is this doc-comment stale?

components/zcash_protocol/src/consensus.rs
```diff
@@ -779,9 +770,7 @@ impl From<BranchId> for u32 {
             BranchId::Nu6 => 0xc8e7_1055,
             BranchId::Nu6_1 => 0x4dec_4df0,
             BranchId::Nu6_2 => 0x5437_f330,
-            #[cfg(zcash_unstable = "nu6.3")]
-            // TODO: Replace this placeholder once the Ironwood / NU6.3 consensus branch ID is chosen.
-            BranchId::Nu6_3 => 0xffff_ffff,
+            BranchId::Nu6_3 => 0x37a5_165b,
```
dannywillems, line 770: How are the consensus branch ID's defined?

dannywillems: [ZIP-0200](https://github.com/zcash/zips/blob/main/zips/zip-0200.rst) only mentions:

```
## CONSENSUS_BRANCH_ID
 
A globally-unique non-zero 32-bit identifier.

Implementations MAY use a value of zero in consensus branch ID fields to indicate the absence of any upgrade (i.e. that the Sprout consensus rules apply).
```

daira: In this case: https://github.com/zcash/zips/pull/1301/changes/d138c0dea9a5270006d2ca932731f4bfab3ac93b

pczt/src/roles/creator/mod.rs
```diff
@@ -207,15 +205,13 @@ impl Creator {
...
 
-        #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
         use zcash_protocol::constants::V6_TX_VERSION;
 
         let tx_version = match parts.version {
             zcash_primitives::transaction::TxVersion::Sprout(_)
             | zcash_primitives::transaction::TxVersion::V3 => None,
             zcash_primitives::transaction::TxVersion::V4 => Some(V4_TX_VERSION),
             zcash_primitives::transaction::TxVersion::V5 => Some(V5_TX_VERSION),
-            #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
```
dannywillems: Review note: as V6 was explictly activated with a disjunction between NU6.3 and NU7, I guess that we do not intent to have - or more precisely, we haven't had merged any work having - a different transaction version/format for the next NU7.

daira: We have not merged any work toward that, but it will likely be ZIP 248.

pczt/src/lib.rs
```diff
@@ -480,24 +460,12 @@ impl Pczt {
                 orchard_bundle,
             ),
         };
-        #[cfg(not(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7")))]
```
dannywillems: Review note: removed as the `cfg` attribute is exclusive.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -528,9 +506,9 @@ impl<A: Authorization> TransactionData<A> {
 
     /// Computes this transaction's digest using the provided digest strategy.
```
dannywillems, line 507: Review notes: I see, the transaction digest defines the transaction ID. ZIP-0244 describes it: https://zips.z.cash/zip-0244.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1266,38 +1221,12 @@ pub mod testing {
...
             BranchId::Nu6_1 => Just(TxVersion::V5).boxed(),
             BranchId::Nu6_2 => Just(TxVersion::V5).boxed(),
-            #[cfg(zcash_unstable = "nu6.3")]
             BranchId::Nu6_3 => Just(TxVersion::V6).boxed(),
             #[cfg(zcash_unstable = "nu7")]
             BranchId::Nu7 => Just(TxVersion::V6).boxed(),
         }
     }
 
-    #[cfg(all(not(zcash_unstable = "nu7"), not(zcash_unstable = "nu6.3")))]
```
dannywillems: Review notes: removed as now NU7 is also supposed to use the same transaction data structure. No reason to have this proptest strategy as the monorepo will only use V6 transaction format onward.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1362,7 +1291,7 @@ pub mod testing {
         }
     }
 
-    #[cfg(zcash_unstable = "nu6.3")]
+    #[cfg(not(zcash_unstable = "nu7"))]
```
dannywillems, line 1294: Review notes: why is it excluding NU7 now?
I initially thought it was because the strategy was parametrized by the consensus branch ID, which does not contain NU7 yet, but therefore, NU7 should have been also in the cfg attribute before?

zcash_primitives/src/transaction/txid.rs
```diff
@@ -52,15 +51,13 @@ const ZCASH_SAPLING_OUTPUTS_NONCOMPACT_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxId
...
-#[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
 const ZCASH_SAPLING_V6_SIGS_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxAuthSapliH_v6";
 
 fn sapling_spends_noncompact_personalization(version: TxVersion) -> &'static [u8; 16] {
     match version {
         TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 | TxVersion::V5 => {
             ZCASH_SAPLING_SPENDS_NONCOMPACT_HASH_PERSONALIZATION
         }
-        #[cfg(any(zcash_unstable = "nu6.3", zcash_unstable = "nu7"))]
         TxVersion::V6 => ZCASH_SAPLING_SPENDS_V6_NONCOMPACT_HASH_PERSONALIZATION,
```
dannywillems, line 61: Review notes: see https://github.com/zcash/zips/blob/main/zips/zip-0229.md for the V6 transaction format.

zcash_primitives/src/transaction/mod.rs
dannywillems: Ah okay, there is another arb_txdata defined for nu7 above, L1230

dannywillems (commented):
If it can be of any help: https://github.com/zcash/librustzcash/pull/2512.

Generated by Claude when asked to fix the CI. The local configuration that I'm on rn didn't add the commit "Co-Authored by"; I can amend it if necessary.

dannywillems:
> If it can be of any help: #2512.
> 
> Generated by Claude when asked to fix the CI. The local configuration that I'm on rn didn't add the commit "Co-Authored by"; I can amend it if necessary.

The missing "Co-Authored by" have been added in https://github.com/zcash/librustzcash/commit/9038ab1492f9dab80e0fc977ddbcc68117a941e9.

components/zcash_protocol/src/consensus.rs
nuttycom: Yes, good catch.

components/zcash_protocol/src/consensus.rs
```diff
@@ -714,6 +721,11 @@ pub enum BranchId {
     Nu6_1,
     /// The consensus rules deployed by [`NetworkUpgrade::Nu6_2`].
     Nu6_2,
+    /// The consensus rules to be deployed by [`NetworkUpgrade::Nu6_3`].
+    ///
+    /// This variant currently uses a placeholder branch ID until the Ironwood /
+    /// NU6.3 consensus branch ID is chosen.
```
nuttycom:
```suggestion
```

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -116,6 +123,8 @@ mod tests {
         assert!(regtest.is_nu_active(NetworkUpgrade::Nu6, expected_nu6));
         assert!(regtest.is_nu_active(NetworkUpgrade::Nu6_1, expected_nu6_1));
         assert!(regtest.is_nu_active(NetworkUpgrade::Nu6_2, expected_nu6_2));
+        // nu6_3 must not be activated at or below the nu6_2 height
+        assert!(!regtest.is_nu_active(NetworkUpgrade::Nu6_3, expected_nu6_2));
```
nuttycom:
```suggestion
        assert!(regtest.is_nu_active(NetworkUpgrade::Nu6_3, expected_nu6_3));
```

pczt/src/roles/creator/mod.rs
```diff
@@ -89,11 +166,25 @@ impl Creator {
...
+                note_version: crate::orchard::NoteVersion::V2,
+                zkproof: None,
+                bsk: None,
+            },
+            ironwood: OrchardBundle {
                 actions: vec![],
                 flags: self.orchard_flags,
                 value_sum: (0, true),
                 anchor: self.orchard_anchor,
+                note_version: NoteVersion::V3,
```
nuttycom, line 196: Why is this not feature-gated as in `OrchardBundle`?

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -75,26 +74,73 @@ pub fn read_v5_bundle<R: Read>(
...
-#[cfg(zcash_unstable = "nu7")]
+/// Reads an [`orchard::Bundle`] from a v5 transaction format.
+///
+/// This deliberately does not take the consensus branch ID to determine the
+/// format to read. Although NU6.3 does not disable v5 transactions, the Orchard
+/// action flags and wire serialization are identical before and after the
+/// upgrade, so a v5 bundle deserializes the same way regardless of activation
+/// height; only the bundle commitment domain varies across upgrades, and that
+/// affects the txid/sighash digests rather than deserialization. Since this is
+/// public API. By contrast, `read_v6_bundle` takes a `bundle_version` argument so
```
nuttycom, line 102: I think this rationale is bad; we should take the consensus branch ID here because we now store it in the in-memory format of the bundle. Nonblocking for this release but we should fix before 0.29.0.

dannywillems:
I'm keeping the branch [dw/merge-ironwood-branch](https://github.com/zcash/librustzcash/tree/dw/merge-ironwood-branch) up-to-date with this branch + main (`spend-index` is not on this branch yet) to have a branch in zallet compatible with NU6.3.
The bump up of librustzcash to [dw/merge-ironwood-branch](https://github.com/zcash/librustzcash/tree/dw/merge-ironwood-branch) will be in an individual commit (TODO add link to PR) that will need to be updated to point to main when this patch is merged.

pczt/src/roles/creator/mod.rs
ValarDragon: Orchard bundle ideally shouldn’t be feature gated. Pczt/orchard.rs should be handling the conditional compilation for note version

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -90,6 +95,7 @@ mod tests {
         let expected_nu6 = BlockHeight::from_u32(7);
         let expected_nu6_1 = BlockHeight::from_u32(8);
         let expected_nu6_2 = BlockHeight::from_u32(9);
+        let expected_nu6_3 = BlockHeight::from_u32(10);
         #[cfg(zcash_unstable = "nu7")]
         let expected_nu7 = BlockHeight::from_u32(10);
```
daira:
```suggestion
        let expected_nu7 = BlockHeight::from_u32(11);
```

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -132,6 +141,7 @@ mod tests {
...
```
daira:
```suggestion
        let expected_nu7 = BlockHeight::from_u32(11);
```

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -203,6 +218,7 @@ mod tests {
...
```
daira:
```suggestion
        let expected_nu7 = BlockHeight::from_u32(11);
```

pczt/src/roles/creator/mod.rs
```diff
@@ -10,17 +10,61 @@ use crate::{
...
+/// Errors that can occur when creating a PCZT.
+#[derive(Debug)]
+pub enum Error {
+    /// The consensus branch ID does not correspond to any known network upgrade.
+    UnknownConsensusBranchId,
+    /// The network upgrade for the consensus branch ID predates the v5
+    /// transaction format, so it cannot be used to create a PCZT.
+    UnsupportedConsensusBranchId,
+    /// The requested Orchard flags cannot be represented under the Orchard bundle
+    /// version implied by the consensus branch ID.
```
daira:
```suggestion
    /// The requested Orchard-protocol flags cannot be represented under the
    /// Orchard or Ironwood bundle version implied by the consensus branch ID.
```

pczt/src/roles/creator/mod.rs
nuttycom: The counterargument here is that `self.orchard_bundle_version.note_version()` is future-proof; otherwise we have to remember to come here and change the constant if we need a V4 note plaintext.

components/zcash_protocol/src/constants.rs
```diff
@@ -37,8 +37,6 @@ pub const V5_TX_VERSION: u32 = 5;
 pub const V5_VERSION_GROUP_ID: u32 = 0x26A7270A;
 
 /// Transaction version 6, specified in [ZIP 230](https://zips.z.cash/zip-0230).
```
daira:
```suggestion
/// Transaction version 6, specified in [ZIP 229](https://zips.z.cash/zip-0229).
```

pczt/src/roles/creator/mod.rs
```diff
@@ -50,20 +109,38 @@ impl Creator {
...
+    ///
+    /// The flags are validated against, and encoded under, the Orchard bundle
+    /// version implied by the consensus branch ID passed to [`Creator::new`]
+    /// (which also fixes the note-plaintext version).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::UnrepresentableOrchardFlags`] if `flags` cannot be encoded
+    /// under that bundle version (e.g. cross-address-enabled flags under a
+    /// post-NU6.3 Orchard version).
```
daira:
```suggestion
    /// post-NU6.3 Orchard version, or cross-address-disabled flags under a pre-NU6.3
    /// Orchard version).
```

pczt/src/roles/creator/mod.rs
daira: @ValarDragon We want *each* pool to be feature-gated (the fact that Sapling isn't is a bug: #1380, #1161, #1162). I agree with @nuttycom.

pczt/src/roles/tx_extractor/mod.rs
```diff
@@ -149,6 +158,7 @@ pub enum Error {
     SaplingRequired,
     SighashMismatch,
     Transparent(TransparentError),
+    Ironwood(IronwoodError),
```
daira:
```suggestion
    Ironwood(IronwoodError),
    Orchard(OrchardError),
    Sapling(SaplingError),
    SaplingRequired,
    SighashMismatch,
    Transparent(TransparentError),
```
Nit: this was currently in alphabetical order (and keeping it that way happens to also keep `Ironwood`, `Orchard`, `Sapling` in reverse-chronological order). Note that GitHub is messing up the diff.

pczt/src/roles/tx_extractor/orchard.rs
```diff
@@ -7,25 +11,43 @@ pub(super) fn verify_bundle(
...
-
-    validator.add_bundle(bundle, sighash);
+    if let Some(vk) = orchard_vk {
+        verify_bundle_with_key(bundle, vk, sighash)
+    } else {
+        // PCZT extraction produces new Orchard bundles using the NU6.2 fixed
+        // circuit.
+        let vk = VerifyingKey::build(OrchardCircuitVersion::FixedPostNu6_2);
+        verify_bundle_with_key(bundle, &vk, sighash)
+    }
```
daira, line 21: (Non-blocking) This should be expressible more simply using
```rust
orchard_vk.unwrap_or_else(|| &VerifyingKey::build(OrchardCircuitVersion::FixedPostNu6_2))
```

components/zcash_protocol/src/constants.rs
nuttycom:
```suggestion
/// Transaction version 6, specified in [ZIP 229](https://zips.z.cash/zip-0229).
```

pczt/src/roles/tx_extractor/orchard.rs
```diff
@@ -7,25 +11,43 @@ pub(super) fn verify_bundle(
...
-        if validator.validate(vk, rng) {
-            Ok(())
-        } else {
-            Err(OrchardError::InvalidProof)
-        }
+        verify_bundle_with_key(bundle, vk, sighash)
+    } else {
+        let vk = VerifyingKey::build(OrchardCircuitVersion::PostNu6_3);
+        verify_bundle_with_key(bundle, &vk, sighash)
+    }
```
daira, line 34: (Non-blocking) Same comment as above (with `PostNu6_3`).

pczt/src/roles/tx_extractor/orchard.rs
nuttycom: @daira related to this, `orchard::ProtocolVersion` now maps `1:1` with `OrchardCircuitVersion` - should we just unify the two? There's not a circumstance where `OrchardCircuitVersion` can add a variant without `orchard::ProtocolVersion` changing, and if we have non-circuit changes to the protocol, it's fine for the circuit code to unify handling for multiple `PrococolVersion`s?

components/zcash_protocol/src/local_consensus.rs
```diff
@@ -116,6 +123,8 @@ mod tests {
         assert!(regtest.is_nu_active(NetworkUpgrade::Nu6, expected_nu6));
         assert!(regtest.is_nu_active(NetworkUpgrade::Nu6_1, expected_nu6_1));
         assert!(regtest.is_nu_active(NetworkUpgrade::Nu6_2, expected_nu6_2));
+        // nu6_3 must not be activated at or below the nu6_2 height
+        assert!(!regtest.is_nu_active(NetworkUpgrade::Nu6_3, expected_nu6_2));
         // nu7 must not be activated at or below the nu6_2 height
         #[cfg(zcash_unstable = "nu7")]
         assert!(!regtest.is_nu_active(NetworkUpgrade::Nu7, expected_nu6_2));
```
daira:
```suggestion
        // nu7 must not be activated at or below the nu6_3 height
        #[cfg(zcash_unstable = "nu7")]
        assert!(!regtest.is_nu_active(NetworkUpgrade::Nu7, expected_nu6_3));
```

pczt/src/roles/creator/mod.rs
```diff
@@ -10,17 +10,61 @@ use crate::{
...
+    use orchard::bundle::BundleVersion;
+
+    match branch_id {
+        BranchId::Nu6_2 => BundleVersion::orchard_v2(),
+        BranchId::Nu6_3 => BundleVersion::orchard_v3(),
+        #[cfg(zcash_unstable = "nu7")]
+        BranchId::Nu7 => BundleVersion::orchard_v3(),
+        // NU5, NU6, and NU6.1 use the original (pre-NU6.2) Orchard pool; pre-NU5
+        // branches are rejected before reaching here.
+        _ => BundleVersion::orchard_insecure_v1(),
```
daira: Make it an exhaustive match anyway, to avoid a future maintenance hazard.

pczt/src/lib.rs
```diff
@@ -74,11 +85,248 @@ pub struct Pczt {
...
+        outputs: Vec::new(),
+    };
+
+    const EMPTY_SAPLING: sapling::Bundle = sapling::Bundle {
+        spends: Vec::new(),
+        outputs: Vec::new(),
+        value_sum: 0,
+        anchor: [0; 32],
+        bsk: None,
+    };
```
daira, line 179: Why EMPTY_TRANSPARENT and EMPTY_SAPLING, but no EMPTY_ORCHARD or EMPTY_IRONWOOD?

daira: Oh I see, it's an API improvement that was made with Orchard and never backported.

zcash_client_backend/lightwallet-protocol/walletrpc/service.proto
```diff
@@ -312,7 +317,7 @@ service CompactTxStreamer {
     rpc GetLatestTreeState(Empty) returns (TreeState) {}
 
     // Returns a stream of information about roots of subtrees of the note commitment tree
-    // for the specified shielded protocol (Sapling or Orchard).
+    // for the specified shielded protocol (Sapling, Orchard, or Ironwood).
```
daira:
```suggestion
    // for the specified shielded pool (Sapling, Orchard, or Ironwood).
```

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6201,7 +6201,9 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, Dsf>(
 
     // Create proofs.
     let sapling_prover = LocalTxProver::bundled();
-    let orchard_pk = ::orchard::circuit::ProvingKey::build();
+    let orchard_pk = ::orchard::circuit::ProvingKey::build(
+        ::orchard::circuit::OrchardCircuitVersion::FixedPostNu6_2,
+    );
```
daira, line 6206: I think there's likely inadequate test coverage of Ironwood in the pool tests. :-(

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -746,7 +746,11 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                 ),
                 #[cfg(feature = "orchard")]
                 &(
-                    ::orchard::builder::BundleType::DEFAULT,
+                    // Preserve the legacy Orchard action-count estimate here.
+                    // If this path targets restrictions that disable
+                    // cross-address transfers, thread the target-height-selected
+                    // `BundleVersion` through this call.
+                    crate::ANY_ORCHARD_BUNDLE_VERSION,
```
daira, line 753: I don't know whether this is correct.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -905,16 +909,27 @@ where
...
     };
     #[cfg(not(feature = "orchard"))]
     let orchard_action_count: usize = 0;
 
     #[cfg(feature = "orchard")]
     let use_orchard = orchard_action_count > 0;
 
+    // The wallet does not yet construct Ironwood bundles, so they contribute no
+    // actions to the fee.
+    let ironwood_action_count: usize = 0;
```
daira, line 931: OK this is how we get away with not making changes that would otherwise clearly be necessary in this file.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: This is correct for this PR; we have not yet released `zcash_client_backend` - those changes will be made on `main`.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -2555,6 +2567,8 @@ fn fake_compact_block_from_compact_tx(
             + cb.vtx.iter().map(|tx| tx.outputs.len() as u32).sum::<u32>(),
         orchard_commitment_tree_size: initial_orchard_tree_size
             + cb.vtx.iter().map(|tx| tx.actions.len() as u32).sum::<u32>(),
+        // The test framework does not generate Ironwood notes.
+        ironwood_commitment_tree_size: 0,
```
daira, line 2571: Inadequate test coverage. Please file an issue. (This is my major high-level conclusion about the PR.)

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -585,7 +585,7 @@ mod tests {
                 &[] as &[Infallible],
             ),
             &(
-                orchard::builder::BundleType::DEFAULT,
+                crate::ANY_ORCHARD_BUNDLE_VERSION,
```
daira, line 588: Inadequate test coverage.

zcash_client_backend/src/lib.rs
```diff
@@ -82,6 +82,15 @@ pub mod tor;
...
 pub use decrypt::{DecryptedOutput, TransferType, decrypt_transaction};
 
+/// Orchard bundle version for fee and change call sites that should not select a concrete
+/// branch-specific policy.
+///
+/// These paths only need a bundle version to drive the Orchard action-count policy; the chosen
+/// value preserves the legacy pre-NU6.3 Orchard behavior those paths used historically.
+#[cfg(feature = "orchard")]
+pub(crate) const ANY_ORCHARD_BUNDLE_VERSION: ::orchard::bundle::BundleVersion =
+    ::orchard::bundle::BundleVersion::orchard_v2();
```
daira, line 92: I don't think this approach will actually work. That is, it will break Orchard-pool bundle creation after NU6.3, because it won't be calculating the number of actions correctly for a same-address-required bundle.

pczt/src/roles/tx_extractor/orchard.rs
daira: There could very easily be non-circuit changes. A serious *non-circuit* consensus bug could require a `ProtocolVersion` change without a corresponding `OrchardCircuitVersion`. You could artificially make the circuits for two `ProtocolVersion`s the same, but there seems to be a valid logical distinction to me.

zcash_client_backend/src/lib.rs
daira: Apparently this was not supposed to work yet, but I don't know how I could have been expected to know that 😕

zcash_primitives/src/transaction/components/orchard.rs
daira: I agree. Also, "Since this is public API." is an incomplete sentence, I don't know what it was trying to say.

components/zcash_protocol/src/consensus.rs
```diff
@@ -834,12 +848,12 @@ impl BranchId {
             BranchId::Nu6_2 => params
                 .activation_height(NetworkUpgrade::Nu6_2)
                 .map(|lower| {
-                    #[cfg(zcash_unstable = "nu7")]
-                    let upper = params.activation_height(NetworkUpgrade::Nu7);
-                    #[cfg(not(zcash_unstable = "nu7"))]
-                    let upper = None;
+                    let upper = params.activation_height(NetworkUpgrade::Nu6_3);
                     (lower, upper)
                 }),
```
nuttycom, line 856: This was actually a bug; it's necessary to conditionally set the upper bound if nu7 is active. The fix however is NU7 gated so not a blocker for this PR.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -244,18 +281,65 @@ impl BuildConfig {
...
+            {
+                None
+            }
+            BuildConfig::Coinbase { .. } => Some(
+                orchard::builder::Builder::new(
+                    orchard::builder::BundleType::Coinbase,
+                    bundle_version,
+                    // Coinbase transactions have `enableSpends = 0`. Every pool for which a
+                    // coinbase Orchard bundle is built (Orchard pre-NU6.3 and Ironwood) permits
+                    // cross-address transfers, so the spends-disabled flag set is representable.
```
daira, line 310: (Non-blocking) What's written in the comment is true but not relevant here, because this code isn't used for Ironwood.
```suggestion
                    // Coinbase transactions have `enableSpends = 0`. Every protocol version
                    // for which a coinbase Orchard-pool bundle can be built (pre-NU6.3) permits
                    // cross-address transfers, so the spends-disabled flag set is representable.
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -265,6 +349,48 @@ impl BuildConfig {
...
+
+fn orchard_bundle_version_for_branch(
+    consensus_branch_id: BranchId,
+) -> orchard::bundle::BundleVersion {
+    match consensus_branch_id {
+        BranchId::Nu6_3 => orchard::bundle::BundleVersion::orchard_v3(),
+        #[cfg(zcash_unstable = "nu7")]
+        BranchId::Nu7 => orchard::bundle::BundleVersion::orchard_v3(),
+        BranchId::Nu6_2 => orchard::bundle::BundleVersion::orchard_v2(),
+        _ => orchard::bundle::BundleVersion::orchard_insecure_v1(),
```
daira, line 390: Make this an exhaustive match to avoid future maintenance hazards.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -398,17 +544,36 @@ impl<P, U> Builder<P, U> {
 
         let orchard_available = version.has_orchard() && self.consensus_branch_id.has_orchard();
         if !orchard_available
-            && self
-                .orchard_builder
-                .as_ref()
-                .is_some_and(|b| !b.spends().is_empty() || !b.outputs().is_empty())
+            && self.orchard_builder.as_ref().is_some_and(|b| {
```
daira, line 547: (Non-blocking) This could be more similar to the Ironwood code by factoring out `orchard_in_use`.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -512,12 +687,42 @@ impl<P: consensus::Parameters> Builder<P, ()> {
...
 impl<P: consensus::Parameters, U> Builder<P, U> {
+    /// Overrides the expiry height for the transaction under construction.
+    ///
+    /// For non-coinbase transactions, setting this to `BlockHeight::from(0)`
+    /// disables transaction expiry. Coinbase builders reject overridden expiry
+    /// heights that do not match the target block height.
+    ///
+    /// Disabling expiry by setting the height to `BlockHeight::from(0)` is not
+    /// recommended: non-expiring transactions are not yet well tested
+    /// end-to-end and may surface edge cases elsewhere in the stack. Callers
```
daira, line 706: (Non-blocking)
```suggestion
    /// end-to-end and are known to cause bugs elsewhere in the stack. Callers
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1156,6 +1542,462 @@ mod tests {
...
+            sapling: Some(BlockHeight::from_u32(2)),
+            blossom: Some(BlockHeight::from_u32(3)),
+            heartwood: Some(BlockHeight::from_u32(4)),
+            canopy: Some(BlockHeight::from_u32(5)),
+            nu5: Some(BlockHeight::from_u32(6)),
+            nu6: Some(BlockHeight::from_u32(7)),
+            nu6_1: Some(BlockHeight::from_u32(8)),
+            nu6_2: Some(BlockHeight::from_u32(9)),
+            nu6_3: None,
+            nu7: Some(BlockHeight::from_u32(10)),
```
daira, line 1585:
```suggestion
            nu6_3: Some(BlockHeight::from_u32(10)),
            nu7: Some(BlockHeight::from_u32(11)),
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1156,6 +1542,462 @@ mod tests {
...
+
+        assert!(builder.orchard_builder.is_none());
+    }
+
+    #[test]
+    #[cfg(all(feature = "circuits", zcash_unstable = "nu7"))]
+    fn nu7_coinbase_builder_does_not_expose_orchard() {
+        let builder = Builder::new(
+            nu7_test_network(),
+            zcash_protocol::consensus::BlockHeight::from_u32(10),
```
daira, line 1649: (If the change above is taken)
```suggestion
            zcash_protocol::consensus::BlockHeight::from_u32(11),
```

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1156,6 +1542,462 @@ mod tests {
...
+        builder
+            .propose_version::<Infallible>(TxVersion::V5)
+            .unwrap();
+
+        let recipient = orchard::keys::FullViewingKey::from(
+            &orchard::keys::SpendingKey::from_bytes([0; 32]).unwrap(),
+        )
+        .address_at(0u32, orchard::keys::Scope::External);
+        builder
+            .add_ironwood_output::<crate::transaction::fees::zip317::FeeRule>(
```
daira, line 1834: (Non-blocking) Arguably this call should fail.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -1156,6 +1542,462 @@ mod tests {
...
+                BranchId::Nu6_3,
+                TxVersion::V5,
+                None
+            ))
+        );
+    }
+
+    /// Test helper: returns a full viewing key, an Orchard note carrying the given
+    /// note plaintext `version`, and a dummy Merkle path, for exercising the
+    /// Ironwood builder's note-version handling.
```
daira, line 1857: Consider adding a test that the commitment and nullifier depend on the version (properly called lead byte), i.e. are different for lead-byte 0x02 and 0x03 notes that otherwise have the same fields.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -176,16 +168,19 @@ impl TxVersion {
...
             TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 => false,
             TxVersion::V5 => true,
-            #[cfg(zcash_unstable = "nu7")]
             TxVersion::V6 => true,
         }
     }
 
+    /// Returns `true` if this transaction version supports the Ironwood protocol.
+    pub fn has_ironwood(&self) -> bool {
+        matches!(self, TxVersion::V6)
```
daira, line 177: (Non-blocking) This is implicitly a non-exhaustive match. Make it an exhaustive one to avoid future maintenance hazards.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -468,8 +520,9 @@ impl<A: Authorization> TransactionData<A> {
                 &self.zip233_amount,
```
daira, line 520: (Non-blocking) This is technically incorrect because if `zip233Amount` were added, it would be in v7, not v6. The issue is that the NU7-gated code previously assumed that v6 belonged to it, but now NU7 does not have its own tx version with the needed boilerplate (and so if the NU7 tests still pass, they only do so by coincidence / insufficient test coverage). Not necessary to fix now.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -981,6 +1062,7 @@ impl Transaction {
         self.write_transparent(&mut writer)?;
         self.write_v5_sapling(&mut writer)?;
         orchard_serialization::write_v6_bundle(self.orchard_bundle.as_ref(), &mut writer)?;
+        orchard_serialization::write_v6_bundle(self.ironwood_bundle.as_ref(), &mut writer)?;
```
daira, line 1065: (Non-blocking) It mildly disturbs me that there's no generic parameter to ensure type safety in places that should take only Orchard-pool bundles or only Ironwood-pool bundles.

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1116,21 +1221,27 @@ pub mod testing {
...
             lock_time in any::<u32>(),
             expiry_height in any::<u32>(),
             transparent_bundle in transparent::arb_bundle(),
             sapling_bundle in sapling::arb_bundle_for_version(version),
             orchard_bundle in orchard::arb_bundle_for_version(version),
+            ironwood_bundle in if version.has_ironwood() {
+                orchard::arb_ironwood_bundle_for_version(version).boxed()
+            } else {
+                Just(None).boxed()
+            },
```
daira, line 1244: I'm confused, doesn't `arb_ironwood_bundle_for_version` work just like `orchard::arb_bundle_for_version`, returning `Just(None).boxed()` itself if `version` doesn't support Ironwood? So this `if` should be unnecessary.

```suggestion
            ironwood_bundle in orchard::arb_ironwood_bundle_for_version(version),
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1142,58 +1253,71 @@ pub mod testing {
...
             expiry_height in any::<u32>(),
+            zip233_amount in 0..=MAX_MONEY,
             transparent_bundle in transparent::arb_bundle(),
             sapling_bundle in sapling::arb_bundle_for_version(version),
             orchard_bundle in orchard::arb_bundle_for_version(version),
+            ironwood_bundle in if version.has_ironwood() {
+                orchard::arb_ironwood_bundle_for_version(version).boxed()
+            } else {
+                Just(None).boxed()
+            },
```
daira, line 1276: As above.

```suggestion
            ironwood_bundle in orchard::arb_ironwood_bundle_for_version(version),
```

zcash_primitives/src/transaction/mod.rs
```diff
@@ -1142,58 +1253,71 @@ pub mod testing {
...
             expiry_height in any::<u32>(),
-            zip233_amount in 0..=MAX_MONEY,
             transparent_bundle in transparent::arb_bundle(),
             sapling_bundle in sapling::arb_bundle_for_version(version),
             orchard_bundle in orchard::arb_bundle_for_version(version),
+            ironwood_bundle in if version.has_ironwood() {
+                orchard::arb_ironwood_bundle_for_version(version).boxed()
+            } else {
+                Just(None).boxed()
+            },
```
daira, line 1308:
```suggestion
            ironwood_bundle in orchard::arb_ironwood_bundle_for_version(version),
```

zcash_primitives/src/transaction/txid.rs
```diff
@@ -47,6 +51,50 @@ const ZCASH_SAPLING_OUTPUTS_NONCOMPACT_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxId
...
+
+fn sapling_auth_includes_anchor(version: TxVersion) -> bool {
+    match version {
+        TxVersion::Sprout(_) | TxVersion::V3 | TxVersion::V4 | TxVersion::V5 => false,
+        TxVersion::V6 => true,
+    }
+}
+
+/// Selects the `(value pool, orchard tx version)` pair that reproduces the
+/// pre-bump `BundleCommitmentDomain` for a given transaction version. The value
```
daira, line 82: Pre-bump?

zcash_primitives/src/transaction/txid.rs
```diff
@@ -94,22 +142,32 @@ pub(crate) fn transparent_outputs_hash<T: Borrow<TxOut>>(vout: &[T]) -> Blake2bH
 ///
 /// Write disjoint parts of each Sapling shielded spend to a pair of hashes:
 /// * \[nullifier*\] - personalized with ZCASH_SAPLING_SPENDS_COMPACT_HASH_PERSONALIZATION
-/// * \[(cv, anchor, rk, zkproof)*\] - personalized with ZCASH_SAPLING_SPENDS_NONCOMPACT_HASH_PERSONALIZATION
+/// * \[(cv, anchor, rk)*\] for pre-v6 transactions, personalized with
```
daira: The previous comment was correct for v4 but incorrect for v5.
```suggestion
/// * \[(cv, anchor, rk, zkproof)*\] for v4 transactions, personalized with
///   ZCASH_SAPLING_SPENDS_NONCOMPACT_HASH_PERSONALIZATION
/// * \[(cv, anchor, rk)*\] for v5 transactions, personalized with
```

zcash_primitives/src/transaction/mod.rs
nuttycom: Yeah, that was a conscious choice; we tried it the other way and that type parameter ends up *everywhere* and Rust is a pain about the polymorphic call sites where you'd actually want to be indifferent to what that parameter is instantiated to.

zcash_primitives/src/transaction/txid.rs
daira, line 145:
```suggestion
/// * \[(cv, anchor, rk)*\] for v5 transactions (v4 is not handled here), personalized with
```

daira (approved):
utACK with non-blocking comments. (There's one suggestion to fix a misleading consensus-relevant code comment.)

zcash_primitives/src/transaction/txid.rs
```diff
@@ -207,11 +265,12 @@ pub(crate) fn hash_transparent_txid_data(
 
 /// Implements [ZIP 244 section T.3](https://zips.z.cash/zip-0244#t-3-sapling-digest)
 fn hash_sapling_txid_data<A: sapling::bundle::Authorization>(
+    version: TxVersion,
```
daira, line 268: The doc header just above (line 266, `Implements [ZIP 244 section T.3]`) is now only accurate for v5. This function takes `version` and, for v6, produces a different Sapling digest -- v6 spend personalization `ZTxIdSSpendNH_v6`, and the spend anchor moved out of the spends digest into the auth digest -- which is specified in **ZIP 229**, not ZIP 244 T.3. Suggest noting the v6 variant, e.g. "...; for v6 the Sapling digest follows ZIP 229." (Same family as the v4/v5 field-list fix on `hash_sapling_spends`.)

Posted with the assistance of Claude Opus 4.8.

zcash_primitives/src/transaction/txid.rs
```diff
@@ -94,22 +142,32 @@ pub(crate) fn transparent_outputs_hash<T: Borrow<TxOut>>(vout: &[T]) -> Blake2bH
...
 /// * \[nullifier*\] - personalized with ZCASH_SAPLING_SPENDS_COMPACT_HASH_PERSONALIZATION
-/// * \[(cv, anchor, rk, zkproof)*\] - personalized with ZCASH_SAPLING_SPENDS_NONCOMPACT_HASH_PERSONALIZATION
+/// * \[(cv, anchor, rk)*\] for pre-v6 transactions, personalized with
+///   ZCASH_SAPLING_SPENDS_NONCOMPACT_HASH_PERSONALIZATION
+/// * \[(cv, rk)*\] for v6 transactions, personalized with
+///   ZCASH_SAPLING_SPENDS_V6_NONCOMPACT_HASH_PERSONALIZATION
 ///
 /// Then, hash these together personalized by ZCASH_SAPLING_SPENDS_HASH_PERSONALIZATION
 pub(crate) fn hash_sapling_spends<A: sapling::bundle::Authorization>(
+    version: TxVersion,
```
daira, line 152: Follow-on to the v4/v5 field-list fix above: the header `Implements [ZIP 244 section T.3a]` (line 141) is v5-specific, but this function now also computes the v6 Sapling-spends digest (`ZTxIdSSpendNH_v6`, anchor omitted), which is specified in **ZIP 229**. The bullet list already distinguishes v5/v6; the header line could add "; the v6 variant is specified in ZIP 229."

Posted with the assistance of Claude Opus 4.8.

zcash_primitives/src/transaction/txid.rs
```diff
@@ -358,10 +506,11 @@ pub struct BlockTxCommitmentDigester;
 impl TransactionDigest<Authorized> for BlockTxCommitmentDigester {
     /// We use the header digest to pass the transaction ID into
     /// where it needs to be used for personalization string construction.
-    type HeaderDigest = BranchId;
+    type HeaderDigest = (TxVersion, BranchId);
```
daira, line 509: Adjacent doc nit: the comment just above (line 507) says the header digest is used "to pass the transaction ID into where it needs to be used for personalization". Neither the old `BranchId` nor the new `(TxVersion, BranchId)` is a transaction ID -- the auth-commitment personalization is `ZTxAuthHash_` followed by the consensus branch ID. Pre-existing wording, but the tuple change makes "the transaction ID" read even less accurately. Could reword to "...to pass the transaction version and consensus branch ID...". Non-blocking.

Posted with the assistance of Claude Opus 4.8.

pczt/src/roles/tx_extractor/orchard.rs
```diff
@@ -7,25 +11,43 @@ pub(super) fn verify_bundle(
...
     sighash: [u8; 32],
 ) -> Result<(), OrchardError> {
-    let mut validator = orchard::bundle::BatchValidator::new();
-    let rng = OsRng;
-
-    validator.add_bundle(bundle, sighash);
+    if let Some(vk) = orchard_vk {
+        verify_bundle_with_key(bundle, vk, sighash)
+    } else {
+        // PCZT extraction produces new Orchard bundles using the NU6.2 fixed
```
daira, line 17: This is a bug. `FixedPostNu6_2` is specific to the NU6.2 circuit, while the sibling `verify_ironwood_bundle` below uses `PostNu6_3`. Per ZIP 258, the NU6.3 Action verifying key (selected by block height) applies to Orchard-*pool* Actions in v5 transactions too -- so a post-NU6.3 Orchard-pool bundle would not verify under `FixedPostNu6_2`.

Posted with the assistance of Claude Opus 4.8.

PR #2511, opened 2026-07-01
Disable the stale zip_0233 test on release/nu6.3_testnet_support

daira:
`transaction::tests::zip_0233` fails on the **Test NU7** CI steps on this branch: the NU6.3 v6 sighash/txid changes moved the computed v6 digest, but the (unchanged) ZIP 233 test vectors still expect the old values.

This marks the test `#[ignore]` with a TODO to regenerate the vectors (or re-enable) when the v7 transaction format and its ZIP 233 changes are settled. Test-only — it still compiles (the `v6_signature_hash` signature and the vector file are unchanged), so `#[ignore]` is the right lever.

🤖 Prepared with Claude Opus 4.8


nuttycom (approved):
ACK, doc-only

PR #2513, opened 2026-07-01
zcash_client_backend: Add per-pool input/output counts to proposal Step

dannywillems:
This code is extracted from https://github.com/zcash/zallet/blob/main/zallet/src/components/json_rpc/methods/z_send_many.rs#L299-L340. It will be reused there later.
This was initially an idea from the work on https://github.com/zcash/zallet/pull/531.

------

Add counting methods on `Step` that parallel the existing `input_in_pool`/`output_in_pool`/`change_in_pool` predicates: `input_count_in_pool`, `output_count_in_pool`, `change_count_in_pool`, and `orchard_action_count`. The latter returns the number of Orchard actions a step requires (the greater of its Orchard spends and its Orchard outputs plus change), so consumers such as zallet can enforce an Orchard action limit without reimplementing the traversal.j

### DIFF
--- a/zcash_client_backend/src/proposal.rs
+++ b/zcash_client_backend/src/proposal.rs
@@ -602,6 +602,59 @@ impl<NoteRef> Step<NoteRef> {
             .iter()
             .any(|c| c.output_pool() == pool_type)
     }
+
+    /// Returns the number of inputs to this step that belong to the given pool.
+    ///
+    /// For a shielded pool this is the number of spent notes of that protocol; for
+    /// [`PoolType::Transparent`] it is the number of transparent inputs.
+    pub fn input_count_in_pool(&self, pool_type: PoolType) -> usize {
+        match pool_type {
+            PoolType::Transparent => self.transparent_inputs().len(),
+            PoolType::SAPLING => self
+                .shielded_inputs()
+                .iter()
+                .flat_map(|s_in| s_in.notes())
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/proposal.rs
```diff
@@ -592,16 +597,73 @@ impl<NoteRef> Step<NoteRef> {
...
+            .filter(|c| c.output_pool() == pool_type)
+            .count()
+    }
+
+    /// Returns the number of Orchard actions required by this step.
+    ///
+    /// Each Orchard action can carry both a spend and an output, so the number of actions is
+    /// the greater of the number of Orchard note spends and the number of Orchard outputs
+    /// (payments plus change) in this step.
+    pub fn orchard_action_count(&self) -> usize {
```
nuttycom, line 661: Note that this will change after Ironwood activation; the orchard transaction builder will create `spends + outputs` actions for Orchard, as the cross-address restriction requires any change output's corresponding dummy input to be signed with the spending key corresponding to the internal IVK (to which the change output is sent).

nuttycom (approved):
utACK, modulo that this will need to change for NU6.3 support

zcash_client_backend/src/proposal.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/2515

PR #2516, opened 2026-07-01
zcash_primitives: Address post-merge review comments from PR #2509

nuttycom:
Addresses @daira's post-merge review comments on #2509:

- `zcash_primitives::transaction::components::orchard::read_v5_bundle` now takes the
  consensus branch ID (as discussed in the `read_v5_bundle` rationale thread), deriving
  the Orchard bundle version via a new public helper `orchard_bundle_version_for_branch`.
  The helper is an exhaustive match, replacing both the wildcard-arm private helper in
  `builder.rs` and the inline mapping in `Transaction::read_v5`, and the incomplete
  "Since this is public API." sentence is gone.
- `TxVersion::has_ironwood` is now an exhaustive match.
- The three `arb_txdata` proptest strategies use `arb_ironwood_bundle_for_version`
  directly (it already yields `None` for versions without Ironwood support).
- Factored out `Builder::orchard_in_use`, mirroring `ironwood_in_use`.
- Applied the suggested comment/doc corrections (coinbase Orchard-builder flag comment,
  `with_expiry_height` warning, `txid.rs` "pre-bump" wording and Sapling spends digest
  doc).
- The nu7 builder test network now activates NU6.3 at height 10 and NU7 at height 11,
  per the suggestions.
- Added a unified `bundle_version_for_branch(branch_id, pool)`, returning the
  `BundleVersion` in effect for an Orchard-protocol value pool under a consensus
  branch, or `None` when the pool is unsupported under that branch (Orchard prior to
  NU5; Ironwood prior to NU6.3). `read_v5_bundle` now takes the consensus branch ID,
  and `read_v6_bundle` the branch ID plus the `orchard::ValuePool` selecting the slot,
  in place of caller-supplied `BundleVersion`s — callers can no longer select a
  wrong-slot bundle version, and a non-empty bundle belonging to an unsupported pool is
  rejected as invalid data. In `Builder::new` the `None` result subsumes the previous
  explicit NU5-activation check.
- Removed `zcash_client_backend::ANY_ORCHARD_BUNDLE_VERSION`: every fee/change call
  site has the proposal's target height available and now derives the Orchard bundle
  version from it, so proposals targeting post-NU6.3 heights count one action per
  Orchard spend or output, matching the transaction builder. (`EmptyBundleView` keeps
  a fixed version, which is sound because an empty bundle yields zero actions under
  every version's policy.) This also answers "does this need to be public?": yes —
  `zcash_client_backend` now consumes `bundle_version_for_branch`.
- Added `note_commitment_and_nullifier_depend_on_note_version`, checking that lead-byte
  0x02 and 0x03 notes with otherwise identical fields produce distinct commitments and
  nullifiers.

- **pczt**: fixed extraction and verification of post-NU6.3 Orchard-pool bundles
  (reported by @daira on #2509). `extract_tx_data` now parses the Orchard-pool bundle
  using the bundle version implied by the PCZT's consensus branch ID, and the
  Transaction Extractor verifies each Orchard-protocol bundle using the circuit
  version implied by the bundle's own version instead of hardcoding the NU6.2 fixed
  circuit; `verify_bundle`/`verify_ironwood_bundle` collapse into one function.
- The branch-to-Orchard-revision consensus knowledge now lives in one place:
  `zcash_protocol::consensus::BranchId::orchard_protocol_revision`, returning the new
  `OrchardProtocolRevision` enum. `zcash_primitives::…::bundle_version_for_branch` and
  `pczt` both derive `BundleVersion`s from it via trivial conversions (pczt cannot
  reuse the `zcash_primitives` helper directly because that dependency is optional).
  Building on this, **all** remaining PCZT roles (Updater, Verifier, Prover, low-level
  Signer) are now branch-aware — the Prover in particular now proves against the
  correct circuit for post-NU6.3 Orchard-pool bundles — and the legacy-pinned
  `into_orchard_parsed` is removed. This completes the first section of #2519.
- Applied the ZIP 244 / ZIP 229 attribution fixes to the Sapling digest documentation
  in `txid.rs`, and corrected the `BlockTxCommitmentDigester::HeaderDigest` doc.
- Added a ZIP 317 change-calculation test covering the post-NU6.3 Orchard
  action-count policy (partially addressing the `fees/zip317.rs` coverage comment;
  the broader Ironwood coverage gaps are tracked in #2518).

Comments deliberately not addressed here (all marked non-blocking):

- `add_ironwood_output` succeeding after `propose_version(V5)` ("arguably this call
  should fail") — add-time version validation is a behavioral design change, deferred.
- The `zip233Amount`-in-v6 characterization and NU7 tx-version boilerplate question —
  per the comment, not necessary to fix now.
- The type-safety suggestion for distinguishing Orchard-pool from Ironwood-pool bundle
  parameters — needs a design discussion.

### DIFF
--- a/components/zcash_protocol/src/consensus.rs
+++ b/components/zcash_protocol/src/consensus.rs
@@ -900,6 +900,44 @@ impl BranchId {
             BranchId::Nu7 => true,
         }
     }
+
+    /// Returns the revision of the Orchard protocol in effect under this consensus
+    /// branch, or `None` for branches that predate NU5 (under which the Orchard
+    /// protocol is not supported).
+    pub fn orchard_protocol_revision(&self) -> Option<OrchardProtocolRevision> {
+        use BranchId::*;
+        match self {
+            Sprout | Overwinter | Sapling | Blossom | Heartwood | Canopy => None,
+            Nu5 | Nu6 | Nu6_1 => Some(OrchardProtocolRevision::InsecureV1),
+            Nu6_2 => Some(OrchardProtocolRevision::V2),
+            Nu6_3 => Some(OrchardProtocolRevision::V3),
+            #[cfg(zcash_unstable = "nu7")]
+            Nu7 => Some(OrchardProtocolRevision::V3),
+        }
+    }
+}
+
+/// The revision of the Orchard protocol deployed by a network upgrade.
+///
+/// The revisions correspond one-to-one to the `orchard` crate's `ProtocolVersion`;
+/// this type exists so that crates that do not depend on `orchard` can express which
+/// protocol revision a consensus branch selects. Use
+/// [`BranchId::orchard_protocol_revision`] to obtain the revision in effect under a
+/// given consensus branch.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum OrchardProtocolRevision {
+    /// The original revision of the Orchard protocol, deployed at NU5 and used prior
+    /// to NU6.2. Uses the historical unsound Orchard circuit; cross-address transfers
+    /// are permitted.
+    InsecureV1,
+    /// The revision of the Orchard protocol deployed at NU6.2. Uses the post-NU6.2
+    /// fixed circuit; cross-address transfers are permitted.
+    V2,
+    /// The revision of the Orchard protocol deployed at NU6.3, which introduces the
+    /// Ironwood value pool. Uses the post-NU6.3 circuit; cross-address transfers are
+    /// prohibited for the Orchard value pool and permitted for the Ironwood value
+    /// pool.
+    V3,
 }
 
 #[cfg(any(test, feature = "test-dependencies"))]
@@ -985,6 +1023,34 @@ mod tests {
     fn branch_id_from_u32() {
         assert_eq!(BranchId::try_from(0), Ok(BranchId::Sprout));
         assert!(BranchId::try_from(1).is_err());
+    }
+
+    #[test]
+    fn orchard_protocol_revision() {
+        use super::OrchardProtocolRevision;
+
+        assert_eq!(BranchId::Canopy.orchard_protocol_revision(), None);
+        assert_eq!(
+            BranchId::Nu5.orchard_protocol_revision(),
+            Some(OrchardProtocolRevision::InsecureV1)
+        );
+        assert_eq!(
+            BranchId::Nu6_1.orchard_protocol_revision(),
+            Some(OrchardProtocolRevision::InsecureV1)
+        );
+        assert_eq!(
+            BranchId::Nu6_2.orchard_protocol_revision(),
+            Some(OrchardProtocolRevision::V2)
+        );
+        assert_eq!(
+            BranchId::Nu6_3.orchard_protocol_revision(),
+            Some(OrchardProtocolRevision::V3)
+        );
+        #[cfg(zcash_unstable = "nu7")]
+        assert_eq!(
+            BranchId::Nu7.orchard_protocol_revision(),
+            Some(OrchardProtocolRevision::V3)
+        );
     }
 
     #[test]
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -13,6 +13,14 @@ workspace.
 ### Changed
 - Migrated to `zcash_protocol 0.10.0-pre.0`, `zcash_transparent 0.9.0-pre.0`,
   `zcash_primitives 0.29.0-pre.0`, `zcash_proofs 0.29.0-pre.0`.
+- Every PCZT role that parses the Orchard-pool bundle derives its bundle
+  version from the PCZT's consensus branch ID, and returns an
+  `UnsupportedConsensusBranchId` error when that branch ID is unrecognized or
+  predates NU5.
+- `pczt::roles::low_level_signer::Signer::sign_orchard_with` now bounds its
+  error parameter by `From<pczt::roles::low_level_signer::OrchardParseError>`
+  instead of `From<orchard::pczt::ParseError>` (as does the new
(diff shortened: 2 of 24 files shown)


zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -91,27 +91,50 @@ fn read_bundle<R: Read>(
...
+///   * NU6.2: fixed circuit, cross-address enabled;
+///   * NU6.3 onward: post-NU6.3 circuit, cross-address disabled (consensus-mandated).
+pub fn orchard_bundle_version_for_branch(consensus_branch_id: BranchId) -> BundleVersion {
+    match consensus_branch_id {
+        BranchId::Sprout
+        | BranchId::Overwinter
+        | BranchId::Sapling
+        | BranchId::Blossom
+        | BranchId::Heartwood
+        | BranchId::Canopy
```
nuttycom: Orchard is not supported prior to NU5; should this return an error?

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -91,27 +91,50 @@ fn read_bundle<R: Read>(
...
-/// public API. By contrast, `read_v6_bundle` takes a `bundle_version` argument so
-/// the caller selects the Orchard or Ironwood v6 slot.
+/// The v5 Orchard flag byte is the same in every epoch (bit 2 reserved), so serialization
+/// is unaffected by this choice, but the `BundleVersion` fixes the bundle's cross-address
+/// semantics and circuit generation, which do follow the consensus epoch:
+///   * pre-NU6.2: historical insecure circuit, cross-address enabled, proof size not
+///     enforced;
+///   * NU6.2: fixed circuit, cross-address enabled;
+///   * NU6.3 onward: post-NU6.3 circuit, cross-address disabled (consensus-mandated).
+pub fn orchard_bundle_version_for_branch(consensus_branch_id: BranchId) -> BundleVersion {
```
nuttycom: Does this need to be public?

zcash_client_backend/src/lib.rs
```diff
@@ -87,6 +87,14 @@ pub use decrypt::{DecryptedOutput, TransferType, decrypt_transaction};
...
+///
+/// NOTE: as a consequence, fee and change calculation undercount Orchard actions for
+/// post-NU6.3 target heights (where the Orchard pool disables cross-address transfers, so
+/// every spend and output occupies its own action). Proposals that spend Orchard-pool notes
+/// at or beyond NU6.3 activation therefore cannot yet be created: the transaction builder
+/// revalidates the fee against the branch-derived bundle version and rejects them
+/// (fail-closed; an underpaying transaction cannot result). These call sites must be made
+/// branch-aware before the next `zcash_client_backend` release.
 #[cfg(feature = "orchard")]
 pub(crate) const ANY_ORCHARD_BUNDLE_VERSION: ::orchard::bundle::BundleVersion =
```
nuttycom: Let's look at options for removing this constant as well; we should determine whether the call sites where it's used can correctly determine the bundle version to use from the available context.

pczt/src/orchard.rs
```diff
@@ -749,8 +749,43 @@ impl Bundle {
     }
 }
 
+/// Returns the Orchard bundle version used by the Orchard pool of the given
+/// (v5-or-later) network upgrade.
+#[cfg(feature = "orchard")]
+pub(crate) fn orchard_bundle_version_for_branch(
```
nuttycom: Reuse zcash_primitives::transaction::components::orchard::bundle_version_for_branch here

pczt/src/orchard.rs
```diff
@@ -749,8 +749,43 @@ impl Bundle {
...
 #[cfg(feature = "orchard")]
 impl Bundle {
+    /// Parses the Orchard-pool bundle using the legacy (pre-NU6.3) Orchard bundle
+    /// version.
+    ///
+    /// TODO: this is incorrect for PCZTs whose consensus branch ID is NU6.3 or later
+    /// (under which the Orchard pool mandates the cross-address restriction); callers
+    /// should instead derive the bundle version from the PCZT's consensus branch ID via
+    /// [`orchard_bundle_version_for_branch`] and use [`Self::into_parsed_with_version`],
+    /// as `Pczt::extract_tx_data` does.
```
nuttycom: What inhibits us from fixing this right now?

pczt/CHANGELOG.md
```diff
@@ -10,6 +10,26 @@ workspace.
 
 ## [0.8.0] - PLANNED
 
+### Fixed
```
nuttycom: Nit: this "fixed" section is not fixing any bug relative to the 0.7.0 release, so it does not need to be included.

components/zcash_protocol/src/consensus.rs
```diff
@@ -900,6 +900,44 @@ impl BranchId {
...
+            Nu7 => Some(OrchardProtocolRevision::V3),
+        }
+    }
+}
+
+/// The revision of the Orchard protocol deployed by a network upgrade.
+///
+/// The revisions correspond one-to-one to the `orchard` crate's `ProtocolVersion`;
+/// this type exists so that crates that do not depend on `orchard` can express which
+/// protocol revision a consensus branch selects. Use
```
daira, line 924: (Non-blocking) The duplication is mildly unfortunate. Could `orchard` depend on `zcash_protocol` and `pub use` this so that it's the same type?

nuttycom: No; that would create a circular dependency between `orchard` and the `librustzcash` repository and when we had that in the past with Sapling it was a maintenance nightmare. This was the least-duplicative solution I could come up with.

nuttycom (commented):
self-utACK 802d33a9013d017832e33b0dfe3bc94db7705d3c

daira (approved):
❤️ utACK

PR #2517, opened 2026-07-01
zcash_client_sqlite: Upgrade imported transparent addresses to their derived form on gap-limit derivation

nuttycom:
## Summary

A transparent address can exist in the `addresses` table both as a standalone (`Foreign`)
import — e.g. a watch-only pubkey/script imported into an account — and as an address the
account itself derives. When gap-limit generation derived such an address,
`store_address_range` inserted a *second* row for the same transparent receiver instead of
reconciling with the existing import. On databases carrying the new `UNIQUE` index on
`addresses.cached_transparent_receiver_address` (#2486) this makes the derive fail outright
with a constraint violation; without it, it silently produced a duplicate record for the
receiver, violating the intended one-record-per-receiver invariant.

This PR:

- **`store_address_range` upgrades in place.** When the address being derived already exists as
  a `Foreign`-scope row for the receiver in the same account, the existing row is upgraded to
  its derived form (`key_scope`, `transparent_child_index`, `diversifier_index_be`,
  `receiver_flags`; the `imported_transparent_receiver_*` columns cleared) instead of inserting
  a duplicate. The row `id` is preserved, so any UTXOs, exposure, and spend-search state already
  attached to the imported receiver carry over. The lookup that decides whether to upgrade reads
  only columns present at every schema version at which `store_address_range` runs, and the
  imported-column `UPDATE` is prepared only when a `Foreign` row actually exists — so this
  function remains usable from the `transparent_gap_limit_handling` migration, which runs before
  those columns are added.

- **Migration exposure carry-over.** The `add_transparent_receiver_address_index` migration
  (#2486) already collapses a pre-existing derived/imported duplicate onto a canonical record;
  it now also carries the earliest `exposed_at_height` among the merged records onto the
  survivor.

- **Cross-account duplicates are resolved by derivation** (review follow-up). Previously a
  receiver duplicated across more than one account was an unconditional migration abort. Now,
  if exactly one of the records reproduces the receiver when its address is derived from its
  own account's viewing key at its recorded child index, that record is definitively correct:
  it is retained, and the received outputs of the other records are repointed to it along with
  their account attribution (the reason cross-account merges were previously unsafe). No
  account is privileged in this determination — an address genuinely derived under the ZIP 32
  account index `0x7FFFFFFF` (used for the `zcashd` legacy account) wins over another account's
  imported record just as any other verified derivation would. Cross-account duplicates for
  which no unique record verifies still abort with `CorruptedData`.

CHANGELOG updated.

## Test plan

- New `store_address_range_upgrades_imported_receiver`: importing a standalone address that is
  subsequently derived upgrades the existing row in place (same `id`, derived scope, import
  columns cleared, no duplicate), and a UTXO attached to it carries over.
- New `repair_carries_min_exposed_at_height`: the migration keeps the earliest exposure height
  when merging a duplicate.
- New `resolves_cross_account_duplicate_by_derivation`: a receiver duplicated across two
  accounts, where one record is genuinely derivable from its account's viewing key, is resolved
  in favor of the verified record; the attached output's `account_id` moves with it.
- New `cross_account_winner_may_be_legacy_account`: the mirror case — the verified derivation
  is under account index `0x7FFFFFFF`, and that record wins over another account's import,
  demonstrating that resolution is not keyed on account identity.
- Pre-existing `rejects_cross_account_duplicates` still passes: a cross-account duplicate in
  which neither record verifies by derivation continues to abort.
- Full `zcash_client_sqlite` suite passes under `--features transparent-key-import,orchard`
  (268 passed, 0 failed); the crate also builds without the `transparent-inputs` feature
  (cross-account resolution requires the key APIs and remains fail-closed there).

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
@@ -104,6 +104,21 @@ impl RusqliteMigration for Migration {
                 )));
             }
 
+            // Carry the earliest observed exposure across the merged records onto the canonical
+            // one, so that upgrading an imported receiver to its derived form does not discard an
+            // earlier exposure height recorded against the imported record. Run this before
+            // deleting the redundant rows, while the whole group is still present.
+            transaction.execute(
+                "UPDATE addresses
+                 SET exposed_at_height = (
+                     SELECT MIN(exposed_at_height)
+                     FROM addresses
+                     WHERE cached_transparent_receiver_address = :addr
+                 )
+                 WHERE id = :canonical",
+                named_params! { ":addr": addr, ":canonical": canonical_id },
+            )?;
+
             for (dup_id, _) in group.iter().skip(1) {
                 for table in [
                     "transparent_received_outputs",
@@ -349,6 +364,67 @@ mod tests {
             )
             .unwrap();
         assert_eq!(imported_count, 0);
+    }
+
+    /// When repairing a duplicate, the surviving canonical record keeps the earliest
+    /// `exposed_at_height` observed across the merged records.
+    #[test]
+    fn repair_carries_min_exposed_at_height() {
+        let network = Network::TestNetwork;
+        let data_file = NamedTempFile::new().unwrap();
+        let mut db_data =
+            WalletDb::for_path(data_file.path(), network, test_clock(), test_rng()).unwrap();
+
+        WalletMigrator::new()
+            .with_seed(Secret::new(vec![0xab; 32]))
+            .ignore_seed_relevance()
+            .init_or_migrate_to(&mut db_data, DEPENDENCIES)
+            .unwrap();
+
+        let account_id = insert_account(&db_data.conn, 0, [0xAA; 16]);
+
+        // Derived (canonical) record exposed at height 200.
+        db_data
+            .conn
+            .execute(
+                "INSERT INTO addresses (account_id, key_scope, diversifier_index_be, address,
+                 transparent_child_index, cached_transparent_receiver_address, receiver_flags,
+                 exposed_at_height)
+                 VALUES (?1, 0, X'00000000000000000000000000', 'addr_derived', 0, 't_exp', 5, 200)",
+                [account_id],
+            )
+            .unwrap();
+        let derived_id = db_data.conn.last_insert_rowid();
+
+        // Imported record exposed earlier, at height 100.
+        db_data
+            .conn
+            .execute(
+                "INSERT INTO addresses (account_id, key_scope, address,
+                 cached_transparent_receiver_address, imported_transparent_receiver_pubkey,
+                 receiver_flags, exposed_at_height)
+                 VALUES (?1, -1, 'addr_imported', 't_exp',
+                 X'020000000000000000000000000000000000000000000000000000000000000001', 5, 100)",
+                [account_id],
+            )
+            .unwrap();
+
+        WalletMigrator::new()
+            .with_seed(Secret::new(vec![0xab; 32]))
+            .ignore_seed_relevance()
+            .init_or_migrate_to(&mut db_data, &[MIGRATION_ID])
+            .unwrap();
+
+        // The canonical (derived) record now carries the earliest exposure.
+        let exposed: Option<i64> = db_data
+            .conn
+            .query_row(
+                "SELECT exposed_at_height FROM addresses WHERE id = ?1",
+                [derived_id],
+                |r| r.get(0),
+            )
+            .unwrap();
+        assert_eq!(exposed, Some(100));
     }
 
     /// A transparent receiver duplicated across more than one account cannot be safely merged, so
--- a/zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
@@ -12,15 +12,19 @@
 //! Any pre-existing duplicates (for example, the same address both HD-derived and standalone
 //! imported) are resolved by the migration: a single canonical record is retained, the
 //! received-output foreign keys of the others are repointed to it, and the redundant records are
-//! deleted. A receiver duplicated across more than one account cannot be safely merged and aborts
-//! the migration. `NULL` values do not participate in SQLite's `UNIQUE` semantics, so rows without
-//! a transparent receiver are unaffected.
+//! deleted. A receiver duplicated across more than one account is resolved by derivation: if
+//! exactly one of the records reproduces the receiver when its address is derived from its own
+//! account's viewing key at its recorded child index, that record is definitively correct and is
+//! retained (with the received outputs of the others reattributed to it); otherwise the conflict
+//! cannot be resolved automatically and the migration aborts. `NULL` values do not participate in
+//! SQLite's `UNIQUE` semantics, so rows without a transparent receiver are unaffected.
 
 use std::collections::HashSet;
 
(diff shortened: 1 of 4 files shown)


zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -104,6 +104,21 @@ impl RusqliteMigration for Migration {
                 )));
             }
```
nuttycom: This should not fail in the specific case that one of the accounts is the "legacy" account - in the case that one record has account ID 0x7FFFFFFF and the other has some other nonnegative index, the record with the nonnegative index should win and the legacy account record should be deleted.

nuttycom: Addressed in c039d084d1, generalized per the follow-up discussion: rather than privileging any account ID, cross-account duplicates are now resolved purely by derivation — the unique record (if any) that reproduces the receiver when derived from its own account's viewing key at its recorded child index wins, and the received outputs of the other records are repointed to it along with their account attribution. Since an address may genuinely have been derived under the 0x7FFFFFFF account index, that case wins symmetrically (covered by the new `cross_account_winner_may_be_legacy_account` test). Cross-account duplicates for which no unique record verifies still abort.

_Comment posted with Claude Code assistance._

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -743,15 +762,55 @@ pub(crate) fn store_address_range<P: consensus::Parameters>(
...
+        let foreign_id: Option<i64> = stmt_lookup_foreign
+            .query_row(
+                named_params![
+                    ":account_id": account_id.0,
+                    ":transparent_address": transparent_address_enc,
+                    ":foreign_scope": KeyScope::Foreign.encode(),
+                ],
+                |row| row.get(0),
+            )
+            .optional()?;
```
nuttycom: This doesn't resolve what happens if the address has been imported under a different account id.

zcash-harry: How about we drop the `account_id = :account_id` filter from `stmt_lookup_foreign` (or add a fallback query without it when the first finds nothing), so it can find a `Foreign` row for this receiver under any account.
If the match's `account_id` differs from the account currently deriving, the upgrade `UPDATE` needs to also set `account_id` = `:account_id `, moving the row to the account that actually owns it and the received-output tables need the same `account_id` cascade that `add_transparent_receiver_address_index::up` already does for the migration's cross-account case, so funds already attached to the old account move with it rather than getting orphaned.

Other than that LGTM

zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -33,9 +37,121 @@ pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x3d4f12d6_3da9_4ace_ac65_
...
+        encoding::AddressCodec as _,
+        keys::{UnifiedFullViewingKey, UnifiedIncomingViewingKey},
+    };
+
+    let Some(idx) = child_index.and_then(NonHardenedChildIndex::from_index) else {
+        return false;
+    };
 
-impl schemerz::Migration<Uuid> for Migration {
+    let derived = match key_scope {
```
nuttycom: nit: use `KeyScope::decode`

zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -33,9 +37,121 @@ pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x3d4f12d6_3da9_4ace_ac65_
...
+            }),
+        // Internal scope: derivable only from the UFVK's account pubkey.
+        1 => ufvk
+            .and_then(|s| UnifiedFullViewingKey::decode(params, s).ok())
+            .and_then(|k| {
+                k.transparent()
+                    .and_then(|apk| apk.derive_internal_ivk().ok())
+            })
+            .and_then(|ivk| ivk.derive_address(idx).ok()),
+        // Foreign and ephemeral records cannot be verified by scope derivation.
```
nuttycom: Incorrect; foreign key scope is `2` - see `KeyScope::decode` comment above.

zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -74,47 +191,75 @@ impl RusqliteMigration for Migration {
...
+                 JOIN accounts ON accounts.id = a.account_id
+                 WHERE a.cached_transparent_receiver_address = :addr
+                 ORDER BY (a.transparent_child_index IS NULL), a.id",
             )?;
-            let group = group_stmt
+            let group: Vec<AddressRecord> = group_stmt
                 .query_map(named_params! { ":addr": addr }, |row| {
-                    Ok((row.get::<_, i64>(0)?, row.get::<_, i64>(1)?))
+                    Ok((
+                        row.get(0)?,
```
nuttycom, line 222: At this point, this is complex enough we should use a bespoke struct.

zcash_client_sqlite/src/wallet/init/migrations/add_transparent_receiver_address_index.rs
```diff
@@ -186,6 +364,325 @@ mod tests {
...
+            .conn
+            .prepare("SELECT id FROM addresses WHERE cached_transparent_receiver_address = ?1")
+            .unwrap()
+            .query_map([&taddr], |row| row.get(0))
+            .unwrap()
+            .collect::<Result<_, _>>()
+            .unwrap();
+        assert_eq!(remaining, vec![ephemeral_id]);
+    }
+
```
zcash-harry, line 685: Unlike `resolves_cross_account_duplicate_by_derivation`, this test doesn't attach a UTXO to the losing (imported) record, so it never confirms the `account_id`/`address_id` repoint actually happens for the ephemeral path, only that the right row survives. The repoint code is scope-agnostic so it should work, but worth mirroring the External-scope test's UTXO + `assert_eq`!((`address_id`, `account_id`), ...) check here too, since that's the part that actually matters for not losing/misattributing funds.

_Comment assisted by Claude_

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -743,15 +768,77 @@ pub(crate) fn store_address_range<P: consensus::Parameters>(
...
+
+        let foreign_row: Option<(i64, i64)> = stmt_lookup_foreign
+            .query_row(
+                named_params![
+                    ":transparent_address": transparent_address_enc,
+                    ":foreign_scope": KeyScope::Foreign.encode(),
+                ],
+                |row| Ok((row.get(0)?, row.get(1)?)),
+            )
+            .optional()?;
```
zcash-harry, line 786: This still assumes at most one Foreign row matches per receiver. Post-migration that's guaranteed by the unique index but the function doc above notes it's also called from `transparent_gap_limit_handling`, which runs before that index exists, so in that window a pre-existing duplicate Foreign row (same account or cross-account) would hit `QueryReturnedMoreThanOneRow` here instead of failing gracefully. 

Worth a defensive LIMIT 1 / take-first here anyway, the "at most one match" guarantee comes from `import_standalone_transparent_pubkey`/`_script` elsewhere, not anything local to this function, so it's not obvious at the call site why it holds.

_Comment assisted by Claude_

Cosmos-Harry (approved):

PR #2520, opened 2026-07-02
pczt: Complete branch-aware v6/Ironwood support

### LINKED ISSUE #2519: pczt: complete branch-aware Orchard bundle handling and Ironwood role support
Follow-up to the review of #2509 and the partial fixes in #2516. The `pczt` crate
(0.8.0 PLANNED, unreleased) has remaining gaps in its NU6.3 support:

## Branch-aware Orchard-pool bundle parsing

#2516 made `Pczt::extract_tx_data` (and hence transaction extraction and
verification) parse the Orchard-pool bundle using the bundle version implied by the
PCZT's consensus branch ID. The remaining roles still parse via
`Bundle::into_orchard_parsed`, which pins the legacy (pre-NU6.3)
`BundleVersion::orchard_v2()` (see the TODO on that method):

- `roles/updater/orchard.rs`
- `roles/verifier/orchard.rs`
- `roles/prover/orchard.rs` (proofs would be created for the wrong circuit /
  flag semantics for post-NU6.3 Orchard-pool bundles)
- `roles/low_level_signer/mod.rs` (note: its error channel is generically bounded
  by `From<orchard::pczt::ParseError>`, so threading a branch-ID parse failure
  requires an API change)

## Ironwood support gaps

- `Creator::build_from_parts` silently discards `parts.ironwood`, so the
  builder→PCZT path loses Ironwood spends/outputs.
- `Creator::new` cannot produce a V6/Ironwood PCZT (hardcodes `V5_TX_VERSION`; no
  Ironwood anchor or flags); its Ironwood placeholder bundle shares the Orchard
  flags/anchor, which is incorrect for the Ironwood pool.
- `extract_tx_data` silently drops a non-empty Ironwood bundle for V5 PCZTs, and
  does not reject a V6 PCZT whose consensus branch ID predates NU6.3.
- The v1 serialization round-trip is lossy for the Ironwood bundle (flags, anchor,
  value_sum, zkproof, bsk are discarded when actions are empty), and the three
  "empty Ironwood bundle" representations (Creator placeholder, v1 decode fallback,
  v2 omitted-bundle fallback) disagree, which breaks `Bundle::merge` in Combiner
  workflows that mix v1-round-tripped and in-memory copies.
- The high-level `Signer` has no `sign_ironwood` method, and `Verifier`/`Updater`
  have no Ironwood equivalents of their Orchard methods.

nuttycom:
Stacked on #2516; resolves the remainder of the pczt gaps identified in the #2509
review. Closes #2519.

## Canonical empty bundle states

The Creator, `Creator::build_from_parts`, the v1/v2 serialization formats, and the IO
Finalizer previously produced four different representations of an "empty" Ironwood
bundle (and inconsistent empty Sapling/Orchard bundles), breaking `Bundle::merge` for
Combiner workflows that mix serialization-round-tripped and in-memory copies of a PCZT.
Each bundle type now has a single canonical empty representation used everywhere:

- The Creator's Ironwood placeholder no longer mirrors the Orchard bundle's flags and
  anchor (its flags are now the Ironwood defaults, with cross-address transfers enabled).
- `build_from_parts` uses the canonical empties (all-zeroes anchors rather than the
  empty-tree root) for absent bundles, and maps `parts.ironwood` through instead of
  discarding it — fixing the builder→PCZT path silently losing Ironwood spends/outputs.
- Empty bundle value sums are non-negative zero, matching re-serialized parsed bundles.
- The IO Finalizer no longer sets `bsk` on an empty Orchard-protocol bundle. (Sapling is
  still always finalized: its Transaction Extractor requires `bsk` even when empty.)
- The v2 omission decision is exact equality with the slot's canonical empty bundle, so
  non-default flags/note-version round-trip losslessly instead of being restored to
  defaults; v1 refuses to encode a PCZT whose Ironwood bundle is not canonically empty.

## Creator v6 support and extraction validation

- `Creator::new` derives the transaction version/version group ID from the consensus
  branch ID (v6 from NU6.3 onward); new `with_ironwood_anchor`/`with_ironwood_flags`
  configure the Ironwood bundle, rejecting use with non-v6 versions.
- Transaction extraction rejects v6 PCZTs whose branch ID predates NU6.3, and non-v6
  PCZTs carrying non-canonical Ironwood data, instead of silently dropping it or
  extracting a consensus-invalid transaction.
- The v1 format refuses to encode v6 PCZTs (parseable by 0.7.0 peers but never
  extractable).

## High-level Ironwood role methods

- `Signer::{sign_ironwood, apply_ironwood_signature}`, `Verifier::with_ironwood`, and
  `Updater::update_ironwood_with` — mechanical mirrors of the Orchard-pool methods, so
  hardware-wallet-style signers no longer need the low-level Signer to authorize
  Ironwood actions.

One deliberate leftover: the Signer's `empty_ironwood` passthrough is likely removable
now that empty-bundle parse/serialize round-trips are canonical, but that simplification
is left for a follow-up rather than folded in here.

### DIFF
--- a/pczt/src/lib.rs
+++ b/pczt/src/lib.rs
@@ -38,7 +38,7 @@ use {
     common::{Global, determine_lock_time},
     zcash_primitives::transaction::{Authorization, TransactionData, TxVersion},
     zcash_protocol::{
-        consensus::BranchId,
+        consensus::{BranchId, OrchardProtocolRevision},
         constants::{V5_TX_VERSION, V5_VERSION_GROUP_ID},
     },
 };
@@ -124,6 +124,12 @@ pub mod v1 {
         type Error = super::EncodingError;
 
         fn try_from(pczt: super::Pczt) -> Result<Self, Self::Error> {
+            // The v1 format predates the v6 transaction format; a parser of the v1
+            // encoding could parse a v6 PCZT but never extract a transaction from it.
+            if pczt.global.tx_version == zcash_protocol::constants::V6_TX_VERSION {
+                return Err(super::EncodingError::UnsupportedTxVersion);
+            }
+
             // The v1 format cannot represent an Ironwood bundle in any state other
             // than the canonical empty one; a parser of the v1 encoding will
             // reconstruct exactly that value.
@@ -149,6 +155,37 @@ pub mod v1 {
                 orchard: pczt.orchard.into(),
                 ironwood: orchard::EMPTY_IRONWOOD,
             }
+        }
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use zcash_protocol::consensus::BranchId;
+
+        use crate::roles::creator::Creator;
+
+        #[test]
+        fn v1_refuses_v6_pczts_and_non_canonical_ironwood_bundles() {
+            // A v6 PCZT cannot be encoded as v1, even when its Ironwood bundle is
+            // canonically empty.
+            let pczt = Creator::new(BranchId::Nu6_3.into(), 10_000_000, 133, [0; 32], [0; 32])
+                .unwrap()
+                .build();
+            assert!(matches!(
+                super::Pczt::try_from(pczt),
+                Err(crate::EncodingError::UnsupportedTxVersion)
+            ));
+
+            // A v5 PCZT carrying non-canonical Ironwood bundle data cannot be encoded
+            // as v1, because the data would be dropped.
+            let mut pczt = Creator::new(BranchId::Nu6.into(), 10_000_000, 133, [0; 32], [0; 32])
+                .unwrap()
+                .build();
+            pczt.ironwood.bsk = Some([1; 32]);
+            assert!(matches!(
+                super::Pczt::try_from(pczt),
+                Err(crate::EncodingError::UnsupportedTxVersion)
+            ));
         }
     }
 }
@@ -388,23 +425,11 @@ impl Pczt {
 
         let consensus_branch_id = BranchId::try_from(global.consensus_branch_id)
             .map_err(|_| ExtractError::UnknownConsensusBranchId)?;
-        let orchard_bundle_version = consensus_branch_id
+        let orchard_protocol_revision = consensus_branch_id
             .orchard_protocol_revision()
-            .map(crate::orchard::orchard_bundle_version_for_revision)
             // The v5 and v6 transaction formats do not exist prior to NU5, so no
             // transaction could be extracted under such a branch in any case.
             .ok_or(ExtractError::UnsupportedConsensusBranchId)?;
-
-        let transparent = transparent
-            .into_parsed()
-            .map_err(ExtractError::TransparentParse)?;
-        let sapling = sapling.into_parsed().map_err(ExtractError::SaplingParse)?;
-        let orchard = orchard
-            .into_parsed_with_version(orchard_bundle_version)
-            .map_err(ExtractError::OrchardParse)?;
-        let ironwood = ironwood
-            .into_ironwood_parsed()
-            .map_err(ExtractError::IronwoodParse)?;
 
         let version = match (global.tx_version, global.version_group_id) {
             (V5_TX_VERSION, V5_VERSION_GROUP_ID) => Ok(TxVersion::V5),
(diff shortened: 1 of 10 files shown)


pczt/src/roles/creator/mod.rs
```diff
@@ -171,13 +236,9 @@ impl Creator {
...
-                actions: vec![],
-                flags: self.orchard_flags,
-                value_sum: (0, true),
-                anchor: self.orchard_anchor,
-                note_version: NoteVersion::V3,
-                zkproof: None,
-                bsk: None,
+                flags: self.ironwood_flags,
+                anchor: self.ironwood_anchor,
+                ..crate::orchard::EMPTY_IRONWOOD
```
ValarDragon, line 247: ooh this is a nice new syntax

pczt/src/roles/creator/mod.rs
```diff
@@ -134,6 +162,43 @@ impl Creator {
...
+    #[cfg(feature = "orchard")]
+    pub fn with_ironwood_flags(
+        mut self,
+        ironwood_flags: orchard::bundle::Flags,
+    ) -> Result<Self, Error> {
+        if self.tx_version != V6_TX_VERSION {
+            return Err(Error::IronwoodNotSupported);
+        }
+        self.ironwood_flags = ironwood_flags
+            .to_byte(orchard::bundle::BundleVersion::ironwood_v3())
```
nuttycom: For orchard we use `self.orchard_bundle_version` - for symmetry, we could create a `self.ironwood_bundle_version` but since we have the consensus branch ID and we know the value pool, we should be able to rename `orchard_bundle_version` to `bundle_version`, add a `ValuePool` argument,  and just base our determination on the consensus branch id and `ValuePool`?

ValarDragon (dismissed):
This looks good to me! Double checked everything still seemed consistent wtih #2492

pczt/src/roles/creator/mod.rs
nuttycom: Yeah default assignment via constants is great.

pczt/src/roles/creator/mod.rs
daira: Probably? It assumes that versions only change at consensus branches, but I think we're unlikely to do one in a soft-fork. (The Orchard Temporary Vulnerability Mitigation did not need to change the version, even if the version mechanism had existed before NU6.2.)

pczt/src/lib.rs
```diff
@@ -144,18 +153,41 @@ pub mod v1 {
...
+
+    #[cfg(test)]
+    mod tests {
+        use zcash_protocol::consensus::BranchId;
+
+        use crate::roles::creator::Creator;
+
+        #[test]
+        fn v1_refuses_v6_pczts_and_non_canonical_ironwood_bundles() {
+            // A v6 PCZT cannot be encoded as v1, even when its Ironwood bundle is
```
daira:
```suggestion
            // A v6 tx cannot be encoded as a v1 PCZT, even when its Ironwood bundle is
```

pczt/src/lib.rs
```diff
@@ -144,18 +153,41 @@ pub mod v1 {
...
+            let pczt = Creator::new(BranchId::Nu6_3.into(), 10_000_000, 133, [0; 32], [0; 32])
+                .unwrap()
+                .build();
+            assert!(matches!(
+                super::Pczt::try_from(pczt),
+                Err(crate::EncodingError::UnsupportedTxVersion)
+            ));
+
+            // A v5 PCZT carrying non-canonical Ironwood bundle data cannot be encoded
+            // as v1, because the data would be dropped.
```
daira:
```suggestion
            // A v5 tx carrying non-canonical Ironwood bundle data cannot be encoded
            // as a v1 PCZT, because the data would be dropped.
```

pczt/src/lib.rs
```diff
@@ -406,33 +425,47 @@ impl Pczt {
...
+        if matches!(version, TxVersion::V6)
+            && orchard_protocol_revision < OrchardProtocolRevision::V3
+        {
+            return Err(ExtractError::UnsupportedConsensusBranchId.into());
+        }
+
+        // Only the v6 transaction format carries an Ironwood bundle.
+        if !matches!(version, TxVersion::V6) && ironwood != crate::orchard::EMPTY_IRONWOOD {
+            return Err(ExtractError::IronwoodNotSupported.into());
+        }
```
daira: These `matches!` are implicitly non-exhaustive. Consider using an exhaustive match to avoid future maintenance hazards, even if it's more verbose. It's not too much more verbose because the same `match` on `version` can cover both `if`s, like this:
```rust
        match version {
                if ironwood != crate::orchard::EMPTY_IRONWOOD {
                    return Err(ExtractError::IronwoodNotSupported.into());
                }
            },
            TxVersion::V6 => {
                if orchard_protocol_revision < OrchardProtocolRevision::V3 {
                    return Err(ExtractError::UnsupportedConsensusBranchId.into());
                }
            },
        }
```

daira (dismissed):
utACK

pczt/src/lib.rs
```diff
@@ -124,7 +124,16 @@ pub mod v1 {
         type Error = super::EncodingError;
 
         fn try_from(pczt: super::Pczt) -> Result<Self, Self::Error> {
-            if !pczt.ironwood.actions.is_empty() {
+            // The v1 format predates the v6 transaction format; a parser of the v1
+            // encoding could parse a v6 PCZT but never extract a transaction from it.
```
daira, line 128: (Non-blocking)
```suggestion
            // encoding could parse a PCZT for v6 but never extract a transaction from it.
```

pczt/src/lib.rs
```diff
@@ -144,18 +153,41 @@ pub mod v1 {
...
     }
+
+    #[cfg(test)]
+    mod tests {
+        use zcash_protocol::consensus::BranchId;
+
+        use crate::roles::creator::Creator;
+
+        #[test]
+        fn v1_refuses_v6_pczts_and_non_canonical_ironwood_bundles() {
```
daira, line 168:
```suggestion
        fn v1_refuses_pczts_for_v6_txns_and_non_canonical_ironwood_bundles() {
```

daira (approved):
Post-hoc ACK

PR #2523, opened 2026-07-02
zcash_client_backend: Wire Ironwood anchors into wallet transaction building

dannywillems:
Rebased 2498. Opening a new patch to compare both changes.


dannywillems:
Continuing on 2498.

PR #2532, opened 2026-07-02
zcash_client_sqlite: Add note version to Orchard received notes

dannywillems:


zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -290,6 +290,27 @@ pub(crate) fn select_unspent_note_meta(
...
+/// See [`ORCHARD_NOTE_VERSION`]. The two versions match the Orchard protocol revision: pre-NU6.3
+/// Orchard notes are version 2, and NU6.3 Ironwood notes are version 3.
+// Referenced by the migration tests and by the invariant assertion below; the Ironwood
+// note-storage write path that uses it at runtime lands with Ironwood scanning.
+#[allow(dead_code)]
+pub(crate) const IRONWOOD_NOTE_VERSION: i64 = 3;
+
+// The two note versions must differ so that an Orchard and an Ironwood action in the same
+// transaction can share an action index under the widened uniqueness constraint.
+const _: () = assert!(ORCHARD_NOTE_VERSION != IRONWOOD_NOTE_VERSION);
```
nuttycom, line 312: We need a pool identifier; the note version is too fragile for disambiguation - even though it *does* disambiguate orchard vs ironwood, an ironwood2 pool could have the same note plaintext version.

nuttycom (changes requested):
If we're going to make `orchard_received_notes` mean notes received using the orchard _protocol_ instead of the orchard _pool_, we need to disambiguate by something more than the note plaintext version. But my inclination here would be to have a separate `ironwood_received_notes` table instead.

zcash_client_sqlite/src/wallet/orchard.rs
nuttycom: I've been thinking about this more and I'm not sure  which the best way to go is. At very least, I do not want the note plaintext version to be used as a _pool_ identifier. Keeping everything in the orchard_received_notes table might be okay, but we should definitely populate the note version from the note itself, and have an explicit pool identifier column (and make the uniqueness constraint cover that instead of the note plaintext version) if we want to do so. 

Reusing `orchard_received_notes` probably ultimately results in less churn, but I'm concerned about the confusion that results from the balance for two separate pools being mixed together in `orchard_received_notes`. I opened https://github.com/zcash/librustzcash/pull/2538 to explore the alternative of adding a separate ironwood_received_notes table.

dannywillems: I agree with your approach; I was also not convinced by the approach in this patch (which follows ValarGroup's). I will have a look at 2538.

dannywillems:
Closing in favor of https://github.com/zcash/librustzcash/pull/2538

PR #2533, opened 2026-07-02
zcash_client_sqlite: Extend scan ranges for Ironwood subtrees

dannywillems:
Stacked on #2532 (base: `dw/ironwood-note-version-migrations`).

When Ironwood notes are detected during scanning, this extends the suggested
scan ranges to include the blocks needed to complete the Ironwood note
commitment tree subtrees containing those notes, mirroring the existing Sapling
and Orchard handling, and removes the corresponding `todo!` in
`update_completed_ranges`.

### Design note

~~Ironwood notes are stored alongside Orchard notes in `orchard_received_notes`,
but their commitment tree is tracked separately (in the `ironwood_*` tables). A
single `ShieldedPool` -> table-prefix mapping (`table_constants`) cannot serve
both the note table and the tree tables for Ironwood, so `extend_range` now
takes the tree table prefix directly instead of deriving it from a
`ShieldedPool`.~~ The Sapling, Orchard, and Ironwood callers pass their respective
prefixes, and the Ironwood extension is gated on NU6.3 activation.

### Tests

Adds `extend_range_uses_ironwood_tree_shards`, verifying that `extend_range`
queries the `ironwood_tree_shards` table and extends the scan range to cover the
subtrees containing the discovered notes.

### DIFF
--- a/zcash_client_sqlite/src/wallet/scanning.rs
+++ b/zcash_client_sqlite/src/wallet/scanning.rs
@@ -28,6 +28,9 @@ use super::{block_max_scanned, wallet_birthday};
 #[cfg(feature = "orchard")]
 use zcash_client_backend::data_api::{IRONWOOD_SHARD_HEIGHT, ORCHARD_SHARD_HEIGHT};
 
+#[cfg(not(feature = "orchard"))]
+use zcash_protocol::PoolType;
+
 pub(crate) fn priority_code(priority: &ScanPriority) -> i64 {
     use ScanPriority::*;
     match priority {
@@ -674,7 +677,6 @@ pub(crate) mod tests {
     };
     use zcash_primitives::block::BlockHash;
     use zcash_protocol::{
-        ShieldedPool,
         consensus::{BlockHeight, NetworkUpgrade, Parameters},
         local_consensus::LocalNetwork,
         value::Zatoshis,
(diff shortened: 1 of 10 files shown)


zcash_client_sqlite/src/wallet/init/migrations/orchard_note_version_uniqueness.rs
```diff
@@ -0,0 +1,262 @@
...
+        "Adds note version to Orchard received-note uniqueness."
+    }
+}
+
+impl RusqliteMigration for Migration {
+    type Error = WalletMigrationError;
+
+    fn up(&self, transaction: &rusqlite::Transaction) -> Result<(), Self::Error> {
+        transaction.execute_batch(
+            "DROP INDEX IF EXISTS idx_orchard_received_notes_account;
```
dannywillems: As we merged https://github.com/zcash/librustzcash/pull/2538/ instead of https://github.com/zcash/librustzcash/pull/2532, this is out-dated.

dannywillems:
Commit message 93c7024b0f208aba50adf62bc3b356894cce266e is wrong. It is still the old design. I can update before merging.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -225,11 +227,13 @@ fn extend_range(
     conn: &rusqlite::Transaction<'_>,
     range: &Range<BlockHeight>,
     required_subtree_indices: BTreeSet<u64>,
-    protocol: ShieldedPool,
```
nuttycom, line 228: Since we added the Ironwood tables we can switch this back to `ShieldedPool`

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -330,7 +346,7 @@ pub(crate) fn scan_complete<P: consensus::Parameters>(
             conn,
             &range,
             required_sapling_subtrees,
-            ShieldedPool::Sapling,
+            SAPLING_TABLES_PREFIX,
```
dannywillems: As we're going back to the shielded pool structure, this should be reverted.

nuttycom (approved):

PR #2535, opened 2026-07-03
pczt: preverified signing parse for the low-level Signer

czarcas7ic:
## Summary

The low-level Signer now parses with a preverified fast path that skips deriving each spend's `FullViewingKey` from its wire `fvk` bytes (a Pallas point decompression plus two `commit_ivk` Sinsemilla hashes per spend). The spend authorization signature does not depend on `fvk`, so signatures are byte-identical to the full parse.

`sign_orchard_with` and `sign_ironwood_with` snapshot each spend's `(rk, fvk)` before parsing and restore the `fvk` after serialization, so signing leaves the wire bytes unchanged. The restore is enforced rather than documented: it returns `OrchardParseError::SigningClosureModifiedActions` if the signing closure changes the action count or reorders actions, detected by comparing each position's wire `rk` to the snapshot, and it makes no changes when it rejects.

The Orchard action parse is also refactored from `.map(...).collect()` into a bounded `#[inline(never)]` loop, which keeps a single per-action parse frame live at a time so a constrained signer such as an embedded hardware wallet does not overflow its task stack. This is behaviorally identical (same order, same first error), and the full `parse` used by the Verifier, Prover, and Updater is unchanged. Callers of the preverified path must have already run the full Verifier checks over the same PCZT bytes; a spend parsed this way has `fvk: None` and must not go to the Verifier check path, the Prover, or an `fvk`-preserving serialization.

## Dependency

This depends on orchard's `parse_preverified_for_signing`, which is not yet in a released orchard. The workspace temporarily `[patch]`es orchard to the branch that adds it, with a matching temporary cargo-vet exemption. Both are placeholders: once that orchard change lands and releases, they are dropped in favor of an orchard version bump.

## Testing

`cargo test -p pczt --all-features`. Two new end-to-end tests (Orchard and Ironwood) prove byte-identical signatures against the full-parse path, wire-`fvk` preservation across a signing pass, and that the produced signature verifies against `rk`. Unit tests cover the positional restore and the reorder guard.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -227,3 +227,6 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(zcash_unstable, values("nu7"))',
   'cfg(live_network_tests)',
 ] }
+
+[patch.crates-io]
+orchard = { git = "https://github.com/zcash/orchard", rev = "20c10904ae84af19fe657bbbb7369c4dac4f142a" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,4 +229,4 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 ] }
 
 [patch.crates-io]
-orchard = { git = "https://github.com/zcash/orchard", rev = "20c10904ae84af19fe657bbbb7369c4dac4f142a" }
+orchard = { git = "https://github.com/zcash/orchard", rev = "ab7c5fae83d3c197ba045b6485fa0206d68dc6f4" }
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -846,12 +846,11 @@ impl Bundle {
         bundle_version: BundleVersion,
         preverified: bool,
     ) -> Result<orchard::pczt::Bundle, orchard::pczt::ParseError> {
-        // Parse each action through an `#[inline(never)]` helper driven by a plain
-        // loop, not `.map(..).collect()`: the latter compiles to a single combined
-        // stack frame (the per-action closure inlined into `try_fold`) tens of KB
-        // deep, overflowing the fixed task stacks of embedded signers. The loop keeps
-        // one bounded helper frame live at a time, and is behaviorally identical
-        // (same order, same first error).
+        // We parse actions through a helper that is specifically `#[inline(never)]`.
+        // This is because if this gets inlined in a loop (e.g. `.map(..).collect()`),
+        // it could compile into a stack frame that is tens of KB deep.
+        // This can overflow stacks of embedded signers for high action count
+        // transactions.
         #[inline(never)]
         fn parse_action_inner(
             action: Action,
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -924,16 +924,7 @@ impl Bundle {
                 action.output.proprietary,
             )?;
 
-            if preverified {
-                orchard::pczt::Action::parse_preverified_for_signing(
-                    action.cv_net,
-                    spend,
-                    output,
-                    action.rcv,
-                )
-            } else {
-                orchard::pczt::Action::parse(action.cv_net, spend, output, action.rcv)
-            }
+            orchard::pczt::Action::parse(action.cv_net, spend, output, action.rcv)
         }
 
         let note_version = self.note_version;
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -227,3 +227,6 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(zcash_unstable, values("nu7"))',
   'cfg(live_network_tests)',
 ] }
+
+[patch.crates-io]
+orchard = { git = "https://github.com/zcash/orchard", rev = "c45692ff2d0a3ff54b578ffc3d2a5afb0fedc7ba" }
```
czarcas7ic: TODO: Replace with orchard tag when merged

pczt/src/orchard.rs
```diff
@@ -795,22 +795,79 @@ pub(crate) fn orchard_bundle_version(global: &crate::common::Global) -> Option<B
...
+    /// selects between the full parse and the preverified signing parse.
+    fn into_parsed_inner(
+        self,
+        bundle_version: BundleVersion,
+        preverified: bool,
+    ) -> Result<orchard::pczt::Bundle, orchard::pczt::ParseError> {
+        // Parse each action through an `#[inline(never)]` helper driven by a plain
+        // loop, not `.map(..).collect()`: the latter compiles to a single combined
+        // stack frame (the per-action closure inlined into `try_fold`) tens of KB
+        // deep, overflowing the fixed task stacks of embedded signers. The loop keeps
```
czarcas7ic: This is not a hypothetical issue, I ran into this with Keystone tests which caused the change

pczt/src/roles/low_level_signer/mod.rs
```diff
@@ -112,6 +138,53 @@ impl Signer {
...
+    bundle
+        .actions()
+        .iter()
+        .map(|action| (action.spend.rk, action.spend.fvk))
+        .collect()
+}
+
+/// Restores the wire `fvk` bytes from [`snapshot_spend_fvks`] into each spend by
+/// position, after proving the signing closure did not resize or reorder the action
+/// list.
```
ValarDragon:
```suggestion
/// position, after checking the signing closure did not resize or reorder the action
/// list.
```

proving is overloaded 😂

pczt/src/orchard.rs
```diff
@@ -795,22 +795,79 @@ pub(crate) fn orchard_bundle_version(global: &crate::common::Global) -> Option<B
...
+        // one bounded helper frame live at a time, and is behaviorally identical
+        // (same order, same first error).
```
ValarDragon:
```suggestion
        // We parse actions through a helper that is specifically `#[inline(never)]`.
        // This is because if this gets inlined in a loop (e.g. `.map(..).collect()`),
        // it could compile into a stack frame that is tens of KB deep.
        // This can overflow stacks of embedded signers for high action count
        // transactions.
```

pczt/src/roles/low_level_signer/mod.rs
```diff
@@ -50,16 +74,18 @@ impl Signer {
 
         let bundle_version = crate::orchard::orchard_bundle_version(&pczt.global)
             .ok_or(OrchardParseError::UnsupportedConsensusBranchId)?;
+        let fvk_snapshot = snapshot_spend_fvks(&pczt.orchard);
         let mut bundle = pczt
             .orchard
             .clone()
```
ValarDragon, line 80: If were hitting issues with performance of this function specifically, this Clone stands out as weird to me.

Granted were hitting problems with stack size, and this _should_ be irrelevant for stack size

pczt/src/roles/low_level_signer/mod.rs
```diff
@@ -112,6 +138,53 @@ impl Signer {
     }
 }
 
+/// A by-position snapshot of each spend's wire `(rk, fvk)` bytes: the `fvk` that the
+/// preverified signing parse drops and must restore, paired with the `rk` used as a
+/// per-position tamper check. See [`restore_spend_fvks`].
+#[cfg(feature = "orchard")]
+type SpendFvkSnapshot = alloc::vec::Vec<([u8; 32], Option<[u8; 96]>)>;
```
ValarDragon: Were removing the rk from the sign call here. Any check, e.g. `rk_1 != rk_2`, can be checked outside the sign call. In the exact sign call we are hitting stack limits.

pczt/src/orchard.rs
czarcas7ic: https://github.com/zcash/librustzcash/pull/2535/commits/93ae3603b6c7e92d880144c3039bf1af4060971c

pczt/src/roles/low_level_signer/mod.rs
nuttycom: Seems like maybe one could use `std::mem::take` to avoid the clone here?

nuttycom (approved):

PR #2538, opened 2026-07-03
zcash_client_sqlite: Add ironwood_received_notes

nuttycom:
This is an alternative to #2532 that explores creating a separate table for Ironwood received note storage.


dannywillems (approved):
utACK. I prefer this to the version in https://github.com/zcash/librustzcash/pull/2532.

Issue #2530, opened 2026-07-02
Finish upstreaming Ironwood / Orchard dummy-ciphertext work (fork backport + #2508/#2498 follow-ups)


dannywillems:
# Tracking: finish upstreaming the Ironwood / Orchard dummy-ciphertext work

Backport the remaining changes from the valargroup fork branch
`adam/qleak-pr44-orchard-dummy-ciphertexts`
(https://github.com/valargroup/librustzcash) into `zcash/librustzcash` `main`,
and close out the follow-up TODOs and deferred tests left by the already-merged
Ironwood PRs **#2508** and **#2498**.

This is written as a **step-by-step plan**: work the phases in order, one at a
time. Each phase is self-contained and can be picked up independently.

---

## Background (why hashes don't match)

The fork branch diverged from `main` at PR #2405 (`32fa8e2568`). Since then
`main` gained **156 commits** and the fork is **76 commits** ahead. Upstream did
**not** cherry-pick the fork; it **re-implemented** the same work as a series of
clean split PRs, so `git cherry` (0 matches) and `git range-diff` (1 fuzzy
match) both fail to pair commits. The remaining work must therefore be judged at
the **feature level**, not by commit hash.

Already-merged upstream PRs that cover the bulk of the fork:

> #2421 ironwood-split-1-foundation, #2429 ironwood-split-2-v6-txid, #2446
> ironwood-catchup, #2485 + #2498 ironwood-tx-builder(+completion), #2508
> ironwood-wallet-fees, #2427 pczt_v1, #2490 pczt-v2-serialization, #2492
> pczt-ironwood-roles, #2520 pczt_ironwood_completion, #2483 orchard_v1, #2493
> orchard-bundle-version, #2506 bump-orchard-0.15.0-pre.1, #2432
> orchard-upstream-pin, #2484 ironwood_zip317, #2501 ironwood_lightwallet_protocol,
> #2513 orchard-limit.

### Do NOT port (already upstream)

Ironwood foundation, v6 txid, tx-builder(+completion), wallet-fees, PCZT
v1/v2/roles/completion (incl. the PR44 `dummy_sk` payload — `pczt/src/orchard.rs`
already carries it, redactor aligned), NU6.3 metadata, orchard bump.

### Do NOT port (obsolete / superseded)

Every `qr_orchard` / "reduced Orchard API" / "legacy Orchard API compatibility
shim" commit. `main` dropped `qr_orchard` entirely and moved to the published
[...]

PR #2534, opened 2026-07-02
zcash_client_sqlite: Report Ironwood outputs as pool code 4

dannywillems:
~~Ironwood notes are stored alongside Orchard notes in `orchard_received_notes`,
distinguished by `note_version` (added in #2532).~~

This assigns Ironwood the
distinct SQLite `output_pool` code 4 and reports Ironwood outputs under it in the
wallet-facing views.

### Changes

- Assign pool code 4 to the Ironwood pool in `pool_code`/`parse_pool_code`
  (transparent 0, Sapling 2, Orchard 3, Ironwood 4), replacing the `todo!`.
- `v_received_outputs` and `v_received_output_spends` derive the pool code from
  `note_version` (Orchard = 3, Ironwood = 4). `v_transactions` and `v_tx_outputs`
  propagate the pool code, so they pick this up automatically.
- A new migration (`ironwood_pool_code_views`) drops and recreates the four views
  at their updated definitions, and repoints any existing sent note that
  corresponds to a received Ironwood note from pool code 3 to 4. It depends on all
  current DAG leaves so the views are recreated at their latest definition.

### Tests

- `pool_code` round-trip and distinctness.
- A proptest (using the reusable `arb_orchard_note` and `arb_ironwood_note` strategies) asserting that,
  regardless of note payload, an Orchard note is reported by `v_received_outputs`
  under pool code 3 and an Ironwood note under pool code 4.
- `verify_schema` confirms the recreated views match their definitions.

Adapted from valargroup/librustzcash@d94102c3ea and the fork's
`ironwood_pool_code_views` migration
(`adam/qleak-pr44-orchard-dummy-ciphertexts`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2539, opened 2026-07-03
Add wallet scanning and storage for the Ironwood shielded pool

dannywillems:
This PR adds wallet support for the Ironwood shielded pool ([ZIP 2005], NU6.3) to
`zcash_client_backend` and `zcash_client_sqlite`: assigning the pool code,
scanning blocks for received Ironwood notes, storing them, and building the
Ironwood note commitment tree.

Ironwood is treated throughout as a pool distinct from Orchard. Its notes are
Orchard-shaped and are decrypted with the account's Orchard viewing keys, but
they carry version 3 note plaintexts and are decrypted under a separate
note-encryption domain. Where Ironwood reuses Orchard's cryptographic primitives
(commitment, nullifier, and tree types), that reuse is expressed through aliases
rather than by modelling Ironwood as a variant of Orchard.

Tracking issue: #2530. Adapted from the Ironwood work in
valargroup/librustzcash (branch `adam/qleak-pr44-orchard-dummy-ciphertexts`,
Adam Tucker).

## What this adds

**Pool code and views**
- Assigns Ironwood the SQLite `output_pool` code 4.
- Adds received Ironwood notes and their spends to the `v_received_outputs` and
  `v_received_output_spends` views.

**Scanning**
- `ScanningKeys` and the compact- and full-block `BatchRunners` gain a distinct
  Ironwood set, so the scanner trial-decrypts `tx.ironwood_actions` and the
  transaction's Ironwood bundle.
- Ironwood outputs are decrypted under the Ironwood note-encryption domain
  (`orchard::note_encryption::IronwoodDomain`), which accepts version 3
  plaintexts and is distinct from the Orchard domain.
- The scanner tracks Ironwood outputs and the Ironwood note commitment tree as a
  pool distinct from Orchard, across `ScannedBlock`, `BlockMetadata`, `WalletTx`,
  and `Nullifiers`.

**Storage and commitment tree**
- Received Ironwood notes are stored in `ironwood_received_notes` (note version
  3), Ironwood nullifiers are tracked for spend detection, and the Ironwood note
  commitment tree is persisted during `put_blocks`, reusing the Orchard shard
  height and `with_ironwood_tree_mut`.
- `ChainState` and the gRPC `TreeState` gain the Ironwood note commitment tree
  frontier; an absent treestate field parses as an empty tree, the correct
  Ironwood treestate at pool activation.
- Cross-pool note-commitment-tree checkpoint reconciliation is generalized from
  two pools to three, so each tree gains a checkpoint at every height
  checkpointed in any pool.
- Received Ironwood notes are included in the account balance: `AccountBalance`
  gains an Ironwood component and `get_wallet_summary` computes it, so scanned
  Ironwood notes are reflected in the wallet's total and spendable balances.

**Tests**
- A property test for the three-pool checkpoint reconciliation.
- A scanning test that decrypts a real version 3 note and asserts it is detected
  as an Ironwood output, distinct from Orchard.
- An end-to-end test that scans a block containing an Ironwood note and asserts
  it is stored in `ironwood_received_notes` as note version 3 and the Ironwood
  note commitment tree is persisted. This required threading the Ironwood tree
  through the block-generation test harness and adding an `IronwoodFvk` test
  viewing key that emits version 3 notes.

## Known limitations / follow-ups

- The Ironwood send/spend path (transaction builder output and change, proposal
  serialization, and note selection for spends) is not implemented; those
  `todo!` sites remain for a later change.

[ZIP 2005]: https://zips.z.cash/zip-2005


zcash_client_backend/src/scanning.rs
```diff
@@ -210,6 +210,22 @@ impl<AccountId> ScanningKeyOps<OrchardDomain, AccountId, orchard::note::Nullifie
...
+///
+/// Ironwood ([ZIP 2005], NU6.3) is a distinct shielded pool with its own note commitment tree,
+/// but its notes are Orchard-shaped: they are decrypted under the Orchard note-encryption domain
+/// and use the Orchard commitment and nullifier types. These aliases name that reuse so that
+/// Ironwood can share the underlying Orchard primitives without being modelled as part of an
+/// "Orchard family"; the two pools are kept strictly separate throughout the scanner.
+///
+/// [ZIP 2005]: https://zips.z.cash/zip-2005
+#[cfg(feature = "orchard")]
+pub(crate) type IronwoodDomain = OrchardDomain;
```
dannywillems: This will need to be changed, as there is an existing `orchard::note_encryption::IronwoodDomain` in the `orchard` crate, version `0.15.0`. The consistency bug has been detected while writing property tests. A fix is coming.

dannywillems: Fixed in [93c401e](https://github.com/zcash/librustzcash/pull/2539/commits/93c401edb79b82a31b647b91ca5e0d49bf4f04f9)

zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -456,22 +510,50 @@ where
...
+
+            let [ensure_sapling, ensure_orchard, ensure_ironwood] = cross_pool_ensure_heights(
+                &sapling_checkpoint_positions.keys().copied().collect(),
+                &orchard_checkpoint_positions.keys().copied().collect(),
+                &ironwood_checkpoint_positions.keys().copied().collect(),
+            );
+
             (
                 ensure_checkpoints(
-                    orchard_checkpoint_positions.keys(),
```
dannywillems, line 470: Was this a bug before?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -3790,6 +3792,8 @@ pub fn birthday_in_anchor_shard<T: ShieldedPoolTester>(
                     sapling_initial_tree,
                     #[cfg(feature = "orchard")]
                     orchard_initial_tree,
+                    #[cfg(feature = "orchard")]
+                    Frontier::empty(),
```
dannywillems: We should define a value `ironwood_initial_tree` to follow the pattern of the others

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4705,6 +4709,8 @@ pub fn truncate_to_chain_state_below_birthday<T: ShieldedPoolTester, Dsf>(
...
```
dannywillems: We should define a value `ironwood_initial_tree` to follow the pattern of the others

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -5288,6 +5296,8 @@ where
...
```
dannywillems: We should define a value `ironwood_initial_tree` to follow the pattern of the others

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -5550,6 +5560,8 @@ where
...
```
dannywillems: We should define a value `ironwood_initial_tree` to follow the pattern of the others

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -6129,6 +6141,8 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, Dsf>(
                     Frontier::empty(),
                     #[cfg(feature = "orchard")]
                     Frontier::empty(),
+                    #[cfg(feature = "orchard")]
+                    Frontier::empty(),
```
dannywillems: We should define a value `ironwood_initial_tree` to follow the pattern of the others

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4849,6 +4855,8 @@ pub fn truncate_to_chain_state_above_scanned<T: ShieldedPoolTester, Dsf>(
         shard2_sapling_frontier,
         #[cfg(feature = "orchard")]
         shard2_orchard_frontier,
+        #[cfg(feature = "orchard")]
+        Frontier::empty(),
```
dannywillems: We should define a value `ironwood_initial_tree` to follow the pattern of the others

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -683,16 +722,25 @@ where
                     },
                 );
 
+            // Ironwood notes are not generated by callers that specify explicit tree sizes, so the
+            // gap-filled prior block carries the prior Ironwood tree forward unchanged.
+            #[cfg(feature = "orchard")]
+            let ironwood_tree = prior_cached_block.chain_state.final_ironwood_tree().clone();
+            let ironwood_end_size = prior_cached_block.ironwood_end_size();
```
dannywillems: Should `ironwood_end_size` be behind the feature flag orchard too?

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -1871,13 +1924,18 @@ impl<Cache, DsFactory: DataStoreFactory> TestBuilder<Cache, DsFactory> {
             #[cfg(feature = "orchard")]
             let _final_orchard_tree_size =
                 initial_state.chain_state.final_orchard_tree().tree_size() as u32;
+            let _final_ironwood_tree_size = 0;
```
dannywillems, line 1937: Should this be behind the feature flag `orchard`?

zcash_client_backend/src/data_api/ll/wallet.rs
dannywillems: No, this was not a bug. With two pools, ensuring each tree carries the other pool's checkpoint heights leaves both trees checkpointed at the union of their heights, which is correct. This change generalizes that to three pools: each tree gains a checkpoint at every height checkpointed in any other pool (the union of the other two). It reduces exactly to the previous two-pool behavior when Ironwood has no checkpoints, which the `ensure_heights_degrade_to_two_pools` proptest verifies.


Answer by Claude 4.8

dannywillems:
## Status update: Ironwood receive and send paths complete

The branch now implements the full Ironwood wallet path end to end (scan -> store -> tree -> balance -> select -> spend), rebased on the latest `main` (which absorbed #2534). Below is what was done, grouped for step-by-step review, followed by the correctness bugs the tests surfaced.

### Scanning / receive
- Ironwood scanning keys and compact/full-block batch runners as a pool distinct from Orchard.
- Block scanner tracks Ironwood outputs, nullifiers, and the Ironwood note commitment tree (`ScannedBlock`, `BlockMetadata`, `WalletTx`, `Nullifiers`).
- Received notes are stored in `ironwood_received_notes` (note version 3), nullifiers tracked for spend detection.
- `ChainState` and the gRPC `TreeState` carry the Ironwood tree frontier; commitment-tree persisted in `put_blocks` (reusing `ORCHARD_SHARD_HEIGHT` and `with_ironwood_tree_mut`); three-way checkpoint reconciliation.
- Received Ironwood notes are included in the account balance.

### Send / spend
- `select_spendable_ironwood_notes` / `get_spendable_ironwood_note`; `ReceivedNotes` gains a distinct `ironwood` list.
- Input-pool exclusivity: shielded inputs come from a single group, **Orchard alone** or **Sapling + Ironwood together**, never combining Orchard with the others; preference is recipient-matching (avoids an unnecessary turnstile output); an amount that would need Orchard plus another pool fails as insufficient funds. Orchard-receiver payments and Orchard change route to the Ironwood bundle post-NU6.3 (the Orchard V3 bundle forbids the cross-address output that Orchard change would require).
- The transaction-builder spend side (`Builder::add_ironwood_spend`, Ironwood anchor) was already upstream; this connects note selection to it.

### Correctness bugs found and fixed (via the tests)
1. **Wrong decryption domain** — Ironwood outputs were decrypted under `OrchardDomain` (version 2), which rejects version 3 plaintexts, so a wallet would have silently detected no Ironwood notes. Fixed to use `orchard::note_encryption::IronwoodDomain`.
2. **Double-spend hazard** — `store_transaction_to_be_sent` never marked Ironwood notes spent, so a spent Ironwood note stayed selectable and the balance did not decrease. Fixed by marking the transaction's Ironwood-bundle nullifiers spent.
3. **Insufficient-funds report** — an Orchard-only wallet short of funds reported `available: 0` instead of its true single-group balance.

### Tests
All new tests are proptests with reusable amount strategies:
- checkpoint reconciliation invariants;
- version-3 note scan detection;
- end-to-end receive storage (note in `ironwood_received_notes`, tree persisted, balance);
- end-to-end spend (note actually spent, fee charged, balance decreases);
- the four input-pool-exclusivity invariants (Orchard never combined with Sapling/Ironwood; Sapling+Ironwood may combine; Orchard-plus-another-pool is insufficient; an Orchard payment routes to the Ironwood bundle and builds).
- proptest regression seeds are checked in.

### Verification
`cargo clippy --all-features --all-targets -- -D warnings`, `cargo fmt --check`, and the full `zcash_client_sqlite` suite (284 passing) are all green.

### Not in this PR
The Ironwood proposal/protobuf `ValuePool` representation (`proposal.rs` / `proto.rs`) is left for a later change; Ironwood inputs ride under `PoolType::ORCHARD` and are distinguished by note version, so those `todo!` arms are currently unreachable.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1316,8 +1316,9 @@ where
     } else {
         OrchardBuildMode::Orchard
     };
-    // Orchard-pool payment and change outputs are routed to the Ironwood bundle
-    // in exactly the Ironwood spend and Ironwood output modes.
+    // Orchard-pool *payment* outputs are routed to the Ironwood bundle in exactly the Ironwood
+    // spend and Ironwood output modes: post-NU6.3, new Orchard payment outputs are forbidden, so
+    // payments to an Orchard receiver are created in the Ironwood bundle.
```
dannywillems: "so payments to an Orchard receiver are created in the Ironwood bundle" should be rephrased.

zcash_client_backend/src/scanning/compact.rs
```diff
@@ -166,27 +212,47 @@ where
                     })
                     .collect::<Result<Vec<_>, _>>()?,
             );
+
+            #[cfg(feature = "orchard")]
```
dannywillems, line 216: LGTM. Copied from the Orchard case above.

zcash_client_backend/src/scanning/compact.rs
```diff
@@ -272,11 +343,30 @@ where
             orchard_spends
         };
 
+        #[cfg(feature = "orchard")]
```
dannywillems, line 367: LGTM. Coped from the Orchard case above.

zcash_client_backend/src/scanning/compact.rs
```diff
@@ -354,12 +444,54 @@ where
         #[cfg(feature = "orchard")]
         orchard_note_commitments.append(&mut orchard_nc);
 
+        #[cfg(feature = "orchard")]
```
dannywillems, line 479: LGTM. Copied from the Orchard case above.

zcash_client_backend/src/scanning/compact.rs
```diff
@@ -508,13 +650,29 @@ impl PositionTracker {
             |m| m.orchard_commitment_tree_size,
         )?;
 
+        #[cfg(feature = "orchard")]
```
dannywillems, line 686: LGTM. Copied from the Orchard case above.

zcash_client_backend/src/scanning/compact.rs
dannywillems, line 721: Note: this field should be renamed into something that contains `orchard`, like `orchard_actions`, to follow the pattern for ironwood (meaningfully coined `ironwood_actions` by Claude).

dannywillems: Would be in a follow-up.

zcash_client_backend/src/scanning/compact.rs
```diff
@@ -572,6 +742,16 @@ impl PositionTracker {
                     computed: self.orchard_tree_position,
                 });
             }
+
+            #[cfg(feature = "orchard")]
```
dannywillems, line 779: LGTM, follow the Orchard case pattern above.

zcash_client_backend/src/scanning/full.rs
```diff
@@ -652,6 +797,17 @@ fn orchard_action_count(tx: &Transaction) -> u32 {
     })
 }
 
+/// Returns the number of Ironwood actions in `tx`.
+///
+/// Note commitment tree sizes are `u32`-bounded by the protocol, so a valid transaction
+/// can never contain more than `u32::MAX` actions.
+#[cfg(feature = "orchard")]
```
dannywillems, line 807: Should be moved into a `impl` declaration.

zcash_client_backend/src/data_api.rs
```diff
@@ -977,28 +1024,49 @@ impl<NoteRef> ReceivedNotes<NoteRef> {
...
+    }
+
     /// Computes the total value of Sapling notes.
     pub fn sapling_value(&self) -> Result<Zatoshis, BalanceError> {
         self.sapling.iter().try_fold(Zatoshis::ZERO, |acc, n| {
             (acc + n.note_value()?).ok_or(BalanceError::Overflow)
         })
     }
 
-    /// Computes the total value of Sapling notes.
```
dannywillems, line 987: Ah, nice catch, Claude.

zcash_client_backend/src/data_api.rs
```diff
@@ -1933,6 +2010,19 @@ pub trait WalletRead {
         )
     }
 
+    /// Returns the nullifiers for Ironwood notes that the wallet is tracking, along with their
+    /// associated account IDs, that are either unspent or have not yet been confirmed as spent.
+    /// Ironwood nullifiers are Orchard-shaped but are tracked as a separate pool.
+    #[cfg(feature = "orchard")]
```
dannywillems, line 2062: LGTM. This follows the Orchard structure above.

zcash_client_backend/src/proto.rs
```diff
@@ -402,6 +402,33 @@ impl service::TreeState {
         }
     }
 
+    /// Deserializes and returns the Ironwood note commitment tree field of the tree state.
+    ///
+    /// The Ironwood tree is Orchard-shaped, but Ironwood is a distinct pool tracked separately
+    /// from Orchard. An empty field yields an empty tree, which is the correct treestate at the
+    /// Ironwood pool's activation.
+    #[cfg(feature = "orchard")]
```
dannywillems, line 415: LGTM. It follows the Orchard structure above.

zcash_client_backend/src/scanning.rs
```diff
@@ -210,6 +210,51 @@ impl<AccountId> ScanningKeyOps<OrchardDomain, AccountId, orchard::note::Nullifie
...
+/// The nullifier type for Ironwood notes. This is the Orchard nullifier type, which does not
+/// depend on the note plaintext version. See [`IronwoodDomain`].
+#[cfg(feature = "orchard")]
+pub(crate) type IronwoodNullifier = orchard::note::Nullifier;
+
+/// An Orchard viewing key trial-decrypts Ironwood outputs under the Ironwood note-encryption
+/// domain. The key material is the same as for Orchard; only the domain (and thus the accepted
+/// note plaintext version) differs.
+#[cfg(feature = "orchard")]
+impl<AccountId> ScanningKeyOps<IronwoodDomain, AccountId, orchard::note::Nullifier>
```
dannywillems, line 234: ScanningKeyOps could get an additional type parameter for the domain, which would later be instantiated with the type `IronwoodDomain` and `OrchardDomain`.

zcash_client_backend/src/scanning.rs
```diff
@@ -345,6 +438,8 @@ pub struct Nullifiers<AccountId> {
     sapling: Vec<(AccountId, sapling::Nullifier)>,
     #[cfg(feature = "orchard")]
     orchard: Vec<(AccountId, orchard::note::Nullifier)>,
+    #[cfg(feature = "orchard")]
+    ironwood: Vec<(AccountId, IronwoodNullifier)>,
```
dannywillems, line 442: Might be too much to have the alias `IronwoodNullifier`.

zcash_client_sqlite/proptest-regressions/wallet/orchard.txt
```diff
@@ -0,0 +1,8 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
dannywillems: Needs your inputs on this @nuttycom

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -436,12 +436,14 @@ pub(crate) fn truncate_to_chain_state_commitment_tree_error<T: ShieldedPoolTeste
             captured.block_hash(),
             foreign.final_sapling_tree().clone(),
             captured.final_orchard_tree().clone(),
+            incrementalmerkletree::frontier::Frontier::empty(),
```
dannywillems: Nit: we could have this in a let-binding before.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -436,12 +436,14 @@ pub(crate) fn truncate_to_chain_state_commitment_tree_error<T: ShieldedPoolTeste
...
         ),
         ShieldedPool::Orchard => ChainState::new(
             capture_height,
             captured.block_hash(),
             captured.final_sapling_tree().clone(),
             foreign.final_orchard_tree().clone(),
+            incrementalmerkletree::frontier::Frontier::empty(),
```
dannywillems: Nit: we could have this in a let-binding before.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -558,12 +560,14 @@ pub(crate) fn put_blocks_commitment_tree_error<T: ShieldedPoolTester>() {
             captured.block_hash(),
             foreign.final_sapling_tree().clone(),
             captured.final_orchard_tree().clone(),
+            incrementalmerkletree::frontier::Frontier::empty(),
```
dannywillems: Nit: we could have this in a let-binding before.

zcash_client_sqlite/src/testing/pool.rs
```diff
@@ -558,12 +560,14 @@ pub(crate) fn put_blocks_commitment_tree_error<T: ShieldedPoolTester>() {
...
         ),
         ShieldedPool::Orchard => ChainState::new(
             from_height - 1,
             captured.block_hash(),
             captured.final_sapling_tree().clone(),
             foreign.final_orchard_tree().clone(),
+            incrementalmerkletree::frontier::Frontier::empty(),
```
dannywillems: Nit: we could have this in a let-binding before.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -818,6 +818,8 @@ pub(crate) mod tests {
                         sapling_initial_tree,
                         #[cfg(feature = "orchard")]
                         orchard_initial_tree,
+                        #[cfg(feature = "orchard")]
+                        Frontier::empty(),
```
dannywillems: This could be using a let-binding.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1233,6 +1237,8 @@ pub(crate) mod tests {
...
```
dannywillems: This could be using a let-binding.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -975,6 +977,8 @@ pub(crate) mod tests {
...
```
dannywillems: This could be using a let-binding.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1426,6 +1432,8 @@ pub(crate) mod tests {
...
```
dannywillems: This could be using a let-binding.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1755,6 +1763,7 @@ pub(crate) mod tests {
                         birthday_prior_block_hash,
                         Frontier::empty(), // the Sapling tree is unused in this test
                         orchard_initial_tree,
+                        Frontier::empty(), // the Ironwood tree is unused in this test
```
dannywillems: This could be using a let-binding.

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -1755,6 +1763,7 @@ pub(crate) mod tests {
                         birthday_prior_block_hash,
                         Frontier::empty(), // the Sapling tree is unused in this test
```
dannywillems: This could be using a let-binding.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -2695,6 +2695,33 @@ pub(crate) fn get_wallet_summary<P: consensus::Parameters>(
         drop(orchard_trace);
     }
 
+    #[cfg(feature = "orchard")]
+    {
```
dannywillems, line 2699: LGTM. This copies the Orchard case above.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3341,6 +3399,13 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
         #[cfg(not(feature = "orchard"))]
         panic!("Sent a transaction with Orchard Actions without `orchard` enabled?");
     }
+    #[cfg(feature = "orchard")]
+    if let Some(bundle) = sent_tx.tx().ironwood_bundle() {
```
dannywillems: Should we follow the same pattern as for Orchard above, with a `panic!`? cc @nuttycom

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -332,12 +378,20 @@ pub(crate) fn parse_note_version(code: i64) -> Option<NoteVersion> {
...
+/// version is recorded in the `note_version` column but does not influence table selection. This
+/// lets the scanner route its separate Orchard and Ironwood output streams to their respective
+/// tables while the decrypted-transaction path selects the table from the note version.
 pub(crate) fn put_received_note<
     T: ReceivedOrchardOutput<AccountId = AccountUuid>,
     P: consensus::Parameters,
 >(
     conn: &rusqlite::Transaction,
     params: &P,
+    table_prefix: &str,
```
dannywillems: Claude decided to move the logic of `table_prefix` to the caller.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5448,6 +5539,8 @@ mod tests {
                         sapling_initial_tree,
                         #[cfg(feature = "orchard")]
                         orchard_initial_tree,
+                        #[cfg(feature = "orchard")]
+                        Frontier::empty(),
```
dannywillems: Nit: we could use a let-binding.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5595,6 +5688,8 @@ mod tests {
...
```
dannywillems: Nit: we could use a let-binding.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5759,6 +5854,8 @@ mod tests {
...
```
dannywillems: Nit: we could use a let-binding.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3422,9 +3487,17 @@ pub(crate) fn store_transaction_to_be_sent<P: consensus::Parameters>(
                 }
                 #[cfg(feature = "orchard")]
                 Note::Orchard(note) => {
+                    // The decrypted-transaction path handles a single Orchard-domain output
```
dannywillems: Note: the logic of the variable `table_prefix` was previously the job of the callee. Claude decided to delegate that to the caller now.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
             .unwrap();
         assert_eq!(total, ((value - spend_value).unwrap() + value3).unwrap());
     }
+
+    /// Property tests for the shielded-input pool-exclusivity rules enforced by input selection
```
dannywillems: Important review, cc @nuttycom. Claude struggled.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+    /// once the Ironwood pool (NU6.3) is active:
+    ///
+    /// 1. Orchard inputs are never combined with Sapling or Ironwood inputs. Spending Orchard is a
```
dannywillems: These are four of the properties we discussed with @nuttycom. Claude has been instructed to implement proptest for these cases.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+                nu6: Some(activation),
+                nu6_1: Some(activation),
+                nu6_2: Some(activation),
+                nu6_3: Some(activation),
+                ..TestBuilder::<(), ()>::DEFAULT_NETWORK
+            }
+        }
+
+        // A single-output ZIP 317 change strategy. Its nominal change pool is Orchard, but once
+        // Ironwood is active any Orchard-pool change is routed into the Ironwood bundle (the
```
dannywillems: > any Orchard-pool change is routed into the Ironwood bundle

This is incorrect. Orchard-pool change MUST stay in the Orchard pool, as we don't want to have the changes crossing the pools.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+        }
+
+        // Counts the shielded input notes selected across every step of a proposal, partitioned by
+        // pool: Sapling, Orchard (note version v2), and Ironwood (note version v3).
+        fn input_pool_counts<FeeRuleT, NoteRef>(
+            proposal: &zcash_client_backend::proposal::Proposal<FeeRuleT, NoteRef>,
+        ) -> (usize, usize, usize) {
+            let (mut sapling, mut v2, mut v3) = (0usize, 0usize, 0usize);
+            for step in proposal.steps() {
+                if let Some(inputs) = step.shielded_inputs() {
```
dannywillems: Note: this logic should be lifted to the Step structure.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+                    .propose_transfer(
+                        account_id,
+                        &input_selector,
+                        &change_strategy,
+                        request,
+                        ConfirmationsPolicy::MIN,
+                    )
+                    .unwrap();
+
+                let (sapling, v2, v3) = input_pool_counts(&proposal);
```
dannywillems: This should be explicitly named with `orchard` and `ironwood`, instead of `v2` and `v3`.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+                    request,
+                    ConfirmationsPolicy::MIN,
+                );
+
+                let err = result.expect_err(
+                    "spending must fail: Orchard cannot be combined with Sapling to reach the amount",
+                );
+                prop_assert!(
+                    format!("{err:?}").contains("InsufficientFunds"),
+                    "expected an InsufficientFunds error, got: {:?}",
```
dannywillems: Should we have a more explicit error?

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+                    .unwrap()
+                    .expect("the sent transaction was stored");
+
+                // The Orchard-receiver payment (and the Orchard-input change) are carried by the
+                // Ironwood bundle; the Orchard bundle carries the Orchard spend.
+                prop_assert!(
+                    tx.ironwood_bundle().is_some(),
+                    "the payment to an Orchard receiver must be routed to the Ironwood bundle",
+                );
+                prop_assert!(
```
dannywillems, line 2593: The change must stay in the Orchard pool, more precisely.

zcash_client_sqlite/src/wallet/orchard.rs
dannywillems: Claude detected some missing methods, and some wrong logic in the proposal. This will be fixed in a commit.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -977,4 +1284,407 @@ pub(crate) mod tests {
...
+                    .propose_transfer(
+                        account_id,
+                        &input_selector,
+                        &change_strategy,
+                        request,
+                        ConfirmationsPolicy::MIN,
+                    )
+                    .unwrap();
+
+                let (sapling, v2, v3) = input_pool_counts(&proposal);
```
dannywillems: This should be explicitly named with orchard and ironwood, instead of v2 and v3.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1041,6 +1063,10 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                 &orchard_view,
                 #[cfg(feature = "orchard")]
                 &ironwood_view,
+                // Once Ironwood is active, Orchard-pool change is routed into the Ironwood bundle
```
dannywillems: > Orchard-pool change is routed into the Ironwood bundle

No, this is not what we discussed with @nuttycom

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1144,29 +1172,95 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+            // is never combined with another pool to make up the difference).
+            let sapling_ironwood_pools = {
+                let mut pools = vec![];
+                if sapling_supported {
+                    pools.push(ShieldedPool::Sapling);
+                }
+                // Ironwood notes can be spent once the Ironwood pool is active at the target
+                // height, and may be combined with Sapling inputs.
+                #[cfg(feature = "orchard")]
+                if orchard_supported
```
dannywillems: Nit: is `ironwood_activate_at` redundant? This is the code above:
```
       let (sapling_supported, orchard_supported) = (true, cfg!(feature = "orchard"));
        // Without the `orchard` feature there are no Orchard-family pools to select from, so
        // `orchard_supported` (always false) is only referenced by Orchard-gated code.
        let _ = orchard_supported;
```

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1144,29 +1172,95 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                        let orchard_value = orchard.total_value()?;
+                        let sapling_ironwood_value = sapling_ironwood.total_value()?;
+                        let orchard_covers = orchard_value >= amount_required;
+                        let sapling_ironwood_covers = sapling_ironwood_value >= amount_required;
+                        // Prefer the input group that matches the payment's pool, to avoid an
+                        // unnecessary cross-pool (turnstile) output: spend the Orchard group when
+                        // the payment targets an Orchard receiver, otherwise spend the
+                        // Sapling+Ironwood group. Fall back to the other group when the preferred
+                        // one cannot cover the amount, and to the larger group when neither covers
+                        // (so the insufficient-funds error is accurate). Orchard is never combined
```
dannywillems: We might want to have variants of `InsufficientFunds` to inform the API user more precisely.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1144,29 +1172,95 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+                            )
+                            .map_err(InputSelectorError::DataSource)?,
+                    )
+                } else {
+                    None
+                };
+
+                shielded_inputs = match orchard {
+                    Some(orchard) => {
+                        let orchard_value = orchard.total_value()?;
```
dannywillems: We never want to create Orchard transactions; it is even forbidden by the protocol from NU6.3.

zcash_client_sqlite/src/wallet/orchard.rs
dannywillems: Done. `999cd31246` adds `Note::pool()` (classifies version-3 Orchard notes as Ironwood, distinct from `Note::protocol` which still reports the cryptographic protocol), implements the `PoolType::Ironwood` arms of `Step::input_in_pool`/`input_count_in_pool`, and switches the Orchard/Sapling arms to classify by pool so Orchard no longer counts Ironwood notes. The test now delegates to `Step::input_count_in_pool`. Lifted one level further to `Proposal::input_count_in_pool` in `7e49ba6c2a`, with helper property tests added in `fe8c003c89`.


Generated by Claude.

zcash_client_sqlite/src/wallet/orchard.rs
dannywillems: Done in `999cd31246`: the pool tallies are now named `orchard`/`ironwood` instead of `v2`/`v3`.

Generaetd by Claude.

zcash_client_sqlite/src/wallet/orchard.rs
dannywillems: Done in `999cd31246`: renamed to `orchard`/`ironwood` here and at the remaining call site.


Generated by Claude.

dannywillems:
## Orchard-spend change must stay in the Orchard pool (privacy fix, in progress)

### Property

When a transaction spends from the Orchard pool and produces change, the change must stay in the Orchard pool. Example: spend a single 2 ZEC Orchard note, pay 1 ZEC to an Ironwood recipient; the ~1 ZEC change must stay in Orchard and only the 1 ZEC payment crosses the turnstile into Ironwood.

### Current behavior (violates the property)

When Ironwood is active, all Orchard-pool outputs (payments *and change*) are routed into the Ironwood bundle. The change routing is explicit at `zcash_client_backend/src/data_api/wallet.rs`:

```rust
if orchard_outputs_are_ironwood {
    // Orchard V3 forbids cross-address outputs, so change to the internal change
    // address cannot be created there; route it to the Ironwood bundle instead.
    builder.add_ironwood_output(internal_ovk, change_address, change_value, memo)?;
```

So today the full 2 ZEC leaves Orchard, instead of only the 1 ZEC payment.

### Protocol constraint (shapes the fix)

Post-NU6.3 the Orchard-pool bundle is `orchard_v3()`, which mandates `disableCrossAddress` (verified in `orchard-0.15`):

```rust
// permits_cross_address_transfers() is false ONLY for (V3, ValuePool::Orchard)
assert!(!BundleVersion::orchard_v3().permits_cross_address_transfers());
```

`disableCrossAddress` requires, per action, that the output note's address equal the spent note's address. So Orchard change can stay in Orchard, but only if sent back to the same address as the spent Orchard note, not to the wallet's internal change address. That is precisely the purpose of `disableCrossAddress`: spend Orchard, send the payment out, keep the change at the same Orchard address without a turnstile crossing. The spent note's address is available at build time (`orchard::Note::recipient()`).

### Plan (confirmed)

1. Build Orchard-pool change via `add_orchard_output` to the spent note's own address (not `Scope::Internal`), never routing Orchard change to Ironwood.
2. Fee/action accounting: count Orchard change as an Orchard action. `orchard_change_to_ironwood` in `compute_balance` goes away for Orchard change (Orchard change never routes to Ironwood); the Sapling+Ironwood input group produces genuine Ironwood change instead.
3. Scope: applies to the Orchard-input group only; the Sapling+Ironwood group is unaffected.

Confirmed design points: change returns to the same address as the spent Orchard note (the only protocol-legal option); for multi-address Orchard spends the change goes to one spent note's address (whichever the builder pairs it with). Property tests will cover the single-note case from the example and the multi-note case.

zcash_client_sqlite/src/wallet/orchard.rs
dannywillems: I instructed Claude to change this. It is working on it now.

dannywillems: See https://github.com/zcash/librustzcash/pull/2539#issuecomment-4879450372

zcash_client_backend/src/wallet.rs
```diff
@@ -582,6 +624,22 @@ impl Note {
...
+    ///
+    /// Unlike [`Note::protocol`], which reports the note's cryptographic protocol (Sapling or
+    /// Orchard), this classifies version-3 Orchard notes as belonging to the Ironwood pool.
+    /// Sapling and version-2 Orchard notes report their own pool.
+    pub fn pool(&self) -> ShieldedPool {
+        match self {
+            Note::Sapling(_) => ShieldedPool::Sapling,
+            #[cfg(feature = "orchard")]
+            Note::Orchard(n) => match n.version() {
+                orchard::note::NoteVersion::V3 => ShieldedPool::Ironwood,
```
nuttycom: We should not be relying exclusively on the note version for the pool determination. Instead, the note should carry the `ShieldedPool` value directly, as we did in zcash_client_sqlite when we separated the tables.

dannywillems:
## Orchard-spend change now stays in the Orchard pool (implemented)

Follow-up to the analysis above. Implemented in `db7429df72` (+ changelog `96ba3ae43f`).

### Behavior

When a transaction spends Orchard notes with Ironwood active, the Orchard-pool change stays in the Orchard pool; only the payment crosses the turnstile into Ironwood. The example holds: spend a 2 ZEC Orchard note, pay 1 ZEC to an Ironwood recipient, and the ~1 ZEC change is retained as an Orchard note.

### How the cross-address restriction is satisfied

`orchard_v3` mandates `disableCrossAddress`, so ordinary Orchard outputs are rejected there. The change is returned to a spent note's own address and added through the builder's dedicated change entry point, `add_orchard_change_output`, which fabricates a same-address paired spend. Using `add_orchard_output` there fails with `CrossAddressDisabled`; the dedicated path is required.

### Changes

- `data_api/wallet.rs`: capture the spent Orchard note's address before the spend loop consumes the inputs; the change loop selects on `(orchard_outputs_are_ironwood, orchard_change_recipient)` — same-address Orchard change when spending Orchard, route-to-Ironwood when there is no Orchard spend, or pre-NU6.3 internal-address Orchard change. Same-address change can land at an external-scoped address, so recorded Orchard change is decrypted with the external IVK as well as the internal one.
- `input_selection.rs`: `orchard_change_to_ironwood = orchard_outputs_are_ironwood && orchard_inputs.is_empty()`, passed to `compute_balance` so the proposal fee matches the built transaction.

### Tests

`orchard_payment_routes_to_ironwood_and_builds` now asserts, at the built-transaction level, that the wallet owns no Ironwood note and that the change is retained as an Orchard note of value `note - payment - fee`. All four Ironwood privacy invariant proptests pass at 128 cases. Clippy (`--all-features --all-targets`) and `cargo fmt --check` are clean; the full `zcash_client_sqlite` lib suite is green.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: This is accurate; the unified address contains an orchard protocol receiver, but the received note's value is tracked in the Ironwood pool.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1316,8 +1316,9 @@ where
     } else {
         OrchardBuildMode::Orchard
     };
-    // Orchard-pool payment and change outputs are routed to the Ironwood bundle
-    // in exactly the Ironwood spend and Ironwood output modes.
+    // Orchard-pool *payment* outputs are routed to the Ironwood bundle in exactly the Ironwood
```
nuttycom:
```suggestion
    // Payment outputs to Orchard receivers are routed to the Ironwood bundle in exactly the Ironwood
```

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1567,6 +1568,14 @@ where
         )?;
     }
 
+    // Post-NU6.3 the Orchard bundle enforces the cross-address restriction, so Orchard-pool
+    // change must be returned to a spent Orchard note's own address to remain in the Orchard pool
+    // (rather than crossing the turnstile into Ironwood). Capture that address before the spend
+    // loop consumes `orchard_inputs`; `None` means no Orchard note is spent, in which case any
+    // Orchard-pool change is routed into the Ironwood bundle instead.
+    #[cfg(feature = "orchard")]
+    let orchard_change_recipient = orchard_inputs.first().map(|(note, _)| note.recipient());
```
nuttycom: This is incorrect; the recipient for the change output should be the default address derived from the internal full viewing key of the account corresponding to the spent note. Then, the `TransactionBuilder::add_orchard_change_output` will be called instead of `add_orchard_output`; this switch should be dependent upon whether or not NU6.3 is active.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1915,42 +1924,73 @@ where
                         .orchard()
                         .cloned()
                         .ok_or(Error::KeyNotAvailable(PoolType::ORCHARD))?;
-                    let change_address =
-                        orchard_fvk.address_at(0u32, orchard::keys::Scope::Internal);
```
nuttycom, line 1919: This was correct and should not be changed.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1915,42 +1924,73 @@ where
...
-                        ))
+                    match (orchard_outputs_are_ironwood, orchard_change_recipient) {
+                        // Post-NU6.3 with an Orchard spend: the change stays in the Orchard pool,
+                        // returned to a spent note's own address so it satisfies the Orchard V3
+                        // cross-address restriction. Only the payment crosses into Ironwood. The
+                        // Orchard V3 bundle forbids ordinary outputs, so the change is added via
+                        // `add_orchard_change_output`, which pairs it with a fabricated same-address
+                        // spend.
+                        (true, Some(change_address)) => {
+                            builder.add_orchard_change_output(
```
nuttycom: This is correct, the code being replaced was wrong.

zcash_client_backend/src/data_api/wallet.rs
nuttycom: In looking more at this I'm going to do a bunch more refactoring around how we handle wallet outputs and notes, to ensure that the separation between "note received in a pool" and "protocol-level note" is clearer. Right now we're dispatching on the note plaintext version in a bunch of places that we shouldn't be. Hopefully I'll have something that's reviewable for you for when you wake up; if not, I'll post incomplete work to a branch but with enough type-level refactoring that you with an agent can propagate the changes through.

zcash_client_backend/src/data_api/wallet.rs
nuttycom, line 1: The preexisting `OrchardBuildMode` handling in this file is wrong; if there are any Ironwood spends or outputs, an Ironwood anchor *must* be chosen from the tree, otherwise anchor selection reveals that the spends are dummies.

zcash_client_backend/src/data_api/wallet.rs
dannywillems: Agreed, it looked incorrect to me too.

zcash_client_backend/src/fees/common.rs
```diff
@@ -275,6 +275,20 @@ where
...
+    // balances), so `change_pool` can only be Ironwood if the caller configured it as
+    // the fallback change pool for a fully-transparent (shielding) flow. Shielding
+    // directly into Ironwood is not supported, and `OutputManifest` cannot represent an
+    // Ironwood change output, so reject it here rather than panicking on the assertion
+    // below.
+    if change_pool == ShieldedPool::Ironwood {
+        return Err(ChangeError::BundleError(
+            "Shielding change directly into the Ironwood pool is not supported.",
+        ));
+    }
```
nuttycom: This is incorrect. Change should go to the ironwood pool when ironwood funds are spent. The surrounding machinery needs to be updated to allow this.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1771,7 +1771,12 @@ where
                     add_transparent_output(&mut builder, &mut transparent_output_meta, to)?;
                 }
                 PoolType::Shielded(ShieldedPool::Ironwood) => {
-                    todo!("Ironwood pool support is not yet implemented")
+                    // Payments to a unified address are never assigned the Ironwood output
+                    // pool: Ironwood notes are delivered to the Orchard receiver (the Orchard
+                    // arm above routes them into the Ironwood bundle when Ironwood is active),
+                    // so step construction only ever assigns Sapling, Orchard, or Transparent
+                    // here. Reaching this arm means the proposal is internally inconsistent.
+                    return Err(Error::ProposalNotSupported);
```
nuttycom, line 1785: This seems incorrect; the proposal should represent the outputs being created in Ironwood; the whole point of a proposal is that a user can inspect it and understand its effects.

zcash_client_backend/src/data_api/wallet/input_selection.rs
nuttycom: Not true; we have to be able to create Orchard input transactions in order to migrate funds out of Orchard. Also this library should be usable before the NU6.3 activation.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1649,47 +1805,36 @@ impl<DbT: InputSource> ShieldingSelector for GreedyInputSelector<DbT> {
...
+                    }
+                }
+                // Unreachable: `resolve_shielded_destination` rejects transparent
+                // destinations earlier with `ShieldingRequiresShieldedRecipient`.
+                _ => {
+                    return Err(InputSelectorError::Proposal(
+                        ProposalError::ShieldingRequiresShieldedRecipient,
+                    ));
+                }
+            };
```
nuttycom, line 2003: Why does this not have a branch for PoolType::IRONWOOD?

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -332,12 +332,20 @@ pub(crate) fn parse_note_version(code: i64) -> Option<NoteVersion> {
...
+/// version is recorded in the `note_version` column but does not influence table selection. This
+/// lets the scanner route its separate Orchard and Ironwood output streams to their respective
+/// tables while the decrypted-transaction path selects the table from the note version.
 pub(crate) fn put_received_note<
     T: ReceivedOrchardOutput<AccountId = AccountUuid>,
     P: consensus::Parameters,
 >(
     conn: &rusqlite::Transaction,
     params: &P,
+    table_prefix: &str,
```
nuttycom: nit: instead of a string table prefix, I would prefer to pass the `ShieldedPool` and then use the existing function to get the table prefix within the scope of `put_received_note`

zcash_client_sqlite/src/wallet.rs
nuttycom: This will change in a later commit when `Note::Orchard` is refactored to contain the pool.

dannywillems:
I'm reverting to previous changes from @nuttycom. My local main wasn't up-to-date when rebasing.

dannywillems:
> I'm reverting to previous changes from @nuttycom. My local main wasn't up-to-date when rebasing.

Rebasing now on top of https://github.com/zcash/librustzcash/commit/ea2992595d944d2fddc0d38ef24e242be7e037ba, current `origin/main`.

zcash_client_backend/src/data_api/testing.rs
nuttycom, line 686: Why is `initial_ironwood_tree_size` not passed here?

zcash_client_backend/src/data_api/testing.rs
nuttycom: The initial ironwood tree size should be an argument like the passed-in orchard and sapling tree sizes.

zcash_client_backend/src/scanning.rs
nuttycom: I think it's fine; I like to have the type aliases for those things that are protocol-consistent, even though the pool-association should be strictly separated at the type level.

zcash_client_sqlite/src/wallet/orchard.rs
nuttycom: I later decided that this was overly restrictive; the condition has now been simplified to the consensus rule "orchard valuePoolBalance must be nonnegative" - funds must exit the Orchard pool after NU6.3

zcash_client_sqlite/src/wallet/orchard.rs
nuttycom: I don't understand why we are formatting then string matching instead of pattern matching against the actual error type.

zcash_client_sqlite/src/wallet/orchard.rs
nuttycom: I think that "The Orchard spend must remain in the Orchard bundle" is correct, this isn't related to change, this is ensuring that the spend of an Orchard note is not mistakenly represented as a spend in the Ironwood bundle (an easy protocol-confusion to make). However, this check is not testing what the message asserts; it's just checking the presence of the Orchard bundle, not the contents of that bundle.

zcash_client_sqlite/src/wallet.rs
nuttycom: Yes. If `tx.ironwood_bundle()` is not feature-gated, then the outer check should not be feature-gated.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1056,6 +1051,8 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                             sapling: use_sapling,
                             #[cfg(feature = "orchard")]
                             orchard: use_orchard,
+                            #[cfg(feature = "orchard")]
+                            ironwood: false,
```
nuttycom: Verify that this is replaced with a `use_ironwood` value by the end of the commit chain.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1412,6 +1409,8 @@ where
         sapling: use_sapling,
         #[cfg(feature = "orchard")]
         orchard: use_orchard,
+        #[cfg(feature = "orchard")]
+        ironwood: false,
```
nuttycom: Verify that this is replaced by `use_ironwood` by the time the PR is complete.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -934,8 +1005,6 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                             &orchard_fees::EmptyBundleView,
                             #[cfg(feature = "orchard")]
                             &orchard_fees::EmptyBundleView,
-                            #[cfg(feature = "orchard")]
```
nuttycom, line 937: I don't understand this removal; `compute_balance` should also have an argument for Ironwood.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -959,8 +1028,6 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                         &orchard_fees::EmptyBundleView,
                         #[cfg(feature = "orchard")]
                         &orchard_fees::EmptyBundleView,
-                        #[cfg(feature = "orchard")]
```
nuttycom, line 962: I don't understand this removal; this should also have an argument for Ironwood.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1144,23 +1202,36 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                 Err(other) => return Err(InputSelectorError::Change(other)),
             }
 
+            // Select from every pool that is spendable at the target height; the
+            // pool-usage trimming at the top of the loop decides which of the selected
+            // notes are actually spent. The turnstile constraint on post-Ironwood Orchard
+            // spends is enforced by the change strategy and by proposal validation, so no
+            // structural restriction on combining pools is needed here.
```
nuttycom: This means that one has to do non-local reasoning when reading the code, which is a common source of bugs. The choice of selectable pools here should be consistent with elsewhere: pool crossing should be minimized; when NU6.3 is active, outputs to Orchard protocol receivers are constructed in the Ironwood bundle, resulting in value moving to the Ironwood pool.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1258,31 +1372,52 @@ where
...
-        orchard_fees::transactional_action_count(
-            orchard_bundle_version_for_height(params, target_height),
-            spendable_notes.orchard.len(),
-            requested_orchard_actions,
-        )
-        .map_err(|e| InputSelectorError::Change(ChangeError::BundleError(e)))?
-    };
+    let recipient_wants_orchard = recipient.can_receive_as(PoolType::ORCHARD);
+    #[cfg(feature = "orchard")]
+    let route_orchard_output_to_ironwood = ironwood_active_at(params, target_height);
```
nuttycom: Rename `route_orchard_output_to_ironwood` is a confusing name, because it's not an Orchard pool output, it's an Ironwood pool output. Rename to `orchard_receivers_fill_ironwood`.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1258,31 +1372,52 @@ where
...
+            orchard::bundle::Flags::ENABLED,
+            spendable_notes.ironwood.len(),
+            usize::from(recipient_wants_orchard && route_orchard_output_to_ironwood),
+        )
+        .map_err(|s| InputSelectorError::Change(ChangeError::BundleError(s)))?;
+    #[cfg(not(feature = "orchard"))]
     let ironwood_action_count: usize = 0;
 
+    #[cfg(feature = "orchard")]
+    let use_ironwood = ironwood_action_count > 0;
```
nuttycom: Instead of `use_ironwood`, `use_sapling`, `use_orchard` I would recast these as `ironwood_bundle_required` etc.

zcash_client_backend/src/fees/common.rs
```diff
@@ -830,6 +855,9 @@ pub(crate) fn check_for_uneconomic_inputs<NoteRefT: Clone, E>(
             transparent: t_allowed + t_extra,
             sapling: s_allowed + s_extra,
             orchard: o_allowed + o_extra,
+            // Ironwood inputs are not passed to this economic check, so no Ironwood dust
+            // allocation is modelled here.
+            ironwood: 0,
```
nuttycom: This is indicative of missing functionality that we need to fix.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4698,13 +4709,19 @@ pub fn truncate_to_chain_state_below_birthday<T: ShieldedPoolTester, Dsf>(
                 .map(|root| CommitmentTreeRoot::from_parts(birthday_height - 100, root))
                 .collect::<Vec<_>>();
 
+            // Ironwood is not active at these test heights, so its tree is empty.
+            #[cfg(feature = "orchard")]
+            let ironwood_initial_tree = Frontier::empty();
```
nuttycom, line 4715: Follow-up: we need these tests to be updated to work correctly after Ironwood activation.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1183,15 +1252,58 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+/// Ironwood (when active) and then Orchard — an Orchard-receiver payment is delivered
+/// via the Ironwood bundle once Ironwood is active — and for other payments it is
+/// Sapling. The legacy Orchard pool comes last otherwise, so that it is drawn upon
+/// only when the more current pools cannot cover the required amount.
+#[cfg(feature = "orchard")]
+fn selectable_pool_preference<ParamsT: consensus::Parameters>(
+    params: &ParamsT,
+    target_height: TargetHeight,
+    sapling_supported: bool,
+    orchard_supported: bool,
```
nuttycom, line 1402: The meaning of these arguments is not clear to me. Is "x_supported" referring to the protocol or the pool? Clarify the documentation.

zcash_client_backend/src/decrypt.rs
```diff
@@ -191,59 +199,134 @@ pub fn decrypt_transaction<'a, P: consensus::Parameters, AccountId: Copy>(
...
+    }
+
+    // Trial-decrypt an Ironwood bundle. Ironwood notes are Orchard-shaped and use the account's
+    // Orchard viewing keys, but they carry version-3 note plaintexts and so must be decrypted
+    // under [`IronwoodDomain`]; decrypting them under [`OrchardDomain`] would silently detect
+    // nothing. Ironwood is a pool distinct from Orchard. This mirrors `decrypt_orchard_bundle`
+    // above; the note, key, and action types are shared, only the domain and pool differ.
     #[cfg(feature = "orchard")]
-    let orchard_outputs = orchard_bundle
+    fn decrypt_ironwood_bundle<AccountId: Copy>(
```
nuttycom: `decrypt_orchard_bundle` and `decrypt_ironwood_bundle` should be a single function, since at the _protocol_ level orchard bundles and ironwood bundles are equivalent in trial decryption; the pool (if necessary) should be passed as an argument.

zcash_client_backend/src/proposal.rs
```diff
@@ -518,12 +575,60 @@ impl<NoteRef> Step<NoteRef> {
...
+                .filter(|(_, pool)| **pool == PoolType::ORCHARD)
+                .map(|(idx, _)| {
+                    transaction_request
+                        .payments()
+                        .get(idx)
+                        .and_then(|p| p.amount())
+                        .expect("all payments previously checked to have amount values")
+                })
+                .try_fold(Zatoshis::ZERO, |acc, a| acc + a)
+                .ok_or(ProposalError::Overflow)?;
```
nuttycom: This should always be zero, because with ironwood active, by the time we get here, we should already know that all these outputs are going to Orchard. As a consequence, we should simply assert that if ironwood is active (which is established above)`!payment_pools.iter().any(|(_, pool)| **pool == PoolType::ORCHARD)` - the *only* orchard outputs are change, which are handled below.

nuttycom:
Shifted back to draft to address problems Fable found.

zcash_client_backend/src/scanning/compact.rs
```diff
@@ -28,6 +28,9 @@ use orchard::{
     tree::MerkleHashOrchard,
 };
 
+#[cfg(feature = "orchard")]
```
nullcopy, line 31: These changes are making me wonder if Ironwood logic be gated under a new Ironwood feature instead of `orchard`? I guess it depends if the `orchard` feature is gating the protocol or the pool. FMV the pool is the feature to be gated, but that's just my opinion.

e.g. maybe I want to use librustzcash to build a post-NU6.3 tool which is only aware of Ironwood pool

nullcopy: Not blocking this PR, but want to raise it for discussion

zcash_client_backend/src/scanning/compact.rs
nullcopy, line 202: (nit) I would prefer to rename `tx.actions` to `tx.orchard_actions` to avoid ambiguity and be consistent with `tx.ironwood_actions`

zcash_client_backend/src/scanning.rs
```diff
@@ -210,6 +210,22 @@ impl<AccountId> ScanningKeyOps<OrchardDomain, AccountId, orchard::note::Nullifie
...
+/// Ironwood can share the underlying Orchard primitives without being modelled as part of an
+/// "Orchard family"; the two pools are kept strictly separate throughout the scanner.
+///
+/// [ZIP 2005]: https://zips.z.cash/zip-2005
+#[cfg(feature = "orchard")]
+pub(crate) type IronwoodDomain = OrchardDomain;
+
+/// The nullifier type for Ironwood notes. See [`IronwoodDomain`].
+#[cfg(feature = "orchard")]
+pub(crate) type IronwoodNullifier = orchard::note::Nullifier;
```
nullcopy, line 228: (nb) The symmetry seeker in me wants to see a corresponding `OrchardNullifier` type alias

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -10,13 +10,20 @@ use schemerz_rusqlite::RusqliteMigration;
...
 pub(super) const MIGRATION_ID: Uuid = Uuid::from_u128(0x64925567_65ae_495e_b6cf_d5f56e99e422);
 
-const DEPENDENCIES: &[Uuid] = &[account_delete_cascade::MIGRATION_ID];
+// This migration invokes `block_max_scanned`, which reads the `ironwood_commitment_tree_size`
+// column of the `blocks` table; that column is added by `ironwood_shardtree`, so this migration
+// must run after it. (The analogous Orchard column is covered because `orchard_shardtree` is
+// already a transitive dependency.)
+const DEPENDENCIES: &[Uuid] = &[
+    account_delete_cascade::MIGRATION_ID,
+    ironwood_shardtree::MIGRATION_ID,
```
nullcopy, line 25: (nit) this requires a graph update in zcash_client_sqlite/src/wallet/init/migration.rs

zcash_client_backend/src/scanning.rs
```diff
@@ -212,20 +212,49 @@ impl<AccountId> ScanningKeyOps<OrchardDomain, AccountId, orchard::note::Nullifie
...
+/// The nullifier type for Ironwood notes. This is the Orchard nullifier type, which does not
+/// depend on the note plaintext version. See [`IronwoodDomain`].
 #[cfg(feature = "orchard")]
 pub(crate) type IronwoodNullifier = orchard::note::Nullifier;
 
+/// An Orchard viewing key trial-decrypts Ironwood outputs under the Ironwood note-encryption
+/// domain. The key material is the same as for Orchard; only the domain (and thus the accepted
+/// note plaintext version) differs.
+#[cfg(feature = "orchard")]
+impl<AccountId> ScanningKeyOps<IronwoodDomain, AccountId, orchard::note::Nullifier>
```
nullcopy, line 234: Shouldn't this be generic on the `IronwoodNullifier` type?

nullcopy, line 234: Would be nice to have a generic `OrchardFamily` trait to abstract these implementations which are common to both

zcash_client_backend/src/data_api/wallet.rs
nullcopy: Orchard & Ironwood inputs both pass into this closure, right? This disagrees with the `involves(Orchard)` filter above, no?

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1447,66 +1358,62 @@ where
...
-        OrchardBuildMode::IronwoodSpends => {
-            // `IronwoodSpends` is constructed only when `ironwood_spend_inputs` is
-            // `Some`, so the inputs are always present here.
-            let Some(inputs) = ironwood_spend_inputs else {
-                return Err(Error::ProposalNotSupported);
-            };
+    let (ironwood_anchor, ironwood_inputs) = if proposal_step
+        .involves(PoolType::Shielded(ShieldedPool::Ironwood))
+    {
+        let Some(inputs) = proposal_step.shielded_inputs() else {
```
nullcopy: Similar comment as above, won't this capture `Orchard` inputs, despite the apparent intent to filter above with `involves(Ironwood)`

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -1447,66 +1358,62 @@ where
...
+    // outputs is indistinguishable from one that spends real Ironwood notes.
     #[cfg(feature = "orchard")]
-    let (ironwood_anchor, ironwood_inputs) = match orchard_build_mode {
-        OrchardBuildMode::IronwoodSpends => {
-            // `IronwoodSpends` is constructed only when `ironwood_spend_inputs` is
-            // `Some`, so the inputs are always present here.
-            let Some(inputs) = ironwood_spend_inputs else {
-                return Err(Error::ProposalNotSupported);
-            };
+    let (ironwood_anchor, ironwood_inputs) = if proposal_step
```
nullcopy: (nit) the shape of this Ironwood section does not match the Orchard section, which begs the question: Why are they different if the logic is ostensibly identical?

nullcopy (commented):
Flushing comments as of 562d5c30dcd9046bc3bd268fbb8ee34e86f36013

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -1180,15 +1267,58 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
...
+    prefer_orchard_family: bool,
+) -> Vec<ShieldedPool> {
+    let ironwood_selectable = orchard_supported && ironwood_active_at(params, target_height);
+    let mut preference = Vec::with_capacity(3);
+    if prefer_orchard_family {
+        if ironwood_selectable {
+            preference.push(ShieldedPool::Ironwood);
```
nullcopy, line 1409: Wouldn't it be better to prefer Orchard first, to eagerly select Orchard coins to be moved to Ironwood?

nullcopy (approved):
utACK 52a19f0ec046b622fbd618b6434e0bdd4d896117

PR #2540, opened 2026-07-04
Optional derived Orchard PCZT fields

czarcas7ic:
## Summary

Shrinks the PCZT a producer sends an air-gapped signer.

- **Recomputable derived fields:** 6 derived Orchard fields and the bundle anchor become optional in v2, so the producer can drop them from the wire and the signer recomputes them byte-identically via `fill_derived_fields()`. `out_ciphertext` stays required because it is RNG-derived and not recomputable. v1 is unchanged and rejects redacted PCZTs with `RequiresV2`.

Every recomputed field is asserted byte-identical to the never-redacted PCZT, preserving the same sighash and signature, for Orchard V2 and Ironwood V3.


pczt/src/lib.rs
```diff
@@ -225,19 +233,18 @@ pub mod v2 {
     }
 
     /// Encodes the in-memory [`super::Pczt`] into the v2 serialization type [`Pczt`],
-    /// omitting empty Transparent, Sapling, and Orchard bundles.
-    impl TryFrom<super::Pczt> for Pczt {
-        type Error = super::EncodingError;
-
-        fn try_from(pczt: super::Pczt) -> Result<Self, Self::Error> {
```
czarcas7ic, line 232: Now that v2 orchard bundle does not re use the v1 wire type due to the optional fields, v2 has fields that are all optional which cannot fail, so "try from" is no longer needed.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -350,6 +353,34 @@ impl BuildConfig {
...
+) -> orchard::builder::BundleType {
+    if bundle_version == orchard::bundle::BundleVersion::ironwood_v3() {
+        orchard::builder::BundleType::Transactional {
+            bundle_required: false,
+            pad_to_minimum: false,
+        }
+    } else {
+        orchard::builder::BundleType::DEFAULT
+    }
+}
```
czarcas7ic: If we agree on orchard also not needing the padding as well, we could remove this entirely and just set BundleType::DEFAULT accordingly

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -754,17 +754,18 @@ mod tests {
...
+        // Ironwood output: + ironwood (1) = 5 actions. At 5000 zat/action that is
+        // 20000 vs 25000.
         assert_eq!(
             without_ironwood.fee_required(),
             Zatoshis::const_from_u64(20000)
         );
         assert_eq!(
             with_ironwood.fee_required(),
-            Zatoshis::const_from_u64(30000)
+            Zatoshis::const_from_u64(25000)
```
czarcas7ic: Goes down due to no padding requirement

PR #2548, opened 2026-07-06
Make PCZT v2 anchor and cv_net optional

czarcas7ic:
PCZT v2 Orchard bundle encoding now represents all-zero bundle anchors and per-action cv_net values as absent optional fields. Missing values decode back to all-zero logical fields, so the in-memory model stays unchanged while the wire format can omit default data.

Test plan: cargo test -p pczt --all-features

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -546,13 +546,15 @@ pub(crate) mod v2 {
         }
     }
 
+    const ZERO_BYTES_32: [u8; 32] = [0; 32];
+
     /// PCZT fields that are specific to producing the transaction's Orchard bundle.
     #[derive(Clone, Debug, Serialize, Deserialize, Getters)]
     pub struct Bundle {
         actions: Vec<Action>,
         flags: u8,
         value_sum: (u64, bool),
-        anchor: [u8; 32],
+        anchor: Option<[u8; 32]>,
         note_version: SerializedNoteVersion,
         zkproof: Option<Vec<u8>>,
         bsk: Option<[u8; 32]>,
@@ -561,7 +563,7 @@ pub(crate) mod v2 {
     /// Information about an Orchard action within a transaction.
     #[derive(Clone, Debug, Serialize, Deserialize)]
     pub(crate) struct Action {
-        cv_net: [u8; 32],
+        cv_net: Option<[u8; 32]>,
         spend: v1::Spend,
         output: Output,
         rcv: Option<[u8; 32]>,
@@ -597,7 +599,7 @@ pub(crate) mod v2 {
                     .collect::<Vec<_>>(),
                 flags: bundle.flags,
                 value_sum: bundle.value_sum,
-                anchor: bundle.anchor,
+                anchor: (bundle.anchor != ZERO_BYTES_32).then_some(bundle.anchor),
                 note_version: bundle.note_version.into(),
                 zkproof: bundle.zkproof,
                 bsk: bundle.bsk,
@@ -615,7 +617,7 @@ pub(crate) mod v2 {
                     .collect(),
                 flags: bundle.flags,
                 value_sum: bundle.value_sum,
-                anchor: bundle.anchor,
+                anchor: bundle.anchor.unwrap_or(ZERO_BYTES_32),
                 note_version: bundle.note_version.into(),
                 zkproof: bundle.zkproof,
                 bsk: bundle.bsk,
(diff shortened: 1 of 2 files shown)


pczt/src/orchard.rs
```diff
@@ -545,18 +546,47 @@ pub(crate) mod v2 {
...
+        anchor: Option<[u8; 32]>,
         note_version: SerializedNoteVersion,
         zkproof: Option<Vec<u8>>,
         bsk: Option<[u8; 32]>,
     }
 
+    /// Information about an Orchard action within a transaction.
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub(crate) struct Action {
+        cv_net: Option<[u8; 32]>,
```
czarcas7ic, line 566: Action copied to make cv_net optional field

pczt/src/orchard.rs
```diff
@@ -545,18 +546,47 @@ pub(crate) mod v2 {
...
+        rcv: Option<[u8; 32]>,
+    }
+
+    /// Information about the output part of an Orchard action.
+    #[serde_as]
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub(crate) struct Output {
+        cmx: [u8; 32],
+        ephemeral_key: [u8; 32],
+        enc_ciphertext: Vec<u8>,
```
czarcas7ic, line 578: In the next PR, we will be modifying enc_ciphertext type in order to save space in PCZTs when transferring to Keystone so we only need to send the memo over the wire when this is an output we know, rather than always including the full enc_ciphertext.

ValarDragon (approved):

pczt/src/orchard.rs
```diff
@@ -587,14 +617,72 @@ pub(crate) mod v2 {
...
                 zkproof: bundle.zkproof,
                 bsk: bundle.bsk,
             }
         }
     }
 
+    impl From<super::Action> for Action {
+        fn from(action: super::Action) -> Self {
+            Self {
+                cv_net: (action.cv_net != ZERO_BYTES_32).then_some(action.cv_net),
```
dannywillems, line 631: Nit: `Some(action.cv_net)` would always have the same effect.

pczt/src/orchard.rs
```diff
@@ -587,14 +617,72 @@ pub(crate) mod v2 {
...
+                output: Output::from(action.output),
+                rcv: action.rcv,
+            }
+        }
+    }
+
+    impl From<Action> for super::Action {
+        fn from(action: Action) -> Self {
+            Self {
+                cv_net: action.cv_net.unwrap_or(ZERO_BYTES_32),
```
dannywillems, line 642: Review note: we match `None -> [0;32]`, and `[0;32]` on the other side has two representations, precisely `None and `Some([0; 32])`.
The encoding from v2 to v1 has two encodings; these are not bijective methods. It is fine.

pczt/src/orchard.rs
dannywillems: Review notes: it makes total sense to move it here as we're in the nested module v2, and it is exactly a change for a v2 of pczt.

pczt/src/orchard.rs
```diff
@@ -545,18 +546,47 @@ pub(crate) mod v2 {
         }
     }
 
+    const ZERO_BYTES_32: [u8; 32] = [0; 32];
```
dannywillems, line 549: Nit: this is used later as a default value for cv_net and anchor between v1 and v2. This could be renamed, even if two values are defined, to have more semantic context for the reviewers/readers.

dannywillems (approved):
utACK 80d9017eb8bae44ba4b1c1642eef5fd466b9b693.

Discussed in person with Dave and Adam.
This is only a follow-up from the larger patch submitted last week changing PCZT to support Ironwood.

pczt/src/orchard.rs
ValarDragon: Fixed in PR that makes cv_net None in the logical type as well

pczt/src/orchard.rs
```diff
@@ -587,14 +617,72 @@ pub(crate) mod v2 {
...
+            Self {
+                cv_net: action.cv_net.unwrap_or(ZERO_BYTES_32),
+                spend: super::Spend::from(action.spend),
+                output: super::Output::from(action.output),
+                rcv: action.rcv,
+            }
+        }
+    }
+
+    impl From<super::Output> for Output {
```
dannywillems, line 650: This exists here because `Output` is already redefined above, and the semantic changes will appear later.

PR #2550, opened 2026-07-07
zcash_client_backend: Support creation of transparent change

### LINKED ISSUE #1839: zcash_client_backend: Support creation of transparent change.
In order to provide full zcashd wallet replacement capabilities, `zallet` needs to be able to function as a fully-transparent wallet. This includes the creation of transparent change.

schell:
Closes #1839.

Adds support for returning change to the transparent pool when the net flows of a transaction under construction are fully transparent, enabling `zallet` to replicate `zcashd`'s `z_sendmany` behavior for fully-transparent spends. Builds on the transparent-spending support for `propose_transaction` added in #2422.

All new API surface is gated on the `transparent-inputs` feature flag.

## Design

- **`fees::TransparentChangePolicy`** — a new policy enum with variants `ShieldChange` (the default, preserving existing behavior) and `TransparentChangeAllowed`. Configured on the ZIP 317 and fixed change strategies via new `with_transparent_change_policy` builder methods. The policy only takes effect when the transaction's net flows are fully transparent; transactions involving shielded flows always shield their change, so no shielded change information is ever leaked to the transparent pool.
- **`ChangeValue::transparent(value)`** — a new non-ephemeral transparent change value, distinct from the ephemeral (ZIP 320) transparent output value. Transparent change is always emitted as a single output (the note-splitting policy applies only to shielded change), the exact-balance no-change case is preserved, zero-valued transparent outputs are never created, and the existing `DustOutputPolicy` machinery applies. The proposal protobuf encoding is wire-compatible: transparent change is encoded with the existing transparent `valuePool` and `isEphemeral` unset, a combination that previously failed to decode.
- **`WalletWrite::reserve_next_n_internal_addresses`** — reserves BIP 44 internal-scope (change) transparent addresses, parallel to the existing ephemeral reservation API. The `zcash_client_sqlite` implementation reuses the existing scope-generic reservation machinery and internal-scope gap limit; no migration is required, since internal-scope gap addresses are already generated at account creation and received change is recorded via the existing `Recipient::InternalTransparent` handling.
- `create_proposed_transactions` (and the PCZT construction/extraction paths) now realize transparent change outputs by reserving internal-scope addresses and recording the outputs against the receiving account.

Because general (non-shielding) transfers exclude coinbase UTXOs from input selection (`CoinbaseFilter::NonCoinbaseOnly`), transparent change can never be produced by a transaction that spends coinbase outputs, structurally satisfying `zcashd`'s coinbase-spend-forbids-change rule.

## Testing

- Fees-level unit tests covering the transparent-change branch, the exact-match no-change case, the shielded-flows no-op, non-splitting of transparent change, and dust rejection.
- End-to-end tests of the full t→t-with-change lifecycle: proposal, proposal serialization round trip, transaction creation, internal-scope address reservation/exposure, and post-mining spendability of the change output at the change address.
- Gap-limit enforcement tests for internal-scope address reservation, including independence from ephemeral-scope reservations.
- Verified with `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, workspace checks across feature combinations (`--no-default-features`, `transparent-inputs` alone and with `orchard`), intra-doc link validation, and the full `zcash_client_sqlite` test suite (319 passed).

## AI disclosure

The commits in this PR were prepared with AI assistance (Claude / opencode) and carry `Co-Authored-By` trailers accordingly. I have reviewed and take responsibility for all changes.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/transparent.rs
+++ b/zcash_client_backend/src/data_api/testing/transparent.rs
@@ -2441,3 +2441,293 @@ where
         .expect("AllFunds gather should succeed");
     assert_eq!(all.len(), n_dust);
 }
+
+/// Tests that [`WalletWrite::reserve_next_n_internal_addresses`] reserves sequential
+/// internal-scope (change) addresses, that reservation observes the internal-scope gap
+/// limit, and that internal-scope reservations are accounted independently of
+/// ephemeral-scope reservations.
+///
+/// This test expects the data store to be configured with the default gap limits, under
+/// which the internal-scope gap limit is 5.
+///
+/// The `is_reached_gap_limit` predicate must return `true` if and only if the provided
+/// error is the backend's exact "reached gap limit" error variant, the scope reported by
+/// that error is [`TransparentKeyScope::INTERNAL`], and the address index reported by that
+/// error equals the provided expected index. It must not match any other error. (A
+/// predicate is used because this test cannot name the backend's concrete error type
+/// without inverting the crate dependency.)
+pub fn reserve_next_n_internal_addresses_gap_limit<DSF>(
+    dsf: DSF,
+    cache: impl TestCache,
+    is_reached_gap_limit: impl Fn(
+        &<DSF::DataStore as crate::data_api::WalletRead>::Error,
+        DSF::AccountId,
+        u32,
+    ) -> bool,
+) where
+    DSF: DataStoreFactory,
+{
+    use std::collections::HashSet;
+    use transparent::keys::NonHardenedChildIndex;
+
+    let mut st = TestBuilder::new()
+        .with_data_store_factory(dsf)
+        .with_block_cache(cache)
+        .with_account_from_sapling_activation(BlockHash([0; 32]))
+        .build();
+    let account_id = st.test_account().cloned().unwrap().id();
+
+    // Seed the chain so that a chain height is known; address reservation records the
+    // exposure height of each reserved address.
+    let not_our_key = ExtendedSpendingKey::master(&[]).to_diversifiable_full_viewing_key();
+    let not_our_value = Zatoshis::const_from_u64(10000);
+    let (start_height, _, _) =
+        st.generate_next_block(&not_our_key, AddressType::DefaultExternal, not_our_value);
+    st.scan_cached_blocks(start_height, 1);
+
+    // Reserving internal addresses yields distinct, sequentially-indexed addresses derived
+    // under the internal (change) key scope.
+    let reserved = st
+        .wallet_mut()
+        .reserve_next_n_internal_addresses(account_id, 3)
+        .unwrap();
+    assert_eq!(reserved.len(), 3);
+    for (i, (_, meta)) in reserved.iter().enumerate() {
+        assert_eq!(meta.scope(), Some(TransparentKeyScope::INTERNAL));
+        assert_eq!(
+            meta.address_index(),
+            Some(NonHardenedChildIndex::const_from_index(
+                u32::try_from(i).unwrap()
+            )),
+        );
+    }
+    // None of the reserved addresses have received funds, so the gap cannot advance: with
+    // the default internal-scope gap limit of 5, only two more addresses may be reserved.
+    // Reservation continues at the next sequential indices, so the returned addresses are
+    // distinct from those of the first batch.
+    let more = st
+        .wallet_mut()
+        .reserve_next_n_internal_addresses(account_id, 2)
+        .unwrap();
+    assert_eq!(more.len(), 2);
+    for (i, (_, meta)) in more.iter().enumerate() {
+        assert_eq!(meta.scope(), Some(TransparentKeyScope::INTERNAL));
+        assert_eq!(
+            meta.address_index(),
+            Some(NonHardenedChildIndex::const_from_index(
+                u32::try_from(reserved.len() + i).unwrap()
+            )),
+        );
+    }
+    let unique_addrs = reserved
+        .iter()
+        .chain(more.iter())
+        .map(|(a, _)| *a)
+        .collect::<HashSet<_>>();
+    assert_eq!(unique_addrs.len(), reserved.len() + more.len());
+
+    assert_matches!(
+        st.wallet_mut().reserve_next_n_internal_addresses(account_id, 1),
+        Err(e) if is_reached_gap_limit(&e, account_id, 5)
+    );
+
+    // Internal-scope reservations must not consume ephemeral-scope gap space.
+    let ephemeral = st
+        .wallet_mut()
+        .reserve_next_n_ephemeral_addresses(account_id, 1)
+        .unwrap();
+    assert_eq!(ephemeral[0].1.scope(), Some(TransparentKeyScope::EPHEMERAL),);
+    assert_eq!(
+        ephemeral[0].1.address_index(),
+        Some(NonHardenedChildIndex::const_from_index(0)),
+    );
+}
+
+/// Tests the full lifecycle of a t->t transfer with transparent change: a change strategy
+/// configured with [`TransparentChangePolicy::TransparentChangeAllowed`] must propose a
+/// non-ephemeral transparent change output, and transaction creation must send that change
+/// to a previously-unexposed internal-scope (change) transparent address of the spending
+/// account, where it is recorded as received and becomes spendable once mined.
+///
+/// [`TransparentChangePolicy::TransparentChangeAllowed`]: crate::fees::TransparentChangePolicy::TransparentChangeAllowed
+pub fn propose_t2t_with_transparent_change<DSF>(dsf: DSF, cache: impl TestCache)
+where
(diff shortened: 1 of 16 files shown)


dannywillems (dismissed):
We want to have https://github.com/zcash/librustzcash/pull/2539 first, as discussed with @nuttycom. Only requesting changes to block the merge - not requesting actual changes for now.

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -2440,3 +2440,272 @@ where
...
+    let unique_addrs = reserved.iter().map(|(a, _)| *a).collect::<HashSet<_>>();
+    assert_eq!(unique_addrs.len(), reserved.len());
+
+    // None of the reserved addresses have received funds, so the gap cannot advance: with
+    // the default internal-scope gap limit of 5, only two more addresses may be reserved.
+    let more = st
+        .wallet_mut()
+        .reserve_next_n_internal_addresses(account_id, 2)
+        .unwrap();
+    assert_eq!(more.len(), 2);
```
nullcopy, line 2512: Need to test that the two addresses returned here are unique from the previously returned addresses

zcash_client_backend/src/data_api/testing/transparent.rs
```diff
@@ -2440,3 +2440,272 @@ where
...
+
+    assert_matches!(
+        st.wallet_mut().reserve_next_n_internal_addresses(account_id, 1),
+        Err(e) if is_reached_gap_limit(&e, account_id, 5)
```
nullcopy, line 2531: Is this pattern guard the best way to assert the error reason? Why not match the exact error code?

e.g. if this returns an unexpected error AND `is_reached_gap_limit() ` trips, then this assert will pass when it shouldn't

zcash_client_backend/src/fees/common.rs
```diff
@@ -521,7 +554,11 @@ where
                         cfg.params,
                         BlockHeight::from(target_height),
                         transparent_input_sizes.clone(),
-                        transparent_output_sizes.clone(),
+                        transparent_output_sizes
+                            .clone()
+                            // Count the standard size of the P2PKH change output when change is
+                            // to be returned to the transparent pool.
+                            .chain(wants_transparent_change.then_some(P2PKH_STANDARD_OUTPUT_SIZE)),
```
nuttycom, line 570: There's  a potential gap here - what happens when the input being spent is a p2sh multisig address? In that case, the transparent change should be returned to the original p2sh multisig. This isn't necessary for the transparent address handling that's needed for `z_sendfromaccount` at the moment, but I think that this needs an issue filed for follow-up.

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -185,6 +185,12 @@ enum PcztRecipient<AccountId> {
     InternalAccount {
         receiving_account: AccountId,
     },
+    // This variant is placed at the end of the enum in order to preserve the encoded
+    // representation of the prior variants.
+    #[cfg(feature = "transparent-inputs")]
+    InternalTransparent {
```
nullcopy, line 192: Does `InternalTransparent` need to be distinct from `InternalAccount'?

zcash_client_backend/src/fees.rs
```diff
@@ -64,9 +64,39 @@ impl FeeRule for StandardFeeRule {
...
+    ///
+    /// This is the default policy.
+    #[default]
+    ShieldChange,
+    /// When the net flows of the transaction are fully transparent, change is returned to the
+    /// transparent pool at an internal-scope (change) transparent address of the wallet, as
+    /// described in [BIP 44].
+    ///
+    /// [BIP 44]: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
+    TransparentChangeAllowed,
```
nullcopy, line 97: If shielded actions are included in the transaction, will change be shielded then?

zcash_client_backend/src/fees/common.rs
```diff
@@ -587,6 +629,22 @@ where
                 NonZeroU64::new(u64::try_from(split_count).expect("usize fits into u64")).unwrap(),
             );
             let simple_case = || {
+                #[cfg(feature = "transparent-inputs")]
+                if wants_transparent_change {
+                    return (
+                        if total_change.is_zero() {
```
nullcopy, line 644: Will this allow creation of dust outputs? Should we omit dust values here?

nullcopy: Actually, this decision needs to be made according to a user specified `DustOutputPolicy`

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -957,6 +1003,262 @@ mod tests {
...
+            result,
+            Ok(balance) if
+                balance.proposed_change() == [ChangeValue::transparent(Zatoshis::const_from_u64(649_0000))] &&
+                balance.fee_required() == Zatoshis::const_from_u64(10000)
+        );
+    }
+
+    #[test]
+    #[cfg(feature = "transparent-inputs")]
+    fn transparent_change_rejects_dust() {
```
nullcopy, line 1354: Dust should not always be rejected. Depends on `DustOutputPolicy`. We should test the different permutations of dust output policies

nullcopy (dismissed):
I have some questions that may need to become follow-up issues, but nothing blocking for now.

utACK 8e370ef1f22fc6633f7959de8cab79197c8ebfa4

zcash_client_backend/src/data_api/testing/transparent.rs
schell: Good catch!

zcash_client_backend/src/data_api/testing/transparent.rs
schell: The indirection is deliberate, though the reason is non-obvious in that the specialization lives at the call site. This test is in `zcash_client_backend`, which can't name `SqliteClientError` (that would invert the crate dependency), so instead it takes a predicate, which is the same pattern as `send_multi_step_proposed_transfer`. The sqlite-side closure matches the exact variant, scope, and index, so an unexpected error can't trip it. I've documented that contract on the test parameter.

zcash_client_backend/src/data_api/wallet.rs
schell: Yes! `InternalAccount` is converted via `into_recipient_with_note` and carries a shielded `Note`, while transparent change goes through the outpoint path and must record the on-chain-observable recipient address. This variant is just the construction-time mirror of the pre-existing `Recipient::InternalTransparent` / `Recipient::InternalAccount` distinction. There is some rationale on `Recipient::InternalTransparent` in `zcash_client_backend/src/wallet.rs`.

zcash_client_backend/src/fees/common.rs
schell: Yeah, that's a gap. I've created https://github.com/zcash/librustzcash/issues/2570 to track it.

zcash_client_backend/src/fees/common.rs
schell: Indeed :) That's what happens :) By the time `simple_case` runs, the dust decision has already been made by matching on the `DustOutputPolicy::action`.

zcash_client_backend/src/fees.rs
schell: Yes :) Change is always shielded whenever the transaction has any shielded inputs or outputs. `wants_transparent_change` in `fees/common.rs` requires `net_flows.is_transparent()`, so the policy only ever takes effect for fully-transparent flows. This is covered by the `transparent_change_policy_has_no_effect_on_shielded_flows` test.

zcash_client_backend/src/data_api/wallet.rs
nullcopy: Hmm, the naming is confusing, because `InternalTransparent` is internal and belonging to the account. Let's not block here, but later we should revisit the naming and perhaps even the deliniation of those two variants. I suspect there's an axis of symmetry we can refactor on here.

zcash_client_backend/src/data_api/testing/transparent.rs
nullcopy: We have some backend defined errors, and I wonder if this would be better as one. Not worth blocking on, but something we should log for a future decision. Maybe just a follow-up issue for now

nullcopy (approved):
Let's create some issues to revisit later for possible optimizations, but this looks good to me!

utACK d739b42733e6c0037e25b376a262ebe948ebd928

zcash_client_backend/src/data_api/testing/transparent.rs
schell: Filed as #2594.

zcash_client_backend/src/data_api/wallet.rs
schell: Filed as #2595.

PR #2551, opened 2026-07-07
Add opt-in unpadded Orchard-pool bundles

czarcas7ic:
Makes the transactional Orchard/Ironwood bundle type configurable instead of flipping the default: `BuildConfig::Standard` gains `orchard_pool_bundle_type`, the ZIP 317 change strategies gain `with_unpadded_orchard_pool_bundles()`, and `create_pczt_from_proposal` takes the matching bundle type. Defaults are unchanged (padded, same fees as main); a pool-migration wallet can opt in to build a single-output migration as 1 action per bundle. Requires the orchard revision that keeps `BundleType::DEFAULT` padded and adds `DEFAULT_UNPADDED`.

It also re-exports the `orchard` crate from `zcash_primitives` so crates that keep their own `orchard` dependency optional can still name the new field's type.

### DIFF
--- a/zcash_client_backend/src/fees/zip317.rs
+++ b/zcash_client_backend/src/fees/zip317.rs
@@ -100,7 +100,7 @@ impl<R, I> SingleOutputChangeStrategy<R, I> {
     /// bundle to the 2-action minimum.
     ///
     /// The transaction executing the proposal must be built with the matching bundle
-    /// type ([`BundleType::DEFAULT_UNPADDED`](orchard::builder::BundleType)), or the
+    /// type ([`BundleType::UNPADDED`](orchard::builder::BundleType)), or the
     /// builder's balance check will fail. Intended for transactions whose shape is
     /// already public (e.g. pool migrations); see the orchard `pad_to_minimum`
     /// documentation for the privacy trade-off.
@@ -163,7 +163,7 @@ where
 
         #[cfg(feature = "orchard")]
         let orchard_pool_bundle_type = if self.unpadded_orchard_pool_bundles {
-            ::orchard::builder::BundleType::DEFAULT_UNPADDED
+            ::orchard::builder::BundleType::UNPADDED
         } else {
             ::orchard::builder::BundleType::DEFAULT
         };
@@ -238,7 +238,7 @@ impl<R, I> MultiOutputChangeStrategy<R, I> {
     /// bundle to the 2-action minimum.
     ///
     /// The transaction executing the proposal must be built with the matching bundle
-    /// type ([`BundleType::DEFAULT_UNPADDED`](orchard::builder::BundleType)), or the
+    /// type ([`BundleType::UNPADDED`](orchard::builder::BundleType)), or the
     /// builder's balance check will fail. Intended for transactions whose shape is
     /// already public (e.g. pool migrations); see the orchard `pad_to_minimum`
     /// documentation for the privacy trade-off.
@@ -306,7 +306,7 @@ where
 
         #[cfg(feature = "orchard")]
         let orchard_pool_bundle_type = if self.unpadded_orchard_pool_bundles {
-            ::orchard::builder::BundleType::DEFAULT_UNPADDED
+            ::orchard::builder::BundleType::UNPADDED
         } else {
             ::orchard::builder::BundleType::DEFAULT
         };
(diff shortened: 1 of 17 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -322,10 +331,12 @@ impl BuildConfig {
...
         match self {
             BuildConfig::Standard {
-                ironwood_anchor, ..
+                ironwood_anchor,
+                orchard_pool_bundle_type,
+                ..
             } => ironwood_anchor.as_ref().map(|a| {
                 orchard::builder::Builder::new(
-                    orchard::builder::BundleType::DEFAULT,
+                    *orchard_pool_bundle_type,
```
ValarDragon, line 335: Why are we deref'ing a pointer?

ValarDragon: Something seems wrong with typesz:         orchard_pool_bundle_type: orchard::builder::BundleType

yet we have to deref?

czarcas7ic: Since this method matches on `&self`, the field binds as `&BundleType`.

`Builder::new` takes `BundleType` by value, and `BundleType` is `Copy`, so the deref just copies it.

ValarDragon: ok sg

zcash_client_backend/src/fees/zip317.rs
```diff
@@ -197,9 +227,26 @@ impl<R, I> MultiOutputChangeStrategy<R, I> {
...
+    /// The transaction executing the proposal must be built with the matching bundle
+    /// type ([`BundleType::DEFAULT_UNPADDED`](orchard::builder::BundleType)), or the
+    /// builder's balance check will fail. Intended for transactions whose shape is
+    /// already public (e.g. pool migrations); see the orchard `pad_to_minimum`
+    /// documentation for the privacy trade-off.
+    #[cfg(feature = "orchard")]
+    pub fn with_unpadded_orchard_pool_bundles(mut self) -> Self {
+        self.unpadded_orchard_pool_bundles = true;
+        self
+    }
```
ValarDragon, line 249: Does this really need to be a pub?

czarcas7ic: We use in Vizor here

https://github.com/valargroup/vizor-wallet/blob/13a72baa0e81d6c419e85a8c228e8e4625b86080/rust/src/wallet/sync/send.rs#L5451-L5453

Its how wallets would set the flag we introduced.

dannywillems:
Note: this has to be read with the context of https://github.com/zcash/orchard/pull/527. I reviewed !527 (adding a field to have a count).

I'll review this after lunch.

zcash_primitives/src/transaction/builder.rs
```diff
@@ -260,6 +260,11 @@ pub enum BuildConfig {
         sapling_anchor: Option<sapling::Anchor>,
         orchard_anchor: Option<orchard::Anchor>,
         ironwood_anchor: Option<orchard::Anchor>,
+        /// The [`BundleType`](orchard::builder::BundleType) for the transactional
```
dannywillems: Usually, we don't add documentation for the field. I would prefer to see this removed for consistency.

zcash_primitives/src/lib.rs
```diff
@@ -26,3 +26,8 @@ pub mod block;
 pub(crate) mod encoding;
 pub mod merkle_tree;
 pub mod transaction;
+
+/// Re-export of the `orchard` crate, whose types appear in this crate's public API
```
dannywillems: See https://github.com/zcash/librustzcash/pull/2558/changes

dannywillems: Talked in person with Adam, I'll push on the branch directly.

dannywillems: Cherry-picked on this branch.

dannywillems (approved):
utACK 035614bf807c3206c70d25dfc628df3ac650b7e8.

dannywillems:
Wondering if we could have used a different design for the padding field. I feel there is a more elegant solution, but I would probably come back to this later when we have more time.

PR #2552, opened 2026-07-07
DNM: WIP: Make remaining Orchard PCZT fields optional

czarcas7ic:
DNM/WIP. This branch isolates the remaining PCZT Orchard optional-field work on a review base that matches the local stack.

It makes the remaining recomputable Orchard-shaped fields optional in the v2 wire encoding, adds `MemoKind` for elided `enc_ciphertext` reconstruction, fills derived fields before parsing/signing, and rejects anchor elision before v6.

Validation:
- `cargo fmt --all`
- `cargo test -p pczt --test end_to_end --all-features`
- `git diff --check`

### DIFF
--- a/pczt/src/lib.rs
+++ b/pczt/src/lib.rs
@@ -191,6 +191,14 @@ pub mod v1 {
 }
 
 /// Types and operations for the v2 Pczt encoding.
+///
+/// In the Orchard-shaped bundles of this encoding the derived fields (`cv_net`,
+/// `nullifier`, `rk`, `cmx`, `ephemeral_key`, `enc_ciphertext`) and the bundle
+/// `anchor` are optional, and each output carries an optional
+/// [`MemoKind`](crate::orchard::MemoKind) tag. A producer elides those fields with the
+/// Redactor to shrink the encoding, and the receiver recomputes them with
+/// `Pczt::fill_derived_fields` (which requires the `orchard` feature; parsing the
+/// bundles also performs the fill implicitly).
 pub mod v2 {
     use alloc::vec::Vec;
     use serde::{Deserialize, Serialize};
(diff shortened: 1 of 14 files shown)


pczt/src/roles/verifier/orchard.rs
```diff
@@ -16,6 +16,7 @@ impl super::Verifier {
 
         let bundle = orchard
             .into_parsed_with_version(
+                global.tx_version,
```
czarcas7ic, line 19: We pass the tx version since leaving out the anchor is only valid in v6 transactions.

pczt/src/lib.rs
```diff
@@ -225,19 +233,18 @@ pub mod v2 {
...
     /// Encodes the in-memory [`super::Pczt`] into the v2 serialization type [`Pczt`],
-    /// omitting empty Transparent, Sapling, and Orchard bundles.
-    impl TryFrom<super::Pczt> for Pczt {
-        type Error = super::EncodingError;
-
-        fn try_from(pczt: super::Pczt) -> Result<Self, Self::Error> {
-            Ok(Self {
+    /// omitting empty Transparent, Sapling, and Orchard bundles. Infallible: the v2
+    /// encoding can represent every in-memory PCZT.
+    impl From<super::Pczt> for Pczt {
```
czarcas7ic, line 238: This is `From` rather than `TryFrom` because v2 now represents missing Orchard derived fields as Optional, so encoding no longer needs to reject them.

pczt/src/orchard.rs
```diff
@@ -404,41 +477,50 @@ pub mod v1 {
...
+            Ok(Self {
+                // This encoding has no representation for an elided derived field;
+                // reject any action that has redacted one (see
+                // [`crate::EncodingError::RequiresV2`]).
+                cv_net: action.cv_net.ok_or(crate::EncodingError::RequiresV2)?,
+                spend: Spend::try_from(action.spend)?,
+                output: Output::try_from(action.output)?,
                 rcv: action.rcv,
-            }
+            })
```
czarcas7ic, line 500: This is `TryFrom` rather than `From` because v1 cannot represent missing Orchard derived fields (not Optional), so encoding must reject redacted actions.

pczt/src/orchard.rs
```diff
@@ -962,14 +1175,247 @@ pub(crate) fn orchard_bundle_version(global: &crate::common::Global) -> Option<B
...
+    /// - An elided `anchor` is refilled with the fixed placeholder
+    ///   [`Anchor::empty_tree`](orchard::Anchor::empty_tree), not a recomputation, so
+    ///   it only reproduces a producer anchor that was already that constant. This
+    ///   method therefore accepts an elided anchor only when `tx_version` is the v6
+    ///   transaction format, where the anchor is excluded from the txid/sighash
+    ///   digest and a signer sees no difference; the extracting wallet must install
+    ///   the real anchor.
+    ///
+    /// On error the bundle may be left partially filled.
+    pub fn fill_derived_fields(&mut self, tx_version: u32) -> Result<(), FillError> {
```
czarcas7ic, line 1278: One of the primary functions to review here, which keystone call to fill in the respective elided fields.

PR #2555, opened 2026-07-07
Add v2 memo plaintext ciphertext resolution

ValarDragon:
Represent Orchard encrypted note plaintext as either ciphertext or stripped memo plaintext in the logical PCZT model, and resolve compact fields before Orchard role parsing.

We plan for more compact fields to get resolved like this before Orchard role parsing. So what I did was add a mutative "resolve_fields" function, for resolving fields that have been redacted, but are needed for Orchard calls.

We add a function in redactor to swap all Orchard ciphertexts with memo's, for all memo's that are decryptable via the action struct.

This saves us ~570 bytes/action in serialization to Keystone's, which is very significant. It halves the size relative to the state after making Anchor optional. This reduces ~4 QR frames per action, and were talking about 90-150 action flows.

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -228,12 +228,6 @@ impl EncCiphertext {
     }
 }
 
-impl From<Vec<u8>> for EncCiphertext {
-    fn from(ciphertext: Vec<u8>) -> Self {
-        EncCiphertext::Encrypted(ciphertext)
-    }
-}
-
 #[cfg(feature = "orchard")]
 fn recover_memo_plaintext_from_ciphertext_and_action(
     action: &Action,
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -4,6 +4,7 @@ use alloc::collections::BTreeMap;
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
+use core::fmt;
 
 #[cfg(feature = "orchard")]
 use ff::PrimeField;
@@ -12,6 +13,9 @@ use getset::Getters;
 use orchard::bundle::BundleVersion;
 #[cfg(feature = "orchard")]
 pub(crate) use orchard::note::NoteVersion;
+use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
+#[cfg(feature = "orchard")]
+use zcash_note_encryption::{Domain, ENC_CIPHERTEXT_SIZE, EphemeralKeyBytes, ShieldedOutput};
 
 use crate::{
     common::{Global, Zip32Derivation},
@@ -86,6 +90,9 @@ pub(crate) const ORCHARD_SPENDS_AND_OUTPUTS_ENABLED: u8 = 0b0000_0011;
 /// and the flag value of an empty bundle for serialization purposes.
 pub(crate) const IRONWOOD_SPENDS_OUTPUTS_AND_CROSS_ADDRESS_ENABLED: u8 = 0b0000_0111;
 
+/// The size in bytes of the memo portion of an Orchard note plaintext.
+pub(crate) const MEMO_SIZE: usize = 512;
+
 /// The canonical empty Orchard-pool bundle: the form the Orchard slot of a PCZT takes
 /// when it carries no Orchard-protocol data. The Creator, the v1 decoder, and the v2
 /// decoder all produce exactly this value for an absent bundle, so that copies of a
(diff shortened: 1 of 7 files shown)


pczt/src/orchard.rs
```diff
@@ -962,8 +1318,94 @@ pub(crate) fn orchard_bundle_version(global: &crate::common::Global) -> Option<B
...
 impl Bundle {
+    /// Resolves every derived or compact field representation this bundle
+    /// carries.
+    ///
+    /// For improved efficiency, callers that will pass the same bundle through
+    /// multiple roles should call this once up front, not in each role. Parsing
+    /// also resolves fields defensively.
+    pub fn resolve_fields(&mut self) -> Result<(), ::orchard::pczt::ParseError> {
+        self.resolve_memo_plaintexts()
+    }
```
ValarDragon, line 1447: I want to add `cv_net` here in a follow-up PR.

pczt/src/roles/redactor/orchard.rs
```diff
@@ -164,6 +164,33 @@ impl ActionRedactor<'_> {
         });
     }
 
+    /// Replaces the output's encrypted note plaintext with a stripped memo
+    /// plaintext.
+    ///
+    /// The receiver can recompute `enc_ciphertext` from this memo, the output note
```
czarcas7ic: Should this say the sender?

pczt/src/orchard.rs
```diff
@@ -1177,7 +1633,9 @@ impl Bundle {
                     output: Output {
                         cmx: output.cmx().to_bytes(),
                         ephemeral_key: output.encrypted_note().epk_bytes,
-                        enc_ciphertext: output.encrypted_note().enc_ciphertext.to_vec(),
+                        enc_ciphertext: EncCiphertext::Encrypted(
+                            output.encrypted_note().enc_ciphertext.to_vec(),
+                        ),
```
ValarDragon, line 1697: Intentional, as we don't need to deal with "was this a more compressed memo representation". That is purely a redaction-time and serialization time concern.

pczt/src/orchard.rs
```diff
@@ -1070,11 +1516,21 @@ impl Bundle {
                 )
             }?;
 
+            action
+                .output
+                .resolve_memo_plaintext(note_version, spend.nullifier().to_bytes())?;
```
ValarDragon: Make this an assign, not a mutative operation

pczt/src/orchard.rs
```diff
@@ -962,8 +1318,94 @@ pub(crate) fn orchard_bundle_version(global: &crate::common::Global) -> Option<B
...
+#[cfg(feature = "orchard")]
+impl Output {
+    /// Recomputes `enc_ciphertext` from memo plaintext, if this output carries it.
+    ///
+    /// This requires the action's spend nullifier because the output note's
+    /// [`rho`](::orchard::note::Rho) is derived from it.
+    fn resolve_memo_plaintext(
+        &mut self,
+        note_version: NoteVersion,
+        spend_nullifier: [u8; 32],
```
ValarDragon: Make non-mutative

pczt/src/roles/redactor/orchard.rs
ValarDragon: I was thinking of it as receiver of the PCZT consumer, let me just say that instead

pczt/src/orchard.rs
ValarDragon: We discussed mut is actually fine I believe, just made the fn name clearer + explained what it edits in code comment.

pczt/src/orchard.rs
ValarDragon: The advantage of making this a mutative operation is actually that we get to avoid an excess 512 byte memo allocation - this appears in constrained devices, so seems worth the mutative API trade-off.

pczt/src/orchard.rs
```diff
@@ -111,6 +118,232 @@ pub(crate) const EMPTY_IRONWOOD: Bundle = Bundle {
...
+            EncCiphertext::MemoPlaintext(_) => None,
+        }
+    }
+}
+
+impl From<Vec<u8>> for EncCiphertext {
+    fn from(ciphertext: Vec<u8>) -> Self {
+        EncCiphertext::Encrypted(ciphertext)
+    }
+}
```
nuttycom: This seems like it should be a `TryFrom` and bound the length. Actually, why is this using a Vec as input to begin with? Shouldn't it be a fixed array?

ValarDragon: I agree, was actually a big point of confusion between @czarcas7ic and I. I expected it to always be a 580 byte array, but it was a vec in the code so we didn't update that.

Happy to do that tho

pczt/src/orchard.rs
```diff
@@ -111,6 +118,232 @@ pub(crate) const EMPTY_IRONWOOD: Bundle = Bundle {
...
+                write!(f, "memo plaintext exceeds {MEMO_SIZE} bytes")
+            }
+            MemoPlaintextError::NotStripped => {
+                write!(f, "memo plaintext has trailing zero bytes")
+            }
+        }
+    }
+}
+
+/// A memo plaintext with all trailing zero bytes stripped.
```
dannywillems, line 144: It is important to note that there is an invariant that the plaintext _maximum_ length (i.e. the size of the inner vector) is `MEMO_SIZE`. All constructors MUST verify this invariant, which is the case at the moment, as it is not enforced by the type.

The second constructor, `from_stripped_bytes`, can construct vectors of any size below `MEMO_SIZE`.

czarcas7ic:
As requested by nuttycom, lets only merge this and next PR until https://github.com/zcash/librustzcash/pull/2539 is merged @dannywillems, we can still get PR approved / ready though

dannywillems (dismissed):
Blocking merge until 2539 is merged, see https://github.com/zcash/librustzcash/pull/2555#issuecomment-4915131094

dannywillems:
> As requested by nuttycom, lets only merge this and next PR until #2539 is merged @dannywillems, we can still get PR approved / ready though

I didn't know. Explicitly requesting changes to avoid anyone else merging, in case of other async review/approvals.

pczt/src/orchard.rs
nuttycom: This matches zcash_note_encryption

pczt/src/orchard.rs
```diff
@@ -111,6 +118,232 @@ pub(crate) const EMPTY_IRONWOOD: Bundle = Bundle {
...
+        let esk = D::derive_esk(note)?;
+
+        try_output_recovery_with_pkd_esk(domain, pk_d, esk, output)
+            .map(|(_, _, memo)| MemoPlaintext::from_memo(memo))
+    }
+
+    let enc_ciphertext = match &action.output.enc_ciphertext {
+        EncCiphertext::Encrypted(ciphertext) => ciphertext.as_slice().try_into().ok()?,
+        // we return None here to avoid excess sets or clone operations, as the caller need not do anything in this case.
+        EncCiphertext::MemoPlaintext(_) => return None,
```
nuttycom, line 285: Why isn't this returning the passed-in memo plaintext returned here? Why are we returning None instead?

nuttycom: That would make the return value non-Optional

nuttycom: Answer from @ValarDragon: it would require a clone.

nuttycom (approved):

PR #2557, opened 2026-07-08
Allow redacting Orchard anchor and cv_net

ValarDragon:
## Summary

This PR builds on `dev/pczt-v2-memo-plaintext` and updates PCZT v2 Orchard/Ironwood handling so `anchor` and `cv_net` can be represented as redacted logical fields.

- Represents Orchard bundle anchors as `Option<[u8; 32]>` in the logical PCZT type.
- Represents Orchard action `cv_net` as `Option<[u8; 32]>` and resolves it from note values plus `rcv` when needed.
- Keeps v1 serialization strict: missing anchors or `cv_net` require v2.
- Makes missing-anchor parse behavior derive from transaction version, using `DEFAULT_ANCHOR` only as a parse-time placeholder.
- Preserves redacted anchors when roles serialize parsed bundles back to PCZT.
- Adds redactor helpers and end-to-end coverage for redacted Ironwood anchor / `cv_net` behavior.

## Role Behavior

Verifier, updater, signer, and low-level signer can continue operating with redacted anchors where they do not require the real value. IO finalizer and prover now explicitly reject missing anchors for non-empty bundles because those roles require the real anchor.

## Validation

- `cargo fmt --all`
- `cargo check -p pczt --features orchard`
- `cargo test -p pczt --features orchard v2_round_trips_memo_plaintext_ciphertext_data`
- `cargo test -p pczt --features orchard,internal-tests,spend-finalizer,zcp-builder --test end_to_end ironwood_low_level_signer_uses_preverified_signing_parse`
- `cargo test -p pczt --features orchard,internal-tests,spend-finalizer,zcp-builder --test end_to_end redacted_anchor_is_not_resolved`
- `git diff --check`

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -63,7 +63,7 @@ pub struct Bundle {
     ///
     /// Set by the Creator.
     #[getset(get = "pub")]
-    pub(crate) anchor: [u8; 32],
+    pub(crate) anchor: Option<[u8; 32]>,
 
     /// The note plaintext version for notes in this bundle.
     pub(crate) note_version: NoteVersion,
@@ -93,6 +93,8 @@ pub(crate) const IRONWOOD_SPENDS_OUTPUTS_AND_CROSS_ADDRESS_ENABLED: u8 = 0b0000_
 /// The size in bytes of the memo portion of an Orchard note plaintext.
 pub(crate) const MEMO_SIZE: usize = 512;
 
+pub(crate) const DEFAULT_ANCHOR: [u8; 32] = [0; 32];
+
 /// The canonical empty Orchard-pool bundle: the form the Orchard slot of a PCZT takes
 /// when it carries no Orchard-protocol data. The Creator, the v1 decoder, and the v2
 /// decoder all produce exactly this value for an absent bundle, so that copies of a
@@ -101,7 +103,7 @@ pub(crate) const EMPTY_ORCHARD: Bundle = Bundle {
     actions: Vec::new(),
     flags: ORCHARD_SPENDS_AND_OUTPUTS_ENABLED,
     value_sum: (0, false),
-    anchor: [0; 32],
+    anchor: None,
     note_version: NoteVersion::V2,
     zkproof: None,
     bsk: None,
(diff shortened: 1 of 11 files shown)


pczt/src/orchard.rs
```diff
@@ -63,7 +63,7 @@ pub struct Bundle {
     ///
     /// Set by the Creator.
     #[getset(get = "pub")]
-    pub(crate) anchor: [u8; 32],
+    pub(crate) anchor: Option<[u8; 32]>,
```
ValarDragon, line 66: Change the logical type

pczt/src/orchard.rs
```diff
@@ -93,6 +93,9 @@ pub(crate) const IRONWOOD_SPENDS_OUTPUTS_AND_CROSS_ADDRESS_ENABLED: u8 = 0b0000_
 /// The size in bytes of the memo portion of an Orchard note plaintext.
 pub(crate) const MEMO_SIZE: usize = 512;
 
+#[cfg(feature = "orchard")]
+const DEFAULT_ANCHOR: [u8; 32] = [0; 32];
```
ValarDragon: I specifically chose to use this, instead of the semantically valid default anchor thats some hash output.

Thats because _there is no_ valid usage of an all 0's anchor, hence feels safe for usage as default anchor when converting None's into some object for the Orchard crate

pczt/src/orchard.rs
```diff
@@ -1431,27 +1451,68 @@ impl Output {
...
+            NoteValue::from_raw(self.spend.value.ok_or(ParseError::InvalidValueCommitment)?);
+        let output_value = NoteValue::from_raw(
+            self.output
+                .value
+                .ok_or(ParseError::InvalidValueCommitment)?,
+        );
+        let rcv =
+            ValueCommitTrapdoor::from_bytes(self.rcv.ok_or(ParseError::InvalidValueCommitment)?)
+                .into_option()
+                .ok_or(ParseError::InvalidValueCommitment)?;
```
nuttycom, line 1477: pczt: Implement recovery of orchard cv_net from spend value, output value, rcv using an orchard library function.

We'd rather not recapitulate a part of the orchard protocol in the PCZT internals, so create an issue to refactor this later.

nuttycom: opened #2566

nuttycom (approved):
utACK modulo that my review of the tests was very light.

PR #2563, opened 2026-07-08
Allow redacting Sapling anchors

ValarDragon:
## Summary

- represent the logical Sapling anchor as optional so v2 PCZT redaction can remove it
- preserve v1 behavior by filling the default anchor only for missing-anchor bundles with no Sapling spends
- add Sapling anchor redaction API and v2 round-trip coverage

## Validation


nuttycom (approved):
utACK

PR #2565, opened 2026-07-08
Make Orchard v2 spend fields (nullifier, rk) options

ValarDragon:
## Summary

- add an Orchard v2 serialization-specific spend type
- encode spend nullifiers and rk values as optional fields in v2
- reject v2 encodings that omit either field until the logical Orchard spend type can represent their absence
- add coverage for missing nullifier and missing rk parse failures

## Validation

- git diff --check

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -764,7 +764,7 @@ pub(crate) mod v2 {
     use serde::{Deserialize, Serialize};
     use serde_with::serde_as;
 
-    use super::{NoteVersion, v1};
+    use super::NoteVersion;
 
     /// A serializable representation of Orchard note plaintext versions.
     #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


pczt/src/orchard.rs
```diff
@@ -801,11 +807,35 @@ pub(crate) mod v2 {
...
     }
 
+    /// Information about the spend part of an Orchard action.
+    #[serde_as]
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub(crate) struct Spend {
+        #[serde_as(as = "Option<[_; 32]>")]
+        nullifier: Option<[u8; 32]>,
+        #[serde_as(as = "Option<[_; 32]>")]
+        rk: Option<[u8; 32]>,
```
ValarDragon, line 822: These two became option, as the change over v1

nuttycom: Open an issue for the approach we discussed of supporting an enum containing either (nullifier, rk) or alpha in the logical type. Also with everything here now being optional we probably need to document better how to use that optionality.

nuttycom (approved):
utACK modulo documentation/issue requests.

pczt/src/orchard.rs
nuttycom: https://github.com/zcash/librustzcash/issues/2575

PR #2567, opened 2026-07-08
zcash_client_sqlite: Add feature-gated ZeWIF wallet import

nuttycom:
Adds a feature-gated (`zewif`) `zcash_client_sqlite::zewif` module that imports a
wallet from a [ZeWIF](https://github.com/zcash/zewif) (Zcash Wallet Interchange
Format) document into a `WalletDb`.

## Overview

`import_wallet` populates an already-initialized `WalletDb` from a
`zewif::Zewif` document. All spending-key material carried by the document is
diverted to a caller-provided `SecretSink` (for example, zallet's age-encrypted
keystore) and is **never** written to the wallet database; the database stores
only viewing keys and public metadata.

The import is lossy by design — a ZeWIF document can describe things the wallet
database cannot represent (Sprout accounts, non-multisig redeem scripts, …).
Rather than failing, those items are skipped and tallied in the returned
`ZewifImportReport`.

## Pipeline

1. Verify that every wallet in the document matches the database's network.
2. Deliver every secret-store entry to the `SecretSink`, indexing the seed and
   viewing-key material that account import can make use of.
3. Import each account: seed-derived accounts whose seed material is present are
   re-derived via HD (preserving the recorded ZIP 32 account index); all others
   are imported from their viewing keys — as spending accounts when the
   corresponding secret was delivered to the sink, and view-only otherwise.
   Account birthdays are reconstructed from tree-state frontiers where present.
4. Register standalone transparent spending keys and P2SH redeem scripts against
   the imported accounts that own their addresses.
5. Mark the transparent addresses recorded in the document as exposed, so that
   address-based recovery includes them.
6. Trial-decrypt and store the document's transactions; those not shown to
   involve any imported account are left for the post-import rescan to recover.

## Limitations

- Sprout accounts are not representable and are skipped (Sprout spending keys
  are still delivered to the sink).
- Address-book entries have no backing store in this crate and are not imported.
- Regtest documents are not currently supported.
- Encrypted secret stores must be decrypted by the caller before import.

## Notes

- Entirely behind the `zewif` feature (off by default); no effect on existing
  builds or the default dependency set.
- The `zewif` crate is currently pulled in as a `[patch.crates-io]` git
  dependency, with corresponding `deny.toml` and `cargo-vet` entries.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) assistance.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -39,6 +39,8 @@ exceptions = [
     { name = "foldhash", allow = ["Zlib"] },
     { name = "inotify", allow = ["ISC"] },
     { name = "inotify-sys", allow = ["ISC"] },
+    { name = "minicbor", allow = ["BlueOak-1.0.0"] },
+    { name = "minicbor-derive", allow = ["BlueOak-1.0.0"] },
     { name = "minreq", allow = ["ISC"] },
     { name = "nanorand", allow = ["Zlib"] },
     { name = "notify", allow = ["CC0-1.0"] },
--- a/zcash_client_sqlite/src/zewif.rs
+++ b/zcash_client_sqlite/src/zewif.rs
@@ -757,16 +757,31 @@ fn account_birthday<P: Parameters, S>(
     }
 }
 
-/// Delivers the document's secret store to the sink and indexes the material
-/// that account import can make use of.
-fn deliver_secrets<S: SecretSink>(
+/// Returns the document's plaintext secret store to draw secret material from,
+/// or an error if the store is encrypted. Delivery of that material to the sink
+/// is performed by the caller.
+fn plaintext_secret_store<E>(
     document: &::zewif::Zewif,
-) -> Result<Option<&::zewif::SecretStore>, ZewifImportError<S::Error>> {
+) -> Result<Option<&::zewif::SecretStore>, ZewifImportError<E>> {
     match document.secrets() {
         None => Ok(None),
         Some(::zewif::Secrets::Plain(store)) => Ok(Some(store)),
         Some(::zewif::Secrets::Encrypted(_)) => Err(ZewifImportError::EncryptedSecrets),
     }
+}
+
+/// Parses the ZIP 32 account index recorded by a derived key source, mapping an
+/// out-of-range value to [`ZewifImportError::InvalidAccountIndex`].
+fn derived_account_index<S>(
+    account_name: &str,
+    source: &::zewif::DerivedKeySource,
+) -> Result<zip32::AccountId, ZewifImportError<S>> {
+    zip32::AccountId::try_from(source.account_index()).map_err(|_| {
+        ZewifImportError::InvalidAccountIndex {
+            account_name: account_name.to_owned(),
+            index: source.account_index(),
+        }
+    })
 }
 
 /// Constructs the ZIP 32 derivation metadata recorded by a derived key source.
@@ -775,12 +790,7 @@ fn zip32_derivation<S>(
     source: &::zewif::DerivedKeySource,
 ) -> Result<Zip32Derivation, ZewifImportError<S>> {
     let seed_fp = decode_seed_fingerprint(source.seed_fingerprint().encoding())?;
-    let account_index = zip32::AccountId::try_from(source.account_index()).map_err(|_| {
-        ZewifImportError::InvalidAccountIndex {
-            account_name: account_name.to_owned(),
-            index: source.account_index(),
-        }
-    })?;
+    let account_index = derived_account_index(account_name, source)?;
     let legacy_address_index = source
         .legacy_address_index()
         .map(|i| {
@@ -842,8 +852,9 @@ where
 
     // Deliver all secret material to the sink, indexing what account import can
     // use.
+    let secret_store = plaintext_secret_store::<S::Error>(document)?;
     let mut available = AvailableSecrets::empty();
-    if let Some(store) = deliver_secrets::<S>(document)? {
+    if let Some(store) = secret_store {
         for entry in store.seeds() {
             sink.store_seed(entry).map_err(ZewifImportError::Sink)?;
             let (fingerprint, seed_bytes) = seed_entry_bytes(entry)?;
@@ -888,13 +899,7 @@ where
         }
     }
 
-    register_transparent_keys(
-        wdb,
-        &params,
-        deliver_secrets::<S>(document)?,
-        &taddrs,
-        &mut report,
-    )?;
+    register_transparent_keys(wdb, &params, secret_store, &taddrs, &mut report)?;
     mark_addresses_exposed(wdb, &params, &taddrs, &mut report)?;
 
     import_transactions(wdb, &params, document, &mut report)?;
@@ -913,16 +918,16 @@ struct TransparentAddressRecords {
     redeem_scripts: Vec<(AccountUuid, Vec<u8>)>,
 }
 
-/// Decodes a WIF-encoded transparent spending key, returning the secret key
-/// and whether the corresponding public key uses the compressed encoding.
-fn decode_wif(expected_prefix: u8, wif: &str) -> Option<(secp256k1::SecretKey, bool)> {
+/// Decodes a WIF-encoded transparent spending key, accepting both the
+/// uncompressed and the compressed (`0x01`-suffixed) payload forms.
+fn decode_wif(expected_prefix: u8, wif: &str) -> Option<secp256k1::SecretKey> {
     let payload = bs58::decode(wif).with_check(None).into_vec().ok()?;
     match payload.as_slice() {
         [prefix, key_data @ ..] if *prefix == expected_prefix && key_data.len() == 32 => {
-            Some((secp256k1::SecretKey::from_slice(key_data).ok()?, false))
+            secp256k1::SecretKey::from_slice(key_data).ok()
         }
         [prefix, key_data @ .., 0x01] if *prefix == expected_prefix && key_data.len() == 32 => {
-            Some((secp256k1::SecretKey::from_slice(key_data).ok()?, true))
+            secp256k1::SecretKey::from_slice(key_data).ok()
         }
         _ => None,
     }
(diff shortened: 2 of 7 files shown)


zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2107 @@
...
+    }
+
+    fn store_unified_key(&mut self, _entry: &::zewif::UnifiedKeyEntry) -> Result<(), Self::Error> {
+        Ok(())
+    }
+}
+
+/// The shielded pool to which an invalid tree frontier belongs.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FrontierPool {
```
nullcopy: Do we need to define this new enum? Can't we use the existing shielded pool enum?

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2107 @@
...
+    /// Seed bytes ready for ZIP 32 derivation, keyed by the canonical 32-byte
+    /// ZIP 32 seed fingerprint, so that textually-differing but equivalent
+    /// fingerprint encodings resolve to the same seed.
+    seeds: HashMap<[u8; 32], SecretVec<u8>>,
+    /// The canonical encodings of Sapling extended full viewing keys whose
+    /// spending keys were delivered to the sink.
+    sapling_fvks: Vec<String>,
+    /// The canonical encodings of unified full viewing keys whose spending keys
+    /// were delivered to the sink.
+    unified_fvks: Vec<String>,
```
nullcopy, line 543: Should we also enumerate Orchard FVKs, or do we only ever accept those as part of a UFVK?

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2107 @@
...
+                chain_state,
+                account
+                    .recover_until_height()
+                    .map(|h| BlockHeight::from(u32::from(h))),
+            ),
+            BirthdayBasis::ChainState,
+        ))
+    } else {
+        // Without a recorded tree state, we can only choose the height at which
+        // scanning begins; the note commitment trees will be rebuilt by the
```
nullcopy, line 721: As a middle-ground, we could estimate birthday from the mined height of any stored transaction involving this account, right?

nullcopy: Not blocking, but worth a follow up

nullcopy (commented):
Left a few non-blocking questions.

zcash_client_sqlite/src/zewif.rs
nuttycom: Those are only defined as part as UFVKs.

zcash_client_sqlite/src/zewif.rs
nuttycom: Usually a wallet receives its first transaction relatively soon after someone sets it up; I don't think this usually amounts to a reasonable savings, we can just use the birthday.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2107 @@
...
+/// The reason an account in a ZeWIF document was not imported.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AccountSkipReason {
+    /// The account's viewing capability is a Sprout viewing key; the Sprout pool
+    /// is not representable in the wallet database.
+    SproutViewingKey,
+    /// The account is a bare set of transparent addresses with no unified key
+    /// structure, and no seed material was available from which its contents
+    /// could be re-derived.
+    TransparentAddressSetWithoutSeed,
```
nuttycom, line 428: It doesn't seem to me that we should skip accounts on this basis; such an account is view-only but it should still be representable as an account. If structural inhibitions to this currently exist, we need to get rid of those inhibitions.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+}
+
+/// How the birthday of an imported account was determined.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BirthdayBasis {
+    /// The document carried the tree state prior to the account's birthday
```
dannywillems, line 434: Nit: we're not used to use comments for enum field. No need to fix it now. It is in the whole patch I think.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+    pub address_book_entries_not_imported: usize,
+}
+
+/// The secret material available for driving account import, indexed by the
+/// public identifiers accounts use to reference it.
+struct AvailableSecrets {
+    /// Seed bytes ready for ZIP 32 derivation, keyed by the canonical 32-byte
+    /// ZIP 32 seed fingerprint, so that textually-differing but equivalent
+    /// fingerprint encodings resolve to the same seed.
+    seeds: HashMap<[u8; 32], SecretVec<u8>>,
```
dannywillems, line 537: Nit: could be more typed. Can be done in follow-ups. No need for a first version.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+        return Err(ZewifImportError::SeedFingerprintMismatch {
+            claimed: claimed_encoding.to_owned(),
+        });
+    }
+    Ok((claimed, SecretVec::new(seed_bytes)))
+}
+
+/// Converts a BIP 39 mnemonic to its 64-byte seed, using the empty passphrase
+/// (as zcashd and zallet do).
+fn mnemonic_to_seed(m: &::zewif::Bip39Mnemonic) -> Result<Vec<u8>, bip0039::Error> {
```
dannywillems, line 603: This doesn't exist somewhere else already?

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+}
+
+/// Constructs an [`AccountBirthday`] for an account from the birthday
+/// information carried by the document.
+fn account_birthday<P: Parameters, S>(
+    params: &P,
+    account: &::zewif::Account,
+) -> Result<(AccountBirthday, BirthdayBasis), ZewifImportError<S>> {
+    let zewif_block_hash =
+        |h: Option<::zewif::BlockHash>| h.map_or(BlockHash([0; 32]), |h| BlockHash(*h.as_bytes()));
```
dannywillems, line 679: Minor: I don't know how the `None` case should be handled. I guess it is fine.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+    let params = wdb.params().clone();
+    let expected = params.network_type();
+
+    // Verify that every wallet in the document belongs to the expected network.
+    for wallet in document.wallets() {
+        match (wallet.network(), expected) {
+            (::zewif::Network::Mainnet, NetworkType::Main) => {}
+            (::zewif::Network::Testnet, NetworkType::Test) => {}
+            (::zewif::Network::Regtest(_), _) => {
+                return Err(ZewifImportError::UnsupportedRegtest);
```
dannywillems, line 839: Nit: Is it expected? We might want for integration-tests.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+struct TransparentAddressRecords {
+    /// Maps each transparent address string to the account that recorded it
+    /// and the height at which it is known to have been exposed.
+    owners: HashMap<String, (AccountUuid, BlockHeight)>,
+    /// The P2SH redeem scripts recorded on imported accounts' addresses.
+    redeem_scripts: Vec<(AccountUuid, Vec<u8>)>,
+}
+
+/// Decodes a WIF-encoded transparent spending key, accepting both the
+/// uncompressed and the compressed (`0x01`-suffixed) payload forms.
```
dannywillems, line 925: I would accept this hypothesis.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+        assert_eq!(
+            report.skipped_accounts[0].reason,
+            AccountSkipReason::SproutViewingKey
+        );
+    }
+
+    #[test]
+    fn secrets_are_delivered_to_the_sink() {
+        let (_file, mut wdb) = test_wallet_db();
+        let ts = test_seed(0);
```
dannywillems, line 1696: Nit: would be nice to use proptest. Good for follow-up.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+
+        let secp = secp256k1::Secp256k1::new();
+        let secret_key = secp256k1::SecretKey::from_slice(&[0x42; 32]).unwrap();
```
dannywillems, line 1699: Nit: would be nice to use proptest. Good for follow-up.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+        doc.add_transaction(txid, ::zewif::Transaction::new(txid));
+
+        let report = import_wallet(&mut wdb, &doc, &mut DiscardSecrets).unwrap();
+        assert_eq!(report.transactions_without_raw_data, 1);
+        assert_eq!(report.transactions_stored, 0);
+        assert_eq!(report.transactions_without_wallet_relevance, 0);
+    }
+
+    /// The raw seed for a [`TestSeed`], as ZIP 32 derivation consumes it.
+    fn test_seed_bytes(ts: &TestSeed) -> Vec<u8> {
```
dannywillems, line 1813: Could be a proptest. For a follow-up. Leaving comment for Claude.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -0,0 +1,2094 @@
...
+                purpose: AccountPurpose::Spending { .. },
+                ..
+            }
+        ));
+    }
+
+    /// Builds a transparent-only transaction paying `value` zatoshis to `to`
+    /// under the consensus rules in force at `height`, returning its txid and
+    /// raw bytes.
+    fn transparent_tx_to(
```
dannywillems, line 1938: I guess these should be existing somewhere else as a proptest.

deny.toml
```diff
@@ -39,6 +39,8 @@ exceptions = [
     { name = "foldhash", allow = ["Zlib"] },
     { name = "inotify", allow = ["ISC"] },
     { name = "inotify-sys", allow = ["ISC"] },
+    { name = "minicbor", allow = ["BlueOak-1.0.0"] },
```
dannywillems, line 42: LGTM

dannywillems (approved):
utACK. Left comments but only to keep track of thoughts.

zcash_client_sqlite/src/zewif.rs
nuttycom: We will want it, but I don't think we currently have zcashd wallet integration tests. Opened https://github.com/zcash/librustzcash/issues/2583

PR #2568, opened 2026-07-08
Support optional PCZT anchors and updater APIs

ValarDragon:
## Summary

- allow PCZT Creator to accept optional Sapling and Orchard anchors, requiring anchors only for non-empty v5 shielded bundles
- add updater APIs for restoring Sapling, Orchard, and Ironwood anchors after signing and before proof creation
- add updater APIs for restoring Sapling, Orchard, and Ironwood spend witnesses before proof creation
- extend PCZT roundtrip and end-to-end tests for redacted anchors and witnesses

## Validation

- cargo fmt --all -- --check
- cargo doc --no-deps -p pczt --all-features --document-private-items
- git diff --check

### DIFF
--- a/pczt/src/roles/updater/mod.rs
+++ b/pczt/src/roles/updater/mod.rs
@@ -59,10 +59,125 @@ impl Updater {
         }
     }
 
+    /// Sets spend witnesses for Sapling spends by spend index.
+    ///
+    /// This may be called after shielded signatures have been added. Sapling
+    /// spend proofs depend on the witnesses, so this must be called before proof
+    /// creation for the updated spends.
+    ///
+    /// Returns an error if any witness references a spend index that does not
+    /// exist, or if a target spend proof is already present.
+    #[cfg(feature = "sapling")]
+    pub fn set_sapling_spend_witnesses(
+        mut self,
+        witnesses: impl IntoIterator<Item = (usize, ::sapling::MerklePath)>,
+    ) -> Result<Self, SpendWitnessUpdateError> {
+        set_sapling_spend_witnesses(&mut self.pczt.sapling.spends, witnesses)?;
+
+        Ok(self)
+    }
+
+    /// Sets spend witnesses for Orchard actions by action index.
+    ///
+    /// This may be called after shielded signatures have been added. Orchard
+    /// proofs depend on the witnesses, so this must be called before proof
+    /// creation.
+    ///
+    /// Returns an error if any witness references an action index that does not
+    /// exist, if the Orchard bundle is not using Orchard note plaintexts, or if
+    /// an Orchard proof is already present.
+    #[cfg(feature = "orchard")]
+    pub fn set_orchard_spend_witnesses(
+        mut self,
+        witnesses: impl IntoIterator<Item = (usize, ::orchard::tree::MerklePath)>,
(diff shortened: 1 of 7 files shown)


pczt/src/roles/updater/mod.rs
```diff
@@ -59,12 +59,245 @@ impl Updater {
...
+    ///
+    /// Returns an error if the PCZT's transaction format does not support this
+    /// update, if any Sapling spend proof is already present, or if the PCZT
+    /// already contains a different Sapling anchor.
+    #[cfg(feature = "sapling")]
+    pub fn set_sapling_anchor(
+        mut self,
+        anchor: ::sapling::Anchor,
+    ) -> Result<Self, AnchorUpdateError> {
+        ensure_anchor_update_supported(&self.pczt.global)?;
```
nuttycom, line 77: nit: if the anchor's already set and equal to the provided anchor, we can skip these checks and just return Ok(self)

pczt/src/roles/updater/mod.rs
```diff
@@ -59,12 +59,245 @@ impl Updater {
...
+    ///
+    /// Returns an error if the PCZT's transaction format does not support this
+    /// update, if an Orchard proof is already present, or if the PCZT already
+    /// contains a different Orchard anchor.
+    #[cfg(feature = "orchard")]
+    pub fn set_orchard_anchor(
+        mut self,
+        anchor: ::orchard::Anchor,
+    ) -> Result<Self, AnchorUpdateError> {
+        ensure_anchor_update_supported(&self.pczt.global)?;
```
nuttycom, line 98: Same thing here, these checks could be skipped in the case the anchor is already set and isn't changing.

pczt/src/roles/updater/mod.rs
```diff
@@ -59,12 +59,245 @@ impl Updater {
...
+    ///
+    /// Returns an error if the PCZT's transaction format does not support this
+    /// update, if an Ironwood proof is already present, or if the PCZT already
+    /// contains a different Ironwood anchor.
+    #[cfg(feature = "orchard")]
+    pub fn set_ironwood_anchor(
+        mut self,
+        anchor: ::orchard::Anchor,
+    ) -> Result<Self, AnchorUpdateError> {
+        ensure_anchor_update_supported(&self.pczt.global)?;
```
nuttycom, line 119: ditto

pczt/tests/end_to_end.rs
```diff
@@ -1231,155 +1202,979 @@ fn ironwood_low_level_signer_uses_preverified_signing_parse() {
...
-                .sign(sighash, &orchard_ask, ChaCha20Rng::from_seed(seed))
-                .expect("signing succeeds");
-            Ok(())
+    let (spend_vk, output_vk) = sapling_prover.verifying_keys();
+    let tx = TransactionExtractor::new(proved)
+        .with_sapling(&spend_vk, &output_vk)
+        .extract()
+        .unwrap();
+    assert_eq!(u32::from(tx.expiry_height()), 10_000_040);
+}
```
nuttycom, line 1299: This diff was super confusing just because of the copy/move - as a follow-up we should have an agent factor out the commonalities of these tests.

nuttycom (approved):
utACK

PR #2572, opened 2026-07-09
zcash_pool_migration: Add an Orchard-to-Ironwood value-pool migration engine

Chlup:
Adds **`zcash_pool_migration`**, a new workspace crate implementing the Orchard→Ironwood value-pool migration engine for mobile wallets: a synchronous, network-free library that plans a note split into self-funding power-of-ten denominations, builds/proves/signs migration transactions as PCZTs (software and external-signer flows), schedules transfers by block height, persists its state in `ext_ironwood_migration_*` tables in the shared wallet DB, and hands the consuming application pre-signed transactions to broadcast. The iOS/Android SDK FFI layers wrap `MigrationContext` directly.

## Provenance

Behavior was proven in the vizor wallet's migration engine (Apache-2.0) and a standalone prototype crate built against the valargroup librustzcash fork; this is a from-scratch port onto current upstream APIs. Notable upstream-era changes vs. the prototype:

- **Transfers use the high-level pipeline** (`InputSelector::propose_transaction` with an explicit anchor and `TxVersion::V6` → `create_pczt_from_proposal(…, target_expiry_height)` → PCZT prove/sign/finalize). The note split keeps a direct-builder path because Orchard-receiver outputs deliberately route to Ironwood post-activation, while the split must mint Orchard-pool change.
- **No unstable cfg flags** (Ironwood is stabilized on main) and **no serde in the crate** (one dependency feature, `zcash_client_sqlite/serde`, is required by `create_pczt_from_proposal`'s `AccountId: serde::Serialize` bound).
- Retained-checkpoint pinning (`ensure_retained`) hardens anchors across multi-transfer signing loops.
- Run completion is persisted (restoring vizor behavior), so completed runs release note locks and later deposits can start fresh runs.

## Verification

- 95 unit tests, including real PCZT prove/sign round trips for the external-signer flow.
- 4 seeded-wallet end-to-end tests (via `zcash_client_backend::data_api::testing` + a file-backed sqlite fixture): the transfer pipeline produces a V6 transaction whose Ironwood bundle carries exactly the crossing value with the exact ZIP-317 self-funding fee (change falls back to Orchard); the note split signs against seeded notes; a migration is driven to persisted completion by mining the extracted transaction; transfer execution is height-gated. This includes runtime validation of the unified `PostNu6_3` proving key for both Orchard and Ironwood bundles.
- Workspace-wide `cargo test --all-features`, clippy `-D warnings`, fmt, and doc builds are green.

## Notes for reviewers

- Three small commits touch sibling crates: a targeted clippy allow in `zcash_client_backend` and precise feature-gating of imports in `zcash_client_sqlite/…/ephemeral.rs` — both fix pre-existing lints that only surface under the feature combination this crate induces. Happy to split them into a preliminary PR if preferred.
- Four `docs/` commits carry the internal design spec/plan/report used to build this; happy to drop them from the series (`git revise`) if unwanted in-tree.
- Two open questions flagged for the team: (1) whether the pinned crates.io `orchard 0.15.0-pre.2` carries the dummy-ciphertext `enc_ciphertext` randomization relevant to migration self-sends; (2) comfort with the vizor (Apache-2.0) lineage under this workspace's dual license.
- Known non-blocking follow-ups are listed in `docs/superpowers/reports/2026-07-09-zcash-pool-migration-final-report.md`, the notable one being expiry detection for a never-broadcast note split.

AI disclosure: commits carry `Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>` per AGENTS.md.


Chlup:
CI failure was merge skew, not a branch defect: main's 63fea5bbd8 made the Orchard-pool bundle type configurable after this branch was cut, changing `BuildConfig::Standard` and `create_pczt_from_proposal`. Merged current main (through #2565) and adapted the three call sites with `BundleType::DEFAULT` — the end-to-end tests' exact fee/crossing assertions confirm unchanged semantics.

One extra commit fixes `pczt/tests/end_to_end.rs`, whose `BuildConfig` initializer was missed by 63fea5bbd8 and fails to compile on current main under `--all-targets` (11/11 tests pass with the fix). Happy to split that out if preferred.

docs/superpowers/plans/2026-07-08-zcash-pool-migration.md
nullcopy, line 1: Please remove this file

docs/superpowers/reports/2026-07-09-zcash-pool-migration-final-report.md
nullcopy, line 1: Please remove this file

docs/superpowers/specs/2026-07-08-zcash-pool-migration-crate-design.md
nullcopy, line 1: Please remove this file

zcash_client_backend/src/data_api/wallet.rs
```diff
@@ -2587,6 +2587,7 @@ where
                     .iter()
                     .enumerate()
                     .filter_map(|(index, input)| {
+                        #[allow(clippy::bind_instead_of_map)]
```
nullcopy, line 2727: Do we need to ignore this lint?

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -26,8 +13,44 @@ use crate::{
 
 use super::next_check_time;
 
+// The imports below back only this module's test / test-dependencies surface (`metadata`,
+// `get_known_ephemeral_addresses`, `find_account_for_ephemeral_address_str`), which is compiled
+// out of ordinary builds. Each is gated to exactly the configurations that use it — rather than
+// blanket-`allow`ing unused imports — so the unused-import lint stays active in every feature
+// combination.
```
nullcopy: No need for this note, the feature gates are self explanatory

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
nullcopy, line 1: Changes like these feel like clean-ups/refactors unrelated to your feature. I would prefer if these were split cleanly into just a few commits, so I can independently verify the refactors (assuming pure refactor) and the feature work. The chain of 34 commits should be squashed into a handful of commits which make sense to review in isolation

nullcopy (changes requested):
Flushing comments as of 17e91c11cd8fae9f8198a38fee559a54d963f2e9

zcash_pool_migration/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+rust-version.workspace = true
+categories.workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+
+[dependencies]
+zcash_address.workspace = true
+zcash_client_backend = { workspace = true, features = ["orchard", "pczt"] }
+# The `serde` feature is required transitively: the high-level `create_pczt_from_proposal`
```
dannywillems, line 20: This comment can be removed.

zcash_pool_migration/src/state.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    #[test]
+    fn phase_strings_round_trip() {
+        for phase in ALL_PHASES {
+            assert_eq!(Phase::parse(phase.as_str()), Some(phase));
+        }
+    }
+
+    #[test]
+    fn known_phase_strings_are_vizor_compatible() {
```
dannywillems, line 142: Can be removed.

nuttycom:
High level comments first:

* ~Transfers should not use the high-level pipeline because we cannot guarantee note selection. Each scheduled transaction should be preconfigured to spend a particular note; the high-level pipeline may choose the wrong notes for the desired transaction transfer size.~ This comment was incorrect; the top level description did not identify that a custom `InputSource` impl is being used; this solves the problem nicely.
* Once a transaction has been created using the zcash_primitives transaction builder, ingest it to the wallet using `decrypt_and_store_transaction` so that the wallet can recognize that the notes have been committed to be spent and does not attempt to spend them in an ordinary transaction.
* Given the previous point, we might need a mechanism to explicitly unlock Orchard notes to permit manual transfer across the pool boundary, ignoring any notes that are already "claimed" by a migration transaction. It must be possible at any point for a user to manually send any portion of their Orchard balance to the Ironwood pool, but this should not be permitted to happen accidentally.

zcash_client_sqlite/src/wallet/transparent/ephemeral.rs
```diff
@@ -26,8 +13,44 @@ use crate::{
...
+#[cfg(any(test, feature = "test-dependencies"))]
```
nuttycom, line 16: Also, collapse this to a single `use { }` block to reduce the feature gate noise.

zcash_pool_migration/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+//!   [`MigrationContext::store_signed_schedule_pczts`] takes the matching [`SignedTransferPczt`]
+//!   slice back and is **all-or-nothing**: every staged transfer must be matched by exactly one
+//!   signed PCZT by id, or nothing is persisted and the staged originals are retained for a
+//!   retry.
+//!
+//! # Security and design notes
+//!
+//! * **Self-funding denominations.** The note split decomposes the spendable Orchard balance
+//!   into power-of-ten ZEC notes (1, 10, 100, ... ZEC), each holding its power-of-ten crossing
+//!   value plus a fee buffer (4× the ZIP-317 marginal fee, covering the 2 Orchard + 2 Ironwood
```
nuttycom, line 131: We may relax the fee buffer requirement in order to reduce the action count that the transfer requires. This involves making migration transactions slightly more distinguishable, but they're already more or less recognizable so this is an acceptable tradeoff in order to be able to fit more transactions into a single Keystone signing session.

zcash_pool_migration/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+//!   actions a migration transfer spends). When such a note is later spent, it pays its own
+//!   transfer fee and exactly the power-of-ten value crosses the Orchard→Ironwood turnstile.
+//! * **Dust stays in Orchard.** Any residual that cannot form a whole self-funding note —
+//!   including dust — is left behind as Orchard change, never folded into a transaction fee:
+//!   folding an identifiable dust amount into a fee would deanonymize a dust-attacked wallet.
+//! * **Natural anchor, pinned during signing.** Every transfer in a schedule shares the wallet's
```
nuttycom, line 137: If I understand what's being suggested here, this is not the desired behavior unless a new anchor will be chosen before sending. Sharing an anchor between multiple transactions links them to one another, allowing an observer of the chain to potentially discover the full balance of a given wallet that is crossing the turnstile.

nuttycom: As a secondary note, we need to ensure that zcash_client_backend's `decrypt_and_store_transaction` and full-block scanning mechanisms can correctly handle the case where a transaction is added to the wallet database, then later has its anchor updated and its proofs rewritten.

dannywillems:
Merging main. This branch won't be committed into `main`, therefore it is fine to have a merge commit.

dannywillems:
This is a branch used by a few people for now, for testing and development. Don't delete it.

nuttycom:
Superseded by https://github.com/zcash/librustzcash/pull/2744

PR #2574, opened 2026-07-09
Add PCZT anchor updater APIs

czarcas7ic:
Adds `Updater` methods for restoring Sapling, Orchard, and Ironwood anchors before proof creation for transaction formats that do not commit shielded signatures to anchors. Existing matching anchors are accepted, while conflicting anchors and already-proved spends or bundles are rejected so callers must clear stale anchors before replacing them.

Tests cover signed PCZT copies whose Sapling or Orchard anchors are restored before proof generation, plus Ironwood anchor restoration across signer finish.

Validation:
- `cargo doc --no-deps -p pczt --all-features --document-private-items`
- `git diff --check`

### DIFF
--- a/pczt/src/roles/updater/mod.rs
+++ b/pczt/src/roles/updater/mod.rs
@@ -59,9 +59,78 @@ impl Updater {
         }
     }
 
+    /// Sets the Orchard bundle anchor for a version 6 PCZT on NU6.3 or later.
+    ///
+    /// Orchard signatures in v6 do not commit to this anchor, so this may be
+    /// called after shielded signatures have been added. Orchard proofs do
+    /// depend on the anchor, so this must be called before proof creation.
+    ///
+    /// Returns an error if the PCZT is not version 6 on NU6.3 or later, or if an
+    /// Orchard proof is already present.
+    #[cfg(feature = "orchard")]
+    pub fn set_v6_orchard_anchor(
+        mut self,
+        anchor: ::orchard::Anchor,
+    ) -> Result<Self, OrchardAnchorUpdateError> {
+        ensure_v6_consensus_branch(&self.pczt.global)?;
+        ensure_no_orchard_proof(&self.pczt.orchard)?;
+        self.pczt.orchard.anchor = Some(anchor.to_bytes());
+        Ok(self)
+    }
+
+    /// Sets the Ironwood bundle anchor for a version 6 PCZT on NU6.3 or later.
+    ///
+    /// Ironwood signatures in v6 do not commit to this anchor, so this may be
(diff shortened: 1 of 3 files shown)


pczt/src/roles/updater/mod.rs
```diff
@@ -59,12 +59,81 @@ impl Updater {
...
+    /// Sets the Orchard bundle anchor for a version 6 PCZT on NU6.3 or later.
+    ///
+    /// Orchard signatures in v6 do not commit to this anchor, so this may be
+    /// called after shielded signatures have been added. Orchard proofs do
+    /// depend on the anchor, so this must be called before proof creation.
+    ///
+    /// Returns an error if the PCZT is not version 6 on NU6.3 or later, or if an
+    /// Orchard proof is already present.
+    #[cfg(feature = "orchard")]
+    pub fn set_v6_orchard_anchor(
```
ValarDragon: I don't think the API needs v6 in the name, it can just error if consensus branch is <= v5 tx format being introduced

ValarDragon: We need the same API for Sapling as well

nuttycom: Right, this will be applicable to subsequent transaction versions as well. It should error if the PCZT already has a conflicting anchor also.

pczt/src/roles/updater/mod.rs
```diff
@@ -74,3 +143,44 @@ impl GlobalUpdater<'_> {
...
     }
 }
+
+/// Errors that can occur while setting Orchard or Ironwood anchors.
+#[cfg(feature = "orchard")]
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum OrchardAnchorUpdateError {
+    /// The PCZT must use the version 6 transaction format for this update.
+    RequiresV6,
```
nuttycom: This shouldn't be specific to transaction version 6; generalize for later versions.

pczt/src/roles/updater/mod.rs
```diff
@@ -74,3 +143,44 @@ impl GlobalUpdater<'_> {
...
+    /// The PCZT's consensus branch ID is unrecognized.
+    UnknownConsensusBranchId,
+    /// The PCZT's consensus branch ID does not support version 6 PCZTs.
```
nuttycom: We currently only have v1 and v2 PCZTs, so I'm not sure what this comment is supposed to mean.

czarcas7ic: I think it was referring to transaction version 6

nuttycom (changes requested):
Minor naming iisues + Sapling API needed, otherwise LGTM

PR #2581, opened 2026-07-09
zcash_client_sqlite: resolve duplicate receiver on standalone transparent import

dannywillems:
## Problem

Importing a standalone transparent pubkey whose receiver address is already recorded (for example because it was derived as an account receiver) fails with:

```
UNIQUE constraint failed: addresses.cached_transparent_receiver_address
```

This surfaced in Zallet's `migrate-zcashd-wallet`: the migration imports a wallet's unified accounts first (deriving and caching each account's transparent receivers), then imports the loose keystore keys as standalone transparent pubkeys. When a loose key's P2PKH address equals one already derived for an account, the standalone import inserts the same `cached_transparent_receiver_address` a second time and aborts.

## Root cause

The UNIQUE index on `addresses.cached_transparent_receiver_address` was added in ef6214bc5e, and the **derivation** path was taught to resolve the resulting duplicates in 3b6a45575c (deriving an address that was already imported upgrades the imported row in place). The **standalone-import** path never got the symmetric resolution: `import_standalone_transparent_pubkey`'s idempotency check keys only on `imported_transparent_receiver_pubkey`, so an address already present as a *derived* row (NULL pubkey) is not matched, and it falls through to a plain `INSERT` that violates the index.

## Fix

- Add an address-existence check before the `INSERT` in `import_standalone_transparent_pubkey`, so an already-recorded receiver is a no-op — the import-direction counterpart of `store_address_range`'s resolution.
- Extract that check into `transparent_receiver_address_exists`.
- `import_standalone_transparent_pubkey` now returns the number of address rows inserted: `1` when a new receiver row was added, `0` when nothing was inserted. The `WalletWrite` impl discards the count, so the public trait API is unchanged.

## Tests

Two proptest cases (random secret key and, for the derived case, a random non-hardened child index above the gap):
- `import_standalone_transparent_pubkey_noop_when_address_derived` — importing a receiver already present as a derived account receiver returns `0` and leaves exactly one row.
- `import_standalone_transparent_pubkey_returns_rows_inserted` — a fresh import returns `1`, and a repeat import of the same pubkey returns `0`.

Existing standalone-import tests, `cargo fmt --check`, and `cargo clippy --tests` all pass.

Generated with Claude.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -24,6 +24,11 @@ workspace.
   exact tree-update logic: `zcash_client_backend::data_api::ll::wallet::`
   `{build_subtrees, checkpoint_positions, ensure_checkpoints,
   cross_pool_ensure_heights, update_tree}`.
+- `zcash_client_backend::data_api::WalletWrite::import_standalone_transparent_pubkeys`
+  (behind the `transparent-key-import` feature flag), a batch variant of
+  `import_standalone_transparent_pubkey` that lets implementations validate the
+  target account once for the whole batch. The default implementation imports
+  each pubkey individually.
 - `zcash_client_backend::proposal::Proposal::proposed_version` and
   `with_proposed_version`. The transaction version requested when a proposal is
   constructed is now recorded on the proposal (and preserved across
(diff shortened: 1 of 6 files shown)


nuttycom (dismissed):
utACK

dannywillems (commented):
Changing the design. Please don't merge.

EDIT: Should be ready now.

nuttycom (approved):
utACK

PR #2585, opened 2026-07-09
Update MSRV to 1.88

nuttycom:
This allows us to update `time` to a version that passes `cargo deny` checks.


dannywillems (approved):
utACK

PR #2586, opened 2026-07-09
zcash_client_backend: extract put_blocks_rows and expose the put_blocks tree-stage helpers

LukasKorba:
As discussed with @nuttycom: Slipstream (the sync engine behind Zodl) implements the
`zcash_client_backend` wallet interfaces and re-routes exactly one method —
`WalletWrite::put_blocks` — so that note commitment trees can be maintained in memory
per scan batch and flushed as a dirty delta, instead of being written through
`WalletCommitmentTrees` block by block. Because `put_blocks` is a monolith over
private helpers, doing that today requires maintaining a line-by-line mirror of its
~1000-line body, of which only the tree stage actually differs. This PR exposes the
seams so that mirror can be deleted and the row-writing logic exists in exactly one
place — which also means the Ironwood arms of this code stay single-sourced.

Behavior is unchanged: `put_blocks` is now expressed as `put_blocks_rows` followed by
the (unchanged) tree stage, and the existing test suites are the oracle.

## Changes

- **Extract `put_blocks_rows`** (`data_api::ll::wallet`): the row-writing stage of
  `put_blocks` — sequential-blocks validation, per-block block/tx/note rows,
  spent-note marking, nullifier tracking and pruning, and (under
  `transparent-inputs`) gap-address maintenance — as a `pub` function over
  `LowLevelWalletWrite` (new `PutBlocksRowsDbT` alias, notably *without* the
  `WalletCommitmentTrees` requirement). It returns `PutBlocksRows`: the accumulated
  per-pool commitments, note positions, and last scanned height — exactly what the
  tree stage consumes.
- **Make the tree-stage helpers `pub`**: `build_subtrees`, `checkpoint_positions`,
  `ensure_checkpoints`, `cross_pool_ensure_heights`, and `update_tree`.
  `update_tree` is already generic over `S: ShardStore`, so alternative stores can
  reuse the exact insertion/checkpoint/anchor-retention logic; the others let a
  caller reproduce the cross-pool checkpoint alignment without copying it.
- CHANGELOG entry under 0.24.0.

## Notes for review

- `put_blocks_rows` carries a free `TE` type parameter so that its errors are
  directly the `PutBlocksError<SE, TE>` of the enclosing `put_blocks` call; happy to
  introduce a narrower error type instead if that's preferred.
- Naming (`put_blocks_rows`, `PutBlocksRows`, `PutBlocksRowsDbT`) is up for
  bikeshedding.
- A thought for a follow-up rather than this PR: `put_blocks`/`put_blocks_rows` grow
  a repeated arm per pool at every pool site — with Ironwood that's three. If the
  per-pool sections were driven by iteration over the active pools, a future pool
  would be a much smaller change for every consumer of these seams. Happy to discuss
  or attempt that separately.

## Follow-ups planned

- A `WalletCommitmentTrees` bulk-flush method (`put_{sapling,orchard,ironwood}_shards`)
  so stores that accumulate tree updates can flush through the interface — removing
  our last direct `zcash_client_sqlite` dependency, per the second half of the same
  discussion.
- An optional perf improvement to the rows stage that we measured on mobile
  (skipping nullifier-map inserts that the next call's prune is guaranteed to
  delete) — kept out of this PR so it stays purely a refactor.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -11,6 +11,19 @@ workspace.
 ## [0.24.0] - PLANNED
 
 ### Added
+- `zcash_client_backend::data_api::ll::wallet::put_blocks_rows` (with the
+  `PutBlocksRows` result type and the `PutBlocksRowsDbT` trait alias): the
+  row-writing stage of `put_blocks`, extracted as a public function over
+  `LowLevelWalletWrite` so that wallet stores that maintain their note
+  commitment trees by other means can reuse it without also implementing
+  `WalletCommitmentTrees`. `put_blocks` itself is unchanged in behavior; it is
+  now expressed as `put_blocks_rows` followed by the note commitment tree
+  updates.
+- The helper functions used by the note commitment tree stage of `put_blocks`
+  are now `pub`, so that alternative `ShardStore`-backed stores can reuse the
+  exact tree-update logic: `zcash_client_backend::data_api::ll::wallet::`
+  `{build_subtrees, checkpoint_positions, ensure_checkpoints,
+  cross_pool_ensure_heights, update_tree}`.
 - `zcash_client_backend::proposal::Proposal::proposed_version` and
   `with_proposed_version`. The transaction version requested when a proposal is
   constructed is now recorded on the proposal (and preserved across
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK

Issue #2584, opened 2026-07-09
zcash_client_sqlite: Support regtest wallets in ZeWIF import


nuttycom:
## Summary

`zcash_client_sqlite::zewif::import_wallet` unconditionally rejects any document
recorded against a regtest network — the `AccountViewingKey`-network match in
`zcash_client_sqlite/src/zewif.rs` returns `ZewifImportError::UnsupportedRegtest`
for `zewif::Network::Regtest(_)` regardless of the wallet database's parameters —
so a regtest wallet cannot be imported at all. This tracks adding regtest support.

Builds on #2567. Related to #2582.

## Why regtest is gated

`zewif::Network::Regtest` carries `RegtestParams { activations: BTreeMap<branch_id, height> }`.
Regtest activation heights are operator-configured (`-nuparams`), and wallet data
recorded against one schedule is in general meaningless against another: the
importer relies on activation heights for consensus-branch-ID selection when
storing raw transactions, for note-commitment-tree interpretation, and for the
Sapling-activation birthday floor. It therefore cannot blindly assume the wallet
database's regtest `Parameters` describe the same chain the document was recorded
against.

## Importer work (primary deliverable)

Replace the blanket rejection with:

- `(Regtest(rp), NetworkType::Regtest)`: if `rp.activations()` is non-empty,
  verify it against the wallet database's params — for each `NetworkUpgrade`,
  compare `params.activation_height(nu)` with
  `rp.activations().get(&u32::from(nu.branch_id()))`; a disagreement is a new
  `ZewifImportError::RegtestActivationMismatch`. If the document carries no
  schedule (the common case today — see below), trust the caller-supplied params,
  since the operator controls both the node's `-nuparams` and the import
  parameters.
- `(Regtest(_), Main | Test)`: `NetworkMismatch`, as today.

Regtest parameters are supplied via `LocalNetwork`
(`zcash_protocol::local_consensus`), whose `network_type()` is
`NetworkType::Regtest` and which exposes `activation_height`.

## Exporter follow-ons (zcash/zewif-zcashd, separate)

- `zewif-zcashd` currently emits `Network::Regtest(Default::default())` — an
  *empty* activation schedule — because a `wallet.dat` does not record the
  schedule (it comes from node config). For a *verifiable* regtest export the
  schedule would need to be captured from an additional input (`-nuparams` /
  `getblockchaininfo`), not just the wallet file.
- HRP inconsistency: regtest Sapling extended keys are encoded with regtest HRPs,
[...]

PR #2587, opened 2026-07-09
zcash_client_sqlite: Support regtest wallets in ZeWIF import

nuttycom:
## Summary

`zcash_client_sqlite::zewif::import_wallet` previously rejected every document
recorded against a regtest network with `UnsupportedRegtest`, because the
equivalence of regtest activation schedules could not be verified. This imports
them instead against the wallet database's regtest `Parameters`.

- A network-upgrade activation schedule carried by the document
  (`zewif::RegtestParams`) is verified against the configured parameters: each
  recorded consensus-branch-ID activation height must equal
  `Parameters::activation_height`, otherwise the new
  `ZewifImportError::RegtestActivationMismatch`.
- A document carrying **no** schedule is trusted against the caller-supplied
  params, since the operator controls both the node's `-nuparams` and the import
  parameters. This is the common case today: an exporter cannot in general
  recover the activation schedule from a `wallet.dat` alone (e.g. `zewif-zcashd`
  emits an empty `RegtestParams`).

`UnsupportedRegtest` is replaced by `RegtestActivationMismatch`. The `zewif`
feature is unreleased, so there is no CHANGELOG entry.

## Tests

- Unit tests of the verifier: empty schedule trusted, matching schedule accepted,
  mismatched height rejected, unknown branch ID rejected.
- Integration tests through `import_wallet` against a `LocalNetwork` database: a
  regtest document imports successfully; a mismatched schedule is rejected; a
  regtest document against testnet params is a `NetworkMismatch`.

## Follow-ons

The exporter-side items (an empty `RegtestParams` today; regtest address/HRP
encoding reconciliation in `zewif-zcashd`) remain tracked in #2584.

Part of #2584.

### DIFF
--- a/zcash_client_sqlite/src/zewif.rs
+++ b/zcash_client_sqlite/src/zewif.rs
@@ -48,8 +48,11 @@
 //!   block scanning can establish. Importing an account seeds the scan queue from
 //!   its birthday; the wallet's balance becomes visible and spendable as the
 //!   post-import rescan progresses.
-//! * Documents recorded against a regtest network are not currently supported,
-//!   as the equivalence of regtest activation schedules cannot be verified here.
+//! * Documents recorded against a regtest network are imported against the
+//!   wallet database's regtest [`Parameters`]. A network-upgrade activation
+//!   schedule carried by the document is verified against those parameters; a
+//!   document that carries no schedule is trusted, as the operator controls both
+//!   the node's activation heights and the import parameters.
 //! * Encrypted secret stores must be decrypted by the caller (using the `zewif`
 //!   crate's decryption support) before import.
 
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/zewif.rs
```diff
@@ -801,6 +816,63 @@ fn zip32_derivation<S>(
     ))
 }
 
+/// Maps a consensus branch ID to the network upgrade it activates, or `None`
+/// for the pre-Overwinter Sprout rules, which have no activation height.
+///
+/// This mirrors the branch-ID assignment in [`zcash_protocol::consensus`], which
```
dannywillems, line 822: Nit: we should be able to make it public then? It is in components/zcash_protocol/src/consensus.rs. Claude will do it later.

dannywillems (approved):
LGTM

zcash_client_sqlite/src/zewif.rs
dannywillems: See https://github.com/zcash/librustzcash/pull/2588

nuttycom:
Closing in favor of #2588

PR #2590, opened 2026-07-09
zcash_client_sqlite: Gate feature-conditional imports to fix clippy warnings

dannywillems:
## Summary

Several imports and helpers in `zcash_client_sqlite` were consumed only by
code compiled under narrower feature sets than the modules that declared them.
As a result, building certain feature combinations produced `unused_imports`
and `dead_code` warnings, for example:

- `cargo clippy -p zcash_client_sqlite --features zewif` (transparent-inputs on,
  test-dependencies off)
- `cargo clippy -p zcash_client_sqlite --all-targets` (default features, orchard
  off)

This gates each item with the exact `cfg` of the code that uses it. No
functional change.

## Changes

- `wallet::transparent::ephemeral`: `metadata` (and the imports it needs) is
  used only by the `test-dependencies`-gated `get_known_ephemeral_addresses`,
  and `find_account_for_ephemeral_address_str` only by the
  `transparent-inputs` test surface. Gated accordingly.
- `lib.rs`: `AddressCodec` is used only by the `WalletTest`-impl method
  `find_account_for_ephemeral_address`; gated to
  `all(any(test, test-dependencies), transparent-inputs)`.
- `wallet::commitment_tree` tests: `SqliteClientError` is used only by the
  `orchard`-gated `HistoricalWitnessGenerator` type alias.
- `wallet::init::migrations` tests: `proptest::prelude::any` and
  `proptest::prop_compose` are used only by the `orchard`-gated note-generation
  strategies.

## Verification

`cargo clippy -p zcash_client_sqlite` is warning-clean across all of:
default, `--features zewif`, `transparent-inputs`, `test-dependencies`,
`spend-index`, `orchard`, and `--all-features`, each also with `--all-targets`.
Affected tests pass under `--all-features`.

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -123,11 +123,16 @@ use {
         fees::StandardFeeRule,
         wallet::TransparentAddressMetadata,
     },
-    zcash_keys::{
-        encoding::AddressCodec,
-        keys::transparent::gap_limits::{AddressStore, GapLimits},
-    },
+    zcash_keys::keys::transparent::gap_limits::{AddressStore, GapLimits},
 };
+
+// `AddressCodec` is used only by `find_account_for_ephemeral_address`, which is
+// part of the `WalletTest` surface.
+#[cfg(all(
+    any(test, feature = "test-dependencies"),
+    feature = "transparent-inputs"
+))]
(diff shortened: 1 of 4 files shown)


nullcopy (approved):
utACK 6f5e7ca6e1968a6ecff6569548d0cec32354db6c

PR #2593, opened 2026-07-09
Make Orchard PCZT cmx optional

ValarDragon:
## Summary

- Allow v2 Orchard PCZTs to redact output note commitments by representing logical and v2 `cmx` as optional.
- Resolve redacted `cmx` before parsed conversion from the output note fields and action spend nullifier, while keeping memo recovery best-effort.
- Keep v1 encoding strict, add `ActionRedactor::clear_cmx`, and update focused tests and changelog entries.

## Validation

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -286,12 +286,6 @@ fn recover_memo_plaintext_from_ciphertext_and_action(
         // we return None here to avoid excess sets or clone operations, as the caller need not do anything in this case.
         EncCiphertext::MemoPlaintext(_) => return None,
     };
-    let output = OutputRecoveryData {
-        cmx: action.output.cmx,
-        ephemeral_key: action.output.ephemeral_key,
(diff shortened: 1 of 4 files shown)


pczt/src/orchard.rs
```diff
@@ -1025,7 +1027,7 @@ pub(crate) mod v2 {
             MemoPlaintext, NoteVersion, ORCHARD_SPENDS_AND_OUTPUTS_ENABLED, Output, Spend,
         };
 
-        fn logical_action(cv_net: Option<[u8; 32]>) -> LogicalAction {
+        fn logical_action(cv_net: Option<[u8; 32]>, cmx: Option<[u8; 32]>) -> LogicalAction {
```
ValarDragon, line 1030: This is in tests

nuttycom (approved):
utACK e53d979720d726ace1106b686069cc6388954825

PR #2596, opened 2026-07-10
zcash_client_backend: rename and restructure Recipient::InternalAccount / BuildRecipient

### LINKED ISSUE #2595: zcash_client_backend: Revisit the naming and delineation of `Recipient::InternalTransparent` vs. `Recipient::InternalAccount`
Follow-up from https://github.com/zcash/librustzcash/pull/2550#discussion_r3549071493.

PR #2550 added `Recipient::InternalTransparent` (`zcash_client_backend/src/wallet.rs`) to record the send side of a non-ephemeral transparent output that the wallet both funded and received (i.e. transparent change), along with construction-time mirrors `BuildRecipient::InternalTransparent` and `PcztRecipient::InternalTransparent` in `zcash_client_backend/src/data_api/wallet.rs`.

The current delineation from `Recipient::InternalAccount` is:

- `InternalAccount` carries a shielded `Note`; the recipient address is not externally meaningful, so wallet-internal shielded outputs are recorded against the receiving account alone (with an optional `external_address` for cross-account sends via an external IVK).
- `InternalTransparent` records the on-chain-observable `recipient_address`, because for transparent outputs the address is visible on chain and must be retained; the output is subsequently tracked via its outpoint.

As noted in review, the naming is confusing: both variants describe outputs that are internal to (belonging to an account of) the wallet, so the `Transparent` / `Account` suffixes do not clearly communicate the actual distinction (shielded note vs. transparent outpoint tracking). There may also be an axis of symmetry to refactor along — e.g. internal/external crossed with pool type — rather than maintaining parallel variants whose differences are encoded in their payloads.

Suggested scope:

- Reconsider the names of the two variants (and their `BuildRecipient` / `PcztRecipient` mirrors) so that the distinction they encode is evident.
- Evaluate whether the variants should be restructured (e.g. a common internal-recipient shape parameterized by pool-specific data) instead of renamed.

Note that `Recipient` is part of the public API of `zcash_client_backend`, so any change here is semver-breaking and should be scheduled accordingly.

schell:
Resolves #2595.

Two commits:

1. Renames `zcash_client_backend::wallet::Recipient::InternalAccount` (and its
   `BuildRecipient`/`PcztRecipient` mirrors) to `InternalShielded`, for symmetry with
   `Recipient::InternalTransparent`. The prior name did not communicate the actual
   distinction between the two internal variants — payload domain (shielded note vs.
   transparent address), not some special relationship to "the account". The
   `Recipient` enum doc comments are expanded to spell out the two independent axes the
   variants vary along. This is a public, semver-breaking rename; a `Changed` entry is
   added to the unreleased section of `zcash_client_backend/CHANGELOG.md`.
   `PcztRecipient`'s variant order (and thus its postcard encoding within a PCZT) is
   unchanged.

2. Splits the private `BuildRecipient` type into `ShieldedBuildRecipient` and
   `TransparentBuildRecipient`, matching the two functions that previously consumed a
   single `BuildRecipient` with two `unreachable!()` arms apiece
   (`into_recipient_with_note`, used only for shielded outputs where a decrypted `Note`
   is available; `into_recipient_with_outpoint`, used only for transparent outputs
   where an `OutPoint` is available). Each split type carries only the variants
   relevant to its domain, and each has a single `into_recipient` method that is a
   total function. `PcztRecipient::from_recipient` is likewise split. This is a
   private, internal-only change with no effect on the public API.

Both follow from review discussion on #2550:
https://github.com/zcash/librustzcash/pull/2550#discussion_r3549071493

## Verification

- `cargo clippy --all-features --all-targets -- -D warnings` (workspace)
- `cargo check -p zcash_client_backend` with `--all-features`, `--no-default-features`,
  `--features transparent-inputs`, `--features orchard`
- `cargo test --workspace --all-features` (all green, no failures)
- `cargo doc --no-deps --workspace --all-features --document-private-items` (nightly):
  no new warnings (4 pre-existing warnings, unrelated to this change)
- Verified commit 1 builds standalone

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -282,6 +282,11 @@ workspace.
   this option must be executed with a matching unpadded builder configuration.
 
 ### Changed
+- `zcash_client_backend::wallet::Recipient::InternalAccount` has been renamed
+  to `Recipient::InternalShielded`, for symmetry with `Recipient::InternalTransparent`
+  and to clarify that the distinguishing feature of this variant is that its
+  payload is a shielded note, not that it is somehow more "the account" than the
+  other internal variant.
 - MSRV is now 1.88
 - `zcash_client_backend::data_api::wallet::create_pczt_from_proposal` now takes an
   `orchard_pool_bundle_type` argument (behind the `pczt` feature flag) selecting
--- a/zcash_client_backend/src/data_api/ll/wallet.rs
+++ b/zcash_client_backend/src/data_api/ll/wallet.rs
@@ -1264,7 +1264,7 @@ where
                 let note = output.to_wallet_note();
                 let value = note.value();
 
-                let recipient = Recipient::InternalAccount {
+                let recipient = Recipient::InternalShielded {
                     receiving_account: output.account_id(),
                     external_address: None,
                     note: Box::new(note),
@@ -1282,7 +1282,7 @@ where
                     let value = note.value();
 
                     // Even if the recipient address is external, record the send as internal.
-                    let recipient = Recipient::InternalAccount {
+                    let recipient = Recipient::InternalShielded {
                         receiving_account: output.account_id(),
                         external_address: Some(external_address(
                             wallet_db,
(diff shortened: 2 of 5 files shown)


nuttycom (approved):
utACK e53d979720d726ace1106b686069cc6388954825

PR #2598, opened 2026-07-10
zcash_client_sqlite: Include Ironwood checkpoints in rewind selection

gustavovalverde:
Wallet rewinds currently choose a checkpoint shared by Sapling and Orchard without requiring Ironwood. When Ironwood lacks the selected height, `truncate_to_checkpoint` leaves that tree unchanged, so it remains ahead of the rest of the wallet and replacement blocks can conflict with its stale frontier.

Include Ironwood in both the selected-checkpoint query and the safe-rewind-height intersection. `rewind_to_chain_state` now also treats Ironwood checkpoint divergence inside the pruning window as corrupted data, matching the existing Orchard safeguard. Focused regressions cover a requested height without an Ironwood checkpoint and a checkpoint set with no safe three-pool height.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -3674,28 +3674,26 @@ pub(crate) fn set_transaction_status<P: consensus::Parameters>(
 }
 
 /// Returns the minimum checkpoint height that exists in all note commitment trees that contain
-/// data. When both trees have checkpoints, returns the minimum of their intersection. When only
-/// one tree has checkpoints, returns that tree's minimum. Returns `None` when both are empty.
+/// data. A height qualifies when every tree that has any checkpoints has a checkpoint at that
+/// height. Returns `None` when all trees are empty.
 fn min_shared_checkpoint_height(
     conn: &rusqlite::Connection,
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/CHANGELOG.md
```diff
@@ -123,6 +123,10 @@ workspace.
   the migration to abort.
 
 ### Fixed
+- Wallet rewinds now select checkpoints shared by the Sapling, Orchard, and Ironwood note
+  commitment trees. Previously, a rewind could select a height without an Ironwood checkpoint,
+  leaving the Ironwood tree ahead of the rest of the wallet and preventing replacement blocks
+  from being scanned after a reorg.
```
nuttycom: Also a fix relative to unreleased code, so it doesn't need to be in the CHANGELOG

nuttycom (commented):
utACK modulo the CHANGELOG fix.

zcash_client_sqlite/CHANGELOG.md
gustavovalverde: Fixed

nuttycom (approved):
utACK

PR #2599, opened 2026-07-10
zcash_client_backend: Fetch Ironwood subtree roots during sync

gustavovalverde:
Wallets that use `zcash_client_backend::sync::run` and start from a birthday after one or more completed Ironwood subtrees receive Sapling and Orchard subtree roots, but not Ironwood roots. Their Ironwood commitment tree therefore starts from incomplete history, so later Ironwood notes can lack the witness path needed for spending.

Fetch the Ironwood roots through the existing lightwalletd `GetSubtreeRoots` RPC and store them through `WalletCommitmentTrees::put_ironwood_subtree_roots`, matching the established Sapling and Orchard sync path. The shared download and decoding logic now serves all three pools, and a lightweight mock gRPC regression requires the sync phase to issue three subtree-root requests.

### DIFF
--- a/zcash_client_backend/src/sync.rs
+++ b/zcash_client_backend/src/sync.rs
@@ -226,6 +226,36 @@ where
 }
 
 #[cfg(feature = "sync")]
+async fn download_subtree_roots<ChT, H>(
+    client: &mut CompactTxStreamerClient<ChT>,
+    protocol: service::ShieldedProtocol,
+) -> Result<Vec<CommitmentTreeRoot<H>>, tonic::Status>
+where
+    ChT: GrpcService<TonicBody>,
+    ChT::Error: Into<StdError>,
+    ChT::ResponseBody: Body<Data = Bytes> + Send + 'static,
+    <ChT::ResponseBody as Body>::Error: Into<StdError> + Send,
+    H: HashSer,
+{
+    let mut request = service::GetSubtreeRootsArg::default();
+    request.set_shielded_protocol(protocol);
+
+    client
+        .get_subtree_roots(request)
+        .await?
+        .into_inner()
+        .and_then(|root| async move {
+            let root_hash = H::read(&root.root_hash[..])?;
+            Ok(CommitmentTreeRoot::from_parts(
+                BlockHeight::from_u32(root.completing_block_height as u32),
+                root_hash,
+            ))
+        })
+        .try_collect()
+        .await
+}
+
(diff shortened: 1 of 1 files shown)


zcash_client_backend/CHANGELOG.md
```diff
@@ -464,6 +464,12 @@ workspace.
   `fee_required` directly or implements the trait must thread through the number
   of Ironwood actions, passing `0` for transactions without an Ironwood bundle.
 
+### Fixed
+- The built-in synchronization flow now downloads Ironwood note commitment subtree roots.
+  Previously, wallets using `sync::run` could start from an incomplete Ironwood tree when their
+  birthday was after completed Ironwood subtrees, preventing them from constructing Ironwood
+  witnesses for notes discovered later.
```
nuttycom: This is only a fix relative to a not-yet-released state, so it doesn't need to be in the changelog.

zcash_client_backend/src/sync.rs
```diff
@@ -283,6 +283,28 @@ where
         db_data
             .put_orchard_subtree_roots(0, &orchard_roots)
             .map_err(Error::WalletTrees)?;
+
```
nuttycom, line 288: Ideally instead of duplicating this code you could factor out a function.

zcash_client_backend/Cargo.toml
```diff
@@ -157,6 +158,7 @@ rand_chacha.workspace = true
 shardtree = { workspace = true, features = ["test-dependencies"] }
 tempfile = "3.5.0"
 tokio = { version = "1.21.0", features = ["rt-multi-thread"] }
+tower = { workspace = true, features = ["util"] }
```
nuttycom: I don't understand why this PR involves these dependency additions?

zcash_client_backend/CHANGELOG.md
gustavovalverde: Fixed

zcash_client_backend/src/sync.rs
gustavovalverde: Refactored

zcash_client_backend/Cargo.toml
gustavovalverde: Removed, pulled from the other PR by mistake :)

nuttycom (approved):
utACK, nice simplification to the types is a bonus!

PR #2600, opened 2026-07-10
zcash_client_backend: add bulk-flush methods to WalletCommitmentTrees

LukasKorba:
This is the second half of the discussion behind #2586: Slipstream implements the
`zcash_client_backend` wallet interfaces and maintains its note commitment trees in an
in-memory store during scanning, flushing the accumulated changes (dirty shards, the tree
cap, and the checkpoint delta) to the wallet database once per batch. Today that flush is
the one place it must reach into `zcash_client_sqlite` concretely; this PR lets it go
through the `WalletCommitmentTrees` interface instead — the "I'd prefer Slipstream didn't
deal with `zcash_client_sqlite` directly" ask.

## Changes

- **`WalletCommitmentTrees::put_sapling_shards`** (and `put_orchard_shards` /
  `put_ironwood_shards` under the `orchard` feature): provided methods that bulk-write
  previously accumulated tree changes — shards (ascending index order), an optional
  replacement tree cap, and a checkpoint delta (removals applied before additions) — to
  the backing [`ShardStore`].
- The **default implementations** apply the changes through the corresponding
  `with_*_tree_mut` methods (the same pattern as `remove_retained_checkpoints_below`), so
  **existing implementations of the trait are unaffected** and `zcash_client_sqlite`
  needs no changes — the default runs inside its usual tree-access transaction context.
  The Ironwood variant follows `with_ironwood_tree_mut`'s convention for backends that do
  not track an Ironwood tree.
- A unit test exercising the flush-and-read-back roundtrip (including
  remove-before-add checkpoint replacement) against the in-memory mock store.
- CHANGELOG entry under 0.24.0.

## Notes for review

- The signature types are written fully qualified (`shardtree::LocatedPrunableTree`,
  `shardtree::store::Checkpoint`, …) deliberately: `WalletCommitmentTrees` is an
  `ambassador`-delegatable trait, and delegation sites receive a textual copy of each
  method signature — unqualified names would force every existing
  `#[derive(Delegate)]` site (in-repo `TestDb`, and downstream users of the pattern)
  to add imports. Verified against the in-repo delegation site.
- Parameters are borrowed (`&[...]`) and cloned per store write, because
  `with_*_tree_mut` takes `FnMut` — and per-shard cloning is exactly what a consumer
  flushing from an accumulation store does today anyway. Happy to change the calling
  convention if you'd prefer owned inputs.
- Independent of #2586 (branched from current `main`); the two only touch adjacent
  CHANGELOG lines.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -3866,6 +3866,138 @@ pub trait WalletCommitmentTrees {
         Ok(())
     }
 
+    /// Bulk-writes previously accumulated changes to the wallet's Sapling note commitment tree.
+    ///
+    /// Writes the provided shards to the backing [`ShardStore`] (`shards` must be in ascending
+    /// shard-index order; stores may reject sequences that would leave gaps in the tree),
+    /// replaces the tree cap if one is provided, and applies the checkpoint delta — removals
+    /// before additions, so that a checkpoint whose data has changed may appear in both lists.
+    ///
+    /// This is intended for wallet stores that accumulate note commitment tree updates outside
+    /// the backing store (for example, in an in-memory tree) and periodically flush the
+    /// accumulated changes. The default implementation applies the changes through
+    /// [`WalletCommitmentTrees::with_sapling_tree_mut`].
+    fn put_sapling_shards(
+        &mut self,
+        shards: &[shardtree::LocatedPrunableTree<sapling::Node>],
+        cap: Option<&shardtree::PrunableTree<sapling::Node>>,
+        checkpoints_remove: &[BlockHeight],
+        checkpoints_add: &[(BlockHeight, shardtree::store::Checkpoint)],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        self.with_sapling_tree_mut(|tree| {
+            for shard in shards {
+                tree.store_mut()
+                    .put_shard(shard.clone())
+                    .map_err(ShardTreeError::Storage)?;
+            }
+            if let Some(cap) = cap {
+                tree.store_mut()
+                    .put_cap(cap.clone())
+                    .map_err(ShardTreeError::Storage)?;
+            }
+            for height in checkpoints_remove {
+                tree.store_mut()
+                    .remove_checkpoint(height)
+                    .map_err(ShardTreeError::Storage)?;
+            }
+            for (height, checkpoint) in checkpoints_add {
+                tree.store_mut()
+                    .add_checkpoint(*height, checkpoint.clone())
+                    .map_err(ShardTreeError::Storage)?;
+            }
+            Ok::<_, ShardTreeError<Self::Error>>(())
+        })
+    }
+
+    /// Bulk-writes previously accumulated changes to the wallet's Orchard note commitment tree.
+    ///
+    /// Writes the provided shards to the backing [`ShardStore`] (`shards` must be in ascending
+    /// shard-index order; stores may reject sequences that would leave gaps in the tree),
+    /// replaces the tree cap if one is provided, and applies the checkpoint delta — removals
+    /// before additions, so that a checkpoint whose data has changed may appear in both lists.
+    ///
+    /// This is intended for wallet stores that accumulate note commitment tree updates outside
+    /// the backing store (for example, in an in-memory tree) and periodically flush the
+    /// accumulated changes. The default implementation applies the changes through
+    /// [`WalletCommitmentTrees::with_orchard_tree_mut`].
+    #[cfg(feature = "orchard")]
+    fn put_orchard_shards(
+        &mut self,
+        shards: &[shardtree::LocatedPrunableTree<orchard::tree::MerkleHashOrchard>],
+        cap: Option<&shardtree::PrunableTree<orchard::tree::MerkleHashOrchard>>,
+        checkpoints_remove: &[BlockHeight],
+        checkpoints_add: &[(BlockHeight, shardtree::store::Checkpoint)],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        self.with_orchard_tree_mut(|tree| {
+            for shard in shards {
+                tree.store_mut()
+                    .put_shard(shard.clone())
+                    .map_err(ShardTreeError::Storage)?;
+            }
+            if let Some(cap) = cap {
+                tree.store_mut()
+                    .put_cap(cap.clone())
+                    .map_err(ShardTreeError::Storage)?;
+            }
+            for height in checkpoints_remove {
+                tree.store_mut()
+                    .remove_checkpoint(height)
+                    .map_err(ShardTreeError::Storage)?;
+            }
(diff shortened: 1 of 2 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -3844,6 +3844,138 @@ pub trait WalletCommitmentTrees {
         Ok(())
     }
 
+    /// Bulk-writes previously accumulated changes to the wallet's Sapling note commitment tree.
```
nullcopy: (nit) I don't think it's useful to describe how the function behaved before

nullcopy: Same for the others

zcash_client_backend/src/data_api.rs
```diff
@@ -3844,6 +3844,138 @@ pub trait WalletCommitmentTrees {
...
+    /// accumulated changes. The default implementation applies the changes through
+    /// [`WalletCommitmentTrees::with_sapling_tree_mut`].
+    fn put_sapling_shards(
+        &mut self,
+        shards: &[shardtree::LocatedPrunableTree<sapling::Node>],
+        cap: Option<&shardtree::PrunableTree<sapling::Node>>,
+        checkpoints_remove: &[BlockHeight],
+        checkpoints_add: &[(BlockHeight, shardtree::store::Checkpoint)],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        self.with_sapling_tree_mut(|tree| {
```
nullcopy, line 3934: All three of these functions do the same thing, but for a different tree... can this be abstracted?

zcash_client_backend/src/data_api.rs
```diff
@@ -3915,6 +4047,92 @@ mod tests {
     use zcash_keys::address::{Address, UnifiedAddress};
     use zip32::DiversifierIndex;
 
+    #[test]
+    fn put_sapling_shards_flushes_through_the_interface() {
```
nullcopy, line 4060: (nit) to be complete, we should test against all pools, but I won't block on this.

Maybe if you are able to abstract the methods to use a common tree update helper, then this test can just exercise the helper to easily cover everything 🤷

zcash_client_backend/src/data_api.rs
```diff
@@ -3844,6 +3844,138 @@ pub trait WalletCommitmentTrees {
...
+        &mut self,
+        shards: &[shardtree::LocatedPrunableTree<sapling::Node>],
+        cap: Option<&shardtree::PrunableTree<sapling::Node>>,
+        checkpoints_remove: &[BlockHeight],
+        checkpoints_add: &[(BlockHeight, shardtree::store::Checkpoint)],
+    ) -> Result<(), ShardTreeError<Self::Error>> {
+        self.with_sapling_tree_mut(|tree| {
+            for shard in shards {
+                tree.store_mut()
+                    .put_shard(shard.clone())
```
nullcopy: (possible to explore later) I wonder if shardtree could be improved with a batch insertion instead of looping over singular insert

nullcopy (commented):
A few minor comments as of e72c9e2d5c4e1562e6f1f3d0f92bee31c61e34e8

Also some conflicts

zcash_client_backend/src/data_api.rs
nuttycom: Yeah, that would be potentially useful. I didn't consider it worth the optimization initially because we were essentially never inserting more than one shard at a time, given that our batch fetches were only 100 or 1000 blocks at a time.

zcash_client_backend/src/data_api.rs
nuttycom: @nullcopy I don't understand your comment, is it stale? I don't see any description of previous behavior of a function; this is the contract that implementations have to enforce, AFACIT?

nuttycom (dismissed):
utACK - deduplicating the function implementation is a nonblocking enhancement but you might as well do it since this requires conflict resolution anyway.

zcash_client_backend/src/data_api.rs
LukasKorba: I read it as the "previously accumulated" phrasing (temporal framing rather than describing the input). Reworded across all three methods and the changelog entry — the docs now state what the batch contains and its ordering contract.

zcash_client_backend/src/data_api.rs
LukasKorba: Extracted a shared `apply_tree_changes` (generic over the node type and backing store); the three methods are now single-statement delegations through their respective `with_*_tree_mut`.

zcash_client_backend/src/data_api.rs
LukasKorba: Covered via the extraction, as you suggested: `check_apply_tree_changes` exercises the shared helper directly over a `MemoryShardStore`, instantiated for both node types (`sapling::Node` and `MerkleHashOrchard` — the latter being the tree shape Orchard and Ironwood share). The existing test keeps covering the trait-level wiring through `MockWalletDb`, and a new test pins that `put_ironwood_shards` is ignored (not an error) on a backend without an Ironwood tree.

zcash_client_backend/src/data_api.rs
LukasKorba: Agreed — that belongs at the shardtree level, independent of this PR (the provided methods here target released shardtree). It matters more for bulk backfill than tip-following: our batches cover 10k-block ranges, so shards arrive several at a time. Two things a batch API could buy: a `ShardStore::put_shards(&[…])` would let SQLite-backed stores use a single transaction / prepared statement instead of N round trips, and on the `ShardTree::insert_tree` path batching would amortize the excess-checkpoint pruning that currently runs inside every call — pruning mid-batch can evict a checkpoint that a later insert in the same batch would have re-established, an ordering we ran into while porting checkpoint retention. Happy to open a shardtree issue.

LukasKorba:
Rebased onto current main (the conflicts were confined to the changelog) and addressed the review comments in 1f100478ed — including the function-body dedup from the utACK.

zcash_client_backend/src/data_api.rs
```diff
@@ -3762,6 +3762,49 @@ pub trait WalletWrite: WalletRead {
...
 
+/// Applies a batch of note commitment tree changes — shards, an optional replacement tree
+/// cap, and a checkpoint delta — directly to the given tree's backing [`ShardStore`].
+///
+/// `shards` must be in ascending shard-index order; stores may reject sequences that would
+/// leave gaps in the tree. Checkpoint removals are applied before additions, so that a
+/// checkpoint whose data has changed may appear in both lists.
+///
+/// This is the shared implementation of the [`WalletCommitmentTrees`] `put_*_shards`
+/// provided methods.
```
nuttycom, line 3773:
```suggestion
/// provided methods.
///
/// NOTE: This procedure must be called only within a the context of a transaction, such as
/// in the scope of a `with_*_tree_mut` call; otherwise, failure of an intermediate step could
/// lead to data corruption.
```

nuttycom (approved):
utACK

PR #2602, opened 2026-07-10
Add PCZT batch signing messages and signature contributions

czarcas7ic:
Adds type-safe Orchard and Ironwood spend authorization signature contributions to the PCZT Signer role, with helpers to extract contributions from a signed PCZT and apply them to the client-retained copy. It also adds request and response types for transporting batches of PCZTs to an external signer while preserving request order. Sapling spend authorization signatures are intentionally not represented.

The public batch request operates on logical `Pczt` values. Its versioned Postcard wire format uses `PCZB || batch_version || pczt_version || body`, with one shared PCZT encoding version and typed, headerless PCZT payloads. This makes mixed-version batches structurally unrepresentable. Responses use the distinct `PCZS || batch_version || body` framing. Request and response correlation is intentionally left to the application transport instead of being duplicated in these domain messages.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3007,8 +3007,9 @@ checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
 
 [[package]]
 name = "orchard"
-version = "0.15.0-pre.2"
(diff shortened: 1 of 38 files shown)


nuttycom (approved):
utACK 878db2074ae8ac2682d3e6c61c00f7018b6adc0c

PR #2603, opened 2026-07-10
fix: preserve shielding proposal anchors

gustavovalverde:
Transparent-to-shielded proposals discarded the checkpoint selected at proposal time whenever they had no shielded inputs. The transaction builder then resolved an anchor from the target height, which may not be checkpointed in a transparent-only chain and fails with `AnchorNotFound`.

Preserve the selected checkpoint for newly generated input-less steps. The zero sentinel still represents legacy deferred anchors, so older serialized proposals retain their existing interpretation.

AI assistance: Codex prepared the change and regression coverage.


zcash_client_backend/src/proposal.rs
```diff
@@ -665,10 +665,18 @@ impl<NoteRef> Step<NoteRef> {
...
 
-        // Only a step that spends shielded notes binds a concrete anchor (needed to witness those
-        // notes). An input-less step defers to interpretation time, so its stored anchor is `None`
-        // regardless of the value passed here.
-        let anchor_height = shielded_inputs.as_ref().map(|_| anchor_height);
+        // Keep the checkpoint selected during proposal construction even for transparent-only
+        // inputs: output-only Orchard and Ironwood bundles still require a tree anchor. The zero
+        // sentinel remains reserved for decoded legacy proposals that deferred their anchor, and
+        // is invalid for steps that spend shielded inputs.
+        let anchor_height = if u32::from(anchor_height) == 0 {
```
nuttycom: This comparison makes me suspicious; I would expect the anchor height to only be zero as an artifact of protobuf deserialization returning the default; however, ideally we would exclude that from the set of valid values for anchor heights, as of course genesis can never be an anchor for any valid spend. It looks to me like this PR might be fixing an error (or an API design issue) somewhere else in the stack.

nuttycom: I will push a fix to this branch.

nuttycom (approved):
utACK 9367a561829c2b2c947b494f06db0ad05d150fd0

zcash_client_backend/src/proposal.rs
```diff
@@ -953,6 +987,69 @@ mod tests {
...
+                )
+                .unwrap(),
+                false,
+                false,
+            ),
+            Err(ProposalError::MissingShieldedAnchor)
+        );
+
+        // A purely transparent step may carry no anchor.
+        assert_matches!(
```
nullcopy, line 1036: (nit) this assertion would be better in a different "fully_transparent_step_passes_without_anchor" test or something. Not worth blocking though.

nullcopy (approved):
utACK 9367a561829c2b2c947b494f06db0ad05d150fd0

PR #2604, opened 2026-07-10
zcash_client_backend: add a nullifier-tracking floor to `put_blocks_rows`

LukasKorba:
Bulk restores spend a startlingly large fraction of their row-writing stage on the nullifier map: `track_block_*_nullifiers` plus the subsequent prune measured ~85% of that stage's cost in our wallet, because the insert-everything-then-prune-to-a-trailing-window cycle inserts thousands of blocks' nullifier rows per batch and deletes all but the window one maintenance call later.

Those insertions are provably unobservable whenever the scanned batch **extends the wallet's contiguous fully-scanned frontier**: the nullifier map exists to detect spends observed before the corresponding note's block has been scanned, which cannot occur below a contiguous frontier — any wallet note spendable in a skipped block was either received in an already-scanned block (its spend is detected against the wallet's own nullifiers, not the map) or is received later in the same ascending batch (the spend is linked when the receiving transaction is processed).

Per review discussion, the skip is derived **internally** rather than exposed as a caller parameter, so the API has no unsound-use surface and every consumer of `put_blocks` / `put_blocks_rows` benefits automatically:

- `LowLevelWalletRead` gains `block_fully_scanned_height` (the `ll` traits are unreleased, so this is not a breaking change); `zcash_client_sqlite` implements it by delegating to the existing `block_fully_scanned` query.
- `put_blocks_rows` skips nullifier-map insertion for blocks more than `NULLIFIER_MAP_RETENTION_BLOCKS` (= 100) below the batch end, **only** when `block_fully_scanned_height() == from_state.block_height()`. The retained trailing window keeps the map aligned with an equal `prune_tracked_nullifiers` pruning depth and comfortably exceeds the maximum reorg depth.
- Every out-of-order range — scanning after a gap, recent-first, chain-tip pre-scans — derives no floor and continues to track the nullifiers of every block. The unit tests pin exactly the failure scenario raised in review (a range after a gap discarding a nullifier belonging to a not-yet-scanned note).

Tests: the derivation and gating helpers are unit-tested (including the after-a-gap counter-example), and the full `zcash_client_sqlite` suite (307 tests, including the out-of-order scanning and cross-range spend-detection tests) passes against the new behavior.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -18,7 +18,11 @@ workspace.
   commitment trees by other means can reuse it without also implementing
   `WalletCommitmentTrees`. `put_blocks` itself is unchanged in behavior; it is
   now expressed as `put_blocks_rows` followed by the note commitment tree
-  updates.
+  updates. `put_blocks_rows` takes a `nullifier_tracking_floor: Option<BlockHeight>`
+  argument that lets callers skip nullifier-map insertion for blocks below the
+  floor when they can prove the skipped entries are unobservable (for example,
+  because an immediately following maintenance prune would delete them);
+  `put_blocks` passes `None`, tracking every block as before.
 - The helper functions used by the note commitment tree stage of `put_blocks`
   are now `pub`, so that alternative `ShardStore`-backed stores can reuse the
   exact tree-update logic: `zcash_client_backend::data_api::ll::wallet::`
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -16,13 +16,24 @@ workspace.
   row-writing stage of `put_blocks`, extracted as a public function over
   `LowLevelWalletWrite` so that wallet stores that maintain their note
   commitment trees by other means can reuse it without also implementing
-  `WalletCommitmentTrees`. `put_blocks` itself is unchanged in behavior; it is
-  now expressed as `put_blocks_rows` followed by the note commitment tree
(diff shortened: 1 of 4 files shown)


nuttycom:
@LukasKorba this can be rebased on main now to fix the build failures.

zcash_client_backend/CHANGELOG.md
```diff
@@ -16,9 +16,23 @@ workspace.
   row-writing stage of `put_blocks`, extracted as a public function over
   `LowLevelWalletWrite` so that wallet stores that maintain their note
   commitment trees by other means can reuse it without also implementing
-  `WalletCommitmentTrees`. `put_blocks` itself is unchanged in behavior; it is
-  now expressed as `put_blocks_rows` followed by the note commitment tree
-  updates.
+  `WalletCommitmentTrees`. `put_blocks` is now expressed as `put_blocks_rows`
+  followed by the note commitment tree updates.
+- `zcash_client_backend::data_api::ll::LowLevelWalletRead::block_fully_scanned_height`,
```
nuttycom: This is a trait API change so it belongs in the changed section, not the added section.

nuttycom (dismissed):
Utack modulo changelog update and rebase to fix ci.

zcash_client_backend/CHANGELOG.md
LukasKorba: Moved to the Changed section (the `NULLIFIER_MAP_RETENTION_BLOCKS` constant stays under Added). Also rebased onto main for the zewif fix — `zcash_client_backend` and the full-feature `zcash_client_sqlite` suites pass locally on the rebase.

nuttycom (approved):
utACK

PR #2606, opened 2026-07-11
Support for Ironwood pool migration

nullcopy:


czarcas7ic:
For SignedTransferPczt lets ensure we create an API that returns compact signatures only instead of the full PCZT.

Another thing to keep in mind for the API is that all the splits should be able to be signed all at once for the sake of Keystone (is probably self evident, but am just saying it anyway).

I am not sure if a padding number has been selected, but in the Vizor mega PR that was made strictly for testing, we made this value 16 (see https://github.com/valargroup/vizor-wallet/pull/73 if at all helpful)

zcash_pool_migration/src/types.rs
```diff
@@ -0,0 +1,559 @@
...
+    /// Consumes this prepared transfer, returning the owned serialized PCZT bytes.
+    pub fn into_pczt_bytes(self) -> Vec<u8> {
+        self.pczt_bytes
+    }
+}
+
+/// An unsigned-but-proven PCZT awaiting an external signer (Keystone-style hardware wallet
+/// flow).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UnsignedTransferPczt {
```
dannywillems, line 276: Note/Nit: we can use a type for the state `Unsigned/Signed`.

dannywillems:
https://github.com/zcash/librustzcash/pull/2625 I'm opening this patch to be able to push on the branch cc @nullcopy.

nullcopy:
Subsumed by https://github.com/zcash/librustzcash/pull/2625

PR #2607, opened 2026-07-11
Migrate to the released zewif 1.0.0-rc.2

nuttycom:
This replaces the `[patch.crates-io]` git-main pin of the `zewif` crate with the `1.0.0-rc.2` release from crates.io.

## Changes

- The workspace dependency is now `zewif = "1.0.0-rc.2"` and the git patch is removed. No code changes were required in `zcash_client_sqlite`: rc.2's changes are wire-level (the tagged-union encoding is flattened to `[variant-id, body?]`, and the codec moved to minicbor 2), and the workspace MSRV already satisfies zewif's new 1.88 floor.
- Lock resolution: minicbor 1.1.0 → 2.2.2 and minicbor-derive 0.17.0 → 0.19.4, which force proc-macro2 1.0.106 and syn 2.0.118. `quote` was opportunistically bumped by cargo but is not required to move, so it is pinned back to 1.0.45, which existing audits cover.
- `cargo vet`: with the git patch gone, the `[policy.zewif] audit-as-crates-io = false` entry no longer applies and is removed in favor of a `[[trusted.zewif]]` trusted-publisher entry (nuttycom), matching the existing equihash/zip32-style entries. The minicbor, minicbor-derive, and syn exemptions move in place to the required versions. `cargo vet` passes.

## Compatibility

ZeWIF documents produced against zewif 1.0.0-rc.1 (or the previously pinned git snapshot) do not decode with rc.2; the container version remains 1 while the format is in release-candidate status.

## Verification

All 347 `zcash_client_sqlite` tests pass with `--features zewif`, and `cargo check --workspace` is clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,7 +249,7 @@ checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.100",
+ "syn 2.0.118",
  "synstructure",
 ]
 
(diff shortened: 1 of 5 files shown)


dannywillems (approved):
utACK

PR #2608, opened 2026-07-12
zcash_client_backend: test that an Orchard change fallback is promoted to Ironwood after NU6.3

dannywillems:
Prompted by a review question on [zallet#585](https://github.com/zcash/zallet/pull/585/changes#r3565054090): "I'm not sure what the behavior is if the fallback change pool is Orchard after nu6.3 activation."

The behavior is correct today, but nothing pinned it. `select_change_pool` has unit tests covering the promotion, and `MultiOutputChangeStrategy` delegates the decision to it, but no test exercised that promotion *through* a change strategy. A regression in the wiring between the strategy and `select_change_pool` (for example dropping the `is_nu_active(Nu6_3, ..)` argument at `fees/common.rs:457`) would not have been caught. Repo-wide, no test asserted an Ironwood `ChangeValue` at all.

### Changes

- Add `orchard_fallback_change_pool_is_promoted_to_ironwood_after_nu6_3`. It builds a `MultiOutputChangeStrategy` with `fallback_change_pool: ShieldedPool::Orchard` and shields a single transparent UTXO. Fully transparent flows are the only case in which the fallback actually decides the change pool, since a transaction with shielded flows infers its change pool from the pool it already uses. The test asserts the change pool either side of activation:
  - before NU6.3, the change is returned to Orchard, as value may still enter that pool;
  - after NU6.3, the Orchard turnstile forbids value from entering the Orchard pool, and as the transaction spends no Orchard notes the change is promoted to Ironwood.

  The fee is identical at both heights: the change output is charged to the Ironwood bundle instead of the Orchard one, and each pads to the same two-action floor.

- Add `ChangeValue::ironwood` (behind `orchard`), parallel to the existing `sapling` and `orchard` constructors, so the expected change value can be named.

The test was checked against a mutation: hardcoding `false` for the `ironwood_active` argument to `select_change_pool` makes it fail, reporting Orchard change where Ironwood is required.

### Note on the existing behavior (for the zallet thread)

The promotion is not unconditional, and the complementary cases are already covered:

- an Orchard *spend* keeps its change in Orchard while the pool balance strictly decreases (change bound below the Orchard input value), and only overflows to Ironwood otherwise;
- a Sapling-only spend keeps its change in Sapling; the promotion only ever fires on an Orchard preference.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -161,6 +161,9 @@ workspace.
   such a change value is represented by the existing transparent `valuePool`
   with `isEphemeral` unset; decoding this combination previously returned
   `ProposalDecodingError::InvalidChangeRecipient`.
+- `zcash_client_backend::fees::ChangeValue::ironwood` (behind `orchard`):
+  constructs an Ironwood change value, parallel to `ChangeValue::sapling` and
+  `ChangeValue::orchard`.
 - `zcash_client_backend::data_api::WalletWrite::reserve_next_n_internal_addresses`
   (behind `transparent-inputs`): reserves the next `n` available
   internal-scope (change) transparent addresses for an account, parallel to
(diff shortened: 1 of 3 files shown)


nuttycom (approved):
utACK

PR #2610, opened 2026-07-12
Allow per-pool Orchard protocol bundle types

czarcas7ic:
Replaces `BuildConfig::Standard`'s single `orchard_pool_bundle_type` field with separate `orchard_bundle_type` and `ironwood_bundle_type` fields, so the bundle type can be selected independently for each Orchard protocol value pool (e.g. a padded Orchard bundle with an unpadded Ironwood bundle, as in migration transactions). This follows the per-pool pattern of the existing anchor fields instead of adding an override constructor. `zcash_client_backend` sets both fields from its existing single parameter, so its API is unchanged.

Validated with the full `zcash_primitives` all-features test suite, a no-default-features check, Clippy across all targets for the affected crates, and rustfmt.


nuttycom (approved):
utACK

PR #2615, opened 2026-07-13
zcash_client_backend: Fix `propose_send_max_transfer` transparent-recipient and edge-case handling

### LINKED ISSUE #2614: zcash_client_backend: `propose_send_max_transfer` fails for transparent recipients and mishandles several edge cases
A review of `propose_send_max` in `zcash_client_backend/src/data_api/wallet/input_selection.rs` found the following defects:

1. **Send-max to a plain transparent recipient always fails.** No payment pool is ever assigned for a bare transparent address (or a unified address with no shielded receiver), so every such proposal fails validation with `ProposalError::PaymentPoolsMismatch`, even though the fee computation explicitly supports these recipients. The only transparent-recipient test bails out earlier on the memo check, so this was uncovered.

2. **TEX recipients are not feature-gated.** The `Address::Tex` arm of the fee computation is not gated on `transparent-inputs`. In builds without that feature, the second-step fee is deducted from the payment but no ephemeral step is constructed, and the proposal fails with the same misleading `PaymentPoolsMismatch` instead of a meaningful "TEX requires transparent-inputs" error.

3. **A wallet whose balance exactly equals the fee proposes a useless transaction.** When `input_total == total_fee_required`, the proposal pays the recipient zero and burns the entire balance as fees (shielded recipients), or fails with an unrelated `ZeroValuedTransparentOutput` error (transparent recipients), instead of reporting `InsufficientFunds`.

4. **UA delivery does not fall back to Sapling in no-`orchard` builds.** Orchard-receiver preemption of Sapling delivery is decided via `ZcashAddress::can_receive_as(ORCHARD)`, which reflects the address contents rather than what the build can pay. A payment to a UA with both Sapling and Orchard receivers is assigned no pool at all in a build without the `orchard` feature, failing with `PaymentPoolsMismatch` instead of being delivered via the Sapling receiver. Relatedly, a UA with no payable receiver fails with the same misleading error instead of `UnsupportedAddress`.

5. **Minor:** the Ironwood action count is computed by calling `num_actions` directly with hardcoded `Flags::ENABLED` rather than via `transactional_action_count` with the bundle version in effect at the target height, as every sibling call site does; and the `tr0_fee + tr1_fee` sum is computed with an `expect` that a caller-supplied fee rule can make reachable.

---
_This issue was filed with Claude Code assistance._

nuttycom:
Closes #2614.

A review of `propose_send_max` found that send-max proposals to plain transparent recipients always failed, and that several edge cases were mishandled. Each fix is a separate commit, made test-first (the new pool tests were observed failing with the exact defect before each fix):

1. **Fix send-max proposals to transparent recipients.** No payment pool was ever assigned for a bare transparent address or a unified address with no shielded receiver, so every such proposal failed validation with `ProposalError::PaymentPoolsMismatch`. Such recipients are now assigned `PoolType::Transparent` (TEX recipients excluded — their payment belongs to the ephemeral second step), and the unified-address arm of the fee computation shares the same predicate so the transparent-output accounting and pool assignment cannot drift apart.

2. **Reject send-max to TEX recipients without `transparent-inputs`.** The `Address::Tex` fee arm was not feature-gated; without the feature, the second-step fee was deducted but no ephemeral step was constructed, and the proposal failed with a misleading error. TEX recipients are now rejected up front with `GreedyInputSelectorError::UnsupportedTexAddress`, consistent with `propose_transfer`. This required widening `ProposeSendMaxErrT`'s note-selection error type from `BalanceError` to `GreedyInputSelectorError` (balance errors are now wrapped in its `Balance` variant).

3. **Require send-max proposals to deliver nonzero value.** When the spendable balance exactly equaled the required fee, the proposal paid the recipient zero and consumed the entire balance as fees. This now reports `InsufficientFunds`, with `required` reflecting the minimum total needed to deliver a nonzero amount.

4. **Deliver via Sapling when Orchard is unavailable.** Orchard-receiver preemption of Sapling delivery was decided via `ZcashAddress::can_receive_as(ORCHARD)`, which reflects the address contents rather than the receivers the build can pay. In a no-`orchard` build, a payment to a UA with both receivers was assigned no pool at all. Orchard receivers now take precedence only when the feature is enabled, and a UA with no payable receiver is rejected with `GreedyInputSelectorError::UnsupportedAddress`. Verified end-to-end through transaction creation in a default-features build.

5. **Minor robustness/consistency:** the Ironwood action count now goes through `transactional_action_count` with the bundle version in effect at the target height (behavior-identical for all current versions), and the `tr0_fee + tr1_fee` sum reports `BalanceError::Overflow` instead of panicking when a caller-supplied fee rule produces fees near the maximum monetary amount.

Seven new pool tests cover the fixed behaviors, instantiated for the Sapling and (where applicable) Orchard pool testers, including `cfg(not(feature = ...))` variants for the feature-gated cases. The full `zcash_client_sqlite` suite passes with `orchard transparent-inputs` (313 tests) and with default features (166 tests), and `zcash_client_backend` compiles warning-free across the `orchard`/`transparent-inputs` feature matrix.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -47,6 +47,10 @@ workspace.
   receiver that the build is able to pay is now rejected with
   `GreedyInputSelectorError::UnsupportedAddress` instead of the same misleading
   error.
+- `zcash_client_backend::data_api::wallet::propose_send_max_transfer` now
+  returns `GreedyInputSelectorError::Balance` instead of panicking when a
+  custom fee rule produces per-transaction fees whose sum exceeds the maximum
+  monetary amount.
 
 ## [0.24.0-rc.1] - 2026-07-12
 
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -10,6 +10,16 @@ workspace.
 
 ## [Unreleased]
 
+### Changed
+- `zcash_client_backend::data_api::wallet::ProposeSendMaxErrT` now uses
+  `GreedyInputSelectorError` (instead of `BalanceError`) as its note-selection
+  error type, so that `propose_send_max_transfer` can report
+  `GreedyInputSelectorError::UnsupportedTexAddress` — consistent with
+  `propose_transfer` — when a TEX recipient is requested and the
+  `transparent-inputs` feature is not enabled. Balance errors encountered
+  during send-max input selection are now wrapped in
+  `GreedyInputSelectorError::Balance`.
+
 ### Fixed
 - `zcash_client_backend::data_api::wallet::propose_send_max_transfer` now
   correctly constructs proposals paying a transparent (non-TEX) recipient — a
@@ -17,6 +27,12 @@ workspace.
   Previously, such proposals failed validation with
   `ProposalError::PaymentPoolsMismatch` because no output pool was assigned to
   the payment.
+- When the `transparent-inputs` feature is not enabled,
+  `zcash_client_backend::data_api::wallet::propose_send_max_transfer` now
+  rejects TEX recipients up front with
+  `GreedyInputSelectorError::UnsupportedTexAddress` instead of constructing an
+  internally inconsistent proposal that failed validation with a misleading
+  error.
 
 ## [0.24.0-rc.1] - 2026-07-12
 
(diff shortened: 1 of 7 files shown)


nullcopy (dismissed):
utACK fcc1e64b0778052b61925bde8df78f8152d69632

pacu (dismissed):
utACK. I would add tests that have more notes and not just a single one

nullcopy (approved):
utACK d83c45fe9b2ff5a629c329eb94b426639dc574af

PR #2620, opened 2026-07-14
Return transparent change to originating P2SH address

### LINKED ISSUE #2570: zcash_client_backend: Return transparent change to the originating P2SH address when spending P2SH inputs
Follow-up from https://github.com/zcash/librustzcash/pull/2550#discussion_r3545542604.

`TransparentChangePolicy::TransparentChangeAllowed` currently always returns transparent change to a BIP-44 internal-scope P2PKH address of the spending account, and the ZIP-317 fee computation in `single_pool_output_balance` counts the change output at `P2PKH_STANDARD_OUTPUT_SIZE`. When the inputs being spent are held at a P2SH (e.g. multisig) address, change sent to a wallet-internal P2PKH address would not be spendable under the same multisig arrangement and the change should instead be returned to the originating P2SH address.

This will require:
- surfacing the script type (and originating address) of selected transparent inputs to the change strategy
- sizing the change output in the fee computation according to the destination script
- a change-destination representation in `ChangeValue`/proposal serialization that can carry "return to originating address" semantics rather than implying internal-scope derivation
- corresponding handling in `create_proposed_transactions` (skipping internal-address reservation for such outputs)

Not currently needed for the `z_sendmany`/`z_sendfromaccount` support in `zallet`, which spends from derived P2PKH addresses.

schell:
## Summary

`TransparentChangePolicy::TransparentChangeAllowed` currently always returns transparent change to a BIP-44 internal-scope P2PKH address of the spending account, and the ZIP-317 fee computation in `single_pool_output_balance` counts the change output at `P2PKH_STANDARD_OUTPUT_SIZE`. When the inputs being spent are held at a P2SH (e.g. multisig) address, change sent to a wallet-internal P2PKH address would not be spendable under the same multisig arrangement.

This adds a new `TransparentChangePolicy::ReturnToOriginatingAddress` policy that, when the net flows of a transaction are fully transparent, returns change to the transparent address that originated the value being spent (selecting the address controlling the largest total input value when the selected inputs span more than one address), instead of an internal-scope address of the wallet.

- `fees::TransparentChangeDestination` represents the two possible non-ephemeral transparent change destinations (`InternalP2pkh` and `OriginatingAddress`), and is now carried by `ChangeValue::transparent` in place of the previous no-argument constructor.
- `fees::common::resolve_transparent_change` determines the destination and the correct standard output size (P2PKH or the newly added `P2SH_STANDARD_OUTPUT_SIZE`) for the transparent change output, and is used by `single_pool_output_balance` for both fee sizing and change emission.
- The proposal protobuf `ChangeValue` message gains an optional `transparentChangeAddress` field carrying the encoded destination address; its absence continues to mean `InternalP2pkh`, preserving compatibility with previously serialized proposals.
- `create_proposed_transactions` sends change destined for an `OriginatingAddress` directly to that address, skipping the internal-scope address reservation used for `InternalP2pkh` change.

Not currently needed for the `z_sendmany`/`z_sendfromaccount` support in `zallet`, which spends from derived P2PKH addresses.

Closes #2570

## Testing

- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `cargo check --workspace --all-features --all-targets`
- `cargo check` for `--no-default-features`, `--features orchard`, `--features transparent-inputs` (both `zcash_client_backend` and `zcash_primitives`)
- `cargo test -p zcash_client_backend --all-features` (99 passed, including new unit and integration-style tests for the resolver, fee sizing, and proposal round trip)
- `cargo test -p zcash_client_sqlite --all-features propose_t2t_with_transparent_change` (confirms the existing `InternalP2pkh` path still works end-to-end)

### DIFF
--- a/zcash_client_backend/proto/proposal.proto
+++ b/zcash_client_backend/proto/proposal.proto
@@ -150,8 +150,12 @@ message TransactionBalance {
 // given a unique t-address.
 //
 // When the transparent value pool is selected and the `isEphemeral` field
-// is not set, the value represents transparent change, which will be sent
-// to an internal-scope (change) transparent address of the wallet.
+// is not set, the value represents transparent change. If
+// `transparentChangeAddress` is absent, the change will be sent to an
+// internal-scope (change) transparent address of the wallet; if present, the
+// change will be sent to that address instead (used to return change to the
+// address that originated the transparent value being spent, e.g. when
+// spending from a P2SH address).
 message ChangeValue {
     // The value of a change or ephemeral output to be created, in zatoshis.
     uint64 value = 1;
@@ -162,6 +166,12 @@ message ChangeValue {
     MemoBytes memo = 3;
     // Whether this is to be an ephemeral output.
     bool isEphemeral = 4;
+    // For non-ephemeral transparent change only: the encoded originating
+    // transparent address to which change should be returned, in place of an
+    // internal-scope (change) address of the wallet. Encoded as 21 bytes: a
+    // 1-byte script type (`0x00` for P2PKH, `0x01` for P2SH) followed by the
+    // 20-byte hash. Must be absent for shielded and ephemeral outputs.
+    optional bytes transparentChangeAddress = 5;
 }
 
 // An object wrapper for memo bytes, to facilitate representing the
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -10,6 +10,43 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_client_backend::fees::TransparentChangePolicy::ReturnToOriginatingAddress`
+  (behind the `transparent-inputs` feature flag): when the net flows of a
+  transaction are fully transparent, returns change to the transparent
+  address that originated the value being spent (selecting the address
+  controlling the largest total input value when the selected inputs are
+  controlled by more than one address), instead of an internal-scope P2PKH
+  address of the wallet. This is the appropriate policy when spending from a
+  P2SH (e.g. multisig) address, where change sent to an internal-scope P2PKH
+  address would not be spendable under the same multisig arrangement.
+- `zcash_client_backend::fees::TransparentChangeDestination` (behind
+  `transparent-inputs`): distinguishes the two possible destinations of a
+  non-ephemeral transparent change output, `InternalP2pkh` and
+  `OriginatingAddress`.
+- `zcash_client_backend::fees::ChangeValue::transparent_change_destination`
+  (behind `transparent-inputs`): returns the `TransparentChangeDestination`
+  for a non-ephemeral transparent change value.
+
+### Changed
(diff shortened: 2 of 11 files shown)


nuttycom:
This is also relevant to https://github.com/zcash/zallet/issues/617

zcash_client_backend/proto/proposal.proto
```diff
@@ -162,6 +166,12 @@ message ChangeValue {
     MemoBytes memo = 3;
     // Whether this is to be an ephemeral output.
     bool isEphemeral = 4;
+    // For non-ephemeral transparent change only: the encoded originating
+    // transparent address to which change should be returned, in place of an
+    // internal-scope (change) address of the wallet. Encoded as 21 bytes: a
+    // 1-byte script type (`0x00` for P2PKH, `0x01` for P2SH) followed by the
+    // 20-byte hash. Must be absent for shielded and ephemeral outputs.
+    optional bytes transparentChangeAddress = 5;
```
nuttycom, line 174: This might be better represented as a string with the canonical text encoding of the address; it will be better for wallets to not have to understand the byte encoding.

nuttycom:
Thanks @schell for taking this on. Comparing this implementation side-by-side with a parallel branch surfaced a few gaps here that led us to continue with the other implementation:

1. **No proposal-level validation of the change address.** A deserialized proposal's `transparentChangeAddress` is accepted as-is, so a tampered or malicious serialized proposal can direct change to an arbitrary address, and `create_proposed_transactions` will send it there while recording the output as wallet-owned change (`Recipient::InternalTransparent`). #2637 validates in `Step::from_parts` that an explicit change recipient matches the address of one of the same step's transparent inputs, on both the trusted and untrusted construction paths.

2. **The feature isn't reachable end-to-end.** Spending from an imported P2SH address currently fails before change policy is consulted: `select_spendable_transparent_outputs` and `get_wallet_transparent_output` in `zcash_client_sqlite` don't attach the redeem-script-derived `known_input_size`, so real P2SH spend proposals fail with `UnknownP2shInputs` (the unit tests here bypass this via a mock `InputView::serialized_size`). #2637 fixes that and adds an end-to-end multisig test through `create_proposed_transactions`.

3. **Decode enforcement doesn't match the field's stated contract.** The proto doc says `transparentChangeAddress` "must be absent for shielded and ephemeral outputs," but only the ephemeral case is rejected; the field set on a shielded change value is silently ignored on decode.

4. **Policy semantics.** #2570 describes internal-P2PKH change for P2SH-funded inputs as incorrect behavior of `TransparentChangeAllowed`; #2637 changes that behavior directly, whereas this PR leaves it in place and adds an opt-in policy. Relatedly, the largest-input-value selection for mixed funding sources silently moves residual value between custody arrangements (e.g. from a solo-controlled address into a multisig, or between two different multisigs); #2637 instead refuses with a typed error when P2SH-funded inputs don't share a single originating address and nonzero change is required, leaving the custody decision to the caller's input selection.

Subsumed by #2637.

*This comment was written with Claude Code assistance.*

PR #2625, opened 2026-07-15
Support for Ironwood pool migration

dannywillems:

PR #2626, opened 2026-07-15
zcash_pool_migration: add crate with public migration types

dannywillems:
First slice of splitting #2625 (the Orchard -> Ironwood migration engine, ~7900 lines) into smaller, independently reviewable PRs.

This PR introduces the `zcash_pool_migration` crate containing **only its public data types** and the crate scaffolding. The migration engine (backend, context, store, scheduling, split, denominations, error) follows in subsequent PRs.

## Contents
- Crate manifest (`Cargo.toml`), `README.md`, `CHANGELOG.md`, and the workspace `members` entry.
- `src/lib.rs`: crate docs plus the public re-exports.
- `src/types.rs`: the public data types exchanged with the platform. Private fields with `from_parts`-style constructors and accessor methods; no `serde`.

## Notes
- **Typestate.** A transfer PCZT tracks its signing state at the type level as `TransferPczt<State>` (`Unsigned` / `Signed`). `TransferPczt::into_signed` is the only `Unsigned -> Signed` transition, so the signing step cannot be skipped or the two states confused at a call site. `UnsignedTransferPczt` / `SignedTransferPczt` are aliases for the two states.
- **Strategies, not tests.** Proptest strategies for every public type live in `types::testing`, behind the `test-dependencies` feature, for reuse by the later engine modules and dependent crates.
- Uses `core` instead of `std` where possible.

## Verification
- `cargo test --all-features`
- `cargo fmt --check`

## Follow-ups
- Broaden the typestate approach across the engine (e.g. `MigrationState` driving the per-step methods).
- Land the remaining modules as further slices of #2625.

### DIFF
--- a/zcash_pool_migration/src/types.rs
+++ b/zcash_pool_migration/src/types.rs
@@ -1,0 +1,533 @@
+//! Public data types for the migration engine.
+//!
+//! Amounts use [`Zatoshis`] and heights use [`BlockHeight`] — the canonical `zcash_protocol`
+//! types, which enforce their own invariants — and transaction identifiers use [`TxId`]. Unlike
+//! the `zodl_ironwood_migration` prototype this crate ports from, none of these types derive
+//! `serde`: every type here has private fields with `from_parts`-style constructors and accessor
+//! methods instead, and the platform's FFI/JNI layer is expected to define its own wire-format
+//! DTOs over those accessors.
+//!
+//! Where a value has a signing lifecycle, that lifecycle is encoded in the type system rather than
+//! carried as a runtime flag: a transfer PCZT is a [`TransferPczt<State>`](TransferPczt) whose
+//! `State` is [`Unsigned`] or [`Signed`], and the only way to reach the signed state is the
+//! [`TransferPczt::into_signed`] transition.
+
+use core::fmt;
+use core::marker::PhantomData;
+
+use zcash_protocol::{TxId, consensus::BlockHeight, value::Zatoshis};
+
+/// A run-scoped identifier for a single migration transfer, or for the one-off note-split
+/// ("prep") transaction that precedes them.
+///
+/// Backed by a string of the form `"<run_id>:<index>"` (built by `TransferId::for_transfer`) or
+/// `"prep:<run_id>"` (built by `TransferId::for_prep`), rather than a bare run id, so a single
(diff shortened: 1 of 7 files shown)


zcash_pool_migration/src/types.rs
```diff
@@ -0,0 +1,533 @@
...
+            id,
+            pczt_bytes,
+            _state: PhantomData,
+        }
+    }
+}
+
+/// Top-level migration state machine surfaced to the app.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MigrationState {
```
dannywillems, line 366: I would like to move this into a type-level state. Leaving like this for now.

zcash_pool_migration/src/types.rs
```diff
@@ -0,0 +1,533 @@
...
+    },
+    /// The input note had already been spent.
+    InvalidNote,
+    /// The transaction's anchor/expiry height had already passed.
+    Expired,
+}
+
+/// Proptest strategies for the public types.
+#[cfg(feature = "test-dependencies")]
+pub mod testing {
```
dannywillems, line 411: We can see later if we removed them. I am verbose on purpose regarding the strategies.

zcash_pool_migration/src/types.rs
```diff
@@ -0,0 +1,533 @@
...
+/// Keystone-style hardware wallet flow) or [`Signed`] (the bytes an external signer returned). Both
+/// states carry the same data — the [`TransferId`] they answer and the serialized `pczt::Pczt`
+/// bytes — but are distinct types, so the compiler enforces the lifecycle: an unsigned value can
+/// only become signed through the [`TransferPczt::into_signed`] transition, and an API that
+/// requires a signed PCZT cannot be handed an unsigned one by mistake.
+///
+/// The two states are also exposed as the aliases [`UnsignedTransferPczt`] and
+/// [`SignedTransferPczt`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TransferPczt<State> {
```
dannywillems, line 294: Thinking about using it in a smart way later.

dannywillems:
Superseded by a same-repo PR from the `feat/pool-migration-types` branch on zcash/librustzcash (the branch now lives on the main repo rather than the fork).

PR #2627, opened 2026-07-15
zcash_ironwood_migration_backend: add crate with public migration types

dannywillems:
First slice of splitting #2625 into small, stacked PRs, now under the new two-crate architecture.

The migration engine is being split into two crates:
- **`zcash_ironwood_migration_backend`** (this crate): backend-agnostic, depends only on `pczt` / `zcash_primitives` / `zcash_keys` (+ `zcash_protocol`). It will expose the trait a wallet implements to support Ironwood migration, and holds the shared logic (types, denomination planning, scheduling).
- `zcash_ironwood_migration_sqlite` (later): the SQLite implementation of that trait, depending on `zcash_client_sqlite`.

This PR introduces the `zcash_ironwood_migration_backend` crate with **only its public data types** and the crate scaffolding.

## Contents
- Crate manifest (`Cargo.toml`), `README.md`, `CHANGELOG.md` (own commit), and the workspace `members` entry.
- `src/lib.rs`: crate docs plus the public re-exports.
- `src/types.rs`: the public data types. Private fields with `from_parts`-style constructors and accessor methods; no `serde`.

## Notes
- **Typestate.** A transfer PCZT tracks its signing state at the type level as `TransferPczt<State>` (`Unsigned` / `Signed`); `TransferPczt::into_signed` is the only `Unsigned -> Signed` transition. `UnsignedTransferPczt` / `SignedTransferPczt` are aliases.
- **Strategies, not tests.** Proptest strategies for every public type live in `types::testing`, behind the `test-dependencies` feature.
- Uses `core` instead of `std` where possible; ASCII-only.

## Verification
- `cargo fmt --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7206,6 +7206,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "zcash_ironwood_migration_backend"
+version = "0.1.0"
+dependencies = [
+ "proptest",
+ "zcash_protocol",
+]
+
+[[package]]
 name = "zcash_keys"
 version = "0.15.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "zcash_client_backend",
     "zcash_client_sqlite",
     "zcash_history",
+    "zcash_ironwood_migration_backend",
     "zcash_keys",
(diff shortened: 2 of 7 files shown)


dannywillems:

PR #2628, opened 2026-07-15
zcash_ironwood_migration_backend: add backend-agnostic error types

dannywillems:
Part of the two-crate split of #2625. **Draft. Stacked on #2627** (base `feat/pool-migration-types`).

Adds `MigrationError` + `InvalidStateError` to the backend-agnostic crate. Reworked from the original single-crate error: **no `zcash_client_sqlite` / `rusqlite` / `shardtree` dependency**. A wallet-backend failure or a storage failure is carried as an opaque message (`Backend(String)` / `Store(String)`) rather than a concrete error type, so it names no backend-specific type; a backend maps its own error in at the trait boundary. Keeps the stable `error_code` FFI mapping, `Display`, `core::error::Error`, and derives no `serde`.

Verified: clippy `-D warnings`, doc `-D warnings`, fmt.


dannywillems:

PR #2629, opened 2026-07-15
zcash_ironwood_migration_backend: add note-split denomination planner

dannywillems:
Part of the two-crate split of #2625. **Draft. Stacked on #2628** (base `feat/pool-migration-errors`).

Adds the internal note-split planner: `plan_denominations` decomposes a spendable Orchard balance into self-funding `{1, 2, 5} * 10^k` ZEC denomination notes, leaving any residual (including dust) as Orchard change. Pure `u64` arithmetic, no deps. `DenominationPlan` has private fields + `pub(crate)` accessors. Internal module with no consumer yet -> temporary `#[allow(dead_code)]`.

Verified: clippy `-D warnings`, doc `-D warnings`, fmt.


dannywillems:

PR #2631, opened 2026-07-15
zcash_ironwood_migration_backend: add height-based transfer scheduler

dannywillems:
Part of the two-crate split of #2625. **Draft. Stacked on #2629** (base `feat/pool-migration-denominations`).

Adds the internal transfer scheduler (`build_schedule`): assigns each crossing value a send window and expiry, shares the wallet's natural anchor, and samples the gap between transfers from an exponential distribution. Also adds `zcash_protocol::consensus::{SECONDS_PER_BLOCK, BLOCKS_PER_HOUR}` so cadence/expiry are expressed in hours and converted to blocks from a protocol-owned source. Deps: `rand` + `rand_distr`. Internal module -> temporary `#[allow(dead_code)]`.

Verified: clippy `-D warnings` (both crates), doc `-D warnings`, fmt.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7210,6 +7210,8 @@ name = "zcash_ironwood_migration_backend"
 version = "0.1.0"
 dependencies = [
  "proptest",
+ "rand 0.8.5",
+ "rand_distr",
  "zcash_protocol",
 ]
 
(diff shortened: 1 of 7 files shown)


dannywillems:

PR #2633, opened 2026-07-15
zcash_client_sqlite: Order add_transparent_value_index after account_delete_cascade

### LINKED ISSUE #2632: zcash_client_sqlite: idx_transparent_received_outputs_value_zat survival depends on migration tie-breaking order
The `add_transparent_value_index` migration (which creates
`idx_transparent_received_outputs_value_zat`) declares a dependency only on
`fix_transparent_received_outputs`. However, the later `account_delete_cascade`
migration also rebuilds the `transparent_received_outputs` table (via `CREATE
TABLE` + `INSERT INTO ... SELECT` + `DROP TABLE` + `ALTER TABLE ... RENAME
TO`), which destroys any index on the old table, and it does not recreate the
value_zat index (it predates it historically, so it only recreates the indexes
that existed at its point in the DAG).

Since `add_transparent_value_index` and `account_delete_cascade` are
incomparable in the migration DAG, whether the index survives a full migration
run depends on the migrator's topological tie-breaking. On current `main` the
iteration order happens to run `account_delete_cascade` first, so the index
survives and `wallet::init::tests::verify_schema` passes — but any change to
the migration set can perturb the tie-break and drop the index. I hit this
while rebasing a feature branch that adds a new migration: with the new
migration present, `add_transparent_value_index` ran before
`account_delete_cascade`, the rebuild silently destroyed the index, and
`verify_schema` failed.

The fix is the same pattern already used by
`add_transparent_receiver_address_index` for the `addresses` table: an
index-creating migration must depend on the topologically-latest migration
that rebuilds its table — here, `account_delete_cascade`.

---

nuttycom:
Closes #2632.

`account_delete_cascade` rebuilds the `transparent_received_outputs` table
without recreating `idx_transparent_received_outputs_value_zat`, but nothing
ordered it before `add_transparent_value_index`: the two migrations were
incomparable in the migration DAG, so whether the index survived a full
migration run depended on the migrator's topological tie-breaking. Current
`main` happens to run them in the safe order, but adding any new migration can
perturb the tie-break, silently drop the index, and fail
`wallet::init::tests::verify_schema` (this surfaced while rebasing a feature
branch that adds a new migration).

This changes `add_transparent_value_index` to depend on
`account_delete_cascade` — the topologically-latest rebuild of the table —
following the same pattern used by `add_transparent_receiver_address_index`
for the `addresses` table. The earlier rebuilds
(`fix_transparent_received_outputs`, `utxos_to_txos`) are reachable
transitively and no longer need to be listed. The migration DAG comment in
`migrations.rs` is updated to match.

Both migrations already ship in published releases, but adding an edge between
them is safe for existing wallets: any wallet that has applied
`add_transparent_value_index` has also applied `account_delete_cascade` (or
will, before the migrator considers the DAG satisfied), and reordering only
affects fresh migration runs.

---

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations.rs
@@ -131,8 +131,8 @@ pub(super) fn all_migrations<
     //                        \      ensure_default_transparent_address    /                                 \
     //                         \                     |                    /                                   \
     //                          `---- fix_transparent_received_outputs --'                                     \
-    //                                    /         /           \    \                                         |
-    //                                   /         /             \   add_transparent_value_index               |
+    //                                    /         /           \                                              |
+    //                                   /         /             \                                             |
     //                                  /         /               \                                            |
     //           support_zcashd_wallet_import    /             fix_v_transactions_expired_unmined              |
     //                \                         /                    /        |         \                      |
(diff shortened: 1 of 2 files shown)


nuttycom (commented):
self-utACK

nullcopy (approved):
utACK 4e30befab3b8499957d5c123aee068b2d8a2ddac

PR #2634, opened 2026-07-15
zcash_ironwood_migration_backend: add WalletMigrationBackend trait

dannywillems:
Part of splitting #2625 into two crates under the new architecture. **Stacked on #2627** (base branch `feat/pool-migration-types`); GitHub retargets to `main` as the chain merges.

Adds the **backend-agnostic wallet interface** the migration engine will be generic over. This is the linchpin of the two-crate split: the shared migration logic lives in `zcash_ironwood_migration_backend` and drives *any* wallet backend through this trait; `zcash_ironwood_migration_sqlite` (and later an in-memory backend, like `zcash_client_memory`) implement it.

## `wallet::WalletMigrationBackend`
```rust
pub trait WalletMigrationBackend {
    type Error: core::error::Error;
    fn pool_balances(&self) -> Result<PoolBalances, Self::Error>;
    fn target_and_anchor_heights(&self) -> Result<(u32, u32), Self::Error>;
    fn is_tx_mined(&self, txid: TxId) -> Result<bool, Self::Error>;
    fn build_note_split_pczt(&mut self, target_values: &[u64], reserved: &[NoteRef])
        -> Result<(pczt::Pczt, SplitOutputs), Self::Error>;
    fn build_transfer_pczt(&mut self, crossing_value: u64, reserved: &[NoteRef],
        target_height: u32, expiry_height: u32) -> Result<Option<TransferBuild>, Self::Error>;
    fn sweep_crossing_value(&self) -> Result<Option<u64>, Self::Error>;
    fn retain_anchor(&mut self, height: u32) -> Result<(), Self::Error>;
    fn release_retained_anchors_below(&mut self, height: u32) -> Result<(), Self::Error>;
}
```

## Design principle (keeps the crate agnostic)
- **No wallet-backend-specific type appears in the interface.** Note references cross as `NoteRef` (`(TxId, output_index)`), never `ReceivedNoteId`; build methods return an **unproven `pczt::Pczt`** plus plain-data metadata. All input selection / `create_pczt_from_proposal` / commitment-tree access stays inside the implementation.
- The engine does the **pure** PCZT prove/sign/finalize/extract itself (those need only `pczt`/`orchard`/keys), so builds return *unproven* PCZTs.
- Associated `type Error: core::error::Error` keeps the error backend-agnostic.
- Plain-data types (`PoolBalances`, `NoteRef`, `SpentNote`, `SplitOutputs`, `TransferBuild`) follow the crate convention: private fields, accessors, no `serde`.

[...]

PR #2635, opened 2026-07-15
zcash_ironwood_migration_backend: add migration Phase

dannywillems:
Part of the two-crate split of #2625. **Draft. Stacked on #2634** (base `feat/pool-migration-wallet-trait`).

Adds `state::Phase`, a migration run's fine-grained 14-value phase, with its stable persisted string form (`as_str` / `parse`). This is the prerequisite for the `MigrationStore` trait (its row types carry `Phase`).

Backend-agnostic: `Phase` names only the phases; it references no wallet or storage type. The `Phase -> MigrationState` mapping is engine logic and lands with the generic `MigrationContext`.

Notes:
- Factual module doc (dropped the "vizor's internal phase / Ported from" phrasing; crate-level attribution lives in `lib.rs`).
- No tests; the fixed-value phase tests were dropped.

## Verification
- `cargo clippy -p zcash_ironwood_migration_backend --all-features --all-targets -- -D warnings`
- `cargo fmt --check`


dannywillems:

PR #2637, opened 2026-07-15
zcash_client_backend: Return transparent change to the originating P2SH address when spending P2SH inputs

### LINKED ISSUE #2570: zcash_client_backend: Return transparent change to the originating P2SH address when spending P2SH inputs

nuttycom:
## Summary

Under `TransparentChangePolicy::TransparentChangeAllowed`, when a transaction's transparent inputs are all funded by a single P2SH (e.g. multisig) address, ZIP 317 change selection now returns transparent change to that originating address — sized by its actual output script — instead of deriving a fresh internal-scope P2PKH change address that would not be spendable under the same multisig arrangement.

- `ChangeValue`'s non-ephemeral transparent variant carries an optional explicit recipient (`ChangeValue::transparent_to_address` / `ChangeValue::transparent_recipient`); the no-recipient form retains internal-scope-derivation semantics.
- `single_pool_output_balance` resolves the change destination from the transparent inputs' funding addresses. When P2SH-funded inputs do not all share a single originating address, change computation fails with the new `ChangeError::TransparentChangeDestinationAmbiguous` — lazily, only when a nonzero transparent change output must actually be emitted, so exact-match (changeless) spends from mixed sources still succeed. P2PKH-only flows are unaffected.
- `Step::from_parts` rejects a transparent change output whose explicit recipient is not the address of one of the same step's transparent inputs (`ProposalError::TransparentChangeRecipientMismatch`), on both trusted and untrusted (deserialized) construction paths, so a tampered proposal cannot redirect change.
- Serialized proposals carry the recipient in a new optional `transparentChangeRecipientScript` field (the destination scriptPubKey); absence preserves the legacy internal-scope semantics, so `PROPOSAL_SER_V1` is unchanged and proposals serialized by older versions decode as before. Malformed or misplaced scripts yield the new `ProposalDecodingError::TransparentChangeRecipientInvalid`.
- `create_proposed_transactions` sends explicit-recipient change directly to that address, skipping internal-scope address reservation (which observes a gap limit and cannot be undone). The PCZT construction path shares this code and records such change correctly (the address is recovered from the output script at extraction).
- `zcash_client_sqlite`: the end-to-end test exposed that `select_spendable_transparent_outputs` and `get_wallet_transparent_output` failed to attach the redeem-script-derived input size for imported P2SH addresses, causing P2SH spend proposals to fail with `UnknownP2shInputs` (or under-count fees). The size attachment is now centralized in the shared row converter, and a recorded redeem script that cannot be parsed or sized surfaces `CorruptedData` instead of silently falling back to the P2PKH size estimate.

Test coverage: ZIP-317 sizing/ambiguity unit tests (constructed so that incorrect P2PKH sizing of the change output produces a different fee), `Step::from_parts` validation tests (exercised in the transparent-inputs-only CI lane), proposal serialization round-trip plus rejection tests, and an end-to-end test that imports a multisig P2SH address, spends from it, verifies the change output pays the originating P2SH script with the exact ZIP-317 fee, verifies no internal-scope address was newly exposed, and confirms the returned change is recorded as spendable.

Closes #2570

## Testing

- `cargo test -p zcash_client_backend --all-features` (99 unit tests + doctests)
- `cargo test -p zcash_client_sqlite --all-features` (379 + 4 tests)
- `cargo check -p zcash_client_backend --no-default-features` (also `--features orchard`, `--features transparent-inputs`)
- `cargo clippy -p zcash_client_backend --all-features --tests` (no new warnings)

### DIFF
--- a/zcash_client_backend/proto/proposal.proto
+++ b/zcash_client_backend/proto/proposal.proto
@@ -151,7 +151,8 @@ message TransactionBalance {
 //
 // When the transparent value pool is selected and the `isEphemeral` field
 // is not set, the value represents transparent change, which will be sent
-// to an internal-scope (change) transparent address of the wallet.
+// to an internal-scope (change) transparent address of the wallet, unless
+// `transparentChangeRecipientScript` is set.
 message ChangeValue {
     // The value of a change or ephemeral output to be created, in zatoshis.
     uint64 value = 1;
@@ -162,6 +163,14 @@ message ChangeValue {
     MemoBytes memo = 3;
     // Whether this is to be an ephemeral output.
     bool isEphemeral = 4;
+    // The scriptPubKey of the transparent address to which this change output is to be
+    // sent. This may only be set when `valuePool` is `Transparent` and `isEphemeral` is
+    // false, and it must be the scriptPubKey of an address that funds one or more of the
+    // transparent inputs of the same proposal step; it is used to return change to its
+    // originating address (e.g. when spending P2SH inputs). When this field is omitted,
+    // transparent change is sent to an internal-scope (change) transparent address of
+    // the wallet. Proposals serialized by older versions omit this field.
+    optional bytes transparentChangeRecipientScript = 5;
 }
 
 // An object wrapper for memo bytes, to facilitate representing the
--- a/zcash_client_backend/src/fees/common.rs
+++ b/zcash_client_backend/src/fees/common.rs
@@ -22,6 +22,10 @@ use super::{
 use super::TransparentChangePolicy;
 #[cfg(feature = "orchard")]
 use super::orchard as orchard_fees;
+#[cfg(feature = "transparent-inputs")]
+use ::transparent::address::{Script, TransparentAddress};
+#[cfg(feature = "transparent-inputs")]
+use zcash_script::script;
 
 pub(crate) struct NetFlows {
     t_in: Zatoshis,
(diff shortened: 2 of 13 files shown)


zcash_client_backend/proto/proposal.proto
```diff
@@ -162,6 +163,14 @@ message ChangeValue {
...
     // Whether this is to be an ephemeral output.
     bool isEphemeral = 4;
+    // The scriptPubKey of the transparent address to which this change output is to be
+    // sent. This may only be set when `valuePool` is `Transparent` and `isEphemeral` is
+    // false, and it must be the scriptPubKey of an address that funds one or more of the
+    // transparent inputs of the same proposal step; it is used to return change to its
+    // originating address (e.g. when spending P2SH inputs). When this field is omitted,
+    // transparent change is sent to an internal-scope (change) transparent address of
+    // the wallet. Proposals serialized by older versions omit this field.
+    optional bytes transparentChangeRecipientScript = 5;
```
nuttycom, line 173: Should this be the scriptpubkey, or can it just be the (text-encoded) address?

schell:
👍 utACK [60abd4d](https://github.com/zcash/librustzcash/pull/2637/commits/60abd4d3dfcda0937a12dffeaae52cd4845a6db2)

zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1139,7 +1139,33 @@ fn to_unspent_transparent_output(
...
+    // If the address has a recorded redeem script, compute the known input size so that the
+    // ZIP 317 fee calculator can size P2SH inputs. The `imported_transparent_receiver_script`
+    // column must be present in the caller's query for this to take effect; callers that do not
+    // select it (an `Err` from `row.get`) or rows without a redeem script leave the size
+    // unknown, falling back to the standard P2PKH estimate. However, a recorded redeem script
+    // that cannot be parsed or sized indicates database corruption (import validation only
+    // permits standard, sizeable scripts to be stored): silently falling back would under-size
+    // the input and could produce an under-fee'd transaction, so it is an error instead.
+    if let Ok(Some(rs_bytes)) =
+        row.get::<_, Option<Vec<u8>>>("imported_transparent_receiver_script")
```
nullcopy: if `row.get()` returns `NULL`, wouldn't that also cause the silent fall-back and under-size issue? Should that be a hard error instead of letting an `Ok()` value?

zcash_client_backend/src/fees/common.rs
nullcopy: We should no longer just assume P2PKH inputs here

nullcopy (changes requested):
Left a few comments. Also need to fix CI.
Reviewed at 60abd4d3dfcda0937a12dffeaae52cd4845a6db2

PR #2639, opened 2026-07-16
zcash_client_sqlite: Add extension-table writes within wallet transactions

### LINKED ISSUE #2638: zcash_client_sqlite: Allow extension-table writes to be atomic with wallet operations
## Problem

`zcash_client_sqlite` supports applications creating their own tables inside the wallet database via `WalletMigrator::with_external_migrations`, with the documented requirement that all external schema use the `ext_` prefix. However, there is no way to *write* to those tables atomically with a wallet operation:

- `WalletDb`'s connection is (deliberately) private, so the API can enforce its invariants.
- `WalletWrite` methods such as `import_account_ufvk` open their own transaction internally, so they cannot run inside a caller-held `rusqlite` transaction.
- The `WalletDb<SqlTransaction>` handle passed to the `transactionally()` closure exposes no way to execute application SQL.

So an application that needs to pair a wallet write with a write to its own table is forced into two separate commits, and has to hand-roll ordering/compensation logic for the window between them.

### Motivating example

Zallet stores encrypted spending key material in `ext_zallet_keystore_*` tables in the wallet database, while the corresponding account lives in the `zcash_client_sqlite` tables. Its `z_importkey` implementation must perform a keystore INSERT and an `import_account_ufvk` in lock-step; because they cannot share a transaction, a failure between the two writes leaves the wallet in an inconsistent state (see zcash/wallet#615 — an exportable key that is never scanned). `z_importaddress` has the same shape with standalone transparent keys. Compensating with `delete_account` is not viable either: `add_account` rewinds the wallet-wide scan queue and note commitment tree state, which `delete_account` does not restore.

## Proposed solution

Add a variant of `transactionally` that additionally passes the closure a restricted statement executor scoped to the same transaction:

```rust
impl<C: BorrowMut<rusqlite::Connection>, P, CL, R> WalletDb<C, P, CL, R> {
    pub fn transactionally_with_extension<F, A, E: From<rusqlite::Error>>(&mut self, f: F) -> Result<A, E>
    where
        F: FnOnce(
            &mut WalletDb<SqlTransaction<'_>, &P, &CL, &mut R>,
            &ExtensionTransaction<'_>,
        ) -> Result<A, E>;
}
```

`ExtensionTransaction` exposes `execute` and `query_row`, and runs every statement under a SQLite authorizer (installed only for the duration of that statement, so wallet statements are never subject to it) that:

- **allows** reads against any table, so extension statements can reference wallet rows (e.g. a foreign key into `accounts`);
- **allows** `INSERT`/`UPDATE`/`DELETE` only against tables whose names begin with the `ext_` prefix already reserved for external migrations; and
- **denies** everything else — DDL, `PRAGMA`, `ATTACH`/`DETACH`, and transaction control — so extension SQL cannot alter the wallet schema or interfere with the enclosing transaction.

This keeps all wallet-table mutations flowing exclusively through `WalletDb`'s methods (the invariants the private connection protects), while making the `ext_` prefix an enforced ownership boundary rather than only a naming convention. Because the boundary is the prefix, the policy stays correct as the wallet schema evolves.

Usage then looks like:

```rust
wallet_db.transactionally_with_extension(|wdb, ext| {
    let account = wdb.import_account_ufvk(name, &ufvk, &birthday, purpose, None)?;
    ext.execute(
        "INSERT INTO ext_myapp_keys (key_id, material) VALUES (:id, :material)",
        named_params! { ... },
    )?;
    Ok(account)
})
```

Known limitations of the proposed policy, for discussion:

- Extension tables using `AUTOINCREMENT` cannot be written through this API, because SQLite services `AUTOINCREMENT` via the internal `sqlite_sequence` table, which the prefix rule denies. A plain `INTEGER PRIMARY KEY` rowid or an explicit key works.
- The initial surface is `execute` + `query_row`. A `prepare`-style API would need the authorizer to remain applied across the `Statement`'s lifetime (rows are evaluated lazily), which is left for future work.
- Reads of wallet tables are allowed wholesale; if that is considered too broad it could be narrowed, though foreign-key enforcement legitimately requires reading referenced wallet rows.

I have an implementation of this (with tests covering commit/rollback atomicity and the authorizer denials) that I'll open as a PR.

nullcopy:
Adds `WalletDb::transactionally_with_extension`, which runs wallet operations and writes to application-owned extension tables atomically in a single database transaction. The closure receives an `ExtensionTransaction` handle sharing that transaction, so an application can pair a wallet write with writes to its own `ext_`-prefixed tables and have both commit or roll back together.

Each statement run through the handle executes under a SQLite authorizer installed only for the duration of that statement. The authorizer allows reads against any table, allows INSERT/UPDATE/DELETE only against tables whose names begin with the `ext_` prefix reserved for external migrations, and denies everything else, including DDL, PRAGMA, ATTACH/DETACH, and transaction control.

Closes https://github.com/zcash/librustzcash/issues/2638
[COR-1583](https://linear.app/zodl/issue/COR-1583)

### DIFF
--- a/zcash_client_sqlite/src/lib.rs
+++ b/zcash_client_sqlite/src/lib.rs
@@ -301,6 +301,139 @@ impl<'a> Borrow<rusqlite::Transaction<'a>> for SqlTransaction<'a> {
     }
 }
 
+/// The prefix reserved for schema (tables, indices, views, ...) created by external
+/// migrations.
+///
+/// The `zcash_client_sqlite` schema promises never to use this prefix for any of its own
+/// names, so any object whose name begins with it is owned by an application rather than
+/// by the wallet.
+const EXTENSION_SCHEMA_PREFIX: &str = "ext_";
+
+/// A restricted statement executor for writing to application-owned extension tables
+/// within a wallet database transaction.
+///
+/// A handle of this type is provided by [`WalletDb::transactionally_with_extension`]
+/// alongside the wallet handle, and shares the same database transaction: statements run
+/// through it either commit atomically with the wallet operations performed in the same
+/// closure, or are rolled back together with them.
+///
+/// # Authorization policy
+///
+/// Every statement executed through this type runs under a SQLite authorizer that is
+/// installed only for the duration of that single statement. The authorizer:
+///
+/// - **allows** reads (`SELECT`, and reads of individual rows and columns) against any
+///   table, so that extension statements may reference wallet data (for example, to
+///   satisfy a foreign key into an account row);
+/// - **allows** `INSERT`, `UPDATE`, and `DELETE` only against tables whose names begin
+///   with the `ext_` prefix reserved for external migrations (see
+///   [`WalletMigrator::with_external_migrations`]); and
+/// - **denies** everything else, including all schema changes (DDL), `PRAGMA`,
+///   `ATTACH`/`DETACH`, and transaction-control actions (`BEGIN`, `COMMIT`, `ROLLBACK`,
+///   `SAVEPOINT`, `RELEASE`), so that extension statements cannot alter the wallet schema
+///   or interfere with the enclosing transaction.
+///
+/// Because writes are restricted to the `ext_` prefix, a statement that inserts into an
+/// `AUTOINCREMENT` extension table is denied: SQLite services `AUTOINCREMENT` by writing
+/// to the internal `sqlite_sequence` table, which does not carry the prefix. Extension
+/// tables that must be written through this API should therefore avoid `AUTOINCREMENT`
+/// (an ordinary `INTEGER PRIMARY KEY` rowid, or an explicitly supplied key, works
+/// without it).
+///
+/// [`WalletMigrator::with_external_migrations`]: crate::wallet::init::WalletMigrator::with_external_migrations
+pub struct ExtensionTransaction<'conn> {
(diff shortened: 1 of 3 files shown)


zcash_client_sqlite/src/lib.rs
```diff
@@ -301,6 +301,139 @@ impl<'a> Borrow<rusqlite::Transaction<'a>> for SqlTransaction<'a> {
...
+}
+
+impl<'conn> ExtensionTransaction<'conn> {
+    /// Runs `f` with the extension authorizer installed on the connection, removing it
+    /// again (even on error or panic) before returning.
+    fn with_authorizer<T>(
+        &self,
+        f: impl FnOnce() -> Result<T, rusqlite::Error>,
+    ) -> Result<T, rusqlite::Error> {
+        self.conn.authorizer(Some(extension_authorizer));
```
nuttycom, line 399: Huh, neat! I had no idea that this was something that rusqlite supports, that's great.

nuttycom (approved):
utACK  3adfaac98eac9c46a3875b50a8fee7b3e152853d

zcash_client_sqlite/src/lib.rs
nullcopy: Yeah, neither did I! This was definitely a wow moment for me

PR #2641, opened 2026-07-16
zcash_ironwood_migration_backend: add pure PCZT pipeline

dannywillems:
Part of the two-crate split of #2625, and the first slice of the generic engine. **Draft. Stacked on #2631** (base `feat/pool-migration-scheduling`).

Adds the **pure PCZT pipeline** the engine drives a built transaction through: `shielded_proving_key`, `prove_pczt`, `sign_all_orchard_spends`, `prove_sign_finalize`, `combine_signed_pczt` (external-signer path), `extract_broadcast_tx`, and the `SignedPcztOutcome` result.

These act on a `pczt::Pczt` and need only `pczt` / `orchard` / `zcash_keys` and the spending key -- no wallet database, no storage -- so the proving/signing logic is backend-agnostic. Adds `orchard` + `zcash_keys` deps and the `pczt` prover/signer/spend-finalizer/tx-extractor roles.

Internal module (`pub(crate)`), no consumer among the extracted slices yet (its caller is the generic `MigrationContext`), so a temporary `#[allow(dead_code)]`.

## Verification
- Builds clean; the crate itself is clippy-clean under `--all-features` (any residual clippy findings are inside the `zcash_keys`/`orchard`/`pczt` deps under a per-crate partial-feature build, which the whole-workspace `--all-features` CI clippy does not hit -- other workspace crates already drive those deps to full features).
- `cargo fmt --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7209,9 +7209,12 @@ dependencies = [
 name = "zcash_ironwood_migration_backend"
 version = "0.1.0"
 dependencies = [
+ "orchard",
+ "pczt",
  "proptest",
  "rand 0.8.5",
  "rand_distr",
+ "zcash_keys",
  "zcash_protocol",
 ]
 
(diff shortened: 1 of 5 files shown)


dannywillems:

PR #2642, opened 2026-07-16
Make Step::orchard_action_count NU6.3-aware

### LINKED ISSUE #2515: zcash_client_backend: Update Step::orchard_action_count for NU6.3 updates
> Note that this will change after Ironwood activation; the orchard transaction builder will create `spends + outputs` actions for Orchard, as the cross-address restriction requires any change output's corresponding dummy input to be signed with the spending key corresponding to the internal IVK (to which the change output is sent). 

 _Originally posted by @nuttycom in [#2513](https://github.com/zcash/librustzcash/pull/2513/changes#r3507172285)_

### LINKED ISSUE #2618: `Step::orchard_action_count` hardcodes the legacy `max(spends, outputs)` formula
Residual API-hygiene item from zcash/librustzcash#2518.

`Step::orchard_action_count` in `zcash_client_backend/src/proposal.rs` still<br>computes the Orchard action count as

```rust
let spends = self.input_count_in_pool(PoolType::ORCHARD);
let outputs = self.output_count_in_pool(PoolType::ORCHARD)
    + self.change_count_in_pool(PoolType::ORCHARD);
spends.max(outputs)
```

This is the legacy (pre-NU6.3) padded-bundle formula. It takes no consensus-branch<br>or bundle-version argument, so for a post-NU6.3 Orchard-pool bundle (where<br>cross-address spend/output pairing is disabled and the action count is<br>`spends + outputs`) it understates the action count. It also has no Ironwood<br>counterpart.

Mitigating context: the method currently has no callers anywhere in the repository.<br>The fee and builder paths compute action counts version-awarely via<br>`fees::orchard::transactional_action_count` (which consults<br>`BundleVersion::default_flags`), and [#2513](<https://github.com/zcash/librustzcash/issues/2513>) gave `Step` generic per-pool helpers<br>(`input_count_in_pool`, `output_count_in_pool`, `change_count_in_pool`) that cover<br>Ironwood. So this is stale public API rather than a live bug — but any external<br>consumer calling it to estimate post-NU6.3 fees would get a wrong answer.

Suggested resolution, either:

* make it branch/version-aware (it would need a `BundleVersion` or target-height +<br>`Parameters` argument, mirroring `transactional_action_count`), or
* deprecate and remove it, pointing users at the per-pool count helpers plus<br>`fees::orchard::transactional_action_count`.

dannywillems:
Closes #2515. Closes #2618.

## Problem

`Step::orchard_action_count` hardcoded the legacy padded-bundle formula:

```rust
spends.max(outputs)
```

From NU6.3 the Orchard pool disables cross-address transfers: a change output's
corresponding dummy input must be signed with the spending key for the internal
IVK the change is sent to, so it cannot share an action with a spend of a note
belonging to a different address. The builder therefore pairs each requested
spend and output with a fabricated zero-valued counterpart, and the bundle
requires `spends + outputs` actions.

The method took no consensus-branch or bundle-version argument, so it
understated the action count for any post-NU6.3 Orchard-pool bundle, and had no
Ironwood counterpart. It has no callers in this repository, so this was stale
public API rather than a live bug, but an external consumer using it to estimate
post-NU6.3 fees would have underpaid.

## Approach

#2618 offered two resolutions: make the method version-aware, or deprecate and
remove it. This takes the first, matching #2515's framing, which resolves both
issues together.

`orchard_action_count` now takes the `BundleType` and `BundleVersion` the
transaction builder will be configured with, mirroring
`fees::orchard::transactional_action_count`, to which it delegates. The version
supplies the flags deciding whether spends and outputs pair; the bundle type
supplies the padding applied on top (`DEFAULT` pads to the ZIP 317 two-action
floor, `UNPADDED` does not). `ironwood_action_count` is added as the
counterpart. Both are gated on the `orchard` feature.

This is a breaking signature change, noted in the CHANGELOG; there are no
in-repo callers to update.

## Constants

The tests needed a realistic ceiling on a bundle's actions, which meant two
sizes that were magic numbers or absent:

- `zcash_protocol::constants::MAX_BLOCK_BYTES`, replacing a private copy in
  `input_selection`.
- `zcash_primitives::transaction::components::orchard::ACTION_SIZE`, which
  existed nowhere despite being what a caller must divide a size budget by to
  bound an action count. It is composed from the sizes of the parts
  `write_action_without_auth` writes, not stated as a literal: the ciphertext
  sizes come from `zcash_note_encryption`, and each Pallas element is named
  (`VALUE_COMMITMENT_BYTE_SIZE`, `NULLIFIER_BYTE_SIZE`, ...). Those per-element
  sizes belong upstream beside the types in `orchard`, which defines all of them
  but exposes no constant for any of their encodings; a comment says so.

`ACTION_SIZE` excludes each action's separately-encoded spend authorization
signature and its share of the bundle proof, so it bounds an action count from
above rather than exactly. That is documented on the constant.

## Tests

- `orchard_action_count_grows_at_nu6_3_activation_height` (proptest): resolves
  the bundle version the way the fee and builder paths do (height -> consensus
  branch -> `BundleVersion`) across the pre-NU6.3 upgrades and NU6.3, asserting
  `max(spends, change)` before and `spends + change` from NU6.3, and that
  Ironwood still pairs. Verified to fail against the old formula (minimal input
  `spends = 1, change = 1` at NU6.3).
- `nu6_3_action_growth_at_scale_depends_on_step_shape`: at the block ceiling the
  change costs a consolidation one extra action (2439 -> 2440), but doubles a
  balanced step, halving what one bundle can carry (2439 spends and outputs
  before, 1219 of each after).
- `counts_beyond_the_block_limit_are_counted_exactly`: a step can be rebuilt from
  a serialized proposal, so its output and change counts are attacker-influenced.
  One claiming roughly eight times a block's capacity must still be counted
  exactly, so the fee rule sees an unaffordable number rather than a wrapped-around
  affordable one.
- `spends_cannot_be_charged_to_a_coinbase_bundle`: the documented error path, and
  the only error a step can provoke, since `default_flags` always enables spends
  and outputs.
- Padding: the whole `u8` floor range including the degenerate zero floor, plus
  `bundle_required` on an empty step.
- Note and change values are generated rather than hardcoded: an action count is
  a function of how many notes and outputs a step has, never of what they are
  worth.
- `action_size_matches_the_encoding` (proptest, `zcash_primitives`): measures a
  real action and holds `ACTION_SIZE` and every per-element size to what the
  encoder actually writes, field by named field.

Also gates an unrelated `TargetHeight` test import in `fees::zip317` on both the
`orchard` and `transparent-inputs` features, matching its only consumer; it
previously warned as unused when building the tests with `orchard` alone. Clippy
is clean across all four combinations of those features, and the `zcash_primitives`
`no_std` build still passes.

### DIFF
--- a/components/zcash_protocol/src/constants.rs
+++ b/components/zcash_protocol/src/constants.rs
@@ -40,3 +40,12 @@ pub const V5_VERSION_GROUP_ID: u32 = 0x26A7270A;
 pub const V6_TX_VERSION: u32 = 6;
 /// The version group ID for Zcash v6 transactions.
 pub const V6_VERSION_GROUP_ID: u32 = 0xD884B698;
+
+/// The maximum size in bytes of a Zcash block, and therefore the maximum size of any single
+/// transaction within one.
+///
+/// It is specified as `MAX_BLOCK_SIZE` in
+/// [§ 7.6 Block Header Encoding and Consensus](https://zips.z.cash/protocol/protocol.pdf#blockheader).
+///
+/// This constant is called `MAX_BLOCK_SIZE` in the zcashd source.
+pub const MAX_BLOCK_BYTES: usize = 2_000_000;
--- a/zcash_client_backend/src/proposal.rs
+++ b/zcash_client_backend/src/proposal.rs
@@ -957,10 +957,14 @@ mod tests {
         value::NoteValue,
     };
     use proptest::prelude::*;
-    use zcash_primitives::transaction::{TxId, components::orchard::bundle_version_for_branch};
+    use zcash_primitives::transaction::{
+        TxId,
+        components::orchard::{ACTION_SIZE, bundle_version_for_branch},
+    };
     use zcash_protocol::{
         PoolType, ShieldedPool,
         consensus::{BlockHeight, BranchId, Network, NetworkUpgrade, Parameters},
+        constants::MAX_BLOCK_BYTES,
         value::Zatoshis,
     };
     use zip321::TransactionRequest;
@@ -1068,7 +1072,7 @@ mod tests {
                 TransactionRequest::empty(),
                 BTreeMap::new(),
                 vec![],
-                shielded_inputs_for(orchard_and_ironwood_notes(1, 0)),
+                shielded_inputs_for(orchard_and_ironwood_notes((1, 10_000), (0, 0))),
                 None,
                 vec![],
                 TransactionBalance::new(vec![], Zatoshis::const_from_u64(10_000)).unwrap(),
@@ -1130,7 +1134,7 @@ mod tests {
         // not the turnstile, is what rejects this.)
         assert_matches!(
             validated_step(
-                orchard_and_ironwood_notes(1, 0),
+                orchard_and_ironwood_notes((1, 10_000), (0, 0)),
                 TransactionBalance::new(
                     vec![shielded_change(ShieldedPool::Orchard, 8_000)],
                     Zatoshis::const_from_u64(4_000),
@@ -1150,7 +1154,7 @@ mod tests {
         // constraint.
         assert_matches!(
             validated_step(
-                orchard_and_ironwood_notes(1, 0),
+                orchard_and_ironwood_notes((1, 10_000), (0, 0)),
                 TransactionBalance::new(
                     vec![shielded_change(ShieldedPool::Orchard, 6_000)],
                     Zatoshis::const_from_u64(4_000),
@@ -1168,7 +1172,7 @@ mod tests {
         // than the step's Orchard inputs remove: 6_000 change < 10_000 input.
         assert_matches!(
             validated_step(
-                orchard_and_ironwood_notes(1, 0),
+                orchard_and_ironwood_notes((1, 10_000), (0, 0)),
                 TransactionBalance::new(
                     vec![shielded_change(ShieldedPool::Orchard, 6_000)],
                     Zatoshis::const_from_u64(4_000),
@@ -1183,7 +1187,7 @@ mod tests {
         // pool balance unchanged, which the turnstile forbids.
         assert_matches!(
             validated_step(
-                orchard_and_ironwood_notes(1, 1),
+                orchard_and_ironwood_notes((1, 10_000), (1, 20_000)),
                 TransactionBalance::new(
                     vec![
                         shielded_change(ShieldedPool::Orchard, 10_000),
@@ -1205,7 +1209,7 @@ mod tests {
         // change at all.
         assert_matches!(
             validated_step(
-                orchard_and_ironwood_notes(0, 1),
+                orchard_and_ironwood_notes((0, 0), (1, 20_000)),
                 TransactionBalance::new(
                     vec![shielded_change(ShieldedPool::Orchard, 16_000)],
                     Zatoshis::const_from_u64(4_000),
@@ -1224,7 +1228,7 @@ mod tests {
         // valid.
         assert_matches!(
             validated_step(
-                orchard_and_ironwood_notes(0, 1),
+                orchard_and_ironwood_notes((0, 0), (1, 20_000)),
                 TransactionBalance::new(
                     vec![shielded_change(ShieldedPool::Orchard, 16_000)],
                     Zatoshis::const_from_u64(4_000),
@@ -1268,7 +1272,7 @@ mod tests {
             request,
             BTreeMap::from([(0usize, pool)]),
             vec![],
-            shielded_inputs_for(orchard_and_ironwood_notes(1, 0)),
+            shielded_inputs_for(orchard_and_ironwood_notes((1, 10_000), (0, 0))),
             Some(BlockHeight::from_u32(100)),
             vec![],
             TransactionBalance::new(vec![], Zatoshis::const_from_u64(4_000)).unwrap(),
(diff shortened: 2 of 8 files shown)


zcash_client_backend/src/proposal.rs
```diff
@@ -958,7 +1026,7 @@ mod tests {
             shielded_inputs,
             anchor_height: Some(BlockHeight::from_u32(100)),
             prior_step_inputs: vec![],
-            balance: TransactionBalance::new(vec![], Zatoshis::ZERO).unwrap(),
+            balance: TransactionBalance::new(change, Zatoshis::ZERO).unwrap(),
```
dannywillems, line 1033: Note: this is in tests.

zcash_client_backend/src/proposal.rs
```diff
@@ -1227,25 +1299,449 @@ mod tests {
...
 
+    // Note and change values, spanning a zero-valued note through a large one. An action count is a
+    // function of how many notes and outputs a step has in each pool, never of what they are worth,
+    // so every value here must give the same answer. The upper bound keeps the total value of even
+    // the largest generated step well inside `MAX_MONEY`, so that `TransactionBalance` construction
+    // cannot fail for a reason these tests are not about.
+    fn arb_note_value() -> impl Strategy<Value = u64> {
+        prop_oneof![
+            1 => Just(0u64),
+            8 => 1u64..1_000_000_000,
```
dannywillems, line 1338: This is in zatoshis, so the limits is actually 21*10^6*10^8. LGTM.

components/zcash_protocol/src/constants.rs
```diff
@@ -40,3 +40,12 @@ pub const V5_VERSION_GROUP_ID: u32 = 0x26A7270A;
...
 pub const V6_VERSION_GROUP_ID: u32 = 0xD884B698;
+
+/// The maximum size in bytes of a Zcash block, and therefore the maximum size of any single
+/// transaction within one.
+///
+/// It is specified as `MAX_BLOCK_SIZE` in
+/// [§ 7.6 Block Header Encoding and Consensus](https://zips.z.cash/protocol/protocol.pdf#blockheader).
+///
+/// This constant is called `MAX_BLOCK_SIZE` in the zcashd source.
+pub const MAX_BLOCK_BYTES: usize = 2_000_000;
```
dannywillems, line 51: I start instructing Claude to move magic numbers out of the codebase/tests environments, and expose them in crates everyone can use.

zcash_client_backend/src/proposal.rs
```diff
@@ -1227,25 +1299,449 @@ mod tests {
...
+        ]
+    }
+
+    // An upper bound on the actions a single bundle can contain on-chain: a transaction may not
+    // exceed a block, and each action costs at least its action description. The true ceiling is
+    // lower, since `ACTION_SIZE` excludes each action's spend authorization signature and the
+    // bundle's proof, but an over-estimate is what these tests want: it exercises counts at least
+    // as large as anything buildable.
+    const MAX_ACTIONS_PER_BUNDLE: usize = MAX_BLOCK_BYTES / ACTION_SIZE;
```
dannywillems, line 1347: LGTM

zcash_client_backend/src/proposal.rs
```diff
@@ -1227,25 +1299,449 @@ mod tests {
...
+            NetworkUpgrade::Nu6,
+            NetworkUpgrade::Nu6_1,
+            NetworkUpgrade::Nu6_2,
+        ])
+    }
+
+    // The network upgrades from which the Orchard pool disables cross-address transfers. `Nu7` is
+    // excluded: it has no activation height on any network yet, so no bundle can be built for it.
+    fn arb_nu6_3_or_later_upgrade() -> impl Strategy<Value = NetworkUpgrade> {
+        prop::sample::select(vec![NetworkUpgrade::Nu6_3])
```
dannywillems, line 1376: Can be extended later.

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -26,6 +28,29 @@ pub const FLAG_SPENDS_ENABLED: u8 = 0b0000_0001;
...
+/// Every such value in an Orchard action — the value commitment, nullifier, randomized
+/// verification key, note commitment, and ephemeral key — is encoded in this many bytes.
+/// [`EphemeralKeyBytes`] is the encoding of one of them (the ephemeral key, a group element),
+/// so its width is that of all of them; note that this is a property of the *encoding*, not of
+/// the in-memory representation, which differs between these types.
+const PALLAS_ENCODING_SIZE: usize = size_of::<EphemeralKeyBytes>();
+
+/// The number of Pallas encodings in an Orchard action description: the value commitment,
+/// nullifier, randomized verification key, note commitment, and ephemeral key.
+const PALLAS_ENCODINGS_PER_ACTION: usize = 5;
```
dannywillems: I would prefer to have something more precise. Like s`ize_of::NoteCommitmentBytes`, etc.

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -26,6 +28,43 @@ pub const FLAG_SPENDS_ENABLED: u8 = 0b0000_0001;
 pub const FLAG_OUTPUTS_ENABLED: u8 = 0b0000_0010;
 pub const FLAGS_EXPECTED_UNSET: u8 = !(FLAG_SPENDS_ENABLED | FLAG_OUTPUTS_ENABLED);
 
+// The encoded size of each element of an Orchard action. Each is a Pallas group element or base
+// field element, both of which encode to 32 bytes; note that this is a property of the *encoding*,
+// and unrelated to the in-memory representation, which differs between these types.
+//
+// These belong upstream in `orchard`, beside the types themselves: it defines each of these
```
dannywillems, line 35: Should be in a follow-up issue. Not part of this patch.

zcash_primitives/src/transaction/components/orchard.rs
```diff
@@ -7,6 +7,7 @@ use corez::io::{self, Read, Write};
 
 use nonempty::NonEmpty;
 
+use core::mem::size_of;
```
schell, line 10: nit: This should go up with the other "core" imports, but the imports organisation here seem a bit wacky anyway.

schell (approved):
utACK [8e37bc5](https://github.com/zcash/librustzcash/pull/2642/commits/8e37bc5310280ce985adc9e1269bf1c1cfd72443)

zcash_primitives/src/transaction/components/orchard.rs
dannywillems: Re the imports: we can use cargo nightly with the option `imports_granularity=Crate` to unify, like Mina does for proof-systems: https://github.com/o1-labs/proof-systems/blob/master/.rustfmt.toml.
I haven't suggested it yet because of the deadlines, and the huge diff that it might create. I keep that in mind for the future.

PR #2643, opened 2026-07-16
Add PCZT signer redaction helper

czarcas7ic:
Adds a wallet helper that creates a smaller PCZT for an external signer by removing wallet-only data and fields that can be regenerated. The wallet keeps the original PCZT and combines the signer’s contribution back into it, while the signer view retains the information needed to verify what is being signed.

### DIFF
--- a/pczt/src/orchard.rs
+++ b/pczt/src/orchard.rs
@@ -443,6 +443,41 @@ impl Action {
             self.output.enc_ciphertext = EncCiphertext::MemoPlaintext(memo);
         }
     }
+
+    pub(crate) fn redact_recomputable_fields(&mut self, note_version: NoteVersion) {
+        let original_enc_ciphertext = self.output.enc_ciphertext.clone();
+        self.replace_enc_ciphertext_with_decrypted_memo_plaintext(note_version);
+        if self.output.enc_ciphertext != original_enc_ciphertext {
+            let mut resolved = self.output.clone();
+            if resolved
+                .encrypt_ciphertext_from_memo(note_version, self.spend.nullifier)
+                .is_err()
+                || resolved.enc_ciphertext != original_enc_ciphertext
+            {
+                self.output.enc_ciphertext = original_enc_ciphertext;
+            }
+        }
+
+        if let Some(original_cv_net) = self.cv_net {
+            let mut resolved = self.clone();
+            resolved.cv_net = None;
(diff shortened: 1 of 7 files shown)


pczt/src/orchard.rs
```diff
@@ -443,6 +443,41 @@ impl Action {
             self.output.enc_ciphertext = EncCiphertext::MemoPlaintext(memo);
         }
     }
+
+    pub(crate) fn redact_recomputable_fields(&mut self, note_version: NoteVersion) {
+        let original_enc_ciphertext = self.output.enc_ciphertext.clone();
+        self.replace_enc_ciphertext_with_decrypted_memo_plaintext(note_version);
```
nullcopy, line 449: This isn't redacting, it's decrypting. I understand the rationale, but we should name the function differently. I expect redaction to remove information from view, but this actually reveals more information, just in a more compact representation.

Perhaps rename to `compact_resolvable_fields()`? This would pair well with `resolve_fields()`.

pczt/src/orchard.rs
```diff
@@ -1411,6 +1446,61 @@ pub(crate) mod v2 {
             );
         }
 
+        #[cfg(feature = "orchard")]
+        #[test]
+        fn recomputable_field_redaction_checks_derived_values() {
+            let mut action = decryptable_action_with_memo([0; MEMO_SIZE]);
```
nullcopy, line 1452: Would prefer to have actual data in the memo

pczt/src/orchard.rs
```diff
@@ -1411,6 +1446,61 @@ pub(crate) mod v2 {
...
+            assert!(matches!(
+                action.output.enc_ciphertext,
+                EncCiphertext::MemoPlaintext(_)
+            ));
+        }
+
+        #[cfg(feature = "orchard")]
+        #[test]
+        fn recomputable_field_redaction_retains_unverifiable_values() {
+            let mut action = decryptable_action_with_memo([0; MEMO_SIZE]);
```
nullcopy, line 1470: Would prefer to have actual data in the memo

nullcopy (approved):
I have a few minor suggestions, but no objections. I would push harder on the name, but since it's used exclusively in the context of a redactor, I'll concede that it's at least consistent.

utACK 9281db6145eaff7aa0ed098434de8e401d974559

PR #2645, opened 2026-07-16
zcash_ironwood_migration_backend: add crate with note-split planning

dannywillems:
First, independent patch for the Orchard -> Ironwood migration. This focuses only on the note splitting; the PCZT, signing, errors, types, etc. come after the note-split strategy, which is the part that drives the most discussion. The planner is customizable over time via a trait, and the rest of the library is built on top of it.

----

## Problem

When a wallet migrates from the Orchard value pool to Ironwood, every note it spends makes a value **cross the turnstile in a transaction a chain observer can see**. The rule that chooses those crossing amounts is privacy-critical: a deterministic split derived from the wallet's balance turns the sequence of amounts into a fingerprint that lets an observer recognise a specific balance, link all the crossings of one migration together, and single out large holders ("whales").

## Decision: ZIP 318 canonical quantization

The near-final [ZIP 318](https://zips.z.cash/zip-0318) settles the denomination scheme on **canonical `{1, 2, 5} * 10^k` quantization**: every crossing amount is drawn from a small, shared denomination set so that many wallets emit identical values that collide and cannot be attributed. Privacy rests on **value collision (k-anonymity)**, explicitly *not* on unpredictability: a random or high-entropy amount would collide with no other wallet and become a near-unique fingerprint, which is why the ZIP rejects random or arbitrary sizing.

This PR is the first reviewable slice of the larger migration effort (#2625): a new **backend-agnostic** crate `zcash_ironwood_migration_backend` containing just its **note-split planner**, so the core logic can be reviewed in isolation.

> An earlier revision of this PR shipped two strategies (a randomized `{1,2,5}` sampler and a power-of-ten canonical) while the scheme was still undecided. It has been reduced to the single strategy that matches the near-final ZIP: the randomized sampler used the per-wallet unpredictability model the ZIP rejects, and the power-of-ten form predates the ZIP's move to the `{1, 2, 5}` series.

## Solution

The composition rule is abstracted behind the object-safe `DenominationStrategy` trait, with **one implementation**:

- **`CanonicalOneTwoFive`** — ZIP 318's canonical quantization. A deterministic, descending greedy decomposition over the `{1, 2, 5} * 10^k` ZEC series, which is exactly the ZIP's greedy decimal-digit expansion (each decimal digit expands into `{5, 2, 1}` times its place value). It works in zatoshi, so it mints sub-1-ZEC denominations down to a dust floor. Verified against the ZIP's worked examples: `540 -> [500, 20, 20]`, `123.45 -> [100, 20, 2, 1, 0.2, 0.2, 0.05]`, `25000 -> [10000, 10000, 5000]`.

The trait is **retained as the seam** for a future variant, such as ZIP 318's *optional* frequency-constrained randomized substitution (which only varies which canonical denomination is chosen, never the values themselves).

It produces a `NoteSplitPlan` of **self-funding notes** (each holding its crossing value plus a fee buffer, so when spent it pays its own migration-transfer fee), bounds every note by a **maximum denomination** (ZIP 318's `DENOM_CAP`, 10,000 ZEC) so a whale's crossings stay within the shared denomination set (larger balances migrate over several runs), and leaves any residual below the dust floor (ZIP 318's `MAX_RESIDUAL_VALUE`, 0.01 ZEC) as source-pool change, never folded into a fee (folding an identifiable dust amount into a fee would deanonymise a dust-attacked wallet).

The other tunables are pluggable too: the per-note fee comes from a `FeePolicy` trait (`Zip317FeePolicy` provided), and the maximum denomination, dust floor, and note cap are strategy parameters. `plan_note_split` is a thin convenience wrapper over the recommended `CanonicalOneTwoFive`.

Note splitting is fundamentally a **constrained integer partition** problem: partitioning a known integer `N` (the migratable balance) into parts `n_1, ..., n_k` (the crossing values) with `N = (n_1 + ... + n_k) + k*f + r` (`f` = per-note fee buffer, `r` = residual change). The top-level module doc states this formally, enumerating the parameters and the five principles each part must obey (canonical, bounded magnitude, bounded count, feasible/self-funding, drained), then places it among neighbouring problems with references (change-making, bounded knapsack, subset-sum, Boltzmann samplers, preferred-number denomination design), since this crate is meant to be reusable.

## Structure

- `note_splitting.rs` — `NoteSplitPlan`, the `DenominationStrategy` and `FeePolicy` traits, `Zip317FeePolicy`, the public constants, and `plan_note_split`.
- `note_splitting/strategies.rs` — the `CanonicalOneTwoFive` strategy and its tests.
- `note_splitting/utils.rs` — the `largest_one_two_five` denomination helper.

## Notes for reviewers

- **`CanonicalOneTwoFive` is exactly ZIP 318's canonical quantization.** The descending greedy `{1, 2, 5} * 10^k` choice equals the ZIP's greedy decimal-digit expansion into `{5, 2, 1}` times each place value; a golden test pins all three ZIP worked examples.
- **The trait is kept deliberately**, even with one implementation, so ZIP 318's optional randomized substitution (or another variant) can be added later by selecting a strategy rather than rewriting code.
- `NoteSplitPlan` keeps private fields with accessors and derives no `serde`.
- Constants reuse protocol values: `COIN` / `MAX_MONEY` (`zcash_protocol`), `MARGINAL_FEE` (`zcash_primitives`); no magic numbers.
- **Pool-agnostic code.** The `note_splitting` code names no specific pool: no `orchard`/`ironwood` dependency, and identifiers use source/destination framing (`NoteSplitPlan::change`, `SOURCE_ACTIONS_PER_TRANSFER` / `DESTINATION_ACTIONS_PER_TRANSFER`). The doc uses Orchard -> Ironwood (NU6.3) as the concrete running example. This keeps the planner reusable for future pool migrations.
- Dependencies: `zcash_protocol`, `zcash_primitives`, `rand_core`; dev-only `proptest`, `rand_chacha`.
- The rest of the migration engine (types, error, scheduling, the wallet/store traits, the PCZT pipeline) lands in follow-up PRs on this crate.

## Tests

A proptest suite drives the trait contract over the whole input range (value conservation, self-funding note = crossing + buffer, note cap, migratable sum, valid + bounded + **non-increasing** `{1, 2, 5} * 10^k` denominations, small residual unless the cap is reached, and RNG-independence), plus golden vectors for ZIP 318's worked examples (`540`, `123.45`, `25000`), the full 1-2-5 series (`12,345 ZEC`), a sub-1-ZEC decomposition (`0.53 ZEC`), and whale capping (`45,000 ZEC`).

## Commits

1. the crate + note-split planner
2. the CHANGELOG
3. an AGENTS.md line requiring named constants over magic numbers
4. reduce the strategies to the single ZIP 318 canonical one
5. the matching CHANGELOG update

## Verification

- `cargo test -p zcash_ironwood_migration_backend`
- `cargo clippy -p zcash_ironwood_migration_backend --all-features --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p zcash_ironwood_migration_backend --no-deps`
- `cargo fmt --check`


zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -0,0 +1,284 @@
...
+//! Whatever the strategy, the result is a [`NoteSplitPlan`]: a multiset of *denomination notes*,
+//! each holding `denomination + fee buffer` so that when it is later spent in a migration transfer
+//! it pays its own fee (the [`FeePolicy`] sizes the buffer). Every note is bounded by a maximum
+//! denomination, so even a whale crosses many bounded, collision-prone amounts rather than one
+//! distinctive one, and a balance beyond a single run's capacity migrates over several runs.
+//! Whatever cannot form a whole self-funding note is left in Orchard as change, never folded into a
+//! transaction fee (folding an identifiable dust amount into a fee would deanonymise a dust-attacked
+//! wallet).
+//!
+//! # Relation to known problems
```
dannywillems, line 46: This is left in the comments in case we want to ask Claude/an agent to implement other solutions. This is for context purposes.

zcash_pool_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,566 @@
...
+        Self {
+            max_notes,
+            max_denomination_zec,
+            buffer_zatoshi: fee.transfer_fee_buffer_zatoshi(),
+        }
+    }
+
+    /// The recommended configuration: [`MIGRATION_MAX_PREPARED_NOTES_PER_RUN`] notes,
+    /// [`MIGRATION_MAX_DENOMINATION_ZEC`] cap, ZIP-317 fees.
+    pub fn recommended() -> Self {
```
dannywillems, line 63: Of course, this can be changed later.

zcash_ironwood_migration_backend/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "zcash_ironwood_migration_backend"
```
dannywillems: The crate can be renamed later. I would be in favor of something pool agnostic.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+            MIGRATION_MAX_PREPARED_NOTES_PER_RUN,
+            MIGRATION_MAX_DENOMINATION_ZEC,
+            &Zip317FeePolicy,
+        )
+    }
+
+    /// The smallest self-funding note: the minimum denomination plus its transfer buffer. A draw
+    /// leaving less than this cannot migrate any more value.
+    fn min_note_zatoshi(&self) -> u64 {
+        MIN_ONE_TWO_FIVE_DENOMINATION_ZEC * COIN + self.buffer_zatoshi
```
dannywillems: buffer_zatoshi would actually be the minimal fee.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+
+impl DenominationStrategy for RandomizedOneTwoFive {
+    fn plan(
+        &self,
+        total_input_zatoshi: u64,
+        prep_fee_zatoshi: u64,
+        rng: &mut dyn RngCore,
+    ) -> NoteSplitPlan {
+        if total_input_zatoshi <= prep_fee_zatoshi {
+            return NoteSplitPlan::empty(total_input_zatoshi, prep_fee_zatoshi, None);
```
dannywillems: I don't know if we want to have an error here.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+            prep_fee_zatoshi,
+            migration_outputs,
+            crossing_values,
+            budget,
+        )
+    }
+}
+
+/// The largest `{1, 2, 5} * 10^k` whole-ZEC value not exceeding `n`, for `n >= 1`.
+fn largest_denomination_zec(n: u64) -> u64 {
```
dannywillems: That could be moved into a utils.rs module later.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+    }
+    pow
+}
+
+/// Every `{1, 2, 5} * 10^k` whole-ZEC denomination `d` with `lo <= d <= hi`, in ascending order.
+/// Empty when `hi == 0`.
+fn denominations_between(lo: u64, hi: u64) -> Vec<u64> {
+    let mut out = Vec::new();
+    let mut pow = 1u64;
+    'outer: loop {
```
dannywillems: Fun Rust feature.
`goto` is back!

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+            Some(p) => pow = p,
+            None => break,
+        }
+    }
+    out
+}
+
+/// The largest power of the radix `p` (a multiple of the power-of-radix `floor`) with
+/// `floor <= p <= hi`, or `0` if `hi < floor`.
+fn largest_power_of_ten(hi: u64, floor: u64) -> u64 {
```
dannywillems: Could be moved in an utils.rs module later.

zcash_pool_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+        (
+            0u64..=MAX_MONEY,
+            0u64..MAX_SAMPLED_PREP_FEE_ZATOSHI,
+            any::<u64>(),
+            1usize..=MAX_SAMPLED_NOTE_CAP,
+        )
+    }
+
+    /// Whether `zat` is a `{1, 2, 5} * 10^k` whole-ZEC amount (in zatoshi).
+    fn is_one_two_five_zat(zat: u64) -> bool {
```
dannywillems, line 138: This could be moved into a utils.rs module.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+        }
+        let mut n = zat / COIN;
+        while n.is_multiple_of(10) {
+            n /= 10;
+        }
+        matches!(n, 1 | 2 | 5)
+    }
+
+    /// Whether `zat` is a power of ten (in zatoshi).
+    fn is_power_of_ten_zat(zat: u64) -> bool {
```
dannywillems: This could be moved into a utils.rs module.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+/// discarded.
+pub struct RandomizedOneTwoFive {
+    max_notes: usize,
+    max_denomination_zec: u64,
+    buffer_zatoshi: u64,
+}
+
+impl RandomizedOneTwoFive {
+    /// A strategy with an explicit note cap, maximum denomination (in whole ZEC), and fee model.
+    pub fn new(max_notes: usize, max_denomination_zec: u64, fee: &dyn FeePolicy) -> Self {
```
nuttycom: It's fine to use `&dyn` here if it's necessary, but I think everywhere else in the codebase we use static polymorphism?

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+    fn min_note_zatoshi(&self) -> u64 {
+        MIN_ONE_TWO_FIVE_DENOMINATION_ZEC * COIN + self.buffer_zatoshi
+    }
+
+    /// One random decomposition draw. Assumes `total_input_zatoshi > prep_fee_zatoshi`.
+    fn draw(
+        &self,
+        total_input_zatoshi: u64,
+        prep_fee_zatoshi: u64,
+        rng: &mut dyn RngCore,
```
nuttycom: Same question here about `dyn` vs static polymorphism.

zcash_ironwood_migration_backend/src/note_splitting.rs
```diff
@@ -0,0 +1,318 @@
...
+/// The default cap on how many notes one migration run prepares. Bounding the note count keeps the
+/// decomposition a bounded problem and bounds each run's transaction and proving cost; a larger
+/// balance migrates over several runs.
+pub const MIGRATION_MAX_PREPARED_NOTES_PER_RUN: usize = 50;
+
+/// The default largest denomination (in whole ZEC) [`RandomizedOneTwoFive`] gives a single note:
+/// `2 * 10^4 = 20_000` ZEC, itself a `{1, 2, 5} * 10^k` value. Capping the top denomination keeps
+/// even a whale's crossings within the shared denomination set, so no single crossing is a
+/// near-unique fingerprint.
+pub const MIGRATION_MAX_DENOMINATION_ZEC: u64 = 20_000;
```
dannywillems: Note: 10_000 is the maximum; decided in a the wallet call now.

zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -0,0 +1,318 @@
...
+use zcash_protocol::value::COIN;
+
+pub mod strategies;
+
+pub use strategies::{CanonicalPowerOfTen, RandomizedOneTwoFive};
+
+/// The default cap on how many notes one migration run prepares. Bounding the note count keeps the
+/// decomposition a bounded problem and bounds each run's transaction and proving cost; a larger
+/// balance migrates over several runs.
+pub const MIGRATION_MAX_PREPARED_NOTES_PER_RUN: usize = 50;
```
dannywillems, line 133: Note: this should be covering outputs + spends.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+            canonical_crossings(12_345 * (COIN / 100)),
+            vec![
+                hundred, ten, ten, one, one, one, tenth, tenth, tenth, tenth, hundredth, hundredth,
+                hundredth, hundredth, hundredth
+            ]
+        );
+    }
+
+    /// Canonical: a balance below the dust floor migrates nothing and keeps it all as change.
+    #[test]
```
dannywillems: LGTM

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+            vec![
+                tenth, tenth, tenth, tenth, tenth, hundredth, hundredth, hundredth
+            ]
+        );
+    }
+
+    /// Canonical reproduces the ZIP draft's full worked example: 123.45 ZEC, including its 0.1 and
+    /// 0.01 parts.
+    #[test]
+    fn canonical_expands_123_point_45_zec() {
```
dannywillems: I feel it should be 10,000 + 2,000 + 200 + 100 + 2 * 10 + 2 * 10 + 5 * 1.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+        let mut rng = ChaCha8Rng::seed_from_u64(0);
+        canonical_no_fee()
+            .plan(total, 0, &mut rng)
+            .crossing_values()
+            .to_vec()
+    }
+
+    /// Canonical: whole-ZEC digit expansion, capped at DENOM_CAP. 250 ZEC = two 100s and five 10s.
+    #[test]
+    fn canonical_expands_250_zec() {
```
dannywillems: I think it should be 2 * 100 + 5 * 10, i.e. only two notes.

dannywillems: Adding a new strategy for this. What I mean is we want 200 + 50.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
dannywillems: Adding a new strategy for this.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+            vec![
+                hundred, ten, ten, one, one, one, tenth, tenth, tenth, tenth, hundredth, hundredth,
+                hundredth, hundredth, hundredth
+            ]
+        );
+    }
+
+    /// Canonical: a balance below the dust floor migrates nothing and keeps it all as change.
+    #[test]
+    fn canonical_below_dust_floor_migrates_nothing() {
```
dannywillems: This could be parametrized by the strategy, as it should always be true for all strategies. We should not have duplicated code.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,695 @@
...
+                return v;
+            }
+        }
+    }
+    pow
+}
+
+/// Every `{1, 2, 5} * 10^k` whole-ZEC denomination `d` with `lo <= d <= hi`, in ascending order.
+/// Empty when `hi == 0`.
+fn denominations_between(lo: u64, hi: u64) -> Vec<u64> {
```
dannywillems: Could be moved into a utils.rs or something like this.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,756 @@
...
+/// A randomized decomposition whose crossing values follow the `{1, 2, 5} * 10^k` ZEC series. The
+/// same balance yields different crossing amounts on different runs, varying both which denominations
+/// appear and their order of magnitude, so 723 ZEC is as likely to be split into hundreds and tens as
+/// into one large note plus smaller ones. Privacy rests on per-wallet unpredictability while every
+/// individual value stays canonical (drawn from the shared 1-2-5 set) and hence collision-prone.
+///
+/// Each plan keeps the best of a few independent draws (`RANDOMIZED_PLAN_ATTEMPTS`, the one that
+/// migrates the most), so an unlucky draw that exhausts the note cap while leaving a remainder is
+/// discarded.
+pub struct RandomizedOneTwoFive {
```
dannywillems: Not possible to mind sub-1 ZEC. Addressing it now.

dannywillems: Fixed.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,795 @@
...
+        // Smallest self-funding note: the dust floor plus its transfer buffer.
+        let min_note = self.dust_floor_zatoshi + buffer;
+        let mut budget = total_input_zatoshi - prep_fee_zatoshi;
+
+        let mut migration_outputs = Vec::new();
+        let mut crossing_values = Vec::new();
+        while budget >= min_note && migration_outputs.len() < self.max_notes {
+            // Largest `{1, 2, 5} * 10^k` denomination whose note fits the budget, capped.
+            let affordable = (budget - buffer).min(self.max_denomination_zatoshi);
+            let crossing = largest_one_two_five(affordable, self.dust_floor_zatoshi);
```
dannywillems, line 93: I think this can be simplified, as I think L212-214 can be replaced with `affordable < self.dust_floor_zatoshi`.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,795 @@
...
+}
+
+impl DenominationStrategy for CanonicalPowerOfTen {
+    fn plan(
+        &self,
+        total_input_zatoshi: u64,
+        prep_fee_zatoshi: u64,
+        _rng: &mut dyn RngCore,
+    ) -> NoteSplitPlan {
+        if total_input_zatoshi <= prep_fee_zatoshi {
```
dannywillems: Maybe this could be shared with `CanonicalOneTwoFive`. Only the `largest_power_of_ten` is different.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,800 @@
...
+            denom_cap_zatoshi: denom_cap_zec * COIN,
+            dust_floor_zatoshi,
+            buffer_zatoshi: fee.transfer_fee_buffer_zatoshi(),
+        }
+    }
+
+    /// The Ironwood migration ZIP draft configuration: [`ZIP_DENOM_CAP_ZEC`] cap,
+    /// [`RESIDUAL_MIGRATION_MIN_ZATOSHI`] dust floor, ZIP-317 fees, and a per-run note cap of
+    /// [`MIGRATION_MAX_PREPARED_NOTES_PER_RUN`].
+    pub fn zip_draft() -> Self {
```
dannywillems: This should be changed as the ZIP changes.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,800 @@
...
+            );
+            let mut a = ChaCha8Rng::seed_from_u64(seed);
+            let mut b = ChaCha8Rng::seed_from_u64(seed ^ 0xdead_beef);
+            prop_assert_eq!(s.plan(total, fee, &mut a), s.plan(total, fee, &mut b));
+        }
+
+        /// The canonical strategy matches an independent reference greedy power-of-ten expansion when
+        /// there is no fee buffer to consume budget.
+        #[test]
+        fn canonical_matches_reference_expansion(
```
dannywillems: This is trivial.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,800 @@
...
+/// zatoshi, so it mints sub-1-ZEC denominations down to a dust floor). The same balance yields
+/// different crossing amounts on different runs, varying both which denominations appear and their
+/// order of magnitude, so 723 ZEC is as likely to be split into hundreds and tens as into one large
+/// note plus smaller ones. Privacy rests on per-wallet unpredictability while every individual value
+/// stays canonical (drawn from the shared 1-2-5 set) and hence collision-prone.
+///
+/// Each plan keeps the best of a few independent draws (`RANDOMIZED_PLAN_ATTEMPTS`, the one that
+/// migrates the most), so an unlucky draw that exhausts the note cap while leaving a remainder is
+/// discarded.
+pub struct RandomizedOneTwoFive {
```
dannywillems: Note: the seed has less and less impact when the amount reaches the maximum, as the set of possible draws decreases.
The indistinguishability between the randomized and the canonical might not be very large. An easy way to get an actual number of the randomness impact is to compute the number of possible draws based on the decomposition.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
schell: This seems sane.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,777 @@
...
+            // since `budget >= min_note` guarantees `budget - buffer >= floor`.
+            let affordable = (budget - buffer).min(self.max_denomination_zatoshi);
+            // Floor the pick so `slots_left` notes could still drain the budget (using the real,
+            // uncapped budget so a whale is pushed toward the cap), but never above the affordable
+            // ceiling. As the budget shrinks the floor drops, widening the choice; in the last slot
+            // it collapses to the largest affordable denomination.
+            let floor_target = ((budget - buffer) / slots_left).max(floor).min(affordable);
+            let floor_denom = largest_one_two_five(floor_target, floor);
+            let candidates = denominations_between(floor_denom, affordable);
+            let crossing = candidates[(rng.next_u64() % candidates.len() as u64) as usize];
```
schell: I didn't catch this but `claude` says:
> Modulo bias in the privacy-critical sampler. strategies.rs:147 uses rng.next_u64() % candidates.len(). candidates.len() is tiny (≤ ~30 denominations), so the bias is negligible in practice, but for a privacy-critical random sampler it's worth either documenting explicitly or using a rejection-sampling / uniform helper. Cheap to make rigorous.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
```diff
@@ -0,0 +1,777 @@
...
+    /// zatoshi, which MUST be a power of ten), and fee model.
+    pub fn new(
+        max_notes: usize,
+        max_denomination_zec: u64,
+        dust_floor_zatoshi: u64,
+        fee: &dyn FeePolicy,
+    ) -> Self {
+        Self {
+            max_notes,
+            max_denomination_zatoshi: max_denomination_zec * COIN,
```
schell: This could be a checked multiply, though it would change the return type.

schell (changes requested):
This is great! 👍 I just had some little comments.

Also - what about the `std`ness of this crate? There's no `#![no_std]` / `extern crate alloc`. The library only needs `alloc (Vec)` and only the tests use std for `HashSet`.

Last nit - in places the comments say we have two implementations of `DenominationStrategy` but there are actually three here.

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
dannywillems: I removed the randomized version, as it is not recommended in the ZIP. This comment is therefore outdated. Resolving the conversation.

dannywillems:
> This is great! 👍 I just had some little comments.
> 
> Also - what about the `std`ness of this crate? There's no `#![no_std]` / `extern crate alloc`. The library only needs `alloc (Vec)` and only the tests use std for `HashSet`.
> 
> Last nit - in places the comments say we have two implementations of `DenominationStrategy` but there are actually three here.

no_std implemented too. Good suggestion!

dannywillems:
> Last nit - in places the comments say we have two implementations of `DenominationStrategy` but there are actually three here.

This has been addressed too. Only one strategy is implemented now, the one following the ZIP318.

zcash_ironwood_migration_backend/Cargo.toml
dannywillems: Done: renamed to `zcash_pool_migration_backend` in 06086eeb2c (description, README, and crate docs generalised; first use remains Orchard -> Ironwood).

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
dannywillems: Addressed.

dannywillems:
Self-ACK, modulo comments.

zcash_pool_migration_backend/CHANGELOG.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Added
+- Initial scaffolding of the backend-agnostic Orchard -> Ironwood value-pool migration engine crate.
+- Note-split planning (the `note_splitting` module): decomposing a spendable source-pool balance into
+  the self-funding notes that cross the turnstile. The module is pool-agnostic (its code names no
+  specific pool); Zcash's first use is the Orchard -> Ironwood migration. The composition rule is
+  abstracted behind the `DenominationStrategy` trait, with one implementation aligned with ZIP 318:
+  - `CanonicalOneTwoFive`: the canonical `{1, 2, 5} * 10^k` quantization of ZIP 318, a deterministic
+    descending greedy decomposition over that series (equivalently, decimal-digit expansion into
+    `{5, 2, 1}` times each place value), e.g. 12,345 ZEC -> 10,000 + 2,000 + 200 + 100 + 20 + 20 + 5.
```
schell, line 18: Maybe a little note about how this decomposition assumes fees are zero?

zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -0,0 +1,339 @@
...
+//! cannot be attributed. Privacy rests on value collision (k-anonymity), explicitly not on
+//! unpredictability: a random or high-entropy amount would collide with no other wallet and become a
+//! near-unique fingerprint, which is why the ZIP rejects random or arbitrary sizing.
+//!
+//! The composition rule is abstracted behind the [`DenominationStrategy`] trait, with one
+//! implementation (in [`strategies`]):
+//!
+//! - [`CanonicalOneTwoFive`]: the ZIP 318 canonical quantization, a deterministic descending greedy
+//!   decomposition over the `{1, 2, 5} * 10^k` ZEC series (equivalently, decimal-digit expansion into
+//!   `{5, 2, 1}` times each place value), e.g. 12,345 ZEC -> 10,000 + 2,000 + 200 + 100 + 20 + 20 + 5.
```
schell, line 26: There's a caveat here that this decomposition can't happen if fees are > 0.

schell (approved):
Both points from my last review are resolved. `no_std` with `extern crate alloc` and `std `gated to `#[cfg(test)]`, and the strategy count is down to one. I like the rename to  `zcash_pool_migration_backend`.

My only last nit (non-blocking) is that in practice only the zero-fee planner would ever produce the canonical 5-2-1 decompositions mentioned in the module docs and changelog. No biggy though, I think everyone understands that.

zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -0,0 +1,339 @@
...
+/// See the module docs for the implementation and its denomination set. Object-safe, so a wallet can
+/// hold a selected `Box<dyn DenominationStrategy>` and a future variant can be added behind it.
+pub trait DenominationStrategy {
+    /// Decompose `total_input_zatoshi`, after reserving `prep_fee_zatoshi` for the note-split
+    /// transaction, into self-funding notes. `rng` is used by randomized strategies and ignored by
+    /// deterministic ones.
+    fn plan(
+        &self,
+        total_input_zatoshi: u64,
+        prep_fee_zatoshi: u64,
```
nuttycom, line 299: Should this take a `FeeRule` instead of a constant fee? Or is the idea that we will reserve a fee sufficient for a transaction padded to the maximum split actions? The latter is fine for the migration, since we'll want to pad anyway.

zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -0,0 +1,339 @@
...
+/// migration_outputs[i] == crossing_values[i] + buffer
+/// ```
+///
+/// They describe that one note at the two phases of the migration:
+/// - `crossing_values[i]` is the denomination that CROSSES the turnstile into the destination pool
+///   when the note is spent (the privacy-relevant value an observer sees; their sum is
+///   [`total_migratable_zatoshi`](Self::total_migratable_zatoshi)).
+/// - `migration_outputs[i]` is the note CREATED in the source pool during the prep phase: the
+///   crossing value plus the buffer, so the note self-funds its own migration transfer (the buffer
+///   pays that transfer's fee, and the crossing value is what remains to cross).
```
nuttycom, line 201: Since the per-note crossing fee is a constant, why do we need parallel vectors? Why not just a single vector of crossing values and then one can map over it to add the fee to get the actual migration outputs?

nuttycom (commented):
pots-hoc utACK

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
dannywillems: This is addressed in https://github.com/zcash/librustzcash/pull/2675

zcash_ironwood_migration_backend/src/note_splitting/strategies.rs
dannywillems: Addressed in https://github.com/zcash/librustzcash/pull/2675

zcash_pool_migration_backend/src/note_splitting.rs
dannywillems: Te latter.

zcash_pool_migration_backend/src/note_splitting.rs
dannywillems: Addressed in https://github.com/zcash/librustzcash/pull/2675

PR #2647, opened 2026-07-16
zcash_pool_migration_backend: note-split preparation planning and PCZT construction

dannywillems:
Stacked on #2645. This is the migration's transaction-construction layer, consolidated into one PR: it plans the note-preparation transactions, builds them (and the value-crossing transfer) as unproven PCZTs, and pre-signs them.

This PR supersedes the earlier three-way split (this branch plus #2656 and #2658, now folded in and closed). The obsolete single-transaction note-split builder `build_split_pczt` has been dropped in favour of the ZIP-318-compliant 16-action preparation pipeline below.

### 1. Note-split preparation planning (the `preparation` module, pure, `no_std`)

`plan_preparation` partitions the work of minting a note-split plan's self-funding notes into note-preparation transactions of exactly `PREP_TX_ACTIONS` (16) Orchard actions (ZIP 318), organised into the fewest sequential layers with parallel transactions inside a layer. It returns a `PreparationPlan` of `PrepTransaction`s (each a same-pool send-to-self referencing wallet notes or earlier layers' outputs). A largest-first greedy feeds each output transaction from the largest note, routes every leftover forward as a feeder, and consolidates dust; once the funding notes are scheduled it consolidates the leftover feeders toward a single residual note (ZIP 318's "one note per part plus at most one residual note").

The single residual is only reachable above the fee threshold: because a note is spent atomically, several spends can each strand a sub-fee remainder whose total is below one transaction fee, and those cannot be merged (the consolidation output would be negative), so the planner guarantees at most one residual worth a fee. Property tests assert this invariant alongside value conservation, the 16-action budget, "each funding note minted once", and the ZIP-318 lower bound on the transaction count. Pure (no cryptography or I/O).

### 2. The PCZT builders and pre-signing (the `build` module, behind the `orchard` feature)

- `build_prep_tx` builds one `PrepTransaction` into a same-pool Orchard send-to-self; every output is a wallet-controlled internal change note, and the bundle is built with `pad_to_minimum` set to 16, so orchard fills it to exactly 16 actions with fabricated dummies (no preparation transaction is distinguishable by its action count).
- `build_transfer_pczt` builds a value-crossing transfer, spending one self-funding note and outputting its crossing value to the account's own internal Ironwood change address (a ZIP 318 MUST).
- `sign_pczt` adds the Orchard spend-authorization signatures up front, signing a finalized but still unproven PCZT (the sighash is fixed independent of the zk proofs), so the migration signs early and proves later at scheduling time.

The shared transaction-builder plumbing (build config, PCZT finalization, action-index mapping) lives in the module root. Each builder is pure (no database or wallet-backend access) and takes the RNG as a parameter. Adds optional `orchard` and `pczt` dependencies, enabled only by the feature.

## Notes for reviewers

- The crate is `no_std`; the `build` module is feature-gated on `orchard`.
- Residual semantics: ZIP 318's "at most one residual note" is only achievable above the fee threshold. When several spends each strand a sub-fee remainder whose total is below one transaction fee, those cannot be consolidated (the output would be negative) and remain as multiple sub-fee dust notes; the planner guarantees at most one residual worth a fee. Worth confirming with the ZIP authors whether "at most one residual" is meant strictly or as "at most one residual worth a fee".
- Padding to exactly 16 actions uses orchard 0.15.0's `BundleType::Transactional { pad_to_minimum }` (already resolved by the workspace), so no dependency change.
- Known test gap (tracked): a multi-spend consolidation-shaped `build_prep_tx` needs a multi-note shared-anchor witness helper; the single-spend paths are covered.
- The transfer's fee/action-count (currently 4 actions = 20,000 zatoshi; ZIP 318 allows leaving the Ironwood bundle unpadded, i.e. 3 actions = 15,000) is a follow-up reconciliation item.

## Tests

Proptest suites for the preparation planner (value conservation, <= 16 actions per transaction, each funding note minted once, at most one residual, acyclic layers) and each builder (exactly-16-action bundles, balanced transfers, pre-signing authorizes only the account's own spend), plus golden cases.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc -p zcash_pool_migration_backend --no-deps`
- `cargo fmt --check`

### DIFF
--- a/zcash_pool_migration_backend/src/build.rs
+++ b/zcash_pool_migration_backend/src/build.rs
@@ -1,0 +1,278 @@
+//! Constructing the migration transactions as unproven PCZTs, and pre-signing them.
+//!
+//! [`build_prep_tx`] builds a note-preparation transaction (restructuring the wallet's notes into the
+//! self-funding notes a migration run needs), [`build_transfer_pczt`] builds a value-crossing
+//! transfer, and [`sign_pczt`] adds the Orchard spend-authorization signatures up front. Each builder
+//! runs purely on the transaction [`Builder`](zcash_primitives::transaction::builder::Builder): no
+//! database or wallet-backend access. Selecting the notes to spend and resolving their witnesses and
+//! the anchor are the wallet backend's job, done separately; here they are inputs, and the returned
+//! PCZT is still to be proven and finalized.
+//!
+//! The shared transaction-builder plumbing (building the config, finishing the PCZT through the
+//! [`Creator`]/[`IoFinalizer`] roles, un-shuffling output positions) lives in this module root, so
+//! every builder reuses it.
+
+use alloc::string::String;
+
+use core::fmt;
+
+use pczt::roles::{creator::Creator, io_finalizer::IoFinalizer};
+use zcash_primitives::transaction::builder::{BuildConfig, PcztParts};
+use zcash_protocol::consensus::Parameters;
+
+mod prep;
+mod sign;
+mod transfer;
+pub use prep::build_prep_tx;
+pub use sign::sign_pczt;
+pub use transfer::build_transfer_pczt;
+
+#[cfg(test)]
+mod end_to_end;
+
+/// An error building a migration PCZT.
+#[derive(Debug)]
+pub enum BuildError {
+    /// The requested outputs cannot be balanced against the selected inputs (the real fee versus the
+    /// planned outputs).
+    Balance(String),
+    /// The transaction builder or the PCZT assembly pipeline failed.
+    Build(String),
+}
+
+impl fmt::Display for BuildError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            BuildError::Balance(m) => write!(f, "cannot balance the transaction: {m}"),
+            BuildError::Build(m) => write!(f, "pczt build failed: {m}"),
+        }
+    }
+}
+
+impl core::error::Error for BuildError {}
+
+/// The [`BuildConfig`] shared by the migration transactions: an Orchard-anchored bundle, with the
+/// destination-pool (Ironwood) bundle anchored only when that phase produces a crossing output
+/// (`None` for the same-pool note split).
+pub(crate) fn build_config(
+    orchard_anchor: orchard::Anchor,
+    ironwood_anchor: Option<orchard::Anchor>,
+) -> BuildConfig {
+    BuildConfig::Standard {
+        sapling_anchor: None,
+        orchard_anchor: Some(orchard_anchor),
+        ironwood_anchor,
+        orchard_bundle_type: orchard::builder::BundleType::DEFAULT,
+        ironwood_bundle_type: orchard::builder::BundleType::DEFAULT,
+    }
+}
+
+/// Finish an unproven PCZT from the transaction builder's parts: run the [`Creator`] and
+/// [`IoFinalizer`] roles.
+pub(crate) fn finalize_pczt<P: Parameters>(parts: PcztParts<P>) -> Result<pczt::Pczt, BuildError> {
+    let created = Creator::build_from_parts(parts)
+        .ok_or_else(|| BuildError::Build("pczt creation failed".into()))?;
+    IoFinalizer::new(created)
+        .finalize_io()
+        .map_err(|e| BuildError::Build(format!("io finalize: {e:?}")))
+}
+
+/// Map an output's request-order position to its real Orchard action index. The Orchard builder
+/// shuffles action positions, so the caller must look up where each requested output landed.
+pub(crate) fn output_action_index(
+    meta: &orchard::builder::BundleMetadata,
+    output: usize,
+) -> Result<u32, BuildError> {
+    meta.output_action_index(output)
+        .map(|i| i as u32)
+        .ok_or_else(|| BuildError::Build(format!("no action index for output {output}")))
+}
+
+/// Test helpers shared by the split and transfer builders: a deterministic account, regtest network
+/// params, and a single-leaf Orchard witness. Kept here so both submodules' tests reuse them.
+#[cfg(test)]
+pub(crate) mod test_util {
+    use alloc::vec::Vec;
+
+    use incrementalmerkletree::{Hashable, Level};
+    use orchard::keys::{FullViewingKey, Scope, SpendingKey};
+    use orchard::note::{ExtractedNoteCommitment, Note, NoteVersion, RandomSeed, Rho};
+    use orchard::tree::{MerkleHashOrchard, MerklePath};
+    use orchard::value::NoteValue;
+    use orchard::{Anchor, NOTE_COMMITMENT_TREE_DEPTH};
+    use rand_chacha::ChaCha8Rng;
+    use rand_core::{RngCore, SeedableRng};
+    use zcash_protocol::consensus::BlockHeight;
+    use zcash_protocol::local_consensus::LocalNetwork;
+
+    /// 32 random bytes from a `seed`-derived RNG, keeping calls deterministic per case.
+    fn draw_bytes(rng: &mut ChaCha8Rng) -> [u8; 32] {
+        let mut bytes = [0u8; 32];
+        rng.fill_bytes(&mut bytes);
+        bytes
+    }
+
+    /// A post-NU6.3 height (past the regtest NU6.3 activation) at which the migration transactions
+    /// are built and their fees computed.
+    pub(crate) const TARGET_HEIGHT: u32 = 100;
+
+    /// An account's Orchard spending key, derived from `seed` so tests can vary the account across
+    /// proptest cases. Draws bytes until they form a valid spending key.
+    pub(crate) fn spending_key(seed: u64) -> SpendingKey {
+        let mut rng = ChaCha8Rng::seed_from_u64(seed);
+        loop {
+            let bytes = draw_bytes(&mut rng);
+            if let Some(sk) = SpendingKey::from_bytes(bytes).into_option() {
+                return sk;
+            }
+        }
+    }
+
+    /// An account's Orchard full viewing key (see [`spending_key`]).
+    pub(crate) fn account(seed: u64) -> FullViewingKey {
+        FullViewingKey::from(&spending_key(seed))
+    }
+
+    /// A regtest network with the pre-NU6.3 upgrades active, and NU6.3 active only when requested.
+    /// The migration builds on a network where NU6.3 (the Ironwood pool) is live.
+    pub(crate) fn regtest_network(nu6_3_active: bool) -> LocalNetwork {
+        let nu6_3 = if nu6_3_active {
+            Some(BlockHeight::from_u32(10))
+        } else {
+            None
+        };
+        LocalNetwork {
+            overwinter: Some(BlockHeight::from_u32(1)),
+            sapling: Some(BlockHeight::from_u32(2)),
+            blossom: Some(BlockHeight::from_u32(3)),
+            heartwood: Some(BlockHeight::from_u32(4)),
+            canopy: Some(BlockHeight::from_u32(5)),
+            nu5: Some(BlockHeight::from_u32(6)),
+            nu6: Some(BlockHeight::from_u32(7)),
+            nu6_1: Some(BlockHeight::from_u32(8)),
+            nu6_2: Some(BlockHeight::from_u32(9)),
+            nu6_3,
+            #[cfg(zcash_unstable = "nu7")]
+            nu7: None,
+        }
+    }
+
+    /// An Orchard note of `value` owned by `fvk`, with its randomness drawn from `rng`.
+    fn orchard_note(fvk: &FullViewingKey, value: u64, rng: &mut ChaCha8Rng) -> Note {
+        let recipient = fvk.address_at(0u32, Scope::External);
+        let note_value = NoteValue::from_raw(value);
+        let rho = loop {
+            let bytes = draw_bytes(rng);
+            if let Some(rho) = Rho::from_bytes(&bytes).into_option() {
+                break rho;
+            }
+        };
+        let rseed = loop {
+            let bytes = draw_bytes(rng);
+            if let Some(rseed) = RandomSeed::from_bytes(bytes, &rho).into_option() {
+                break rseed;
+            }
+        };
+        Note::from_parts(recipient, note_value, rho, rseed, NoteVersion::V2)
+            .into_option()
+            .expect("valid note parts")
+    }
+
+    /// An Orchard note of `value` owned by `fvk`, with its randomness derived from `seed` (so
+    /// proptest varies the note across cases), placed as the sole leaf of an otherwise-empty
+    /// note-commitment tree, with the matching anchor. The authentication path uses the empty-subtree
+    /// roots for a single leaf at position 0, so `add_orchard_spend`'s anchor check
+    /// (`path.root(cmx) == anchor`) accepts it.
+    pub(crate) fn single_note_witness(
+        fvk: &FullViewingKey,
+        value: u64,
+        seed: u64,
+    ) -> (Note, MerklePath, Anchor) {
+        let mut rng = ChaCha8Rng::seed_from_u64(seed);
+        let note = orchard_note(fvk, value, &mut rng);
+        let cmx = ExtractedNoteCommitment::from(note.commitment());
+        let auth_path = core::array::from_fn(|level| {
+            let level = Level::from(level as u8);
+            MerkleHashOrchard::empty_root(level)
+        });
+        let path = MerklePath::from_parts(0, auth_path);
+        let anchor = path.root(cmx);
+        (note, path, anchor)
+    }
+
+    /// `values.len()` Orchard notes owned by `fvk`, placed as the first leaves of one note-commitment
+    /// tree, each paired with its authentication path against the single shared anchor. Randomness is
+    /// derived from `seed`. This lets a test build a multi-spend transaction whose spends all anchor to
+    /// the same root. Assumes `values` is non-empty.
+    pub(crate) fn shared_anchor_witnesses(
+        fvk: &FullViewingKey,
+        values: &[u64],
+        seed: u64,
+    ) -> (Vec<(Note, MerklePath)>, Anchor) {
+        // The notes and their leaf hashes.
+        let notes: Vec<Note> = values
+            .iter()
+            .enumerate()
+            .map(|(i, &v)| {
+                let mut rng = ChaCha8Rng::seed_from_u64(seed.wrapping_add(i as u64));
+                orchard_note(fvk, v, &mut rng)
+            })
+            .collect();
+        let leaves: Vec<MerkleHashOrchard> = notes
+            .iter()
+            .map(|n| MerkleHashOrchard::from_cmx(&ExtractedNoteCommitment::from(n.commitment())))
+            .collect();
+
+        // Filled subtree roots per level: `levels[l][p]` is the root of the subtree at level `l`,
+        // position `p`. Level 0 is the leaves; each higher level combines pairs, using the
+        // empty-subtree root for a missing right sibling. Positions past `levels[l].len()` are empty.
+        let mut levels: Vec<Vec<MerkleHashOrchard>> =
+            Vec::with_capacity(NOTE_COMMITMENT_TREE_DEPTH + 1);
+        levels.push(leaves);
+        for l in 0..NOTE_COMMITMENT_TREE_DEPTH {
+            let level = Level::from(l as u8);
+            let cur = &levels[l];
+            let mut next = Vec::with_capacity(cur.len().div_ceil(2));
+            let mut p = 0;
+            while p < cur.len() {
+                let left = cur[p];
+                let right = cur
+                    .get(p + 1)
+                    .copied()
+                    .unwrap_or_else(|| MerkleHashOrchard::empty_root(level));
+                next.push(MerkleHashOrchard::combine(level, &left, &right));
+                p += 2;
+            }
+            levels.push(next);
+        }
+
+        // Each leaf's authentication path: the sibling subtree root at every level (its computed value
+        // when filled, else the empty-subtree root).
+        let witnesses: Vec<(Note, MerklePath)> = notes
+            .into_iter()
+            .enumerate()
+            .map(|(i, note)| {
+                let auth_path = core::array::from_fn(|l| {
+                    let level = Level::from(l as u8);
+                    let sibling = (i >> l) ^ 1;
+                    levels[l]
+                        .get(sibling)
+                        .copied()
+                        .unwrap_or_else(|| MerkleHashOrchard::empty_root(level))
+                });
+                (note, MerklePath::from_parts(i as u32, auth_path))
+            })
+            .collect();
+
+        // Every leaf's path yields the same root; assert it so a broken helper fails loudly rather
+        // than silently producing mismatched anchors.
+        let root_of = |(note, path): &(Note, MerklePath)| {
+            path.root(ExtractedNoteCommitment::from(note.commitment()))
+        };
+        let anchor = root_of(&witnesses[0]);
+        for w in &witnesses {
+            assert_eq!(root_of(w), anchor, "shared-anchor witnesses inconsistent");
+        }
+        (witnesses, anchor)
+    }
+}
--- a/zcash_pool_migration_backend/src/build/end_to_end.rs
+++ b/zcash_pool_migration_backend/src/build/end_to_end.rs
@@ -1,0 +1,86 @@
+//! End-to-end test of the migration pipeline for a typical wallet, using only the crate's public API.
+//! It doubles as a usage example: plan the note split, plan the preparation transactions, build and
+//! pre-sign one, then build and pre-sign a pool-crossing transfer.
+
+use orchard::keys::{FullViewingKey, SpendAuthorizingKey};
+use rand_chacha::ChaCha8Rng;
+use rand_core::SeedableRng;
+use zcash_protocol::value::COIN;
+
+use super::test_util::{TARGET_HEIGHT, regtest_network, single_note_witness, spending_key};
+use super::{build_prep_tx, build_transfer_pczt, sign_pczt};
+use crate::note_splitting::{FeePolicy, Zip317FeePolicy, plan_note_split};
+use crate::preparation::{PREP_TX_ACTIONS, PrepInput, plan_preparation};
+
+/// note split -> preparation plan -> build + sign a preparation transaction -> build + sign a
+/// pool-crossing transfer, for a typical single-note wallet.
+#[test]
+fn migration_pipeline_end_to_end() {
+    let seed = 42u64;
+    let params = regtest_network(true);
+
+    // The account, and (conceptually) its single spendable Orchard note holding the whole balance.
+    let sk = spending_key(seed);
+    let fvk = FullViewingKey::from(&sk);
+    let ask = SpendAuthorizingKey::from(&sk);
+    let balance = 78 * COIN;
+
+    // 1. Note split: decompose the balance into canonical self-funding denominations.
+    let prep_fee = PREP_TX_ACTIONS as u64 * Zip317FeePolicy.marginal_fee_zatoshi();
+    let split = {
+        let mut rng = ChaCha8Rng::seed_from_u64(seed);
+        plan_note_split(balance, prep_fee, &mut rng)
+    };
+    let funding = split.migration_outputs();
+    assert!(!funding.is_empty(), "the balance yields funding notes");
+
+    // 2. Preparation: plan the send-to-self transactions that mint those funding notes. A typical
+    //    wallet (one note, a handful of funding notes) prepares in a single transaction.
+    let prep = plan_preparation(&[balance], funding, split.prep_fee_zatoshi())
+        .expect("the balance funds the preparation");
+    assert_eq!(prep.layer_count(), 1);
+    assert_eq!(prep.transaction_count(), 1);
+
+    // 3. Build + pre-sign the preparation transaction. The wallet backend resolves the transaction's
+    //    single input (PrepInput::Wallet(0)) to the account's note and its witness against the anchor.
+    let tx = &prep.layers()[0][0];
+    assert!(matches!(tx.inputs(), [PrepInput::Wallet(0)]));
+    let (note, path, anchor) = single_note_witness(&fvk, balance, seed);
+    let (prep_pczt, placed) = build_prep_tx(
+        &params,
+        TARGET_HEIGHT,
+        &fvk,
+        anchor,
+        vec![(note, path)],
+        tx.outputs(),
+        ChaCha8Rng::seed_from_u64(seed + 1),
+    )
+    .expect("the preparation transaction builds");
+    assert_eq!(
+        prep_pczt.orchard().actions().len(),
+        PREP_TX_ACTIONS,
+        "the preparation bundle is padded to exactly 16 actions"
+    );
+    assert_eq!(placed.len(), tx.outputs().len(), "every output is located");
+    sign_pczt(prep_pczt, &ask).expect("pre-signing the preparation transaction");
+
+    // 4. Build + pre-sign a pool-crossing transfer for one funding note (witnessed directly here, as
+    //    it would be once the preparation transaction is mined). It sends exactly the canonical
+    //    denomination into the Ironwood pool.
+    let crossing = split.crossing_values()[0];
+    let funding_value = funding[0];
+    let (fnote, fpath, fanchor) = single_note_witness(&fvk, funding_value, seed + 2);
+    let transfer_pczt = build_transfer_pczt(
+        &params,
+        TARGET_HEIGHT,
+        TARGET_HEIGHT + 40_000,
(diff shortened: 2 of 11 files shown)


zcash_ironwood_migration_backend/src/build/split.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// The `(output_index, value)` of the Orchard change output, if the split produced one.
+    pub fn change(&self) -> Option<(u32, u64)> {
+        self.change
+    }
+}
+
+/// Exact ZIP-317 fee (in zatoshi) for a note-split transaction. The bundle disables cross-address
+/// transfers, so each spend and each output occupies its own action:
+/// `actions = n_spends + n_outputs`, floored at the ZIP-317 grace count.
```
dannywillems: Important: for the migration, we cap the number of actions! To discuss later.

zcash_pool_migration_backend/src/build/transfer.rs
```diff
@@ -0,0 +1,136 @@
...
+///
+/// # Errors
+///
+/// Returns [`BuildError::Build`] if the builder or PCZT pipeline fails (including when the note's
+/// buffer does not cover the transfer fee, which unbalances the transaction).
+#[allow(clippy::too_many_arguments)]
+pub fn build_transfer_pczt<P, R>(
+    params: &P,
+    target_height: u32,
+    expiry_height: u32,
```
dannywillems, line 50: Note: this will be an important parameter defined at the scheduling level.

zcash_pool_migration_backend/src/build/split.rs
```diff
@@ -0,0 +1,501 @@
...
+///
+/// The caller supplies `rng` (a cryptographically secure RNG in production, e.g. `OsRng`; tests can
+/// pass a seeded one), keeping this builder pure.
+pub fn build_split_pczt<P, R>(
+    params: &P,
+    target_height: u32,
+    orchard_fvk: &FullViewingKey,
+    anchor: orchard::Anchor,
+    spends: Vec<(orchard::note::Note, orchard::tree::MerklePath)>,
+    output_values: &[u64],
```
dannywillems: Note: this comes from the note splitting. That's the output we actually want to have to cross the turnstile. The role of this method is to consolidate the notes into the split outputs to simplify the construction of PCZT that would be crossing the turnstile.

zcash_pool_migration_backend/src/build/end_to_end.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use super::test_util::{TARGET_HEIGHT, regtest_network, single_note_witness, spending_key};
+use super::{build_prep_tx, build_transfer_pczt, sign_pczt};
+use crate::note_splitting::{FeePolicy, Zip317FeePolicy, plan_note_split};
+use crate::preparation::{PREP_TX_ACTIONS, PrepInput, plan_preparation};
+
+/// note split -> preparation plan -> build + sign a preparation transaction -> build + sign a
+/// pool-crossing transfer, for a typical single-note wallet.
+#[test]
+fn migration_pipeline_end_to_end() {
```
dannywillems, line 18: This is to test the user API.

zcash_pool_migration_backend/src/build.rs
```diff
@@ -0,0 +1,278 @@
...
+    use orchard::tree::{MerkleHashOrchard, MerklePath};
+    use orchard::value::NoteValue;
+    use orchard::{Anchor, NOTE_COMMITMENT_TREE_DEPTH};
+    use rand_chacha::ChaCha8Rng;
+    use rand_core::{RngCore, SeedableRng};
+    use zcash_protocol::consensus::BlockHeight;
+    use zcash_protocol::local_consensus::LocalNetwork;
+
+    /// 32 random bytes from a `seed`-derived RNG, keeping calls deterministic per case.
+    fn draw_bytes(rng: &mut ChaCha8Rng) -> [u8; 32] {
```
dannywillems, line 109: Thinking about moving the test helpers later.

zcash_pool_migration_backend/src/build.rs
```diff
@@ -0,0 +1,278 @@
...
+            nu6_1: Some(BlockHeight::from_u32(8)),
+            nu6_2: Some(BlockHeight::from_u32(9)),
+            nu6_3,
+            #[cfg(zcash_unstable = "nu7")]
+            nu7: None,
+        }
+    }
+
+    /// An Orchard note of `value` owned by `fvk`, with its randomness drawn from `rng`.
+    fn orchard_note(fvk: &FullViewingKey, value: u64, rng: &mut ChaCha8Rng) -> Note {
```
dannywillems, line 161: This could be moved in a test utils in the crate `orchard`.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
+//! Note-preparation transaction planning: how to restructure a wallet's spendable source-pool notes
```
dannywillems, line 1: This is the core of the PR.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+//! more dust notes than one transaction can consume, needs **layers**. A layer is a set of
+//! transactions with no dependencies between them (buildable, provable, and broadcastable in
+//! parallel); a later layer may spend the outputs of an earlier one, but only after they are mined
+//! and a boundary passes, so each extra layer extends the preparation phase by roughly one anchor
+//! bucket. The planner therefore prefers fewer layers (which dominate the wall-clock) over fewer
+//! transactions.
+//!
+//! # The strategy
+//!
+//! The planner is a largest-first layered greedy. In each layer it feeds each output transaction from
```
dannywillems, line 29: We should have a trait later, that allows building different strategy. I see the trait as the actual NP problem, and an implementation as a solution to the NP problem.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+//! fee; any further residue is sub-fee dust.
+//!
+//! When a single note can produce every funding note, the planner takes a fan-out fast path: it splits
+//! that note through a BALANCED tree (fanning out by [`FUNDING_OUTPUTS_PER_TX`] per layer), so the
+//! depth is logarithmic in the funding-note count rather than linear in it. The balanced tree uses more
+//! transactions, and so more fee, than a linear feeder chain would; it trades that for fewer layers,
+//! which dominate the wall-clock. Every other shape (many notes, mixed sizes, dust) uses the layered
+//! greedy above.
+//!
+//! This is a pure planner: it works in note *values* (in zatoshi) and does no cryptography or I/O. It
```
dannywillems, line 51: This is purely providing a solution to the note preparation problem. The callers of the algorithms are implemented in future (already opened) patches.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+use alloc::vec::Vec;
+
+use core::fmt;
+
+/// The exact number of Orchard actions in every note-preparation transaction ([ZIP 318]): each is
+/// padded up to this count, so no preparation transaction is distinguishable from another by its
+/// action count, and one transaction handles at most this many notes in total (spends plus outputs).
+///
+/// [ZIP 318]: https://zips.z.cash/zip-0318
+pub const PREP_TX_ACTIONS: usize = 16;
```
dannywillems, line 67: Note that all of these parameters could become part of the trait of the NP problem, as mentioned earlier.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+pub const FUNDING_OUTPUTS_PER_TX: usize = PREP_TX_ACTIONS - 2;
+
+/// The most notes one transaction consolidates: the action budget less the single output it produces
+/// (`16 - 1`).
+pub const CONSOLIDATION_INPUTS_PER_TX: usize = PREP_TX_ACTIONS - 1;
+
+/// A note a preparation transaction spends: either one of the wallet's original spendable notes, or a
+/// note an earlier layer produced.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PrepInput {
```
dannywillems, line 80: See `PrepInput` as variables of a multi-linear system of equations, as described in the ZIP, or nodes in the DAG. `Wallet(usize)` represents nodes without fan-in, i.e. "leaves", where `Prior` represents the internal nodes.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    Prior {
+        layer: usize,
+        transaction: usize,
+        output: usize,
+    },
+}
+
+/// A note a preparation transaction produces.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PrepOutput {
```
dannywillems, line 93: Also see these as nodes in the DAG. `Intermediate` are strict internal nodes, `Funding` and `Change` are nodes without fan-out.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    Funding(u64),
+    /// An intermediate ("feeder") note, spent by a later layer to route value forward.
+    Intermediate(u64),
+    /// Leftover value returned to the source pool.
+    Change(u64),
+}
+
+impl PrepOutput {
+    /// The note value this output carries.
+    pub fn value(&self) -> u64 {
```
dannywillems, line 104: This simply gets the value of the node in the DAG.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+type PoolNote = (PrepInput, u64);
+
+/// Plan the note-preparation transactions that mint `funding` (the self-funding note values, in
+/// zatoshi) from `available` (the wallet's spendable source-pool note values, in zatoshi), reserving
+/// `fee_per_tx` zatoshi for each transaction (the ZIP-317 fee of a padded [`PREP_TX_ACTIONS`]-action
+/// transaction).
+///
+/// Returns an empty plan when `funding` is empty, and [`PrepError::InsufficientFunds`] when the
+/// available value cannot cover the funding notes plus the per-transaction fees.
+pub fn plan_preparation(
```
dannywillems, line 262: This is actually the algorithm that would build the DAG.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    use crate::note_splitting::{FeePolicy, Zip317FeePolicy};
+
+    /// A representative padded [`PREP_TX_ACTIONS`]-action ZIP-317 fee reserve for the tests (each
+    /// action costs one [`Zip317FeePolicy`] marginal fee). The planner treats it opaquely.
+    fn fee_per_tx() -> u64 {
+        PREP_TX_ACTIONS as u64 * Zip317FeePolicy.marginal_fee_zatoshi()
+    }
+
+    /// The funding values a plan mints, sorted, for multiset comparison against the request.
+    fn sorted_funding(plan: &PreparationPlan) -> Vec<u64> {
```
dannywillems, line 710: This could be moved at the structure level, in the `impl`.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+        let mut out = plan.funding_notes();
+        out.sort_unstable();
+        out
+    }
+
+    /// Assert every structural invariant of a plan against its inputs.
+    fn assert_plan_valid(plan: &PreparationPlan, available: &[u64], funding: &[u64], fee: u64) {
```
dannywillems, line 717: This could also be moved into the `impl` of the structure `PreparationPlan`.

zcash_pool_migration_backend/src/preparation.rs
dannywillems, line 262: Different data structures might be more suitable for the complexity. I haven't computed the actual memory footprint nor the theoretical complexity.

I would focus first on having an actual implementation with working tests, and after that improving the complexity of the algorithms.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    fee_per_tx: u64,
+) -> Result<PreparationPlan, PrepError> {
+    // Funding values still to produce, largest first (so `last()` is the smallest).
+    let mut remaining: Vec<u64> = funding.iter().copied().filter(|&v| v > 0).collect();
+    remaining.sort_unstable_by(|a, b| b.cmp(a));
+
+    // Exact-match pass: a wallet note already equal to a funding value IS that funding note, so it is
+    // used directly, with no preparation transaction and no fee. The matched notes are removed from
+    // both the funding still to produce and the notes available to spend.
+    let mut used = vec![false; available.len()];
```
dannywillems, line 274: Probably this could be using a HashSet or something else more efficient.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    let mut direct_funding: Vec<(usize, u64)> = Vec::new();
+    remaining.retain(|&f| {
```
dannywillems, line 276: This does a `filter` on the existing notes and populate `direct_funding/used`.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+            Some(i) => {
+                used[i] = true;
+                direct_funding.push((i, f));
+                false
+            }
+            None => true,
+        }
+    });
+
+    let mut layers: Vec<Vec<PrepTransaction>> = Vec::new();
```
dannywillems, line 291: In the case we have only direct funding notes, perfect case. We have already the solution to the problem.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    // The notes available to spend in the current layer (layer 0: the wallet's own notes not already
+    // used directly as funding notes).
+    let mut current: Vec<PoolNote> = available
+        .iter()
+        .enumerate()
+        .filter(|(i, _)| !used[*i])
+        .map(|(i, &v)| (PrepInput::Wallet(i), v))
+        .collect();
+
+    while !remaining.is_empty() {
```
dannywillems, line 333: This is where the interesting logic happens. We build the DAG here.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+            direct_funding,
+        });
+    }
+
+    // Fan-out fast path: when a single wallet note can produce every remaining funding note, split it
+    // through a balanced tree (depth logarithmic in the note count) rather than the linear feeder chain
+    // the layered loop below would build for a lone large note. Only that case takes this path;
+    // everything else (many notes, mixed sizes, dust) falls through to the layered greedy unchanged.
+    // Trade-off: the balanced tree uses more transactions (fees) than the chain, buying fewer layers.
+    if let Some((idx, big)) = available
```
dannywillems, line 304: `available` represents here the remaining notes of the wallet, filtered by the notes that can already be used to fund equally valued outputs/funding notes. We could define the variable above. These would be the actual "starting nodes" in the DAG to build the non-trivial paths.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    /// preparation transaction produces one output per scheduled part (up to `FUNDING_OUTPUTS_PER_TX`),
+    /// not one transaction per part (that is the phase-2 transfer). So this is one layer, one
+    /// transaction, with several funding outputs.
+    #[test]
+    fn common_case_is_one_layer_one_tx() {
+        let funding = [500u64, 200, 100, 20, 5];
+        let total: u64 = funding.iter().sum::<u64>() + fee_per_tx() + 10_000;
+        let plan = plan_preparation(&[total], &funding, fee_per_tx()).unwrap();
+        assert_eq!(plan.layer_count(), 1);
+        assert_eq!(plan.transaction_count(), 1);
```
dannywillems, line 874: Do we actually need to create transactions? I don't think so... The notes are already present.

EDIT: Ah! Yes, we want as it is actually a single note with the total of the funding notes.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+        let funding = [100u64, 100, 100, 100];
+        let whale = 400 + 3 * fee_per_tx();
+        let plan = plan_preparation(&[whale], &funding, fee_per_tx()).unwrap();
+        assert_plan_valid(&plan, &[whale], &funding, fee_per_tx());
+        assert_eq!(sorted_funding(&plan), vec![100, 100, 100, 100]);
+    }
+
+    /// The planner is deterministic: identical inputs always produce an identical plan.
+    #[test]
+    fn deterministic() {
```
dannywillems, line 1215: This could be actually a proptest.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    #[test]
+    fn varied_funding_from_one_whale() {
+        let funding = [500_000u64, 200_000, 100_000, 20_000, 5_000, 2_000];
+        let whale = funding.iter().sum::<u64>() + fee_per_tx() + 12_345;
+        let plan = plan_preparation(&[whale], &funding, fee_per_tx()).unwrap();
+        assert_plan_valid(&plan, &[whale], &funding, fee_per_tx());
+        assert_eq!(plan.layer_count(), 1);
+        assert_eq!(plan.transaction_count(), 1);
+    }
+
```
dannywillems, line 1281: I feel there isn't enough example. The reviewer is encouraged to give more ideas.

dannywillems:
Self-ACK, modulo comments.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+//! its spends plus its outputs, so one transaction can consume and produce at most
+//! [`PREP_TX_ACTIONS`] notes in total (for example 15 spends and one output when consolidating, or
+//! one spend and 15 outputs when splitting). A splitting transaction therefore mints MANY funding
+//! notes at once, one output per scheduled part (up to [`FUNDING_OUTPUTS_PER_TX`]); the number of
+//! funding notes in a preparation transaction is not one. The one-transaction-per-part shape belongs
+//! to the phase-2 crossing transfers (each spends a single funding note), not to preparation.
+//!
+//! A single transaction therefore cannot always turn the wallet's notes into every funding note: a
+//! note that must fan out into more outputs than one transaction holds, or a balance spread across
+//! more dust notes than one transaction can consume, needs **layers**. A layer is a set of
```
nuttycom, line 20: If the wallet holds dust-valued notes, we should *only* consume these if the wallet would otherwise be creating dummy padding notes.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+//! and a boundary passes, so each extra layer extends the preparation phase by roughly one anchor
+//! bucket. The planner therefore prefers fewer layers (which dominate the wall-clock) over fewer
+//! transactions.
+//!
+//! # The strategy
+//!
+//! The planner is a largest-first layered greedy. In each layer it feeds each output transaction from
+//! the largest available note it can (one big note funds up to [`FUNDING_OUTPUTS_PER_TX`] funding
+//! notes), routes every leftover forward as an intermediate ("feeder") note, and consolidates notes
+//! too small to fund anything on their own into feeder notes. Once all funding notes are scheduled it
```
nuttycom, line 32: As a documentation nit, it would be helpful to distinguish between notes and intermediate balances used in determination of what actual notes will be consumed and created. I would prefer those intermediate balances ("feeder" notes, if I'm reading this correctly?) to be called something other than "notes" - let's reserve the term "note" exclusively with notes in the protocol sense. "Feeder" is a reasonable name for this interstitial type; we just shouldn't refer to a feeder as a note.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+//! consolidates the feeders that no layer spent into a single residual note, matching ZIP 318's
+//! "one note per part plus at most one residual note". For a typical wallet (a few notes, a handful of
+//! funding notes) this is a single layer; extra layers appear only for a lone large note fanning out
+//! into many funding notes, or a dust-heavy balance.
```
nuttycom, line 36: "dust" is also a term that needs to be used with care; it exclusively refers to notes with value below the marginal fee, i.e. those that do not have any economic value and that can only be spent where a dummy would otherwise require a fee to be paid anyway.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    mut pool: Vec<PoolNote>,
+    layer: usize,
+    fee: u64,
+    txs: &mut Vec<PrepTransaction>,
+    next: &mut Vec<PoolNote>,
+) -> Vec<PoolNote> {
+    if pool.len() < 2 {
+        return pool;
+    }
+    pool.sort_unstable_by(|a, b| b.1.cmp(&a.1));
```
nuttycom, line 499: nit: `PoolNote` being an alias for a tuple makes this slightly harder to read, it would be better as a struct.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+                f.write_str("available notes cannot fund the requested notes plus preparation fees")
+            }
+        }
+    }
+}
+
+impl core::error::Error for PrepError {}
+
+/// A note available to spend in the current layer: a reference and its value.
+type PoolNote = (PrepInput, u64);
```
nuttycom, line 253: This is confusingly named, because we always use "Pool" to refer to "shielded pool" - also, why does PrepInput not just carry its value along with it? As far as I can tell, we never construct a PrepInput in a place where we don't know its value.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+        let mut partial: Vec<(PrepInput, Vec<u64>, u64)> = Vec::new();
+        let mut consolidatable: Vec<PoolNote> = Vec::new();
+
+        for (in_ref, value) in current.drain(..) {
+            if remaining.is_empty() {
+                // Everything is already scheduled; this note stays unspent in the wallet.
+                continue;
+            }
+            let smallest = *remaining.last().expect("remaining is non-empty");
+            if value <= fee_per_tx || value - fee_per_tx < smallest {
```
nuttycom, line 352: I'm not sure about the "consolidation" concept here - I think that this is mistaken; unless there are padding inputs where these can be consumed, we will not consolidate uneconomic notes at all; we will leave them in the wallet.

zcash_pool_migration_backend/src/preparation.rs
dannywillems: > I would prefer those intermediate balances ("feeder" notes, if I'm reading this correctly?) 

They are actually protocol notes. They're "changes" notes in the DAG construction.

nuttycom (commented):
Additional notes related to uneconomic inputs.

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Agreed. I asked Claude to rename the non-dust occurrences into `sub-quantum`, in the idea that "denomination" is a "quantum".

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+        sizes.push(take);
+        n -= take;
+    }
+    sizes
+}
+
+/// Consolidate `pool` into feeder notes: append one consolidation transaction per batch (of at most
+/// [`CONSOLIDATION_INPUTS_PER_TX`] inputs) to `txs` in layer `layer`, with its feeder pushed to
+/// `next`. Returns any notes whose batch could not cover the fee (too small to consolidate).
+fn consolidate(
```
nuttycom, line 489: Building separate consolidation transactions doesn't make sense for individual notes below the marginal fee, so I would recast consolidation as filling in for dummy notes in an existing transaction; once all economic value is exhausted, *all* the remaining sub-economic notes should be just left in the wallet. Does it simplify things to filter out all the sub-economic notes first, and then sprinkle them in where otherwise dummies would be introduced as padding, at the end?

zcash_pool_migration_backend/src/preparation.rs
dannywillems: It must be applied at different places, including the tests and variable names.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -0,0 +1,1300 @@
...
+    fn dust_is_consolidated_first() {
+        // Twenty notes each far below the single 100-unit funding note (plus fees).
+        let per = fee_per_tx() + 20;
+        let available: Vec<u64> = core::iter::repeat_n(per, 20).collect();
+        let funding = [100u64];
+        let plan = plan_preparation(&available, &funding, fee_per_tx()).unwrap();
+        assert!(plan.layer_count() >= 2, "dust must consolidate first");
+        assert_eq!(plan.residual_count(), 1, "one residual note");
+        assert_plan_valid(&plan, &available, &funding, fee_per_tx());
+    }
```
nuttycom, line 906: Hm, I see. So a number of notes that have change value that *can* fund their own spend can be consolidated, but only if the consolidation yields a note that can fund its own fee? Such a note can never "cross the threshold" on its own, because its value is less than the smallest pool-crossing note that we support, currently 0.01 ZEC. 

From a user perspective, it only makes sense to spend a note if doing so contributes to the value of a pool-crossing amount. Otherwise, the note should be left in the wallet. Is that's the behavior that's implemented here? It looks like this is focused on the fee amount instead of the pool-crossing value.

nuttycom (approved):
utACK. Overall I'm convinced that this is the right direction - the only improvement I see is to generate fewer transactions in the case that spending small notes ends up not being economic - if spending any number of notes does not result in at least 0.01015000 ZEC (the minimum value for pool-crossing, including fee) then we should leave those notes unspent.

dannywillems:
Merging, as approved. And I will address individual comments! There are still cases we might want to think about.

zcash_pool_migration_backend/src/build/transfer.rs
```diff
@@ -0,0 +1,141 @@
...
+) -> Result<pczt::Pczt, BuildError>
+where
+    P: Parameters + Clone,
+    R: RngCore + CryptoRng,
+{
+    let target = BlockHeight::from_u32(target_height);
+    let expiry = BlockHeight::from_u32(expiry_height);
+    // The Ironwood bundle is output-only (no spend to anchor against); the empty tree is the
+    // output-only anchor convention.
+    let config = build_config(anchor, Some(orchard::Anchor::empty_tree()));
```
nuttycom, line 66: Hmm, okay, this is fine for these particular migration transactions I guess because we're not relying upon them being indistinguishable from Ironwood spends.

nuttycom: If we instead passed a real anchor, it would be possible for another wallet to "disguise" a zero-change payment to an Ironwood recipient as a migration transaction - say a 0.01 ZEC reveal across the turnstile, with an additional amount funded from ironwood, but this isn't critical. Using the empty tree *does* leak more information than necessary, so ideally we should provide the ironwood anchor as an argument as well; it would be the anchor at the same height as the Orchard anchor being used.

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Addressed in https://github.com/zcash/librustzcash/pull/2679

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Addressed in https://github.com/zcash/librustzcash/pull/2679

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Addressed in https://github.com/zcash/librustzcash/pull/2679

nuttycom (commented):
post-hoc utACK for the remainder of the commits that I missed reviewing the first time around. Using a real ironwood anchor instead of the empty tree is desirable.

dannywillems:
> utACK. Overall I'm convinced that this is the right direction - the only improvement I see is to generate fewer transactions in the case that spending small notes ends up not being economic - if spending any number of notes does not result in at least 0.01015000 ZEC (the minimum value for pool-crossing, including fee) then we should leave those notes unspent.

As discussed in a call, I don't fully agree. If we have 15 notes of 0.01 ZEC ($5.3 as of today), we might still want to consolidate them into one of 0.15 (~$75 as of today) - reaching the 16 actions per note (modulo fees). And after that migrating to Ironwood in one transaction. And the 15 notes of 0.01 ZEC might be something from a real use case if we think about buying cups of coffee, where we get small changes.

dannywillems:
> post-hoc utACK for the remainder of the commits that I missed reviewing the first time around. Using a real ironwood anchor instead of the empty tree is desirable.

Opening follow-up https://github.com/zcash/librustzcash/issues/2680

zcash_pool_migration_backend/src/build/transfer.rs
dannywillems: Linking in https://github.com/zcash/librustzcash/issues/2680

dannywillems:
> > utACK. Overall I'm convinced that this is the right direction - the only improvement I see is to generate fewer transactions in the case that spending small notes ends up not being economic - if spending any number of notes does not result in at least 0.01015000 ZEC (the minimum value for pool-crossing, including fee) then we should leave those notes unspent.
> 
> As discussed in a call, I don't fully agree. If we have 15 notes of 0.01 ZEC ($5.3 as of today), we might still want to consolidate them into one of 0.15 (~$75 as of today) - reaching the 16 actions per note (modulo fees). And after that migrating to Ironwood in one transaction. And the 15 notes of 0.01 ZEC might be something from a real use case if we think about buying cups of coffee, where we get small changes.

Open https://github.com/zcash/librustzcash/issues/2681

dannywillems:
> utACK. Overall I'm convinced that this is the right direction - the only improvement I see is to generate fewer transactions in the case that spending small notes ends up not being economic - if spending any number of notes does not result in at least 0.01015000 ZEC (the minimum value for pool-crossing, including fee) then we should leave those notes unspent.

Open https://github.com/zcash/librustzcash/issues/2681

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Open https://github.com/zcash/librustzcash/issues/2681

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Open https://github.com/zcash/librustzcash/issues/2681

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Open https://github.com/zcash/librustzcash/issues/2681

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Open https://github.com/zcash/librustzcash/issues/2682

zcash_pool_migration_backend/src/build.rs
dannywillems: Open follow-up https://github.com/zcash/librustzcash/issues/2683

zcash_pool_migration_backend/src/preparation.rs
dannywillems: Open https://github.com/zcash/librustzcash/issues/2684

zcash_pool_migration_backend/src/preparation.rs
dannywillems: See https://github.com/zcash/librustzcash/issues/2684

zcash_pool_migration_backend/src/preparation.rs
dannywillems: https://github.com/zcash/librustzcash/issues/2686

zcash_pool_migration_backend/src/preparation.rs
dannywillems: See https://github.com/zcash/librustzcash/issues/2687

zcash_pool_migration_backend/src/preparation.rs
dannywillems: See https://github.com/zcash/librustzcash/issues/2687

zcash_pool_migration_backend/src/preparation.rs
dannywillems: https://github.com/zcash/librustzcash/issues/2688

dannywillems:
Follow-ups open for all threads.

PR #2648, opened 2026-07-16
Add Ironwood historical note selection

czarcas7ic:
Adds `WalletDb::get_unspent_ironwood_notes_at_historical_height` so callers can retrieve Ironwood notes that existed and were unspent at a requested height. The Orchard and Ironwood selectors share one query implementation while preserving their separate tables and pool identities.


czarcas7ic:
@dannywillems this is good to go now, thanks

dannywillems (approved):
utACK. This copies the existing orchard variant by generalizing the table name.

PR #2650, opened 2026-07-16
zcash_client_backend: accept expiry_height in create_proposed_transactions

Cosmos-Harry:
## Summary

`create_pczt_from_proposal` already accepts an `expiry_height: Option<BlockHeight>` override, letting a caller replace the builder-derived expiry height before finalization. The non-PCZT sibling `create_proposed_transactions` (used directly by, e.g., zallet's `z_sendmany`) had no equivalent — the expiry was always derived purely from the proposal's `min_target_height`, with no way for a caller to widen or override it.

This PR adds the same `expiry_height` parameter to `create_proposed_transactions` (and the private `create_proposed_transaction` / `build_proposed_transaction` helpers it shares with the PCZT path), applying it via the existing, already-tested `Builder::with_expiry_height` immediately after `Builder::new` — before any inputs are added or anything is signed. Same validation as the PCZT sibling: a nonzero override below the proposal's target height is rejected with `Error::ExpiryHeightBelowTargetHeight`; a zero override (disable expiry) is exempt.

The PCZT call site is left behaviorally unchanged — it continues to apply its override post-build via `pczt_parts.expiry_height` as before; this PR just threads `None` through the now-shared parameter list for that path so both callers of `build_proposed_transaction` compile against one signature.

Per review, the parameter is named `expiry_height` rather than `target_expiry_height` — it literally just sets the expiry, with the default computed from the target height as before. The final commit also renames the pre-existing `create_pczt_from_proposal` parameter to match (parameter names are not part of the Rust API surface, so this is not a breaking change).

## Motivation — updated

This was originally motivated by zcash/zallet#634, where `z_sendmany` built already-expired transactions on a live chain because zallet's cached view of the chain tip had fallen ~1000 blocks behind reality. **That specific bug turned out to have a different, simpler root cause**: zallet's `zebra` backend was pinned to `zebra-rpc` 11.0.0, which had a since-fixed non-finalized-block-syncer bug (ZcashFoundation/zebra#10841, merged and released in `zebra-rpc` 11.1.0). Bumping that dependency alone fully resolves the reproduced issue — no zallet-side code change is needed after all, and no zallet PR is forthcoming on top of this one.

So this PR is no longer a "prerequisite fix" for an active bug. What remains is a smaller, standalone case for it: `create_proposed_transactions` has no way to defend itself if a caller's cached tip view goes stale again for some other reason in the future (a slow restart, a different regression, network hiccups during catch-up) — it simply trusts `min_target_height` and a fixed ~40-block buffer. Adding the same override the PCZT path already has closes that gap and keeps the two builders consistent, but I want to be upfront that it isn't fixing a currently-active defect.

## Test plan

- [x] `cargo test -p zcash_client_backend --features "pczt,transparent-inputs,orchard" --lib` — 74/74 passing
- [x] `cargo fmt -p zcash_client_backend -- --check` — clean
- [x] `cargo clippy -p zcash_client_backend --features "pczt,transparent-inputs,orchard" --all-targets -- -D warnings` — one pre-existing unrelated warning at `wallet.rs:2758` (`bind_instead_of_map`), confirmed present on `main` before this change too
- [x] CHANGELOG entry added under `[Unreleased] / Changed` (breaking signature change)

Used Claude for the investigation, implementation, and this description; scope and correctness reviewed by me.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -13,6 +13,8 @@ workspace.
 ### Added
 - `zcash_client_backend::proposal::Step::ironwood_action_count`, the Ironwood-pool
   counterpart to `Step::orchard_action_count`. Requires the `orchard` feature.
+
+### Changed
 - `zcash_client_backend::data_api::wallet::create_proposed_transactions` now
   takes an `expiry_height: Option<BlockHeight>` parameter, mirroring the
   one already accepted by `create_pczt_from_proposal`. This lets callers
@@ -20,8 +22,6 @@ workspace.
   construction too — useful when a caller's view of the chain tip used to
   derive `min_target_height` may lag the real tip by more than the default
   expiry buffer covers.
-
-### Changed
 - `zcash_client_backend::proposal::Step::orchard_action_count` now takes the
   `orchard::builder::BundleType` and `orchard::bundle::BundleVersion` that the
   transaction builder will be configured with, and returns a `Result`. It
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -13,6 +13,13 @@ workspace.
 ### Added
 - `zcash_client_backend::proposal::Step::ironwood_action_count`, the Ironwood-pool
   counterpart to `Step::orchard_action_count`. Requires the `orchard` feature.
+- `zcash_client_backend::data_api::wallet::create_proposed_transactions` now
+  takes an `expiry_height: Option<BlockHeight>` parameter, mirroring the
+  one already accepted by `create_pczt_from_proposal`. This lets callers
+  override the builder-derived expiry height for non-PCZT transaction
+  construction too — useful when a caller's view of the chain tip used to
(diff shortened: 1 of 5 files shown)


zcash_client_backend/CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_client_backend::data_api::wallet::create_proposed_transactions` now
```
nuttycom, line 18: This should go in the Changed section; this is a breaking API change, not an addition.

zcash-harry: Fixed in 143f962d — moved to Changed.

zcash-harry:
Rebased on main and renamed the parameter to `expiry_height` throughout. Since parameter names aren't part of the Rust API surface, the final commit also renames the pre-existing `create_pczt_from_proposal` parameter to match, so the two siblings stay consistent — no changelog impact from that part.

nuttycom (approved):

PR #2651, opened 2026-07-17
Patch zewif to the CBOR tag container framing revision

nuttycom:
Update to zewif 0.1.0-rc.3

zewif 0.1.0-rc.3 frames ZeWIF documents as self-described CBOR
(55799(133133([version, payload]))) in place of the ASCII magic +
version header. No code changes are needed: the zewif import module
consumes an in-memory ::zewif::Zewif document, and the container framing
is fully encapsulated by the zewif crate's serialize/deserialize.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7438,9 +7438,9 @@ dependencies = [
 
 [[package]]
 name = "zewif"
-version = "1.0.0-rc.2"
+version = "1.0.0-rc.3"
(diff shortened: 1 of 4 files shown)


Cosmos-Harry (approved):
PR description seems stale, talks about [patch.crates-io] and audit-as-crates-io = false that aren't in the diff. 
Otherwise ACK

PR #2654, opened 2026-07-17
Split AccountBalance unshielded balance into regular and coinbase buckets

### LINKED ISSUE #2653: AccountBalance cannot distinguish transparent coinbase from regular funds; immature coinbase counted spendable
`AccountBalance` exposes a single `unshielded_balance: Balance`, with no way to distinguish coinbase funds (which consensus requires to be spent to shielded outputs, and which are subject to the 100-block maturity rule) from regular transparent funds. In addition, `zcash_client_sqlite`'s `add_transparent_account_balances` applies no coinbase-maturity rule, so immature coinbase value is reported as *spendable* in wallet summaries.

Downstream impact: no zallet RPC can answer "how much can I currently pay to a transparent recipient?" — see zcash/zallet#635, where this took down a mining pool's payout automation (`z_sendmany` failing with `Insufficient balance (have 0, …)` while the wallet reported 100+ ZEC of "spendable" transparent balance).

Proposed: split the unshielded balance into regular (non-coinbase) and coinbase buckets, using the existing `tx_index == 0` coinbase convention (`CoinbaseFilter`), and classify immature coinbase as pending rather than spendable.

pacu:
Fixes #2653.

## Summary

Splits `AccountBalance`'s single unshielded balance into **regular** (non-coinbase) and **coinbase** buckets, and classifies transparent coinbase correctly in the wallet-summary account balances.

Motivated by zcash/zallet#635: a mining pool found that no zallet RPC can answer "how much can I pay to a transparent recipient?", because `AccountBalance::unshielded_balance()` folds coinbase (which consensus requires to be shielded before further spending) into the only transparent number available, and the summary SQL applied no coinbase-maturity rule (immature coinbase was reported *spendable*).

## API changes (`zcash_client_backend`)

- `AccountBalance` now carries `unshielded_regular_balance` + `unshielded_coinbase_balance`, with accessors and `with_*_mut` mutators following the existing per-pool pattern (invariant-checked via `check_total`). Fields are deliberately **not** feature-gated, matching the existing pattern (only the sqlite population is gated on `transparent-inputs`).
- `unshielded_balance()` now returns the combined `Balance` **by value** (regular + coinbase).
- `with_unshielded_balance_mut` is removed (no single field remains for it to mutate).
- Deprecated `unshielded()` keeps returning the combined total, unchanged.
- `total()` / `uneconomic_value()` sum both buckets; `spendable_value()` & co. remain shielded-only as documented.

## Behavior changes (`zcash_client_sqlite`)

`add_transparent_account_balances` classifies each output by the existing conventions:
- coinbase = `IFNULL(t.tx_index, 1) == 0` (same predicate as `CoinbaseFilter`; unknown `tx_index` counts as regular, so light-client data degrades to today's behavior),
- **mature** coinbase (>= `COINBASE_MATURITY_BLOCKS` confirmations) -> coinbase `spendable`,
- **immature** coinbase -> coinbase `pending` (`value_pending_spendability`) — fixing the latent bug where it was counted spendable,
- dust routes to the matching bucket's `uneconomic_value`.

`get_transparent_balances` (the per-address map) is intentionally left unsplit — it feeds a different API surface and the per-address coinbase split is not needed for the motivating issue.

## Tests

- New shared test fns `transparent_coinbase_balance_split` and `transparent_coinbase_balance_dust` (run against the real SQLite store): immature->pending under both the minconf-0 and pending-pass routings, maturity transition after 100 blocks, NULL-`tx_index` -> regular, per-bucket dust routing.
- Pure unit tests for the new `AccountBalance` API surface (mutator routing, by-value sum, MAX_MONEY overflow rejection across the two fields).
- `check_balance` extended so all 15 existing call sites now also pin regular==expected / coinbase==ZERO / deprecated-total-unchanged.

`cargo test -p zcash_client_backend` and `-p zcash_client_sqlite` (with `transparent-inputs orchard test-dependencies`) pass; both crates also compile without `transparent-inputs` and with `--no-default-features`.

No version bumps are included (left to the release process). An end-to-end regtest proof (zebrad + zaino + zallet) of the full flow accompanies the zallet PR.

https://claude.ai/code/session_017WB1TU6tyLmMa1n4SQ9XgK

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,6 +28,30 @@ workspace.
   construction too — useful when a caller's view of the chain tip used to
   derive `min_target_height` may lag the real tip by more than the default
   expiry buffer covers.
+- `zcash_client_backend::data_api::AccountBalance` now tracks transparent funds
+  in two separate buckets: regular (non-coinbase) funds and funds in coinbase
+  outputs, replacing its single internal unshielded balance:
+  - New accessors `AccountBalance::unshielded_regular_balance` and
+    `AccountBalance::unshielded_coinbase_balance`, and new mutators
+    `AccountBalance::with_unshielded_regular_balance_mut` and
+    `AccountBalance::with_unshielded_coinbase_balance_mut`. Transparent outputs
+    whose transaction index within their block is unknown are classified as
+    regular (non-coinbase) funds.
+  - `AccountBalance::unshielded_balance` now returns a `Balance` by value,
+    computed as the sum of the regular and coinbase buckets, instead of a
+    `&Balance` reference.
+  - Value in immature transparent coinbase outputs is now expected to be
+    reported in the `value_pending_spendability` field of the coinbase bucket
+    rather than as spendable value; it becomes spendable only once the output
+    reaches coinbase maturity.
+- `zcash_client_backend::data_api::wallet::ProposeSendMaxErrT` now uses
+  `GreedyInputSelectorError` (instead of `BalanceError`) as its note-selection
+  error type, so that `propose_send_max_transfer` can report
+  `GreedyInputSelectorError::UnsupportedTexAddress` — consistent with
+  `propose_transfer` — when a TEX recipient is requested and the
+  `transparent-inputs` feature is not enabled. Balance errors encountered
+  during send-max input selection are now wrapped in
+  `GreedyInputSelectorError::Balance`.
 - `zcash_client_backend::proposal::Step::orchard_action_count` now takes the
   `orchard::builder::BundleType` and `orchard::bundle::BundleVersion` that the
   transaction builder will be configured with, and returns a `Result`. It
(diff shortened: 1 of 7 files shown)


nuttycom (dismissed):
utACK, looks good.

nuttycom (dismissed):
re-utACK d8f38a7cd0983af511bde637b138d65146f936b2

nuttycom (approved):
re-utACK ee47ffb11562a60a9792a2ae18bf161be36d672e

nuttycom (dismissed):
re-utACK https://github.com/zcash/librustzcash/pull/2654/commits/19f065fd6a559245b2ee4db708b1565de0b8fc07

pacu:
The `Test NU7` / `Slow Test NU7` CI failures were a compile error in the new `fake_transparent_coinbase_tx` test helper: under `--cfg zcash_unstable="nu7"` with `zip-233` enabled (the NU7 jobs build `--all-features`), `TransactionData::from_parts` takes a ZIP 233 burn-amount argument the helper didn't pass.

Pushed `47344ee188` on top with the fix:
- the helper passes `Zatoshis::ZERO` under `#[cfg(all(zcash_unstable = "nu7", feature = "zip-233"))]`, matching the existing `zcash_client_sqlite` call sites;
- `zcash_client_backend` gains a `zip-233 = ["zcash_primitives/zip-233"]` feature to gate on (it previously had no `TransactionData` constructor callers, so the feature never existed there);
- `zcash_client_sqlite`'s `zip-233` now also enables `zcash_client_backend/zip-233`, so a sqlite-only `zip-233` build can't see the 9-argument signature while backend still makes the 8-argument call.

Verified locally with `RUSTFLAGS='--cfg zcash_unstable="nu7"' cargo test --all-features` over `zcash_client_backend`/`zcash_client_sqlite`/`pczt`/`zcash_primitives` (597 tests passing) on the previous base, and re-checked compilation after rebasing the fix onto the current branch head.

nuttycom (approved):
re-utACK 47344ee188da8412ccb5f934a3ab1922f2f1a835

PR #2656, opened 2026-07-17
zcash_pool_migration_backend: add note-preparation transaction planner

dannywillems:
Stacked on #2647. Adds the note-preparation transaction planner: how a wallet's spendable source-pool notes are restructured into the exact self-funding notes a migration run needs, as transactions that each obey the [ZIP 318](https://zips.z.cash/zip-0318) action budget.

## Problem

[ZIP 318] requires every note-preparation transaction to be padded to **exactly 16 Orchard actions** (a mobile-proving-time and on-chain-uniformity constraint). Under NU6.3 a bundle's action count is its spends plus its outputs, so one transaction can consume and produce at most 16 notes in total (e.g. 15 spends + 1 output when consolidating, or 1 spend + 15 outputs when splitting).

A single transaction therefore cannot always turn the wallet's notes into every self-funding note: a note that must fan out into more outputs than one transaction holds, or a balance spread across more dust notes than one transaction can consume, needs several transactions organised into **layers**. A layer is a set of transactions with no dependencies between them (built, proved, and broadcast in parallel); a later layer may spend an earlier layer's outputs, but only after they are mined and a boundary passes, so each extra layer extends the preparation phase by roughly one anchor bucket. (The prior `build::build_split_pczt` fans up to 50 outputs into one transaction, which this planner replaces the shape of.)

## Solution

A **pure** `preparation` module (no cryptography, no I/O, `no_std`):

```rust
pub fn plan_preparation(available: &[u64], funding: &[u64], fee_per_tx: u64)
    -> Result<PreparationPlan, PrepError>
```

- `available`: the wallet's spendable source-pool note values (zatoshi).
- `funding`: the target self-funding note values from the note-split plan (each already `crossing + buffer`).
- `fee_per_tx`: the ZIP-317 fee reserved per padded 16-action transaction; the builder later absorbs the real fee into the change.

It returns a `PreparationPlan` of layers of `PrepTransaction`s. Each input is a `PrepInput::Wallet(i)` (an original note) or a `PrepInput::Prior { layer, transaction, output }` (an earlier layer's output); each output is a `PrepOutput::Funding` (a final self-funding note), `Intermediate` (a feeder routed to a later layer), or `Change`.

**Largest-first layered greedy**: in each layer, feed each output transaction from the largest note it can (1 input + up to `FUNDING_OUTPUTS_PER_TX = 14` funding notes + a feeder/change slot = 16 actions), route a large remainder forward as a feeder note, and consolidate notes too small to fund anything (15 inputs -> 1 feeder). Anything below a whole funding note is left as source-pool change. A final pass relabels any feeder that no later layer ends up spending as change, so value is conserved end to end. For a typical wallet (a few notes, a handful of funding notes) this is a single layer.

## Structure

- `preparation.rs` — the constants (`PREP_TX_ACTIONS`, `FUNDING_OUTPUTS_PER_TX`, `CONSOLIDATION_INPUTS_PER_TX`), the `PrepInput` / `PrepOutput` / `PrepTransaction` / `PreparationPlan` / `PrepError` types (with accessors and an `input_value` resolver), and `plan_preparation`.

## Notes for reviewers

- **Pure and `no_std`.** No crypto or I/O; a later slice wires the `build` module to emit one PCZT per `PrepTransaction`, and the engine drives the layers (build/sign a layer together, then prove + broadcast it only after the prior layer is mined and a boundary passes).
- **Fees.** The planner reserves `fee_per_tx` out of each transaction's inputs (the caller passes `16 * MARGINAL_FEE`, the fee of a padded 16-action tx); the real fee is smaller and the builder absorbs it into change.
- **Known limitation (documented in the module doc).** The greedy is a heuristic, not a layer-count optimiser: for a lone, very large note fanning out into many funding notes it chains feeders linearly (one layer per 14 funding notes) rather than fanning the note into several parallel feeders first, so in that extreme it can use more layers than the minimum (e.g. ~4 where 2 would do for 50 notes from one note). The common cases are already minimal; the lone-whale fan-out is flagged as future work.

## Tests

A proptest suite drives the structural contract over arbitrary inputs: every transaction stays within the 16-action budget, value is conserved per transaction (`inputs = outputs + fee`) and globally (`wallet spent = funding + change + fees`), every requested funding note is minted exactly once, and every `Prior` input references a strictly earlier layer (acyclic). A second generator that always includes one note large enough to fund everything asserts the planner never spuriously fails when the value is present. Golden tests cover the one-layer common case, a whale single note fanning across layers, dust consolidated before it can fund, an insufficient-value `Err`, and the consolidation batching (never a lone note, never over budget).

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc -p zcash_pool_migration_backend --no-deps`
- `cargo fmt --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7261,6 +7261,9 @@ dependencies = [
 name = "zcash_pool_migration_backend"
 version = "0.1.0"
 dependencies = [
+ "incrementalmerkletree",
+ "orchard",
+ "pczt",
  "proptest",
  "rand_chacha 0.3.1",
  "rand_core 0.6.4",
--- a/zcash_pool_migration_backend/CHANGELOG.md
+++ b/zcash_pool_migration_backend/CHANGELOG.md
@@ -53,3 +53,15 @@ and this library adheres to Rust's notion of
   unsigned. It signs a finalized but still UNPROVEN PCZT: the spend-authorization signature is over
   the transaction's sighash, which is fixed independently of the zk proofs, so the migration signs up
   front (capturing the account's authorization) and proves later, at scheduling time.
+- Note-preparation transaction planning (the `preparation` module): a pure planner that partitions the
+  work of minting a note-split plan's self-funding notes into note-preparation transactions of exactly
+  `PREP_TX_ACTIONS` (16) Orchard actions, per ZIP 318, organised into the fewest sequential layers
+  (with parallel transactions inside a layer). `plan_preparation` takes the wallet's spendable
+  source-pool note values, the target funding-note values, and a per-transaction fee reserve, and
+  returns a `PreparationPlan` of `PrepTransaction`s (each a same-pool send-to-self referencing wallet
+  notes or earlier layers' outputs). It uses a largest-first greedy: feed each output transaction from
+  the largest note, route every leftover forward as a feeder note, and consolidate dust; once the
+  funding notes are scheduled it consolidates the leftover feeders into a single residual note, per
+  ZIP 318's "one note per part plus at most one residual note". A typical wallet needs one layer, and
+  extra layers appear only for a lone large note fanning out into many funding notes or a dust-heavy
+  balance. It is pure (no cryptography or I/O) and `no_std`.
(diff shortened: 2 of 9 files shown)


dannywillems:
Folded into #2647, which now contains the note-preparation planner and all the PCZT builders as one PR.

PR #2658, opened 2026-07-17
zcash_pool_migration_backend: add the note-preparation PCZT builder

dannywillems:
Stacked on #2656. Adds the note-preparation PCZT builder that turns the preparation planner's transactions into unproven PCZTs.

## Problem

[ZIP 318](https://zips.z.cash/zip-0318) requires each note-preparation transaction to be **exactly 16 Orchard actions** (padding if necessary), so no preparation transaction is distinguishable from another by its action count.

## Solution

`build::build_prep_tx` (behind the `orchard` feature): it takes one `preparation::PrepTransaction`'s **resolved** input notes (the wallet backend maps each `PrepInput` to its `(Note, MerklePath)` witness) and its `outputs`, and builds a same-pool Orchard send-to-self as an unproven `pczt::Pczt`.

- Every output (a `Funding`, `Intermediate` feeder, or `Change`/residual note) is a wallet-controlled **internal change note**.
- The Orchard bundle is built with `BundleType::Transactional { pad_to_minimum: Some(16), .. }`, so orchard fills it to **exactly 16 actions with fabricated dummies** (orchard 0.15.0 already ships `pad_to_minimum`, and the workspace already resolves that release, so this needs no dependency change).
- The spent notes must total the outputs plus the ZIP-317 fee of a 16-action transaction (which the planner reserves per transaction), or the build does not balance.
- It returns each requested output's real **post-shuffle action index** so the caller can locate the notes after mining and spend the feeders in a later layer. The dummy actions are not returned.

It reuses the shared build plumbing (`finalize_pczt`, `output_action_index`), is pure (no database/wallet-backend access), and takes the RNG as a parameter.

## Tests

A proptest builds a single-spend funding transaction fanned into up to 13 outputs and asserts the built bundle has **exactly 16 Orchard actions**, and that each output is reported at a distinct action index with its planned value. Plus: a minimal `1 spend + 1 output` transaction fills to 16; and the builder rejects an empty spend set, an empty output set, and an over-budget action count.

Known test gap (tracked): a multi-spend consolidation-shaped transaction (many spends + 1 output) needs a multi-note shared-anchor witness helper; the single-spend path is covered here.

## Verification

- `cargo fmt --check`

### DIFF
--- a/zcash_pool_migration_backend/CHANGELOG.md
+++ b/zcash_pool_migration_backend/CHANGELOG.md
@@ -65,3 +65,12 @@ and this library adheres to Rust's notion of
   ZIP 318's "one note per part plus at most one residual note". A typical wallet needs one layer, and
   extra layers appear only for a lone large note fanning out into many funding notes or a dust-heavy
   balance. It is pure (no cryptography or I/O) and `no_std`.
+- The pure PCZT note-preparation builder (`build::build_prep_tx`, behind the `orchard` feature): turns
+  one `preparation::PrepTransaction`'s resolved input notes and its outputs into an unproven
+  `pczt::Pczt` for a same-pool Orchard send-to-self. Every output (funding, feeder, or residual) is a
+  wallet-controlled internal change note, and the Orchard bundle is built with `pad_to_minimum` set to
+  `PREP_TX_ACTIONS` (16), so orchard fills it to exactly that many actions with fabricated dummies and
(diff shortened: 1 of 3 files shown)


dannywillems:
Folded into #2647, which now contains the note-preparation planner and all the PCZT builders as one PR (the obsolete single-transaction split builder was dropped).

PR #2659, opened 2026-07-17
zcash_client_backend: Add a backend-agnostic network-privacy layer

### LINKED ISSUE #2665: zcash_client_backend: Add a backend-agnostic network-privacy layer
Generalize zcash_client_backend::tor into a backend-agnostic network-privacy layer, so that the privacy transport used for lightwalletd point queries, HTTP requests, and exchange-rate fetches can be selected at runtime — Tor today, with alternative backends (e.g. the Nym mixnet) pluggable without further changes to this crate.

At the same time, we should extract the duplicated functionality from zcash-android-wallet-sdk and zcash-swift-wallet-sdk.

nuttycom:
This PR generalizes `zcash_client_backend::tor` into a backend-agnostic network-privacy layer, so that the privacy transport used for lightwalletd point queries, HTTP requests, and exchange-rate fetches can be selected at runtime — Tor today, with alternative backends (e.g. the Nym mixnet) pluggable without further changes to this crate.

## Motivation

Both mobile SDKs (zcash-swift-wallet-sdk and zcash-android-wallet-sdk) carry near-identical ~290-line hand-rolled blocking wrappers (`TorRuntime` + `LwdConn`) over `zcash_client_backend::tor`, and both already select their transport at runtime above the FFI (`ServiceMode`). Meanwhile, the only Tor-specific capability the gRPC and HTTP machinery in this crate actually uses is "open a byte stream to `host:port`" — a single `connect()` call in each of `tor/grpc.rs` and `tor/http.rs`. This PR makes that seam explicit and moves the duplicated blocking wrapper here so both SDKs can share it.

## What's included

- **`zcash_client_backend::privacy`** (new `lightwalletd-privacy` feature):
  - `PrivateNetwork` — the core trait: `connect(&self, host, port) -> impl Future<Output = Result<Stream, Error>>` (RPITIT), `isolated_handle()`, `set_dormant(DormantMode)`.
  - `DynPrivateNetwork` / `BoxedStream` — an object-safe mirror with a blanket impl, so FFI layers can hold `Arc<dyn DynPrivateNetwork>`; the `Arc<dyn>` form itself implements `PrivateNetwork`.
  - A crate-owned `DormantMode` (re-exported at `tor::DormantMode`; `From` conversions with `arti_client::DormantMode` under the `tor` feature).
- **`privacy::grpc` / `privacy::http` / `privacy::http::cryptex`** — the existing gRPC-over-Tor, HTTP-over-Tor, and exchange-rate modules, moved and parameterized over `PrivateNetwork`. The old `tor::http::*` paths (`HttpError`, `Retry`, `cryptex`) remain as re-exports, and `tor::Client`'s methods delegate, so the module's public API stays source-compatible.
- **`privacy::blocking`** (new `lightwalletd-blocking` feature, independent of `tor`): `PrivacyRuntime` (owns a Tokio runtime + an erased backend; `create_tor(tor_dir, dangerously_trust_everyone)` convenience under the `tor` feature) and `LwdConn` (a blocking lightwalletd client carrying the union of the two mobile SDKs' wrapper methods). This is the deduplication target: both SDKs' FFI layers delegate to these types unchanged in their corresponding integration branches.
- `tor::Client` implements `PrivateNetwork`; `tor::Error` gains a `Backend` variant. `HttpError::Spawn` is removed (connection tasks now spawn on the ambient Tokio runtime).

## Verification

- `cargo test -p zcash_client_backend --all-features` passes; two new unit tests exercise the erased layer end-to-end over in-memory duplex streams (HTTP round-trip through both a concrete and an `Arc<dyn>` backend, gRPC helper instantiation over the erased form).
- Feature independence: `--no-default-features` plus each of `tor`, `lightwalletd-blocking`, `tor,lightwalletd-blocking`, `lightwalletd-privacy` compile.
- `cargo check --workspace --all-features` clean; clippy clean and rustfmt applied on touched code.

## Downstream consumers

Integration branches exist for both mobile SDKs (stacked on their `maint/ironwood-2.5.x` branches) that delete the hand-rolled wrappers and delegate the existing Tor FFI symbols to `privacy::blocking` with unchanged FFI surfaces, pinning the librustzcash family to this branch's tip via `[patch.crates-io]` until it lands in a published release (zcash/zcash-swift-wallet-sdk#1809, zcash/zcash-android-wallet-sdk#2014). A stacked draft PR (#2660) adds experimental Nym backends implementing `PrivateNetwork`, useful as a cross-reference for reviewing the abstraction.

Resolves MOB-1492 and MOB-1491

---

### DIFF
--- a/zcash_client_backend/src/privacy.rs
+++ b/zcash_client_backend/src/privacy.rs
@@ -1,0 +1,215 @@
+//! Backend-agnostic network privacy for lightwallet clients.
+//!
+//! Wallets frequently need to hide the network-level metadata (in particular their IP
+//! address) that they would otherwise reveal to a `lightwalletd` server or to an exchange
+//! rate provider. This module abstracts over the transport that provides this privacy so
+//! that the same gRPC, HTTP, and exchange-rate machinery can run over any such transport.
+//!
+//! The central abstraction is the [`PrivateNetwork`] trait, which exposes the single
+//! capability the higher layers require: opening a byte stream to a remote `host:port`.
+//! The [`crate::tor`] module provides the reference implementation (Tor, via `arti`);
+//! other backends (for example a dVPN tunnel or a mixnet proxy) can be implemented in
+//! separate crates without depending on `arti`.
+//!
+//! Because [`PrivateNetwork`] uses return-position `impl Trait` and an associated stream
+//! type, it is not object-safe. FFI layers that need to store a backend behind a pointer
+//! can instead use the object-safe [`DynPrivateNetwork`] mirror, for which a blanket
+//! implementation is provided for every [`PrivateNetwork`]. `Arc<dyn DynPrivateNetwork>`
+//! itself implements [`PrivateNetwork`], so the generic helpers in this module accept
+//! either a concrete backend or its erased form.
+
+use std::{fmt, future::Future, pin::Pin, sync::Arc};
+
+use tokio::io::{AsyncRead, AsyncWrite};
+
+/// The level of background activity a [`PrivateNetwork`] should maintain.
+///
+/// This is a crate-owned generalization of `arti_client::DormantMode`; the [`crate::tor`]
+/// backend maps between the two. Backends without a native notion of dormancy may treat a
+/// [`DormantMode::Soft`] request as a hint to tear down idle resources and rebuild them
+/// lazily on next use.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum DormantMode {
+    /// The backend functions as normal, and background tasks run periodically.
+    #[default]
+    Normal,
+    /// Background tasks are suspended, conserving CPU usage. Attempts to use the backend
+    /// will wake it back up again.
+    Soft,
+}
+
+#[cfg(feature = "tor")]
+impl From<DormantMode> for arti_client::DormantMode {
+    fn from(mode: DormantMode) -> Self {
+        match mode {
+            DormantMode::Normal => arti_client::DormantMode::Normal,
+            DormantMode::Soft => arti_client::DormantMode::Soft,
+        }
+    }
+}
+
+#[cfg(feature = "tor")]
+impl From<arti_client::DormantMode> for DormantMode {
+    fn from(mode: arti_client::DormantMode) -> Self {
+        match mode {
+            arti_client::DormantMode::Soft => DormantMode::Soft,
+            // `arti_client::DormantMode` is `#[non_exhaustive]`; treat anything we don't
+            // specifically recognise as the fully-awake mode.
+            _ => DormantMode::Normal,
+        }
+    }
+}
+
+/// A byte stream produced by a [`PrivateNetwork`], type-erased for use behind a pointer.
+///
+/// This is the stream type carried by [`DynPrivateNetwork`] and by the blanket
+/// `PrivateNetwork` implementation for `Arc<dyn DynPrivateNetwork>`.
+pub trait NetworkStream: AsyncRead + AsyncWrite + Unpin + Send {}
+impl<T: AsyncRead + AsyncWrite + Unpin + Send> NetworkStream for T {}
+
+/// A boxed [`NetworkStream`].
+pub type BoxedStream = Box<dyn NetworkStream + 'static>;
+
+/// A network transport that provides connection-level privacy for wallet traffic.
+///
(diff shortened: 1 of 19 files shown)


zcash_client_backend/src/privacy/blocking.rs
```diff
@@ -0,0 +1,617 @@
...
+
+    /// Returns a new runtime handle whose backend is isolated from this one's, sharing the
+    /// same Tokio runtime.
+    ///
+    /// See [`PrivateNetwork::isolated_handle`] for the isolation semantics.
+    #[must_use]
+    pub fn isolated_client(&self) -> Self {
+        Self {
+            runtime: self.runtime.clone(),
+            net: self.net.dyn_isolated_handle(),
```
zcash-harry: **zcash_client_backend/src/privacy/blocking.rs:100,114**

`self.net.dyn_isolated_handle() `doesn't dispatch to the concrete backend. Rust resolves it to the blanket impl on `Arc<dyn DynPrivateNetwork>` itself, so it wraps the current handle in another Arc layer instead of returning a fresh one. 

Chaining `isolated_client()` calls grows this unboundedly (5 calls → 6 layers, O(N²) cost). 

Fix: use `self.net.isolated_handle() `instead — same file, already correct.

zcash_client_backend/src/privacy/http.rs
```diff
@@ -0,0 +1,392 @@
...
+{
+    debug!("Making request");
+    let (mut sender, connection) = conn::http1::handshake(TokioIo::new(stream))
+        .await
+        .map_err(HttpError::from)?;
+
+    // Spawn a task to poll the connection and drive the HTTP state. Every consumer of
+    // this crate's privacy layer runs on a Tokio runtime (`arti` uses one internally,
+    // and the blocking facade owns one), so a Tokio task is always available here.
+    tokio::spawn(async move {
```
zcash-harry: `tokio::spawn(...)` here is unconditional now, replacing the old fallible `PreferredRuntime::current()?.spawn(...) (→ HttpError::Spawn)`. 

`tokio::spawn` panics with no ambient Tokio runtime instead of returning Err. Since this is now public API meant for non-Tor backends, a caller off the Tokio runtime will panic instead of getting a recoverable error.

zcash_client_backend/src/privacy/blocking.rs
nuttycom: Fixed as suggested at both sites (folded into the facade commit, branch force-pushed with lease). One nuance for the record: the concrete backend's `isolated_handle` *was* still being reached — the blanket impl on the `Arc` calls `<Arc<dyn _> as PrivateNetwork>::isolated_handle`, which dispatches through `&**self` to the inner trait object — so isolation semantics held. But the extra `Arc` layer per call (and the resulting unbounded growth/indirection when chaining) was real, and `self.net.isolated_handle()` is the right call. Also added a doc warning on `DynPrivateNetwork` about this method-resolution footgun (the blanket impl applying to `Arc<dyn DynPrivateNetwork>` itself) so future call sites don't reintroduce it.

*(Reply drafted with Claude Code assistance.)*

zcash_client_backend/src/privacy/http.rs
nuttycom: Fixed (folded into the gRPC/HTTP generalization commit, branch force-pushed with lease): the connection task is now spawned via `tokio::runtime::Handle::try_current()`, and a missing ambient runtime surfaces as a new recoverable `HttpError::Runtime(TryCurrentError)` variant instead of a panic. The changelog's `HttpError::Spawn`-removal entry now points at the replacement variant.

*(Reply drafted with Claude Code assistance.)*

zcash-harry (dismissed):
utACK

zcash_client_backend/src/privacy.rs
```diff
@@ -0,0 +1,243 @@
...
+    fn from(mode: DormantMode) -> Self {
+        match mode {
+            DormantMode::Normal => arti_client::DormantMode::Normal,
+            DormantMode::Soft => arti_client::DormantMode::Soft,
+        }
+    }
+}
+
+#[cfg(feature = "tor")]
+impl From<arti_client::DormantMode> for DormantMode {
```
pacu, line 59: Can't this be under the tor.rs file so it's all feature gated by module?

pacu (dismissed):
non blocking nits

nuttycom:
Draft until after the final zcash_client_backend-0.24.0 release

PR #2660, opened 2026-07-17
zcash_client_privacy_nym: Add experimental Nym network-privacy backends

### LINKED ISSUE #2664: Add Nym support for lightwalletd server connections.
At present, we support using Tor for lightwalletd connections; we should generalize and extend this support to permit use of a Nym backend, extracting common functionality from zcash-swift-wallet-sdk and zcash-android-wallet-sdk.

nuttycom:
**Draft, stacked on #2659.** This PR adds experimental Nym-network backends implementing the `PrivateNetwork` trait introduced there. It exists primarily as a cross-reference while reviewing the base PR's abstraction: two backends with very different characteristics (fixed-endpoint Sphinx mixnet; arbitrary-host WireGuard dVPN) both fit the trait without changes to it.

Both crates are **excluded from the cargo workspace** (own lockfiles) so their large dependency trees and differing MSRV never affect the main workspace or CI, and are `publish = false`.

## `zcash_client_privacy_nym` — Sphinx mixnet, proxy mode (working)

`MixnetProxyNetwork` sends traffic through the full Nym mixnet (Sphinx mixing + cover traffic) using `nym-sdk` from crates.io. The mixnet is client-to-client, so every reachable endpoint must be fronted by a proxy listener (included as `proxy_listener`, run adjacent to the lightwalletd it serves); the backend is configured with a routing table mapping `(host, port)` to the proxy's Nym `Recipient`, and unconfigured endpoints fail fast with `Error::NoRoute`. Isolation = a fresh ephemeral mixnet identity per isolated handle; dormancy = teardown + lazy rebuild (Nym clients cannot reconnect). No credentials are required in the current free tier.

Verified live against the mainnet mixnet (test committed as `#[ignore]`): in-process proxy listener + wallet client, full round-trip through the real mixnet in ~11s (client bootstrap ~4s, stream open ~7s).

## `zcash_client_privacy_nym_dvpn` — 1/2-hop WireGuard dVPN (build-blocked, committed as an artifact)

`DvpnNetwork` wraps `nym-smol-dvpn` (git-pinned; unpublished until nymtech releases `nym-bridges`): a userspace WireGuard datapath over Nym gateways exposing `tcp_connect(host, port)` to arbitrary hosts — no proxy infrastructure, and it handles bulk throughput (Nym's own example streams compact blocks from a public lightwalletd through it). Registration requires zk-nym ticketbooks from a funded NYX account; provisioning is a separate `provision` module so no credential material touches the datapath. Privacy semantics are documented honestly: multi-hop VPN, **not** mixnet (no mixing/cover traffic).

The crate currently **does not resolve**: `zcash_primitives` pins `crypto-common = "=0.2.0-rc.1"` while the Nym credential stack requires `>=0.2.0-rc.5` (via `jwt-simple`/`superboring`; the full version matrix has no escape). It unblocks when the workspace advances `crypto-common` past rc.1 (edition-2024 migration) or Nym relaxes its credential-crypto floor. Details in the crate README.

## Open questions

- Crate naming (`zcash_client_privacy_nym{,_dvpn}` vs lightwalletd-scoped alternatives).
- Whether the build-blocked dvpn crate should stay in-tree as a documented artifact or wait out the `crypto-common` conflict.

---


zcash_client_privacy_nym_dvpn/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+    "lightwalletd-privacy",
+    "lightwalletd-tonic-tls-webpki-roots",
+] }
+
+tokio = { version = "1", features = ["rt", "net", "io-util", "sync", "macros"] }
+tokio-util = "0.7"
+tracing = "0.1"
+
+# The Nym `smol-dvpn` datapath and its (funded) registration path, pinned to the monorepo
+# revision studied when this crate was written. Both are `git`-only today.
```
mmsinclair, line 41: there are two blockers here:

- we are actively working on getting https://github.com/nymtech/nym-bridges published to crates.io
- and https://github.com/nymtech/nym/pull/6953 will remain draft until this is done, and we need to make a minor change to our license restrictions to allow the AGPL license for `nym-bridges`

zcash_client_privacy_nym_dvpn/CHANGELOG.md
```diff
@@ -0,0 +1,32 @@
...
+    `nym-sdk-session` `Session` into a `DvpnConfig`.
+
+### Known issues
+
+- **This crate does not currently build.** Its Nym `git` credential stack requires
+  `crypto-common >= 0.2.0-rc.5`, while `zcash_client_backend` pins
+  `crypto-common = "=0.2.0-rc.1"`; `cargo` cannot resolve a lockfile satisfying both. See
+  `README.md` for the full analysis and the conditions that would unblock it. The source is
+  retained as a complete, reviewed artifact against the Nym monorepo revision
+  `1fd9ae8817b2c6e49283e8e5597f031ce7f6091c`.
```
mmsinclair, line 29: in `librustzcash` where else do you use Nym's `crypto-common`?

once `nym-smol-dvpn` is on crates.io hopefully this will be resolved by using the correctly matched versions

zcash_client_privacy_nym_dvpn/README.md
```diff
@@ -0,0 +1,103 @@
...
+3. an upstream Nym release that removes the `jwt-simple`/`superboring` credential path from
+   the `smol-dvpn` datapath.
+
+Until then this crate is retained as a complete, reviewed artifact. Its `Cargo.toml` pins
+the studied Nym monorepo revision (`1fd9ae8817b2c6e49283e8e5597f031ce7f6091c`).
+
+## MSRV
+
+Like its sibling, this crate would require **Rust >= 1.89** once buildable (the Nym tree
+pulls `libcrux-psq 0.0.8`, which does not compile on 1.88).
```
mmsinclair, line 94: what is the minimum version of Rust you can build with?

we could experiment with seeing if we can set a lower minimum version (might not be possible, depending on the constraints)

zcash_client_privacy_nym/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+zcash_client_backend = { path = "../zcash_client_backend", default-features = false, features = [
+    "lightwalletd-privacy",
+    "lightwalletd-tonic-tls-webpki-roots",
+] }
+
+tokio = { version = "1", features = ["rt", "net", "io-util", "sync", "macros"] }
+tracing = "0.1"
+
+# mixnet-proxy backend (crates.io).
+nym-sdk = { version = "1.21", optional = true }
```
mmsinclair, line 45: we have a fix inbound soon that will correct an MTU mismatch issue with the IPR - please use the next package version (will let you know when it is ready) because otherwise it will be unstable for users when the requests are bigger than the IPR's tunnel MTU

mmsinclair (commented):
Very nicely done @nuttycom - we are working on getting missing crates onto crates.io. Will let you know when they are there to use.

zcash_client_privacy_nym_dvpn/Cargo.toml
mmsinclair: nym-bridges is now on https://crates.io/crates/nym-bridges

zcash_client_privacy_nym_dvpn/CHANGELOG.md
mfahampshire: @nuttycom this will go out today - I will drop a comment on here when it is

zcash_client_privacy_nym/Cargo.toml
mfahampshire: @nuttycom ditto with this - I will drop the exact version once its released (it'll be an `-rc.x` until the next platform release)

nuttycom: Awesome, thank you!

zcash_client_privacy_nym_dvpn/README.md
nuttycom: We can bump to 1.89 in librustzcash. We're generally pretty conservative in our MSRV upgrades, but this is a small enough bump that it would be fine.

zcash_client_privacy_nym_dvpn/CHANGELOG.md
mfahampshire: Its out - please point at https://crates.io/crates/nym-smoldvpn - note the rename from `smoldvpn` to `nym-smoldvpn`; I will be yanking the `-rc` for this crate as this naming follows our naming conventions.

mfahampshire: Let me know if you still need to pull `crypto-common` and I will look into it, if the `nym-smoldvpn` import doesn't fix it

zcash_client_privacy_nym/Cargo.toml
mfahampshire: https://crates.io/crates/nym-sdk/1.21.5-rc.3 this `-rc.3` version has the fix

emersonian:
Bandwidth use as configured is worth documenting somewhere, and encouraging an end-user notice.

With the SDK defaults this backend runs the main Poisson stream at a 20 ms average (plus 200 ms loop cover), so a connected client pushes a constant ~1 Mbps / ~430 MB per hour upstream whether or not the wallet is doing anything. That's the correct mixnet behavior, but set_dormant is the only thing that stops it, and a mobile integrator won't learn that from the current docs.

Poisson stream settings should be configurable by end users in some way, perhaps a "Less bandwidth" vs "Safest" slider similar to what Tor Browser offers.

That said, I'm excited for Nym!

Also, Zec.rocks operates a few experimental Nym Service Providers pointed at lightwalletd that we're happy to circulate for testing.

PR #2661, opened 2026-07-17
zcash_pool_migration_backend: transfer scheduling and anchor selection

dannywillems:
Stacked on #2647. This is Phase 2's transfer SCHEDULING and anchor-selection layer: a pure `scheduling` module implementing the ZIP 318 "Transfer scheduling" strong requirements. It works only in block heights and part indices (no cryptography, no note tree, no I/O) and takes the RNG as a parameter (a CSPRNG in production).

## What it adds

- Shuffle (MUST): a uniform Fisher-Yates shuffle of the parts, so the broadcast order of denominations is independent of the balance.
- Delays (MUST): a truncated exponential inter-arrival delay (mean `MEAN_DELAY` = 144 blocks, discard-and-redraw above `MAX_DELAY` = 576) sampled by inverse-CDF, and cumulative per-part broadcast heights from the commit height.
- Boundary anchor at proving time (MUST): a recency-weighted Geometric(1/2) draw over the candidate boundaries (above NU6.3 activation, at or after the funding note's creation, strictly below the most recent boundary), capped at `ANCHOR_AGE_CAP` = 16. The most recent boundary is never used (modal age 1, mean 2), so transfers cluster into cohorts. It returns the chosen boundary height; the wallet backend resolves the anchor tree-state.
- Cohorts + K_MAX (SHOULD, still an open ZIP issue): grouping a wallet's own transfers by boundary and a provisional per-wallet multiplicity cap.
- Expiry (MUST): the canonical rolling window (`EXPIRY_MODULUS` = 34560, window `2 * EXPIRY_MODULUS`), a pure function of the height, giving 1 to 2 months of validity.

Every numeric parameter is a named `const`. All the MUST invariants are property-tested (delays bounded; broadcast heights non-decreasing and at least the commit height; anchor age in `[1, ANCHOR_AGE_CAP]` with the boundary in the candidate set and strictly below the most recent; expiry in the correct rolling window; shuffle is a permutation).

## Notes for reviewers

- The crate is `no_std`, so the exponential draw's natural log uses `libm` (declared at the workspace level).
- Engine-enforced MUSTs are out of scope for this pure module and documented as such: a background window must be used EITHER to synchronize OR to broadcast (never both), and at most one overdue transfer may be sent at wallet open.
- Open questions for the ZIP authors: (1) `K_MAX`'s value and the fallback when a wallet's candidate set is saturated; (2) whether `most_recent_boundary` should exclude very recent (reorg-risk) boundaries; (3) delay-rounding tie-breaking, and whether a zero-block delay (two transfers at one height) is intended.

## Tests / verification

- `cargo test -p zcash_pool_migration_backend` and `--features orchard`: 58 / 65 green.
- `cargo clippy -p zcash_pool_migration_backend --all-targets` (both feature sets), `cargo fmt --check`, and `RUSTDOCFLAGS="-D warnings" cargo doc -p zcash_pool_migration_backend --no-deps`.

### DIFF
--- a/zcash_pool_migration_backend/src/scheduling.rs
+++ b/zcash_pool_migration_backend/src/scheduling.rs
@@ -1,0 +1,865 @@
+//! Transfer scheduling and anchor selection: WHEN each migration transfer is broadcast, WHICH
+//! Orchard anchor it proves against, and WHEN it expires.
+//!
+//! This module is pool-agnostic and pure arithmetic plus RNG draws (no cryptography, no note tree,
+//! no I/O); it works only in block heights and part indices. All randomness comes from a caller-
+//! supplied `rng` (a CSPRNG in production; the tests pass a seeded `rand_chacha::ChaCha8Rng`).
+//! Every function is deterministic given its `rng`.
+//!
+//! # The problem
+//!
+//! The note split ([`crate::note_splitting`]) decides the crossing VALUES (quantized denominations)
+//! so many wallets emit colliding amounts. Scheduling decides the TEMPORAL and ANCHOR dimensions of
+//! the same crossings, which leak just as much if left predictable:
+//!
+//! - If the sequence of denominations were a fixed function of the balance, a chain observer could
+//!   read the balance back out of the order the transfers appear in.
+//! - If every transfer proved against the latest tree state, the anchor would timestamp the transfer
+//!   and shrink its anonymity set to the wallets active in that exact block.
+//!
+//! [ZIP 318] settles this with randomization drawn from principled distributions, which this module
+//! implements ("Transfer scheduling" and the anchor-selection rules):
+//!
+//! 1. SHUFFLE ([`shuffle_indices`]): the quantized parts are broadcast in a uniformly random order,
+//!    so the temporal sequence of denominations is independent of the balance.
+//! 2. DELAYS ([`draw_delay`]): the gap between successive transfers is an exponential inter-arrival
+//!    time (mean [`MEAN_DELAY`], capped at [`MAX_DELAY`]), so broadcasts look like an unremarkable
+//!    Poisson process rather than a burst.
+//! 3. CUMULATIVE ([`schedule_broadcast_heights`]): each transfer's scheduled height is the running
+//!    sum of independent delays from the commit height.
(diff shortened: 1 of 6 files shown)


zcash_pool_migration_backend/src/scheduling.rs
```diff
@@ -0,0 +1,806 @@
...
+    // Standard downward Fisher-Yates: for i from len-1 down to 1, swap i with a uniform j in [0, i].
+    let mut i = len - 1;
+    while i > 0 {
+        let j = gen_index(rng, i + 1);
+        slice.swap(i, j);
+        i -= 1;
+    }
+}
+
+/// Draw a uniform integer in `[0, bound)` from `rng` (`bound > 0`) using Lemire's unbiased
```
dannywillems, line 218: Avoid division.

dannywillems:
Self-ACK (additional tests - proptest and golden test vectors - are being added by an agent. AGENTS.md will be also updated to take into account some new engineering principles I do believe are good to be respected.

dannywillems (commented):
Self-tACK 9b644c3661b956cc309248a166f00a0ec67086eb

zcash_pool_migration_backend/src/scheduling.rs
```diff
@@ -0,0 +1,1223 @@
...
+pub fn draw_anchor_boundary<R: RngCore>(
+    nu63_activation: u32,
+    funding_creation_height: u32,
+    most_recent_boundary: u32,
+    rng: &mut R,
+) -> Option<u32> {
+    let (lowest, highest) = candidate_boundary_bounds(
+        nu63_activation,
+        funding_creation_height,
+        most_recent_boundary,
```
nuttycom: `most_recent_boundary` is not checked to be a multiple of `BOUNDARY_MODULUS` and this public function's semantics will break down if this invariant is violated. This needs to be explicitly checked.

nuttycom: We should pass the chain tip height instead of the boundary, and determine the boundary height ourselves. 

Also, in general, we should use the `BlockHeight` type from `zcash_protocol` (and also `Zatoshis`) instead of bare integers in public APIs.

nuttycom (approved):
utACK 0193d5aa20dc6c5fea8dad10e46e23956d9392b7

PR #2662, opened 2026-07-17
Add PCZT batch signer redaction helper

czarcas7ic:
Adds a wallet helper for PCZTs sent through the signatures only batch protocol. It layers on general signer redaction, removes Orchard and Ironwood full viewing keys and existing signatures, and removes alpha only from actions already authorized in the retained authoritative PCZT while preserving unsigned zero value wallet spends for signing. This centralizes the capability policy without weakening the general signer path.

It also renames the Orchard protocol redactor operation to `compact_resolvable_fields` so the API reflects that it replaces decryptable ciphertext with memo plaintext as well as omitting fields.

### DIFF
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -11,8 +11,8 @@ workspace.
 ## [Unreleased]
 
 ### Added
-- `pczt::roles::redactor::orchard::OrchardRedactor::redact_recomputable_fields`,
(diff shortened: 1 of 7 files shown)


nuttycom (approved):
utACK

nullcopy (approved):
utACK 93b5cce0522429358353f8892b14a38a72591745

PR #2663, opened 2026-07-17
zcash_pool_migration_backend: the migration engine (planning, state, commit preparation)

dannywillems:
Stacked on #2695 (`feat/deferred-shielded-anchors`, which depends on zcash/orchard#534). Part of the Orchard → Ironwood pool migration (ZIP 318). Merge zcash/orchard#534 → #2695 first.

Adds the `engine` module: the orchestration layer that ties the note-split, preparation, and scheduling planners together behind a `MigrationBackend` trait, so the engine drives a migration without knowing how the wallet stores notes, holds keys, or persists state.

## What's here
- `plan_migration` (pure, `no_std`): reads the account's spendable notes and the chain tip from the backend, decomposes the balance into denominations, plans the preparation transactions, schedules the transfers — broadcast order SHUFFLED per ZIP 318, so the on-chain sequence of denominations is independent of the balance — and reconciles the split against the preparation fees (drop-smallest when they do not fit), returning a `MigrationPlan` preview for user consent.
- The persisted state model: a `MigrationState` of `MigrationTransaction`s, each storing its pre-signed PCZT plus schedule, dependencies, drawn anchor boundary, and lifecycle state, behind `PoolMigrationRead`/`PoolMigrationWrite` store traits. The durable artifact is the pre-signed PCZT, so a wallet resumes a migration entirely from the store after a restart — and a live (non-terminal) migration is never overwritten by a re-commit.
- **ONE signing phase** (`commit_preparation`): every transaction — each preparation layer in topological order, then every transfer — is built and signed in a single pass, before anything is broadcast. This rests on ZIP 374: the V6 txid and sighash exclude shielded anchors, so every PCZT is built with ABSENT anchors and witnesses (via #2695's `DeferredPcztBuilder`), and a spent note's plaintext fully determines the signed data — including a note minted by a still-unmined earlier migration transaction, whose plaintext the engine recovers from the built bundle (its rho is the paired spend's nullifier and its rseed is drawn at build; mining only assigns the tree position, which matters only to the proof). Mining therefore gates only the BROADCAST order, which the state machine walks by dependencies and the drawn schedule; the consumer proves each transaction — installing its anchor (for a transfer, its drawn boundary) and witnesses through the PCZT `Updater` role — just before broadcasting it.
- The external-signer seam: `build_preparation_unsigned` returns the same transactions UNSIGNED, in topological order, and `batch_unsigned_by_action_budget` splits them into signing sessions bounded by the device's per-interaction action budget (16 actions per preparation, 3 per transfer) — consecutive prefixes, never gated on mining. `MigrationState::apply_signature` applies each signed PCZT back.
- A wallet-backed adapter (`wallet` module, behind the `wallet` feature): `WalletMigration` implements the backend/crypto/store traits over any `zcash_client_backend` wallet (`WalletRead` + `InputSource` only — the migration never touches the note commitment tree) plus the account's `UnifiedSpendingKey` and a caller-supplied store.

## Follow-ups (not in this PR)
- The proving/broadcast driver: install each due transaction's anchor and witnesses via the PCZT `Updater`, prove, and broadcast at the drawn heights.
- Reconciliation-on-launch (at most one overdue transfer released on reopen) and rebuild-on-expiry (rebuilding a preparation redraws its outputs' rseeds, so its dependents rebuild with it).
- A concrete store (`zcash_pool_migration_sqlite`).

Planning is pure (`no_std`); building and signing sit behind the `orchard` feature.

Description updated with Claude Code assistance.

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -46,9 +46,9 @@ use crate::preparation::{PrepError, PreparationPlan, plan_preparation};
 use crate::scheduling::{self, Schedule};
 
 /// What the migration engine needs from a wallet to PLAN a migration: the account's spendable notes and
-/// the chain state. Later slices add the methods for building (note witnesses, viewing keys), signing,
-/// persistence, and reconciliation; a backend implements this trait over its own note store and chain
-/// view (for example a `zcash_client_backend` wallet, or the migration's own SQLite store).
+/// the chain state. Following the `zcash_client_backend` pattern, a later slice replaces this with the
+/// wallet's own note-source and chain-view traits (`WalletRead` / `InputSource`), so any such wallet is a
+/// migration wallet; for now a backend implements it directly over its note store and chain view.
 pub trait MigrationBackend {
     /// The backend's own error type (a store or chain-access failure).
     type Error;
@@ -60,14 +60,27 @@ pub trait MigrationBackend {
 
     /// The current chain-tip height, from which the transfer schedule's delays accumulate.
     fn chain_tip_height(&self) -> Result<BlockHeight, Self::Error>;
+}
 
-    /// Persist the migration state: every transaction as its pre-signed PCZT plus the metadata the
-    /// application needs to prove, schedule, and broadcast it. Storing the pre-signed transactions,
-    /// not just the plan, is what lets a wallet resume a migration after being closed or restarted.
-    fn store_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error>;
+/// Read access to a persisted pool migration: the store side of the migration interface, mirroring
+/// `zcash_client_backend`'s `WalletRead`. A store implements this over its own tables (the
+/// `zcash_pool_migration_sqlite` crate does so as a migration registered into `zcash_client_sqlite`'s
+/// `WalletDb`). The committed migration is a set of pre-signed PCZTs plus their schedule and lifecycle
+/// state, so a wallet resumes a migration entirely from the store after being closed or restarted.
+pub trait PoolMigrationRead {
+    /// The store's own error type.
+    type Error;
 
-    /// Load the persisted migration state, if a migration is in progress.
-    fn load_migration(&self) -> Result<Option<MigrationState>, Self::Error>;
+    /// The migration currently in progress, if any.
+    fn get_migration(&self) -> Result<Option<MigrationState>, Self::Error>;
+}
+
+/// Write access to a persisted pool migration, mirroring `zcash_client_backend`'s `WalletWrite`.
+pub trait PoolMigrationWrite: PoolMigrationRead {
+    /// Persist a committed migration: every transaction as its pre-signed PCZT plus the metadata the
+    /// application needs to prove, schedule, and broadcast it. Storing the pre-signed transactions, not
+    /// just the plan, is what lets a wallet resume a migration after being closed or restarted.
+    fn put_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error>;
 
     /// Advance one stored transaction's lifecycle state (for example after the application broadcasts
     /// it, or the chain mines it, or it expires).
@@ -420,7 +433,10 @@ pub fn commit_preparation<P, B, R>(
 ) -> Result<MigrationState, CommitError<<B as MigrationBackend>::Error>>
 where
     P: zcash_protocol::consensus::Parameters + Clone,
-    B: MigrationBackend + MigrationCrypto<Error = <B as MigrationBackend>::Error>,
+    B: MigrationBackend
+        + MigrationCrypto<Error = <B as MigrationBackend>::Error>
+        + PoolMigrationRead<Error = <B as MigrationBackend>::Error>
+        + PoolMigrationWrite,
     R: RngCore + rand_core::CryptoRng,
 {
     use crate::build::build_prep_tx;
@@ -508,7 +524,7 @@ where
         transactions,
     };
     backend
-        .store_migration(&state)
+        .put_migration(&state)
         .map_err(CommitError::Backend)?;
     Ok(state)
 }
@@ -531,13 +547,16 @@ pub fn commit_transfers<P, B, R>(
 ) -> Result<MigrationState, CommitError<<B as MigrationBackend>::Error>>
 where
     P: zcash_protocol::consensus::Parameters + Clone,
-    B: MigrationBackend + MigrationCrypto<Error = <B as MigrationBackend>::Error>,
+    B: MigrationBackend
+        + MigrationCrypto<Error = <B as MigrationBackend>::Error>
+        + PoolMigrationRead<Error = <B as MigrationBackend>::Error>
+        + PoolMigrationWrite,
     R: RngCore + rand_core::CryptoRng,
 {
     use crate::build::build_transfer_pczt;
 
     let mut state = backend
-        .load_migration()
+        .get_migration()
         .map_err(CommitError::Backend)?
         .ok_or(CommitError::NoMigrationInProgress)?;
 
@@ -596,7 +615,7 @@ where
     }
 
     backend
-        .store_migration(&state)
+        .put_migration(&state)
         .map_err(CommitError::Backend)?;
     Ok(state)
 }
@@ -647,14 +666,20 @@ mod tests {
         fn chain_tip_height(&self) -> Result<BlockHeight, Self::Error> {
             Ok(self.tip)
         }
+    }
 
-        fn store_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error> {
+    impl PoolMigrationRead for MockBackend {
+        type Error = core::convert::Infallible;
+
+        fn get_migration(&self) -> Result<Option<MigrationState>, Self::Error> {
+            Ok(self.stored.clone())
+        }
+    }
+
+    impl PoolMigrationWrite for MockBackend {
+        fn put_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error> {
             self.stored = Some(state.clone());
             Ok(())
-        }
-
-        fn load_migration(&self) -> Result<Option<MigrationState>, Self::Error> {
-            Ok(self.stored.clone())
         }
 
         fn update_transaction(
@@ -702,7 +727,7 @@ mod tests {
     #[test]
     fn stores_loads_and_updates_a_migration() {
         let mut backend = MockBackend::new(Vec::new(), 0);
-        assert!(backend.load_migration().unwrap().is_none());
+        assert!(backend.get_migration().unwrap().is_none());
 
         let mut rng = ChaCha8Rng::seed_from_u64(1);
         let note_split =
@@ -723,11 +748,11 @@ mod tests {
             funding_notes: Vec::new(),
             transactions: vec![tx],
         };
-        backend.store_migration(&state).unwrap();
+        backend.put_migration(&state).unwrap();
 
         // The stored transactions round-trip, and a state update persists.
         let loaded = backend
-            .load_migration()
+            .get_migration()
             .unwrap()
             .expect("a migration is stored");
         assert_eq!(loaded.status, MigrationStatus::Committed);
@@ -741,7 +766,7 @@ mod tests {
                 },
             )
             .unwrap();
-        let loaded = backend.load_migration().unwrap().unwrap();
+        let loaded = backend.get_migration().unwrap().unwrap();
         assert_eq!(
             loaded.transactions[0].state,
             MigrationTxState::Mined {
@@ -794,14 +819,20 @@ mod commit_tests {
         fn chain_tip_height(&self) -> Result<BlockHeight, Self::Error> {
             Ok(BlockHeight::from_u32(2_000_000))
         }
+    }
 
-        fn store_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error> {
+    impl PoolMigrationRead for CommitMock {
+        type Error = core::convert::Infallible;
+
+        fn get_migration(&self) -> Result<Option<MigrationState>, Self::Error> {
+            Ok(self.stored.clone())
+        }
+    }
+
+    impl PoolMigrationWrite for CommitMock {
+        fn put_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error> {
             self.stored = Some(state.clone());
             Ok(())
-        }
-
-        fn load_migration(&self) -> Result<Option<MigrationState>, Self::Error> {
-            Ok(self.stored.clone())
         }
 
         fn update_transaction(
@@ -942,6 +973,6 @@ mod commit_tests {
             );
             assert!(tx.pczt.as_ref().is_some_and(|b| !b.is_empty()));
         }
-        assert!(backend.load_migration().unwrap().is_some());
+        assert!(backend.get_migration().unwrap().is_some());
     }
 }
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -395,11 +395,6 @@ pub enum CommitError<E> {
     Backend(E),
     /// Building or serializing a transaction failed.
     Build(alloc::string::String),
-    /// Retained for API compatibility; no longer returned. Multi-layer preparation is now committed
-    /// phase by phase: [`commit_preparation`] signs layer 0 and [`commit_pending_preparation`] signs
-    /// each later layer once its predecessor mines, so a plan whose later layers spend earlier layers'
-    /// outputs is fully supported.
-    UnsupportedMultiLayer,
     /// No committed migration was found to build the transfers for (nothing was loaded from storage).
     NoMigrationInProgress,
 }
@@ -410,9 +405,6 @@ impl<E: fmt::Display> fmt::Display for CommitError<E> {
         match self {
             CommitError::Backend(e) => write!(f, "wallet backend error: {e}"),
             CommitError::Build(m) => write!(f, "building the migration failed: {m}"),
-            CommitError::UnsupportedMultiLayer => {
-                f.write_str("multi-layer preparation cannot yet be pre-signed (single-layer only)")
-            }
             CommitError::NoMigrationInProgress => {
                 f.write_str("no committed migration is in progress")
             }
@@ -591,7 +583,8 @@ where
 ///
 /// Loads the committed migration, finds the earliest still-`Planned` preparation layer (a `layer > 0`)
 /// whose whole immediately-prior layer is [`Mined`](MigrationTxState::Mined), builds and pre-signs
-/// EVERY transaction of that one layer (resolving each transaction's [`PrepInput::Prior`] spends to the
+/// EVERY transaction of that one layer (resolving each transaction's
+/// [`PrepInput::Prior`](crate::preparation::PrepInput::Prior) spends to the
 /// prior layer's now-witnessable feeder notes by value), and persists the result. If no layer is ready
 /// (the next layer's dependencies are not all mined, or every preparation layer is already built), it
 /// returns the migration unchanged, so a caller can poll it. Call it once per newly-mined layer until
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -178,6 +178,47 @@ pub enum MigrationStatus {
     Failed,
 }
 
+impl AsRef<str> for MigrationStatus {
+    /// The stable lowercase wire name of the status, as stored by a backend and parsed back with
+    /// [`TryFrom<&str>`](Self). Borrow-free: it returns a `&'static str`, so encoding a status
+    /// allocates nothing.
+    fn as_ref(&self) -> &str {
+        match self {
+            MigrationStatus::Planning => "planning",
+            MigrationStatus::Committed => "committed",
+            MigrationStatus::InProgress => "in_progress",
+            MigrationStatus::Complete => "complete",
+            MigrationStatus::Failed => "failed",
+        }
+    }
+}
+
+/// The error returned when a string does not name a [`MigrationStatus`] (its [`TryFrom<&str>`] impl).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ParseMigrationStatusError;
+
+impl fmt::Display for ParseMigrationStatusError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("unrecognized migration status")
+    }
+}
+
+impl TryFrom<&str> for MigrationStatus {
+    type Error = ParseMigrationStatusError;
+
+    /// Parses the lowercase wire name produced by [`AsRef<str>`](AsRef).
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        Ok(match s {
+            "planning" => MigrationStatus::Planning,
+            "committed" => MigrationStatus::Committed,
+            "in_progress" => MigrationStatus::InProgress,
+            "complete" => MigrationStatus::Complete,
+            "failed" => MigrationStatus::Failed,
+            _ => return Err(ParseMigrationStatusError),
+        })
+    }
+}
+
 /// The persisted state of a migration: the note split (for the preview and residual accounting) and
 /// every transaction, each as its pre-signed PCZT and metadata. A wallet resumes a migration entirely
 /// from this state after being closed or restarted; this is what a [`MigrationBackend`] stores.
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -666,6 +666,11 @@ pub enum CommitError<E> {
     Build(alloc::string::String),
     /// No committed migration was found to build the transfers for (nothing was loaded from storage).
     NoMigrationInProgress,
+    /// A resolved wallet note's value does not match the value the plan recorded for it: the
+    /// wallet's spendable set changed between planning and commit (the plan captures notes by
+    /// their index into that set, so any receipt or spend since planning shifts them), and the
+    /// migration must be re-planned.
+    StalePlan,
     /// A non-terminal migration is already stored. A committed migration is resumed from the
     /// store (or cancelled), never rebuilt over: overwriting it would orphan its pre-signed —
     /// and possibly already broadcast — transactions, and a rebuilt layer 0 would double-spend
@@ -682,6 +687,10 @@ impl<E: fmt::Display> fmt::Display for CommitError<E> {
             CommitError::NoMigrationInProgress => {
                 f.write_str("no committed migration is in progress")
             }
+            CommitError::StalePlan => f.write_str(
+                "a wallet note no longer matches the plan; the spendable set changed since \
+                 planning and the migration must be re-planned",
+            ),
             CommitError::MigrationInProgress => f.write_str(
                 "a non-terminal migration is already stored; resume or cancel it instead of \
                  committing a new one",
@@ -908,10 +917,19 @@ where
                 let mut spends = Vec::with_capacity(prep_tx.inputs().len());
                 for input in prep_tx.inputs() {
                     match input {
-                        PrepInput::Wallet { index, .. } => {
+                        PrepInput::Wallet { index, value } => {
                             let witness = backend
                                 .resolve_wallet_note(*index, anchor_height)
                                 .map_err(CommitError::Backend)?;
+                            // The plan captured this note by its index into the spendable set at
+                            // PLANNING time; any receipt or spend since then shifts the indices,
+                            // so a resolved note whose value differs from the planned one means
+                            // the plan is stale — caught here as a typed error rather than as an
+                            // opaque balance failure (or, for an equal-valued interloper, a
+                            // silently signed spend of a note the plan reserved elsewhere).
+                            if witness.0.value().inner() != u64::from(*value) {
+                                return Err(CommitError::StalePlan);
+                            }
                             spends.push(witness);
                         }
                         // A layer-0 transaction spends only wallet notes; a Prior input here is a
@@ -1172,10 +1190,15 @@ where
         let mut spends = Vec::with_capacity(prep_tx.inputs().len());
         for input in prep_tx.inputs() {
             match input {
-                PrepInput::Wallet { index, .. } => {
+                PrepInput::Wallet { index, value } => {
                     let witness = backend
                         .resolve_wallet_note(*index, anchor_height)
                         .map_err(CommitError::Backend)?;
+                    // As in `commit_preparation_inner`: a value mismatch means the spendable
+                    // set shifted since planning and the plan is stale.
+                    if witness.0.value().inner() != u64::from(*value) {
+                        return Err(CommitError::StalePlan);
+                    }
                     spends.push(witness);
                 }
                 PrepInput::Prior { .. } => {
@@ -1966,6 +1989,57 @@ mod commit_tests {
             &mut rng,
         )
         .expect("replaces a terminal migration");
+    }
+
+    /// A wallet note is resolved by its index into the CURRENT spendable set; if that set
+    /// changed since planning (the value at the planned index no longer matches), the commit
+    /// reports a stale plan instead of building against the wrong note.
+    #[test]
+    fn commit_preparation_detects_a_stale_plan() {
+        let seed = 17u64;
+        let sk = spending_key(seed);
+        let fvk = FullViewingKey::from(&sk);
+        let balance = 78 * COIN;
+
+        let plan = {
+            let (note, path, anchor) = single_note_witness(&fvk, balance, seed);
+            let planner = CommitMock {
+                notes: vec![Zatoshis::from_u64(balance).expect("test balance is valid")],
+                witnesses: vec![(note, path)],
+                anchor,
+                fvk: fvk.clone(),
+                ask: SpendAuthorizingKey::from(&sk),
+                stored: None,
+            };
+            let mut rng = ChaCha8Rng::seed_from_u64(seed);
+            plan_migration(&regtest_network(true), &planner, &mut rng).expect("plans a migration")
+        };
+
+        // The spendable set shifted between planning and commit: the note at index 0 now has a
+        // different value than the plan recorded.
+        let mut values = vec![balance + COIN];
+        values.extend(plan.funding_notes().iter().map(|&v| u64::from(v)));
+        let (witnesses, anchor) = shared_anchor_witnesses(&fvk, &values, seed);
+        let mut backend = CommitMock {
+            notes: vec![Zatoshis::from_u64(balance).expect("test balance is valid")],
+            witnesses,
+            anchor,
+            fvk,
+            ask: SpendAuthorizingKey::from(&sk),
+            stored: None,
+        };
+
+        let mut rng = ChaCha8Rng::seed_from_u64(seed + 1);
+        assert!(matches!(
+            commit_preparation(
+                &regtest_network(true),
+                BlockHeight::from_u32(TARGET_HEIGHT),
+                &mut backend,
+                &plan,
+                &mut rng,
+            ),
+            Err(CommitError::StalePlan)
+        ));
     }
 
     /// A wallet mock for the MULTI-LAYER preparation test. The first `n_wallet` witnesses are the
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -484,39 +484,32 @@ where
     let (prep_tx_fee, transfer_fee_buffer) =
         canonical_fees(params, commit_height).map_err(MigrationError::Fee)?;
 
-    let note_values: Vec<u64> = notes.iter().map(|&n| u64::from(n)).collect();
     // The preparation-layout capability the decomposition consults at each step: how many
     // preparation transactions minting a candidate funding multiset takes, or `None` when the
     // wallet's notes cannot mint it (so the split stops or steps down a denomination).
-    let prep_tx_count = |funding: &[u64]| {
-        plan_preparation(&note_values, funding, u64::from(prep_tx_fee))
+    let prep_tx_count = |funding: &[Zatoshis]| {
+        plan_preparation(&notes, funding, prep_tx_fee)
             .ok()
             .map(|plan| plan.transaction_count())
     };
     let note_split = plan_note_split(
-        u64::from(balance),
-        u64::from(transfer_fee_buffer),
-        u64::from(prep_tx_fee),
+        balance,
+        transfer_fee_buffer,
+        prep_tx_fee,
         &prep_tx_count,
         rng,
     );
-    let funding_notes: Vec<u64> = note_split.migration_outputs();
+    let funding_notes = note_split.migration_outputs();
     if funding_notes.is_empty() {
         return Err(MigrationError::NothingToMigrate);
     }
 
     // The decomposition verified this multiset against the preparation planner at every step, so
     // this final planning pass succeeds by construction; the error path is kept for safety.
-    let preparation = plan_preparation(&note_values, &funding_notes, u64::from(prep_tx_fee))
+    let preparation = plan_preparation(&notes, &funding_notes, prep_tx_fee)
         .map_err(MigrationError::Preparation)?;
 
     let schedule = scheduling::schedule(commit_height, funding_notes.len(), rng);
-
-    let funding_notes = funding_notes
-        .into_iter()
-        .map(Zatoshis::from_u64)
-        .collect::<Result<Vec<_>, _>>()
-        .map_err(MigrationError::InvalidBalance)?;
 
     Ok(MigrationPlan {
         note_split,
@@ -1019,7 +1012,7 @@ where
     // resolve them together so each feeder note (distinct by tree position, even at equal value) is
     // matched to exactly one spend across the layer. Wallet inputs (not expected past layer 0) resolve
     // individually.
-    let mut prior_values: Vec<u64> = Vec::new();
+    let mut prior_values: Vec<Zatoshis> = Vec::new();
     for &(_, plan_index) in &targets {
         for input in state.preparation.layers()[ready_layer][plan_index].inputs() {
             if let PrepInput::Prior { value, .. } = input {
@@ -1027,11 +1020,6 @@ where
             }
         }
     }
-    let prior_values = prior_values
-        .into_iter()
-        .map(Zatoshis::from_u64)
-        .collect::<Result<Vec<_>, _>>()
-        .map_err(|_| CommitError::Build("prior input value out of range".into()))?;
     let mut prior_notes = backend
         .resolve_funding_notes(&prior_values, anchor_height)
         .map_err(CommitError::Backend)?
@@ -1191,8 +1179,12 @@ where
         state.note_split.migration_outputs().first(),
         state.note_split.crossing_values().first(),
     ) {
-        (Some(funding), Some(crossing)) => funding.saturating_sub(*crossing),
-        _ => 0,
+        // A prepared note is its crossing plus the buffer by construction, so the checked
+        // subtraction underflows only for an inconsistently stored plan.
+        (Some(&funding), Some(&crossing)) => (funding - crossing).ok_or_else(|| {
+            CommitError::Build("stored note split has a crossing exceeding its note".into())
+        })?,
+        _ => Zatoshis::ZERO,
     };
 
     for tx in state.transactions.iter_mut() {
@@ -1207,7 +1199,10 @@ where
         let (note, merkle_path) = witnesses.get(crossing).cloned().ok_or_else(|| {
             CommitError::Build(format!("no funding note for crossing {crossing}"))
         })?;
-        let crossing_value = u64::from(state.funding_notes[crossing]).saturating_sub(buffer);
+        // The funding note is its crossing plus the buffer by construction (see above).
+        let crossing_value = (state.funding_notes[crossing] - buffer).ok_or_else(|| {
+            CommitError::Build("stored funding note is smaller than the fee buffer".into())
+        })?;
 
         let pczt = build_transfer_pczt(
             params,
@@ -1278,7 +1273,7 @@ mod tests {
 
     /// A count-only preparation-layout stub for tests that exercise the split in isolation: one
     /// padded transaction per [`FUNDING_OUTPUTS_PER_TX`] funding notes.
-    fn prep_tx_count_stub(notes: &[u64]) -> Option<usize> {
+    fn prep_tx_count_stub(notes: &[Zatoshis]) -> Option<usize> {
         Some(notes.len().div_ceil(FUNDING_OUTPUTS_PER_TX))
     }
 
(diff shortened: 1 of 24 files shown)


zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,893 @@
...
+    /// application needs to prove, schedule, and broadcast it. Storing the pre-signed transactions,
+    /// not just the plan, is what lets a wallet resume a migration after being closed or restarted.
+    fn store_migration(&mut self, state: &MigrationState) -> Result<(), Self::Error>;
+
+    /// Load the persisted migration state, if a migration is in progress.
+    fn load_migration(&self) -> Result<Option<MigrationState>, Self::Error>;
+
+    /// Advance one stored transaction's lifecycle state (for example after the application broadcasts
+    /// it, or the chain mines it, or it expires).
+    fn update_transaction(
```
dannywillems, line 111: We could use type state variables. Nit.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,893 @@
...
+pub enum MigrationTxKind {
+    /// A note-preparation transaction: the `index`-th transaction of preparation `layer`.
+    Preparation { layer: usize, index: usize },
+    /// A phase-2 pool-crossing transfer of the `crossing`-th funding note.
+    Transfer { crossing: usize },
+}
+
+/// Where a migration transaction is in its lifecycle.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MigrationTxState {
```
dannywillems, line 151: Could be type-state variables.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,893 @@
...
+    /// The boundary height whose tree state the transaction proves against, drawn at proving time (for
+    /// a transfer); `None` until proved.
+    pub anchor_boundary: Option<u32>,
+    /// The transaction's lifecycle state.
+    pub state: MigrationTxState,
+}
+
+/// The overall status of a migration.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MigrationStatus {
```
dannywillems, line 242: Could be type state variables.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,893 @@
...
+}
+
+/// Where a migration transaction is in its lifecycle.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MigrationTxState {
+    /// Built but not yet signed.
+    Planned,
+    /// Pre-signed (the account's spend authorization is attached), not yet proved.
+    Signed,
+    /// Proved against a real anchor, ready to broadcast.
```
dannywillems, line 161: There is `AwaitingSignature` in https://github.com/zcash/librustzcash/pull/2669/changes/9ed123a6989874c130f816cd069bb9ba8ea8b87b. Keeping it separately to avoid rebasing too much for now.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+    fn update_transaction(
+        &mut self,
+        id: MigrationTxId,
+        state: MigrationTxState,
+    ) -> Result<(), Self::Error>;
+}
+
+/// A stable identifier for a migration transaction within a migration.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct MigrationTxId(pub u32);
```
dannywillems: Could we have a more precise type than a u32, if it is an actual Zcash transaction ID?

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+    /// signed PCZT is returned. Only the external-signing path
+    /// ([`build_preparation_unsigned`]/[`build_transfers_unsigned`]) produces this state; the in-process
+    /// commit functions sign immediately and go straight to [`Signed`](Self::Signed).
+    AwaitingSignature,
+    /// Pre-signed (the account's spend authorization is attached), not yet proved.
+    Signed,
+    /// Proved against a real anchor, ready to broadcast.
+    Proved,
+    /// Broadcast to the network, with its transaction id.
+    Broadcast { txid: [u8; 32] },
```
dannywillems: Could we have a more precise type than 32 bytes, if it represents a Zcash transaction ID?

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+            _ => return Err(ParseMigrationStatusError),
+        })
+    }
+}
+
+/// The persisted state of a migration: the note split (for the preview and residual accounting) and
+/// every transaction, each as its pre-signed PCZT and metadata. A wallet resumes a migration entirely
+/// from this state after being closed or restarted; this is what a [`MigrationBackend`] stores.
+#[derive(Clone, Debug)]
+pub struct MigrationState {
```
dannywillems, line 300: Could be using a type-state variable.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+        .chain_tip_height()
+        .map_err(MigrationError::Backend)?;
+
+    let note_split = plan_note_split(balance, prep_fee_zatoshi, rng);
+    if note_split.migration_outputs().is_empty() {
+        return Err(MigrationError::NothingToMigrate);
+    }
+
+    // Reconcile the note split against the preparation fees. The split reserves for a single prep
+    // transaction, but preparation may need several; when its fees do not fit the balance, drop the
```
dannywillems: We must have tests that ensure that the next time, the funding note will appear. I will add an integration-tests for that. We must have the case where the migration spans over multiple plans.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+#[cfg(feature = "orchard")]
+pub trait MigrationCrypto {
+    /// The backend's error type (shared with its [`MigrationBackend`] impl).
+    type Error;
+
+    /// The account's Orchard full viewing key.
+    fn orchard_fvk(&self) -> Result<orchard::keys::FullViewingKey, Self::Error>;
+
+    /// A current Orchard anchor to build against; every spend's witness resolves against this same tree
+    /// state. The proof is re-anchored to a drawn boundary at proving time, and the anchor is not in the
```
dannywillems: We must have a test for this statement.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+{
+    match signing {
+        Signing::InProcess => {
+            let signed = backend.sign(pczt).map_err(CommitError::Backend)?;
+            let bytes = signed
+                .serialize()
+                .map_err(|e| CommitError::Build(format!("serialize: {e:?}")))?;
+            Ok((bytes, MigrationTxState::Signed))
+        }
+        Signing::External => {
```
dannywillems, line 766: This handles Keystone support.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+            let bytes = pczt
+                .serialize()
+                .map_err(|e| CommitError::Build(format!("serialize: {e:?}")))?;
+            Ok((bytes, MigrationTxState::AwaitingSignature))
+        }
+    }
+}
+
+/// Commit a planned migration's PREPARATION: build and pre-sign the FIRST layer of preparation
```
dannywillems: Note that we sign and broadcast all the transactions for the preparation first, before migrating any note to Ironwood.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+/// engine builds and signs the next one (see [`commit_pending_preparation`]) whose spends reference
+/// that layer's now-witnessable feeder notes, until the whole preparation is mined and the transfers
+/// are built (see [`commit_transfers`]).
+///
+/// A single-layer plan (the common case: a few wallet notes fanning into a handful of funding notes)
+/// signs its one layer here and records only the transfer placeholders; nothing is deferred. A
+/// multi-layer plan (a lone whale fanning out, or a dust-heavy balance) signs layer 0 here and defers
+/// the rest.
+///
+/// For an EXTERNAL signer (a hardware wallet), use [`build_preparation_unsigned`] instead, which builds
```
dannywillems: This will be the method to be used by the mobile SDK.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+                let depends_on = layer_ids
+                    .last()
+                    .cloned()
+                    .expect("a layer after layer 0 has a preceding layer");
+                transactions.push(MigrationTransaction {
+                    id,
+                    kind: MigrationTxKind::Preparation { layer, index },
+                    pczt: None,
+                    depends_on,
+                    scheduled_height: target_height,
```
dannywillems: Wondering if the height should be different for the other layers. Might they be different? Maybe the scheduled_height? `expiry_height` is large enough.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+        })
+        .min();
+
+    let Some(ready_layer) = ready_layer else {
+        // Nothing to build this step (the next layer is not mined yet, or all layers are built).
+        return Ok(state);
+    };
+
+    let fvk = backend.orchard_fvk().map_err(CommitError::Backend)?;
+    let anchor = backend.orchard_anchor().map_err(CommitError::Backend)?;
```
dannywillems: Note: the anchor MUST be changed when proving. This must be verified in a test layer.

zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -249,6 +249,30 @@ impl NoteSplitPlan {
         }
     }
 
+    /// Reassemble a plan from its stored fields, exactly as they were persisted. This is the inverse
```
dannywillems, line 213: This could have been added directly the PR adding the structure.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -118,6 +118,13 @@ pub struct PrepTransaction {
 }
 
 impl PrepTransaction {
+    /// Construct a transaction from its spent and produced notes. Used to reconstruct a persisted plan
```
dannywillems, line 136: This could have been added directly the PR adding the structure.

zcash_pool_migration_backend/src/preparation.rs
```diff
@@ -146,6 +153,20 @@ pub struct PreparationPlan {
 }
 
 impl PreparationPlan {
+    /// Reconstruct a plan from its parts: the layers in dependency order (see [`layers`](Self::layers))
+    /// and the direct-funding notes (see [`direct_funding_notes`](Self::direct_funding_notes)). Used by
+    /// a store to round-trip a persisted plan; the caller supplies parts a valid plan could have
+    /// produced (no validation beyond what the accessors expose is done here).
+    pub fn from_parts(
```
dannywillems, line 175: This could have been added directly the PR adding the structure.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1664 @@
...
+    use crate::preparation::{PREP_TX_ACTIONS, PrepInput, plan_preparation};
+
+    fn prep_fee() -> u64 {
+        PREP_TX_ACTIONS as u64 * Zip317FeePolicy.marginal_fee_zatoshi()
+    }
+
+    /// A wallet holding the account's key and a set of note witnesses against one anchor: index 0 is the
+    /// source note the preparation spends, and the rest are the funding notes the transfers spend. It
+    /// signs with its own spend-authorizing key and stores the migration in memory.
+    struct CommitMock {
```
dannywillems, line 1411: Moving in their own crate for testing.

dannywillems: See https://github.com/zcash/librustzcash/pull/2673

zcash_pool_migration_backend/src/state.rs
```diff
@@ -0,0 +1,663 @@
...
+    pub action: Option<NextAction>,
+    /// Why it is not yet actionable, when it is waiting (and not already broadcast or mined).
+    pub blocked_on: Option<Blocker>,
+    /// The height it was mined at, once mined.
+    pub mined_height: Option<u32>,
+    /// The transaction id (raw internal bytes), once broadcast.
+    pub txid: Option<[u8; 32]>,
+}
+
+impl MigrationState {
```
dannywillems, line 112: This could have been moved next to the `MigrationState` declaration.

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,380 @@
...
+            store,
+        }
+    }
+
+    /// Recover the store.
+    pub fn into_store(self) -> St {
+        self.store
+    }
+
+    /// The anchor checkpoint height the migration builds against (a checkpoint guaranteed to exist in
```
dannywillems: I think this could be simplified. There might be unnecessary wrappers around methods, and we could directly use `WalletRead`. Only a thought, not a confirmed statement.

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,380 @@
...
+impl<'a, W, St> MigrationBackend for WalletMigration<'a, W, St>
+where
+    W: WalletRead + InputSource + WalletCommitmentTrees,
+    <W as InputSource>::AccountId: Copy,
+    <W as WalletCommitmentTrees>::Error: fmt::Debug,
+    St: PoolMigrationRead,
+{
+    type Error = AdapterError<W, St>;
+
+    fn spendable_orchard_note_values(&self) -> Result<Vec<u64>, Self::Error> {
```
dannywillems: As mentioned above, these might be unnecessary wrappers. Just a thought, not confirmed statement. Could be checked in a follow-up.

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    fn orchard_anchor(&self) -> Result<Anchor, Self::Error> {
+        Ok(self.witness(&[])?.0)
+    }
+
+    fn resolve_wallet_note(
+        &self,
+        index: usize,
+        _anchor: Anchor,
```
dannywillems: Why do we have the anchor?

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,380 @@
...
+        let notes = self.spendable_orchard()?;
+        let &(note, position, _) = notes.get(index).ok_or(Error::NoteNotFound(index))?;
+        let (_, mut paths) = self.witness(&[position])?;
+        Ok((note, paths.remove(0)))
+    }
+
+    fn resolve_funding_notes(
+        &self,
+        values: &[u64],
+        _anchor: Anchor,
```
dannywillems: Why do we have the anchor?

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,380 @@
...
+
+    /// Compile-time proof that `WalletMigration` over ANY `zcash_client_backend` wallet `W` and ANY
+    /// migration store `St` satisfies every trait bound `commit_preparation` / `commit_transfers`
+    /// require (backend + crypto + store, all sharing one error type). Naming the two generic
+    /// functions instantiated at `WalletMigration<W, St>` forces the type checker to verify that
+    /// instantiation's bounds hold; if the four trait impls ever stop lining up with the commit path,
+    /// this stops compiling. It is never called and needs no wallet instance, so it pulls in no
+    /// test-only wallet dependency (which would otherwise force `zcash_client_backend`'s Orchard
+    /// feature on across the whole workspace's test build).
+    #[allow(dead_code)]
```
dannywillems, line 268: Purely user API level, and dead code as we only want the compiler.

zcash_pool_migration_backend/src/state.rs
```diff
@@ -0,0 +1,663 @@
...
+            "a Failed (cancelled) migration must not be revived to InProgress"
+        );
+        assert!(s.is_terminal());
+
+        // The next step for a terminal migration is Complete (no action), so a driver never
+        // broadcasts or builds for it; a Signed transaction is NOT offered for broadcast.
+        assert_eq!(s.next_step(100), AdvanceStep::Complete);
+
+        // Detecting a mined transaction still does not resurrect it.
+        s.mark_mined(MigrationTxId(0), 10);
```
dannywillems: Well. That should never happened.

dannywillems (commented):
self-utACK 962e2952b0cec037c3fad36edb296248df12467b. Different minor comments. I think this could be improved. It could be done in follow-up.

zcash_pool_migration_backend/src/engine.rs
dannywillems: This is now here.

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,388 @@
...
+        notes.sort_by_key(|(_, pos, _)| *pos);
+        Ok(notes)
+    }
+
+    /// Resolve the anchor and a witness for each requested tree position, all against one checkpoint.
+    fn witness(
+        &self,
+        positions: &[Position],
+    ) -> Result<(Anchor, Vec<MerklePath>), AdapterError<W, St>> {
+        let anchor_height = self.anchor_height()?;
```
nuttycom: This looks wrong; any proof we're making (and consequently any witness) should be computed against the anchor at the bucketed height, not against what `get_target_and_anchor_heights` returns. At least, this is true of the witnesses for the transactions that cross the turnstile. This function, if it's needed, should take the anchor height as an argument so that we can be certain that the correct anchor is being chosen for each call site.

zcash_pool_migration_backend/src/engine.rs
nuttycom: Probably would add complexity without much benefit, given that we have to return in `Result` anyway and whenever you add type-directed state it makes a bunch of call sites more complicated.

zcash_pool_migration_backend/src/engine.rs
nuttycom: Agreed, this should just use an ordinary `TxId`; the transaction id is computed exclusively from effecting data that is available at the time of migration transaction construction.

zcash_pool_migration_backend/src/engine.rs
nuttycom: `TxId` can be computed as soon as the transaction can be signed.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1746 @@
...
+    Mined { height: BlockHeight },
+    /// Expired before it could be mined, and to be rebuilt.
+    Expired,
+}
+
+/// One transaction of a committed migration: its pre-signed PCZT plus the metadata the consuming
+/// application needs to prove it against a fresh anchor, wait for its dependencies, broadcast it at
+/// its scheduled height, and track its state.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MigrationTransaction {
```
nuttycom, line 173: We do not permit `pub` fields in `pub` structs. Use constructors and accessors (the getset crate is used in PCZT, so it is in the dependency tree already).

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1746 @@
...
+    /// dependency graph; empty for an independent transaction).
+    pub depends_on: Vec<MigrationTxId>,
+    /// The height at which to broadcast (for a transfer; a preparation transaction waits for its
+    /// dependencies to mine and a boundary to pass rather than a fixed height).
+    pub scheduled_height: BlockHeight,
+    /// The height after which the transaction is invalid and must be rebuilt.
+    pub expiry_height: BlockHeight,
+    /// The boundary height whose tree state the transaction proves against, drawn at proving time (for
+    /// a transfer); `None` until proved.
+    pub anchor_boundary: Option<BlockHeight>,
```
nuttycom: The boundary can be identified as part of the scheduling process; it doesn't have to wait for proving time. We know in advance the scheduled block height, and that combined with the scheduled height of the transaction that funds the migration transaction fully determines the anchor height.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1746 @@
...
+    }
+}
+
+/// The persisted state of a migration: the note split (for the preview and residual accounting) and
+/// every transaction, each as its pre-signed PCZT and metadata. A wallet resumes a migration entirely
+/// from this state after being closed or restarted; this is what a [`MigrationBackend`] stores.
+#[derive(Clone, Debug)]
+pub struct MigrationState {
+    /// The overall status.
+    pub status: MigrationStatus,
```
nuttycom: Same comment about pub fields.

nuttycom (commented):
Flushing partial review (up to zcash_pool_migration_backend/src/engine.rs `plan_migration`)

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1746 @@
...
+        .ok_or(MigrationError::InvalidBalance(BalanceError::Overflow))?;
+    if balance == Zatoshis::ZERO {
+        return Err(MigrationError::NothingToMigrate);
+    }
+    let commit_height = backend
+        .chain_tip_height()
+        .map_err(MigrationError::Backend)?;
+
+    let note_values: Vec<u64> = notes.iter().map(|&n| u64::from(n)).collect();
+    let note_split = plan_note_split(u64::from(balance), u64::from(prep_fee), rng);
```
nuttycom: `plan_note_split` is a public function and should use proper `Zatoshis` types for balance.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,1746 @@
...
+/// transaction, which the note split and the preparation planner both reserve. `rng` must be a
+/// cryptographically secure RNG (the schedule's shuffle, delays, and the note split's optional
+/// randomization draw from it).
+///
+/// This is pure orchestration of the note-split, preparation, and scheduling planners: no cryptography,
+/// and nothing is built, signed, or persisted. The result is the [`MigrationPlan`] preview to present
+/// for user consent before committing the migration.
+pub fn plan_migration<B, R>(
+    backend: &B,
+    prep_fee: Zatoshis,
```
nuttycom: Taking the prep fee as an argument here says there's something wrong with the preparation further down the stack; you can't know the fee until you know how many notes you have to migrate and how many transactions it will take to migrate them.

nuttycom: Since decomposition is an iterative process, this should take the fee rule, and then the plan_note_split function should compute the fees at each step of its operation.

zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -231,25 +203,33 @@ impl NoteSplitPlan {
...
+    /// of the accessors below: a store (for example `zcash_pool_migration_sqlite`) reads the columns
+    /// back and reconstructs the plan verbatim, so `total_migratable_zatoshi` is taken as stored
+    /// rather than recomputed (the caller is responsible for having persisted a consistent set, which
+    /// for a plan produced by `Self::from_notes` means `total_migratable_zatoshi` equals the sum of
+    /// `crossing_values`).
+    pub fn from_stored_parts(
+        crossing_values: Vec<u64>,
+        note_fee_buffer_zatoshi: u64,
         change: Option<u64>,
+        prep_fees_zatoshi: u64,
```
nuttycom: These also need to be `Zatoshis`. The `change` field is also unclear what it's representing; is this also a Zatoshis value?

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2013 @@
...
+                });
+            }
+        }
+        layer_ids.push(this_layer_ids);
+    }
+
+    // Every transfer waits for the last preparation layer to be mined: a layer is broadcast only after
+    // its predecessor mines, so the last layer mining implies every layer (hence every funding note) is
+    // mined and witnessable. An empty preparation (every funding note used directly) leaves this empty.
+    let last_layer_ids: Vec<MigrationTxId> = layer_ids.last().cloned().unwrap_or_default();
```
nuttycom: Ideally we wouldn't need to wait for all the layers to be mined; as soon as we have notes prepared for pool-crossing, those transactions should be able to go at their scheduled times. We do lose a little bit if a lwd-compromising adversary is able to determine the full size of the preparation transaction set for a single wallet, but scheduling the layered prep transactions interleaved with the pool-crossing transactions requires more bookkeeping and likely leaks in different ways.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2013 @@
...
+    // signed later, once the preparation is mined (see `commit_transfers`). This persists the drawn
+    // schedule (which is not reproducible) as part of the committed migration.
+    //
+    // The boundary anchor is drawn HERE, at scheduling time, because the schedule fully determines
+    // it: the candidate set lies strictly above the NU6.3 activation, at or after the height the
+    // funding preparation is built to mine (`target_height` bounds the funding notes' creation from
+    // below), and strictly below the most recent boundary at the transfer's scheduled broadcast
+    // height (the chain view the wallet will have observed when it proves, just before broadcast).
+    // `None` means no candidate boundary exists at that scheduled height (a transfer scheduled very
+    // close to activation); the application then draws against the proving-time view instead.
```
nuttycom: The schedule *must* ensure that at least one valid boundary exists for every possible scheduled transaction. The fallback described here shouldn't be necessary.

zcash_pool_migration_backend/src/scheduling.rs
```diff
@@ -74,6 +74,19 @@ pub const MEAN_DELAY: u32 = 144;
 /// an unbounded time. 576 blocks is `4 * MEAN_DELAY`, about twelve hours. See [`draw_delay`].
 pub const MAX_DELAY: u32 = 576;
 
+/// Mean of the exponential inter-arrival delay between successive PREPARATION transactions, in
+/// blocks (~15 minutes at the ~75-second target spacing). Preparation transactions are fully
+/// shielded self-sends: they need TEMPORAL decoupling from one another (a burst of identically
+/// shaped transactions from one wallet is a linkable cluster), but no anchor bucketing — only the
+/// pool-crossing transfers anchor to boundaries — so their spacing can be much tighter than the
+/// transfers' [`MEAN_DELAY`]. Provisional; not yet specified by ZIP 318.
+pub const PREP_MEAN_DELAY: u32 = 12;
```
nuttycom: Let's double this (also doubling pre_max_delay)

zcash_pool_migration_backend/src/state.rs
```diff
@@ -0,0 +1,723 @@
...
+                            (true, Some(NextAction::BuildAndSign), None)
+                        } else {
+                            (false, None, Some(Blocker::Dependencies))
+                        }
+                    }
+                    // Built for an external signer and waiting for its signed PCZT; the wallet's
+                    // automatic driver takes no action (the external-signing caller drives it via
+                    // `apply_signature`), so it is neither ready nor blocked on the chain.
+                    MigrationTxState::AwaitingSignature => (false, None, Some(Blocker::Signature)),
+                    MigrationTxState::Signed | MigrationTxState::Proved => {
```
nuttycom, line 265: I'm confused about this case - *both* signing and proving must be complete in order for a transaction to be ready to be broadcast. Why is this an "or"?

nuttycom (dismissed):
utACK 36d78385ebe117f0913a42df39470c0d0c0e8dec

In retrospect it would have been good to review this before the PR that it was stacked upon merged, because it makes additional changes to some of that code, but due to time constraints that wasn't feasible.

nuttycom (commented):
Thorough review of the engine layer. The architecture is sound and the planner reuse is clean, but I found one critical privacy defect and four high-severity lifecycle bugs, mostly at the seams where the engine glues the planners to real chain state. The eight most severe findings are inline comments; additional confirmed items below.

**Additional confirmed findings (below the inline cut):**

- `engine.rs:445`, `engine.rs:481`, `build/end_to_end.rs:144` — these doc lines (added in this PR) still describe the transfer as "2-Orchard + 2-Ironwood" / "2+2-action", but the code now deliberately builds a single unpadded Ironwood action (3 logical actions, 15,000-zat fee; `DESTINATION_ACTIONS_PER_TRANSFER = 1`). Beyond the doc fix, the shape choice deserves a conscious decision: by the PR's own principle at engine.rs:481 ("a nonstandard fee would partition the anonymity set"), diverging from the Orchard builder's default padded shape that other ZIP 318 implementations would produce (4 actions, 20,000 zats) risks exactly that partition.
- `engine.rs:1098`/`1110` — `state.preparation.layers()[ready_layer][plan_index]` bracket-indexes values loaded from the store with no bounds check, and the `from_parts` constructors explicitly do not validate; a buggy/corrupt store panics in library code where the sibling path (engine.rs:1263, 1278) returns `CommitError`. AGENTS.md requires a typed error rather than a panic at public boundaries.
- `engine.rs:641` — `CommitError::Build(String)` funnels structured `BuildError`s and distinct recoverable conditions (NU6.3-inactive, which `MigrationError` models structurally as `Nu63NotActive`; the stale-plan "must be re-planned" case at engine.rs:973-978) into strings a caller cannot match on.
- Derive-don't-duplicate: `engine.rs:1257-1267` re-derives the fee buffer that `NoteSplitPlan::note_fee_buffer()` already exposes (the underflow branch is unreachable and the `Zatoshis::ZERO` fallback only fires with no crossings); `engine.rs:1043-1066` re-implements `MigrationState::ready_prep_layer()`/`deps_mined()` verbatim; `MigrationState.funding_notes` always equals `note_split.migration_outputs()` — and the comments at engine.rs:362/367-369 plus the `<=` assertion at engine.rs:1468 describing a before/after-reconciliation distinction are stale (reconciliation now happens inside `plan_note_split`).
- `wallet.rs:283` — `resolve_wallet_note` re-runs the full spendable-note query and opens a fresh tree session per input (up to ~50x per layer-0 build); the batched pattern already exists in `resolve_funding_notes`. Relatedly `commit_transfers_inner` has no early return when nothing is `Planned` and rewrites every PCZT blob via `put_migration` unconditionally.
- Minor: `schedule_prep_broadcast_heights` (scheduling.rs:289) is a line-for-line copy of `schedule_broadcast_heights` differing only in the delay draw; the engine test `test_net()` (engine.rs:1330) duplicates `build::test_util::regtest_network` byte-for-byte (a shared home needs to be ungated — `build` is orchard-gated while engine tests are not); engine.rs:1949 hardcodes `+ 10` for the in-scope `EST_PREP_LAYER_MINING_BLOCKS` under a comment claiming to mirror `plan_migration` exactly; `wallet::Error::Sign(String)` Debug-formats the concrete non-generic `BuildError` that could be carried structurally (`Tree(String)`'s generics justification does not apply to it); the CHANGELOG paragraph about `UnsupportedMultiLayer` refers to an error variant that no longer exists.

For the record, two plausible-looking candidates were checked and refuted: `commit_transfers_inner` re-invocation after partial progress cannot hit spent funding notes in the current driver flow (state persists once, after the loop), and replacing `selection_target`'s `chain_height()+1` with `get_target_and_anchor_heights` would be wrong (it bundles min-confirmations anchor policy; the adapter's boundary-height anchors are deliberate).

---
*This review was prepared with Claude Code assistance.*

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2142 @@
...
+    // gaps intact while making an empty candidate set impossible for a plan committed at (or
+    // reasonably near) its planning height.
+    let nu63_activation = params
+        .activation_height(zcash_protocol::consensus::NetworkUpgrade::Nu6_3)
+        .ok_or(MigrationError::Nu63NotActive)?;
+    let schedule_base = commit_height.max(scheduling::earliest_broadcast_height(
+        nu63_activation,
+        est_last_prep_height,
+    ));
+    let schedule = scheduling::schedule(schedule_base, funding_notes.len(), rng);
```
nuttycom: **[critical] The ZIP 318 SHUFFLE is never applied; broadcast order deterministically leaks the migrated balance.**

`shuffle_indices`/`shuffle_in_place` exist in scheduling.rs (with docs quoting the "SHUFFLE MUST") but their only callers are scheduling.rs's own tests. `plan_migration` passes the greedy largest-first split (strategies.rs: "the parts are non-increasing … a pure function of the balance; the `rng` argument is ignored") straight to `scheduling::schedule`, whose contract says it returns heights "in the (already shuffled) part order the caller passes". The order survives to chain: `Transfer { crossing: i }` is paired index-by-index with the non-decreasing cumulative heights and `next_broadcastable` releases in stored order, so on-chain time order = descending denominations = the balance's digit expansion — the exact leak the module doc warns about at scheduling.rs:15-16. The engine docs at engine.rs:4 and engine.rs:485 currently claim the shuffle happens.

Fix direction: draw a permutation with `shuffle_indices` here and apply it when pairing crossings with scheduled heights.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2142 @@
...
+                        PrepInput::Prior { .. } => {
+                            return Err(CommitError::Build(
+                                "layer 0 preparation transaction spends a prior-layer output"
+                                    .into(),
+                            ));
+                        }
+                    }
+                }
+
+                let (pczt, _placed) = build_prep_tx(
```
nuttycom: **[high] Preparation PCZTs embed nExpiryHeight = target+40 but are scheduled for broadcast beyond it; the stored expiry metadata masks the wedge.**

`build_prep_tx` (build/prep.rs:136) never calls `with_expiry_height` (unlike `build_transfer_pczt` at transfer.rs:75), so the Builder default `DEFAULT_TX_EXPIRY_DELTA = 40` applies — yet prep broadcasts are drawn at cumulative mean-24 delays (P(schedule > target+40) is roughly 0.17 / 0.50 / 0.77 for the 1st/2nd/3rd tx in a layer), and engine.rs:907 stores `expiry_height(scheduled_height)` (the ~1-2 month canonical window) as the transaction's metadata. `next_broadcastable` gates only on `scheduled_height <= target`, so it hands the consumer a network-invalid pre-signed transaction forever; nExpiryHeight is under the ZIP 244 sighash so it cannot be patched post-signing, and nothing ever sets `Expired` or rebuilds — the migration wedges.

Fix direction: thread the per-transaction expiry into `build_prep_tx` and set `with_expiry_height(expiry_height(scheduled_height))` so the embedded expiry matches the stored one.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2142 @@
...
+
+    let mut state = backend
+        .get_migration()
+        .map_err(CommitError::Backend)?
+        .ok_or(CommitError::NoMigrationInProgress)?;
+    let mut unsigned: Vec<UnsignedMigrationTx> = Vec::new();
+
+    let fvk = backend.orchard_fvk().map_err(CommitError::Backend)?;
+    // Witness against the BUCKETED anchor: the most recent boundary at the build height.
+    let anchor_height = scheduling::most_recent_boundary(target_height);
```
nuttycom: **[high] Transfers are witnessed at `most_recent_boundary(target)`, which almost always predates the preps' mining; there is no wait signal and the drawn `anchor_boundary` is never consulted.**

`next_step` dispatches `BuildTransfers` the moment `all_preparations_mined()` (state.rs:289-290) with no boundary gating, but the funding notes' positions do not exist at the last 144-boundary at/below the tip (~143/144 of the time on prompt dispatch), so shardtree hard-errors `QueryError::NotContained` (via wallet.rs:206), surfacing as a stringly `Error::Tree` inside `CommitError::Backend` — indistinguishable from a fatal error. It self-heals only under blind polling for up to ~144 blocks (~3 hours). The per-transfer `anchor_boundary` drawn at engine.rs:966-979 to satisfy the at-or-after-funding-creation condition is never read here, and the engine-test mocks ignore `anchor_height` entirely, so no test exercises this.

Fix direction: gate `BuildTransfers` (in `next_step`) on `most_recent_boundary(target)` reaching the preps' mined heights, and/or return a typed wait-until error here.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2142 @@
...
+    }
+
+    let state = MigrationState {
+        status: MigrationStatus::Committed,
+        note_split: plan.note_split().clone(),
+        funding_notes,
+        preparation: plan.preparation().clone(),
+        transactions,
+    };
+    backend
```
nuttycom, line 1124: **[high] `commit_preparation_inner` persists via `put_migration` without checking for an existing non-terminal migration.**

state.rs:212-214 documents that a non-terminal migration "must not be overwritten", but no entry point checks `get_migration`/`is_terminal`, the `put_migration` trait doc carries no non-overwrite contract, `CommitError` has no already-in-progress variant, and both the wallet adapter and the test stores overwrite blindly. Given the crate's own crash-resume design, a re-plan-and-recommit after a crash silently orphans already-broadcast pre-signed transactions (their txids are lost, so `mark_mined` can never run) and produces a new layer 0 double-spending the same notes.

Fix direction: check `get_migration` at the top of `commit_preparation_inner` and return a typed `CommitError::MigrationInProgress` unless the existing state is terminal.

zcash_pool_migration_backend/src/state.rs
```diff
@@ -0,0 +1,723 @@
...
+    pub fn transaction_statuses(&self, target_height: BlockHeight) -> Vec<TransactionStatus> {
+        self.transactions
+            .iter()
+            .map(|t| {
+                let deps_ok = self.deps_mined(&t.depends_on);
+                // `Planned`/`Expired` need building; `Signed`/`Proved` need broadcasting. In both cases
+                // a transaction is actionable only once its dependencies (the prior anchor bucket) are
+                // mined, and a broadcastable one only once it is also due.
+                let (ready, action, blocked_on) = match t.state {
+                    MigrationTxState::Planned | MigrationTxState::Expired => {
```
nuttycom: **[high] `Expired` is a documented-but-dead trap state: statuses advertise it as `BuildAndSign`-ready while every driver path ignores it.**

The `update_transaction` trait doc (engine.rs:113-114) explicitly invites the consumer to set it ("…or it expires") and the variant doc promises "to be rebuilt", but `has_unbuilt_transfer` (state.rs:287), `ready_prep_layer` (state.rs:154), `next_broadcastable` (state.rs:176), and both commit functions (engine.rs:1274, 1056/1086) match only `Planned`. A consumer following the docs lands in a permanent, self-contradictory deadlock: `transaction_statuses` says build, `next_step` returns `Waiting` forever (never `Complete` — state.rs:296 requires all `Mined`), commits no-op, and the migration can neither complete nor be replaced.

Fix direction: either drop the variant until the rebuild-on-expiry slice lands, or wire it consistently through `next_step` and the commit paths — but not the current halfway state.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2142 @@
...
+            id: MigrationTxId(next_id),
+            kind: MigrationTxKind::Transfer { crossing },
+            pczt: None,
+            depends_on: last_layer_ids.clone(),
+            scheduled_height: schedule.broadcast_height(),
+            expiry_height: schedule.expiry_height(),
+            anchor_boundary: Some(
+                scheduling::draw_anchor_boundary(
+                    nu63_activation,
+                    target_height,
```
nuttycom: **[medium] At commit time each transfer's persisted `anchor_boundary` is drawn with `target_height` as the funding-creation floor, admitting boundaries that predate the funding notes' existence.**

`plan_migration` floors the same draw at `est_last_prep_height` (engine.rs:555-573), but here the draw uses the commit build height, so it samples a strict superset that includes boundaries before the last prep layer mines (~33-50% of draws for the earliest transfers; the test at engine.rs:1737-1740 only asserts `b >= boundary_at_or_after(TARGET_HEIGHT)`). In this PR the field is never read for witnessing (see the comment at engine.rs:1244), but its doc says it is "the boundary height whose tree state the transaction proves against" and `MigrationCrypto` promises re-anchoring to the drawn boundary at proving time (engine.rs:602) — impossible to honor for a boundary predating the funding note, so the deferred proving slice inherits unusable recorded anchors.

Fix direction: floor the commit-time draw at the last prep layer's estimated mining height, mirroring `plan_migration`.

zcash_pool_migration_backend/src/wallet.rs
```diff
@@ -0,0 +1,389 @@
...
+        }
+    }
+
+    fn resolve_wallet_note(
+        &self,
+        index: usize,
+        anchor_height: BlockHeight,
+    ) -> Result<(OrchardNote, MerklePath), Self::Error> {
+        let notes = self.spendable_orchard()?;
+        let &(note, position, _) = notes.get(index).ok_or(Error::NoteNotFound(index))?;
```
nuttycom: **[medium] Note identity crosses the backend seam as a bare index or bare value with no cross-check, so a spendable-set change during the consent gap resolves the wrong note.**

The engine discards the stored `value` when resolving `PrepInput::Wallet { index, .. }` (engine.rs:862) and this adapter blindly does `notes.get(index)` against a freshly re-queried, position-sorted spendable list whose membership can change during the unbounded plan-to-commit gap; `resolve_funding_notes` (wallet.rs:300-303) matches by bare `u64` equality against ALL spendable notes; and the builder's `placed` output-location data that would pin identity is discarded at engine.rs:879/1130. Exact-balance enforcement in `build_prep_tx` makes ordinary staleness fail-safe but opaque (an unexplained balance error with no stale-plan indication); the silent residue is a privacy defeat (a transfer spending a coincidental equal-value unprepared note) and rare mid-flight wedges after layer-0 fees are sunk.

Fix direction (minimal): verify the resolved note's value against the plan's stored value and return a typed stale-plan error on mismatch. Full (txid, action-index) provenance is a bigger follow-up.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -0,0 +1,2142 @@
...
+            u32::from(target_height),
+            &fvk,
+            anchor,
+            spends,
+            &outputs,
+            &mut *rng,
+        )
+        .map_err(|e| CommitError::Build(format!("{e}")))?;
+
+        let signed = backend.sign(pczt).map_err(CommitError::Backend)?;
```
nuttycom: **[medium] `commit_pending_preparation` signs in-process unconditionally, so an external-signer backend with a multi-layer plan dead-ends at layer 1.**

`build_preparation_unsigned` happily commits multi-layer plans (later layers are recorded as `Planned` placeholders regardless of signing mode, engine.rs:911-931), `AdvanceStep::BuildPreparationLayer` directs consumers only here (state.rs:34-35), and `apply_signature` matches only `AwaitingSignature` while deferred layers go straight to `Signed` (engine.rs:1148). So a watch-only/hardware backend that starts a 2-layer migration via the unsigned path gets `backend.sign()` called on a backend with no spend authority after layer 0 mines — the only escape is cancelling, with funds already consolidated by layer 0. The limitation is acknowledged only in the CHANGELOG, not in the API docs, and nothing rejects multi-layer plans on the external path.

Fix direction: either reject multi-layer plans in `build_preparation_unsigned` with a typed error until unsigned deferred layers exist, or provide the unsigned variant of this function.

nuttycom (approved):
utACK 0c2775f5ed1561360b164b3971ecf0cf734e75eb

dannywillems:
As discussed in the last meeting this morning/night, I am merging this patch, and I will go through each change to review/suggest changes in follow-up PRs/issues.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/state.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/wallet.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/wallet.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/wallet.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/wallet.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/wallet.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/state.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/scheduling.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Addressed on main; tracked in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Addressed on main; tracked in #2701.

zcash_pool_migration_backend/src/state.rs
dannywillems: Addressed on main; tracked in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/wallet.rs
dannywillems: Follow-up in #2701.

zcash_pool_migration_backend/src/engine.rs
dannywillems: Follow-up in #2701.

PR #2666, opened 2026-07-18
zcash_client_sqlite: don't mark never-exposed ZeWIF addresses as exposed

### LINKED ISSUE #2668: Zcashd migration over-eagerly marks addresses as exposed, causing gap limit issues
See original report here: https://github.com/zcash/zallet/issues/637

nullcopy:
Closes https://github.com/zcash/librustzcash/issues/2668
[COR-1601](https://linear.app/zodl/issue/COR-1601/zcashd-migration-over-eagerly-marks-addresses-as-exposed-causing-gap)

### DIFF
--- a/zcash_client_sqlite/src/zewif.rs
+++ b/zcash_client_sqlite/src/zewif.rs
@@ -524,6 +524,10 @@ pub struct ZewifImportReport {
     /// wallet does not recognize as receivers of any imported account, and so
     /// could not be marked as exposed.
     pub addresses_not_recognized: usize,
+    /// The number of transparent addresses recorded in the document with no
+    /// exposure height and appearing in none of the document's transactions,
+    /// and so deliberately left unexposed.
+    pub addresses_never_exposed: usize,
     /// The number of transactions from the document that were stored because
     /// they involve the imported accounts.
     pub transactions_stored: usize,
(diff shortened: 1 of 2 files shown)


zcash_client_sqlite/src/zewif.rs
```diff
@@ -524,6 +524,15 @@ pub struct ZewifImportReport {
     /// wallet does not recognize as receivers of any imported account, and so
     /// could not be marked as exposed.
     pub addresses_not_recognized: usize,
+    /// The number of transparent addresses recorded in the document with no
+    /// exposure height and appearing in none of the document's transactions,
+    /// and so deliberately left unexposed. These are addresses that were never
+    /// handed out (for example zcashd keypool reserves); marking them exposed
```
nullcopy: Only the first sentence is useful here. Drop the rest.

zcash_client_sqlite/src/zewif.rs
```diff
@@ -1142,12 +1193,20 @@ fn import_transactions<DbT, P, S>(
...
     report: &mut ZewifImportReport,
-) -> Result<(), ZewifImportError<S>>
+) -> Result<std::collections::HashSet<TransparentAddress>, ZewifImportError<S>>
 where
     DbT: WalletWrite<AccountId = AccountUuid, Error = SqliteClientError>,
     P: consensus::Parameters,
     S: std::error::Error,
 {
+    // The transparent addresses that appear as recipients in the document's
+    // transactions. Any such address has been exposed on-chain, regardless of
```
nullcopy: What if a stored transaction is an outgoing payment, and the change address is wallet owned. Will this logic also detect those addresses and mark them as exposed?

zcash_client_sqlite/src/zewif.rs
```diff
@@ -1100,6 +1125,27 @@ where
...
+                    // document's transactions, it was exposed on-chain: importing
+                    // those transactions has already marked it exposed at its true
+                    // mined height, so it must not be treated as a never-exposed
+                    // reserve. An address that appears in no transaction is a
+                    // keypool reserve that was never handed out; leave it
+                    // unexposed so it does not consume the gap limit. It stays a
+                    // tracked receiver and will be marked exposed if and when it
+                    // is reserved or later observed on-chain.
+                    if !tx_transparent_recipients.contains(&taddr) {
+                        report.addresses_never_exposed += 1;
```
nuttycom: This isn't sufficient; it needs to take into account the gap limit. If any address at a higher index has been exposed, this address should also be considered to have been exposed.

nuttycom (approved):
utACK

PR #2667, opened 2026-07-18
zcash_pool_migration_backend: split persistence into store traits

dannywillems:
Stacked on #2663 (base: `feat/pool-migration-engine`). Part of the Orchard ->
Ironwood pool migration (ZIP 318). Merge #2645 -> #2647 -> #2661 -> #2663 first.

First step of reshaping the migration engine to integrate the way every other
wallet operation does - the `zcash_client_backend` (traits + logic) /
`zcash_client_sqlite` (tables) split - so a consuming app (zallet, or any other)
uses it the way it already builds and stores transactions.

This PR extracts the migration PERSISTENCE from `MigrationBackend` into two
traits, mirroring `zcash_client_backend`'s `WalletRead` / `WalletWrite`:

- `PoolMigrationRead::get_migration` loads the in-progress migration.
- `PoolMigrationWrite::put_migration` records a committed migration (its
  pre-signed PCZTs + schedule + per-transaction lifecycle state), and
  `update_transaction` advances a transaction's state.

A store implements these over its own tables; `MigrationBackend` keeps only the
planning queries. `commit_preparation` / `commit_transfers` now require the store
traits alongside the crypto. The state types (`MigrationState` /
`MigrationTransaction`) are the traits' data model, unchanged.

Follow-up (one PR, per the plan): a new `zcash_pool_migration_sqlite` crate
implements these traits over SQLite tables (which `zcash_client_sqlite` registers
into `WalletDb`), and the build path moves onto the wallet's own
`WalletRead` / `InputSource` / `WalletCommitmentTrees` (replacing `MigrationCrypto`).
Design + the concrete cycle/serialization recipe: `migration-engine-b-design.md`.

61 / 69 tests.

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -44,9 +44,9 @@ use crate::preparation::{PrepError, PreparationPlan, plan_preparation};
 use crate::scheduling::{self, Schedule};
 
 /// What the migration engine needs from a wallet to PLAN a migration: the account's spendable notes and
-/// the chain state. Later slices add the methods for building (note witnesses, viewing keys), signing,
-/// persistence, and reconciliation; a backend implements this trait over its own note store and chain
-/// view (for example a `zcash_client_backend` wallet, or the migration's own SQLite store).
+/// the chain state. Following the `zcash_client_backend` pattern, a later slice replaces this with the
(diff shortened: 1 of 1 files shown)


dannywillems:
Folded into #2663: the persistence store-trait split was a 60-line refactor of a trait #2663 itself introduces, with no consumer in between, so it now lands as a commit within the engine PR. #2669 has been retargeted onto feat/pool-migration-engine.

PR #2669, opened 2026-07-18
zcash_pool_migration_sqlite: SQLite persistence for pool migrations

dannywillems:
Based on `main` (the engine stack #2645 -> #2647 -> #2661 -> #2663 and the
`WalletMigration` wallet adapter have all merged). This PR is the SQLite
persistence layer for the pool-migration engine, in the shape every other wallet
operation uses: the `zcash_client_backend` (traits + logic) / `zcash_client_sqlite`
(tables) split. It implements `zcash_pool_migration_backend`'s `PoolMigrationRead`
/ `PoolMigrationWrite` over SQLite, so a `zcash_client_sqlite` wallet persists and
resumes an in-progress migration from its own `wallet.db`.

## New crate `zcash_pool_migration_sqlite`

A committed migration is its pre-signed PCZTs plus schedule and lifecycle state,
stored as one migration row and one row per transaction, so a wallet resumes
after being closed or restarted.

Structure (one generic store, one public submodule per pool):

- The generic, pool-agnostic machinery (the DDL builders, the blob encode/decode
  helpers, and the `Store` type carrying the read/write/update logic) lives in a
  private `store` module, parameterized over a `Tables` descriptor of per-pool
  table/index names. The generic type never appears in the public API.
- Each pool migration is a public submodule that instantiates the store with its
  own table names. Currently the only one is `orchard_ironwood` (the Orchard ->
  Ironwood migration, ZIP 318), the crate's only public surface:
  `zcash_pool_migration_sqlite::orchard_ironwood::{PoolMigrations, MIGRATION_ID, init_migration_tables, Error}`,
  over tables `orchard_ironwood_migrations` / `orchard_ironwood_migration_transactions`
  indexed by `(state, scheduled_height)`.
- This lets future pool migrations reuse the same machinery under their own
  tables by adding a sibling submodule, addressing the review comments asking the
  pools to be reflected in the tables and surface (the pool name lives in the
  module path and the table names, keeping the crate the generic home).

## `zcash_client_sqlite` registration

`zcash_client_sqlite` depends on this crate (never the reverse: the store crate
owns the DDL + a `MIGRATION_ID` and never names `zcash_client_sqlite`). It
registers a thin `schemerz` migration that runs
`orchard_ironwood::init_migration_tables`, depends on `ironwood_received_notes`,
maps the error, and exposes the store through `WalletDb`, so the tables live in
the same `wallet.db` and share its schema versioning. The golden schema mirror in
`db.rs` tracks the installed tables.

## Tests

- `zcash_pool_migration_sqlite`: 7/7 (public round-trip through
  `orchard_ironwood::PoolMigrations`, plus the internal preparation-codec
  round-trip).
- `zcash_client_sqlite`: `verify_schema` green (the golden mirror matches the
  store DDL, including the renamed tables); crate builds clean.
- fmt + clippy (`-D warnings`) clean; `cargo doc` clean.


zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -33,6 +33,7 @@ mod nullifier_map;
 mod orchard_note_version;
 mod orchard_received_notes;
 mod orchard_shardtree;
+mod pool_migration_tables;
```
dannywillems: This is actually a ironwood related. We should mirror this in the name.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -589,6 +589,49 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
     transaction_id ASC
 )";
 
+/// Stores the note-split decomposition and status of the single in-progress Orchard -> Ironwood
```
dannywillems, line 592: We should mirror the pools name in the table.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
+//! SQLite persistence for the Orchard -> Ironwood value-pool migration (ZIP 318).
```
dannywillems: The pools should be renamed in the crate name.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+    MigrationState, MigrationStatus, MigrationTransaction, MigrationTxId, MigrationTxKind,
+    MigrationTxState, PoolMigrationRead, PoolMigrationWrite,
+};
+use zcash_pool_migration_backend::note_splitting::NoteSplitPlan;
+use zcash_pool_migration_backend::preparation::{
+    PrepInput, PrepOutput, PrepTransaction, PreparationPlan,
+};
+
+/// The `schemerz` migration id for the pool-migration tables. `zcash_client_sqlite` registers its
+/// (thin) migration under this id, depending on `ironwood_received_notes::MIGRATION_ID`.
```
dannywillems: Should it depend on more than just that migration?

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+/// so it is stored as one row; a future multi-account model would replace this with an account key.
+pub const SINGLETON_ID: i64 = 0;
+
+/// DDL for the `pool_migrations` table: the note-split decomposition and overall status of the single
+/// active migration. The `*_values` / `funding_notes` columns hold little-endian `u64` arrays (see
+/// [`encode_u64s`]); the `preparation` column holds the tagged encoding of the preparation plan (see
+/// [`encode_preparation`]), retained so deferred preparation layers can be rebuilt after their prior
+/// layer mines; zatoshi and height scalars fit in SQLite's signed 64-bit integer.
+pub const CREATE_POOL_MIGRATIONS: &str = "
+    CREATE TABLE IF NOT EXISTS pool_migrations (
```
dannywillems: The pool names should be reflected.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+        preparation BLOB NOT NULL
+    )";
+
+/// DDL for the `pool_migration_transactions` table: one row per migration transaction, its pre-signed
+/// PCZT (`pczt`, `NULL` until built/signed under two-phase signing), its dependency graph
+/// (`depends_on`, a little-endian `u32` array of transaction ids), schedule, and lifecycle `state`.
+/// `kind` is `'preparation'` (with `layer`/`tx_index`) or `'transfer'` (with `crossing`); `state`
+/// carries `txid`/`mined_height` for the `broadcast`/`mined` states.
+pub const CREATE_POOL_MIGRATION_TRANSACTIONS: &str = "
+    CREATE TABLE IF NOT EXISTS pool_migration_transactions (
```
dannywillems: The pool names should be reflected.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+/// that are due to prove or broadcast without scanning the table.
+pub const CREATE_TX_DUE_INDEX: &str = "
+    CREATE INDEX IF NOT EXISTS idx_pool_migration_tx_due
+        ON pool_migration_transactions (state, scheduled_height)";
+
+/// Create the pool-migration tables (and the due-transaction index) on `conn`. This is the body a
+/// `zcash_client_sqlite` `schemerz` migration's `up()` calls; it is idempotent (`IF NOT EXISTS`).
+pub fn init_migration_tables(conn: &Connection) -> rusqlite::Result<()> {
+    conn.execute_batch(&format!(
+        "{CREATE_POOL_MIGRATIONS};\n{CREATE_POOL_MIGRATION_TRANSACTIONS};\n{CREATE_TX_DUE_INDEX};"
```
dannywillems: Is it the common way to achieve a multi-table creation?

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+    }
+}
+
+impl From<rusqlite::Error> for Error {
+    fn from(e: rusqlite::Error) -> Self {
+        Error::Db(e)
+    }
+}
+
+/// The pool-migration store: a [`PoolMigrationRead`] / [`PoolMigrationWrite`] over a
```
dannywillems: There might be better way to achieve it. The whole code is mostly wrappers. Not fully happy with the result.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+            ":funding_notes": encode_u64s(&state.funding_notes),
+            ":preparation": encode_preparation(&state.preparation),
+        },
+    )?;
+
+    for mtx in &state.transactions {
+        let (kind, layer, tx_index, crossing) = encode_tx_kind(mtx.kind);
+        let (tx_state, txid, mined_height) = encode_tx_state(&mtx.state);
+        tx.execute(
+            "INSERT INTO pool_migration_transactions
```
dannywillems: The table name should reflect the pools.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+
+    fn store() -> PoolMigrations<Connection> {
+        let conn = Connection::open_in_memory().expect("in-memory db");
+        init_migration_tables(&conn).expect("create tables");
+        PoolMigrations::new(conn)
+    }
+
+    /// A two-layer preparation plan exercising every input tag (Wallet, Prior), every output tag
+    /// (Funding, Intermediate, Change), multiple layers, and a direct-funding note.
+    fn sample_preparation() -> PreparationPlan {
```
dannywillems: We should share these with the ones available in the `zcash_pool_migration_backend` crate.

dannywillems: And these could be proptest.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+        }
+    }
+
+    #[test]
+    fn get_migration_empty_is_none() {
+        assert!(store().get_migration().expect("read").is_none());
+    }
+
+    #[test]
+    fn put_then_get_round_trips() {
```
dannywillems: All of these tests could be shared per backend. We could have a single test framework.

zcash_pool_migration_sqlite/src/lib.rs
```diff
@@ -0,0 +1,884 @@
...
+        let mut trailing = encoded.clone();
+        trailing.push(0);
+        assert!(matches!(
+            decode_preparation(&trailing),
+            Err(Error::Corrupt("preparation.trailing"))
+        ));
+    }
+
+    #[test]
+    fn put_replaces_previous_migration() {
```
dannywillems: Should we call `put` `replace` then? Or should we add a comment on the method to mention this property? Should be a method in the test framework.

dannywillems (commented):
Blocking as we should reflect the pool names in the crate but also in the tables. Otherwise, LGTM.

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,425 @@
...
+            anchor_boundary,
+            state,
+        ));
+    }
+    Ok(out)
+}
+
+/// Persist `state` as the single active migration in the tables named by `t`, replacing any existing
+/// one. Runs inside the caller's transaction so the replacement is atomic.
+fn write_migration(
```
dannywillems: The name and/or the return type should mention that it updates/replaces.

zcash_client_sqlite/src/wallet/init/migrations.rs
dannywillems: Done: the schemerz migration module is now `orchard_ironwood_migration_tables` (75d2f1b).

zcash_client_sqlite/src/wallet/db.rs
dannywillems: Done: the tables are now `orchard_ironwood_migrations` / `orchard_ironwood_migration_transactions`, and the golden schema mirror in `db.rs` matches (`verify_schema` passes).

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Addressed by making the pool a public submodule rather than renaming the crate: the generic store machinery is kept crate-internal, and `zcash_pool_migration_sqlite::orchard_ironwood` is the only public surface, over the `orchard_ironwood_*` tables. So the pool name lives in the module path and the table names, and the crate stays the generic home for future pool submodules.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Checked the dependency graph: the two tables have only the internal `migration_id` FK and the store touches only its own two tables, so no additional dependency is required for schema correctness. `ironwood_received_notes` is kept as the single frontier dependency because its DAG closure transitively covers the Orchard source and Ironwood destination note/shardtree tables the engine needs (verified the reachability). Added a comment on `DEPENDENCIES` documenting this (bb487d09a).

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Done: reflected as `orchard_ironwood_*`.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Done: reflected as `orchard_ironwood_*`.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Confirmed idiomatic: `execute_batch` with concatenated CREATE statements is the standard multi-table pattern here (51 migrations use it, including `ironwood_received_notes`). No change.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Partly addressed: the store is now a generic `store` core plus a public `orchard_ironwood` facade, and the byte (de)serialization moved onto the backend types (canonical `write`/`read` via `zcash_encoding`), leaving only the SQL column-mapping here, which removes most of the wrapper feel. Leaving this open for you to judge whether the new structure is enough.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Done: `orchard_ironwood_migration_transactions`.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Done: the fixtures are now shared `arb_*` strategies in `zcash_pool_migration_backend::testing` (behind the `test-dependencies` feature), reused here instead of duplicated.

dannywillems: Done: the store round-trips are now proptests over `arb_migration_state`.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Done: a shared store-conformance suite (`assert_empty_is_none` / `assert_put_get_roundtrip` / `assert_put_replaces` / `assert_update_transaction`) lives in `zcash_pool_migration_backend::testing`, generic over `PoolMigrationRead`/`PoolMigrationWrite`, so every store (this one, and the in-memory backend) runs the same suite.

zcash_pool_migration_sqlite/src/lib.rs
dannywillems: Done: renamed to `replace_migration` (the `PoolMigrationWrite` trait method and every impl) (c45a638).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done: renamed the internal helper to `replace_migration` too (c45a638).

dannywillems:
Closing. Cherry-picked in multiple patches, and merged.

PR #2672, opened 2026-07-20
Document database write atomicity requirements for agents

dannywillems:
## Motivation

Two atomicity defects were found by audit in Zallet, which writes to the wallet database through this crate's API: an age-recipient insert loop that autocommitted per row behind a one-shot guard, and `z_importkey` writing the keystore and the wallet account in two separate transactions. Both are fixed. This adds the check to `AGENTS.md` so the class is caught when the next such code is written rather than after.

## Approach

The section is a procedure rather than an exhortation. A rule saying "make writes atomic" would not have caught either defect, because the property that made them unrecoverable is second-order: what a retry does after a partial commit, and whether a one-shot guard or an asymmetric read path turns that partial state into something permanent. Those are now questions to answer in the PR description whenever an operation issues more than one write.

Beyond the checklist it records:

- which primitive to reach for (`transactionally`, `transactionally_with_extension`), and that intra-crate write helpers should take a `&Transaction` rather than opening their own, since a helper that opens its own can never be composed atomically;
- that a `rusqlite` transaction cannot be held across an `.await`, so an API forcing async work into the middle of a write sequence cannot be used atomically;
- that `WalletWrite` trait methods with multi-part updates should state their atomicity requirement in rustdoc, as `apply_tree_changes` does, since implementors and callers cannot otherwise see it;
- a test-evidence requirement, pointing at `transactionally_with_extension_rolls_back_on_error` as the template.

## Two hazards worth a second look

Both are invisible at the call site, and both are now written down:

1. There are no savepoints anywhere in `zcash_client_sqlite/src/lib.rs`. The `WalletDb<SqlTransaction<'_>>` implementations take none per method, so rollback depends entirely on the error propagating out of the `transactionally` closure. A discarded error (`let _ = ...`, `.ok()`, a `match` arm that continues) commits partial state, and the enclosing `transactionally` still returns `Ok`.
2. Atomicity does not compose across calls: `WalletWrite` and `WalletCommitmentTrees` on `WalletDb<C: BorrowMut<Connection>>` each open their own transaction, so two sequential calls are two commits with an inconsistent window between them.

## Testing

Documentation only; no code change.

## Related

Context, not closed by this PR: zcash/zallet#615, zcash/zallet#616, and #2638 which added `transactionally_with_extension`. A companion PR adds the caller-facing version of this section to Zallet's `AGENTS.md`: zcash/zallet#658.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -308,6 +308,96 @@ Type safety is paramount. This is a security-critical codebase.
 - **`testing` submodules**: Exposed via `test-dependencies` feature for cross-crate
   test utilities (proptest strategies, mock implementations).
 
+## Database Write Atomicity (`zcash_client_sqlite`)
+
+`rusqlite` autocommits every statement executed outside an explicit transaction. A
+sequence of writes that must land together will therefore commit one at a time unless
+it is wrapped, and a failure partway through (an I/O error, a full disk, a crash, or an
+error returned by a later step) leaves a subset committed. Downstream wallets have
+shipped real bugs of this shape against this crate's API, so it is checked on every
+change that writes to the wallet database.
+
+**Before writing or reviewing any code that writes to the wallet database, answer these
+four questions in order.** Answer them in the PR description when the answer to the
+first is "more than one".
+
+1. **How many writes does this operation issue?** Count every `INSERT`, `UPDATE`, and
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
ACK

PR #2673, opened 2026-07-20
zcash_pool_migration_memory: extract the in-memory test backends into a crate

dannywillems:
Extracts the in-memory mock backends (`MockBackend`, `CommitMock`, `LayeredMock`)
and the witness/key test helpers (`spending_key`, `regtest_network`,
`single_note_witness`, `shared_anchor_witnesses`, `prep_fee`) out of
`zcash_pool_migration_backend`'s `#[cfg(test)]` modules into a new, shared,
**test-only** crate `zcash_pool_migration_memory`, so other crates can reuse
them. The crate's docs state prominently that it is for testing only and not for
production.

## Notes for reviewers

- **Unit tests -> integration tests.** A dev-dependency cycle rebuilds the
  backend library under `cfg(test)` as a distinct crate instance, so mock trait
  impls built against the plain library do not satisfy the test binary's
  same-named traits. The mock-consuming engine tests therefore move from unit
  tests into integration tests under `zcash_pool_migration_backend/tests/`, which
  link the same library instance as the memory crate. This mirrors how
  `zcash_client_memory` is consumed. The build-module tests are unaffected (their
  helpers cross only shared `orchard`/`pczt` types) and re-export the moved
  helpers.
- **One public API addition:** `MigrationPlan::from_parts`, because a relocated
  test hand-builds a plan. It mirrors the existing `PreparationPlan::from_parts`
  and `NoteSplitPlan::from_stored_parts`. No production behavior changes; the test
  count is unchanged (94).

Based on #2663.


dannywillems:
Closing as we are consolidating in the SQLite PR.

PR #2674, opened 2026-07-20
zcash_pool_migration_backend: test reconciliation on a many-equal-note source

dannywillems:
The reconciliation loop in `plan_migration` drops the smallest funding notes
until the preparation fits, but no unit test exercised a source shape that forces
a drop; the general planning test uses a balance whose whole split is fundable, so
the drop path was only covered indirectly by an integration test.

Adds a test for the "exchange" shape: ten identical 5-ZEC notes (50 ZEC) whose
greedy split has a small tail the equal source notes cannot fund. It asserts
reconciliation actually drops for this shape, drops only the smallest
denominations (a sub-multiset of the split, from the bottom), never creates value,
and leaves a preparation that is fundable from the source notes.

Based on #2663.

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -1318,6 +1318,83 @@ mod tests {
         ));
     }
 
+    /// Reconciliation on a many-equal-note source (the "exchange" wallet shape). Ten identical 5-ZEC
+    /// notes (50 ZEC) decompose into `[20, 20, 5, 2, 2, 0.5, ...]`, but equal source notes cannot fund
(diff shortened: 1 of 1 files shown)


zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1128,6 +1128,69 @@ mod tests {
         ));
     }
 
+    /// Reconciliation on a many-equal-note source (the "exchange" wallet shape). Ten identical 5-ZEC
```
dannywillems, line 1321: It is an interesting case.

dannywillems:
Closing as we are consolidating in the SQLite PR.

PR #2675, opened 2026-07-20
zcash_pool_migration_backend: address the note-split review comments

dannywillems:
Addresses nuttycom's review on #2645 (the note-split slice, already merged).

- `CanonicalOneTwoFive::new` takes `&impl FeePolicy` instead of `&dyn FeePolicy`.
  It is a constructor, not a trait-object method, so static polymorphism (as the
  rest of the codebase uses) fits.
- `DenominationStrategy::plan` is generic over `R: RngCore` instead of taking
  `&mut dyn RngCore`. The trait is no longer object-safe, and the aspirational
  `Box<dyn DenominationStrategy>` note is dropped; a trait object can be
  reintroduced when a strategy is actually selected at runtime.
- `NoteSplitPlan` no longer stores the migration outputs and crossing values as
  two parallel vectors. Since every prepared note is exactly its crossing value
  plus one constant fee buffer, the plan stores the crossing values and that
  buffer and derives `migration_outputs()` on demand, making the
  `output = crossing + buffer` invariant structural rather than something the two
  vectors could drift from.
- Documents why the note split reserves a single constant prep fee (the fee for a
  note-split transaction padded to the maximum split action count) rather than a
  fee rule: the migration pads that transaction to the maximum anyway.

Tests pass; fmt and clippy clean.

### DIFF
--- a/zcash_pool_migration_backend/src/note_splitting.rs
+++ b/zcash_pool_migration_backend/src/note_splitting.rs
@@ -184,28 +184,23 @@ impl FeePolicy for Zip317FeePolicy {
 /// cross the turnstile, and the residual kept in the source pool. Produced by a
 /// [`DenominationStrategy`].
 ///
-/// [`migration_outputs`](Self::migration_outputs) and [`crossing_values`](Self::crossing_values) are
-/// PARALLEL (index `i` describes the same prepared note) and differ by exactly the per-note fee
-/// buffer (see [`FeePolicy::transfer_fee_buffer_zatoshi`]), so for every `i`:
-///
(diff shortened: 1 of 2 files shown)


dannywillems (commented):
Addressing comments from 2645. This will require changes in other open PRs. I would be in favor of merging it after the PR train is merged.

nuttycom (approved):
utACK

dannywillems:
> Addressing comments from 2645. This will require changes in other open PRs. I would be in favor of merging it after the PR train is merged.

Well. Too late... 😄

PR #2678, opened 2026-07-20
zcash_pool_migration_backend: fix the test build after migration_outputs change

dannywillems:
main's `cargo test --all-features -p zcash_pool_migration_backend` currently
fails to compile. Two independently-green PRs collided:

- #2647 added `build/end_to_end.rs`, which does
  `let funding = split.migration_outputs(); plan_preparation(&[balance], funding, ...)`.
- The note-split review change made `NoteSplitPlan::migration_outputs()` return
  an owned `Vec<u64>` (it now derives the prepared-note values from the crossing
  values plus the fee buffer) instead of a borrowed `&[u64]`.

`plan_preparation` takes `&[u64]`, so passing the `Vec` no longer compiles.
One-character fix: borrow it (`&funding`). Verified `cargo test --all-features`
is green again (46 passed).


nuttycom (approved):
utACK

PR #2695, opened 2026-07-21
zcash_primitives, pczt: defer shielded anchors to proving time (ZIP 374)

nuttycom:
Depends on zcash/orchard#534, pinned here via `[patch.crates-io]` until it lands in an orchard release (per the usual policy, a merge non-blocker but a publish blocker).

For V6 (NU6.3) transactions neither the txid nor the sighash commits to a shielded anchor (ZIP 374 moved them to the authorizing-data digest), and witnesses are pure proving inputs — but the transaction builder demanded every Orchard-family anchor in its `BuildConfig` and a rooting Merkle witness per spend at add-spend time. Flows that pre-sign transactions long before proving them — notably the ZIP 318 pool-migration transfers (#2663), whose proving anchor is a bucketed boundary drawn near each transfer's future broadcast height — were forced to witness against whatever tree state existed at build time, or to wait after their funding notes mined for an anchor they do not actually need at signing.

This adds `zcash_primitives::transaction::builder::DeferredPcztBuilder`, a builder crafted exclusively for that flow rather than a mode grafted onto the general-purpose `Builder` (whose API is unchanged):

- V6-only by construction: `new` fails under any earlier format, whose signatures commit to anchors.
- Takes no anchors at all; spends are added as bare `(fvk, note)` pairs through orchard's new deferred-anchor support (`Builder::new_with_anchor_deferred` / `add_spend_unwitnessed`).
- Covers exactly the two Orchard-family pools: their nullifiers are rho-based, so a spend can be signed before its witness is known; a Sapling nullifier commits to its note's position, so Sapling can never defer, and is excluded along with transparent inputs.
- The emitted PCZT carries ABSENT anchor and witness fields for the deferred bundles (`pczt`'s serialization honors orchard's deferral flag; padding dummies keep their witnesses, which the prover requires and the consistency check exempts). From there the existing ZIP 374 machinery takes over unchanged: IoFinalizer and Signer already operate with absent V6 anchors, the Updater's `set_{orchard,ironwood}_anchor` / `set_*_spend_witnesses` install the real values after signing, and the Prover enforces witness-to-anchor consistency — the same check an anchored build performs at add-spend time, relocated to where the real anchor first exists.

The end-to-end test builds an Orchard-spend/Ironwood-output V6 transaction with both anchors deferred, signs it, installs the real anchor and witness afterwards, and proves and extracts it, asserting the sighash is unchanged by the installation and the pre-made spend-auth signature still verifies.

Follow-up: the #2663 migration engine's `commit_transfers` adopts this builder, so transfers need no anchor or witness at commit time and the persisted per-transfer `anchor_boundary` becomes purely the proving-time anchor.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3008,8 +3008,7 @@ checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
 [[package]]
 name = "orchard"
 version = "0.15.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cca2ede6a4a77bd729eed3be9303d98a08b217edc85190dcf4dbe004086d5a65"
+source = "git+https://github.com/zcash/orchard.git?rev=3bd2b736d1273226595773265313fc3f3428af16#3bd2b736d1273226595773265313fc3f3428af16"
 dependencies = [
  "aes",
  "bitvec",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,3 +229,9 @@ unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(zcash_unstable, values("nu7"))',
   'cfg(live_network_tests)',
 ] }
+
+[patch.crates-io]
+# ZIP 374 deferred-anchor builder support (orchard PR branch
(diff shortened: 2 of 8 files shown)


nuttycom (commented):
self-utACK

PR #2698, opened 2026-07-21
feat(tor): enable .onion lightwalletd connections via arti

y4ssi:
## Problem

When a wallet app configures a `.onion` address as the lightwalletd endpoint, two things block the connection:

1. The `tor` feature in `zcash_client_backend` does not pull in `arti-client/onion-service-client`, so the onion-service client code is not compiled in. Arti returns \"feature not compiled in\" at runtime.
2. Even with the feature compiled in, `StreamPrefs` defaults to `BoolOrAuto::Auto` for `connect_to_onion_services`, which arti treats as disabled unless the user has explicitly opted in via config. The gRPC connector must set it to `BoolOrAuto::Explicit(true)` to allow the connection.

## Changes

- `zcash_client_backend/Cargo.toml`: add `"arti-client/onion-service-client"` to the `tor` feature
- `zcash_client_backend/src/tor/grpc.rs`: switch from `connect()` to `connect_with_prefs()` with `StreamPrefs::connect_to_onion_services(BoolOrAuto::Explicit(true))`

## Testing

Tested end-to-end on an Android emulator (x86_64, API 34) against a lightwalletd instance exposed via a Tor v3 onion service. The wallet successfully bootstrapped Arti, connected to the `.onion` endpoint, and synced compact blocks. Before this fix, Arti rejected the connection with `allow_onion_addrs disabled in stream preferences`.


zcash_client_backend/src/tor/grpc.rs
```diff
@@ -70,7 +70,12 @@ impl Service<Uri> for HttpTcpConnector {
             let (_, host, port) = parsed?;
 
             debug!("Connecting through Tor to {}:{}", host, port);
-            let stream = client.inner.connect((host.as_str(), port)).await?;
+            let mut prefs = StreamPrefs::new();
+            prefs.connect_to_onion_services(BoolOrAuto::Explicit(true));
```
nuttycom: Is this something that the library user should have more granular control over? If a user of the library explicitly does *not* want to support .onion connections, they don't really have a choice here as the controlling argument is a string.

nuttycom (changes requested):
I think that the option whether or not to connect to .onion services should remain an explicit one for the user of the library; maybe the `HttpTcpConnector` type should carry a `StreamPrefs` as well as the client, and then expose a constructor `HttpTcpConnector::with_onion_capability` that sets the preferences up correctly, in addition to whatever other constructors are exposed?

zcash_client_backend/src/tor/grpc.rs
```diff
@@ -25,6 +25,17 @@ impl Client {
 
         let is_https = http::url_is_https(&endpoint)?;
 
+        let is_onion = endpoint
+            .host()
+            .map(|h| h.ends_with(".onion"))
+            .unwrap_or(false);
```
nuttycom: This still doesn't fully resolve the issue - it's a string match, not an argument to the `connect_to_lightwalletd` function that the caller has access to. The option needs to be propagated out all the way to where the *application developer* gets to choose, so that it's not dependent on the host string.

zcash_client_backend/src/tor/grpc.rs
y4ssi: Updated — refactored twice per your feedback. allow_onion_services is now an explicit bool argument to connect_to_lightwalletd, propagated from the caller rather than inferred from the host string.

Good call on the deeper principle: a user can run Tor in their own environment (system daemon, SOCKS proxy, Tor Browser) and still configure a .onion endpoint. Auto-detecting .onion and forcing the Arti path would break that case — their external Tor setup handles resolution transparently without needing Arti involved. With the explicit bool, the caller decides which world they're in: Arti embedded (true) or external Tor (false/default).

nuttycom (approved):
utACK 6721534b766eeee91b58a943a31a10adbfbb81a6

PR #2704, opened 2026-07-21
zcash_protocol: canonical Zatoshis binary codec

dannywillems:
Adds `Zatoshis::write`/`read` (an unsigned 64-bit little-endian integer over `corez::io`) plus a `to_u64_le_bytes` accessor: the canonical on-wire form, living with the type. `read` maps an out-of-range amount to `InvalidData`.

Extracted from #2669 as an independent, additive change.

PR #2703, opened 2026-07-21
zcash_encoding: round-trip proptest helper

dannywillems:
Adds `testing::check_roundtrip` behind a new `test-dependencies` feature: a generic write-then-read identity helper for codec proptests. It asserts a `write`/`read` pair are exact inverses and that the encoding is stable across a round-trip (catching non-canonical or lossy codecs).

Extracted from #2669 as an independent, additive change.

PR #2705, opened 2026-07-21
zcash_pool_migration: canonical codecs on the migration types

dannywillems:
Extracted from #2669 (B1 of the review split). Adds canonical `write`/`read` binary codecs onto the pool-migration types (`PreparationPlan`, `NoteSplitPlan`, `PrepInput`/`Output`/`Transaction`, `MigrationTxKind`, `MigrationTxId`, `MigrationTxState`), built on `zcash_encoding` + the `Zatoshis` codec, so a store serializes the types through their own methods. Stacked on the `Zatoshis` codec PR (#2704) which is stacked on the `zcash_encoding` helper PR (#2703).

### DIFF
--- a/zcash_pool_migration_backend/src/note_splitting.rs
+++ b/zcash_pool_migration_backend/src/note_splitting.rs
@@ -119,7 +119,9 @@
 
 use alloc::vec::Vec;
 
+use corez::io::{self, Read, Write};
 use rand_core::RngCore;
+use zcash_encoding::{Optional, Vector};
 
 use zcash_protocol::value::{BalanceError, COIN, Zatoshis};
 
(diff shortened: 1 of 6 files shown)


zcash_pool_migration_backend/src/note_splitting.rs
```diff
@@ -342,3 +381,58 @@ pub fn plan_note_split<R: RngCore>(
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    use zcash_encoding::testing::check_roundtrip;
+
+    /// A [`Zatoshis`] amount well within the money supply, for round-trip fixtures.
+    fn arb_zatoshis() -> impl Strategy<Value = Zatoshis> {
```
dannywillems, line 393: These will be extracted later. Claude didn't pick it up correctly from the 2669 patch.

nuttycom (changes requested):
As discussed elsewhere, we should consider whether we actually want to store binary blobs in the database; they're not queryable. For this application I would prefer to have a normalized database structure, so that we can query these.

dannywillems:
Closing, see https://github.com/zcash/librustzcash/pull/2713#issuecomment-5037913455

PR #2706, opened 2026-07-21
zcash_pool_migration: shared store-conformance test framework

dannywillems:
Extracted from #2669 (B2 of the review split). Adds the `testing` module behind `test-dependencies`: reusable `arb_*` strategies and a backend-agnostic `assert_*` store-conformance suite over the `PoolMigrationRead`/`Write` traits, so every store implementation shares one test suite. Stacked on the type-codecs PR (#2705).


zcash_pool_migration_backend/src/testing.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// An arbitrary [`BlockHeight`], within the range heights realistically take.
+pub fn arb_block_height() -> impl Strategy<Value = BlockHeight> {
+    (0u32..5_000_000).prop_map(BlockHeight::from_u32)
+}
+
+/// An arbitrary transaction id (32 random bytes), for the [`MigrationTxState::Broadcast`] payload.
+pub fn arb_txid() -> impl Strategy<Value = TxId> {
+    any::<[u8; 32]>().prop_map(TxId::from_bytes)
+}
```
nuttycom: These proptest generators should be with their types in `zcash_protocol` - in fact, I'd be very surprised if they don't already exist?

nuttycom (changes requested):
This looks good to me with the exception that the generators for `zcash_protocol` types should move to `zcash_protocol` (if they don't already exist; I'm pretty sure they do?) and then this can be rebased on `main` since we're not sure we want the encodings from `codec/pool-migration-type-codecs`

zcash_pool_migration_backend/src/testing.rs
dannywillems: Agreed, some commits haven't been picked up from 2669, like 2d3cc824aa45fa22d075f4a591b68907aa94b5c7

zcash_pool_migration_backend/CHANGELOG.md
```diff
@@ -7,119 +7,4 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
-### Added
```
nuttycom, line 10: We don't need a CHANGELOG entry for the initial release. The CHANGELOG is explicitly to inform users how to adapt to a changed API, and there's no base API that they're changing from until the initial release.

components/zcash_protocol/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub mod local_consensus;
 pub mod memo;
 pub mod value;
 
-mod txid;
+pub mod txid;
```
nuttycom: Instead of making this public, make a `pub use` inside a `testing` module at the root.

nuttycom (dismissed):
utACK d40fe80b2e15b23b6b131d7cdbe0cdd6ca633995

dannywillems (commented):
utACK https://github.com/zcash/librustzcash/commit/d40fe80b2e15b23b6b131d7cdbe0cdd6ca633995

nuttycom (approved):
re-ACK aea4652215f2cf6b671c87fb11fb76114b37a1ee

dannywillems (commented):
utACK aea4652215f2cf6b671c87fb11fb76114b37a1ee

nuttycom:
Weird, the bitrot test failed on the subsequent PR, not this PR, but the failure was here.

PR #2707, opened 2026-07-21
zcash_pool_migration_memory: in-memory test backends crate

dannywillems:
Extracted from #2669 (B3 of the review split). Two atomic, semantically independent commits:

1. `zcash_pool_migration: rename put/write_migration to replace_migration` — a pure rename of the store trait's persist method and its callers, no behavior change.
2. `zcash_pool_migration_memory: extract the in-memory test backends into a crate` — move the `MockBackend`/`CommitMock` test backends and the seed-derived Orchard witness helpers out of the engine crate's inline test module into a dedicated `zcash_pool_migration_memory` test-support crate (like `zcash_client_memory`), so the engine's integration tests and any downstream store reuse the same mocks.

Stacked on the store-conformance framework PR (#2706).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,6 +341,9 @@ jobs:
       - name: Add zcash_primitives as a dependency of the synthetic crate
         working-directory: ./ci-build
         run: cargo add --no-default-features --path ../crates/zcash_primitives
+      - name: Add zcash_pool_migration_backend as a dependency of the synthetic crate
+        working-directory: ./ci-build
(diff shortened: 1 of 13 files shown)


nuttycom (dismissed):
utACK 98aa8738423051e4d4d616ffcb7a5b5c6f2bf4db

dannywillems (commented):
self-ACK

nuttycom (approved):
re-utACK f664d2febe4c9eaf01415b6289f74de666ec5357

Issue #2701, opened 2026-07-21
Pool migration engine: follow-ups from the #2663 review


dannywillems:
Consolidated follow-ups from the review of #2663 (the pool-migration engine PR, now merged): the unresolved inline threads plus the additional findings in nuttycom's [review](https://github.com/zcash/librustzcash/pull/2663#pullrequestreview-4740806985). Each item links back to its source comment. To be worked through one by one; the source threads are resolved and point here.

## Correctness
- [ ] **Transfer anchor/witness timing** - transfers are witnessed at `most_recent_boundary(target)`, with no wait signal, and the drawn `anchor_boundary` is never consulted ([r3619093776](https://github.com/zcash/librustzcash/pull/2663#discussion_r3619093776)). Tied to the retention + proving work (#2700).
- [ ] **`anchor_boundary` floor** drawn with `target_height`, admitting boundaries that predate the funding notes ([r3619093784](https://github.com/zcash/librustzcash/pull/2663#discussion_r3619093784)).
- [ ] **Preparation PCZT expiry** - `build_prep_tx` may not set `with_expiry_height` (preps embed `target+40` yet broadcast beyond it) ([r3619093775](https://github.com/zcash/librustzcash/pull/2663#discussion_r3619093775)).
- [ ] **Note identity across the backend seam** - resolved by bare index/value with no cross-check; a spendable-set change during the consent gap resolves the wrong note ([r3619093787](https://github.com/zcash/librustzcash/pull/2663#discussion_r3619093787)).
- [ ] **External-signer multi-layer dead-end** - `commit_pending_preparation` signs in-process unconditionally ([r3619093792](https://github.com/zcash/librustzcash/pull/2663#discussion_r3619093792)).
- [ ] **Panic on corrupt store** - `engine.rs:1098/1110` bracket-indexes store-loaded values with no bounds check; should return `CommitError` ([review](https://github.com/zcash/librustzcash/pull/2663#pullrequestreview-4740806985)).
- [ ] **Transfer shape/fee decision** - the transfer builds one unpadded Ironwood action (3 actions, 15,000 zat) vs the padded 4-action/20,000-zat shape other ZIP 318 impls produce; diverging risks partitioning the anonymity set. Needs a conscious decision + doc fix ([review](https://github.com/zcash/librustzcash/pull/2663#pullrequestreview-4740806985)).

## Error modeling / API
[...]

PR #2708, opened 2026-07-21
zcash_pool_migration: typed CommitError + transfer-shape doc fix

dannywillems:
Extracted from #2669 — two small, independent backend improvements to the merged pool-migration engine, in one PR:

**Transfer-shape doc fix** — the CHANGELOG described the transfer as `four logical actions / 2 source + 2 destination`; the code builds three (`2 source + 1 destination`, the single unpadded Ironwood output, per ZIP 318). One-line correction.

**CommitError error modeling** (two atomic commits):
- Replace the `plan.prep_schedule()[layer][index]` out-of-bounds *panic* on a malformed/inconsistent stored plan with a typed `CommitError::InconsistentPlan` at the public boundary.
- De-stringify `CommitError::Build(String)`: carry the structured `build::BuildError` in `Build`, add a `Serialize(EncodingError)` variant, and model NU6.3-not-active as `Nu63NotActive`; the re-plan case shares `StalePlan`. Callers can now match on the condition instead of parsing a message.

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -640,6 +640,12 @@ pub enum CommitError<E> {
     /// and possibly already broadcast — transactions, and a rebuilt layer 0 would double-spend
     /// the same wallet notes.
     MigrationInProgress,
+    /// The migration plan is internally inconsistent: two of its parallel structures disagree
+    /// (for example a preparation layer has no matching entry in the preparation schedule), so
+    /// it cannot be committed. A plan assembled through `from_parts` is not validated, so a
+    /// malformed one reaches the commit boundary as this typed error rather than a panic; the
+    /// string names which structure and index disagreed, for diagnosis.
+    InconsistentPlan(alloc::string::String),
 }
 
 #[cfg(feature = "orchard")]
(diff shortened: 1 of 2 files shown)


dannywillems:
See issue https://github.com/zcash/librustzcash/issues/2701.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -626,39 +626,58 @@ pub trait MigrationCrypto {
...
-    /// Building or serializing a transaction failed.
-    Build(alloc::string::String),
+    /// Building a migration transaction failed. Carries the structured builder error.
+    Build(crate::build::BuildError),
+    /// Serializing a built migration PCZT (for storage or an external signer) failed.
+    Serialize(pczt::EncodingError),
+    /// NU6.3 is not active on this network, so there is no destination pool to migrate into. The
+    /// planning side models the same recoverable condition as
+    /// [`MigrationError::Nu63NotActive`](MigrationError::Nu63NotActive).
+    Nu63NotActive,
```
dannywillems, line 636: I realize that these might be used later in 2669... 😕

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -758,15 +777,11 @@ where
     match signing {
         Signing::InProcess => {
             let signed = backend.sign(pczt).map_err(CommitError::Backend)?;
-            let bytes = signed
-                .serialize()
-                .map_err(|e| CommitError::Build(format!("serialize: {e:?}")))?;
+            let bytes = signed.serialize().map_err(CommitError::Serialize)?;
```
nullcopy, line 780: (nb) I wonder if we should consider something like [thiserror](https://docs.rs/thiserror/latest/thiserror/) to simplify error mapping

nullcopy (approved):
utACK c1386234a202dc37dc1c380d1318ad245e6dbdf8

PR #2709, opened 2026-07-21
zcash_pool_migration: release each crossing on its own funding note

dannywillems:
Extracted from #2669. A pool-crossing transfer's dependency set was the whole last preparation layer, so a crossing could not broadcast until *every other* crossing's preparation had mined. Record the preparation transaction that mints each funding note as that note's producer, and depend each transfer only on its own producer (or nothing, for a direct-funding wallet note); a crossing then releases at its scheduled height as soon as its own funding note is on-chain, independently of the others.

This follows ZIP 318's per-note availability MUST ("wait until the boundary that closes the anchor-height bucket in which a note-preparation transaction was mined has passed before treating ITS output notes as available") and consciously relaxes the more conservative SHOULD that all preparation complete before Phase 2 begins. The relaxation is safe: the schedule is floored at the estimated last-preparation height, and the boundary-passed half of the MUST is still enforced downstream by the proving-time anchor draw. Independent of #2708 (both off `main`; they touch the same commit function so whichever merges second will rebase cleanly).

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -799,6 +799,21 @@ where
     }
 }
 
+/// A spendable note recovered from an already-built preparation transaction, or a direct-funding
+/// wallet note, tracked by the commit so a later transaction can spend it BY VALUE before it is
+/// mined. Its signable plaintext is fixed at build time (its `rho` is the paired spend's nullifier
+/// and its `rseed` is drawn by the builder); only its tree position awaits mining, and that matters
+/// only to the proof (deferred to proving time, ZIP 374). `consumed` guards against spending it
+/// twice, and `producer` is the preparation transaction that mints it (or `None` for a
+/// direct-funding wallet note), so a transfer depends only on its own funding note's producer.
+#[cfg(feature = "orchard")]
+struct MintedNote {
+    value: Zatoshis,
+    note: orchard::note::Note,
+    consumed: bool,
+    producer: Option<MigrationTxId>,
+}
+
 /// Commit a planned migration: build and pre-sign EVERY transaction — each preparation layer, in
 /// topological order, then every transfer — in this one pass, and persist the whole committed
 /// migration through the backend. Anchors and witnesses are deferred to proving time (ZIP 374), so
@@ -925,18 +940,13 @@ where
     // BROADCAST order only: every transaction is built and signed here, in this one pass.
     let mut layer_ids: Vec<Vec<MigrationTxId>> = Vec::new();
 
-    // The spendable notes minted by already-built preparation transactions, RECOVERED from their
-    // built bundles (a minted note's plaintext is fixed at build time — its rho is the paired
-    // spend's nullifier and its rseed is drawn by the builder; mining only assigns its tree
-    // position, which matters only to the proof). A later layer's Prior spends and the transfers'
-    // funding notes draw from this pool by value, each note spent at most once. This is what
-    // makes the WHOLE migration signable in topological order before anything is broadcast:
-    // signing sessions are bounded by the signer's action budget, never by mining.
-    // Each entry is `(value, recovered note, consumed?, producer)`. `producer` is the id of the
-    // preparation transaction that mints the note, or `None` for a direct-funding wallet note that
-    // needs no preparation; a transfer then depends only on its own funding note's producer, rather
-    // than on the whole last layer.
-    let mut minted: Vec<(Zatoshis, orchard::note::Note, bool, Option<MigrationTxId>)> = Vec::new();
+    // The pool of notes minted by already-built preparation transactions (and the direct-funding
+    // wallet notes), which later layers' `Prior` spends and the transfers draw from BY VALUE, each
+    // spent at most once (see [`MintedNote`]). Because a minted note is signable before it is
+    // mined, this pool is what lets the WHOLE migration be built and signed in one topological pass,
+    // before anything is broadcast — signing sessions bounded by the signer's action budget, never
+    // by mining.
+    let mut minted: Vec<MintedNote> = Vec::new();
 
     for (layer, prep_layer) in plan.preparation().layers().iter().enumerate() {
         let mut this_layer_ids: Vec<MigrationTxId> = Vec::with_capacity(prep_layer.len());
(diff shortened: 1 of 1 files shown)


zcash_pool_migration_backend/src/engine.rs
```diff
@@ -920,7 +920,11 @@ where
     // funding notes draw from this pool by value, each note spent at most once. This is what
     // makes the WHOLE migration signable in topological order before anything is broadcast:
     // signing sessions are bounded by the signer's action budget, never by mining.
-    let mut minted: Vec<(Zatoshis, orchard::note::Note, bool)> = Vec::new();
+    // Each entry is `(value, recovered note, consumed?, producer)`. `producer` is the id of the
+    // preparation transaction that mints the note, or `None` for a direct-funding wallet note that
+    // needs no preparation; a transfer then depends only on its own funding note's producer, rather
+    // than on the whole last layer.
+    let mut minted: Vec<(Zatoshis, orchard::note::Note, bool, Option<MigrationTxId>)> = Vec::new();
```
dannywillems: Nit: would be nice to have a type for this. The whole method `commit_preparation_inner` could be simplified, I think.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -920,7 +920,11 @@ where
     // funding notes draw from this pool by value, each note spent at most once. This is what
     // makes the WHOLE migration signable in topological order before anything is broadcast:
     // signing sessions are bounded by the signer's action budget, never by mining.
-    let mut minted: Vec<(Zatoshis, orchard::note::Note, bool)> = Vec::new();
```
dannywillems, line 935: I would be in favor of rewriting this method later, or splitting it.

dannywillems: This will be done in a follow-up.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1005,7 +1019,12 @@ where
             for (_action_index, output, note) in placed {
                 match output {
                     PrepOutput::Funding(value) | PrepOutput::Intermediate(value) => {
-                        minted.push((value, note, false));
+                        minted.push(MintedNote {
+                            value,
+                            note,
+                            consumed: false,
+                            producer: Some(id),
```
dannywillems, line 1038: This is one of the main change.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1042,14 +1061,29 @@ where
...
             return Err(CommitError::StalePlan);
         }
-        minted.push((value, note, false));
+        // A direct-funding wallet note already exists on-chain, so its transfer has no producer to
+        // wait on.
+        minted.push(MintedNote {
+            value,
+            note,
+            consumed: false,
+            producer: None,
```
dannywillems, line 1082: This is one of the main change of this patch.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1585,9 +1623,19 @@ mod commit_tests {
                     assert!(tx.anchor_boundary.is_none());
                 }
                 MigrationTxKind::Transfer { .. } => {
+                    // A transfer depends only on the ONE preparation transaction that mints its
```
dannywillems, line 1638: Makes sense.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1793,13 +1842,187 @@ mod commit_tests {
...
+    }
+
+    /// Consolidating many small notes needs a DEEP preparation — here four layers — and the state
+    /// machine still walks the broadcasts strictly layer by layer: each preparation layer depends on
+    /// the whole layer before it and broadcasts only once that predecessor mines, and the transfers
+    /// come only after the last layer.
+    #[test]
+    fn commits_a_deep_multi_layer_migration() {
+        let seed = 11u64;
+        // Thirty 100-ZEC notes consolidate through four preparation layers.
```
dannywillems, line 1932: Because of the limit of 16 actions.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -799,6 +799,21 @@ where
     }
 }
 
+/// A spendable note recovered from an already-built preparation transaction, or a direct-funding
```
dannywillems, line 802: Too verbose. But LGTM.

zcash_pool_migration_backend/src/engine.rs
dannywillems: See https://github.com/zcash/librustzcash/pull/2711/

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1793,13 +1842,187 @@ mod commit_tests {
...
         }
     }
 
+    /// A crossing releases as soon as ITS OWN funding note's producer mines — not once the whole
+    /// preparation completes. A two-layer whale's funding notes come from more than one producer, so
+    /// mining ONE producer makes the crossings it funds releasable while the crossings funded by the
+    /// other, still-unmined producer stay blocked.
+    #[test]
+    fn a_crossing_releases_when_its_own_producer_mines() {
+        let seed = 11u64;
```
dannywillems, line 1858: I want to simplify that later, when we moved into a test engine.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1793,13 +1842,187 @@ mod commit_tests {
...
+
+        // The state machine broadcasts each preparation layer in order — a layer only once its
+        // predecessor has mined — then, once the last layer mines, the transfers.
+        let mut state = state;
+        let target = BlockHeight::from_u32(2_100_000);
+        let mut height = 2_000_000u32;
+        for layer in 0..layer_count {
+            let ids = layer_ids(&state, layer);
+            match state.next_step(target) {
+                crate::state::AdvanceStep::Broadcast { id } => assert!(
```
dannywillems, line 1986: These are the lines we care.

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1793,13 +1842,187 @@ mod commit_tests {
...
+                    ids.contains(&id),
+                    "layer {layer} broadcasts once its predecessor has mined"
+                ),
+                other => panic!("expected a layer-{layer} broadcast, got {other:?}"),
+            }
+            // The others stay BLOCKED: every LATER preparation layer still depends (transitively)
```
dannywillems, line 1992: And these ones.

dannywillems:
We could even add more granularity I think, but leaving it per layer for now.

nuttycom (approved):
utACK, looks good

PR #2710, opened 2026-07-21
zcash_pool_migration: consult the drawn anchor boundary at proving time

dannywillems:
## Summary

A migration transfer draws an Orchard anchor boundary at scheduling time and
persists it on the transaction, but until now nothing consulted it: proving was
left entirely to the consumer, so the drawn boundary never determined the tree
state a transfer proved against (a #2663 review finding).

This adds a proving step so the engine finally consults each transfer's
persisted `anchor_boundary` at proving time, via a new `MigrationCrypto` seam:

- `MigrationCrypto` gains `prove_transfer(pczt, anchor_boundary)`.
- A free engine function `prove_transfer(crypto, state, id)` reads a transfer's
  persisted `anchor_boundary` (drawn at scheduling time), hands it and the
  stored PCZT to the crypto backend to install the Orchard source anchor and the
  funding note's witness against that boundary and the Ironwood destination
  anchor (through the PCZT `Updater` role, ZIP 374) and prove both bundles, then
  stores the proven PCZT and moves the transaction `Signed -> Proved` via
  `MigrationState::set_transaction_proved`.
- A new `ProveError` guards the flow: it rejects an unknown transaction, a
  preparation transaction (`NotATransfer`), a non-`Signed` transaction
  (`NotReady`), a transfer with no drawn boundary (`NoAnchorBoundary`), and
  surfaces parse/serialize/crypto failures.

The `WalletMigration` adapter's leg needs the drawn boundary's commitment-tree
checkpoint to still exist at proving time, so it is gated on migration
anchor-checkpoint retention (#2700) for now and returns a new
`Error::ProvingUnsupported`; the in-memory mock stands in for the prover to
exercise the engine flow, and an inline test drives a committed transfer
`Signed -> Proved` and checks the `NotReady` / `NotATransfer` guards.

Extracted from #2669.

## Notes

No `orchard` patch is needed: the code does not call the #535 `Updater` /
`Prover` API (`set_anchor` / `set_spend_witness` appear only in doc comments),
so it compiles against main's `orchard 0.15.0` with no `[patch.crates-io]`
change.

`cargo test --release --all-features -p zcash_pool_migration_backend` is green
(99 tests), and stable `cargo fmt` is clean.

PR #2712, opened 2026-07-21
zcash_client_sqlite: key the pool-migration store by account

Chlup:
Keys the SQLite pool-migration store by account, so each account's Orchard -> Ironwood migration (ZIP 318) is tracked independently — the prerequisite for a wallet running concurrent migrations across accounts (e.g. a software account and an imported hardware-wallet account over one wallet database).

Re-implemented against the store that landed on `main` in #2713 (`zcash_client_sqlite::pool_migration`), which stores a single migration per wallet ("a future multi-account model would replace this with an account key" — `store.rs`). This PR is that model:

- **Schema**: `orchard_ironwood_migrations` gains `account_uuid BLOB NOT NULL` plus a `UNIQUE` index (`idx_orchard_ironwood_migrations_account`) — at most one migration per account, enforced by the database. The six child tables are untouched: they already address rows through the parent's synthetic `id`.
- **API**: `PoolMigrations::for_account(conn, AccountUuid)` replaces `PoolMigrations::new(conn)` (unreleased surface). Scoping happens at construction; the engine's `PoolMigrationRead` / `PoolMigrationWrite` traits are unchanged.
- **Schema migration**: `orchard_ironwood_migration_account_key` — a no-op when `account_uuid` already exists (fresh databases get the keyed shape directly from `init_migration_tables`), and a drop-and-recreate for a database that ran the singleton-shaped predecessor (these tables have not shipped in a release; at most one in-progress migration re-plans).
- **Tests**: the shared store-conformance suite now runs per-account; four new isolation proptests (cross-account `get`/`replace`/`update_transaction` scoping — including the case where two accounts share a `tx_id` — and same-account replace semantics); migration-path tests for the legacy-recreate and new-shape-no-op branches plus the full-DAG `migrate` check. `cargo test -p zcash_client_sqlite --all-features`: **402 passed, 0 failed** (baseline before this change: 395; +7 new). Backend/memory crates untouched and green (94+4+4).

This PR is now single-purpose: the earlier proving commits that rode this branch (an early cut of the work now under review in #2710) have moved to the stacked branch `michal/MOB-1455/MOB-1495-pool-migration-proving` (this branch + that pair), which the mobile SDK consumes until #2710 lands.

(Base changed from `feat/pool-migration-sqlite` to `main` on 2026-07-22, after that feature branch's absorption into `main`.)


dannywillems:
Note: this will need to be backported in https://github.com/zcash/librustzcash/pull/2713

zcash_client_sqlite/src/pool_migration/store.rs
```diff
@@ -72,6 +72,7 @@ fn create_migrations_sql(t: &Tables) -> String {
     format!(
         "CREATE TABLE IF NOT EXISTS {} (
             id INTEGER PRIMARY KEY,
+            account_uuid BLOB NOT NULL,
```
dannywillems, line 79: We might want to use `TEXT` (or the same type as the other table for the account UUID) to be able to add an index and query it efficiently.

dannywillems: `BLOB` is used at the other places. Resolving this thread.

zcash_client_sqlite/src/pool_migration/orchard_ironwood.rs
```diff
@@ -182,5 +197,118 @@ mod tests {
                 .expect_err("no such transaction");
             prop_assert!(matches!(err, Error::Corrupt(_)));
         }
+
+        /// Two accounts sharing one connection are isolated: writing account A's migration
+        /// creates no row visible to account B (which reads back `None`, exactly as an untouched
+        /// store would), while account A itself round-trips normally.
+        #[test]
+        fn accounts_are_isolated(state in arb_migration_state()) {
```
dannywillems, line 392: Follow-up: we will want E2E tests for multiaccounts too.

zcash_client_sqlite/src/pool_migration/store.rs
```diff
@@ -58,12 +59,11 @@ pub(crate) struct Tables {
     pub transaction_deps: &'static str,
     /// The index over `(state, scheduled_height)` on the transactions table.
     pub tx_due_index: &'static str,
+    /// The unique index over `account_uuid` on the migrations table, enforcing at most one
+    /// migration per account.
+    pub account_index: &'static str,
 }
 
-/// The primary-key value of the single active migration. There is at most one migration in progress,
-/// so it is stored as one row; a future multi-account model would replace this with an account key.
```
dannywillems, line 64: So multi-account model was already in mind 😄

zcash_client_sqlite/src/pool_migration/store.rs
```diff
@@ -276,34 +296,58 @@ impl<C: BorrowMut<Connection>> Store<C> {
 // Read
 // ---------------------------------------------------------------------------
 
-fn read_migration(conn: &Connection, t: &Tables) -> Result<Option<MigrationState>, Error> {
+/// Resolve the primary key of `account`'s row in `t.migrations`, if one exists. Every child table
+/// is addressed by this key rather than by `account_uuid` directly, so a write operation resolves
+/// it once and reuses it for every subsequent query.
+fn resolve_migration_id(
```
dannywillems, line 306: LGTM.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -594,11 +594,14 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
...
-/// The single active migration's status and the scalar fields of its note-split plan. The crossing
-/// values are an ordered list in `orchard_ironwood_migration_crossing_values`.
+/// One row per account's active migration: its status and the scalar fields of its note-split
+/// plan. The crossing values are an ordered list in `orchard_ironwood_migration_crossing_values`.
+/// `account_uuid` is enforced unique by `INDEX_ORCHARD_IRONWOOD_MIGRATIONS_ACCOUNT`, so an account
+/// has at most one migration in progress.
 pub(super) const TABLE_ORCHARD_IRONWOOD_MIGRATIONS: &str = "
 CREATE TABLE orchard_ironwood_migrations (
     id INTEGER PRIMARY KEY,
+    account_uuid BLOB NOT NULL,
```
dannywillems, line 605: Missing a foreign key reference.

dannywillems (approved):
utACK 71e25f52165e6dec1b44d3a355f9731648684157.

Adding 3 commits on top of @Chlup work, handling cascading account deletion.

zcash_client_sqlite/src/pool_migration/store.rs
```diff
@@ -58,20 +59,24 @@ pub(crate) struct Tables {
...
 
 fn create_migrations_sql(t: &Tables) -> String {
+    // `account_uuid` is not a foreign key into the wallet's `accounts` table: this store is
+    // deliberately standalone and pool-agnostic (it is exercised over a connection that need not
+    // hold an `accounts` table at all). The wallet's account-deletion path removes an account's
+    // migration explicitly instead; see `orchard_ironwood::delete_account_migration`.
     format!(
         "CREATE TABLE IF NOT EXISTS {} (
             id INTEGER PRIMARY KEY,
+            account_uuid BLOB NOT NULL,
```
nuttycom, line 79: Why is this not a foreign key to the `accounts` table?

zcash_client_sqlite/src/wallet/db.rs
nuttycom: This should have been an `account_id INTEGER REFERENCES accounts(id)` instead.

PR #2713, opened 2026-07-21
zcash_pool_migration_sqlite: add table schema and register it in zcash_client_sqlite

dannywillems:
Adds the SQLite schema for an in-progress Orchard -> Ironwood value-pool migration (ZIP 318), split into two commits.

## Design: fully normalized, BLOB only for the PCZT

Every structured value is stored in typed columns and child tables so it can be queried directly, rather than as an opaque blob. The only `BLOB` column is the pre-signed transaction (`pczt`), which is genuinely unstructured, already-versioned bytes. All amounts are zatoshi `INTEGER` columns; the broadcast `txid` is stored as hex `TEXT`.

Nine tables (all prefixed `orchard_ironwood_migration[s]_`):

| Table | Holds |
|---|---|
| `orchard_ironwood_migrations` | the singleton migration: `status` + the note-split scalars (`note_split_fee_buffer`, `note_split_change`, `note_split_prep_fees`, `note_split_total_input`, `note_split_total_migratable`) |
| `..._crossing_values` | the note split's crossing values (ordered list) |
| `..._funding_notes` | the reconciled funding-note values (ordered list) |
| `..._prep_transactions` | the preparation plan's `(layer, tx_index)` grid |
| `..._prep_inputs` | each prep transaction's inputs (`source` = `wallet`/`prior`, with the variant's columns) |
| `..._prep_outputs` | each prep transaction's outputs (`role` = `funding`/`intermediate`/`change`) |
| `..._prep_direct_funding` | wallet notes used directly as funding notes (no prep transaction) |
| `..._migration_transactions` | one row per migration transaction: `kind` (+ `kind_layer`/`kind_index`/`kind_crossing`), `pczt` (the only BLOB), schedule, `anchor_boundary`, lifecycle `state`, hex `txid`, `mined_height` |
| `..._transaction_deps` | the transaction dependency edges |

plus `idx_orchard_ironwood_migration_tx_due (state, scheduled_height)`.

## Commits

- **`zcash_pool_migration_sqlite`: add the value-pool migration table schema.** A new crate owning the DDL above. The generic `store` module builds the statements from per-pool table names; the `orchard_ironwood` submodule instantiates them with this pool's names and exposes its canonical `MIGRATION_ID` and an idempotent `init_migration_tables`. The crate depends only on `rusqlite` and `uuid` (it does not depend on `zcash_pool_migration_backend`, so it is fully decoupled from the store codecs).
- **`zcash_client_sqlite`: register the Orchard -> Ironwood pool-migration tables.** A thin `schemerz` migration whose `up()` runs that DDL inside the wallet schema (same `wallet.db`, shared schema versioning), depending on `ironwood_received_notes`. Registered in the migration list and as a current leaf, with the golden schema for all nine tables and the index added to the expected-schema test.

## Scope: tables only

The store implementation (the `PoolMigrationRead` / `PoolMigrationWrite` logic that maps the engine types to and from these tables) is intentionally deferred to a follow-up, since it depends on work not yet accepted. This PR wires only the schema.


dannywillems:
## Mobile SDK database access for the Orchard → Ironwood migration

Context for this schema PR: I audited how the iOS (ZcashLightClientKit) and Android (zcash-android-wallet-sdk) SDKs access the wallet database for the migration, across their ironwood-migration branches.

**Bottom line: neither SDK issues its own SQL against these tables.** Both hand the wallet-DB *path* across the FFI and delegate 100% of migration DB access to librustzcash (`zcash_pool_migration::MigrationContext`), which owns every query. The tables this PR adds (`orchard_ironwood_migrations` / `orchard_ironwood_migration_transactions`) are never named in either SDK, in any layer (Rust FFI, Kotlin, or Swift). There is no raw SQL and no `rusqlite`/SQLite handle in the migration FFI code.

### Delegated migration DB operations (identical surface on iOS and Android)

Each FFI entry point builds `MigrationContext::new(db_path, network, account)` and calls one crate method; the SQL lives in the crate, not the SDK.

| `MigrationContext` method | R/W | Purpose |
|---|:---:|---|
| `migration_state` | R | current state-machine state |
| `migration_progress` | R | completed/total, remaining Orchard |
| `is_note_split_needed` | R | does the Orchard note need splitting first |
| `has_overdue_transfers` / `has_invalid_transfers` | R | scheduling checks |
| `prepare_note_split` | R | plan the denomination split |
| `residual_after_migration` | R | leftover value below the dust floor |
| `propose_transfers` / `propose_immediate_transfers` | R | build the transfer schedule |
| `next_due_transfer` / `pending_transfer_proposal` | R | next scheduled transfer |
| `is_sync_required_before_next_transfer` | R | gate on sync |
| `sign_note_split` | W | build + sign + persist the split PCZT |
| `sign_and_store_migration_schedule` | W | persist the signed schedule |
| `record_transfer_result` | W | mark broadcast / mined |
| `restart_current_migration_step` / `refresh_stale_transfers` | W | recover a step |
| `create_unsigned_*` / `store_signed_*` (Keystone / hardware) | R/W | offline-signer PCZT export / import |
| `ironwood_activation_height` | — | pure network param, no DB |

### The SDKs' own direct DB access (all pre-existing, none migration-related)

| SDK | Direct DB access | Tables / views | Migration? |
|---|---|---|:---:|
| Both | read-only metadata views over the wallet `data.db` | `v_transactions`, `v_tx_outputs`, `blocks`, `transactions` | no |
| iOS only | a **separate** SDK-owned SQLite database | `tx_submit_plans` (`SubmitPlanStore.swift`) | no |
| Android | no Room at all; `SupportSQLiteDatabase` over the views above | — | no |

### Branch nuance

- **Productized integration** (Android `feature/orchard_migration`, iOS `michal/MOB-1455/MOB-1495-*`): the full `MigrationContext` delegation above. (iOS `MOB-1455-2/-3` are an earlier prototype over `zodl_ironwood_migration`.)
- **Interface / planner branch** (`harry/ironwood-migration-sdk-interface`, both SDKs): a pure-arithmetic denomination planner (`plan_denominations` / `zcashlc_plan_orchard_denomination_split`) with **no DB access at all**; actual transfers reuse the standard `zcash_client_sqlite` proposal API, and the SDK / app owns the split + anchor logic.

### Relevance to this PR

Since neither SDK queries these tables directly, the schema here (and the eventual `PoolMigrationRead` / `PoolMigrationWrite` store implementation) stays entirely librustzcash-owned. The tables-only slice can land ahead of the codecs with zero SDK impact.

dannywillems:
## Note: `idx_orchard_ironwood_migration_tx_due` is not exercised by any query yet

Flagging for reviewers, since this PR adds the index but nothing queries it in the current design.

The store's read path loads the **whole** migration and computes all lifecycle logic in Rust, rather than filtering in SQL:

```sql
SELECT status, note_split, funding_notes, preparation
  FROM orchard_ironwood_migrations WHERE id = 0;
SELECT tx_id, kind, pczt, depends_on, scheduled_height, ...
  FROM orchard_ironwood_migration_transactions
  WHERE migration_id = 0 ORDER BY tx_id;
```

The state machine (`MigrationState::next_broadcastable`) then selects the next transaction **in memory**, filtering on `state ∈ {Signed, Proved}` AND `scheduled_height <= tip` AND `deps_mined(depends_on)`. So `idx_..._tx_due (state, scheduled_height)` — which would accelerate exactly that `state` + `scheduled_height` predicate — is never hit by any query. It is speculative today.

This is the right design, not a bug: a migration is a singleton with ~tens of rows, so the whole-object load is trivially cheap and keeps the lifecycle rules as one pure, backend-agnostic source of truth (reused verbatim by the in-memory backend) instead of splitting the predicate between SQL and Rust. (The only filter that would *need* a schema change to push into SQL is `deps_mined`, since `depends_on` is a BLOB and would require a normalized edge table — not worth it at this scale.)

So the index is either:

- **forward-provisioning** for an eventual "give me just the due transaction" SQL query (my lean: keep it — it documents the intended hot path and costs ~nothing on a tiny table), or
- **droppable** until such a query exists.

No change needed for this tables-only PR; just noting it so the index is not mistaken for a load-bearing part of the current read path.

---

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -589,6 +589,42 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
...
+/// Stores the note-split decomposition and status of the single in-progress Orchard -> Ironwood
+/// pool migration (ZIP 318). The table DDL and store live in the `zcash_pool_migration_sqlite`
+/// crate; this golden copy tracks the schema those tables install into `wallet.db`. The `note_split`
+/// column holds the whole serialized note-split plan; the `funding_notes` column holds a serialized
+/// vector of little-endian `u64` amounts; `preparation` holds the serialized preparation plan.
+pub(super) const TABLE_ORCHARD_IRONWOOD_MIGRATIONS: &str = "
+CREATE TABLE orchard_ironwood_migrations (
+    id INTEGER PRIMARY KEY,
+    status TEXT NOT NULL,
+    note_split BLOB NOT NULL,
```
dannywillems: These are the blobs which makes https://github.com/zcash/librustzcash/pull/2705/changes blocked for now.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -589,6 +589,42 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
...
+    preparation BLOB NOT NULL
+)";
+/// Stores one row per transaction of the in-progress pool migration: its pre-signed PCZT (`pczt`,
+/// always present, since every transaction is built when the migration is committed under one-phase
+/// signing), dependency graph, schedule, and lifecycle state. See [`TABLE_ORCHARD_IRONWOOD_MIGRATIONS`].
+pub(super) const TABLE_ORCHARD_IRONWOOD_MIGRATION_TRANSACTIONS: &str = "
+CREATE TABLE orchard_ironwood_migration_transactions (
+    migration_id INTEGER NOT NULL REFERENCES orchard_ironwood_migrations(id) ON DELETE CASCADE,
+    tx_id INTEGER NOT NULL,
+    kind BLOB NOT NULL,
```
dannywillems: See https://github.com/zcash/librustzcash/pull/2705/changes#diff-5d7cb9b0bcb065dfba5f5422d470b5a479d920f6fe1488525377e7b9d37e6038R156

dannywillems:
I misunderstood the initial concern; I thought that we would save in blob except with we needed to actually have some logical queries on the data, at the cost of expanding the schema in more tables. Using blobs was a solution to not have more than two tables. The final design decision is to not use any blob, except for the PCZT which are actually (unstructured) bytes/data. Blobs should only be used for unstructured data.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -589,6 +589,122 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
...
+// `zcash_pool_migration_sqlite` crate; these golden copies track the normalized schema those tables
+// install into `wallet.db`. Every structured value is stored in typed columns and child tables; the
+// only `BLOB` is the pre-signed transaction (`pczt`), which is already-versioned, unstructured bytes.
+
+/// The single active migration's status and the scalar fields of its note-split plan. The crossing
+/// values are an ordered list in `orchard_ironwood_migration_crossing_values`.
+pub(super) const TABLE_ORCHARD_IRONWOOD_MIGRATIONS: &str = "
+CREATE TABLE orchard_ironwood_migrations (
+    id INTEGER PRIMARY KEY,
+    status TEXT NOT NULL,
```
dannywillems, line 602: Could be an integer, but if we do use one, and we remove one of the status, the logic will need to be adapted.

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,901 @@
...
+/// Decode a stored `INTEGER` zatoshi amount, erroring on a negative or out-of-range value.
+fn zat(v: i64) -> Result<Zatoshis, Error> {
+    u64::try_from(v)
+        .ok()
+        .and_then(|u| Zatoshis::from_u64(u).ok())
+        .ok_or(Error::Corrupt("zatoshi value"))
+}
+
+/// Lowercase-hex encode raw bytes (used for the broadcast `txid` column).
+fn hex_encode(bytes: &[u8]) -> String {
```
dannywillems: Could it be implemented somewhere else? Useful?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,901 @@
...
+// Value helpers
+// ---------------------------------------------------------------------------
+
+/// A zatoshi amount as the `i64` stored in an `INTEGER` column (the money cap fits in `i64`).
+fn unzat(z: Zatoshis) -> i64 {
+    z.into_u64() as i64
+}
+
+/// Decode a stored `INTEGER` zatoshi amount, erroring on a negative or out-of-range value.
+fn zat(v: i64) -> Result<Zatoshis, Error> {
```
dannywillems: Could it be implemented somewhere else? Useful?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,901 @@
...
+    }
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Value helpers
+// ---------------------------------------------------------------------------
+
+/// A zatoshi amount as the `i64` stored in an `INTEGER` column (the money cap fits in `i64`).
+fn unzat(z: Zatoshis) -> i64 {
```
dannywillems: Could it be implemented somewhere else? Useful?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,901 @@
...
+    let mut s = String::with_capacity(bytes.len() * 2);
+    for &b in bytes {
+        s.push(char::from_digit((b >> 4) as u32, 16).expect("nibble is 0..16"));
+        s.push(char::from_digit((b & 0x0f) as u32, 16).expect("nibble is 0..16"));
+    }
+    s
+}
+
+/// Decode a 64-character lowercase-hex string into a 32-byte array, or `None` if malformed.
+fn hex_decode_32(s: &str) -> Option<[u8; 32]> {
```
dannywillems: Could it be implemented somewhere else? Useful?

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Good call — replaced with the `hex` crate (already a workspace dependency), dropping the hand-rolled version.

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Good call — replaced with the `hex` crate (already a workspace dependency), dropping the hand-rolled version.

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Removed — this is just `Zatoshis::into_u64` / `Zatoshis::from_u64` (already in `zcash_protocol`) over rusqlite's native `u64` support, matching how `zcash_client_sqlite` stores amounts. A new `From<BalanceError> for Error` lets the range check propagate via `?`, so no wrapper is needed.

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Removed — this is just `Zatoshis::into_u64` / `Zatoshis::from_u64` (already in `zcash_protocol`) over rusqlite's native `u64` support, matching how `zcash_client_sqlite` stores amounts. A new `From<BalanceError> for Error` lets the range check propagate via `?`, so no wrapper is needed.

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+
+    // The transactions in each layer, in `(layer, tx_index)` order.
+    let coords: Vec<(usize, usize)> = {
+        let mut stmt = conn.prepare(&format!(
+            "SELECT layer, tx_index FROM {} WHERE migration_id = ? ORDER BY layer, tx_index",
+            t.prep_transactions
+        ))?;
+        let rows = stmt.query_map(params![SINGLETON_ID], |row| {
+            Ok((
+                row.get::<_, i64>(0)? as usize,
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+                row.get::<_, i64>(1)? as usize,
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+          WHERE migration_id = ? AND layer = ? AND tx_index = ?
+          ORDER BY ordinal",
+        t.prep_inputs
+    ))?;
+    let rows = stmt.query_map(
+        params![SINGLETON_ID, layer as i64, tx_index as i64],
+        |row| {
+            Ok((
+                row.get::<_, String>(0)?,
+                row.get::<_, Option<i64>>(1)?,
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+                row.get::<_, Option<i64>>(2)?,
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+                row.get::<_, Option<i64>>(3)?,
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+                row.get::<_, Option<i64>>(4)?,
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+    tx_index: usize,
+) -> Result<Vec<PrepOutput>, Error> {
+    let mut stmt = conn.prepare(&format!(
+        "SELECT role, value FROM {}
+          WHERE migration_id = ? AND layer = ? AND tx_index = ?
+          ORDER BY ordinal",
+        t.prep_outputs
+    ))?;
+    let rows = stmt.query_map(
+        params![SINGLETON_ID, layer as i64, tx_index as i64],
```
dannywillems: Could it be u64?

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+        |row| Ok((row.get::<_, String>(0)?, row.get::<_, u64>(1)?)),
+    )?;
+    let mut out = Vec::new();
+    for r in rows {
+        let (role, value) = r?;
+        let value = Zatoshis::from_u64(value)?;
+        let output = match role.as_str() {
```
dannywillems: This could go at the `PrepOutput` level.

zcash_pool_migration_sqlite/src/store.rs
```diff
@@ -0,0 +1,864 @@
...
+                mined_height: row.get(11)?,
+            })
+        })?;
+        mapped.collect::<Result<_, _>>()?
+    };
+
+    let mut out = Vec::with_capacity(rows.len());
+    for r in rows {
+        let id = MigrationTxId::new(r.tx_id);
+        let kind = match r.kind.as_str() {
```
dannywillems: This could go at the MigrationKind level.

zcash_client_sqlite/src/pool_migration/store.rs
```diff
@@ -0,0 +1,864 @@
...
+                    ":depends_on_tx_id": u32::from(*dep),
+                },
+            )?;
+        }
+    }
+    Ok(())
+}
+
+/// Insert an ordered list of zatoshi amounts as `(ordinal, value)` rows into a child table.
+fn insert_zatoshi_list(
```
dannywillems, line 778: Wondering if a iter or a map or something iterating over the `values` could do the same job.

dannywillems (commented):
We need to ensure that atomicity is handled property when inserting.

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done in `c9ba7eb141` — switched the index columns to `u64` (they are non-negative, and it matches the value columns, which are also `u64`).

zcash_client_sqlite/src/wallet/db.rs
dannywillems: Kept as `TEXT`: it reuses `MigrationStatus`'s existing `AsRef<str>` / `TryFrom<&str>` (the canonical serialization), is self-describing, and stays stable if the enum is reordered — an int would couple the mapping to the variant order, exactly as you note. Happy to switch if you prefer the compactness.

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done — moved the role mapping onto the type: `PrepOutput::from_role` + `AsRef<str>` (with `ParsePrepOutputError`), matching `MigrationTxState`. The store now uses `output.as_ref()` on write and `PrepOutput::from_role` on read.

zcash_pool_migration_sqlite/src/store.rs
dannywillems: Done — moved the reconstruction onto the type: `MigrationTxKind::from_stored` + `AsRef<str>` + `preparation_indices` / `transfer_crossing` accessors (with `ParseMigrationTxKindError`). The store uses those instead of the inline match.

zcash_client_sqlite/src/pool_migration/store.rs
dannywillems: It is a side-effecting `INSERT` per element (each does an `execute(...)?`), so a `for` loop is the idiomatic form here; a `.try_for_each` would be equivalent but reads less clearly for a fallible side effect. Left as the loop.

dannywillems:
Re: atomicity when inserting — this is handled. `Store::replace_migration` runs the whole delete-all + insert-all inside a single `rusqlite::Transaction`:

```rust
let tx = self.conn.borrow_mut().transaction()?;
replace_migration(&tx, tables, state)?;  // all DELETEs + INSERTs
tx.commit()?;
```

The `Transaction` commits only at the end; if any statement fails, the `?` returns early and the transaction is dropped, which rolls back (rusqlite `Transaction` rolls back on drop by default), so a partially-written migration never persists. `update_transaction` is a single `UPDATE`, and `get_migration` is read-only.

zcash_pool_migration_sqlite/README.md
nuttycom, line 1: I don't understand the rationale of having a separate crate here, given that `zcash_client_sqlite` maintains the table. Either this should be an external crate and use `ext_` tables, or it should be a part of `zcash_client_sqlite` and not a separate crate that users must depend on. I think that latter is the better path.

zcash_client_sqlite/src/pool_migration/store.rs
```diff
@@ -0,0 +1,221 @@
...
+            migration_id INTEGER NOT NULL,
+            layer INTEGER NOT NULL,
+            tx_index INTEGER NOT NULL,
+            PRIMARY KEY (migration_id, layer, tx_index),
+            FOREIGN KEY (migration_id, layer)
+                REFERENCES {}(migration_id, layer) ON DELETE CASCADE
+        )",
+        t.prep_transactions, t.prep_layers
+    )
+}
```
nuttycom: These two tables seem a bit superfluous. They serve to enforce the uniqueness of the (migration_id, layer) and (migration_id, layer, tx_index) tuples but I don't think they're critical to the design; the `prep_inputs` and `prep_outputs` tables seem sufficient.

nuttycom (commented):
Flushing comments.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -589,6 +589,132 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
...
+    PRIMARY KEY (migration_id, ordinal)
+)";
+/// The reconciled funding-note values (an ordered list of zatoshi amounts).
+pub(super) const TABLE_ORCHARD_IRONWOOD_MIGRATION_FUNDING_NOTES: &str = "
+CREATE TABLE orchard_ironwood_migration_funding_notes (
+    migration_id INTEGER NOT NULL REFERENCES orchard_ironwood_migrations(id) ON DELETE CASCADE,
+    ordinal INTEGER NOT NULL,
+    value INTEGER NOT NULL,
+    PRIMARY KEY (migration_id, ordinal)
+)";
```
nuttycom: This table seems redundant with respect to the migration_crossing_values table; consider how to deduplicate.

zcash_client_sqlite/src/pool_migration/store.rs
nuttycom: Removed both tables in a0e937819e / 2bfb063a96: the layers/transactions grid is now reconstructed from the `(layer, tx_index)` coordinates on the `prep_inputs` / `prep_outputs` rows. This relies on invariants a real plan always satisfies (every preparation transaction has at least one input and one output; no layer is empty) — the backend's proptest strategies now guarantee them (8e909544f9), and the store rejects a violating state on write with `Error::Unrepresentable` rather than silently renumbering the grid (which would misdirect prior-output references).

🤖 Posted with [Claude Code](https://claude.com/claude-code) assistance.

zcash_client_sqlite/src/wallet/db.rs
nuttycom: Removed in 8e909544f9: the redundancy was at the type level — `MigrationState.funding_notes` was always `note_split.migration_outputs()` (each crossing value plus the fee buffer; reconciliation already happens inside `plan_note_split`), so the field is gone from `MigrationState` / `MigrationPlan`, `funding_notes()` is now a derived accessor, and the store persists only the note split — no funding-notes table.

🤖 Posted with [Claude Code](https://claude.com/claude-code) assistance.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -589,6 +589,108 @@ CREATE INDEX idx_ironwood_received_note_spends_transaction_id ON ironwood_receiv
...
+/// hex broadcast `txid` and `mined_height`. Dependencies are edges in
+/// `orchard_ironwood_migration_transaction_deps`.
+pub(super) const TABLE_ORCHARD_IRONWOOD_MIGRATION_TRANSACTIONS: &str = "
+CREATE TABLE orchard_ironwood_migration_transactions (
+    migration_id INTEGER NOT NULL REFERENCES orchard_ironwood_migrations(id) ON DELETE CASCADE,
+    tx_id INTEGER NOT NULL,
+    kind TEXT NOT NULL,
+    kind_layer INTEGER,
+    kind_index INTEGER,
+    kind_crossing INTEGER,
```
nuttycom, line 668: Why are these `kind_`-prefixed?

nuttycom: Figured it out, these are encoded with a discriminant.

nuttycom (approved):
utACK 2bfb063a965640500634c4d062177f5e7d3a4cf9

PR #2714, opened 2026-07-21
feat(pool-migration): estimate the number of migration runs a wallet needs

dannywillems:
## Motivation

When a migration plan is started, a wallet SDK wants to know **how many
migration runs ("rounds")** the user's balance will take, and — for mobile
wallets that sign through a **capacity-limited hardware signer (Keystone)** —
how many **signing interactions** that implies. A balance beyond a single run's
capacity (the note cap times the maximum denomination, e.g. `50 * 10,000 ZEC`)
migrates over several runs, each run's change rolling forward. Nothing exposed
either count. This adds them.

## What this adds

`zcash_pool_migration_backend::engine::estimate_migration_runs(params, backend, rng)`
returns a `MigrationRunEstimate`: a preview computed before anything is planned,
signed, or committed.

- **`MigrationRunEstimate`**: `run_count()`, `runs() -> &[RunEstimate]`,
  `final_residual()`, and aggregates `total_migratable()`, `total_crossings()`,
  `total_prep_layers()`, `total_prep_transactions()`, `total_transactions()`,
  and `total_signing_sessions(max_per_session)`.
- **`RunEstimate`** (one per run) carries **all three** dimensions:
  - note-split side: `migratable()`, `crossings()`
  - note-preparation side: `prep_layers()`, `prep_transactions()`
  - signing side: `transactions()`, `signing_sessions(max_per_session)`

## How it works

- **Depends on the wallet's note structure, not just its total value.** Each run
  is decomposed with `plan_note_split` backed by the real `plan_preparation`
  over the *current* note set (exact feasibility and fees, exactly as
  `plan_migration` plans one run).
- **Note-splitting and note-preparation are computed per run and compared.** Each
  `RunEstimate` records the crossings the split produces and the layers /
  transactions preparing them costs.
- **The note structure evolves across runs.** After each run the next run's notes
  are the wallet notes preparation did not spend (funding notes cross out) plus
  the preparation residuals, so a whale drains over several rounds.
- **Signer capacity is a query parameter.** A hardware signer (e.g. Keystone) can
  sign only so many transactions per interaction — a user-configured hard limit.
  `total_signing_sessions(limit)` reports the number of signing interactions the
  whole migration needs. It is a pure function of the already-computed per-run
  transaction counts, so an SDK can evaluate it for any capacity without
  re-running the planners. Sessions are **summed per run**, not
  `ceil(total / limit)`, because a later run's transactions spend notes an
  earlier run must mine first — each run is signed on its own. The note-split run
  count is unchanged by the signer's capacity.
- A zero or fully sub-quantum balance yields a zero-run preview rather than an
  error.

## Cost note

The estimator iterates the note-split and preparation planners once per run
(~one `plan_migration` per round), so a max-supply whale runs them ~40+ times.
That is fine for a one-off preview but should not be called in a tight loop.
Computing `total_signing_sessions` for a different limit does **not** re-run the
planners.

## Tests

Adds five engine tests (crate suite passing):

- `estimates_a_single_run_with_both_sides`
- `empty_balance_estimates_zero_runs`
- `whale_migrates_over_several_runs`
- `estimate_depends_on_wallet_note_structure` (same 200 ZEC balance as one note
  vs. 200 notes -> strictly more preparation work)
- `signing_sessions_follow_the_signer_capacity` (one session per transaction at
  capacity 1, one per run above every run's tx count, monotone as the limit
  tightens, summed per run)

`cargo clippy` and `cargo fmt` are clean (repo's pinned stable toolchain).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -43,6 +43,7 @@
 
 use alloc::vec::Vec;
 use core::fmt;
+use core::num::NonZeroUsize;
 
 use corez::io;
 
@@ -638,6 +639,23 @@ impl RunEstimate {
     pub fn prep_transactions(&self) -> usize {
         self.prep_transactions
     }
+
+    /// The total number of transactions this run builds and signs: its preparation transactions plus
+    /// one pool-crossing transfer per funding note.
+    pub fn transactions(&self) -> usize {
+        self.prep_transactions + self.crossings
+    }
+
+    /// The number of signing sessions this run needs when an external signer (for example a Keystone
+    /// hardware wallet) can sign at most `max_per_session` transactions in one interaction:
+    /// `ceil(transactions / max_per_session)`. All of a run's transactions are built and signed
+    /// together (anchors and witnesses are deferred to proving time, [ZIP 374]), so they pool into
+    /// sessions bounded only by the signer's capacity.
+    ///
+    /// [ZIP 374]: https://zips.z.cash/zip-0374
+    pub fn signing_sessions(&self, max_per_session: NonZeroUsize) -> usize {
+        self.transactions().div_ceil(max_per_session.get())
+    }
 }
 
 /// An estimate of migrating a wallet's whole spendable balance across successive migration RUNS
@@ -648,6 +666,12 @@ impl RunEstimate {
 /// structure (see [`estimate_migration_runs`]). Each run carries BOTH sides so an application can
 /// compare them: the note-split crossings it migrates and the note-preparation layers and
 /// transactions it costs.
+///
+/// A capacity-limited external signer adds a third dimension: given how many transactions such a
+/// signer can sign in one interaction, [`total_signing_sessions`](Self::total_signing_sessions) gives
+/// the number of signing interactions the whole migration requires. That limit is a query parameter,
+/// not part of the estimate, so an SDK can evaluate it for any signer capacity without re-running the
+/// planners.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MigrationRunEstimate {
     runs: Vec<RunEstimate>,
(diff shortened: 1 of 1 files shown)


zcash_pool_migration_backend/src/engine.rs
```diff
@@ -606,6 +606,230 @@ where
...
+/// ("rounds"): one [`RunEstimate`] per run and the value left un-migrated at the end.
+///
+/// A balance beyond one run's capacity (the note cap times the maximum denomination) migrates over
+/// several runs, each run's spent notes and preparation residuals forming the next run's note
+/// structure (see [`estimate_migration_runs`]). Each run carries BOTH sides so an application can
+/// compare them: the note-split crossings it migrates and the note-preparation layers and
+/// transactions it costs.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MigrationRunEstimate {
+    runs: Vec<RunEstimate>,
```
nullcopy, line 653: This naming is confusing. A run is a vector of runs?

Perhaps rename this struct to `TotalMigrationEstimate` which is a collection of `RunEstimate`s?

zcash_pool_migration_backend/src/engine.rs
```diff
@@ -606,6 +606,230 @@ where
...
+    }
+
+    /// The total number of note-preparation transactions across all runs (the sum of each run's
+    /// [`prep_transactions`](RunEstimate::prep_transactions)).
+    pub fn total_prep_transactions(&self) -> usize {
+        self.runs.iter().map(RunEstimate::prep_transactions).sum()
+    }
+}
+
+/// Estimate how the account the `backend` represents will migrate its whole spendable balance: the
```
nullcopy, line 704: Why is this only an estimate and not a precise migration schedule? Is it just because we don't know if the user will spend notes manually and deviate from the schedule, or something else?

nullcopy (commented):
I know I'm a bit late with this, but to avoid confusion with other forms of migration, should we be more distinct in our naming? To make it clear that we're talking about ZIP318/Ironwood migration and not something else? e.g:
```
estimate_zip318_migration_runs()
Zip318RunEstimate
Zip318MigrationEstimate
```

Flushing comments as of 974763fe53ff61a27c73f5eaf8130ea8c2aa8afb

nullcopy (approved):
Approving, because my prior comments are non-blocking. The logic looks sound to me.

utACK 974763fe53ff61a27c73f5eaf8130ea8c2aa8afb

zcash_pool_migration_backend/src/engine.rs
dannywillems: Agreed. This could be simplified. Keeping for a follow-up.

dannywillems (commented):
self-ACK. We can improve later if needed. This is an informational feature.

PR #2715, opened 2026-07-21
zcash_pool_migration_backend: remove the K_MAX multiplicity cap

nuttycom:
Implements the ZIP editors' decision on [zcash/zips#1317](https://github.com/zcash/zips/pull/1317/files#r3625913319) to drop ZIP 318's per-wallet `K_MAX`-per-cohort multiplicity bound: so long as anchor ages come from independent random draws from the exponential distribution, truncating the outcome of those draws with an arbitrary cap serves no purpose and only distorts the distribution.

This removes `K_MAX`, `draw_anchor_boundary_bounded`, `BoundaryCounts`, and `group_by_boundary` from the scheduling module — the migration engine only ever used the plain `draw_anchor_boundary` — along with their tests, golden vectors, and proptests, and updates the module docs to state that per-wallet multiplicity on a boundary is deliberately unbounded. None of the removed items were referenced outside the scheduling module or mentioned in the CHANGELOG.

Two cleanup commits fix latent issues surfaced while verifying this change:

- The `testing` module imports `zcash_primitives::transaction::testing::arb_txid`, but the `test-dependencies` feature only forwarded `zcash_protocol/test-dependencies`, so `cargo check -p zcash_pool_migration_backend --all-features` failed with an unresolved import on `main`. The feature now forwards `zcash_primitives/test-dependencies` as well; it is ordered first so every commit builds cleanly under `--all-features`.
- `pczt`'s `verify_witnesses_root_to_anchor` and its `AnchorConsistencyError` (Orchard-protocol and Sapling variants) are only used by the `Prover` role but were gated only on their pool feature, emitting dead-code warnings in pool-enabled builds without `prover`. They are now gated on the pool feature combined with `prover`, and the previously unlogged `AnchorConsistencyError` additions are recorded in the pczt changelog.

### DIFF
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -16,6 +16,10 @@ workspace.
 - `pczt::roles::redactor::orchard::OrchardRedactor::compact_resolvable_fields`,
   behind the `orchard` feature, which compacts the Orchard-protocol (Orchard and
   Ironwood) action fields that `pczt::orchard::Bundle::resolve_fields` can restore.
+- `pczt::orchard::AnchorConsistencyError` and `pczt::sapling::AnchorConsistencyError`,
+  behind the `prover` feature combined with the `orchard` / `sapling` feature
+  respectively: the witness-to-anchor consistency errors the `Prover` role
+  reports via its `InconsistentWitness` error variants.
 
 ### Changed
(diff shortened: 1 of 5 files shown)


nullcopy (approved):
utACK 0836df3b11471191e706fdfbc8ca50e27200f80d

PR #2716, opened 2026-07-22
feat: note locking

### LINKED ISSUE #2161: zcash_client_backend: Support note locking

schell:
## Note locking

Adds a note/UTXO locking facility to `zcash_client_backend` and `zcash_client_sqlite`, allowing a wallet to reserve the inputs selected by an in-flight transaction proposal or PCZT so that concurrent proposals for the same account do not select the same outputs. This addresses the gap described in the Zallet audit, where overlapping `z_sendmany`/`z_shieldcoinbase` operations on one account could select the same inputs during the proving-long selection→store window and each build, sign, and broadcast — wasting proving work and producing a self-healing but confusing conflict.

Continues the work in #2162 (author @nuttycom); part of the ZIP 318 / Ironwood migration stack.

Closes #2161.

### What this adds

- **`OutputRef`** (`zcash_client_backend::wallet`): a cross-pool reference to a received output `(txid, pool, output_index)`, with `From<NoteId>`. `ProposalError::ChainDoubleSpend` now wraps an `OutputRef` instead of an ad-hoc `(PoolType, TxId, u32)` tuple.
- **`WalletWrite::lock_outputs` / `unlock_output` / `clear_locked_outputs`**, plus `WalletTest::get_locked_outputs`. `lock_outputs` takes a sequence and locks it atomically (all-or-nothing), failing with `LockError::LockFailure` if any output is already locked; expired locks are replaced. `clear_locked_outputs` releases every lock for an account, as a recovery path for lost proposals (e.g. the app was killed before the transaction could be built).
- A **`lock_expiry_height` column** on the four received-notes/outputs tables (`sapling`, `orchard`, `ironwood`, `transparent`), via a new migration. A note is locked while `lock_expiry_height >= target_height`; locks expire automatically as the chain advances.
- An **`include_locked: bool` parameter** on the `InputSource` selection/metadata methods, threaded through to the SQL queries so locked notes are excluded from selection (and from spendable balance) by default.
- A **`lock_for_blocks: Option<u32>` parameter** on `propose_transfer`, `propose_standard_transfer_to_address`, `propose_send_max_transfer`, and `propose_shielding`. When `Some(n)`, the proposal's inputs are locked until `target_height + n`; when `None`, no locking occurs. `unlock_proposal_inputs` releases them on the abandon/reject path.
- **Balance** gains a `locked_value` component (`Balance`, `AccountBalance`). Locked funds move out of the spendable total but remain part of `total()`.
- `store_transactions_to_be_sent` unlocks any outputs recorded as spent by the stored transactions, since the spend records then prevent re-selection.

### Design notes

- **`lock_for_blocks: Option<u32>` vs `lock_notes: bool`.** #2161 originally proposed a `lock_notes: bool`. This PR uses the height-based `Option<u32>` instead: `Some(n)` is "lock with an expiry window", `None` is "do not lock". It is strictly more expressive and matches the `lock_expiry_height` column. This is intentional and documented on `propose_transfer`.
- **Concurrency / TOCTOU.** Note selection and locking cannot be one atomic step above the storage layer, so two callers may select overlapping inputs before either locks them. The conflict is resolved at the storage layer: the second `lock_outputs` fails with `LockError::LockFailure`, surfaced through proposal creation as `ProposalError::ChainDoubleSpend`. The losing caller has locked nothing partially and should treat it as "account busy" and retry. Documented on `WalletWrite::lock_outputs` and the proposal functions.
- **Selection/balance boundary.** The selection filter (`lock_expiry_height < target_height` ⇒ selectable) is the exact complement of the locked-balance condition (`lock_expiry_height >= target_height` ⇒ locked), so a note is never simultaneously spendable and counted as locked. A boundary test exercises `lock_expiry_height == target_height` and `== target_height - 1`.
- **Atomicity.** In the SQLite backend, `lock_outputs`, `unlock_output`, and `clear_locked_outputs` each run within a single storage transaction; the batch lock is all-or-nothing.

### Breaking changes

- `InputSource` selection/metadata methods gain `include_locked: bool` (downstream `InputSource` implementers must update their signatures).
- The four `propose_*` functions gain `lock_for_blocks: Option<u32>` and now require `WalletWrite` rather than `WalletRead`.
- `ProposalError::ChainDoubleSpend` now wraps `OutputRef`.
- `Balance::total` / `AccountBalance::total` now include locked value; callers computing a total from `spendable_value` alone must add `locked_value`.

See the `zcash_client_backend` and `zcash_client_sqlite` CHANGELOGs for the full list.

### Testing

- `explicit_note_locking`, `proposal_level_note_locking`, `note_locking_height_boundary`, and `clear_locked_outputs` pool tests, run for both Sapling and Orchard, plus the migration test.
- Full `zcash_client_sqlite` and `zcash_client_backend` suites pass with `--all-features`; `clippy --all-features --all-targets` is clean; verified across `--no-default-features`, `transparent-inputs`, `orchard`, and `orchard,transparent-inputs` feature combinations.

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1763,14 +1763,11 @@ pub trait InputSource {
     ///
     /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or would not be
     /// spendable in a transaction mined in the block at the target height.
-    /// Locked outputs are selected according to `lock_filter` (see [`LockFilter`]; a
-    /// [`LockFilter::Policy`] carrying the default `Exclude` selects none).
     #[cfg(feature = "transparent-inputs")]
     fn get_unspent_transparent_output(
         &self,
         _outpoint: &OutPoint,
         _target_height: TargetHeight,
-        _lock_filter: LockFilter<'_>,
     ) -> Result<Option<WalletTransparentOutput<Self::AccountId>>, Self::Error> {
         unimplemented!(
             "InputSource::get_spendable_transparent_output must be overridden for wallets to use the `transparent-inputs` feature"
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1660,18 +1660,21 @@ pub trait InputSource {
     /// specified shielded protocol.
     ///
     /// Returns `Ok(None)` if the note is not known to belong to the wallet or if the note
-    /// is not spendable as of the given height.
+    /// is not spendable as of the given height. When `include_locked` is `false`, locked
+    /// notes are also excluded.
     fn get_spendable_note(
         &self,
         txid: &TxId,
         protocol: ShieldedPool,
         index: u32,
         target_height: TargetHeight,
+        include_locked: bool,
     ) -> Result<Option<ReceivedNote<Self::NoteRef, Note>>, Self::Error>;
 
     /// Returns a list of spendable notes sufficient to cover the specified target value, if
     /// possible. Only spendable notes corresponding to the specified shielded protocol will
-    /// be included.
+    /// be included. When `include_locked` is `false`, locked notes are excluded.
+    #[allow(clippy::too_many_arguments)]
     fn select_spendable_notes(
         &self,
         account: Self::AccountId,
@@ -1680,16 +1683,18 @@ pub trait InputSource {
         target_height: TargetHeight,
         confirmations_policy: ConfirmationsPolicy,
         exclude: &[Self::NoteRef],
+        include_locked: bool,
     ) -> Result<ReceivedNotes<Self::NoteRef>, Self::Error>;
 
     /// Returns the list of notes belonging to the wallet that are unspent as of the specified
-    /// target height.
+    /// target height. When `include_locked` is `false`, locked notes are excluded.
     fn select_unspent_notes(
         &self,
         account: Self::AccountId,
         sources: &[ShieldedPool],
         target_height: TargetHeight,
         exclude: &[Self::NoteRef],
+        include_locked: bool,
     ) -> Result<ReceivedNotes<Self::NoteRef>, Self::Error>;
 
     /// Returns metadata describing the structure of the wallet for the specified account.
@@ -1698,6 +1703,7 @@ pub trait InputSource {
     /// - notes that are not considered spendable as of the given `target_height`
     /// - unspent notes excluded by the provided selector;
     /// - unspent notes identified in the given `exclude` list.
+    /// - when `include_locked` is `false`, locked notes.
     ///
     /// Implementations of this method may limit the complexity of supported queries. Such
     /// limitations should be clearly documented for the implementing type.
@@ -1707,6 +1713,7 @@ pub trait InputSource {
         selector: &NoteFilter,
         target_height: TargetHeight,
         exclude: &[Self::NoteRef],
+        include_locked: bool,
     ) -> Result<AccountMeta, Self::Error>;
 
     /// Fetches the transparent output corresponding to the provided `outpoint` if it is considered
@@ -1714,11 +1721,13 @@ pub trait InputSource {
     ///
     /// Returns `Ok(None)` if the UTXO is not known to belong to the wallet or would not be
     /// spendable in a transaction mined in the block at the target height.
+    /// When `include_locked` is `false`, locked outputs are also excluded.
     #[cfg(feature = "transparent-inputs")]
     fn get_unspent_transparent_output(
         &self,
         _outpoint: &OutPoint,
         _target_height: TargetHeight,
+        _include_locked: bool,
     ) -> Result<Option<WalletTransparentOutput<Self::AccountId>>, Self::Error> {
         unimplemented!(
             "InputSource::get_spendable_transparent_output must be overridden for wallets to use the `transparent-inputs` feature"
@@ -1738,6 +1747,7 @@ pub trait InputSource {
     ///
     /// Any output that is potentially spent by an unmined transaction in the mempool should be
     /// excluded unless the spending transaction will be expired at `target_height`.
+    /// When `include_locked` is `false`, locked outputs are also excluded.
     #[cfg(feature = "transparent-inputs")]
     fn get_spendable_transparent_outputs(
         &self,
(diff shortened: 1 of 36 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -291,7 +309,10 @@ impl Balance {
 
     /// Returns the total value of funds represented by this [`Balance`].
     pub fn total(&self) -> Zatoshis {
-        (self.spendable_value + self.change_pending_confirmation + self.value_pending_spendability)
+        (self.spendable_value
+            + self.locked_value
```
dannywillems, line 316: Let's ensure that `spendable_value` is not including `locked_value` already.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -148,12 +150,22 @@ pub(super) fn all_migrations<
...
-    //                                                /                  \              |
-    //                                               /                    \     ironwood_pool_code_views
+    //                                                  /              \                      \
+    //                                                 /                \               ironwood_received_notes
+    //                                                /                  \                     /           \
+    //                                               /                    \    ironwood_pool_code_views   note_locking
+    //                                              /                      \
     //                                          ivk_item_cache    add_transparent_receiver_address_index
     //
+    //    note_locking only appends a nullable lock_expiry_height column to the four
```
dannywillems: Nit: do we want this comment?

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5264,6 +5289,335 @@ pub(crate) fn get_block_range(
...
+    for table in [
+        "sapling_received_notes",
+        "orchard_received_notes",
+        "ironwood_received_notes",
+        "transparent_received_outputs",
+    ] {
+        rows_updated += conn.execute(
+            &format!(
+                "UPDATE {table} SET lock_expiry_height = NULL
+                 WHERE lock_expiry_height IS NOT NULL
```
dannywillems, line 5444: Review comment: even if the lock expiry height has passed, this will clean it. LGTM.

zcash_client_backend/src/data_api.rs
```diff
@@ -243,13 +246,28 @@ impl Balance {
...
+    /// Returns the value in the account that is currently "locked".
+    ///
+    /// The outputs that comprise this balance are seen by the wallet as being committed to be
+    /// spent by a transaction proposal or PCZT.
+    pub fn locked_value(&self) -> Zatoshis {
+        self.locked_value
+    }
+
     /// Adds the specified value to the spendable total, checking for overflow.
     pub fn add_spendable_value(&mut self, value: Zatoshis) -> Result<(), BalanceError> {
```
dannywillems, line 261: This should be changed to not include the locked value?

zcash_client_sqlite/src/wallet/common.rs
```diff
@@ -265,14 +287,17 @@ where
              AND rn.recipient_key_scope IS NOT NULL
              AND rn.nf IS NOT NULL
              AND rn.commitment_tree_position IS NOT NULL
-             AND rn.id NOT IN ({})
+             AND rn.id NOT IN ({}) -- the note is unspent
+             AND ({}) -- the note is not locked
              GROUP BY rn.id",
-            spent_notes_clause(table_prefix)
+            spent_notes_clause(table_prefix),
+            output_unlocked_condition("rn"),
```
dannywillems: Seems that `rn` is for the `received notes`.

zcash_client_sqlite/src/lib.rs
```diff
@@ -845,6 +857,7 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> InputSour
         &self,
         outpoint: &OutPoint,
         target_height: TargetHeight,
+        _include_locked: bool,
```
dannywillems: This is always ignored.

zcash_client_sqlite/src/lib.rs
```diff
@@ -845,6 +857,7 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> InputSour
...
     ) -> Result<Option<WalletTransparentOutput<Self::AccountId>>, Self::Error> {
         wallet::transparent::get_wallet_transparent_output(
```
dannywillems, line 861: The method should also support the locked parameter I think.

zcash_client_backend/src/data_api.rs
nuttycom: That isn't a change in this location; it's a change at the caller.

zcash_client_backend/src/data_api.rs
nuttycom: Also an issue for the caller; this is just the balance struct.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5264,6 +5289,335 @@ pub(crate) fn get_block_range(
     .map_err(SqliteClientError::from)
 }
 
+#[cfg(any(test, feature = "test-dependencies"))]
+pub(crate) fn get_locked_outputs(
```
nuttycom, line 5294: There's a lot of benign duplication in this function; we should reduce it in a follow-up.

zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -148,12 +151,24 @@ pub(super) fn all_migrations<
...
+    // off to the side of the tree above:
+    //
+    //                                     ironwood_received_notes
+    //                                                |
+    //                              orchard_ironwood_migration_tables
+    //                                                |
+    //                           orchard_ironwood_migration_account_key
+    //                                                |
+    //                            orchard_ironwood_migration_lock_owner
+    //
```
nuttycom: Human TODO: integrate this into the graph.

schell:
👀

nuttycom (approved):
tentative utACK a271fe918d0620014e595fba6c3e39781d7b3798

schell:
No blocking issues from me, utACK [a271fe9](https://github.com/zcash/librustzcash/pull/2716/commits/a271fe918d0620014e595fba6c3e39781d7b3798).

PR #2717, opened 2026-07-22
zcash_pool_migration_backend: test single-quantum migration

dannywillems:
Adds test coverage for migrating a balance of exactly **one quantum + fees**,
the smallest migratable balance. A quantum is a canonical `{1, 2, 5} * 10^k`
denomination; its fees are the ZIP-317 transfer buffer that funds the crossing
note plus one padded preparation-transaction fee.

Before this PR, only `below_min_note_migrates_nothing` covered the region
around this boundary, and it only exercises the *just-below* case (migrates
nothing). Neither the split layer nor the full pipeline had an at-boundary test
pinning that exactly one quantum + fees migrates that one denomination cleanly.

### Note-split unit test (`note_splitting/strategies.rs`)

The `CanonicalOneTwoFive` split in isolation: a balance of exactly one quantum
plus its fees splits into exactly one crossing note of that denomination, with
no change.

- `single_quantum_plus_fees_migrates_exactly_one_note` — a few example quanta
  from the 0.01 ZEC minimum denomination to the 10,000 ZEC cap.
- `single_quantum_plus_fees_is_one_note` — a proptest over an `arb_quantum`
  strategy that enumerates every valid `{1, 2, 5} * 10^k` denomination, with the
  sampled preparation fee bounded below the smallest inter-denomination gap
  (`RESIDUAL_MIGRATION_MIN`) so the greedy must pick exactly the quantum.

### Full-migration integration test (`tests/engine_plan.rs`)

`full_migration_of_one_quantum_is_one_layer_one_transaction` drives the whole
`plan_migration` pipeline (note splitting then preparation) for a lone source
note of exactly one quantum plus its fees. It asserts the result is **one
preparation layer holding exactly one transaction**, minting the single funding
note, with no source-pool change and a single scheduled transfer.

The canonical fees depend only on the network and height (not the balance), so
they are discovered from a probe migration and reused to build the exact
per-quantum balance across a few example denominations.

### Verification

- Full `zcash_pool_migration_backend` suite green (unit + doc + integration).
- `cargo clippy --tests` clean.
- `cargo fmt --check` clean.

### Follow-up (separate PR)

A prover-backed full end-to-end test built on top of `prove/pool-migration-transfer`
(#2710), exercising split -> preparation -> transfer proving rather than only
planning.

### DIFF
--- a/zcash_pool_migration_backend/src/note_splitting/strategies.rs
+++ b/zcash_pool_migration_backend/src/note_splitting/strategies.rs
@@ -332,6 +332,115 @@ mod tests {
         assert_eq!(p.change(), Some(zat(below)));
     }
 
+    /// The largest reserved preparation fee the single-quantum tests sample. Kept below the smallest
+    /// gap between adjacent denominations (`RESIDUAL_MIGRATION_MIN`, the 0.01 -> 0.02 ZEC step) so
+    /// that `quantum + fee` never rounds up to the next `{1, 2, 5} * 10^k` denomination: the plan
+    /// must then pick exactly the quantum. Realistic ZIP-317 preparation fees are far smaller (a
+    /// handful of marginal fees).
+    const MAX_SINGLE_QUANTUM_PREP_FEE_ZATOSHI: u64 = COIN / 200; // 0.005 ZEC, half the minimum denom
+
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK. It would be useful to have a similar test that, instead of balancing exactly (which may be a rare special case) instead results in a single transfer plus a residual.

PR #2718, opened 2026-07-22
zcash_pool_migration: prove the minimal one-quantum migration end to end

dannywillems:
**Stacked PR** (draft). Depends on, and is based on:
- #2710 (`prove/pool-migration-transfer`) — the real Orchard+Ironwood transfer proving and the E2E chain-simulation harness this builds on.
- #2717 (`test/migration-single-quantum`) — the planner-only single-quantum tests, cherry-picked here so this branch has both. Its two commits appear in this PR's range and will drop out once #2717 merges.

Base is set to `prove/pool-migration-transfer` so the diff shows only the incremental work. Once #2710 (and #2717) merge, this will be rebased onto `main` and retargeted.

### What this adds

The prover-backed companion to #2717's planner-only
`full_migration_of_one_quantum_is_one_layer_one_transaction`: a wallet funded
with exactly **one quantum plus its fees** migrates that one denomination as a
single preparation transaction and a single transfer, proven end to end against
the wallet's own Orchard commitment tree.

A quantum is a canonical `{1, 2, 5} * 10^k` denomination; its fees are the
ZIP-317 transfer buffer that funds the crossing note plus one padded
preparation-transaction fee.

`single_quantum_migration_proves_end_to_end` (in #2710's
`tests/prove_chain_sim.rs`) reuses the chain-simulation harness (wallet factory,
block cache, migration store, prove/mine/scan helpers) and:

1. Discovers the canonical ZIP-317 fees from a cheap `MockBackend` probe (the
   fees use a fixed fee rule, so they are network- and height-independent, and
   `canonical_fees` is private), then funds a real in-memory `WalletDb` with
   exactly `quantum + transfer_buffer + prep_tx_fee`.
2. Asserts the plan is one crossing of exactly the quantum, no source-pool
   change, one preparation layer, one transaction.
3. Proves the single preparation (asserts Orchard-only), mines and scans it,
   advances to the transfer's drawn boundary, proves the transfer, and asserts
   both its Orchard and Ironwood bundles verify.

### Verification

- Both E2E proving tests pass (`2 passed`, ~33s with real halo2 proving).
- `cargo clippy --test prove_chain_sim` clean.
- `cargo fmt --check` clean.

### DIFF
--- a/zcash_pool_migration_backend/tests/prove_chain_sim.rs
+++ b/zcash_pool_migration_backend/tests/prove_chain_sim.rs
@@ -49,10 +49,14 @@ use zcash_pool_migration_backend::engine::{
     PoolMigrationWrite,
 };
 use zcash_pool_migration_backend::wallet::{WalletMigration, WalletMigrationProver};
+use zcash_pool_migration_memory::{MockBackend, regtest_network};
 
 /// Every network upgrade (through NU6.3, which activates the Ironwood pool) is active from this
 /// height, so a migration built at or above it is post-NU6.3 and its transfers cross into Ironwood.
 const ACTIVATION: u32 = 100_000;
+/// The single quantum the minimal end-to-end migration moves: the 0.01 ZEC minimum denomination
+/// (the smallest `{1, 2, 5} * 10^k` value the note-split planner emits).
+const SINGLE_QUANTUM_ZATOSHI: u64 = COIN / 100;
 /// Empty blocks scanned after a note is received so its commitment-tree shard completes and an
 /// anchor at that height is available.
 const SHARD_COMPLETION_BLOCKS: usize = 5;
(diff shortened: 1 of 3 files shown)


zcash_pool_migration_backend/tests/prove_chain_sim.rs
```diff
@@ -401,3 +405,184 @@ fn migration_proves_end_to_end_against_a_funded_wallet() {
...
+///
+/// This is the prover-backed companion to the planner-only
+/// `full_migration_of_one_quantum_is_one_layer_one_transaction` (in `tests/engine_plan.rs`): where
+/// that test asserts the PLAN shape, this one funds a real wallet with exactly that balance and
+/// proves the resulting single preparation and single transfer end to end against the wallet's own
+/// Orchard commitment tree. A quantum is a canonical `{1, 2, 5} * 10^k` denomination; its fees are
+/// the ZIP-317 transfer buffer that funds the crossing note plus one padded preparation-transaction
+/// fee, so the whole balance migrates as one crossing note with no source-pool change.
+#[test]
+fn single_quantum_migration_proves_end_to_end() {
```
dannywillems, line 563: Note: we're rewriting the test framework in 2710, and therefore it will be required to update this test.

PR #2719, opened 2026-07-22
zcash_pool_migration_backend: require CryptoRng for planning randomness

dannywillems:
## Summary

Closes a type-level gap found in a security review of the pool-migration
randomness: the commit/build entry points already bound their RNG as
`RngCore + CryptoRng`, but the planning surface accepted any `RngCore` and
enforced the documented "must be a cryptographically secure RNG" requirement
only in prose.

The planning draws decide the privacy-relevant observables of a migration:
the broadcast order (ZIP 318 SHUFFLE), the exponential inter-arrival delays,
the anchor-boundary selection, and (for a randomized strategy) the crossing
denominations. A predictable generator there would let a chain observer
reconstruct the schedule, so the `CryptoRng` marker bound is now required on
the public planning API:

- `engine::plan_migration` and `engine::estimate_migration_runs`
- `scheduling::{draw_delay, draw_prep_delay, shuffle_indices, shuffle_in_place, schedule_broadcast_heights, schedule_prep_broadcast_heights, schedule, draw_anchor_boundary}`
- `note_splitting::plan_note_split` and `note_splitting::DenominationStrategy::plan`

Private helpers keep the plain `RngCore` bound; they are reachable only
through the bounded public functions.

No behavior change and no caller changes: the crate's tests already use
`rand_chacha::ChaCha8Rng`, which implements `CryptoRng`. The crate is
unreleased, so no changelog entry beyond the existing "Initial release"
section.

## Verification

- `cargo build --release -p zcash_pool_migration_backend --all-features`
- `cargo test --release -p zcash_pool_migration_backend --all-features` (106 passed)
- `cargo clippy --release -p zcash_pool_migration_backend -p zcash_pool_migration_memory --all-features --all-targets` (clean)
- `cargo build --release -p zcash_client_sqlite --all-features`
- `cargo fmt -p zcash_pool_migration_backend -- --check`

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -642,7 +642,7 @@ pub fn plan_migration<P, B, R>(
 where
     P: zcash_protocol::consensus::Parameters,
     B: MigrationBackend,
-    R: RngCore,
+    R: RngCore + rand_core::CryptoRng,
 {
     let notes = backend
         .spendable_orchard_note_values()
@@ -913,7 +913,7 @@ pub fn estimate_migration_runs<P, B, R>(
 where
     P: zcash_protocol::consensus::Parameters,
     B: MigrationBackend,
-    R: RngCore,
+    R: RngCore + rand_core::CryptoRng,
 {
     let height = backend
         .chain_tip_height()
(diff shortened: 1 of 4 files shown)


nuttycom (approved):
utACK with mild reservations; usually we require `CryptoRng` at the use site, so as not to inhibit testing with arbitrary RNGs. This is a design question that we've gone back and forth on; it results in stuff like https://github.com/zcash/librustzcash/blob/44984d48caf4cbc5f8f0c1502fa0501567fd8df7/zcash_primitives/src/transaction/builder.rs#L1784

PR #2720, opened 2026-07-22
zcash_client_sqlite: key the pool-migration store by account row

nuttycom:
Follow-up to #2712, correcting a design decision made there.

#2712 keyed the SQLite pool-migration store by the account UUID, stored as a bare `account_uuid BLOB` column with no foreign key into `accounts`. This uses a row-oriented foreign key instead, matching every other account-owned table in the schema.

## What changed

- **Schema**: `orchard_ironwood_migrations.account_uuid BLOB NOT NULL` → `account_id INTEGER NOT NULL REFERENCES accounts(id) ON DELETE CASCADE`, with the unique index now over `account_id`.
- **API boundary**: `PoolMigrations::for_account` still takes an `AccountUuid`, but now resolves it to the account's row (`AccountRef`) up front and keys the migration by that row internally. It became fallible (`Result<_, Error>`); a new `Error::AccountUnknown` covers a UUID with no matching account.
- **Deletion**: the `ON DELETE CASCADE` removes an account's migration (and its child rows in turn) when the account is deleted, so the explicit `delete_account_migration` helper and its call in `delete_account` are gone. A test asserts the cascade fires.
- **Migration**: since these tables have not shipped in a release, the fix is applied at the source — the separate `orchard_ironwood_migration_account_key` migration is deleted, and the initial `orchard_ironwood_migration_tables` migration now creates the account-keyed shape directly.

## Notes for reviewers

- **`for_account` is now fallible.** Downstream consumers of this unreleased surface (e.g. the mobile SDK) will need to handle the `Result`.
- **Developer databases from this unreleased window must be recreated.** The fix lives in the initial migration with no catch-up, so a database already on the singleton or `account_uuid` shape will not auto-upgrade. Nothing in a public release is affected (latest release is `0.22.0-rc.1`; these tables are in `[Unreleased]`).

## Testing

`cargo test -p zcash_client_sqlite --all-features`: 400 passed, 0 failed (+ doc-tests). Includes `verify_schema` (builds a real wallet DB through the full migration chain and matches the golden schema), the migration-DAG leaf-consistency tests, and the new account-deletion cascade test with foreign keys enforced. `cargo clippy --all-features --all-targets` and `cargo fmt` are clean.


dannywillems (approved):
Makes more sense. utACK [ff15da7](https://github.com/zcash/librustzcash/pull/2720/commits/ff15da7c8fa0fd4bf9e087996b05e29adefb65c3)

nuttycom (commented):
self-utACK

PR #2723, opened 2026-07-22
zcash_pool_migration_backend: detect and rebuild expired migration transactions

dannywillems:
## Motivation

A ZIP 318 audit of the migration code flagged that the state machine ignores
transaction expiry. A pre-signed migration transaction can only be mined in a
block at or below its expiry height (ZIP 203), and the migration's expiry
window is a rolling one-to-two months. Two concrete gaps followed:

- `next_broadcastable` offered an **expired** transaction for broadcast. A
  wallet resumed after its broadcast windows lapsed (dormant past the expiry
  window) would be told to broadcast a transaction the node rejects, with no
  recovery path.
- `next_step` had no way to express "this transaction must be rebuilt", so an
  expired transaction blocking its dependents would leave the migration on
  `Waiting` indefinitely.

This PR closes both halves of that finding: the pure state machine now
**detects** expiry and surfaces a rebuild decision, and the engine now
**rebuilds** the expired transfer to act on it.

Review of an earlier revision surfaced the constraint that shapes the
solution: the transaction signature hash covers the expiry height, so nothing
of an expired artifact is reusable — the part must be carried by an entirely
new transaction, **signed anew**. Signing needs the account's spend authority
(unlike proving and broadcasting, which need only the viewing key, the
commitment tree, and the network), which drives both the in-process/external
split of the rebuild API and the restriction of the advance step to transfers.

### Detection and advance decision (`zcash_pool_migration_backend/src/state.rs`)

- `is_expired(tx, target_height)`: a not-yet-mined transaction is expired once
  `expiry_height < target_height`. An `expiry_height` of `0` never expires; a
  `Mined` transaction is never expired (it was included in time). Expiry is
  **computed** from the persisted height and the tip, not stored as a
  `MigrationTxState` variant, so it cannot go stale as the chain advances.
- `next_provable` and `next_broadcastable` skip expired transactions.
- `next_step` returns a new `AdvanceStep::Rebuild { id }` for an expired
  **transfer**, after any still-valid prove/broadcast work and in preference
  to `Waiting`, so the migration surfaces the rebuild rather than stalling.
  Only a transfer — a leaf of the dependency graph — is surfaced: an expired
  **preparation** has no single-transaction remediation (its dependents'
  pre-signatures commit to the notes it would have minted, so rebuilding it
  means re-signing the whole dependent subtree, a follow-on slice); it stays
  visible through `Blocker::Expired` and `expired_transactions`.
- `transaction_statuses` reports a new `Blocker::Expired` and exposes each
  transaction's `expiry_height`, so a wallet UI can show "needs rebuild" and
  expiry proximity.
- `expired_transactions(target_height)`: the on-launch detection a wallet runs
  to list the transactions whose parts must be re-carried.

### Rebuild (`zcash_pool_migration_backend/src/engine.rs`)

- `rebuild_expired_transfer` makes the `AdvanceStep::Rebuild` signal
  actionable by constructing an entirely **new** transaction for the part: it
  reschedules the transfer forward from the current tip with a fresh
  memoryless delay, derives the new canonical expiry, draws a fresh boundary
  anchor, builds a new PCZT against the same funding note and crossing value,
  and signs it **anew in-process**, storing it back as `Signed`. This is
  ZIP 318's expired-transaction handling: a new transaction with a fresh
  anchor and expiry for the affected part, its denomination unchanged.
- `rebuild_expired_transfer_unsigned` serves migrations signed by an
  **external** (hardware or offline) signer, whose spend authority is not
  available in-process: it constructs the same new transaction but leaves it
  in `AwaitingSignature` and returns an `UnsignedMigrationTx` to route to the
  device (batchable by action budget), with `apply_signature` completing it to
  `Signed` — mirroring `commit_preparation` / `build_preparation_unsigned`.
- The funding note is recovered by **identity**, not value: the expired PCZT's
  one real spend reveals the note's nullifier, which is matched among the
  wallet's spendable Orchard notes. Denominations repeat across a migration,
  so a value-based match could grab a sibling transfer's funding note and turn
  the rebuilt transfer into a double-spend of that still-valid sibling.
  Anchors and witnesses stay deferred (ZIP 374): the fresh anchor is installed
  at proving time, exactly as for an originally committed transfer. The caller
  persists the updated state afterwards, as after proving.
- A typed `RebuildError` rejects a still-valid, mined, or non-transfer
  transaction, and reports (via `FundingNoteUnavailable`) a funding note that
  is gone (spent outside the migration) so the caller re-plans the remaining
  balance instead of rebuilding the part — never substituting a different
  spendable note of coincidentally equal value.

## Scope

This is the detection, reporting, advance-decision, and rebuild slice of the
audit's "error-handling states" finding, all in the pure/backend-agnostic state
module and the engine, so it is testable without a chain and usable by every
consumer (a mobile wallet using the crates directly, or Zallet). Expired
**preparation** remediation (re-signing the affected subtree), invalid
(spent-outside-the-migration) input detection, reorg handling, and the
one-overdue-per-open throttle remain separate follow-on slices.

## Testing

- Detection: an expired transaction is not proved or broadcast; `next_step`
  returns `Rebuild` for an expired transfer after still-valid work, and never
  for an expired preparation (reported via `Blocker::Expired` /
  `expired_transactions` instead); `Blocker::Expired` and `expiry_height` are
  surfaced; a mined-past-expiry transaction is not expired; a terminal
  (cancelled) migration is never offered for rebuild; and
  `expiry_height == 0` never expires.
- Rebuild: `rebuild_expired_transfer` reschedules, re-anchors, and re-signs an
  expired transfer back to `Signed`; the rebuilt transfer spends exactly the
  note its expired PCZT spent (matched by nullifier) even when a sibling
  transfer's funding note has the same value; a same-value-but-different note
  yields `FundingNoteUnavailable`; `rebuild_expired_transfer_unsigned` leaves
  the rebuilt transfer awaiting the external signature (no in-process
  authorization of the account's spend) and round-trips through
  `apply_signature`; `RebuildError` rejects a still-valid, unknown, or
  non-transfer transaction.
- Two engine lifecycle tests used a target height past the schedule's expiry
  window purely to mean "everything is due"; they now derive their target from
  the latest scheduled height, so they exercise dependency ordering rather
  than tripping the expiry path.
- `cargo test -p zcash_pool_migration_backend --all-features` green (109 unit
  tests, the integration walk, and the real-proving chain simulation); clippy
  `-D warnings` clean; `zcash_pool_migration_memory` and `zcash_client_sqlite`
  build clean.

Stacked on `prove/pool-migration-transfer`.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -274,8 +274,11 @@ pub struct MigrationTransaction {
     /// dependencies to mine and a boundary to pass rather than a fixed height).
     #[getset(get_copy = "pub")]
     pub(crate) scheduled_height: BlockHeight,
-    /// The height after which the transaction is invalid and must be rebuilt
-    /// (rebuild-on-expiry is grown by a later slice, alongside reconciliation-on-launch).
+    /// The height after which the transaction can no longer be mined (ZIP 203) and its part must
+    /// be carried by an entirely new, freshly signed transaction: a transfer via
+    /// [`rebuild_expired_transfer`] / [`rebuild_expired_transfer_unsigned`] (detection via
+    /// [`MigrationState::expired_transactions`]); an expired preparation awaits its own
+    /// remediation slice.
     #[getset(get_copy = "pub")]
     pub(crate) expiry_height: BlockHeight,
     /// The boundary height whose tree state the transaction proves against. For a transfer this is
@@ -455,8 +458,9 @@ pub struct MigrationState {
     #[getset(get = "pub")]
     pub(crate) note_split: NoteSplitPlan,
     /// The preparation plan (its layers and direct-funding notes), retained for auditability and
-    /// for the rebuild-on-expiry slice. A `Preparation { layer, index }` transaction's spends
-    /// resolve against `preparation.layers()[layer][index]`.
+    /// for rebuilding expired PREPARATION transactions (a follow-on slice). A
+    /// `Preparation { layer, index }` transaction's spends resolve against
+    /// `preparation.layers()[layer][index]`.
     #[getset(get = "pub")]
     pub(crate) preparation: PreparationPlan,
     /// Every migration transaction, in dependency order.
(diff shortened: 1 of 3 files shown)


zcash_pool_migration_backend/src/engine.rs
```diff
@@ -1341,6 +1066,225 @@ where
...
+        }
+    }
+}
+
+#[cfg(feature = "orchard")]
+impl<E: core::error::Error> core::error::Error for RebuildError<E> {}
+
+/// Rebuild an EXPIRED migration transfer in place, moving it back to
+/// [`Signed`](MigrationTxState::Signed) with a fresh boundary anchor and canonical expiry and an
+/// unchanged denomination.
```
nuttycom: It's not possible to rebuild an expired migration transfer and move it back to Signed, because the transaction signature hash covers the expiry height. An expired transaction must be entirely reconstructed and signed anew.

nuttycom (changes requested):
Rebuild is incorrect; new signatures must be created since the signature hash covers the expiry height.

nuttycom (approved):
utACK 034255bfbc32170cb3d63ee27a6cd51a531f3de1

PR #2726, opened 2026-07-22
Note locking: review fixes, SQL predicate proptests, and scenario tests

dannywillems:
Builds on #2716 (targets `schell/note-locking`) with fixes and test hardening from a deep review of the note-locking feature. Work in progress; opened as draft so progress is visible.

## Fixes

- **Locked proposals now round-trip through their serialized form.** `try_into_standard_proposal` retrieved inputs with `include_locked: false`, so a proposal created with `lock_for_blocks: Some(_)` (which locks its own inputs) failed to decode with `InputNotFound` for every locked shielded input; this broke the persist-and-restore flow the locking feature exists to support. Decoding now retrieves inputs with `include_locked: true`; a `locked_proposal_proto_roundtrip` regression test was verified to fail against the previous behavior.
- **SQLite `get_unspent_transparent_output` now honors `include_locked`.** The impl accepted the parameter but ignored it, violating the trait contract and diverging from the shielded path.
- **`ProposalError::InputsLocked`.** Lock conflicts during proposal creation previously surfaced as `ChainDoubleSpend` ("attempts to spend the same output twice"), which is misleading for a transient account-busy condition; they now surface as a dedicated variant callers can match on for retry logic.

## Documentation

- `lock_outputs`: inputs must be distinct; the transaction-scoped SQLite impl relies on caller rollback for the all-or-nothing contract.
- `clear_locked_outputs` / `unlock_proposal_inputs`: locks record no owner, so these can release locks held by still-in-flight proposals; documented when they are safe to call.
- `propose_*`: lock expiry re-opens the selection race if proving outlasts the lock window; choose `lock_for_blocks` conservatively.

## Tests

Property-based (proptest), evaluating the actual SQL fragments in SQLite against executable models:

- `include_locked: true` disables lock filtering entirely.
- The selection predicate is the exact complement of the locked-balance predicate: an output is never both selectable and counted as locked, and never neither. This pins the passed-expiry semantics: once `target_height` exceeds `lock_expiry_height`, the output is selectable again with no unlock call.
- The lock-acquisition guard matches its model, including the NULL-chain-tip case.
- No lock stealing: a replaceable lock is always already expired for selection at `target_height = chain_tip + 1`.

Deterministic pool tests (Sapling + Orchard):

- `locked_proposal_proto_roundtrip`: regression test for the decoding fix.
- `lock_expiry_restores_spendability`: chain advance past the expiry restores spendability with no unlock call, and an expired lock is replaced by a fresh `lock_outputs`.
- `lock_conflict_and_batch_atomicity`: `LockFailure` on active locks (including same-caller re-lock and duplicated refs in one batch); a failed batch locks nothing; `unlock_output` return-value semantics.
- `unlock_proposal_inputs_releases_locks`: the abandoned-proposal recovery path.

The four per-pool `UPDATE` statements in `lock_outputs`/`unlock_output` were also deduplicated via a shared `received_outputs_table` lookup and `output_lockable_condition` helper (which is what makes the acquisition guard property-testable).

## More property-based tests

- **Model-based stateful proptest** (`note_locking_model`, Sapling + Orchard): random sequences of `lock`/`unlock`/`clear`/`mine` operations (expiry deltas and mining counts drawn small so sequences routinely cross expiry boundaries) run against both the real store and a trivial in-memory model of per-note lock expiry plus the chain tip. After every operation the store must agree with the model on the operation outcome (including all-or-nothing failure of conflicting batches), the `get_locked_outputs` set, and the balance decomposition (locked value = sum of model-locked notes, spendable = remainder, total unaffected by lock state).
- **`Balance` bucket invariants**: adds succeed exactly while the overflow guard permits, mutate only the requested bucket, and leave the balance unchanged on failure; `total()` is always the sum of the four participating buckets (uneconomic value never contributes and is guarded only against its own overflow); `Balance + Balance` is componentwise with the equivalent overflow condition.
- **`OutputRef` identity**: `From<NoteId>` preserves every component; identity is exactly the (txid, pool, output index) triple with `Ord` consistent with `Eq` (the `BTreeSet` double-spend check and the lock tables key on this).

## Transparent path

- `transparent_note_locking`: a locked UTXO is excluded from `get_unspent_transparent_output` and `get_spendable_transparent_outputs` unless `include_locked` is set (regression test for the fixed parameter, verified to fail against the old behavior), is reported as locked value in the per-address balances with an unchanged total, conflicts with a second lock, and returns to spendability once the chain tip passes the expiry height.

## Cross-connection concurrency

- `concurrent_handles_resolve_lock_conflict`: two independent `WalletDb` connections to the same database file, simulating the shared select-before-lock state of two concurrent proposal flows. Both handles observe the same spendable note; only the first lock succeeds; the loser's failure names the contested note; lock/unlock changes are immediately visible across handles; and the ownerless-lock property is pinned across connections.

## Edge behaviors pinned

- Locking an unknown output fails with `LockFailure`; locking an already-spent note currently succeeds with no balance effect (pinned so any future tightening of the contract is a deliberate change).
- CHANGELOG "Fixed" entries added for the two behavior fixes.

## Owner-keyed locking (`LockOwner` / `LockRequest`)

Locks now record the flow that acquired them, via an opaque caller-held `LockOwner` token stored in a `lock_owner` column (added by a second migration):

- A lock may be acquired when the output holds no lock, when the existing lock has expired as of the chain tip, or when the existing lock is held by the **same owner** (an idempotent acquire/extend, making crash-retry of a flow that already locked its inputs safe). Acquisition fails only on an active lock held by a **different** owner, preserving the no-stealing invariant (property-tested: a foreign lock is replaceable exactly when it is already expired for selection).
- `unlock_output` / `unlock_proposal_inputs` release only locks held by the given owner, closing the release-another-proposal's-locks hazard. `clear_locked_outputs` remains the owner-agnostic recovery hammer.
- The `propose_*` functions take `lock_inputs: Option<LockRequest>` (owner + expiry window) in place of `lock_for_blocks: Option<u32>`, and `propose_shielding_coinbase` gains the same parameter (it previously offered no locking, forcing Zallet to re-implement it).
- `lock_outputs` now takes `&[OutputRef]` instead of `impl Iterator`, restoring dyn-compatibility of `WalletWrite`.
- All predicate proptests, the model-based stateful proptest, and the deterministic/cross-connection tests gained an owner dimension (same-owner re-lock, cross-owner conflict and atomicity, owner-scoped unlock, owner takeover of expired locks).

Also fixed here, found by the integration suite below: the **coinbase branch of the transparent balance tally ignored locks**; a mature coinbase UTXO locked by an in-flight shielding proposal was reported as spendable while selection refused it.

## Downstream: Zallet integration and Python end-to-end suite

- **Zallet**: zcash/zallet#672 ([branch `danny/note-locking-zallet`](https://github.com/zcash/wallet/tree/danny/note-locking-zallet)) wires locking into the spend RPCs (`builder.note_lock_blocks` config, default 40 blocks; one `LockOwner` per operation with `LockRequest` into the propose calls; retryable "account is busy" error mapping; owner-scoped unlock-on-failure guards; a `z_clearlockedoutputs` recovery RPC; `lockedZat` in `z_getbalanceforaccount`). The Zallet branch is based on Zallet `main` and pins this PR's branch directly; the note-locking integration is fully independent of the pool-migration work.
- **Integration tests**: [`zcash/integration-tests` branch `danny/note-locking-integration`](https://github.com/zcash/integration-tests/tree/danny/note-locking-integration) (PR zcash/integration-tests#180) adds ten note-locking scenarios across five `qa/rpc-tests` suites (locked-balance visibility, concurrent-send double-spend prevention, unlock-on-broadcast, expiry restoring funds, locks surviving a SIGKILL restart, clear-locks recovery, account isolation, all-notes-locked errors, shielding with locking, and the short-lock-window self-heal trade-off). Both downstream branches are based on their repos' `main` and are independent of the pool-migration work. **All ten scenarios pass live** against the owner-keyed API, e.g. `ZALLET_BACKEND=zebra uv run ./qa/rpc-tests/wallet_note_locking.py`; the shielding suite behaviorally confirms the locked-coinbase balance fix.

Note for consumers pinning this branch: its `zcash_primitives` builds against the ZIP 374 deferred-anchor `orchard` builder API, so downstream manifests must mirror the workspace's git `orchard` entry (rev `3bd2b736`) in `[patch.crates-io]` alongside the librustzcash pins.

## Note for reviewers

`cargo test --release -p zcash_client_backend --all-features` has one pre-existing failure unrelated to this branch: `proposal::tests::orchard_pool_payment_with_ironwood_active_is_a_programming_error` expects a `debug_assert!` panic, which is compiled out under `--release`. It passes in debug mode and fails identically on the base branch.

### DIFF
--- a/zcash_client_backend/src/data_api/wallet/input_selection.rs
+++ b/zcash_client_backend/src/data_api/wallet/input_selection.rs
@@ -746,6 +746,10 @@ impl<DbT> GreedyInputSelector<DbT> {
             ),
         };
         *amount_at_transparent_gather = amount_at_gather;
+        // Input selection must never draw on locked outputs: a locked output belongs to
+        // another in-flight proposal, and spending it would recreate the conflict that
+        // locking exists to prevent.
+        let include_locked = false;
         Ok(wallet_db
             .select_spendable_transparent_outputs(
                 account,
@@ -756,7 +760,7 @@ impl<DbT> GreedyInputSelector<DbT> {
                 target_value,
                 shielding_max_inputs(self.shielding_block_space_percent),
                 &StandardFeeRule::Zip317,
-                false,
+                include_locked,
             )
             .map_err(InputSelectorError::DataSource)?
             .into_iter()
@@ -1322,6 +1326,10 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                         && required > amount_at_transparent_gather
                     {
                         let address_allow_list = transparent.address_allow_list();
+                        // Input selection must never draw on locked outputs: a locked output belongs to
+                        // another in-flight proposal, and spending it would recreate the conflict that
+                        // locking exists to prevent.
+                        let include_locked = false;
                         transparent_inputs = wallet_db
                             .select_spendable_transparent_outputs(
                                 account,
@@ -1332,7 +1340,7 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                                 TargetValue::AtLeast(required),
                                 shielding_max_inputs(self.shielding_block_space_percent),
                                 &StandardFeeRule::Zip317,
-                                false,
+                                include_locked,
                             )
                             .map_err(InputSelectorError::DataSource)?
                             .into_iter()
@@ -1356,6 +1364,10 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
             // by one policy: pool crossing is minimized, and a payment to an Orchard
             // receiver draws on the pool its output is constructed in (the Ironwood
             // pool once NU6.3 is active).
+            // Input selection must never draw on locked outputs: a locked output belongs to
+            // another in-flight proposal, and spending it would recreate the conflict that
+            // locking exists to prevent.
+            let include_locked = false;
             shielded_inputs = wallet_db
                 .select_spendable_notes(
                     account,
@@ -1364,7 +1376,7 @@ impl<DbT: InputSource> InputSelector for GreedyInputSelector<DbT> {
                     target_height,
                     confirmations_policy,
                     &exclude,
-                    false,
+                    include_locked,
                 )
                 .map_err(InputSelectorError::DataSource)?;
 
@@ -1488,6 +1500,10 @@ where
     InputSourceT: InputSource,
     FeeRuleT: FeeRule + Clone,
 {
+    // Input selection must never draw on locked outputs: a locked output belongs to
+    // another in-flight proposal, and spending it would recreate the conflict that
+    // locking exists to prevent.
+    let include_locked = false;
     let spendable_notes = wallet_db
         .select_spendable_notes(
             source_account,
@@ -1496,7 +1512,7 @@ where
             target_height,
             confirmations_policy,
             &[],
-            false,
+            include_locked,
         )
         .map_err(InputSelectorError::DataSource)?;
 
@@ -2176,13 +2192,17 @@ where
     // one query per address (including for the many addresses that have no spendable outputs),
     // which is prohibitively expensive for wallets that hold large numbers of transparent
     // addresses.
+    // Input selection must never draw on locked outputs: a locked output belongs to
+    // another in-flight proposal, and spending it would recreate the conflict that
+    // locking exists to prevent.
+    let include_locked = false;
     let mut utxos = wallet_db
         .get_spendable_transparent_outputs_for_addresses(
             source_addrs,
             target_height,
             confirmations_policy,
             output_filter,
-            false,
+            include_locked,
         )
         .map_err(InputSelectorError::DataSource)?;
 
--- a/zcash_client_backend/src/proto.rs
+++ b/zcash_client_backend/src/proto.rs
@@ -805,6 +805,14 @@ impl proposal::Proposal {
                 };
 
                 let target_height = TargetHeight::from(self.min_target_height);
+
+                // A proposal created with `lock_for_blocks` locks its own inputs, so
+                // input retrieval during decoding must not filter locked outputs;
+                // otherwise a locked proposal would fail to round-trip through its
+                // serialized form. Double-spend protection is enforced when the
+                // proposal's transactions are created, not here.
+                let include_locked = true;
+
                 // Steps are checked against the Orchard turnstile when Ironwood is
                 // active at the height for which the proposal was constructed.
                 #[cfg(feature = "orchard")]
@@ -875,7 +883,7 @@ impl proposal::Proposal {
                                                     .get_unspent_transparent_output(
                                                         &outpoint,
                                                         target_height,
-                                                        false,
+                                                        include_locked,
                                                     )
                                                     .map_err(ProposalDecodingError::InputRetrieval)?
                                                     .ok_or({
@@ -896,7 +904,7 @@ impl proposal::Proposal {
                                                 protocol,
                                                 out.index,
                                                 target_height,
-                                                false,
+                                                include_locked,
                                             )
                                             .map_err(ProposalDecodingError::InputRetrieval)
                                             .and_then(|opt| {
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -28,6 +28,15 @@ workspace.
 - `zcash_client_backend::data_api::error::LockError`
 - `zcash_client_backend::wallet::OutputRef`
 - `impl From<zcash_client_backend::wallet::NoteId> for zcash_client_backend::wallet::OutputRef`
+- `zcash_client_backend::wallet::LockOwner`, an opaque token identifying the
+  holder of an output lock. Locks record their owner: unlocking is scoped to
+  the owner that took the lock, and re-locking by the same owner is an
+  idempotent acquire/extend (supporting crash-retry of a flow that had already
+  locked its inputs), while an active lock held by a different owner cannot be
+  acquired or released until it expires.
+- `zcash_client_backend::data_api::wallet::LockRequest`, the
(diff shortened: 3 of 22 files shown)


zcash_client_backend/src/proto.rs
```diff
@@ -805,6 +805,14 @@ impl proposal::Proposal {
                 };
 
                 let target_height = TargetHeight::from(self.min_target_height);
+
+                // A proposal created with `lock_for_blocks` locks its own inputs, so
+                // input retrieval during decoding must not filter locked outputs;
+                // otherwise a locked proposal would fail to round-trip through its
+                // serialized form. Double-spend protection is enforced when the
+                // proposal's transactions are created, not here.
+                let include_locked = true;
```
dannywillems, line 814: This is a semantic change.

zcash_client_backend/src/proto.rs
```diff
@@ -875,7 +883,7 @@ impl proposal::Proposal {
                                                     .get_unspent_transparent_output(
                                                         &outpoint,
                                                         target_height,
-                                                        false,
```
dannywillems, line 878: See above. This is a semantic change.

zcash_client_backend/src/proto.rs
```diff
@@ -896,7 +904,7 @@ impl proposal::Proposal {
                                                 protocol,
                                                 out.index,
                                                 target_height,
-                                                false,
```
dannywillems, line 899: See above. This is a semantic change.

dannywillems:
## Design note: `LockOwner` vs a foreign key, and a proposed follow-up

While reviewing the ownership design we considered whether `lock_owner` should be a foreign key into an existing table instead of an opaque 32-byte token. Recording the analysis here for discussion.

### Why no existing table works as the owner

- **`transactions`**: at proposal time no transaction exists yet (and `txid` is NOT NULL UNIQUE, so no placeholder row is possible). The lock covers exactly the selection -> build -> prove -> sign -> store window, i.e. the phase *before* a txid is fixed. Once a transaction is stored, the spend tables already express "this note is held by transaction X" and `store_transactions_to_be_sent` clears the locks; a transaction-keyed lock would duplicate the spend tables while leaving the pre-transaction window unnamed. A multi-step proposal also builds several transactions, and a transaction rebuilt after a crash generally has a different txid, which would defeat the idempotent same-owner re-lock.
- **`accounts`**: too coarse; the point of ownership is distinguishing concurrent flows *within* one account.

So the entity that ought to own a lock, an in-flight spend operation, is not persisted anywhere in the schema today. The opaque token stands in for it. (Where a durable transaction identity does exist, e.g. a persisted PCZT whose effecting data is final, `impl From<TxId> for LockOwner` lets the txid serve as the token, so that flow can re-derive its owner after a restart.)

### Follow-up proposal: a persisted `tx_proposals` table

The design where a real FK becomes possible is a persisted in-flight-operations entity:

```sql
CREATE TABLE tx_proposals (
    id INTEGER PRIMARY KEY,
    account_id INTEGER NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
    lock_owner BLOB NOT NULL UNIQUE,  -- keeps LockOwner as the stable public identity
    created_height INTEGER NOT NULL,
    lock_expiry_height INTEGER NOT NULL,
    proposal BLOB  -- optionally: the serialized proposal itself
);
-- on the four received-notes tables, lock_owner/lock_expiry_height become:
-- lock_proposal_id INTEGER REFERENCES tx_proposals(id) ON DELETE SET NULL
```

This buys, beyond referential integrity:

- cascade release of all of a proposal's locks by deleting one row;
- enumerable in-flight operations (recovery UIs can list and selectively abandon them, instead of the all-or-nothing `clear_locked_outputs`);
- the expiry stored once per proposal instead of denormalized per note row;
- and, if the proposal blob is stored, restart-resumable operations: today a wallet whose async operations are memory-only (Zallet) cannot complete an in-flight proposal after a restart at all. A persisted proposal plus the locked-input decoding fix in this PR would make that flow work end to end. Our integration suite has a skipped scenario waiting on exactly this.

Because the table would keep `LockOwner` as its public identity (a UNIQUE column), the `WalletWrite` API in this PR survives the upgrade unchanged; the trait never needs to expose backend-internal row ids.

### Open questions (why this is a follow-up, not part of this PR)

- Row lifecycle: who deletes a `tx_proposals` row, the wallet (on store/expiry) or the caller (on abandon)? Is an expired-but-present row an error state or a resumable one?
- Should the serialized proposal live in the wallet DB at all, or remain the application's responsibility (with only the lock session persisted)?
- Migration story for the two lock columns this PR adds (they would move into the new table).

Recommendation: land this PR with the opaque token (everything ownership must guarantee is enforced and property-tested with it: no cross-flow release, idempotent crash-retry, no lock stealing), and treat `tx_proposals` as the follow-up where persisted operations and lock FKs arrive together.

zcash_client_backend/src/data_api/wallet/input_selection.rs
```diff
@@ -746,6 +746,10 @@ impl<DbT> GreedyInputSelector<DbT> {
             ),
         };
         *amount_at_transparent_gather = amount_at_gather;
+        // Input selection must never draw on locked outputs: a locked output belongs to
+        // another in-flight proposal, and spending it would recreate the conflict that
+        // locking exists to prevent.
+        let include_locked = false;
```
nuttycom, line 752: There's a problem with hardcoding `include_locked` to false in input selection - this needs to be something that the caller has control over, because there are legitimate cases where the lock should be overridden. Specifically, in the wallet migration flow, it must *always* remain possible for a user to choose to migrate their funds immediately, discarding privacy; this requires overriding the locks. What we want to be able to do is *preferentially* choose unlocked notes, but if required by the user, we should also be able to choose locked notes (those that have been reserved for use by a pool-migration PCZT).

schell:
Looking good! I'm still reviewing in depth but this is a a good improvement 👍

PR #2728, opened 2026-07-22
zcash_client_backend: retain anchor checkpoints across large scan batches

### LINKED ISSUE #2727: Anchor-retention checkpoints are pruned when a scan batch exceeds the checkpoint budget
## Summary

`update_tree` (in `zcash_client_backend/src/data_api/ll/wallet.rs`) registers anchor-retention checkpoints *after* `insert_tree`, but `insert_tree` prunes to `max_checkpoints` (the `PRUNING_DEPTH` checkpoint budget) during insertion. When a single `put_blocks` scan batch is larger than that budget — routine during initial sync and catch-up — an interval-aligned anchor checkpoint is pruned before its retention id is registered. Its root and the witnesses anchored to it then become permanently unrecoverable.

Anchor retention happens to work only when scans proceed in batches no larger than `PRUNING_DEPTH` (100) blocks, because retention is then registered before the *next* batch's prune.

Anchor retention is currently dormant on mainnet (its floor is the NU6.3 activation height, which is unassigned), so no released wallet is affected yet — but the write path is incorrect.

## Reproduction

With NU6.3 active, receive a note, then scan a single batch far enough that an interval-aligned anchor ages more than `PRUNING_DEPTH` checkpoints behind the tip. A witness for the note as of that anchor is then `None` (the checkpoint was pruned), even though the anchor's id is still present in `retained_checkpoints()`.

## Fix

Register anchor retention for a batch's checkpoint heights before `insert_tree`, so the pruning that runs during insertion exempts them. `ShardTree::ensure_retained` records a checkpoint id even when no checkpoint with that id yet exists.

---

nuttycom:
`update_tree` registered anchor retention *after* `insert_tree`, but `insert_tree` prunes to `max_checkpoints` (the `PRUNING_DEPTH` checkpoint budget) during insertion. A single `put_blocks` scan batch larger than that budget — routine during initial sync and catch-up — therefore pruned an interval-aligned anchor checkpoint before its retention id was registered, leaving its root and anchored witnesses permanently unrecoverable. Anchor retention only happened to work for batches no larger than the budget, because retention was then registered before the *next* batch's prune. (Retention is dormant on mainnet until NU6.3 has an assigned activation height, so no released wallet is affected yet.)

The fix registers anchor retention for a batch's checkpoint heights before `insert_tree`, so the pruning during insertion exempts them. `ShardTree::ensure_retained` records a checkpoint id even when no checkpoint with that id yet exists, so recording it ahead of insertion is safe.

Commits (TDD, red then green):
1. Add a backend-generic, wallet-level regression test — run for Sapling and Orchard against the SQLite backend — that drives the real batched `put_blocks`/`insert_tree` path. It **fails** at that commit, demonstrating the bug. The interval and pruning depth are read from the implementation, so the test tracks whatever values it defines.
2. Reorder the retention registration to fix the bug; the test passes.

Closes #2727

---

### DIFF
--- a/zcash_client_backend/src/data_api/ll/wallet.rs
+++ b/zcash_client_backend/src/data_api/ll/wallet.rs
@@ -1606,7 +1606,7 @@ fn should_track_nullifiers(
 
 /// The interval, in blocks, at which checkpoints are retained as durable "anchors" once anchor
 /// retention is active. At 75-second blocks this is roughly every 6 hours (4 per day).
-const ANCHOR_RETENTION_INTERVAL: u32 = 288;
+pub(crate) const ANCHOR_RETENTION_INTERVAL: u32 = 288;
 
 /// Returns whether the checkpoint at `height` should be retained as a durable anchor.
 ///
--- a/zcash_client_backend/src/data_api/ll/wallet.rs
+++ b/zcash_client_backend/src/data_api/ll/wallet.rs
@@ -1698,12 +1698,15 @@ where
     retain_anchor_checkpoint(tree, anchor_retention_height, frontier_height)?;
 
     for (subtree, checkpoints) in subtrees {
-        // Capture the checkpoint heights before the checkpoint map is consumed by `insert_tree`.
-        let checkpoint_heights = checkpoints.keys().copied().collect::<Vec<_>>();
-        tree.insert_tree(subtree, checkpoints)?;
-        for height in checkpoint_heights {
-            retain_anchor_checkpoint(tree, anchor_retention_height, height)?;
+        // Register anchor retention for this batch's checkpoint heights *before* `insert_tree`,
+        // which prunes down to `max_checkpoints` during insertion. A batch larger than the
+        // checkpoint budget would otherwise prune an anchor before it could be retained, so
+        // retention must be recorded first; `ShardTree::ensure_retained` accepts a checkpoint
+        // height whose checkpoint does not yet exist.
+        for height in checkpoints.keys() {
+            retain_anchor_checkpoint(tree, anchor_retention_height, *height)?;
         }
+        tree.insert_tree(subtree, checkpoints)?;
     }
 
     // Ensure we have a tree checkpoint for each checkpointed block height.
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -4358,6 +4358,194 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, Dsf: DataStoreFactory>(
     );
 }
 
+/// A wallet-level test for note-commitment-tree *anchor retention*: once NU6.3 (Ironwood) is
+/// active, checkpoints on the anchor-retention interval are retained as durable anchors, exempt
+/// from the ordinary `PRUNING_DEPTH`-checkpoint pruning budget, so that their roots and the
+/// witnesses anchored to them remain computable even after they age far behind the chain tip.
+///
+/// The retention interval and pruning depth are read from `crate::data_api::ll::wallet`, so this
+/// test tracks whatever values the implementation defines rather than assuming a particular one.
+///
+/// The test:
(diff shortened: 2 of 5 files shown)


zcash_client_backend/src/data_api/ll/wallet.rs
```diff
@@ -1698,12 +1698,15 @@ where
     retain_anchor_checkpoint(tree, anchor_retention_height, frontier_height)?;
 
     for (subtree, checkpoints) in subtrees {
-        // Capture the checkpoint heights before the checkpoint map is consumed by `insert_tree`.
-        let checkpoint_heights = checkpoints.keys().copied().collect::<Vec<_>>();
-        tree.insert_tree(subtree, checkpoints)?;
-        for height in checkpoint_heights {
-            retain_anchor_checkpoint(tree, anchor_retention_height, height)?;
+        // Register anchor retention for this batch's checkpoint heights *before* `insert_tree`,
```
dannywillems, line 1701: Follow-up: I would suggest to add a PBT in `shardtree`.

dannywillems: Working on it. An agent has been spawned.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4358,6 +4358,194 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, Dsf: DataStoreFactory>(
...
+///   above the buried anchor *was* pruned, exactly the interval-aligned anchors at/above the floor
+///   are retained, and the full `PRUNING_DEPTH` window of checkpoints at the chain tip survives.
+pub fn anchor_checkpoints_retained_across_deep_scan<
+    T: ShieldedPoolTester,
+    Dsf: DataStoreFactory,
+>(
+    ds_factory: Dsf,
+    cache: impl TestCache,
+) {
+    use std::collections::BTreeSet;
```
dannywillems, line 4388: Nit: we should stop using `use` statements in the body of methods. This should be added to AGENTS.md, and a PR should be opened to go through the whole repo to fix this.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4358,6 +4358,194 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, Dsf: DataStoreFactory>(
...
+    fn tree_anchor_state<S, const DEPTH: u8, const SHARD_HEIGHT: u8>(
+        tree: &mut ShardTree<S, DEPTH, SHARD_HEIGHT>,
+        note_position: Position,
+        anchor_height: BlockHeight,
+    ) -> Result<(bool, BTreeSet<BlockHeight>, BTreeSet<BlockHeight>), ShardTreeError<S::Error>>
+    where
+        S: ShardStore<CheckpointId = BlockHeight>,
+        S::H: incrementalmerkletree::Hashable + Clone + PartialEq,
+    {
+        let witness_computable = tree
```
dannywillems, line 4407: Follow-up: we should come up with some Rust idiomatic `Builder`, so we know where the setup begins/ends, where the calls begins/ends and where the checks/asserts begins/ends.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -4358,6 +4358,194 @@ pub fn checkpoint_gaps<T: ShieldedPoolTester, Dsf: DataStoreFactory>(
...
+                survivors.insert(*cid);
+                Ok(())
+            })
+            .map_err(ShardTreeError::Storage)?;
+        Ok((witness_computable, survivors, retained))
+    }
+
+    // A network on which NU6.3 (Ironwood) is active from the Sapling activation height, so anchor
+    // retention is live with its floor at the account birthday.
+    let activation = BlockHeight::from_u32(100_000);
```
dannywillems, line 4430: Nit: there should be a factory for this, behind the feature flag `testing`.

dannywillems (approved):
utACK 8e3b63e2536c9e41b7d345b339c8adfdc87e1eb3

PR #2729, opened 2026-07-22
zcash_client_backend: reduce the anchor retention interval to 144 blocks

nuttycom:
Halve the anchor retention interval from 288 to 144 blocks (~every 3 hours at 75-second blocks, 8 per day), and update the doc comment to match.

Stacked on #2728 (the anchor-retention pruning fix); this PR targets that branch, so its diff is just the interval change. It should be rebased onto `main` once #2728 merges.

---


dannywillems (dismissed):
utACK 7b6c94ccdc314f9238d41132e0abd03ece65a3cf. Discussed in a call.

dannywillems (approved):
re utACK 7b6c94ccdc314f9238d41132e0abd03ece65a3cf.

PR #2730, opened 2026-07-22
zcash_primitives: cache the Orchard proving key across transaction builds

### LINKED ISSUE #2722: zcash_primitives: Ensure that the proving key is not reconstructed multiple times by the transaction builder.
@ValarDragon reports: "LRZ rebuilds the orchard proving key every single tx in the builder" - we should lazily load the proving key and reuse it when building multiple transactions.

dannywillems:
## Summary

Fixes #2722. As reported by @ValarDragon, the transaction builder rebuilt
the Orchard proving key on every call to `Builder::build`, so constructing
many transactions in a process repeatedly paid the cost of reconstructing
this expensive object.

The builder now obtains the proving key from a process-wide cache keyed by
circuit version. The circuit version is still derived from the consensus
branch (unchanged); the key is built lazily on first use via `OnceLock` and
reused thereafter. Building many transactions therefore reuses a single key
per circuit version.

## Details

- Added a `cached_orchard_proving_key(circuit_version)` helper that memoizes
  one `ProvingKey` per `OrchardCircuitVersion` in `OnceLock` statics. Keying
  by circuit version is required because the proving/verifying keys differ
  between versions.
- `build_internal` now maps the derived circuit version through the cache
  (`Option<&ProvingKey>`) instead of calling `ProvingKey::build` inline.
- The cache relies on `std::sync::OnceLock`, so it is gated on the `std`
  feature. Without `std` there is no thread-safe cache, so the key is built
  once per transaction (previous behavior); proving requires `std` in
  practice, so that fallback is not exercised when actually creating proofs.
- Mirrors the existing caching pattern in `pczt/src/roles/prover/orchard.rs`.

## Testing

- Added `cached_orchard_proving_key_reuses_one_instance_per_version`, which
  asserts pointer identity across repeated calls for the same circuit version
  and distinct instances for different versions.
- `cargo test -p zcash_primitives --all-features transaction::builder` passes.
- `cargo clippy -p zcash_primitives --all-features --all-targets` is clean.
- Verified both the `std` and `--no-default-features --features circuits`
  (non-`std`) build paths compile.

### DIFF
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -1566,29 +1566,41 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
 
         // The Orchard and Ironwood circuit version is fixed by the transaction's
         // consensus branch (both pools share the post-NU6.3 circuit), so derive it
-        // once from the branch rather than from a bundle. Only build the key when
+        // once from the branch rather than from a bundle. Only obtain the key when
         // an Orchard or Ironwood bundle is actually present.
-        let orchard_proving_key = {
-            let build_proving_key = unauthed_tx.orchard_bundle.is_some();
-            let build_proving_key = build_proving_key || unauthed_tx.ironwood_bundle.is_some();
+        let orchard_circuit_version = {
+            let build_proving_key =
+                unauthed_tx.orchard_bundle.is_some() || unauthed_tx.ironwood_bundle.is_some();
             build_proving_key.then(|| {
-                orchard::circuit::ProvingKey::build(
-                    bundle_version_for_branch(
-                        unauthed_tx.consensus_branch_id,
-                        orchard::ValuePool::Orchard,
-                    )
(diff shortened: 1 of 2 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -1566,29 +1566,41 @@ impl<P: consensus::Parameters, U: sapling::builder::ProverProgress> Builder<P, U
...
+                .circuit_version()
             })
         };
 
+        // Reuse a process-wide cached proving key rather than reconstructing it on
+        // every `build` call. See `cached_orchard_proving_key`.
+        #[cfg(feature = "std")]
+        let orchard_proving_key: Option<&orchard::circuit::ProvingKey> =
+            orchard_circuit_version.map(cached_orchard_proving_key);
+        // Without `std` there is no thread-safe cache; build the key once for this
```
dannywillems, line 1589: This is the old behavior, but it is ok.

nuttycom (approved):
utACK bf4bf2856057385c46dd6287ce742da0414d1de8

ValarDragon:
Awesome thank you!

PR #2735, opened 2026-07-22
zcash_client_sqlite: treat empty note-commitment trees as vacuously consistent in rewind_to_chain_state

nesence-m:
Since #2598, rewind_to_chain_state's per-pool truncation window computation
requires the Sapling, Orchard, and Ironwood checkpoint tables to agree on the
minimum checkpoint at or above the truncation floor, failing with
CorruptedData if any pool's window floor diverges from Sapling's.

On an upgraded wallet, the ironwood_shardtree migration (like
orchard_shardtree before it) creates the Ironwood tables empty, with no
backfill of historical checkpoints. Parity across pools is only restored the
next time blocks are scanned, via each scanned block adding a matching
checkpoint to every pool's tree. Until then, any rewind that lands within the
gap deterministically fails the parity check on a database that is not
actually corrupted.

An empty tree cannot be "left ahead" of another pool's checkpoint window, and
ShardTree::truncate_to_checkpoint is a safe no-op on an empty tree, so
extending the parity check with a tolerance for empty trees cannot reintroduce
the inconsistency #2598 was written to prevent. This mirrors the tolerance
min_shared_checkpoint_height and select_truncation_height already apply via
`OR NOT EXISTS (...)` for pools with no checkpoints at all.

Observed as a wallet-breaking crash on Android during Keystone account import
("Sapling and Ironwood should have the same checkpoints"), where the failure
also left the importing wallet permanently stuck in a non-syncing state.

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -147,7 +147,7 @@ use {
 };
 
 #[cfg(feature = "orchard")]
-use zcash_client_backend::data_api::ORCHARD_SHARD_HEIGHT;
+use zcash_client_backend::data_api::{IRONWOOD_SHARD_HEIGHT, ORCHARD_SHARD_HEIGHT};
 
 #[cfg(feature = "zcashd-compat")]
 use {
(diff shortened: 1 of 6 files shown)


nuttycom (changes requested):
Blocking, there are still bugs.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -691,6 +691,27 @@ pub(crate) fn min_checkpoint_id_at_or_above(
...
+    table_prefix: &'static str,
+) -> Result<(), Error> {
+    conn.execute_batch(&format!(
+        "DELETE FROM {table_prefix}_tree_checkpoint_marks_removed;
+         DELETE FROM {table_prefix}_tree_checkpoints;
+         DELETE FROM {table_prefix}_tree_shards;
+         DELETE FROM {table_prefix}_tree_cap;"
+    ))
+    .map_err(Error::Query)
+}
```
nuttycom: This discards downloaded subtree roots; this will make it impossible to create proofs until the roots have been re-downloaded.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -5441,20 +5679,39 @@ mod tests {
...
+                 transaction_id INTEGER,
+                 commitment_tree_position INTEGER);
+             CREATE TABLE ironwood_received_notes (
+                 id INTEGER PRIMARY KEY,
+                 transaction_id INTEGER,
+                 commitment_tree_position INTEGER);",
         )
         .unwrap();
         conn
     }
```
nuttycom, line 5795: This fixture is fragile; it can diverge from the behavior of the actual committed schema. The test schema should be established by running the real migrations.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -3853,14 +3894,132 @@ pub(crate) fn truncate_to_height<P: consensus::Parameters>(
...
+                ),
+                named_params![":height": u32::from(rescan_floor)],
+                |row| row.get::<_, bool>(0),
+            )?;
+            if loses_witnesses {
+                Err(SqliteClientError::CorruptedData(format!(
+                    "truncation to height {truncation_height} would empty the {pool:?} note \
+                     commitment tree, destroying witness data for notes received at or below \
+                     height {rescan_floor} that no rescan would re-create"
+                )))
```
nuttycom: The `CorruptedData` result here is not the correct way to express the result; this is an expected potential outcome of valid scan history. Instead, `TreeTruncation` should be expanded to include cases to indicate that the truncation/rewind cannot be executed.

nuttycom (approved):
utACK 82bbc3af5055622db0b78fe60b7eb99e70f76d4c

PR #2736, opened 2026-07-22
Pool migration: post-#2710 review follow-ups

dannywillems:
Collects the follow-up changes agreed after PR #2710 was approved and merged,
addressing the remaining review comments from @nuttycom one commit at a time.

## Changes

- **`zcash_pool_migration_backend`: select notes at the chain tip when
  proving.** `prove_orchard` now takes the chain tip and derives the
  `TargetHeight` as `tip + 1` for `select_unspent_notes`, matching how the
  wallet selects spendable notes elsewhere (r3632783151, r3632780651).

- **`zcash_client_sqlite`: find the highest rooted checkpoint without
  probing.** The `highest_rooted_orchard_checkpoint` test helper now reads the
  checkpoint set directly via `ShardStore::for_each_checkpoint` (bounded by
  `checkpoint_count()`), instead of probing the tree at every height down from
  `from`.

Built and validated with the `prove_chain_sim` integration test (passes in
~264s).

### DIFF
--- a/zcash_client_sqlite/src/testing.rs
+++ b/zcash_client_sqlite/src/testing.rs
@@ -130,24 +130,26 @@ where
     W: zcash_client_backend::data_api::WalletCommitmentTrees,
 {
     use shardtree::error::ShardTreeError;
+    use shardtree::store::ShardStore;
     use zcash_client_backend::data_api::WalletCommitmentTrees;
 
-    let mut height = u32::from(from);
-    loop {
-        let bh = BlockHeight::from_u32(height);
-        let rooted = db
-            .with_orchard_tree_mut::<_, _, ShardTreeError<<W as WalletCommitmentTrees>::Error>>(
-                |tree| Ok(tree.root_at_checkpoint_id(&bh)?.is_some()),
-            )
-            .expect("queries the Orchard tree");
-        if rooted {
-            return Some(bh);
-        }
-        if height == 0 {
-            return None;
-        }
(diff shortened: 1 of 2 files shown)


zcash_pool_migration_backend/src/engine.rs
nuttycom, line 1: This first commit overspecializes to my "at the chain tip" comment. This should use the standard anchor height chosen by `get_target_and_anchor_heights`.

nuttycom (changes requested):

zcash_client_backend/CHANGELOG.md
```diff
@@ -909,7 +909,7 @@ workspace.
   - `TransactionDataRequest::transactions_involving_address`
   - `wallet::ConfirmationsPolicy`
   - `wallet::SpendingKeys`
-  - `wallet::TargetHeight`
+  - `wallet::TargetHeight` (re-exported from `zcash_protocol::consensus`)
```
nuttycom: This incorrectly alters the CHANGELOG of a long-since-published release. Revert.

components/zcash_protocol/src/consensus.rs
nuttycom, line 1: I'm confused because the moved type does not actually appear to now be used in the migration crate; if it was already usable where needed under the `wallet` feature flag, then there's no need to move it.

nuttycom (approved):

PR #2711, opened 2026-07-21
zcash_pool_migration: extract a Committer for commit_preparation_inner

dannywillems:
Hosts the shared build state that `commit_preparation_inner` threaded through its stages (the minted-note pool, the transaction-id counter, and the `transactions`/`unsigned`/`layer_ids` accumulators, plus the owned backend, rng, resolved FVK, and NU6.3 activation) in a `Committer` struct, so each stage becomes a method and the function reads as an orchestration:

    start
      -> build_preparation_layers
      -> add_direct_funding
      -> build_transfers
      -> into_state

- `start` performs the "migration already in progress" guard and resolves the Orchard FVK and NU6.3 activation height.
- `into_state` consumes the committer, which releases the `&mut backend` reborrow so the caller can `put_migration` afterward.
- `plan` is kept a method parameter, not a struct field, so iterating `plan.preparation().layers()` while calling `next_id()`/`resolve_prep_spends()` never borrows `self` both immutably and mutably at once.

Pure refactor: no behavior change, and all 99 `zcash_pool_migration_backend` tests pass. `commit_preparation` / `build_preparation_unsigned` public signatures are unchanged.

Stacked on #2709; addresses r3623833443's "the whole method could be simplified".

PR #2738, opened 2026-07-23
zcash_pool_migration_backend: use TargetHeight for above-tip heights

dannywillems:
Threads the `TargetHeight` type through the migration state machine and commit
path, so above-tip heights (`chain_tip + 1`) are type-distinguished from
ordinary block heights. Addresses the review nits on PRs #2710 and #2711 noting
that a dedicated `TargetHeight` type exists for heights above the chain tip.

## Commits

1. **`zcash_protocol`: move `TargetHeight` into the consensus module.** The type
   wraps a `BlockHeight` and has no dependency on `zcash_client_backend`, so it
   belongs in the lower `zcash_protocol` crate next to `BlockHeight`,
   re-exported from `zcash_client_backend::data_api::wallet` so existing call
   sites are unaffected.

   This move is a prerequisite for commit 2: `zcash_pool_migration_backend` is
   `#![no_std]` and its `state.rs` compiles unconditionally, but
   `zcash_client_backend` (the type's previous home) is an *optional* dependency
   enabled only by the `wallet` feature. `state.rs` could not adopt
   `TargetHeight` from there without breaking the default/no-feature build;
   sourcing it from the always-present `zcash_protocol` fixes that.

2. **`zcash_pool_migration_backend`: use `TargetHeight` for above-tip heights.**
   The `state.rs` decision methods (`next_step`, `next_provable`,
   `next_broadcastable`, `transaction_statuses`, `expired_transactions`, and the
   private `is_expired`/`prove_ready`/`next_rebuildable`) and the `engine.rs`
   commit functions + `Committer` now take `TargetHeight` instead of a bare
   `BlockHeight`. The engine.rs import is gated on `orchard` (all its uses are),
   keeping the no-orchard build warning-clean; `state.rs` uses it unconditionally.

## Testing

- 116 lib + 3 `engine_commit` tests pass; `prove_chain_sim` E2E passes (~276s).
- Clean builds for `orchard`, `orchard,wallet`, `--no-default-features`, and
  downstream `zcash_client_sqlite --features orchard`.

### DIFF
--- a/zcash_pool_migration_backend/src/engine.rs
+++ b/zcash_pool_migration_backend/src/engine.rs
@@ -51,6 +51,8 @@ use getset::{CopyGetters, Getters};
 use rand_core::RngCore;
 use zcash_protocol::TxId;
 use zcash_protocol::consensus::BlockHeight;
+#[cfg(all(test, feature = "orchard"))]
+use zcash_protocol::consensus::TargetHeight;
 use zcash_protocol::value::{BalanceError, Zatoshis};
 
 use zcash_primitives::transaction::fees::FeeRule as _;
@@ -3058,7 +3060,7 @@ mod commit_tests {
 
         // Advance the chain tip past the transfer's expiry, so it can no longer be mined.
         backend.tip = old.expiry_height + 1;
-        let target = backend.chain_tip_height().unwrap() + 1;
+        let target = TargetHeight::from(backend.chain_tip_height().unwrap() + 1);
         assert!(
             state.expired_transactions(target).contains(&id),
             "the transfer has expired at the new tip"
@@ -3495,6 +3497,7 @@ mod commit_tests {
             .iter()
             .map(|t| t.scheduled_height)
             .max()
+            .map(TargetHeight::from)
             .expect("the committed migration has transactions");
         match state.next_step(target) {
             crate::state::AdvanceStep::Prove { id }
@@ -3675,6 +3678,7 @@ mod commit_tests {
             .iter()
             .map(|t| t.scheduled_height)
             .max()
+            .map(TargetHeight::from)
             .expect("the committed migration has transactions");
         let mut height = 2_000_000u32;
         for layer in 0..layer_count {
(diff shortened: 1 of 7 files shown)


zcash_pool_migration_backend/src/engine.rs
```diff
@@ -2258,7 +2262,9 @@ where
             .prep_schedule()
             .last()
             .and_then(|layer| layer.last())
-            .map_or(self.target_height, |&h| h + EST_PREP_LAYER_MINING_BLOCKS);
+            .map_or(BlockHeight::from(self.target_height), |&h| {
```
dannywillems: Should be called differently than target_height, if TargetHeight is not used here?

zcash_pool_migration_backend/src/state.rs
```diff
@@ -266,12 +266,12 @@ impl MigrationState {
...
     /// The id of the next transaction ready to BROADCAST: already `Proved`, its dependencies mined,
     /// and scheduled at or before `target_height` (`chain_tip + 1`).
-    pub fn next_broadcastable(&self, target_height: BlockHeight) -> Option<MigrationTxId> {
+    pub fn next_broadcastable(&self, target_height: TargetHeight) -> Option<MigrationTxId> {
         self.transactions
             .iter()
             .find(|t| {
                 matches!(t.state, MigrationTxState::Proved)
-                    && t.scheduled_height <= target_height
+                    && t.scheduled_height <= BlockHeight::from(target_height)
```
dannywillems, line 274: Should it be `TargetHeight`?

zcash_pool_migration_backend/src/state.rs
```diff
@@ -453,7 +453,7 @@ impl MigrationState {
                         MigrationTxState::Proved => {
                             if !deps_ok {
                                 (false, None, Some(Blocker::Dependencies))
-                            } else if t.scheduled_height <= target_height {
+                            } else if t.scheduled_height <= BlockHeight::from(target_height) {
```
dannywillems, line 456: Should it be TargetHeight?

PR #2741, opened 2026-07-23
zcash_history: fix height-range underflow and remove zcashd-era dead code

### LINKED ISSUE #2676: zcash_history: integer underflow bug in leaf_count() — NodeData::read() doesn't validate height invariants; zcashd-era code causing maintenance issues
While reviewing `zcash_history` I found an integer underflow bug in `leaf_count()` and noticed several pieces of stale code left over from when this crate was used via FFI from `zcashd`. With `zcashd` having reached its End-of-Support halt on ~July 18th 2026 and Zebra now being the sole consumer of this crate, some of this code appears to be dead weight. Filing as one issue since the crate is small (3 source files) and Zebra is the only downstream consumer I'm aware of.

## The bug

`NodeData::read()` (`node_data.rs:152–153`) deserialises `start_height` and `end_height` without checking that `start_height <= end_height` or that `start_height >= 1`:

```rust
data.start_height = Self::read_compact(r)?;
data.end_height = Self::read_compact(r)?;
// no validation
```

`Entry::leaf_count()` (`entry.rs:44`) then does:

```rust
V::end_height(&self.data) - (V::start_height(&self.data) - 1)
```

This is unsigned `u64` subtraction. If `start_height > end_height + 1` it underflows. If `start_height == 0` the subexpression `(start_height - 1)` underflows on its own. In debug this panics with "attempt to subtract with overflow". In release (Zebra sets `overflow-checks = false` in its release profile) it wraps silently — `leaf_count()` returns a garbage value (~1.8 × 10^19), `complete()` returns the wrong answer, and `append_leaf()` fails with `ExpectedInMemory`.

I confirmed this by constructing a `NodeData` with `start_height = 200, end_height = 5`, serialising it, deserialising it via `Entry::from_bytes()` (which succeeds — no validation), and calling `leaf_count()`. Debug: panics. Release: returns `18446744073709551422`.

Under normal operation on mainnet/testnet this can't happen — Zebra sets `start_height = end_height = coinbase_height` from validated block data, and Heartwood activates at 903,000+ so `start_height >= 1` always holds. But `NodeData::read()` and `Entry::from_bytes()` are public functions, and any corrupted or untrusted bytes passed to them would hit the underflow instead of getting a clean `Err`. On regtest with configurable activation heights, the `start_height == 0` case doesn't even require corruption — if Heartwood is set to activate at height 0, the genesis leaf has `start_height == 0` and the underflow fires when the second block is processed.

It seems like adding validation in `NodeData::read()` after reading both height fields would be the simplest fix, but I'll defer to the team on the exact approach.

## Stale zcashd-era code

The crate was written in 2019 for the original `zcashd` FFI layer. `zcashd` reached its End-of-Support halt on ~July 18th 2026 (block height 3,417,100) and is no longer maintained — Zebra is now the sole consumer of `zcash_history`. Several things in the crate appear to be leftovers from that era:

**1. Duplicate CompactSize implementation.** `node_data.rs:84–114` reimplements Bitcoin's CompactSize variable-length integer encoding by hand (`read_compact`/`write_compact`). The workspace already ships `zcash_encoding::CompactSize` (`components/zcash_encoding/src/lib.rs:29–111`), which implements the same wire format and is used by `zcash_primitives`, `zcash_client_backend`, and `zcash_client_sqlite`. Both implementations produce identical bytes for canonical values — the only difference is that `zcash_encoding::CompactSize::read` rejects non-canonical encodings that `zcash_history::read_compact` currently accepts silently. Could these be consolidated?

**2. `byteorder` dependency.** The crate depends on `byteorder` solely for `ReadBytesExt`/`WriteBytesExt` convenience methods (`read_u16::<LittleEndian>()`, `write_u32::<LittleEndian>()`, etc.). There are 35 call sites across 3 source files. The rest of `librustzcash` appears to have migrated off `byteorder` in favour of direct `from_le_bytes`/`to_le_bytes` calls. Is this dependency still needed, or could it be removed?

**3. Dead `#[repr(C)]` attributes and FFI doc comment.** `EntryLink` and `EntryKind` carry `#[repr(C)]` (`lib.rs:42`, `lib.rs:61`), and the crate-level doc comment reads "To be used in zebra and via FFI bindings in zcashd" (`lib.rs:3`). There are zero `extern "C"` functions or `#[no_mangle]` exports in the crate — the FFI was in the `librustzcash` C bindings layer. Now that `zcashd` is end-of-life and Zebra accesses these types through the Rust API only, are the `#[repr(C)]` attributes and the doc comment still needed?

**4. `std::io` instead of `corez::io`.** The crate uses `std::io::{Read, Write, Cursor}` throughout, while the rest of the workspace has migrated to `corez::io` for `no_std` compatibility (PR #10522). `zcash_history` appears to be one of the last holdouts. Would migrating this to `corez::io` be straightforward?

**5. `primitive-types::U256` dependency.** Used only for `subtree_total_work` — the crate calls `from_little_endian`, `to_little_endian`, and `Add` on it. This is a narrow surface that might be served by a `[u8; 32]` wrapper, though I understand this would touch the public `NodeData` struct and might need to be a separate semver-breaking change.

## Summary of potential dependency changes

| Dependency | Current | If all changes above land |
|------------|---------|---------------------------|
| `primitive-types` | `0.12` | unchanged (deferred, semver-breaking) |
| `byteorder` | `1` | removed |
| `blake2b_simd` | workspace | unchanged |
| `zcash_encoding` | not a dependency | added |
| `corez` | not a dependency | added |

Happy to open a PR for any or all of these if the maintainers are interested. I'm not sure which of these changes the team would consider in scope vs. out of scope, so guidance would be appreciated.

> _Drafted with AI assistance (Claude). Human author: @craftsoldier._

craftsoldier:
Closes #2676.

This PR fixes the integer underflow bug in `zcash_history::Entry::leaf_count()` and removes several pieces of dead code left over from the `zcashd` FFI era.

## Bug fix

- `NodeData::read()` now validates that `end_height - start_height + 1` fits in a `u64` before accepting a deserialized node. Descending ranges and overflowing ranges are rejected with `io::ErrorKind::InvalidData`.
- `Entry::leaf_count()` now uses `checked_sub`/`checked_add(1)` instead of raw `u64` subtraction. This fixes the underflow that occurs when `start_height == 0` (regtest with Heartwood at height 0) or when the range is otherwise invalid.

## Cleanup

- Remove dead `#[repr(C)]` attributes from `EntryLink` and `EntryKind`, and update the crate doc comment to reflect that `zcashd` is end-of-life.
- Migrate from `std::io` to `corez::io` (with the `std` feature enabled, this is a no-op type swap that aligns the crate with the rest of the workspace).
- Replace the hand-rolled CompactSize implementation with `zcash_encoding::CompactSize`. The new implementation additionally rejects non-canonical encodings.
- Remove the `byteorder` dependency and convert the 16 remaining call sites to `from_le_bytes`/`to_le_bytes` + `read_exact`/`write_all`.

## Deliberately deferred

Item 5 from #2676 — replacing `primitive_types::U256` with `[u8; 32]` — is **not** included because it is a semver-breaking change to the public `NodeData` struct. Zebra already depends on `primitive-types` independently (via `uint::construct_uint!` for `ExpandedDifficulty`), so the dependency reduction is only at the `zcash_history` crate level. That change should be done separately as part of a future `0.6.0` release if the team wants it.

## Testing

- `cargo test -p zcash_history --all-features`: 23/23 pass, including new regression tests for the `start_height == 0` case.
- `cargo clippy -p zcash_history --all-features --all-targets -- -D warnings`: clean.
- `cargo check --workspace --all-features`: clean.
- Zebra's `zebra-chain` history-tree tests pass against the patched crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7262,6 +7262,7 @@ dependencies = [
  "corez",
  "primitive-types",
  "proptest",
+ "zcash_encoding",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7259,6 +7259,7 @@ dependencies = [
  "assert_matches",
  "blake2b_simd",
  "byteorder",
+ "corez",
  "primitive-types",
  "proptest",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7258,7 +7258,6 @@ version = "0.5.0"
 dependencies = [
  "assert_matches",
  "blake2b_simd",
- "byteorder",
  "corez",
  "primitive-types",
  "proptest",
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ graph TB
             equihash
             f4jumble
             zcash_encoding
+            zcash_history
         end
     end
 
@@ -99,6 +100,7 @@ graph TB
     %% zcash_keys --> zcash_encoding
     %% zcash_primitives --> zcash_encoding
     %% zcash_address --> zcash_encoding
+    zcash_history --> zcash_encoding
     zcash_protocol --> zcash_encoding
 
     zcash_primitives --> equihash
@@ -158,6 +160,7 @@ graph TB
     click zcash_transparent "https://docs.rs/zcash_transparent/" _blank
     click zip321 "https://docs.rs/zip321/" _blank
     click zip32 "https://docs.rs/zip32/" _blank
+    click zcash_history "https://docs.rs/zcash_history/" _blank
 ```
 
 <!-- END mermaid-dependency-graph -->
--- a/components/zcash_encoding/CHANGELOG.md
+++ b/components/zcash_encoding/CHANGELOG.md
@@ -8,12 +8,20 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Added
+- `zcash_encoding::CompactSize::{read_unbounded, write_unbounded}`, which use the
+  same wire format as `CompactSize::{read, write}` but span the full `u64` range
+  instead of enforcing the `MAX_COMPACT_SIZE` consensus limit.
 - `zcash_encoding::testing::check_roundtrip` (behind the new `test-dependencies`
   feature), a helper that asserts a `write`/`read` codec pair are exact inverses
   and that the encoding is stable across a round-trip.
 
 ### Changed
 - MSRV is now 1.88
+- `zcash_encoding::CompactSize::write` now returns an error when the provided
+  value exceeds `MAX_COMPACT_SIZE`, rather than silently writing a value that
+  `CompactSize::read` would reject. This mirrors the bound already enforced by
+  `CompactSize::read`; callers that may legitimately encode larger values should
+  use the new `CompactSize::write_unbounded`.
 
 ## [0.4.0] - 2026-04-23
 
--- a/components/zcash_encoding/src/lib.rs
+++ b/components/zcash_encoding/src/lib.rs
@@ -32,13 +32,34 @@ pub const MAX_COMPACT_SIZE: u32 = 0x02000000;
 pub struct CompactSize;
 
 impl CompactSize {
-    /// Reads an integer encoded in compact form.
-    pub fn read<R: Read>(mut reader: R) -> io::Result<u64> {
+    /// Reads an integer encoded in compact form, enforcing the
+    /// [`MAX_COMPACT_SIZE`] consensus limit.
+    ///
+    /// Use [`CompactSize::read_unbounded`] to decode values that are permitted
+    /// to exceed this limit.
+    pub fn read<R: Read>(reader: R) -> io::Result<u64> {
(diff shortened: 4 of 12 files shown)


craftsoldier:
Thanks @nuttycom! The unbounded CompactSize addition to `zcash_encoding` and the `no_std` support are great catches — the consensus-bounded `CompactSize` would indeed reject spec-valid ZIP 221 nodes with cumulative tx counts exceeding `MAX_COMPACT_SIZE`. Good call on also fixing the write/read asymmetry in `CompactSize::write`.

Happy to leave the rest in your hands — let me know if there's anything else you'd like me to address on this branch.

nuttycom (dismissed):
utACK 41c99955cbf3c6fe90fa5d38a2074efa1b907205

nuttycom:
Marking as draft to ensure this doesn't get merged before current crate releases as the semver break in zcash_encoding will complicate the release process.

PR #2742, opened 2026-07-23
Centralize note locking into dedicated modules and extract OutputLockStore

dannywillems:
Refactors the note-locking feature (introduced in #2716) out of the ~20 files
it had grown across and into one module per crate, then extracts the storage
contract as a trait. Move-only commits (relocation + re-exports, zero semantic
change) are kept strictly separate from semantic commits, and every commit
builds and passes the relevant test suites.

- `zcash_client_backend::data_api::locking`: the locking vocabulary
  (`LockOwner`, `LockError`, `LockRequest`, `unlock_proposal_inputs`,
  `LockedInputPolicy`, `LockFilter`) with the semantics stated once in the
  module rustdoc plus an integration map of every touchpoint; all old paths
  preserved via re-exports.
- `zcash_client_sqlite::wallet::locking`: the lock-mutation statements, the
  SQL eligibility/acquisition fragments, and (new, small semantic change) an
  `is_locked_at` predicate replacing the four hand-inlined balance
  classifications.
- Test consolidation: the shared locking scenarios live in
  `data_api::testing::pool::locking` (re-exported), and all SQLite lock tests
  (predicate proptests, the cross-connection concurrency test, and the
  per-pool instantiations) live in `wallet::locking`'s tests. Test counts
  verified unchanged by the moves.
- `OutputLockStore` extraction: `WalletWrite` no longer declares
  `lock_outputs` / `unlock_output` / `clear_locked_outputs`; they move (docs
  and all) to the new trait, which `WalletWrite` requires with its
  `AccountId`/`Error` tied to `WalletRead`'s, and
  `get_locked_outputs` moves off `WalletTest`. Breaking relative to
  unreleased main; see the CHANGELOG entry.

Gates run per commit: `cargo check --release --workspace --all-features
--all-targets` clean; final state additionally clippy-clean for
`zcash_client_backend` + `zcash_client_sqlite`, backend tests green apart from
the pre-existing `--release` failure of
`proposal::tests::orchard_pool_payment_with_ironwood_active_is_a_programming_error`,
and the full `zcash_client_sqlite` suite green.


zcash_client_sqlite/src/wallet/transparent.rs
```diff
@@ -1933,10 +1926,7 @@ pub(crate) fn add_transparent_account_balances(
             let add_pending = |bal: &mut Balance| {
                 if value <= zip317::MARGINAL_FEE {
                     bal.add_uneconomic_value(value)
-                } else if lock_expiry_height
```
dannywillems, line 1937: Oh wow, that was using an `iter` on an option before.

zcash_client_sqlite/src/wallet/locking.rs
```diff
@@ -233,6 +233,17 @@ pub(crate) fn unlock_spent_notes(
...
 }
 
+/// Returns whether an output bearing the given `lock_expiry_height` is locked as of
+/// `target_height`.
+///
+/// An output is locked while `lock_expiry_height >= target_height`; selection eligibility is
+/// the exact complement of this predicate (see [`output_eligible_condition`]). Balance
+/// computations use this single definition to tally locked value separately from spendable
+/// value.
+pub(crate) fn is_locked_at(lock_expiry_height: Option<u32>, target_height: TargetHeight) -> bool {
```
dannywillems, line 244: Better now.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1602,6 +1595,37 @@ impl<C: Borrow<rusqlite::Connection>, P: consensus::Parameters, CL, R> WalletTes
     }
 }
 
+impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R: RngCore>
```
dannywillems: We should move the type refinement in a where statement.

zcash_client_sqlite/src/lib.rs
```diff
@@ -1854,13 +1878,53 @@ impl<C: BorrowMut<rusqlite::Connection>, P: consensus::Parameters, CL: Clock, R:
...
         as_of_height: BlockHeight,
-    ) -> Result<(), Self::Error> {
+    ) -> Result<(), <Self as WalletRead>::Error> {
         self.transactionally(|wdb| wdb.notify_output_verified_unspent(outpoint, as_of_height))
     }
 }
 
 /// This impl block is only usable when you already have an [`SqlTransaction`], meaning
 /// you are inside a [`WalletDb::transactionally`] block with a lock on the database.
+impl<P: consensus::Parameters, CL: Clock, R: RngCore> OutputLockStore
```
dannywillems: We should move the type refinement in a where statement.

dannywillems (commented):
Self-ACK

dannywillems:
A refacto of the tests is coming, to decrease the number of lines. I am making a few follow-ups. The changes here are, on purpose, limited to moving/extracting some small pieces of repeated code.

zcash_client_sqlite/src/wallet/transparent.rs
nuttycom: You can tell a Haskeller because they use `any` | `all` | `traverse` on every kind of data structure. :laughing:

nuttycom: See also https://wiki.haskell.org/Fairbairn_threshold - knowing powerful combinators is better than knowing special cases or having built-in language support for constructs! If only everyone knew about `Applicative` and Kleisli composition, software woud be _better_.

nuttycom (changes requested):
The no-default-features configuration currently fails Clippy because `TxId` is imported unconditionally in `zcash_client_sqlite/src/wallet/locking.rs`, while its only use is gated behind `test` / `test-dependencies`.

Reproduction:

```console
$ cargo clippy -p zcash_client_sqlite --no-default-features --lib -- -D warnings
error: unused import: `zcash_primitives::transaction::TxId`
  --> zcash_client_sqlite/src/wallet/locking.rs:24:5
```

Please gate the import consistently with its use, or use the fully-qualified path in the gated function.

dannywillems:
> 45e8919



> The no-default-features configuration currently fails Clippy because `TxId` is imported unconditionally in `zcash_client_sqlite/src/wallet/locking.rs`, while its only use is gated behind `test` / `test-dependencies`.
> 
> Reproduction:
> 
> ```
> $ cargo clippy -p zcash_client_sqlite --no-default-features --lib -- -D warnings
> error: unused import: `zcash_primitives::transaction::TxId`
>   --> zcash_client_sqlite/src/wallet/locking.rs:24:5
> ```
> 
> Please gate the import consistently with its use, or use the fully-qualified path in the gated function.

Agent spawned and commit https://github.com/zcash/librustzcash/pull/2742/changes/45e89193d2b7df0c9d7b9df52891226091e4cb30 pushed

dannywillems:
> The no-default-features configuration currently fails Clippy because `TxId` is imported unconditionally in `zcash_client_sqlite/src/wallet/locking.rs`, while its only use is gated behind `test` / `test-dependencies`.
> 
> Reproduction:
> 
> ```
> $ cargo clippy -p zcash_client_sqlite --no-default-features --lib -- -D warnings
> error: unused import: `zcash_primitives::transaction::TxId`
>   --> zcash_client_sqlite/src/wallet/locking.rs:24:5
> ```
> 
> Please gate the import consistently with its use, or use the fully-qualified path in the gated function.

As a follow-up, adding the job in the CI

dannywillems:
Rebased to have the `--no-default-features` job.

nuttycom (dismissed):
The no-default-features failure is fixed. I verified the updated head with:

```console
cargo clippy -p zcash_client_sqlite --no-default-features --lib -- -D warnings
```

This now passes cleanly.

nuttycom (approved):
utACK, this is a really nice cleanup, well done

PR #2745, opened 2026-07-24
Replace per-pool BundleType fields with a coinbase-free BundlePadding

nuttycom:
## Summary

`BuildConfig::Standard` (and `DeferredPcztBuilder::new`) selected each Orchard-family pool's transactional bundle with an `orchard::builder::BundleType`, which also admits `BundleType::Coinbase`. With a separate field per pool, the type could express states that cannot occur:

- a coinbase bundle inside a `Standard` (non-coinbase) transaction, and
- one pool coinbase while the other is not.

Whether a transaction is coinbase is a property of the *whole* transaction — already chosen by the `BuildConfig` variant — so it must not be selectable per pool. The assertion that had previously rejected a coinbase override was removed when the per-pool fields were introduced, leaving only a misleading `.expect()` that would panic with a message claiming the flags were the problem.

## Changes

- Add `zcash_primitives::transaction::builder::BundlePadding`, carrying only the transactional padding axis (`bundle_required` / `pad_to_minimum`), with `DEFAULT` / `UNPADDED` constants mirroring the corresponding `BundleType` values and a `bundle_type()` bridge that always yields a `Transactional` bundle type.
- `BuildConfig::Standard` now stores `orchard_padding` / `ironwood_padding: BundlePadding`, and `DeferredPcztBuilder::new` takes two `BundlePadding`s. The per-pool coinbase state is now unrepresentable, and the two builder `.expect()`s are genuinely unreachable (relabeled to say so).
- Propagate the type through `zcash_client_backend::data_api::wallet::create_pczt_from_proposal` (`orchard_pool_padding`), closing the same hole at the public boundary.
- The `fees` module keeps its internal `BundleType` locals: they are needed for `num_actions` fee counting and never construct a coinbase bundle.

Per-pool *padding* remains supported (pool migration builds a padded Orchard source bundle alongside an unpadded Ironwood destination); only the nonsensical per-pool *coinbase* selection is removed.

## Testing

- `cargo check --workspace --tests`
- `cargo test -p zcash_primitives --features circuits,transparent-inputs` (the per-pool padding builder tests)
- `cargo test -p pczt`

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) assistance.

### DIFF
--- a/zcash_primitives/src/transaction/builder.rs
+++ b/zcash_primitives/src/transaction/builder.rs
@@ -262,6 +262,52 @@ impl Progress {
     }
 }
 
+/// The padding policy for a transactional Orchard-family (Orchard or Ironwood) bundle.
+///
+/// This selects only the parameters that vary between transactional bundles: whether a
+/// bundle is produced even when empty, and the minimum action count it is padded to. It
+/// deliberately cannot express an [`orchard::builder::BundleType::Coinbase`] bundle:
+/// coinbase construction is a property of the whole transaction, selected by
+/// [`BuildConfig::Coinbase`], never of an individual pool. Storing padding here rather
+/// than a full `BundleType` keeps the per-pool coinbase state unrepresentable.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct BundlePadding {
+    /// Produce a bundle even when no spends or outputs have been added; the resulting
+    /// bundle then consists entirely of dummy actions.
+    pub bundle_required: bool,
+    /// The minimum number of actions to pad the bundle to. `None` uses the orchard
+    /// builder's default 2-action minimum; `Some(1)` leaves the bundle unpadded, so its
+    /// action count reveals the transaction shape (acceptable when that shape is already
+    /// public, e.g. pool migrations).
+    pub pad_to_minimum: Option<u8>,
+}
+
+impl BundlePadding {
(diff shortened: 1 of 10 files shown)


zcash_primitives/src/transaction/builder.rs
```diff
@@ -262,15 +262,61 @@ impl Progress {
...
+        bundle_required: false,
+        pad_to_minimum: Some(1),
+    };
+
+    /// Returns the orchard [`BundleType`](orchard::builder::BundleType) selecting this
+    /// padding. The result is always a `Transactional` bundle type.
+    pub fn bundle_type(self) -> orchard::builder::BundleType {
+        orchard::builder::BundleType::Transactional {
+            bundle_required: self.bundle_required,
+            pad_to_minimum: self.pad_to_minimum,
```
schell, line 306: nit?: Does it make sense to make `BundleType::Transactional` a tuple variant that holds `BundlePadding`, instead of duplicating the records? I'm not sure.

schell (dismissed):
LGTM. Just a non-blocking question.

zcash_primitives/src/transaction/builder.rs
nuttycom: `BundleType` is in the `orchard` crate; it's not here. I don't think it's worth it to move `BundlePadding` there.

schell (approved):

PR #2752, opened 2026-07-24
zcash_client_sqlite: record internal-scope addresses for wallet-internal shielded outputs

nesence-m:
## Problem

Wallet-internal shielded outputs (change, and ZIP 318 Orchard → Ironwood pool-migration outputs) never get an `addresses` row: `ensure_address` deliberately returns `None` for `Scope::Internal` in both `wallet/orchard.rs` and `wallet/sapling.rs`. As a result `v_tx_outputs.to_address` is NULL for every self-transfer output, and a wallet frontend that renders a transaction's recipient from that view shows a blank recipient for pool-migration transactions.

Observed live on a testnet wallet in the Zashi Android app: the ZIP 318 migration transaction's detail screen had an empty "Sent to" field, because all of its outputs (the Orchard note-split outputs and the Ironwood crossing output) are internal-scope and had no linked address row.

## Fix

**Commit 1 — link internal-scope addresses in `ensure_address`:**

- `orchard.rs`: derive the internal-scope IVK from the account's FVK (`fvk.to_ivk(Scope::Internal)`) — the account's `uivk()` exposes the external IVK, whose `diversifier_index()` cannot decrypt an internal receiver. Build a single-receiver UA from the note's own recipient and upsert it under `KeyScope::INTERNAL`. This path also covers Ironwood notes, since `put_received_note` is shared by both pools.
- `sapling.rs`: same idea, with one wrinkle: `DiversifiableFullViewingKey::decrypt_diversifier`'s internal-scope branch verifies its candidate against the *external* address (`self.address(j)` rather than `self.derive_internal().address(j)`), so it never matches an internal receiver (sapling-crypto 0.7.0; appears worth an upstream report there). The fix derives the internal DFVK directly via `sapling_derive_internal_fvk` and decrypts the diversifier index against it.
- `upsert_address` gains an explicit `key_scope: KeyScope` parameter (it previously hardcoded `EXTERNAL`); existing call sites pass `KeyScope::EXTERNAL`, preserving behavior.

**Commit 2 — receiver-union merge in `upsert_address`:**

The first commit exposed a latent collision: the row keyed by `(account_id, key_scope, diversifier_index_be)` can be written by more than one subsystem. Transparent gap-limit provisioning stores the *bare transparent* internal change address at each child index (per `generate_address_list`'s documented behavior for internal scope), and an internal shielded note's diversifier index — index 0 for standard change — collides with it. `upsert_address`'s conflict arm keeps the first writer's encoding unless `force_update_address`, so the shielded note linked to a bare `tm…` string, and the wallet displayed a transparent address as the recipient of a fully-shielded migration output.

`upsert_address` now detects an existing row whose stored encoding lacks a receiver the incoming address carries, re-encodes the union of both receiver sets as a unified address, and updates the row in place (address text, receiver flags, exposure height). Row id and the transparent bookkeeping columns (`cached_transparent_receiver_address`, `transparent_child_index`) are preserved — and remain valid, because the union keeps the transparent receiver they cache.

## Testing

- New regression test `internal_address_upsert_merges_with_preexisting_transparent_row` drives a `TransferType::AccountInternal` Orchard note at diversifier index 0 through the real `put_received_note` path against the account fixture's genuine gap-limit row, asserting: same row id (no duplicate row), `receiver_flags` become the union, `cached_transparent_receiver_address` unchanged, and the stored encoding decodes as a UA carrying both receivers.
- Existing test expectations that the old behavior had been masking are updated (details in the first commit message).
- Full `zcash_client_sqlite` suite green with `--features "orchard,transparent-inputs,unstable,serde,test-dependencies"`.
- Verified end-to-end on an Android testnet wallet built against this branch: after a fresh sync, the migration transaction's outputs resolve to the account's internal-scope unified address in `v_tx_outputs` (previously NULL), and the pre-existing gap-limit row upgraded in place (`receiver_flags` 1 → 9, cached transparent receiver intact).


zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -376,13 +379,48 @@ pub(crate) fn ensure_address<
...
+        // need an address row: without one, `v_tx_outputs` has no `to_address` for the note, and
+        // self-transfers show up with a blank recipient. The external ivk above can't decrypt an
+        // internal-scope receiver's diversifier index — external and internal scopes are derived
+        // from distinct diversifier keys — so derive the internal-scope ivk from the account's
+        // full viewing key instead. An internal-scope output can only have been decrypted if the
+        // account holds a full viewing key, since deriving an internal ivk requires spend
+        // authority that a view-only (incoming-viewing-key-only) account does not have.
+        let ufvk = account.ufvk().expect(
+            "an internal-scope output was decrypted, so the account holds a full viewing key",
+        );
```
nuttycom, line 399: NACK. Internal addresses must *NEVER* be recorded in the addresses table.

nuttycom (changes requested):

PR #2753, opened 2026-07-24
zcash_primitives 0.30.0 post-release merge

nuttycom:


dannywillems:
> Note: this review was generated by Claude Opus 4.8 (Claude Code) on my behalf, and reviewed by me before posting.

Reviewed both commits. LGTM.

**16b1abd (CHANGELOG rewrite).** Collapsing `0.8.0-rc.1` into `0.8.0`, normalizing the subsection order to Added / Changed / Removed, grouping the Ironwood entries per role, and promoting the serde removal into its own `Removed` section (with the clarification that the `sapling` and `transparent` logical types keep their impls). No dangling `0.8.0-rc.1` link reference is left behind.

I checked the dropped entries against the source at this commit, and they are all correct to drop:

- `pczt::orchard::Bundle::resolve_memo_plaintexts` does not exist in the crate. `pczt/src/orchard.rs` has only `pub fn resolve_fields`, and the sole occurrence of the name in the repo was the changelog line itself.
- "`Creator::build_from_parts` now uses an all-zeroes anchor for absent Sapling and Orchard bundles" is stale. The canonical empty bundles carry `anchor: None` (`pczt/src/sapling.rs:54`, `pczt/src/orchard.rs:106` and `:117`); `DEFAULT_ANCHOR` survives only as a parse-time placeholder for redacted empty bundles (`pczt/src/sapling.rs:484`).
- The empty-bundle `value_sum` and IO Finalizer `bsk` details are implementation detail rather than public API. `bsk` is `pub(crate)` with no `#[getset(get = "pub")]`, so it is not reachable by a user of the crate at all. Per the changelog policy in `AGENTS.md` ("never describe implementation details or contracts that are not visible to a user of the public API"), these belong out of the entry; the user-visible consequence, that empty bundles have a single canonical representation so copies taken through different serialization paths still merge, is retained.
- The batch-signer wire format is compressed to "a versioned Postcard wire format", but the full encoding is documented in the module rustdoc (`pczt/src/roles/signer/batch.rs:8-12`).

The rc.1 collapse also lines up with the same policy's requirement that entries never carry interstitial changes to APIs that were modified more than once since the last release.

The additions look like real gaps being closed: `ExtractError::{IronwoodExtract, IronwoodParse}`, `tx_extractor::{IronwoodError, Error::Ironwood}`, `io_finalizer::Error::IronwoodFinalize`, `Prover::{requires_ironwood_proof, create_ironwood_proof}`, `Redactor::redact_ironwood_with`, the full `creator::Error` variant list, and the `prover::IronwoodError` to `prover::orchard::IronwoodError` path correction.

**6469caa (cargo vet metadata).** Verified each assertion against crates.io: `pczt` 0.8.0, `zcash_primitives` 0.30.0, and `zcash_proofs` 0.30.0 were all published 2026-07-24 by `nuttycom`, matching the updated `[[publisher.*]]` stanzas. The three `[[unpublished.*]]` stanzas are correctly dropped now that those versions exist.

dannywillems (approved):
utACK

PR #2755, opened 2026-07-24
Add storage-trait operations needed by external sync engines

nuttycom:
Adds three operations to the `zcash_client_backend` storage traits, with `zcash_client_sqlite` implementations. Each replaces an operation that an out-of-tree sync engine built on these crates would otherwise need direct SQL against wallet-internal tables to express:

- **`WalletRead::get_wallet_recover_until`** — the latest `recover_until` height among the wallet's accounts (the wallet-wide recovery horizon), mirroring `get_wallet_birthday`'s wallet-wide shape. Drives an "is the wallet still recovering?" signal.
- **`WalletWrite::prune_scan_queue_below`** — removes (or trims to a height floor) scan-queue entries below a given height. With `retain_with_priority: None`, nothing below the height is retained, irrespective of priority; with `Some(priority)`, entries at or above that priority are retained, as are the `Scanned`/`Ignored` bookkeeping entries (the store's own record of chain coverage — e.g. the canonical `Ignored` range spanning activation → wallet birthday — which the backend maintains itself; pruning those would make the queue shape depend on how often the wallet was opened). Primary use case: discarding `Historic` ranges that no remaining account justifies after `delete_account`, which deliberately does not modify the scan queue. The sqlite implementation keeps the common no-op case read-only (a probe inside the deferred transaction), so callers pruning at every wallet open never contend for the write lock.
- **`WalletCommitmentTrees::{get_sapling_subtree_root, get_orchard_subtree_root, get_ironwood_subtree_root}`** — paired readers for the existing `put_*_subtree_roots` methods, returning the store's record of a completed subtree's root hash, if known. The Ironwood variant has a default implementation returning `Ok(None)`, mirroring `with_ironwood_tree_mut`.

Includes mock implementations in `data_api::testing` (Ironwood mock modeling is tracked separately in #2756), unit tests for all three operations in `zcash_client_sqlite`, and changelog entries.

---

### DIFF
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -92,7 +92,7 @@ use zip32::{DiversifierIndex, fingerprint::SeedFingerprint};
 
 use self::{
     chain::{ChainState, CommitmentTreeRoot},
-    scanning::ScanRange,
+    scanning::{ScanPriority, ScanRange},
 };
 use crate::{
     data_api::{
@@ -2057,6 +2057,15 @@ pub trait WalletRead {
     /// or `Ok(None)` if the wallet has no initialized accounts.
     fn get_wallet_birthday(&self) -> Result<Option<BlockHeight>, Self::Error>;
 
+    /// Returns the height at which the wallet as a whole will have exited recovery mode.
+    ///
+    /// This returns the latest `recover_until` height among accounts maintained by this
+    /// wallet (see [`AccountBirthday::recover_until`]), or `Ok(None)` if no account has a
+    /// recovery horizon set (for example, in a wallet whose accounts were all created at
+    /// the chain tip rather than restored from backup). Heights below the returned value,
+    /// exclusive, are in scope for wallet recovery for at least one account.
+    fn get_wallet_recover_until(&self) -> Result<Option<BlockHeight>, Self::Error>;
+
     /// Returns a [`WalletSummary`] that represents the sync status and the wallet balances as of
     /// the chain tip given the specified confirmation policy for all accounts known to the wallet,
     /// or `Ok(None)` if the wallet has no summary data available.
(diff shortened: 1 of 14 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -3614,6 +3623,25 @@ pub trait WalletWrite: WalletRead {
...
+    /// entries deleted or trimmed.
+    ///
+    /// This is a queue-hygiene operation. The primary use case is discarding historic scan
+    /// ranges that no remaining account justifies: [`WalletWrite::delete_account`] does not
+    /// modify the scan queue, so the deep ranges queued for a since-deleted account's
+    /// birthday would otherwise still be scanned even though no remaining account can have
+    /// notes below its own birthday. In that case, pass the wallet birthday
+    /// ([`WalletRead::get_wallet_birthday`]) as `below` and [`ScanPriority::Historic`] as
+    /// `priority`; ranges required for note witnesses (e.g. [`ScanPriority::FoundNote`])
+    /// carry other priorities and are preserved.
```
nuttycom: The semantics here should be `retain_with_priority` and should be an Option<ScanPriority> - if None, then no ranges below the given height should be retained irrespective of their priority. If Some priority is specified, all ranges at that priority and greater should be retained. I would also name this method `prune_scan_queue_below` and make the argument `height` for the BlockHeight.

zcash_client_backend/src/data_api/testing.rs
```diff
@@ -3552,4 +3584,25 @@ impl WalletCommitmentTrees for MockWalletDb {
 
         Ok(())
     }
+
+    #[cfg(feature = "orchard")]
+    fn get_orchard_subtree_root(
```
nullcopy, line 3589: Also need `get_ironwood_subtree_root()`, no?

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -44,6 +44,56 @@ pub(crate) fn priority_code(priority: &ScanPriority) -> i64 {
...
+         WHERE priority > :floor AND priority < :retain AND block_range_start < :height",
+        named_params![":floor": floor_code, ":retain": retain_code, ":height": height],
+        |r| r.get(0),
+    )?;
+    if candidates == 0 {
+        return Ok(0);
+    }
+    let deleted = conn.execute(
+        "DELETE FROM scan_queue
+         WHERE priority > :floor AND priority < :retain AND block_range_end <= :height",
```
nullcopy: (nit) Would be nice to avoid repeating yourself here

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -44,6 +44,56 @@ pub(crate) fn priority_code(priority: &ScanPriority) -> i64 {
...
+        named_params![":floor": floor_code, ":retain": retain_code, ":height": height],
+    )?;
+    // The queue's global non-overlap invariant means at most one entry can straddle
+    // `height`, so the trimmed start is unique.
+    let trimmed = conn.execute(
```
nullcopy: Doesn't this leave a gap between the start_height of this range, and the end_height of the preceding range?

nullcopy: It doesn't look like any of the tests look for gaps. Maybe I'm wrong about that invariant :shrug:

nullcopy (changes requested):
Unless I'm mitaken, this can introduce gaps in the scan queue. Which we want to avoid, right?

Reviewed as of 1f831bc3ae40f653e250a87ff8e9a0131fbb1222

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: No, you're right about the invariant. This should be a two stage process - first, delete lower ranges, then second, reset the start height of the range with the minimum height. The priority conditions here might throw a wrench into that.

zcash_client_backend/src/data_api/testing.rs
dannywillems: And the `put_ironwood_subtree_roots` is missing I think too.

dannywillems: It is in testing though.

zcash_client_backend/src/data_api.rs
```diff
@@ -3614,6 +3623,40 @@ pub trait WalletWrite: WalletRead {
...
+    /// irrespective of their priority. If it is `Some(priority)`, entries with that
+    /// priority and greater are retained (left untouched, even where they straddle
+    /// `height`), as are entries with the bookkeeping priorities
+    /// [`ScanPriority::Scanned`] and [`ScanPriority::Ignored`] — those record which
+    /// regions of the chain the backend has already covered or deliberately skips, and
+    /// removing them would cause the backend to forget coverage state it maintains
+    /// itself (use the `None` form when that full reset is the intent). Entries with
+    /// priorities between the bookkeeping ones and the retained threshold are pruned.
+    ///
+    /// Pruning must not leave a gap in the queue's coverage: implementations are required
```
nullcopy, line 3639: Can we add a test that tries some pruning corner cases and checks for gaps?

nuttycom (commented):
self-utACK ab87e6a7f50300340e18aab41660306698d811d6

zcash_client_backend/src/data_api/testing.rs
nuttycom: I had a follow-up filed for this: https://github.com/zcash/librustzcash/issues/2756

zcash_client_sqlite/src/wallet/scanning.rs
```diff
@@ -44,6 +44,125 @@ pub(crate) fn priority_code(priority: &ScanPriority) -> i64 {
...
+
+    if replacement == existing {
+        return Ok(0);
+    }
+
+    conn.execute(
+        "DELETE FROM scan_queue WHERE block_range_start < :height",
+        named_params![":height": u32::from(height)],
+    )?;
+    insert_queue_entries(conn, replacement.iter())?;
```
nullcopy, line 158: Isn't this pretty expensive? It attempts to reinsert the full set of retained ranges, including ranges which were not touched by the delete operation. This could be a lot of unnecessary queries, right?

nullcopy (commented):
flushing comments as of 95acc6d177e767b4be53148e74dcb9a2bf679ef0

zcash_client_sqlite/src/wallet/scanning.rs
nuttycom: The ranges typically don't end up all that fragmented because we don't jump around that much in scanning; I haven't seen the queue get to larger than say 10 or so ranges.

nullcopy (approved):
I'd like to see a gap check test, but the logic looks correct to me.
utACK ab87e6a7f50300340e18aab41660306698d811d6

dannywillems (approved):
utACK

PR #2759, opened 2026-07-24
Make the pool-migration anchor bucketing interval configurable

### LINKED ISSUE #2758: Pool migration: make the anchor bucketing interval configurable
The ZIP 318 anchor bucketing interval is hard-coded as 144 blocks in two independent places:

- `zcash_client_backend::data_api::ll::wallet::ANCHOR_RETENTION_INTERVAL`, which decides which note commitment tree checkpoints are retained as durable anchors; and
- `zcash_pool_migration::scheduling::BOUNDARY_MODULUS` (defined as an alias of `MEAN_DELAY`), which decides which heights a migration transfer may anchor to.

These two must agree. They meet in `WalletMigrationProver::prove_orchard`, at `tree.root_at_checkpoint_id(&anchor_height)`: if the scheduler's grid is not a subset of the retention grid, transfers become unprovable at run time with a bare `WalletProveError::AnchorNotFound` and no compile-time or commit-time signal.

Being pinned to 144 blocks is also a practical obstacle to testnet testing. At the ~75-second target spacing a single bucket is about three hours, so exercising a migration end to end against a live test network means waiting out multiple buckets per transfer.

Both constants should be replaced by types threaded as function arguments and configuration, so the interval can be set to an arbitrary block interval. The design should eliminate the opportunity for the two grids to diverge rather than merely documenting that they must match; static or lazily initialized constants are not a satisfactory solution.

The transfer and preparation inter-arrival delay constants (`MEAN_DELAY`, `MAX_DELAY`, `PREP_MEAN_DELAY`, `PREP_MAX_DELAY`) are in scope for the same reason: a short bucket interval combined with a 144-block mean delay still spreads a migration over an impractical span for testing. `ANCHOR_AGE_CAP` and the expiry-window constants are out of scope.

---

nuttycom:
Closes #2758.

The ZIP 318 anchor bucketing interval was a hard-coded 144 blocks in two independent places, which had to agree or migration transfers became unprovable:

| | |
|---|---|
| `zcash_client_backend::data_api::ll::wallet::ANCHOR_RETENTION_INTERVAL` | which checkpoints survive pruning |
| `zcash_pool_migration::scheduling::BOUNDARY_MODULUS` (an alias of `MEAN_DELAY`) | which heights a transfer may anchor to |

They meet at `tree.root_at_checkpoint_id(&anchor_height)` in `WalletMigrationProver::prove_orchard`. A divergence surfaced only at proving time, as a bare `AnchorNotFound`.

## Approach

Newtypes carrying the arithmetic, threaded as arguments and configuration; no static or lazily initialized constants.

**`zcash_client_backend`** gains `data_api::anchor_retention::{AnchorRetentionInterval, AnchorRetention}`. The interval owns the boundary arithmetic — `is_boundary`, `boundary_at_or_below`, `boundary_at_or_above` — replacing the hand-written modulo expressions previously scattered across the retention and scheduling code. `put_blocks` and `update_tree` take `Option<AnchorRetention>` in place of a bare activation height.

**`zcash_pool_migration`** gains `AnchorBucketInterval`, `DelayDistribution` (a mean and the cap above which a draw is rejected, which must not be below it), and `SchedulingParams`. `MEAN_DELAY`, `MAX_DELAY`, `PREP_MEAN_DELAY`, `PREP_MAX_DELAY` and `BOUNDARY_MODULUS` are removed; `SchedulingParams::ZIP_318` carries the specified values.

**`zcash_client_sqlite`** gains `WalletDb::with_anchor_retention_interval`, following the `GapLimits` pattern.

## How divergence is closed

The interval is never configured on the migration side. `MigrationBackend::scheduling_params` is a **required** trait method — a default would silently reinstate the hazard — and `WalletMigration` implements it as `SchedulingParams::new(wallet.anchor_retention_interval().into(), …)`. On the wallet-backed path the interval is therefore unreachable from the caller; only the delays are settable, via `with_scheduling_delays`. `WalletRead::anchor_retention_interval` is the accessor a migration reads the grid back through, and it is defaulted, so downstream backends are unaffected.

Changing the interval mid-migration is caught rather than left to fail obscurely: the committed grid is recorded on `MigrationState` (and in a new `anchor_bucket_interval` column), and a mismatch is reported as `ProveError::AnchorIntervalMismatch` or `RebuildError::AnchorIntervalMismatch`. The rebuild check runs ahead of any per-transfer condition, since a moved grid invalidates every transfer rather than making one repairable.

The reverse `From<AnchorBucketInterval> for AnchorRetentionInterval` is deliberately omitted. It would have to call `AnchorRetentionInterval::custom`, which is `unstable`-gated, forcing `zcash_pool_migration`'s `wallet` feature to enable `zcash_client_backend/unstable`; it is also the wrong direction, since the wallet is the side that retains and so is the authority on the grid.

## Testing

A proptest across the `From` seam asserts the two implementations of the boundary arithmetic agree for any height and any interval — a drift there is precisely the failure this design exists to prevent.

`prove_chain_sim` gains an end-to-end test: a `WalletDb` configured with a 12-block retention interval commits a real migration, its preparations are mined, and each transfer's boundary is buried more than the pruning depth behind the chain tip before it is proved. The proof succeeds only because the wallet retained a checkpoint on its *configured* grid. Stubbing out `should_retain_anchor` makes it fail with `no root at the anchor checkpoint`, confirming the test is load-bearing rather than incidentally passing.

`anchor_checkpoints_retained_across_deep_scan` now runs at both the ZIP 318 interval and a short one, so the wallet is shown to retain the grid it was configured with. This makes that test somewhat slower; nothing else verified end to end that `with_anchor_retention_interval` takes effect.

## Note for review

The new `anchor_bucket_interval` column is added to the `orchard_ironwood_migration_tables` migration **in place** rather than via a follow-on migration, on the basis that it is not yet released: it was added on 2026-07-21 in a0e937819e, after the 0.22.0-rc.1 release of 2026-07-12, and sits under `[Unreleased]` in the changelog. If that rc reached users with the table present, this should become a separate migration instead.

---

### DIFF
--- a/zcash_client_backend/src/data_api/anchor_retention.rs
+++ b/zcash_client_backend/src/data_api/anchor_retention.rs
@@ -1,0 +1,222 @@
+//! Durable anchor-checkpoint retention: the block-height grid on which note commitment tree
+//! checkpoints are kept alive past the ordinary pruning window.
+//!
+//! A [ZIP 318] pool migration proves each pool-crossing transfer against the note commitment tree
+//! state at a BOUNDARY block rather than at the chain tip, so that many wallets' transfers share a
+//! small set of common anchors instead of each pinning a unique recent block. A transfer is proved
+//! long after its boundary has passed, so the boundary's checkpoint must still be present in the
+//! wallet's tree at proving time — ordinary checkpoint pruning would have discarded it. Retention
+//! exempts the boundary checkpoints from that pruning.
+//!
+//! [`AnchorRetentionInterval`] defines the grid, and owns the arithmetic that decides whether a
+//! height is a boundary. The migration scheduler MUST draw its anchors from the same grid the
+//! wallet retains: a transfer anchored to a height the wallet did not retain cannot be proved. See
+//! [`WalletRead::anchor_retention_interval`], the accessor through which a migration reads the grid
+//! back off the wallet that maintains it.
+//!
+//! [ZIP 318]: https://zips.z.cash/zip-0318
+//! [`WalletRead::anchor_retention_interval`]: super::WalletRead::anchor_retention_interval
+
+use core::num::NonZeroU32;
+
+use zcash_protocol::consensus::BlockHeight;
+
+/// The interval, in blocks, between the durable anchor checkpoints a wallet retains.
+///
+/// A height `h` is a BOUNDARY of this interval exactly when `h` is a multiple of it (see
+/// [`Self::is_boundary`]); those are the heights whose checkpoints are retained, and the only tree
+/// states a migration transfer may anchor to. The interval is a modulus, so it is necessarily
+/// non-zero.
+///
+/// The value recommended for use with this crate is supplied by the [`Default`] implementation,
+/// which is [`Self::ZIP_318`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct AnchorRetentionInterval(NonZeroU32);
+
+impl AnchorRetentionInterval {
+    /// The interval specified by [ZIP 318]: 144 blocks, roughly three hours (8 per day) at the
+    /// Zcash ~75-second target block spacing.
+    ///
+    /// [ZIP 318]: https://zips.z.cash/zip-0318
+    pub const ZIP_318: Self = Self(NonZeroU32::new(144).expect("144 is nonzero"));
+
+    /// Constructs an interval other than the [ZIP 318] one.
+    ///
+    /// A wallet on the production network MUST use [`Self::ZIP_318`]: the anonymity set a shared
+    /// anchor provides is exactly the set of transfers that chose the same boundary, so a wallet
+    /// retaining (and anchoring to) a different grid than its peers is distinguishable from them.
+    /// A shorter interval is useful on test networks, where waiting out 144-block buckets makes
+    /// exercising a migration impractical.
+    ///
+    /// This constructor is only available under the `unstable` feature, as it is not recommended
+    /// for general use.
+    ///
+    /// [ZIP 318]: https://zips.z.cash/zip-0318
+    #[cfg(any(test, feature = "test-dependencies", feature = "unstable"))]
+    pub const fn custom(blocks: NonZeroU32) -> Self {
+        Self(blocks)
+    }
+
+    /// Returns the interval as a number of blocks.
+    pub fn block_count(&self) -> NonZeroU32 {
+        self.0
+    }
+
+    /// Returns whether `height` is a boundary of this interval, i.e. whether a checkpoint at
+    /// `height` is retained as a durable anchor.
+    pub fn is_boundary(&self, height: BlockHeight) -> bool {
+        u32::from(height) % self.0 == 0
+    }
+
+    /// Returns the greatest boundary height that does not exceed `height`, i.e. `height` rounded
+    /// DOWN to a multiple of this interval.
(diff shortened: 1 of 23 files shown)


zcash_pool_migration/src/scheduling.rs
```diff
@@ -63,38 +63,238 @@
...
+            cap: NonZeroU32::new(576).expect("576 is nonzero"),
+        },
+        preparation_delay: DelayDistribution {
+            mean: NonZeroU32::new(24).expect("24 is nonzero"),
+            cap: NonZeroU32::new(96).expect("96 is nonzero"),
+        },
+    };
+
+    /// Constructs scheduling parameters from their constituent parts.
+    pub fn new(
```
nuttycom, line 280: We should also add a `new_with_default_distributions` that takes a bucket interval and computes the distributions based on the same ratios relative to the bucket interval that ZIP 318 uses; this will be useful for testing.

zcash_client_sqlite/src/lib.rs
```diff
@@ -464,13 +466,33 @@ impl<P, CL, R> WalletDb<rusqlite::Connection, P, CL, R> {
...
     }
 }
 
+impl<C, P, CL, R> WalletDb<C, P, CL, R> {
+    /// Sets the interval on which this wallet retains note commitment tree checkpoints as durable
+    /// anchors, exempt from ordinary checkpoint pruning.
+    ///
+    /// A ZIP 318 pool migration run over this wallet reads the interval back through
+    /// [`WalletRead::anchor_retention_interval`] and draws its transfers' anchors from the same
+    /// grid, so the two cannot disagree. Note that the interval is not persisted: an application
```
nuttycom: The configured retention interval is not persisted. After reopening a custom-grid wallet without reapplying this setter, a pruning-capable scan can pass a required custom boundary without retaining it. Reapplying the custom setting later makes the migration interval check pass, but the checkpoint is already missing, recreating the unexplained `AnchorNotFound` failure this change aims to prevent. Could we persist the interval, or validate the active migration’s stored interval before any scan that may prune checkpoints?

zcash_client_backend/src/data_api/anchor_retention.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Returns the greatest boundary height that does not exceed `height`, i.e. `height` rounded
+    /// DOWN to a multiple of this interval.
+    pub fn boundary_at_or_below(&self, height: BlockHeight) -> BlockHeight {
+        let h = u32::from(height);
+        BlockHeight::from_u32(h - (h % self.0))
+    }
+
+    /// Returns the least boundary height that is not below `height`, i.e. `height` rounded UP to a
+    /// multiple of this interval. Saturates at [`u32::MAX`].
+    pub fn boundary_at_or_above(&self, height: BlockHeight) -> BlockHeight {
```
nuttycom, line 91: `boundary_at_or_above` can return a value that is not a boundary near `u32::MAX`. For example, rounding `u32::MAX` on the ZIP-318 grid saturates to `u32::MAX`, which is not divisible by 144. The duplicate implementation in `zcash_pool_migration::scheduling` has the same issue. This violates the public API’s boundary guarantee; could this return `Option`/`Result` when no representable upper boundary exists, or otherwise preserve the invariant?

nuttycom: Heights near u32::MAX are not really reachable in practice (such heights are in the distant future) so I think this is ignorable for now; the `Option` here would clutter things up.

Cosmos-Harry (approved):
utACK

PR #2765, opened 2026-07-25
zcash_pool_migration: signing rounds bounded by a per-interaction action budget

dannywillems:
## Summary

Adds signing-round packing to `zcash_pool_migration`: group a migration's
transactions into external-signer **signing rounds** bounded by a
per-interaction **action budget**, and expose the counts a wallet needs to
preview the work for the user (interactions and signing workload).

This models the Keystone hardware wallet's limit of at most **96 total Orchard
actions across all transactions in one signing round** (a per-round total, not a
per-transaction cap): e.g. 6 preparation transactions of 16 actions, or 32
transfers of 3 actions, or any mix summing to ≤ 96.

## The user model: balance → quanta → runs → rounds

- The starting point is the wallet **balance**. The canonical planner quantizes
  it into crossings (the "quanta"), each a canonical denomination, **after**
  reserving the per-crossing transfer buffer and preparation fees. So a 2 ZEC
  balance is not one crossing: it becomes 7 quanta and migrates 1.99 ZEC.
- A balance beyond one **run**'s note cap (50 quanta) migrates over several runs;
  runs are serialized by mining.
- Each run's transactions are packed into signing **rounds** bounded by the
  signer's action budget. Rounds cannot span runs, so interactions are summed
  per run.

Invariant enforced throughout: a migration always migrates every funding note it
prepares, so `transfers >= 1` (never a zero-transfer plan).

## API (all additive; transparent to existing callers)

`MigrationPlan` stays a concrete type and `plan_migration` /
`estimate_migration_runs` keep their signatures. The budget is a **query
parameter**, so a plan is evaluated for any signer without re-planning.

New module `zcash_pool_migration::signing_rounds`:

- `SigningRoundBudget` — a `NonZeroU32` newtype with `KEYSTONE` (96) and
  `DEFAULT` (512) constants (no "unlimited" case), passed at the API call.
- `SigningRoundStrategy` trait (mirrors `DenominationStrategy`) with the named
  solutions `MinRounds` (the optimal packing) and `NextFit` (order-preserving
  greedy).
- `PlannedTx`, `PlannedSigningRound`, `min_signing_rounds`, `min_budget_for_rounds`.

On `MigrationPlan` (additive methods): `signing_rounds(budget)` /
`signing_round_count(budget)` / `signing_rounds_with(strategy, budget)`,
`group_unsigned(unsigned, budget)` (drives the real signing sessions, matching
the preview by id), `planned_transactions()`, and totals
(`total_transactions`, `preparation_tx_count`, `transfer_tx_count`,
`total_actions`, `value_migrated`, `residual`, `min_budget_for_single_round`).

On the estimate side, the previous **wrong** per-transaction model
(`RunEstimate::signing_sessions`) is replaced by the per-round **action** model
(`signing_rounds` / `total_signing_rounds`).
`batch_unsigned_by_action_budget` now takes a typed `SigningRoundBudget`.

### Signing workload (time estimate)

`RunEstimate::actions()` and `MigrationRunEstimate::total_actions()` give the
total Orchard actions the user must sign — the signing **workload**, a proxy for
signing time (multiply by the device's per-action time). This is distinct from
the number of interactions (the rounds): a signer's per-round budget only splits
the same workload into more or fewer rounds.

## The optimization

Grouping into the fewest rounds is one-dimensional **bin packing** (NP-hard), but
this instance has only two distinct item sizes (preparation = 16, transfer = 3),
a polynomial special case. `MinRounds` computes the exact optimum with a small
dynamic program, so the number of signer interactions is minimal, not a greedy
approximation. Reordering across rounds is sound: every transaction is
independent at signing time (anchors/witnesses deferred, ZIP 374), so the
prep→transfer dependency gates **broadcast** order only, not signing.

## Tests

- **Packer golden vectors** (`SIGNING_ROUND_GOLDEN_VECTORS` + `assert_golden_*`):
  inputs `(n_prep, n_transfer, budget)` → hand-verified optimal round counts,
  every regime; checked against brute force; budgets named from the
  `KEYSTONE`/`DEFAULT`/`minimum_feasible` constants (no magic numbers).
- **Balance-keyed scenarios** (`MIGRATION_SCENARIOS`): the fee-aware source
  balances and their expected plan shape are defined **once** in the `testing`
  module and reused by **both** the real-proving chain simulation
  (`prove_chain_sim`) and the signing-round end-to-end test, so a balance
  decomposes identically everywhere.
- **Evolution tables** (plan-only): `SIGNING_ROUND_EVOLUTION` shows how the
  Keystone round count steps up with the action total within one run (1 round at
  46 actions → 3 at 230); `MULTI_RUN_EVOLUTION` shows the totals grow across runs
  (2 runs / 359 actions / 5 rounds up to 11 runs / 2399 actions / 32 rounds).
- **End-to-end usage tutorial** (`tests/signing_rounds_e2e.rs`): the full
  Keystone flow (plan → preview → build → group → sign → apply), the software
  single-round path, preview-only estimate + inverse-budget + workload queries,
  and strategy selection.
- The reusable conformance suite (`arb_*` / `assert_*`) lives in the `testing`
  module (feature `test-dependencies`) so downstream crates inherit it.

All tests pass (130 lib + 3 + 5 + 7 e2e), plus the real-proving `prove_chain_sim`
(2 tests) after the scenario refactor. `cargo fmt`, `clippy` (incl. `wallet`
targets), and `cargo doc` are clean. The SQLite store interface is unaffected
(it depends only on the persisted types and the store traits, none of which
changed; its 13 store-conformance tests pass).

## Notes for review

Two decisions to confirm:
- `DEFAULT = 512` (finite, above one run's action total) — easy to change.
- `MigrationPlan` is kept concrete rather than made a trait, prioritizing
  transparency for the API user; the pluggable-strategy requirement lives in
  `SigningRoundStrategy`.

### DIFF
--- a/zcash_pool_migration/src/signing_rounds.rs
+++ b/zcash_pool_migration/src/signing_rounds.rs
@@ -498,52 +498,6 @@ fn div_ceil_u64(a: u64, b: u64) -> u64 {
 mod tests {
     use super::*;
 
-    fn synth(n_prep: usize, n_transfer: usize) -> Vec<PlannedTx> {
-        let mut txs = Vec::new();
-        let mut id = 0u32;
-        for i in 0..n_prep {
-            txs.push(PlannedTx::new(
-                MigrationTxId::new(id),
-                MigrationTxKind::Preparation { layer: 0, index: i },
-            ));
-            id += 1;
-        }
-        for c in 0..n_transfer {
-            txs.push(PlannedTx::new(
-                MigrationTxId::new(id),
-                MigrationTxKind::Transfer { crossing: c },
-            ));
-            id += 1;
-        }
-        txs
-    }
-
-    // A brute-force minimum round count for small instances, to check `MinRounds` optimality.
-    fn brute_min_rounds(n_prep: usize, n_transfer: usize, cap: u32) -> usize {
-        // Try increasing round counts; feasible if preparations fit (<= mp each) and transfers fit
-        // in the residual capacity, maximizing residual by an inner DP identical to production.
-        if n_prep == 0 && n_transfer == 0 {
-            return 0;
-        }
-        let w_hi = PREPARATION_ACTIONS;
-        let w_lo = TRANSFER_ACTIONS;
-        if cap < w_hi {
-            let per = (cap / w_lo).max(1) as usize;
-            return n_prep + n_transfer.div_ceil(per);
-        }
-        let mp = (cap / w_hi) as usize;
-        let lo_cap = |k: usize| -> usize { ((cap - (k as u32) * w_hi) / w_lo) as usize };
-        for r in 1.. {
-            if r * mp < n_prep {
-                continue;
-            }
-            if best_prep_distribution(n_prep, mp, &lo_cap, r, n_transfer).is_some() {
-                return r;
-            }
-        }
-        unreachable!()
-    }
-
     fn assert_valid(
         strategy: &dyn SigningRoundStrategy,
         n_prep: usize,
@@ -551,7 +505,7 @@ mod tests {
         cap: u32,
     ) {
         let budget = SigningRoundBudget::new(NonZeroU32::new(cap).unwrap());
-        let txs = synth(n_prep, n_transfer);
+        let txs = crate::testing::planned_txs(n_prep, n_transfer);
         let rounds = strategy.pack(&txs, budget);
         // Coverage: every id exactly once.
         let mut ids: Vec<u32> = rounds
@@ -591,16 +545,13 @@ mod tests {
     }
 
     #[test]
-    fn min_rounds_matches_brute_force() {
+    fn min_rounds_matches_the_reference() {
         for n_prep in 0..14 {
             for n_transfer in 0..40 {
                 for &cap in &[16u32, 18, 24, 48, 96, 100] {
-                    let want = brute_min_rounds(n_prep, n_transfer, cap);
-                    let got = min_signing_rounds(
-                        n_prep,
-                        n_transfer,
-                        SigningRoundBudget::new(NonZeroU32::new(cap).unwrap()),
-                    );
+                    let budget = SigningRoundBudget::new(NonZeroU32::new(cap).unwrap());
+                    let want = crate::testing::min_rounds_reference(n_prep, n_transfer, budget);
+                    let got = min_signing_rounds(n_prep, n_transfer, budget);
                     assert_eq!(
                         got, want,
                         "n_prep={n_prep} n_transfer={n_transfer} cap={cap}"
--- a/zcash_pool_migration/src/signing_rounds.rs
+++ b/zcash_pool_migration/src/signing_rounds.rs
@@ -576,15 +576,18 @@ mod tests {
         );
     }
 
+    // The reusable golden vectors (inputs + expected optimal outputs, in `crate::testing`) validate
+    // the optimal packer against a fixed, human-checked table, and every strategy produces a valid
+    // packing on those inputs.
     #[test]
-    fn keystone_pure_cases() {
-        // 6 preparations of 16 == 96 fits one round; 32 transfers of 3 == 96 fits one round.
-        assert_eq!(min_signing_rounds(6, 0, SigningRoundBudget::KEYSTONE), 1);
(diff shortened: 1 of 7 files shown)


zcash_pool_migration/src/testing.rs
```diff
@@ -362,3 +368,556 @@ pub fn assert_update_transaction<S: PoolMigrationWrite>(
...
+/// An arbitrary set of [`PlannedTx`] for one run: `n_prep` preparation transactions followed by
+/// `n_transfer` transfers, in the canonical commit order with sequential ids (the same order a
+/// [`MigrationPlan`](crate::engine::MigrationPlan) enumerates).
+pub fn arb_planned_txs() -> impl Strategy<Value = Vec<PlannedTx>> {
+    arb_signing_round_counts().prop_map(|(n_prep, n_transfer)| planned_txs(n_prep, n_transfer))
+}
+
+/// Build the canonical planned-transaction list for `n_prep` preparation and `n_transfer` transfer
+/// transactions (preparation first, then transfers), with sequential ids.
+pub fn planned_txs(n_prep: usize, n_transfer: usize) -> Vec<PlannedTx> {
```
dannywillems, line 405: Duplicated with synth in signing_rounds.rs

zcash_pool_migration/tests/signing_rounds_e2e.rs
```diff
@@ -0,0 +1,412 @@
...
+
+/// Plan a migration for a wallet holding notes of the given `values` (in ZEC).
+fn plan_for(seed: u64, values_zec: &[u64]) -> (CommitMock, MigrationPlan) {
+    let notes: Vec<u64> = values_zec.iter().map(|v| v * COIN).collect();
+    plan_notes(seed, &notes)
+}
+
+/// Sign one round's unsigned PCZTs on the "device" and apply the signatures back, exactly as a
+/// wallet does with the bytes a Keystone returns.
+fn sign_round_and_apply(
```
dannywillems: I guess this could be exported:

dannywillems: To show a real usage, there is a trait to implement with the sign method.

zcash_pool_migration/src/signing_rounds.rs
```diff
@@ -0,0 +1,668 @@
...
+                MigrationTxId::new(id),
+                MigrationTxKind::Transfer { crossing: c },
+            ));
+            id += 1;
+        }
+        txs
+    }
+
+    // A brute-force minimum round count for small instances, to check `MinRounds` optimality.
+    fn brute_min_rounds(n_prep: usize, n_transfer: usize, cap: u32) -> usize {
```
dannywillems: Maybe this could move outside the testing env.

zcash_pool_migration/src/signing_rounds.rs
```diff
@@ -0,0 +1,668 @@
...
+
+fn div_ceil_u64(a: u64, b: u64) -> u64 {
+    a.div_ceil(b)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn synth(n_prep: usize, n_transfer: usize) -> Vec<PlannedTx> {
```
dannywillems: This seems duplicated with another method in another test.

zcash_pool_migration/tests/signing_rounds_e2e.rs
dannywillems: Done in b91d6c210b: `sign_round_and_apply` now signs through the `MigrationCrypto::sign` trait (the CommitMock backend implements it), instead of the lower-level `build::sign_pczt` helper, so the example shows the real signer seam.

zcash_pool_migration/src/testing.rs
dannywillems: Done in b91d6c210b: removed the private `synth` builder and reuse `testing::planned_txs`.

zcash_pool_migration/src/signing_rounds.rs
dannywillems: Done in b91d6c210b: moved it to the reusable `testing` module as `min_rounds_reference`, and made it self-contained (it no longer calls the production packer's internals), so it is a genuine cross-check a downstream strategy can also use.

nuttycom (approved):
utACK b91d6c210b18d11d1cabc18a027e91bc8a7e0e66

PR #2767, opened 2026-07-26
zcash_pool_migration: denomination and transfer-id terminology, and expose the value a transfer crosses

nuttycom:
Two defects in the pool-migration API surface, found while reviewing the Swift SDK's FFI over this engine.

## 1. "Note split" names only half the operation

The planner decides the *values* of the self-funding notes a run mints, and reaching those values may split a large note **or consolidate several small ones** — `preparation` does both, up to 15 spends into one output when consolidating. The name leaked all the way out to the mobile SDKs' user-facing APIs, where it tells a user with many small notes that their wallet is about to "split" them.

Renamed to the vocabulary the module already speaks internally (`MIGRATION_MAX_DENOMINATION_ZEC`, `strategies::CanonicalOneTwoFive`) and that ZIP 318 uses for the 1-2-5 series — neutral between splitting and consolidating:

| before | after |
|---|---|
| `note_splitting` (module) | `denomination` |
| `NoteSplitPlan` | `DenominationPlan` |
| `plan_note_split` | `plan_denominations` |
| `MigrationState::note_split` / `MigrationPlan::note_split` | `::denominations` |
| `arb_note_split_plan` | `arb_denomination_plan` |

`crossing_values`, `migration_outputs`, `note_fee_buffer` and the `preparation` module keep their names.

The pool-migration store's `note_split_*` SQL columns keep theirs too: they are created by `CREATE TABLE IF NOT EXISTS` in databases already written by the published `0.1.0-rc.1`, so changing them is a schema migration rather than a rename, and they are invisible to API consumers. Happy to do that separately if you'd rather the storage vocabulary match.

## 2. There was no accessor for what a transfer actually moves

`funding_notes()` is a **spend-side** value: a transfer spends its funding note, of which the fee buffer pays that transaction's own fee and only the remainder crosses. It was also the only accessor indexed by crossing, so a consumer marshaling a transfer for display reached for it and reported an amount one transfer fee too large — and not the round `{1,2,5}×10ⁿ` denomination the user consented to. Both mobile SDKs hit this; one had already open-coded `funding_note - note_fee_buffer` with a comment explaining the trap.

Added:

```rust
MigrationState::crossing_values()           -> &[Zatoshis]
MigrationState::transfer_crossing_value(tx) -> Option<Zatoshis>
MigrationPlan::crossing_values()            -> &[Zatoshis]
```

The per-transaction accessor indexes the crossing values by the transaction's own `transfer_crossing()`, so no consumer re-derives the arithmetic or needs to know the buffer exists. Both slices stay index-aligned with `funding_notes()` and, on the plan, with `schedule()`, so amount/height pairing is unchanged. The spend-side accessors keep their names and now state which side they are on.

## 3. `MigrationTxId` invited confusion with a real transaction id

The name reads as a Zcash transaction id, and consumers marshaling a migration transaction for their platform reached for it as though it were one. It is a row key into the persisted migration, meaningful before a transaction is built at all. Renamed to **`MigrationTransferId`** (and `testing::arb_migration_tx_id` → `arb_migration_transfer_id`).

The distinction is load-bearing, and the type's docs now say which way it runs: a `TxId` identifies one broadcast **attempt**, while this id identifies the **transfer**. `rebuild_expired_transfer` keeps the transfer id and produces a new transaction with a new `TxId` — so a consumer that keyed its own records on the `TxId` loses the thread exactly when a transfer expires and is rebuilt. Use the `TxId` to talk to the network about a transaction, and this id to talk about the transfer.

As with the denomination rename, the persisted `tx_id` SQL columns keep their names — they exist in databases written by the published `0.1.0-rc.1`.

## Notes

- Three commits, one per change, each verified to build alone.
- No `CHANGELOG.md` entry: `AGENTS.md` says a newly added crate's changelog carries **only** the initial-release line, and `zcash_pool_migration`'s still does. Say the word if you'd rather rc.1 adopters got an explicit rename table there.
- No version bump — that belongs to the release PR that cuts the next rc.
- `cargo test -p zcash_pool_migration --all-features` passes (two new unit tests cover the accessor); `cargo clippy --all-features --all-targets` clean on the touched crates; `cargo fmt --all --check` clean; each commit builds on its own (`git rebase --exec`).

---

### DIFF
--- a/zcash_client_sqlite/src/pool_migration.rs
+++ b/zcash_client_sqlite/src/pool_migration.rs
@@ -6,7 +6,7 @@
 //! committed migration is a set of pre-signed PCZTs plus their schedule and lifecycle state, so a
 //! wallet resumes a migration entirely from these tables after being closed or restarted.
 //!
-//! The schema is fully NORMALIZED: every structured value (the note-split plan, the preparation
+//! The schema is fully NORMALIZED: every structured value (the denomination plan, the preparation
 //! plan's transaction inputs/outputs and direct-funding notes, the transaction kind, and the
 //! dependency graph) is stored in typed columns and child tables, so it can be queried directly.
 //! The `BLOB` columns are the pre-signed transaction (`pczt`), which is genuinely unstructured,
@@ -42,7 +42,7 @@
 //!
 //! There is at most one migration in progress per pool per account, stored as a row in the pool's
 //! migrations table keyed by an `account_id` foreign key into `accounts` (with `ON DELETE CASCADE`,
-//! so an account's migration is removed with the account), with its transactions, note split, and
+//! so an account's migration is removed with the account), with its transactions, denomination plan, and
 //! preparation plan in the pool's child tables (addressed through that row's synthetic primary
 //! key). The pool's `PoolMigrations` type is the store: construct it over a `rusqlite::Connection`
 //! (the same one [`WalletDb`](crate::WalletDb) uses) and the [`AccountUuid`](crate::AccountUuid)
--- a/zcash_client_sqlite/src/pool_migration/orchard_ironwood.rs
+++ b/zcash_client_sqlite/src/pool_migration/orchard_ironwood.rs
@@ -162,10 +162,10 @@ mod retention_follows_the_committed_migration {
         TestBuilder, orchard::OrchardPoolTester, pool::ShieldedPoolTester,
     };
     use zcash_client_backend::data_api::{Account as _, WalletCommitmentTrees, WalletRead};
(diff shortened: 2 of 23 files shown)


nuttycom (commented):
self-utACK

dannywillems (dismissed):
utACK. Mostly renaming + adding some accessors.

dannywillems (dismissed):
utACK afa5636705e29510754810f53f30045fdc2db36d. Agree for the renaming, confusing.

dannywillems (approved):
utACK a1f00fd3b8fc9d7dbc8dc9855d7cec32f95f92f7

PR #2769, opened 2026-07-26
Select PCZT signer views by receiver capability; minimal-encoding serialization

### LINKED ISSUE #2768: PCZT signer redaction and serialization defaults are unsafe for deployed hardware signers
`zcash_client_backend::data_api::wallet::redact_pczt_for_signer` produces only the compact signer view, whose receiver requirements (v2 PCZT encoding, `resolve_fields` support) deployed hardware-signer firmware does not meet in its ordinary signing flow. Separately, `pczt 0.8.0` changed `Pczt::serialize()` from the v1-only encoding to unconditional v2, silently changing the wire bytes SDKs send to signers when they upgraded from pczt 0.7.

Observed effect: mobile-SDK Keystone sends fail at finalization with `Orchard(Extract(MissingSpendAuthSig))` — the device returns the PCZT unsigned, and the Combiner merges without complaint since a field missing on one side is not a conflict (see zcash/zcash-swift-wallet-sdk#1863, which adopted the compact view and surfaced this).

Proposed fix (PR to follow): give `redact_pczt_for_signer` an explicit `SignerView { Full, Compact }` capability argument, and make `Pczt::serialize()` emit the minimal encoding version capable of representing its content (v1 for v5 content, v2 otherwise). Investigating the Full view's v1-representability also surfaced a latent bug: `v1::Pczt::try_from` required an Orchard anchor even for an empty Orchard bundle, wrongly rejecting Sapling-only v5 PCZTs from the v1 encoding.

nuttycom:
Closes #2768.
Resolves MOB-1552

Two changes restoring a workable wire contract for deployed hardware signers:

**`pczt`: minimal-encoding `serialize()`.** `Pczt::serialize` previously always emitted the v2 encoding; it now emits the v1 encoding whenever the content is representable in it (pre-v6 transaction, canonical-empty Ironwood bundle, no compact-only field state), falling back to v2 otherwise. `parse()` already accepts both, so this is strictly compatibility-improving. Callers that need a pinned encoding use `pczt::v1::Pczt` / `pczt::v2::Pczt` directly.

Implementation note: v1-representability is decided by attempting `v1::Pczt::try_from` on a clone (gated behind cheap tx-version/Ironwood pre-checks) rather than by a parallel representability predicate — the full conditions live in private per-bundle `TryFrom` impls, and a duplicated predicate would drift. The clone only occurs on the pre-v6 path, where it is small relative to the serialization work itself.

**`zcash_client_backend`: explicit `SignerView` argument.** `redact_pczt_for_signer` now takes `SignerView::{Full, Compact}`. `Compact` is the previous behavior verbatim (and remains what `redact_pczt_for_batch_signer` composes on). `Full` is a conservative view for Signers that predate the compact view — including deployed hardware-signer firmware in its ordinary signing flow: it retains proofs, binding-signature keys, and anchors, clears Orchard/Ironwood/Sapling spend witnesses and dummy signing keys, and performs no field compaction, so its view of a v5 transaction stays representable in the v1 encoding (which the new `serialize()` then selects).

Along the way this surfaced and fixes a latent `pczt` bug: `v1::Pczt::try_from` required an Orchard anchor even for an empty Orchard bundle, wrongly rejecting Sapling-only v5 PCZTs from the v1 encoding; the empty bundle now gets the same placeholder-anchor fallback the Sapling bundle already had. Three field accessors (`orchard::Bundle::zkproof`, `orchard::Spend::dummy_sk`, `sapling::Spend::witness`) are added for the new pool-test assertions and as legitimate Signer-facing API.

Release implications: `pczt 0.9.0` (breaking: serialize behavior), `zcash_client_backend 0.24.0-rc.3` (breaking within the rc line: `redact_pczt_for_signer` signature). CHANGELOGs updated in the same commits as the API changes.

### DIFF
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -10,6 +10,11 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `pczt::orchard::Bundle::zkproof`
+- `pczt::orchard::Spend::dummy_sk`
+- `pczt::sapling::Spend::witness`
+
 ### Changed
 - `pczt::Pczt::serialize` now emits the minimal encoding version capable of
   representing the PCZT's content: the v1 encoding whenever the content is
@@ -19,6 +24,14 @@ workspace.
   with receivers that predate the v2 encoding, such as deployed hardware
   signers. Callers that require a specific encoding version should use
   `pczt::v1::Pczt` or `pczt::v2::Pczt` directly.
+
+### Fixed
+- `pczt::v1::Pczt::try_from` (and thus `pczt::Pczt::serialize`'s v1 encoding
+  selection) incorrectly required an Orchard anchor even for a PCZT with an
+  empty Orchard bundle, returning `EncodingError::RequiresV2` for such PCZTs
+  even though they carry no Orchard-protocol data. An empty Orchard bundle
+  now falls back to a placeholder anchor for the v1 encoding, matching the
+  existing behavior for an empty Sapling bundle.
 
 ## [0.8.0] - 2026-07-23
 
--- a/pczt/CHANGELOG.md
+++ b/pczt/CHANGELOG.md
@@ -10,6 +10,16 @@ workspace.
 
 ## [Unreleased]
 
+### Changed
+- `pczt::Pczt::serialize` now emits the minimal encoding version capable of
+  representing the PCZT's content: the v1 encoding whenever the content is
+  representable in it (a pre-v6 transaction with a canonical-empty Ironwood
+  bundle and no compact-only field state), and the v2 encoding otherwise.
+  Previously it always emitted the v2 encoding. This maximizes compatibility
+  with receivers that predate the v2 encoding, such as deployed hardware
+  signers. Callers that require a specific encoding version should use
+  `pczt::v1::Pczt` or `pczt::v2::Pczt` directly.
+
 ## [0.8.0] - 2026-07-23
 
 ### Added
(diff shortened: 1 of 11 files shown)


dannywillems (commented):
I see. Keystone still uses pczt 0.2.1: https://github.com/KeystoneHQ/keystone3-firmware/blob/master/rust/apps/zcash/Cargo.toml#L23. Should we update it?

nuttycom:
> I see. Keystone still uses pczt 0.2.1: https://github.com/KeystoneHQ/keystone3-firmware/blob/master/rust/apps/zcash/Cargo.toml#L23. Should we update it?

This is a question for @nullcopy

pczt/src/orchard.rs
```diff
@@ -750,6 +752,12 @@ pub mod v1 {
                 return Err(crate::EncodingError::UnsupportedOrchardNoteVersion);
             }
 
+            let anchor = match bundle.anchor {
+                Some(anchor) => anchor,
+                None if bundle.actions.is_empty() => super::DEFAULT_ANCHOR,
```
dannywillems, line 757: Note: this is also a diff in the patch. LGTM.

pczt/CHANGELOG.md
```diff
@@ -10,6 +10,29 @@ workspace.
 
 ## [Unreleased]
 
+### Added
```
dannywillems, line 13: Nit: should we provide aliases prefixed by `keystone_` or something that would be useful for upstream users? And probably AI agents too?

dannywillems (approved):
tACK 27bb53290423b010f8a65d3d8ce472c1957142d0.
As a follow-up, after Ironwood, it would be nice to update the pczt library in the firmware, if possible.

PR #2770, opened 2026-07-26
zcash_client_sqlite: pin the change pool for a mixed Orchard/Sapling spend

nuttycom:
Adds a test pinning which pool receives the change when a single transaction spends from both the Orchard and Sapling pools post-NU6.3.

### Background

`select_change_pool` prefers the Orchard pool whenever Orchard notes are spent, and demotes that preference to Ironwood only when the change would equal or exceed the Orchard input value (the turnstile bound). The nearest existing coverage, `orchard_combines_with_other_pools_within_turnstile`, exercises exactly this shape — an Orchard note and a Sapling note, neither sufficient alone — but asserts only the turnstile bound and that the transaction builds. Which pool actually received the change was unpinned, so a change in that policy would have gone unnoticed.

### What the test covers

A payment funded from both pools and delivered to an Orchard receiver (routed through the Ironwood bundle post-NU6.3), where the change is small enough that the turnstile permits its return to Orchard. It asserts the change assignment on the proposal, then the three bundle value balances on the **built** transaction, which is where the privacy property is actually observable:

| bundle | valueBalance |
|---|---|
| Orchard | +80,000 |
| Sapling | +100,000 |
| Ironwood | −150,000 |

The Orchard change output leaves `valueBalanceOrchard` a *lower bound* on the value of the notes spent — an observer cannot distinguish spending 100,000 and retaining 20,000 from spending 80,000 outright. Routing the change to Sapling instead would leave no Orchard output to absorb any of the input value, publishing the spent notes' total exactly. Neither assignment discloses the change amount itself: the value entering the Ironwood bundle is exactly the payment either way.

Test-only; no library changes and no CHANGELOG entry.

---

### DIFF
--- a/zcash_client_sqlite/src/wallet/orchard.rs
+++ b/zcash_client_sqlite/src/wallet/orchard.rs
@@ -1773,8 +1773,10 @@ pub(crate) mod tests {
         use zcash_keys::address::Address;
         use zcash_primitives::block::BlockHash;
         use zcash_protocol::{
-            PoolType, ShieldedPool, consensus::BlockHeight, local_consensus::LocalNetwork,
-            value::Zatoshis,
+            PoolType, ShieldedPool,
+            consensus::BlockHeight,
+            local_consensus::LocalNetwork,
+            value::{ZatBalance, Zatoshis},
         };
         use zip321::{Payment, TransactionRequest};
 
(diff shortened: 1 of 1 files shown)


zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -2742,6 +2744,149 @@ pub(crate) mod tests {
...
+            assert_eq!(
+                i64::from(sapling_value_balance),
+                i64::from(ZatBalance::from(sapling_note_value)),
+                "the Sapling pool is drained in full",
+            );
+            assert_eq!(
+                i64::from(ironwood_value_balance),
+                -i64::from(ZatBalance::from(payment_value)),
+                "exactly the payment enters the Ironwood pool, so the change amount is hidden",
+            );
```
schell, line 2887: nit: These three assertions could be written as something like `assert_eq!(orchard_value_balance, ZatBalance::from(orchard_note_value) - ZatBalance::from(expected_change))` and `assert_eq!(ironwood_value_balance, -ZatBalance::from(payment_value)),` dropping the i64::from(...) conversions, which fits the repo conventions better, but no biggy.

schell (approved):

PR #2771, opened 2026-07-26
pczt: test minimal-encoding serialization against Keystone firmware

dannywillems:
Layered on top of #2769.

Confirms that the minimal-encoding `Pczt::serialize()` restored by #2769 stays
wire-compatible with the deployed Keystone 3 firmware, which uses the `pczt`
0.2.1 crate.

## What it does

Adds `pczt/tests/firmware_compat.rs`, an integration test that pins the
canonical Orchard->transparent PCZT vector embedded in the firmware (produced
by the firmware's `pczt` 0.2.1, a **v1** encoding) and asserts:

- the firmware vector **parses** under this PR;
- the **default** `Pczt::serialize()` (the path a wallet SDK calls) emits the
  minimal (v1) encoding **byte-for-byte identical** to the firmware vector;
- the explicit `v1::Pczt` path also matches, while the `v2` path differs (the
  pre-fix default that diverged: v2 2748 bytes vs v1 2731 bytes);
- `serialize()` is **idempotent** across a parse round-trip.

Empirically: before #2769 the default emitted v2 and diverged from the
firmware (the regression that broke deployed signers, surfacing as
`Orchard(Extract(MissingSpendAuthSig))`); with #2769 it emits v1 that matches.

## Shared vector

The same bytes are guarded on the firmware side, so if either implementation's
v1 layout drifts, one of the two round-trip assertions fails. Vector source
(pinned commit):
https://github.com/KeystoneHQ/keystone3-firmware/blob/8420703ad8674745ca8f086590371661747b7dca/rust/apps/zcash/src/pczt/sign.rs#L214

### DIFF
--- a/pczt/tests/firmware_compat.rs
+++ b/pczt/tests/firmware_compat.rs
@@ -1,0 +1,111 @@
+//! Cross-implementation PCZT wire-compatibility with the Keystone 3 firmware.
+//!
+//! This test is a reviewer's aid for PR #2769 (issue #2768): it confirms that
+//! the minimal-encoding `Pczt::serialize()` restored by this PR produces bytes
+//! that the deployed firmware (which uses the `pczt` 0.2.1 crate) already
+//! parses and produces.
+//!
+//! The vector below is the canonical Orchard->transparent PCZT embedded in the
+//! keystone3-firmware source, produced by the firmware's `pczt` 0.2.1. It is a
+//! **v1** encoding (`PCZT` magic + little-endian version `1`). Keeping the
+//! exact bytes here and in the firmware test makes this a shared encoding
+//! vector: if either side's v1 layout drifts, one of the two round-trip
+//! assertions fails.
+//!
+//! Source of the vector (KeystoneHQ/keystone3-firmware, pinned commit):
+//! <https://github.com/KeystoneHQ/keystone3-firmware/blob/8420703ad8674745ca8f086590371661747b7dca/rust/apps/zcash/src/pczt/sign.rs#L214>
+//!
+//! The crucial claim is on the DEFAULT `serialize()` path a wallet SDK calls.
+//! Before this PR that default emitted v2 (2748 bytes) and diverged from the
+//! firmware; with this PR it emits minimal-encoding v1 (2731 bytes) that is
+//! byte-identical to the firmware vector.
+
+fn encoding_version(bytes: &[u8]) -> u32 {
+    assert_eq!(&bytes[..4], b"PCZT", "not a PCZT (bad magic)");
+    u32::from_le_bytes(bytes[4..8].try_into().unwrap())
+}
+
+/// Byte-for-byte the same constant used by the firmware unit test
+/// (`PCZT_V1_ORCHARD_TO_TRANSPARENT`). Do not regenerate casually: a change
+/// here is a wire-format change that must be mirrored in the firmware.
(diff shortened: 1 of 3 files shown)


pczt/tests/firmware_compat.rs
```diff
@@ -0,0 +1,131 @@
+//! Cross-implementation PCZT wire-compatibility with the Keystone 3 firmware.
+//!
+//! This test is a reviewer's aid for PR #2769 (issue #2768): it confirms that
```
dannywillems, line 3: I don't know if we want to keep this information in the patch. The reviewer can decide.

pczt/tests/firmware_compat.rs
```diff
@@ -0,0 +1,131 @@
...
+//!
+//! Source of the vector (KeystoneHQ/keystone3-firmware, pinned commit):
+//! <https://github.com/KeystoneHQ/keystone3-firmware/blob/8420703ad8674745ca8f086590371661747b7dca/rust/apps/zcash/src/pczt/sign.rs#L214>
+//!
+//! The crucial claim is on the DEFAULT `serialize()` path a wallet SDK calls.
+//! Before this PR that default emitted v2 (2748 bytes) and diverged from the
+//! firmware; with this PR it emits minimal-encoding v1 (2731 bytes) that is
+//! byte-identical to the firmware vector.
+
+fn unhex(s: &str) -> Vec<u8> {
```
dannywillems: Well... I guess this could be using `hex` no?

pczt/tests/firmware_compat.rs
```diff
@@ -0,0 +1,131 @@
...
...cc6e5ce73e43b51e054860d7939ca13d77813b372070fd24cdd9c0e2fad7567471c0279bba19a76f0cdbd3107220821dd676c1df6524c15b87c1318eda418d65f8c66d2a77a65f6894199d44611e60c0291c330d1692bd521aef0e316e2b3f8c377b0d6873b3b645196ba74a79c6e0509869ac66276c3e2dfefd54a12365b5945406e7b673321ed36e89a14a194ae8b864e9ac4684655bae7fcd3123a226f282ac6ac82ca88d6a383d8be90f87f4cb85225f697932abfb4c05cda3b6dadb003621fee663f3fcb8f1c96320a3f148bc106ec231961a8f5142dd614317eef16b81492668a8b8795b85d7b0f737fa8d79e9dc3d78840d158a73dc6d1700ce3a8de2a9f93ff1bc8108703b94fd5bd230a19dd0fd821b832d3508b335e07bac28e95c3ab0eb637334bf166fa2a440ea35c0372bb5a745ee86c727a80f0d0d080fef6642ae7aae1407d6a25c3050c498a52ae300105bded1f19829b10df00e7ba301a9aef2c99ad7c5338b0e259ab97ea852630606b8d59709ca067d32698c8761e0f7d5b76ac07d4860b0fe2992010ba88827bb37cf4e3436488580e79101b366d454f29aa2bdf76725130baa08b38af3a71c251521809c84fe3d086943f39f01d760884b6342fac60c010001c54930d4f4f9946dfe91ac3e94cf5b513871c4a5c0c21137959482da796d2d280000000001c4666732084baff2e402ed7d3e457303c73b77dbd4aa5bc943ac7ca96f3779070398a2e304004aed48232c44dbd0b0b5404063ecc4679436f28c6251cbba91e29388fcd98d0e0001dc2be19f4118dbb7500df3a95e304733b247cea7f8c681f6aaafceb8fc1d7d28";
```
dannywillems, line 31: Follow-up: we could have something cross-repo that works better.

The bytes could be from a file in the repository, and in the CI, we always fetch `master` of the keystone firmware which contains the test vectors the keystone firmware is tested against. That way, we can be warned if the repo impl drifts.

pczt/tests/firmware_compat.rs
dannywillems: Good call, done. Switched to `hex::decode` and dropped the hand-rolled `unhex`; added `hex` (already a workspace dep) to the pczt dev-dependencies and updated Cargo.lock.

pczt/tests/firmware_compat.rs
```diff
@@ -0,0 +1,111 @@
...
+    let v1 = pczt::v1::Pczt::try_from(pczt.clone())
+        .expect("firmware PCZT is v1-representable")
+        .serialize();
+    assert_eq!(encoding_version(&v1), 1);
+    assert_eq!(
+        v1, bytes,
+        "explicit v1 encoding matches the firmware vector"
+    );
+
+    let v2 = pczt::v2::Pczt::try_from(pczt)
```
dannywillems, line 84: LGTM for this check.

pczt/tests/firmware_compat.rs
```diff
@@ -0,0 +1,111 @@
...
+        .expect("v2 encode")
+        .serialize();
+    assert_eq!(encoding_version(&v2), 2);
+    assert_ne!(
```
dannywillems, line 88: LGTM

pczt/tests/firmware_compat.rs
nuttycom: Opened https://github.com/zcash/librustzcash/issues/2772

pczt/tests/firmware_compat.rs
nuttycom: I don't care for now, I want to get past CI runs faster.

nuttycom (approved):

PR #2776, opened 2026-07-26
PCZT signer-view follow-ups: leak-regression tests and v1 anchor normalization

### LINKED ISSUE #2775: PCZT signer-view follow-ups: full-view leak-regression tests, empty-bundle v1 anchor normalization
Follow-ups identified in the final review of #2769 (see also #2768):

1. **Full-view proprietary-redaction regression tests** (Important): the `SignerView::Full` pool-test block asserts sighash equality, witness/dummy clearing, and retention properties, but has no assertions that wallet-proprietary metadata (`proposal_info`/`output_info`) is removed — the redaction's primary purpose, and the one redaction path with deployed hardware-signer exposure. The compact-view blocks have these assertions; the full view should too. Also exercise the Full view against v6/Ironwood content (currently only Compact is exercised there).
2. **Local pczt unit test** for the empty-Orchard-bundle v1 anchor fallback introduced in #2769 (currently covered only transitively via zcash_client_sqlite pool tests).
3. **v1-decode normalization**: `From<pczt::orchard::v1::Bundle>` reconstructs `anchor: Some([0;32])` where the logical original was `None` for an empty bundle, diverging from the documented `EMPTY_ORCHARD` invariant ("the v1 decoder produces exactly this value"). Functionally inert (placeholder anchors merge and re-encode stably) but observable via `anchor().is_none()`. Normalize the decode; the Sapling side has the same pre-existing gap and needs its own compatibility analysis (deliberately not addressed here).
4. Doc accuracy touch-ups that follow from 3 (`Pczt::serialize` rustdoc, `EMPTY_ORCHARD` doc comment).

nuttycom:
Closes #2775. Follow-ups from the final review of #2769.

**Commit 1 (test-only):** adds the missing leak-regression assertions to the `SignerView::Full` pool-test block — global `proposal_info` and per-output `output_info` proprietary metadata must be absent from the full signer view (the compact block already asserted this; the full view is the one with deployed hardware-signer exposure and had no such coverage). Also exercises the Full view against v6/Ironwood content in `create_pczt_supports_ironwood_output`: sighash equality, Ironwood witness/dummy clearing, v6 anchor retention, and proprietary removal.

**Commit 2:** normalizes the v1 decode of an empty Orchard bundle — a placeholder anchor written by the v1 encoder's canonicalization now decodes back to `anchor: None`, restoring the documented `EMPTY_ORCHARD` invariant that all decode paths produce exactly the canonical empty bundle. Adds the local pczt unit test for the #2769 anchor fallback (positive case beside `v1_rejects_missing_anchor_and_cv_net`) extended to assert the exact round trip. The Sapling side has the same pre-existing asymmetry and is deliberately untouched pending its own compatibility analysis (tracked in #2775).

Changelog note: `main`'s `pczt` `Unreleased` section still contains the (already-published-as-0.9.0) #2769 content, so the new Fixed bullet currently sits in that section; it belongs under the post-0.9.0 Unreleased heading once the release stamping lands on `main` — happy to rebase/adjust when that merge is in.

Tests: `cargo test -p pczt --all-features` (69), sqlite pool `pczt` tests (6), `cargo clippy --all-features --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all clean.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -8089,6 +8089,32 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, Dsf>(
             .iter()
             .all(|spend| spend.witness().is_none())
     );
+    // Redaction's primary purpose is removing the wallet's proprietary metadata,
+    // which the Signer has no use for and should never see.
+    assert!(
+        !full_view
+            .global()
+            .proprietary()
+            .contains_key("zcash_client_backend:proposal_info")
+    );
+    for bundle in [full_view.orchard(), full_view.ironwood()] {
+        assert!(bundle.actions().iter().all(|action| {
+            !action
+                .output()
+                .proprietary()
+                .contains_key("zcash_client_backend:output_info")
+        }));
+    }
+    assert!(full_view.sapling().outputs().iter().all(|output| {
+        !output
+            .proprietary()
+            .contains_key("zcash_client_backend:output_info")
+    }));
+    assert!(full_view.transparent().outputs().iter().all(|output| {
+        !output
+            .proprietary()
+            .contains_key("zcash_client_backend:output_info")
+    }));
     if *full_view.global().tx_version() == zcash_protocol::constants::V5_TX_VERSION {
         let full_view_bytes = full_view.clone().serialize().unwrap();
         assert!(pczt::v1::Pczt::try_from(full_view).is_ok());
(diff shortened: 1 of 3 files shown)


zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -8089,6 +8089,32 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, Dsf>(
...
+    assert!(full_view.sapling().outputs().iter().all(|output| {
+        !output
+            .proprietary()
+            .contains_key("zcash_client_backend:output_info")
+    }));
+    assert!(full_view.transparent().outputs().iter().all(|output| {
+        !output
+            .proprietary()
+            .contains_key("zcash_client_backend:output_info")
+    }));
```
schell, line 8117: Might be a good idea to put these in a constant so we only have to change it in one place?

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -9508,6 +9534,42 @@ where
...
+    }
+    assert!(full_view.sapling().outputs().iter().all(|output| {
+        !output
+            .proprietary()
+            .contains_key("zcash_client_backend:output_info")
+    }));
+    assert!(full_view.transparent().outputs().iter().all(|output| {
+        !output
+            .proprietary()
+            .contains_key("zcash_client_backend:output_info")
```
schell, line 9570: Ditto here about the constant.

schell (approved):
LGTM. Just a nit.

nuttycom (commented):
self-utACK

PR #2778, opened 2026-07-26
Stamp ZIP 32 derivation on wallet-controlled zero-value spends in PCZTs

### LINKED ISSUE #2777: create_pczt_from_proposal omits zip32_derivation on wallet-controlled zero-value spends, leaving v6 PCZTs unsignable by external signers
In a post-NU6.3 (v6, dual-bundle) transaction, the vanilla-Orchard-pool bundle pairs change outputs with fabricated **wallet-controlled zero-value spends** (`orchard` builder, `SpendInfo { dummy_sk: None, fvk, .. }` — deliberate, since the legacy pool's cross-address rule precludes random-keyed dummy notes there). These spends must be signed by the wallet's spend authorizing key through the normal signing flow.

However, `create_pczt_from_proposal` stamps `zip32_derivation` only on the indices of the *requested* real spends (`orchard_spends` / `ironwood_spends`). The wallet-controlled zero-value spends receive no derivation. An external Signer that identifies its actions by derivation (e.g. Keystone firmware, which explicitly supports signing wallet-controlled zero-value spends that retain `alpha`) correctly skips them as "not ours", and nothing else can sign them — the IO Finalizer's dummy pre-signing does not apply (`dummy_sk` is absent by design). Transaction extraction then fails with `Orchard(Extract(MissingSpendAuthSig))` on those actions.

Byte-level evidence from a captured failing testnet request PCZT (v6/v2, all-Orchard funds, payment resolving to an Ironwood receiver): the Orchard bundle held one real spend (value + derivation present) and three zero-value spends with `alpha` present but `dummy_sk`, `spend_auth_sig`, and `zip32_derivation` all absent; the Ironwood bundle's padding dummies were correctly pre-signed by the IO Finalizer. Reproduced with both zcash_client_backend 0.24.0-rc.2/pczt 0.8.0 and current main — the bug predates the signer-view work (zcash/librustzcash#2768/[#2769](<https://github.com/zcash/librustzcash/issues/2769>)) and only manifests in v6 dual-bundle builds (v5 padding uses pre-signed `dummy_sk` dummies throughout, which is why v5 hardware flows never hit it).

Proposed fix: in `create_pczt_from_proposal`'s Orchard and Ironwood updater loops, stamp `zip32_derivation` on every action whose spend still requires a signature (`spend_auth_sig` is `None`) rather than only the requested-spend indices — after IO finalization, protocol dummies already carry signatures, and every remaining unsigned spend in a wallet-created PCZT belongs to the wallet's account. With derivations present, deployed hardware-signer firmware signs these actions without modification.

Observed downstream as zcash/zcash-swift-wallet-sdk#1869's residual failure mode.

nuttycom:
Closes #2777.

**The bug:** in post-NU6.3 (v6) transactions, the vanilla-Orchard-pool bundle pairs change outputs with wallet-controlled zero-value spends (`dummy_sk: None` by design — the legacy pool's cross-address rule precludes random-keyed dummies, so these spends use the wallet's own key and must be signed through the normal signing flow). `create_pczt_from_proposal` stamped `zip32_derivation` only on the requested real-spend indices, so these spends carried no derivation; derivation-matching external Signers (e.g. Keystone firmware, which explicitly supports signing wallet-controlled zero-value spends) correctly skip them as "not ours", nothing else can sign them, and transaction extraction fails with `Orchard(Extract(MissingSpendAuthSig))`. Confirmed byte-for-byte against a captured failing testnet request PCZT.

**The fix:** stamp the derivation on every action whose spend still lacks `spend_auth_sig` at update time — the Updater runs post-IO-finalization, so pre-signed protocol dummies are excluded by that check, and in a wallet-created PCZT every remaining unsigned spend belongs to the wallet's account. The previously-used `orchard_spends`/`ironwood_spends` index sets became dead and are removed. TDD: the Ironwood-output pool test first failed on a new assertion that every action is pre-signed or derivable (with a non-vacuity guard proving the scenario produces the zero-value wallet-spend shape), and passes with the fix.

**Why no automated test caught it:** upstream PCZT tests sign with local/software signers, which match spends by FVK and succeed regardless of derivation metadata; the firmware-compat test checks wire-format fidelity, not signer-metadata semantics. This PR therefore also adds the general signability invariant to `pczt_single_step` (all pool variants, via the Verifier role): every shielded spend must be already-signed, dummy-key-carrying, or ZIP 32-derivable whenever the account's derivation is known — this assertion alone would have caught the bug on day one and now guards the fix.

Also included: a changelog housekeeping commit moving the v1-decode anchor-normalization entry out of the published 0.9.0 heading (it shipped after 0.9.0; the entry belongs under Unreleased).

Tests: pool `pczt` suite (6/6, including the new invariant across Sapling/Orchard variants), `cargo clippy --all-features --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all clean.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -10,6 +10,13 @@ workspace.
 
 ## [Unreleased]
 
+### Fixed
+- PCZTs created by `create_pczt_from_proposal` for post-NU6.3 (v6) transactions
+  now carry ZIP 32 derivation metadata on wallet-controlled zero-value spends,
+  so external Signers can identify and sign them; previously such actions
+  were unsignable and transaction extraction failed with a missing
+  spend-auth signature.
+
 ## [0.24.0-rc.3] - 2026-07-26
 
 ### Added
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -7996,6 +7996,46 @@ pub fn pczt_single_step<P0: ShieldedPoolTester, P1: ShieldedPoolTester, Dsf>(
         consensus::BranchId::try_from(*pczt_created.global().consensus_branch_id())
             .expect("the PCZT carries a valid consensus branch ID");
 
+    // Signability invariant: the test account's derivation is known, so every
+    // shielded spend must be signable by SOME party: already signed
+    // (IO-finalized protocol dummies), carrying the dummy key that lets the IO
+    // Finalizer sign it, or exposing the ZIP 32 derivation metadata by which
+    // an external Signer identifies the spend as its own. A spend meeting none
+    // of these is unsignable and dooms transaction extraction.
+    fn assert_signable_orchard_bundle(bundle: &orchard::pczt::Bundle) {
+        for action in bundle.actions() {
+            assert!(
+                action.spend().spend_auth_sig().is_some()
+                    || action.spend().dummy_sk().is_some()
+                    || action.spend().zip32_derivation().is_some(),
+                "unsignable Orchard-protocol spend: no signature, dummy key, or derivation",
+            );
+        }
+    }
+    pczt::roles::verifier::Verifier::new(pczt_created.clone())
+        .with_orchard::<Infallible, _>(|bundle| {
+            assert_signable_orchard_bundle(bundle);
+            Ok(())
+        })
+        .unwrap()
+        .with_ironwood::<Infallible, _>(|bundle| {
+            assert_signable_orchard_bundle(bundle);
+            Ok(())
+        })
+        .unwrap()
+        .with_sapling::<Infallible, _>(|bundle| {
+            for spend in bundle.spends() {
+                assert!(
+                    spend.spend_auth_sig().is_some()
+                        || spend.proof_generation_key().is_some()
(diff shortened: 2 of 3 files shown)


nuttycom (commented):
self-utACK

zcash-harry:
Two gaps this doesn't close:

1. **Migration-path builders still need the same treatment.**
   `zcash_pool_migration_backend`'s `build_prep_tx`/`build_transfer_pczt` don't go
   through `create_pczt_from_proposal`, so they're unaffected by this fix and still
   produce PCZTs whose wallet-controlled zero-value spends carry no derivation.
   Consumers currently work around it by annotating downstream; the same
   `spend_auth_sig().is_none()` stamping should be applied in those builders.

2. **`redact_pczt_for_batch_signer` can still strip a spend into an unsignable state.**
   It clears `spend_auth_sig` on *every* action. A padding dummy reaches it signed (so
   correctly gets no derivation under the new predicate) with `dummy_sk` already cleared
   by `redact_pczt_for_signer` — afterwards it has no signature, no dummy key, and no
   derivation. It should skip clearing signatures on actions that carry no
   `zip32_derivation`, since no external signer will ever reproduce them. Note the new
   invariant added here won't catch this, as it runs pre-redaction.

_Findings by Claude_

PR #2783, opened 2026-07-27
Close the two remaining paths to unsignable wallet-controlled spends

### LINKED ISSUE #2782: Wallet-controlled zero-value spends remain unsignable in the migration builders and the batch signer view
Follow-up to zcash/librustzcash#2777, whose fix (`create_pczt_from_proposal` stamping ZIP 32 derivation on every spend that still needs a signature) is on `main`. Two paths to the same "unsignable action" defect were not covered by it.

### 1. `zcash_pool_migration`'s builders

`build_prep_tx` and `build_transfer_pczt` do not go through `create_pczt_from_proposal`, so they still emit PCZTs whose wallet-controlled zero-value spends — and whose real spends — carry no derivation. A preparation transaction pairs one such spend with every internal change output, since the post-NU6.3 vanilla-Orchard-pool cross-address rule precludes the random-keyed dummies used for padding.

Measured shapes of the built PCZTs (post-IO-finalization):

| built PCZT | actions | pre-signed | carrying derivation |
| -- | -- | -- | -- |
| preparation (1 spend, 2 outputs) | 16 Orchard | 13 | 0 |
| preparation (1 spend, 15 outputs) | 16 Orchard | 0 | 0 |
| transfer | 2 Orchard + 1 Ironwood | 1 + 1 | 0 |

This is reachable: the engine's `Signing::External` path hands exactly these PCZTs to an outside signer. A Signer matching by derivation path skips the unstamped spends as "not ours", nothing else can sign them, and extraction fails with a missing spend-auth signature.

### 2. `redact_pczt_for_batch_signer`

It clears `spend_auth_sig` on every action. A protocol padding dummy reaches it already signed — so under the new predicate it correctly carries no derivation — with its `dummy_sk` already cleared by the compact signer view it builds on. Clearing its signature too leaves an action with no signature, no dummy key, no derivation, and (as an already-authorized action) no randomizer: nothing the batch Signer or the wallet can ever authorize.

The signability invariant added to `pczt_single_step` alongside the zcash/librustzcash#2777 fix cannot catch this, as it runs pre-redaction.

---

nuttycom:
Closes #2782. Addresses the two gaps raised in review of #2778 (whose fix is on `main`).

## 1. `redact_pczt_for_batch_signer` retains existing signatures

It cleared `spend_auth_sig` on every action. A protocol padding dummy reaches it already signed — so under the new derivation predicate it correctly carries no derivation — with its `dummy_sk` already cleared by the compact signer view it builds on. Clearing its signature too left an action with no signature, no dummy key, no derivation, and (as an already-authorized action) no randomizer: nothing the batch Signer or the wallet could ever authorize.

Only already-authorized actions carry a signature at that point, and the function already snapshots their indices, so the fix is to drop the blanket `clear_spend_auth_sig`. Every action in the returned view is now either already authorized or still authorizable.

**Consequence worth noting:** the batch Signer's response now carries the retained padding-dummy signatures back alongside the new ones. Re-applying an identical signature to the authoritative PCZT is a no-op, which the pool test exercises end to end — the flow already applied every returned signature and unwrapped. The test's expected signature positions became "every action in both bundles," which directly encodes the invariant: after the batch round, nothing is left unauthorized.

The signability invariant added to `pczt_single_step` alongside the #2778 fix runs pre-redaction and could not catch this, so the batch-view assertions in `create_pczt_supports_ironwood_output` gain the post-redaction form.

## 2. `zcash_pool_migration`'s builders stamp derivation

`build_prep_tx` and `build_transfer_pczt` do not go through `create_pczt_from_proposal`, so the #2778 fix did not reach them; see #2782 for the measured shapes. This is reachable via the engine's `Signing::External` path.

Stamping goes in `finalize_pczt`, which both builders funnel through, using the same predicate as `create_pczt_from_proposal`: after IO finalization, set the derivation on every action whose spend still lacks `spend_auth_sig`, in both the Orchard and Ironwood bundles. That covers the real spends and the wallet-controlled zero-value spends while excluding the pre-signed padding dummies.

The derivation crosses the API as a new `build::AccountDerivation`, holding only the seed fingerprint and account index; the builders compose `m/32'/coin_type'/account'` from their own consensus parameters, so a caller cannot address a migration to the wrong coin type. The migration crate keeps its own type rather than taking `zcash_client_backend`'s `Zip32Derivation`, which would pull the wallet layer into the backend-agnostic core that `MigrationCrypto` exists to keep at arm's length; a `From` conversion behind the `wallet` feature bridges the two.

`MigrationCrypto` gains a required `account_derivation` method, threaded to all three build sites. It is required rather than defaulting to `Ok(None)` so an implementor cannot silently reintroduce the bug.

`WalletMigration`'s `MigrationCrypto` impl now requires `W: WalletRead<AccountId = <W as InputSource>::AccountId>`, since reading the account record needs the two ids to agree — the same bound form already used in `zcash_client_backend::data_api::wallet`. `zcash_client_sqlite` satisfies it.

## Tests

TDD throughout. Disabling the stamping turns all five new/extended assertions red: `stamps_derivation_*` in both builders, the end-to-end pipeline test, `rebuild_expired_transfer_unsigned_awaits_the_external_signature`, and the new `externally_signed_migrations_identify_every_spend_to_the_signer`, which covers the whole commit path (preparation layers and transfers alike) and its converse under a derivation-less backend.

`cargo test` clean for `zcash_client_backend`, `zcash_pool_migration`, `zcash_pool_migration_memory`, and `zcash_client_sqlite` (460 passed); `cargo clippy --workspace --all-features --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean; `git rebase --exec 'cargo check --workspace --all-features --all-targets'` confirms each commit builds on its own.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7349,6 +7349,7 @@ dependencies = [
  "zcash_pool_migration_memory",
  "zcash_primitives",
  "zcash_protocol",
+ "zip32",
 ]
 
 [[package]]
@@ -7362,6 +7363,7 @@ dependencies = [
  "rand_core 0.6.4",
  "zcash_pool_migration",
  "zcash_protocol",
+ "zip32",
 ]
 
 [[package]]
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -25,6 +25,17 @@ workspace.
   `sync::Error`, `sync::decryptor::TryQueueError`, `tor::Error`,
   `tor::grpc::GrpcError`, and `tor::http::HttpError`.
 
+### Fixed
+- `zcash_client_backend::data_api::wallet::redact_pczt_for_batch_signer` no
+  longer removes existing spend authorization signatures. It previously
+  stripped the signatures of the pre-signed protocol padding dummies while
+  also removing their randomizers and (via the compact signer view) their
+  dummy signing keys, leaving actions that neither the batch Signer nor the
+  wallet could authorize. Every action in the returned view is now either
+  already authorized or still authorizable. A batch Signer's response
+  consequently carries the retained signatures back alongside the new ones;
+  re-applying them to the authoritative PCZT is a no-op.
+
 ## [0.24.0-rc.4] - 2026-07-26
 
 ### Changed
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -9657,10 +9657,20 @@ where
         assert_eq!(batch_bundle.len(), original_bundle.len());
         for (original, batch) in original_bundle.iter().zip(batch_bundle) {
             assert!(!batch.has_fvk);
-            assert!(!batch.has_signature);
+            // A pre-signed protocol padding dummy keeps its signature: the batch view has
+            // already dropped its `dummy_sk` and it carries no derivation, so nothing
+            // downstream could ever reproduce that signature.
+            assert_eq!(batch.has_signature, original.has_signature);
             assert_eq!(
                 batch.has_alpha,
                 original.has_alpha && !original.has_signature
+            );
+            // The invariant the redaction must preserve: every action is either already
+            // authorized or still authorizable. Neither the Signer nor the wallet can
+            // rescue an action that is left with no signature and no randomizer.
+            assert!(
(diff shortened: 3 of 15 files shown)


schell (approved):
LGTM. Nothing worth holding this back. Claude had some nits, but they seemed like noise to me. 👍

Issue #2785, opened 2026-07-27
Reduce SDK<->librustzcash duplication: move shared FFI logic upstream


dannywillems:
Over the past few months, AI coding agents working in parallel on both mobile
SDKs have replicated a large amount of Rust logic that would be better shared or
moved upstream into librustzcash, and this issue tracks consolidating it.

> Note: this audit and issue were authored by an AI agent (Claude); the
> file:line references and recommendations should be reviewed before acting.

## Summary

The Android (`zcash-android-wallet-sdk`) and Swift (`zcash-swift-wallet-sdk`)
SDK Rust FFI layers contain a large amount of logic that is duplicated between
the two SDKs, and in some cases reimplements logic that already exists here in
librustzcash. This is a tracking issue for moving the shared, host-neutral logic
upstream so both SDKs (and zallet) consume one maintained implementation.

This is a maintainability / tech-debt epic, not a bug report. The SDKs generally
reuse librustzcash correctly (ZIP-321, the `eip681` crate, `init_wallet_db`
migrations, Tor bootstrap/HTTP/TLS, note selection are all consumed, not
reimplemented). The problem is *parallel duplication*: Android (JNI) and Swift
(C-ABI) each hand-write near-identical Rust wrapping the same upstream calls, and
the copies have begun to drift.

Audit basis: android & swift `maint/v2.7.x`, librustzcash `main` (`6d0cae86f5`).
File:line references below are from those branches.

## A. Genuine logic missing upstream, copied in both SDKs -> move upstream

- [ ] **A1** `propose_orchard_to_ironwood` pool-migration builder -- android `migration.rs:66-152`, swift `migration.rs:40-99`. Home: `zcash_pool_migration` / the `zcash_ironwood_migration_backend` effort already in flight.
- [ ] **A2** ZIP-325 account + private-use metadata keys -- swift `derivation.rs:577-737`, android `lib.rs:2900-3018`. Home: `zcash_keys` (e.g. an `AccountMetadataKey` type). Security-sensitive; must stay bit-identical.
- [ ] **A3** ZIP-32 arbitrary wallet/account keys -- swift `derivation.rs:763-826`, android `lib.rs:3020-3075`. Home: `zcash_keys`.
- [ ] **A4** PCZT multi-pool proving orchestration (Orchard->Ironwood->Sapling) -- android `lib.rs:2573-2653`, swift `lib.rs:2975-3075`. Home: `pczt::Prover`.
[...]

PR #2789, opened 2026-07-27
zcash_client_backend: add standard ZIP 317 split change strategy

dannywillems:
Part of #2785 (item **A6**).

The Android and Swift wallet SDKs each hand-write an identical `zip317_helper`
that builds a ZIP 317 multi-output change strategy splitting change into at
least 4 notes of value at least 0.1 ZEC. The two copies have already drifted
(`.expect("4 is nonzero")` on Android vs `.unwrap()` on Swift), and the `4` /
`0.1 ZEC` constants are duplicated magic numbers.

This moves the shared logic upstream into `zcash_client_backend::fees` so both
SDKs (and zallet) consume one maintained implementation:

- `SplitPolicy::standard()`, plus the `SplitPolicy::STANDARD_OUTPUT_COUNT` (4)
  and `SplitPolicy::STANDARD_MIN_OUTPUT_VALUE` (0.1 ZEC, expressed as
  `COIN / 10`) constants.
- `MultiOutputChangeStrategy::<StandardFeeRule, I>::standard(change_memo)`,
  capturing the whole helper's change strategy: ZIP 317 fee rule, Orchard
  change fallback, default dust policy, and `SplitPolicy::standard()`. It is
  reachable through the existing `fees::standard::MultiOutputChangeStrategy<I>`
  type alias.

The trivial `GreedyInputSelector::new()` half of the SDK helper is intentionally
left in the SDKs (input selection is not a fee concern); each helper collapses
to `(MultiOutputChangeStrategy::standard(memo), GreedyInputSelector::new())`.

A new unit test pins the `4` and `10_000_000`-zatoshi values so any future
change to the wallet default is deliberate rather than silent.

Verified: builds with default features and with `orchard,transparent-inputs`;
`fees` tests pass; `cargo fmt --check` and `cargo clippy` clean.

### DIFF
--- a/zcash_client_backend/src/fees/zip317.rs
+++ b/zcash_client_backend/src/fees/zip317.rs
@@ -296,6 +296,26 @@ impl<R, I> MultiOutputChangeStrategy<R, I> {
     }
 }
 
+impl<I> MultiOutputChangeStrategy<StandardFeeRule, I> {
+    /// Constructs the standard general-purpose multi-output change strategy.
+    ///
+    /// This is the change strategy used by the mobile wallet SDKs. It applies the ZIP 317 fee
+    /// rule, falls back to the Orchard pool for change when the transaction has no shielded
+    /// inputs, uses the default [`DustOutputPolicy`], and splits change according to
+    /// [`SplitPolicy::standard`]. It is equivalent to a
+    /// [`crate::fees::standard::MultiOutputChangeStrategy`] pre-configured with the wallet
+    /// defaults; use [`MultiOutputChangeStrategy::new`] to customize any of them.
+    pub fn standard(change_memo: Option<MemoBytes>) -> Self {
+        Self::new(
+            StandardFeeRule::Zip317,
+            change_memo,
+            ShieldedPool::Orchard,
+            DustOutputPolicy::default(),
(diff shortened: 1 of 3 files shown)


zcash_client_backend/src/fees/zip317.rs
```diff
@@ -296,6 +296,26 @@ impl<R, I> MultiOutputChangeStrategy<R, I> {
...
+    /// rule, falls back to the Orchard pool for change when the transaction has no shielded
+    /// inputs, uses the default [`DustOutputPolicy`], and splits change according to
+    /// [`SplitPolicy::standard`]. It is equivalent to a
+    /// [`crate::fees::standard::MultiOutputChangeStrategy`] pre-configured with the wallet
+    /// defaults; use [`MultiOutputChangeStrategy::new`] to customize any of them.
+    pub fn standard(change_memo: Option<MemoBytes>) -> Self {
+        Self::new(
+            StandardFeeRule::Zip317,
+            change_memo,
+            ShieldedPool::Orchard,
```
dannywillems, line 312: This would need to be changed to Ironwood by default. It is already handled in the code to redirect to Ironwood after NU6.3 activation.

PR #2792, opened 2026-07-27
zcash_client_backend: apply timeouts to Tor network operations

### LINKED ISSUE #2791: Tor transports apply no timeouts to network operations
Neither the HTTP-over-Tor client (`zcash_client_backend::tor::http`) nor the
gRPC-over-Tor connector (`zcash_client_backend::tor::grpc`) bounds how long a
network operation may take.

`arti` already bounds opening the Tor stream itself, via
`StreamTimeoutConfig::connect_timeout` (10s by default), so `tor_client.connect()`
is covered. The gap is in the layers above it:

- **TLS handshake** — `TlsConnector::connect()` in `one_http_request` has no deadline.
  A server that reads the ClientHello and then goes silent holds the future forever.
- **HTTP exchange** — `conn::http1::handshake` and `sender.send_request()` have no
  deadline, so a server that accepts the request and never sends response headers
  hangs indefinitely.
- **Response body** — `parse_response(body).await` is unbounded, so a server that
  trickles a body forever is never cut off. `http_get_json` collects the whole body.
- **gRPC** — `Endpoint::from(endpoint)` sets neither `connect_timeout` nor `timeout`,
  and no HTTP/2 keep-alive, so both connection setup and mid-stream stalls are
  unbounded.

This matters more over Tor than on the clearnet: a half-open circuit can persist for
a long time without the peer's TCP stack delivering a reset, so the OS-level
detection an application developer might expect doesn't fire.

`Client::get_latest_zec_to_usd_rate` aggregates several exchanges with `join!` and
`join_all`, so a single unresponsive exchange is enough to hang the whole rate query.

Fix should apply per-attempt deadlines at each of these layers, with the values
configurable rather than hard-coded, since appropriate durations depend on expected
circuit latency. For gRPC the request deadline should not cap streaming methods such
as `GetBlockRange`; HTTP/2 keep-alive is the right mechanism for detecting a peer that
stalls partway through a stream.

---

nuttycom:
Closes #2791.

Neither the HTTP-over-Tor client nor the gRPC-over-Tor connector bounded how long a
network operation could take. A server that accepted a connection and then went
silent — never completing the TLS handshake, never sending response headers, or
trickling a response body forever — would leave the corresponding future pending
indefinitely.

`arti` already bounds opening the Tor stream itself via
`StreamTimeoutConfig::connect_timeout` (10s by default), so the exposure was in the
layers above: the TLS handshake, the HTTP exchange, and the response body.

## Changes

- Adds `tor::Timeouts`, carried by `tor::Client` and propagated through
  `isolated_client`. `Client::create` keeps its signature and uses the defaults;
  the new `Client::create_with_timeouts` lets callers tune them.

  | Knob | Default | Bounds |
  |---|---|---|
  | `connect` | 30s | Tor stream + TLS handshake, both transports |
  | `request` | 60s | Sending the request through receiving response headers |
  | `response_body` | 60s | Reading a complete HTTP response body (HTTP only) |
  | `grpc_keepalive` | 30s / 20s | HTTP/2 ping interval + pong deadline (gRPC only) |

- HTTP: the Tor stream and TLS handshake are bounded together, so `connect` means the
  same thing here as it does for gRPC (where `tonic` applies it to the connector as a
  whole). Deadlines are enforced per attempt and surface as the new
  `HttpError::Timeout(TimeoutPhase)`, which `retry_filter` sees like any other error.

- gRPC: `Endpoint::connect_timeout`, `Endpoint::timeout`, and HTTP/2 keep-alive.

## Notes for review

The `Timeouts` fields are private with `with_*` setters rather than public, so future
knobs aren't breaking changes.

`Endpoint::timeout` bounds only the wait for response *headers*, not the response
itself, so streaming methods such as `GetBlockRange` are not capped by it. Those are
covered instead by the HTTP/2 keep-alive pings, which detect a peer that stalls partway
through a stream — the one failure mode no request deadline can catch without also
capping legitimate long streams. This piece goes slightly beyond what #2791 describes,
but it's the case that matters most for a syncing wallet.

The 30/60/60 defaults are reasoned for typical Tor circuit latency with headroom rather
than derived from measurements; if there are real numbers from downstream wallets they
should probably override these.

Adds a hermetic regression test (`silent_server_times_out`) that drives
`make_http_request` over a duplex stream whose far end never responds. No network or
Tor client needed, so it runs in normal CI.

---

### DIFF
--- a/zcash_client_backend/src/tor/grpc.rs
+++ b/zcash_client_backend/src/tor/grpc.rs
@@ -22,6 +22,14 @@ impl Client {
     /// Tor hidden services (`.onion` addresses). The caller is responsible for deciding
     /// whether onion connections are appropriate for the given endpoint; this crate
     /// does not infer that from the endpoint host.
+    ///
+    /// The returned client applies this `Client`'s [`Timeouts`] to every request it makes.
+    /// Note that the request deadline bounds the wait for a response's headers, not the
+    /// duration of the response itself, so long-running streaming methods such as
+    /// `GetBlockRange` are not capped by it; a peer that stalls partway through a stream
+    /// is instead detected by the HTTP/2 keep-alive.
+    ///
+    /// [`Timeouts`]: super::Timeouts
     pub async fn connect_to_lightwalletd(
         &self,
         endpoint: Uri,
@@ -37,7 +45,12 @@ impl Client {
             HttpTcpConnector::new(self.clone())
         };
 
-        let channel = Endpoint::from(endpoint);
+        let channel = Endpoint::from(endpoint)
+            .connect_timeout(self.timeouts.connect)
+            .timeout(self.timeouts.request)
+            .http2_keep_alive_interval(self.timeouts.grpc_keepalive_interval)
+            .keep_alive_timeout(self.timeouts.grpc_keepalive_timeout)
+            .keep_alive_while_idle(false);
         let channel = if is_https {
             channel
                 .tls_config(ClientTlsConfig::new().with_webpki_roots())
--- a/zcash_client_backend/src/tor/http.rs
+++ b/zcash_client_backend/src/tor/http.rs
@@ -1,6 +1,6 @@
 //! HTTP requests over Tor.
 
-use std::{fmt, future::Future, io, sync::Arc};
+use std::{fmt, future::Future, io, sync::Arc, time::Duration};
 
 use arti_client::TorClient;
 use futures_util::task::SpawnExt;
@@ -21,7 +21,7 @@ use tokio_rustls::{
 use tor_rtcompat::PreferredRuntime;
 use tracing::{debug, error};
 
-use super::{Client, Error};
+use super::{Client, Error, Timeouts};
 
 pub mod cryptex;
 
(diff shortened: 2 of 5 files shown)


zcash_client_backend/src/tor/grpc.rs
```diff
@@ -37,7 +45,12 @@ impl Client {
             HttpTcpConnector::new(self.clone())
         };
 
-        let channel = Endpoint::from(endpoint);
+        let channel = Endpoint::from(endpoint)
+            .connect_timeout(self.timeouts.connect)
+            .timeout(self.timeouts.request)
+            .http2_keep_alive_interval(self.timeouts.grpc_keepalive_interval)
+            .keep_alive_timeout(self.timeouts.grpc_keepalive_timeout)
+            .keep_alive_while_idle(false);
```
nullcopy, line 53: Might improve performance to set this true

nullcopy (dismissed):
utACK

zcash_client_backend/src/tor/http.rs
```diff
@@ -258,30 +322,52 @@ where
...
-            .await
-            .map_err(HttpError::Tls)?;
-        make_http_request(stream, url, request, body).await
+        let stream = with_timeout(timeouts.connect, TimeoutPhase::Connect, async {
+            let stream = tor_client.connect((host.as_str(), port)).await?;
+
+            // On apple-darwin targets there's an issue with the native TLS implementation
+            // when used over Tor circuits. We use Rustls instead.
+            //
+            // https://gitlab.torproject.org/tpo/core/arti/-/issues/715
```
dannywillems, line 337: This seems to be fixed.

dannywillems (dismissed):
tACK, and see https://github.com/zcash/librustzcash/pull/2793.

dannywillems (approved):
utACK b463f5d0582de74f336feaac5e2505af02354904. re-ACK after rebase

PR #2793, opened 2026-07-27
zcash_client_backend: test Tor HTTP timeouts against an adversarial server

dannywillems:
Stacked on top of #2792 (`fix/tor_timeouts`); please review/merge that first.

## What

Adds unit and property tests that drive the HTTP-over-Tor timeout machinery introduced in #2792 against a fake, adversarial server, with no real Tor client and no sockets.

The HTTP transport is generic over the byte stream, so an in-memory `tokio::io::duplex` pair stands in for a Tor `DataStream`: one end is handed to the client code, the other is driven as the server.

## Why the duplex is the right adversary (the Tor-specific bit)

Every stalling scenario **holds the peer end open and silent**. That is the failure mode this branch exists to survive: over Tor a stalled peer delivers neither more bytes nor a TCP reset, and a half-open circuit persists, so the future hangs forever without a deadline. Dropping the peer instead would deliver an EOF, which is the clearnet case, not the one under test.

## Coverage

- `Request` timeout: server reads the request, then goes silent (no response head).
- `ResponseBody` timeout: server sends headers promising N body bytes, sends fewer, then stalls. Asserts the phase boundary is the arrival of the response head (a `ResponseBody`, not a `Request`, timeout).
- `Connect` timeout: to test this without a `TorClient`, the TLS handshake is split out of `one_http_request` into a `tls_handshake` helper, exercised against a peer that never sends its `ServerHello`.
- Well-behaved control: a responsive server must not be timed out.
- A proptest over adversarial and well-behaved servers asserting the invariant: the phase named by the resulting `Timeout` error is exactly the phase that stalled, else the body is returned.

Deadlines fire in virtual time under tokio's `start_paused`, so the suite is deterministic and adds no real delay (needs tokio's `macros` + `test-util` dev features).

## Out of scope (needs a real `TorClient`, marked in comments, not faked)

- Opening the Tor stream (the other half of the `Connect` phase; `arti` bounds that itself via `StreamTimeoutConfig::connect_timeout`).
- The circuit-isolating retry loop in `http_request`.
- The gRPC transport (timeouts delegated to `tonic`; a fake stalling `tower::Service<Uri>` connector is a possible follow-up).

Tests-only plus a private refactor and a dev-dependency feature bump, so no CHANGELOG entry.

### DIFF
--- a/zcash_client_backend/src/tor/grpc.rs
+++ b/zcash_client_backend/src/tor/grpc.rs
@@ -163,3 +163,64 @@ impl From<tonic::transport::Error> for GrpcError {
         GrpcError::Tonic(e)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::{
+        future::{Future, pending},
+        pin::Pin,
+        task::{Context, Poll},
+        time::Duration,
+    };
+
+    use hyper_util::rt::TokioIo;
+    use tokio::io::DuplexStream;
+    use tonic::transport::{Endpoint, Uri};
+    use tower::Service;
+
+    /// A connector that accepts a connection request and then never completes it, modelling
+    /// a Tor circuit that hangs while opening. Unlike the HTTP transport, the gRPC deadlines
+    /// live in `tonic`'s `Endpoint`, so this checks that the `connect_timeout` we configure
+    /// there actually bounds a connector that never returns.
+    #[derive(Clone)]
+    struct StallingConnector;
+
+    impl Service<Uri> for StallingConnector {
+        // The connector never returns a stream, so this type is only a phantom to satisfy
+        // tonic's bounds; `TokioIo<DuplexStream>` is a convenient one that is `Read + Write`.
+        type Response = TokioIo<DuplexStream>;
+        type Error = std::io::Error;
+        type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+        fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+            Poll::Ready(Ok(()))
+        }
+
+        fn call(&mut self, _: Uri) -> Self::Future {
+            Box::pin(pending::<Result<Self::Response, Self::Error>>())
+        }
+    }
+
+    /// The gRPC `connect_timeout` must bound a connector that never completes, so a Tor
+    /// circuit that hangs on open cannot leave `connect_to_lightwalletd` pending forever.
+    /// The outer timeout is a safety net that must not be the one to fire; if it does, the
+    /// `connect_timeout` we set on the `Endpoint` is not being applied.
+    #[tokio::test(start_paused = true)]
+    async fn connect_timeout_bounds_a_stalling_connector() {
+        let endpoint = Endpoint::from_static("http://198.51.100.1:8232")
+            .connect_timeout(Duration::from_secs(5))
+            .timeout(Duration::from_secs(60));
+
+        let connect = endpoint.connect_with_connector(StallingConnector);
+        let res = tokio::time::timeout(Duration::from_secs(30), connect).await;
+
+        assert!(
+            res.is_ok(),
+            "connect_timeout did not bound the stalling connector; the 30s safety net fired",
+        );
+        assert!(
+            res.unwrap().is_err(),
+            "expected the connect to fail once connect_timeout elapsed",
+        );
+    }
+}
--- a/zcash_client_backend/src/tor/http.rs
+++ b/zcash_client_backend/src/tor/http.rs
@@ -543,7 +543,7 @@ mod tests {
     use std::{future::pending, time::Duration};
 
     use http_body_util::{BodyExt, Empty};
-    use hyper::{Uri, body::Bytes};
+    use hyper::{StatusCode, Uri, body::Bytes};
     use proptest::prelude::*;
     use tokio::io::{AsyncReadExt, AsyncWriteExt, DuplexStream};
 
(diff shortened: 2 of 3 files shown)


zcash_client_backend/src/tor/http.rs
```diff
@@ -496,36 +512,145 @@ impl From<futures_util::task::SpawnError> for HttpError {
...
+        pending::<()>().await;
+    }
 
-        let res = with_timeout(
-            Duration::from_millis(100),
+    /// Consumes bytes from `peer` up to the end of the HTTP request head (`\r\n\r\n`). The
+    /// GET requests issued here carry no body, so this reads the whole request.
+    async fn read_request_head(peer: &mut DuplexStream) {
+        let mut buf = Vec::new();
+        let mut chunk = [0u8; 256];
```
dannywillems: Nit: have this magic number call CHUNK_SIZE. Even if it is in tests, magic numbers are evil.

zcash_client_backend/src/tor/http.rs
```diff
@@ -496,36 +512,145 @@ impl From<futures_util::task::SpawnError> for HttpError {
...
+    // The absolute values are irrelevant under `start_paused`; they only need to be
+    // positive so the deadline exists to be advanced to.
+    const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+    const RESPONSE_BODY_TIMEOUT: Duration = Duration::from_secs(5);
+    const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+
+    /// How the fake server behaves once a client has connected. Each stalling variant
+    /// models a half-open Tor circuit: the peer accepted the stream, but no further bytes
+    /// and no close will ever arrive.
+    enum Behavior {
```
dannywillems, line 613: We should have more adversarial cases in follow-ups. I'll think about a few (or spawn an agent...).

zcash_client_backend/src/tor/http.rs
```diff
@@ -496,36 +512,145 @@ impl From<futures_util::task::SpawnError> for HttpError {
...
 
-        let res = with_timeout(
-            Duration::from_millis(100),
+    /// Consumes bytes from `peer` up to the end of the HTTP request head (`\r\n\r\n`). The
+    /// GET requests issued here carry no body, so this reads the whole request.
+    async fn read_request_head(peer: &mut DuplexStream) {
+        let mut buf = Vec::new();
+        let mut chunk = [0u8; 256];
+        loop {
+            let n = peer.read(&mut chunk).await.expect("read from client");
```
dannywillems, line 650: Peers only send by 256 bytes?

zcash_client_backend/src/tor/http.rs
```diff
@@ -496,36 +512,145 @@ impl From<futures_util::task::SpawnError> for HttpError {
...
+    /// How the fake server behaves once a client has connected. Each stalling variant
+    /// models a half-open Tor circuit: the peer accepted the stream, but no further bytes
+    /// and no close will ever arrive.
+    enum Behavior {
+        /// Read the request, then go silent without sending any response.
+        StallBeforeHead,
+        /// Send response headers promising `content_length` body bytes, send `sent` of
+        /// them (`sent < content_length`), then go silent. Models a circuit that trickles
+        /// and then half-opens partway through the body.
+        StallDuringBody { sent: usize, content_length: usize },
```
dannywillems: `sent` is an adjective that can be used for many things. Renaming into something related to the length makes more sense, and is semantically better for the reader.

zcash_client_backend/src/tor/http.rs
```diff
@@ -496,36 +512,145 @@ impl From<futures_util::task::SpawnError> for HttpError {
...
+    /// Serves one connection according to `behavior`.
+    async fn serve(mut peer: DuplexStream, behavior: Behavior) {
+        read_request_head(&mut peer).await;
+        match behavior {
+            Behavior::StallBeforeHead => hold_open(peer).await,
+            Behavior::StallDuringBody {
+                sent,
+                content_length,
+            } => {
+                let head = format!("HTTP/1.1 200 OK\r\nContent-Length: {content_length}\r\n\r\n");
```
dannywillems: The header should be more typed, so we can handle more cases (more HTTP versions, more HTTP response types, etc).

zcash_client_backend/src/tor/http.rs
```diff
@@ -496,36 +512,145 @@ impl From<futures_util::task::SpawnError> for HttpError {
...
+                peer.write_all(head.as_bytes()).await.expect("write head");
+                peer.write_all(&vec![b'a'; sent])
+                    .await
+                    .expect("write partial body");
+                peer.flush().await.expect("flush");
+                hold_open(peer).await;
+            }
+            Behavior::Complete { body } => {
+                let head = format!("HTTP/1.1 200 OK\r\nContent-Length: {}\r\n\r\n", body.len());
```
dannywillems: See https://github.com/zcash/librustzcash/pull/2793/changes#r3657565157

zcash_client_backend/src/tor/http.rs
```diff
@@ -534,6 +659,142 @@ mod tests {
...
+    async fn stall_during_body_times_out_as_response_body() {
+        let res = run_get(Behavior::StallDuringBody {
+            sent: 8,
+            content_length: 1024,
+        })
+        .await;
+        assert!(
+            matches!(
+                res,
+                Err(Error::Http(HttpError::Timeout(TimeoutPhase::ResponseBody)))
```
dannywillems: Outch, starts to be verbose 😄

zcash_client_backend/src/tor/http.rs
```diff
@@ -534,6 +659,142 @@ mod tests {
...
+            "expected the body back, got {res:?}",
+        );
+    }
+
+    /// A peer that accepts the stream but never completes the TLS handshake must surface
+    /// as a `Connect` timeout. Models a half-open circuit mid-handshake: the `ServerHello`
+    /// never arrives.
+    #[tokio::test(start_paused = true)]
+    async fn tls_handshake_stall_times_out_as_connect() {
+        let (client, peer) = tokio::io::duplex(16 * 1024);
```
dannywillems: Again magic numbers here.

zcash_client_backend/src/tor/http.rs
dannywillems: Done: named `CHUNK_SIZE` (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: No, they can send however they like: the reader loops and accumulates across reads, so the buffer only bounds a single read, not what the peer may send. Renamed to `CHUNK_SIZE` (bumped to 1024) with a comment saying exactly that (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: Renamed `sent` -> `sent_len` (parallel to `content_length`) across `Behavior`, `Scenario`, and the strategy (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: Done: the response head is now built via a small typed `ResponseHead` (hyper `StatusCode`), so status/version/length are fields rather than a hand-written format string, leaving room for more cases (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: Same fix as the typed `ResponseHead` above; both `serve` arms go through it now (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: Trimmed: added a `timeout_phase()` extractor so the asserts read `assert_eq!(timeout_phase(&res), Some(TimeoutPhase::_))` (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: Named `DUPLEX_BUF_SIZE`; both duplex call sites use it now (06f4644).

zcash_client_backend/src/tor/http.rs
dannywillems: Agreed. Keeping this as a follow-up rather than growing this PR: candidates are a gRPC stalling `tower::Service<Uri>` connector, a slow-trickle head, a malformed/partial head, and a Content-Length/body-length mismatch. Resolving here; the follow-up lives outside this PR.

dannywillems:
I'll rebase when the branch `fix/tor_timeouts` has been merged.

zcash_client_backend/src/tor/grpc.rs
```diff
@@ -163,3 +163,64 @@ impl From<tonic::transport::Error> for GrpcError {
...
+        }
+    }
+
+    /// The gRPC `connect_timeout` must bound a connector that never completes, so a Tor
+    /// circuit that hangs on open cannot leave `connect_to_lightwalletd` pending forever.
+    /// The outer timeout is a safety net that must not be the one to fire; if it does, the
+    /// `connect_timeout` we set on the `Endpoint` is not being applied.
+    #[tokio::test(start_paused = true)]
+    async fn connect_timeout_bounds_a_stalling_connector() {
+        let endpoint = Endpoint::from_static("http://198.51.100.1:8232")
```
dannywillems, line 210: Nit: we could use a generator that respects the RFC 1123 instead of a fixed IP.

PR #2795, opened 2026-07-27
zcash_client_backend: reject non-HTTP(S) URLs in the Tor transport

dannywillems:
## What

`tor::http::url_is_https` rejected only a URL with a *missing* scheme. A URL with an explicit scheme other than `http`/`https` (`ftp`, `ws`, `file`, ...) was accepted and its `is_https` treated as `false`, so `http_get` / `http_post` / `connect_to_lightwalletd` would issue a **plaintext HTTP request** to such an endpoint instead of refusing it.

That contradicts the `HttpError::NonHttpUrl` contract, which documents "Only HTTP or HTTPS URLs are supported."

## Fix

Reject any scheme that is not `http` or `https`, alongside the missing-scheme case that was already rejected. Adds unit tests for default-port inference (443/80), an explicit port, and the rejected schemes.

## Notes

- Found while writing adversarial tests for the Tor timeout work in #2793; this is the independent, `main`-targeted fix for that finding.
- Behavior change surfaced as a `### Fixed` CHANGELOG entry (dedicated commit).

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -32,6 +32,11 @@ workspace.
   `tor::grpc::GrpcError`, and `tor::http::HttpError`.
 
 ### Fixed
+- The Tor HTTP and gRPC transports now reject a URL whose scheme is neither
+  `http` nor `https` (for example `ftp` or `ws`) with `tor::http::HttpError::NonHttpUrl`.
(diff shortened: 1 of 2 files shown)


nullcopy (approved):
utACK a95fdfc29ddb0fdfaa5e02fdd88cb85a6f3469ac

PR #2797, opened 2026-07-27
AGENTS.md: require disclosures be classified against the security policy

### LINKED ISSUE #2796: AGENTS.md gives agents no pointer to the security severity rubric
`AGENTS.md` says nothing about security vulnerability disclosures, so an agent
helping someone draft or file a report against this repository has no pointer to
the severity rubric it should be classifying against.

The gap is made worse by where the policy lives. There is no `SECURITY.md` in
this repository; GitHub's security policy tab resolves to the organization-level
`zcash/.github` repo. An agent instructed to "read SECURITY.md" therefore finds
nothing locally, and falls back on CVSS or a general intuition about impact —
neither of which matches the rubric the maintainers actually use to prioritize.

Two consequences show up in practice:

- Findings are filed at the wrong severity. The policy explicitly warns that
  overstating severity may make a reporter ineligible for bounty compensation,
  which is a poor outcome for a well-intentioned reporter who simply had no
  access to the rubric.
- Findings are filed that the policy classifies as "Not Vulnerability" at all —
  in particular, bugs whose worst effect is temporary unavailability of funds
  recoverable by a rescan, and misbehavior on data supplied by a light wallet
  server, which the light wallet threat model already treats as trusted.

`AGENTS.md` should require that the policy be read before a report is drafted,
filed, or assessed, and that the finding be classified against its rubric with
the matching category and bullet stated explicitly.

---

nuttycom:
Closes #2796.

Adds a `## Security Vulnerability Disclosures` stanza to `AGENTS.md`, placed with
the other policy gates rather than among the technical conventions.

It requires an agent to read the Zcash security policy in full before drafting,
filing, or assessing a vulnerability report, and to classify the finding against
that policy's rubric — naming both the category and the specific bullet it
matches.

## Notes for review

**Where the policy is read from.** There is no `SECURITY.md` in this repository;
the security tab resolves to the org-level `zcash/.github` repo. The stanza says
so explicitly and gives a command that works without web access:

```sh
gh api repos/zcash/.github/contents/SECURITY.md --jq '.content' | base64 -d
```

with the raw URL as a fallback. Without this, an agent told to "read SECURITY.md"
looks locally, finds nothing, and invents a rubric.

**Three points called out beyond the bare requirement**, because they are what
goes wrong in practice:

- `Not Vulnerability` is a real verdict, covering rescan-recoverable fund
  unavailability and behavior already conceded by the light wallet threat model.
  The latter matters here specifically: light wallet servers are explicitly
  trusted for the correctness of the data they return.
- Overstating severity has a stated bounty-ineligibility penalty, so a finding
  straddling two categories is assigned the lower one.
- Critical reports route to the Signal group named in the policy, not to a
  GitHub advisory. An agent that classifies something Critical and then opens a
  draft advisory has routed it wrongly.

**Slightly wider than severity alone**, the closing paragraph requires factual
claims be checked against the cited file and line and against the dependency
versions this workspace pins. A claim that is not true has no severity to
assign. Happy to drop that paragraph if it reads as scope creep.

No code changes; documentation only.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,51 @@ Example:
 Co-Authored-By: Claude <noreply@anthropic.com>
 ```
 
+## Security Vulnerability Disclosures
+
+Before helping a user draft, file, or assess a security vulnerability report against this
+repository, the agent MUST read the Zcash security policy in full. It is **not** checked
+into this repository — it is inherited from the organization-level `zcash/.github` repo,
+so read it from there:
+
+```sh
+gh api repos/zcash/.github/contents/SECURITY.md --jq '.content' | base64 -d
+```
+
+If `gh` is unavailable, fetch
+<https://raw.githubusercontent.com/zcash/.github/main/SECURITY.md>. The rendered version
+is linked from <https://github.com/zcash/librustzcash?tab=security-ov-file>.
+
+The agent MUST then classify the finding against the rubric in that policy — Critical,
+High, Moderate, Low, or **Not Vulnerability** — and MUST state, both in the report and to
+the user, which category it assigned and which specific bullet of the rubric the finding
+matches. A report that does not name its category and the reasoning for it is not ready
+to file.
+
+That rubric is not a general-purpose severity scale. Substituting CVSS, a vendor
+severity label, or an intuition about impact produces the wrong answer. In particular:
+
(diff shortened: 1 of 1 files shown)


nullcopy (approved):
ACK 87a54e92f0526dcc1ab80e9e602516218c11d476

dannywillems (commented):
Post-ACK

PR #2798, opened 2026-07-27
ci: add a no-default-features clippy job

dannywillems:
The existing `clippy` / `clippy-beta` jobs run with `--all-features`, so any
lint that only fires under a narrower feature set is masked. The most common
case is an import used only behind a feature gate: with `--all-features` it
looks used, but a `--no-default-features` build reports it as unused.

This adds a `clippy-nodefault` job that runs
`cargo clippy --no-default-features -- -D warnings` (checking the default
lib/bin targets, matching the reproduction below), and wires it into
`required-checks`.

Motivated by a review on #2742: `cargo clippy -p zcash_client_sqlite
--no-default-features --lib -- -D warnings` reported an unused `TxId` import
that the all-features clippy did not catch. Verified locally that this job is
clean on `main` and that it flags that class of issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -390,6 +390,25 @@ jobs:
           args: >
             -D warnings
 
+  clippy-nodefault:
+    name: Clippy (no default features)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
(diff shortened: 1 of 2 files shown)


dannywillems (commented):
See new job [CI / Clippy (no default features) (pull_request)](https://github.com/zcash/librustzcash/actions/runs/30281084034/job/90027338697?pr=2798)

nuttycom (approved):

Issue #2630, opened 2026-07-15
Orchard → Ironwood value-pool wallet migration engine (`zcash_pool_migration`)


nuttycom:
## Summary

Implement an engine that migrates a wallet's funds from the Orchard value pool to the Ironwood value pool (introduced in NU6.3 / v6 transactions). This issue tracks that work: the engine itself, the wallet-backend/PCZT primitives it builds on, and the remaining gaps and follow-ups.

The engine should be **synchronous and network-free**. Given a wallet SQLite database it should:

1. **Split** the spendable Orchard balance into notes sent to the wallet's internal viewing key in fixed denominations: powers of  `10^n * {1, 2, 5}` each carrying its crossing value plus a fee buffer, leaving dust/residual value below 0.01 ZEC as Orchard change.
2. **Schedule** each pool-crossing transaction as an individually-timed Orchard→Ironwood transfer, sampling inter-transfer gaps from an exponential distribution.
4. **Build and sign** every transfer as a PCZT (software-signing or external/Keystone-style signer), persisting all state in five additive `ext_ironwood_migration_*` tables.
5. **Provide functionality for anchor update**  Before a transaction is sent, its proofs must be updated to use the anchor at a  recent "anchor block"  at height $h$ where $h \in H,\quad H = \lbrace n \in \mathbb{Z} \mid n \bmod 144 = 0 \rbrace$
6.  Provide the platform pre-signed transactions to **broadcast** and report results back, modeled as a state machine; relevant states are (`NotStarted` / `SplitPendingConfirmation` / `ReadyToPropose` / `InProgress` / `Complete` / `RequiresAttention`). This is the wallet-side realization of the single-confirmation Orchard→Ironwood migration flow (zcash/zips#1317), and consumes the pre-signed v6 transaction-chain primitives tracked in the Epic #2529.

## What this PR must deliver

- A set of traits that may be implemented by wallet backends to provide persistence and data retrieval for operations related to the migration. This should include APIs for retrieving metadata about the Orchard notes currently held by the wallet and persisting and retrieving migration PCZTs and other migration state information.
- Self-funding `power-of-ten * {1,2,5}` denomination planning (`split`, `denominations`).
- Height-based transfer scheduling.
- Both software-signing and external-signer (staged proven-but-unsigned PCZT) flows.
- Seeded-wallet end-to-end tests

## Enabling work already merged

[...]

PR #2801, opened 2026-07-27
zcash_pool_migration: minimal sync/proving wake-up schedule

### LINKED ISSUE #2800: zcash_pool_migration: minimal sync/proving wake-up schedule for background-constrained wallets
Part of #2630.

Wallet applications need to optimize their use of background-processing time; on older iOS versions in particular, background scheduling options are very limited. The migration engine already produces the schedule of transfer **transmission** heights, but offers no easily-consumable schedule of the **sync and proving** work that must precede each transmission.

Each pool-crossing transfer's broadcast height is known at scheduling time, and its ZIP 318 anchor boundary is drawn then as well, so every transfer has a predetermined proving window: after its bucketed anchor height settles and strictly before its broadcast height. When a wallet wakes to sync and prove, it should prove every transfer whose anchor height is in the past.

`zcash_pool_migration` should expose APIs that compute, from a committed migration, the **minimum** set of wake-ups covering every pending transfer — each wake-up scheduled ~10 blocks after a bucketed anchor height, with jitter to prevent thundering-herd load on the light wallet servers.

---

nuttycom:
Closes #2800. Part of the #2630 epic.

## Summary

- Adds `scheduling::schedule_sync_wakeups`: computes the **provably minimal** set of sync/proving wake-up heights implied by a committed transfer schedule. Each pending transfer's drawn ZIP 318 anchor boundary and broadcast height bound a proving window `[anchor + settle_margin, broadcast − 1]`; the fewest wake-ups piercing every window (the classic minimum piercing-set problem) are found by the optimal greedy over deadline-sorted windows. Each wake-up lands at a bucketed anchor height + 10 blocks, plus uniform jitter (bounded by both the configured cap and the group's slack, drawn from a `CryptoRng`) as a thundering-herd defense, so concurrent wallets don't converge on the light wallet servers in synchronized bursts. A mandatory immediate wake-up at the caller's current height collects overdue transfers and absorbs any window containing that height, preserving minimality.
- Adds `WakeupParams` (provisional defaults: 10-block settle margin, 12-block jitter cap), `SyncWakeup<T>` (wake height + the transfers it is responsible for proving), and `WakeupScheduleError<T>`.
- Adds `MigrationState::sync_wakeup_schedule`: the one-call consumer API over a committed migration's transfers that still need proofs (`Signed`/`AwaitingSignature`, unexpired), recomputed on demand from persisted state rather than persisted itself.

## Test plan

- Unit tests for window construction, greedy grouping, overdue absorption, and the degenerate-parameter clamps.
- Property tests: every transfer is covered exactly once inside its window, heights strictly increase and never fall in the past, and the wake-up count equals a brute-force piercing optimum (including overdue cases).
- Golden vectors pinned under seeded `ChaCha8Rng` with hand-verified derivation comments.
- `cargo test -p zcash_pool_migration --all-features`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo fmt --check` all green.

### DIFF
--- a/zcash_pool_migration/src/scheduling.rs
+++ b/zcash_pool_migration/src/scheduling.rs
@@ -37,6 +37,12 @@
 //!    validity as a pure function of the current height, so the expiry height itself carries no
 //!    per-wallet information.
 //!
+//! Beyond the ZIP 318 draws, [`schedule_sync_wakeups`] derives from a committed transfer schedule
+//! the MINIMAL set of sync/proving wake-ups a background-constrained wallet needs: every transfer
+//! is proved after its drawn anchor boundary settles and strictly before its broadcast height, and
+//! each wake-up lands a settle margin past a bucketed anchor height plus an anti-thundering-herd
+//! jitter ([`WakeupParams`]).
+//!
 //! # Cohorts
 //!
 //! Transfers (across all wallets) that prove against the same boundary anchor form a COHORT: to an
@@ -63,6 +69,7 @@
 //! [ZIP 318]: https://zips.z.cash/zip-0318
 
 use alloc::vec::Vec;
+use core::fmt;
 use core::num::NonZeroU32;
 
 use rand_core::{CryptoRng, RngCore};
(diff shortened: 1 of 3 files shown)


zcash_pool_migration/src/scheduling.rs
```diff
@@ -349,6 +356,107 @@ impl Default for SchedulingParams {
...
+
+impl WakeupParams {
+    /// The provisional defaults: a 10-block settle margin (about 12.5 minutes at the target block
+    /// spacing, so the synced boundary is one a reorg can no longer plausibly displace) and a
+    /// 72-block jitter cap (half a ZIP 318 anchor bucket, about 1.5 hours). Anchor boundaries sit
+    /// on a GLOBAL grid, so without jitter every migrating wallet would wake at the same heights
+    /// and converge on the light wallet servers together.
+    pub const DEFAULT: Self = Self {
+        settle_margin: 10,
+        jitter_cap: 72,
```
nullcopy: I wonder if jitter should be block denominated? It would be better to jitter some wall-time value, to avoid thundering herds on given blocks. And then you could arguably tighten the jitter window

zcash_pool_migration/src/scheduling.rs
```diff
@@ -532,6 +640,143 @@ pub fn expiry_height(current_height: BlockHeight) -> BlockHeight {
...
+/// (sync and broadcast never share a wake window), giving it a proving WINDOW of heights. The
+/// fewest wake-ups piercing every window is the classic minimum piercing-set problem, solved
+/// optimally by a greedy pass over the windows in deadline order; each wake-up lands at the
+/// latest window-opening height of the group it covers — under real parameters, a bucketed anchor
+/// height plus the settle margin — plus a uniform random jitter bounded by both
+/// [`WakeupParams::jitter_cap`] and the group's slack. The jitter is drawn from a [`CryptoRng`]
+/// because wake-up times are observables: anchor boundaries sit on a global grid, and un-jittered
+/// wake-ups would both stampede the light wallet servers and mark the wallet as migrating.
+///
+/// No wake-up is scheduled below `not_before` (the caller's current target height, or the commit
```
nullcopy: I don't like the name `not_before`, because it's used to determine if something is overdue... would rename to `current_tip` or something

zcash_pool_migration/src/scheduling.rs
```diff
@@ -349,6 +356,115 @@ impl Default for SchedulingParams {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct WakeupParams {
+    settle_margin: u32,
+    jitter_cap: u32,
+}
+
+impl WakeupParams {
+    /// The provisional defaults: a 10-block settle margin (about 12.5 minutes at the target block
+    /// spacing, so the synced boundary is one a reorg can no longer plausibly displace) and a
+    /// 24-block jitter cap (about 30 minutes).
```
nullcopy: Do we really need 30 minutes of jitter? I would think < 10 minutes is sufficient to avoid thundering herd. We already randomize the broadcast schedule, so we already don't expect many clients to wake at the same time.

nullcopy (dismissed):
utACK 05c02610a52bd4d7942daf1ee327cc18b010c175

I have my doubts about the jitter. At minimum, I would prefer if jitter was not injected into the wake heights, but instead a wall-time that clients would inject into their scheduled wake times. This is not serious enough to block, though.

Also, one minor naming nit.

zcash_pool_migration/src/state.rs
```diff
@@ -281,6 +283,45 @@ impl MigrationState {
...
 
+    /// The minimal schedule of sync/proving wake-ups for the transfers that still need proofs, as
+    /// of `target_height` (`chain_tip + 1`): each entry is a height at which to wake, sync, and
+    /// prove (see [`crate::scheduling::schedule_sync_wakeups`], which defines the windows, the
+    /// minimality guarantee, the jitter, and the immediate wake-up that collects overdue
+    /// transfers). This is the schedule a background-constrained wallet registers with its OS,
+    /// alongside the (independent) broadcast heights the transfers themselves carry.
+    ///
+    /// Covered are transfers in the `Signed` or `AwaitingSignature` state — the latter cannot be
+    /// proved until its signature arrives, but its wake-up is still needed — while `Proved`,
```
nuttycom: This is not quite accurate: proving and signature application are entirely *independent* operations, but I think this is just a documentation failure, it doesn't change the algorithm.

zcash_pool_migration/src/scheduling.rs
nuttycom: We *do* expect to have similar wake times, because the anchor heights are bucketed and shared across wallets. A wallet is going to want to notify the user to create its proofs as soon as it can after the bucket height at which it arrives, to give the maximum window for sync and proving before broadcast. So I would expect wallets in general to want to prove as soon as the anchor height is stable (and 10 blocks is a conservative anchor stability window).

nullcopy (approved):
utACK c62e513855196668ab80f517a027da95fbc02f80

zcash_pool_migration/src/scheduling.rs
```diff
@@ -532,6 +655,145 @@ pub fn expiry_height(current_height: BlockHeight) -> BlockHeight {
...
+        }
+        windows = surviving;
+    }
+
+    // Greedy grouping in deadline order: a window joins the open group iff it still contains the
+    // group's first (smallest) deadline; the group's wake-up point will lie in
+    // `[max_ready, first_deadline]`, which every member's window contains. A window that fails the
+    // open group fails every earlier (smaller-deadline) group too, so a new group is opened
+    // exactly when the classic optimal greedy would place a new piercing point.
+    struct Group<T> {
```
dannywillems, line 748: Nit: move this outside of the function, even if the structure is only used nested.

PR #2804, opened 2026-07-28
zcash_client_backend: Implement Debug, Display and Error for BirthdayError

### LINKED ISSUE #2803: BirthdayError implements neither Display nor Error
`zcash_client_backend::data_api::BirthdayError` is `#[non_exhaustive]` but implements neither `Display` nor `std::error::Error`, so a consumer's fallback arm cannot say anything about the variant it caught.

Concretely, matching on the result of `AccountBirthday::from_treestate` now requires a wildcard arm, and that arm has nothing to report:

```rust
AccountBirthday::from_treestate(treestate, recover_until).map_err(|e| match e {
    BirthdayError::HeightInvalid(e) => anyhow!("Invalid TreeState: Invalid height: {}", e),
    BirthdayError::Decode(e) => anyhow!("Invalid TreeState: Invalid frontier encoding: {}", e),
    // Neither `{}` nor `{:?}` compiles here, so the error's own information is lost.
    _ => anyhow!("Invalid TreeState: unrecognized birthday error"),
})?;
```

A `Display` impl plus `impl std::error::Error` (with `source()` returning the inner `TryFromIntError` / decode error where applicable) would let callers surface unknown variants faithfully, and would let the type flow through `anyhow`/`thiserror` without a hand-written bridge. `Debug` would help too, for the same reason.

This matters more now that the type is `#[non_exhaustive]`: every consumer must write the arm that cannot report anything, and each new variant is silently flattened into it.

---

nuttycom:
`data_api::BirthdayError` is `#[non_exhaustive]` but implemented none of
`Debug`, `Display` or `std::error::Error`, so the wildcard arm that every
consumer must now write when matching the result of
`AccountBirthday::from_treestate` had nothing it could report — neither `{}`
nor `{:?}` compiled.

This adds `#[derive(Debug)]`, a `Display` impl that interpolates the inner
error into a message naming the stage that failed, and an `Error` impl whose
`source()` returns the inner `TryFromIntError` / `io::Error`. Consumers can
then surface unrecognized variants faithfully, and the type flows through
`anyhow`/`thiserror` without a hand-written bridge. Variant docs and a
CHANGELOG entry are included.

`std::error::Error` is referred to by its full path because `data_api` already
has a child module named `error`.

Closes #2803

---

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -12,6 +12,8 @@ workspace.
 
 ### Added
 - `impl Debug for zcash_client_backend::data_api::ll::wallet::PutBlocksError`
+- `zcash_client_backend::data_api::BirthdayError` now implements `Debug`,
+  `Display` and `std::error::Error`.
 - `zcash_client_backend::tor::Timeouts`
 - `zcash_client_backend::tor::Client::create_with_timeouts`
 - `zcash_client_backend::tor::http::TimeoutPhase`
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -66,7 +66,7 @@ use nonempty::NonEmpty;
 use secrecy::SecretVec;
 use std::{
     collections::{HashMap, HashSet},
-    fmt::Debug,
+    fmt::{self, Debug},
     hash::Hash,
     io,
     num::{NonZeroU32, TryFromIntError},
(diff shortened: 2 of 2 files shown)


Cosmos-Harry (approved):

Cosmos-Harry (approved):
utACK

schell (approved):

nullcopy (approved):

dannywillems (approved):
There aren't enough approvals for this highly sensitive change.

Image

PR #2806, opened 2026-07-28
Gate the testing imports, and lint every crate against each of its features

dannywillems:
Two commits: a small import fix, and the CI job that would have caught it.

## The fix

Eight imports in `zcash_client_backend/src/data_api/testing/` sit at the top of
an UNGATED `use` block but are used only by `orchard`- and
`transparent-inputs`-gated scenarios: `PoolType`, `TxId`, `InputSource`,
`WalletRead`, `WalletWrite`, `TargetHeight`, `LockFilter`, `LockedInputPolicy`.

Under `--all-features` every gate is on and the imports are used, so nothing
complained. A downstream crate that takes `test-dependencies` WITHOUT those pool
features compiles a configuration in which they are dead, and warns.
`zcash_pool_migration`'s dev-dependency is exactly such a consumer, which is
where they surfaced.

Each name moves under the gate of the code that uses it. `TxId` in `locking`
stays ungated: unlike the others it is used by scenarios that are not
transparent-specific.

## The CI gap

`Clippy (MSRV)` lints `--all-features`, `Clippy (no default features)` lints the
empty set. **The two endpoints, and nothing in between** -- so an import used
only under a feature can sit in an ungated block indefinitely.

This is not a one-off. The same class has now appeared three times: these eight
imports, plus two in a recent refactor (an ungated `use getset::...` whose only
consumer was orchard-gated, and a `pub mod` inserted between a `#[cfg]` and its
target, silently transferring the gate). All three were invisible at both
endpoints.

`clippy-feature-combos` lints each crate against each of its own features. Crate
list AND feature list both come from `cargo metadata`, so a new crate or feature
is covered the day it lands rather than when someone remembers to edit CI. One
job per crate loops its features in-process, so the dependency build is paid
once and reused across configurations instead of once per runner.

Single features only: the powerset is not an option (2^19 for
`zcash_client_backend`) and pairwise interactions are deliberately out of scope.
Single features are what catch this class.

## The debt this uncovered

I swept all 115 crate/feature configurations before writing the job. **45
already fail**, in three kinds:

| kind | example | what it means |
|---|---|---|
| does not compile | `zcash_client_sqlite --features orchard` is missing `get_locked_outputs`; `zcash_primitives` benches need `sapling::prover` | real feature-gating bugs |
| dead code | most of `pczt` (`PLACEHOLDER_ANCHOR` never used) | same class as the fix above |
| ordinary lints | `zcash_keys --features orchard`: an unneeded `return`, a loop that never loops | reachable only in that configuration |

Twelve of eighteen crates are already fully clean. The six that are not are
recorded in `.github/workflows/clippy-known-failures.txt`, which is explicitly
DEBT and not policy: a listed configuration may fail, anything else may not. So
the job is green today and fails on anything new. It prints a notice when a
listed configuration starts passing, so the list cannot rot silently. Burning
one down is: fix it, delete the line.

I verified the job's exact verdict logic against the real sweep data: **0
unexpected failures, 0 stale entries.**

## Testing

- All four `zcash_client_backend` configurations that the fix touches lint clean
  with `-D warnings`: `test-dependencies` alone, plus `,orchard`,
  `,transparent-inputs`, and all three together.
- The full 115-configuration sweep is the basis for the allowlist.
- YAML validated; `required-checks` updated to include both new jobs.

Happy to split the CI job into its own PR if you would rather land the import
fix on its own.


dannywillems:
The 45 pre-existing configurations recorded in `clippy-known-failures.txt` are now tracked in #2807, which groups them into their ~8 root causes with a suggested order. Nothing in this PR fixes them; it only stops the list from growing.

Worth flagging one from that triage: `zcash_client_sqlite` fails to **compile** in 11 of its 13 configurations because `OutputLockStore::get_locked_outputs` is `#[cfg]`-gated on the trait in one crate and on the impl in another, and the two gates are evaluated against different feature sets. That is a structural hazard rather than a lint, and it is the first thing I would fix.

.github/workflows/ci.yml
```diff
@@ -443,6 +443,91 @@ jobs:
...
+      # metadata` means a new crate or feature is covered the day it lands.
+      #
+      # One job per crate, looping its features in-process: the dependency build is paid once and
+      # reused across every configuration, rather than once per runner.
+      #
+      # Configurations listed in clippy-known-failures.txt are pre-existing debt and may fail; see
+      # that file for what they are and how to burn them down. Anything else must lint clean.
+      - name: Clippy each feature
+        run: |
+          set -u
```
dannywillems: Move this into a bash script so we can re-execute it locally easily. See `.github/helpers`.

zcash_keys/src/address.rs
```diff
@@ -539,11 +539,11 @@ pub mod testing {
 
     #[cfg(not(feature = "sapling"))]
     pub fn arb_addr(request: UnifiedAddressRequest) -> impl Strategy<Value = Address> {
-        return prop_oneof![
```
dannywillems, line 542: 😄

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
-
         if self.cmx.is_some() {
             return Ok(());
         }
 
         let recipient = Address::from_raw_address_bytes(
             self.recipient
                 .as_ref()
-                .ok_or(ParseError::InvalidExtractedNoteCommitment)?,
+                .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?,
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
         )
         .into_option()
-        .ok_or(ParseError::InvalidExtractedNoteCommitment)?;
+        .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?;
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
         let rho = Rho::from_bytes(&spend_nullifier)
             .into_option()
-            .ok_or(ParseError::InvalidExtractedNoteCommitment)?;
+            .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?;
         let rseed = RandomSeed::from_bytes(
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
             *self
                 .rseed
                 .as_ref()
-                .ok_or(ParseError::InvalidExtractedNoteCommitment)?,
+                .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?,
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
             &rho,
         )
         .into_option()
-        .ok_or(ParseError::InvalidExtractedNoteCommitment)?;
+        .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?;
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
         let note = Note::from_parts(
             recipient,
             NoteValue::from_raw(
                 self.value
-                    .ok_or(ParseError::InvalidExtractedNoteCommitment)?,
+                    .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?,
             ),
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1874,48 +1875,41 @@ impl Output {
...
             rho,
             rseed,
             note_version,
         )
         .into_option()
-        .ok_or(ParseError::InvalidExtractedNoteCommitment)?;
+        .ok_or(::orchard::pczt::ParseError::InvalidExtractedNoteCommitment)?;
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

pczt/src/orchard.rs
```diff
@@ -1991,26 +1982,26 @@ impl Output {
...
         if self.cv_net.is_some() {
             return Ok(());
         }
 
-        let spend_value: NoteValue =
-            NoteValue::from_raw(self.spend.value.ok_or(ParseError::InvalidValueCommitment)?);
+        let spend_value: NoteValue = NoteValue::from_raw(
+            self.spend
+                .value
+                .ok_or(::orchard::pczt::ParseError::InvalidValueCommitment)?,
```
dannywillems: Import `::orchard::pczt::ParseError` instead.

zcash_client_sqlite/src/wallet/init/migrations/ironwood_pool_code_views.rs
```diff
@@ -159,6 +159,30 @@ impl RusqliteMigration for Migration {
...
+    #[cfg(feature = "orchard")]
+    use tempfile::NamedTempFile;
+    #[cfg(feature = "orchard")]
+    use zcash_keys::keys::UnifiedSpendingKey;
+    #[cfg(feature = "orchard")]
+    use zcash_protocol::consensus::Network;
+    #[cfg(feature = "orchard")]
+    use {
+        crate::WalletDb, crate::testing::db::test_clock, crate::testing::db::test_rng,
+        crate::wallet::init::WalletMigrator,
```
dannywillems: Could gather crate::wallet import here.

zcash_client_sqlite/src/wallet/init/migrations/ironwood_pool_code_views.rs
```diff
@@ -159,6 +159,30 @@ impl RusqliteMigration for Migration {
...
+        crate::wallet::init::migrations::tests::ArbIronwoodNote,
+        crate::wallet::init::migrations::tests::ArbOrchardNote,
+        crate::wallet::init::migrations::tests::arb_ironwood_note,
+        crate::wallet::init::migrations::tests::arb_orchard_note,
+    };
+    #[cfg(feature = "orchard")]
+    use {
+        proptest::prelude::ProptestConfig, proptest::prop_assert_eq,
```
dannywillems: Could gather proptest:: import here.

zcash_client_sqlite/src/wallet/init/migrations/witness_stabilized_notes.rs
```diff
@@ -105,6 +105,9 @@ mod tests {
 
     #[cfg(feature = "orchard")]
     use zcash_client_backend::data_api::ORCHARD_SHARD_HEIGHT;
+    use zcash_client_backend::data_api::scanning::ScanPriority;
+    use zcash_protocol::ShieldedPool;
+    use {crate::wallet::scanning::mark_stabilized_notes, crate::wallet::scanning::priority_code};
```
dannywillems: Could gather crate::wallet together here.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1646,6 +1646,20 @@ mod tests {
...
     use crate::error::SqliteClientError;
+    #[cfg(feature = "orchard")]
+    use ::orchard::tree::MerkleHashOrchard;
+    #[cfg(feature = "orchard")]
+    use incrementalmerkletree::frontier::Frontier;
+    #[cfg(feature = "orchard")]
+    use rand::SeedableRng;
+    #[cfg(feature = "orchard")]
+    use rand_chacha::ChaChaRng;
+    use shardtree::error::ShardTreeError;
```
dannywillems: Could gather shardtree import together here.

zcash_client_sqlite/src/wallet/commitment_tree.rs
```diff
@@ -1646,6 +1646,20 @@ mod tests {
...
+    use shardtree::store::ShardStore;
+    use std::collections::BTreeSet;
+    #[cfg(feature = "orchard")]
+    use zcash_client_backend::data_api::ORCHARD_SHARD_HEIGHT;
```
dannywillems: Could gather zcash_client_backend together here.

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -682,7 +682,51 @@ pub(crate) mod tests {
...
+    use std::convert::Infallible;
+    #[cfg(feature = "orchard")]
+    use zcash_client_backend::decrypt_transaction;
+    #[cfg(feature = "orchard")]
+    use zcash_keys::address::Address;
+    use zcash_primitives::block::BlockHash;
+    #[cfg(feature = "orchard")]
+    use zcash_protocol::value::Zatoshis;
+    use {crate::TxRef, crate::testing::db::TestDbFactory};
+    use {
```
dannywillems: Could be gathered too...

zcash_client_sqlite/src/wallet/orchard.rs
```diff
@@ -1733,10 +1675,43 @@ pub(crate) mod tests {
...
         use crate::testing::{
             BlockCache,
             db::{TestDb, TestDbFactory},
         };
+        use crate::wallet::orchard::select_spendable_ironwood_notes;
+        use rand_core::OsRng;
+        use std::collections::HashMap;
+        use std::num::NonZeroU32;
+        use transparent::builder::TransparentSigningSet;
+        use zcash_client_backend::data_api::TargetValue;
```
dannywillems: Could be gathered.

pczt/src/orchard.rs
```diff
@@ -1870,49 +1869,42 @@ impl Output {
...
+    ) -> Result<(), OrchardParseError> {
         if self.cmx.is_some() {
             return Ok(());
         }
 
         let recipient = Address::from_raw_address_bytes(
             self.recipient
                 .as_ref()
-                .ok_or(ParseError::InvalidExtractedNoteCommitment)?,
+                .ok_or(OrchardParseError::InvalidExtractedNoteCommitment)?,
```
pacu, line 1880: Does error aliasing affect how they are decribed upstream? 

If a caller gets this error, does the error display `OrchardParseError` or just `ParseError`?

I think if the error name doesn't help to locate the code it doesn't help the developers much.

pczt/src/orchard.rs
```diff
@@ -1,21 +1,32 @@
...
+
 #[cfg(feature = "orchard")]
-use zcash_note_encryption::{Domain, ENC_CIPHERTEXT_SIZE, EphemeralKeyBytes, ShieldedOutput};
+use {
+    ::orchard::{
+        Address, Note, ValuePool,
+        bundle::BundleVersion,
+        note::{ExtractedNoteCommitment, Nullifier, RandomSeed, Rho},
+        note_encryption::{CompactAction, IronwoodDomain, OrchardDomain, OrchardNoteEncryption},
+        pczt::ParseError as OrchardParseError,
```
dannywillems, line 20: We should remove this alias. I don't see any conflict with another existing `ParseError`, therefore I think it is unnecessary.

pczt/src/orchard.rs
dannywillems: I added the comment https://github.com/zcash/librustzcash/pull/2806/changes/BASE..5a23f40495cb9f5f378776d2c7eabda39143b310#r3690385465.

> Does error aliasing affect how they are decribed upstream?

No, it is only local to the file.

> If a caller gets this error, does the error display OrchardParseError or just ParseError?

`ParseError`.

.github/helpers/clippy-each-feature.sh
nuttycom, line 1: I would move this to a top-level `scripts` directory to make it useful outside of CI

components/zcash_address/src/kind/unified/address.rs
```diff
@@ -253,7 +253,7 @@ pub mod testing {
     }
 }
 
-#[cfg(any(test, feature = "test-dependencies"))]
```
nuttycom, line 256: Do we actually not have any tests exercising these test vectors? We should fix that!

zcash_keys/src/keys.rs
```diff
@@ -1715,24 +1704,23 @@ impl UnifiedIncomingViewingKey {
...
+        #[cfg(feature = "sapling")]
         loop {
-            let res = self.address(j, request);
-            match res {
-                Ok(ua) => {
-                    return Ok((ua, j));
-                }
-                #[cfg(feature = "sapling")]
+            match self.address(j, request) {
+                Ok(ua) => return Ok((ua, j)),
```
nuttycom, line 1712: What edition did this syntax finally become allowed in??? Yay!

nuttycom (approved):
utACK with a couple nonblocking comments, this is a nice regularization.

.github/helpers/clippy-each-feature.sh
dannywillems: See https://github.com/zcash/librustzcash/issues/2875

pczt/src/orchard.rs
dannywillems: See https://github.com/zcash/librustzcash/issues/2876

dannywillems:
Opened follow-up issues, and I will spawn agents to address them.

components/zcash_address/src/kind/unified/address.rs
dannywillems: See https://github.com/zcash/librustzcash/issues/2877

PR #2799, opened 2026-07-27
zcash_pool_migration: advisory note locks at prove; wallet record at broadcast

dannywillems:
**Taken over and reworked** per the cancel-stack design (plan D7); now stacked on #2923 (B2) ← #2921 (B1) ← #2914 (A). The original two-part design ("a migration reserves the notes it commits to, and the user always wins a conflict over them") survives intact in Part 1; Part 2 (cancel-on-conflict) moves to the Phase C PR that follows, where it merges with the user-initiated cancel as one function.

## Part 1 (original work, rebased): advisory locks at prove

`prove_transfer` / `prove_preparation` lock the transaction's spent notes via `MigrationProver::lock_spent_notes` once the proof succeeds, under a token derived (BLAKE2b-256 over the sorted spent-note refs — crash-recoverable, idempotent re-lock) and expiring at the transaction's own `expiry_height` (exact ZIP 203 alignment). Locks are advisory: default note selection steers away; a `LockedInputPolicy` override reaches every reserved note; a rival lock fails the prove and discards the proof. Ported onto the current engine: the token now rides `ProvedTransaction` so the artifact and its reservation are recorded in one act, and the lock step lives inside the `ProveOutcome` flow (no lock is taken on a `NotYetProvable`/`MarkedUnsatisfiable` answer).

## Part 2 (new): the wallet record binds at BROADCAST, not prove

Store-at-prove served the activity view but dragged in hard spend marks — freezing the migration's input notes out of the user's balance for the whole prove-to-broadcast window (days to weeks under ZIP 318), against the rule that the user always has full access to their funds. With locks as the reservation:

- `store_proved_transaction` records the proof in the migration store alone.
- New **`take_transaction_for_broadcast`** finalizes the proved PCZT, records the wallet-side transaction (sent outputs, spend marks, status-queue entry), and returns the broadcastable bytes in one database transaction — **binding at the attempt**, so a consumer that dies mid-submit has already left the record the `Proved`-row promotion sweep relies on. Moving the record also moves the `queue_tx_status` enqueue, closing a live privacy leak: the wallet previously asked lightwalletd about txids that had never been broadcast.
- A **backfill schema migration** removes the wallet records earlier releases wrote at prove for never-broadcast transactions of TERMINAL migrations (including the rows the SDK's hand-rolled `Failed`-as-cancel left in the field); pending migrations' records are kept and recreated at broadcast.

[...]

PR #2808, opened 2026-07-28
zcash_pool_migration: split the testing module into three

dannywillems:
A pure move: `testing.rs` becomes `testing/{strategies,conformance,scenarios}.rs`,
re-exported from `testing/mod.rs`. No logic changes.

## Motivation

The module held three things whose consumers are disjoint:

- **`proptest` generators** (`arb_*`), used by the crate's own codec proptests
  and by any downstream store crate;
- **the backend-agnostic conformance suites** (`assert_*`), which a store crate
  points at to inherit the `PoolMigrationRead` / `PoolMigrationWrite` contract,
  and which pin what a `SigningRoundStrategy` must satisfy;
- **the fixed golden data** (`MIGRATION_SCENARIOS`, the signing-round vectors),
  shared between the fast accounting tests and the slow proving tests so both
  speak about the same balances.

At 972 lines, a reader after any one of those waded through the other two.

## No API break

Everything is re-exported from `testing/mod.rs`, so every existing path
(`zcash_pool_migration::testing::arb_migration_state`, ...) resolves unchanged.

The check that matters: `git diff origin/main -- zcash_client_sqlite
zcash_pool_migration/tests zcash_pool_migration_memory` is **empty**. No
consumer needed an edit.

## The split also makes the layering visible

`conformance` and `strategies` both depend on `scenarios`, and not on each
other. That says what these three things are: the scenarios are the data, the
strategies generate more of it, and the conformance suites are the satisfaction
relation over both.

Two small cross-module imports encode that and are commented where they appear:
`conformance` takes the golden vectors and the workload builder it replays, and
`strategies` takes the same workload builder so the random and the fixed paths
construct runs identically.

## Testing

- `zcash_pool_migration --all-features`: 168 pass (unchanged).
- `zcash_client_sqlite --all-features`: 461 (2 ignored) + 2 + 4 (1 ignored),
  unchanged. This is the suite that actually consumes the conformance module.
- `cargo fmt`, and clippy with `-D warnings` under `--no-default-features`,
  `--features test-dependencies`, and `--all-features`: clean.
- Intra-doc links: no new failures (the two on this crate are pre-existing on
  `main`, and are fixed in #2799).

## Context

This is one step of a larger reorganization of the crate; #2805 did the layering
(`model` / `backend` / `lifecycle` / driver) and is independent of this. Neither
depends on the other, but both touch the crate, so whichever lands second will
want a rebase.


zcash_pool_migration/src/testing/scenarios.rs
```diff
@@ -0,0 +1,387 @@
...
+use zcash_protocol::value::COIN;
+
+use crate::engine::{MigrationTransferId, MigrationTxKind};
+use crate::signing_rounds::{PREPARATION_ACTIONS, PlannedTx, SigningRoundBudget, TRANSFER_ACTIONS};
+
+use alloc::vec::Vec;
+
+/// Build the canonical planned-transaction list for `n_prep` preparation and `n_transfer` transfer
+/// transactions (preparation first, then transfers), with sequential ids.
+pub fn planned_txs(n_prep: usize, n_transfer: usize) -> Vec<PlannedTx> {
```
dannywillems, line 19: I think this should be out of the scenarios.rs file/testing env. There might also be existing methods for this.

nullcopy (approved):

PR #2810, opened 2026-07-28
zcash_client_sqlite: flag previously-received Ironwood notes as change

### LINKED ISSUE #2815: zcash_client_sqlite: `flag_previously_received_change` never flags Ironwood notes, so wallets show their own change as a recipient
## The missing line

`flag_previously_received_change` (`zcash_client_sqlite/src/wallet.rs:5301-5330`) builds a per-pool closure and then invokes it for exactly two pools:

```rust
flag_received_change(ShieldedPool::Sapling)?;
#[cfg(feature = "orchard")]
flag_received_change(ShieldedPool::Orchard)?;

Ok(())
```

There is no `ShieldedPool::Ironwood` arm. Everything the arm would need already exists: `table_constants` returns `IRONWOOD_TABLE_CONSTANTS` (`wallet/common.rs:71,94`) gated on `orchard` exactly as the Orchard arm is, the prefix resolves to `ironwood` (`wallet.rs:5601`), and `ironwood_received_notes` carries both columns the statement touches, `is_change INTEGER NOT NULL` and `recipient_key_scope INTEGER` (`wallet/init/migrations/ironwood_received_notes.rs:60,63`). The omission is a missed line, not a missing capability.

## What the function is for

A note received on the internal key scope is not by itself change. It is change only when the same account also *spent* in that transaction, and that second fact is not always available when the note is first written.

`is_change` has three writers, all combining monotonically through `is_change = MAX(:is_change, is_change)` (`wallet/orchard.rs:477,499`):

| \# | Writer | Condition it uses | Pools |
| -- | -- | -- | -- |
| 1 | `store_sent_tx`, `wallet.rs:3591-3636` | recipient classified `Recipient::InternalShielded`, stored as `TransferType::AccountInternal` | generic |
| 2 | `scan_block`, `zcash_client_backend/src/scanning.rs:975` | `spent_from_accounts.contains(key.account_id())` | generic |
| 3 | `flag_previously_received_change`, `wallet.rs:5301` | internal scope **and** a same-account `sent_notes` row | **Sapling and Orchard only** |

Writer 3 is the backstop for writer 2. `spent_from_accounts` is derived by matching a block's nullifiers against the wallet's currently known unspent set (`scanning.rs:457-466`), so it returns false whenever a block is scanned before the block that created its inputs. That is not an edge case in this codebase: scanning is driven by a `ScanPriority` queue rather than front to back, and `scanning.rs:846,866` collects `unlinked_nullifiers` for exactly the nullifiers the wallet cannot yet resolve. When the transaction is later enhanced and its `sent_notes` rows are written, writer 3 repairs what writer 2 could not know.

## Why the omission is permanent

`is_change` is only ever raised, never lowered, and nothing revisits a received note once its transaction has been recorded. An internal-scope Ironwood note that missed the repair therefore keeps `is_change = 0` for the life of the wallet. A code fix alone does not help a note already written; that needs a migration.

Because the flag is monotone, a missing writer can only under-approximate. This can leave change unmarked; it can never mark a genuine payment as change. That bounds the severity: the failure mode is "your own change appears as a recipient", never "a real recipient disappears".

## The bugs this produces

Every one is a classification error. `account_balance_delta` does not read the flag, so balances stay correct throughout. What degrades is the account the wallet gives of the value.

1. **The wallet offers the user's own change as a recipient.** A wallet renders "sent to" from the `v_tx_outputs` rows that are not change. An unflagged change note is in that set, so the user is shown their own internal address as the recipient of their own transaction.
2. `v_transactions` **miscounts the transaction.** `has_change` reports false; `received_note_count` counts the change note as an incoming payment; the `sent_note_counts` CTE filters on `COALESCE(ro.is_change, 0) = 0`, so `sent_note_count` and `memo_count` count it as a note the wallet sent.
3. **A shielding transaction stops being recognised as one.** `is_shielding` requires `MAX(COALESCE(sent_note_counts.sent_notes, 0)) = 0`, which the leaked change output falsifies. This one moves displayed **amounts**, not labels: a wallet that renders shielding transactions from `total_spent` and ordinary sends from `account_balance_delta` switches formulas when the flag flips.
4. **An Orchard-to-Ironwood turnstile crossing reads as an incoming payment.** This is the widest case. After NU6.3 no value may be added to the Orchard pool, so every post-NU6.3 shielded change output lands in Ironwood: the pool this repair skipped is the pool change now always uses. The crossing pays the account's own internal receiver and has no external recipient at all, yet unflagged it reports one, and the whole crossed balance reads as money arriving from outside.

## Affected wallets

Those holding an internal-scope Ironwood note written by scanning at a point where the transaction's inputs were not yet linkable. A wallet that created the transaction locally is unaffected, because writer 1 flags the note at send time. Ironwood shipped in `zcash_client_sqlite 0.22.0-rc.1`, so no released `0.21.x` wallet is affected and no `maint/` backport is needed.

## Root cause worth recording

The correctness condition is "this list names every pool with a `{prefix}_received_notes` table", and nothing enforces it. Adding Ironwood satisfied every compiler check because the list is data, not types. The same crate has the same two-pool list a second time, in `fix_bad_change_flagging`, which is the migration-time form of this repair.

An audit for this class found three further instances: zcash/librustzcash#2812, zcash/librustzcash#2813 and zcash/librustzcash#2814. The compiler enforces pool coverage exactly where a pool is a *type*, which is why none of them is a `match` arm; it enforces nothing where a pool is *data*, such as a `UNION` of per-pool SQL branches or a list of table-name strings.

## Fix

zcash/librustzcash#2810 adds the Ironwood arm, adds `fix_bad_ironwood_change_flagging` to repair wallets already holding a misclassified note, and covers the four bugs above with scenario tests that assert the wrong state before the change and the right one after.

dannywillems:
Closes #2815.

## Problem

`flag_previously_received_change` repairs the classification of a received note that was recorded before the wallet could link its transaction's spends to itself. It ran over the Sapling and Orchard received-note tables, but not the Ironwood one.

`is_change` has three writers, and only this one was missing a pool:

| Writer | Condition | Pools |
| --- | --- | --- |
| `store_sent_tx` (`wallet.rs`) | recipient classified `Recipient::InternalShielded` | generic |
| `scan_block` (`zcash_client_backend/src/scanning.rs:975`) | `spent_from_accounts.contains(account)` | generic |
| `flag_previously_received_change` | internal scope + same-account `sent_notes` row | **Sapling, Orchard only** |

The third is the backstop for the second. `spent_from_accounts` is derived by matching a block's nullifiers against the wallet's known unspent set, so it is false whenever a block is scanned before the block that created its inputs. That is routine: scanning is driven by a `ScanPriority` queue rather than front to back, and `scanning.rs` collects `unlinked_nullifiers` for exactly the nullifiers the wallet cannot yet resolve.

The same two-pool list appears twice in the crate. `fix_bad_change_flagging` is the migration-time form of this repair and hardcodes the same pair. This PR adds the Ironwood arm to the runtime repair and the corresponding sibling migration.

## Why the omission is permanent

`is_change` is only ever raised (`is_change = MAX(:is_change, is_change)`), and nothing revisits a received note after its transaction has been written. An internal-scope Ironwood note that the scanner could not classify therefore keeps `is_change = 0` for the life of the wallet, and a code fix alone would not repair one already recorded.

Because the flag is monotone, a missing writer can only under-approximate: this can leave change unmarked, but can never mark a genuine payment as change.

## Why this is wider than it looks: the turnstile

After NU6.3 no value may be added to the Orchard pool, so every post-NU6.3 shielded change output lands in Ironwood. The pool this repair skipped is the pool change now always uses.

The Orchard-to-Ironwood crossing is the sharpest case. It pays the account's own internal receiver, so its output is exactly the note class affected, and it has no external recipient at all. Unflagged, the wallet reports one anyway, and the whole crossed balance reads as money arriving from outside.

## Observable effects

All of them are classification errors. `account_balance_delta` does not read the flag, so balances stay correct throughout; what degrades is the account the wallet gives of the value.

- `v_transactions` counts the change note in `received_note_count` and `sent_note_count` instead of reporting `has_change`.
- `v_tx_outputs` reports the account's own change as a non-change output. A wallet renders "sent to" from exactly that set, so the user is shown their own internal address as the recipient of their own transaction.
- `is_shielding` requires that the wallet know of no external outputs, which the leaked change output falsifies. This one moves displayed **amounts** rather than labels: a wallet that renders shielding transactions from `total_spent` and ordinary sends from `account_balance_delta` switches formulas when the flag flips.

## Change

Two commits.

**`b73dcd823c`** adds the `ShieldedPool::Ironwood` arm to `flag_previously_received_change`, gated on `orchard` to match `table_constants`, plus a comment recording why an omission there is permanent rather than transient.

**`3cdfa190b7`** adds `fix_bad_ironwood_change_flagging`, the Ironwood counterpart of `fix_bad_change_flagging`, running that migration's statement across the Ironwood received-note table so that wallets already holding a misclassified note are repaired on upgrade. Idempotent for the same monotonicity reason, so it is safe on an unaffected wallet and safe to reapply after a partial upgrade. Deliberately **not** feature-gated: the Ironwood tables exist in the schema regardless of `orchard` (`lib.rs:176-181`), so a wallet written by an Orchard-enabled build is repaired even by a build that is not.

Sapling and Orchard get no equivalent pass. They were in the runtime repair's pool list throughout, so the predicate would find nothing to change. Running it for them would be harmless rather than wrong; say the word if you would rather have the invariant re-asserted for all three.

## Tests

Ten in total, against a fully migrated wallet database rather than a hand-rolled schema, with table names written out literally so the assertions do not go through the same mapping as the code under test.

Four cover the runtime repair, one per pool plus the scope restriction. With the Ironwood arm removed:

```
test wallet::tests::flags_previously_received_orchard_change ... ok
test wallet::tests::flags_previously_received_sapling_change ... ok
test wallet::tests::flags_previously_received_ironwood_change ... FAILED
    Ironwood: an internal-scope note in a self-funded transaction must be flagged as change
```

Six cover the migration, of which three are scenarios asserting the wrong state before it and the right one after:

| Test | What it pins |
| --- | --- |
| `scenario_change_is_offered_as_a_recipient` | the account's own internal address offered as recipient of its own transaction |
| `scenario_shielding_transaction_is_no_longer_recognised` | `is_shielding` flips, moving displayed amounts |
| `scenario_turnstile_crossing_is_reported_as_an_incoming_payment` | the crossing reads as incoming; also pins `account_balance_delta` unchanged either side, to record that the value accounting never depended on the flag |
| `repairs_and_is_idempotent` | applying twice is a no-op |
| `leaves_external_scope_notes_alone` | a self-payment on the external scope stays visible as an output |
| `migrate` | the migration applies within the DAG |

## Database write atomicity

Per AGENTS.md. The runtime change adds one `UPDATE` to a function that already issues two; all three are the same idempotent statement against different tables, and both call sites (`insert_sent_output`, `put_sent_output`) already run inside a `rusqlite::Transaction`, so no new atomicity requirement arises. The migration is a single statement. Both only ever raise the flag, so a retry after a partial commit converges.

## Verification

- `cargo test -p zcash_client_sqlite --all-features`: `wallet::tests` 28 passed / 0 failed, `wallet::init::migrations` 84 passed / 0 failed
- `cargo clippy -p zcash_client_sqlite --all-features --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `cargo check -p zcash_client_sqlite --no-default-features`: clean

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zcash_client_sqlite/src/wallet.rs
+++ b/zcash_client_sqlite/src/wallet.rs
@@ -5322,9 +5322,16 @@ fn flag_previously_received_change(
         .map_err(SqliteClientError::from)
     };
 
+    // Every pool with a `{prefix}_received_notes` table must appear here. Omitting one is not
+    // merely a missed opportunity to set the flag at this call: `is_change` is only ever
+    // raised, never lowered, and nothing revisits the row afterwards, so for any note whose
+    // spends were not linkable to the wallet at the time it was scanned the omission is
+    // permanent.
     flag_received_change(ShieldedPool::Sapling)?;
     #[cfg(feature = "orchard")]
     flag_received_change(ShieldedPool::Orchard)?;
+    #[cfg(feature = "orchard")]
+    flag_received_change(ShieldedPool::Ironwood)?;
 
     Ok(())
 }
@@ -5798,7 +5805,7 @@ pub mod testing {
 mod tests {
     use std::num::NonZeroU32;
 
-    use rusqlite::Connection;
+    use rusqlite::{Connection, named_params};
     use sapling::zip32::ExtendedSpendingKey;
     use secrecy::{ExposeSecret, SecretVec};
     use uuid::Uuid;
@@ -5817,7 +5824,10 @@ mod tests {
         testing::{BlockCache, db::TestDbFactory},
     };
 
-    use super::{account_birthday, min_shared_checkpoint_height, select_truncation_height};
+    use super::{
+        KeyScope, ShieldedPool, TxRef, account_birthday, flag_previously_received_change,
+        min_shared_checkpoint_height, select_truncation_height,
+    };
 
     #[cfg(feature = "orchard")]
     use {crate::testing::db::TestDb, zcash_protocol::local_consensus::LocalNetwork};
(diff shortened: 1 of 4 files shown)


nuttycom (dismissed):
utACK

nuttycom:
@dannywillems can you add a migration that performs the repair of change flags?

dannywillems:
Don't merge. Adding migration + scenarii tests.

dannywillems:
Pushed `3cdfa190b7`, which adds the migration this PR previously deferred, so the "Not included" section above no longer applies: wallets already holding a misclassified note are now repaired on upgrade and no rescan is required. The CHANGELOG entry has been updated to say so.

The migration is `fix_bad_ironwood_change_flagging`, the Ironwood counterpart of the existing `fix_bad_change_flagging`, running that migration's statement against `ironwood_received_notes`. Same idempotence argument as the runtime repair: `is_change` only ever rises, so it is safe on an unaffected wallet and safe to reapply after a partial upgrade. It is deliberately not feature-gated, because the Ironwood tables exist in the schema regardless of `orchard`, so a wallet written by an Orchard-enabled build is repaired even by a build that is not. Sapling and Orchard get no equivalent pass; they were in the runtime repair's pool list throughout, so the predicate would find nothing.

Three scenario tests were added, each asserting the wrong state before the migration and the right one after, to pin what the flag actually costs a wallet:

- **The account's own change is offered as a recipient.** A wallet renders "sent to" from the non-change rows of `v_tx_outputs`, so an unflagged change note shows the user their own internal address as the recipient of their own transaction.
- **A shielding transaction stops being recognised as shielding.** `is_shielding` requires that the wallet know of no external outputs, which the leaked change output falsifies. This one moves displayed amounts rather than labels: a wallet that renders shielding transactions from `total_spent` and ordinary sends from `account_balance_delta` switches formulas when the flag flips.
- **An Orchard-to-Ironwood turnstile crossing reads as an incoming payment.** This is the widest case and the reason the pool that was missing matters more than it looks. After NU6.3 no value may be added to the Orchard pool, so every post-NU6.3 shielded change output lands in Ironwood: the pool the repair skipped is the pool change now always uses. The crossing has no external recipient at all, yet unflagged it reports one, and the whole crossed balance reads as money arriving from outside.

That last test also pins `account_balance_delta` either side of the migration, to record that the value accounting never depended on the flag. What the bug corrupts is the account of the value, not the value.

Verification on the updated branch: `wallet::init::migrations` 84 passed / 0 failed, `clippy --all-features --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

zcash_client_sqlite/src/wallet/init/migrations/fix_bad_ironwood_change_flagging.rs
```diff
@@ -0,0 +1,528 @@
...
+        () => {
+            TestBuilder::new()
+                .with_data_store_factory(TestDbFactory::default())
+                .with_account_from_sapling_activation(BlockHash([0; 32]))
+                .build()
+        };
+    }
+
+    /// The `accounts.id` of the wallet's only account, and the id of an address belonging to it.
+    fn account_and_address_ids(conn: &rusqlite::Connection) -> (i64, i64) {
```
dannywillems, line 139: Wondering if this could be extracted from the testing environment.

zcash_client_sqlite/src/wallet.rs
```diff
@@ -7137,4 +7147,223 @@ mod tests {
             0
         );
     }
+
+    /// The name of the received-note table for a pool, written out rather than derived from
+    /// `table_constants`, so that these tests do not assert through the same mapping the code
+    /// under test uses.
+    fn received_notes_table(pool: ShieldedPool) -> &'static str {
```
dannywillems, line 7154: Could this be extracted out of the testing env?

nuttycom (approved):
utACK 3ed308ba1635ee413cf58b656428a0a70554d5cd

PR #2816, opened 2026-07-28
zcash_client_sqlite: record Ironwood receipts as uses of their address

### LINKED ISSUE #2812: zcash_client_sqlite: `v_address_uses` has no Ironwood branch, so Ironwood receipts do not count as address uses
## Summary

`v_address_uses` (`zcash_client_sqlite/src/wallet/db.rs:1802-1821`) unions `orchard_received_notes`, `sapling_received_notes` and `transparent_received_outputs`, keyed by `address_id`. There is no `ironwood_received_notes` branch, so a note received in the Ironwood pool does not register as a use of the address it was received at.

## This is not hypothetical

Ironwood notes carry the column the view joins on. `orchard::put_received_note` (`wallet/orchard.rs:438-492`) takes the `shielded_pool` as a parameter and serves both Orchard and Ironwood, and it populates `address_id` via `ensure_address`. The rows exist; the view does not select them.

## Consumers affected

- **`involved_accounts`** (`wallet.rs:1583-1592`) resolves which accounts a set of transactions involves. An account whose only involvement in a transaction is an Ironwood note received at one of its addresses is not reported.
- **`v_address_first_use`** (`db.rs:1824-1836`) feeds the transparent gap-limit search at `wallet/transparent.rs:450`. An address whose only use is an Ironwood receipt has a NULL `first_use_height`, so it reads as never used.

## Why the blast radius is larger than it looks

After NU6.3 a payment to an Orchard receiver is delivered in the Ironwood bundle. Shielded receipts at a unified address therefore become invisible to address-use tracking in the ordinary case, not an exotic one. Treating a used address as unused is an address-reuse hazard, so this is a privacy property rather than a balance one; balances are unaffected because they do not read this view.

## Likely cause

The `ironwood_pool_code_views` migration recreated `v_received_outputs` and `v_received_output_spends` to add the Ironwood branch. `v_address_uses` is the third view in the crate that unions the per-pool received-note tables, and it was not recreated. Neither that migration nor `db.rs` records a reason to exclude it, so this reads as an oversight rather than a decision.

## Suggested fix

Recreate `v_address_uses` with an Ironwood branch, both in `db.rs` and in a new migration, following the shape of `ironwood_pool_code_views`.

## Context

Found while auditing for the class of bug behind #2810, where a hand-maintained per-pool list omitted Ironwood. The compiler enforces pool coverage where a pool is a *type* (an exhaustive `match` breaks when a variant is added); it enforces nothing where a pool is *data*, such as a `UNION` of per-pool SQL branches. This is that second category.

dannywillems:
Closes #2812. Resolves MOB-1594

## The bug

`v_address_uses` is the wallet's record of which of its addresses have been seen
on chain. It unions `orchard_received_notes`, `sapling_received_notes` and
`transparent_received_outputs`, keyed by `address_id`, and has no
`ironwood_received_notes` branch.

This is not hypothetical. `wallet::orchard::put_received_note` takes a
`shielded_pool: ShieldedPool` and serves both Orchard and Ironwood, setting
`address_id` through `ensure_address` in either case. Ironwood notes therefore
carry the column the view joins on; their rows are simply not selected.

After NU6.3 every payment to an Orchard receiver is delivered in the Ironwood
bundle, so an Ironwood-only receipt is the shape of an ordinary received
payment, not an exotic one.

## Consumers affected

- `v_address_first_use` aggregates `v_address_uses` into the first height at
  which each address was used. `wallet::transparent::find_gap_start` selects
  only addresses with `first_use_height IS NOT NULL`, so an address whose sole
  receipt is an Ironwood note is invisible to the transparent gap-limit search
  and stays eligible to be handed out again. That is an address-reuse hazard.
- `wallet::involved_accounts` reads `v_address_uses` to decide which of the
  wallet's accounts a transaction concerns, and returned the empty set for a
  transaction whose only wallet-relevant output was an Ironwood note.
- `wallet::transparent::utxo_query_height` reads `v_address_uses` directly to
  decide where to start searching for UTXOs; it inherits the same blind spot.

## The fix

An Ironwood branch is added to `v_address_uses`, both in the current-schema
definition in `wallet/db.rs` and in a new migration,
`v_address_uses_ironwood`, that recreates the view for existing wallets.

`v_address_first_use` is deliberately **not** recreated. SQLite stores a view's
definition as text and resolves its references to other views by name when a
query is compiled, so it picks up the new `v_address_uses` on its own. The
existing `account_delete_cascade` migration already relies on this: it drops and
recreates `v_address_uses` and leaves `v_address_first_use` alone, and the
schema-comparison test in `wallet/init.rs` passes on that basis today.

The migration depends on `ironwood_received_notes`, which transitively orders it
after `account_delete_cascade` (`account_delete_cascade` ->
`witness_stabilized_notes` -> `orchard_note_version` ->
`ironwood_received_notes`), so it cannot be overwritten by the older definition.
It becomes a new leaf, so `CURRENT_LEAF_MIGRATIONS` and the DAG comment are
updated; `current_leaf_migrations_are_the_dag_leaves` checks that.

## Database write atomicity

The migration issues no `INSERT`, `UPDATE` or `DELETE`. Its two statements are
DDL (`DROP VIEW` then `CREATE VIEW`), and they run in the `rusqlite::Transaction`
that `schemerz` hands to `up`, so they land together or not at all. Answering the
four questions anyway:

1. **How many writes?** Zero data writes; two DDL statements in one supplied
   transaction.
2. **Must they land together?** Yes, and they do: a committed `DROP VIEW`
   without its `CREATE VIEW` would leave the wallet without `v_address_uses`.
   The enclosing transaction guarantees it.
3. **What happens on retry?** The migration is only ever run to completion or
   rolled back, so a retry starts from the pre-migration schema and succeeds.
4. **Does any read path see only one side?** No: no reader can observe the
   intermediate state, because it is never committed.

## Tests

Three tests in the new migration module, all run under `--all-features`:

- `migrate` (`test_migrate`) - the migration applies to an empty wallet.
- `scenario_ironwood_receipt_does_not_mark_its_address_used`
- `scenario_ironwood_receipt_leaves_its_account_uninvolved`

The scenario tests migrate a wallet to the state immediately preceding this
migration, seed an account, an address, a mined transaction and an Ironwood note
received at that address directly with `named_params` (the state under test is
one an upgrading wallet is in), assert the wrong observation, then apply the
migration and assert the right one. They assert against `v_address_uses`,
`v_address_first_use` and `involved_accounts`, the surfaces consumers read, not
the underlying table.

### Proof that the tests catch the bug

With the Ironwood branch removed from both `wallet/db.rs` and the migration's
`up`, and nothing else changed:

```
running 3 tests
test wallet::init::migrations::v_address_uses_ironwood::tests::migrate ... ok
test wallet::init::migrations::v_address_uses_ironwood::tests::scenario_ironwood_receipt_does_not_mark_its_address_used ... FAILED
test wallet::init::migrations::v_address_uses_ironwood::tests::scenario_ironwood_receipt_leaves_its_account_uninvolved ... FAILED

---- ...scenario_ironwood_receipt_does_not_mark_its_address_used stdout ----
assertion `left == right` failed: the Ironwood receipt must be recorded as a use of the address
  left: 0
 right: 1

---- ...scenario_ironwood_receipt_leaves_its_account_uninvolved stdout ----
assertion `left == right` failed: the receiving account must be involved in the transaction that paid it
  left: 0
 right: 1

test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 463 filtered out
```

With the fix restored, all three pass.

## Verification

- `cargo fmt --all -- --check` - clean
- `cargo clippy -p zcash_client_sqlite --all-features --all-targets -- -D warnings` - zero warnings
- `cargo test -p zcash_client_sqlite --all-features` - 464 passed, 0 failed (lib);
  2 passed (`pool_migration_prove_chain_sim`); 4 passed (doc tests)

`cargo check -p zcash_client_sqlite --tests` without `--all-features` fails on
`OutputLockStore::get_locked_outputs`, which reproduces identically on an
unmodified `main` and is unrelated to this change (being addressed in #2811).

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations.rs
@@ -53,6 +53,7 @@ mod tx_retrieval_queue_expiry;
 mod ufvk_support;
 mod utxos_table;
 mod utxos_to_txos;
+mod v_address_uses_ironwood;
 mod v_received_output_spends_account;
 mod v_sapling_shard_unscanned_ranges;
 mod v_transactions_additional_totals;
@@ -150,13 +151,13 @@ pub(super) fn all_migrations<
     //     add_transparent_value_index  v_tx_outputs_key_scopes  standalone_p2sh    witness_stabilized_notes
     //                                                    /          \         \
     //                                                   /            \      orchard_note_version
-    //                                                  /              \                      \
-    //                                                 /                \               ironwood_received_notes
-    //                                                /                  \                     /        |    \
-    //                                               /                    \    ironwood_pool_code_views |  note_locking
+    //                                                  /              \           \
+    //                                                 /                \  ironwood_received_notes -- note_locking
+    //                                                /                  \       |               |------|     |
+    //                                               /                    \    ironwood_pool_code_views |  v_address_uses_ironwood
     //                                              /                      \                            |
-    //                                             /                        \         fix_bad_ironwood_change_flagging
-    //                                         ivk_item_cache    add_transparent_receiver_address_index
+    //                                             /                        \    fix_bad_ironwood_change_flagging
+    //                                          ivk_item_cache    add_transparent_receiver_address_index
     //
     let rng = Rc::new(Mutex::new(rng));
     vec![
@@ -261,6 +262,7 @@ pub(super) fn all_migrations<
         Box::new(ironwood_received_notes::Migration),
         Box::new(ironwood_pool_code_views::Migration),
         Box::new(fix_bad_ironwood_change_flagging::Migration),
+        Box::new(v_address_uses_ironwood::Migration),
         Box::new(orchard_ironwood_migration_tables::Migration),
         Box::new(tree_retained_checkpoints::Migration),
         Box::new(note_locking::Migration),
@@ -411,6 +413,7 @@ pub const CURRENT_LEAF_MIGRATIONS: &[Uuid] = &[
     add_transparent_value_index::MIGRATION_ID,
     ironwood_pool_code_views::MIGRATION_ID,
     fix_bad_ironwood_change_flagging::MIGRATION_ID,
+    v_address_uses_ironwood::MIGRATION_ID,
     orchard_ironwood_migration_tables::MIGRATION_ID,
     tree_retained_checkpoints::MIGRATION_ID,
     note_locking::MIGRATION_ID,
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -47,6 +47,14 @@ workspace.
   in the Ironwood bundle, so this affected ordinary received payments. A
   migration corrects the affected view; no caller action is required beyond
   upgrading.
+- The funding account reported for a transparent output by
+  `zcash_client_backend::wallet::WalletTransparentOutput::funding_account` now
+  takes value spent from the Ironwood pool into account. Ironwood inputs were
+  not counted, so an output whose creating transaction was funded entirely from
+  Ironwood reported no funding account, and one funded from several pools could
+  report an account other than the largest contributor. Post-NU6.3 wallets hold
+  their shielded value in Ironwood, so this affected ordinary spends rather than
+  only unusual ones.
 
 ## [0.22.0-rc.4] - 2026-07-26
 
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -39,6 +39,14 @@ workspace.
   which presents it to the user as a recipient of their own transaction. Balances
   were not affected. Notes already recorded with the wrong classification are
   repaired on upgrade by a new migration; no rescan is required.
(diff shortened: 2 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations.rs
```diff
@@ -151,8 +152,8 @@ pub(super) fn all_migrations<
     //                                                   /            \      orchard_note_version
     //                                                  /              \                      \
     //                                                 /                \               ironwood_received_notes
-    //                                                /                  \                     /           \
-    //                                               /                    \    ironwood_pool_code_views   note_locking
+    //                                                /                  \                       /       |       \
```
dannywillems: 😢

nuttycom (dismissed):
utACK ac0b8ef23b9fbd413392d9a244dd445d2a7fe69f

nuttycom (approved):
utACK 5cbd4e366528e64af6de17a09f62b335e5e9b276

PR #2817, opened 2026-07-28
zcash_client_sqlite: count Ironwood inputs when attributing a funding account

### LINKED ISSUE #2813: zcash_client_sqlite: `list_funding_accounts` does not count Ironwood inputs
## Summary

`list_funding_accounts` (`zcash_client_sqlite/src/wallet/transparent.rs:1058-1093`) sums contributed input value per account with an inner `UNION ALL` over `transparent_received_outputs`, `sapling_received_notes` and `orchard_received_notes`, each joined to its spend table. There is no Ironwood branch, so value spent from the Ironwood pool contributes nothing to the tally.

## Consumer affected

`to_unspent_transparent_output` (`transparent.rs:1134-1137`) takes the largest contributor as the `funding_account` of a `WalletTransparentOutput`:

```rust
// `WalletTransparentOutput` records at most a single funding account; when
// multiple wallet accounts contributed inputs to the creating transaction we
// pick the largest contributor.
let funding_account = list_funding_accounts(conn, creating_tx_id)?
    .into_iter()
    .next()
    .map(|(account, _)| account);
```

Two failure modes:

* A transparent output created by a transaction funded **entirely** from Ironwood notes gets `funding_account = None`.
* On a mixed-pool transaction, the largest contributor is computed from an incomplete tally, so the wrong account wins whenever the Ironwood share would have dominated.

## Why this is the normal case post-NU6.3

After NU6.3 no value may be added to the Orchard pool, so a wallet that has crossed the turnstile holds its shielded value in Ironwood. Any transparent output it subsequently creates is funded from Ironwood, which is exactly the case this query cannot see.

## Suggested fix

Add the Ironwood branch to the `UNION ALL`, joining `ironwood_received_notes` to `ironwood_received_note_spends`. Local to one query; no migration required.

## Context

Found while auditing for the class of bug behind zcash/librustzcash#2810, where a hand-maintained per-pool list omitted Ironwood. See also zcash/librustzcash#2812, which is the same class in a view definition.

dannywillems:
Closes #2813. Resolves MOB-1595

## The bug

`list_funding_accounts` (`zcash_client_sqlite/src/wallet/transparent.rs`) totals
the value each wallet account contributed to a transaction, by unioning the
wallet's spent-output tables. It carried three branches:

- `transparent_received_outputs` / `transparent_received_output_spends`
- `sapling_received_notes` / `sapling_received_note_spends`
- `orchard_received_notes` / `orchard_received_note_spends`

There was no `ironwood_received_notes` / `ironwood_received_note_spends` branch,
so value spent from the Ironwood pool contributed nothing to any account's total.

## The consumer affected

`to_unspent_transparent_output`, immediately below it, takes the largest
contributor as the `funding_account` of the `WalletTransparentOutput` it builds.
That gives two distinct failure modes:

1. A transparent output whose creating transaction was funded **entirely** from
   Ironwood reports `funding_account = None`, because the tally is empty.
2. On a transaction funded from **several pools**, the largest contributor is
   chosen from an incomplete tally, so an account whose Ironwood contribution
   would have dominated loses to an account that contributed less. This is the
   worse case: the answer is not absent but wrong, and nothing downstream can
   tell that it is.

This is not an edge case after NU6.3. No value may be added to the Orchard pool
once the turnstile is crossed, so a wallet that has crossed it holds its shielded
value in Ironwood and funds its ordinary spends from there.

Note that `WalletRead::get_funding_accounts` (the nullifier-based detection in
`zcash_client_backend`) already handles Ironwood, and is covered by
`wallet::orchard::tests::ironwood_privacy_invariants::get_funding_accounts_detects_ironwood_only_spends`.
This SQL helper is a separate path and was not covered.

## The fix

A fourth `UNION ALL` branch over `ironwood_received_notes` joined to
`ironwood_received_note_spends` on `ironwood_received_note_id`, mirroring the
Orchard branch exactly.

**No migration is required, and this was verified rather than assumed.** Both
tables are installed by the existing `ironwood_received_notes` migration and are
present in the golden schema (`wallet/db.rs`, `TABLE_IRONWOOD_RECEIVED_NOTES` and
`TABLE_IRONWOOD_RECEIVED_NOTE_SPENDS`). They are not feature-gated: they exist
whether or not the `orchard` feature is enabled. The change is confined to one
prepared statement, reads only tables that already exist, and writes nothing.

## Why the branches stay explicit

A hand-maintained per-pool list is what caused this bug, so the alternative of
driving the query from the crate's cross-pool views (`v_received_output_spends`
joined to `v_received_outputs`, the pattern `v_transactions` already uses) was
measured rather than dismissed. `EXPLAIN QUERY PLAN` on a migrated wallet:

Explicit branches (this PR) - index lookups only:

```
CO-ROUTINE contribs
  CO-ROUTINE (subquery-4)
    COMPOUND QUERY
      LEFT-MOST SUBQUERY
        SEARCH tros USING INDEX idx_transparent_received_output_spends_transaction_id (transaction_id=?)
        SEARCH tro USING INTEGER PRIMARY KEY (rowid=?)
        SEARCH t USING INTEGER PRIMARY KEY (rowid=?)
      UNION ALL
        SEARCH srns USING INDEX idx_sapling_received_note_spends_transaction_id (transaction_id=?)
        ...
      UNION ALL
        SEARCH orns USING INDEX idx_orchard_received_note_spends_transaction_id (transaction_id=?)
        ...
      UNION ALL
        SEARCH irns USING INDEX idx_ironwood_received_note_spends_transaction_id (transaction_id=?)
        SEARCH irn USING INTEGER PRIMARY KEY (rowid=?)
        SEARCH t USING INTEGER PRIMARY KEY (rowid=?)
```

Same query written over the shared views - full scans of every pool table:

```
MATERIALIZE v_received_outputs
  COMPOUND QUERY
    LEFT-MOST SUBQUERY
      SCAN sapling_received_notes
      ...
    UNION USING TEMP B-TREE
      SCAN orchard_received_notes
      ...
    UNION USING TEMP B-TREE
      SCAN ironwood_received_notes
      ...
    UNION USING TEMP B-TREE
      SCAN u
      ...
SCAN ros
BLOOM FILTER ON ro (pool=? AND id_within_pool_table=?)
SEARCH ro USING AUTOMATIC COVERING INDEX (pool=? AND id_within_pool_table=?)
```

Because the view has to be joined by output id, SQLite materializes it in full;
the predicate on the spending transaction cannot be pushed into it.
`to_unspent_transparent_output` runs this query once per candidate output while
gathering UTXOs, so the view form would turn a per-output index lookup into a
per-output scan of every note the wallet holds. That is the wrong trade for a
selection path, so the branches stay explicit and the hazard is instead recorded
in a doc comment on `list_funding_accounts`: every pool with a received-output
table must appear here, a missing branch is silent rather than an error, and the
views are deliberately not used.

If the maintainers would rather remove the hazard structurally, the shape that
would work is a `value` (and creating-transaction) column on
`v_received_output_spends` so the spend view alone is sufficient. That is a
migration and a separate change.

## Tests

Two scenario tests in `wallet::transparent::tests::funding_accounts`, gated with
the rest of the module behind `transparent-inputs`. Each builds the transparent
output through the wallet's own `put_received_transparent_utxo` path and then
seeds the creating transaction's spent notes directly, because the wallet cannot
yet be driven to build a transaction that spends Ironwood value.

- `scenario_ironwood_funded_output_reports_no_funding_account`
- `scenario_largest_ironwood_contributor_loses_to_a_smaller_one`

Both were proved to catch the bug by removing the new `UNION ALL` branch and
re-running them:

```
---- ...::scenario_ironwood_funded_output_reports_no_funding_account stdout ----
assertion `left == right` failed: the bug: an output funded entirely from Ironwood
reports no funding account, because the funding-account query counts no Ironwood value
  left: None
 right: Some(AccountUuid(81e224f0-46f8-410b-935c-fff87c46efc0))

---- ...::scenario_largest_ironwood_contributor_loses_to_a_smaller_one stdout ----
assertion `left != right` failed: the bug: the smaller Sapling contributor is reported
as the funding account, because the larger Ironwood contribution is not counted
  left: Some(AccountUuid(a30a5ea2-1845-4a95-b9f2-b75973c0ea54))
 right: Some(AccountUuid(a30a5ea2-1845-4a95-b9f2-b75973c0ea54))

test result: FAILED. 0 passed; 2 failed; 463 filtered out
```

(The second is an `assert_ne!`, so both sides printing the same value is the
failure: the reported funding account *is* account B, the smaller contributor.)

With the branch restored, both pass.

## Database write atomicity

Not applicable: this change issues **zero** writes. `list_funding_accounts` is a
read-only `SELECT`, and the fix adds a fourth read branch to it. The scenario
tests write only to their own throwaway in-memory wallets.

## Verification

- `cargo fmt --all -- --check` - clean
- `cargo clippy -p zcash_client_sqlite --all-features --all-targets -- -D warnings` - zero warnings
- `cargo test -p zcash_client_sqlite --all-features` - `463 passed; 0 failed; 2 ignored`, plus `2 passed` in `pool_migration_prove_chain_sim` and `4 passed` doc-tests

`cargo check -p zcash_client_sqlite --no-default-features --features transparent-inputs --tests` still fails on the pre-existing `OutputLockStore::get_locked_outputs` gap that #2811 addresses; that is the only error reported, and it is unrelated to this change.

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -47,6 +47,14 @@ workspace.
   in the Ironwood bundle, so this affected ordinary received payments. A
   migration corrects the affected view; no caller action is required beyond
   upgrading.
+- The funding account reported for a transparent output by
+  `zcash_client_backend::wallet::WalletTransparentOutput::funding_account` now
+  takes value spent from the Ironwood pool into account. Ironwood inputs were
+  not counted, so an output whose creating transaction was funded entirely from
+  Ironwood reported no funding account, and one funded from several pools could
+  report an account other than the largest contributor. Post-NU6.3 wallets hold
+  their shielded value in Ironwood, so this affected ordinary spends rather than
+  only unusual ones.
 
 ## [0.22.0-rc.4] - 2026-07-26
 
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -1067,6 +1067,16 @@ pub(crate) fn utxo_query_height(
 /// are ordered by total contributed value descending; ties are broken in favor of
 /// the account whose oldest contributed input has the lowest mined height (with
 /// unmined inputs sorting last), then by `accounts.id`.
+///
+/// The inner `UNION ALL` must carry one branch per pool in which the wallet can record a
+/// received output: transparent, Sapling, Orchard, and Ironwood. A missing branch does not
+/// produce an error, it silently under-counts the accounts that funded the transaction, so
+/// a pool added to the schema without a branch here changes which account this reports.
+///
+/// The pools are enumerated here rather than read from the cross-pool `v_received_outputs`
+/// view because that view has to be materialized in full to be joined by output id, which
+/// scans every received-note table. This query is run once per candidate output, so it is
+/// written to be satisfied from the spend tables' `transaction_id` indexes instead.
 fn list_funding_accounts(
     conn: &rusqlite::Connection,
     creating_tx_id: i64,
@@ -1099,6 +1109,13 @@ fn list_funding_accounts(
                    ON orns.orchard_received_note_id = orn.id
                  JOIN transactions t ON t.id_tx = orn.transaction_id
                  WHERE orns.transaction_id = :creating_tx_id
+                 UNION ALL
+                 SELECT irn.account_id, irn.value, t.mined_height
+                 FROM ironwood_received_notes irn
+                 JOIN ironwood_received_note_spends irns
+                   ON irns.ironwood_received_note_id = irn.id
+                 JOIN transactions t ON t.id_tx = irn.transaction_id
+                 WHERE irns.transaction_id = :creating_tx_id
              )
              GROUP BY account_id
          ) contribs ON contribs.account_id = a.id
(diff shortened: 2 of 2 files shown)


nuttycom (approved):
utACK fce63f222ade4e493402d328a20703b0ee575c04

PR #2818, opened 2026-07-28
zcash_client_sqlite: cover every shielded pool in the imported-receiver reattribution

### LINKED ISSUE #2814: zcash_client_sqlite: account re-attribution in `store_address_range` skips `ironwood_received_notes`
## Summary

`store_address_range` (`zcash_client_sqlite/src/wallet/transparent.rs:826-830`) re-attributes outputs when an address previously imported as a standalone `Foreign` transparent receiver is later derived by a different account. Deriving the address proves ownership, so the row and the outputs received at it move to the deriving account:

```rust
// If the import was recorded under a different account, the outputs received at
// the address follow the (derivation-proven) attribution to this account.
if foreign_account != account_id.0 {
    for table in [
        "transparent_received_outputs",
        "sapling_received_notes",
        "orchard_received_notes",
    ] {
```

`ironwood_received_notes` is not in the list, although it has the `address_id` column the `UPDATE` matches on and `orchard::put_received_note` populates it for Ironwood notes just as it does for Orchard ones.

## Honest severity

Weaker than zcash/librustzcash#2812 and zcash/librustzcash#2813, and worth stating plainly: a `Foreign` row is a transparent-only import, so it is not obvious that a shielded note can reference it before the upgrade happens. If no shielded note can, then the `sapling_received_notes` and `orchard_received_notes` entries are themselves defensive.

The argument for adding Ironwood is therefore consistency rather than a demonstrated failure. Whatever reasoning put the other two shielded pools in this loop applies unchanged to Ironwood: if they are defensive, Ironwood should be too; if they are load-bearing, so is it. Leaving one shielded pool out is the one option that is wrong under either reading.

A reviewer who knows that a shielded note can never carry a `Foreign` row's `address_id` may prefer to close this and instead drop all three shielded tables from the loop, which would be an equally consistent resolution.

## Suggested fix

Either add `"ironwood_received_notes"` to the list, or establish that no shielded note can reference a `Foreign` address row and remove the two that are there. State which, in a comment.

## Context

Found while auditing for the class of bug behind zcash/librustzcash#2810, where a hand-maintained per-pool list omitted Ironwood. See zcash/librustzcash#2812 and zcash/librustzcash#2813 for the two higher-severity findings from the same audit.

dannywillems:
Closes #2814.

## Investigation first: can a shielded note reference a `Foreign` address row?

No, and this is the part that decides what the right fix is.

The reattribution loop in `store_address_range` moves the account attribution of the
outputs received at an address when a standalone (`Foreign`) transparent import is
upgraded in place by a *different* account deriving the same receiver. It walked
`transparent_received_outputs`, `sapling_received_notes` and `orchard_received_notes`,
but not `ironwood_received_notes`, even though that table has the same `address_id`
column and `orchard::put_received_note` populates it for Ironwood notes exactly as it
does for Orchard ones. Issue #2814 asks whether that is a real mis-attribution bug.

It is not, and neither are the Sapling and Orchard arms reachable. Three independent
legs, each verified against the source:

1. **The schema forbids the shape.** `addresses` carries
   `CONSTRAINT ck_addr_foreign_or_diversified CHECK ((diversifier_index_be IS NULL) ==
   (key_scope = -1))` (`wallet/db.rs:232-234`). Every `Foreign` row therefore has a NULL
   diversifier index. `upsert_address` (`wallet.rs:1504-1576`) hard-codes `key_scope` to
   `KeyScope::EXTERNAL` in both the `INSERT` and the
   `ON CONFLICT (account_id, diversifier_index_be, key_scope)` target, and always binds a
   non-NULL `diversifier_index_be`. It can neither create nor match a `Foreign` row.

2. **`upsert_address` is the sole writer of shielded `address_id`.** Both
   `sapling::ensure_address` (`wallet/sapling.rs:293-334`) and `orchard::ensure_address`
   (`wallet/orchard.rs:348-387`, the path shared by Orchard and Ironwood) return either
   `None` or `upsert_address(..)`. The migration backfills go through the same function
   (`transparent_gap_limit_handling.rs:522` and `:586`); `ironwood_received_notes` is
   created empty with no backfill at all. The only `address_id` resolved by joining on a
   transparent address targets `transparent_received_outputs` alone
   (`transparent_gap_limit_handling.rs:618-621`).

3. **No row is ever demoted to `Foreign`.** `key_scope = -1` is only ever written at
   `INSERT` time, by `import_standalone_transparent_pubkey` (`wallet.rs:909`, flags
   `P2PKH`) and `import_standalone_transparent_script` (`wallet.rs:1002`, flags `P2SH`);
   both bail out early if the receiver already has a row. The one `UPDATE` that writes
   `key_scope` is the upgrade in `store_address_range` itself, which moves a row *out of*
   `Foreign`. So a `Foreign` row also never carries the `SAPLING` or `ORCHARD` receiver
   flag: it is a transparent-only import with no shielded receiver to receive a note at.

### Where the three-table list came from

It was copied from the duplicate-merge loop in `add_transparent_receiver_address_index`
(`add_transparent_receiver_address_index.rs:275-297`), as the commit that added the
runtime loop says (8754bde). **There the shielded entries are genuinely reachable**, and
that is the asymmetry that explains the whole thing: that loop repoints the rows of
*duplicate* address records, selected only by sharing a
`cached_transparent_receiver_address`, and a duplicate can be a fully derived,
external-scoped record belonging to another account, which certainly can carry shielded
notes. A `Foreign` record is explicitly excluded from being the canonical one
(`record_derives_receiver` returns `None` for it). That migration is also right to omit
`ironwood_received_notes`: whichever topological order the migrator picks for the two
sibling branches under `standalone_p2sh`, the Ironwood table is empty at that point,
since it is created empty and only the runtime scanner writes to it.

`store_address_range` matches its row with `AND key_scope = :foreign_scope`, which
restricts it to exactly the case leg 1 rules out.

## Resolution

Leaving one shielded pool out is wrong under either reading, so the choice was between
removing the shielded arms and adding Ironwood. **Ironwood is added**, with a comment
recording the invariant that makes all three shielded arms redundant.

The reasoning is asymmetric risk, not aesthetics. If the analysis above is right, the
added statement costs one no-op `UPDATE` on a path that only runs when an import is
upgraded across accounts. If it is wrong in some corner not found here, removing the
Sapling and Orchard arms would mis-attribute shielded funds to the wrong account, which
is silent and permanent. The existing code also plainly means "every pool that has a
received-note table"; Ironwood is the anomaly, and the same omission has already been a
real bug once in the sibling pool list in `flag_previously_received_change`.

The comment makes the redundancy deliberate rather than accidental, so a future reader
does not re-derive this analysis from scratch, and the tests below keep it honest.

### Schema-version safety

`store_address_range` is reachable from a migration (`transparent_gap_limit_handling.rs:147`),
so naming a table that does not yet exist there would be a hard error. It cannot happen:
the branch runs only when a `Foreign` row exists, which requires the
`imported_transparent_receiver_*` columns first added by `support_zcashd_wallet_import`,
which is a strict descendant of `transparent_gap_limit_handling` in the migration DAG.
`ironwood_received_notes` is downstream of that again. At the only migration-time call
site there are no `Foreign` rows at all and the branch is never entered; at runtime the
database is fully migrated.

## Tests

- `store_address_range_reattributes_shielded_notes_of_imported_receiver` seeds a
  `Foreign` record under account B with one directly-inserted note in each of the three
  shielded tables, has account A derive the address, and asserts all three move to
  account A. Verified to fail on Ironwood with the one-line change reverted
  (`ironwood_received_notes was not reattributed to the deriving account`), and to pass
  with it.
- `foreign_addresses_cannot_carry_a_diversifier_index` pins the invariant that makes the
  first test unreachable in production: the schema rejects a `Foreign` record carrying a
  diversifier index, and `upsert_address` produces an external-scoped record with one.
  This keeps the dead arms provably dead rather than quietly becoming load-bearing.

Both pass with `--all-features` and with `transparent-inputs,transparent-key-import`
(no `orchard`).

## Database write atomicity

The operation issues more than one write, so, per `AGENTS.md`:

1. **How many writes?** Per derived address that matches an import: one `UPDATE addresses`,
   plus (only when the import was under a different account) one `UPDATE` per pool table,
   now four instead of three. Otherwise one `INSERT`. This change adds one statement and
   does not change the shape.
2. **Must they land together?** Yes. A record upgraded to the deriving account whose
   outputs still carry the importing account's `account_id` is a wrong-balance state that
   the rest of the wallet would read as valid. They already do land together:
   `store_address_range` takes a `&rusqlite::Transaction` and never opens its own, so all
   of these run inside the caller's transaction and roll back as a unit.
3. **What happens on retry?** The operation is idempotent, and a partial commit is not
   reachable given (2). Re-deriving the same address finds the row already in its derived
   form, so the `Foreign` lookup no longer matches and the `INSERT ... ON CONFLICT DO
   NOTHING` arm is a no-op. If the row were somehow left half-updated, a retry would find
   no `Foreign` row and would not repair the notes, which is exactly why the single
   enclosing transaction is load-bearing.
4. **Does any read path see only one side?** Balance and note-selection queries read
   `*_received_notes.account_id` directly, while address-scoped queries join through
   `addresses.account_id`, so a partial write would be visible through one and not the
   other. Again, only reachable if the transaction were split.

No `WalletWrite` trait method changes, so no rustdoc atomicity sentence is added or
altered.

## No CHANGELOG entry

There is no user-visible change: the added statement cannot match a row in any state the
crate can produce. A `Fixed` entry would tell users something was broken for them, and a
`Changed` entry would describe an implementation detail invisible through the public API,
both of which `AGENTS.md` rules out. Happy to add one if you would rather record it.

## Open question for a maintainer, not addressed here

`ensure_address` builds the account's UA with `UnifiedAddressRequest::ALLOW_ALL`, so
`upsert_address` binds a `cached_transparent_receiver_address`. If a `Foreign` row already
holds that same transparent receiver, the `INSERT` conflicts on
`idx_addresses_cached_transparent_receiver_address`, which is not the `ON CONFLICT`
target, so the statement raises rather than upserting. That is a hard error on the scan
path rather than a mis-attribution, so it does not change the answer above, but it looks
like the shielded-side counterpart of the situation `store_address_range` handles on the
transparent side. Left alone deliberately to keep this PR to one concern.

### DIFF
--- a/zcash_client_sqlite/src/wallet/transparent.rs
+++ b/zcash_client_sqlite/src/wallet/transparent.rs
@@ -823,11 +823,23 @@ pub(crate) fn store_address_range<P: consensus::Parameters>(
 
             // If the import was recorded under a different account, the outputs received at
             // the address follow the (derivation-proven) attribution to this account.
+            //
+            // Only the transparent update can match today: a `Foreign` row is a transparent-only
+            // import, so no shielded note can be attached to it. A shielded note's `address_id`
+            // is only ever produced by `upsert_address`, which always writes the external key
+            // scope and a non-null diversifier index, whereas the `addresses` check constraint
+            // requires a `Foreign` row to have a null diversifier index; and no code path demotes
+            // an existing row to the `Foreign` scope. The shielded pools are nonetheless updated
+            // here so that this reattribution stays complete if that invariant is ever relaxed:
+            // mis-attributing a shielded note to the wrong account is silent and permanent,
+            // whereas these statements cost nothing on a path that only runs when an import is
+            // upgraded across accounts.
             if foreign_account != account_id.0 {
                 for table in [
                     "transparent_received_outputs",
                     "sapling_received_notes",
                     "orchard_received_notes",
+                    "ironwood_received_notes",
                 ] {
                     conn.execute(
                         &format!(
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK 9badf46c10ea675ade63b229300643b486dec2a4

PR #2820, opened 2026-07-28
zcash_client_sqlite: generate the migration DAG comment from DEPENDENCIES

### LINKED ISSUE #2819: zcash_client_sqlite: generate and check the migration DAG comment instead of hand-maintaining it
## Problem

The migration DAG in `zcash_client_sqlite/src/wallet/init/migrations.rs` is a hand-drawn ASCII comment. The source of truth is elsewhere: each migration module declares `const DEPENDENCIES: &[Uuid]`. Nothing keeps the two in step, and they have drifted.

Extracting the real DAG from the `DEPENDENCIES` declarations and comparing it against the comment shows three migrations that are **absent from the diagram entirely**:

- `orchard_ironwood_migration_tables` (a child of `ironwood_received_notes`)
- `tree_retained_checkpoints` (a child of `witness_stabilized_notes`)
- `ironwood_shardtree` (a child of `orchard_shardtree` and `wallet_summaries`)

`ironwood_received_notes` is drawn with two children and actually has four. None of this is caught by anything: the comment is prose, so no test, lint or type checks it.

`CURRENT_LEAF_MIGRATIONS` was checked by the same extraction and is correct, so the drift is confined to the diagram.

## Why it matters

The diagram is the only place a reader can see the shape of the migration graph. A reviewer adding a migration uses it to decide what to depend on, and a wrong picture is worse than no picture. This is also the third instance this week of a hand-maintained list that the compiler cannot see: see #2815, #2812 and #2813, all of which were per-pool lists rather than graphs, but the same failure mode.

## Rendering properties of the current diagram

Measured from the file rather than assumed, so a generator can reproduce or deliberately replace them:

- Comment prefix `    //`; node names on the header row start at column 12.
- **Maximum line width 118.** Every line in the pre-existing diagram is at or under this. There is no `rustfmt.toml`, and rustfmt does not reflow comments, so the budget is a convention rather than a tool constraint.
- Diagonals move exactly one column per row, `/` down-left and `\` down-right.
- A fan-out is drawn as `/ | \` on the row below the parent, children on the row after.
- Long names that do not fit beside their siblings hang from their own connector on a later row.

The binding constraint is width. Node names reach 37 characters (`orchard_ironwood_migration_tables`), so three siblings cannot share a row, and a depth-indented tree at two spaces per level reaches 148 columns because the graph is about 25 levels deep.

## Proposal

A tool that derives the diagram from `DEPENDENCIES` and either checks or rewrites the comment:

- `--check`: fail if the comment does not match the generated rendering, and separately report any migration missing from it and any mismatch between the computed leaves and `CURRENT_LEAF_MIGRATIONS`.
- `--write`: rewrite the comment block in place.
- `--mermaid` / `--dot`: for documentation, where a real layout engine can do a better job than ASCII.

Reproducing the existing hand-drawn art automatically is not realistic. The rendering should therefore be a canonical form chosen for determinism and completeness rather than beauty: a depth-indented tree using the primary parent, with the remaining edges listed separately so the tree stays inside the width budget.

## Open question for maintainers

Two directions, and this issue does not presume one:

1. **Check only.** Keep the art hand-drawn and have the tool verify that every migration and every edge appears. Preserves a diagram people like; the check is weaker, since it cannot verify the art's shape.
2. **Generate.** Replace the art with generated output. Stronger guarantee and no drift possible, at the cost of a less expressive picture.

Direction 2 also permits layout changes that direction 1 forbids, such as reordering leaf nodes so a branch can run past them, which is what the current art needs in order to fit the missing nodes.

## Where it should live

A Rust integration test under `zcash_client_sqlite/tests/` would run in existing CI with no new job and no new language dependency, and could regenerate under an env var. A standalone script is faster to write and easier to iterate on but needs a CI step. Worth deciding before this is finalised.

dannywillems:
Draft. Closes #2819.

## Problem

The migration DAG in `migrations.rs` is drawn by hand, while the graph it depicts is declared by each migration's `DEPENDENCIES`. Nothing kept the two in step, and they had drifted:

- `ironwood_shardtree`, `orchard_ironwood_migration_tables` and `tree_retained_checkpoints` were absent from the diagram entirely.
- `ironwood_received_notes` was drawn with two children when it has three.

`CURRENT_LEAF_MIGRATIONS` was checked by the same extraction and was already correct, so the drift was confined to the picture.

## Approach: check, do not generate

An earlier revision of this branch generated the diagram. That is the stronger guarantee, but the output is markedly worse to read than the hand-drawn art, so this takes the other direction from #2819: **the diagram stays hand-drawn and the tool verifies it.**

`zcash_client_sqlite/tools/migration_dag.py` reads the `DEPENDENCIES` of every migration module and checks:

- every migration appears somewhere in the diagram;
- every migration-looking name in the diagram is a real module, so a rename or deletion cannot leave a ghost behind;
- the leaves computed from the graph match `CURRENT_LEAF_MIGRATIONS`;
- no diagram line exceeds the 118-column budget (measured from the diagram itself; every pre-existing line was at or under it).

It also has `--mermaid` and `--dot` for documentation, where a real layout engine does better than ASCII.

### What it deliberately does not check

The shape of the drawing. Whether an edge is depicted, and whether it is depicted in the right place, is not recoverable from free-form ASCII art. This catches a migration that was **forgotten**, not one drawn in the wrong place. The script's docstring says so explicitly, so nobody mistakes its silence for a stronger guarantee than it gives. Generation is the only way to close that gap, and it costs the art.

## Diagram changes

Unchanged in style. `orchard_ironwood_migration_tables` is drawn as the third child of `ironwood_received_notes`, hanging from its own connector, because three names of that length do not fit on one row inside the budget.

The other two edges are recorded in a note beneath the drawing rather than drawn:

```
    // Two edges are not drawn above, for want of horizontal room. They are still
    // checked by tools/migration_dag.py:
    //   witness_stabilized_notes -> tree_retained_checkpoints
    //   orchard_shardtree, wallet_summaries -> ironwood_shardtree
```

Placing those two in the art needs a relayout of the surrounding structure, which would touch nodes this change has no other reason to move. A line saying they exist is better than a diagram that omits them silently. Happy to do the relayout instead if you would rather have them drawn.

## CI

A `migration-dag` job beside the existing `uuid` check, added to `required-checks` so drift blocks rather than merely reports. No new toolchain: it runs `python3`, already present on the runner, and needs no cargo build.

If you would rather not add Python to a pure-Rust CI, the alternative is a Rust integration test under `tests/` doing the same parsing, which would fold into the existing jobs. Say the word and I will port it.

## Verification

- `tools/migration_dag.py zcash_client_sqlite`: passes (57 migrations, 8 leaves, widest line 118 of 118)
- Negative test: deleting one recorded edge makes it fail with `ironwood_shardtree ... missing from the diagram` and exit 1
- `cargo fmt --all -- --check`: clean
- The workflow YAML parses

## Note on history

Two commits: the first took the generation approach, the second replaces it with this one. I can squash them into a single commit if you would prefer the branch to read as one change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -549,6 +549,16 @@ jobs:
             uuidparse -n -o uuid $U4 | sort | uniq | wc -l
           )
 
+  migration-dag:
+    name: Migration DAG comment
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Check the migration DAG comment against each migration's DEPENDENCIES
+        run: python3 zcash_client_sqlite/tools/migration_dag.py zcash_client_sqlite
+
   required-checks:
     name: Required status checks have passed
     needs:
@@ -563,6 +573,7 @@ jobs:
       - fmt
       - protobuf
       - uuid
+      - migration-dag
     if: ${{ always() }}
     runs-on: ubuntu-latest
     steps:
--- a/zcash_client_sqlite/src/wallet/init/migrations.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations.rs
@@ -89,73 +89,91 @@ pub(super) fn all_migrations<
     rng: R,
     seed: Option<Rc<SecretVec<u8>>>,
 ) -> Vec<Box<dyn RusqliteMigration<Error = WalletMigrationError>>> {
-    //                                   initial_setup
-    //                                   /           \
-    //                          utxos_table         ufvk_support
-    //                             |                 /         \
-    //                             |    addresses_table   sent_notes_to_internal
-    //                             |          /                /
-    //                           add_utxo_account             /
-    //                                        \              /
-    //                                     add_transaction_views
-    //                                               |
-    //                                       v_transactions_net
-    //                                               |
-    //                                            received_notes_nullable_nf---------------------.
-    //                                            /           |                                   \
-    //                                           /            |                                    \
-    //           ,-------------- shardtree_support    sapling_memo_consistency                    nullifier_map
-    //          /                     /           \                       \                                  |
-    // orchard_shardtree   add_account_birthdays   receiving_key_scopes   v_transactions_transparent_history |
-    //   |                    |                 \            |                     |                         |
-    //   |   v_sapling_shard_unscanned_ranges    \           |       v_tx_outputs_use_legacy_false           |
-    //   |                    |                   \          |                     |                         |
-    //   |            wallet_summaries             \         |      v_transactions_shielding_balance         /
-    //   \                    \                     \        |                     |                        /
-    //    \                    \                     \       |      v_transactions_note_uniqueness         /
-    //     \                    \                     \      |        /                                   /
-    //      \                    `------------------- full_account_ids                                   /
-    //       \                                        /               \                                 /
-    //        \                         orchard_received_notes        spend_key_available              /
-    //         \                            /          \                      /                       /
-    //          \     ensure_orchard_ua_receiver     utxos_to_txos           /                       /
-    //           \                          \              |                /                       /
-    //            \                          \     ephemeral_addresses     /                       /
-    //             \                          \            |              /                       /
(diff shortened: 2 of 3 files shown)


nuttycom:
Instead of this kind of check, I think that I would like to regenerate the entire migration DAG via Mermaid.

PR #2822, opened 2026-07-28
zcash_client_sqlite: classify pool-crossing transactions in v_transactions

### LINKED ISSUE #2821: v_transactions has no classification or amount for wallet-internal pool-crossing transactions
For a wallet-internal transfer that moves an account's own funds between shielded pools — a ZIP 318 Orchard → Ironwood migration transfer, or a user-initiated payment to the wallet's own address routed across the turnstile — `v_transactions.account_balance_delta` is just the negated fee. That is correct, but it is not the quantity a wallet wants to display: the user-meaningful amount is the value that crossed pools.

Nothing in `v_transactions` identifies such transactions (there is no analogue of `is_shielding`) or carries the crossed amount, so SDK clients can only improvise from `total_spent` or `total_received`, both of which overstate the crossing whenever the transaction also returns change to a pool it spent from. This contributed to wallets displaying an Orchard → Ironwood transaction's value as the total of the notes spent rather than the pool-crossing amount (see zcash/zcash-swift-wallet-sdk#1899 for the client-side half).

`v_transactions` should classify these transactions and expose the crossed value, so that clients can display it directly.

---

nuttycom:
Closes #2821. Resolves MOB-1599.

For a wallet-internal transfer that moves an account's own funds between shielded pools — a ZIP 318 Orchard → Ironwood migration transfer, or a user-initiated payment to the wallet's own address routed across the turnstile — `account_balance_delta` is just the negated fee, which is correct but not the quantity a wallet wants to display. Nothing in `v_transactions` identified such transactions or carried the crossed amount, so clients could only improvise from `total_spent` or `total_received`, both of which overstate the crossing whenever the transaction also returns change to a pool it spent from.

This PR adds two columns to `v_transactions`, mirroring the shape of the existing `is_shielding` classification:

- `is_pool_crossing`: true when every wallet-spent note and wallet-received output in the transaction is shielded, the account spent at least one note, at least one output was received in a pool the account spent nothing from, and no external outputs of the transaction are known.
- `pool_crossing_value`: the total value received in pools the account did not spend from — the amount that crossed — when `is_pool_crossing` is true; NULL otherwise.

The implementation adds a `spent_note_pools` CTE (the distinct pools each account spent from, per transaction) and a per-pool LEFT JOIN, leaving the view's existing aggregation structure untouched. The view is recreated by a new `v_transactions_pool_crossing` migration depending on `account_delete_cascade` (the prior owner of the view definition) and `ironwood_pool_code_views` (the Ironwood arms of the underlying views).

One timing property to be aware of: a payment that returns value to one of the wallet's own addresses is classified once the wallet has observed the returned output — the scanner marks an output received by an account that funded the transaction as change, at which point no external outputs remain. While such a transaction is unmined, its returned output is not yet known and it is treated as an ordinary payment.

The first commit adds regression coverage for the transaction-history value accounting of every Orchard → Ironwood transaction shape (external payment, payment to the wallet's own receiver, and mined ZIP 318 migration transfers — the proving chain simulation previously extracted transfers but never mined them). The second commit adds the classification, with the chain simulation asserting that each mined transfer is classified with its crossing value, and the two payment tests pinning down the not-classified and classified-once-observed cases.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -8570,12 +8570,12 @@ pub fn orchard_to_ironwood_payment_reports_net_value_delta<Dsf>(
             assert!(tx.has_change(), "the transaction has change ({})", $phase);
             // The payment has an external recipient, so it is not a pool crossing even though
             // its value moves from the Orchard pool into the Ironwood pool.
-            assert!(
-                !tx.is_pool_crossing(),
+            assert_eq!(
+                tx.pool_crossing_value(),
+                None,
                 "a payment to an external recipient is not a pool crossing ({})",
                 $phase,
             );
-            assert_eq!(tx.pool_crossing_value(), None);
         }};
     }
 
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -8457,7 +8457,7 @@ where
 
 /// The transaction history entry for a payment funded from the Orchard pool and delivered
 /// through the Ironwood pool reports the amount that left the account (payment plus fee) as
-/// its balance delta — not the total value of the notes spent, which would ignore the change
+/// its balance delta, not the total value of the notes spent, which would ignore the change
 /// returned to the wallet.
 #[cfg(feature = "orchard")]
 pub fn orchard_to_ironwood_payment_reports_net_value_delta<Dsf>(
@@ -8466,7 +8466,7 @@ pub fn orchard_to_ironwood_payment_reports_net_value_delta<Dsf>(
 ) where
     Dsf: DataStoreFactory,
 {
-    // A network on which NU6.3 — the version 6 transaction format — is active from height 100_000.
+    // A network on which NU6.3, the version 6 transaction format, is active from height 100_000.
     let ironwood_active_network = {
         let activation = BlockHeight::from_u32(100_000);
         LocalNetwork {
@@ -8576,15 +8576,15 @@ pub fn orchard_to_ironwood_payment_reports_net_value_delta<Dsf>(
         }};
     }
 
-    // The history entry is correct as soon as the transaction is stored…
+    // The history entry is correct as soon as the transaction is stored...
     check_history!("before mining");
 
-    // …and remains correct once the transaction is mined and scanned…
+    // ...and remains correct once the transaction is mined and scanned...
     let (h, _) = st.generate_next_block_including(txid);
     st.scan_cached_blocks(h, 1);
     check_history!("after mining");
 
-    // …and remains correct after transaction enhancement, in which the wallet retrieves the
+    // ...and remains correct after transaction enhancement, in which the wallet retrieves the
     // full transaction and re-stores it via `decrypt_and_store_transaction` (as the mobile
     // SDKs do to recover memos and fee information).
     let tx = st
@@ -8597,8 +8597,8 @@ pub fn orchard_to_ironwood_payment_reports_net_value_delta<Dsf>(
     check_history!("after enhancement");
 }
 
-/// A migration of the wallet's own funds from the Orchard pool to the Ironwood pool — a payment
-/// to the wallet's own external Orchard receiver, delivered through the Ironwood pool — leaves
+/// A migration of the wallet's own funds from the Orchard pool to the Ironwood pool, namely a
+/// payment to the wallet's own external Orchard receiver delivered through the Ironwood pool, leaves
 /// the account balance unchanged except for the fee. The payment output returned to the wallet
 /// is not recorded at transaction-creation time (it is an address payment, which the wallet
 /// expects to detect by scanning), so this exercises detection of the wallet's own Ironwood
@@ -8610,7 +8610,7 @@ pub fn orchard_to_ironwood_self_migration_reports_fee_only_delta<Dsf>(
 ) where
     Dsf: DataStoreFactory,
 {
-    // A network on which NU6.3 — the version 6 transaction format — is active from height 100_000.
+    // A network on which NU6.3, the version 6 transaction format, is active from height 100_000.
     let ironwood_active_network = {
         let activation = BlockHeight::from_u32(100_000);
         LocalNetwork {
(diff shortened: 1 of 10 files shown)


zcash_client_sqlite/src/wallet/init/migrations/v_transactions_pool_crossing.rs
```diff
@@ -0,0 +1,235 @@
...
+//! example a ZIP 318 Orchard → Ironwood migration transfer — has an `account_balance_delta` of
+//! only the transaction fee, which is correct but is not the quantity a wallet wants to present:
+//! the user-meaningful amount is the value that crossed pools. Nothing in the view identified
+//! such transactions or carried that amount, so clients could only improvise from `total_spent`
+//! or `total_received`, both of which overstate the crossing whenever the transaction also
+//! returns change to a pool it spent from.
+//!
+//! This migration recreates `v_transactions` with two new columns, mirroring the shape of the
+//! existing `is_shielding` classification:
+//! - `is_pool_crossing`: true when every wallet-spent note and wallet-received output in the
```
dannywillems: I guess the two values are complementary. `is_pool_crossing = true iff pool_cross_value is not NULL` and `is_pool_crossing = false iff pool_cross_value is NULL`.

zcash_client_backend/src/data_api/testing/pool.rs
```diff
@@ -7920,7 +7922,6 @@ pub fn propose_v5_payment_to_orchard_receiver_is_rejected<Dsf>(
 ) where
     Dsf: DataStoreFactory,
 {
-    use super::orchard::OrchardPoolTester;
     use crate::data_api::wallet::{input_selection::SpendPolicy, propose_transfer};
```
dannywillems, line 7925: I would like to move all the imports at the top.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1427,6 +1467,10 @@ SELECT accounts.uuid                AS account_uuid,
             -- We do not know about any external outputs of the transaction.
             AND MAX(COALESCE(sent_note_counts.sent_notes, 0)) = 0
        ) AS is_shielding,
+       -- The value that crossed pools, when this transaction is a wallet-internal transfer
+       -- between shielded pools; NULL when it is not such a transfer. A transaction is one
+       -- exactly when this column is non-NULL.
+       pool_crossings.crossing_value AS pool_crossing_value,
```
dannywillems, line 1473: That's the column sent back to the user.

zcash_client_sqlite/tests/pool_migration_prove_chain_sim.rs
```diff
@@ -439,12 +439,85 @@ impl Run {
...
                 scenario.label
             );
-            ironwood_notes.push(
+            let crossing_value =
                 Zatoshis::from_u64(i64::from(ironwood.value_balance()).unsigned_abs())
-                    .expect("a valid Ironwood note value"),
+                    .expect("a valid Ironwood note value");
+            ironwood_notes.push(crossing_value);
+
+            // The transfer's Orchard bundle is spend-only, so its value balance is the spent
```
dannywillems, line 447: We should create more helpers in the DSL to reuse them here, and have shorter bodies for tests.

zcash_client_sqlite/src/wallet/db.rs
```diff
@@ -1393,6 +1404,35 @@ sent_note_counts AS (
...
+-- Identifies the transactions that are wallet-internal transfers moving an account's own
+-- funds between shielded pools, and reports the value that crossed. `crossing_value` is
+-- non-NULL exactly for such a transaction, so it carries both the classification and the
+-- amount; see the `pool_crossing_value` column below.
+pool_crossings AS (
+    SELECT notes_by_pool.account_id     AS account_id,
+           notes_by_pool.transaction_id AS transaction_id,
+           CASE WHEN (
+                -- Every note spent and every output received by the wallet is shielded.
+                SUM(CASE WHEN notes_by_pool.pool = 0 THEN notes_by_pool.spent_note_count + notes_by_pool.received_note_count ELSE 0 END) = 0
```
dannywillems, line 1416: Could we extract this in a helper? It was hard for me to read.

dannywillems, line 1416: We could add a view like for `notes_by_pool`.

I'm opening follow-ups for these.

zcash_client_sqlite/tests/pool_migration_prove_chain_sim.rs
```diff
@@ -439,12 +439,85 @@ impl Run {
...
+            let history = self
+                .st
+                .wallet()
+                .get_tx_history()
+                .expect("reads the transaction history");
+            let entry = history
+                .iter()
+                .find(|t| t.txid() == tx.txid())
+                .expect("the mined transfer appears in the transaction history");
+            assert_eq!(
```
dannywillems, line 477: Are these checks really important?

zcash_client_sqlite/tests/pool_migration_prove_chain_sim.rs
```diff
@@ -439,12 +439,85 @@ impl Run {
...
+                "{}: transfer total_received",
+                scenario.label
+            );
+            assert_eq!(
+                entry.account_value_delta(),
+                -ZatBalance::from(transfer_fee),
+                "{}: transfer balance delta is only the fee",
+                scenario.label
+            );
+            // The transfer is classified as a pool crossing carrying the crossing value, which
```
dannywillems, line 495: That's the check we actually care.

zcash_client_sqlite/tests/pool_migration_prove_chain_sim.rs
```diff
@@ -439,12 +439,85 @@ impl Run {
...
+            // is the amount a wallet should display for it (the balance delta is only the fee).
+            assert!(
+                entry.is_pool_crossing(),
+                "{}: the transfer is a pool crossing",
+                scenario.label
+            );
+            assert_eq!(
```
dannywillems, line 502: And this one.

zcash_client_sqlite/tests/pool_migration_prove_chain_sim.rs
```diff
@@ -439,12 +439,85 @@ impl Run {
...
+                entry.pool_crossing_value(),
+                Some(crossing_value),
+                "{}: transfer pool_crossing_value",
+                scenario.label
             );
         }
 
+        // After every transfer is mined and scanned, the wallet holds the migrated value (now in
```
dannywillems, line 510: Sure, that can be added.

dannywillems (approved):
utACK 8af3e41556a16cca47b5d32ed1d529c48f73cf6e, modulo comments for follow-up.

zcash_client_sqlite/src/wallet/init/migrations/v_transactions_pool_crossing.rs
dannywillems: This is an important change from the initial state of @nuttycom

PR #2823, opened 2026-07-28
zcash_client_sqlite: anchor points for external migrations

nuttycom:
External migrations registered via `WalletMigrator::with_external_migrations` are applied as part
of a single graph alongside the internal ones, so an external migration that reads or extends
internal schema must order itself after the migration that creates that schema. The anchors this
crate offers for that are the per-release leaf constants — but they stopped at 0.19.0, so nothing
added since had any published state to anchor against, including every Ironwood table.

Two commits, addressing that from both ends.

## Add the leaf sets for the releases since 0.19.0

Release candidates are published crate versions, and a wallet built against one is as real as any
other, so they get constants too. Three states have been exposed since 0.19.0:

| Constant | Covers | Change |
| --- | --- | --- |
| `V_0_20_0` | 0.20.0 – 0.21.1 | `v_tx_outputs_key_scopes`, `ivk_item_cache`, `witness_stabilized_notes` |
| `V_0_22_0_RC1` | 0.22.0-rc.1 | adds the two transparent indexes, `ironwood_pool_code_views` and `tree_retained_checkpoints`; `witness_stabilized_notes` gains dependents |
| `V_0_22_0_RC2` | 0.22.0-rc.2 – rc.4 | adds `orchard_ironwood_migration_tables` and `note_locking` |

Versions whose migration state is unchanged relative to the preceding constant are omitted, which
is the convention the module documentation already records.

The leaves were computed from each published version's own migration graph, by the rule
`current_leaf_migrations_are_the_dag_leaves` uses: every migration that no other migration in that
graph depends on. `migrate_between_releases_without_data` then walks the new states in order and
asserts the migration state changes at each, exercising them against the real migrator.

Note that `0.22.0-rc.1`, `-rc.3` and `-rc.4` are published on crates.io but have no git tag; only
`zcash_client_sqlite-0.22.0-rc.2` is tagged. Their leaf sets here were computed from the published
crate sources. It made no difference to the outcome — rc.3 and rc.4 match rc.2 — but the missing
tags look like a release-process slip worth fixing separately.

## Expose the identifier of each internal migration, behind `unstable`

A release constant cannot name schema added since the most recent release, so an external
migration that depends on such schema still has no correct dependency to declare: it either
anchors too early, or hard-codes a private identifier that would resolve to
`DependencyError::UnknownId` — a wallet that fails to open — if that migration were reshaped
before release.

Each migration's `MIGRATION_ID` becomes public within its still-private module, re-exported
through a new `migrations::ids` facade:

```rust
use zcash_client_sqlite::wallet::init::migrations::ids;

fn dependencies(&self) -> HashSet<Uuid> {
    [ids::IRONWOOD_RECEIVED_NOTES].into_iter().collect()
}
```

The facade sits behind the `unstable` feature, because these identifiers are unstable by nature:
each names an individual migration rather than a state of the graph that a published release
exposed, so the set of them changes between releases. The intended shape is to depend on an
identifier while developing against unreleased schema, and to move the anchor to the release
constant that covers it once that release exists — which is what the first commit makes possible
for everything released so far.

A facade rather than 59 public modules keeps the public names decoupled from the internal module
names, keeps rustdoc to one extra module, and gives a single place to document how to choose an
anchor. `#![deny(missing_docs)]` means each identifier also carries a summary of what its
migration does; five already had hand-written docs and keep them, and the rest take the first
sentence of their module's `//!` doc, or of `description()` for the two modules that have none.

`ids_module_covers_every_migration` lists the exported identifiers explicitly and compares them
against the graph, in the same style as `current_leaf_migrations_are_the_dag_leaves`. Listing
rather than deriving is deliberate: a migration added without an `ids` entry fails the test rather
than silently going unexported.

---

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -11,6 +11,15 @@ workspace.
 ## [Unreleased]
 
 ### Added
+- `zcash_client_sqlite::wallet::init::migrations::ids`, behind the `unstable`
+  feature flag, exposing the identifier of each individual internal migration.
+  External migrations registered via `WalletMigrator::with_external_migrations`
+  can be anchored against a specific internal migration where the per-release
+  constants (such as `migrations::V_0_19_0`) are not precise enough — in
+  particular when depending on schema added since the most recent release. These
+  identifiers are unstable: the set of them changes between releases, so they are
+  intended for development against unreleased schema, with the anchor moved to
+  the release constant that covers it once that release exists.
 - `zcash_client_sqlite::error::SqliteClientError::BackendError` and the
   `zcash_client_sqlite::error::BackendError` it wraps, reported when a
   `zcash_client_backend` error value carries a variant this crate has no
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -11,6 +11,12 @@ workspace.
 ## [Unreleased]
 
 ### Added
+- `zcash_client_sqlite::wallet::init::migrations::V_0_20_0`, `V_0_22_0_RC1` and
+  `V_0_22_0_RC2`, the leaf migration sets for the releases published since
+  0.19.0, so that an external migration can anchor against any published state
+  of the migration graph. Release candidates are published crate versions and so
+  are covered here; versions that did not change the migration state relative to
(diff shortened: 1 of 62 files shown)


schell (approved):

nuttycom (commented):
self-utACK

PR #2824, opened 2026-07-28
zcash_client_sqlite: correct the v_tx_outputs_key_scopes description

nuttycom:
`v_tx_outputs_key_scopes`'s `description` is a copy of `account_delete_cascade`'s, so the migrator
reports that this migration adds `ON DELETE CASCADE` to foreign keys to support account deletion.

It adds missing key scope information to the `v_received_outputs` and `v_tx_outputs` views, as its
module documentation says.

---


nuttycom (commented):
doc-only self-ACK

PR #2825, opened 2026-07-28
Limit transaction status requests to fully-transparent transactions

pacu:
## Summary

`TransactionDataRequest::GetStatus` is a query by transaction id, and it exists for transactions that ordinary compact-block scanning cannot detect — i.e. fully-transparent ones. A transaction with any shielded component is detected by scanning, which sets its mined height, so a status request for it is redundant and out of scope. Two paths did not hold to that scoping:

- `transaction_data_requests` synthesized a `GetStatus` request for **every** transaction whose mined height was unknown, regardless of its bundles (the fallback added in 406875adc, first shipped in `zcash_client_sqlite` 0.18.5).
- `queue_tx_retrieval` created `Status`-type queue entries for transactions with shielded components whenever raw transaction data was already present — which happens for scanned wallet transactions and for transparent input prevouts (e.g. the shielded first step of a ZIP 320 pair).

Additionally, two places classified scannability from Sapling and Orchard bundles only, omitting Ironwood. Since NU6.3 carries payments involving Orchard receivers in the Ironwood bundle, ordinary transactions were affected.

## Changes

- `transaction_data_requests` now reads exclusively from `tx_retrieval_queue`; the blanket fallback over unmined transactions is removed.
- Because the queue is now the sole source of requests, `set_transaction_status` retains a `Status`-type queue entry when a not-yet-mined answer is not terminal, so an unresolved fully-transparent transaction keeps being queried. Terminal means: mined; or confirmed unmined at or beyond a known nonzero expiry height; or, for an unknown expiry height, confirmed unmined at or beyond `min_observed_height + PRUNING_DEPTH + DEFAULT_TX_EXPIRY_DELTA`. Expiry height 0 is polled indefinitely. This mirrors the semantics the removed clause encoded.
- `queue_tx_retrieval` enforces the scoping directly: when raw transaction data is available it is parsed, and no request is created for a transaction with a Sapling, Orchard, or Ironwood bundle, nor for one already known to be mined.
- `zcash_client_backend`'s `store_decrypted_tx` and the `tx_retrieval_queue` migration now count Ironwood bundles when classifying scannability, matching `store_transaction_to_be_sent`.
- A new migration removes previously-queued status-check entries for transactions with shielded components from existing wallet databases.

## Tests

- `shielded_send_generates_no_status_requests` (per pool): an unmined shielded send produces no `GetStatus` or `Enhancement` request at any chain tip height, including past expiry, and scanning alone sets its mined height.
- `transparent_status_requests_persist_until_terminal` (per pool): a fully-transparent ZIP 320 second-step transaction is queued and keeps being polled across not-yet-mined answers, stopping only once mined or confirmed unmined at its expiry height; the shielded first step is not queued.
- `status_request_retention_terminal_conditions`: the terminal-condition matrix for expiry height 0, known nonzero, and unknown expiry heights.
- Migration tests covering that shielded and unsatisfiable entries are purged while fully-transparent `Status` entries and all `Enhancement` entries are preserved.

`cargo test -p zcash_client_sqlite --all-features` (485 passed) and `cargo test -p zcash_client_backend --all-features` (119 passed) are green, as are `cargo fmt --all -- --check` and `cargo clippy --all-features --all-targets -- -D warnings`.

A backport of these commits onto the `zcash_client_sqlite 0.22.0-rc.4` release commit is available at `pacu/status-request-scoping-0.22.0-rc.4` if an interim branch is useful.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -32,6 +32,15 @@ workspace.
   `tor::grpc::GrpcError`, and `tor::http::HttpError`.
 
 ### Fixed
+- `data_api::ll::wallet::store_decrypted_tx` now classifies a transaction
+  containing an Ironwood bundle as detectable via chain scanning. Previously
+  only Sapling and Orchard bundles were considered, so an as-yet-unmined
+  Ironwood-only transaction stored via transaction enhancement was queued for
+  txid-based status retrieval, even though ordinary compact-block scanning
+  detects such a transaction and sets its mined height, making the status
+  request redundant. Since NU6.3 every
+  shielded spend or output involving an Orchard receiver is carried in the
+  Ironwood bundle, so this affected ordinary transactions.
 - The Tor HTTP and gRPC transports now reject a URL whose scheme is neither
   `http` nor `https` (for example `ftp` or `ws`) with `tor::http::HttpError::NonHttpUrl`.
   Previously such a URL was silently treated as plaintext HTTP, contradicting the
--- a/zcash_client_backend/src/data_api/ll/wallet.rs
+++ b/zcash_client_backend/src/data_api/ll/wallet.rs
@@ -1088,6 +1088,9 @@ where
         #[cfg(feature = "orchard")]
         let detectable_via_scanning =
             detectable_via_scanning | d_tx.tx().orchard_bundle().is_some();
+        #[cfg(feature = "orchard")]
+        let detectable_via_scanning =
+            detectable_via_scanning | d_tx.tx().ironwood_bundle().is_some();
 
         if d_tx.mined_height().is_none() && !detectable_via_scanning {
             wallet_db.queue_tx_retrieval(std::iter::once(d_tx.tx().txid()), None)?
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -3852,6 +3852,238 @@ where
     );
 }
 
+/// A wallet's own shielded send must never generate a `GetStatus` transaction data request, at
+/// any chain tip height: its mined status is learned via ordinary chain scanning, and a
+/// txid-scoped status query for it is redundant; status requests are limited to
+/// fully-transparent transactions, which scanning cannot detect.
+pub fn shielded_send_generates_no_status_requests<T: ShieldedPoolTester, Dsf: DataStoreFactory>(
+    dsf: Dsf,
+    cache: impl TestCache,
+) {
+    use zcash_primitives::transaction::builder::DEFAULT_TX_EXPIRY_DELTA;
+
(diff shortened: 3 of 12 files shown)


nuttycom (changes requested):
The new invariant that any transaction containing a shielded bundle is detectable via compact-block scanning does not hold, and making `tx_retrieval_queue` the sole source of status requests also loses tracking across reorgs. I think both issues are blocking.

1. **Transparent-funded sends to an external shielded receiver are not necessarily detectable by scanning.** Compact scanning only constructs a wallet transaction when it matches a wallet-owned shielded nullifier or decrypts a wallet-owned shielded output. Bundle presence alone is insufficient. In particular, `shielding_coinbase_to_orchard_receiver_delivers_via_ironwood` spends only wallet transparent inputs, sends the sole Ironwood output to a separate spending key, and has no change. Its Ironwood spend side is dummy padding, so the sender can neither match a real wallet nullifier nor decrypt the output. The assertion added to that test that no status request should exist therefore removes the only mechanism by which the sending wallet can learn the mined outcome. The same problem applies to transparent-to-Sapling/Orchard/Ironwood payments whose shielded outputs are entirely external. If such a transaction is mined, the wallet can eventually misclassify it as expired-unmined, or leave it pending indefinitely when expiry is zero. The migration also deletes existing status entries for these transactions based solely on bundle presence.

   Please base `detectable_via_scanning` on an actual wallet-observable shielded spend or output, rather than `*_bundle().is_some()`. A useful regression test would mine and scan the external-Ironwood shielding transaction and assert that its `mined_height` is updated; as currently constructed, compact scanning should not detect it.

2. **A reorg permanently loses status tracking for fully-transparent transactions.** `set_transaction_status(..., Mined(...))` deletes the queue entry. `truncate_to_height` can later clear that transaction’s `mined_height`, but it does not recreate the queue entry. Previously, the fallback query over unmined transaction rows regenerated `GetStatus`; after this PR, the queue is the sole request source. Because a fully-transparent transaction is invisible to compact scanning, the wallet can no longer learn whether it is re-mined on the replacement chain.

   Please preserve or recreate the status-tracking intent across rewinds. One possible design is to retain dormant status entries for mined transactions while filtering them from emitted requests until `mined_height` becomes null again. Please add a test that marks a fully-transparent transaction mined, rewinds below that height, and verifies that `GetStatus` is emitted again.

nuttycom:
I have an alternate implementation that an agent is going to propose.

schell:
Huh - I figured I would get tests passing here before reviewing, but if you're going to present an alternative I'll just close mine.

nuttycom:
Opened https://github.com/zcash/librustzcash/pull/2828 as an alternative.

nuttycom:
> Huh - I figured I would get tests passing here before reviewing, but if you're going to present an alternative I'll just close mine.

For this sort of thing, just take over the PR, and even rebase/force-push - we'd rather have the clean commit history and not wait for a round-trip for reviewing just the fixes.

nuttycom:
Closed in favor of #2828

PR #2826, opened 2026-07-28
Make manual Orchard -> Ironwood crossings indistinguishable from migration transfers

### LINKED ISSUE #2725: zcash_client_backend: Make manual ironwood->orchard transfers indistinguishable from automated ones.
In order to reduce the amount of data that must be transferred to and from a hardware wallet device for the Orchard -> Ironwood migration,  we do not require Ironwood to be padded to two actions. However, this makes automated pool-crossing transfers explicitly distinguishable from manual ones, which have two Ironwood actions by default.

In the case that an input note(s) are old enough to be witnessed against an anchor that satisfies the ZIP 318 anchor bucketing criterion, and the pool-crossing value is in the set of valid ZIP 318 quantized values, we should emit only a single Ironwood output and should use a ZIP 318 bucketed anchor whenever possible in manually generated transactions.

nuttycom:
Closes #2725.

When a manual payment crosses the Orchard turnstile and happens to look exactly like a ZIP 318 migration transfer, this makes it *be* one: a single unpadded Ironwood action against a bucketed anchor, rather than the two-action default that marks it as manual.

## What qualifies

A step is a *canonical crossing* only when its whole shape matches a migration transfer. Every condition is required:

- exactly one Orchard input, so the Orchard bundle is two actions;
- no Ironwood spends (a payment funded from Ironwood crosses nothing);
- no Ironwood change, leaving a single Ironwood output;
- that output's value is a canonical `{1, 2, 5} * 10^k` denomination within `[0.01 ZEC, 10,000 ZEC]`;
- the anchor lies on the wallet's ZIP 318 bucket grid.

Dropping the padding on a weaker test would not buy anonymity but destroy it: an unpadded Ironwood bundle beside an Orchard bundle of the wrong size, or against a chain-tip anchor, is a shape nothing else emits.

## How it is decided

`propose_transfer` decides *before* keeping a proposal, rather than discovering by failure. It attempts selection restricted to the Orchard pool under a bucketed confirmations policy, and keeps the result only if `Step::is_canonical_crossing` holds; otherwise it re-proposes normally.

Two things drove that design:

- The common miss is not insufficient funds but a **successful multi-note selection**, which funds fine yet is not canonical. A fallback triggered only by insufficient funds would leave such a transaction carrying a bucketed anchor — up to one interval of extra confirmations — while still being built padded.
- Restricting to Orchard is **required, not preferred**. The ordinary selector avoids crossing pools, so it funds an Ironwood-destined payment from Ironwood once the wallet holds any. Without the restriction, a user self-migrating by sending themselves canonical amounts stalls at whatever the first crossing moved.

Bucketing is expressed as a raised confirmation requirement rather than a separately lowered anchor, because the anchor and the bound on which notes are spendable are the same quantity (`target_height` less required confirmations). Moving one number moves both, so a note without a witness at the bucketed anchor cannot be selected.

## Structure

1. `zcash_protocol::zip318` (**purely additive**) — the ZIP 318 constants, the denomination predicate, the anchor bucket interval, and an unsealed `PoolMigrationConstants` trait with ZIP-valued defaults.
2. `zcash_pool_migration` and `zcash_client_backend` adopt it, collapsing the interval type they duplicated. This removes a proptest that existed solely to catch drift between the two implementations — one type cannot drift from itself.
3. The send path: Ironwood padding derived from the proposal, and the bucketed-anchor proposal above.

The wallet is the single authority on the grid (`WalletRead::pool_migration_params`). Reading it from the network defaults in some places and the wallet in others agrees on mainnet by coincidence and diverges for any wallet with a custom retention interval.

## Notes for review

- `ChangeStrategy::compute_balance` and `InputSelector::{propose_transaction, propose_shielding}` gain `anchor_height` and `zip318` parameters; `Step::{orchard_action_count, ironwood_action_count}` take `BundlePadding` instead of `BundleType`; `with_unpadded_orchard_pool_bundles` is removed. All in the CHANGELOGs.
- The last commit fixes feature-flagging that was broken independently of this work, including a pre-existing `cargo clippy -p zcash_client_sqlite --all-targets` failure on `main`.
- Verified: `fmt --check`, default and `--all-features` clippy with `-D warnings`, rustdoc link check, `cargo build -p zcash_protocol --no-default-features`, and the full test suite.


components/zcash_protocol/src/zip318.rs
```diff
@@ -0,0 +1,452 @@
...
+
+    /// Returns whether `value` is a canonical crossing denomination under these parameters: a
+    /// `{1, 2, 5} * 10^k` amount within
+    /// `[max_residual_value, denomination_cap]`.
+    fn is_canonical_denomination(&self, value: Zatoshis) -> bool {
+        is_canonical_within(value, self.max_residual_value(), self.denomination_cap())
+    }
+}
+
+impl PoolMigrationConstants for NetworkType {}
```
nuttycom: This uses the mainnet defaults for `TestNetwork`.

ValarDragon (dismissed):
utACK Action count + anchor interval look right to me!

TODO still: Expiry height

zcash-harry:
Would `create_pczt_from_proposal `needs the same `is_canonical_crossing check `?

nuttycom (commented):
re-utACK

nullcopy (approved):
utACK

Issue #2525, opened 2026-07-02
pczt: make the shielded bundle anchor deferrable in v6 PCZTs (anchor is authorizing data)


nuttycom:
## Motivation

In the v6 transaction format the Sapling/Orchard/Ironwood anchors are authorizing data (ZIP 229): the txid digest — and therefore the sighash — omits them ([`includes_anchor_in_txid_digest` is true only for `OrchardV5`](https://github.com/zcash/orchard/blob/840cd68e99f2/src/bundle/commitments.rs#L98-L108)). So for v6 the anchor can be chosen **after** signatures are collected — at broadcast time when re-anchoring to the canonical bucket anchors of the migration flow (zcash/zips#1317), or when spending outputs of a parent transaction that had not been mined at signing time (#2524).

## Current behavior

The PCZT data model fixes the anchor at creation:

- `Bundle::anchor: [u8; 32]` is required and documented "Set by the Creator" ([`pczt/src/orchard.rs#L58-L62`](https://github.com/zcash/librustzcash/blob/eb828caadb4b/pczt/src/orchard.rs#L58-L62));
- the Combiner refuses to merge copies whose anchors differ ([`pczt/src/orchard.rs#L673`](https://github.com/zcash/librustzcash/blob/eb828caadb4b/pczt/src/orchard.rs#L673));
- the Signer and IO Finalizer read the field when assembling `TransactionData` for sighash computation (harmless for v6, where the value does not reach the sighash, but the field must be present and parseable).

## Proposed change

- Make the anchor deferrable for v6 PCZTs: `anchor: Option<[u8; 32]>`, "set by the Updater before proving" (companion Updater issue: #2526).
- PCZT serialization format version bump; the v1/v2 decoders map the existing required field to `Some`.
- Combiner: `None` merges with `Some` (same semantics as the other deferred fields); unequal `Some` values still conflict.
- Signer / IO Finalizer: tolerate `None` for v6 (any placeholder is sound for sighash assembly); reject a missing anchor for v5, where the sighash commits to it.
- Sapling parallel: v6 also makes `anchorSapling` authorizing data, so the same deferral applies to the Sapling bundle (sapling-crypto crate) — though only for re-anchoring spends of *mined* notes: Sapling nullifiers are position-dependent, so spends of unmined Sapling notes cannot be pre-signed regardless.

Related: #2467, #2524.

---
Filed with the assistance of Claude Code (Claude Fable 5).

nuttycom:
[...]

PR #2828, opened 2026-07-29
wallet: preserve transaction status observation intent

nuttycom:
## Summary

- record an explicit durable status-observation intent for transactions whose outcome cannot be learned through compact-block scanning
- determine scan observability from actual wallet-owned shielded spends and outputs instead of shielded bundle presence
- keep status intent dormant while mined and reactivate it automatically after a chain rewind
- represent status observation and transaction enhancement as independent intents
- migrate obsolete inferred status entries

This supersedes the bundle-presence approach discussed in #2825. In particular, it handles external shielding transactions, out-of-order scanning, transaction discovery/recovery, enhancement of a transaction already under status observation, and reorgs without causing the wallet to silently stop observing transaction outcomes.

## Database write atomicity

1. **How many writes?** Sending or storing a transaction already performs multiple wallet writes; this change adds the status-intent write within that existing operation. Status updates can update transaction metadata and remove completed queue intent. The migration deletes obsolete rows and transactionally rebuilds the queue with a `(txid, query_type)` uniqueness constraint.
2. **Must they land together?** Yes for transaction storage, status updates, and the schema migration. Runtime operations execute through existing transaction-scoped wallet APIs. The migration receives a `rusqlite::Transaction`, so its cleanup and table replacement commit or roll back together.
3. **What happens on retry?** Queue insertion is an idempotent upsert per intent. Status updates and queue cleanup are also idempotent. Migration rollback preserves the old table and schemerz retries the complete migration.
4. **Can reads see only one side?** No committed partial state is exposed because each multi-write path shares an enclosing SQLite transaction. Status and enhancement may intentionally coexist as independent rows. Mined status intents remain stored but are filtered from request reads until a rewind clears the mined height.

## Testing

- `cargo check --workspace --all-features`
- `cargo check -p zcash_client_sqlite --no-default-features`
- `cargo test -p zcash_client_sqlite --all-features`
- focused regression coverage for coexisting `Status` and `Enhancement` intents
- `git diff --check`

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -30,8 +30,14 @@ workspace.
   `proto::{CompactFormatError, ProposalDecodingError}`, `scanning::ScanError`,
   `sync::Error`, `sync::decryptor::TryQueueError`, `tor::Error`,
   `tor::grpc::GrpcError`, and `tor::http::HttpError`.
+- `zcash_client_backend::data_api::ll::LowLevelWalletWrite` has added method
+  `queue_tx_status`, which records durable status-observation intent for
+  transactions whose mined status cannot be learned by compact-block scanning.
 
 ### Fixed
+- `data_api::ll::wallet::store_decrypted_tx` now distinguishes a wallet-observable
+  shielded spend or output from mere shielded bundle presence when deciding
+  whether txid-based status observation is required.
 - The Tor HTTP and gRPC transports now reject a URL whose scheme is neither
   `http` nor `https` (for example `ftp` or `ws`) with `tor::http::HttpError::NonHttpUrl`.
   Previously such a URL was silently treated as plaintext HTTP, contradicting the
--- a/zcash_client_backend/src/data_api/ll.rs
+++ b/zcash_client_backend/src/data_api/ll.rs
@@ -618,6 +618,15 @@ pub trait LowLevelWalletWrite: LowLevelWalletRead {
         dependent_tx_ref: Option<Self::TxRef>,
     ) -> Result<(), Self::Error>;
 
+    /// Adds a [`TransactionDataRequest::GetStatus`] request for a transaction whose mined status
+    /// cannot be learned through ordinary compact-block scanning.
+    ///
+    /// The request intent must remain durable while the transaction is mined, so that it becomes
+    /// active again if a chain rewind un-mines the transaction.
+    ///
+    /// [`TransactionDataRequest::GetStatus`]: super::TransactionDataRequest
+    fn queue_tx_status(&mut self, txid: TxId) -> Result<(), Self::Error>;
+
     /// Adds a [`TransactionDataRequest::TransactionsInvolvingAddress`] request to the transaction
     /// data request queue. When the transparent output of `tx_ref` at output index `output_index`
     /// (which must have been received at `receiving_address`) is detected as having been spent,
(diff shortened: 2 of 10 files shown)


nuttycom:
This is an alternative to #2525

zcash_client_backend/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ workspace.
 
 ### Added
 - `impl Debug for zcash_client_backend::data_api::ll::wallet::PutBlocksError`
+- `zcash_client_backend::data_api::ll::LowLevelWalletWrite::queue_tx_status`,
```
nuttycom: This is a change to a trait; it should go in the ## Changed## section

nuttycom (commented):
self-utACK with changelog nit.

Cosmos-Harry (commented):
queue_tx_retrieval's ON CONFLICT never upgrades an existing row's query_type — it only merges dependent_transaction_id. So if a txid already has a Status row, a later Enhancement request for it is silently swallowed; query_type stays Status forever.

This is reachable in practice: transaction A gets a Status row (no wallet-detectable output). Later, transaction B spends A's output as a prevout and pays the wallet — queue_transparent_input_retrieval then queues A for Enhancement (to check if A was wallet-created, for gap-limit discovery). Since A already has a Status row, that request is dropped and A's raw data never gets fetched.

Should ON CONFLICT upgrade Status → Enhancement here, since Enhancement is a strict superset?

nuttycom:
Addressed the review in `5124eb25c2`.

Rather than making `Enhancement` overwrite `Status`, the queue now permits independent `(txid, query_type)` rows. This preserves the durable status intent across enhancement and later rewinds, while allowing the same txid to be fetched for transparent relevance. The regression test queues `Status` first and then `Enhancement`, and asserts that both requests are returned.

Validated with the focused migration/regression tests, `cargo check --workspace --all-features`, and Clippy for all `zcash_client_sqlite` targets/features.

Cosmos-Harry (approved):

zcash-harry:
utACK

Cosmos-Harry (changes requested):
Actually claude found more 

thread 'wallet::init::tests::verify_schema' panicked
  left (actual, post-migration):   CREATE TABLE "tx_retrieval_queue" ( txid BLOB NOT NULL, query_type INTEGER NOT NULL,
                                    dependent_transaction_id INTEGER, FOREIGN KEY (dependent_transaction_id)
                                    REFERENCES transactions (id_tx), CONSTRAINT tx_retrieval_intent UNIQUE (txid, query_type) )
  right (expected, from db.rs):    CREATE TABLE "tx_retrieval_queue" ( txid BLOB NOT NULL UNIQUE, query_type INTEGER NOT NULL,
                                    dependent_transaction_id INTEGER REFERENCES transactions (id_tx) ON DELETE CASCADE )
Two distinct bugs are visible in that one diff:

1. The reference schema file wasn't updated (the one you already knew about, now confirmed as a hard failure)
zcash_client_sqlite/src/wallet/db.rs's TABLE_TX_RETRIEVAL_QUEUE constant is the "expected fully-migrated schema" used by wallet::init::tests::verify_schema to catch exactly this kind of drift — it still says txid BLOB NOT NULL UNIQUE (old single-row-per-txid constraint). It needs updating to match the new CONSTRAINT tx_retrieval_intent UNIQUE (txid, query_type).

2. A second, more serious regression: ON DELETE CASCADE got dropped from the foreign key.
The new migration's CREATE TABLE tx_retrieval_queue_new declares:


FOREIGN KEY (dependent_transaction_id) REFERENCES transactions(id_tx)
— no ON DELETE CASCADE. But the schema this migration is supposed to preserve has ... REFERENCES transactions(id_tx) ON DELETE CASCADE. This isn't cosmetic: it means deleting a transaction row that a queue entry depends on no longer automatically cleans up that queue entry. Depending on SQLite's foreign-key enforcement setting, that either leaves a dangling row behind or blocks the parent delete outright — a real referential-integrity regression introduced by the table-rebuild pattern this migration uses (copy structure → drop old → rename new), which evidently copied from an older/incomplete version of the table definition rather than the fully-current one.

nuttycom (commented):
Ugh, just saw. Okay.

nuttycom:
Addressed both schema findings in `a85cf492b3`.

- Updated `TABLE_TX_RETRIEVAL_QUEUE` to use `UNIQUE (txid, query_type)`.
- Preserved `ON DELETE CASCADE` on `dependent_transaction_id` in the rebuilt table.
- Also recreated `idx_tx_retrieval_queue_dependent_tx`, which would otherwise be lost when dropping the old table.

`wallet::init::tests::verify_schema`, the migration test, and the coexisting status/enhancement regression all pass.

Cosmos-Harry (approved):
utACK

PR #2829, opened 2026-07-29
Release zcash_client_sqlite version 0.22.0-rc.5

nuttycom:
Release preparation for `zcash_client_sqlite 0.22.0-rc.5` and the crates beneath it, covering everything merged to `main` since the rc.4 round — principally the ZIP 318 canonical crossing work (#2826) and durable transaction-status observation intent (#2828).

| Crate | Version |
| --- | --- |
| `zcash_protocol` | 0.10.1 → 0.10.2 |
| `zcash_keys` | 0.16.0 → 0.16.1 |
| `zcash_client_backend` | 0.24.0-rc.4 → 0.24.0-rc.5 |
| `zcash_pool_migration` | 0.1.0-rc.3 → 0.1.0-rc.4 |
| `zcash_client_sqlite` | 0.22.0-rc.4 → 0.22.0-rc.5 |

`zcash_protocol` and `zcash_keys` are patch releases: `zcash_protocol` gains the `zip318` module (which `zcash_client_backend` and `zcash_pool_migration` now source their ZIP 318 constants from, so its published minimum must be 0.10.2), and `zcash_keys` gains `impl Debug for GapAddressesError`.

Each release commit moves that crate's `## [Unreleased]` changelog content under a dated version heading, bumps the crate version and the workspace dependency requirement, and refreshes `Cargo.lock` and the `cargo-vet` metadata. Where a released crate's version is semver-visible to a dependent, the dependent's `## [Unreleased]` section records the migration.

### DIFF
--- a/zcash_client_backend/Cargo.toml
+++ b/zcash_client_backend/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "zcash_client_backend"
 description = "APIs for creating shielded Zcash light clients"
-version = "0.24.0-rc.4"
+version = "0.24.0-rc.5"
 authors = [
     "Jack Grigg <jack@z.cash>",
     "Kris Nuttycombe <kris@electriccoin.co>"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7332,7 +7332,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_pool_migration"
-version = "0.1.0-rc.3"
+version = "0.1.0-rc.4"
 dependencies = [
  "corez",
  "getset",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7428,7 +7428,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_protocol"
-version = "0.10.1"
+version = "0.10.2"
 dependencies = [
  "corez",
  "document-features",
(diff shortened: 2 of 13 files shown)


Cosmos-Harry (approved):
utACK

nullcopy (dismissed):
need changelog for zcash_protocol

zcash_client_backend/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "zcash_client_backend"
 description = "APIs for creating shielded Zcash light clients"
-version = "0.24.0-rc.4"
+version = "0.24.0-rc.5"
```
nullcopy, line 4: Remind me why these are rc bumps and not version bumps

nullcopy (approved):
utACK cb356a7def26d0bd8e1f21709951aeea137f58fa

zcash_client_backend/Cargo.toml
nuttycom: These really should have been betas - I got anxious and thought we were closer than we were. But for anything but x.y.z, each release is considered a distinct version for semver; cargo won't bump you between RC versions automatically, so the advantage of these being RCs is that we can still make API-breaking changes between them.

PR #2831, opened 2026-07-29
zcash_pool_migration: condition the drawn broadcast schedule

dannywillems:
## Problem

A migration's inter-arrival delay is a continuous exponential rounded to whole blocks, so a drawn delay can be **zero**: `P(delay rounds to 0) = 1 - exp(-1 / (2 * mean))`, about one draw in 288 at the ZIP 318 transfer mean of 144 blocks and one in 48 at the tighter preparation mean of 24. Since broadcast heights are the cumulative sum of those delays, a zero gap schedules two of the wallet's transactions in the **same block**.

That is the correlation the drawn delays exist to prevent. Migration transactions have a distinctive, near-identical shape, so two of them mined together are far more plausibly one wallet's pair than two unrelated wallets' coincidence — and for the pool-crossing transfers, a shared block additionally reveals that their individually unremarkable (quantized) values sum to one balance.

ZIP 318 specifies the exponential and its `MAX_DELAY` rejection but is silent on collisions, so this is an addition to it rather than a reading of it.

## Approach

Rather than special-case that one collision, this expresses it as an instance of a general condition on a drawn schedule, which is also what applications need: a maintenance window they will not be online for, a height range the user has blocked out.

- **`ScheduleConstraint`** — `admits(&Placement) -> bool`, with a provided `is_valid(start, &heights)` that is the conjunction of `admits` over every prefix: the whole-schedule check corresponding to what the samplers enforce, for validating a schedule from elsewhere (a persisted plan, a hand-assembled one).
- **`Placement`** — the candidate broadcast height plus the partial schedule it would extend (`height`, `index`, `start`, `preceding`, `previous`, `gap`). This is the extension point: a condition needing more than the height gains another accessor here, without touching the trait or the samplers.
- **`MinBlockGap`**, with `DISTINCT_BLOCKS` as the crate default — the no-two-in-one-block condition, now one built-in among others.
- Composition by conjunction: closures are conditions, and `()`, slices, and 2-/3-tuples conjoin their members, so a caller writes `(MinBlockGap::DISTINCT_BLOCKS, my_closure)`.
- `_with` forms throughout (`schedule_with`, `schedule_broadcast_heights_with`, `schedule_prep_broadcast_heights_with`, `plan_migration_with`) taking a caller-supplied condition. The existing names keep their behaviour under the default.

The samplers redraw a rejected placement, so a schedule is returned **only if** the condition holds on it, and `ConstraintUnsatisfied` (surfaced as `MigrationError::Unschedulable`) is returned when no draw satisfies it within `CONSTRAINT_ATTEMPTS`.

Two design points:

- A condition is a predicate on a **prefix plus one candidate**, not on a finished schedule. That is what makes it enforceable during sampling: each placement is accepted or redrawn as it is drawn, at one redraw per rejection rather than redrawing whole schedules, and the retained delays keep their distribution conditioned on the predicate. A predicate that can only be evaluated once every later height is known cannot be expressed this way, and the docs say so.
- Collisions are **redrawn**, not clamped up to one block. Clamping would move the zero mass onto a gap of exactly one, making one-block gaps roughly twice as likely as the exponential predicts; redrawing leaves the shape of the retained support untouched.

## Breaking changes

`schedule`, `schedule_broadcast_heights`, and `schedule_prep_broadcast_heights` now return `Result<_, ConstraintUnsatisfied>` and apply `MinBlockGap::DISTINCT_BLOCKS`. Broadcast heights are consequently strictly increasing rather than merely non-decreasing. `MigrationError` gains an `Unschedulable` variant. `plan_migration` is unchanged.

## Testing

`cargo test -p zcash_pool_migration --all-features` passes (179 tests), fmt/clippy/rustdoc clean. New coverage:

- the zero-delay rate is measured, so the condition is answering a real outcome rather than a theoretical one;
- `is_valid` as the prefix conjunction, composition, and `Placement` accessors;
- a proptest that a returned schedule satisfies an arbitrary caller condition, drawn under a mean-1-block distribution where the redraw path fires constantly;
- unsatisfiable conditions (a predicate nothing meets, and a minimum gap wider than the delay cap) return no schedule, naming where they gave up;
- an end-to-end `plan_migration_with` blackout-window test that first asserts the **unconstrained** draw for that seed lands in the window, so it cannot pass vacuously.

Existing golden vectors are unchanged: those seeds happen to draw no zeros.

### DIFF
--- a/zcash_pool_migration/CHANGELOG.md
+++ b/zcash_pool_migration/CHANGELOG.md
@@ -8,12 +8,15 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Added
-- `zcash_pool_migration::scheduling::{ScheduleConstraint, Placement, MinBlockGap,
+- `zcash_pool_migration::scheduling::{ScheduleConstraint, Placement, Predicate, MinBlockGap,
   ConstraintUnsatisfied, CONSTRAINT_ATTEMPTS}`, conditions a drawn broadcast schedule must satisfy.
   A condition is a predicate on one candidate height plus the partial schedule it extends; the
   samplers redraw a rejected placement, so a schedule is returned only if the condition holds on it
   (`ScheduleConstraint::is_valid`) and otherwise fails with `ConstraintUnsatisfied`. Conditions
-  compose by conjunction (tuples, slices), and a closure is a condition.
+  compose by conjunction (tuples, slices), forward through references and boxes (so a set assembled
+  at runtime can hold `Box<dyn ScheduleConstraint>`), and a closure becomes one via `Predicate`.
+  Note that a condition beyond the default narrows this wallet's schedule distribution away from
+  other wallets', which costs anonymity.
 - `zcash_pool_migration::scheduling::{schedule_with, schedule_broadcast_heights_with,
   schedule_prep_broadcast_heights_with}` and `zcash_pool_migration::engine::plan_migration_with`,
   which draw a schedule subject to a caller-supplied condition. The existing forms apply the crate
--- a/zcash_pool_migration/CHANGELOG.md
+++ b/zcash_pool_migration/CHANGELOG.md
@@ -7,6 +7,28 @@ and this library adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+- `zcash_pool_migration::scheduling::{ScheduleConstraint, Placement, MinBlockGap,
+  ConstraintUnsatisfied, CONSTRAINT_ATTEMPTS}`, conditions a drawn broadcast schedule must satisfy.
+  A condition is a predicate on one candidate height plus the partial schedule it extends; the
+  samplers redraw a rejected placement, so a schedule is returned only if the condition holds on it
+  (`ScheduleConstraint::is_valid`) and otherwise fails with `ConstraintUnsatisfied`. Conditions
+  compose by conjunction (tuples, slices), and a closure is a condition.
+- `zcash_pool_migration::scheduling::{schedule_with, schedule_broadcast_heights_with,
+  schedule_prep_broadcast_heights_with}` and `zcash_pool_migration::engine::plan_migration_with`,
+  which draw a schedule subject to a caller-supplied condition. The existing forms apply the crate
+  default, `MinBlockGap::DISTINCT_BLOCKS`.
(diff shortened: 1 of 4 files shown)


dannywillems:
Self-review turned up two defects in the first commit, now fixed in 4bec0c2f54, plus one omission:

1. **The blanket `impl ScheduleConstraint for F: Fn(&Placement) -> bool` foreclosed every forwarding impl.** `&C`, `Box<C>`, and `Arc<C>` implement `Fn` whenever `C` does, so those impls overlap the blanket one and cannot be added. That made the runtime case the slice impl advertised — a set of conditions assembled from user settings, which is heterogeneous and so must be held as `Box<dyn ScheduleConstraint>` — fail to compile. Closures now go through `Predicate`, and references and boxes forward, so a slice of boxed trait objects is a condition. Covered by a new unit test and a doctest.

2. **An end-to-end assertion stated an invariant the code does not have.** It ran `is_valid` over `MigrationPlan::schedule`, which is not in draw order: the engine shuffles it, precisely so the on-chain sequence of denominations does not spell out the balance. `is_valid` reads a sequence in draw order, so on a shuffled one a descending pair reports a zero gap and the condition rejects. It passed only because that fixture plans a single transfer; on a balance splitting into ten it fails. The test now plans a multi-transfer balance and asserts what the sampler actually guarantees there — pairwise distinct heights — keeping the sequence form for the preparation layers, which are not shuffled.

3. **The anonymity cost of a non-default condition was undocumented.** The schedule draws exist to make every migrating wallet's broadcasts look alike, so the anonymity set is the set of wallets drawing from the same distribution. A condition peculiar to one application, or to one user's settings, is a signature of the schedules it produces — the same partitioning this module exists to prevent, applied by the wallet to itself. Now stated on the module, on the trait, and on both `_with` entry points, recommending the default on mainnet.

Follow-ups filed rather than fixed here: #2832 (greedy per-step rejection is not the conditioned joint law, and the docs currently overstate it), #2833 (conditions do not reach the sync/proving wake-up schedule, which is where the availability motivation actually bites), #2834 (attempt budget is a fixed const with an over-broad probability claim, `is_valid` is unsealed, first-placement gap is measured against a non-transaction, prefix-scanning conditions are quadratic, cross-sequence non-collision is incidental, and the constraint/condition vocabulary is inconsistent).

Marked draft while those are considered.

PR #2836, opened 2026-07-29
zcash_client_sqlite: repair orchard_ironwood_migrations schema drift

dannywillems:
Resolves MOB-1598
Resolves MOB-1592

## Problem

A wallet built from a git revision in `a0e937819e..b24a42723b` fails **every** scan with:

```
Error while scanning blocks: The underlying datasource produced the following error:
no such column: anchor_bucket_interval in
SELECT DISTINCT anchor_bucket_interval FROM orchard_ironwood_migrations WHERE status <> ? at offset 16
```

`b24a42723b` added `anchor_bucket_interval` to the `orchard_ironwood_migration_tables` DDL **in place**, on the stated assumption that the migration was "not yet released". That holds for tagged releases (every tagged `zcash_client_sqlite` containing `a0e937819e` also contains `b24a42723b`), but not for a wallet built from a git revision inside that 165-commit window. Such a wallet has the table without the column and has already recorded `7b2f6a41-…` as applied, so its `CREATE TABLE IF NOT EXISTS` never runs again and the column can never appear. There is no `ALTER TABLE` anywhere in the tree.

The blast radius is not confined to pool migration. `WalletDb::put_blocks` reads the committed grids on every scan once NU6.3 has an activation height (`lib.rs:2175`), and "no such column" is a **prepare-time** error, so it fires even when no migration is in progress and the table is empty. Every scan then fails, no block is written, and no transaction ever acquires a mined height — which surfaces to the user as every transaction being stuck "pending" indefinitely, and later flipping to "expired" once the tip passes `expiry_height`.

## Fix

A new `orchard_ironwood_migration_anchor_interval` migration adds the column where it is absent. The presence check is load-bearing rather than defensive: on a wallet whose table was created after the DDL edit the column already exists, and `ADD COLUMN` would fail with "duplicate column name".

I chose `ADD COLUMN` over a table rebuild because `orchard_ironwood_migrations` has five child tables with `ON DELETE CASCADE`, and dropping the parent to add one column is a lot of risk for the benefit. To keep a repaired table and a freshly created one byte-identical in `sqlite_schema` — SQLite cannot add a `NOT NULL` column without a default — the creating DDL carries a matching `DEFAULT`. `verify_schema` confirms the two agree. The store always binds the column explicitly, so no insert relies on the default.

### On the backfill value

The grid an existing row was committed under is **not recoverable**: it came from the planning `SchedulingParams` (`engine.rs:949`, `:2637`), which is not recorded anywhere the migration can read. The ZIP 318 grid (144) is the only defensible reconstruction, and it is exact on the production network, where a wallet MUST use it and `AnchorBucketInterval::custom` is documented test-network-only.

It can be wrong on a **test** network, where a shorter grid is legitimate — note that `with_anchor_retention_interval` (`78229dd3d8`) itself landed inside the affected window. Such a row is then rejected by the committed-vs-configured check as `ProveError::AnchorIntervalMismatch` (`engine.rs:1566`) and must be re-planned. That is loud and diagnosable rather than silent, and it does not prune checkpoints: `put_blocks` unions the committed grids with the wallet's *current* interval, so a testnet wallet still configured at its custom grid keeps retaining it.

The alternative — making the column nullable and treating `NULL` as "unknown, skip the check" — is strictly more accurate for affected wallets, but changes `MigrationState::anchor_bucket_interval` and the store from `NonZeroU32` to `Option<NonZeroU32>`, weakening the invariant `b24a42723b` introduced for every wallet to accommodate a handful of dev testnet ones. Happy to take that direction instead if reviewers prefer it.

## Tests

- New: upgrade path adds the column and backfills an existing row; fresh path is a no-op; `test_migrate`.
- `verify_schema` — confirms repaired and fresh schemas agree.
- `current_leaf_migrations_are_the_dag_leaves`, `ids_module_covers_every_migration`, `migrate_between_releases_without_data`.
- 15 `pool_migration::orchard_ironwood` store tests.

`cargo fmt --check` and `cargo clippy --all-features --all-targets` are clean. I ran the tests above rather than the full crate suite.

## Worth a second look

Any other recent commit that edited an already-applied migration's DDL in place on the same "not yet released" reasoning has the same defect. Tag status is not a sufficient test when consumers build from git.

### DIFF
--- a/zcash_client_sqlite/src/wallet/init/migrations/orchard_ironwood_migration_anchor_interval.rs
+++ b/zcash_client_sqlite/src/wallet/init/migrations/orchard_ironwood_migration_anchor_interval.rs
@@ -1,0 +1,202 @@
+//! Adds the `anchor_bucket_interval` column to `orchard_ironwood_migrations` where it is missing.
+//!
+//! The column records the anchor retention grid a pool migration was committed under. It was
+//! introduced by editing the [`orchard_ironwood_migration_tables`] DDL in place, on the assumption
+//! that no wallet had yet applied that migration. That assumption does not hold for a wallet built
+//! from a git revision between the two commits: such a wallet has `orchard_ironwood_migrations`
+//! without the column and has already recorded [`orchard_ironwood_migration_tables`] as applied, so
+//! its `CREATE TABLE IF NOT EXISTS` never runs again and the column can never appear.
+//!
+//! The consequence is not confined to pool migration. [`WalletDb::put_blocks`] reads the committed
+//! grids on every scan once NU6.3 has an activation height, and "no such column" is a prepare-time
+//! error, so it fires even when no migration is in progress and the table is empty. Every scan then
+//! fails, no block is ever written, and no transaction ever acquires a mined height.
+//!
+//! This migration adds the column where it is absent. On a wallet whose table was created after the
+//! DDL edit the column is already present and this is a no-op, so both paths converge on the schema
+//! in [`crate::wallet::db::TABLE_ORCHARD_IRONWOOD_MIGRATIONS`].
+//!
+//! [`WalletDb::put_blocks`]: crate::WalletDb
+
+use std::collections::HashSet;
+use std::num::NonZeroU32;
+
+use rusqlite::named_params;
+use schemerz_rusqlite::RusqliteMigration;
+use uuid::Uuid;
+use zcash_protocol::zip318::AnchorBucketInterval;
+
+use super::orchard_ironwood_migration_tables;
+use crate::wallet::init::WalletMigrationError;
+
+/// Adds the `anchor_bucket_interval` column to `orchard_ironwood_migrations` where it is missing.
+pub const MIGRATION_ID: Uuid = Uuid::from_u128(0x1ab3caf9_ef1e_482c_93a3_a3f1080038df);
+
+const DEPENDENCIES: &[Uuid] = &[orchard_ironwood_migration_tables::MIGRATION_ID];
+
+pub(super) struct Migration;
+
+impl schemerz::Migration<Uuid> for Migration {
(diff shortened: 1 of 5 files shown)


zcash_client_sqlite/src/wallet/init/migrations/orchard_ironwood_migration_anchor_interval.rs
```diff
@@ -0,0 +1,202 @@
...
+            [],
+            |row| row.get::<_, bool>(0),
+        )
+        .unwrap()
+    }
+
+    /// The upgrade path: the column is added, and an existing row is backfilled with the ZIP 318
+    /// grid rather than being dropped.
+    #[test]
+    fn adds_column_and_backfills_existing_rows() {
```
dannywillems, line 159: LGTM.

zcash_client_sqlite/src/wallet/init/migrations/orchard_ironwood_migration_anchor_interval.rs
```diff
@@ -0,0 +1,202 @@
...
+            )
+            .unwrap();
+        assert_eq!(id, 1);
+        assert_eq!(interval, default_interval().get());
+    }
+
+    /// The fresh path: the table already carries the column, so `up` must leave it alone rather
+    /// than fail with "duplicate column name".
+    #[test]
+    fn is_a_no_op_when_the_column_is_present() {
```
dannywillems, line 191: LGTM. Sure...
I don't know if it exists in SQL a way to check if the schema has been modified or not after an operation.

dannywillems (commented):
self-utACK e78bddaa1dc09f6b410ec9b4314bc1e8680dbef6

nuttycom (approved):
utACK

PR #2839, opened 2026-07-29
zcash_client_backend: hoist the remaining pool-test imports to the top

dannywillems:
Follow-up to the review comment on #2822 asking that the imports in
`data_api/testing/pool.rs` be declared at the top rather than inside each test
function. That PR hoisted only `OrchardPoolTester`; this one finishes the job.

58 function-local `use` statements, 31 distinct after dedup, spread across 8
different `cfg` contexts, move to the module header gated to match the
functions that need them.

Four groups were not simple moves:

- `PoolType` and `OrchardPoolTester` were already imported at module level, so
  the local copies were shadowing duplicates and are dropped.
- `PoolType`'s module-level import was gated on `orchard` **because** the
  `not(orchard)` function shadowed it with a local import. With that gone the
  outer import is genuinely needed in every configuration, so the gate comes
  off. This is the one that breaks the Sapling-only build if missed.
- `MaxSpendMode`, `OutputOfSentTx`, `TransactionStatus` and `NonEmpty` each
  already appear in an existing grouped import, so the hoisted copies fold into
  those groups instead of sitting alongside them.
- `GreedyInputSelectorError` was imported under both `transparent-inputs` and
  `not(transparent-inputs)`, so it is unconditional.

One place the move makes the code worse, worth a reviewer's opinion:
`ProposalError` is already in the `transparent-inputs` group, so the `orchard`
arm has to become `all(orchard, not(transparent-inputs))` to avoid a duplicate
definition. That reads worse than the local import it replaces; happy to leave
that single one function-local if preferred.

Checked against every feature combination that selects a different arm:
`test-dependencies` alone, and with each of `orchard`, `transparent-inputs` and
`pczt`, then combined. All compile warning-free, as do
`--all-features --all-targets` and `--no-default-features` under
`clippy -D warnings`.

No behaviour change: this only relocates imports.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool.rs
+++ b/zcash_client_backend/src/data_api/testing/pool.rs
@@ -59,17 +59,8 @@ use crate::{
 
 use super::{DataStoreFactory, Reset, TestCache, TestFvk, TestState};
 
-#[cfg(feature = "orchard")]
-use super::orchard::OrchardPoolTester;
-
 use crate::data_api::ll::wallet::PRUNING_DEPTH;
 use crate::data_api::wallet::input_selection::GreedyInputSelectorError;
-#[cfg(all(feature = "pczt", feature = "transparent-inputs"))]
-use crate::data_api::wallet::input_selection::LockedInputPolicy;
-#[cfg(feature = "orchard")]
-use crate::data_api::wallet::{input_selection::SpendPolicy, propose_transfer};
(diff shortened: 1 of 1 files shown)


nullcopy (commented):
I'm loving these cleanups
utACK 3a8819052e47b525941a14d48b8dad0546c45595

nullcopy (dismissed):

nuttycom (approved):
utACK, nice cleanup

PR #2842, opened 2026-07-29
ci: reduce test suite runtime

nuttycom:
## Summary

- stacked on #2854, which resolves the repository's existing zizmor findings and provides safe structured feature outputs
- tier CI so pull requests execute the broad `all-features` configuration while compiling every other active Linux feature configuration across all targets
- retain the complete active Linux runtime matrix, representative cross-platform coverage, and full-workspace expensive `all-pools` suites outside pull requests
- run all currently inactive NU7 runtime coverage only on the nightly schedule, including regular Linux, Windows, and expensive suites
- cancel superseded runs, improve cache reuse, reduce test debug information, and cap nested test/prover concurrency
- use in-memory SQLite databases and Sapling mock provers for routine tests
- add the additive `ignore-expensive-tests` feature, which keeps expensive tests compiled and visible but ignored in broad `--all-features` runs
- report failed scheduled expensive-test runs to Slack through the `SLACK_WEBHOOK_URL` repository secret

## Motivation

Recent CI behavior showed the SQLite all-feature unit suite taking 777.77 seconds on Linux and over 3000 seconds on an uncached Windows NU7 job. The overall observed workflow consumed roughly 704 runner-minutes, with repeated feature/platform configurations paying substantially overlapping setup and execution costs.

With these changes, the all-feature SQLite library suite completes locally with CI-like thread limits in 187.32 seconds, a roughly 76% reduction for the primary bottleneck.

The pruned pull-request matrix preserves runtime coverage for the broad active configuration most likely to expose behavioral differences. The remaining partial configurations still compile all test targets on every pull request. Merge groups, `main`, scheduled, and manual runs retain the broader active runtime coverage. NU7 runtime coverage is scheduled nightly while NU7 is not under active development. Expensive tests run as complete workspace suites rather than fragile name-filtered subsets.

## Slack setup

Create a Slack incoming webhook for the desired channel and save its URL as the `SLACK_WEBHOOK_URL` Actions repository secret. Only a failed or cancelled scheduled expensive-test matrix sends a message; pull-request, push, merge-group, and manual runs do not.

## Validation

- `cargo test -p zcash_client_sqlite --all-features --lib -- --test-threads=2` (480 passed, 2 ignored)
- `cargo test -p zcash_client_backend --all-features --lib` (132 passed)
- `cargo check -p zcash_client_sqlite --all-features --all-targets`
- verified `--all-features` reports the isolated chain and proving tests as ignored with an explicit reason
- verified the nightly-like `expensive-tests` feature set contains no ignored copies of those tests
- exercised both accepted feature-argument forms through `.github/scripts/cargo-with-features.sh`
- file-backed concurrent-lock test passed
- workflow YAML parse and `git diff --check`
- `uvx zizmor==1.28.0 .github` on the combined stacked tree — no findings

The current GitHub Actions run provides the end-to-end hosted-runner validation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,36 @@ on:
   push:
     branches: main
   merge_group:
+  schedule:
+    - cron: '23 8 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: >-
+    ${{
+      format(
+        'ci-{0}-{1}-{2}',
+        github.workflow,
+        github.event_name,
+        github.event.pull_request.number || github.ref
+      )
+    }}
+  cancel-in-progress: >-
+    ${{
+      github.event_name == 'pull_request'
+      || (
+        github.event_name == 'push'
+        && github.ref == 'refs/heads/main'
+      )
+    }}
+
+env:
+  # Full debug information substantially increases optimized test linking time and
+  # cache size, particularly on Windows. Line tables retain useful CI backtraces.
+  CARGO_PROFILE_TEST_DEBUG: 1
+  # Proving code uses Rayon internally. Capping both layers prevents each
+  # concurrent libtest case from trying to occupy the whole runner.
+  RAYON_NUM_THREADS: 2
 
 jobs:
   required-test:
@@ -19,40 +49,26 @@ jobs:
       matrix:
         target:
           - Linux
-        state:
-          - transparent
-          - transparent-inputs-only
-          - transparent-key-encoding-only
-          - Sapling-only
-          - Orchard
-          - all-pools
-          - all-features
-          - NU6.3
-          - NU7
+        state: >-
+          ${{
+            fromJSON(
+              github.event_name == 'schedule'
+              && '["transparent", "transparent-inputs-only", "transparent-key-encoding-only", "Sapling-only", "Orchard", "all-pools", "all-features", "NU7"]'
+              || (
+                github.event_name == 'pull_request'
+                && '["all-features"]'
+                || '["transparent", "transparent-inputs-only", "transparent-key-encoding-only", "Sapling-only", "Orchard", "all-pools", "all-features"]'
+              )
+            )
+          }}
 
         include:
           - target: Linux
             os: ${{ vars.UBUNTU_LARGE_RUNNER || 'ubuntu-latest' }}
 
-          - state: transparent
-            transparent-pool: true
-          - state: transparent-inputs-only
-            extra_flags: transparent-inputs
-          - state: transparent-key-encoding-only
-            extra_flags: transparent-key-encoding
-          - state: Orchard
-            orchard-pool: true
-          - state: all-pools
-            all-pools: true
-          - state: all-features
-            all-features: true
-          - state: NU7
-            all-features: true
-            rustflags: '--cfg zcash_unstable="nu7"'
-
     env:
-      RUSTFLAGS: ${{ matrix.rustflags }}
-      RUSTDOCFLAGS: ${{ matrix.rustflags }}
+      RUSTFLAGS: ${{ matrix.state == 'NU7' && '--cfg zcash_unstable="nu7"' || '' }}
+      RUSTDOCFLAGS: ${{ matrix.state == 'NU7' && '--cfg zcash_unstable="nu7"' || '' }}
 
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -61,15 +77,25 @@ jobs:
       - id: prepare
         uses: ./.github/actions/prepare
         with:
-          transparent-pool: ${{ matrix.transparent-pool || false }}
-          orchard-pool: ${{ matrix.orchard-pool || false }}
-          all-pools: ${{ matrix.all-pools || false }}
-          all-features: ${{ matrix.all-features || false }}
-          extra-features: ${{ matrix.extra_flags || '' }}
+          transparent-pool: ${{ matrix.state == 'transparent' }}
+          orchard-pool: ${{ matrix.state == 'Orchard' }}
+          all-pools: ${{ matrix.state == 'all-pools' }}
+          all-features: ${{ matrix.state == 'all-features' || matrix.state == 'NU7' }}
+          extra-features: >-
+            ${{
+              matrix.state == 'transparent-inputs-only'
+              && 'transparent-inputs'
+              || (
+                matrix.state == 'transparent-key-encoding-only'
+                && 'transparent-key-encoding'
+                || ''
+              )
+            }}
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
           shared-key: msrv
-          save-if: ${{ matrix.state == 'all-features' }}
+          # NU7 has a distinct RUSTFLAGS-derived cache key.
+          save-if: ${{ matrix.state == 'all-features' || matrix.state == 'NU7' }}
       - name: Run tests
         shell: bash
         env:
@@ -82,12 +108,70 @@ jobs:
           elif [[ -n "$FEATURES" ]]; then
             feature_flags+=(--features "$FEATURES")
           fi
-          cargo test --workspace "${feature_flags[@]}"
+          cargo test --workspace "${feature_flags[@]}" -- --test-threads=2
       - name: Verify working directory is clean
         run: git diff --exit-code
 
+  feature-check:
+    name: Check ${{ matrix.state }} test targets
+    runs-on: ${{ vars.UBUNTU_LARGE_RUNNER || 'ubuntu-latest' }}
+    strategy:
+      matrix:
+        state: >-
+          ${{
+            fromJSON(
+              github.event_name == 'pull_request'
+              && '["transparent", "transparent-inputs-only", "transparent-key-encoding-only", "Sapling-only", "Orchard", "all-pools"]'
+              || '["covered-by-runtime-matrix"]'
+            )
+          }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - id: prepare
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: ./.github/actions/prepare
+        with:
+          transparent-pool: ${{ matrix.state == 'transparent' }}
+          orchard-pool: ${{ matrix.state == 'Orchard' }}
+          all-pools: ${{ matrix.state == 'all-pools' }}
+          extra-features: >-
+            ${{
+              matrix.state == 'transparent-inputs-only'
+              && 'transparent-inputs'
+              || (
+                matrix.state == 'transparent-key-encoding-only'
+                && 'transparent-key-encoding'
+                || ''
+              )
+            }}
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        if: ${{ github.event_name == 'pull_request' }}
+        with:
+          shared-key: msrv
+          save-if: "false"
+      - name: Check all test targets
+        if: ${{ github.event_name == 'pull_request' }}
+        shell: bash
+        env:
+          ALL_FEATURES: ${{ steps.prepare.outputs.all-features }}
+          FEATURES: ${{ steps.prepare.outputs.feature-list }}
+        run: |
+          feature_flags=()
+          if [[ "$ALL_FEATURES" == true ]]; then
+            feature_flags+=(--all-features)
+          elif [[ -n "$FEATURES" ]]; then
+            feature_flags+=(--features "$FEATURES")
+          fi
+          cargo check --workspace --all-targets "${feature_flags[@]}"
+      - name: Feature configuration runtime coverage
+        if: ${{ github.event_name != 'pull_request' }}
+        run: echo "The full feature matrix runs tests for this event."
+
   test:
     name: Test ${{ matrix.state }} on ${{ matrix.target }}
+    if: ${{ github.event_name != 'pull_request' }}
     runs-on: ${{ matrix.os }}
     continue-on-error: true
     strategy:
@@ -95,14 +179,14 @@ jobs:
         target:
           - macOS
           - Windows
-        state:
-          - transparent
-          - Sapling-only
-          - Orchard
-          - all-pools
-          - all-features
-          - NU6.3
-          - NU7
+        state: >-
+          ${{
+            fromJSON(
+              github.event_name == 'schedule'
+              && '["all-features", "NU7"]'
+              || '["all-features"]'
+            )
+          }}
 
         include:
           - target: macOS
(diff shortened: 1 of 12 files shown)


zcash_client_sqlite/Cargo.toml
```diff
@@ -220,6 +220,9 @@ unstable = ["zcash_client_backend/unstable"]
 ## A feature used to isolate tests that are expensive to run. Test-only.
 expensive-tests = []
 
+## A feature used to ensure that expensive tests do not run with `--all-features`.
+no-expensive-tests = []
```
nullcopy: This breaks my brain. `--all-features` would enable `expensive-tests` and `no-expensive-tests` at the same time, but `no-expensive-tests` exists solely to modify `all-features` to not compile expensive tests.

Maybe it's just my own bias, but I've always considered cargo features to be add things, and not remove things... Maybe we can just chalk it up to my frustrations around cargo features in general.

nullcopy: I asked Claude to help flesh out my thoughts, and I found this helpful:

> The reason it "breaks your brain" is that it deliberately violates Cargo's additivity convention: enabling a feature is supposed to only add things, and no-expensive-tests exists to remove things. That's normally dangerous because of feature unification — if crate A wants expensive-tests and crate B enables no-expensive-tests, B silently wins for the whole build. But here both features are test-only flags on leaf crates gating #[test] items, not public API, so the worst-case failure mode is "some tests silently don't run in a build nobody executes tests from." That's precisely the intended effect for --all-features, and it's why the same pattern already ships on main in zcash_primitives and zcash_proofs — this PR just extends the established idiom to zcash_client_sqlite.


The "some tests silently don't run" scenario seems undesirable

.github/workflows/ci.yml
```diff
@@ -61,25 +54,70 @@ jobs:
...
+          for feature_set in "${feature_sets[@]}"; do
+            cargo check --workspace --features "${common_features[*]} ${feature_set}"
+          done
+      - name: Feature configurations are covered by runtime jobs
+        if: ${{ github.event_name != 'pull_request' }}
+        run: echo "The complete unique feature matrix runs for this event."
+
   test:
     name: Test ${{ matrix.state }} on ${{ matrix.target }}
+    if: ${{ github.event_name != 'pull_request' }}
```
nullcopy, line 174: I'm wondering if `test` and `test-slow` should be merged now. I don't understand why slow tests aren't run on all the other non-PR events

.github/workflows/ci.yml
```diff
@@ -133,79 +150,61 @@ jobs:
...
+          ${{ steps.prepare.outputs.feature-flags }} \
+          -- --test-threads=2
       - name: Verify working directory is clean
         run: git diff --exit-code
 
   test-slow:
-    name: Slow Test ${{ matrix.state }} on ${{ matrix.target }}
-    runs-on: ${{ matrix.os }}
+    name: Scheduled expensive tests
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
```
nullcopy: I generally like the idea of moving expensive tests to a nightly runner, but how do we know if the tests failed? Does it gate anything?

.github/workflows/ci.yml
```diff
@@ -61,25 +54,71 @@ jobs:
...
+        shell: bash
+        env:
+          FEATURE_FLAGS: ${{ steps.prepare.outputs.feature-flags }}
+        run: |
+          read -ra feature_flags <<< "$FEATURE_FLAGS"
+          cargo test --workspace "${feature_flags[@]}" -- --test-threads=2
       - name: Verify working directory is clean
         run: git diff --exit-code
 
+  feature-check:
```
nullcopy, line 115: From Claude:
> `feature-check` doesn't compile test code. `cargo check --workspace --features` ... covers libs/bins only. Partial-feature breakage in `#[cfg(test)]` modules or tests/ (historically the common failure mode for these configs — that's why the matrix existed) won't surface until the merge group, costing a full queue round-trip. Use --all-targets.

.github/workflows/ci.yml
```diff
@@ -61,25 +54,71 @@ jobs:
...
+          persist-credentials: false
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        if: ${{ github.event_name == 'pull_request' }}
+        with:
+          shared-key: msrv
+          save-if: "false"
+      - name: Check unique partial feature configurations
+        if: ${{ github.event_name == 'pull_request' }}
+        shell: bash
+        run: |
```
nullcopy, line 160: This duplicates the `prepare` action's feature list in a bash array. When someone adds a feature to .github/actions/prepare/action.yml, this list will diverge

.github/workflows/ci.yml
```diff
@@ -133,79 +151,62 @@ jobs:
...
         with:
           shared-key: msrv
           save-if: "false"
-      - name: Run slow tests
-        run: >
-          cargo test
-          --workspace
-          ${{ steps.prepare.outputs.feature-flags }}
-          --features expensive-tests
+      - name: Run legacy transaction round-trip properties
```
nullcopy: These substring filters exit 0 if they match zero tests. A rename of tx_serialization_roundtrip_* (or chain::tests, test_sprout_constraints below) would cause this check to silently pass without actually testing anything

.github/workflows/ci.yml
```diff
@@ -133,79 +151,62 @@ jobs:
...
+        run: |
+          cargo test \
+          -p zcash_primitives \
+          --features 'expensive-tests test-dependencies' \
+          transaction::tests::tx_serialization_roundtrip_ \
+          -- --test-threads=1
+      - name: Run Sprout constraint vectors
```
nullcopy: Since these steps are independent suites, could the later ones get `if: always()`? As written, a failure in the round-trip step skips the remaining three, so a fail at the beginning hides the other results which won't run again until the next night

.github/workflows/ci.yml
```diff
@@ -133,79 +151,62 @@ jobs:
...
+          read -ra feature_flags <<< "$FEATURE_FLAGS"
+          cargo test --workspace "${feature_flags[@]}" -- --test-threads=2
       - name: Verify working directory is clean
         run: git diff --exit-code
 
   test-slow:
-    name: Slow Test ${{ matrix.state }} on ${{ matrix.target }}
-    runs-on: ${{ matrix.os }}
+    name: Scheduled expensive tests
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
```
nullcopy: `tests-slow` was previously run for "nu7", but not anymore. Is that intentional?

.github/workflows/ci.yml
```diff
@@ -133,79 +151,62 @@ jobs:
...
+    runs-on: ${{ vars.UBUNTU_LARGE_RUNNER || 'ubuntu-latest' }}
     continue-on-error: true
```
nullcopy: Now that this job is the sole coverage for the expensive tests, `continue-on-error: true` means a scheduled failure never turns anything red. How do we know if a fail occurred?

nullcopy (changes requested):
Reviewed as of 9691c33eefffca654f103a6044b9f3de509427e4

I'm afraid this makes our tests more fragile in a few ways, and I'm not convinced that the benefit justifies that cost. CI time was particularly painful in the recent rush to deliver a LOT of changes in time for a VERY aggressive network upgrade... but that should be rare, right? In fact, if that situation happens again, I would rather just bypass CI for those urgent PRs, while still ensuring we have CI run for the actual release.

zcash_client_sqlite/Cargo.toml
nuttycom: Ah yeah, we definitely don't want features that violate additivity, that's always a recipe for pain.

.github/workflows/ci.yml
nuttycom: I agree, I don't think the "slow" classification is meaningful any more, at least as it is currently configured.

.github/workflows/ci.yml
nuttycom: I also had this question.

zcash_client_sqlite/Cargo.toml
nuttycom: What we could do here instead is use a `cfg` flag rather than a feature flag, like we do for `zcash_unstable`. But that would mean basically never running these locally unless you were specifically trying to test something that was under that flag. `cfg` flags aren't subject to the additivity rule, but they do apply throughout the *entire* dependency tree, not just the current workspace, which makes them a footgun. Still `cfg(zcash_slow_tests)` is an option we could consider.

.github/workflows/ci.yml
nuttycom: This has been a recurring problem with our feature flags. Once we're through the NU6.3 upgrade, one of the things I want to do is reconsider the `orchard` feature flag specifically. We need to look at what is needed in terms of our crate stack for Keystone, since they're orchard-only - I would like to eliminate redundancies between their stack and our crates, reorganizing stuff in our tree if necessary, but I think at this point with Ironwood formal verification making it THE POOL, the `orchard` feature's time of being useful is past.

nuttycom: We will probably keep the feature flags in `zcash_keys`, but only keep them there (or lower). And we will also keep `transparent-inputs`; turning off receiving to transparent is still something I want people to be able to do.

nuttycom: The counterargument to removing feature flags is that they will serve us well in the future when we specifically want to turn pools off. But maybe we handle that strictly via long-lived maintenance branches in the codebase, instead of feature-flagging?

zcash_client_sqlite/Cargo.toml
nuttycom: Replaced this with the additive `ignore-expensive-tests` feature. `expensive-tests` still controls compilation; the new feature uses `cfg_attr(..., ignore)` so `--all-features` compiles and lists the tests as ignored instead of removing them.

nuttycom: Agreed. The inverted feature is gone. I verified that `--all-features` reports each isolated test as ignored with an explicit reason, while the nightly-like feature set has no ignored copies.

.github/workflows/ci.yml
nuttycom: Reworked this. Expensive tests now run on every non-PR event (push, merge group, schedule, and manual dispatch), in full-workspace all-pools and NU7 configurations. The ordinary cross-platform job remains separate because its OS matrix and cache behavior differ.

.github/workflows/ci.yml
nuttycom: The expensive matrix no longer uses `continue-on-error`, so failures turn the workflow red and gate merge-group/main runs normally. Scheduled failures additionally notify Slack through `SLACK_WEBHOOK_URL`; the notification links directly to the failed run.

.github/workflows/ci.yml
nuttycom: Fixed: each pruned pull-request configuration now runs `cargo check --workspace --all-targets`, so test modules, integration tests, examples, benches, libraries, and binaries are compiled.

.github/workflows/ci.yml
nuttycom: Fixed: the compile matrix now calls the existing `prepare` action. Feature arguments are passed through an environment variable into a small validated wrapper, avoiding both list duplication and direct step-output interpolation in shell commands.

.github/workflows/ci.yml
nuttycom: Removed all substring filters. Each expensive configuration now runs `cargo test --workspace ... --features expensive-tests`, so renaming a test cannot silently remove it from coverage.

.github/workflows/ci.yml
nuttycom: The independent filtered steps have been replaced by a single full-workspace Cargo invocation per matrix configuration. Libtest runs the complete selected suite, so there are no later workflow steps for an early failure to skip.

.github/workflows/ci.yml
nuttycom: NU7 is restored in the expensive matrix. The two full-workspace configurations are all-pools and NU7, with the latter retaining `zcash_unstable="nu7"`.

.github/workflows/ci.yml
nuttycom: Fixed: `continue-on-error` was removed. Non-PR failures turn CI red, and scheduled failures also trigger the Slack notification job.

nuttycom:
Review changes are pushed in e3bd098173. The inverted feature is now additive, partial configurations compile all targets, expensive coverage runs as full-workspace all-pools and NU7 suites on every non-PR event, and scheduled failures notify Slack. I also removed direct step-output interpolation from the modified jobs and pinned the new Slack action; local zizmor reports no new findings from this patch. Ready for another look once CI settles.

nuttycom:
Rebased onto current `main` and force-pushed the PR as a single commit, 78c8741dbc. NU7 runtime coverage is now nightly-only: regular Linux, Windows, and expensive NU7 jobs run on `schedule`, and NU7 is absent from pull-request, push, merge-group, and manual-dispatch matrices. The squashed workflow has three fewer zizmor template-injection findings than current `main` and introduces none; its remaining findings are unchanged upstream lines covered by #2843.

nuttycom:
This is now a true stacked PR on #2854. The overlap was resolved using #2854’s structured `feature-list` / `all-features` outputs, so the extra wrapper is no longer needed. #2842 remains one commit (`d9844f6e88`) above #2854, and the combined tree has zero zizmor findings.

nuttycom:
Rebased the single #2842 commit onto amended #2854 after its output-interface cleanup. New head: e5fa382245. The combined stacked tree passes YAML parsing, formatting, all-feature/all-target compilation, and the full zizmor audit with zero findings.

.github/workflows/ci.yml
```diff
@@ -19,40 +34,15 @@ jobs:
...
-          - transparent
-          - transparent-inputs-only
-          - transparent-key-encoding-only
-          - Sapling-only
-          - Orchard
-          - all-pools
-          - all-features
-          - NU6.3
-          - NU7
+        state: ${{ fromJSON(github.event_name == 'schedule' && '["transparent", "transparent-inputs-only", "transparent-key-encoding-only", "Sapling-only", "Orchard", "all-pools", "all-features", "NU7"]' || (github.event_name == 'pull_request' && '["all-features"]' || '["transparent", "transparent-inputs-only", "transparent-key-encoding-only", "Sapling-only", "Orchard", "all-pools", "all-features"]')) }}
```
nuttycom: Is it at all possible to format this more legibly?

.github/workflows/ci.yml
nuttycom: Will these post to Slack when they fail as well?

nuttycom (commented):
self-utACK

nullcopy (approved):
utACK ecfab77e139b7e54e8658427e4de4a09f5e4e574

nuttycom:
Superseded by #2858, which uses the identical head commit (`ecfab77e139b7e54e8658427e4de4a09f5e4e574`) from an upstream-owned branch and targets #2854’s `ci/fix-zizmor-alerts` branch as a proper stacked PR.

PR #2843, opened 2026-07-29
ci: resolve zizmor security findings

nuttycom:
## Summary

- pass generated feature flags and matrix package names through environment variables instead of expanding them directly into shell source
- pin every `dtolnay/rust-toolchain` invocation to the reviewed master commit and select stable, beta, or nightly through the explicit `toolchain` input
- apply a seven-day cooldown to routine Dependabot GitHub Actions updates

## Security impact

This resolves all currently open zizmor template-injection, unpinned-action, and Dependabot cooldown findings. It also fixes the same patterns in the mutation workflow before they become separate alerts.

## Validation

- `uvx zizmor==1.28.0 .github` — no findings
- parsed every workflow and `dependabot.yml` as YAML
- `git diff --check`

The cooldown applies to routine version updates; Dependabot security updates are unaffected.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -29,6 +29,12 @@ outputs:
   feature-flags:
     description: 'Feature flags'
     value: ${{ steps.prepare.outputs.flags || steps.prepare-all.outputs.flags }}
+  feature-list:
+    description: 'Space-separated feature names, empty when all features are enabled'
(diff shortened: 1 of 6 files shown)


nuttycom:
Superseded by #2854, which uses the identical commit (`f49137dc4f1f67c960f019c1e7900c2aee23506d`) from an upstream-owned branch so dependent CI work can use it as a clean stacked-PR base.

Issue #2848, opened 2026-07-29
Abandon the canonical ZIP 318 crossing attempt when the bucketed anchor's checkpoint is missing


nuttycom:
## Summary

The fix for #2846 prevents new holes in the retained anchor grid, but wallets that scanned past NU6.3 activation before it permanently lack the checkpoints at grid boundaries whose blocks were shielded-empty. A canonical ZIP 318 crossing anchored at such a boundary is proposed successfully (note eligibility is a height comparison) and then fails at build time with `ProposalError::AnchorNotFound`.

**Backfilling the missing checkpoints is not reliably possible from local state.** The note commitment tree prunes interior nodes that no retained checkpoint or mark references, so by the time a hole is discovered the node data a late-added checkpoint would need may already be gone; reconstructing it would require querying a light wallet server for subtree data. This is therefore not a database migration.

## Approach

Abandon the canonical attempt immediately when the wallet holds no checkpoint at the chosen boundary, falling back to an ordinary (non-canonical, padded) crossing exactly as when no sufficiently-old note exists. Since #2846's fix repairs checkpoint storage going forward, an upgraded wallet only lacks the current age-1 boundary until newly scanned boundaries rotate into position — on the order of one grid interval (~3 hours) after upgrade — so a brief window of non-canonical crossings is an acceptable degradation, and the failure mode changes from a hard send error to the same graceful fallback used everywhere else in the canonical path.

*Filed with Claude Code assistance.*

PR #2849, opened 2026-07-29
zcash_client_backend: checkpoint empty anchor-retention boundaries

### LINKED ISSUE #2846: Anchor-retention grid boundaries falling on shielded-empty blocks are never checkpointed
## Summary

Once NU6.3 is active, `put_blocks` retains note-commitment-tree checkpoints on the anchor-retention grid (the ZIP 318 anchor bucket boundaries) so that pool-crossing transfers can be proved against them. However, scanning only creates a checkpoint for a block via that block's **last note commitment** (`zcash_client_backend/src/scanning.rs`), and the cross-pool ensure step in `put_blocks` only propagates heights that are checkpointed in *some* pool. A block containing **no shielded outputs in any pool** therefore never receives a checkpoint in any tree, and `ensure_retained` is never invoked for its height: a grid boundary landing on such a block is permanently absent from the retained set.

Block-by-block scanning masks the defect — every batch checkpoints its starting frontier, so each height is checkpointed as the next block's batch begins — which is why the existing test suite (whose filler blocks each deliberately carry a commitment) never caught it. Real wallets scan in multi-block batches and exhibit it.

## Evidence (mainnet)

Post-NU6.3 grid boundaries in a wallet synced over this range, against the shielded output counts of the boundary blocks themselves (chain-global facts, independently verifiable):

| Boundary | Shielded outputs in that block | Retained? |
|---|---|---|
| 3428208 | 12 Orchard + 2 Ironwood | yes |
| 3428352 | 16 Orchard | yes |
| 3428496 | none in any pool | **no** |
| 3428640 | none in any pool | **no** |
| 3428784 | 18 Orchard + 2 Ironwood | yes |
| 3428928 | none in any pool | **no** |
| 3429072 | none in any pool | **no** |
| 3429216 | 38 Orchard + 8 Ironwood | yes |

The correlation is exact: every missing boundary block is shielded-empty. Since emptiness is a chain property, **every wallet scanning those ranges loses the same boundaries**, thinning the shared ZIP 318 anchor grid for all participants.

## Consequences

- Note eligibility at a bucketed anchor is a height comparison; neither proposal creation nor the canonical-crossing check verifies that the anchor's checkpoint exists. A canonical ZIP 318 crossing (or a real migration transfer) anchored at a missing boundary is therefore proposed successfully and then **fails hard at build time with `ProposalError::AnchorNotFound`**.
- The tree state at a boundary height is well defined regardless of whether the boundary block added commitments (it is the state as of the last commitment at or before it), so there is no protocol reason for the hole.

Related: #2700 introduced the retention machinery; this is a residual gap in it.

Existing wallets scanned before a fix retain the holes permanently; repairing them is tracked separately.

nuttycom:
Closes #2846.

Scanning attaches a checkpoint to a block's last note commitment, so a block containing no shielded outputs in any pool produced no checkpoint at all, and the cross-pool ensure step had nothing to copy. An anchor-retention grid boundary landing on such a block was therefore never checkpointed: the retained grid gained a permanent hole, and a ZIP 318 crossing anchored there failed at build time with `ProposalError::AnchorNotFound`. On the affected mainnet range half the post-activation boundaries were lost this way (see the issue for the block-by-block evidence).

`put_blocks` now adds every retained height in the scanned range to the checkpoint-ensure sets of all three pools; the ensured checkpoint takes the tree position of the last commitment at or before the boundary, which is exactly the state a ZIP 318 anchor at that height commits to. The new `AnchorRetention::retained_in_range` enumerates those heights.

Block-by-block scanning masked the defect in the test suite — every batch checkpoints its starting frontier — so the new tests scan their boundary-crossing ranges in a single batch with the boundary blocks generated empty:

- `empty_boundary_blocks_are_checkpointed_and_retained` (Sapling and Orchard variants) asserts the boundary checkpoints exist, are retained, and support witness construction — and that the Ironwood tree, which holds no commitments at all, retains the same heights.
- `canonical_crossing_builds_at_empty_boundary_block` proposes and **builds** a canonical crossing whose bucketed anchor is an empty boundary block; without the fix it reproduces the `AnchorNotFound` failure exactly.

Repair of pre-existing holes in already-synced wallets is tracked in #2848.

### DIFF
--- a/zcash_client_backend/src/data_api/anchor_retention.rs
+++ b/zcash_client_backend/src/data_api/anchor_retention.rs
@@ -92,6 +92,40 @@ impl AnchorRetention {
     pub fn retains(&self, height: BlockHeight) -> bool {
         height >= self.from_height && self.intervals.iter().any(|i| i.is_boundary(height))
     }
+
+    /// Returns every height in `range` whose checkpoint this policy retains, ascending.
+    ///
+    /// This is the enumeration form of [`Self::retains`]: a height is in the result exactly when
+    /// `range` contains it and [`Self::retains`] holds for it. It exists so that a caller adding a
+    /// scanned range of blocks can CREATE the checkpoints the policy will need retained — a
+    /// boundary block containing no note commitments produces no checkpoint of its own, and a
+    /// policy can only keep alive a checkpoint that exists.
+    pub fn retained_in_range(
+        &self,
+        range: core::ops::RangeInclusive<BlockHeight>,
+    ) -> BTreeSet<BlockHeight> {
+        let start = u64::from(u32::from(core::cmp::max(*range.start(), self.from_height)));
+        let end = u64::from(u32::from(*range.end()));
+        self.intervals
+            .iter()
+            .flat_map(|interval| {
+                // A height is a boundary of an interval exactly when it is a multiple of it, so
+                // the boundaries in range are the multiples of `step` from the first at or above
+                // `start`. The arithmetic is in `u64` so that neither the round-up nor the
+                // iterator's one-past-the-end probe can overflow near the top of the height range.
+                let step = u64::from(interval.block_count().get());
+                let first = start.div_ceil(step) * step;
+                (0u64..)
+                    .map(move |k| first + k * step)
+                    .take_while(move |boundary| *boundary <= end)
+                    .map(|boundary| {
+                        BlockHeight::from_u32(
+                            u32::try_from(boundary).expect("bounded by a u32 height"),
+                        )
+                    })
+            })
+            .collect()
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


zcash_client_backend/src/data_api/anchor_retention.rs
```diff
@@ -92,6 +92,40 @@ impl AnchorRetention {
...
+    /// This is the enumeration form of [`Self::retains`]: a height is in the result exactly when
+    /// `range` contains it and [`Self::retains`] holds for it. It exists so that a caller adding a
+    /// scanned range of blocks can CREATE the checkpoints the policy will need retained — a
+    /// boundary block containing no note commitments produces no checkpoint of its own, and a
+    /// policy can only keep alive a checkpoint that exists.
+    pub fn retained_in_range(
+        &self,
+        range: core::ops::RangeInclusive<BlockHeight>,
+    ) -> BTreeSet<BlockHeight> {
+        let start = u64::from(u32::from(core::cmp::max(*range.start(), self.from_height)));
```
nullcopy, line 107: If they request a range which starts before `self.from_height`, should that be an error?

nullcopy (commented):
Flushing comments as of 81c5a8bbb90ad8aaa31774cd1e61e0f042e03f3c

zcash_client_backend/src/data_api/anchor_retention.rs
nullcopy: nvm, thinking about this more, silently clipping to `self.start` is fine

nullcopy (approved):
utACK 81c5a8bbb90ad8aaa31774cd1e61e0f042e03f3c

PR #2850, opened 2026-07-29
Migration-shaped note selection: chain-age order, single-note canonical funding

### LINKED ISSUE #2847: Note selection: order by chain age and prefer single-note funding for canonical ZIP 318 crossings
## Summary

Two gaps in note selection prevent ordinary wallets from reliably producing canonical ZIP 318 pool crossings:

1. **Selection order is discovery order, not age.** The running-sum window in `zcash_client_sqlite`'s `select_spendable_notes` accumulates notes in row-id order (and carries no window `ORDER BY` at all in the unfiltered case). Row ids are assigned in *scan* order, and priority scanning visits recent blocks before back-filling history, so a restored wallet's newest notes carry its lowest ids: selection spends freshly received change while far older notes sit idle — despite the query's stated intent to select the oldest notes first. The correct age key is the note's commitment tree position, which is assigned in strict chain order.

2. **No single-note preference for canonical crossings.** A ZIP 318 migration transfer spends exactly one note. The canonical-crossing attempt in `propose_transfer` runs ordinary greedy accumulation, which reaches the target through several small notes whenever the oldest notes are small; the resulting multi-input proposal is then discarded as non-canonical and the payment is built padded, even when the wallet holds a single note that could have funded it. The attempt should prefer the oldest single note covering the payment plus its fee — falling back to accumulation when none exists — so that the canonical shape is produced whenever it is achievable.

Age *filtering* needs no new machinery: the bucketed anchor already raises the confirmation floor so only sufficiently-old notes are eligible. What is missing is age *ordering* and single-note *preference*.

nuttycom:
Closes #2847. Stacked on #2849; only the top two commits are new here, and this should be rebased onto `main` once #2849 merges.

Two changes that together let ordinary wallets reliably produce canonical ZIP 318 crossings:

**`zcash_client_sqlite`: order note selection by chain age, not discovery.** The running-sum selection window ordered by row id (or not at all in the unfiltered case). Ids are assigned in scan order, and priority scanning visits recent blocks before back-filling history, so a restored wallet's newest notes carry its lowest ids: selection spent freshly received change while far older notes sat idle. The window now orders by note commitment tree position — assigned in strict chain order — with the lock-tier key, where present, remaining the leading key. The new test scans two notes out of chain order (so the newer note takes the lower id) and asserts that a selection either note could satisfy returns the older one.

**`zcash_client_backend`: fund canonical crossings from a single note.** A migration transfer spends exactly one note, but the canonical-crossing attempt ran ordinary greedy accumulation, which reaches the target through several small notes whenever the oldest notes are small — funding the payment while losing the canonical shape. `SpendPolicy` now carries a `NoteSelection` preference; under `PreferSingle` the selector first asks the data source for the single oldest note covering the required amount alone (the new `InputSource::select_single_spendable_note`, with a best-effort default and a direct-query sqlite implementation sharing the eligibility CTE), falling back to accumulation when no pool holds one. `propose_transfer` sets the preference only on its canonical attempt, so no changes are needed in the FFI, the mobile SDKs, or applications. The new test funds a wallet with two older small notes plus one larger covering note and asserts the canonical payment takes the single-note shape while a non-canonical amount still accumulates.

Age *filtering* required no new code: the bucketed anchor already raises the confirmation floor so only sufficiently-old notes are eligible for the canonical attempt.

### DIFF
--- a/zcash_client_sqlite/src/testing/db.rs
+++ b/zcash_client_sqlite/src/testing/db.rs
@@ -42,6 +42,7 @@ use zcash_primitives::{
 };
 use zcash_protocol::{
     ShieldedPool, consensus, consensus::BlockHeight, local_consensus::LocalNetwork, memo::Memo,
+    value::Zatoshis,
 };
 use zip32::DiversifierIndex;
 
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -12,6 +12,18 @@ workspace.
 
 ### Added
 - `zcash_client_backend::data_api::anchor_retention::AnchorRetention::retained_in_range`
+- `zcash_client_backend::data_api::wallet::input_selection::NoteSelection`
+- `zcash_client_backend::data_api::wallet::input_selection::SpendPolicy::{with_note_selection, note_selection}`
+- `zcash_client_backend::data_api::InputSource::select_single_spendable_note`, with a
+  best-effort default implementation
+- `zcash_client_backend::data_api::ReceivedNotes::{is_empty, into_single_covering}`
+
+### Changed
+- `zcash_client_backend::data_api::wallet::propose_transfer` now funds a canonical
+  ZIP 318 crossing attempt from the single oldest Orchard note that covers the
+  payment and its fee, falling back to ordinary accumulation when no such note
+  exists. Payments of canonical denominations that previously lost the canonical
+  shape to multi-note funding now take it whenever a single covering note exists.
 
 ### Fixed
 - `zcash_client_backend::data_api::ll::wallet::put_blocks` now creates (and retains) a
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1171,6 +1171,75 @@ impl<NoteRef> ReceivedNotes<NoteRef> {
             .ok_or(BalanceError::Overflow);
     }
 
+    /// Returns whether the collection contains no notes in any pool.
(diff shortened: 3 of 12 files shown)


nuttycom:
The single-note query should preserve the caller's lock-tier preference. `ValueSelection::SingleCovering` currently orders candidates only by `commitment_tree_position`, so when `PreferUnlocked` or `PreferLocked` admits both tiers, `.next()` can select an older note from the non-preferred tier even when the preferred tier contains a covering note. In particular, `PreferUnlocked` may reuse an acknowledged in-flight locked input despite an unlocked alternative. Please order by the lock tier first and chain age second, matching the accumulation path, and add coverage for single-note selection under both lock-preference policies.

Relevant code: `zcash_client_sqlite/src/wallet/common.rs`, the `ValueSelection::SingleCovering` query.

zcash_client_sqlite/src/testing/db.rs
```diff
@@ -42,6 +42,7 @@ use zcash_primitives::{
 };
 use zcash_protocol::{
     ShieldedPool, consensus, consensus::BlockHeight, local_consensus::LocalNetwork, memo::Memo,
+    value::Zatoshis,
```
schell, line 45: Why was this needed?

schell (approved):
Seems good. I thought I smelled something at first with ordering by `commitment_tree_position` because it wasn't obvious that `commitment_tree_position` was non-null. I had to trace down the paths that set it. There's also an import of `Zatoshis` that I'm surprised doesn't cause a clippy warning.

PR #2851, opened 2026-07-29
zcash_client_backend: model dummy outputs in proposals

Cosmos-Harry:
## Summary

Models the exact number of dummy outputs for the Sapling, Orchard, and Ironwood bundles in transaction proposals, and preserves those counts across proposal serialization.

## Root cause

A canonical Orchard-to-Ironwood crossing has no Ironwood dummy output and is priced with one Ironwood action (15,000 zats). Proposal protobuf serialization previously omitted this transaction shape. On deserialization, the builder used the default two-action shape, which added one dummy output, required 20,000 zats, and failed with an apparent 5,000-zat insufficient-funds error.

This affects ordinary software-wallet sends as well: they cross the proposal protobuf boundary before being built as PCZTs. Older proposals that omit the new counts retain the previous default behavior.

## Validation

- `cargo test -p zcash_client_backend --lib --features orchard,pczt`
- `cargo test -p zcash_client_sqlite canonical_crossing_is_bucketed_and_unpadded --features orchard`
- Reproduced with the affected mobile wallet database and recipient: proposal creation, PCZT creation, and transaction authorization all succeeded with this change (without broadcast).

### DIFF
--- a/zcash_client_backend/proto/proposal.proto
+++ b/zcash_client_backend/proto/proposal.proto
@@ -139,6 +139,16 @@ message TransactionBalance {
     repeated ChangeValue proposedChange = 1;
     // The fee to be paid by the proposed transaction, in zatoshis.
     uint64 feeRequired = 2;
+    // The exact number of dummy outputs in each shielded bundle. Omitted by
+    // proposals created before transaction shape was modelled explicitly.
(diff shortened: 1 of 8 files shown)


nuttycom (dismissed):

nuttycom (approved):
utACK

PR #2852, opened 2026-07-29
zcash_client_backend: abandon canonical crossings without a provable anchor

### LINKED ISSUE #2848: Abandon the canonical ZIP 318 crossing attempt when the bucketed anchor's checkpoint is missing

nuttycom:
Closes #2848. Stacked on #2850 (and transitively #2849); only the top commit is new here, and this should be rebased as the stack beneath it merges.

Note eligibility at the bucketed anchor is a height comparison, so a canonical ZIP 318 crossing proposal was kept without checking that the wallet actually holds the note commitment tree checkpoint its anchor needs; building then failed hard with `ProposalError::AnchorNotFound`. Wallets that scanned past NU6.3 activation before #2849's fix are permanently missing the boundaries whose blocks carried no shielded outputs, and the holes cannot be backfilled from local state — the tree prunes node data that no retained checkpoint references, so reconstruction would require refetching subtree data from a light wallet server. That is why this is a proposal-time gate rather than a repair migration.

`propose_transfer` now consults the new `WalletRead::has_tree_checkpoint` (defaulting to `true` to preserve behavior for backends that cannot answer; answered directly from the checkpoints table in `zcash_client_sqlite`) and abandons the canonical attempt when the boundary's checkpoint is absent — the same graceful fallback to an ordinary, padded crossing taken when no sufficiently-old note exists. Under repaired retention the degradation is temporary: freshly scanned boundaries rotate into the age-1 position within about a grid interval (~3 hours) of upgrading.

The test (`canonical_crossing_abandoned_without_anchor_checkpoint`) drives the canonical path once to bind the boundary the production arithmetic chooses, deletes that boundary's checkpoint records through a backend-supplied hole-punching callback to simulate the legacy wallet, and asserts the same payment then proposes as an ordinary crossing and **builds** — without the gate it reproduces the `AnchorNotFound` failure.

### DIFF
--- a/zcash_client_backend/CHANGELOG.md
+++ b/zcash_client_backend/CHANGELOG.md
@@ -17,6 +17,7 @@ workspace.
 - `zcash_client_backend::data_api::InputSource::select_single_spendable_note`, with a
   best-effort default implementation
 - `zcash_client_backend::data_api::ReceivedNotes::{is_empty, into_single_covering}`
+- `zcash_client_backend::data_api::InputSource::anchor_computable`
 
 ### Changed
 - `zcash_client_backend::data_api::wallet::propose_transfer` now funds a canonical
@@ -24,6 +25,11 @@ workspace.
   payment and its fee, falling back to ordinary accumulation when no such note
   exists. Payments of canonical denominations that previously lost the canonical
   shape to multi-note funding now take it whenever a single covering note exists.
+- `zcash_client_backend::data_api::wallet::propose_transfer` abandons the canonical
+  ZIP 318 crossing attempt when no anchor is computable at the bucketed boundary
+  height (per `InputSource::anchor_computable`), proposing an ordinary crossing
+  instead of a canonical proposal whose build would fail with
+  `ProposalError::AnchorNotFound`.
 
 ### Fixed
 - `zcash_client_backend::data_api::ll::wallet::put_blocks` now creates (and retains) a
--- a/zcash_client_backend/src/data_api.rs
+++ b/zcash_client_backend/src/data_api.rs
@@ -1781,6 +1781,23 @@ pub trait InputSource {
         lock_filter: LockFilter<'_>,
     ) -> Result<Option<ReceivedNote<Self::NoteRef, Note>>, Self::Error>;
 
+    /// Returns whether an anchor is COMPUTABLE at `height` for spends from the given pool: whether
+    /// this data source can produce the note commitment tree root, and witnesses to it, as of the
+    /// end of that block.
+    ///
+    /// A height inside the wallet's scanned range need not qualify: tree states are only
+    /// materialized at the heights the wallet chose to retain, and a wallet that scanned past
+    /// NU6.3 activation before boundary checkpointing was repaired is permanently missing the
+    /// anchor-retention boundaries whose blocks carried no shielded outputs. Such a hole cannot be
+    /// backfilled from local state, so a caller deciding whether to anchor at a retained boundary
+    /// should consult this before committing to it, and fall back rather than propose a
+    /// transaction that cannot be built.
+    fn anchor_computable(
+        &self,
+        protocol: ShieldedPool,
+        height: BlockHeight,
+    ) -> Result<bool, Self::Error>;
+
     /// Returns a list of spendable notes sufficient to cover the specified target value, if
     /// possible. Only spendable notes corresponding to the specified shielded protocol will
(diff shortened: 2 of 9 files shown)


zcash_client_backend/src/data_api.rs
```diff
@@ -2219,6 +2219,30 @@ pub trait WalletRead {
...
+    /// missing the grid boundaries whose blocks carried no shielded outputs, and such a hole
+    /// cannot be backfilled from local state, because the tree prunes node data that no retained
+    /// checkpoint references. A caller deciding whether to anchor at a retained boundary should
+    /// therefore consult this before committing to it, and fall back rather than propose a
+    /// transaction that cannot be built.
+    ///
+    /// The default implementation returns `true`, preserving the historical behavior for backends
+    /// that cannot answer; such backends retain the risk that building at a missing anchor fails
+    /// with an error instead of falling back.
+    fn has_tree_checkpoint(
```
nuttycom: This looks like it should be on `WalletCommitmentTrees` instead. Also, it should not have a default implementation.

zcash_client_backend/CHANGELOG.md
```diff
@@ -17,13 +17,19 @@ workspace.
 - `zcash_client_backend::data_api::InputSource::select_single_spendable_note`, with a
   best-effort default implementation
 - `zcash_client_backend::data_api::ReceivedNotes::{is_empty, into_single_covering}`
+- `zcash_client_backend::data_api::InputSource::anchor_computable`
```
nuttycom, line 20: This is a change to a trait (semver break), not an addition.

nuttycom (commented):
self-utACK

zcash_client_backend/CHANGELOG.md
nuttycom: I will fix this in the release PR.

zcash_client_backend/src/data_api.rs
```diff
@@ -1781,6 +1781,23 @@ pub trait InputSource {
...
+    /// NU6.3 activation before boundary checkpointing was repaired is permanently missing the
+    /// anchor-retention boundaries whose blocks carried no shielded outputs. Such a hole cannot be
+    /// backfilled from local state, so a caller deciding whether to anchor at a retained boundary
+    /// should consult this before committing to it, and fall back rather than propose a
+    /// transaction that cannot be built.
+    fn anchor_computable(
+        &self,
+        protocol: ShieldedPool,
+        height: BlockHeight,
+    ) -> Result<bool, Self::Error>;
```
schell, line 1799: It's hard for me to tell what happened, but it seems like the intention was to put this on `WalletRead` as `WalletRead::has_tree_checkpoint` and give it a default implementation. Is that not the case?

nullcopy (approved):
utACK 67cf73ed177e233f6307f19e8215defc904ada18

zcash_client_backend/src/data_api.rs
nuttycom: `WalletRead` is kind of the wrong place for it - this is metadata about the wallet's view what it has scanned, so I think it belongs with InputSource than WalletRead.

nuttycom: `InputSource` is also what is used by the `InputSelector` machinery, which is what consumes this.

PR #2854, opened 2026-07-29
ci: resolve zizmor security findings

nuttycom:
## Summary

- pass generated feature flags and matrix package names through environment variables instead of expanding them directly into shell source
- pin every `dtolnay/rust-toolchain` invocation to the reviewed master commit and select stable, beta, or nightly through the explicit `toolchain` input
- apply a seven-day cooldown to routine Dependabot GitHub Actions updates

## Security impact

This resolves all currently open zizmor template-injection, unpinned-action, and Dependabot cooldown findings. It also fixes the same patterns in the mutation workflow before they become separate alerts.

## Validation

- `uvx zizmor==1.28.0 .github` — no findings
- parsed every workflow and `dependabot.yml` as YAML
- `git diff --check`

The cooldown applies to routine version updates; Dependabot security updates are unaffected.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -26,9 +26,12 @@ inputs:
     required: false
     default: true
 outputs:
-  feature-flags:
-    description: 'Feature flags'
-    value: ${{ steps.prepare.outputs.flags || steps.prepare-all.outputs.flags }}
+  feature-list:
+    description: 'Space-separated feature names, empty when all features are enabled'
+    value: ${{ steps.prepare.outputs.features }}
+  all-features:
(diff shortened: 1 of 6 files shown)


nuttycom:
Follow-up review cleanup is included in ec3890a6a3: removed the obsolete quoted `feature-flags` output, its `flags` writes, and the now-unused `prepare-all` helper. All workflow consumers use only the structured `feature-list` and `all-features` outputs. Full `.github` zizmor audit remains clean.

nuttycom (commented):
self-utACK

nullcopy (approved):
utACK ec3890a6a30c2e2e959591650d962698cfc67cb3

PR #2855, opened 2026-07-30
Adopt the revised ZIP 318 migration timing

nuttycom:
Adopts the parameter revisions from zcash/zips#1343:

- `TRANSFER_DELAY_MEAN` is now 66 blocks (previously 144); the 576-block cap is
  unchanged. Since the caps are no longer a fixed multiple of the means
  (576/66 ≈ 8.7 for transfers, 96/16 = 6 for preparations), `DELAY_CAP_RATIO`
  and `PREP_MEAN_DIVISOR` are removed, and
  `SchedulingParams::new_with_default_distributions` now scales each ZIP 318
  delay mean and cap by the ratio the given interval bears to the ZIP 318 one
  (reproducing `SchedulingParams::ZIP_318` exactly at the standard 144-block
  interval).
- `ANCHOR_AGE_CAP` is now 4 boundaries (previously 16).
- ZIP 318 specifies the preparation delay, so those constants move from
  `zcash_pool_migration` to `zcash_protocol::zip318` as `PREP_DELAY_MEAN`
  (16 blocks, previously a provisional 24) and `PREP_DELAY_CAP` (96 blocks),
  with a matching `PoolMigrationConstants::preparation_delay` accessor.

The revision's relaxation of the anchor-age lower bound from $a \geq 1$ to
$a \geq 0$ is deliberately **not** adopted: anchoring to the most recent
boundary would let a light wallet server use IP address correlation to link a
wallet's sync traffic to a transfer broadcast near a boundary. See
https://github.com/zcash/zips/pull/1343#issuecomment-5124302101 for details.

Also fixes the `zcash_client_sqlite` `pczt-tests` feature to enable the
`orchard` and `transparent-inputs` features it requires: enabling it without
both previously failed to compile, including under the exact feature set
listed in the migration proving simulation's `#![cfg]` gate.

### DIFF
--- a/components/zcash_protocol/CHANGELOG.md
+++ b/components/zcash_protocol/CHANGELOG.md
@@ -10,6 +10,21 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- `zcash_protocol::zip318::{PREP_DELAY_MEAN, PREP_DELAY_CAP}`, the ZIP 318
+  preparation inter-arrival delay distribution.
+- `zcash_protocol::zip318::PoolMigrationConstants::preparation_delay`
+
+### Changed
+- `zcash_protocol::zip318::TRANSFER_DELAY_MEAN` is now 66 blocks (previously
+  144) and `zcash_protocol::zip318::ANCHOR_AGE_CAP` is now 4 boundaries
+  (previously 16), adopting the revised ZIP 318 migration timing.
+
+### Removed
+- `zcash_protocol::zip318::DELAY_CAP_RATIO`. ZIP 318 no longer relates each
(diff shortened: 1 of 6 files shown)


nuttycom (commented):
self-utACK; note that we retain the minimum anchor age as being >= 1 period, see https://github.com/zcash/zips/pull/1343#issuecomment-5124302101

nullcopy (dismissed):
utACK 61e93ea244e823c927a5801f8f744332b85d4d15

nullcopy (approved):
utACK f9566d9d84368836c1c111d83f7cce0f1c11aa28

PR #2856, opened 2026-07-30
zcash_client_sqlite: preserve lock-tier preference in single-note selection

nuttycom:
Addresses the post-merge review feedback on #2850: https://github.com/zcash/librustzcash/pull/2850#issuecomment-5124366640

The `ValueSelection::SingleCovering` query ordered its candidates only by `commitment_tree_position`, so when `PreferUnlocked` or `PreferLocked` admitted both lock tiers, the single oldest covering note could come from the non-preferred tier — in particular, `PreferUnlocked` could reuse an acknowledged in-flight locked input despite an unlocked alternative.

The lock tier is now the primary sort key and chain age the secondary, matching the accumulation path's window order. Because the single-covering tail orders over the CTE (where the notes-table alias is out of scope), the tier expression is materialized as a CTE column and the sort direction applied at the ordering site: `locked_tier_expr` replaces `locked_tier_order_key`, returning the expression and direction separately, with the former composition inlined at the callers that still order on the expression directly.

The new scenario, `single_note_selection_honors_lock_tier_preference` (registered for both shielded pools), locks the older of two covering notes and asserts that single-note selection returns the unlocked note under `PreferUnlocked`, the locked note under `PreferLocked`, and the unlocked note under `Exclude`. Without the fix it fails exactly as the review describes: `PreferUnlocked` reaches for the older locked note.

### DIFF
--- a/zcash_client_backend/src/data_api/testing/pool/locking.rs
+++ b/zcash_client_backend/src/data_api/testing/pool/locking.rs
@@ -14,19 +14,18 @@ use zcash_protocol::{PoolType, TxId, consensus::BlockHeight, value::Zatoshis};
 use zip321::Payment;
 
 #[cfg(feature = "transparent-inputs")]
-use crate::data_api::{
-    InputSource, WalletRead, WalletWrite,
-    wallet::{
-        TargetHeight,
-        input_selection::{LockFilter, LockedInputPolicy},
-    },
-};
+use crate::data_api::WalletWrite;
 use crate::{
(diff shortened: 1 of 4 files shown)


nuttycom (commented):
self-utACK

nullcopy (approved):
Does this need a CHANGELOG entry/tweak?
Code looks good to me

utACK a8ba05fcaf61b15bb8f684f368bab1e4ef67eade

PR #2857, opened 2026-07-30
zcash_client_sqlite: report transparent outputs at their transparent receiver in v_tx_outputs

### LINKED ISSUE #2845: zcash_client_sqlite: `v_tx_outputs.to_address` reports the enclosing unified address for transparent outputs received by the wallet
## Summary

For transparent outputs received by the wallet at an external-scope receiver, `v_tx_outputs.to_address` reports the **enclosing unified address** rather than the transparent address the output actually paid. For wallet-internal transparent sends (a wallet paying one of its own transparent receivers) this is compounded by the view's dedup step: the sent-side row carries the correct `t…` address from `sent_notes.to_address`, but the merge takes `MAX(to_address)`, and under SQLite's BINARY collation `'u'` (0x75) > `'t'` (0x74), so the unified address deterministically shadows the transparent one (mainnet `u1…` > `t1…`, testnet `utest1…` > `tm…`).

The user-visible symptom in a wallet frontend that surfaces `to_address`: send funds to one of your own transparent addresses, and the transaction detail screen shows one of your unified addresses as the recipient.

## Root cause

Two independent defects in [`VIEW_TX_OUTPUTS`](https://github.com/zcash/librustzcash/blob/732654e15d69ee1cb9b282bc6cace1e6f73ff96a/zcash_client_sqlite/src/wallet/db.rs#L1542-L1608):

**1. The received-outputs arm resolves the address through `addresses.address`, which is pool-inappropriate for transparent outputs.**

```sql
    SELECT ...
           a.address                    AS to_address,
           ...
    FROM v_received_outputs ro
    ...
    LEFT JOIN addresses a ON a.id = ro.address_id
```

For an external-scope transparent receiver, the `addresses` row stores the **UA** in `address`; the transparent receiver itself lives only in [`cached_transparent_receiver_address`](https://github.com/zcash/librustzcash/blob/732654e15d69ee1cb9b282bc6cace1e6f73ff96a/zcash_client_sqlite/src/wallet/db.rs#L161-L166). So every transparent output received at such a receiver — not just self-sends — reports the UA. (Ironically, `transparent_received_outputs.address` holds the correct `t…` string, written by [`put_transparent_output`](https://github.com/zcash/librustzcash/blob/732654e15d69ee1cb9b282bc6cace1e6f73ff96a/zcash_client_sqlite/src/wallet/transparent.rs#L2784-L2815), but `v_received_outputs` forwards only `address_id`, so the view never sees it.)

This also contradicts the view's own column documentation ("`address`: … for received outputs, this is the address at which the output was received"): the address at which a transparent output is received is the transparent receiver, not the UA that contains it.

**2. The sent/received dedup merges `to_address` by lexicographic `MAX`.**

```sql
-- merge duplicate rows while retaining maximum information
SELECT ...
    MAX(to_address)             AS to_address,
    ...
FROM unioned
GROUP BY transaction_id, output_pool, output_index
```

A wallet-internal transparent send produces two rows for the same `(transaction_id, 0, output_index)`:

* sent arm: `sent_notes.to_address` = the actual recipient `t…` address (recorded faithfully for every `Recipient` variant by [`recipient_params`](https://github.com/zcash/librustzcash/blob/732654e15d69ee1cb9b282bc6cace1e6f73ff96a/zcash_client_sqlite/src/wallet.rs#L5303-L5362); the send side of a self-received transparent output is recorded as `Recipient::InternalTransparent` with the literal recipient address, see [`ll/wallet.rs`](https://github.com/zcash/librustzcash/blob/732654e15d69ee1cb9b282bc6cace1e6f73ff96a/zcash_client_backend/src/data_api/ll/wallet.rs#L1427-L1436));
* received arm: the UA, per defect 1.

`MAX` then picks the UA by byte order. Even once defect 1 is fixed, `MAX` remains a fragile way to choose between "the address the sender addressed" and "the address at which the wallet received"; preferring the sent arm's value when present would make the merged row's meaning independent of collation accidents.

## Provenance

Regression introduced by the `v_tx_outputs_return_addrs` migration (e02889e2f8, the implementation of #1951), first released in `zcash_client_sqlite 0.18.1` and carried forward through `account_delete_cascade` and `v_tx_outputs_key_scopes` to current `main` (and the `0.22.0-rc` line). Prior generations of the view selected the literal `t…` string (`utxos.address`) — or `NULL` in the interim schema — for the transparent received arm, so `sent_notes.to_address` survived the merge and self-sends displayed correctly.

## Impact

* Wallet-internal transparent sends display a unified address instead of the transparent recipient.
* All transparent outputs received at external-scope receivers report the enclosing UA in `to_address`, misrepresenting what was observably paid on chain.
* Downstream consumers cannot recover the transparent address: it never leaves SQL. (E.g. an SDK that classifies the recipient from the returned string will additionally misclassify these outputs as shielded/unified recipients.)

## Suggested fix

A new view migration:

1. In the received arm, select the pool-appropriate receiver: `CASE WHEN ro.pool = 0 THEN a.cached_transparent_receiver_address ELSE a.address END AS to_address` (or expose `transparent_received_outputs.address` through `v_received_outputs`).
2. In the merge, prefer the sent arm's `to_address` over the received arm's rather than relying on `MAX` — e.g. by tagging the arms and using an ordered aggregate, so that the recorded send recipient is authoritative when both exist.

---

nuttycom:
The received-output arm of `v_tx_outputs` resolved every receiving address through `addresses.address`, which for an external-scope entry holds the unified encoding; the transparent receiver lives only in `addresses.cached_transparent_receiver_address`. Every transparent output received by the wallet at such an address was therefore reported with the enclosing unified address as its `to_address`.

For a payment the wallet made to one of its own transparent addresses the effect was compounded: the transaction produces both a `sent_notes` row carrying the transparent address that was actually paid and a received-output row carrying the unified address, and the view merged the two with `MAX(to_address)` — under SQLite's BINARY collation `'u' > 't'`, so the unified address deterministically shadowed the recorded recipient.

The new `v_tx_outputs_transparent_addresses` migration recreates the view so that:
- the received-output arm reports the pool-appropriate receiver (`cached_transparent_receiver_address` for transparent outputs, `address` otherwise), and
- the sent/received merge prefers the recipient address recorded when the wallet created the output over the receiving address — via an `is_sent_row` discriminator and `COALESCE` — rather than choosing between the two encodings by byte order.

A regression test receives a UTXO at the account's own external-scope transparent receiver and asserts `to_address` both before and after the send side is recorded as `Recipient::InternalTransparent`; it fails against the previous view definition in both positions and passes with the migration applied.

Fixes #2845.

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -48,6 +48,13 @@ workspace.
   commitment tree position (chain order). Previously notes were drawn in
   scan-discovery order, which for a restored wallet prefers its most recently
   discovered — typically newest — notes.
+- The `to_address` column of the `v_tx_outputs` view now reports transparent
+  outputs received by the wallet at the transparent receiver address itself,
+  rather than at a unified address containing that receiver, and for outputs
+  the wallet created, the recipient address recorded at transaction
+  construction time now takes precedence over the receiving address.
+  Previously, a payment to one of the wallet's own transparent addresses was
+  reported with the receiving account's unified address as its `to_address`.
 
 ## [0.22.0-rc.5] - 2026-07-28
 
--- a/zcash_client_sqlite/src/wallet/db.rs
+++ b/zcash_client_sqlite/src/wallet/db.rs
@@ -1527,8 +1527,11 @@ GROUP BY notes.account_id, notes.transaction_id";
 ///   output of this view, one for each such account.
 /// - `to_account_uuid`: The UUID of the wallet account that received the output, if any; for
 ///   outgoing transaction outputs this will be `NULL`.
-/// - `address`: The address to which the output was sent; for received outputs, this is the
-///   address at which the output was received, or `NULL` for wallet-internal outputs.
+/// - `address`: The address to which the output was sent. For outputs created by the wallet,
+///   this is the recipient address recorded when the transaction was created. For other
+///   received outputs it is the address at which the output was received — for transparent
+///   outputs, the transparent receiver itself rather than a unified address containing it —
+///   or `NULL` for wallet-internal outputs.
 /// - `diversifier_index_be`: The big-endian representation of the diversifier index (or, for
 ///   transparent addresses, the BIP 44 change-level index of the derivation path) of the receiving
 ///   address. This will be `NULL` for outgoing transaction outputs.
@@ -1557,7 +1560,13 @@ WITH unioned AS (
            ro.output_index              AS output_index,
            from_account.uuid            AS from_account_uuid,
            to_account.uuid              AS to_account_uuid,
-           a.address                    AS to_address,
+           -- for a transparent output, the address at which it was received is
+           -- the transparent receiver itself, not a unified address containing it
+           CASE ro.pool
+                WHEN 0 THEN a.cached_transparent_receiver_address
+                ELSE a.address
+           END                          AS to_address,
+           0                            AS is_sent_row,
            a.diversifier_index_be       AS diversifier_index_be,
            ro.value                     AS value,
            ro.is_change                 AS is_change,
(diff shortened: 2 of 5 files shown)


nuttycom (commented):
self-utACK

nullcopy (approved):
utACK 7e6e1af7fc61c36a2550d46acda59d1e5f8dca15

PR #2862, opened 2026-07-30
Release zcash_client_sqlite version 0.22.0-rc.6

nuttycom:
Release round for the ZIP 318 timing revision and the pool-crossing wallet work merged since
`zcash_client_sqlite 0.22.0-rc.5`.

| Crate | Version |
| --- | --- |
| `zcash_protocol` | 0.10.2 → 0.10.3 |
| `zcash_client_backend` | 0.24.0-rc.5 → 0.24.0-rc.6 |
| `zcash_pool_migration` | 0.1.0-rc.4 → 0.1.0-rc.5 |
| `zcash_client_sqlite` | 0.22.0-rc.5 → 0.22.0-rc.6 |

## What this ships

- **Revised ZIP 318 migration timing** (zcash/zips#1343): `TRANSFER_DELAY_MEAN` is now 66 blocks
  (was 144), `ANCHOR_AGE_CAP` is 4 boundaries (was 16), and the preparation delay is specified
  rather than provisional (`PREP_DELAY_MEAN`/`PREP_DELAY_CAP`). #2855 also makes
  `SchedulingParams::new_with_default_distributions` scale every delay parameter by the interval
  ratio instead of deriving delays from fixed cap/mean ratios.
- **Migration-shaped note selection** (#2850, #2856): `propose_transfer` funds a canonical
  crossing from the single oldest covering Orchard note, via the new
  `SpendPolicy::with_note_selection` / `NoteSelection::PreferSingle`.
- **Transaction shape recorded as dummy-output counts** (#2851): a `ChangeStrategy` now records
  the exact per-pool dummy-output count it costed (`fees::DummyOutputCounts`), replacing the
  Ironwood-specific `TransactionBalance::with_ironwood_bundle_padding`. The counts serialize into
  the proposal wire format, so every construction path reproduces the shape the fee was charged
  against without re-running the fee model.
- **Anchor availability fixes** (#2849, #2852): `put_blocks` checkpoints every anchor-retention
  grid height in a scanned range, including boundaries whose blocks carry no note commitments, and
  `propose_transfer` abandons a canonical crossing when no anchor is computable at the bucketed
  boundary rather than proposing one whose build fails with `ProposalError::AnchorNotFound`.
- **Pool-crossing transaction history** (#2822): `v_transactions` gains a `pool_crossing_value`
  column classifying wallet-internal shielded pool crossings and reporting the amount that crossed.
- **`v_tx_outputs` transparent addresses** (#2857): transparent outputs are reported at the
  transparent receiver itself rather than at a unified address containing it.
- **Stuck-scan repair** (#2836): a wallet built from a revision that predates the
  `anchor_bucket_interval` column never acquired it, and the committed-grid query in `put_blocks`
  then failed at prepare time on *every* scan, so no block was written and every transaction stayed
  pending. A new migration installs the column where it is missing.

## Notes for review

- `zcash_protocol` is a **patch** bump. `DELAY_CAP_RATIO` is deprecated rather than removed
  (7bf6378), which keeps the release semver-compatible and confines this round to four crates —
  a breaking `zcash_protocol` release would have forced `zcash_address`, `zcash_primitives`,
  `zcash_transparent`, `zcash_keys`, `pczt`, `zip321`, and `zcash_proofs` into it as well.
  `cargo semver-checks` against 0.10.2 passes 196 checks with no violations.
- The changelog audit moved the #2836 fix entry out of the already-published `0.22.0-rc.5`
  section, where it had originally landed, into this release's `Fixed` section, and added the
  previously-unlogged `fees`/`proto` dummy-output API changes and the `pczt-tests` feature change.
- `cargo semver-checks` reports nothing for the three prerelease crates: with a prerelease
  baseline it assumes a major bump and skips all 253 lints, so their audit rests on the diff sweep.
- Publish bottom-up: `zcash_protocol`, then `zcash_client_backend`, then `zcash_pool_migration`,
  then `zcash_client_sqlite`.
- Tags are added after merge, one per crate on that crate's own release commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7332,7 +7332,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_pool_migration"
-version = "0.1.0-rc.4"
+version = "0.1.0-rc.5"
 dependencies = [
  "corez",
  "getset",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7186,7 +7186,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_client_sqlite"
-version = "0.22.0-rc.5"
+version = "0.22.0-rc.6"
 dependencies = [
  "ambassador",
  "assert_matches",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7428,7 +7428,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_protocol"
-version = "0.10.2"
+version = "0.10.3"
 dependencies = [
  "corez",
  "document-features",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7111,7 +7111,7 @@ dependencies = [
 
 [[package]]
 name = "zcash_client_backend"
-version = "0.24.0-rc.5"
+version = "0.24.0-rc.6"
 dependencies = [
  "ambassador",
  "arti-client",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ categories = ["cryptography::cryptocurrencies"]
 # Intra-workspace dependencies
 equihash = { version = "0.3", path = "components/equihash", default-features = false }
 zcash_address = { version = "0.13.0", path = "components/zcash_address", default-features = false }
-zcash_client_backend = { version = "0.24.0-rc.5", path = "zcash_client_backend" }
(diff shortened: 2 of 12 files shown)


nullcopy (approved):
utACK f370a1f250b726a5d4df5d820cd0768bbe58db99

PR #2864, opened 2026-07-30
zcash_client_sqlite: emit the documented diversifier_index_be column from v_tx_outputs

### LINKED ISSUE #1073: Expose `diversifier` column from `sapling_received_notes` in the `v_tx_outputs` view
The diversifier should be exposed in the 'public' view so that wallets can recover the original diversifier index used to generate the address that was used to send money to this wallet.

### LINKED ISSUE #2863: zcash_client_sqlite: `v_tx_outputs` documents a `diversifier_index_be` column that the view does not emit
## Summary

The documentation of `v_tx_outputs` describes a `diversifier_index_be` column ([db.rs#L1535-L1537](https://github.com/zcash/librustzcash/blob/6c07e5f3297febf469e5cc8d0b91321e0767cdd7/zcash_client_sqlite/src/wallet/db.rs#L1535-L1537)):

> - `diversifier_index_be`: The big-endian representation of the diversifier index (or, for transparent addresses, the BIP 44 change-level index of the derivation path) of the receiving address. This will be `NULL` for outgoing transaction outputs.

but the view never emits it. Both arms of the `unioned` CTE select the column ([db.rs#L1570](https://github.com/zcash/librustzcash/blob/6c07e5f3297febf469e5cc8d0b91321e0767cdd7/zcash_client_sqlite/src/wallet/db.rs#L1570) and the sent arm's `NULL`), and the outer merge projection ([db.rs#L1608-L1630](https://github.com/zcash/librustzcash/blob/6c07e5f3297febf469e5cc8d0b91321e0767cdd7/zcash_client_sqlite/src/wallet/db.rs#L1608-L1630)) then drops it — its column list goes directly from `to_address` to `value`. `SELECT diversifier_index_be FROM v_tx_outputs` fails with "no such column", and the CTE selection is dead code.

## Provenance

The column entered the CTE with the `v_tx_outputs_return_addrs` migration (e02889e2f8), the implementation of #1951 — whose title asks to expose *the address and diversifier index* at which each output is received. The address half shipped (and was recently corrected in #2845 / #2857); the diversifier-index half made it into the CTE and the column documentation, but never into the view's output, and has been carried forward verbatim ever since (first affected release: `zcash_client_sqlite 0.18.1`).

Because every query naming the column has always failed immediately, no consumer can currently depend on it; projecting it now is purely additive.

## Suggested fix

Recreate the view with the column projected through the outer merge, e.g. `MAX(diversifier_index_be) AS diversifier_index_be` following `to_address`. The merge semantics work out: for an output the wallet both sent and received, the received arm's non-`NULL` index wins over the sent arm's `NULL`, and rows for outputs sent to external recipients remain `NULL`. The documented `NULL` condition should be restated accordingly ("not received at one of the wallet's diversified addresses" rather than "outgoing"), since a wallet-internal transparent send now carries the receiving address's index.

---

nuttycom:
The `diversifier_index_be` column has been selected into the `unioned` CTE of `v_tx_outputs` — and documented as part of the view's output — since the receiving-address columns were added in the implementation of #1951, but the outer merge projection never carried it through: `SELECT diversifier_index_be FROM v_tx_outputs` has always failed with "no such column", and the CTE selection was dead code. Since every query naming the column failed immediately, no consumer can depend on its absence; projecting it is purely additive.

The new `v_tx_outputs_diversifier_index` migration (depending on `v_tx_outputs_transparent_addresses` from #2857) recreates the view with `MAX(diversifier_index_be)` projected through the merge. For an output the wallet both sent and received, the received arm's non-`NULL` index survives; rows for outputs sent to external recipients remain `NULL`. The column documentation's `NULL` condition is restated accordingly ("not received at one of the wallet's diversified addresses" rather than "outgoing"), since a wallet-internal send now carries the receiving address's index.

A regression test receives a UTXO at the account's own external-scope transparent receiver and asserts that the view's `diversifier_index_be` matches the `addresses` table's value for the receiving address, both before and after the send side of the output is recorded; it fails with "no such column" against the previous view definition.

Fixes #2863.

Closes #1073, which asked for this column so that a wallet could recover the diversifier index used to generate a receiving address. That issue's alternative suggestion — populating `to_address` for inbound notes rather than exposing the index — is already satisfied by the received arm's address projection, so this change answers it on both readings.

### DIFF
--- a/zcash_client_sqlite/CHANGELOG.md
+++ b/zcash_client_sqlite/CHANGELOG.md
@@ -10,6 +10,14 @@ workspace.
 
 ## [Unreleased]
 
+### Added
+- The `v_tx_outputs` view now emits the `diversifier_index_be` column that its
+  documentation has described since the receiving-address columns were added:
+  the big-endian diversifier index of the receiving address, `NULL` for outputs
+  not received at one of the wallet's diversified addresses. The column was
+  previously computed internally but omitted from the view's output, so any
+  query naming it failed with "no such column".
+
 ## [0.22.0-rc.6] - 2026-07-29
 
 ### Added
--- a/zcash_client_sqlite/src/wallet/db.rs
+++ b/zcash_client_sqlite/src/wallet/db.rs
@@ -1534,7 +1534,8 @@ GROUP BY notes.account_id, notes.transaction_id";
 ///   or `NULL` for wallet-internal outputs.
 /// - `diversifier_index_be`: The big-endian representation of the diversifier index (or, for
 ///   transparent addresses, the BIP 44 change-level index of the derivation path) of the receiving
-///   address. This will be `NULL` for outgoing transaction outputs.
+///   address. This will be `NULL` for outputs that were not received at one of the wallet's
+///   diversified addresses (in particular, for outputs sent to external recipients).
 /// - `value`: The value of the output, in zatoshis.
 /// - `is_change`: `0` for outgoing outputs and outputs received at external-facing addresses, `1`
 ///   for outputs received at wallet-internal addresses. This represents a best-effort judgement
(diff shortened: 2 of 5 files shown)


str4d (approved):
utACK 62269968f3380fd416ab233c8f959b669c2f3868

Issue #2700, opened 2026-07-21
Pool migration: anchor bucketing needs deeper checkpoint retention than PRUNING_DEPTH (AnchorNotFound on every transfer)


dannywillems:
## Summary

The pool-migration scheduler (`zcash_pool_migration_backend::scheduling`, merged in #2663) draws each transfer's Orchard proving anchor from coarse BOUNDARY blocks that are **always at least 144 blocks, and up to ~2304 blocks, below the chain tip**. The wallet's checkpoint retention (`PRUNING_DEPTH = 100`) is shallower than that. Since the migration defers anchors and witnesses to proving time (ZIP 374) and the consumer must install the anchor at the drawn boundary, a wallet on the default retention **cannot produce a witness/root at any boundary the scheduler picks**, giving `AnchorNotFound` at proving time on essentially every transfer.

This is a cross-layer gap, not a bump that lives elsewhere: nothing raises `PRUNING_DEPTH`, there is no migration-time retention mechanism, and `scheduling` does not document this as a consumer responsibility. It is an engine/scheduling issue (from #2663), **not** introduced by the `zcash_pool_migration_sqlite` store PR (#2669), which only persists what the scheduler decides. Reported by the mobile team while wiring the SDK integration.

## Verified facts

- `BOUNDARY_MODULUS = MEAN_DELAY = 144` blocks (`scheduling.rs:70,94`).
- `draw_anchor_boundary`: `candidate = most_recent_boundary - age * 144`, `age >= 1` (age 0 excluded), so the anchor is always `>= 144` blocks below the tip (`scheduling.rs:378-415`).
- `ANCHOR_AGE_CAP = 16` -> oldest drawable anchor up to `16 * 144 = 2304` blocks below `most_recent_boundary`, ~2 days (`scheduling.rs:100`).
- `PRUNING_DEPTH = 100` (`zcash_client_sqlite/src/lib.rs:169`), not bumped anywhere.
- Migration builds UNPROVEN PCZTs and defers anchor+witness to proving time (ZIP 374); the wallet adapter never resolves a witness, the consumer installs them via the PCZT `Updater` (`wallet.rs:13-14,97`). `anchor_boundary` is only drawn/stored (`engine.rs:1098`, doc at `engine.rs:201`), never consumed in the build path.

## Arithmetic

- Newest possible anchor (`age=1`, tip on a boundary): `tip - 144`. Already `144 > 100`.
- Oldest (`age=16`): `~tip - 2447`.
[...]

PR #2867, opened 2026-07-30
zcash_pool_migration: Broadcast before proving in `MigrationState::next_step`

### LINKED ISSUE #2866: zcash_pool_migration: `next_step` surfaces proving ahead of due broadcasts
`MigrationState::next_step` returns `AdvanceStep::Prove` in preference to `AdvanceStep::Broadcast`. That ordering was designed around the incorrect presumption that the note commitment tree's default pruning window bounded the period during which a transfer could be proved, making proving time-critical: it is not — the wallet durably retains the boundary checkpoints that committed transfers anchor to (#2700), so a transfer remains provable from the moment its drawn boundary settles until it is broadcast.

The ordering that actually serves wallets is the reverse. ZIP 318 requires that a background wake window be used either to sync or to broadcast, never both; broadcasting a stored proven transaction requires no sync, while proving is inherently sync-bound. A wallet that wakes to find a proven transaction due should therefore be told to submit it immediately — without any sync operations being initiated in that session — and be offered proving work only once no broadcast is due.

The rustdoc stating the pruning rationale (`AdvanceStep::Prove`, `NextAction::Prove`, `Blocker::AnchorBoundary`, `prove_ready`, `next_provable`, and the `state` module header) needs the same correction, and `next_step`'s contract — the drive loop, the ordering, and the session-separation implications for wallet construction — should be fully documented for consumers.

nuttycom:
Closes #2866.

`MigrationState::next_step` previously surfaced proving work ahead of due broadcasts, a design premised on the incorrect idea that the note commitment tree's default pruning window bounded the period during which a transfer could be proved. Boundary checkpoints are durably retained until a migration's transfers are proven (#2700), so proving has no deadline pressure; the constraint that matters is the ZIP 318 sync/broadcast session separation, under which a wallet that wakes with a proven transaction due should submit it immediately, without initiating sync operations in that session.

This PR:

- Reorders `next_step` to surface a due `AdvanceStep::Broadcast` ahead of `AdvanceStep::Prove`, with a CHANGELOG entry for the behavioral change.
- Rewrites the rustdoc that stated the pruning rationale (`AdvanceStep`, `NextAction`, `Blocker::AnchorBoundary`, `prove_ready`, `next_provable`, the `state` module header, and `prove_transfer`'s caller-timing note).
- Fully documents `next_step`: the drive-loop contract mapping each step to the operation that discharges it, the ordering rationale, and the wallet-construction implication that enforcing the session separation is the consumer's runtime policy (`next_step` itself is memoryless about sessions).
- In a second commit, fixes stale rustdoc predating the one-signing-phase design: `Blocker::Dependencies`, the `TransactionStatus` overview, and the preparation-planner module doc still described preparation layers as signed per-layer and serialized on anchor-bucket boundaries.

An audit of the crate for other design elements premised on the pruning misunderstanding found none in behavior: the sync/proving wake-up schedule bounds each proving window by the broadcast height (sync/broadcast decoupling), not by pruning, and the anchor-interval-mismatch machinery is already premised on durable retention.

Part of #2630.

### DIFF
--- a/zcash_pool_migration/src/state.rs
+++ b/zcash_pool_migration/src/state.rs
@@ -12,9 +12,12 @@
 //!
 //! Every transaction is built and pre-signed when the migration is committed (one signing phase;
 //! anchors and witnesses are deferred to proving time per ZIP 374), so the state machine's only
-//! job is to ORDER the broadcasts: a transaction becomes broadcastable once its dependencies (the
-//! preparation layers that mint its inputs) have mined and its scheduled height has arrived, and
-//! the consumer proves it — installing its anchor and witnesses — just before broadcasting.
+//! job is to ORDER the remaining work: a transaction is proved — installing its anchor and
+//! witnesses — once its anchor is resolvable (for a transfer, once its drawn boundary settles),
+//! and becomes broadcastable once it is proved, its dependencies (the preparation layers that
+//! mint its inputs) have mined, and its scheduled height has arrived. See
+//! [`MigrationState::next_step`] for how the two are ordered and the sync/broadcast session
+//! separation that ordering is designed around.
 
 use alloc::vec::Vec;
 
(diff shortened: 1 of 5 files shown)


dannywillems (dismissed):
utACK

nuttycom:
Converted to draft to add a distinction between prep and transfer transactions to the prove result - when we prove a prep transaction, we can immediately broadcast.

zcash_pool_migration/src/state.rs
```diff
@@ -267,17 +274,20 @@ impl MigrationState {
...
     /// sync wake-up well before its scheduled broadcast height, keeping the sync work and the
     /// broadcast in separate waking sessions.
     pub fn next_provable(&self, target_height: BlockHeight) -> Option<MigrationTransferId> {
-        self.transactions
-            .iter()
-            .find(|t| {
-                matches!(t.state, MigrationTxState::Signed) && self.prove_ready(t, target_height)
-            })
-            .map(|t| t.id)
+        self.next_provable_tx(target_height).map(|t| t.id)
```
dannywillems, line 290: Why not inlining?

dannywillems (approved):
utACK, modulo comment.

zcash_pool_migration/src/state.rs
dannywillems: Okay, I see. Closing.

